PR #3041, opened 2017-02-14
Unpin CI agent

gpeng:
Whitehall was pinned to ci-agent-3 but the performance fixes have been
deployed to all agents now so we can remove this restriction.


andrewgarner:
Great work! 👍

PR #3043, opened 2017-02-15
Add sync check failure for 4xx and 5xx errors

bevanloon:
Sync checks are reporting no failures when a base path returns a 4xx or 5xx against the content stores.

Note there was already a test named `test_returns_empty_array_if_response_not_200`. I've renamed this test to reflect the 301 nature that it actually tests - I'm assuming there was some reason we wanted not to report failures on "moved permanently".

I think it's sensible to flag up where we don't find items in the content store so we don't get a bunch of false positives in the sync checks, but given the previous test noted above, happy to have a discussion around it.

### DIFF
--- a/lib/sync_checker/checks/http_status_check.rb
+++ b/lib/sync_checker/checks/http_status_check.rb
@@ -1,0 +1,15 @@
+module SyncChecker
+  module Checks
+    HttpStatusCheck = Struct.new(:disallowed_response_codes) do
+      def call(response)
+        failures = []
+
+        if Array(disallowed_response_codes).include?(response.response_code)
+          failures << "http response code #{response.response_code} returned for #{response.request.base_url}"
+        end
+
+        failures
+      end
+    end
+  end
+end
--- a/lib/sync_checker/formats/edition_base.rb
+++ b/lib/sync_checker/formats/edition_base.rb
@@ -1,6 +1,8 @@
 module SyncChecker
   module Formats
     class EditionBase
+      HTTP_400_AND_500_ERRORS = 400..599
+
       def self.scope
         klass = self.name.demodulize.sub(/Check$/, '')
         Document.where(id: klass.constantize.all.pluck(:document_id).uniq)
@@ -70,7 +72,8 @@ module SyncChecker
               expected_details_hash(edition_expected_in_draft, locale)
             end
           ),
-          Checks::TranslationsCheck.new(edition_expected_in_draft.available_locales)
+          Checks::TranslationsCheck.new(edition_expected_in_draft.available_locales),
+          Checks::HttpStatusCheck.new(HTTP_400_AND_500_ERRORS)
         ]
       end
 
@@ -115,7 +118,8 @@ module SyncChecker
           Checks::UnpublishedCheck.new(document),
           Checks::TranslationsCheck.new(edition_expected_in_live.available_locales),
           Checks::TopicsCheck.new(edition_expected_in_live,
-                                  topic_blacklist: DraftTopicContentIds.fetch)
+                                  topic_blacklist: DraftTopicContentIds.fetch),
+          Checks::HttpStatusCheck.new(HTTP_400_AND_500_ERRORS)
         ]
       end
 
--- a/test/unit/sync_checker/checks/http_status_check_test.rb
+++ b/test/unit/sync_checker/checks/http_status_check_test.rb
@@ -1,0 +1,43 @@
+require 'minitest/autorun'
+require 'mocha/setup'
+require 'active_support/json'
+require_relative '../../../../lib/sync_checker/checks/http_status_check'
+
+module SyncChecker::Checks
+  class HttpStatusCheckTest < Minitest::Test
+    def response(response_code)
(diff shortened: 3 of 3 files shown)


bevanloon:
👍  @andrewgarner @gpeng 
I've pulled this together in 983c68b55a0783e63c3ea83ff43f5230dfd25259. 

What are everyone's thoughts on adding this to EditionBase as per 77a6aecde641d5724439b0c5dc42b53e5a54b10a? I'm just as happy to pull it into WLNA only since that's where I was hitting the problem if it seems too intrusive to include it in the base class.

lib/sync_checker/formats/edition_base.rb
```diff
@@ -66,7 +66,8 @@ def checks_for_draft(locale)
               expected_details_hash(edition_expected_in_draft)
             end
           ),
-          Checks::TranslationsCheck.new(edition_expected_in_draft.available_locales)
+          Checks::TranslationsCheck.new(edition_expected_in_draft.available_locales),
+          Checks::HttpStatusCheck.new(edition_expected_in_draft)
```
andrewgarner: It's not clear why we'd initialise this with the edition as we don't need to know anything about the edition, we just need a response object that is provided when the check is called.

bevanloon:
My thoughts behind the object's name was that we could, should we choose, use it to test for any number of http status codes.....rather than only 400 and 500. I tried to keep something more generic since the existing code seems like that to me (TopLevelChecks, LinksCheck etc). 

If we go down the road of naming it something like ResponseOk check then we are implying that we specifically only accept a 200 type response.....which is not entirely correct given the existing tests that imply we are happy with no errors on a 301 (https://github.com/alphagov/whitehall/blob/master/test/unit/sync_checker/checks/top_level_check_test.rb#L41)

I thought about very specifically named 4xxCheck and 5xxCheck, but then we risk an explosion of modules and files.

Any suggestions on a better name?

lib/sync_checker/formats/edition_base.rb
bevanloon: Perhaps the route of LinksCheck is the way forward - we could pass in a range of http codes that we want to fail on - is that what you meant by https://github.com/alphagov/whitehall/pull/3043#commitcomment-20910818?

andrewgarner: Yep. All of the other checks follow the convention of specifying what is expected. We seem to be unclear about which codes are expected, although we are clear on which codes should not be expected so maybe we could provide a range of disallowed response codes?

```rb
module SyncChecker
  module Checks
    HttpStatusCheck = Struct.new(:disallowed_response_codes) do
      def call(response)
        failures = []

        if (disallowed_response_codes).include?(response.response_code)
          failures << "http response code #{response.response_code} returned for #{response.request.base_url}"
        end

        failures
      end
    end
  end
end

HttpStatusCheck.new(400..599).call(response)
```

lib/sync_checker/checks/http_status_check.rb
```diff
@@ -0,0 +1,21 @@
+module SyncChecker
+  module Checks
+    HttpStatusCheck = Struct.new(:disallowed_response_codes) do
+      def call(response)
+        failures = []
+
+        if (disallowed_response_codes_array).include?(response.response_code)
```
andrewgarner: You don't need the parentheses here.

lib/sync_checker/formats/edition_base.rb
```diff
@@ -190,6 +192,10 @@ def latest_edition_is_rejected?(document)
           document.pre_publication_edition &&
           document.pre_publication_edition.rejected?
       end
+
+      def disallowed_http_codes
+        (400..599)
```
andrewgarner: Did you consider expressing this as a constant?

lib/sync_checker/checks/http_status_check.rb
andrewgarner: You could probably just inline the coercion as it is pretty expressive?

```rb
if Array(disallowed_response_codes).include?(response.response_code)
# ...
end
```

lib/sync_checker/formats/edition_base.rb
bevanloon: Yep...Personal preference here - it looked odd to me to have a constant range. 

Also, perhaps over thinking the future of the code, but if we wanted to construct custom ranges / arrays of status codes and treat them differently, the method expresses that opportunity, whereas a constant feels more final and as if it closes off code paths. I know it's the same effort to refactor a constant. Like I say, just a feeling. Nothing particularly logical.

I happy to change it though - thoughts?

andrewgarner: I don't think we should worry about it looking odd, at the moment it is something that is constant to our check so I think we should express it using that syntax.

andrewgarner: http://www.virtuouscode.com/2011/08/18/do-we-need-constants/

bevanloon: nice - TIL `module_function`

PR #3044, opened 2017-02-16
Present speeches to publishing api

steventux:
https://trello.com/c/J9H6FRK1/611-speech-migration-2-implement-publishing-of-format-to-publishing-api

Presenters for Speech format.

### DIFF
--- a/app/presenters/publishing_api/speech_presenter.rb
+++ b/app/presenters/publishing_api/speech_presenter.rb
@@ -1,0 +1,82 @@
+module PublishingApi
+  class SpeechPresenter
+    include UpdateTypeHelper
+
+    attr_accessor :item
+    attr_accessor :update_type
+
+    def initialize(item, update_type: nil)
+      self.item = item
+      self.update_type = update_type || default_update_type(item)
+    end
+
+    def content_id
+      item.content_id
+    end
+
+    def content
+      content = BaseItemPresenter.new(item).base_attributes
+      content.merge!(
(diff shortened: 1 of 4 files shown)


app/presenters/publishing_api/speech_presenter.rb
```diff
@@ -0,0 +1,82 @@
...
+    def image_payload
+      {
+        image: {
+          alt_text: speaker.name,
+          url: speaker.image.url,
+        }
+      }
+    end
+
+    def links_for_policies
```
gpeng: I know we had a conversation about this but it looks like everything else is using `related_policies` for this, passed as a key to the `LinksPresenter`

steventux: There's actually quite a bit of variation in the content schemas, speech schema and others use the key `policies` https://github.com/alphagov/govuk-content-schemas/blob/master/formats/speech/publisher/links.json#L11 see https://trello.com/c/rXXuFPfS/848-rename-publications-related-policies-links-key for the detail on the differences. Happy to change it but conversations I had with Finding Things team yesterday pointed to retaining the `policies` key and removing the `related_policies` one.

steventux:
Thanks for the review @gpeng updated

app/presenters/publishing_api/speech_presenter.rb
gpeng: Ok. Fair enough.

PR #3045, opened 2017-02-16
Fix attachment locales

gpeng:
We are correctly filtering `HtmlAttachment`s based on locale on `Publication` documents but this filtering needs to also be applied to `FileAttachment`s.


gpeng:
I think it failed a test originally if we included `FileAttachment` but that was because of the incorrect assumption I'd made about `nil` locale attachments.

PR #3046, opened 2017-02-16
Add speeches sync check

steventux:

I've branched this from `present-speeches-to-publishing-api` in order to run the check so the first 2 commits can be ignored.
Relies on https://github.com/alphagov/whitehall/pull/3044


gpeng:
Rebased from master to bring in changes merged from #3044

PR #3048, opened 2017-02-17
Fix attachments and world location news articles

mgrassotti:
Trello: https://trello.com/c/BBPaIzcQ/588-check-htmlattachment-routes-3

Some attachment translations were not correctly migrated. We are
deleting them from publishing API and republishing them in Whitehall.

Paired with @bevanloon, @gpeng


db/data_migration/20170217094518_republish_problematic_translations.rb
```diff
@@ -0,0 +1,662 @@
...
+  /government/world-location-news/274722.zh
+  /government/world-location-news/282389.zh
+  /government/world-location-news/282404.zh
+  /government/world-location-news/286926.zh
+  /government/world-location-news/291036.zh
+)
+
+base_paths = base_paths_200_302 + base_paths_bad_wlna
+
+Edition.where(base_path: base_paths).joins(:document).distinct.pluck(:id).each do |id|
```
gpeng: This doesn't work.  Firstly, `Edition` doesn't have a `base_path` field in Whitehall either so you can't query on it.

Secondly, if the query did work we need `document_id` to republish with and this is returning `Edition#id`. I also don't think the join/distinct bit is necessary as it doesn't help with the `document_id`. Once you get the right query to return an edition AR relation you can just `pluck(:document_id).uniq`.

mgrassotti: @gpeng yes sorry I already noticed that and modify it to use Document ids instead. Thanks

PR #3050, opened 2017-02-17
Find the missing children

gpeng:
Changes introduced in #3045 have caused some (err... loads) of unexpected failures in the `Publication` sync checks. We need to filter the `FileAttachment`s by locale when presenting but we don't want to include them in the sync check of the `children` links as publishing-API is unaware of them and won't create a link.

So... the first commit excludes them from the check.

The second commit filters the documents passed to the details check for `Publication`s by locale. This worryingly doesn't seem to have been causing failures but they are [filtered by locale in the presenter](https://github.com/alphagov/whitehall/blob/master/app/presenters/publishing_api/publication_presenter.rb#L76-L91) so we should filter them in the check too.

### DIFF
--- a/lib/sync_checker/formats/consultation_check.rb
+++ b/lib/sync_checker/formats/consultation_check.rb
@@ -1,6 +1,6 @@
 module SyncChecker::Formats
   class ConsultationCheck < EditionBase
-    def expected_details_hash(consultation)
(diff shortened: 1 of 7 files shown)


fofr:
db837ec41085a30501c06b7d49149298

PR #3052, opened 2017-02-17
Data migration to republish some Publications

gpeng:
As part of the migration of the `Publication` format, we need to republish them to update the rendering app. We are doing this in stages. This is the second batch which will republish ~37000 documents.

### DIFF
--- a/db/data_migration/20170217153845_republish_publications_migration_two.rb
+++ b/db/data_migration/20170217153845_republish_publications_migration_two.rb
@@ -1,0 +1,18 @@
+document_scope = Document.where(
+  id: Publication.where(state: %w(draft published withdrawn)).pluck(:document_id)
+)
+
+lowest_id_for_this_republish = 260000
+highest_id_for_this_republish = 330000
+document_scope = document_scope.where(
+  id:  lowest_id_for_this_republish..highest_id_for_this_republish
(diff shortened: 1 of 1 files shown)


db/data_migration/20170217153845_republish_publications_migration_two.rb
```diff
@@ -0,0 +1,18 @@
+document_scope = Document.where(
+  id: Publication.where(state: %w(draft published withdrawn)).pluck(:document_id)
+)
+
+lowest_id_for_this_republish = 260000
+highest_id_for_this_republish = 330000
+document_scope = document_scope.where(
```
bevanloon: is this just here for clarity? i.e. couldn't we get the result set we wanted with 
`Publication.where(state: %w(draft published withdrawn), document_id: 260000..330000)`?

gpeng: It was to avoid snarky remarks about 'magic numbers' from some of our more hilarious colleagues 😆

bevanloon: 👍 carry on then 😆

dougdroper: 👂

PR #3019, opened 2017-02-07
Add topics section to document review page

MatMoore:
# Background
We're going to be showing a tagging interface for the new education taxonomy to publishers in selected organisations, ahead of deploying a new navigation beta.

Tagging will eventually be mandatory to publish, but it is currently optional.

This functionality is behind a feature flag as the tagging interface itself is going to be iterated.

# This change
When reviewing a document, you should be able to see all the topics the document is currently tagged to. These get rendered breadcrumb style so publishers can understand the browse paths that will lead to the document.

## Screenshots
transport to education and training for people aged 16 to 18 publication admin gov uk
transport to education and training for people aged 16 to 18 publication admin gov uk_2

Trello: https://trello.com/c/sKsbTkKk/464-add-topics-section-to-whitehall-document-overview-page
Prototype: https://govuk-tagging.herokuapp.com/publication-review?
Paired with @klssmith

PR #3053, opened 2017-02-17
Add live breadcrumbs preview

Davidslv:
The goal of this commit is to show the uses how their choice of the taxons affects
the document in the taxon and also in the breadcrumbs.

Trello: https://trello.com/c/REchyOAj/463-add-live-breadcrumb-preview-to-whitehall-tagging-interface

Paired with @MatMoore 

### Screenshot

screen shot 2017-02-17 at 15 52 57

### DIFF
--- a/app/assets/javascripts/admin/modules/breadcrumb_preview.js
+++ b/app/assets/javascripts/admin/modules/breadcrumb_preview.js
@@ -1,0 +1,67 @@
+(function(Modules) {
+  "use strict";
+
+  Modules.BreadcrumbPreview = function() {
+    var that = this;
+
+    that.fetchCheckedCheckboxes = function() {
+      return $('.topic-tree :checked');
+    };
+
+    that.buildBreadcrumbsStructure = function(checkboxes) {
+      var structure = $.map(checkboxes, function(checkbox) {
+        var $checkbox = $(checkbox);
+        var ancestors = $checkbox.data('ancestors').split('|');
+
+        return [ancestors];
+      });
+
+      return structure;
+    };
+
+    that.filterBreadcrumbs = function(breadcrumbs) {
+      var longestFirst = breadcrumbs.sort(function (a, b) {
+        return b.length - a.length;
+      });
+
+      var breadcrumbsToDisplay = [];
+
+      $.each(longestFirst, function(index, breadcrumb) {
+        var visited = false;
+
+        $.each(breadcrumbsToDisplay, function(index, visitedBreadcrumb) {
+          var breadcrumbString = JSON.stringify(breadcrumb);
+          var visitedBreadcrumbString = JSON.stringify(visitedBreadcrumb.slice(0, breadcrumb.length));
+
+          if (breadcrumbString === visitedBreadcrumbString) {
+            visited = true;
+          }
+        });
+
+        if (!visited) {
+          breadcrumbsToDisplay.push(breadcrumb);
+        }
+      });
+
+      return breadcrumbsToDisplay;
+    }
+
+    that.start = function(element) {
+      var $topicTree = $('.topic-tree');
+      $(element).removeClass('hidden');
+
+      $topicTree.on('change', function() {
+        var checkboxes = that.fetchCheckedCheckboxes();
+        var breadcrumbsArray = that.buildBreadcrumbsStructure(checkboxes);
+        var breadcrumbsToDisplay = that.filterBreadcrumbs(breadcrumbsArray);
+
+        $(element).mustache(
+          'admin/edition_tags/_breadcrumb_list',
+          { breadcrumbs: breadcrumbsToDisplay }
+        );
+      });
+
+      $topicTree.trigger('change');
+    }
+  };
+})(window.GOVUKAdmin.Modules);
--- a/app/assets/javascripts/admin/modules/breadcrumb_preview.js
+++ b/app/assets/javascripts/admin/modules/breadcrumb_preview.js
@@ -2,13 +2,13 @@
   "use strict";
 
   Modules.BreadcrumbPreview = function() {
-    var that = this;
+    var preview = this;
 
-    that.fetchCheckedCheckboxes = function() {
+    preview.fetchCheckedCheckboxes = function() {
       return $('.topic-tree :checked');
     };
 
-    that.buildBreadcrumbsStructure = function(checkboxes) {
+    preview.buildBreadcrumbsStructure = function(checkboxes) {
       var structure = $.map(checkboxes, function(checkbox) {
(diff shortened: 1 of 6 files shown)


lib/taxonomy.rb
```diff
@@ -42,6 +42,24 @@ def descendants
       tree.tap(&:shift)
     end
 
+    # Get ancestors of a taxon
+    #
+    # @return [Array] all taxons in the path from the root of the taxonomy to the parent taxon
+    def ancestors
```
Davidslv: This functionality will be soon removed from here and we will use the gem.
That will be part of code refactor. 

https://github.com/alphagov/govuk_taxonomy_helpers/blob/master/lib/govuk_taxonomy_helpers/linked_content_item.rb#L56-L69

app/assets/javascripts/admin/modules/breadcrumb_preview.js
```diff
@@ -0,0 +1,66 @@
...
+        return b.length - a.length;
+      });
+
+      var seenAlready = [];
+
+      $.each(longestFirst, function(index, breadcrumb) {
+        var seen = false;
+
+        $.each(seenAlready, function(index, seenBreadcrumb) {
+          var breadcrumbString = JSON.stringify(breadcrumb);
```
Davidslv: You can't compare two arrays in Javascript, using `JSON.stringify` seemed to be the only reasonable solution, fun facts on the screenshot underneath
screen shot 2017-02-17 at 15 06 14

MatMoore: 😱

app/views/admin/edition_tags/edit.html.erb
```diff
@@ -1,7 +1,7 @@
 <% page_title "Tagging: " + @edition.title %>
 
 <div class="row">
-  <div class="col-md-8">
+  <div class="col-md-12">
```
nickcolley: You may be able to remove .row and .col-*-12 since it's full width.

app/assets/javascripts/admin/modules/breadcrumb_preview.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+      });
+
+      return seenAlready;
+    }
+
+    that.start = function(element) {
+      var $topicTree = $('.topic-tree');
+
+      $topicTree.on('change', function () {
```
nickcolley: Since this is not [debounced](https://remysharp.com/2010/07/21/throttling-function-calls), I'd worry there are performance problems with this approach.

Have you profiled this on a slower computer? You could try CPU Throttling to x10+ (https://umaar.com/dev-tips/88-cpu-throttling/)

app/views/admin/edition_tags/edit.html.erb
```diff
@@ -33,6 +33,9 @@
         </div>
       </div>
 
+      <div class="content content-bordered" data-module="breadcrumb-preview">
```
nickcolley: @fofr do we allow JavaScript only interface in the admin interfaces?

app/assets/javascripts/admin/modules/breadcrumb_preview.js
nickcolley: Might be worth making a helper method (or importing one from [lodash](https://lodash.com/docs/4.17.4#isEqual)) so from the code it's obvious what you're trying to do.

Either that or could add a comment.

app/views/admin/edition_tags/edit.html.erb
MatMoore: The intention is for this widget to be javascript only, but the form as a whole should work without javascript.

If you don't have javascript, you will see the same browse paths after saving on the document overview page. This widget displays the same information but you get feedback before you click save.

app/assets/javascripts/admin/modules/breadcrumb_preview.js
Davidslv: screen shot 2017-02-20 at 11 27 18


js-interaction

MatMoore: We tested with 10x and it seems OK...

Although the page itself is slow to render, I didn't notice a delay in the widget itself, and it doesn't seem like there's jankiness when you interact with the checkboxes. Is there anything else we should check?

app/assets/javascripts/admin/modules/breadcrumb_preview.js
```diff
@@ -0,0 +1,67 @@
...
+      });
+
+      return breadcrumbsToDisplay;
+    }
+
+    that.start = function(element) {
+      var $topicTree = $('.topic-tree');
+      $(element).removeClass('hidden');
+
+      $topicTree.on('change', function() {
```
fofr: Could this be extracted into a named function so it's clear what happens when a topicTree changes?
eg `renderUpdatedBreadcrumbs`?

test/javascripts/unit/admin/modules/breadcrumb_preview_test.js
```diff
@@ -0,0 +1,23 @@
+module("BreadcrumbPreview", function(){
```
fofr: These tests seem minimal. Is there missing coverage?

Davidslv: The intent of this test is indeed to be minimal. We believe that testing the other methods wasn't beneficial as both would depend on building a fixture for them to be tested. I'm not keen on fixture testing as the structure of the DOM can change. It's likely that the next developer touching this code doesn't know about it. The fixture can still be a passing fixture even thought is not representing the real thing.

app/assets/javascripts/admin/modules/breadcrumb_preview.js
nickcolley: Sounds good to me 👍

MatMoore:
I've just noticed we are not displaying a message when there are no tags selected, so this shouldn't be merged yet.

MatMoore:
Have added another commit that shows a message when there are no tags.

@joelanman Still needs to look at this - for now I've just made it look the same as on the document overview page for education documents (see https://github.com/alphagov/whitehall/pull/3019).

no-tags

MatMoore:
@nickcolley @fofr Have made some more changes to this. Are you happy for this to merged as it is now?

I've reviewed the overall workflow with Joe and I'll be making some small follow up changes soon, but this doesn't affect any of the functionality and I can raise this in a separate PR if it's simpler.

tijmenb:
Merging this to get it out to integration. If there are any further issues we can fix them in post.

PR #3059, opened 2017-02-21
Send speech body as html

steventux:

Another small change to the presenter to present body content similar to other formats.

### DIFF
--- a/test/unit/presenters/publishing_api/speech_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/speech_presenter_test.rb
@@ -43,8 +43,17 @@ class PublishingApi::SpeechPresenterTest < ActiveSupport::TestCase
       create(:speech,
              title: "Speech title",
              summary: "The description",
-             body: "Some content",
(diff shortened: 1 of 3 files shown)


test/unit/presenters/publishing_api/speech_presenter_test.rb
```diff
@@ -43,10 +43,19 @@ def iso8601_regex
       create(:speech,
              title: "Speech title",
              summary: "The description",
-             body: "Some content",
+             body: "# Woo!\nSome content",
              role_appointment: role_appointment)
     end
 
+    let(:expected_body) do
+      expected_body = <<-HTML.strip_heredoc
```
andrewgarner: I think you can call `chomp` after `strip_heredoc` like this:
```rb
<<-HTML.strip_heredoc.chomp
<div>foo</div>
HTML
# => "<div>foo</div>"
```

steventux: :+1: good to know

PR #3061, opened 2017-02-21
Switch News Article rendering app to government-frontend

Rosa-Fox:
[Trello](https://trello.com/c/4tpycpbA/617-news-article-migration-5-final-tasks-deploy-1)

### DIFF
--- a/test/unit/presenters/publishing_api/generic_edition_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/generic_edition_presenter_test.rb
@@ -29,7 +29,7 @@ module PublishingApi
         need_ids: [],
         public_updated_at: edition.updated_at,
         publishing_app: 'whitehall',
-        rendering_app: 'whitehall-frontend',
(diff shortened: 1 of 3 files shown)


test/unit/presenters/publishing_api/generic_edition_presenter_test.rb
```diff
@@ -29,7 +29,7 @@ def present(edition, options = {})
         need_ids: [],
         public_updated_at: edition.updated_at,
         publishing_app: 'whitehall',
-        rendering_app: 'whitehall-frontend',
+        rendering_app: 'government-frontend',
```
Davidslv: You could use `Whitehall::RenderingApp::GOVERNMENT_FRONTEND` as the value there, so if it get's changed there's no need to change the test.

Also none of those constants are frozen https://github.com/alphagov/whitehall/blob/master/lib/whitehall/rendering_app.rb#L3

PR #3062, opened 2017-02-21
Add event tracking to checkboxes on topic tagging page

klssmith:
This sends data to Google Analytics when taxon checkboxes are checked or
unchecked. The data sent includes:

- whether the checkbox is checked or unchecked
- the taxon name of the checkbox
- the public path of the document
- the content format of the document
- the content id of the document

Trello card: https://trello.com/c/pbtbMZer/469-add-event-tracking-to-checkboxes-on-topic-tagging-page

Paired with @Davidslv

### DIFF
--- a/app/assets/javascripts/admin/modules/taxonomy_tree_checkboxes.js
+++ b/app/assets/javascripts/admin/modules/taxonomy_tree_checkboxes.js
@@ -43,11 +43,36 @@
       return $checkedSiblings.length > 0;
     };
 
+    var checkboxTrackClick = function(action, options) {
+      var root = window;
+      root.ga('send', 'event', {
+        eventCategory: options.eventCategory,
+        eventAction: action,
+        eventLabel: options.eventLabel,
+        eventValue: 1,
+        cd1: options.cd1,
+        cd2: options.cd2,
+        cd4: options.cd4
+      });
+    };
+
     this.start = function(element) {
-      $(element).on('click', 'input:checkbox', function() {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/modules/taxonomy_tree_checkboxes.js
```diff
@@ -56,8 +81,10 @@
         and the content could be shown to users seeking any of them.
         */
         if (checked) {
+          checkboxTrackClick("checkboxClickedOn", options);
```
MatMoore: When we automatically check/uncheck the ancestors/descendants, does this trigger another event that causes those checkboxes to be tracked as click?

I think we'd want to track just the user behaviour, since we are using number of clicks as a proxy for how confident the users are when using the interface, and the actual number of checked/unchecked depends on where in the taxonomy you're tagging to.

Davidslv: @MatMoore this only triggers when the user clicks the actual checkbox, the ancestors will not be triggering this, checking the ancestors is responsibility of `checkAncestors(this)` function

Here I've clicked on "Maths" and "GCSE subject content and requirements"
screen shot 2017-02-22 at 11 05 30


Here I've unchecked "GCSE subject content and requirements"

screen shot 2017-02-22 at 11 06 06


You can see from the screenshots that the event is only triggered once per click.
I hope that answers your question.

PR #3063, opened 2017-02-21
Use `links#policies` 

gpeng:
https://github.com/alphagov/govuk-content-schemas/pull/532 makes all Whitehall formats consistently use `links#policies` rather than `links#related_policies`. 

This PR updates the format presenters to use the correct link name and the sync checks for edition formats to check it correctly.

Needs [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas/pull/532) to be deployed to get the tests to pass.

### DIFF
--- a/app/presenters/publishing_api/case_study_presenter.rb
+++ b/app/presenters/publishing_api/case_study_presenter.rb
@@ -33,7 +33,7 @@ module PublishingApi
         [
           :organisations,
           :parent,
-          :related_policies,
+          :policies,
           :topics,
           :world_locations,
           :worldwide_organisations,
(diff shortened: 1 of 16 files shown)


fofr:
Does government-frontend need updating?

gpeng:
I don't think we need to update `government-frontend` until we remove `related_policies` from the schemas altogether. Looks like it already supports both.

https://github.com/alphagov/government-frontend/blob/master/app/presenters/linkable.rb#L11-L12

gpeng:
Closing this as we've decided to go with `related_policies` for everything instead 🙀 which avoids clashes with the `working_group` -> `policies` reverse links in publishing api.

PR #3066, opened 2017-02-22
Fix collections

mgrassotti:
Trello: https://trello.com/c/39Aq7Oxm/598-whitehall-routes-for-document-collections

Some collections were not correctly migrated, so we republish them in
Whitehall.


Davidslv:
is there a schema.rb change to go with this commit?

mgrassotti:
@Davidslv I don't think so, as this is just a data migration

PR #3006, opened 2017-02-01
Remove Consultation presentation

andrewgarner:
Consultations have been migrated and are now rendered by [Government Frontend](alphagov/government-frontend). This removes the frontend code that is no longer needed.

[Trello](https://trello.com/c/LYFL1ZH9/)

PR #3067, opened 2017-02-22
Revert "Remove Consultation presentation"

gpeng:
Reverts alphagov/whitehall#3006 We need to contnue to use Whitehall for preview until we implement the draft links as consultations can have `HtmlAttachment`s.


fofr:
@gpeng Can be bring this back now?

PR #3070, opened 2017-02-22
Prevent page crashing when edition has no expanded_links

Davidslv:
On whitehall, when a user creates a new document, and that document is then associated with an Education organisation (under lead organisations) and saves the page, it crashes the next page.

This is because 

>A draft can be tagged by the new taxonomy tagging form, but it doesn't receive other kinds of tags (set in the edit form) until publish time. All the tags except for taxons are coupled to the 2i workflow at the moment.

This commit rescues `GdsApi::HTTPNotFound ` and uses a new object, `MissingExpandedLinks`,
this object is responsible to solely return an empty array and prevent the view to crash.

Trello:
https://trello.com/c/nsuNcWNC/504-fix-bug-with-editing-a-draft-document-and-expanded-links

### Before
screen shot 2017-02-22 at 18 16 26

### After

screen shot 2017-02-22 at 18 16 49

### DIFF
--- a/app/services/expanded_links_fetcher.rb
+++ b/app/services/expanded_links_fetcher.rb
@@ -9,6 +9,18 @@ class ExpandedLinksFetcher
     ExpandedLinks.new(
       Whitehall.publishing_api_v2_client.get_expanded_links(content_id)
     )
+
+  rescue GdsApi::HTTPNotFound
+    MissingExpandedLinks.new
+  end
+
+  # TODO: This is a workaround, because Publishing API
+  # returns 404 when the document exists but there are no links
+  # This can be removed when that changes.
+  class MissingExpandedLinks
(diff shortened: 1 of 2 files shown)


tijmenb:
> This is because Publishing API didn't know about the expanded links of this draft document.

Shouldn't the publishing-api know the links for a draft? It should be possible to tag drafts and then publish them.

MatMoore:
@tijmenb 

A draft can be tagged by the new taxonomy tagging form, but it doesn't receive other kinds of tags (set in the edit form) until publish time. All the tags except for taxons are coupled to the 2i workflow at the moment.

So we had this case where the topics section can't be rendered until you've either visited the new tagging form, or published the document (and you can't do either if this page errors)

PR #3071, opened 2017-02-22
Migrate needs to the Publishing API

cbaines:
Recently, some work was done regarding Maslow and the Need API, to turn needs in to content in the Publishing API.

Whitehall stores, and can edit what needs are associated with detailed guides. Currently this is done by storing a list of numerical identifiers (need ids), and sending them to the Publishing API. The new way of doing this is to store this data in the Publishing API in the form of links.

This branch includes the changes to make that happen. A one off rake task to make the patch links requests to the Publishing API to migrate the data, and changes to the Whitehall models and views to fetch and modify the data in the Publishing API. For the users of Whitehall, the interface on the detailed guides form will change, it will now look like this:

new-detailed-guide-user-needs


cbaines:
I've marked this as DO NOT MERGE, as testing on integration has shown that it just makes creating/editing detailed guides unusable, as the Publishing API linkables endpoint times out.

cbaines:
> I've marked this as DO NOT MERGE, as testing on integration has shown that it just makes creating/editing detailed guides unusable, as the Publishing API linkables endpoint times out.

Two improvements to this have now been merged [1][2], so I'll remove the DO NOT MERGE notice.

1: https://github.com/alphagov/publishing-api/pull/799
2: https://github.com/alphagov/publishing-api/pull/800

app/models/edition.rb
```diff
@@ -604,19 +601,17 @@ def body_required?
   end
 
   def has_associated_needs?
-    associated_needs.any?
+    respond_to?(:need_ids) && need_ids.try(:any?)
   end
 
   def associated_needs
```
cbaines: > We weren't checking if the object responds_to?(:need_ids) before. Is that necessary?

If you don't have that check, and you call that `has_user_needs?` on some Edition subclass that is not a DetailedGuide, then you would get an error.

One alternative that I see would be to move the needs related code out of the DetailedGuide class and up to the Edition? That would then mean that need_ids would be defined for all Editions.

app/models/detailed_guide.rb
```diff
@@ -62,6 +65,35 @@ def related_detailed_guide_ids=(detailed_guide_ids)
     self.related_documents = other_related_documents + detailed_guide_documents
   end
 
+  def need_ids
+    return @need_ids if @need_ids
+    return [] unless document # TODO: Remove
```
cbaines: So I obviously missed the TODO, as that was meant for me... Anyway, looking again at this, the issue is that with new Editions, they do not have a document, and therefore, I don't think you can look up the corresponding needs.

I've moved this around a bit, and made the comment more useful. Updated patch here b8ae67778eda989243ee76cec6c59b41387f7e51 .

app/models/detailed_guide.rb
```diff
@@ -62,6 +65,35 @@ def related_detailed_guide_ids=(detailed_guide_ids)
...
+      document.content_id
+    )
+
+    return unless response
+
+    @need_ids = response["links"]["meets_user_needs"]
+  end
+
+  def patch_meets_user_needs_links
+    return unless @need_ids
```
cbaines: That seems sensible, new patch here b8ae67778eda989243ee76cec6c59b41387f7e51 .

app/models/detailed_guide.rb
cbaines: I've updated this again now ce95dea6aeea0fb096a41d368384639db9c762ce .

app/models/detailed_guide.rb
cbaines: Actually, now that the test suite has finished running, there are quite a few new errors somewhat related to this. If the `patch_meets_user_needs_links` method uses `need_ids` rather than accessing the value directly, if you just have an Edition, and you save it, `patch_meets_user_needs_links` might lookup the current needs by calling `need_ids` and then proceed to call out to the Publishing API.

However, If you access the value directly, you avoid the possibility that what you are considering saving has actually just been fetched from the Publishing API. I've changed this back for now, what do you think?

app/models/edition.rb
gpeng: It seems like this should either all go in `DetailedGuide` if that is the only format that has needs or else it should all go in the base class (or preferably a separate class as `Edition` is already huge) if we're going to allow all formats to have needs now they're in the publishing API. Also, all `Edition`s will respond to `need_ids` and return an `Array` so if we are going to have `has_associated_needs?` it's unnecessarily defensive.

db/schema.rb
```diff
@@ -430,7 +430,6 @@
     t.datetime "opening_at"
     t.datetime "closing_at"
     t.integer  "corporate_information_page_type_id",          limit: 4
-    t.string   "need_ids",                                    limit: 255
```
gpeng: I think you need to check the schema back out as the migration that caused this is no longer here.

app/models/detailed_guide.rb
gpeng: Is this being caused by the schema issue in my other comment? Either way, while you've still got a database field for `need_ids` this might be difficult to do.

db/schema.rb
cbaines: I'm unsure what you mean by "check the schema back out", can you clarify?

app/models/edition.rb
cbaines: I've collected all the needs related code in one place now, and removed `has_associated_needs?`.

app/models/detailed_guide.rb
cbaines: I think the current code does manage, though this does require the rather ugly `return if @need_ids.is_a? String`. If the `need_ids` column is removed from Whitehall, this can then be removed.

db/schema.rb
gpeng: It looks like a migration has been run to remove this field but the migration isn't in this PR. I meant checkout the schema from master to put this field back and reset this change.

cbaines: So I deleted this entry, as I thought this might convince ActiveRecord not to select it from the database. This is probably wrong, so I've removed the commit deleting this line.

cbaines:
Could you take another look at this @gpeng :)

cbaines:
@gpeng @boffbowsh I checked with Tijmen, and content-tagger currently GDS only, so linking to that from Whitehall is currently not an option.

gpeng:
So I think the next option we discussed was to take all this out of the edition workflow and add do it at the document level. That would mean we could also keep it as an 'old school' link rather than and 'edition' type.

rubenarakelyan:
Closing this for the moment as it is not being actively worked on.

elenatanasoiu:
Picking this up again for: [Trello card](https://trello.com/c/3JN3SdI3/193-migrate-needs-to-the-publishing-api)

elenatanasoiu:
The form for adding/editing needs has been moved from the Editions edit page to its own page. This has been done following a discussion with @gpeng where we decided that we wanted to bypass the whole edition workflow of creating a draft, publishing etc. That's because you don't really need to create a new edition draft just to edit the needs of a document.

This part of the change to move need assignment workflow from the Edition level to the Document level. (done here: https://github.com/alphagov/whitehall/pull/3071/commits/29deb5e4295f5e494db82bba48b6cc10fb36625f)

However, since the document does not have a show/edit page on Whitehall, we've still had to keep the form at the edition level, so we've added a "Add/remove needs" button on the Edition show page. 

## 1. Step 1: Edition show page

Notice the "Add/remove needs" that you can click to get to the "Editing needs" form
screen shot 2017-08-10 at 14 26 56

## 2. Step 2: Editing needs form 
screen shot 2017-08-10 at 14 27 08

Clicking inside the select gives you more needs to choose from:
screen shot 2017-08-10 at 14 27 19

elenatanasoiu:
Tests are failing at the moment because of an issue with Jenkins.

app/controllers/admin/documents_controller.rb
```diff
@@ -9,4 +9,19 @@ def by_content_id
...
+
+  def update_needs
+    @edition = Edition.find(params[:edition_id])
+    document = @edition.document
+
+    document.need_ids = params[:need_ids] || []
+    document.patch_meets_user_needs_links
+
+    redirect_to admin_edition_path(@edition)
+  end
```
h-lame: The REST purist in me wonders if these could be show and update on a needs controller that we hit via `GET /:content_id/needs/` and `PUT /:content_id/needs` ?  It feels slightly odd that we use the edition id in a documents controller.

lib/tasks/migrate_need_ids.rake
```diff
@@ -0,0 +1,51 @@
...
+          WHERE editions.document_id = documents.id
+          AND need_ids IS NOT NULL
+          ORDER BY editions.id DESC LIMIT 1) need_ids
+  FROM documents) results
+WHERE need_ids IS NOT NULL;")
+
+  not_found_need_ids = {}
+
+  results.each do |content_id, raw_need_ids|
+    need_ids = raw_need_ids.lines.map { |line|
```
gpeng: `do end` syntax here would match the other blocks

lib/tasks/migrate_need_ids.rake
```diff
@@ -0,0 +1,51 @@
...
+      rescue GdsApi::HTTPNotFound
+        nil
+      end
+    end
+
+    mapping = need_ids.zip(need_content_ids).map { |n, c| "  #{n} -> #{c}" }.join("\n")
+    puts "#{content_id}:\n#{mapping}\n"
+
+    not_found_count = need_content_ids.count(nil)
+    not_found_need_ids[content_id] = not_found_count if not_found_count.nonzero?
```
gpeng: I'm finding this hard to read, partly because of the naming here but also I think it would be better to maybe split some of this rake task out into smaller methods, the naming of which would help make it easier to follow.

app/helpers/admin/taggable_content_helper.rb
```diff
@@ -55,6 +55,22 @@ def taggable_ministerial_role_appointments_container
     end
   end
 
+  def taggable_needs_container
+    Rails.cache.fetch("need.linkables", expires_in: 1.minute) do
+      publishing_api_with_low_timeout = Services.publishing_api.dup.tap do |client|
```
davidbasalla: Just curious, what's the reasoning for caching this?

elenatanasoiu: I'm assuming it's because we want to make needs more robust when it comes to the Publishing API going down, as the same technique is used for policies (https://github.com/alphagov/whitehall/commit/b69d842c4111fcc7cfa39c08bbda623d2fc7262c) but maybe @cbaines can explain this a bit more in detail.

cbaines: > Just curious, what's the reasoning for caching this?

As @elenatanasoiu says, I think I copied it from the code that works with Policies, and caching will help with reliability.

https://github.com/alphagov/whitehall/blob/64549b845d58e3d4f60509f3b42ef88f93f4d728/app/models/policy.rb#L44-L57

lib/tasks/migrate_need_ids.rake
elenatanasoiu: The `do end` syntax is rejected by Rubocop unfortunately.

features/step_definitions/needs_steps.rb
```diff
@@ -0,0 +1,15 @@
...
+end
+
+When(/^I choose the first need in the dropdown$/) do
+  option = first('#need_ids option').text
+  select option, from: 'need_ids'
+  click_button 'Save needs'
+end
+
+Then(/^I should see the first need in the list of associated needs$/) do
+  page.find("h2:contains('Attachments')")
```
elenatanasoiu: It was checking whether we were on the right page (since that would have "Attachments"). But that could change so I'm removing it.

app/controllers/admin/documents_controller.rb
elenatanasoiu: I've moved the actions for editing/updating needs into a needs_controller. Unfortunately I still  need an edition ID in order to return the person to the right edition show page after they've finished choosing their needs. 

TL;DR The transition from "Edit form for needs" to -> "Show page for edition" requires that I know the edition ID.

lib/tasks/migrate_need_ids.rake
elenatanasoiu: @gpeng I've tried to use better naming and I've separated out as much functionality as I think could be done in order to make the steps clearer. Let me know what you think.

app/models/document.rb
```diff
@@ -111,6 +113,37 @@ def humanized_document_type
     document_type.underscore.gsub('_', ' ')
   end
 
+  def need_ids
+    @need_ids ||= get_user_needs_from_publishing_api
+  end
+
+  def get_user_needs_from_publishing_api
```
elenatanasoiu: I've moved this into its own concern (Document::Needs)

cbaines:
> The rake task has been run successfully on integration.

Awesome :tada:

elenatanasoiu:
After running the rake task on integration/staging/production we've got the following two documents with needs that don't exist: 

```
Documents where need content ids could not be found:
  5f616c6e-7631-11e4-a3cb-005056011aef: 1
  06e683e5-f026-4214-819b-7ebd232a884c: 1
```

1. I've traced the first document (5f616c6e-7631-11e4-a3cb-005056011aef) as having the following missing need: 101095.

This need does not exist in Maslow, although need 101094 and 101096 do exist, which might indicated that need was deleted at some point. The document is also connected to a SINGLE, DELETED edition from 06-06-2014, so having a missing need will not affect any published document. 

2. Second document (06e683e5-f026-4214-819b-7ebd232a884c) had the missing need: 103837. 

This need also doesn't exist in Maslow and might have been deleted. The document is associated to a SINGLE, DRAFT edition from 15-02-2017. Same as before, having a missing need will not affect it as the document has never been published.

PR #3073, opened 2017-02-23
Use current government when Speech#delivered_on is future date or nil...

steventux:

`Speech#delivered_on` is used to determine the Government attributed to the item,
if this date is nil or in the future the government won't be attributed to the
speech. This causes problems for schema validation and doesn't make sense in
the frontend. So if no delivered on date is present save the speech against
the current government, and if the delivered_on date is in the future also
assume this to be a speech from the current government.

### DIFF
--- a/app/models/speech.rb
+++ b/app/models/speech.rb
@@ -52,7 +52,11 @@ class Speech < Announcement
 private
 
   def date_for_government
-    delivered_on.try(:to_date)
+    if delivered_on && delivered_on.past?
+      delivered_on.to_date
+    else
+      super
+    end
   end
 
   def skip_organisation_validation?
--- a/test/unit/speech_test.rb
+++ b/test/unit/speech_test.rb
@@ -178,9 +178,15 @@ class SpeechTest < ActiveSupport::TestCase
     assert_equal previous_government, speech.government
   end
 
-  test '#government returns nil for an speech without a delivered_on' do
+  test '#government returns the current government for an speech delivered at an unspecified time' do
+    current_government = create(:current_government)
     speech = create(:imported_speech, delivered_on: nil)
-    assert_nil speech.government
+    assert_equal current_government, speech.government
   end
 
+  test '#government returns the current government for an speech in the future' do
+    current_government = create(:current_government)
+    speech = create(:imported_speech, delivered_on: 2.weeks.from_now)
+    assert_equal current_government, speech.government
+  end
 end


app/models/speech.rb
```diff
@@ -52,7 +52,13 @@ def delivered_by_minister?
 private
 
   def date_for_government
-    delivered_on.try(:to_date)
+    present_date = Time.zone.now.to_date
+
+    if delivered_on && present_date >= delivered_on
```
gpeng: I think you could use `delivered_on.past?` here

test/unit/speech_test.rb
```diff
@@ -178,9 +178,15 @@ class SpeechTest < ActiveSupport::TestCase
...
+  test '#government returns the current government for an speech without a delivered_on' do
+    current_government = create(:current_government)
     speech = create(:imported_speech, delivered_on: nil)
-    assert_nil speech.government
+    assert_equal current_government, speech.government
   end
 
+  test '#government returns the current government for an speech with a future delivered_on' do
+    current_government = create(:current_government)
+    speech = create(:imported_speech, delivered_on: 2.weeks.since(Date.today))
```
gpeng: `2.weeks.from_now` might be clearer.

app/models/speech.rb
```diff
@@ -52,7 +52,13 @@ def delivered_by_minister?
 private
 
   def date_for_government
-    delivered_on.try(:to_date)
+    present_date = Time.zone.now.to_date
```
andrewgarner: Did you consider using `super`?

```rb
if delivered_on.try(:future?)
  delivered_on.to_date
else
  super
end 
```

test/unit/speech_test.rb
```diff
@@ -178,9 +178,15 @@ class SpeechTest < ActiveSupport::TestCase
...
 
-  test '#government returns nil for an speech without a delivered_on' do
+  test '#government returns the current government for an speech without a delivered_on' do
+    current_government = create(:current_government)
     speech = create(:imported_speech, delivered_on: nil)
-    assert_nil speech.government
+    assert_equal current_government, speech.government
   end
 
+  test '#government returns the current government for an speech with a future delivered_on' do
```
andrewgarner:
```rb
test '#government returns the current government for a speech delivered in the future'
```

steventux:
thanks @andrewgarner good idea

test/unit/speech_test.rb
```diff
@@ -178,9 +178,15 @@ class SpeechTest < ActiveSupport::TestCase
     assert_equal previous_government, speech.government
   end
 
-  test '#government returns nil for an speech without a delivered_on' do
+  test '#government returns the current government for an speech without a delivered_on' do
```
andrewgarner:
```rb
test '#government returns the current government for a speech delivered at an unspecified time'
```

app/models/speech.rb
```diff
@@ -52,7 +52,11 @@ def delivered_by_minister?
 private
 
   def date_for_government
-    delivered_on.try(:to_date)
+    if delivered_on && delivered_on.past?
+      delivered_on
```
andrewgarner: `delivered_on` is a time, in the `super` method `to_date` is always called, I assume so that a Government can be found at the start of the day the speech was delivered on. We should probably stick with that approach here.

PR #3075, opened 2017-02-23
[WIP] Adds Metadata page to HTML attachment print

elenatanasoiu:
[Trello card](https://trello.com/c/88zaXS8J/47-create-print-styles-for-official-documents)

### 1. New metadata page 

An extra page has been added. It contains:
- the copyright logo
- document metadata: Web ISBN & contact address
- copyright text

The page will only be visible when printing HTML attachments.
screen shot 2017-03-02 at 11 39 53

Here is the PDF version on which it was based: 

screen_shot_2016-09-14_at_14 32 28

### 2. Changes to Edit form for HTML Attachment

- Adds a `:meta_data_contact_address` field to the HtmlAttachment form.
- Renames "ISBN" label to "ISBN (web version)" to make it clear that it is not the Print ISBN.

** Before **

screen shot 2017-02-24 at 16 56 36

** After **

screen shot 2017-03-02 at 11 45 36

## Assumptions

1. The year of the copyright notice is considered the year when the document was created

2. There are two sentences that might not make sense when someone is printing a document from their own printer, and have been removed from the final page:

```
Printed on paper containing 75% recycled fibre content minimum.

Printed in the UK by the Williams Lea Group on behalf of the Controller of Her Majesty's Stationery Office
```

3. An assumption was made regarding classifying the ISBN field as a Web ISBN. This is because
in the PDF version there are two ISBNs present in the Metadata page: the Print ISBN and the
Web ISBN. Checking the database, we can find 10 distinct ISBNs assigned to 74 different
HtmlAttachment documents. Through a google lookup, we can identify at least 4 that are
marked as Web ISBNs on pdf documents belonging to gov.uk. Two others are just marked as ISBN. Here are the URLs below:

- Web ISBNs:

https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/446159/Annual_Report_201415.pdf
https://www.google.co.uk/url?sa=t&rct=j&q=&esrc=s&source=web&cd=3&cad=rja&uact=8&ved=0ahUKEwjZiPWejanSAhUhD8AKHRXOBKwQFggiMAI&url=http%3A%2F%2Fwww.milton-keynes.gov.uk%2Fassets%2Fattach%2F22362%2FCode_of_Practice_on_illegal_immigrants_and_private_rented_accommodation__web_.pdf&usg=AFQjCNHtjH9YEn4gT9AwQDAIC9gIfE4Y_A&sig2=jhIz4a1Njaw0hJ8b6SAjlw&bvm=bv.148073327,d.ZGg
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/505928/53949_Armed_Forces_Pay_Review_2016_Accessible.pdf
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/538537/Annual_Report_2016.pdf

- ISBN:

http://content.digital.nhs.uk/catalogue/PUB18117/nhs-out-fram-ind-sep-15-comm.pdf
http://www.housinglin.org.uk/_library/Resources/Housing/OtherOrganisation/An_introduction_to_the_Home_Building_Fund.pdf


app/controllers/admin/attachments_controller.rb
```diff
@@ -103,9 +103,10 @@ def build_file_attachment
 
   def attachment_params
     params.fetch(:attachment, {}).permit(
-      :title, :locale, :isbn, :unique_reference, :command_paper_number,
-      :unnumbered_command_paper, :hoc_paper_number, :unnumbered_hoc_paper,
-      :parliamentary_session, :order_url, :price, :accessible, :external_url,
+      :title, :locale, :isbn, :meta_data_contact_address, :unique_reference,
+      :command_paper_number, :unnumbered_command_paper, :hoc_paper_number,
+      :unnumbered_hoc_paper, :parliamentary_session, :order_url, :price, :accessible,
+      :external_url,
```
h-lame: Totally minor, but as this isn't alphabetical there's no reason to insert the new attr in the middle, so it would be less noisy (and thus clearer what the change was) if you added `:meta_data_contact_address` at the end of the list rather than in the middle.

features/attachment-preview.feature
```diff
@@ -1,4 +1,4 @@
-Feature: Previewing CSV attachcments
+Feature: Previewing CSV attachments
```
h-lame: :+1:

features/edition-attachments.feature
```diff
@@ -54,3 +54,17 @@ Feature: Managing attachments on editions
     And a draft publication "Standards on Beard Grooming" exists
     When I try and upload an attachment but there are validation errors
     Then I should be able to submit the attachment without re-uploading the file
+
+  Scenario: Editing metadata on attachments
+    Given I am an writer
+    And I start drafting a new publication "Standard Beard Lengths"
+    When I start editing the attachments from the publication page
+    And I upload an html attachment with the title "Beard Length Graphs 2012" and the isbn "123456789" and the contact address "Address 1"
+
```
h-lame: There's no `Then` here, so it feels like the feature doesn't actually describe a complete scenario. I know that it *does* test something, but it doesn't really read like it does. Can we combine this test and the next one?

features/step_definitions/attachment_steps.rb
```diff
@@ -123,3 +123,28 @@
...
+  create(:html_attachment, attachable: publication, title: attachment_title, meta_data_contact_address: contact_address)
+end
+
+When /^I preview the html attachment "(.*?)" in print mode$/ do |attachment_title|
+  html_attachment = HtmlAttachment.find_by(title: attachment_title)
+  publication = html_attachment.attachable
+  visit publication_html_attachment_path(publication_id: publication.slug, id: html_attachment.slug, medium: "print")
+end
+
+Then /^I should see the metadata page with contact address "(.*?)"$/ do |contact_address|
```
h-lame: You could probably combine this `Then` with the previous `When` so they're more obviously connected.  `Then previewing the html attachment "..." in print mode includes the contact address "..." and isbn "..."` - something like that?  I reckon it'd be good to include checking for the isbn too as it's also a dynamic part of that page.

db/migrate/20170223180546_add_metadata_fields_to_attachments.rb
```diff
@@ -0,0 +1,5 @@
+class AddMetadataFieldsToAttachments < ActiveRecord::Migration
+  def change
+    add_column :attachments, :meta_data_contact_address, :string
```
h-lame: Maybe this should be `print_meta_data_contact_address`, so it's obvious that this is to be used for print only.

app/views/admin/attachments/_reference_fields.html.erb
```diff
@@ -3,7 +3,10 @@
   <p class="help-block">This determines the translations of the publication that the attachment will appear in.</p>
   <%= form.select :locale, options_for_locales(attachable.translated_locales), include_blank: 'All languages' %>
 <% end %>
-<%= form.text_field :isbn, label_text: 'ISBN' %>
+
+<%= form.text_field :isbn, label_text: "ISBN (web version)" %>
+<%= form.text_area :meta_data_contact_address, rows: 5, label_text: "Organisation's Contact Details" %>
```
h-lame: Worth checking with the content team about this UI, but it might benefit from a hint about what sort of things to include here.  We can put the address, an email etc... - should probably make that clear.

h-lame: And indicate what, if any, styling options they have, because there's probably an expectation that any text box is a govspeak box (maybe not though - definitely check with content folk).

app/views/html_attachments/_print_meta_data.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  © Crown copyright <%= Date.today.year %>
+</p>
+<p>
+  This publication is licensed under the terms of the Open Government Licence v3.0 except where otherwise stated. To view this licence, visit <a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3">nationalarchives.gov.uk/doc/open-government-licence/version/3</a> or write to the Information Policy Team, The National Archives, Kew, London TW9 4DU, or email: <a href="mailto:psi@nationalarchives.gsi.gov.uk" class="meta-data-email">psi@nationalarchives.gsi.gov.uk</a>.
+</p>
+<p>
+  Where we have identified any third party copyright information you will need to obtain permission from the copyright holders concerned.
+</p>
+<p>
+  This publication is available <a href="www.gov.uk/government/publications">www.gov.uk/government/publications</a>
```
h-lame: We're missing an `at` here.  The URL could also be the full one to the attachment (or it's parent edition) rather than the index of all publications.

app/views/html_attachments/_print_meta_data.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  <p>
+    Web ISBN: <%= @html_attachment.isbn %>
+  </p>
+<% end %>
+<p>
+  Printed on paper containing 75% recycled fibre content minimum
+</p>
+<p>
+  Printed in the UK by the Williams Lea Group on behalf of the Controller of Her Majesty's Stationery Office
+</p>
```
h-lame: Worth checking with content team/whomever requested the story, but these last two paragraphs seem out of place given they're not going to be true for anyone that visits the website and prints the html attachment on their printer at home.

app/views/html_attachments/_print_meta_data.html.erb
```diff
@@ -0,0 +1,40 @@
...
+<br/>
+<br/>
+<p>
+<%= image_tag "open-government-licence.png", id: "meta-data-licence" %>
+</p>
+<br/>
+<br/>
+<br/>
+<p>
+  © Crown copyright <%= Date.today.year %>
```
h-lame: I don't know the answer to this, but should this be the year that the html attachment / it's edition was first published?  Or do we want to keep it constantly updated?

app/assets/stylesheets/frontend/print/_html-publication.scss
```diff
@@ -53,4 +53,20 @@
       padding: 0;
     }
   }
+
+  #meta-data {
+    page-break-before: always;
+    page-break-after: always;
+
+    #meta-data * {
```
h-lame: Does this selector actually do anything?  I think scss will turn this into `#meta-data #meta-data *` which won't match, so I suspect we can drop this entirely.

app/assets/stylesheets/frontend/print/_html-publication.scss
```diff
@@ -53,4 +53,20 @@
...
+      visibility: visible;
+      height: 100%;
+      width: 100%;
+    }
+
+    #meta-data-licence {
```
h-lame: This could probably be done with a class rather than an id.

app/assets/stylesheets/frontend/print/_html-publication.scss
```diff
@@ -53,4 +53,20 @@
       padding: 0;
     }
   }
+
+  #meta-data {
```
h-lame: Like previous comment about the contact field, this might want to be called `print-meta-data` so it's more obvious what it is (`meta-data` is a pretty generic name and could be re-used on other elements).

features/step_definitions/attachment_steps.rb
```diff
@@ -123,3 +123,28 @@
...
+
+Given(/^the publication "(.*?)" has an html attachment "(.*?)" with the contact address "(.*?)"$/) do |publication_title, attachment_title, contact_address|
+  publication = Publication.find_by(title: publication_title)
+  create(:html_attachment, attachable: publication, title: attachment_title, meta_data_contact_address: contact_address)
+end
+
+When /^I preview the html attachment "(.*?)" in print mode$/ do |attachment_title|
+  html_attachment = HtmlAttachment.find_by(title: attachment_title)
+  publication = html_attachment.attachable
+  visit publication_html_attachment_path(publication_id: publication.slug, id: html_attachment.slug, medium: "print")
```
h-lame: `medium: 'print'` is super useful - good sleuthing!

features/edition-attachments.feature
elenatanasoiu: done

app/controllers/admin/attachments_controller.rb
elenatanasoiu: done

features/step_definitions/attachment_steps.rb
elenatanasoiu: thanks!

features/step_definitions/attachment_steps.rb
elenatanasoiu: done

db/migrate/20170223180546_add_metadata_fields_to_attachments.rb
elenatanasoiu: done

app/views/admin/attachments/_reference_fields.html.erb
elenatanasoiu: Added hint, will ask content

app/views/html_attachments/_print_meta_data.html.erb
elenatanasoiu: I've changed to be the year when the html attachment was created. I've added that to the assumptions.

app/views/html_attachments/_print_meta_data.html.erb
elenatanasoiu: added

app/views/html_attachments/_print_meta_data.html.erb
elenatanasoiu: Will ask content team

elenatanasoiu: I've removed it for now and added in assumptions

app/assets/stylesheets/frontend/print/_html-publication.scss
elenatanasoiu: Renamed

app/assets/stylesheets/frontend/print/_html-publication.scss
elenatanasoiu: done

app/assets/stylesheets/frontend/print/_html-publication.scss
elenatanasoiu: Removed

app/views/admin/attachments/_reference_fields.html.erb
elenatanasoiu: The hint is `This can contain the address, email, or telephone number`

app/views/html_attachments/_print_meta_data.html.erb
elenatanasoiu: I've checked with Theo, who wrote the requirements. He said it makes sense to have it as the year when the html attachment was created.

app/assets/stylesheets/frontend/layouts/html-publication.scss
```diff
@@ -26,7 +26,7 @@
   }
 }
 
-#meta-data {
+#print-meta-data {
```
h-lame: This is pedantic, but these changes (where you rename css id's, css classes, and file names from `meta-data` to `print-meta-data`) would be better squashed down into the previous commit.  They're really not much to do with this commit because if for some strange reason we decided we didn't want the new fields you add here we'd still want the classes and such to be called `print-meta-data`.

h-lame:
Good stuff.  I particularly like the explicit calling out of your assumptions in the PR notes!

fofr:
This is a nice PR :100:

Do you have some screenshots for what happens when viewing the page without printing?

app/assets/stylesheets/frontend/layouts/html-publication.scss
```diff
@@ -25,3 +25,9 @@
     padding-top: 0;
   }
 }
+
+#print-meta-data {
```
fofr: A heads up, these styles don't pass our sass lint checks (which are in government-frontend)

Sass linting prevents:
* styling based on ID
* adding units to `0`s

Why did you opt for `visibility: hidden;` over `display: none;`?

elenatanasoiu: @fofr Thanks for the heads up! 

> Why did you opt for visibility: hidden; over display: none;?
It was the preferred way to hide items in a CSS Tricks article for print version because it wasn't affecting the layout. But it makes no difference here (that I can see) so I've switched to `display:none;`

I've reimplemented this for government-frontend, where I:
- use `display:none;`
- avoid styling based on IDs
- no units added to zeros

Please see the corresponding PR on `government-frontend`: https://github.com/alphagov/government-frontend/pull/288

PR #3078, opened 2017-02-23
Squish alternative_path in PublishingApiGoneWorker

bevanloon:
Trailing spaces in `alternative_urls` are being sent to the publishing api. These are stored verbatim in the content store (with trailing spaces).

Sync checks subsequently fail because, although the sync checker strips off trailing spaces from the data returned from whitehall's database, it does not strip off spaces in alternative_urls brought back from the content store.

Fix the problem by squishing the data before we send it to the publishing api.

[Trello](https://trello.com/c/DgBRDuIK)

### DIFF
--- a/app/workers/publishing_api_gone_worker.rb
+++ b/app/workers/publishing_api_gone_worker.rb
@@ -5,9 +5,11 @@ class PublishingApiGoneWorker < PublishingApiWorker
         .new.govspeak_to_html(explanation)
     end
 
+    alternative_path_with_no_trailing_space = alternative_path.rstrip if alternative_path
+
     Whitehall.publishing_api_v2_client.unpublish(
       content_id,
-      alternative_path: alternative_path,
+      alternative_path: alternative_path_with_no_trailing_space,
       explanation: rendered_explanation,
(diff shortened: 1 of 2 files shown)


app/workers/publishing_api_gone_worker.rb
```diff
@@ -5,6 +5,8 @@ def perform(content_id, alternative_path, explanation, locale, allow_draft = fal
         .new.govspeak_to_html(explanation)
     end
 
+    alternative_path.squish! if alternative_path
```
gpeng: It probably won't cause any issues in this instance but using the `squish!` here is going to mutate the argument that has been passed in outside of the scope of this class. I think it would be 'better form' to not do that.

bevanloon: Have changed the implementation in c7b0f842c7185ab007b4bd6f5a8c2b68bb80b4ae

I decided to use `rstrip` at the same time instead of the original `squish` just to explicitly remove only the trailing whitespace, rather than the blanket removal and normalisation of whitespace that `squish` does.

PR #3082, opened 2017-02-24
Update slimmer to 10.1.3

thomasleese:


tijmenb:
You can also remove https://github.com/alphagov/whitehall/blob/19cd7d72de32454d532c195f35b027fa1b3ba6ac/config/environments/production.rb#L101, that doesn't do anything anymore.

PR #3083, opened 2017-02-24
[DO NOT MERGE] Fix other migrated formats

mgrassotti:
https://trello.com/c/PX2ouzoX/597-check-migrated-routes

Some pages still reference to`whitehall-frontend` in the router. This should not happen for migrated formats, so we are going to republish
them.

HtmlAttachments and Collections were already fixed in other PRs.


gpeng:
Does this actually fix them? All of these documents have already been republished without the routes being updated.

mgrassotti:
It should be together with https://github.com/alphagov/router-api/pull/100, but yes I still have to test it in integration.

mgrassotti:
not needed as the routes still on whitehall-frontend result draft or withdrawn

PR #3091, opened 2017-02-28
Remove related_policies links check for Speeches

steventux:

`PublishingApi::SpeechPresenter` doesn't present `related_policies` links as these
are keyed as `policies` so don't check for them.


steventux:
Apparently this is the opposite of true. Closing.

PR #3094, opened 2017-02-28
Allow gov.scot offsite links

NeilvB:
[Trello card](https://trello.com/c/DtFNAqmQ/116-allow-external-links-to-scottish-government)


app/models/offsite_link.rb
```diff
@@ -46,6 +46,14 @@ def to_s
 
 private
 
+  def government_or_whitelisted_url?(host)
```
h-lame: :+1: - this is a nice improvement.

PR #3096, opened 2017-03-02
Track organisation pages as `finding` page in Google Analytics

suzannehamilton:
Add meta tag which marks the organisation homepage and individual organisation pages as being in the `finding` (navigation) stage of a user journey, as opposed to the `thing` (content) stage.

This will enable us to analyse how users navigate the site so that we can work out whether the new navigation pages are helping users find what they need.

https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/views/organisations/index.html.erb
+++ b/app/views/organisations/index.html.erb
@@ -1,6 +1,7 @@
 <% page_title "Departments, agencies and public bodies" %>
 <% page_class "index-list-page" %>
 <%= api_link_tag api_organisations_url %>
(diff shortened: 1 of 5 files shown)


suzannehamilton:
Closing this PR because we're taking a different approach: defining a document supertype in [govuk_document_types](https://github.com/alphagov/govuk_document_types).

suzannehamilton:
It turns out that we still need to do this for organisation homepages, because they don't have content store items. I've extracted that part of the change and opened PR #3099.

PR #3099, opened 2017-03-02
Track organisation homepage as a `finding` page in Google Analytics

suzannehamilton:
Add meta tag which marks the organisation homepage as being in the `finding` (navigation) stage of a user journey, as opposed to the `thing` (content) stage.

This will enable us to analyse how users navigate the site so that we can work out whether the new navigation pages are helping users find what they need.

Most pages use the shared meta tags in static to track this value, but organisation homepages are not in the content store so this meta tag must be added separately.

https://trello.com/c/SCsj2A9D/425-add-custom-dimension-to-page-view-tracking-on-new-navigation-pages

### DIFF
--- a/app/views/organisations/index.html.erb
+++ b/app/views/organisations/index.html.erb
@@ -1,6 +1,7 @@
 <% page_title "Departments, agencies and public bodies" %>
 <% page_class "index-list-page" %>
 <%= api_link_tag api_organisations_url %>
+<meta name="govuk:user-journey-stage" content="finding" />
 
 <div class="organisations-index">
   <div class="block-1">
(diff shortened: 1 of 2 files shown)


tijmenb:
If it's not too much trouble, we could also add the page to the content store. Would be one less snowflake to worry about.

suzannehamilton:
@tijmenb Sure! Does that mean adding it to this list like we did for the A/B example page? (In whitehall rather than frontend this time.) https://github.com/alphagov/whitehall/blob/master/lib/tasks/publishing_api.rake ❄️

tijmenb:
Yep. They're defined here: https://github.com/alphagov/whitehall/blob/master/lib/publish_static_pages.rb

suzannehamilton:
@tijmenb I've raised #3101.

I think we need this change as well, though, because the the new content item will have the format `placeholder`, which won't be tagged with the `finding` document supertype. Or have I missed something?

suzannehamilton:
Replaced by #3102 and #3101.

PR #3108, opened 2017-03-09
Remove NewsArticle public frontend code 

mgrassotti:
https://trello.com/c/WTk6X44o/626-news-article-migration-final-tasks-6-deploy-2-0-5days


app/controllers/news_articles_controller.rb
```diff
@@ -1,18 +1,8 @@
...
-    set_meta_description(@document.summary)
-  end
-
   private
 
-  def document_class
-    NewsArticle
-  end
-
   def analytics_format
```
fofr: Does this need to stay? Can the whole file go?

mgrassotti: 👍 yes you are right

mgrassotti:
yes it was... strange 🤔

gpeng:
Rebased from master as #3109 introduced a little conflict. @bevanloon could you just give this another quick once over?

gpeng:
I've got another PR brewing with a couple of other remnants. I'll remove the route in that.

PR #3109, opened 2017-03-09
Remove speech rendering

steventux:
https://trello.com/c/tXv5QqZG/615-speech-migration-final-tasks-6-deploy-2

Speeches are being rendered by government-frontend now so remove Whitehall frontend speech rendering code.


test/integration/document_locale_param_canonicalisation_test.rb
```diff
@@ -10,7 +10,7 @@ def with_locale_param(path, locale)
     u.to_s
   end
 
-  announcement_redir_document_types = %w(news_article speech)
```
fofr: Should this part be deleted? My interpretation is that It relates to the index of speeches which hasn't changed?

app/views/speeches/show.html.erb
```diff
@@ -1,6 +0,0 @@
-<% page_title edition_page_title(@document), t_display_type(@document, 10) %>
-<%= render  template: 'documents/show',
-            locals: {
-              header_title: t_display_type(@document),
-              show_share_links: true
```
fofr: At what point can the `show_share_links: true` part of the `show` template be removed? I think News articles, Speeches and Consultations were the only uses and they've all been migrated?

features/step_definitions/speech_steps.rb
```diff
@@ -61,15 +61,15 @@
...
   select first_policy, from: "Policies"
   select second_policy, from: "Policies"
   click_button "Save"
 end
 
 Then /^the published speech should remain unchanged$/ do
-  visit public_document_path(@speech)
-  assert page.has_css?('h1', text: @speech.title)
-  assert page.has_css?('.body', text: @speech.body)
+  latest_edition = @speech.latest_edition
```
bevanloon: perhaps just a small rename here to be clear that it is a `draft_edition`. Or even cutting it out completely? We have a unit test on Edition that sort of does something similar: https://github.com/alphagov/whitehall/blob/master/test/unit/edition_test.rb#L43
Maybe that's enough of a test for this kind of scenario?

test/integration/document_locale_param_canonicalisation_test.rb
steventux: @fofr good point :+1:, the test will error because the `SpeechController` has been removed as part of this PR and line 25 `get extra_path` var attempts to call the show action on the controller. So perhaps it's a case of asserting what the correct path is rather than requesting it. Will fix.

PR #3110, opened 2017-03-09
Present edition links in `DocumentCollectionPresenter#content`

gpeng:
Publishing API now supports links within the `put_content` call.

This commit adds links to the payload sent for `DocumentCollection`. Once all of the `DocumentCollection` documents have been republished this will enable us to switch over preview to government-frontend as the draft content store will have the correct links.

This is the initial step in the process. We will need to republish in a separate step and then switch over `DocumentCollection#rendering_app` to return `government-frontend` as this controls which app is used for preview.

We will also be able to remove the [`patch_links` step ](https://github.com/alphagov/whitehall/blob/send-draft-links/app/workers/publishing_api_worker.rb#L28) from the worker once this work is complete for all formats.

[Trello](https://trello.com/c/s7LjBWav/634-use-draft-links-for-documentcollections)

Requires https://github.com/alphagov/govuk-content-schemas/pull/567 to be deployed before the tests will pass.

### DIFF
--- a/app/presenters/publishing_api/document_collection_presenter.rb
+++ b/app/presenters/publishing_api/document_collection_presenter.rb
@@ -33,7 +33,7 @@ module PublishingApi
       links = LinksPresenter.new(item).extract(
         %i(organisations policy_areas topics related_policies parent)
       )
-      links.merge!(documents: item.documents.map(&:content_id))
+      links.merge!(documents: item.documents.map(&:content_id).uniq)
       links.merge!(PayloadBuilder::TopicalEvents.for(item))
     end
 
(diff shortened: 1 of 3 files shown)


app/presenters/publishing_api/document_collection_presenter.rb
```diff
@@ -33,7 +33,7 @@ def links
       links = LinksPresenter.new(item).extract(
         %i(organisations policy_areas topics related_policies parent)
       )
-      links.merge!(documents: item.documents.map(&:content_id))
+      links.merge!(documents: item.documents.map(&:content_id).uniq)
```
boffbowsh: Any reason not to use `pluck` here?

steventux: It looks like the unit test for this stubs documents and returns an array of OpenStructs which isn't to say it can't be refactored to work with `pluck`. I'll do this in a subsequent PR.

PR #3113, opened 2017-03-10
Reorder publishing api calls

gpeng:
The `PublishingApiWorker` was calling `publish` before `patch_links` which meant that the item put onto the rabbit queue (on publish) was being sent before the links were updated.

This change in order should fix that.

/cc @jennyd @kevindew


jennyd:
Sounds reasonable, thanks! @dwhenry will test it next week in Integration to check that it resolves the issue, but AFAIK this should be fine to go to Production before that if necessary since it's such a small change.

PR #3116, opened 2017-03-13
[DO NOT MERGE] Enable tagging to draft taxons

rubenarakelyan:
This commit splits the taxons on the taxonomy tagging page into live and draft taxons to make it clear to publishers which taxons will not appear on the live site.

Depends on https://github.com/alphagov/publishing-api/pull/833.

Trello: https://trello.com/c/LmVHv7Jv/537-build-design-for-tagging-to-draft-taxons-in-whitehall

Paired with @Davidslv 

screen shot 2017-03-13 at 11 42 10

### DIFF
--- a/app/models/edition_taxonomy_tag_form.rb
+++ b/app/models/edition_taxonomy_tag_form.rb
@@ -37,7 +37,15 @@ class EditionTaxonomyTagForm
   end
 
   def education_taxons
-    Taxonomy.education
+    @education_taxons ||= Taxonomy.education
+  end
+
+  def live_education_taxons
+    education_taxons.children.reject(&:draft)
+  end
+
+  def draft_education_taxons
+    education_taxons.children.select(&:draft)
   end
 
   # Ignore any taxons that already have a more specific taxon selected
--- a/app/views/admin/edition_tags/edit.html.erb
+++ b/app/views/admin/edition_tags/edit.html.erb
@@ -25,13 +25,27 @@
             <%= @edition_tag_form.education_taxons.name %>
           </p>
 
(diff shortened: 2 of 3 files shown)


app/models/edition_taxonomy_tag_form.rb
```diff
@@ -37,7 +37,15 @@ def publish!
   end
 
   def education_taxons
-    Taxonomy.education
+    @education_taxons ||= Taxonomy.education
+  end
+
+  def live_education_taxons
+    education_taxons.children.select { |taxon| !taxon.draft }
```
tijmenb: `reject(&:draft)` would work as well.

app/models/edition_taxonomy_tag_form.rb
```diff
@@ -37,7 +37,15 @@ def publish!
...
+  end
+
+  def draft_education_taxons
+    education_taxons.children.select { |taxon| taxon.draft }
```
tijmenb: You could also use `select(&:draft)` here

app/models/edition_taxonomy_tag_form.rb
rubenarakelyan: Fixed

app/models/edition_taxonomy_tag_form.rb
rubenarakelyan: Fixed, the linter caught that too!

Davidslv:
@rubenarakelyan It should be good to open after #3128 is merged.

Davidslv:
Superseeded by https://github.com/alphagov/whitehall/pull/3128

PR #3120, opened 2017-03-14
Remove the tax year (2015-2016) from the slug

mgrassotti:
https://trello.com/c/rutKfldX/638-remove-year-reference-from-slug-for-p55-tax-form


mgrassotti:
👍 done and tested on integration

PR #3122, opened 2017-03-14
Use content store for policy links instead of linkables

kevindew:
This won't build green until https://github.com/alphagov/govuk-content-schemas/pull/563 is merged in.

This is a system that replaces Whitehall using the Publishing API's linkables endpoint to render frontend content. It works by publishing the featured policies as links to the organisations placeholder content item). Then when rendering a page it'll access the content item.

For a painless transition this falls back to linkables if there aren't links in the content item. Once we have republished everything we could remove the fall back.

### DIFF
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -33,7 +33,7 @@ class OrganisationsController < PublicFacingController
         if @organisation.live?
           @recently_updated = recently_updated_source.with_translations(I18n.locale).limit(3)
           @feature_list = OrganisationFeatureListPresenter.new(@organisation, view_context)
-          @policies = @organisation.featured_policies.order('ordering').limit(5)
+          @policies = featured_policies
           set_meta_description(@organisation.summary)
 
           expire_on_next_scheduled_publication(@organisation.scheduled_editions)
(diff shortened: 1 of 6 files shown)


app/controllers/organisations_controller.rb
```diff
@@ -1,6 +1,8 @@
 class OrganisationsController < PublicFacingController
   include CacheControlHelper
 
+  CONTENT_NOT_FOUND = Class.new
```
kevindew: It's used as a means to memoize a value that might not exist: https://github.com/alphagov/whitehall/blob/c3c2aa5ddce1521ec48c24a531522f436c043ad2/app/controllers/organisations_controller.rb#L141-L148 as it returning nil won't allow memoize. We had similar confusion when we used this pattern on Pub API too so do wonder if there is a more idiomatic ruby pattern we can use 🤔 - any different approaches to solving this issue in Whitehall that are more consistent?

kevindew:
Thanks for the review @steventux

app/controllers/organisations_controller.rb
steventux: Ah I didn't think about the memoization of the rescue block return value. Makes more sense now.
Perhaps for the sake of clarity a custom class with a descriptive name could be used?
`MissingContentItem`, `NonExistentContentItem` or  something similar (admittedly neither of these suggestions really stands out).

kevindew:
Thanks @bevanloon 

I've updated the approach to memoizing the organisation content, what do you guys think?

PR #3125, opened 2017-03-20
Add ministers to consultations and fatality notices

bevanloon:
The migrated formats for Consultations and Fatality Notices should both send `ministers` in the Links hash.

Since `ministers` was already being sent in News Article and had some nice work around it as a class, extract this into a payload builder and use it for News Article, Consultations and Fatality Notices.

Included in the work is a migration to republish only those Consultations and Fatality Notices that have ministers associated.

[Trello](https://trello.com/c/On535pWu)


db/data_migration/20170317143146_republish_fatality_notices_with_ministers.rb
```diff
@@ -0,0 +1,9 @@
+ids = [
+  42786,
+  54892,
+  309262
+]
+Document.where(id: ids).each do |document|
+  PublishingApiDocumentRepublishingWorker
+    .perform_async_in_queue("bulk_republishng", document.id)
```
gpeng: Typo... `bulk_republishng` -> `bulk_republishing`

bevanloon: https://github.com/alphagov/whitehall/pull/3131

PR #3128, opened 2017-03-20
Request draft taxons to publishing api

Davidslv:
In this commit we have hardcoded the content_id of two taxons that we know are currently in draft state.
The intention is to allow users to tag documents to those two taxons, for that, we need to make a separate request to Publishing API for each draft taxon.

History

We thought of making two separate requests to Publishing API, the first
would only have live taxons and the second would have live and draft
taxons. We would then identify the ones not in the first request and
mark them as draft. Although this wasn't ideal because it was looking
into the Education Taxonomy only, that means that any draft taxon that
is not a child of Education wouldn't be considered.

Once the functionality has been tested we can then move it into https://github.com/alphagov/govuk_taxonomy_helpers

Trello: https://trello.com/c/LmVHv7Jv/537-build-design-for-tagging-to-draft-taxons-in-whitehall
### Look & Feel

screen shot 2017-03-22 at 13 01 56
screen shot 2017-03-22 at 13 03 16

### DIFF
--- a/app/models/edition_taxonomy_tag_form.rb
+++ b/app/models/edition_taxonomy_tag_form.rb
@@ -40,9 +40,17 @@ class EditionTaxonomyTagForm
     Taxonomy.education
   end
 
+  def draft_taxons
+    Taxonomy.drafts
+  end
+
+  def all_taxons
+    education_taxons.tree + draft_taxons.flat_map(&:tree)
+  end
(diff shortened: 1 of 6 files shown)


Davidslv:
This superseeds https://github.com/alphagov/whitehall/pull/3116

PR #3129, opened 2017-03-21
Add links for email to migrated formats

bevanloon:
This will not build until https://github.com/alphagov/govuk-content-schemas/pull/573 is merged.

It also relies on https://github.com/alphagov/whitehall/pull/3125 which includes Ministers in Consultation and Fatality Notice.
 
Adds people and roles to Links hash in `presenters/publishing_api`:

* Consultation
* Fatality Notice
* News Article
* Publication
* Speech

In order to send emails, Email Alert Api relies on the presence of links in data that the Publishing Api sends to the rabbit queues. Whitehall in turn needs to present these links to the Publishing Api. Roles and People are currently required but missing. This will allow Email Alert Api to take over from Govuk Delivery and ultimately allow us to retire Govuk Delivery.

[Trello](https://trello.com/c/WizHAZBT)


app/presenters/publishing_api/payload_builder/people.rb
```diff
@@ -0,0 +1,44 @@
...
+        return {} unless people_present?
+
+        { key => people_from_role_appointments + person_from_role_appointment }
+      end
+
+      def people_present?
+        people_from_role_appointments.present? || person_from_role_appointment.present?
+      end
+
+      def people_from_role_appointments
```
gpeng: Could this be refactored to a single method something like:

```
(Array(ed.try(:role_appointments)) + Array(ed.try(:role_appointment)))
  .map(&:person)
  .map(&:content_id)
```

I'm not sure that's actually an improvement. Using `Array(item.try...` in each method might be though.

bevanloon: 👍  good idea

My take on that ends up like this:
2a6a5e3766e992f60b45aa147b5a048e431cde20
5ea077e49067e40d71d2f55c7565850c5fc74441

I think it cleans the code up a bit.
I've kept the idea of `people` and `role_appointments`, for example in two separate methods because I think it expresses what the code is doing fairly clearly, but could squash it into one?

gpeng: Cool :+1:

PR #3036, opened 2017-02-14
Pdf report

NeilvB:
[Trello card](https://trello.com/c/JIT8o1Tn/90-pdf-report-for-natalie-baron-com)

This task creates a CSV summarising counts of all PDFs that were published on GOV.UK, with their associated (live) organisation.

PR #3130, opened 2017-03-21
Improve performance of generating a pdf report

mybits:
- reduce number of queries on Organisation call for pdf report
- change AR queries to SQL
- use Struct instead of class
- Use AR scoping so we can use find_each method to fetch only 1000 records at each time

### DIFF
--- a/lib/pdf_attachment_reporter.rb
+++ b/lib/pdf_attachment_reporter.rb
@@ -31,13 +31,13 @@ class PDFAttachmentReporter
   def pdfs_by_organisation
     second_time_period_date = @last_time_period_days.days.ago.to_date
 
-    live_organisation_names = Organisation.where(govuk_status: 'live').map(&:name) << POLICY_GROUPS
+    live_organisation_names = Organisation.where(govuk_status: 'live').includes(:translations).map(&:name) << POLICY_GROUPS
 
     live_organisation_published_pdfs_total_counts_hash = Hash[live_organisation_names.map { |o| [o, 0] }]
     live_organisation_published_pdfs_since_first_period_counts_hash = Hash[live_organisation_names.map { |o| [o, 0] }]
     live_organisation_published_pdfs_since_second_period_counts_hash = Hash[live_organisation_names.map { |o| [o, 0] }]
 
-    unique_published_pdf_attachments.each do |attachment|
+    unique_published_pdf_attachments.find_each do |attachment|
       pdf_attachment_data = find_pdf_attachment_data(attachment)
 
       if pdf_attachment_data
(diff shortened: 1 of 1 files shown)


boffbowsh:
Can you add a progress bar as well please?

mybits:
Hi @boffbowsh I was trying to find example of using progress bar in Whitehall (as you mentioned), but I couldn't find anything. Can you point me to where exactly it is used? 
Also do we need progress bar to run in a console or in some more sophisticated way?

boffbowsh:
Here's one example: https://github.com/alphagov/whitehall/blob/8b88ecee31ccb0848e4c108b4bf0258b2a4bbe27/lib/sync_checker/sync_check.rb It just uses the https://github.com/jfelchner/ruby-progressbar gem.

lib/pdf_attachment_reporter.rb
```diff
@@ -31,13 +32,18 @@ def initialize(opts = {})
...
 
-    live_organisation_names = Organisation.where(govuk_status: 'live').map(&:name) << POLICY_GROUPS
+    live_organisation_names = Organisation.where(govuk_status: 'live').includes(:translations).map(&:name) << POLICY_GROUPS
 
     live_organisation_published_pdfs_total_counts_hash = Hash[live_organisation_names.map { |o| [o, 0] }]
     live_organisation_published_pdfs_since_first_period_counts_hash = Hash[live_organisation_names.map { |o| [o, 0] }]
     live_organisation_published_pdfs_since_second_period_counts_hash = Hash[live_organisation_names.map { |o| [o, 0] }]
 
-    unique_published_pdf_attachments.each do |attachment|
+    unique_pdf_attachments = unique_published_pdf_attachments
```
boffbowsh: I'm not sure what the purpose of this assignment is? The method only returns an `ActiveRecord::Relation` object, so no queries will run until a "kicker" method is called, such as `each` or `count`.

lib/pdf_attachment_reporter.rb
```diff
@@ -31,13 +32,18 @@ def initialize(opts = {})
...
+
+    progress_bar.total = unique_pdf_attachments.length
```
boffbowsh: This should be `count` instead of `length`. `length` is the equivalent of `.to_a.size`:

```
irb(main):001:0> Edition.published.length
  Edition Load (3744.4ms)  SELECT `editions`.* FROM `editions` WHERE (`editions`.`state` != 'deleted') AND `editions`.`state` = 'published'
^CIRB::Abort: abort then interrupt!
irb(main):002:0> Edition.published.count
   (95.2ms)  SELECT COUNT(*) FROM `editions` WHERE (`editions`.`state` != 'deleted') AND `editions`.`state` = 'published'
=> 184235
```

lib/pdf_attachment_reporter.rb
```diff
@@ -31,13 +32,18 @@ def initialize(opts = {})
...
+    progress_bar.start
+
+    unique_pdf_attachments.each do |attachment|
```
boffbowsh: Using `find_each` will fetch only 1000 records at a time, keeping the memory usage down

boffbowsh:
Added some comments. You might want to review the comments I left on https://github.com/alphagov/whitehall/pull/3036/

chriscarter90:
Hi @boffbowsh - I've addressed your comments and double checked that everything on #3036 has been too. Hopefully it should be good to go. 👍

PR #3134, opened 2017-03-23
Republish organisation about pages when dependees change

andrewgarner:
The contents of an Organisation's about Corporate Information Page can change for the following reasons:

- the Organisation changes it's chart URL
- the Organisation publishes a Corporate Information Page
- the Organisation publishes a Corporate Report Publication
- the Organisation publishes a Freedom of Information Release Publication
- the Organisation publishes a Transparency Data Publication

These changes ensure that the Organisation's about page is republished in any such case.

This behaviour currently exists at render [here](https://github.com/alphagov/whitehall/blob/master/app/views/corporate_information_pages/index.html.erb#L51-L53) and [here](https://github.com/alphagov/whitehall/blob/master/app/views/organisations/_corporate_information.html.erb). With the Publishing API the equivalent of a render is presenting the about page during republishing.

[See Trello](https://trello.com/c/ND2S0vBm)

### DIFF
--- a/app/services/service_listeners/publishing_api_pusher.rb
+++ b/app/services/service_listeners/publishing_api_pusher.rb
@@ -35,13 +35,28 @@ module ServiceListeners
         api.discard_draft_async(edition)
       end
 
+      handle_corporate_information_pages(event)
       handle_html_attachments(event)
+      handle_publications(event)
     end
 
   private
 
+    def handle_corporate_information_pages(event)
+      return unless edition.is_a?(CorporateInformationPage)
+
+      PublishingApiCorporateInformationPagesWorker
+        .perform_async(edition.id, event)
+    end
+
     def handle_html_attachments(event)
       PublishingApiHtmlAttachmentsWorker.perform_async(edition.id, event)
+    end
+
+    def handle_publications(event)
+      return unless edition.is_a?(Publication)
+
+      PublishingApiPublicationsWorker.perform_async(edition.id, event)
     end
 
     def handle_translations
--- a/app/workers/publishing_api_corporate_information_pages_worker.rb
+++ b/app/workers/publishing_api_corporate_information_pages_worker.rb
@@ -1,0 +1,39 @@
+class PublishingApiCorporateInformationPagesWorker
+  extend Forwardable
+  include Sidekiq::Worker
+
+  def perform(edition_id, event)
+    self.corporate_information_page =
+      CorporateInformationPage.unscoped.find(edition_id)
+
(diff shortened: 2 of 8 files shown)


app/services/service_listeners/publishing_api_pusher.rb
```diff
@@ -35,15 +35,30 @@ def push(event:, options: {})
...
       end
 
+      handle_corporate_information_pages(event)
       handle_html_attachments(event)
+      handle_publications(event)
     end
 
   private
 
+    def handle_corporate_information_pages(event)
```
gpeng: Would it be more explicit to call this (and the publication one) `republish_` rather than `handle_` as that is they do?

app/workers/publishing_api_corporate_information_pages_worker.rb
```diff
@@ -0,0 +1,39 @@
+class PublishingApiCorporateInformationPagesWorker
+  extend Forwardable
+  include Sidekiq::Worker
+
+  def perform(edition_id, event)
```
gpeng: Do we need to pass the event in here? If all events end up aliased to `publish` it just seems to add complexity to 'handle' all of the events.

app/workers/publishing_api_publications_worker.rb
```diff
@@ -0,0 +1,50 @@
+class PublishingApiPublicationsWorker
+  extend Forwardable
+  include Sidekiq::Worker
+
+  def perform(edition_id, event)
```
gpeng: And here 🤣

app/services/service_listeners/publishing_api_pusher.rb
andrewgarner: It doesn't republish the corporate information pages or publications, it republished the about pages of the organisation's related to the corporate information pages or publications. Since that behaviour is contained within those two workers all we know at this stage is that we need to invoke the worker.

app/workers/publishing_api_corporate_information_pages_worker.rb
andrewgarner: There are events that we don't care about such as `schedule`, `unschedule`, `force_schedule`.

app/services/service_listeners/publishing_api_pusher.rb
gpeng: :+1:

app/workers/publishing_api_corporate_information_pages_worker.rb
gpeng: :+1:

PR #3138, opened 2017-03-24
Make tagging documents to taxonomy mandatory

Davidslv:
We want to enforce users to tag documents to a topic (taxon) in the
taxonomy.
This commit shows an error when publishing or force publishing education documents without
adding tags.

screen shot 2017-03-24 at 12 15 51
screen shot 2017-03-24 at 12 16 02

screen shot 2017-03-28 at 12 09 07

Trello:
https://trello.com/c/dBDSxqtJ/497-make-tagging-to-taxons-mandatory-with-special-cases-for-parenting-and-corporate-information


lib/whitehall.rb
```diff
@@ -162,6 +162,10 @@ def self.organisations_in_tagging_beta
       YAML.load_file(Rails.root + "config/organisations_in_tagging_beta.yml")["organisations_in_tagging_beta"]
   end
 
+  def self.tagging_taxonomy_enabled?
```
tijmenb: Why the feature flag?

app/models/edition.rb
```diff
@@ -424,6 +425,22 @@ def can_be_tagged_to_taxonomy?
     false
   end
 
+  def expanded_links
```
tijmenb: Would it be possible to make a single call to `publishing_api.get_links()` to see if the item has been tagged? Then we don't need the gem & the `LinkedContentItem` setup.

app/validators/taxon_validator.rb
```diff
@@ -0,0 +1,18 @@
+class TaxonValidator < ActiveModel::Validator
+  def validate(edition)
+    if missing_taxons?(edition)
+      edition.errors.add(
+        :base,
+        "The document needs to be in one or more topics in the new taxonomy before publishing."
```
tijmenb: Could we link to the tagging page from here?

lib/whitehall.rb
Davidslv: We just moved it inside whitehall module

app/validators/taxon_validator.rb
Davidslv: I don't think that's the responsibility of the validator.

lib/whitehall.rb
tijmenb: It wasn't moved, it was duplicated. I also don't understand why we're still using a feature flag here, because it should be visible on all environments.

Davidslv: Hum... you are right, we didn't removed it from the base_controller, I had the impression we had done that when we added the method here. I don't think we have a good reason to keep this anyway, will remove it and then I'll open a PR on govuk-puppet to remove the env variable.

tijmenb: 👍

Davidslv:
@tijmenb I think I've addressed your comments.
There's a failing test which I believe comes from the change here https://github.com/alphagov/govuk-content-schemas/pull/569

Let me know your toughts

app/controllers/admin/edition_workflow_controller.rb
```diff
@@ -71,9 +71,16 @@ def publish
   end
 
   def confirm_force_publish
+    unless @edition.valid?(:publish)
```
tijmenb: Can you add a small test for this flow? Might be easy to break.

Davidslv:
@tijmenb this should now be good for a second review and possible merge?

Davidslv:
Made a change in the validator text.

app/controllers/admin/edition_workflow_controller.rb
Davidslv: Done 👍

PR #3144, opened 2017-03-28
Render Coporate Information Pages with Government Frontend

Rosa-Fox:


test/unit/helpers/public_document_routes_helper_test.rb
```diff
@@ -122,7 +122,7 @@ class PublicDocumentRoutesHelperTest < ActionView::TestCase
   test "Creates a preview URL with cachebust and edition parameters" do
     edition = create(:corporate_information_page)
     preview_url = preview_document_url(edition)
-    assert_equal "http://test.host/government/organisations/#{edition.organisation.slug}/about/publication-scheme?cachebust=#{Time.zone.now.getutc.to_i}&preview=#{edition.id}", preview_url
+    assert_equal "http://draft-origin.test.alphagov.co.uk/government/organisations/#{edition.organisation.slug}/about/publication-scheme", preview_url
```
steventux: Is this change necessary because of other test.alphagov.co.uk config or just to stick with convention?
Seems odd to be changing this to a domain we don't use.

gpeng: It's the URL that gets generated for preview as a result of switching the rendering app to government-frontend. It's a combination of https://github.com/alphagov/whitehall/blob/master/app/helpers/public_document_routes_helper.rb#L48-L50 and https://github.com/alphagov/whitehall/blob/master/app/helpers/public_document_routes_helper.rb#L85-L98

We can probably get rid of some of that `document_url` stuff now, house of cards notwithstanding. Don't need to attempt it in this PR though 😱

PR #3164, opened 2017-04-07
Send emails via SMTP by defining SMTP_PORT

rubenarakelyan:
This commit changes the whitehall config to send emails using SMTP in development by default, and in production when the `SMTP_PORT` env var is defined. After https://github.com/alphagov/govuk-puppet/pull/5745 is merged, this will allow sent emails to be viewed locally in MailHog.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -16,6 +16,12 @@ Whitehall::Application.configure do
   config.action_controller.perform_caching = false
   config.action_controller.action_on_unpermitted_parameters = :raise
 
+  # Send emails to the local MailHog instance
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = {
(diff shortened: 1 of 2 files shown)


tijmenb:
Is there any risk of actually sending email with this system? Would it make sense to run the training system in a different `Rails.env` to make sure it won't?

Also perhaps good to link to somewhere that explains how to access mailhog.

rubenarakelyan:
We don't have any SMTP server in the development VM and the SES credentials aren't there either so I don't think there's a risk of actual emails being sent there. We will be running the training environment with `RAILS_ENV=production`, so we'll set `SMTP_PORT=1025` there, and any developers running the VM will get this by default.

I have updated the development VM README in https://github.com/alphagov/govuk-puppet/pull/5745 to explain how to access MailHog.

PR #3166, opened 2017-04-11
[WIP] Integration with Link checker api

kevindew:
[Trello Card](https://trello.com/c/RkRvwP9k/969-complete-whitehall-integration)


app/controllers/admin/link_checker_api_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      report.update_from_batch_report(params) if report
+    end
+
+    head :no_content
+  end
+
+private
+
+  def verify_webhook
+    # TODO
```
thomasleese: Hehe

app/models/link_checker_api_report/link.rb
```diff
@@ -0,0 +1,16 @@
+class LinkCheckerApiReport::Link < ActiveRecord::Base
+  serialize :check_errors, Hash
+  serialize :check_warnings, Hash
+
+  belongs_to :report, class_name: LinkCheckerApiReport
+
+  def self.attributes_from_link_report(payload)
```
thomasleese: Can we use the `LinkReport` model from GDS API Adapters here? Or in similar places?

db/schema.rb
```diff
@@ -83,7 +83,6 @@
   add_index "attachments", ["attachable_id", "attachable_type"], name: "index_attachments_on_attachable_id_and_attachable_type", using: :btree
   add_index "attachments", ["attachable_type", "attachable_id", "ordering"], name: "no_duplicate_attachment_orderings", unique: true, using: :btree
   add_index "attachments", ["attachment_data_id"], name: "index_attachments_on_attachment_data_id", using: :btree
-  add_index "attachments", ["content_id"], name: "index_attachments_on_content_id", using: :btree
```
thomasleese: Wonder why this happened here 🤔

test/functional/admin/link_check_reports_controller_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    login_as :writer
+    @publication = create(:publication, body: "[link](http://www.example.com)")
+    Rails.backtrace_cleaner.remove_silencers!
+
+    link_checker_endpoint = "#{Plek.find('link-checker-api')}/batch"
+    body = link_checker_api_batch_report_hash(
+      id: 5,
+      links: [{ uri: "http://www.example.com" }],
+    )
+    stub_request(:post, %r{\A#{link_checker_endpoint}})
```
thomasleese: There are some helper methods which can stub the whole request, have you tried those?

db/schema.rb
kevindew: oh yes that's not good - good spot

app/models/link_checker_api_report/link.rb
kevindew: This uses the output of the webhook as well so it's designed to be compatible with both a strong params object and the output from GDS API - not sure if that might bite at some point though.

test/functional/admin/link_check_reports_controller_test.rb
kevindew: Whitehall uses a different hostname (alphagov.co.uk) than is stubbed by GDS API adapters (dev.gov.uk) which is super annoying.

app/models/link_checker_api_report/link.rb
thomasleese: I was thinking actually turning the input from the web hook into a `BatchReport` object and then just using that -- it's what we use in LLM.

kevindew:
I'll close this for now since it's not being actively worked on.

thomasleese:
Will be working on this during the fire break.

rubenarakelyan:
Closing for the moment since it's not being actively worked on.

PR #3175, opened 2017-04-20
Fix GOVUK_APP_DOMAIN

gpeng:
This is defined in a few places and was using `alphagov.co.uk`. This was causing inconsistencies depending on how tests were run. It is redefined in the Rakefile so tests run with rake were picking up different values from Plek than tests run directly. This resulted in tests that would fail when run individually in the VM but pass on CI and vice versa if the test was updated to pass locally.

This removes some of the definitions and updates the default value to `dev.gov.uk`. It also updates tests and factories to use the new URLs.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,8 +6,6 @@
 # to log level :warn reduces logging and increases execution speed.
 ENV["LOG_LEVEL"] = "warn"
 
-ENV['GOVUK_APP_DOMAIN'] = 'dev.gov.uk' unless ENV['GOVUK_APP_DOMAIN']
-
 require File.expand_path('../config/application', __FILE__)
 require 'ci/reporter/rake/minitest' if Rails.env.development? or Rails.env.test?
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,7 +82,7 @@ node {
       if (params.IS_SCHEMA_TEST) {
         echo "Skipping precompile step because this is a schema test"
       } else {
-        sh("RAILS_ENV=production GOVUK_ASSET_ROOT=http://static.test.alphagov.co.uk bundle exec rake assets:precompile --trace")
+        sh("RAILS_ENV=production GOVUK_ASSET_ROOT=https://static.test.gov.uk bundle exec rake assets:precompile --trace")
       }
     }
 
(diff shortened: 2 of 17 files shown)


rubenarakelyan:
We seem to use `test.gov.uk` rather than `dev.gov.uk` for tests (e.g. https://github.com/alphagov/content-tagger/blob/master/spec/rails_helper.rb#L9)

Rakefile
```diff
@@ -6,7 +6,7 @@
 # to log level :warn reduces logging and increases execution speed.
 ENV["LOG_LEVEL"] = "warn"
 
-ENV['GOVUK_APP_DOMAIN'] = 'dev.gov.uk' unless ENV['GOVUK_APP_DOMAIN']
+ENV['GOVUK_APP_DOMAIN'] = 'test.gov.uk' unless ENV['GOVUK_APP_DOMAIN']
```
gpeng: @rubenarakelyan I've updated this to use `test.gov.uk` throughout. To get that to work I've had to make this change. I'm worried this might have unwanted side effects as this will be set for all `rake` tasks. Should probably be OK. What do you think? I'll squash this into the original commit before we merge if you're happy with it.

rubenarakelyan: Could set it to `test.gov.uk` using `if Rails.env.test?` maybe?

rubenarakelyan:
@gpeng Could this be an issue when running stuff in the dev VM that uses this env var? Could set it to `test.gov.uk` `if Rails.env.test?` maybe.

Rakefile
gpeng: Fixed in ef5129f

rubenarakelyan:
Looks like `Rails` isn't defined at that point :(

Is there any issue with having the Rakefile pointing at `dev.gov.uk` while all the other test stuff is `test.gov.uk`?

gpeng:
Yes. The issue is that tests run via Rake will have different URLs generated than tests run individually so CI (running Rake) and developers running subsets using `mtest` get different test results. We might not need to define it at all in there. Maybe it's better to just define it in the `/envronment/(test|developement).rb` files.

rubenarakelyan:
That makes sense to me

PR #3179, opened 2017-04-24
Use featured image for speeches

gpeng:
Currently the `Speech` format always uses the speaker's 'profile' image in the left side bar. When the speech has been featured with an image the expectation is that the image will be used on the speech page too.

This commit adds support for this functionality effectively allowing the publisher to add a more relevant image for featured content.

[Trello](https://trello.com/c/zhiEl78z/57-enable-custom-images-in-speeches-apr)


fofr:
Awesome 💯

PR #3183, opened 2017-04-27
Create global taxonomy root

whoojemaflip:
This enables us to have multiple Topics without hard-coding the `content_id` of each one.

It uses the `expanded_links['child_taxons']` array of the Homepage as the root of the gov.uk taxonomy tree.

Before deploying, ensure that the  `education, training and skills` taxon has been promoted to a first-class taxonomy topic via the rake task in `content-tagger`

```
$ rake taxonomy:promote_taxon_to_taxonomy_root[c58fdadd-7743-46d6-9629-90bb3ccc4ef0]
```

- [ ] tagged in production
- [ ] tagged in staging
- [ ] tagged in integration

[Trello](https://trello.com/c/wHZQz5Yk/64-build-joe-s-designs-to-allow-publishers-to-tag-across-multiple-themes-in-whitehall)

Paired with @klssmith


lib/tasks/patch_taxonomy_root.rake
```diff
@@ -0,0 +1,10 @@
+desc "One time task to setup the Homepage as taxonomic root"
```
tijmenb: Should we do this in content tagger?

whoojemaflip: Happily

lib/taxonomy/publishing_api_root_taxon_parser.rb
```diff
@@ -0,0 +1,27 @@
+module Taxonomy
+  module PublishingApiRootTaxonParser
```
tijmenb: Perhaps this is nicer as a class to avoid the `private_class_method` thing (but I don't feel very strongly about this).

lib/taxonomy/taxon.rb
```diff
@@ -0,0 +1,67 @@
...
+      @content_id = content_id
+      @base_path = base_path
+      @children = []
+    end
+
+    def children
+      @children.sort_by(&:name)
+    end
+
+    def <<(child_node)
```
tijmenb: Can we give this a name? `append_child_taxon` or something like that?

test/unit/taxonomy/publishing_api_root_taxon_parser_test.rb
```diff
@@ -0,0 +1,105 @@
...
+  #       }
+  #     ]
+  #   }
+  # }
+
+  test ".parse_taxons returns an empty array when there are no child taxons" do
+    assert parsed_result({ "expanded_links" => {} }).empty?
+  end
+
+  test ".parse_taxons parses a single child_taxon" do
```
tijmenb: The tests here might read easier if the test values are inlined, so you have all the information you need inside the `test` block:

```ruby
test ".parse_taxons parses a single child_taxon" do
   single_root_node = {
    "expanded_links" => {
      "child_taxons" => [
        node
      ]
    }
  }

  result = parsed_result(single_root_node)
   
  assert is_an_array_of_taxons(result)
  assert result.length == 1
end
```

lib/taxonomy/publishing_api_root_taxon_parser.rb
whoojemaflip: Same. The reason we didn't make it a class is to avoid it being instantiated. For now. With more refactoring on the way, I'd be surprised if this sticks around in it's current form for very long.

lib/tasks/patch_taxonomy_root.rake
whoojemaflip: https://github.com/alphagov/content-tagger/pull/348

test/unit/taxonomy/publishing_api_root_taxon_parser_test.rb
whoojemaflip: Done

lib/taxonomy/publishing_api_linked_edition_parser.rb
```diff
@@ -0,0 +1,38 @@
+module Taxonomy
+  class PublishingApiLinkedEditionParser
+    attr_accessor :linked_edition
+
+    def initialize(edition_response)
+      @linked_edition = Taxon.new(edition_response.to_h.symbolize_keys)
+      @name_field = name_field
+    end
+
+    def add_expanded_links(expanded_links_response)
```
andrewgarner: There is some duplication between `add_expanded_links` and `parse_nested_item` we could refactor maybe? Possibly by adding some recursion?

- Both look for a `child_taxons` key, but from either `expanded_links_response["expanded_links"]` or `nested_item["links"]`
- Both add a to the `children` collection, but either `linked_edition` or `nested_linked_edition`.

lib/taxonomy/taxon.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def initialize(title:, base_path:, content_id:, **_args)
+      @name = title
+      @content_id = content_id
+      @base_path = base_path
+      @children = []
+    end
+
+    def tree
+      return [self] if @children.empty?
```
andrewgarner: We have a `attr_accessor` for `@children` so this should just be `children`

andrewgarner: `@children` will always be empty?

lib/taxonomy/publishing_api_linked_edition_parser.rb
whoojemaflip: I plan on deleting this class once I get the chance. It's copy-pasta that's only here because handling Draft taxonomies seemed outside of the scope of this PR.

I'd rather not touch it beyond extracting the class into it's own file, since it's not tested and I hope it will go away very soon.

lib/taxonomy/taxon.rb
whoojemaflip: `children` is set by the parser in `Taxonomy::Tree`.

whoojemaflip: FWIW this class is also copy-pasta, and I plan on improving it later. Didn't want to rewrite the whole world in this first PR.

lib/taxonomy/taxon.rb
```diff
@@ -13,9 +13,7 @@ def initialize(title:, base_path:, content_id:, **_args)
     end
 
     def tree
-      return [self] if @children.empty?
-
-      @children.each_with_object([self]) do |child, tree|
+      children.each_with_object([self]) do |child, tree|
```
andrewgarner: If `@children` is always empty why don't you just return `[self]`?

whoojemaflip: `@children` isn't always empty!

PR #3186, opened 2017-04-28
Add SVG support

gpeng:
We are going to recommend to publishers that they use the SVG file format for graphs, charts etc but currently Whitehall doesn't support the format.

This adds support in the publishing app and for formats still rendered by Whitehall.

[Trello](https://trello.com/c/xgjd9XDu/64-investigate-adding-svg-support-to-attachments-feb)

### DIFF
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -6,25 +6,30 @@ class ImageUploader < WhitehallUploader
   end
 
   def extension_white_list
-    %w(jpg jpeg gif png)
+    %w(jpg jpeg gif png svg)
   end
 
-  version :s960 do
+  version :s960, if: :bitmap? do
     process resize_to_fill: [960, 640]
   end
-  version :s712, from_version: :s960 do
+  version :s712, from_version: :s960, if: :bitmap? do
     process resize_to_fill: [712, 480]
   end
-  version :s630, from_version: :s712 do
+  version :s630, from_version: :s712, if: :bitmap? do
     process resize_to_fill: [630, 420]
   end
-  version :s465, from_version: :s630 do
+  version :s465, from_version: :s630, if: :bitmap? do
     process resize_to_fill: [465, 310]
   end
-  version :s300, from_version: :s465 do
+  version :s300, from_version: :s465, if: :bitmap? do
     process resize_to_fill: [300, 195]
   end
-  version :s216, from_version: :s300 do
+  version :s216, from_version: :s300, if: :bitmap? do
     process resize_to_fill: [216, 140]
+  end
+
+  def bitmap?(new_file)
+    return if new_file.nil?
+    !(new_file.content_type =~ /svg/)
   end
 end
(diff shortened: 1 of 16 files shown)


test/unit/image_uploader_test.rb
```diff
@@ -22,17 +22,27 @@ class ImageUploaderTest < ActiveSupport::TestCase
...
     [[712, 480], [630, 420], [465, 310], [300, 195], [216, 140]].each do |(width, height)|
       assert_image_size [width, height], @uploader.send(:"s#{width}")
     end
   end
 
+  test "stores the original svg only" do
+    model = stub("AR Model", id: 1)
+    @uploader = ImageUploader.new(model, "mounted-as")
+
+    @uploader.store!(fixture_file_upload('minister-of-funk.960x640.jpg', ''))
```
davidbasalla: Should this be an svg file rather than a jpg?

test/functional/admin/organisations_controller_test.rb
```diff
@@ -83,7 +83,7 @@ def example_organisation_attributes
   test 'POST :create can set a custom logo' do
     post :create, organisation: example_organisation_attributes.merge(
       organisation_logo_type_id: OrganisationLogoType::CustomLogo.id,
-      logo: fixture_file_upload('logo.png')
+      logo: fixture_file_upload('logo.png', 'image/png')
```
davidbasalla: Why did you have to add the type here as a second param?

gpeng: If you don't provide it `content_type` returns `nil` and I needed to check it to decide whether to do the thumbnails or not.

test/unit/image_uploader_test.rb
gpeng: Whoops... it should and the assertion needs changing too. Well spotted :+1:

app/uploaders/image_uploader.rb
```diff
@@ -6,25 +6,30 @@ class ImageUploader < WhitehallUploader
...
-  version :s300, from_version: :s465 do
+  version :s300, from_version: :s465, if: :image? do
     process resize_to_fill: [300, 195]
   end
-  version :s216, from_version: :s300 do
+  version :s216, from_version: :s300, if: :image? do
     process resize_to_fill: [216, 140]
   end
+
+  def image?(new_file)
```
davidbasalla: Wondering about naming here... could be a bit confusing if we're saying that an SVG is not an image (and it that case it shouldn't be in the ImageUploader)? Could we instead ask something like `has_thumbnail?` and base it on a thumbnail whitelist (I guess basically the white list but excluding 'svg')? Would that get around the necessity of passing the second argument to the test helpers in the other test files? Haven't done anything with Carrierwave before though so not sure if this helpful...

test/unit/image_uploader_test.rb
gpeng: Fixed in bf235c6

app/uploaders/image_uploader.rb
gpeng: Fixed in 5fe89ab I went for `bitmap?` which is what we are effectively checking.

davidbasalla: ah nice 👍 do we still need the line `return if new_file.nil?` now that the `fixture_file_upload` calls in the tests have been updated with the content type?

gpeng: Yes.. I think so. It looks like it gets called quite a bit throughout the tests with nil image. Not sure it would happen outside of testing but it seems like a safer bet to leave it in there.

PR #3189, opened 2017-05-02
Update sync check docs

gpeng:
Adds documentation for the `-b` range/between option and updates the auto sync check docs to note that they are currently disabled.

### DIFF
--- a/docs/migration_sync_checks.md
+++ b/docs/migration_sync_checks.md
@@ -52,7 +52,8 @@ while to complete for the larger formats.
 
 ## Automatically queued checks
 
-All content that is sent to publishing API now queues a check to run 5 minutes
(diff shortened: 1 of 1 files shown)


docs/migration_sync_checks.md
```diff
@@ -48,7 +52,8 @@ while to complete for the larger formats.
 
 ## Automatically queued checks
 
-All content that is sent to publishing API now queues a check to run 5 minutes
-after the changes are saved. The results of these checks are saved in the
-`sync_check_results` table and will be used in the future for ongoing
-monitoring.
+There is functionality to automatically queues a check to run 5 minutes
```
davidbasalla: `queues` should be `queue`?

gpeng: Fixed in 445d582

PR #3192, opened 2017-05-04
Replace observers with concerns

rubenarakelyan:
`rails-observers` were removed in Rails 4 but carried on being used in whitehall via an extracted gem. This is no longer supported in Rails 5, so this commit replaces observers with native Rails concerns.

This is a precursor to the full upgrade to Rails 5.


test/unit/models/ministerial_role/minsterial_role_reindexing_concern_test.rb
```diff
@@ -19,17 +19,6 @@ class MinisterialRoleSearchIndexObserverTest < ActiveSupport::TestCase
     organisation.save
   end
 
-  test 'should not reindex all ministerial roles when an organisation is updated if observer is disabled' do
```
davidbasalla: Would this 'disabling of an observer' ever have happened in Production?

gpeng: I can't imagine why it would and there doesn't seem to be anywhere that it is happening from within the Whitehall code.

gpeng:
Actually... could we get rid of the `rails-observers` gem in this PR too?

test/unit/models/ministerial_role/minsterial_role_reindexing_concern_test.rb
rubenarakelyan: Yeah, the only place the disabling was happening was in the tests

rubenarakelyan:
Yup, missed that since I was doing it in a separate branch to upgrade to Rails 5

app/models/organisation/organisation_search_index_concern.rb
```diff
@@ -0,0 +1,25 @@
...
+private
+
+  def going_live_on_govuk?
+    govuk_status_changed? && govuk_status == 'live'
+  end
+
+  def leaving_live_on_govuk?
+    govuk_status_changed? && govuk_status_was == 'live'
+  end
+end
```
dwhenry: should concerns live in `models/concerns` so that we follow the `rails` pattern for their location.

rubenarakelyan: Technically yes, but we already had an existing concern in this location so I went with consistency in the project for this PR.

PR #3193, opened 2017-05-04
Add UI support for multiple taxonomies

whoojemaflip:
It is now possible to tag an edition to multiple taxonomies.

By default, the taxonomy trees are collapsed when the page is loaded.

screen shot 2017-05-04 at 17 58 39

However since we only have a single published taxonomy in production today, it will load the page with that taxonomy tree fully expanded by default. 

screen shot 2017-05-04 at 17 55 17

[Trello](https://trello.com/c/wHZQz5Yk/64-build-joe-s-designs-to-allow-publishers-to-tag-across-multiple-themes-in-whitehall)


nickcolley:
How accessible is this?

PR #3194, opened 2017-05-05
Allow news articles to be created as translations only

davidbasalla:
For https://trello.com/c/fZPPfJhf/86-allow-news-articles-to-be-created-with-only-non-english-content

Mirrors the cucumber test introduced in https://github.com/alphagov/whitehall/pull/576 --> https://github.com/alphagov/whitehall/blob/master/features/non-english-world-location-news.feature, but excludes the last line `And I should only be able to view the news article article in French` as news articles are rendered in government-frontend now. Also remove references to world organisations as these do not (yet) associate to news articles.

### DIFF
--- a/app/models/news_article.rb
+++ b/app/models/news_article.rb
@@ -52,6 +52,10 @@ class NewsArticle < Newsesque
     Whitehall::RenderingApp::GOVERNMENT_FRONTEND
   end
 
+  def locale_can_be_changed?
+    new_record?
+  end
+
   private
 
   def only_news_article_allowed_invalid_data_can_be_awaiting_type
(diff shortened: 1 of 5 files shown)


davidbasalla:
Hmm good point. Currently testing it on Integration and they don't appear in the Announcements, but I'm not quite sure _why_, so will keep digging... But also wondering whether that would be easier to implement once we've created the new NewsArticle type?

davidbasalla:
@gpeng I had a chat with @rubenarakelyan about this, and current thinking is that these sort of translation-only news articles will only be created in conjunction with the new (yet to be implemented) world news article sub type. We'll then be able to filter out these translated pages in the same way as we filter out worldwide news currently. We'll be communicating this change to the content team as well. So while currently they _would_ show up in the Announcements page, we don't think it's a big issue.

PR #3201, opened 2017-05-08
Pass strings rather than classes to ActiveRecord

rubenarakelyan:
Rails 5 deprecates the passing of classes to ActiveRecord. In preparation for that upgrade, this commit changes code to pass strings instead.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -57,6 +57,7 @@ class Document < ApplicationRecord
   end
 
   def self.at_slug(document_types, slug)
+    document_types = Array(document_types).map(&:to_s)
     where(document_type: document_types, slug: slug).first
   end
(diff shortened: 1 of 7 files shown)


app/models/document.rb
```diff
@@ -57,6 +57,8 @@ def self.published
   end
 
   def self.at_slug(document_types, slug)
+    document_types = [document_types] unless document_types.is_a?(Array)
```
davidbasalla: Could this be `Array(document_types)` instead?
```
Array(1)
=> [1]
Array([1,2,3])
=> [1, 2, 3]
```

rubenarakelyan: Great, I've used `to_a` to do this

PR #3209, opened 2017-05-09
Remove unused `NewsArticleType`

gpeng:
This removes a `NewsArticleType` that is no longer used. (`NewsArticleType::ImportedAwaitingType`). It also removes some related tests for `NewsArticle` in imported state which is also no longer used.

Loosely related to -> [Trello](https://trello.com/c/VygNbQmD/88-create-a-new-worldwide-news-article-subtype)

### DIFF
--- a/app/models/news_article.rb
+++ b/app/models/news_article.rb
@@ -6,7 +6,6 @@ class NewsArticle < Newsesque
   include Edition::CanApplyToLocalGovernmentThroughRelatedPolicies
 
   validates :news_article_type_id, presence: true
-  validate :only_news_article_allowed_invalid_data_can_be_awaiting_type
 
   def self.subtypes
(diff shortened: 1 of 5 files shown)


davidbasalla:
There's a validation for NewsArticles called [`:only_news_article_allowed_invalid_data_can_be_awaiting_type`](https://github.com/alphagov/whitehall/blob/master/app/models/news_article.rb#L9), do you reckon that can be removed/updated as well?

davidbasalla:
We could remove https://github.com/alphagov/whitehall/blob/ef457b634f1ce5bfb2ef4652bdefb17e492ce544/app/models/news_article_type.rb#L14 as well

gpeng:
@davidbasalla Well spotted... Thanks. I've removed that too in 9dd263b

PR #3217, opened 2017-05-11
Upgrade gems

rubenarakelyan:
This PR updates gems in preparation for the Rails 5 upgrade, and also changes the `friendly_id` initializer to use the built-in sequential slugger, which was re-added in version 5.2.0, rather than the whitehall monkey-patch.


rubenarakelyan:
@gpeng we have these: https://github.com/alphagov/whitehall/blob/master/test/unit/whitehall/slugging_test.rb#L10-L20

PR #3219, opened 2017-05-12
Upgrade admin frontend gems

rubenarakelyan:
This PR upgrades gems that are used in rendering admin pages. It also makes a minor consequential change to a `kaminari` test.


rubenarakelyan:
Yes good idea - @fofr or @nickcolley?

nickcolley:
We could split this into changes that only affect the admin side?

I agree with @gpeng we should be cautious

rubenarakelyan:
I've updated the PR to only change the admin UI gems. It still updates `jquery-rails` as a dependency but not `jquery-ui-rails` any more.

fofr:
Looks good to me. Can't see anywhere that the breaking changes in Kaminari or govuk_admin_template will cause problems.

fofr:
Looks like this broke some analytics tracking in Whitehall, I'm guessing because of events that are being tracked without a category:
 
screen shot 2018-01-23 at 07 57 50

PR #3225, opened 2017-05-15
Add worldwide organisations to news articles

davidbasalla:
For https://trello.com/c/428XJAua/87-allow-worldwide-news-articles-to-be-tagged-to-worldwide-organisations

Adds the worldwide organisation association to NewsArticles, specifically to the World News Story subtype. The fields get hidden in the UI by JavaScript. The PublishingApi presenters also get updated to send worldwide organisations as links. The sync checks also get updated.

Also adds the logic for _not_ associating Organisations to 'World news article' News articles. Again, JS is used to hide the whole Organisation form fieldset when selecting the 'World news story' sub type and model validations also get added.

---------------------

### Preview of new validation errors:

screen shot 2017-05-18 at 15 03 28

### DIFF
--- a/app/models/news_article.rb
+++ b/app/models/news_article.rb
@@ -10,6 +10,7 @@ class NewsArticle < Newsesque
   validates :news_article_type_id, presence: true
   validates :worldwide_organisations, absence: true, unless: :world_news_story?
   validate :non_english_primary_locale_only_for_world_news_story
+  validate :organisations_are_not_associated, if: :world_news_story?
   validate :policies_are_not_associated, unless: :can_be_related_to_policies?
 
   def self.subtypes
@@ -74,11 +75,21 @@ class NewsArticle < Newsesque
     !world_news_story?
   end
 
+  def skip_organisation_validation?
+    world_news_story?
+  end
+
   def can_be_related_to_policies?
     !world_news_story?
   end
 
 private
+
+  def organisations_are_not_associated
+    unless edition_organisations.empty?
+      errors.add(:base, "You can't tag a world news story to organisations, please remove organisation")
+    end
+  end
 
   def policies_are_not_associated
     unless edition_policies.empty?
--- a/app/models/news_article.rb
+++ b/app/models/news_article.rb
@@ -4,9 +4,11 @@ class NewsArticle < Newsesque
   include ::Attachable
   include Edition::AlternativeFormatProvider
   include Edition::CanApplyToLocalGovernmentThroughRelatedPolicies
+  include Edition::WorldwideOrganisations
 
   validate :ministers_are_not_associated, if: :world_news_story?
   validates :news_article_type_id, presence: true
+  validates :worldwide_organisations, absence: true, unless: :world_news_story?
   validate :non_english_primary_locale_only_for_world_news_story
   validate :policies_are_not_associated, unless: :can_be_related_to_policies?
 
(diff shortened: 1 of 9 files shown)


app/models/news_article.rb
```diff
@@ -4,9 +4,11 @@ class NewsArticle < Newsesque
   include ::Attachable
   include Edition::AlternativeFormatProvider
   include Edition::CanApplyToLocalGovernmentThroughRelatedPolicies
+  include Edition::WorldwideOrganisations
 
   validates :news_article_type_id, presence: true
   validate :non_english_primary_locale_only_for_world_news_story
+  validate :worldwide_organisation_association_only_for_world_news_story
```
gpeng: I think we can refactor this to clean it up quite a bit. If we had:
```
def world_news_story?
...etc
```
Then we can use the normal `validates:` with `presence: true` if that method ^ returns true and `absence: true` if that method returns false. The method on 67 would also just return `!!world_news_story?`.

test/unit/news_article_test.rb
```diff
@@ -82,4 +83,30 @@ class NewsArticleTest < ActiveSupport::TestCase
...
+    non_world_news_story_news_article_types = [
+      NewsArticleType::NewsStory,
+      NewsArticleType::PressRelease,
+      NewsArticleType::GovernmentResponse,
+    ]
+
+    non_world_news_story_news_article_types.each do |news_article_type|
+      news_article = build(:news_article, news_article_type: news_article_type)
+      news_article.worldwide_organisations.build(name: "Zimbabwean Embassy")
+      refute news_article.valid?
```
gpeng: I usually try to check that the errors are on the bit we're testing in these type of tests too. You can get false passes/failures if the object is invalid for a different reason than we are testing.

davidbasalla:
Thanks @gpeng, I've updated the PR. Could you have another look, especially at the sync checks since I haven't worked with them at all yet

gpeng:
I've realised there's a bit of a bug/usability issue with this. If you create a news story draft with an organisation and save it, then open it for editing again and change the type to 'world news story', when you save the draft again the validation tells you that 'Organisations must be blank' but the organisations section of the form is now hidden so the user can't remove them.

davidbasalla:
ah, I completely missed that the UI allows for changing type of existing News articles... thanks for pointing that out, will fix!

rubenarakelyan:
The JavaScript code that hides the fields should also clear them out. In the long-term, I don't see what the case is for changing a type after publishing, we should think about removing this ability.

gpeng:
I suspect it might happen while the item is in draft and gets saved and edited. We'll probably have to add an empty hidden element for the 'no longer valid' fields to get them to be removed. That should fix the problem and then it won't be a dev job in the future if publishers need to change the type for some reason.

davidbasalla:
As discussed, I've removed the JavaScript code for now, moved it to here -> https://github.com/alphagov/whitehall/pull/3235

davidbasalla:
Hmm the form still let's me submit a World news story, even with Organisations on it. In the show page I see the info that it's not actually valid, but the form shouldn't allow the submission in the first place. Need to investigate...

davidbasalla:
Made the required changes to make this work, turns out I needed to check against `news_article.edition_organisations`, not `news_article.organisations`... Updated implementation follows the other 'World news story' validations too.

test/unit/news_article_test.rb
```diff
@@ -88,19 +88,46 @@ class NewsArticleTest < ActiveSupport::TestCase
...
     news_article = build(:news_article_world_news_story)
     news_article.primary_locale = 'fr'
     assert news_article.valid?
   end
 
-  test "#world_news_story returns true" do
-    article = build(:news_article_world_news_story)
-
-    assert article.world_news_story?
+  test "should be invalid when not associating a worldwide organisation" do
```
gpeng: I think our general consensus (and possibly style guide) is to avoid 'should'. (I realise there is a lot of it about in Whitehall). "is invalid without a worldwide organisation" (or "it is"). Same in the couple of others below.

davidbasalla: Thanks, I've updated the tests (and fixed them in the rest of the test file)

PR #3232, opened 2017-05-17
Add policies validation for world news stories

gpeng:
Adds a validation rule for `NewsArticle` of type `NewsArticle::WorldNewsStory` that validates for the absence of related policies. We are currently migrating `WorldLocationNewsArticle` to this format. This will maintain the behaviour of that format after migration. 

image

[Trello](https://trello.com/c/OW1PrFWc/89-add-specific-validation-rules-for-the-new-world-news-article-subtype)

### DIFF
--- a/app/models/news_article.rb
+++ b/app/models/news_article.rb
@@ -8,6 +8,8 @@ class NewsArticle < Newsesque
   validates :news_article_type_id, presence: true
   validate :non_english_primary_locale_only_for_world_news_story
 
+  validate :policies_are_not_associated, unless: :can_be_related_to_policies?
+
   def self.subtypes
     NewsArticleType.all
   end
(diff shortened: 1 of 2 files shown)


app/models/news_article.rb
```diff
@@ -8,6 +8,8 @@ class NewsArticle < Newsesque
   validates :news_article_type_id, presence: true
   validate :non_english_primary_locale_only_for_world_news_story
 
+  validate :policies_are_not_associated, unless: :can_be_related_to_policies?
```
davidbasalla: Would `validates :edition_policies, absence: true, if: :world_news_story?` work as well? That's why I'm using in my (as of yet unmerged) [PR](https://github.com/alphagov/whitehall/pull/3225)

gpeng: It does but it was messy trying to get the error message to make sense, particularly since I asked for a bit of content input on the correct text for the message.

PR #3235, opened 2017-05-18
Add js to toggle world news story fields visibility

davidbasalla:
For https://trello.com/c/0TpwGP3Z/134-toggle-visibility-on-relevant-form-fields-for-world-news-story-news-article

Hides the relevant organisation/ world organisation fields added in https://github.com/alphagov/whitehall/pull/3225

This adds JS to toggle visibility and reset fields based on the selected ‘News Article’ type. To help with selecting the DOM elements, we add specific class names to the existing fieldsets and add fieldsets to elements that did not have them.

Working the JQuery Chzn select fields was a bit fiddly. The main realisation was that for clearing out select fields that support ‘multiple’ selected options you had to use `.val({})` (rather than `.val('')` as used for single selection fields). We also only clear the field when a value is set, otherwise it leads to weirdness such as truncated placeholder text…

------------------------------------------------

# Before

### Association form fields for 'News story' and 'World news story'
screen shot 2017-05-23 at 15 06 41

# After 

### Association form fields for 'News story'

screen shot 2017-05-23 at 15 05 31

### Association form fields for 'World news story'

screen shot 2017-05-23 at 15 05 57

### DIFF
--- a/app/assets/javascripts/admin/views/editions/_form.js
+++ b/app/assets/javascripts/admin/views/editions/_form.js
@@ -6,11 +6,13 @@
     init: function init(params) {
       this.$form = $(params.selector);
       this.rightToLeftLocales = params.right_to_left_locales;
+      this.$newsTypeSelect = $('select#edition_news_article_type_id')
 
       this.showChangeNotesIfMajorChange();
       this.showFormatAdviceForSelectedSubtype();
       this.toggleLanguageSelect();
-      this.toggleNonEnglishSupport();
+      this.toggleWorldNewsStoryVisibility();
+      this.resetWorldNewsStoryFields();
       this.toggleFirstPublishedDate();
 
       GOVUK.formChangeProtection.init($('#edit_edition'), 'You have unsaved changes that will be lost if you leave this page.');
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -121,6 +91,147 @@
...
     },
 
+    toggleWorldNewsStoryVisibility: function toggleWorldNewsStoryVisibility() {
+      // only toggle fields when news article type is editable
+      if ( $( "select#edition_news_article_type_id" ).length == 0 ) return;
+
+      var nonWorldNewsStoryFieldsets = this.findFieldsets(this.nonWorldNewsStoryFieldSelectors())
+      var worldNewsStoryFieldsets = this.findFieldsets(this.worldNewsStoryFieldSelectors())
+
+      var _this = this
```
gpeng: I think we should probably choose a different name here as we conventionally use `_var_name` to indicate that a variable isn't used. Maybe we should go with `var that = this`

app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -6,11 +6,13 @@
     init: function init(params) {
       this.$form = $(params.selector);
       this.rightToLeftLocales = params.right_to_left_locales;
+      this.newsTypeSelect = $('select#edition_news_article_type_id')
```
gpeng: Looks like the existing JS prior to this was using `$varName` to denote jQuery dom elements. If we're going to have that then we should use it consistently in the new stuff that we have added.

app/assets/javascripts/admin/views/editions/_form.js
davidbasalla: As discussed, I'll leave it as `_this` for now, as there are more references to it through the rest of the code base

davidbasalla:
@gpeng have addressed your comments and squished the commits into one

PR #3239, opened 2017-05-19
Fix person_helper bug

gpeng:
Many of the `Person` names in the Whitehall DB have a trailing space. The `PersonHelper` generates a list of person names for select options and where there is more than one person with the same name it disambiguates the name by appending the person's role.

This fixes a bug where if there are two people with the same name but one has a trailing space and one doesn't the disambiguation the names would not be disambiguated and only one of the users would appear in the select options in the admin UI.

[Trello](https://trello.com/c/RZwgbes3/135-bug-adding-a-new-person-to-an-organisation)

### DIFF
--- a/app/helpers/person_helper.rb
+++ b/app/helpers/person_helper.rb
@@ -18,7 +18,10 @@ module PersonHelper
   end
 
   def forename_surname(person)
-    [person.forename, person.surname].join(' ')
+    [
+      person.forename.try(:strip),
+      person.surname.try(:strip)
+    ].join(' ')
   end
 
   def disambiguate_option(person)
--- a/test/unit/helpers/person_helper_test.rb
+++ b/test/unit/helpers/person_helper_test.rb
@@ -1,0 +1,24 @@
+require "test_helper"
+
+class PersonHelperTest < ActionView::TestCase
+  include ApplicationHelper
+
+  test "it disambiguates names with trailing spaces" do
+    person_one = build(:person, forename: "Peter ", surname: "Jones")
+    person_two = build(:person, forename: "Peter", surname: "Jones")
+    person_one.stubs(:name_with_disambiguator).returns(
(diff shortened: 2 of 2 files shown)


test/unit/helpers/people_helper_test.rb
```diff
@@ -0,0 +1,20 @@
+require "test_helper"
+
+class PersonHelperTest < ActionView::TestCase
+  include ApplicationHelper
+
+  test "it disambiguates names with trailing spaces" do
+    person_one = build(:person, forename: "Peter ", surname: "Jones")
```
davidbasalla: I'm probably being dense, but what's the significance of having 'Peter Jones' for person_one and person_two, but then stubbing them to return 'Vic' and 'Bob'? I'll stare at it a bit longer, probably all will become clear...

Great that we're adding a test for this!

gpeng: The people have to have the same name to trigger the disambiguation call. I stubbed the response as otherwise I'd have had to create a load of other objects that get involved, `Role`, `RoleAppointment` probably a load of other stuff. The return value of `#name_with_disambiguator` is insignificant. We only care that `disambiguated_people_names` returns the values in its results if the `Person`s have the same name but one has a trailing space. The test is testing that the names are considered the same really. Perhaps I could have named it more clearly.

davidbasalla: ah I see! Maybe instead of 'Vic' & 'Bob' it could be 'Peter Jones - Executive Director, Prisons' & 'Peter Jones - Director, Overseas Territories' to more closely match the result of `person.name_with_disambiguator`?

gpeng: Fixed in a7875e5

gpeng:
Rebased

PR #3240, opened 2017-05-22
World news stories are always political

gpeng:
WorldNewsArticles are always considered political. As part of the migration of this format to NewsArticle this PR updates `PoliticalContentIdentifier` to always consider `NewsArticle` of type `NewsArticleType::WorldNewsStory` as political.

[Trello](https://trello.com/c/voHKjF2S/107-tag-all-world-news-stories-as-political)

### DIFF
--- a/lib/political_content_identifier.rb
+++ b/lib/political_content_identifier.rb
@@ -60,10 +60,15 @@ private
   end
 
   def always_political_format?
-    ALWAYS_POLITICAL_FORMATS.include?(edition.class)
+    ALWAYS_POLITICAL_FORMATS.include?(edition.class) ||
+      edition_is_a_world_news_story?
   end
 
   def never_political_format?
     edition.is_a?(FatalityNotice) || stats_publication?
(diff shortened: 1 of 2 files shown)


lib/political_content_identifier.rb
```diff
@@ -60,10 +60,15 @@ def political_publication_type?
...
+    ALWAYS_POLITICAL_FORMATS.include?(edition.class) ||
+      edition_is_a_world_news_story?
   end
 
   def never_political_format?
     edition.is_a?(FatalityNotice) || stats_publication?
   end
+
+  def edition_is_a_world_news_story?
+    edition.is_a?(NewsArticle) && edition.news_article_type == NewsArticleType::WorldNewsStory
```
davidbasalla: We could use the `world_news_story?` method here

gpeng: ooh yes... We could. You're right there. Fixed in 341d408

PR #3243, opened 2017-05-24
Enable B variant for world locations A/B test

rubenarakelyan:
This commit adds support for displaying the B variant of worldwide locations based on the content `worldwide_publishing_taxonomy_ab_test_content.yml`.

Trello: https://trello.com/c/IcRVNu4D/132-change-the-worldwide-location-index-controller-to-work-with-the-a-b-test

### DIFF
--- a/app/controllers/world_locations_controller.rb
+++ b/app/controllers/world_locations_controller.rb
@@ -1,3 +1,5 @@
+require 'yaml'
+
 class WorldLocationsController < PublicFacingController
   enable_request_formats index: [:json], show: [:atom, :json]
   before_action :load_world_location, only: :show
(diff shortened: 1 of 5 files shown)


app/controllers/world_locations_controller.rb
```diff
@@ -28,6 +30,23 @@ def show
         set_meta_description("What the UK government is doing in #{@world_location.name}.")
         set_slimmer_world_locations_header([@world_location])
         set_slimmer_organisations_header(@world_location.worldwide_organisations_with_sponsoring_organisations)
+        # Display the "B" variant of this page for users in the "B" bucket
+        ab_test = GovukAbTesting::AbTest.new("WorldwidePublishingTaxonomy", dimension: 45)
+        @requested_variant = ab_test.requested_variant(request.headers)
+        @requested_variant.configure_response(response)
+        if @requested_variant.variant_b?
+          parent_taxon = {
```
davidbasalla: We could move `parent_taxon` and `accordion_content` into private methods? Might declutter this controller action a bit?

rubenarakelyan: Indeed, done

app/controllers/world_locations_controller.rb
```diff
@@ -44,4 +56,15 @@ def load_world_location
...
 
+  def parent_taxon
+    {
+      title: @world_location.title,
+      description: "Accessing UK services from #{@world_location.name}, advice for travelling to the UK, and help with trading between the UK and #{@world_location.name}."
+    }
+  end
+
+  def accordion_content
+    YAML.load_file(Rails.root + "config/worldwide_publishing_taxonomy_ab_test_content.yml")[@world_location.slug]
```
davidbasalla: I think this could be memoized, since we're calling it twice in the controller method

rubenarakelyan: Nice - done!

PR #3244, opened 2017-05-24
Add task to republish political content

gpeng:
Editions that are `political: true` need to be republished following the election if we update the government record. This will enable history mode on that content and display the 'Published under' banner.

This commit adds a rake task to facilitate this.

[Trello](https://trello.com/c/mAfGFT4x/156-political-content-needs-to-be-republished-post-government-change)

### DIFF
--- a/lib/tasks/election.rake
+++ b/lib/tasks/election.rake
@@ -10,4 +10,21 @@ namespace :election do
       puts "changed to #{person.name}"
     end
   end
+
+  task republish_political_content: :environment do
+    political_document_ids = Edition
+      .where(political: true)
+      .pluck(:document_id)
+      .uniq
+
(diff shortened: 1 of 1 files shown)


davidbasalla:
Looks good to me 👍

danielroseman:
I don't know if it it matters, but this seems to be republishing too much. Surely there's no need to republish all the content from before the 2015 election, even if it is political; it's just 2015-7 stuff, which is marked as "current", that needs resending.

gpeng:
We did discuss reducing the scope as you're suggesting @danielroseman but because of the `date_for_government` logic and republishing large amounts of content being a pretty well worn path now after migration we went with just republishing everything political rather than risk missing some content. Happy to have another look at it if you'd prefer we reduce the volume of republishing.

PR #3247, opened 2017-05-25
Redirect B bucket worldwide organisation pages

gpeng:
As part of the worldwide taxonomy A/B we no longer want visitors assigned to the 'B' group to use the worldwide organisation pages for the countries under test.

This PR redirects 'B' group members to an appropriate world location URL where they will be served the new taxon page under test. In most cases this is the `WorldLocation` that is associated with the requested organisation. In a small number of cases the organisation is related to more than one location so these are hard coded to redirect to the correct location for the test.

[Trello](https://trello.com/c/6mugcukv/159-redirect-users-in-the-b-bucket-of-the-a-b-test-who-land-on-worldwide-organisation-pages)

### DIFF
--- a/app/controllers/worldwide_organisations_controller.rb
+++ b/app/controllers/worldwide_organisations_controller.rb
@@ -6,6 +6,7 @@ class WorldwideOrganisationsController < PublicFacingController
   respond_to :html, :json
 
   def show
+    redirect_if_required_for_ab_test
     respond_to do |format|
       format.html do
         set_expiry 5.minutes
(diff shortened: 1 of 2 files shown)


gpeng:
They should do... I'll add an explicit test though.

davidbasalla:
So for countries that are not in the test, what will happen to them in the B bucket? Looks like they will also get redirected?

davidbasalla:
For WorldLocations and WorldLocationNews (the bit I'm working on), we keep the old ('A' bucket) behaviour for non-english locales. Wondering if that would be relevant for this as well...

davidbasalla:
It might be worth adding another test to check that organisations belonging to countries in the B bucket that are _not_ part of the approved list don't get redirected

gpeng:
@davidbasalla I've addressed your comments and rebased into more logical commits.

davidbasalla:
LGTM!

PR #3248, opened 2017-05-25
Add world news page for a b test

davidbasalla:
For https://trello.com/c/KV9diS41/158-create-a-new-version-of-the-worldwide-location-template-with-only-news

This adds a new News page for world locations. It will be used for the A/B test for a number of selected countries. It’s a slimmed down version of the WorldLocation page, showing just featured and latest documents.

Also updates the WorldLocationController to use the latest `WorldwideAbTestHelper` class

----------------------------------
screen shot 2017-05-25 at 16 48 37

### DIFF
--- a/app/controllers/world_location_news_controller.rb
+++ b/app/controllers/world_location_news_controller.rb
@@ -1,0 +1,29 @@
+class WorldLocationNewsController < PublicFacingController
+  enable_request_formats index: [:atom, :json]
+  before_action :load_world_location, only: :index
+
+  def index
+    recently_updated_source = @world_location.published_editions.with_translations(I18n.locale).in_reverse_chronological_order
+    respond_to do |format|
+      format.html do
+        set_meta_description("What the UK government is doing in #{@world_location.name}.")
+        set_slimmer_world_locations_header([@world_location])
+
+        @recently_updated = recently_updated_source.limit(3)
+        @feature_list = FeatureListPresenter.new(@world_location.feature_list_for_locale(I18n.locale), view_context).limit_to(5)
+      end
+      format.json do
+        redirect_to api_world_location_path(@world_location, format: :json)
+      end
+      format.atom do
+        @documents = EditionCollectionPresenter.new(recently_updated_source.limit(10), view_context)
+      end
+    end
+  end
+
+private
+
+  def load_world_location
+    @world_location = WorldLocation.with_translations(I18n.locale).find(params[:world_location_id])
+  end
+end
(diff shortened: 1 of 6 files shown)


app/models/world_location.rb
```diff
@@ -113,6 +113,10 @@ def self.geographical
     where(world_location_type_id: WorldLocationType.geographic.map(&:id)).ordered_by_name
   end
 
+  def worldwide_publishing_taxonomy_ab_test_content
```
bevanloon: This looks like something I could use :)
Was considering some sort of helper class for the YAML business since I'm doing lots of nasty stuff to get my content out of it here https://github.com/alphagov/whitehall/pull/3250/files#diff-5886670c1b6ac5a5f3997948532b898dR62

app/controllers/world_location_news_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  enable_request_formats index: [:atom, :json]
+  before_action :load_world_location, only: :index
+  before_action :setup_a_b_test, only: :index
+
+  def index
+    recently_updated_source = @world_location.published_editions.with_translations(I18n.locale).in_reverse_chronological_order
+    respond_to do |format|
+      format.html do
+        if render_b_variant?
+          setup_a_b_test
```
rubenarakelyan: Do we need to call this here as well as in the `before_action`?

davidbasalla: Absolutely not, that shouldn't be there! Will fix!

davidbasalla: Fixed

test/functional/world_location_news_controller_test.rb
```diff
@@ -0,0 +1,138 @@
...
+    with_variant WorldwidePublishingTaxonomy: "A", assert_meta_tag: false do
+      get :index, world_location_id: @world_location
+
+      assert_redirected_to world_location_path(@world_location)
+    end
+  end
+
+  view_test "should redirect to the world location page if the user is in 'B' cohort for country not in ab test list" do
+    with_variant WorldwidePublishingTaxonomy: "B", assert_meta_tag: false do
+      world_location = create(:world_location)
```
gpeng: Might be better to set the slug to one we know isn't in the test here and not stub the `WorldwideAbTestHelper` as this is a higher level test. That will exercise the most of the stack and the content data file that we add.

davidbasalla: Thanks, I've updated the test so as to not stub `WorldwideAbTestHelper` at all

PR #3249, opened 2017-05-25
Add primary language and translations available to CSV export

h-lame:
For: https://trello.com/c/p3x8gE9F/162-add-translation-column-to-whitehall-csv-export

Not all editions have english as their primary language, and not all
editions have been translated.  Adding these columns to the CSV export
lets content analysts filter documents to find translated content quickly.

Note that in the presenter we use `edition.translated_locales.count > 1`
as a guard clause instead of the seemingly more obvious
`edition.available_in_multiple_languages?`.  The latter assumes english
as a default and reports true if the edition has any non-english
translations.  There are some editions that are not available in
english at all and so available_in_multiple_languages? reports true, when
they may only be available in one language.  It might be that
`Edition#available_in_multiple_languages?` would be better renamed as
`Edition#available_in_languages_other_than_english?` to avoid this
confusion.

### DIFF
--- a/app/presenters/document_list_export_presenter.rb
+++ b/app/presenters/document_list_export_presenter.rb
@@ -28,7 +28,9 @@ class DocumentListExportPresenter
       'Specialist sectors',
       'Collections',
       'Can have history-mode',
-      'History-mode applied'
+      'History-mode applied',
+      'Primary language',
+      'Translations available',
     ]
   end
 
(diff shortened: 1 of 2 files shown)


h-lame:
Flagged as DNM because I want to check that the export is ok with content designers first who will be the main users of it.

h-lame:
Removed DNM as I've checked with content folk and they're cool with the new columns

PR #3250, opened 2017-05-25
Add embassy contact page template

bevanloon:
This PR adds a new Embassy Contact page that will replace the worldwide organisation pages for users in the B group of the worldwide taxonomy A/B test.

Users in the B group attempting that request a translated version of the page will be shown the current page as we aren't supporting translations for the A/B test.

Trello: https://trello.com/c/9AWsaa3N/154-add-embassy-contact-page-template-to-whitehall

### DIFF
--- a/app/controllers/world_locations_controller.rb
+++ b/app/controllers/world_locations_controller.rb
@@ -53,7 +53,7 @@ class WorldLocationsController < PublicFacingController
     end
   end
 
-  private
+private
 
   def load_world_location
     @world_location = WorldLocation.with_translations(I18n.locale).find(params[:id])
--- a/app/controllers/worldwide_organisations_controller.rb
+++ b/app/controllers/worldwide_organisations_controller.rb
@@ -2,6 +2,7 @@ class WorldwideOrganisationsController < PublicFacingController
   include CacheControlHelper
   enable_request_formats show: :json
   before_action :load_worldwide_organisation
+  before_action :setup_a_b_test
 
   respond_to :html, :json
 
@@ -24,17 +25,25 @@ class WorldwideOrganisationsController < PublicFacingController
   end
 
   def show_b_variant
-    render template: 'embassies/show'
+    @world_location = WorldLocation.with_translations(I18n.locale).find(params[:world_location_id])
+    @embassy_data = embassy_test_data(params[:id])
+    @primary_role = primary_role
+    @other_roles = ([secondary_role] + office_roles).compact
+
+    set_meta_description(@worldwide_organisation.summary)
+    set_slimmer_organisations_header([@worldwide_organisation] + @worldwide_organisation.sponsoring_organisations)
+
+    if b_variant? && @embassy_data
+      render template: 'embassies/show'
+    else
+      redirect_to @worldwide_organisation
+    end
   end
 
 private
 
   def redirect_if_required_for_ab_test
-    ab_test = GovukAbTesting::AbTest.new("WorldwidePublishingTaxonomy", dimension: 45)
-    requested_variant = ab_test.requested_variant(request.headers)
-    requested_variant.configure_response(response)
-
-    if requested_variant.variant_b? &&
+    if b_variant? &&
         worldwide_test_helper.is_under_test?(@worldwide_organisation) &&
         locale_is_en?
       redirect_to worldwide_test_helper.location_for(@worldwide_organisation)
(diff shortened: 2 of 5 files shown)


rubenarakelyan:
Adding `?ABTest-WorldwidePublishingTaxonomy=A` to any links to translated content will push people visiting those pages back into the "A" cohort.

bevanloon:
@gpeng I think 8ee790c8e227206a99baf1e6bb250f52cac0dabd should fix up the redirect.

app/controllers/world_locations_controller.rb
```diff
@@ -79,4 +79,8 @@ def parent_taxon
       description: "Accessing UK services from #{@world_location.name}, advice for travelling to the UK, and help with trading between the UK and #{@world_location.name}."
     }
   end
+
+  def b_variant_page_content
```
bevanloon: Don't think we need this anymore given https://github.com/alphagov/whitehall/blob/master/app/controllers/world_locations_controller.rb#L43

config/routes.rb
```diff
@@ -150,6 +150,7 @@ def external_redirect(path_prefix, target)
 
     resources :world_locations, path: 'world', only: [:index, :show], localised: true do
       resources :world_location_news, path: 'news', only: [:index]
+      resources :worldwide_organisations, path: '', only: [:show], to: 'worldwide_organisations#show_b_variant'
```
davidbasalla: What does the empty string for `path` signify there?

app/controllers/worldwide_organisations_controller.rb
```diff
@@ -23,8 +23,36 @@ def show
...
+  def show_b_variant
+    @world_location = WorldLocation.with_translations(I18n.locale).find(params[:world_location_id])
+    @embassy_data = embassy_test_data(params[:id])
+    @primary_role = primary_role
+    @other_roles = ([secondary_role] + office_roles).compact
+
+    set_meta_description(@worldwide_organisation.summary)
+    set_slimmer_organisations_header([@worldwide_organisation] + @worldwide_organisation.sponsoring_organisations)
+
+    if show_b_variant? && @embassy_data
```
davidbasalla: So the `&& @embassy_data` is effectively checking if there's any data for this organisation in the YAML file? In the other controllers we used the `is_under_test?` method from the WorldwideABTestHelper. Might not make a huge difference here though

app/controllers/worldwide_organisations_controller.rb
```diff
@@ -60,4 +88,28 @@ def secondary_role
...
+  end
+
+  def configure_ab_variant
+    ab_test = GovukAbTesting::AbTest.new("WorldwidePublishingTaxonomy", dimension: 45)
+    @requested_variant = ab_test.requested_variant(request.headers)
+    @requested_variant.configure_response(response)
+  end
+
+  def show_b_variant?
+    configure_ab_variant if @requested_variant.nil?
```
davidbasalla: Wouldn't this always be `nil`? Wondering if we need the `if` clause...

Also there's a bit of duplication of setting up AB test stuff (see method `redirect_if_required_for_ab_test` use in the `show` controller method). I added a before_action where I do all ab test setup in https://github.com/alphagov/whitehall/blob/master/app/controllers/world_location_news_controller.rb#L35-L39, not sure if that's neater?

app/controllers/worldwide_organisations_controller.rb
```diff
@@ -60,4 +88,28 @@ def secondary_role
   def office_roles
     @worldwide_organisation.office_staff_roles.map { |office_staff| RolePresenter.new(office_staff, view_context) }
   end
+
+  def test_helper_content
+    ab_test_helper = WorldwideAbTestHelper.new
```
davidbasalla: We can use the existing `worldwide_test_helper` method here

config/routes.rb
bevanloon: It suppresses `/worldwide_organisations` in the url so we get a url in the form of `/government/world/<location>/<organisation>` rather than `/government/world/<location>/worldwide_organisations/<organisation>`

Not sure if there's a better way to do that?

app/controllers/worldwide_organisations_controller.rb
bevanloon: I think if we use `is_under_test?` we could have a situation where we have, for example, an `organisation` in a `world_location` where the `world_location` is in the YAML file, but the `organisation` isn't.....in that instance `is_under_test?` would return true (because it's testing for existence of the `world_location` in the YAML) but there would be no content.

For example - if our YAML was

```
australia:
  taxonomy:
    - title: Help for British nationals in Australia
      description: Travel documents and help with emergencies.
      base_path: help-for-british-nationals
      tagged_content:
        - title: Get a new or replacement UK passport
          description: Forms, prices and application
          base_path: /apply-renew-passport
        - title: Get emergency UK travel documents
          description: If you're abroad, need to travel and can't get a passport in time.
          base_path: /emergency-travel-document
  embassies:
    british-high-commission-canberra:
      - title: British High Commission Canberra
        summary: The summary
        body: |
          The High Commission in Canberra....
brazil:
  taxonomy:
.
.
.
```

and we hit `/world/australia/british-high-commission-melbourne`, we would end up with `is_under_test?` returning true because Australia is in the YAML, but there would be nothing for the `british-high-commission-melbourne`

I did think of pushing the test for having actual embassy data into the test helper, but I thought it might get messy to start iterating through the YAML structure looking for data so I left it in the controller for now.

davidbasalla: Ah yes, I see! For some reason I though that `is_under_test?` also checks whether the organisation is included in YAML but such appears not to be the case!

app/controllers/worldwide_organisations_controller.rb
bevanloon: 👍 should be fixed in d5f7478e73591b779b667398e5a30b9522402067

app/controllers/worldwide_organisations_controller.rb
bevanloon: yes - much neater - I've implemented that in d5f7478e73591b779b667398e5a30b9522402067

bevanloon:
@davidbasalla I think I've addressed everything :)
fixed an additional bug while I was at it 👍

PR #3251, opened 2017-05-25
Add helper class for Worldwide taxonomy A/B test

gpeng:
Centralises logic for redirects and access to the YAML content file to help with the implementation of the Worldwide taxonomy A/B test.

Also provides methods to simplify checking if a particular location or organisation should be included in the test.

### DIFF
--- a/lib/worldwide_ab_test_helper.rb
+++ b/lib/worldwide_ab_test_helper.rb
@@ -1,8 +1,8 @@
 class WorldwideAbTestHelper
   attr_reader :content
 
-  def initialize
-    @content = YAML.load_file(file_path)
+  def initialize(content_file_path = nil)
+    @content = YAML.load_file(content_file_path || file_path)
   end
 
   def has_content_for?(location_slug)
--- a/lib/worldwide_ab_test_helper.rb
+++ b/lib/worldwide_ab_test_helper.rb
@@ -1,0 +1,59 @@
+class WorldwideAbTestHelper
+  attr_reader :content
+
+  def initialize
+    @content = YAML.load_file(file_path)
+  end
+
+  def has_content_for?(location_slug)
+    !!content_for(location_slug)
+  end
+
+  def content_for(location_slug)
+    content[location_slug]
+  end
+
+  def location_for(worldwide_organisation)
+    hard_coded_location_for(worldwide_organisation) ||
+      worldwide_organisation.world_locations.first
+  end
+
+  def is_under_test?(testable_object)
+    location = testable_object
+    if testable_object.respond_to?(:world_locations)
+      location = location_for(testable_object)
+    end
+    has_content_for?(location.slug)
+  end
+
+private
+
+  def root
+    Rails.root
+  end
+
+  def file_path
+    File.join(root, "config/worldwide_publishing_taxonomy_ab_test_content.yml")
+  end
+
+  def hard_coded_location_slugs
+    {
+      "british-high-commission-pretoria" => "south-africa",
+      "british-consulate-general-los-angeles" => "usa",
+      "did-south-africa" => "south-africa",
+      "british-deputy-high-commission-kolkata" => "india",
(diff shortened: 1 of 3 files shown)


lib/worldwide_ab_test_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+    }
+  end
+
+  def hard_coded_location_for(worldwide_organisation)
+    wwo_slug = worldwide_organisation.slug
+    hard_coded_value = hard_coded_location_slugs[wwo_slug]
+    return unless hard_coded_value.present?
+
+    locations = worldwide_organisation.world_locations
+    matching_locations = locations.select do |location|
```
davidbasalla: Could you use `.detect` or `.find` here?

lib/worldwide_ab_test_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+  def content_for(location_slug)
+    content[location_slug]
+  end
+
+  def location_for(worldwide_organisation)
+    hard_coded_location_for(worldwide_organisation) ||
+      worldwide_organisation.world_locations.first
+  end
+
+  def is_under_test?(worldwide_organisation)
```
davidbasalla: Wondering if the naming of this method should be more descriptive, since since other things apart from  worldwide_organisations can be under test too (eg world locations). But I guess we can adapt the method names as we go.

test/unit/worldwide_ab_test_helper_test.rb
```diff
@@ -0,0 +1,63 @@
+require 'test_helper'
+
+class WorldwideAbHelperTest < ActiveSupport::TestCase
+  def subject
+    WorldwideAbTestHelper.new
+  end
+
+  test "content returns the parsed file " do
+    result = subject.content
+    assert result[:sample].present?
```
davidbasalla: We're relying on the actual yml file here, so it sort of acts like a fixture. We could change the `WorldwideAbTestHelper` class so that we initialise it with some dummy content instead. Something like:

```
dummy_content = {
  sample: {
    something: "stuff"
  }
}

WorldwideAbTestHelper.new(dummy_content)
```

That might make the tests clearer in terms of what data is legitimate and what isn't. Although that would just be changing the code to aid testing, so not sure if there's any benefit to it (other than someone reading this having to look at the yml file). Probably not a big deal, the reliance on the yml file just jumped out at me...

lib/worldwide_ab_test_helper.rb
gpeng: Fixed in f5be0ad

lib/worldwide_ab_test_helper.rb
gpeng: Updated in f5be0ad. I've made it support both and left the name.

test/unit/worldwide_ab_test_helper_test.rb
gpeng: I ummed and arred a bit over that myself. There's probably value in testing this over the actual content file as it will go some way to verifying the contents of the file is parseable etc. That said I also agree with your points. I've added another commit with an inline fixture. See what you think and we can decide whether to keep it in.

test/unit/worldwide_ab_test_helper_test.rb
```diff
@@ -0,0 +1,148 @@
...
+      organisation = stub(
+        slug: "british-deputy-high-commission-kolkata",
+        world_locations: [location, hard_coded_location]
+      )
+
+      assert_equal hard_coded_location, subject(file_path).location_for(organisation)
+    end
+  end
+
+  test "is_under_test? returns true if location if first location" do
```
davidbasalla: small typo here, should be 'if location is first location'

gpeng: Fixed in 00e27a7

PR #3252, opened 2017-05-26
Fix incorrect rendering of multiple same Id

bevanloon:
The current use of `content_tag_for` in the `worldwide_organisation` partial is causing html to be rendered such that we have multiple Ids that are the same on a single page. This is incorrect.

Remove `content_tag_for` and replace it with a hand coded div.
Drop the use of `id` in the div.

The current html being produced will look something like this:

```
<div class="organisation sponsoring-organisation" id="organisation_13">
```

As can be seen on, for example, https://www.gov.uk/government/world/france

The fix will create html like so:

```
<div class="organisation sponsoring-organisation">
```

### DIFF
--- a/app/views/worldwide_organisations/_worldwide_organisation.html.erb
+++ b/app/views/worldwide_organisations/_worldwide_organisation.html.erb
@@ -2,11 +2,11 @@
   <div class="worldwide-organisation-detail">
     <div class="content">
       <% worldwide_organisation.sponsoring_organisations.each do |organisation| %>
-        <%= content_tag_for(:div, organisation, :class => 'sponsoring-organisation') do %>
+        <div class="organisation sponsoring-organisation">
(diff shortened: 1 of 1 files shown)


h-lame:
Does changing the class have any impact?  There are `.organisation` definitions in the css (but I confess I haven't properly searched for to make sure they would actually apply in this case).

bevanloon:
@h-lame I don't think so - at least I couldn't spot any issues. Hopefully @fofr can weigh in with his front end nous. 

Happy to fix the problem a different way though if you think it would be safer? Could drop the `content_tag_for` and just code the raw div up without an `id`. I'm not sure if that has any other implications....

h-lame:
`content_tag_for` is going away in rails 5, so maybe hand coding it would be for the best anyway?

fofr:
IDs look like they were incidental rather than something useful. :+1:

PR #3259, opened 2017-05-31
Send `primary_publishing_organisation` to publishing-api

tijmenb:
This sends the `primary_publishing_organisation` to the publishing-api. It uses the first of the lead organisations to do this.

The field will be indexed in search and used when auditing content.

Tests will fail until https://github.com/alphagov/govuk-content-schemas/pull/622 is deployed.

https://trello.com/c/86sDRf3t

### DIFF
--- a/test/unit/presenters/publishing_api/case_study_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/case_study_presenter_test.rb
@@ -64,7 +64,7 @@ class PublishingApi::CaseStudyPresenterTest < ActiveSupport::TestCase
       presented_item.content[:details].delete(:body)
     )
     assert_equal expected_content[:details], presented_item.content[:details].except(:body)
-    assert_equal expected_links, presented_item.links
+    assert_hash_includes presented_item.links, expected_links
     assert_equal case_study.document.content_id, presented_item.content_id
   end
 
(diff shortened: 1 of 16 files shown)


test/unit/presenters/publishing_api/case_study_presenter_test.rb
```diff
@@ -64,7 +64,7 @@ def present(edition)
       presented_item.content[:details].delete(:body)
     )
     assert_equal expected_content[:details], presented_item.content[:details].except(:body)
-    assert_equal expected_links, presented_item.links
+    assert_hash_includes presented_item.links, expected_links
```
gpeng: Aren't we weakening the test a bit with this? Before it was testing that all of the expected links and only the expected links were being sent and now it is testing that at least the expected links are being sent. We could be sending others that we don't expect and this test still passes?

tijmenb: Yeah, fair question. I suppose the thinking is that a test should only test what it cares about - in this case that the `expected_links` are present. While there _could_ be instances that you want to make sure that the payload matches exactly, I can't really think of a thing that could wrong here. Especially because all the payloads are tested against the content schemas.

gpeng: Fair enough :+1:

tijmenb:
@gpeng green now!

PR #3262, opened 2017-05-31
Apply vertical link padding only to main content of worldwide taxon page

davidbasalla:
Previously the foreign language link would be offset vertically due to the css rules which were introduced for the taxon page. This commit removes the 3px padding from all links and applies it just to the links in the main body, identified by class 'child-topic-contents'.

This is for the new taxon pages that only show for selected countries in the B bucket of the Worldwide A/B test.

## Before

screen shot 2017-05-31 at 15 39 02

## After 

screen shot 2017-05-31 at 15 58 24

### DIFF
--- a/app/assets/stylesheets/frontend/views/_worldwide_publishing_taxonomy.scss
+++ b/app/assets/stylesheets/frontend/views/_worldwide_publishing_taxonomy.scss
@@ -56,7 +56,8 @@
 
   a {
     margin: -3px;
-    padding: 3px;
+    padding-right: 3px;
+    padding-left: 3px;
     outline-color: transparent;
     display: block;
   }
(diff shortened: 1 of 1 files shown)


nickcolley:
Nice, what does the focus style look like?

davidbasalla:
@nickcolley See below (not sure if I'm css-ing right...)
screen shot 2017-05-31 at 15 52 30

davidbasalla:
@nickcolley I've made another change, moving just the _vertical_ padding down to the main content. Doesn't feel great but retains the horizontal padding of the language link...

nickcolley:
It seems like the a selector you removed that padding from is pretty widely scoped so I'm wondering if we should move that margin too, where else are these anchors being styled this way.

davidbasalla:
This css file is only being used for our A/B test for worldwide publishing, so only the worldwide_publishing_taxonomy show template. So the scope of this css file is fairly limited.

I've had a look at moving the margin it's not really making any difference so would be inclined to leave until it causes noticeable problems...

PR #3264, opened 2017-05-31
Return AB test path for WorldwideOrganisation

gpeng:
For the worldwide AB test we require that the usual worldwide organisation path for embassies that are part of the test redirect to the parent location.

In certain circumstances we still need to link to the embassy from pages e.g. corporate information pages and this requires that the links are updated with the new path to avoid the redirect and go to the new embassy pages.

This commit temporarily overrides the path and url helpers for `WorldwideOrganisation` to achieve this for the duration of the test.

[Trello](https://trello.com/c/Biya5HFW/167-fix-link-to-embassies-under-test-from-corporate-information-pages)

This helper is used to generate links on the embassy pages (which are replaced in the test with hard coded links to the new URL) and world location news article pages in the metadata component. The latter links should probably point to the embassy contact pages for the test. There may be other inline links but there don't appear to be any internally when checked with a Google link search. External links for B group users will be directed to the country taxon pages as intended.

### DIFF
--- a/app/helpers/worldwide_organisations_helper.rb
+++ b/app/helpers/worldwide_organisations_helper.rb
@@ -1,9 +1,17 @@
+require "localised_url_path_helper"
+
 module WorldwideOrganisationsHelper
+  include LocalisedUrlPathHelper
+
   def worldwide_organisation_path(worldwide_organisation, options = {})
-    worldwide_organisation_url(
-      worldwide_organisation,
-      options.merge(only_path: true)
-    )
+    if ab_test_helper.is_under_test?(worldwide_organisation)
+      worldwide_organisation_url(
+        worldwide_organisation,
+        options.merge(only_path: true)
+      )
+    else
+      super
+    end
   end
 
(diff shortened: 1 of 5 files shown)


app/helpers/worldwide_organisations_helper.rb
```diff
@@ -0,0 +1,27 @@
+module WorldwideOrganisationsHelper
+  def worldwide_organisation_path(worldwide_organisation, options = {})
+    worldwide_organisation_url(
+      worldwide_organisation,
+      options.merge(only_path: true)
```
davidbasalla: What does the `only_path: true` part do?

gpeng: Yep. That's it.

davidbasalla:
As discussed in person, might be worth considering if we can check the A/B Bucket somehow

bevanloon:
Closing in favour of https://github.com/alphagov/whitehall/pull/3276

PR #3261, opened 2017-05-31
Add contact info to group B embassy pages

bevanloon:
New embassy pages should also show contact information as it currently displays on the worldwide organisation pages.

#### With main and home page offices
with_home_page_offices

#### With main offices but without home page offices
without_home_page_offices

Had a quick chat with Lucy and she advised to include both in the page since.

PR #3265, opened 2017-06-01
Revert "Add contact info to group B embassy pages"

bevanloon:
Reverts alphagov/whitehall#3261

In discussion, we have decided to only show the `main_office` rather than both `main_office` and `home_office`


bevanloon:
Closing in favour of https://github.com/alphagov/whitehall/pull/3266 so we keep a record of the conversation and decision

PR #3267, opened 2017-06-01
Remove option to create WorldLocationNewsArticle

davidbasalla:
For https://trello.com/c/mrgVViLU/165-disable-creation-of-new-wlnas

Removes the option to create a World Location News Article
Remove tests and feature tests relevant to creating World Location News Articles
Changed the feature that was testing for emails from WLNAs for WorldLocations to test against NewsArticles instead...

This PR does not disable the functionality to _edit_ World Location News Articles...

Do not merge as we need to give 2 days notice...

--------------------------

## Before
screen shot 2017-06-01 at 17 08 03

## After
screen shot 2017-06-01 at 17 07 32


davidbasalla:
I haven't removed any sort of controller logic since I wasn't able to find any specific implementation for just World Location News Articles, since it seems to mostly inherit its behaviour...

davidbasalla:
thanks @gpeng, have fixed. 

Good point about the email thing for NewsArticles, I'll check that tomorrow!

bevanloon:
Should we be removing the admin routes at this point or is that happening later in the epic?
We could, for example, still manually visit `/admin/world-location-news/new` and create a WLNA if we were a savvy user.

davidbasalla:
Good point @bevanloon, I'll see if I can restrict the new and create routes!

davidbasalla:
@bevanloon I've updated the routes.rb file in that last commit, what do you think?

bevanloon:
@davidbasalla I think that looks good. I've just been looking at the epic in Trello - I think once we've migrated `WLNA`s to be `NewsArticle`s, we will need to revisit and remove a bit of admin functionality around search and edit.

davidbasalla:
Added Do not merge tag, as we need to give FCO a 2 day warning before we deploy this.

@gpeng, are the changes okay with you? I'm still looking into the email thing for 'World news story' News articles, but that shouldn't hold up this PR.

davidbasalla:
Made another change - I changed the email sign up feature to test against a NewsArticle rather than a WorldLocationNewsArticle. Turns out the emails do get sent for NewsArticles linked to WorldLocations already, there just wasn't a test for it.

PR #3271, opened 2017-06-05
Allow publishers to tag stats announcements in whitehall

klssmith:
Publishers are now able to tag statistical announcements to the new taxonomy.
As with tagging documents, this is only enabled for statistical announcements belonging 
to certain organisations. Tagging statistical announcements is not mandatory.

[Trello card](https://trello.com/c/HSMA7TyX/87-allow-publishers-to-tag-stats-announcements-in-whitehall)

### Screenshots
screen shot 2017-06-05 at 10 53 58

screen shot 2017-06-05 at 10 54 40


app/controllers/admin/statistics_announcement_tags_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def edit
+    @govuk_taxonomy = Taxonomy::GovukTaxonomy.new
+    @tag_form = TaxonomyTagForm.load(@statistics_announcement.content_id)
+  end
+
+  def update
+    @tag_form = TaxonomyTagForm.new(
+      content_id: @statistics_announcement.content_id,
+      selected_taxons: params["taxonomy_tag_form"].fetch("taxons", []).reject(&:blank?),
```
whoojemaflip: this is sufficiently complex to warrant it's own method.

app/controllers/admin/statistics_announcements_controller.rb
```diff
@@ -8,6 +8,10 @@ def index
     @statistics_announcements = @filter.statistics_announcements
   end
 
+  def show
+    fetch_expanded_links if @statistics_announcement.can_be_tagged_to_taxonomy?
```
whoojemaflip: might it be more explicit to set the `@expanded_links` variable here?

app/models/statistics_announcement.rb
```diff
@@ -207,8 +207,24 @@ def requires_redirect?
...
+  end
 
 private
+
+  def organisations_in_tagging_beta?(org_content_ids)
+    return false if org_content_ids.empty?
+
+    organisations_in_tagging_beta = Whitehall.organisations_in_tagging_beta
+
+    org_content_ids.any? do |id|
```
whoojemaflip: This runs at  O(n<sup>2</sup>) so maybe very slow if either of the arrays is sizeable. 

Could it work with an intersection instead? `(organisations_in_tagging_beta & org_content_ids).present?`

klssmith:
@whoojemaflip Thanks - I've addressed your comments

app/views/admin/statistics_announcements/show.html.erb
```diff
@@ -69,6 +69,33 @@
         <%= link_to "Cancel statistics release", [:cancel, :admin, @statistics_announcement], class: "btn btn-default pull-right" %>
       <% end %>
     </div>
+
+    <% if @statistics_announcement.can_be_tagged_to_taxonomy? %>
+      <section class="taxonomy-topics" id="topic-new-taxonomy">
```
whoojemaflip: this is quite a complex bit of view code. Could it be made into a partial? How much does it have in common with the `edition` version?

Issue #2577, opened 2016-05-04
Upgrade carrierwave to mitigate CVE-2016-3714


alexmuller:
https://github.com/carrierwaveuploader/carrierwave/issues/1933 indicates that there's a vulnerability in the way we handle uploads. When a fix is available we need to upgrade carrierwave.

boffbowsh:
What's different about this that's not mitigated by setting a policy.xml for imagemagick?

alexmuller:
The `policy.xml` will restrict what ImageMagick is able to do but there's an underlying problem in that we were able to get the exploit through whitehall by simply changing the file extension from `mvg` to `jpg`. I think that's what the linked issue is referring to - carrierwave should be more robust when checking what kind of file is being uploaded, and we should blacklist things that don't match image formats we accept.

floehopper:
This might've been fixed in #3273.

PR #3273, opened 2017-06-05
Upgrade to Rails 5

davidbasalla:
For https://trello.com/c/eg61wStW/78-upgrade-whitehall-to-a-supported-version-of-rails


app/assets/javascripts/all.js
```diff
@@ -5,7 +5,7 @@
 // the compiled file.
 //
 //= require govuk-admin-template
-//= require vendor/jquery/jquery-ui-1.10.2.custom
+//= require jquery.ui.all
```
davidbasalla: The commit message for this change mentions a gem. Would that be `jquery-ui-rails`? Out of interest what prompted this change?

rubenarakelyan: Yeah, `jquery-ui-rails` was upgraded to work with Rails 5 and I also took the opportunity to switch to using the version of jQuery UI that the gem provides rather than a static version built into whitehall

app/uploaders/attachment_uploader.rb
```diff
@@ -12,6 +12,14 @@ class AttachmentUploader < WhitehallUploader
 
   before :cache, :validate_zipfile_contents!
 
+  process :set_content_type
+  def set_content_type
+    content_type = MIME::Types.type_for(full_filename(file.file)).first.content_type
```
tuzz: Will this always return a mime type?

app/presenters/publishing_api/working_group_presenter.rb
```diff
@@ -53,7 +53,11 @@ def details
 
     def body
       # It looks 'wrong' using the description as the body, but it isn't
-      Whitehall::GovspeakRenderer.new.govspeak_with_attachments_to_html(item.description, item.attachments, item.email)
+      if item.description.present?
+        Whitehall::GovspeakRenderer.new.govspeak_with_attachments_to_html(item.description, item.attachments, item.email)
```
tuzz: Why wasn't this failing before the upgrade? Or was it failing but we didn't notice?

app/views/home/how_government_works.html.erb
```diff
@@ -18,7 +18,7 @@
       <div class="halves-width-section prime-minister">
         <div class="one-half">
           <div class="inner-block">
-            <% if @prime_minister.present? %>
+            <% if @prime_minister.present? && @prime_minister.image_url(:s465).present? %>
```
tuzz: Presumably the real data has an image for the prime minister? Was this change because of a test failure?

test/unit/whitehall/publishing_api_test.rb
```diff
@@ -156,6 +158,8 @@ class Whitehall::PublishingApiTest < ActiveSupport::TestCase
   test ".bulk_republish_async publishes to the Publishing API as a 'republish'" do
     take_part_page = create(:take_part_page)
     presenter = PublishingApiPresenters.presenter_for(take_part_page, update_type: 'republish')
+    WebMock.reset!
```
tuzz: 👍

lib/data_hygiene/dup_filename_attachment_fixer.rb
```diff
@@ -1,70 +0,0 @@
...
-      def basename
-        File.basename(attachment.file.path, ".#{extension}")
-      end
-
-      def extension
-        attachment.file_extension
-      end
-    end
-  end
-end
```
tuzz: Should we drop commit https://github.com/alphagov/whitehall/pull/3273/commits/140b4ea59b432868e5272990d48822a4e28e49f7 ?

test/unit/edition/topic_test.rb
```diff
@@ -9,6 +9,10 @@ class EditionWithTopics < Edition
     @topic = create(:topic)
   end
 
+  test "includes PublishesToPublishingApi" do
+    assert Topic.new.is_a?(PublishesToPublishingApi)
+  end
```
tuzz: Is it worth testing the include? Is there some effect we could test that is only provided by including this module? (e.g. that it publishes when the record is saved)

test/support/publishing_api_test_helpers.rb
```diff
@@ -47,4 +47,19 @@ def expect_no_republishing(*editions)
...
+      klass.any_instance.stubs(:publish_gone_to_publishing_api)
+    end
+
+    yield
+
+    all_classes.each do |klass|
+      klass.any_instance.unstub(:publish_to_publishing_api)
+      klass.any_instance.unstub(:publish_gone_to_publishing_api)
+    end
+  end
```
tuzz: I think given the alternatives, this is a sensible approach.

If we did want to change the helpers to support `times` or change Whitehall to use queueing, we'd easily be able to find all calls to `disable_publishes_to_publishing_api`.

We should make sure we have at least some tests that use that module.

app/views/admin/people/show.html.erb
```diff
@@ -11,7 +11,7 @@
 
     <article class="person">
 
-      <% if @person.image %>
+      <% if @person.image_url %>
```
tuzz: Is it likely there'll be other cases like this? Might be something we need to check carefully.

tijmenb:
Well done email team!! 🎉

floehopper:
@tuzz, @gpeng, @rubenarakelyan: I see that you upgraded carrierwave as part of this. Do you know whether the upgrade fixes CVE-2016-3714 (see #2577)?

PR #3274, opened 2017-06-05
Migrate WorldLocationNewsArticle to NewsArticle

gpeng:
This PR adds a data migration that converts all existing `WorldLocationNewsArticle` documents to `NewsArticle`.

The conversion involves a few steps:

* change the `Document#document_type` to `NewsArticle`.
* set the `Document#sluggable_string` to the existing slug. This causes it to be regenerated. If there is an existing `NewsArticle` with the slug of the `WorldNewsArticle` the slug will be disambiguated in the same way as articles with the same title are treated (appending `--n` to the slug). The document is saved at this point.
* Update the `type` STI field on all of the documents `Edition`s. This is done using `update_all` to bypass validation as changing of the document's type when it is in `published` state is disallowed.
* Update the `news_article_type` of all of the editions to `NewsArticleType::WorldNewsStory`
* Republish the document to update publishing API and content store and to create a redirect to the new URL. `/government/world-location-news/<slug>` -> `/government/news/<slug>`

Also updates the `NewsArticle` sync check to handle `nil` policies which is valid for WLNA.

[Trello](https://trello.com/c/9odHKy0o/139-migrate-existing-wlna-to-newsarticle-format)

### DIFF
--- a/lib/tasks/convert_wlna_to_news_article.rake
+++ b/lib/tasks/convert_wlna_to_news_article.rake
@@ -1,0 +1,61 @@
+namespace :wlna do
+  task :migrate_to_news_article, [:start_document_id, :end_document_id] => :environment do |_t, args|
+    start_id = args[:start_document_id]
+    end_id = args[:end_document_id]
+    raise "start_document_id and end_document_id must be supplied" unless start_id && end_id
+
+    documents = Document.where(id: start_id..end_id)
+      .where(document_type: "WorldLocationNewsArticle")
+
+    puts "Migrating #{documents.count} WLNA -> NewsArticle"
+    documents.each do |wlna_document|
+      begin
+        #setting sluggable_string causes #should_generate_new_friendly_id?
+        #to return true. This will regenerate the slug and
+        #disambiguate if there is already a news article with the slug
+        wlna_document.sluggable_string = wlna_document.slug
+        wlna_document.update_attributes(
+          document_type: "NewsArticle",
+        )
+      rescue ActiveRecord::RecordNotUnique
+        puts "NotUnique: #{wlna_document.id}"
+      end
+
+      editions = editions_including_deleted(wlna_document.id)
+
+      update_db_records_to_news_article(editions)
+      update_unpublishings(editions)
+      update_rummager(editions)
+
+      PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", wlna_document.id)
+
+      print "."
+    end
+
+    unless documents.empty?
+      puts "First document id #{documents.first.id}"
+      puts "Last document id #{documents.last.id}"
+    end
+  end
+end
+
+def update_db_records_to_news_article(editions)
+  editions.update_all(
+    type: "NewsArticle",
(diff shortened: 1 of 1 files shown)


db/data_migration/20170522134843_convert_wlna_to_news_articles.rb
```diff
@@ -0,0 +1,39 @@
...
+  Edition.unscoped.where(document_id: document_id)
+end
+
+world_location_news_articles = WorldLocationNewsArticle.all
+wlna_documents = Document.where(id: world_location_news_articles.pluck(:document_id).uniq)
+
+puts "Migrating #{wlna_documents.count} WLNA -> NewsArticle"
+
+wlna_documents.each do |wlna_document|
+  wlna_document.document_type = "NewsArticle"
```
davidbasalla: What does assigning `"NewsArticle"` to `document_type` do in this context? Can't spot if this is getting saved anywhere...

db/data_migration/20170522134843_convert_wlna_to_news_articles.rb
```diff
@@ -0,0 +1,39 @@
...
+  #updating sluggable_string will regenerate the slug and  de-duplicate
+  #if there is already a news article with this slug
+  begin
+    wlna_document.update_attributes(sluggable_string: wlna_document.slug)
+  rescue ActiveRecord::RecordNotUnique
+    puts "NotUnique: #{wlna_document.id}"
```
davidbasalla: So we output if there's a slug already taken for a WLNA. Does anything else happen? The comment suggests that a new slug would be generated for this?

db/data_migration/20170522134843_convert_wlna_to_news_articles.rb
gpeng: The `Document` gets saved when we call `#update_attributes` on line 26. In addition to making the `Document` be the right 'type', setting it here is necessary for the slug to be 'seen' as a duplicate (the unique constraint on `slug` in the DB is scoped to `document_type`).

db/data_migration/20170522134843_convert_wlna_to_news_articles.rb
gpeng: The `ActiveRecord::RecordNotUnique` error is what was being thrown originally when we attempted to change the `document_type` on a `Document` that had a duplicate slug. I caught it here initially so as the migration continued to run and then I was left on subsequent runs in development with only WLNA that had duplicate slugs. When we run this now it should never happen but I left it in to ensure that we don't leave the DB half migrated. It will just leave the 'offenders' which we can address individually. The slug does get changed on the object but it doesn't get persisted as, if this error is raised, it has violated the unique key constraint on the database.

db/data_migration/20170522134843_convert_wlna_to_news_articles.rb
gpeng: I've amended that in 2fd9b6c

davidbasalla:
Looks good, merge away!

PR #3284, opened 2017-06-08
Add an election task to clear ministerial roles

gpeng:
In the event that there is a continuation of current government following the election it may be necessary to clear the role appointments, with the exception of the prime ministerial role, to allow
the content team to rebuild the ministerial list without ending the current government.

This commit adds a rake task to accomplish this.

### DIFF
--- a/lib/tasks/election.rake
+++ b/lib/tasks/election.rake
@@ -27,4 +27,32 @@ namespace :election do
       )
     end
   end
+
+  desc "
+  Removes all current ministerial appointments with the exception
+  of the prime minister
+  Usage
+    rake election:end_ministerial_appointments
+
+    or to specify an end date other than today
+
+    rake election:end_ministerial_appointments[2017-06-09]
+  "
+  task :end_ministerial_appointments, [:end_date] => :environment do |_t, args|
+    appointments = RoleAppointment.current.for_ministerial_roles
+    end_date = Date.today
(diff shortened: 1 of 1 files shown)


lib/tasks/election.rake
```diff
@@ -27,4 +27,22 @@ namespace :election do
...
     end
   end
+
+  task :end_ministerial_appointments, [:end_date] => :environment do |_t, args|
+    appointments = RoleAppointment.current.for_ministerial_roles
+    end_date = Date.today
+    prime_ministerial_role_id = 1
+    begin
+      if args[:end_date]
+        end_date = Date.parse(args[:end_date])
```
davidbasalla: What would the date input look like in the console? "YYYY-MM-DD"? Just wondering if this could cause user input errors. Could we add an example or some error handling? Or will it just be you running this?

bevanloon: won't it fall through the the rescue block if it can't parse the date?

gpeng: Yeah.. that would be the format. The intention is that we won't need to provide a date and `today` will be fine but as we don't want to be deploying tomorrow or over the weekend I wanted to give it a bit more flexibility just in case.

I've added some documentation in b1a854f

gpeng: And yes.. it will fall through to the rescue.

PR #3288, opened 2017-06-08
Remove AB testing logic from world location news pages

davidbasalla:
For https://trello.com/c/jL5RbbKM/180-create-new-news-pages

We've finished the AB test so can remove any logic related to it.

Note: This should only be merged once we finish the AB test for WorldwidePublishing.

### DIFF
--- a/test/functional/world_location_news_controller_test.rb
+++ b/test/functional/world_location_news_controller_test.rb
@@ -3,7 +3,6 @@ require "test_helper"
 class WorldLocationNewsControllerTest < ActionController::TestCase
   include FilterRoutesHelper
   include FeedHelper
-  include GovukAbTesting::MinitestHelpers
 
   should_be_a_public_facing_controller
 
@@ -12,10 +11,7 @@ class WorldLocationNewsControllerTest < ActionController::TestCase
   end
 
   setup do
-    # for this test, we assume that the user is in bucket 'B' of the AB test for
-    # WorldwidePublishingTaxonomy and is viewing one of the test countries
     @world_location = create(:world_location, title: "UK and India", slug: "india")
-    setup_ab_variant("WorldwidePublishingTaxonomy", "B")
   end
 
   view_test "index displays world location title" do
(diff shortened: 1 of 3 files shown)


test/functional/world_location_news_controller_test.rb
```diff
@@ -98,37 +114,28 @@ def assert_featured_editions(editions)
...
 
-  view_test "should redirect to the world location page if the user is in 'A' cohort" do
-    with_variant WorldwidePublishingTaxonomy: "A", assert_meta_tag: false do
-      get :index, world_location_id: @world_location
+    get :index, world_location_id: world_location.id, format: :atom, locale: 'fr'
 
-      assert_redirected_to world_location_path(@world_location)
+    assert_select_atom_feed do
+      with_locale :fr do
+        assert_select_atom_entries([translated_edition])
```
bevanloon: Is the intent here to ensure that there are no atom feed entries in any other locale?

davidbasalla: Exactly... I copied it from the [WorldLocationsController tests](https://github.com/alphagov/whitehall/blob/master/test/functional/world_locations_controller_test.rb#L322-L335) since this page is essentially a trimmed down copy of it

bevanloon: hmm - okay, just sort of peering through those helper methods - I'm not sure this is actually testing that intent; I think it is ensuring that the atom feed exists in `fr` but is not ensuring that **only** `fr` exists?

davidbasalla: I guess the fact that the other edition does not show up in the results is supposed to test that documents with locale `en` don't show up in the feed for `fr` documents? If I change the test from:
```
assert_select_atom_entries([translated_edition])
```
to
```
assert_select_atom_entries([translated_edition, untranslated_edition])
```
(where I assign the other edition to `untranslated_edition`) it fails with `Expected exactly 2 elements matching "feed > entry", found 1..` Do you reckon it needs to be more explicit than that?

bevanloon: aah okay - I see. No I think it's okay 👍

PR #3289, opened 2017-06-09
Add rake task to generate report

elenatanasoiu:
Creates a one off report of how many Whitehall documents were initially published by authors who don't belong to the document's primary organisation.

For: [Trello card](https://trello.com/c/mqzkdPsn/168-report-on-publishing-organisations-for-whitehall-content)

### DIFF
--- a/lib/tasks/report_primary_organisation.rake
+++ b/lib/tasks/report_primary_organisation.rake
@@ -1,0 +1,95 @@
+namespace :report_primary_organisation do
+  desc "A one-off CSV report of how many documents were initially published by authors who don't belong to the document's primary organisation"
+  task :generate_csv, [:user_email, :start_date] => :environment do |_task, args|
+    VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
+
+    if args[:user_email].present? && args[:start_date].present? && args[:user_email].match(VALID_EMAIL_REGEX)
+      require 'csv'
+      require 'ruby-progressbar'
+
+      class CSVReport
+        def author_belongs_to_diff_org?(author, organisation)
+          author.organisation.present? && organisation.present? && organisation.id != author.organisation.id
+        end
+
+        def header_row
+          ['Admin URL',
+           'First published',
+           'Title',
+           'Format',
+           'Lead organisation',
+           'First Author Name',
+           'First Author Email Address',
+           'First Author Organisation',
+           'Last Author Name',
+           'Last Author Email Address',
+           'Last Author Organisation']
+        end
+
+        def get_last_author(edition)
+          edition.published_by
+        end
+
+        def filtered_editions(start_date)
+          @filtered ||= Edition.published.where("first_published_at >= ?", start_date)
+        end
+
+        def progress_bar
+          @progress ||= ProgressBar.create(
+            autostart: false,
+            format: "%e [%b>%i] [%c/%C]"
+          )
+        end
+
+        def generate_csv(start_date = nil)
+          editions = filtered_editions(start_date)
+
+          CSV.generate do |csv|
+            progress_bar.total = editions.size
+            progress_bar.start
+
+            csv << header_row
+            editions.find_each do |e|
+              progress_bar.log("Processing edition ##{e.id}...")
+
+              first_author = e.creator
+              primary_organisation = e.type == 'CorporateInformationPage' ? e.owning_organisation : e.lead_organisations.first
+
+              if author_belongs_to_diff_org?(first_author, primary_organisation)
+                last_author = get_last_author(e)
+                row = []
+                row << Whitehall.url_maker.admin_edition_url(e)
+                row << e.first_published_at.to_date
+                row << e.title
+                row << e.type.titleize
+                row << primary_organisation
+                row << first_author.try(:name) || 'Name missing'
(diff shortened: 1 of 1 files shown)


lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
...
+         "Last Author email address",
+         "Last Author organisation"]
+      end
+
+      def get_last_author(edition)
+        edition.latest_edition.published_by
+      end
+
+      def filtered_editions
+        Edition.where("state = 'published' AND first_published_at >= '2014-01-01'")
```
h-lame: There's almost certainly a scope for published - I don't think it's gonna be more complicated than `state = 'published'` - but it might be.  In general it's better to use scopes that already exist and to speak in the language of the app, rather than the language of the framework.  Not least because if how we decide what is published changes in the future, we'll get that behaviour if we use the scope, but not if we write our own `where` clause.

It would be worth parameterising the date too (there might be a scope for it too, who knows).  Even though this is a "one-off" report for now it's the kind of thing that has a tendency to want to be run multiple times.

lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
+namespace :report_primary_organisation  do
+  desc "A one-off CSV report of how many documents were initially published by authors who don't belong to the document's primary organisation"
+  task :generate_csv, [:user_email] => :environment do |t, args|
```
h-lame: Could be worth checking for the presence of the `user_email` argument and failing fast if it's not there (or even doesn't look like an email address?)

lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
...
+    require 'csv'
+
+    class CSVReport
+      def get_authors(edition)
```
h-lame: Looks like this method isn't used anymore.

lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
...
+
+      def generate_csv
+        CSV.generate do |csv|
+          csv << header_row
+          filtered_editions.first(20).each do |e|
+            puts "Edition: #{e.id}"
+            presenter = DocumentListExportPresenter.new(e)
+
+            first_author = e.creator
+            primary_organisation = if e.type == 'CorporateInformationPage'
```
h-lame: Can you use the `presenter.lead_organisations` method?  It returns a string or an array, but I reckon we could work around that.

h-lame: Alternatively, maybe we don't need the presenter at all?  We're only using it for `admin_url` and `content_type` and those are simple enough that we could inline them.

lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
...
+         "First Author name",
+         "First Author email address",
+         "First Author organisation",
+         "Last Author name",
+         "Last Author email address",
+         "Last Author organisation"]
+      end
+
+      def get_last_author(edition)
+        edition.latest_edition.published_by
```
h-lame: Does `edition.published_by` do the same thing here?

elenatanasoiu: Yes, changed

lib/tasks/report_primary_organisation.rake
elenatanasoiu: Removed

lib/tasks/report_primary_organisation.rake
elenatanasoiu: Added check

lib/tasks/report_primary_organisation.rake
elenatanasoiu: Regrettably there isn't a scope for 'published', only for 'force_published' and 'not_published'.

lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
...
+      end
+
+      def filtered_editions
+        Edition.where("state = 'published' AND first_published_at >= '2014-01-01'")
+      end
+
+      def generate_csv
+        CSV.generate do |csv|
+          csv << header_row
+          filtered_editions.first(20).each do |e|
```
elenatanasoiu: Added `find_each`

lib/tasks/report_primary_organisation.rake
h-lame: There is, it's just "magic" and not defined directly on the class.  We get it from [the transitions gem](https://github.com/troessner/transitions#automatic-scope-generation) which we used to define the state machine in `Edition::Workflow`.

lib/tasks/report_primary_organisation.rake
```diff
@@ -0,0 +1,79 @@
...
+            puts "Edition: #{e.id}"
```
elenatanasoiu: Added progress_bar

lib/tasks/report_primary_organisation.rake
elenatanasoiu: Removed presenter

lib/tasks/report_primary_organisation.rake
elenatanasoiu: Added the scope and parameterized the date

h-lame:
Looks good.

PR #3293, opened 2017-06-15
Send custom dimension including organisation slug to ga

h-lame:
For: https://trello.com/c/svo2ZN6s/131-custom-dimension-send-user-email-domain-to-ga-for-signon-apps

We want to know which departments are using our publishing apps and we've agreed with the performance team that the best way to do this is to send the organisation slug of the logged in user to GA as a custom dimension. If the user doesn't have an organisation we'll send `(not set)` instead.  Many users don't have organisations in signon, but they're mostly local government users that are unlikely to use any of the apps we're interested in measuring at the moment.  As a test we'll be doing this to signon and whitehall to see what kinds of results we get.

This relies on alphagov/govuk_admin_template#152 and will need the first and final commits squashing once that PR is available in a released version of govuk_admin_template.

### DIFF
--- a/test/integration/admin_analytics_test.rb
+++ b/test/integration/admin_analytics_test.rb
@@ -1,0 +1,67 @@
+require 'test_helper'
+require 'capybara/rails'
+
+class AdminAnalyticsTest < ActionDispatch::IntegrationTest
+  include Capybara::DSL
+
+  def log_out
+    GDS::SSO.test_user = nil
+  end
+
+  setup do
+    ENV['GDS_SSO_MOCK_INVALID'] = '1'
+  end
+
+  teardown do
+    ENV.delete('GDS_SSO_MOCK_INVALID')
+  end
+
+  # GA is noisy in tests because all GA calls become console.log so we
+  # don't want it enabled for all tests, use this to selectively turn it on
+  def with_ga_enabled
+    begin
+      GovukAdminTemplate.configure { |c| c.enable_google_analytics_in_tests = true }
+      yield
+    ensure
+      GovukAdminTemplate.configure { |c| c.enable_google_analytics_in_tests = false }
+    end
+  end
+
+  def refute_dimension_is_set(dimension)
+    refute_match(/#{Regexp.escape("GOVUKAdmin.setDimension(#{dimension}")}/, page.body)
+  end
+
+  def assert_dimension_is_set(dimension, with_value: nil)
+    dimension_set_js_code = "GOVUKAdmin.setDimension(#{dimension}"
+    dimension_set_js_code += ", \"#{with_value}\")" if with_value.present?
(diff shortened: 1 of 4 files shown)


cbaines:
Signon contains an organisation for each user, I would imagine that this would meet the requirements similarly to the email address domain, but hopefully better as that is what its meant for. Have you considered using the organisation data?

h-lame:
@cbaines Good question; we were asked specifically to use the email domain by the perf team, so I'll go check with them if the signon org would be as useful for them.

h-lame:
@cbaines one reason might be that of ~6,500 active users, ~4,200 of them don't have an organisation in signon.  Gonna dig some more though.

h-lame:
@cbaines - thanks for asking - yes we can use the org info from signon instead of the email domain.  Lots of users don't have an org, but they're not the users we're interested in (at this point anyway).  If we do become interested in them maybe we'll do something else (or model their orgs in signon properly).  Thanks!

h-lame:
Redone to use org slug instead of email domain as per above comments.

cbaines:
> @cbaines - thanks for asking - yes we can use the org info from signon instead of the email domain. Lots of users don't have an org, but they're not the users we're interested in (at this point anyway). If we do become interested in them maybe we'll do something else (or model their orgs in signon properly). Thanks!

Great, thanks for looking in to this Murray :)

test/integration/admin_analytics_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  def refute_dimension_is_set(dimension)
+    refute_match(/#{Regexp.escape("GOVUKAdmin.setDimension(#{dimension}")}/, page.body)
+  end
+
+  def assert_dimension_is_set(dimension, with_value: nil)
+    dimension_set_js_code = "GOVUKAdmin.setDimension(#{dimension}"
+    dimension_set_js_code += ", \"#{with_value}\")" if with_value.present?
```
elenatanasoiu: So if there's no `with_value` present, we're missing a closing bracket for the previous line ?

Would a better solution be:
`dimension_set_js_code +=  with_value.present? ? ", \"#{with_value}\")" : ")"`
?

It hasn't failed because we do regex matching.

h-lame: The JS will always have a value so it'll always be: `GOVUKAdmin.setDimension(something, "something")`, never `GOVUKAdmin.setDimension(something)`.  We don't always know (or care) what that value will be in our tests so we can't regexp for the whole bit of JS which is why we regexp for a truncated bit of JS when `with_value` is not supplied.

It's far from ideal that we check this by regexping for a chunk of JS on the page, but I couldn't come up with anything better 😞

elenatanasoiu: ah, makes sense :+1:

PR #3296, opened 2017-06-16
Add organisations eligible for world tagging

bevanloon:
Add FCO, UKVI, DIT, DFID, DExEU to the list of organisations allowed to
tag documents in Whitehall admin.

Added some hierarchy to the YAML file containing the list of organisations for whom tagging is allowed so we differentiate between organisations that are in the education related tagging beta and those that will be tagging worldwide content. This is because the logic to decide which document and publication types for worldwide content is slightly different to the education related workflow.

[Trello](https://trello.com/c/6MObnZSc/211-enable-the-whitehall-tagging-interface-for-worldwide-content)

Paired with @davidbasalla

### DIFF
--- a/app/models/edition/taggable_organisations.rb
+++ b/app/models/edition/taggable_organisations.rb
@@ -4,15 +4,15 @@ module Edition::TaggableOrganisations
   def can_be_tagged_to_taxonomy?
     organisations_content_ids = organisations.map(&:content_id)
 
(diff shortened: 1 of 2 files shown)


davidbasalla:
Looks good (but I'm biased as I looked over @bevanloon's shoulder). Tested this in Integration and everything still works!

PR #3297, opened 2017-06-16
Add worldwide tagging

davidbasalla:
Enable the tagging interface for certain content types created by organisations that create the majority of worldwide content. The organisations are FCO, UKVI, DIT, DFID and DExEU. The content types we want to allow for `/world` tagging are Detailed Guide, Publication:Guidance and Publication:Form.

For https://trello.com/b/1h7Tvb3W/govuk-worldwide-publishing-doing

Related to https://github.com/alphagov/whitehall/pull/3296 

Pairing with @bevanloon 

Note that at the time of merging this will have no visibly effect since we have removed the Organisations from the yml that enable this new tagging behaviour. We'll be adding those organisations back in when we're ready to go live. See https://github.com/alphagov/whitehall/pull/3305

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -425,6 +425,10 @@ class Edition < ApplicationRecord
     false
   end
 
+  def must_be_tagged_to_taxonomy?
+    false
+  end
+
   def has_been_tagged?
     api_response = Services.publishing_api.get_links(content_id)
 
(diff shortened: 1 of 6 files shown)


davidbasalla:
Have fixed the merge conflicts, should be ready to go. Don't want to merge this yet though, since it would show the Tagging interface for documents associated to worldwide organisations.

PR #3303, opened 2017-06-20
Make world location base_path less

gpeng:
WorldLocation pages are being converted to worldwide taxonomy and will no longer be renderable. To facilitate this we will now consider them base_path-less content.

This commit updates the schema for this format and modifies the presenter to match.

WorldLocations of type InternationalDelegation will continue to be rendered by Whitehall so the presenter will still present a base_path for those to the publishing API.

The problem with this approach is that it will break pages that are using this link in the UI as it will no longer be usable to generate a URL. We are planning to implement an association between the world location and the taxon that can be expanded in the world location link and used for generating links in the UI. Removing the World Location link type altogether is not feasible currently as not all content will be tagged to the taxon (e.g. news). In the short term we will add a lookup based link helper to government-frontend.

[Trello](https://trello.com/c/6tio72jl/221-split-world-locations-from-international-delegations-in-whitehall)

### DIFF
--- a/app/presenters/publishing_api/world_location_presenter.rb
+++ b/app/presenters/publishing_api/world_location_presenter.rb
@@ -25,14 +25,22 @@ module PublishingApi
         document_type: item.class.name.underscore,
         public_updated_at: item.updated_at,
         rendering_app: Whitehall::RenderingApp::WHITEHALL_FRONTEND,
-        schema_name: "placeholder",
+        schema_name: "world_location",
       )
-      content.merge!(PayloadBuilder::PolymorphicPath.for(item))
+      if international_delegation?
+        content.merge!(PayloadBuilder::PolymorphicPath.for(item))
+      end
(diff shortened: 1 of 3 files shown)


test/unit/world_location_test.rb
```diff
@@ -161,13 +161,19 @@ class WorldLocationTest < ActiveSupport::TestCase
...
   end
 
-  test "has removeable translations" do
-    stub_any_publishing_api_call
+  test "translated world locations appear to only have :en translated locales" do
+    world_location = create(:world_location, translated_into: [:fr, :en])
+    assert_equal [:en], world_location.translated_locales
+  end
+
+  test "translated world locations appear to only have :en available locales" do
```
davidbasalla: Looks like a double up from the test above

gpeng: This is fixed now despite what github seems to think. This test is for `available_locales` (now I've fixed it). The one above is `translated_locales`

PR #3304, opened 2017-06-21
Fix an over-specific scope joining StatisticsAnnouncements to Publications

whoojemaflip:
The `StatisticsAnnouncements.without_published_publication` scope was failing
to include announcements associated with deleted publications, because
the 'publication_id' referenced in the `where` clause is not NULL in that
case.

By changing the WHERE clause to reference the `id` field of the `included`
table, ActiveRecord builds a left-outer join which gives us the resultset
we expect.

[Fixes](https://govuk.zendesk.com/agent/tickets/2214782)

### DIFF
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -113,7 +113,7 @@ class StatisticsAnnouncement < ApplicationRecord
   def self.without_published_publication
     includes(:publication).
       references(:editions).
-      where("publication_id IS NULL || editions.state NOT IN (?)", Edition::POST_PUBLICATION_STATES)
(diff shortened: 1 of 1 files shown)


whoojemaflip:
No tests for any of the methods in this 263 line class :(

h-lame:
You've hurt [this test's](https://github.com/alphagov/whitehall/blob/0b0c5074add1a96ce9884311f7f6b17287f5bd5c/test/unit/statistics_announcement_test.rb) feelings @whoojemaflip, shame on you! (Not that it has tests for the method you changed though).

whoojemaflip:
How I missed that I do not know @h-lame. This test has my unreserved apologies.

PR #3308, opened 2017-06-22
Add redirect checker

davidbasalla:
For https://trello.com/c/Rn6Z8W7d/227-add-automated-checks-for-world-redirects

We'll be redirecting all worldwide related content from `/government/world` to `/world`. This checker class will test that the redirects are correct, by making a http request through the whole stack. The idea is that you can add arrays of hashes with a `url` to check and the expected `redirect_url`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,7 @@ gem 'equivalent-xml', '~> 0.6.0', require: false
 gem 'govuk_ab_testing', '~> 2.2.0'
 gem 'mlanett-redis-lock'
 gem 'deprecated_columns', '0.1.1'
+gem 'faraday'
 
 if ENV['GDS_API_ADAPTERS_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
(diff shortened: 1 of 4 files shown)


davidbasalla:
@gpeng considering the output will only print an error message for those entries, we'll know to ignore those when we run this before the `available_locales` patch has been reverted?

gpeng:
Yep... that's fair enough.

davidbasalla:
@gpeng I've added https://github.com/alphagov/whitehall/pull/3308/commits/e2607c80c9907e5163157a93ef546384bdbaff0c to address your comment, is that what you had in mind?

gpeng:
Yes, that's pretty much it. I was thinking you could maybe break out a method for getting the `(available|translated)_locales` that uses `original_available_locales` if it is there or falls back to `translated_locales` if not as that would always work. That said, the patch is now in the first stuff we'll be deploying so I suspect it will be there throughout the process.

PR #3317, opened 2017-06-27
Add rake task for redirecting WorldLocations

gpeng:
Redirect all the WorldLocation translations. They are no longer supported so are redirected to the non-translated taxon page. This will also cause them to appear unpublished to the publishing API which will remove them from the links of other content.

This also adds a patch to make all world locations appear untranslated. This will remove the translation link on the world location page until we remove the translations themselves.

### DIFF
--- a/test/unit/world_location_test.rb
+++ b/test/unit/world_location_test.rb
@@ -165,13 +165,19 @@ class WorldLocationTest < ActiveSupport::TestCase
     assert_equal [link_6, link_2, link_1, link_4, link_3], world_location.featured_links.only_the_initial_set
   end
 
-  test "has removeable translations" do
-    stub_any_publishing_api_call
+  test "translated world locations appear to only have :en translated locales" do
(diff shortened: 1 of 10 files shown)


test/unit/world_location_test.rb
```diff
@@ -161,13 +161,19 @@ class WorldLocationTest < ActiveSupport::TestCase
...
   end
 
-  test "has removeable translations" do
-    stub_any_publishing_api_call
+  test "translated world locations appear to only have :en translated locales" do
+    world_location = create(:world_location, translated_into: [:fr, :en])
+    assert_equal [:en], world_location.translated_locales
+  end
+
+  test "translated world locations appear to only have :en available locales" do
```
bevanloon: this looks the same as the test above on 164?

gpeng: You are, of course, right. Fixed in 4cc3ed3. Thanks

PR #3322, opened 2017-06-29
Stop publishing headings in HTML publications

fofr:
Depends on https://github.com/alphagov/govuk-content-schemas/pull/631 (tests will fail until this is merged/deployed)
Follow up to https://github.com/alphagov/government-frontend/pull/384

A follow-up PR should check whether the logic to calculate and store headers in Whitehall can be removed entirely.

@nickcolley @andysellick @rubenarakelyan

### DIFF
--- a/app/presenters/publishing_api/html_attachment_presenter.rb
+++ b/app/presenters/publishing_api/html_attachment_presenter.rb
@@ -53,7 +53,6 @@ module PublishingApi
     def details
       {
         body: body,
-        headings: headings,
         public_timestamp: public_timestamp,
         first_published_version: first_published_version?
       }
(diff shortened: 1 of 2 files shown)


nickcolley:
Maybe we should at a 'TODO' so someone in the future knows to investigate removing the underlying code?

fofr:
Tests passing now https://github.com/alphagov/govuk-content-schemas/pull/631 merged.

fofr:
Code to stop heading generation can't be removed until this is completed:
https://trello.com/c/cLVn6spv/902-switch-html-publications-to-use-draft-stack-when-previewing-content

PR #3325, opened 2017-06-30
[Do Not Merge]Remove World Location translation monkey patch

bevanloon:
World Locations have now had all the translations redirected. We should
remove the monkey patch. We also need the available locales on World
Locations so that the World Location News pages are translated.


bevanloon:
Closing in favour of https://github.com/alphagov/whitehall/pull/3329

PR #3329, opened 2017-07-03
Show translations on world location news pages

bevanloon:
World Locations are now Taxonomies rendered by Collections. World Location News Pages are, however, still rendered by Whitehall. World Locations no longer require translations, but World Location News Pages do. The translations for World Location News is still retrieved from the available translations for World Locations. World Locations are sent to Publishing Api as base pathless and we do not want to send all the translations, so we have created a World Location News Translation model to proxy the information to the frontend.

[Trello](https://trello.com/c/HBCgTrfl/269-add-language-toggle-back-to-world-location-news-pages)

## Before
screen shot 2017-07-03 at 09 49 15

### Translated
screen shot 2017-07-03 at 09 49 35

### In English
screen shot 2017-07-03 at 09 52 07

### Country with no translations
screen shot 2017-07-03 at 09 50 18

### DIFF
--- a/app/models/world_location_news_translation.rb
+++ b/app/models/world_location_news_translation.rb
@@ -1,0 +1,11 @@
+class WorldLocationNewsTranslation
+  attr_accessor :translated_locales
+
+  def initialize(translated_locales = [])
+    @translated_locales = Array(translated_locales)
+  end
+
+  def available_in_multiple_languages?
+    translated_locales.count > 1
+  end
+end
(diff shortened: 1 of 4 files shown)


app/models/world_location_news_translation.rb
```diff
@@ -0,0 +1,11 @@
+class WorldLocationNewsTranslation
+  attr_accessor :translated_locales
+
+  def initialize(translated_locales = [])
+    @translated_locales = [].push(translated_locales).flatten
```
gpeng: Is this because `translated_locales` isn't always an array? If so you could use `Array(translated_locales)`.

bevanloon: I think it should always be an array, but I thought some defensive coding might be better than assumptions - fixed in 3b7e179

PR #3331, opened 2017-07-06
Redirect non-en world locations

gpeng:
We no longer support translated world location pages. They were redirected but re-appeared when we moved them from `/government/world/<slug>` to `/world/<slug>`.

This PR adds a rake task to redirect them to their :en equivalent again. 

This happens in two steps. Firstly, we update the publishing API to change the base_path of the document to `/world/<slug>.<locale>`. Then we make the `unpublish` call with the redirect. We can't just redirect via publishing API as that would create a redirect from the original url
not the new `/world` one.

As the presenter no longer exposes a base path for WorldLocations we can't use the existing worker classes for publishing. This PR adds a worker which reinstates the base path by adding it to the payload from the presenter. This worker can be removed once this rake task has been run.

[Trello](https://trello.com/c/LqIOECF3/17-non-english-locale-worldlocation-pages-still-being-rendered-at-world-sluglocale)

### DIFF
--- a/app/workers/world_location_redirect_worker.rb
+++ b/app/workers/world_location_redirect_worker.rb
@@ -1,0 +1,37 @@
+class WorldLocationRedirectWorker < WorkerBase
+  def perform(world_location_id)
+    world_location = WorldLocation.find_by(id: world_location_id)
+    return if world_location.nil?
+
+    base_path_prefix = "/world"
+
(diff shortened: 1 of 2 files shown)


gpeng:
@rubenarakelyan Could you give this another look. Had to take a different approach. I've tested this on integration and it all looks good.

PR #3335, opened 2017-07-07
Delete frontend parts of document collections

fofr:
* Document collections are rendered by government-frontend
* The template and controller aren’t used
* The route is still used to generate the base_path
* Whitehall is still responsible for the `/series` redirect

## Notes

* There's cache-control logic regarding when to clear caches for document collections based on publish times of documents within the collection: https://github.com/alphagov/whitehall/pull/3335/files#diff-beaddba005049d725f1232dc1f4c3dc4L58 – this logic doesn't exist in government-frontend
* There's a presumption that the logic around groups (eg in this test: https://github.com/alphagov/whitehall/pull/3335/files#diff-beaddba005049d725f1232dc1f4c3dc4L30) is covered by the publishing API's presentation of draft links (eg government-frontend rejects empty groups – https://github.com/alphagov/government-frontend/blob/master/app/presenters/document_collection_presenter.rb#L15-L24)


fofr:
Ready to review. `[Do not merge]` flag is for more thorough testing with a branch deploy to Integration.

fofr:
Deployed to integration and tested previewing/creating/editing document collections.

PR #3337, opened 2017-07-12
Fix two email subscription bugs

tuzz:
https://trello.com/c/MXcrS2DC/20-emails-not-being-generated-and-or-sent-for-world-locations

1) Users subscribed to pages with
worldwide_location filters selected were not
receiving emails for editions that only match on
the worldwide location.

2) Users subscribed to pages with a department
filter were not receiving emails for editions that
have no topics (and vice versa).

This change extracts Product and Powerset classes
so they can be unit tested.

**Benchmarks**

Subscription urls are generated combinatorially. This change means we generate more urls than we did before when editions are tagged to worldwide locations. We did some benchmarking:

- it takes ~1.3 second to generate 1000 urls
- about 100 published editions have > 1000 urls
- the worst case is Edition `id=207271` which has 6780 urls
- it takes ~10 seconds to generate the urls for the worst case edition
- previously this edition had ~600 urls so there's a 10x increase in this case

The downstream systems seem to cope with this volume. govuk-delivery maps these `feed_urls` to `topic_ids` so in practice, only a handful of topics are sent to govdelivery. Therefore, I think we should merge/deploy this change.

### DIFF
--- a/lib/whitehall/gov_uk_delivery/subscription_url_generator.rb
+++ b/lib/whitehall/gov_uk_delivery/subscription_url_generator.rb
@@ -94,16 +94,6 @@ module Whitehall
         } if Whitehall::PublicationFilterOption.find_by_search_format_types(edition.search_format_types)
       end
 
-      def all_combinations_of_args(args)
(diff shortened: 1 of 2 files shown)


tuzz:
Thanks, I've moved those classes inside the module and added some benchmark notes.

PR #3343, opened 2017-07-18
Move summary column next to title for CSV export

elenatanasoiu:
For [Trello card](https://trello.com/c/s5lEAwdw/187-add-summary-field-to-whitehall-csv-export)


elenatanasoiu:
There's a danger of confusing tools written to consume these PDF exports, so it's probably safer to leave the summary column at the end, rather than having it next to the title column.

PR #3346, opened 2017-07-20
Create JSON endpoint to show Policy Area

whoojemaflip:
Helps to build up the relationships between Policy Areas and Policies for the Q2 taxonomy work we are doing.

[Trello](https://github.com/alphagov/whitehall/pull/3346)

### DIFF
--- a/app/controllers/admin/topics_controller.rb
+++ b/app/controllers/admin/topics_controller.rb
@@ -1,5 +1,12 @@
 class Admin::TopicsController < Admin::ClassificationsController
 
+  def show
+    respond_to do |format|
+      format.html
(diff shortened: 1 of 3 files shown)


app/controllers/admin/topics_controller.rb
```diff
@@ -1,5 +1,11 @@
 class Admin::TopicsController < Admin::ClassificationsController
 
+  def show
+    if request.format.json?
```
carvil: Do you need to return something like `415 Unsupported Media Type` when the request isn't JSON?

whoojemaflip: Good point. Updated to a more idiomatic Rails thing.

PR #3352, opened 2017-07-24
Make country optional for contacts

rubenarakelyan:
This commit makes the country field optional for contacts. Since the UK was removed as a world location in https://github.com/alphagov/whitehall/pull/3347, addresses in the UK will not have an associated country.

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -11,7 +11,7 @@ class Contact < ApplicationRecord
 
   validates :title, :contact_type, presence: true
   validates :contact_form_url, uri: true, allow_blank: true
-  validates :street_address, :country_id, presence: true, if: -> r { r.has_postal_address? }
+  validates :street_address, presence: true, if: -> (r) { r.has_postal_address? }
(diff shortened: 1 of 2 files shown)


h-lame:
Do we know this is the only other side-effect of having removed the UK from allowed world locations?  Might it be better to re-instate it and then exclude it from tagging / subscribing?  We're gonna have a bunch of contact objects with invalid `country_ids` now.

At some point we might move Whitehall to use the countries register so that WLs only live in one place, and that register does include a UK so we might want to think about a solution to the tagging problem that doesn't involve pretending the UK doesn't exist.

WorldLocations already have an "active" property which is used to make sure we don't link to them in [the /world index template](https://github.com/alphagov/whitehall/blob/master/app/views/world_locations/_world_location.html.erb) - perhaps the tagging stuff can use the same flag?

rubenarakelyan:
@h-lame Good point. I'll close this one; take a look at https://github.com/alphagov/whitehall/pull/3353 which re-adds the UK with the same ID but disallows tagging to inactive world locations which should solve both issues.

PR #3353, opened 2017-07-24
Disallow tagging to inactive world locations

rubenarakelyan:
This commit changes the world location tagging logic to disallow tagging to inactive world locations. This allows us to re-add the United Kingdom to enable its use as a country for contacts without allow tagging to it.

### DIFF
--- a/app/helpers/admin/taggable_content_helper.rb
+++ b/app/helpers/admin/taggable_content_helper.rb
@@ -101,7 +101,7 @@ module Admin::TaggableContentHelper
   # of the array consists of two values: the location name and its ID
   def taggable_world_locations_container
     Rails.cache.fetch(taggable_world_locations_cache_digest, expires_in: 1.day) do
-      WorldLocation.ordered_by_name.map {|w| [w.name, w.id] }
+      WorldLocation.ordered_by_name.where(active: true).map { |w| [w.name, w.id] }
     end
   end
 
--- a/test/unit/helpers/admin/taggable_content_helper_test.rb
+++ b/test/unit/helpers/admin/taggable_content_helper_test.rb
@@ -147,13 +147,14 @@ class Admin::TaggableContentHelperTest < ActionView::TestCase
     ], taggable_statistical_data_sets_container
   end
 
-  test '#taggable_world_locations_container returns an array of label/ID pairs for all world locations' do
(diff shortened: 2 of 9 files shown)


app/helpers/admin/taggable_content_helper.rb
```diff
@@ -101,7 +101,7 @@ def taggable_statistical_data_sets_container
   # of the array consists of two values: the location name and its ID
   def taggable_world_locations_container
     Rails.cache.fetch(taggable_world_locations_cache_digest, expires_in: 1.day) do
-      WorldLocation.ordered_by_name.map {|w| [w.name, w.id] }
+      WorldLocation.ordered_by_name.select(&:active).map { |w| [w.name, w.id] }
```
tuzz: Should we make this a `where(active: true)`? Although I'm guessing there aren't many `WorldLocation` records to load into memory.

test/unit/helpers/admin/taggable_content_helper_test.rb
```diff
@@ -147,13 +147,14 @@ class Admin::TaggableContentHelperTest < ActionView::TestCase
...
 
-  test '#taggable_world_locations_container returns an array of label/ID pairs for all world locations' do
-    location_a = create(:world_location, name: 'Andora')
-    location_c = create(:world_location, name: 'Croatia')
-    location_b = create(:world_location, name: 'Brazil')
+  test '#taggable_world_locations_container returns an array of label/ID pairs for all active world locations' do
+    location_a = create(:world_location, name: 'Andorra', active: true)
+    location_c = create(:world_location, name: 'Croatia', active: true)
+    location_b = create(:world_location, name: 'Brazil', active: true)
+    create(:world_location, name: 'United Kingdom', active: false)
```
tuzz: 👍

app/helpers/admin/taggable_content_helper.rb
rubenarakelyan: Yes good point, changed

rubenarakelyan:
All the taggings to the UK were removed in https://github.com/alphagov/whitehall/pull/3347. Contacts are still tagged to the UK but since they can't be edited at the moment (until this change is deployed), there shouldn't be any data inconsistency issues. Also, ID 202 which is being used in the data migration here is the same ID that the UK had before we removed it.

PR #3354, opened 2017-07-24
[DO NOT MERGE] Add special route for /government/uploads

cbaines:
To route requests through to whitehall-admin. This is currently done
using NGinx configured through govuk-puppet.

Adding a specific route makes it clearer that these requests are
handled by whitehall-admin, rather than whitehall-frontend.

I haven't tested this locally, I first wanted to put this out there to see if anyone else through it was a good idea.

### DIFF
--- a/lib/whitehall/rendering_app.rb
+++ b/lib/whitehall/rendering_app.rb
@@ -1,7 +1,7 @@
 module Whitehall
   module RenderingApp
-    WHITEHALL_ADMIN = 'whitehall-admin'
-    WHITEHALL_FRONTEND = 'whitehall-frontend'
-    GOVERNMENT_FRONTEND = 'government-frontend'
+    WHITEHALL_ADMIN = 'whitehall-admin'.freeze
+    WHITEHALL_FRONTEND = 'whitehall-frontend'.freeze
+    GOVERNMENT_FRONTEND = 'government-frontend'.freeze
   end
 end
--- a/lib/whitehall/rendering_app.rb
+++ b/lib/whitehall/rendering_app.rb
@@ -1,5 +1,6 @@
 module Whitehall
   module RenderingApp
+    WHITEHALL_ADMIN = 'whitehall-admin'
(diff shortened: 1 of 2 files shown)


lib/whitehall/rendering_app.rb
```diff
@@ -1,5 +1,6 @@
 module Whitehall
   module RenderingApp
+    WHITEHALL_FRONTEND = 'whitehall-admin'
```
rubenarakelyan: Did you mean `WHITEHALL_ADMIN`?

cbaines: Yep, I've updated the branch.

cbaines:
This requires a change in the govuk-content-schemas: https://github.com/alphagov/govuk-content-schemas/pull/648

danielroseman:
As noted on the content-schemas PR, I don't think this is worth spending time on, as everything will be moving to S3 at some point. Making changes now will probably just confuse things.

cbaines:
> As noted on the content-schemas PR, I don't think this is worth spending time on, as everything will be moving to S3 at some point. Making changes now will probably just confuse things.

Ok, I am aware of some stuff happening with the asset-manager app, but I'm not aware of the connection to Whitehall. Is there somewhere I can read up on this?

danielroseman:
Here: https://docs.google.com/document/d/107UCPLM3grUTtoZ6nGIlA6d-QEe0SClxpnaixci_TEg/edit#heading=h.c3ulv3w2exzh

Basically, the idea is to make asset-manager use S3 then move Whitehall to use asset-manager, so we only have one way to upload and serve media.

cbaines:
> Basically, the idea is to make asset-manager use S3 then move Whitehall to use asset-manager, so we only have one way to upload and serve media.

Thanks for the link. I was thinking that this change might be a good stepping stone to a point where there is some kind of proxying or redirect situation for requests to content under /government/uploads. But, thinking about it, if redirects are created, then anything past the router isn't relevant, and the reverse proxying situation is too unconstrained for me to work out if this would help or not.

PR #3357, opened 2017-07-26
Send update type with put content

thomasleese:
Sending it on publish has been deprecated in favour of including it when sending a put content.

[Trello Card](https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk)


thomasleese:
@steventux Yeah, the next stage of this will be to set up an alert which checks to see if we missed anything.

PR #3358, opened 2017-07-27
Populate the analytics meta tags for finders

carvil:
This commit makes sure we set the `@content_item` instance variable with
the content store response related to the finder the user is on, so that
the analytics meta tag component gets populated correctly and all
necessary meta tags added to the page.

This will allow us to identify what page we are on in static and
simplify the code that looks for the number of links and sections on the
page.

The logic to populate the component is already in the layout: https://github.com/alphagov/whitehall/blob/master/app/views/layouts/_frontend_base.html.erb#L13-L14

### DIFF
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -10,6 +10,11 @@ class PublicationsController < DocumentsController
 
     respond_to do |format|
       format.html do
+        @content_item = Whitehall
+          .content_store
+          .content_item("/government/publications")
+          .to_hash
+
         @filter = PublicationFilterJsonPresenter.new(
           @filter, view_context, PublicationesquePresenter
         )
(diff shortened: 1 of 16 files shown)


app/controllers/publications_controller.rb
```diff
@@ -9,6 +9,11 @@ def index
     @filter.publications_search
 
     respond_to do |format|
+      @content_item = Whitehall
```
whoojemaflip: Should this be in the `format.html` block?

carvil: It wouldn't be incorrect to be outside, but since this is only used in ERB partials to add meta tags to the `<head>` section of the page, we don't need to fetch it for the other media types.

By adding it for all media types, I would also need to change a lot more tests to stub the content store call, so I decided to keep it isolated.

Does that make sense?

whoojemaflip: sorry, I should have been clearer. I think that in this instance, the `@content_item` initialisation has been pasted into the wrong place!

carvil: Ah, sorry, my mistake, yes it should :) Will fix it now.

PR #3362, opened 2017-07-27
Preview html attachments using the draft stack

tuzz:
https://trello.com/c/cLVn6spv/31-switch-html-publications-to-use-draft-stack-when-previewing-content

---

The HtmlAttachment#url method needs to return the full URL
in preview mode because the links need to go to draft-origin.
When not previewing (.e.g presenting the attachment to send
to Publishing API), we can use a relative path because that'll
be served alongside the edition on GOV.UK.

---

Previewing is now handled by the draft stack so we
can remove this code from Whitehall. We leave the
routes so we can use them as path helpers, but set
the controller action to _#_ to indicate Whitehall
isn’t able to serve those requests.


rubenarakelyan:
It would be good to get someone from the old Core team to take a look at this since they should have more context about HTML Publications and their migration to publishing-api.

tuzz:
I think Graham's back on Monday so I'll leave this for now and run it past him.

tuzz:
After some manual testing, I noticed we also need to change the html attachment links on published editions as well as drafts. These should point to GOV.UK in production and `www-origin` in integration/staging.

The `documents/_attachment.html.erb` partial is used when we present the code to send downstream to Publishing API. We need to make sure we use a relative path in this context as this is how it behaves now. Otherwise, we should use absolute paths to force the user's browser away from whitehall-admin and to `GOV.UK` or `www-origin`.

We did some work to remove the publication/statistics frontend code which should hopefully make this simpler because there are fewer things using that partial.
(https://trello.com/c/kfyYY4tf/73-9-publications-migration-final-tasks-deploy-2-05-days)

PR #3364, opened 2017-07-31
Republish all published corporate information pages

rubenarakelyan:
This commit republishes all published corporate information pages to apply the change made in https://github.com/alphagov/whitehall/pull/3363.


rubenarakelyan:
I copied another data migration that did the same thing :)

PR #3136, opened 2017-03-24
Remove publications frontend code

mgrassotti:
https://trello.com/c/kfyYY4tf/368-9-publications-migration-final-tasks-deploy-2-0-5-days

Remove any publications frontend code in Whitehall that cannot be fact-checked

PR #3370, opened 2017-08-02
Remove publication templates

tuzz:
https://trello.com/c/kfyYY4tf/73-9-publications-migration-final-tasks-deploy-2-05-days

The rendering of these pages is now handled by government-frontend. This work removes this code from Whitehall and re-works some of the tests to equivalently test the things where assertions were previously made against these views.

It's probably easiest to review this commit-by-commit and read the commit messages.

The supersedes https://github.com/alphagov/whitehall/pull/3136 which deleted more code than was necessary.


rubenarakelyan:
🎉 

Nice

screen shot 2017-08-03 at 09 28 10

app/views/publications/show.html.erb
```diff
@@ -1,51 +0,0 @@
...
-
-  <div>
-    <div class="inner-block">
-      <%= render partial: "document_summary", locals: { document: @document } %>
-    </div>
-  </div>
-
-  <div class="block-2 heading-block">
-    <div class="inner-block">
-      <%= render partial: "documents/attachment_full_width",
```
fofr: Can this partial also be removed? Looks like it was only referenced on publications and statistics.

fofr:
Just the one thing I've found that could also be removed.

:+1: :100:

app/views/publications/show.html.erb
tuzz: Removed and rebased.

tuzz:
Deployed to integration. Everything looks ok. Merging.

PR #3379, opened 2017-08-07
Add speech type explanation to payload

davidbasalla:
The content schema field `speech_type_explanation` for Speeches has not been populated from Whitehall. This commit updates the SpeechPresenter to send the relevant data to the Publishing Platform. With this fix, the speech pages will render the explanation as part of the 'Delivered on:' text:

screen shot 2017-08-07 at 15 05 00

For https://trello.com/c/fyS5Hct7/62-investigate-re-adding-speech-type-to-template


davidbasalla:
@gpeng thanks, have updated the first commit to add in the change for the sync check

PR #3383, opened 2017-08-08
Upgrade sprockets rails

h-lame:
This should unblock two stories: 

1. upgrading to rails 5
2. adding SRI attributes to assets (see: https://trello.com/c/8o2zFPsC/147-enable-subresource-integrity-sri-on-whitehall-s)

To do the sprockets-rails upgrade we need to replace the [test_track JS runner](https://github.com/alphagov/test_track) with an alternative that works with sprockets-rails 3.x.  We've chosen [teaspoon](https://github.com/jejacks0n/teaspoon).

Our requirements are that the alternative:

1. supports qunit and sinon (the JS testing framework our tests are written in)
2. uses the asset pipeline for tests and source (some of our source files require compilation, and one of our test files does too).
3. works with sprockets-rails 3.x
4. supports rails 4 and 5 (to make upgrades easier)

Teaspoon is the only one I found that fits the bill.  It's not clear if it's under active development so we may find ourselves in the same situation again at some future date where this dependency blocks us again.  We may want to consider porting the JS tests to jasmine so we can use jasmine-rails which is what most of our other apps use for JS testing in rails.

Best reviewed commit-by-commit as there are further notes on some of the things we've done as we go (it's not a straightforward set of changes).


h-lame:
I've given this a poke on integration and all the assets seem to compile and be delivered properly, so I'm gonna merge this to unblock the other work.

PR #3388, opened 2017-08-08
[DO NOT MERGE] Fix "View on website" link for world location news

binaryberry:
The link in the admin was still pointing to the world location page
(www.gov.uk/world/country) instead of the world location article page
(www.gov.uk/world/country/news).

cc @gpeng 

https://govuk.zendesk.com/agent/tickets/2294928

### DIFF
--- a/app/views/admin/world_locations/show.html.erb
+++ b/app/views/admin/world_locations/show.html.erb
@@ -7,7 +7,7 @@
       <span class="name"><%= @world_location.name %></span>
       &ldquo;<span class="title"><%= @world_location.title %></span>&rdquo;
     </h1>
-    <p><%= view_on_website_link_for @world_location %></p>
+    <p><%= link_to "View on website", "/world/#{@world_location.name}/news" %></p>
   </div>
 
   <section class="world-location-details">


app/views/admin/world_locations/show.html.erb
```diff
@@ -7,7 +7,7 @@
       <span class="name"><%= @world_location.name %></span>
       &ldquo;<span class="title"><%= @world_location.title %></span>&rdquo;
     </h1>
-    <p><%= view_on_website_link_for @world_location %></p>
```
fofr: Should `view_on_website_link_for` behave differently when passed a world_location?

davidbasalla: I think the problem is that we're after the URL for the WorldLocationNews page (rather than the WorldLocation), which isn't really backed up by a model. Making the `view_on_website_link_for` method support this specific behaviour might be tricky...

app/views/admin/world_locations/show.html.erb
```diff
@@ -7,7 +7,7 @@
...
+    <p><%= link_to "View on website", "/world/#{@world_location.name}/news" %></p>
```
davidbasalla: I think this needs to be `@world_location.slug` in order return the correct URL. Calling `.name` for example would return "São Tomé and Principe" whereas `.slug` would return "sao-tome-and-principe"

davidbasalla:
Added 'Do not merge' tag as we might want to do a bit more with this, see https://trello.com/c/bkdPtKOI/110-link-to-world-location-news-page-from-whitehall-admin-ui

gpeng:
Closed in favour of https://github.com/alphagov/whitehall/pull/3425

PR #3393, opened 2017-08-09
Add named params to tests

rubenarakelyan:
This commit adds named params to all the tests that are missing them as part of the upgrade to Rails 5.

Trello: https://trello.com/c/eg61wStW/4-upgrade-whitehall-to-a-supported-version-of-rails


davidbasalla:
Nice 👍 Merge when 🍏

rubenarakelyan:
Looks like this only works in Rails 5 so we can't split it out.

PR #3396, opened 2017-08-10
Update rummager with world news stories

gpeng:
This PR adds a data migration that sends all of the `NewsArticle` of type `WorldNewsStory` to rummager and removes the entries for their previous base path (these were recently migrated from
`WorldLocationNewsArticle` so their path has changed).

### DIFF
--- a/db/data_migration/20170810111203_update_world_news_stories_in_rummager.rb
+++ b/db/data_migration/20170810111203_update_world_news_stories_in_rummager.rb
@@ -1,0 +1,21 @@
+world_news_story_type_id = 4
+
+editions = Edition
+  .where(state: ["withdrawn", "published"])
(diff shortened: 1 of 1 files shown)


db/data_migration/20170810111203_update_world_news_stories_in_rummager.rb
```diff
@@ -0,0 +1,21 @@
+world_news_story_type_id = 4
+
+editions = Edition
```
davidbasalla: Could you use `NewsArticle.by_subtype(NewsArticleType::WorldNewsStory)` here as well? Probably not a biggie as I assume the outcome would be the same

gpeng: Yep. Could have done by the look of it. I didn't know about that method.

PR #3408, opened 2017-08-17
Make sure attachments always use the public asset host

h-lame:
Attachments are always served by the whitehall-backend machines because
the NFS mount they are stored on is not available to the
whitehall-frontend machines.  This means any rendered page (such as CSV
previews) is rendered with by that rails app, and gets the admin assets
host.  The CORS settings for the admin asset hosts means that the browser
will reject those assets as not being available to pages on the public
site and so styling and functionality breaks.  We change the asset_host
for production.rb to use a proc that makes sure to always use the public
asset host for any attachments paths (those with system/uploads in the
path) and this should fix the problem.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -52,7 +52,17 @@ Whitehall::Application.configure do
   config.cache_store = :dalli_store
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server
-  config.action_controller.asset_host = Whitehall.asset_root
+  # Make sure we always use the public_asest_host for uploads as these are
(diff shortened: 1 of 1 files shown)


tuzz:
LGTM

For posterity, this PR was created as part of this incident:
https://docs.google.com/document/d/15xTawGPq1UFJ5e6Zi6PCJciZVZrpUYuXAXbiFm921PA/edit#

This may be something for the incident review, but perhaps we could look at extracting these `asset_host` procs into a class and unit test that to mitigate future regression. We can potentially do this later.

PR #3412, opened 2017-08-18
Add data migration to convert unknown news articles

davidbasalla:
The NewsArticleType Unknown is a legacy concept. These are NewsArticles that were never categorised into one of the other NewsArticleTypes. It's not possible to create one or even edit existing ones without choosing another NewsArticleType. These are surfaced on the frontend via the [Announcements page](www.gov.uk/government/announcements) and do get shown when filtering by PressRelease or NewsStory. Therefore it seems acceptable to convert these into those two types. This will enable us to remove the ‘Unknown’ NewsArticleType as well as some code that goes with this special case (in a future PR).

Related to https://trello.com/c/K1OAcdZ1/118-documents-with-an-unknown-newsarticletype-are-interfering-with-matching

### DIFF
--- a/db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
+++ b/db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
@@ -1,0 +1,21 @@
+begin
+  # Update NewsArticles in Whitehall
+  unknown_news_articles = NewsArticle.by_subtype(NewsArticleType::Unknown)
+  document_ids = unknown_news_articles.map { |n| n.document_id }.uniq
+  press_releases = unknown_news_articles.includes(:document).select { |n| n.slug.include?("press-release") }
+  news_stories = unknown_news_articles - press_releases
+
+  press_releases.update_all(news_article_type_id: NewsArticleType::PressRelease.id)
+  press_releases.where(minor_change: false, change_note: nil).update_all(change_note: "This news article was converted to a press release")
+
+  news_stories.update_all(news_article_type_id: NewsArticleType::NewsStory.id)
+  news_stories.where(minor_change: false, change_note: nil).update_all(change_note: "This news article was converted to a news story")
+
+  # Republish updated NewsArticles to PublishingAPI
+  document_ids.each do |id|
+    PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", id)
+    print "."
+  end
+rescue NameError => e
+  puts "Can't apply data migration: #{e.message}"
+end


db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    if !news_article.minor_change && news_article.change_note.nil?
+      news_article.change_note = "This NewsArticle was converted to a News Story"
+    end
+
+    news_article.save!
+  end
+
+  # Republish updated NewsArticles to PublishingAPI
+  document_ids = unknown_news_articles.map { |n| n.document.id }
```
gpeng: You should probably `uniq` these as there will be duplicates. Also if you use `n.document_id` it'll might loading the `Document`.

db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  ids.each do |id|
```
gpeng: `document_ids` here not `ids`

db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
```diff
@@ -0,0 +1,41 @@
...
+  unknown_news_articles = NewsArticle.by_subtype(NewsArticleType::Unknown)
+  document_ids = unknown_news_articles.map { |n| n.document_id }.uniq
+  press_releases = unknown_news_articles.includes(:document).select { |n| n.slug.include?("press-release") }
+  news_stories = unknown_news_articles - press_releases
+
+  press_releases.each do |news_article|
+    news_article.news_article_type = NewsArticleType::PressRelease
+
+    if !news_article.minor_change && news_article.change_note.nil?
+      news_article.change_note = "This NewsArticle was converted to a Press Release"
```
gpeng: Should this be 'news article' and 'press release'?

db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
```diff
@@ -0,0 +1,41 @@
+begin
+  # Update NewsArticles in Whitehall
+  unknown_news_articles = NewsArticle.by_subtype(NewsArticleType::Unknown)
+  document_ids = unknown_news_articles.map { |n| n.document_id }.uniq
+  press_releases = unknown_news_articles.includes(:document).select { |n| n.slug.include?("press-release") }
+  news_stories = unknown_news_articles - press_releases
+
+  press_releases.each do |news_article|
```
gpeng: Could we do a couple of `update_all`s here instead. I think that would just issue one SQL statement and avoid the validation that you're skipping on 18. Maybe `press_releases.update_all(news_article_type_id: NewsArticleType::PressRelease.id)` and then `press_releases.where(minor_change: false, change_note: nil).update_all(change_note: <etc>)`? No problem if you don't want to make this change but it would be much faster if it is taking a long time to run.

db/data_migration/20170818104227_convert_unknown_news_articles_to_valid_news_articles.rb
```diff
@@ -0,0 +1,41 @@
...
+      news_article.change_note = "This NewsArticle was converted to a Press Release"
+    end
+
+    # set validate to `false` here as otherwise we get the following errors:
+    # - News article type cannot be modified when edition is in the published state
+    # - Policy area at least one required
+    news_article.save(validate: false)
+  end
+
+  news_stories.each do |news_article|
```
gpeng: Same two points here about the `update_all` and the text in the change note.

davidbasalla:
Thanks @gpeng, have updated and rebased

PR #3413, opened 2017-08-18
Switch to a more reliable way of getting draft taxons

cbaines:
Rather than relying on equality of the items within the arrays,
explicitly reject any items which appear in the list of published
taxons, matching the taxons up using the content id.

This fixes an issue where the Childcare and Parenting taxonomy branch
appears as both draft and live within the Whitehall interface.

https://trello.com/c/AledQjCF/168-whitehall-tagging-interface-shows-childcare-and-parenting-in-both-live-and-draft-topics


tijmenb:
Can you add a test for this?

cbaines:
> Can you add a test for this?

Yep, I'll have a look...

PR #3417, opened 2017-08-21
Document Smart Answers dependencies

alecgibson:
We recently had an incident where the Worldwide API was updated in such
a way that its contract with the Smart Answers application was broken.

This change adds a couple of comments documenting this dependency in the
README.


alecgibson:
[Document Smartanswers' external dependencies](https://trello.com/c/mYBWEYtb/723-document-smartanswers-external-dependencies)

PR #3419, opened 2017-08-21
Removed deprecated need_ids column from Editions

elenatanasoiu:
This is the final step in removing any reference to using the Needs API in `whitehall`. The actual work of disconnecting Needs API and converting needs into content items that can be fetched from the Publishing API was done in a previous PR: https://github.com/alphagov/whitehall/pull/3071.

This commit is only concerned with dropping the now unused `need_ids` column from the `Editions` table.


h-lame:
Merge when green

PR #3425, opened 2017-08-24
Update world location news view link

gpeng:
`WorldLocation`s are no longer rendered explicitly and are now represented as world location news pages. This commit updates the 'View on website' link to point to this URL.

[Trello](https://trello.com/c/bkdPtKOI/129-link-to-world-location-news-page-from-whitehall-admin-ui)


davidbasalla:
Does this update the link for all tabs? I vaguely remember that for the feature tabs, the code for adding the link was somewhere else... also would it handle the translations?

gpeng:
It just updates one place. Looks like the others, at least the ones I looked at, were already pointing to news. The one that this changes doesn't get localised either.

PR #3427, opened 2017-08-25
Bump slimmer to 11.0.2

davidbasalla:
There was a problem with upgrading to Slimmer 11.0.1, see alphagov/slimmer#206. The issue was fixed in Slimmer 11.0.2 so we upgrade to that.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -192,7 +192,7 @@ GEM
     htmlentities (4.3.4)
     http-cookie (1.0.3)
       domain_name (~> 0.5)
-    i18n (0.8.1)
+    i18n (0.8.6)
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'validates_email_format_of'
 gem 'friendly_id', '~> 5.2.1'
 gem 'babosa', '1.0.2'
 gem 'nokogiri', '~> 1.7.2'
-gem 'slimmer', '~> 11.0'
+gem 'slimmer', '~> 11.0.2'
```
gpeng: We don't need to change this. `bundle update slimmer` will pick up the new `x.x.n` version.

PR #3428, opened 2017-08-29
Push html attachments on save

davidbasalla:
Save the edition upon adding/updating an HTML attachment which will trigger it updating the PublishingAPI. This should enable users to preview an attachment after adding it (previously a user would have to save the edition first before being able to preview the attachment).

Effectively reverts https://github.com/alphagov/whitehall/pull/3157 but it's not entirely clear to me why this functionality (of pushing to the publishing api after adding an attachment) was removed in the first place. I tried doing a straight revert but other parts of the code have changed too much since. 

For https://trello.com/c/7SU9Uf54/130-html-attachment-previewing-problems

### DIFF
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -16,7 +16,7 @@ class Admin::AttachmentsController < Admin::BaseController
   def new; end
 
   def create
-    if attachment.save
+    if save_attachment
       redirect_to attachable_attachments_path(attachable), notice: "Attachment '#{attachment.title}' uploaded"
     else
       render :new
@@ -25,7 +25,7 @@ class Admin::AttachmentsController < Admin::BaseController
 
   def update
     attachment.attributes = attachment_params
-    if attachment.save
+    if save_attachment
       message = "Attachment '#{attachment.title}' updated"
       redirect_to attachable_attachments_path(attachable), notice: message
     else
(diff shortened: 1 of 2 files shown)


app/controllers/admin/attachments_controller.rb
```diff
@@ -174,4 +174,12 @@ def handle_duplicate_key_errors_caused_by_double_create_requests(exception)
       raise
     end
   end
+
+  def save_attachment
+    if attachable_is_an_edition?
+      attachment.save && Whitehall.edition_services.draft_updater(attachable).perform!
```
gpeng: Should we just include `PublishesToPublishingApi` on `HtmlAttachment` and then we can just `save`?

I think there's also potentially an issue, albeit a bit of an edge case, with this. If the user saves the attachment on a new draft and then cancels creating the draft it will have been already saved to the publishing API. Whitehall will let users create something else at that base path but publishing API will then conflict.

gpeng: Actually, having discussed this with @davidbasalla offline that issue isn't an issue. The draft edition would have to exist already to get to this point.

tuzz:
The publication QA document[1] contains this step:
> Click modify attachments, add an html attachment and save the document

Are we sure it's not intended behaviour that you need to save the draft edition first?

[1](https://docs.google.com/spreadsheets/d/1iJ-65U9D_n-FFl0zkxXsUR-OfRHr6P9OlSgwWZQNnzU/edit#gid=0)

tuzz:
Discussed with @davidbasalla offline. We're going to update the QA document.

PR #3435, opened 2017-09-04
Fix PDF thumbnail timeout

alecgibson:
Using `Open3` inside a `Timeout` doesn’t quite work as expected, and
will not actually time out.

This is because `Open3` has an `ensure` block that catches the raised
`Timeout::Error`, and renders the `Timeout` useless.

For more information, see: https://bugs.ruby-lang.org/issues/4681

This change removes the use of `Open3`, which means that the `Timeout`
will actually time out in time for a user to not receive a `504` error.

### DIFF
--- a/app/uploaders/attachment_uploader.rb
+++ b/app/uploaders/attachment_uploader.rb
@@ -1,7 +1,5 @@
 # encoding: utf-8
 
-require 'open3'
-
 class AttachmentUploader < WhitehallUploader
   PDF_CONTENT_TYPE = 'application/pdf'
   INDEXABLE_TYPES = %w(csv doc docx ods odp odt pdf ppt pptx rdf rtf txt xls xlsx xml)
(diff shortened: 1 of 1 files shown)


app/uploaders/attachment_uploader.rb
```diff
@@ -46,8 +46,13 @@ def pdf?(file)
...
 
   def get_first_page_as_png(width, height)
+    # Open3 has an `ensure` block that catches the `Timeout::Error` raised by `Timeout`, which basically nullifies
+    # the Timeout. An ugly (but one-line) workaround is to put a second `Timeout` around the first.
+    # See: https://bugs.ruby-lang.org/issues/4681
     output, status = Timeout.timeout(THUMBNAIL_GENERATION_TIMEOUT) do
-      Open3.capture2e(pdf_thumbnail_command(width, height))
+      Timeout.timeout(THUMBNAIL_GENERATION_TIMEOUT) do
+        Open3.capture2e(pdf_thumbnail_command(width, height))
+      end
```
tuzz: Wow, nice find! 🎉 

I wonder if this would be simpler if we avoided using `Open3` all together and did something like:

```ruby
output = `#{pdf_thumbnail_command(width, height)`
status = $?.exitstatus
```

That way we wouldn't need nested timeouts, I think?

app/uploaders/attachment_uploader.rb
```diff
@@ -46,8 +46,13 @@ def pdf?(file)
...
     end
     unless status.success?
```
gpeng: Could we notify Sentry here and in the `rescue` block? I think there's possibly other issues as I was looking at a problematic file the other day and the `gs` shell bit ran really quickly but it still timed out when the ruby code was involved. Might be fixed by getting rid of `Open3` but it would be good to have more visibility.

alecgibson: How do we notify Sentry? Still using Airbrake? Do you have an example somewhere I can copy?

app/uploaders/attachment_uploader.rb
alecgibson: Wow. Didn't even know this was a thing - why were we using `Open3` in the first place then? I'll try it out on integration. Would be great to bin double timeouts.

alecgibson: @tuzz done - feels much nicer!

app/uploaders/attachment_uploader.rb
alecgibson: @gpeng - I think I've added the change, can you please look it over? Is there any information in particular you want added to the log? (eg I don't know if we have things like file size available?)

PR #3437, opened 2017-09-06
Redirect routes for duplicate 'About us' corporate info pages

tuzz:
https://trello.com/c/9JFXOc5y/145-worldwide-organisations-are-not-updated-when-their-about-us-summary-is-changed-added?menu=filter&filter=label:Support

Whitehall has some special redirect logic around
the `/about` corporate info pages for organisations
and worldwide organisations. These pages are now
served by government-frontend and so we can't
rely on this redirect logic any more and need to
set up these routes in a more conventional way.

These `/about` pages are a bit of a mess. Some
orgs. have redirects that go to `/about/about`
which then falls through to Whitehall which then
serves duplicate content at this path. Many of
them have no content at the `/about` paths.

We can sort out this mess by updating or creating
redirects via Publishing API for these `/about` and
`/about/about` paths that redirect to the canonical
'About us' paths. This is much more conventional
approach that gives visibility of these redirects
to Publishing API, Content Store and Router.

### DIFF
--- a/db/data_migration/20170906102634_redirect_cips.rb
+++ b/db/data_migration/20170906102634_redirect_cips.rb
@@ -1,0 +1,63 @@
+class RedirectCIPs
+  def redirect_all
+    about_us = ::CorporateInformationPageType::AboutUs
+
+    about_pages = ::CorporateInformationPage
+      .where(corporate_information_page_type_id: about_us.id)
+
+    about_pages.find_each do |page|
+      page.available_locales.each do |locale|
+        redirect_one(page, locale)
+      end
+    end
+  end
+
+  def redirect_one(page, locale)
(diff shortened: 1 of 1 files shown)


db/data_migration/20170906102634_redirect_cips.rb
```diff
@@ -0,0 +1,64 @@
+class RedirectCIPs
+  def redirect_all
+    about_us = CorporateInformationPageType::AboutUs
+
+    published_about_pages = CorporateInformationPage
+      .published
```
gpeng: There are a few `withdrawn` ones too. Do we need to include those? There are 5 as far as I can see e.g. https://www.gov.uk/government/organisations/oil-and-gas-authority/about/about

PR #3439, opened 2017-09-07
Republish speeches with world locations

tuzz:
We weren't sending world locations through for speeches.
We fixed in this #3438.


tuzz:
Tested on integration, speeches now contain world locations, e.g.
https://www-origin.integration.publishing.service.gov.uk/api/content/government/speeches/eu-transport-council--15?asd

PR #3442, opened 2017-09-08
Send topical_events in the links hash for consultations

tuzz:
https://trello.com/c/4y8S9j1S/85-transition-whitehalls-email-sending-to-use-email-alert-api

Otherwise we don’t trigger email alerts when a
consultation tagged to a topical event is
published.


db/data_migration/20170908110927_republish_consultations_with_topical_events.rb
```diff
@@ -0,0 +1,6 @@
+document_ids = Consultation.find_each.map { |c| c.document_id if c.topical_events.present? }.compact
```
gpeng: You might want to `uniq` these too... Probably not a big issue as there aren't that many but you'll get duplicate document ids in here I think.

PR #3445, opened 2017-09-13
Add people drop down to announcements

bevanloon:
Adds an extra filter to the `/government/announcements` search criteria.

* Adds the drop down to the frontend
* Sends the param through to the rummager search
* Update the atom and email urls to include the new param
* Adds some methods to filter by people against Whitehall database for cases when locale is not `en`

This PR does **not** address the actual creation of the email topics in the email alert service.

[Trello](https://trello.com/c/VvCBrEUe/57-5-add-drop-down-people-filter-to-announcements)

## Before screenshot
announcements_without_people

## After screenshot
announcements_with_people

### DIFF
--- a/app/assets/javascripts/application/document_filter.js
+++ b/app/assets/javascripts/application/document_filter.js
@@ -203,7 +203,7 @@ if(typeof window.GOVUK === 'undefined'){ window.GOVUK = {}; }
       $filterSummary.mustache('documents/_filter_selections', context);
     },
     removeFilters: function(field, removed){
-      var selects = ['topics', 'departments', 'world_locations', 'official_document_status'],
+      var selects = ['topics', 'departments', 'world_locations', 'official_document_status', 'people_ids'],
           inputs = ['keywords', 'from_date', 'to_date'];
 
       if($.inArray(field, selects) > -1){
--- a/app/helpers/document_filter_helper.rb
+++ b/app/helpers/document_filter_helper.rb
@@ -11,6 +11,11 @@ module DocumentFilterHelper
     unsorted_grouped_options_for_select(filter_options.for(:organisations).grouped, selected_values)
   end
 
+  def people_filter_options(selected_person = [])
+    selected_value = selected_person ? selected_person.map(&:slug) : ["all"]
+    filter_option_html(filter_options.for(:people), selected_value)
+  end
+
   def publication_type_filter_options(selected_publication_filter_option = nil)
     selected_value = selected_publication_filter_option ? selected_publication_filter_option.slug : "all"
     filter_option_html(filter_options.for(:publication_type), selected_value)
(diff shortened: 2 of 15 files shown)


app/assets/javascripts/application/document_filter.js
```diff
@@ -203,7 +203,7 @@ if(typeof window.GOVUK === 'undefined'){ window.GOVUK = {}; }
       $filterSummary.mustache('documents/_filter_selections', context);
     },
     removeFilters: function(field, removed){
-      var selects = ['topics', 'departments', 'world_locations', 'official_document_status'],
+      var selects = ['topics', 'departments', 'world_locations', 'official_document_status', 'people_ids'],
```
gpeng: Should this just be `people` to be consistent with the others.

app/helpers/document_filter_helper.rb
```diff
@@ -11,6 +11,11 @@ def organisation_filter_options(selected_organisations = [])
     unsorted_grouped_options_for_select(filter_options.for(:organisations).grouped, selected_values)
   end
 
+  def people_filter_options(selected_person = [])
```
gpeng: If this is an array (admittedly one that will only have one thing in it) it should probably be plural `selected_people`.

lib/whitehall/document_filter/options.rb
```diff
@@ -76,6 +77,13 @@ def options_for_topics
         @options_for_topics ||= StructuredOptions.new(all_label: "All policy areas", grouped: Classification.grouped_by_type)
       end
 
+      def options_for_people
+        @options_for_people ||= StructuredOptions.new(
+          all_label: "All people",
+          ungrouped: Person.all.sort_by(&:name).map { |o| [o.name, o.slug] }
```
gpeng: You could `.order(:name).pluck(:name, :slug)` here for performance goodness.

app/assets/javascripts/application/document_filter.js
leenagupte: I think we kept the name `people_ids` because that's what it was originally called here, and to keep it distinct from the people controller.

lib/whitehall/document_filter/options.rb
leenagupte: @gpeng I haven't been able to get this to work. `:name` seems to return the forename and surname of the person from Person, which is then used to populate the Atom feed filter url.

leenagupte:
@gpeng @tuzz I've addressed most of the comments you raise (thanks!) in another branch: https://github.com/alphagov/whitehall/compare/add-people-drop-down-to-announcements-v2.

It's Whitehall, I didn't want to force push over working code. I'll get another PR ready for review tomorrow.

tuzz:
Ok, great 👍 

Closing this as #3451 supersedes it.

PR #3446, opened 2017-09-13
Use email-alert-api instead of govuk-delivery for email subscriptions

tuzz:
https://trello.com/c/c2QpoCie/152-remove-calls-to-create-topics-from-the-whitehall-emailsignup-model

This change modifies the `EmailSignup` model to use `email-alert-api` for the creation / retrieval of subscriber lists as part of the subscription journey. It had previously used `govuk-delivery` which is being retired.

We now have no need for `govuk-delivery` in Whitehall so I've removed it. This includes all files in the the `lib/whitehall/gov_uk_delivery/` directory except for the `feed_url_validator` which I've now inlined next to where it's used in the `EmailSignup` model.

This also removes the `publish` even hook to notify `govuk-delivery`. Previously, the request to send an email was explicitly made from Whitehall, but we're now implicitly using Publishing API for this – an event is pushed onto a queue which is consumed by `email-alert-service`.

This change affects a few Cucumbers which can no longer make assertions about sending notifications because this happens outside Whitehall.


jennyd:
💥 🎆 🎉 ✨ ✉️ 📪

PR #3450, opened 2017-09-14
Redirect seed-enterprise-investment-scheme-procedures

bevanloon:
Migration to unpublish /guidance/seed-enterprise-investment-scheme-procedures
and provide an alternate url of /guidance/venture-capital-schemes-apply-to-use-the-seed-enterprise-investment-scheme
as per request in zendesk ticket https://govuk.zendesk.com/agent/tickets/2406584

The publisher cannot currently edit this document themselves because of a bug which is documented in https://github.com/alphagov/whitehall/pull/3457


bevanloon:
@gpeng Thanks - I've updated the approach now. This should take care of the zendesk ticket.
The actual bug that is stopping the publisher from doing this will hopefully be taken care of by https://github.com/alphagov/whitehall/tree/return-empty-array-when-expanded-links-needs-id-are-nil once it all builds and reviewed.

PR #3451, opened 2017-09-15
Add people drop down to announcements v2

leenagupte:
Supersedes: #3445 

Adds an extra filter to the `/government/announcements` search criteria.

* Adds the drop down to the frontend
* Sends the param through to the rummager search
* Update the atom and email urls to include the new param
* Adds some methods to filter by people against Whitehall database for cases when locale is not `en`

This PR does **not** address the actual creation of the email topics in the email alert service.

[Trello](https://trello.com/c/VvCBrEUe/57-5-add-drop-down-people-filter-to-announcements)

## Before screenshot
announcements_without_people

## After screenshot
announcements_with_people

### DIFF
--- a/lib/whitehall/document_filter/options.rb
+++ b/lib/whitehall/document_filter/options.rb
@@ -30,6 +30,7 @@ module Whitehall
         announcement_type: 'announcement_filter_option',
         official_documents: 'official_document_status',
         locations: 'world_locations',
+        people: 'people',
       }.freeze
 
       def valid_option_name?(option_name)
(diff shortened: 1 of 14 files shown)


lib/whitehall/document_filter/options.rb
```diff
@@ -76,6 +77,13 @@ def options_for_topics
         @options_for_topics ||= StructuredOptions.new(all_label: "All policy areas", grouped: Classification.grouped_by_type)
       end
 
+      def options_for_people
+        @options_for_people ||= StructuredOptions.new(
+          all_label: "All people",
+          ungrouped: Person.all.sort_by(&:name).map { |o| [o.name, o.slug] }
```
gpeng: I see why the `.order(:name).pluck(:name, :slug)` thing doesn't work now. `name` is a method not a db field 🤔 `.order` would probably still be faster before doing the map but I'm not sure it would be that significant. In digging about in this though I realise now that there are > 2800 people in the Whitehall database. That's a lot of options in a dropdown. The departments one has > 1000 too. This isn't good UI/UX. Hopefully this will get moved to finder-frontend in the next quarter.

tuzz:
🎉  ⭐️  🌈

lib/whitehall/document_filter/options.rb
leenagupte: Here's hoping. It doesn't help that some people are ordered by title and not first name.

PR #3452, opened 2017-09-18
Compile mustache templates to fix failing feature

bevanloon:
Tests were failing when merging to master in speeches feature when "Trying to save a speech that has been changed by another user". This seems to fix it.


gpeng:
Can you give a bit more context about this? Where does it come from? It looks like something that should happen as part of the build and we should be checking in a reviewable version but I might be missing the point.

bevanloon:
So basically I ran the tests locally on master branch with `bundle exec cucumber features/speeches.feature` and along with the failing tests, I also got the following:

```
ERROR: Scenarios tagged @javascript require mustache templates to be compiled first. Execute 'rake shared_mustache:compile' before running these scenarios.
```

I ran the `rake` command and tests passed.

We actually have this in whitehall https://github.com/alphagov/whitehall/blob/c6e5a11f4cc0623460d0f3b80bbb6cba54f7c80d/lib/tasks/test_javascript.rake

and the source of the error
https://github.com/alphagov/whitehall/blob/19cd7d72de32454d532c195f35b027fa1b3ba6ac/features/support/missing_mustache_templates_helper.rb

more than that, I'm not really sure - perhaps @h-lame has some ideas since I think he pushed the task in `test_javascript.rake`?

Yes a build step sounds like a better solution.

tuzz:
I think these templates are already building on CI so this is possibly not needed. I'm going to close this for now, though we can re-open and reconsider if the build starts failing again.

PR #3455, opened 2017-09-18
[DO NOT MERGE] Republish orgs. when CIPs change

tuzz:
We embed information from the 'About us' corporate information page
under the organisation listing on pages like this:

https://www.gov.uk/world/spain#/world/trade-and-invest-spain

This change makes it so we republish the corresponding organisation
for a corporate information page whenever that page changes, e.g.
it is published, withdrawn, etc.

This might be a bit heavy-handed, but it's 'cheap' to republish a
document to the Publishing API, so this might help keep things in
sync if things change in the future, e.g. we embed content from
other CIPs on the org. page.

https://trello.com/c/9JFXOc5y/145-worldwide-organisations-are-not-updated-when-their-about-us-summary-is-changed-added


tuzz:
This isn't the right approach because this data is actually served from Rummager on the Taxon pages. Closing this.

PR #3469, opened 2017-10-03
Add report of content by organisation

boffbowsh:
Produces a CSV report of all Documents created or updated within a
supplied date range, and splits them by Organisation. The CSV has 2
columns per month - one for documents initially created that month and
one for the number of updates to documents.


tuzz:
@boffbowsh: who's this report for? Can you give any more context?

boffbowsh:
Sorry, yes, it's for Sheila Fitzsimons so she can see how often depts are publishing on GOV.UK. I've already run the report with data from my VM, this is just contributing for the future in case it's useful.

PR #3472, opened 2017-10-03
Fix colour contrast in publication headers

vanitabarrett:
Publication headers do not meet WCAG AA when using $light-blue. Changing to $govuk-blue improves the contrast ratio to meet AA standards.


fofr:
Goes with https://github.com/alphagov/government-frontend/pull/497

Issue #3223, opened 2017-05-13
"Coming soon" content items in the past


fofr:
There are 142 content items with the schema_name "Coming soon" which tell users that this content will be published on a date that's in the past.

Example:
https://www.gov.uk/government/case-studies/ww1-australian-vc-recipient-corporal-alexander-stewart-burton

![screen shot 2017-05-13 at 11 03 32](https://cloud.githubusercontent.com/assets/319055/26024570/d2e7644a-37cb-11e7-9b07-f041b1b480b2.png)

## See list of 142
```
all = ContentItem.where(schema_name: 'coming_soon').to_a
y all.select {|c| Date.parse(c.details[:publish_time]) < Date.today }.map {|c| [c._id, c.details[:publish_time]] }
```

cc @gpeng @rubenarakelyan @nickcolley

rubenarakelyan:
I've added this to our support board

ChrisBAshton:
No loner seems to be a thing:

```
ContentItem.where(schema_name: 'coming_soon').to_a
=> []
```

PR #3482, opened 2017-10-06
Use `Edition#content_id` in `ComingSoonPresenter`

gpeng:
The `ComingSoonPresenter` was generating a random content id when presenting content to the publishing API. This was causing `ComingSoon` pages to be orphaned if the content was unscheduled.

This commit uses the edition `content_id` instead which resolves the issue.

[Trello](https://trello.com/c/ef1EoOZQ/189-unscheduled-report-still-shows-as-scheduled)

### DIFF
--- a/app/presenters/publishing_api/coming_soon_presenter.rb
+++ b/app/presenters/publishing_api/coming_soon_presenter.rb
@@ -24,7 +24,7 @@ module PublishingApi
     def initialize(item, update_type: nil)
       self.item = item
       self.update_type = update_type || "major"
-      @content_id = SecureRandom.uuid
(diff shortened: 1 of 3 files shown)


tuzz:
Is it worth cleaning up the orphaned docs. in publishing api, too?

fofr:
Will this fix https://github.com/alphagov/whitehall/issues/3223?

gpeng:
@fofr Yes, well I think it will fix the cause of them being left there. @tuzz Yes, I'll look at that in a separate PR. I don't think we have anything to clean up in Whitehall as they are only created in publishing api/content store.

PR #3486, opened 2017-10-11
Remove SRI from admin JS

gpeng:
We've received a number of support tickets relating to issues with the pop-up menus in Whitehall admin not popping up. These seem to be caused by the admin JS not loading for some users since we added the SRI attributes.

We need to make the JS dependent functionality more resilient but in the short term this commit removes the attributes from the JS. 

A ticket will be added to the Whitehall support backlog to review the use of JS in the Whitehall admin interface and make changes where necessary before reverting this commit.

[Trello](https://trello.com/c/2wYn3eJi/202-new-document-tab-does-not-display-dropdown)

cc: @fofr

### DIFF
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -109,7 +109,7 @@
 <% end %>
 
 <% content_for :body_end do %>
-  <%= javascript_include_tag "admin", integrity: true, crossorigin: 'anonymous' %>
+  <%= javascript_include_tag "admin" %>
   <%= render_mustache_templates if Rails.env.development? %>
   <% initialise_script "GOVUK.adminGlobalInitialiser" %>
   <%= javascript_tag(yield :javascript_initialisers) %>


app/views/layouts/admin.html.erb
```diff
@@ -109,7 +109,7 @@
 <% end %>
 
 <% content_for :body_end do %>
-  <%= javascript_include_tag "admin", integrity: true, crossorigin: 'anonymous' %>
```
h-lame: You're only doing it for the JS, but the CSS also has SRI, is it worth removing it from there too?  Maybe not if we've not had any reports about the site not looking right.

gpeng: Yeah... I considered that. The tickets we've had have been specifically about the 'New Document' and 'More' tabs not working with no mention of styling so I left it on there on the assumption that it isn't causing any problems.

gpeng:
Ticket to fix and revert this PR: https://trello.com/c/Ly0hzsOT/1098-some-functionality-in-whitehall-admin-doesnt-work-at-all-without-javascript

app/views/layouts/admin.html.erb
h-lame: :+1:

PR #3487, opened 2017-10-11
Reslug UKTI organisations to DIT

rubenarakelyan:
This commit changes the slugs of all the worldwide UKTI organisations to DIT following the change of name of the department.

Trello: https://trello.com/c/K6rCadFY/205-75-organisation-changes


db/data_migration/20171011123033_reslug_ukti_to_dit.rb
```diff
@@ -0,0 +1,84 @@
+require "data_hygiene/organisation_reslugger"
```
gpeng: I don't think we need to require this. The environment gets included when these get run https://github.com/alphagov/whitehall/blob/master/lib/tasks/data_migrator.rake#L6

rubenarakelyan: Ah right, was copying from prior art!

PR #3492, opened 2017-10-12
Reslug the-business-and-property-court -> ...-courts

tuzz:
https://govuk.zendesk.com/agent/tickets/2447338
https://trello.com/c/LU45fBKr/204-some-url-changes


db/data_migration/20171012152847_reslug_to_pluralize_court.rb
```diff
@@ -0,0 +1,5 @@
+require "data_hygiene/organisation_reslugger"
```
gpeng: You don't need this. The rake task that runs these includes the environment.

PR #3495, opened 2017-10-16
Fix jQuery UI after upgrading the Rails version

tuzz:
https://trello.com/c/cDOjXWjP/231-editor-view-for-adding-items-to-collection-has-bad-formatting-post-rails-5-upgrade

Previously, the 'jquery.ui.all' module loaded
these modules recursively. This doesn't seem to
work anymore so require them explicitly.

Longer term, we shouild try upgrading jQuery UI
to the latest version:
https://trello.com/c/je5OJmaY/1113-upgrade-jquery-ui-in-whitehall-admin

**Before:**
screen shot 2017-10-16 at 12 29 12

**After:**
screen shot 2017-10-16 at 12 29 53

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -1,5 +1,21 @@
 @import 'govuk_admin_template';
-@import 'jquery.ui.all';
+
+@import 'jquery.ui.core';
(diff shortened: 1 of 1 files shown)


fofr:
jQuery UI 😞

PR #3496, opened 2017-10-16
Clarify whether email alerts go out when unwithdrawing

tuzz:
We've had a couple of Zendesk tickets asking about this.

https://trello.com/c/nUE37uD0/230-user-query-does-whether-unwithdrawing-trigger-an-email

### DIFF
--- a/app/views/admin/edition_workflow/_unwithdraw_modal.html.erb
+++ b/app/views/admin/edition_workflow/_unwithdraw_modal.html.erb
@@ -9,7 +9,7 @@
           <h3 class="modal-title">Are you sure you want to unwithdraw this?</h3>
         </div>
         <div class="modal-body">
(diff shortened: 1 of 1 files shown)


app/views/admin/edition_workflow/_unwithdraw_modal.html.erb
```diff
@@ -9,7 +9,7 @@
           <h3 class="modal-title">Are you sure you want to unwithdraw this?</h3>
         </div>
         <div class="modal-body">
-          <p>This will create a new edition and immediately publish it.</p>
+          <p>This will create a new edition and immediately publish it silently. This will not trigger email alerts.</p>
```
gpeng: I think you can probably lose the 'silently'. It's a less clear way of saying 'this will not trigger email alerts'.

PR #3501, opened 2017-10-18
Add a migration to fix GOV.UK Pay URL

thomasleese:
The URL for GOV.UK Pay is https://www.payments.service.gov.uk but there currently exists old guidance in Whitehall about Pay which can be found from Google. This commit adds a migration which removes the old guidance pages and redirects them to the URL for GOV.UK Pay instead.

[Trello Card](https://trello.com/c/933P4qhQ/240-redirect-url-and-remove-original-page-from-whitehall-govuk-pay)

### DIFF
--- a/db/data_migration/20171018095703_fix_govuk_pay_url.rb
+++ b/db/data_migration/20171018095703_fix_govuk_pay_url.rb
@@ -1,0 +1,20 @@
+slug = "govuk-pay"
+redirect_path = "https://www.payments.service.gov.uk"
+
+document = Document.find_by(slug: slug)
+exit unless document
+
+attachment_content_ids = document.editions.flat_map do |edition|
+  edition.attachments.map(&:content_id)
+end.uniq
+
+content_ids = [document.content_id] + attachment_content_ids
+
+document.delete
+
+puts "#{slug} -> #{redirect_path}"
+
+content_ids.each do |content_id|
+  puts "- #{content_id}"
+  PublishingApiRedirectWorker.new.perform(content_id, redirect_path, "en")
+end


db/data_migration/20171018095703_fix_govuk_pay_url.rb
```diff
@@ -0,0 +1,20 @@
...
+document = Document.find_by(slug: slug)
+exit unless document
+
+attachment_content_ids = document.editions.flat_map do |edition|
+  edition.attachments.map(&:content_id)
+end.uniq
+
+content_ids = [document.content_id] + attachment_content_ids
+
+document.delete
```
gpeng: I don't think you should delete the document in this. I think the correct thing to do would be to unpublish it with a redirect unpublish which will achieve the same thing but leave the history of the document in Whitehall. Is there a reason you didn't do that?

tijmenb:
Is there a reason this can't be done via the Whitehall UI? Seems like we could redirect it.

thomasleese:
I tried unpublishing the document in Whitehall, but it gave me an error because the destination URL is not of the form `gov.uk/example` (I assume redirecting to external GOV.UK URLs isn't support in the UI?).

This was also the reason I went with this approach in the migration, @gpeng, but perhaps I can use code to unpublish the edition and it won't complain?

gpeng:
🤔 We could do it in code except the validation that you're tripping up on is [on the `Unpublishing` model](https://github.com/alphagov/whitehall/blob/master/app/models/unpublishing.rb#L10) which we'd still need to create so that's a bit of a problem.

gpeng:
Discussed this offline... To keep the data consistent in Whitehall -> publishing api -> content store -> router we would ideally change the validation to allow `*.gov.uk` domains. We going to discuss this with the PM.

PR #3503, opened 2017-10-19
Remove Organisation logos from Asset Manager

chrisroos:
Removing assets from Asset Manager when they're removed from Whitehall should allow us to keep the assets in the two systems synchronised while we're running them in parallel.

The requests to remove assets from Asset Manager are placed on a queue to allow us to retry jobs that fail due to network problems.

We've started by removing Organisation logos to reduce the scope of this change.

### DIFF
--- a/app/workers/asset_manager_delete_asset_worker.rb
+++ b/app/workers/asset_manager_delete_asset_worker.rb
@@ -1,0 +1,9 @@
+class AssetManagerDeleteAssetWorker < WorkerBase
+  def perform(legacy_url_path)
+    gds_api_response = Services.asset_manager.whitehall_asset(legacy_url_path)
+    asset_url = gds_api_response['id']
+    asset_id = asset_url[/\/assets\/(.*)/, 1]
+
+    Services.asset_manager.delete_asset(asset_id)
+  end
+end
--- a/lib/whitehall/asset_manager_and_quarantined_file_storage.rb
+++ b/lib/whitehall/asset_manager_and_quarantined_file_storage.rb
@@ -5,6 +5,27 @@ class Whitehall::AssetManagerAndQuarantinedFileStorage < CarrierWave::Storage::A
   end
 
   def retrieve!(identifier)
-    Whitehall::QuarantinedFileStorage.new(uploader).retrieve!(identifier)
+    asset_manager_file = Whitehall::AssetManagerStorage.new(uploader).retrieve!(identifier)
+    quarantined_file = Whitehall::QuarantinedFileStorage.new(uploader).retrieve!(identifier)
+
+    File.new(asset_manager_file, quarantined_file)
+  end
+
+  class File
+    delegate :empty?, :path, :content_type, :filename, to: :@quarantined_file
+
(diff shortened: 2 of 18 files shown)


chrisroos:
@chrislo: I've made a few changes since we paired on this so I'd appreciate you taking a look to make sure everything makes sense.

@floehopper: I'd value your input on this given that you haven't seen it before.

lib/whitehall/asset_manager_storage.rb
```diff
@@ -1,20 +1,35 @@
...
-  def retrieve!(_identifier)
-    raise "We're not currently reading assets from Asset Manager so this shouldn't be called."
+  def retrieve!(identifier)
+    asset_path = uploader.store_path(identifier)
+    File.new(asset_path)
+  end
+
+  class File
+    def initialize(asset_path)
+      @legacy_url_path = '/government/uploads/' + asset_path
```
chrislo: It might be better to use `File.join` here

app/workers/asset_manager_delete_asset_worker.rb
```diff
@@ -0,0 +1,9 @@
+class AssetManagerDeleteAssetWorker < WorkerBase
+  def perform(legacy_url_path)
+    gds_api_response = Services.asset_manager.whitehall_asset(legacy_url_path)
+    asset_url = gds_api_response['id']
+    asset_id = asset_url[/\/assets\/(.*)/, 1]
```
chrislo: Not for this PR - but it might be nice to add this parsing of the URL ID to the API adapter gem at some point. Perhaps open an issue against that library?

lib/whitehall/asset_manager_and_quarantined_file_storage.rb
```diff
@@ -5,10 +5,14 @@ def store!(file)
   end
 
   def retrieve!(identifier)
-    asset_manager_file = Whitehall::AssetManagerStorage.new(uploader).retrieve!(identifier)
     quarantined_file = Whitehall::QuarantinedFileStorage.new(uploader).retrieve!(identifier)
 
-    File.new(asset_manager_file, quarantined_file)
+    if Whitehall.use_asset_manager
```
chrislo: I think this looks fine - another option might be to conditionally switch the storage engine used based on the feature flag in the Uploaders. But given that this can go away when the feature flag is enable it may not be worth the effort.

chrisroos:
I think I've found a problem with this branch. It seems that although the asset is being deleted from the file system and Asset Manager, the asset filename isn't being removed from the ActiveRecord column (e.g. `Organisation#logo`). I've added a failing test to this branch to prompt me to investigate further tomorrow.

lib/whitehall/asset_manager_storage.rb
chrisroos: Done!

lib/whitehall/asset_manager_and_quarantined_file_storage.rb
chrisroos: We've now decided to remove the feature flag in this branch so this no longer applies.

PR #3506, opened 2017-10-25
Vanish coming soon pages for unscheduled editions

rubenarakelyan:
This commit changes the current behaviour of publishing a “410 Gone” item for a “coming soon” when its associated scheduled publication is unscheduled. Now it will remove the item completely which will result in a “404 Not Found” error.

404 is correct here since nothing has been published yet, and so nothing has “gone”.

Trello: https://trello.com/c/BinSjhTs/261-update-logic-for-removing-coming-soon-pages

### DIFF
--- a/app/workers/publishing_api_vanish_worker.rb
+++ b/app/workers/publishing_api_vanish_worker.rb
@@ -1,0 +1,12 @@
+class PublishingApiVanishWorker < PublishingApiWorker
+  def perform(content_id, locale)
+    Services.publishing_api.unpublish(
+      content_id,
+      type: "vanish",
+      locale: locale
+    )
+  rescue GdsApi::HTTPNotFound, GdsApi::HTTPUnprocessableEntity
+    # nothing to do here as we can't unpublish something that doesn't exist
+    nil
+  end
+end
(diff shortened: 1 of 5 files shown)


app/workers/publishing_api_vanish_worker.rb
```diff
@@ -0,0 +1,12 @@
+class PublishingApiVanishWorker < PublishingApiWorker
+  def perform(content_id, locale)
+    Services.publishing_api.unpublish(
+      content_id,
+      type: "vanish",
```
cbaines: From what I remember about the intention behind supporting this as an unpublishing type, it was about removing sensitive content from the site.

From the commit message, I gather that this is about getting a 404, rather than a 410, but whatever content has been published was still published. Is there something odd about the content that is being unpublished here?

rubenarakelyan: To clarify, this makes the "coming soon" page a 404 because the content that was coming soon was never published. So the content itself isn't there, and it doesn't make sense for the page that was announcing some upcoming content to be replaced with a 410.

PR #3507, opened 2017-10-25
Delete previously gone coming soon content items

rubenarakelyan:
These content items were previously unpublished in https://github.com/alphagov/whitehall/pull/3483 with a 410 Gone status. This commit deletes them completely from the content-store so that they return a 404 Not Found error instead. This is because these items referred to content that was never published, therefore it doesn’t make sense to show a 410 error.

Trello: https://trello.com/c/pJ6TWbCV/262-update-unpublished-coming-soon-pages-with-404-http-responses


rubenarakelyan:
As a note, I have copied the array of items from the previous PR and have kept the base paths and dates although they are not used for reference purposes so that we know that the content IDs refer to.

PR #3509, opened 2017-10-25
Reslug digital-data-and-technology-professions

gpeng:
Apparently shouldn't have the last 's'.


gpeng:
I noticed after deploying this that updating `organisations.name` doesn't actually change the name displayed on the page. This uses `organisations.logo_formatted_name`. This was fixed in the console.

PR #3514, opened 2017-10-27
Use Asset Manager to get URL of asset

chrisroos:
This will ensure that once we switch an uploader from using `AssetManagerAndQuarantinedFileStorage` to `AssetManagerStorage` we'll start rendering the assets in whitehall-admin from Asset Manager instead of the file system.

I'm not entirely sure about making this request to the Asset Manager API to get the URL. It feels correct to do this because it means that Whitehall doesn't need to know anything about how the URLs are created. On the other hand, I wonder whether the additional network request is overkill given that we know the path (which is `/government/uploads/ + @legacy_url_path`) and all we need to prepend is the assets host (e.g. assets-origin.)

### DIFF
--- a/test/unit/whitehall/asset_manager_storage_test.rb
+++ b/test/unit/whitehall/asset_manager_storage_test.rb
@@ -70,4 +70,12 @@ class Whitehall::AssetManagerStorage::FileTest < ActiveSupport::TestCase
 
     file.delete
   end
+
+  test 'uses the asset-manager api to return the url of the file' do
+    asset_url = 'http://asset-host/asset.png'
+    Services.asset_manager.stubs(:whitehall_asset).with('/government/uploads/path/to/asset.png').returns('file_url' => asset_url)
+
+    file = Whitehall::AssetManagerStorage::File.new('path/to/asset.png')
+    assert_equal asset_url, file.url
+  end
 end
--- a/test/unit/whitehall/asset_manager_storage_test.rb
+++ b/test/unit/whitehall/asset_manager_storage_test.rb
@@ -63,19 +63,22 @@ class Whitehall::AssetManagerStorageTest < ActiveSupport::TestCase
 end
 
 class Whitehall::AssetManagerStorage::FileTest < ActiveSupport::TestCase
+  setup do
+    asset_path = 'path/to/asset.png'
+    @asset_url_path = "/government/uploads/#{asset_path}"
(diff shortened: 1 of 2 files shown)


chrisroos:
@floehopper / @chrislo: What do you think about the call to the Asset Manager API in this branch? Is it OK or should we construct the path to the asset some other way?

floehopper:
It might be useful to investigate how other publishing apps work. I have a feeling they store the URL returned from the Asset Manager. If so, why would we *not* do that?

test/unit/whitehall/asset_manager_storage_test.rb
```diff
@@ -70,4 +70,12 @@ class Whitehall::AssetManagerStorage::FileTest < ActiveSupport::TestCase
 
     file.delete
   end
+
+  test 'uses the asset-manager api to return the url of the file' do
+    asset_url = 'http://asset-host/asset.png'
+    Services.asset_manager.stubs(:whitehall_asset).with('/government/uploads/path/to/asset.png').returns('file_url' => asset_url)
+
+    file = Whitehall::AssetManagerStorage::File.new('path/to/asset.png')
```
floehopper: Minor: Consider extracting this into a common setup block.

chrisroos:
> I definitely prefer obtaining the URL from Asset Manager versus constructing it in Whitehall, because the latter would indirectly couple Whitehall to Asset Manager's implementation. It might be useful to investigate how other publishing apps use Asset Manager. I have a feeling they store the URL returned from the Asset Manager. If so, why would we not do that? It's probably also worth considering how we'd like the Asset Manager API to work in the future - this might give us a steer on how best to approach this.

You're correct that other apps using Asset Manager store the URL returned from Asset Manager. And it's possible we might want to do that in Whitehall in future. Unfortunately, it's not particularly easy to do that at the moment because of our use of CarrierWave. In particular our use of CarrierWave to create multiple versions of some types of uploaded files: each of these versions would become a new asset in Asset Manager and so we'd need some way of storing an arbitrary number of Asset Manager URLs in Whitehall.

chrisroos:
Thanks for the review, @floehopper. I've rebased on master and will get this merged to master once all the tests have passed.

test/unit/whitehall/asset_manager_storage_test.rb
chrisroos: I've added another commit to extract some of the common code into a `setup` block.

PR #3513, opened 2017-10-27
Migrate organisation logos to asset manager

chrislo:
This PR introduces a rake task `asset_manager:migrate_organisation_logos` which creates a sidekiq job for each file under `system/uploads/organisation/logo/` (organisation logos) to create an asset in the Asset Manager API (if one doesn't already exist for this logo).

There are around 60 such files in production in the moment and the jobs will be created on a low-priority queue when this rake task is run.

Fixes: https://github.com/alphagov/asset-manager/issues/276

PR #3515, opened 2017-10-31
Fix migration of organisation logos to Asset Manager

chrisroos:
This builds on the changes made in PR #3513 by:

* ensuring we send the file content to Asset Manager
* setting the legacy etag to a value that matches the value generated by Nginx

### DIFF
--- a/lib/migrate_assets_to_asset_manager.rb
+++ b/lib/migrate_assets_to_asset_manager.rb
@@ -63,9 +63,7 @@ class MigrateAssetsToAssetManager
     end
 
     def legacy_etag
-      etag = Digest::MD5.hexdigest(read)
-      rewind
-      etag
+      '%x-%x' % [legacy_last_modified, size]
     end
   end
 end
--- a/lib/migrate_assets_to_asset_manager.rb
+++ b/lib/migrate_assets_to_asset_manager.rb
@@ -63,7 +63,9 @@ class MigrateAssetsToAssetManager
     end
 
     def legacy_etag
-      Digest::MD5.hexdigest(read)
+      etag = Digest::MD5.hexdigest(read)
+      rewind
+      etag
     end
   end
 end
--- a/test/unit/migrate_assets_to_asset_manager_test.rb
+++ b/test/unit/migrate_assets_to_asset_manager_test.rb
@@ -42,7 +42,10 @@ class MigrateAssetsToAssetManagerTest < ActiveSupport::TestCase
   end
 
   test 'it calls create_whitehall_asset with the legacy etag' do
-    expected_etag = Digest::MD5.hexdigest(@organisation_logo_file.read)
+    expected_etag = [
+      File.stat(@organisation_logo_path).mtime.to_i.to_s(16),
+      File.stat(@organisation_logo_path).size.to_i.to_s(16),
+    ].join('-')
 
     Services.asset_manager.expects(:create_whitehall_asset).with(
       has_entry(:legacy_etag, expected_etag)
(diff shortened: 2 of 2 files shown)


test/unit/migrate_assets_to_asset_manager_test.rb
```diff
@@ -17,7 +17,7 @@ class MigrateAssetsToAssetManagerTest < ActiveSupport::TestCase
   end
 
   test 'it calls create_whitehall_asset for each file in the list' do
-    Services.asset_manager.expects(:create_whitehall_asset).with(has_entry(:file, responds_with(:path, @organisation_logo_path)))
+    Services.asset_manager.expects(:create_whitehall_asset).with(has_entry(:file, responds_with(:read, File.read(@organisation_logo_path))))
```
floehopper: It would be nice to avoid the very long line and improve the readability. Some ideas:

* Stub `Services.asset_manager` in the setup and return a mock object stored in an `@asset_manager` instance variable.
* Extract `File.read(@organisation_logo_path)` into a `content` local variable.
* Extract `responds_with(:read, content)` into a private method, `def file_with(content)`.

floehopper:
Minor: In the "Generate legacy etags according to the algorithm used by Nginx" commit note, "on mass" should be "en masse"!

lib/migrate_assets_to_asset_manager.rb
```diff
@@ -63,9 +63,9 @@ def legacy_last_modified
     end
 
     def legacy_etag
-      etag = Digest::MD5.hexdigest(read)
-      rewind
-      etag
+      mtime_hex = legacy_last_modified.to_i.to_s(16)
+      file_size_hex = size.to_s(16)
+      [mtime_hex, file_size_hex].join('-')
```
floehopper: Minor:

> I've copied the tests for this behaviour from the equivalent behaviour
added in Asset Manager[1]. I've implemented it without referring to the
implementation in Asset Manager.

It feels a bit perverse to copy the tests, but not the implementation, from Asset Manager. Do you think there's something wrong with the implementation in Asset Manager? Why wouldn't we just copy the implementation?

test/unit/migrate_assets_to_asset_manager_test.rb
```diff
@@ -93,3 +96,29 @@ class OrganisationLogoFilePathsTest < ActiveSupport::TestCase
     end
   end
 end
+
+class OrganisationLogoFileTest < ActiveSupport::TestCase
```
floehopper: Minor: It feels a bit weird that this test of a nested class is in the top-level namespace. Is there any reason not to nest it inside the test case for the parent class?

lib/migrate_assets_to_asset_manager.rb
floehopper: I've just discovered that you ended up copying the implementation from Asset Manager in the next commit, so my previous comment can be disregarded.

Personally, I'd prefer to merge the two commits. If you decide not to do that, I'd be inclined to explain why you haven't copied the implementation in the note for this commit.

test/unit/migrate_assets_to_asset_manager_test.rb
chrisroos: > Minor: It feels a bit weird that this test of a nested class is in the top-level namespace. Is there any reason not to nest it inside the test case for the parent class?

I copied the test that was already there for `OrganisationLogoFilePathsTest` in that it is also at the top-level rather than nested. I don't feel particularly strongly about it so I'm going to leave it as it is.

chrisroos:
> Minor: In the "Generate legacy etags according to the algorithm used by Nginx" commit note, "on mass" should be "en masse"!

Updated. Thanks!

test/unit/migrate_assets_to_asset_manager_test.rb
chrisroos: I've reduced the length of the line by moving the `.with` onto the next line.

lib/migrate_assets_to_asset_manager.rb
chrisroos: I've squashed the last two commits.

chrisroos:
Thanks @floehopper. I've addressed your feedback, rebased on master and force pushed. I'll get this merged once the tests pass.

PR #3522, opened 2017-11-06
Allow other assets to be migrated using a rake task

chrislo:
This PR makes the specific organisation logo migration rake task introduced in db3a6e9de2a general by allowing it to take a directory as a parameter. For example:

    rake asset_manager:migrate_assets[system/uploads/image_data]

will migrate all assets under `Whitehall.clean_uploads_root + system/uploads/image_data`.

### DIFF
--- a/lib/tasks/asset_manager.rake
+++ b/lib/tasks/asset_manager.rake
@@ -1,6 +1,16 @@
 namespace :asset_manager do
   desc "Migrates Assets to Asset Manager."
-  task migrate_assets: :environment do
-    MigrateAssetsToAssetManager.new.perform
+  task :migrate_assets, [:target_dir] => :environment do |_, args|
+    abort(usage_string) unless args[:target_dir]
+    MigrateAssetsToAssetManager.new(args[:target_dir]).perform
+  end
+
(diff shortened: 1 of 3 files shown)


lib/tasks/asset_manager.rake
```diff
@@ -1,6 +1,9 @@
 namespace :asset_manager do
-  desc "Migrates Organisation logos to Asset Manager."
-  task migrate_organisation_logos: :environment do
-    MigrateAssetsToAssetManager.new.perform
+  desc "Migrates Assets to Asset Manager."
+  task :migrate_assets, [:target_dir] => :environment do |_, args|
+    raise "Usage: asset_manager:migrate_assets[system/uploads/organisation/logo]" unless args[:target_dir]
```
chrisroos: I think using "/system/uploads/organisation/logo" as an example is a bit confusing. Can you think of a way of being clearer about what this argument should be? Or do you think it's OK as it is?

chrislo: Thanks @chrisroos - I've improved the documentation string and will rebase before merging.

PR #3523, opened 2017-11-07
Use Asset Manager for ConsultationResponseFormUploader

chrislo:
This commit switches the ConsultationResponseFormUploader to use the
`asset_manager_and_quarantined_file_storage` engine for storage -
meaning that assets will be stored on disk as before, but also sent to
Asset Manager.

We made a similar change to the LogoUploader in 794542b1.

We have to stub calls to the Asset Manager API in the
ConsultationsControllerTest but otherwise no changes to the tests are
required.

We've verified in development that:
 - adding an attachement to a Consultation
 - removing an attachment from a Consultation
 - replacing an attachment on a Consultation

all result in the correct calls being made to Asset Manager. The
existing behaviour when a draft consultation is discarded is to not
remove the associated asset from disk. It is not clear whether this is
intended behaviour, but switching the storage engine preserves this
behaviour (that is, no DELETE call is made to the Asset Manager for
the associated asset).

### DIFF
--- a/app/uploaders/consultation_response_form_uploader.rb
+++ b/app/uploaders/consultation_response_form_uploader.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
 
 class ConsultationResponseFormUploader < WhitehallUploader
+  storage :asset_manager_and_quarantined_file_storage
+
   def extension_whitelist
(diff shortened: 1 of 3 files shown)


chrislo:
Rebasing on master and force-pushing before merging.

PR #3524, opened 2017-11-07
Initial Dockerfile

stephenharker:
Docker image should build but may need some tweaks.


chrislo:
Would you be able to give [a bit more context](https://github.com/alphagov/styleguides/blob/master/pull-requests.md) for this PR @stephenharker?

kevindew:
@thomasleese if we don't specify it falls back to development anyway. I have wondered about whether we benefit from it being set though as it causes a pain when you try and run tests as it tries to run them in development. I think when initially added it was a way of making concrete that it would run in development by default which I guess still stands.

PR #3527, opened 2017-11-08
Catch all kinds of 500 errors from GDS API

thomasleese:
Currently we only catch 502 errors, and all others get sent
through to Sentry. Looking at the history of why we picked only
502 it seems to be because this is the code we get when Rummager
times out.

Since then we've integrated with Publishing API which will often
time out and come through to us as a 504 error. It doesn't seem
useful to log this as an exception since there is nothing we can
do about it in Whitehall.

It seems to me that not reporting any kind of external server error
seems like the right thing to do since it's not in our control and
increases the noise when looking through the exception log. Instead
this commit will mean that we only see client errors which is more
likely to expose a problem on our side.

### DIFF
--- a/app/controllers/public_facing_controller.rb
+++ b/app/controllers/public_facing_controller.rb
@@ -11,12 +11,8 @@ class PublicFacingController < ApplicationController
     log_error_and_render_500 exception
   end
 
-  rescue_from GdsApi::HTTPErrorResponse do |exception|
-    if exception.code == 502
-      log_error_and_render_500 exception
-    else
-      raise
-    end
+  rescue_from GdsApi::HTTPServerError do |exception|
+    log_error_and_render_500 exception
   end
 
   # Allows additional request formats to be enabled.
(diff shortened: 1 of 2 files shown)


app/controllers/public_facing_controller.rb
```diff
@@ -12,7 +12,7 @@ class PublicFacingController < ApplicationController
   end
 
   rescue_from GdsApi::HTTPErrorResponse do |exception|
```
tijmenb: Could you also do `rescue_from GdsApi::HTTPServerError` (the 5XX superclass) and not have the `if` statement?

tijmenb:
> Since then we've integrated with Publishing API which will often time out and come through to us as a 504 error. It doesn't seem useful to log this as an exception since there is nothing we can do about it in Whitehall.

Is this correct? The `PublicFacingController` presumably does only frontend stuff, which shouldn't talk to the publishing-api.

thomasleese:
@tijmenb Er, good point, commit could be clearer, it was actually Content Store I meant.

tijmenb:
Cool. To connect this to https://github.com/alphagov/govuk-rfcs/pull/87, which error type do you think this is?

https://github.com/alphagov/govuk-rfcs/blob/errors/rfc-087-dealing-with-errors.md

thomasleese:
I would say they are "Intermittent errors without user impact".

PR #3528, opened 2017-11-09
Allow RIS files to be uploaded

rubenarakelyan:
This commit whitelists `.ris` files for uploading as attachments. These are text files that contain bibliographical citations in an industry-recognised format.

Trello: https://trello.com/c/wI76F5mI/313-add-citations-file-to-reports

### DIFF
--- a/app/uploaders/attachment_uploader.rb
+++ b/app/uploaders/attachment_uploader.rb
@@ -6,7 +6,7 @@ class AttachmentUploader < WhitehallUploader
 
   THUMBNAIL_GENERATION_TIMEOUT = 10.seconds
   FALLBACK_PDF_THUMBNAIL = File.expand_path("../../assets/images/pub-cover.png", __FILE__)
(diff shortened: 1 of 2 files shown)


fofr:
Is there an example where these files would be uploaded?

In what industry are they recognised?

Is there sufficient guidance for users who do not know what this file type is?

Is this a proprietary format?

fofr:
(The Trello story isn't public by default so I can't see the story’s context)

rubenarakelyan:
@fofr The specific request was about PHE research and analysis (e.g. https://www.gov.uk/government/publications/typhoid-and-paratyphoid-laboratory-confirmed-cases-in-england-wales-and-northern-ireland) which normally include a bibliography of references. A RIS file including these references can be imported into reference-management systems such as EndNote so they can be examined, rather than having to be extracted from PDFs manually. Since these files are in addition to the bibliographies already included in the relevant documents, most users don't need to use them, but they are useful to academics, researchers and (in this case) doctors who wish to easily examine referenced documents. The format itself is text-based and can be imported/exported by multiple systems (see https://en.wikipedia.org/wiki/RIS_(file_format)).

fofr:
Thanks @rubenarakelyan :+1:

PR #3534, opened 2017-11-14
Alert on publishing by CO NSS Watchkeeper account

rubenarakelyan:
This commit adds an email alert to content second line if the CO NSS Watchkeeper account publishes anything. This account is not meant to publish anything except in emergencies.

Trello: https://trello.com/c/lDbWmnE0/333-set-up-alerts-if-co-nss-watchkeepers-account-publishes-anything

### DIFF
--- a/app/mailers/notifications.rb
+++ b/app/mailers/notifications.rb
@@ -42,6 +42,12 @@ class Notifications < ActionMailer::Base
     mail from: no_reply_email_address, to: @author.email, subject: subject
   end
 
+  def edition_published_by_monitored_user(user)
+    @user = user
+    subject = "Account holder #{@user.name} (#{user.email}) has published to live"
+    mail from: no_reply_email_address, to: content_second_line_email_address, subject: subject
+  end
+
   def broken_link_reports(zip_path, recipient_address)
     filename = File.basename(zip_path)
     attachments[filename] = File.read(zip_path)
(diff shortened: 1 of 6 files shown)


app/models/edition/audit_trail.rb
```diff
@@ -28,17 +28,34 @@ def self.acting_as(actor)
...
 
+  def alert!(user)
+    if should_alert_for?(user)
+      Notifications.edition_published_by_monitored_user(user).deliver_now
+    end
+  end
+  private :alert!
+
+  def should_alert_for?(user)
+    ENV['CO_NSS_WATCHKEEPER_EMAIL_ADDRESS'].present? &&
```
thomasleese: Do we need the check for `present?` if we're checking it's equal to the email address?

rubenarakelyan: Nope, equality is good. I'll also only add this env var on production so we don't send emails for other environments.

app/mailers/notifications.rb
```diff
@@ -67,4 +73,8 @@ def no_reply_email_address
     address.display_name = name
     address.format
   end
+
+  def content_second_line_email_address
+    "second-line-content@digital.cabinet-office.gov.uk"
```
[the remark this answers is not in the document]
rubenarakelyan: No reason, maybe we should do but now it's in a public PR! I guess it's also not really a sensitive email address whereas the Watchkeeper one sounds like it may be.

PR #3535, opened 2017-11-14
Fix two HCA publication URLs

rubenarakelyan:
Trello: https://trello.com/c/J160swVJ/343-2-url-changes


rubenarakelyan:
Haha, I couldn't let that go in without fixing it!

PR #3542, opened 2017-11-17
Fix problem with replacing assets in asset manager store

chrisroos:
We've seen [an exception in production](https://sentry.io/govuk/app-whitehall/issues/391658602/events/9208224189/) when someone tried to replace an organisation logo with a file of the same name. The file wasn't removed from Asset Manager and so the replacement couldn't be uploaded.

I've replicated the problem in the AssetManagerIntegrationTest and fixed it by adding the `AssetManagerStorage::File#path` method. I'm assuming that CarrierWave checks whether the path has changed when deciding whether it needs to upload a new file during replacement.

The fix is in the first commit in the branch. The subsequent commits all contain small improvements to the Asset Manager integration tests.

### DIFF
--- a/lib/whitehall/asset_manager_storage.rb
+++ b/lib/whitehall/asset_manager_storage.rb
@@ -35,5 +35,9 @@ class Whitehall::AssetManagerStorage < CarrierWave::Storage::Abstract
     def filename
       @legacy_url_path
     end
+
+    def path
+      @legacy_url_path
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


chrisroos:
Thanks @floehopper. I've added a link to the exception in Sentry in the first commit message and in the PR description.

I've rebased on master in preparation for getting it merged.

chrisroos:
This build failed on Jenkins after I rebased and force-pushed:

```
Failing Scenarios:
cucumber features/speeches.feature:19 # Scenario: Trying to save a speech that has been changed by another user
```

I'm guessing this might be an intermittent error as it passes locally. I've kicked the build to see whether it'll pass on a subsequent run.

chrisroos:
The tests all passed in the subsequent build so I'm going to get this merged.

PR #3543, opened 2017-11-20
Monitor external links added to editions

SebAshton:
GOV.UK must be a reputable source of information for users. Their safety, security and privacy must be protected while using the site so all links to external links must also be safe, and secure.

Pushing data into the service is the first step to achieving this goal.

### DIFF
--- a/app/services/create_edition_link_monitor.rb
+++ b/app/services/create_edition_link_monitor.rb
@@ -1,0 +1,46 @@
+class CreateEditionLinkMonitor
+  def initialize(edition)
+    @edition = edition
+    @uris = extract_links
+  end
+
+  def perform!
+    if can_perform?
+      upsert_resource_monitor
+
+      true
+    end
+  end
+
+  def can_perform?
+    !failure_reason
+  end
+
+  def failure_reason
+    "This edition has no links" unless has_links?
+  end
+
+private
+
+  attr_reader :edition, :uris
+
+  def upsert_resource_monitor
+    Whitehall.link_checker_api_client.upsert_resource_monitor(
+      uris,
+      :whitehall,
+      edition_reference
+    )
+  end
+
+  def edition_reference
+    "edition:#{edition.id}"
+  end
+
+  def has_links?
+    extract_links.count > 0
+  end
+
+  def extract_links
+    @extract_links ||= Govspeak::LinkExtractor.new(edition.body).links
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 20170821152429) do
 
-  create_table "about_pages", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+  create_table "about_pages", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
     t.integer  "topical_event_id"
     t.string   "name"
     t.text     "summary",             limit: 65535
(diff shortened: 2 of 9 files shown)


test/unit/services/create_edition_link_monitor_test.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class CreateEditionLinkMonitorTest < ActiveSupport::TestCase
+  setup do
+    Whitehall.link_checker_api_client.stubs(:upsert_resource_monitor).returns {}
+  end
+
+  def expect_link_checker_api_call
+    Whitehall.link_checker_api_client.expects(:upsert_resource_monitor)
+      .with(any_parameters)
+  end
```
tuzz: Would it be worth asserting the arguments to this call to ensure the links have been extracted correctly?

tuzz: I wonder if it'd be worth pushing some of these assertion-y things into [gds-api-adapters](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/link_checker_api.rb)? [This](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/publishing_api_v2.rb#L167-L170) is an example of an `assert` method in this style.

The benefit(?) is that it tests the assertion at the HTTP layer rather than on the client, so it tests the integration a little bit past the client boundary, rather than at it. It also promotes re-use across the code base / other code bases. What are your thoughts?

app/services/create_edition_link_monitor.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  def can_perform?
+    !failure_reason
+  end
+
+  def failure_reason
+    "This edition has no links" unless has_links?
+  end
```
tuzz: Initially I thought it'd be better not to enqueue jobs for things that don't have links, but it's nice to keep the knowledge of how to extract links in here so I guess that makes sense. 👍

app/services/create_edition_link_monitor.rb
```diff
@@ -0,0 +1,48 @@
...
+  def initialize(edition)
+    @edition = edition
+    @uris = extract_links
+  end
+
+  def perform!
+    if can_perform?
+      ActiveRecord::Base.transaction do
+        upsert_resource_monitor
+      end
```
tuzz: Why is the API call transactionalised?

test/unit/workers/edition_create_link_monitor_worker_test.rb
```diff
@@ -0,0 +1,22 @@
...
+    CreateEditionLinkMonitor.any_instance.stubs(:perform!).returns id: 1
+  end
+
+  test "#perform calls CreateEditionLinkMonitor when valid" do
+    CreateEditionLinkMonitor.expects(:new).with(@edition).returns(@link_monitor)
+
+    worker_invocation = EditionCreateLinkMonitorWorker.new.perform(@edition.id)
+
+    assert worker_invocation
+    assert_equal worker_invocation, id: 1
```
tuzz: I don't understand why this returns `{ id: 1 }`?

`EditionCreateLinkMonitorWorker#perform` returns `CreateEditionLinkMonitor#perform!` and that returns `can_perform?` which is presumably a boolean?

tuzz: Oh, nevermind. It's stubbed.

We should probably swap the order of `expected`, `actual` around.

db/schema.rb
```diff
@@ -1183,8 +1182,8 @@
   end
 
   create_table "worldwide_services", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
-    t.string   "name",            null: false
-    t.integer  "service_type_id", null: false
+    t.string   "name",            default: "", null: false
```
thomasleese: Don't think this change belongs here.

SebAshton: 😕  yea that looks odd

app/services/create_edition_link_monitor.rb
SebAshton: Just to protect from mid-call updates, this could be overly defensive though.

SebAshton:
Closing this PR for now until we need it

PR #3546, opened 2017-11-20
Fix deprecation warnings

chrisroos:
The deprecation warnings noise in the Jenkins output makes it harder to see any real failures.

Deprecation warnings that aren't addressed in this branch because I don't know what to do about them:

* Deprecation warning: Custom radio buttons and checkboxes (released in GOV.UK Elements 3.0.0) no longer require this JavaScript.

* DEPRECATION WARNING: Method sort_by is deprecated and will be removed in Rails 5.1, as `ActionController::Parameters` no longer inherits from hash. Using this deprecated behavior exposes potential security problems. If you continue to use this method you may be creating a security vulnerability in your app that can be exploited. Instead, consider using one of these documented methods which are not deprecated: http://api.rubyonrails.org/v5.0.4/classes/ActionController/Parameters.html (called from reorder at /var/lib/jenkins/workspace/ll_fix-deprecation-warnings-JEGZJVEIVBK6GVMYWBUEUSCML5UH5GK4KLZOVJ46NXDKJVYUBEAA/app/controllers/admin/feature_lists_controller.rb:9)


chrisroos:
Thanks, @floehopper! I've rebased on master and force pushed in preparation for merging.

PR #3547, opened 2017-11-21
Link checker api integration

camillavk:
Integrating the [LinkChecker API](https://github.com/alphagov/link-checker-api/blob/master/docs/api.md) into Whitehall to ensure links on editions are safe and secure. This integration was initially started in April but was closed due to a lack of dev time. It was resurrected in November, requiring a rebase with the master branch to catch it up with the changes that had been made in the meantime.

More detailed documentation about this integration can be found in the [Link Checker API integration docs](docs/link_checker_api_integration.md)

Areas that were incomplete when the PR was resurrected were:
- [x] ~When publishing edition use LinkCheckerApiService~ Links were only being checked for presence when publishing an edition so retaining current functionality and not changing this
- [x] Verify webhook
- [x] `broken_link_reporter` tests were broken


db/schema.rb
```diff
@@ -656,7 +661,37 @@
     t.integer  "organisation_id"
   end
 
+<<<<<<< HEAD
```
thomasleese: Is it possible we can get rid of the merge conflicts here?

db/migrate/20170704073821_add_problem_summary_and_suggested_fix_to_link_checker.rb
```diff
@@ -0,0 +1,6 @@
+class AddProblemSummaryAndSuggestedFixToLinkChecker < ActiveRecord::Migration
+  def change
+    add_column :link_checker_api_report_links, :problem_summary, :text
```
thomasleese: I wonder if we should make this `null: false` (with a default of `""` to backfill the records), because then we only have one kind of empty string not two (`nil` and `''`).

app/models/edition.rb
```diff
@@ -37,6 +37,7 @@ class Edition < ApplicationRecord
   has_many :authors, through: :edition_authors, source: :user
   has_many :classification_featurings, inverse_of: :edition
   has_many :link_check_reports, as: :link_reportable, class_name: LinkCheckerApiReport
+  has_many :links_reports, as: :link_reportable
```
thomasleese: I assume the plan here is to remove this in a further PR? Could we add a comment just to explain this is temporary.

features/step_definitions/admin_broken_link_reporting_steps.rb
```diff
@@ -30,6 +31,7 @@
 end
 
 Then(/^I should see a list of the broken links$/) do
+  binding.pry
```
thomasleese: Extra `binding.pry` here.

features/support/link_checker_api_helper.rb
```diff
@@ -6,20 +6,37 @@ module LinkCheckerApiHelper
...
-    post "/government/admin/link_checker_api_callback", hash
+    body = link_checker_api_batch_report_hash(batch_report_hash)
+
+    headers = {
+      "Content-Type": "application/json",
+      "X-LinkCheckerApi-Signature": generate_signature(body.to_json, Rails.application.secrets.link_checker_api_secret_token)
+    }
+
+    page.driver.headers = headers
+    binding.pry
```
thomasleese: Extra `binding.pry` here.

thomasleese:
Would it be possible to include a detailed description on this PR as well explaining how the integration works? It would be great if you could update or add any relevant document as well: https://github.com/alphagov/whitehall/tree/master/docs

db/migrate/20170704073821_add_problem_summary_and_suggested_fix_to_link_checker.rb
SebAshton: After trying to set these attributes - and getting errors - it turns out we can't set a default on a `BLOB/TEXT` field in MySQL:

>BLOB and TEXT columns cannot have DEFAULT values.

From the [docs](https://dev.mysql.com/doc/refman/5.7/en/blob.html)

thomasleese: Oh, that's annoying 😢 Can we back fill the current reports manually with empty strings and then make it `null: false`?

SebAshton: We could, Would this not then throw an error when a link comes back as ok from the service?
 As `:problem_summary` or `:suggested_fix` come back as `nil` in the payload.

SebAshton: Will set an default in the fetch to avoid this

camillavk:
I think it will be hard to tidy up the PR in such a way that we no longer need the `Rebase with master` commit while retaining an audit of what has happened and why. Would it make it better if we added an explanation to the PR description as to why a rebase was required (i.e. resurrecting a previously closed PR etc)

thomasleese:
Yeah, I understand the problem there.

I was thinking that it should be possible to at least get rid of the merge conflicts from https://github.com/alphagov/whitehall/pull/3547/commits/a54c5405fe1602e85cba49068a43d85be58f3494 so that if someone needed to rollback to that commit for whatever reason, it still works.

If there still needs to be a rebase commit after that (which I suspect there will), then that would be okay with a small description about why it's there.

bevanloon:
@camillavk @SebAshton should this PR be closed now that https://github.com/alphagov/whitehall/pull/3557 is open?

bevanloon:
closing in favour of https://github.com/alphagov/whitehall/pull/3557

PR #3552, opened 2017-11-21
Remove organisation logos from NFS

chrislo:
See: alphagov/asset-manager#272

Organisation logos are now being saved to, and served from[1], Asset
Manager. Any new logos are not being saved to NFS[2]. Existing logos
have been migrated to Asset Manager[3]. This means we can remove these
assets from the file system.

This PR adds `AssetRemover#remove_organisation_logos` to remove
the directory that "clean" organisation logos were moved to after
virus scanning, and all of its contents. We also add a Rake task `asset_manager:remove_organisation_logos` to allow this code to be run via Jenkins. 

[1] alphagov/asset-manager#283
[2] alphagov/asset-manager#274
[2] alphagov/asset-manager#276

### DIFF
--- a/lib/asset_remover.rb
+++ b/lib/asset_remover.rb
@@ -1,6 +1,8 @@
 class AssetRemover
   def remove_organisation_logos
     target_dir = File.join(Whitehall.clean_uploads_root, 'system', 'uploads', 'organisation', 'logo')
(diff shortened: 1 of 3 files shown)


chrisroos:
I've rebased this on master in preparation for merging.

PR #3553, opened 2017-11-22
Improve rendering on developer docs

tuzz:
GitHub treats newlines within paragraphs as spaces
whereas developer docs preserves the newlines.


kevindew:
That's weird, developer docs [shouldn't do that](http://spec.commonmark.org/0.27/#soft-line-break) is there a bug in the markdown parsing there?

tuzz:
I've opened an issue: https://github.com/alphagov/govuk-developer-docs/issues/587

PR #3555, opened 2017-11-23
Update README

rubenarakelyan:
Update the README to be more in line with our other apps, link to all the other whitehall docs and remove information about finders that has been moved to a separate document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,104 +1,56 @@
-# Whitehall App
+# whitehall
 
-"Whitehall" is the code name for the
-[Inside Government](https://www.gov.uk/government/) project, which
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,104 +1,56 @@
...
 
 ## Running the application
 
-The application can be started with
-
 ```
-bundle exec rails s
+$ ./startup.sh
 ```
+If you are using the GDS development virtual machine then the application will be available on the host at http://whitehall-admin.dev.gov.uk/
```
thomasleese: Might be worth linking to the docs for the dev vm here?

rubenarakelyan: Yup, done

PR #3554, opened 2017-11-23
Add docs for LinkCheckerApi integration

camillavk:

PR #3557, opened 2017-11-23
Link Checker API Integration

SebAshton:
Integrating the [LinkChecker API](https://github.com/alphagov/link-checker-api/blob/master/docs/api.md) into Whitehall to ensure links on editions are safe and secure. This integration was initially started in April but was closed due to a lack of dev time. It was resurrected in November, requiring a rebase with the master branch to catch it up with the changes that had been made in the meantime.

This PR is only modifying the current on demand link checking functionality and does not address the link reporter functionality. This will be addressed in a future PR.

More detailed documentation about this integration can be found in the [Link Checker API integration docs](docs/link_checker_api_integration.md)

Closes #3554 
Closes #3547


app/models/edition.rb
```diff
@@ -36,6 +36,7 @@ class Edition < ApplicationRecord
   has_many :edition_authors, dependent: :destroy
   has_many :authors, through: :edition_authors, source: :user
   has_many :classification_featurings, inverse_of: :edition
+  has_many :link_check_reports, as: :link_reportable, class_name: LinkCheckerApiReport
   has_many :links_reports, as: :link_reportable
```
thomasleese: Can we avoid 9bdd5e8 by just getting rid of the removal change in the earlier commit?

config/routes.rb
```diff
@@ -401,6 +401,7 @@ def external_redirect(path_prefix, target)
         end
 
         resources :sitewide_settings
+        post "/link-checker-api-callback" => "link_checker_api#callback"
```
thomasleese: Can we get rid of 3667027 and fix the path in the earlier commits?

config/initializers/link_checker_api.rb
```diff
@@ -0,0 +1,6 @@
+require "gds_api/link_checker_api"
+
+Whitehall.link_checker_api_client = GdsApi::LinkCheckerApi.new(
+  Plek.find("link-checker-api"),
+  bearer_token: ENV["LINK_CHECKER_API_BEARER_TOKEN"] || "example"
```
thomasleese: We don't need a bearer token when interacting with the Link Checker API since it's internal.

config/secrets.yml
```diff
@@ -12,11 +12,14 @@
...
 
 test:
   secret_key_base: 70ebfd8ab10715f055bea44e49057544e6d232e374ba191b54a3f490d86cc8a5327d329d8cb6014b3968ade5efacc48563ec16d85d2305ee72a973307e130506
+  link_checker_api_secret_token: ENGgg9TEmRRxZg9FEAC6qrVu
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
+  link_checker_api_secret_token: <%= ENV["LINK_CHECKER_API_SECRET_TOKEN"] %>
```
thomasleese: We'll need to add this to [`govuk-puppet`](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/whitehall.pp#L128) to make sure it gets set.

app/models/link_checker_api_report.rb
```diff
@@ -0,0 +1,34 @@
...
+  def in_progress?
+    !completed?
+  end
+
+  def has_problems?
+    links.any? { |l| %w(caution broken).include?(l.status) }
+  end
+
+  def broken_links
+    links.select { |l| l.status == "broken" }
```
tuzz: Perhaps `status` could be an [enum](http://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html) so that dependent classes are a bit less coupled to the specifics of this field?

app/services/link_checker_api_service.rb
```diff
@@ -0,0 +1,20 @@
...
+    Govspeak::LinkExtractor.new(reportable.body).links
+  end
+
+  def self.check_links(reportable, webhook_uri)
+    uris = extract_links(reportable)
+    raise "Reportable has no links to check" unless uris.count > 0
+
+    batch_report = Whitehall.link_checker_api_client.create_batch(
+      uris,
+      webhook_uri: webhook_uri,
```
thomasleese: We need to pass in the `webhook_secret_token` here.

app/controllers/admin/link_checker_api_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    return head :bad_request unless given_signature
+    body = request.raw_post
+    signature = OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new("sha1"), webhook_secret_token, body)
+    head :bad_request unless Rack::Utils.secure_compare(signature, given_signature)
+  end
+
+  def webhook_secret_token
+    Rails.application.secrets.link_checker_api_secret_token
+  end
+end
```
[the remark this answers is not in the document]
camillavk: This sounds like a good idea, but I think we would prefer to implement it in a following PR so we can get this one merged in (same for your following comment)

SebAshton:
Good spot @kevindew I've fixed that. Thanks.

@bevanloon I believe @camillavk had previously addressed your comments.

PR #3562, opened 2017-11-27
Add seeds to allow for use within end-to-end testing

RyanMacG:
We require a user to satisfy mocked GDS-SSO and an organisation for use as part of the manuals publishing flow. The manuals publisher fetches the organisation for the publishing user based on the slug assigned to them (we mirror this organisation slug in the manuals publisher)

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,22 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
-#
-# Examples:
-#
-#   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
-#   Mayor.create(name: 'Emanuel', city: cities.first)
+if User.where(name: "Test user").present?
+  puts "Skipping because user already exists"
+else
+  gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+  User.create!(
+    name: "Test user",
+    permissions: ["signin", "GDS Admin"],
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -1,7 +1,25 @@
...
+    name: "Test Organisation",
+    slug: "test-organisation",
+    acronym: "TO",
+    organisation_type_key: :other,
+    logo_formatted_name: "Test"
+  )
+end
+
+
+
```
kevindew: We seem to have a bunch of empty lines here

db/seeds.rb
```diff
@@ -1,7 +1,25 @@
...
-#
-#   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
-#   Mayor.create(name: 'Emanuel', city: cities.first)
+if User.where(name: "Test user").present?
+  puts "Skipping because user already exists"
+else
+  gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+  User.create!(
+    name: "Test user",
+    permissions: ["signin", "GDS Admin"],
```
kevindew: Are you confident these are the perms you need? I don't know Whitehall perms very well myself so not sure if GDS Admin supersedes ones like Editor and GDS Editor - @gpeng do you happen to know?

RyanMacG: We found using `gds_editor` didn't allow us to create organisations within the frontend so we needed it for manual testing at least

gpeng: I think `GDS Admin` is about the highest permission so unless it's the permissions we're testing, which seemingly it isn't, that's probably a good choice.

PR #3564, opened 2017-11-27
Do not overwrite invisible draft taxons

davidbasalla:
For https://trello.com/c/MNOyFuby/186-fix-whitehall-admin-ui-overwrites-draft-links-when-editing-taxon-tagging

Fixes the problem that invisible taxons that were tagged to content (eg from legacy taxonomies) would be overwritten when a publisher edits the taxon taggings for a piece of content.

The commits doing most of the work are https://github.com/alphagov/whitehall/pull/3564/commits/71af956f5784d178d24993776c084908b618544f and https://github.com/alphagov/whitehall/pull/3564/commits/817cb69f8dea4bc20faee79ad0379b095e6804ea. Essentially we compare the content item's tagged taxons to all taxons from the govuk taxonomy. By filtering out the matching published and visible draft taxons, the remainder makes up the invisible draft taxons (this doesn't seem ideal, but currently we have no other way of determining the 'state' (published/visible draft/invisible draft) of a content item's taxons). Finally we add the list of invisible taxon content ids to the edit tag form via a hidden field, and add those links to the final payload before the patch links call gets made to the PublishingAPI.

### DIFF
--- a/app/models/taxonomy_tag_form.rb
+++ b/app/models/taxonomy_tag_form.rb
@@ -1,7 +1,7 @@
 class TaxonomyTagForm
   include ActiveModel::Model
 
-  attr_accessor :selected_taxons, :all_taxons, :content_id, :previous_version
+  attr_accessor :selected_taxons, :invisible_taxons, :all_taxons, :content_id, :previous_version
 
   def self.load(content_id)
     begin
@@ -41,7 +41,7 @@ class TaxonomyTagForm
       .publishing_api
       .patch_links(
         content_id,
-        links: { taxons: most_specific_taxons },
+        links: { taxons: most_specific_taxons + invisible_taxons },
         previous_version: previous_version
       )
   end
--- a/app/models/taxonomy_tag_form.rb
+++ b/app/models/taxonomy_tag_form.rb
@@ -36,32 +36,6 @@ class TaxonomyTagForm
     selected_taxons - (published_taxons + visible_draft_taxons)
   end
 
-  def publish!
-    Services
-      .publishing_api
-      .patch_links(
-        content_id,
-        links: { taxons: most_specific_taxons + invisible_taxons },
-        previous_version: previous_version
-      )
-  end
-
-  # Ignore any taxons that already have a more specific taxon selected
-  def most_specific_taxons
-    all_taxons.each_with_object([]) do |taxon, list_of_taxons|
-      content_ids = taxon.descendants.map(&:content_id)
-
-      any_descendants_selected = selected_taxons.any? do |selected_taxon|
-        content_ids.include?(selected_taxon)
-      end
-
-      unless any_descendants_selected
-        content_id = taxon.content_id
-        list_of_taxons << content_id if selected_taxons.include?(content_id)
-      end
-    end
-  end
-
 private
 
   def govuk_taxonomy
--- a/app/models/taxonomy_tag_form.rb
+++ b/app/models/taxonomy_tag_form.rb
@@ -24,6 +24,18 @@ class TaxonomyTagForm
     )
   end
 
+  def published_taxons
+    govuk_taxonomy.matching_against_published_taxons(selected_taxons)
+  end
+
+  def visible_draft_taxons
+    govuk_taxonomy.matching_against_visible_draft_taxons(selected_taxons)
+  end
+
+  def invisible_draft_taxons
+    selected_taxons - (published_taxons + visible_draft_taxons)
+  end
+
   def publish!
     Services
       .publishing_api
@@ -48,5 +60,11 @@ class TaxonomyTagForm
         list_of_taxons << content_id if selected_taxons.include?(content_id)
       end
     end
+  end
+
+private
+
+  def govuk_taxonomy
+    @_taxonomy ||= Taxonomy::GovukTaxonomy.new
   end
 end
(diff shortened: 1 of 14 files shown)


app/models/taxonomy_tag_form.rb
```diff
@@ -24,29 +24,47 @@ def self.load(content_id)
...
-        content_id = taxon.content_id
-        list_of_taxons << content_id if selected_taxons.include?(content_id)
-      end
+  def invisible_draft_taxons
+    selected_taxons - published_taxons - visible_draft_taxons
+  end
+
+private
+
+  def filter_against_taxonomy_branch(selected_taxons_content_ids, taxon, matched_taxons)
```
davidbasalla: I'm going to move this method to the GovukTaxonomy class instead...

cbaines:
You've got "FIX" in the bottom of the commit message of the 2nd commit " Add taxon retrieval methods to TaxonomyTagForm", does that mean something?

davidbasalla:
@cbaines good catch, and no, it was just a leftover from some rebasing

lib/taxonomy/govuk_taxonomy.rb
```diff
@@ -21,10 +21,39 @@ def all_taxons
...
+      @_draft_visible_taxons ||= matching_against_taxonomy_branches(taxons, draft_child_taxons)
+    end
+
   private
 
     def build_tree(taxon_hash)
       @tree_builder_class.new(taxon_hash).root_taxon
     end
+
+    def matching_against_taxonomy_branches(taxons, taxononomy_branches)
```
cbaines: `s/taxononomy/taxonomy` (unless that is a word??)

lib/taxonomy/govuk_taxonomy.rb
```diff
@@ -21,10 +21,39 @@ def all_taxons
...
+      matching_taxons = []
+      taxononomy_branches.each do |branch|
+        matched_taxons = filter_against_taxonomy_branch(taxons, branch, [])
+        matching_taxons.concat(matched_taxons) if matched_taxons.any?
+      end
+      matching_taxons
```
cbaines: This looks like a good candidate for a `flat_map`, e.g.

```ruby
taxonomy_branches.flat_map do |branch|
  filter_against_taxonomy_branch(taxons, branch, [])
end
```

davidbasalla: Sounds good, will give it a go!

lib/taxonomy/govuk_taxonomy.rb
```diff
@@ -21,10 +21,39 @@ def all_taxons
...
+    def filter_against_taxonomy_branch(selected_taxons_content_ids, taxon, matched_taxons)
+      if selected_taxons_content_ids.include?(taxon.content_id)
+        matched_taxons << taxon.content_id
+      end
+
+      taxon.children.each do |child_taxon_branch|
+        filter_against_taxonomy_branch(selected_taxons_content_ids, child_taxon_branch, matched_taxons)
+      end
+
+      matched_taxons
```
cbaines: This mutates the `matched_taxons` array if I'm not mistaken, which is why I guess you can get away with not using the return value when doing the recursive call.

Mutation isn't great in my opinion, as it often makes code unobviously complex, but this is private, so it's not too bad here. Especially using `flat_map`, I don't think this would be too difficult to write avoiding mutation, but I'll leave the decision up to you.

test/support/taxonomy_helper.rb
```diff
@@ -94,7 +98,7 @@ def draft_taxon_1
     {
       "title" => "About your organisation",
       "base_path" => "/about-your-organisation",
-      "content_id" => "draft_taxon_1",
+      "content_id" => draft_taxon_1_content_id,
```
cbaines: This doesn't match up with the `draft_taxon_1_content_id` method in this module, which I find confusing?

app/services/edition_taxon_publisher.rb
```diff
@@ -0,0 +1,32 @@
+class EditionTaxonPublisher
```
cbaines: Maybe `EditionTaxonLinkPatcher` would more clearly indicate what this class does?

cbaines: I realise that doesn't match the previous method name, but I think that is another good change you'd be making.

davidbasalla: Sounds good

test/support/taxonomy_helper.rb
davidbasalla: I'm not sure what you mean, this aims to follow what's being done for the other taxons and their content_ids, eg `child_taxon_content_id`

cbaines: The `draft_taxon_1_content_id` method gives `"draft_taxon_1"` as the content id [1], which doesn't match the value here. Does that help clarify my confusion?

1; https://github.com/alphagov/whitehall/pull/3564/files/b3a19a50f5fcb9145b0b452d9b875063f5b92782..6d626ae9e2eb598877567e563136fc046b07a9b6#diff-e2203318ea7fdfd069f24a1c05980ce0R19

cbaines: I think I've just been reading it wrong, nothing to see here, sorry for the confusion.

davidbasalla:
Thanks for the review @cbaines. I've addressed your comments (and rebased)

- removed 'FIX' from commit message
- fixed `taxononomy` typo
- reworked the recursive method to not mutate the `matched_taxons` array and used `flat_map`
- renamed `EditionTaxonPublisher` to `EditionTaxonLinkPatcher`

PR #3566, opened 2017-11-28
Add broken link fix on hover

tsteer:
This adds the suggested fix for the broken link when hovering over a broken link in the Whitehall edition link checker. This is to help speed up the process of fixing links by providing the user with a faster way of finding the current issue

### DIFF
--- a/app/views/admin/link_check_reports/_link_check_report.html.erb
+++ b/app/views/admin/link_check_reports/_link_check_report.html.erb
@@ -19,7 +19,7 @@
       <p>We’ve found links in this document that may be broken:</p>
       <ul>
         <% report.broken_links.each do |link| %>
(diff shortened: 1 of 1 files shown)


kevindew:
HI @tsteer can we revert this please? We would very much prefer it if you had these reviewed by someone within GOV.UK  - ideally the team that own Whitehall - before these being merged. As this is a user interface change too it'd be good to have gone through the product owner here.

Also it's worth noting that the suggested fix might not always exist and as the link is truncated before this users may well be using this to differentiate between links that are truncated identically. On a semantic level it doesn't really feel like a title used be used for this additional information regarding the link - you would expect it would tell you more about the link rather than what you need to do with it.

nickcolley:
Might not be related but see https://silktide.com/i-thought-title-text-improved-accessibility-i-was-wrong/

PR #3567, opened 2017-11-28
[DO NOT MERGE] Spike content review

chao-xian:
This is a proof of concept prototype and must not be merged into master!

Shows a list of content items needing review (review by date expiring
in the next 7 days). Then links through to allow you update that
review date. All review date data is stored in the Content
Performance Manager.

And link to the edit functions for each item. This is part of a spike to
prototype "life expectancy" of content.

This goes with this PR in Content Performance Manager: https://github.com/alphagov/content-performance-manager/pull/413

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ gem 'activemodel-serializers-xml'
 gem 'deprecated_columns', '~> 0.1.1'
 gem 'record_tag_helper', '~> 1.0'
 gem 'govuk_ab_testing', '~> 2.4x'
+gem 'rest-client'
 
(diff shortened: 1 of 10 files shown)


chao-xian:
[Spike #1: (Management - Life expectancy) If we set a review date on content then we know what we need to review (and audit) at right time and feed into archive process because we can use this to alert publishers and time it with key dates (e.g budget) [2 people, 2 days, dev and competitor analysis]](https://trello.com/c/A3cMuEY6/658-spike-1-management-life-expectancy-if-we-set-a-review-date-on-content-then-we-know-what-we-need-to-review-and-audit-at-right-tim)

PR #3569, opened 2017-11-29
Add link fix below broken link

tsteer:
This separates links based on status: broken or caution. It adds the suggested fix as an expansion below the link in the Whitehall edition link checker. This is to help speed up the process of fixing links by providing the user with a faster way of finding the current issue. 

screen shot 2017-12-04 at 16 00 35

### DIFF
--- a/app/assets/stylesheets/admin/helpers/_broken_links_report.scss
+++ b/app/assets/stylesheets/admin/helpers/_broken_links_report.scss
@@ -5,3 +5,45 @@
     }
   }
 }
+
+.broken-links-report {
+  .issue-summary {
+    font-weight: normal;
+  }
+
+  .issue-list {
+    list-style: none;
+    padding-left: 0;
+
+    li {
+      margin-top: 5px;
+    }
+  }
+
+  .issue-status-description {
+    margin-top: 10px;
+  }
+
+  .issue-list + .issue-status-description {
+    margin: 15px 0;
+  }
+
+  .display-issue-details {
+    display: block;
+    font-weight: bold;
+    padding-top: 5px;
+  }
+
+  .display-issue-details::-webkit-details-marker {
+    display: none;
+  }
+
+  .display-issue-details::before {
+    content: '\25B6';
+    padding-right: 3px;  
+  }
+
+  details[open] > .display-issue-details::before {
+    content: '\25BC';
+  }
+}
--- a/app/views/admin/link_check_reports/_link_check_report.html.erb
+++ b/app/views/admin/link_check_reports/_link_check_report.html.erb
@@ -13,15 +13,27 @@
                   class: 'js-broken-links-refresh js-hidden',
                   remote: true %>
     </section>
-  <% elsif report.broken_links.present? %>
+  <% elsif report.broken_links.any? || report.caution_links.any? %>
     <section class="alert alert-warning">
-      <h3 class="remove-top-margin">Broken links</h3>
-      <p>We’ve found links in this document that may be broken:</p>
-      <ul>
-        <% report.broken_links.each do |link| %>
-          <li><%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %></li>
-        <% end %>
-      </ul>
+      <h3 class="remove-top-margin"><%= t "broken_links.title" %></h3>
+      <% report.links.sort_by(&:status).group_by(&:status).each do |status, links| %>
+        <% next unless %w(broken caution).include? status %>
+        <p class="issue-status-description"><%= t "broken_links.#{status}.subheading" %></p>
+        <ul class="issue-list">
+          <% links.each do |link| %>
+            <li>
+              <%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %>
+              <details>
+                <summary class="display-issue-details">See more details about this link</summary>
+                <p class="issue-summary"><%= link.problem_summary %></p>
+                <% if link.suggested_fix %>
+                  <p class="issue-summary">Suggested fix: <%= link.suggested_fix %></p>
+                <% end %>
+              </details>
+            </li>
+          <% end %>
+        </ul>
+      <% end %>
       <%= render 'admin/link_check_reports/form', reportable: report.link_reportable, button_text: 'Check again' %>
     </section>
   <% else %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -486,3 +486,9 @@ en:
     transparency: See all our transparency data
     written_statement: See all our written statements to Parliament
     authored_article: See all our authored articles
+  broken_links:
+    title: Broken links
+    broken: 
+      subheading: We’ve found links in this document that may be broken
+    caution:
+      subheading: We’ve found some issues with links that may indicate problems, you should apply caution in linking to them
--- a/config/locales/he.yml
+++ b/config/locales/he.yml
@@ -370,6 +370,12 @@ he:
     opendocument:
       help_html:
     see_more:
+  broken_links:
+    broken:
+      subheading:
+    caution:
+      subheading:
+    title:
   change_notes:
     page_history:
     published_at:
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -471,6 +471,12 @@ ar:
     opendocument:
       help_html:
     see_more:
+  broken_links:
+    broken:
+      subheading:
+    caution:
+      subheading:
+    title:
   change_notes:
     page_history:
     published_at:
--- a/config/locales/az.yml
+++ b/config/locales/az.yml
@@ -60,6 +60,12 @@ az:
     opendocument:
       help_html:
     see_more:
+  broken_links:
+    broken:
+      subheading:
+    caution:
+      subheading:
+    title:
   change_notes:
     page_history:
     published_at:
--- a/config/locales/be.yml
+++ b/config/locales/be.yml
@@ -369,6 +369,12 @@ be:
     opendocument:
       help_html:
     see_more:
+  broken_links:
+    broken:
+      subheading:
+    caution:
+      subheading:
+    title:
   change_notes:
     page_history:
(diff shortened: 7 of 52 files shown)


kevindew:
Hey - thanks for reverting https://github.com/alphagov/whitehall/pull/3566

Also super useful that you included a screenshot in this PR 👍 

So I think you're going to hit an issue with this implementation of the idea you've got. The suggested fix field wasn't intended to be isolated from the description of what the problem is and thus will often be of quite obtuse without the context of whats wrong

To demonstrate here are some example suggested fixes:

- Determine if this is a temporary issue or the resource is no longer available. 
- Contact the site administrator to see if they have an issue that can be fixed.
- Link to an alternative location of this resource, which doesn't require credentials.
- Speak to your system administrator.

Which would all be difficult to a user to understand, whereas with the context of what's wrong you have:

- This redirects to a page that isn't loading. Suggested fix: Determine if this is a temporary issue or the resource is no longer available. 
- This page has a redirect loop and won't open. Suggested fix: Contact the site administrator to see if they have an issue that can be fixed.
- This link contains login credentials which you may not want to share publicly. Suggested fix: Link to an alternative location of this resource, which doesn't require credentials.
- Check failed. Suggested fix: Speak to your system administrator.

I do appreciate that this a challenge given the amount of UI space available.

app/views/admin/link_check_reports/_link_check_report.html.erb
```diff
@@ -19,7 +19,12 @@
       <p>We’ve found links in this document that may be broken:</p>
       <ul>
         <% report.broken_links.each do |link| %>
-          <li><%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %></li>
+          <li>
+            <%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %>
+            <% if link.suggested_fix %>
+              <p class="suggested-fix"><%= link.problem_summary %>. Suggested fix: <%= link.suggested_fix %></p>
```
thomasleese: I think it would be good to always show the problem summary, but only show the suggested fix if there is one.

app/views/admin/link_check_reports/_link_check_report.html.erb
```diff
@@ -17,9 +17,17 @@
...
+      <ul class="issue-list">
         <% report.broken_links.each do |link| %>
-          <li><%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %></li>
+          <li>
+            <%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %>
+            <% if link.suggested_fix %>
+            <details>
+              <summary class="issue-summary">See more details about this link</summary>
+              <p class="issue-fix"><%= link.problem_summary %>. Suggested fix: <%= link.suggested_fix %></p>
+            </details
```
thomasleese: Missing `>` here.

app/views/admin/link_check_reports/_link_check_report.html.erb
```diff
@@ -17,9 +17,17 @@
     <section class="alert alert-warning">
       <h3 class="remove-top-margin">Broken links</h3>
```
thomasleese: Could we group the links into broken links and caution links? And show the broken links at the top?

app/views/admin/link_check_reports/_link_check_report.html.erb
```diff
@@ -17,9 +17,17 @@
...
       <p>We’ve found links in this document that may be broken:</p>
-      <ul>
+      <ul class="issue-list">
         <% report.broken_links.each do |link| %>
-          <li><%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %></li>
+          <li>
+            <%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %>
+            <% if link.suggested_fix %>
```
thomasleese: I think we always want to see the problem summary, but only include the suggested fix part if there is a suggested fix.

app/assets/stylesheets/admin/helpers/_broken_links_report.scss
```diff
@@ -5,3 +5,36 @@
...
+    }
+  }
+
+  .issue-summary {
+    display: block;
+    font-weight: bold;
+    padding-top: 1rem;
+  }
+
+  .issue-summary::-webkit-details-marker {
```
nickcolley: Can you explain what this is doing? `::-webkit-details-marker` looks to be removing the triangle that indicates if the detail element is open or close, but why do we need to remove and re add it?

app/views/admin/link_check_reports/_link_check_report.html.erb
tsteer: I don't think we differentiate between broken and caution links?

thomasleese: The link reports get a status of `broken` or `caution`. - https://github.com/alphagov/link-checker-api/blob/master/docs/api.md#linkreport-entity

app/assets/stylesheets/admin/helpers/_broken_links_report.scss
tsteer: I think webkit is the only browser to implement the drop down arrow by default, so by hiding it we are able to give a more consistent cross browser look and feel

config/locales/en.yml
```diff
@@ -486,3 +486,11 @@ en:
     transparency: See all our transparency data
     written_statement: See all our written statements to Parliament
     authored_article: See all our authored articles
+  broken_links:
+    broken: 
+      title: Broken links
+      subheading: There are broken links in this document
```
thomasleese: Would be good to stick with "We’ve found links in this document that may be broken:" here.

config/locales/en.yml
```diff
@@ -486,3 +486,11 @@ en:
...
+    caution:
+      title: Links with warnings
+      subheading: There are links with warnings in this document
```
thomasleese: This could be "We've found some issues with links that may indicate problems, you should apply caution in linking to them:"

config/locales/en.yml
```diff
@@ -486,3 +486,11 @@ en:
     transparency: See all our transparency data
     written_statement: See all our written statements to Parliament
     authored_article: See all our authored articles
+  broken_links:
+    broken: 
+      title: Broken links
+      subheading: There are broken links in this document
+    caution:
+      title: Links with warnings
```
thomasleese: I don't think it's necessary to have a title on the warning, I think it would be better something like:

# Broken links
> We’ve found links in this document that may be broken:

- url
- ...

> We've found some issues with links that may indicate problems, you should apply caution in linking to them:

- url
- ...

Similar to the design in https://trello.com/c/qKlOIUfB/36-make-whitehall-reports-nice-for-users

app/views/admin/link_check_reports/_link_check_report.html.erb
```diff
@@ -15,11 +15,25 @@
     </section>
   <% elsif report.broken_links.present? %>
```
kevindew: I see that you handle links with caution in the loop but I think this implies that if there are only caution links they won't be shown. Is that the intention?

config/locales/he.yml
```diff
@@ -2,245 +2,151 @@ he:
   document:
     headings:
       attachments:
-        one: "מסמך"
-        two:
-        many:
-        other: "מסמכים"
```
kevindew: Looks like we've lost a bunch of hebrew translations here, which I don't think should be removed.

app/views/admin/link_check_reports/_link_check_report.html.erb
```diff
@@ -15,11 +15,25 @@
...
+              <%= link_to link.uri.truncate(50), link.uri, title: link.uri, class: 'link-inherit' %>
+              <details>
+                <summary class="display-issue-details">See more details about this link</summary>
+                <p class="issue-summary"><%= link.problem_summary %></p>
+                <% if link.suggested_fix %>
+                  <p class="issue-summary">Suggested fix: <%= link.suggested_fix %></p>
+                <% end %>
+              </details>
+            </li>
+          <% end %>
```
kevindew: This HTML feels a little strange where we have headers inside the list.

I feel like this would feel more semantic were you to have it like:
```
<h3 class="remove-top-margin"><%= t "broken_links.broken.title" %></h3>
<p><%= t "broken_links.broken.subheading" %></p>
<ul class="issue-list">
 ....
</ul>
```
and then just have the issues in the list, as that is really the part of this that is a list. What do you think?

config/locales/en.yml
```diff
@@ -486,3 +486,9 @@ en:
     transparency: See all our transparency data
     written_statement: See all our written statements to Parliament
     authored_article: See all our authored articles
+  broken_links:
+    broken: 
+      title: Broken links
```
thomasleese: I would move this entry out of the `broken` section.

config/locales/en.yml
```diff
@@ -486,3 +486,9 @@ en:
     transparency: See all our transparency data
     written_statement: See all our written statements to Parliament
     authored_article: See all our authored articles
+  broken_links:
```
thomasleese: Could we get the english translations added in the previous commit, then we can review it together? Regenerating the remaining translations is fine for this commit.

config/locales/en.yml
```diff
@@ -486,3 +486,9 @@ en:
...
+    title: Broken links
+    broken: 
+      subheading: We’ve found links in this document that may be broken
+    caution:
+      subheading: We've found some issues with links that may indicate problems, you should apply caution in linking to them
```
thomasleese: I think we should use fancy quotes here like in the broken subheading: `We’ve`

thomasleese: Ah sure, in the broken subheading the `We’ve` has got a curly quote compared to the `We've` in the caution subheading. We should make it so the caution also has that curly quote. You should be able to just copy and paste it from the broken subheading.

PR #3574, opened 2017-11-30
Add check all editions links worker

SebAshton:
This PR adds two new workers. `CheckOrganisationLinksWorker` which queues up batch requests for public Editions, these Editions can either be per organisation or not. `CheckAllOrganisationsLinksWorker` is a scheduled worker that runs at 2 a.m. to spin up instances of `CheckOrganisationLinksWorker` for each organisation.

This is the first part of a feature to move the broken link checker to use the Link Checker API.

The next steps are;

- Add a new class to generate the CSVs, and email

Scheduling this means that we can proactively generate the current status of external links in each public edition. This information will be shown on `edition#show` giving the user greater visibility on the errors, so they will not have to run an initial on-demand check. Later this information will be used to provide UI enhancements to the [Documents filter page](https://trello.com/c/otfDCs5y/94-add-a-pill-link-to-editions-with-broken-links-on-documents-view)

## Execution times

| Worker | Time (MM:SS:MS) |
|---------|------|
| CheckAllOrganisationsLinksWorker | 0:00:78 |
| CheckOrganisationLinksWorker [for one org] |  16:46:49 |


lib/whitehall/edition_batch_link_checker.rb
```diff
@@ -0,0 +1,30 @@
+# Calls the Link Checker API to verify all links in an edition per organisation
+module Whitehall
+  class EditionBatchLinkChecker
+    include Rails.application.routes.url_helpers
+
+    attr_reader :logger, :organisation
```
thomasleese: We can probably make these private.

lib/whitehall/edition_batch_link_checker.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def initialize(logger = Rails.logger, organisation = nil)
```
thomasleese: I'd be tempted to make `organisation` a keyword argument:

```ruby
def initialize(logger = Rails.logger, organisation: nil)
```

lib/whitehall/edition_batch_link_checker.rb
```diff
@@ -0,0 +1,30 @@
+# Calls the Link Checker API to verify all links in an edition per organisation
+module Whitehall
+  class EditionBatchLinkChecker
```
thomasleese: This class feels like a worker to me, is there a reason it shouldn't be? It would mean we can then create a rake task to run this task manually, and also get Sidekiq to run it every 24 hours.

app/workers/edition_batch_link_checker.rb
```diff
@@ -0,0 +1,31 @@
+# Calls the Link Checker API to verify all links in an edition per organisation
+require "sidekiq-scheduler"
+
+class EditionBatchLinkChecker
+  include Sidekiq::Worker
+  include Rails.application.routes.url_helpers
+
+  def perform(organisation_id = nil)
+    organisation = find_organisation(organisation_id)
```
thomasleese: I wonder why we need to make it possible to pass in the organisation here, the only place it's used right now it will always be `nil`. We could add support for specifying organisations later if we need it.

app/workers/edition_batch_link_checker.rb
```diff
@@ -0,0 +1,31 @@
+# Calls the Link Checker API to verify all links in an edition per organisation
+require "sidekiq-scheduler"
```
thomasleese: Do we need this in the worker class directly? Can we move it into the initialisation for Sidekiq?

app/workers/edition_batch_link_checker.rb
```diff
@@ -0,0 +1,31 @@
...
+
+class EditionBatchLinkChecker
```
thomasleese: Would be good to have `Worker` the end of the class name, perhaps this should be called `CheckAllEditionsLinksWorker`.

app/workers/edition_batch_link_checker.rb
SebAshton: This is the suggested usage from the [repo](https://github.com/moove-it/sidekiq-scheduler#hello-world) follows how we've used in the [Link Checker API](https://github.com/alphagov/link-checker-api/blob/master/app/workers/cleanup_worker.rb#L1)

thomasleese: 😢 That's annoying, it looks strange having to put it in the worker when the worker doesn't care if it's scheduled or not!

app/workers/check_all_editions_links_scheduled_worker.rb
```diff
@@ -0,0 +1,18 @@
+# Enqueues CheckAllEditionsLinksWorker for all organisations
+require "sidekiq-scheduler"
+
+class CheckOrganisationLinksScheduledWorker
```
thomasleese: I would call this `CheckAllOrganisationsLinksWorker`.

app/workers/check_all_editions_links_worker.rb
```diff
@@ -0,0 +1,29 @@
+# Calls the Link Checker API to verify all links in public editions, either per organisation or not
+class CheckAllEditionsLinksWorker
```
thomasleese: I would make this class `CheckOrganisationLinksWorker` and make it a requirement to pass in an organisation ID, since users can use the `CheckOrganisationLinksScheduledWorker` worker if they want to check all the links.

config/sidekiq.yml
```diff
@@ -12,3 +12,7 @@
   - bulk_republishing
   - sync_checks
   - asset_migration
+:schedule:
+  check_editions_links:
```
thomasleese: Can we name this as close as possible to the worker name but without the `Worker` suffix?

test/test_helper.rb
```diff
@@ -112,6 +112,11 @@ def self.disable_database_queries
     end
   end
 
+  def self.enable_url_helpers
```
thomasleese: Since we already have one of these in the active mailer class, do you think we could refactor it and extract it into a mixin class that gets `include`d in both?

test/support/url_helpers.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  module ClassMethods
+    def enable_url_helpers
+      # See http://jakegoulding.com/blog/2011/02/26/using-named-routes-in-actionmailer-tests-with-rails-3/
+      include Rails.application.routes.url_helpers
+      Rails.application.routes.default_url_options[:host] = "example.com"
+
+      def default_url_options
+        Rails.application.routes.default_url_options
+      end
```
tuzz: Is this method nested to avoid mixing it into classes that use this module?

tuzz:
As a general approach, I was wondering if we already have a mechanism for running these kinds of scheduled tasks? (@gpeng)

Otherwise, this looks good to me. 🚢  👍

SebAshton:
@thomasleese have added benchmarking from integration

test/support/url_helpers.rb
SebAshton: Yep, wanted to keep it as isolated as it was before I extracted it into a module

tuzz: So is the idea that calling `#default_url_options` in a view context usually calls the Rails built-in method, but if you call `#enable_url_helpers` first it calls the one from `UrlHelpers` ?

Seems a bit odd to dynamically define a method by calling another one. I've not seen this pattern before.

PR #3576, opened 2017-11-30
Add a CSV generator service class for LinkCheckerReports

camillavk:
This commit adds a new CSV generating service class that will generate a BrokenLinkReporter CSV. This service class uses the new `LinkCheckerApiReport` to populate the CSV.

This is part of a feature to move the current broken link checker to use the Link Checker API, linked to https://github.com/alphagov/whitehall/pull/3574.

### DIFF
--- a/app/services/link_reporter_csv_service.rb
+++ b/app/services/link_reporter_csv_service.rb
@@ -1,0 +1,69 @@
+class LinkReporterCsvService
+  def initialize(reports_dir:, organisation:)
+    @organisation = organisation
+    @reports_dir = reports_dir
+  end
+
+  def generate
+    CSV.open(file_path, "w", encoding: "UTF-8") do |csv|
+      csv << headings
+      public_editions.each do |edition|
+        csv << row_for_edition(edition) if broken_links(edition).any?
+      end
+    end
+  end
+
+private
+
+  attr_reader :organisation, :reports_dir
+
+  def public_editions
+    Edition.publicly_visible.with_translations.in_organisation(organisation)
+  end
+
+  def file_path
+    reports_dir.join("#{@organisation.slug}_links_report.csv")
+  end
+
+  def headings
+    ["page", "admin link", "public timestamp", "format", "broken link count", "broken links"]
+  end
+
+  def row_for_edition(edition)
+    [
+      public_url(edition),
+      admin_url(edition),
+      timestamp(edition),
+      edition.type,
+      broken_links(edition).size,
+      broken_links(edition).join("\r\n")
+    ]
+  end
+
+  def public_url(edition)
+    Whitehall.url_maker.public_document_url(edition, host: public_host, protocol: "https")
+  end
+
+  def admin_url(edition)
+    Whitehall.url_maker.admin_edition_url(edition, host: admin_host, protocol: "https")
+  end
+
+  def timestamp(edition)
+    edition.public_timestamp.to_s
+  end
+
+  def broken_links(edition)
+    return [] unless edition.link_check_reports.present?
+    edition.link_check_reports.last.broken_links.map(&:uri)
+  end
+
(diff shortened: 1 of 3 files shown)


app/services/link_reporter_csv_service.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def timestamp(edition)
+    edition.public_timestamp.to_s
+  end
+
+  def broken_links(edition)
+    edition.link_check_reports.last.broken_links.map(&:uri)
+  end
+
+  # These hosts are hardcoded because we run this on integration but want the
```
thomasleese: I think it's better that we get integration URLs on integration, staging URLs on staging, and then production URLs on production.

app/services/link_reporter_csv_service.rb
```diff
@@ -0,0 +1,70 @@
+class LinkReporterCsvService
+  delegate :<<, :close, to: :csv
+
+  def initialize(reports_dir:, organisation:)
+    @organisation = organisation
+    @reports_dir = reports_dir
+  end
+
+  def generate
+    csv = CSV.open(file_path, "w", encoding: "UTF-8")
```
thomasleese: You can use

```ruby
CSV.open(...) do
  csv << ...
  ...
end
```

to avoid needing the `ensure`.

app/services/link_reporter_csv_service.rb
```diff
@@ -0,0 +1,70 @@
+class LinkReporterCsvService
+  delegate :<<, :close, to: :csv
```
thomasleese: Do we actually need this?

app/services/link_reporter_csv_service.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def file_path
+    reports_dir.join("#{@organisation.slug}_links_report.csv")
+  end
+
+  def headings
+    ["page", "admin link", "public timestamp", "format", "broken link count", "broken links"]
+  end
+
+  def add_edition_row(csv, edition)
```
thomasleese: I think it would be better if this method just returned the csv row for an edition, say, `row_for_edition`, and then the `csv <<` part happens in `generate`.

app/services/link_reporter_csv_service.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def row_for_edition(edition)
+    [
+      public_url(edition),
+      admin_url(edition),
+      timestamp(edition),
+      edition.type,
+      broken_links(edition).size,
+      broken_links(edition).join("\r\n")
```
thomasleese: Does having new lines here work with the CSV output?

camillavk: Yup, and to make sure the new CSV remained consistent with the old ones I had copied this from here: https://github.com/alphagov/whitehall/blob/master/lib/whitehall/broken_link_reporter.rb#L44

app/services/link_reporter_csv_service.rb
```diff
@@ -0,0 +1,66 @@
...
+  def admin_url(edition)
+    Whitehall.url_maker.admin_edition_url(edition, host: admin_host, protocol: "https")
+  end
+
+  def timestamp(edition)
+    edition.public_timestamp.to_s
+  end
+
+  def broken_links(edition)
+    edition.link_check_reports.last.broken_links.map(&:uri)
```
thomasleese: We might have to put in a check here for if the edition has no link check reports? Not sure whether the best thing to do is to include it in the CSV file with a note or just not include it...

camillavk: 👍 good spot, I've stuck with the existing functionality of not including the edition in the CSV if there is no report. This is only really to handle very edge cases because as the worker task will be running every night and so it will be very rare that an edition doesn't have a link check report.

PR #3577, opened 2017-12-01
Fix document filter presenter previous/next URL

rubenarakelyan:
This commit adds a new previous/next URL to the document filter presenter that always points to the “web” (human-readable) version of the page rather than being context-driven. Currently, this is a JSON URL when called from the frontend JavaScript which means the previous/next buttons on the relevant finders point to JSON URLs.

Trello: https://trello.com/c/qmKw5mOK/420-fix-whitehall-finders-paginating-to-json-when-changing-filters


tijmenb:
Thanks @alextea for notifying us about this bug!

PR #3580, opened 2017-12-04
Use link checker csv service for rake task

camillavk:
With the new integration with the Link Checker API, and the addition of the nightly check for broken links on public editions in https://github.com/alphagov/whitehall/pull/3574 (the results from these checks are saved in the LinkCheckReports table), the BrokenLinkReporter is no longer required.

The rake task now uses the LinkReporterCsvService class added in https://github.com/alphagov/whitehall/pull/3576 which uses the edition's LinkCheckReports to populate the CSV file with the broken links for each edition.


thomasleese:
We also probably shouldn't merge or deploy this until #3574 is running.

PR #3582, opened 2017-12-06
Add pill to broken and caution links on documents view

tsteer:
Display most urgent status of links (broken or caution) on the document overview table. This will make it easier for users to prioritise which documents need most urgent review

screen shot 2017-12-07 at 12 03 34

screen shot 2017-12-07 at 12 03 20

PR #3585, opened 2017-12-08
Add ability to filter editions by broken links

camillavk:
This commit adds the ability to filter editions by broken links on the index page so users are able
to quickly see editions with broken links. This expands on the work done on https://github.com/alphagov/whitehall/pull/3582
along with the other LinkCheckerAPI integration work.

screen shot 2017-12-08 at 11 20 18

### DIFF
--- a/app/models/admin/edition_filter.rb
+++ b/app/models/admin/edition_filter.rb
@@ -108,6 +108,7 @@ module Admin
       editions = editions.in_world_location(selected_world_locations) if selected_world_locations.any?
       editions = editions.from_date(from_date) if from_date
       editions = editions.to_date(to_date) if to_date
+      editions = editions.only_broken_links if only_broken_links
 
       @unpaginated_editions = editions
     end
@@ -219,6 +220,10 @@ module Admin
         sentence = selected_world_locations.map { |l| WorldLocation.friendly.find(l).name }.to_sentence
         " about #{sentence}"
       end
+    end
+
+    def only_broken_links
+      options[:only_broken_links].present?
     end
   end
 end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -195,6 +195,14 @@ class Edition < ApplicationRecord
     where("editions.updated_at <= ?", date)
   end
 
+  # used by Admin::EditionFilter
+  def self.only_broken_links
+    joins("INNER JOIN link_checker_api_reports ON link_checker_api_reports.link_reportable_id = editions.id")
+    .joins("INNER JOIN link_checker_api_report_links ON link_checker_api_report_links.link_checker_api_report_id = link_checker_api_reports.id")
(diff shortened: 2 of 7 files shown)


app/models/edition.rb
```diff
@@ -195,6 +195,12 @@ def self.to_date(date)
     where("editions.updated_at <= ?", date)
   end
 
+  # used by Admin::EditionFilter
+  def self.only_broken_links
+    joins("INNER JOIN link_checker_api_reports ON link_checker_api_reports.link_reportable_id = editions.id")
```
bevanloon: do we need to include some sort of `where` clause against `link_reportable_type`? Sorry I've not followed the development closely, so I'm not sure what the intention of the `link_reportable_type` is, but if it ever contains anything other than `Edition`, our sql here would potentially return non edition based items that match on the id.

app/models/admin/edition_filter.rb
```diff
@@ -220,5 +221,9 @@ def location_matches
         " about #{sentence}"
       end
     end
+
+    def only_broken_links
+      true if options[:only_broken_links].present?
```
gpeng: You could do away with the `true if` here. Also, possibly the `.present?`. What happens if it is `.present?` but `false`? (I'm guessing that shouldn't happen). Maybe you could have `!!options[:only_broken_links]` that would always return a boolean but also respect a value of `false`

app/models/edition.rb
```diff
@@ -195,6 +195,12 @@ def self.to_date(date)
...
+    .joins("INNER JOIN link_checker_api_report_links ON link_checker_api_report_links.link_checker_api_report_id = link_checker_api_reports.id")
```
kevindew: Should one of these joins be checking that there is a status of broken?

app/models/admin/edition_filter.rb
gpeng: Actually... `.present?` returns `false` if the value is `false` so you could leave it as just `options[:only_broken_links].present?`

app/models/edition.rb
camillavk: Yup, this is a good point, I've added a `where` clause to deal with this

app/models/edition.rb
camillavk: Correct, I've added a `where` clause to catch this

camillavk:
I've addressed the other comments and added a unit test @gpeng

app/models/edition.rb
```diff
@@ -195,6 +195,14 @@ def self.to_date(date)
...
+    .where("link_checker_api_reports.link_reportable_type = 'Edition' AND link_checker_api_report_links.status != 'Ok'")
```
thomasleese: Should this definitely be `Ok` with a capital 'o'?

camillavk: @thomasleese I've amended this

PR #3587, opened 2017-12-08
Stagger the checking of organisation edition links

SebAshton:
This commit reduces the load incurred by checking all the links in editions for every organisation by staggering the calls to the `CheckOrganisationLinksWorker`.

### DIFF
--- a/app/workers/check_all_organisations_links_worker.rb
+++ b/app/workers/check_all_organisations_links_worker.rb
@@ -5,8 +5,9 @@ class CheckAllOrganisationsLinksWorker
   include Sidekiq::Worker
 
   def perform
-    organisations.each do |organisation|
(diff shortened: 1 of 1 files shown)


thomasleese:
It's worth noting that this should help to prevent the impact of the continuous "broken link checker in progress" box since the link reports will be marked as "in progress" much closer to the time when they are likely to get a response back from the link checker api.

thomasleese:
Actually, I've just thought... if each worker takes 16 minutes to run, will this solve the problem? Since they're likely to run with minute gaps anyway.

thomasleese:
I believe we have about 1000, so this would run over 17 hours.

SebAshton:
Closing in favour of #3631

PR #3589, opened 2017-12-08
Store ImageUploader attachments in Asset Manager as well as NFS

chrislo:
This PR switches the `ImageUploader` to use the `asset_manager_and_quarantined_file_storage` storage engine - ensuring that any new files that are uploaded using this uploader will be stored in Asset Manager. This is a precursor to migrating all of the existing files that use this uploader to Asset Manager and eventually removing them from NFS altogether. 

The change itself (in the first commit) is simple. Subsequent commits add some additional integration testing to make sure that Create, Update and Delete operations on a model that mounts this uploader (`Person`) work as expected. 

Note that the `ImageUploader` is configured to not remove files when they are replaced with new ones.

Fixes: https://github.com/alphagov/asset-manager/issues/325

### DIFF
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -1,4 +1,6 @@
 class ImageUploader < WhitehallUploader
+  storage :asset_manager_and_quarantined_file_storage
+
   include CarrierWave::MiniMagick
 
   configure do |config|
--- a/test/integration/asset_manager_test.rb
+++ b/test/integration/asset_manager_test.rb
@@ -58,6 +58,52 @@ class AssetManagerIntegrationTest
     end
   end
 
+  class CreatingAPersonImage < ActiveSupport::TestCase
+    setup do
+      @filename = 'minister-of-funk.960x640.jpg'
+      @person = FactoryBot.build(
+        :person,
+        image: File.open(fixture_path.join(@filename))
+      )
+
+      Services.asset_manager.stubs(:create_whitehall_asset)
+    end
+
+    test 'sends the person image to Asset Manager' do
+      Services.asset_manager.expects(:create_whitehall_asset).with do |args|
+        args[:file].is_a?(File) &&
+          args[:legacy_url_path] =~ /#{@filename}/
+      end
+
+      @person.save!
+    end
+
+    test 'sends each version of the person image to Asset Manager' do
+      ImageUploader.versions.each_key do |version_prefix|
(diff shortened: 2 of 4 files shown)


chrislo:
In the case where a file is uploaded using ImageUploader with the same filename (but different contents) the `gds-api-adapter` will throw an `GdsApi::HTTPUnprocessableEntity` exception because the `Legacy url path is already taken`. 

I'm going to close this PR for now and re-open when I've fixed this issue.

floehopper:
Minor: There's a typo in the first commit note: "Assert Manager" -> "Asset Manager".

floehopper:
Minor: There's a typo in the commit note for "Integration test for replacing an ImageUploader image": "but I have varied that setting" -> "but I have verified that setting".

floehopper:
I think it would be useful to summarise the impact of the `remove_previously_stored_files_after_update` uploader config setting in the PR description. In particular I'd like to know whether there will be any difference from a URL point of view for updated assets stored on Whitehall NFS vs those stored in Asset Manager.

floehopper:
Review summary:

* I'm slightly unsure about the integration tests which only check the number of calls to the Asset Manager API, but maybe they're OK.
* I still find calls to `Mocha::Expectation#with` with a block argument hard to read. We've talked about addressing separately it in previous PRs, but this PR adds more occurrences.
* I think it would be useful to summarise the impact of the `remove_previously_stored_files_after_update` uploader config setting in the PR description. In particular I'd like to know whether there will be any difference for updated assets from a URL point of view for assets stored on Whitehall NFS vs those stored in Asset Manager.

I can't actually add a review until this is re-opened.

floehopper:
Review summary:

* I'm slightly unsure about the integration tests which only check the number of calls to the Asset Manager API, but maybe they're OK.
* I still find calls to Mocha::Expectation#with with a block argument hard to read. We've talked about addressing separately it in previous PRs, but this PR adds more occurrences.
* I think it would be useful to summarise the impact of the remove_previously_stored_files_after_update uploader config setting in the PR description. In particular I'd like to know whether there will be any difference for updated assets from a URL point of view for assets stored on Whitehall NFS vs those stored in Asset Manager.

I can't actually add a review until this is re-opened.

chrislo:
I've fixed the typos, refactored the `Mocha::Expectation#with` calls and added some more detail to the PR description. I can't reopen this PR as I've added new commits, so have opened #3602 instead.

PR #3602, opened 2017-12-12
Use Asset Manager and NFS to store files in ImageUploader

chrislo:
I've reopened this PR after addressing the comments in #3589.

This PR switches the `ImageUploader` to use the `asset_manager_and_quarantined_file_storage` storage engine - ensuring that any new files that are uploaded using this uploader will be stored in Asset Manager. This is a precursor to migrating all of the existing files that use this uploader to Asset Manager and eventually removing them from NFS altogether. 

The change itself (in the first commit) is simple. Subsequent commits add some additional integration testing to make sure that Create, Update and Delete operations on a model that mounts this uploader (`Person`) work as expected. 

Note that the `ImageUploader` is configured to not remove files when they are replaced with new ones. This change to the `remove_previously_stored_files_after_update` config variable was added in 32932249 to ensure that already published links to assets that had been replaced would continue to work with the switch to the publishing API. This is an issue because URLs to assets are "burnt in" to the published content in the publishing API and the relationship to the included assets is not modelled - so when the asset URLs change the published content does not.

In terms of asset manager, when an asset is replaced by a file with a different name, two assets with different `legacy_url_path` will exist in asset manager and links to the old assets will continue to work. If an asset is replaced by a new file with the *same* name, the old asset will be soft-deleted in asset manager and the new asset will take its `legacy_url_path`. This change was introduced in https://github.com/alphagov/asset-manager/pull/336.

Fixes: https://github.com/alphagov/asset-manager/issues/325

### DIFF
--- a/test/integration/asset_manager_test.rb
+++ b/test/integration/asset_manager_test.rb
@@ -58,6 +58,52 @@ class AssetManagerIntegrationTest
     end
   end
 
+  class CreatingAPersonImage < ActiveSupport::TestCase
+    setup do
+      @filename = 'minister-of-funk.960x640.jpg'
+      @person = FactoryBot.build(
+        :person,
+        image: File.open(fixture_path.join(@filename))
+      )
+
+      Services.asset_manager.stubs(:create_whitehall_asset)
+    end
+
+    test 'sends the person image to Asset Manager' do
+      Services.asset_manager.expects(:create_whitehall_asset).with do |args|
+        args[:file].is_a?(File) &&
+          args[:legacy_url_path] =~ /#{@filename}/
+      end
+
+      @person.save!
+    end
+
+    test 'sends each version of the person image to Asset Manager' do
+      ImageUploader.versions.each_key do |version_prefix|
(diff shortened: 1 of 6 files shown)


test/integration/asset_manager_test.rb
```diff
@@ -104,6 +104,52 @@ class CreatingAPersonImage < ActiveSupport::TestCase
...
+      @person.reload
+
+      @asset_id = 'asset-id'
+      Services.asset_manager.stubs(:whitehall_asset).returns('id' => "http://asset-manager/assets/#{@asset_id}")
+    end
+
+    test 'removes the person image and all its versions from asset manager' do
+      # Creating a person creates one entry in asset manager for the
+      # uploaded asset and one entry for each of the versions defined
+      # in ImageUploader
```
floehopper: Minor: I found the use of the word "entry" a bit confusing in this comment block. I think it might be clearer to say something like "asset record"...

chrislo: I'll fix that when I rebase, thanks!

chrislo:
Thanks @floehopper. Rebasing and force-pushing before merging.

PR #3614, opened 2017-12-14
MigrateAssetsToAssetManager#to_s returns a count

chrislo:
Before this commit MigrateAssetsToAssetManager#to_s returned a newline
separated list of all of the files to be migrated. When we migrate
`system/uploads/image_data/file`, for example, there will be almost
500,000 files[1]. It is probably not a good idea to generate this much
output in the rake task that calls this method, and a simple count
will be sufficient to verify that the correct number of assets have
been migrated.

[1] https://github.com/alphagov/asset-manager/issues/341

### DIFF
--- a/lib/migrate_assets_to_asset_manager.rb
+++ b/lib/migrate_assets_to_asset_manager.rb
@@ -1,4 +1,6 @@
 class MigrateAssetsToAssetManager
+  include ActionView::Helpers::TextHelper
+
   def initialize(target_dir)
     @file_paths = AssetFilePaths.new(target_dir)
   end
@@ -10,7 +12,7 @@ class MigrateAssetsToAssetManager
   end
 
   def to_s
-    @file_paths.join("\n")
+    "Migrating #{pluralize(@file_paths.size, 'file')}"
   end
 
   class Worker < WorkerBase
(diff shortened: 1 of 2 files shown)


lib/migrate_assets_to_asset_manager.rb
```diff
@@ -40,7 +42,7 @@ def asset_exists?(file)
   end
 
   class AssetFilePaths
-    delegate :each, :join, to: :file_paths
+    delegate :each, :join, :size, to: :file_paths
```
floehopper: Minor: I think you could stop delegating `join` given that it's no longer used in `MigrateAssetsToAssetManager#to_s`.

test/unit/migrate_assets_to_asset_manager_test.rb
```diff
@@ -95,6 +95,10 @@ class AssetFilePathsTest < ActiveSupport::TestCase
     assert @subject.respond_to?(:join)
   end
 
+  test 'delegates size to file_paths' do
```
floehopper: Minor: I might say this should be more like "implements size method", because it's not actually testing the delegation.

lib/migrate_assets_to_asset_manager.rb
chrislo: I'll remove that and rebase.

chrislo:
Thanks for the review @floehopper - I hadn't considered passing in an `IO` object that's a good idea. I don't think I'll make that change now but it'll be useful in the future.

PR #3626, opened 2017-12-18
Return 406 error for attachment preview XHR requests

rubenarakelyan:
This commit restricts attachment previews to only return content for requests that ask for an HTML response. All other response formats will generate a 406 error, incuding XHR requests.

Previously, XHR requests triggered partial HTML to be sent to slimmer, resulting in an exception where slimmer could not find the wrapper element.

Trello: https://trello.com/c/CAHJcse6/463-show-an-error-message-for-attachment-preview-xhr-requests

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -4,18 +4,23 @@ class AttachmentsController < PublicUploadsController
   before_action :reject_non_previewable_attachments, only: :preview
 
   def preview
-    if attachment_visible? && attachment_visibility.visible_edition
-      expires_headers
-      @edition = attachment_visibility.visible_edition
-      @attachment = attachment_visibility.visible_attachment
-      @csv_preview = CsvPreview.new(upload_path)
-      render layout: 'html_attachments'
-    else
-      fail
+    respond_to do |format|
+      format.html do
+        if attachment_visible? && attachment_visibility.visible_edition
+          expires_headers
+          @edition = attachment_visibility.visible_edition
+          @attachment = attachment_visibility.visible_attachment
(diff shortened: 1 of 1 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -4,18 +4,23 @@ class AttachmentsController < PublicUploadsController
...
+          render layout: 'html_attachments'
+        else
+          fail
+        end
+      end
     end
-
   rescue CsvPreview::FileEncodingError, CSV::MalformedCSVError
     render layout: 'html_attachments'
+  rescue ActionController::UnknownFormat
```
gpeng: Does rails do this automatically once you add the `respond_to` block?

app/controllers/attachments_controller.rb
```diff
@@ -4,18 +4,23 @@ class AttachmentsController < PublicUploadsController
...
+          @attachment = attachment_visibility.visible_attachment
+          @csv_preview = CsvPreview.new(upload_path)
+          render layout: 'html_attachments'
+        else
+          fail
+        end
+      end
     end
-
   rescue CsvPreview::FileEncodingError, CSV::MalformedCSVError
```
bevanloon: will this still get triggered if the failure happens inside the `do` block above?

bevanloon: nevermind - spun up a quick test to satisfy my curiosity it does 👍

app/controllers/attachments_controller.rb
rubenarakelyan: I tried that and got an unhandled exception and stack trace

PR #3627, opened 2017-12-19
Redirect asset requests to asset host

chrisroos:
This branch ensures that assets (but not attachments) are all served from the asset host (e.g. assets.publishing.service.gov.uk). This addresses https://github.com/alphagov/asset-manager/issues/348 and will allow us to configure things like caching and routing in a single place.

### DIFF
--- a/test/functional/public_uploads_controller_test.rb
+++ b/test/functional/public_uploads_controller_test.rb
@@ -1,0 +1,26 @@
+require "test_helper"
+
+class PublicUploadsControllerTest < ActionController::TestCase
+  setup do
+    Plek.any_instance.stubs(:public_asset_host).returns('http://asset-host.com')
+  end
+
+  test "redirects asset requests that aren't made via the asset host" do
+    request.host = 'not-asset-host.com'
+
+    get :show, params: { path: 'asset', format: 'txt' }
+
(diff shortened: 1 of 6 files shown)


test/functional/public_uploads_controller_test.rb
```diff
@@ -0,0 +1,24 @@
...
+    get :show, params: { path: 'asset', format: 'txt' }
+
+    assert_redirected_to 'http://asset-host.com/government/uploads/asset.txt'
+  end
+
+  test 'does not redirect asset requests that are made via the asset host' do
+    asset_filesystem_path = File.join(Whitehall.clean_uploads_root, 'asset.txt')
+    FileUtils.touch(asset_filesystem_path)
+
+    Plek.any_instance.stubs(:public_asset_host).returns('http://asset-host.com')
```
floehopper: Minor: This line appears in both tests and could be extracted into a `setup` block.

chrisroos: Good spot. I've moved this to a `setup` block.

chrisroos:
Thanks, @floehopper. I've addressed your comment and force pushed in preparation for merging.

PR #3631, opened 2017-12-20
Limit number of checks sent over to the Link Checker API

SebAshton:
This PR limits the number of editions sent to be checked to 25 per organisation. It will take any new publication first, and then the editions with the oldest checks.

The idea is to monitor the number of jobs created and increase over time.

If every `Organisation` has more than 25 public editions then this limit would ensure only `25750` checks were created on the nightly schedule.

### DIFF
--- a/app/workers/check_organisation_links_worker.rb
+++ b/app/workers/check_organisation_links_worker.rb
@@ -1,13 +1,14 @@
 # Calls the Link Checker API to verify all links in public editions, either per organisation or not
 class CheckOrganisationLinksWorker
   include Sidekiq::Worker
+  ORGANISATION_EDITION_LIMIT = 1000
 
   sidekiq_options queue: "link_checks"
 
   def perform(organisation_id)
     organisation = find_organisation(organisation_id)
 
-    public_editions(organisation).find_each do |edition|
+    public_editions(organisation).each do |edition|
       next unless LinkCheckerApiService.has_links?(edition)
 
       LinkCheckerApiService.check_links(edition, callback)
@@ -21,7 +22,7 @@ private
   end
 
   def public_editions(organisation)
-    Edition.publicly_visible.with_translations.in_organisation(organisation)
+    Edition.includes(:link_check_reports).publicly_visible.with_translations.in_organisation(organisation).order('link_checker_api_reports.updated_at').limit(ORGANISATION_EDITION_LIMIT)
   end
 
   def callback
--- a/test/unit/workers/check_organisation_links_worker_test.rb
+++ b/test/unit/workers/check_organisation_links_worker_test.rb
@@ -23,14 +23,57 @@ class CheckOrganisationLinksWorkerTest < ActiveSupport::TestCase
   end
 
   test "when given an organisation it only calls LinkCheckerApiService with editions for that organisation" do
-    stub_published_publication
+    new_edition = stub_published_publication
 
     CheckOrganisationLinksWorker.new.perform(@hmrc.id)
 
     assert_equal 1, LinkCheckerApiReport.count
+    assert_requested new_edition
+  end
+
+  test "when an organisation contains more new editions than the limit" do
+    new_edition = stub_published_publication
+    existing_edition = create_and_stub_an_edition_with_checks(@hmrc)
+    another_existing_edition = create_and_stub_an_edition(@hmrc)
+
+    assert_report_count_increased
+
+    assert_requested new_edition
+    assert_requested another_existing_edition
+    assert_not_requested existing_edition
+  end
+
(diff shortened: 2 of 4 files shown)


app/workers/check_organisation_links_worker.rb
```diff
@@ -1,13 +1,14 @@
 # Calls the Link Checker API to verify all links in public editions, either per organisation or not
 class CheckOrganisationLinksWorker
   include Sidekiq::Worker
+  ORGANISATION_EDITION_LIMIT = 25
```
thomasleese: I would definitely aim for something more like `1000` here.

app/workers/check_organisation_links_worker.rb
```diff
@@ -1,13 +1,14 @@
...
 
   sidekiq_options queue: "link_checks"
 
   def perform(organisation_id)
     organisation = find_organisation(organisation_id)
 
-    public_editions(organisation).find_each do |edition|
+    public_editions(organisation).each do |edition|
```
thomasleese: Can we not use `find_each` here?

test/unit/workers/check_organisation_links_worker_test.rb
```diff
@@ -23,15 +23,59 @@ class CheckOrganisationLinksWorkerTest < ActiveSupport::TestCase
   end
 
   test "when given an organisation it only calls LinkCheckerApiService with editions for that organisation" do
-    stub_published_publication
+    checked_edition = stub_published_publication
```
thomasleese: Should this really be a `non_checked_edition`? As I understand it, it's a new edition that hasn't been checked yet?

app/workers/check_organisation_links_worker.rb
SebAshton: Sure, was just being conservative - perhaps too much 😄

app/workers/check_organisation_links_worker.rb
SebAshton: Under the hood `find_each` uses `find_in_batches` which ignores the `limit` 

> You can’t set the limit either, that’s used to control the the batch sizes.
https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_in_batches

test/unit/workers/check_organisation_links_worker_test.rb
SebAshton: the variable probably needs a better name :)

test/unit/workers/check_organisation_links_worker_test.rb
```diff
@@ -23,15 +23,59 @@ class CheckOrganisationLinksWorkerTest < ActiveSupport::TestCase
...
+  test "when an organisation contains new and old editions" do
+    new_edition = stub_published_publication
+    checked_edition = create_and_stub_an_edition_with_checks(@hmrc)
+
+    assert_report_count_increased
+
+    assert_requested new_edition
+    assert_requested checked_edition
+  end
+
```
thomasleese: Extra empty line!

test/unit/workers/check_organisation_links_worker_test.rb
```diff
@@ -23,15 +23,59 @@ class CheckOrganisationLinksWorkerTest < ActiveSupport::TestCase
...
     CheckOrganisationLinksWorker.new.perform(@hmrc.id)
 
     assert_equal 1, LinkCheckerApiReport.count
+    assert_requested new_edition
+  end
+
+  test "when an organisation contains more new editions than the limit" do
+    new_edition = stub_published_publication
+    checked_edition = create_and_stub_an_edition_with_checks(@hmrc)
+    another_checked_edition = create_and_stub_an_edition(@hmrc)
```
thomasleese: Can this be called `existing_edition`?

test/unit/workers/check_organisation_links_worker_test.rb
```diff
@@ -23,15 +23,59 @@ class CheckOrganisationLinksWorkerTest < ActiveSupport::TestCase
...
+    assert_report_count_increased
+
+    assert_requested new_edition
+    assert_requested another_checked_edition
+    assert_not_requested checked_edition
+  end
+
+  test "when an organisation contains new and old editions" do
+    new_edition = stub_published_publication
+    checked_edition = create_and_stub_an_edition_with_checks(@hmrc)
```
thomasleese: Could this be called `existing_edition` perhaps?

SebAshton:
Have addressed your comments @thomasleese 👍

thomasleese:
We'll also need to re-enable the worker.

SebAshton:
👍  Will be a follow-up PR

PR #3634, opened 2017-12-20
Use the asset host in consultation response form URLs

chrisroos:
We want Whitehall assets to be served by the asset host (instead of
either by the asset host or gov.uk) so that we can make caching and
routing changes in a single place.

PR #3627 adds a redirect to ensure that asset requests made to gov.uk
are redirect to the asset host. This change will ensure that we're using
the canonical asset host when writing consultation content to the
Content Store.

These URLs are rendered in the "Ways to respond" section on a
consultation page.

### DIFF
--- a/app/presenters/publishing_api/consultation_presenter.rb
+++ b/app/presenters/publishing_api/consultation_presenter.rb
@@ -310,7 +310,8 @@ module PublishingApi
 
         path = File.join(dirname, basename)
 
-        url_helpers.public_upload_url(path, format: extension.delete('.'))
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks, @floehopper. I've rebased on master in preparation for merging.

> Out of interest, do we need to make this change to be able to migrate more Whitehall asset types to Asset Manager?

I don't think we need to make this sort of change before we can migrate more Whitehall assets. Particularly now that PR #3627 has been merged but also because it's not clear that links to other assets are being written into the Content Store in the same way.

PR #3636, opened 2017-12-21
Avoid redirecting requests for HMRC assets

chrisroos:
We host a number of assets for [HMRC's Basic PAYE Tools software][1] on
gov.uk. Some of these files are requested from within the Basic PAYE
Tools app - to check for updates, for example. It's not clear how the
app will behave if the request responds with a 302 redirect instead of a
200 with the content of the file. We plan to ask HMRC to update the URL
that they're checking from within the app but we should avoid
redirecting them in the short term to avoid any problems.

I initially used `=~` to check the regexp in the
`PublicUploadsController` but have changed it to `match?` as instructed
by RuboCop.

I explored reducing the duplication of 'government/uploads' in the app
but it became clear that it was going to take quite some time to do it.
It's not obvious that it's worth the investment at the moment given that
some of this code will go away as we migrate more assets to Asset
Manager.

[1]: https://www.gov.uk/basic-paye-tools

### DIFF
--- a/app/controllers/public_uploads_controller.rb
+++ b/app/controllers/public_uploads_controller.rb
@@ -79,9 +79,15 @@ class PublicUploadsController < ApplicationController
   end
 
   def redirect_to_asset_host
+    return if request_for_an_hmrc_asset?
+
     asset_host = URI.parse(Plek.new.public_asset_host).host
     unless request.host == asset_host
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks, @floehopper. I'll get this merged.

PR #3637, opened 2017-12-21
Ensure asset directory exists before creating an asset in test

chrisroos:
This test was failing intermittently on Jenkins. It would fail if the
`Whitehall.clean_uploads_root` didn't exist. I replicated this locally
by deleting the `tmp/test` directory and running the test. I saw the
following error:

```
PublicUploadsControllerTest#test_does_not_redirect_asset_requests_that_are_made_via_the_asset_host:
Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/chrisroos/Code/alphagov/whitehall/tmp/test/env_1/clean-uploads/asset.txt
```

I believe the test would have passed if it was run after a call to
`VirusScanHelpers#simulate_virus_scan` as that method will create the
upload directory.


chrisroos:
Thanks @thomasleese :-) I'm going to get this merged.

PR #3646, opened 2018-01-02
Allow CSV to be created for docs without orgs

camillavk:
When generating a broken link report CSV for documents, create it with a filepath indicating
if no organisation is attached to the document


thomasleese:
Do you know if https://github.com/alphagov/whitehall/pull/3646/files#diff-ed997bee5317a30158a01a493bda65c4R20 still works if `organisation` is `nil`?

camillavk:
@thomasleese I've updated the test and the code to take into account if `organisation` is `nil` there as well

PR #3658, opened 2018-01-04
Ensure hidden files are migrated to asset manager

chrislo:
C.f alphagov/asset-manager#368

We noticed that some assets in Whitehall have file names which include a leading dot[1] These files are treated as "hidden" by Ruby's `Dir.glob` by default and were therefore not being included in the `asset_manager:migrate_assets` rake task. 

In this PR I first refactor the tests for the migration code to allow me to more easily fix an issue with hidden NFS lock files that occurs when running the tests on the development VM on an OS X host. I then fix that issue. In the final commit I fix the issue described above. 

We have already run the rake task against a number of directories. In a subsequent PR I'll introduce another task that [ensures that any hidden files we've missed are also migrated](https://github.com/alphagov/asset-manager/issues/382).
  
[1] e.g. `/data/uploads/whitehall/clean/system/uploads/consultation_response_form_data/file/257/.doc_Response_form.docx`.

### DIFF
--- a/test/unit/migrate_assets_to_asset_manager_test.rb
+++ b/test/unit/migrate_assets_to_asset_manager_test.rb
@@ -143,8 +143,9 @@ end
 class AssetFileTest < ActiveSupport::TestCase
   setup do
     @path = Rails.root.join('test/fixtures/logo.png')
-    file = MigrateAssetsToAssetManager::AssetFile.open(@path)
-    @parts = file.legacy_etag.split('-')
+    MigrateAssetsToAssetManager::AssetFile.open(@path) do |file|
+      @parts = file.legacy_etag.split('-')
+    end
   end
 
   test 'returns string made up of 2 parts separated by a hyphen' do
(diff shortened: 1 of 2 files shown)


test/unit/migrate_assets_to_asset_manager_test.rb
```diff
@@ -117,6 +117,13 @@ class AssetFilePathsTest < ActiveSupport::TestCase
     assert_equal 2, subject.file_paths.size
   end
 
+  test '#files includes hidden files' do
+    hidden_path = File.join(organisation_logo_dir, '.hidden.jpg')
+    FileUtils.cp(dummy_asset_path, hidden_path)
+
+    assert_equal 2, @subject.file_paths.size
```
floehopper: Minor: This assertion seems quite vague and not easy to understand at first glance. Would it not be better to check for the inclusion of the hidden file's path? I see that some of the existing tests take a similar approach, so I guess it would be worth changing them as well...

chrislo: Good point thanks! I've addressed this in b5d5770.

chrislo:
@floehopper thanks for the review. I've addressed you comments about the commit message in eba3dde. 

Rebasing and force-pushing before merging.

PR #3663, opened 2018-01-05
Remove the conditional for precompiling assets in Dockerfile

RyanMacG:
As part of the work to use the built images pushed up to Docker Hub for E2E testing, where we run the tests in the production RAILS_ENV, we will need the image pushed up to contain the compiled assets.

We can't see a reason to keep the conditional for now as any performance hit will be experienced only as part of CI.


kevindew:
This one seems a bit of a pain with the use of asset host as part of the assets :( will mean it'll not be safe to use. 

I wonder how much pain is involved in dropping the asset_host dependency

kevindew:
I had a bit of a look further and the thing that references the items is just this flash thing that's looking pretty dated: https://www.gov.uk/government/history/10-downing-street#take-the-tour which looks like it might be removed in the [not too distant future](https://trello.com/c/mCfPlz3z/1089-remove-flash-player).

I'd suggest we put the asset host in to get past this hurdle but we leave a comment and a todo in the dockerfile to indicate that it will stop some assets working in other environments and that once that flash is removed those constants should be removed.

Dockerfile
```diff
@@ -21,8 +21,10 @@ RUN bundle install
 
 ADD . $APP_HOME
 
-ARG COMPILE_ASSETS=false
-RUN if [ "$COMPILE_ASSETS" = "true" ] ; then bundle exec rails assets:precompile ; fi
+# TODO: GOVUK_ASSET_HOST is required to serve the Flash SWF on https://www.gov.uk/government/history/10-downing-street#take-the-tour
+# Once the flash player has been removed (https://trello.com/c/mCfPlz3z/1089-remove-flash-player) this var can also be removed
+# see: https://github.com/alphagov/whitehall/pull/3663#issuecomment-355626368
+RUN GOVUK_ASSET_HOST=http://assets-origin.dev.gov.uk GOVUK_ASSET_ROOT=http://assets-origin.dev.gov.uk GOVUK_APP_DOMAIN=www.gov.uk RAILS_ENV=production bundle exec rails assets:precompile
```
kevindew: Sorry nitpicking. Can you use the production hosts for consistency `https://assets.publishing.service.gov.uk` so it matches GOVUK_APP_DOMAIN.

PR #3672, opened 2018-01-10
Add logging to LinkCheckerApiService#check_links

SebAshton:


app/services/link_checker_api_service.rb
```diff
@@ -11,11 +11,23 @@ def self.check_links(reportable, webhook_uri)
...
 
+    Rails.logger.info "Calling create_batch for Edition #{reportable.id}"
+
     batch_report = Whitehall.link_checker_api_client.create_batch(
       uris,
       webhook_uri: webhook_uri,
       webhook_secret_token: webhook_secret_token
     )
+
+    Rails.logger.info "Batch for Edition #{reportable.id} with id #{batch_report.id}"
```
thomasleese: Might be good to extract `batch_report.id` out to a variable because I think it should be `batch_report_id = batch_report.fetch('id')`.

app/services/link_checker_api_service.rb
```diff
@@ -11,11 +11,23 @@ def self.check_links(reportable, webhook_uri)
...
+
+    existing_report = LinkCheckerApiReport.find_by(batch_id: batch_report.id)
+
+    if existing_report
+      Rails.logger.info "LinkCheckerApiReport exists for #{batch_report.id}"
+      Rails.logger.info "It belongs to Edition #{existing_report.link_reportable.id} not #{reportable.id}"
```
thomasleese: Can we just use `existing_report.link_reportable` and `reportable` in the string then we might get info such as the type of document.

app/services/link_checker_api_service.rb
```diff
@@ -11,11 +11,25 @@ def self.check_links(reportable, webhook_uri)
     uris = extract_links(reportable)
     raise "Reportable has no links to check" unless uris.count > 0
 
+    Rails.logger.info "Calling create_batch for Edition #{reportable.id}"
```
thomasleese: Perhaps we should include information about the URIs here too.

```ruby
Rails.logger.info "Checking links for #{reportable} which has #{uris.count} URIs, including: #{uris[0]}"
```

thomasleese:
Shouldn't be needed once #3675 is merged in.

PR #3677, opened 2018-01-11
Upgrade gds-api-adapters to 50.9.1

chrisroos:
So that I can get the change in
https://github.com/alphagov/gds-api-adapters/pull/794 which should allow
me to fix https://github.com/alphagov/asset-manager/issues/384.


thomasleese:
Done in #3678.

PR #3682, opened 2018-01-15
Avoid saving consultation response form data to NFS

chrisroos:
We've recently switched to serving ConsultationResponseFormData assets from Asset Manager (see https://github.com/alphagov/asset-manager/issues/297 for more info) so we can stop saving them to NFS.

This is similar to the change for Organisation logos in PR #3532.


chrisroos:
Thanks for reviewing, @floehopper. I've rebased on master and force pushed in preparation for merging.

PR #3685, opened 2018-01-16
Remove unused LinksReports code

tsteer:


db/migrate/20180116115825_drop_links_reports.rb
```diff
@@ -0,0 +1,5 @@
+class DropLinksReports < ActiveRecord::Migration[5.0]
```
thomasleese: This migration should be deployed afterwards in a separate PR otherwise we might get a situation where old code is still trying to access the table while the deployment is happening.

PR #3689, opened 2018-01-16
Remove LinksReport model

tsteer:
Remove LinksReport model that is not used in the code base; this is to keep code concise.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171206121121) do
+ActiveRecord::Schema.define(version: 20180116144233) do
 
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -78,7 +78,6 @@
     t.index ["attachable_id", "attachable_type"], name: "index_attachments_on_attachable_id_and_attachable_type", using: :btree
     t.index ["attachable_type", "attachable_id", "ordering"], name: "no_duplicate_attachment_orderings", unique: true, using: :btree
     t.index ["attachment_data_id"], name: "index_attachments_on_attachment_data_id", using: :btree
-    t.index ["content_id"], name: "index_attachments_on_content_id", using: :btree
```
thomasleese: I don't think we want this change.

PR #3691, opened 2018-01-16
Delete unused code

tsteer:
Remove LinksChecker code that has no function/is not used in the code base; this is to keep code concise.


config/initializers/authed_hosts_for_link_checker.rb
```diff
@@ -1,4 +0,0 @@
-# This ENV var is only populated for the integration deployment environment.
-if ENV.has_key?("BASIC_AUTH_CREDENTIALS")
-  LinksChecker.authed_domains = { "www.#{ENV['GOVUK_APP_DOMAIN']}" => ENV["BASIC_AUTH_CREDENTIALS"] }
```
thomasleese: Would be good to get rid of `BASIC_AUTH_CREDENTIALS` from govuk-puppet since it's not needed anymore.

PR #3693, opened 2018-01-16
Pass relative paths to MigrateAssetsToAssetManager::Worker

chrislo:
See: https://github.com/alphagov/asset-manager/issues/414

Before this commit we were creating `MigrateAssetsToAssetManager::Worker` jobs with the full path to each asset to be migrated. Due to the way capistrono works and `Whitehall.clean_uploads_root` is defined these paths took the form

    /data/vhost/whitehall-admin/releases/20180116155427/clean-uploads/system/uploads/person/1/image.jpg

If a deployment happened while these jobs were on the queue, they would start to fail as the release directory would change.

With this commit we pass the relative path to the asset (e.g. `system/uploads/person/1/image.jpg`) to the worker and join it to the current `Whitehall.clean_uploads_root` when the job is run.

I've changed the test to refute that `AssetFilePaths#file_paths` does not return directories to use the absolute path. The way it was written before it could have given a false positive.

### DIFF
--- a/lib/migrate_assets_to_asset_manager.rb
+++ b/lib/migrate_assets_to_asset_manager.rb
@@ -2,17 +2,17 @@ class MigrateAssetsToAssetManager
   include ActionView::Helpers::TextHelper
 
   def initialize(target_dir)
-    @file_paths = AssetFilePaths.new(target_dir)
+    @relative_file_paths = AssetFilePaths.new(target_dir)
   end
 
   def perform
-    @file_paths.each do |file_path|
-      Worker.perform_async(file_path)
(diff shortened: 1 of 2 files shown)


chrislo:
I deployed this branch (using the `deploy` task) to integration and ran the `asset_manager:migrate_assets[system/uploads/classification_featuring_image_data/file]` task to enqueue around 15,000 assets for migration. I monitored the size of the `asset_migration` queue and the `RetrySet` size while the migration ran. During the migration I redeployed this branch. The queue continued to be processed and no exceptions were raised. 

I'm reasonably confident that this worker now survives redeployment. I'll rebase and force push before merging.

PR #3698, opened 2018-01-18
Don't hide flash notices after 3 seconds

fofr:
* It may take longer than 3 seconds to read the notice
* A user may be a slow reader
* The removal of the box causes the page to jump, which may interrupt another user task (eg button moving)
* There are no accessibility hooks for these messages, hiding the message makes it inaccessible
* Clicking the box to hide it isn't intuitive – eg a user might be copying the text, this also prevents us from putting links in the flash notices

(`showNotice` and `showAlert` functions also in the flashnotice.js file are not used)

screen shot 2018-01-18 at 10 27 38

cc @BlancaTortajada @benhazell


BlancaTortajada:
Agree, we've started documenting notifications at https://paper.dropbox.com/doc/In-service-notifications-and-alerts-XMfiGjjgDxBvLHJUPJ6aN and the ones used by GDS services do not disappear after a while.

huminmiah:
Happy to make these change now. However, our focus is to have the current workflow in the prototype and then ensure these bug fixes are addressed in the prototype. This is because we might want to re think how we use banners for notices in the workflow based on the patterns GOV.UK elements uses. @BlancaTortajada and @benhazell has started to document these.

PR #3702, opened 2018-01-19
Remove 'placeholder' image alt texts

selfthinker:
If no image has been selected for news articles, a placeholder image gets added instead. And when that placeholder image is used, its alt text reads 'placeholder'.
That is bad for screen reader users as that content doesn't make sense. It's much better to leave it empty so screen readers can ignore it.

The build is currently failing but I don't have Whitehall set up, so cannot test or fix it. Can someone take this over and finish it (unless it's more complicated than I think)?


tlwr:
I'm closing this PR as it has been open for a considerable amount of time and we're trying to cut down the number of orphaned PRs in alphagov.

PR #3704, opened 2018-01-22
Additional rake tasks to remove assets

chrislo:
We are migrating ConsultationResponseFormUploader (alphagov/asset-manager#297) and ImageUploader (alphagov/asset-manager#405) assets to Asset Manager. This means they can be removed from NFS. This PR adds rake tasks to remove these assets.


chrislo:
Thanks @chrisroos. Rebasing and force-pushing before merging.

PR #3708, opened 2018-01-23
Track all form error messages

fofr:
Example event:

category: `form-error`
action: `publication-error`
label: `Title can't be blank`

Example markup that generates event:

```html
<li data-module="auto-track-event" data-track-category="form-error" data-track-action="statistical-data-set-error" data-track-label="Title can't be blank">Title can't be blank</li>
```

Example errors that will be tracked:

screen shot 2018-01-23 at 11 26 18

Notes:
* `object.class.name.demodulize.underscore.humanize` is used to present the error summary, so the dash variant will also work for keeping note of the format the error was on
* Update `strip_tags` to `ActionController::Base.helpers.strip_tags` so it can be used anywhere

Example view of real time events being tracked:

screen shot 2018-01-23 at 11 46 15

### DIFF
--- a/lib/whitehall/form_builder.rb
+++ b/lib/whitehall/form_builder.rb
@@ -1,5 +1,7 @@
 module Whitehall
   class FormBuilder < ActionView::Helpers::FormBuilder
+    include Admin::AnalyticsHelper
+
     def label(method, text = nil, options = {})
       if calculate_required(method, options)
         unless !options[:required].nil? && options[:required] == false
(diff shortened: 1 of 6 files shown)


lib/whitehall/form_builder.rb
```diff
@@ -36,7 +38,7 @@ def errors
     def error_list
       @template.content_tag(:ul, "class" => "errors disc") do
         object.errors.full_messages.each do |msg|
-          @template.concat @template.content_tag(:li, msg)
+          @template.concat @template.content_tag(:li, msg, data: track_analytics_data('form-error', "#{object.class.name.demodulize.underscore.dasherize}-error", msg))
```
andrewgarner: You can calculate this outside the enumerator.

fofr: Updated

PR #3710, opened 2018-01-23
Fix Whitehall link checker for admin URLs

suzannehamilton:
Publishers can link to whitehall URLs in the body of documents, and these are converted to publicly-available GOV.UK URLs when the document is published.

This change adds the same conversion to the link checking feature so that it checks the GOV.UK URL rather than failing because it does not have permission to reach the admin URL.

This is probably easiest to review commit-by-commit - the first four commits are refactorings and new tests.

https://trello.com/c/bmMwHljS/33-2-fix-whitehall-link-checker-for-admin-links

### DIFF
--- a/lib/whitehall/admin_link_lookup.rb
+++ b/lib/whitehall/admin_link_lookup.rb
@@ -1,0 +1,24 @@
+module Whitehall
+  class AdminLinkLookup
+    include Govspeak::EmbeddedContentPatterns
+
+    def self.find_edition(admin_url)
+      case admin_url
+      when ADMIN_ORGANISATION_CIP_PATH
+        organisation = Organisation.find_by(slug: $1)
+        corporate_info_page(organisation: organisation, corporate_info_slug: $2)
+      when ADMIN_WORLDWIDE_ORGANISATION_CIP_PATH
+        organisation = WorldwideOrganisation.find_by(slug: $1)
+        corporate_info_page(organisation: organisation, corporate_info_slug: $2)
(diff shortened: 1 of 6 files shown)


lib/whitehall/admin_link_lookup.rb
```diff
@@ -0,0 +1,24 @@
+module Whitehall
+  class AdminLinkLookup
+    include Govspeak::EmbeddedContentPatterns
+
+    def self.find_edition(admin_url)
+      case admin_url
+      when ADMIN_ORGANISATION_CIP_PATH
+        organisation = Organisation.find_by(slug: $1)
```
h-lame: Using these `$` variables gave me pause as it took me a while to realise that `ADMIN_ORGANISATION_CIP_PATH` is probably a regex with a group in it and `$1` and `$2` will be extracting the slugs.

I wonder if changing them to use named groups (`/(?<name>...)` style) would be clearer (except you have to get them from `$~['name']` which is, probably, even more incomprehensible.  Maybe swapping the case for an `if` with explicit `matchdata = regexp.match(admin_url)` would be clearer.

It's a bigger change than we might want to tackle here though, and I'm not sure how much clearer it'd be.

suzannehamilton: That's a good point. I added the keyword parameters in an attempt to make it clearer, but it's still not obvious that it's performing a regex match.

I'm going to leave this for now. We can revisit it if we think of a way to fix this in a readable way.

PR #3718, opened 2018-01-25
Fix tagging content to world taxons

cbaines:
The World taxons were removed from the Topic taxonomy [1], but this
has broken the tagging interface in Whitehall for users from the
Foreign and Commonwealth Office, so lets just add it back from the
perspective of Whitehall in the short term to fix this functionality.

1: https://trello.com/c/AwIIPzrU/51-remove-the-world-branch-from-the-topic-taxonomy


cbaines:
The Jenkins job has passed, but is stuck on something Docker related, so I've overridden the branch protection and merged it.

PR #3722, opened 2018-01-26
Initial check that scheduled edition has appeared

dwhenry:
This only checks that the edition is present on the site and that any
assocaited announcement has been correctly redirected to the edition.

https://trello.com/c/2G7KW0qS/34-spike-additional-monitoring-for-scheduled-publications-1-day

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,6 +88,7 @@ group :development, :test do
   gem 'pry-byebug'
   gem 'pry-rails'
   gem 'ruby-prof'
+  gem 'sinatra'
   gem 'teaspoon-qunit'
   gem 'test-queue', '~> 0.2.13'
   # teaspoon has coffee assets that mean we need coffee script in order
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -10,7 +10,10 @@ class HealthcheckController < ActionController::Base
 
   def overdue
     # Check the number of overdue editions
-    render json: { 'overdue' => Edition.due_for_publication.count }
+    render json: {
+      'overdue' => Edition.due_for_publication.count,
+      'missing_from_site' => [],
+    }
   end
 
 private
--- a/lib/scheduled_publishing_checker.rb
+++ b/lib/scheduled_publishing_checker.rb
@@ -1,0 +1,76 @@
+class ScheduledPublishingChecker
+  MISSING_EDITION = 'missing_edition'.freeze
+  MISSING_REDIRECT = 'missing_redirect'.freeze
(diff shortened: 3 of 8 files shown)


boffbowsh:
I can't see where this code is actually called, or is this PR still a WIP?

Gemfile
```diff
@@ -88,6 +88,7 @@ group :development, :test do
   gem 'pry-byebug'
   gem 'pry-rails'
   gem 'ruby-prof'
+  gem 'sinatra'
```
h-lame: Huh?  Do we need to add sinatra for this?

app/controllers/healthcheck_controller.rb
```diff
@@ -10,7 +10,10 @@ def check
 
   def overdue
     # Check the number of overdue editions
-    render json: { 'overdue' => Edition.due_for_publication.count }
+    render json: {
+      'overdue' => Edition.due_for_publication.count,
+      'missing_from_site' => [],
```
h-lame: Should this be a fetch of `MissingScheduledPublishing` instances that aren't `SUCCESSFULLY_PUBLISHED` ?

lib/scheduled_publishing_checker.rb
```diff
@@ -0,0 +1,76 @@
...
+      status = check_url(url, announcement_url)
+
+      MissedScheduledPublishing.create(
+        url: url,
+        announcement_url: announcement_url,
+        scheduled_publication: edition.scheduled_publication,
+        status: status,
+        found_at_first_attempt: status == SUCCESSFULLY_PUBLISHED,
+      )
+    end
```
h-lame: Feels weird to me to create all the instances and check them, only to immediately after this loop also check all the ones that haven't succeeded.  Could we create them in a "just added" state in this loop and then pick them up with the rest of them in the next loop.

Or is the intention that new things do get two checks in quick succession?

test/unit/scheduled_publishing_checker_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    setup_server(urls: [], redirects: [])
+
+    ScheduledPublishingChecker.new.check
+
+    assert_equal 1, MissedScheduledPublishing.count
+    assert_equal ScheduledPublishingChecker::MISSING_EDITION, MissedScheduledPublishing.last.status
+  end
+
+  def setup_server(urls:, redirects:)
```
h-lame: Oh, I see now why we're using sinatra - for a fake server to hit in tests.  Could we use webmock instead as it's already in the gemfile?

dwhenry:
We are going to pivot and use the content store report that relies on intent instead.

PR #3723, opened 2018-01-26
Track button, markdown help and external link clicks

fofr:
https://trello.com/c/gIo5fFHw/58-external-link-tracking-in-whitehall-publisher
https://trello.com/c/Uidbttma/62-accordion-tracking-formating-help-in-whitehall-publisher
https://trello.com/c/hlai5gvY/59-button-tracking-in-whitehall-publisher

* Create a track-button-click module for tracking clicks on links or input buttons, use to track buttons on edition show pages and save buttons when editing
* Add click tracking to markdown guidance
* Add tracking to all external links (any link beginning with http and
not linking to the current host)

The JS for button tracking is a modified version of https://github.com/alphagov/govuk_admin_template/blob/master/app/assets/javascripts/govuk-admin-template/modules/track_click.js

The JS for external link tracking is a modified version of https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/external-link-tracker.js


app/assets/javascripts/admin/modules/external_link_tracker.js
```diff
@@ -0,0 +1,35 @@
...
+      }
+
+      GOVUKAdmin.trackEvent('external-link-clicked', href, options);
+    }
+
+    function getLinkFromEvent (evt) {
+      var $target = $(evt.target)
+
+      if (!$target.is('a')) {
+        $target = $target.parents('a')
```
andysellick: This seems really unlikely, but is there any chance this could ever return more than one element, which could potentially break the rest of the code? For that to happen you'd need to have nested a link within a link, which would be odd, so I'm probably just being paranoid. But you could do $target.parents('a').first or something?

fofr: It hasn't been a problem for the version running on www.gov.uk

app/assets/javascripts/admin/modules/track_button_click.js
```diff
@@ -0,0 +1,18 @@
+(function(Modules) {
+  "use strict";
+
+  Modules.TrackButtonClick = function() {
+    this.start = function(container) {
```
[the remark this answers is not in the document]
fofr: It hasn't been a convention in the admin scripts. No reason it shouldn't be.

Issue #3728, opened 2018-01-29
ActionView::Template::Error when asset legacy_url_path includes non-ascii characters


chrislo:
https://sentry.io/govuk/app-whitehall/issues/443169539/

We're seeing `ActionView::Template::Error` in `OrganisationsController#show` when the `legacy_url_path` of an asset contains non-ascii characters. 

```
ActionView::Template::Error: URI must be ascii only "/government/uploads/system/uploads/feature/image/59558/s300_Autonomous_and_electric_shuttle_shown_on_the_esplanade_of_La_De\u0301fense__Paris__2017_by_Sebastien_Durand_Shutterstock.jpg"
```

From https://www.gov.uk/government/organisations/centre-for-connected-and-autonomous-vehicles

PR #3729, opened 2018-01-29
Fix URI::InvalidURIError in AssetManagerStorage

chrislo:
https://github.com/alphagov/whitehall/issues/3728

When the `legacy_url_path` contains non-ascii characters, generating a URI with `URI.join` raises an `URI::InvalidURIError`. We fix this by percent-encoding `legacy_url_path` before joining.

We fixed a similar issue[1] in `gds_api_adapters` in this way.

[1] https://github.com/alphagov/gds-api-adapters/pull/794/


chrislo:
@chrisroos - thank you! That's what I think, yes. Although we were using the asset manager file storage before the `url` method was being delegated to the carrier wave storage.

PR #3732, opened 2018-01-29
Avoid prepending asset host to inline image URLs

chrisroos:
This fixes the broken URLs we were generating when embedding inline images into markdown.

We've recently updated the `ImageUploader` to use the `AssetManagerStorage` engine. The `AssetManagerStorage::File#url` method already includes the asset host so we don't need to prepend it again.

The problem was highlighted by [this end-to-end test][1]

[1]: https://github.com/alphagov/publishing-e2e-tests/blob/master/spec/whitehall/uploading_attachment_spec.rb

### DIFF
--- a/test/unit/helpers/admin/admin_govspeak_helper_test.rb
+++ b/test/unit/helpers/admin/admin_govspeak_helper_test.rb
@@ -93,15 +93,14 @@ class Admin::AdminGovspeakHelperTest < ActionView::TestCase
   end
 
   test "should allow attached images to be embedded in admin html" do
-    images = [OpenStruct.new(alt_text: "My Alt", url: "/image.jpg")]
+    images = [OpenStruct.new(alt_text: "My Alt", url: "https://some.cdn.com/image.jpg")]
     html = govspeak_to_admin_html("!!1", images)
-    assert_select_within_html html, ".govspeak figure.image.embedded img[src=?]", Whitehall.public_asset_host + "/image.jpg"
+    assert_select_within_html html, ".govspeak figure.image.embedded img[src=?]", "https://some.cdn.com/image.jpg"
   end
 
-  test "prefixes embedded image urls with asset host if present" do
-    Whitehall.stubs(:public_asset_host).returns("https://some.cdn.com")
+  test "should allow attached images to be embedded in edition body" do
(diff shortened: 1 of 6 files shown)


chrisroos:
Hey @RyanMacG. All being well this should fix the problem you discovered with inline image URLs. Thanks again for letting us know about it.

RyanMacG:
Thanks @chrisroos can confirm this resolves the E2E scenario we noticed it in.

test/unit/helpers/admin/admin_govspeak_helper_test.rb
```diff
@@ -93,15 +93,14 @@ class Admin::AdminGovspeakHelperTest < ActionView::TestCase
...
 
   test "should allow attached images to be embedded in admin html" do
-    images = [OpenStruct.new(alt_text: "My Alt", url: "/image.jpg")]
+    images = [OpenStruct.new(alt_text: "My Alt", url: "https://some.cdn.com/image.jpg")]
     html = govspeak_to_admin_html("!!1", images)
-    assert_select_within_html html, ".govspeak figure.image.embedded img[src=?]", Whitehall.public_asset_host + "/image.jpg"
+    assert_select_within_html html, ".govspeak figure.image.embedded img[src=?]", "https://some.cdn.com/image.jpg"
   end
 
   test "prefixes embedded image urls with asset host if present" do
```
adrianclay: The name no longer seems to match the behaviour, might be worth renaming this test?

chrisroos: Good spot, @adrianclay. I'll update the test name.

chrisroos:
> Are there any implications for the occurrences where the exception happened? Or are they all related to jobs and will they just retry and succeed?

I think the implication is that editors would've seen broken images in whitehall-admin when they attached an image to a document (e.g. a consultation) and then used the `!!1` markdown syntax to embed that image in the body of the document.

I'm not aware of these exceptions occurring during any Sidekiq jobs but may be wrong.

chrisroos:
Thanks for reviewing, @floehopper. I've updated the test name highlight by @adrianclay, rebased on master and force pushed in preparation for merging.

PR #3734, opened 2018-01-30
Use AssetManagerAndQuarantinedFileStorage for attachments

chrisroos:
This PR configures the `AttachmentUploader` to store attachments on the file system and in Asset Manager. This addresses https://github.com/alphagov/asset-manager/issues/440.

I've tested this change as working in the development vm. I created a consultation and uploaded a single attachment and a single image. The attachment was created in the draft state while the images were created in a non-draft state:

We're not going to be ready to serve these attachment assets from Asset Manager for a while so marking them as draft seems like the safest option in the meantime.

### DIFF
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -1,7 +1,9 @@
 class AssetManagerCreateWhitehallAssetWorker < WorkerBase
   def perform(file_path, legacy_url_path)
     file = File.open(file_path)
-    Services.asset_manager.create_whitehall_asset(file: file, legacy_url_path: legacy_url_path)
+    asset_options = { file: file, legacy_url_path: legacy_url_path }
+    asset_options[:draft] = true if legacy_url_path.match?(/attachment_data/)
+    Services.asset_manager.create_whitehall_asset(asset_options)
     FileUtils.rm(file)
     FileUtils.rmdir(File.dirname(file))
   end
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -1,8 +1,8 @@
 class AssetManagerCreateWhitehallAssetWorker < WorkerBase
-  def perform(file_path, legacy_url_path)
+  def perform(file_path, legacy_url_path, draft = false)
     file = File.open(file_path)
     asset_options = { file: file, legacy_url_path: legacy_url_path }
-    asset_options[:draft] = true if legacy_url_path.match?(/attachment_data/)
+    asset_options[:draft] = true if draft
     Services.asset_manager.create_whitehall_asset(asset_options)
     FileUtils.rm(file)
     FileUtils.rmdir(File.dirname(file))
--- a/test/unit/whitehall_uploader_test.rb
+++ b/test/unit/whitehall_uploader_test.rb
@@ -1,0 +1,7 @@
+require 'test_helper'
+
+class WhitehallUploaderTest < ActiveSupport::TestCase
+  test 'indicates that assets are not protected' do
+    refute WhitehallUploader.new.assets_protected?
+  end
+end
--- a/app/uploaders/attachment_uploader.rb
+++ b/app/uploaders/attachment_uploader.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
 
 class AttachmentUploader < WhitehallUploader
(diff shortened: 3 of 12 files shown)


app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -1,7 +1,9 @@
 class AssetManagerCreateWhitehallAssetWorker < WorkerBase
   def perform(file_path, legacy_url_path)
     file = File.open(file_path)
-    Services.asset_manager.create_whitehall_asset(file: file, legacy_url_path: legacy_url_path)
+    asset_options = {file: file, legacy_url_path: legacy_url_path}
```
floehopper: Very minor: It's usual to have spaces after the opening brace and before the closing brace of a Hash. I would've expected Rubocop to complain about this.

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -1,7 +1,9 @@
...
+    asset_options[:draft] = true if legacy_url_path =~ /attachment_data/
```
floehopper: Minor: I'm not keen on trailing `if` statements - I think the code is clearer if you use the block form.

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -1,8 +1,8 @@
 class AssetManagerCreateWhitehallAssetWorker < WorkerBase
   def perform(file_path, legacy_url_path)
     file = File.open(file_path)
-    asset_options = {file: file, legacy_url_path: legacy_url_path}
-    asset_options[:draft] = true if legacy_url_path =~ /attachment_data/
+    asset_options = { file: file, legacy_url_path: legacy_url_path }
+    asset_options[:draft] = true if legacy_url_path.match?(/attachment_data/)
```
floehopper: Minor: These changes seem orthogonal to the commit note. I think they'd be better combined into the first commit.

test/unit/whitehall_uploader_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class WhitehallUploaderTest < ActiveSupport::TestCase
+  test 'indicates that assets are protected' do
```
floehopper: Typo: I think this should be "indicates that assets are *not* protected"

floehopper:
@chrisroos: One other thought. It would be worth explaining somewhere what the rationale is for marking *all* new attachments as draft, irrespective of their state in Whitehall. Presumably we'll need to update the draft status of these assets when we've implemented https://github.com/alphagov/asset-manager/issues/447.

app/workers/asset_manager_create_whitehall_asset_worker.rb
chrisroos: Rubocop did complain! I fixed it but put it in the subsequent commit by mistake. I've now updated this commit.

app/workers/asset_manager_create_whitehall_asset_worker.rb
chrisroos: I generally agree if it's a long line but I'm happy that this is short enough not to be too unreadable so I'm going to leave it as is.

app/workers/asset_manager_create_whitehall_asset_worker.rb
chrisroos: Good spot. I incorrectly added it to this commit. I've now moved it to the previous commit.

test/unit/whitehall_uploader_test.rb
chrisroos: Correct! I've updated the test name.

chrisroos:
> One other minor thought: it would be nice to use keyword arguments for worker perform method now there are three of them. However, I imagine the switchover might cause problems with queued jobs, so perhaps it's not worth it.

I agree that keyword args would be better but that switching over would be a bit painful so I'm not going to do anything about this for now.

chrisroos:
> One other thought. It would be worth explaining somewhere what the rationale is for marking all new attachments as draft, irrespective of their state in Whitehall.

I've added a note to the commit that marks these attachments as draft in asset-manager. I've added the same note to the description of this PR.

> Presumably we'll need to update the draft status of these assets when we've implemented alphagov/asset-manager#447.

Yes. Or at least before we finally switch the nginx config to serve these attachments.

chrisroos:
Thanks for the review, @floehopper. I've addressed your comments, rebased on master and force pushed in preparation for merging.

PR #3735, opened 2018-01-30
Ensure presenters use correct image urls

chrislo:
This PR fixes two instances where presenters were incorrectly prepending the asset host to image urls. We haven't needed to do that since we switched the storage engine in a024c0f. We fixed another one of these in a746706ac when the end-to-end tests picked up the problem. In the case of this PR the tests were giving false posiitives. This PR adds tests to catch the problem. 

There are two other places where we prepend the asset host in `test/unit/presenters/publishing_api/world_location_news_article_presenter_test.rb:334` where we use the path instead of the URL and in `test/unit/presenters/publishing_api/news_article_presenter_test.rb:241` where we are working with a static image. In both these cases the prepending is correct.

### DIFF
--- a/app/presenters/publishing_api/case_study_presenter.rb
+++ b/app/presenters/publishing_api/case_study_presenter.rb
@@ -66,7 +66,7 @@ module PublishingApi
 
     def image_details
       {
-        url: Whitehall.public_asset_host + presented_case_study.lead_image_path,
+        url: presented_case_study.lead_image_path,
         alt_text: presented_case_study.lead_image_alt_text,
         caption: presented_case_study.lead_image_caption,
       }
--- a/test/unit/presenters/publishing_api/take_part_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/take_part_presenter_test.rb
@@ -8,7 +8,7 @@ class PublishingApi::TakePartPresenterTest < ActiveSupport::TestCase
   test "take part presentation includes the correct values" do
     take_part_page = create(:take_part_page, content_id: SecureRandom.uuid)
 
-    image_url = Whitehall.public_asset_host + take_part_page.image_url(:s300)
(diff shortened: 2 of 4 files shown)


test/unit/presenters/publishing_api/take_part_presenter_test.rb
```diff
@@ -45,6 +45,8 @@ def present(record)
     assert_equivalent_html expected_hash[:details].delete(:body),
       presented_content[:details].delete(:body)
 
+    assert_equal expected_hash[:details][:image][:url], presented_content[:details][:image][:url]
```
chrisroos: Do we need this new assertion given that the line below checks that `expected_hash` equals `presented_content`?

app/presenters/publishing_api/case_study_presenter.rb
```diff
@@ -66,7 +66,7 @@ def body
 
     def image_details
       {
-        url: Whitehall.public_asset_host + presented_case_study.lead_image_path,
+        url: presented_case_study.lead_image_path,
```
chrisroos: In this commit note (for "Use correct image URL in CaseStudyPresenter") you've said, "I've added one [test] so that we prevent this issue again.", but I can't see any new tests. Did you intend to add a test or decide that it wasn't worth it in the end?

chrislo: I think I meant "I've changed the tests to prevent this issue again". I'll fix the commit note.

test/unit/presenters/publishing_api/take_part_presenter_test.rb
chrislo: No. Removed in fixup.

chrislo:
Thanks @chrisroos - rebasing on master and force pushing.

PR #3741, opened 2018-01-31
Reduce n+1 queries on document index

h-lame:
For: https://trello.com/c/59LQ8d4L/35-spike-investigate-whitehall-slowdown-since-christmas-1-day

The document index didn't preload last_author or link_check_reports
associations so it could take some time to render the page as we
load all the extra data for each edition on screen.  Adding some
optional `include(...)` statements to the edition filter will help
speed this up.

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -302,7 +302,17 @@ private
   end
 
   def filter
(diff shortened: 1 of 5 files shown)


h-lame:
Closing because the branch name has a `+` in it and the CI build tries to build a docker using the branch name as a tag, and docker doesn't like `+` chars in tag names so the CI fails.  Replaced by #3747

PR #3742, opened 2018-01-31
Enable uploading of attachments to Asset Manager

chrisroos:
I'm adding this so that we can migrate all existing Whitehall attachments to Asset Manager (see https://github.com/alphagov/asset-manager/issues/441).

We want to mark these attachment assets as draft in Asset Manager to avoid the possibility of exposing potentially sensitive data.

I've tried this in my dev-vm and confirmed that calling the Rake task with the additional `draft` parameter set to `true` results in assets being created in Asset Manager in the draft state.

### DIFF
--- a/lib/tasks/asset_manager.rake
+++ b/lib/tasks/asset_manager.rake
@@ -7,6 +7,10 @@ namespace :asset_manager do
     migrator.perform
   end
 
+  task migrate_attachments: :environment do
+    MigrateAssetsToAssetManager.migrate_attachments
+  end
+
   %i(remove_attachment_file
      remove_consultation_response_form_file
      remove_edition_organisation_image_data_file
--- a/lib/migrate_assets_to_asset_manager.rb
+++ b/lib/migrate_assets_to_asset_manager.rb
@@ -1,6 +1,16 @@
 class MigrateAssetsToAssetManager
   include ActionView::Helpers::TextHelper
 
+  def self.migrate_attachments
+    clean_uploads_root = Pathname.new(Whitehall.clean_uploads_root)
+    attachments_parent_dir = clean_uploads_root.join('system', 'uploads', 'attachment_data', 'file')
+    Pathname.glob(attachments_parent_dir.join('*')).each do |attachment_dir|
+      relative_attachment_dir = attachment_dir.relative_path_from(clean_uploads_root)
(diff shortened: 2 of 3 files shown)


lib/tasks/asset_manager.rake
```diff
@@ -28,9 +29,10 @@ namespace :asset_manager do
   private
 
   def usage_string
-    %{Usage: asset_manager:migrate_assets[<path>]
+    %{Usage: asset_manager:migrate_assets[<path>,<draft>]
 
       Where <path> is a subdirectory under Whitehall.clean_uploads_root e.g. `system/uploads/organisation/logo`
+      And <draft> is an optional Boolean representing whether the uploaded asset should be in the draft state
```
floehopper: Very minor: maybe note default is `false` and that `true` should be represented by the string `'true'`. I might've expected `'1'` to work, but I don't think it will.

chrisroos:
I've updated this PR to add the `asset_manager:migrate_attachments` Rake task.

@chrislo: Is this the sort of change you were imagining?

chrislo:
@chrisroos - I think so yes. I'm not sure if it'll mean that the task runs quicker, but it should mean that we don't have to build an array with around 1.5m entries before we can start adding jobs to the queue.

chrisroos:
Thanks chaps. I've rebased on master and force pushed in preparation for merging.

chrisroos:
>> I've tried this in my dev-vm and confirmed that calling the Rake task with the additional draft parameter set to true results in assets being created in Asset Manager in the draft state.

> How sure are you that passing `draft: true` into `create_whitehall_asset` is creating a draft asset, because `draft` is set to `true` rather than the fact that `:draft` key exists at all? I'm not sure how this stuff is processed when it goes via `gds-api-adapters`.

As discussed in person. I'm not 100% sure but I suspect that `draft` will need to be set to something truthy in order for Rails to set it to `true` in https://github.com/alphagov/asset-manager/blob/3bf8d830cc30e225c24f45332e8e777a70b1d719/app/controllers/whitehall_assets_controller.rb#L28 (i.e. it's probably not enough for the `:draft` key to exist).

lib/tasks/asset_manager.rake
chrisroos: Note that I've since removed the commit that introduced this change.

PR #3743, opened 2018-01-31
Rake tasks to delete non database backed assets from file system

chrislo:
See: alphagov/asset-manager#216

There are a number of assets under `system/uploads/` that are associated with mounted carrierwave uploaders that have subsequently been deleted from the code base. The assets however still have public URLs. We are migrating those assets to Asset Manager and once complete, the rake tasks added in this PR can be run to remove the files from NFS. 

I've also removed some of the rake tasks that have already been run.

### DIFF
--- a/lib/asset_remover.rb
+++ b/lib/asset_remover.rb
@@ -1,4 +1,39 @@
 class AssetRemover
+  def remove_attachment_file
+    target_dir = File.join(Whitehall.clean_uploads_root, 'system', 'uploads', 'attachment', 'file')
+    remove_asset_dir(target_dir)
+  end
+
+  def remove_consultation_response_form_file
+    target_dir = File.join(Whitehall.clean_uploads_root, 'system', 'uploads', 'consultation_response_form', 'file')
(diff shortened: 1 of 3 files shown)


lib/asset_remover.rb
```diff
@@ -1,4 +1,39 @@
 class AssetRemover
+  def remove_attachment_file
+    target_dir = File.join(Whitehall.clean_uploads_root, 'system', 'uploads', 'attachment', 'file')
```
floehopper: Just to double-check: This is distinct from the attachments which are still managed by the app, right? They've got `attachment_data` in the path, I think.

floehopper:
@chrislo Note my question is in a comment on an outdated diff in case you missed it!

lib/asset_remover.rb
chrislo: Yes, that's right. This directory `system/uploads/attachment/file` has 6882 files in it and hasn't had any new files added to it since at least mid-December last year. It looks like the mounted uploader was removed from `Attachment` in dde83ebde0.

PR #3746, opened 2018-02-01
[DO NOT MERGE] Spike to import HMCTS forms

suzannehamilton:
The code is a bit messy and full of TODOs - I'm just opening this PR to get some feedback on the approach, which is:

- Developer exports the spreadsheet as a CSV and copies it to a whitehall machine
- A new whitehall rake task reads the CSV
- For each form in the CSV, it creates a new draft publication
- It downloads all the forms for the publication
- It saves each form as a `FileAttachment` linked to the publication which, as far as I can tell on my dev VM, uploads them to asset-manager for virus scanning

It doesn't yet save all the publication data from the CSV. I concentrated on the attachments and the mandatory fields.

My main question is whether it's OK to download the attachments directly to the whitehall machine.

https://trello.com/c/ZLzjhOhB/57-spike-into-how-we-might-bulk-creating-draft-whitehall-content-based-on-a-spreadsheet-from-hmcts-1-day

### DIFF
--- a/lib/import/hmcts_csv_parser.rb
+++ b/lib/import/hmcts_csv_parser.rb
@@ -1,0 +1,71 @@
+module Import
+  class HmctsCsvParser
+    CSV_HEADERS = %i(
+      publication_page_id
+      artefact_code
+      category
+      artefact_file_name
+      artefact_type
+      artefact_title
+      artefact_link
+      artefact_url
+      publication_type
+      publication_title
+      publication_summary
+      publication_body
+      published_before
+      has_images
+      alternative_format_email
+      policy
+      policy_areas
+      lead_organisations
+      supporting_organisations
+      excluded_nations
+      access_limiting
+      schedule_publishing
+      attachment_accessible
+      new_publication_url
+      new_attachment_url
+    ).freeze
+
+    PUBLICATION_TYPE_SLUGS = {
+      "Form" => "forms",
+      "Guidance" => "guidance",
+    }.freeze
+
+    def self.publications(path)
+      csv = CSV.read(path, headers: CSV_HEADERS)
+
+      Enumerator.new do |yielder|
+        page_id = nil
+        publication_details = {}
+
+        # Skip the header row
+        csv[1..-1].each do |row|
+          if row[:publication_page_id] == page_id
+
+          else
+            yielder << publication_details unless page_id.nil?
+
+            publication_details = {
+              publication_type: publication_type_slug(row[:publication_type]),
+              title: row[:publication_title],
+              summary: row[:publication_summary],
+              body: row[:publication_body],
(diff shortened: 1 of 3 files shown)


lib/import/hmcts_csv_parser.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def self.publications(path)
+      csv = CSV.read(path, headers: CSV_HEADERS)
+
+      Enumerator.new do |yielder|
+        page_id = nil
+        publication_details = {}
+
+        # Skip the header row
+        csv[1..-1].each do |row|
```
h-lame: If we use `CSV.read(path, headers: true, header_converters: :symbol)` then the first row is consumed and used as the headers definition and the strings are turned into lowercased symbols.  Meaning we can just do `csv.each` here. A tiny part of my brain wonders if the `[1..-1]` means we have to read in the whole file first to convert this to an array and then do the range thing, meaning we need a bunch of memory.  A different tiny part of my brain assumes that CSV and Enumerable do something smart so this isn't the case, but it's a smaller part of my brain and the first part is noisier.

lib/import/hmcts_importer.rb
```diff
@@ -40,7 +40,8 @@ def self.default_organisation
     end
 
     def self.create_attachments(attachments, publication)
-      temp_directory = "/tmp/#{SecureRandom.uuid}"
+      # TODO: Is it safe to save these straight to a whilehall machine? If so, what directory?
+      temp_directory = "/tmp/hmcts_attachments_#{SecureRandom.uuid}"
```
h-lame: I think it'd be ok, but also we can use [`require 'tmpdir'; Dir.mktmpdir`](https://ruby-doc.org/stdlib-2.3.1/libdoc/tmpdir/rdoc/Dir.html) to get a definitely temp dir (and I think if we use the block form it'd be removed by ruby for us when we're done).

lib/import/hmcts_csv_parser.rb
suzannehamilton: Hmm. I thought CSV did lazy-loading, but it looks like it stores everything in a big array. So the Enumerator's a bit unnecessary, and the `[1..-1]` presumably makes us loop over the data twice. 😞 

I don't think we can use `header_converters` here, unfortunately. Some of the columns have the same name, and others are quite verbose. I'm also expecting HMCTS to change a couple of them. It seemed less error-prone to just define them ourselves (even though we'll have to update the list if they add new columns).

h-lame: ah, shame about the headers thing :(

suzannehamilton:
I'm closing this one because we don't want to merge this spike. I won't delete the branch, though.

PR #3747, opened 2018-02-01
Reduce n+1 queries on document index

h-lame:
For: https://trello.com/c/59LQ8d4L/35-spike-investigate-whitehall-slowdown-since-christmas-1-day

The document index didn't preload last_author or link_check_reports
associations so it could take some time to render the page as we
load all the extra data for each edition on screen.  Adding some
optional `include(...)` statements to the edition filter will help
speed this up.

Replaces #3741 because CI tries to build a docker using the branch name as a tag, but docker doesn't allow `+`s in tags so it fails.

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -302,7 +302,17 @@ private
   end
 
   def filter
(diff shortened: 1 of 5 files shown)


h-lame:
We reverted this because it made things worse, not better.  It's probably because the link_check_reports table is a) massive, b) had no indexes.  Once we resolve that (see: #3824 for some ideas) we can probably revert the revert on this.

PR #3748, opened 2018-02-01
Add size to AssetManagerAndQuarantinedFileStorage

chrisroos:
To avoid the following exception when calling `AttachmentData#file_is_not_empty`.

    NoMethodError
    undefined method `size' for
    #<Whitehall::AssetManagerAndQuarantinedFileStorage::File:0x00007f185b9de250>

This should fix the [exceptions we've been seeing in production][1] since deploying the changes in
https://github.com/alphagov/whitehall/pull/3734.

[1]: https://sentry.io/govuk/app-whitehall/issues/446471956/events/13256559568/


chrisroos:
Thanks @floehopper. I'll get this merged and deployed.

PR #3750, opened 2018-02-01
Delete New Relic config and dependency

suzannehamilton:
We no longer use New Relic, even in development mode, so the gem and its config can be removed.


suzannehamilton:
Nobody's objected, so I'm going to merge this. I'll revert and update the docs if it turns out that people are using New Relic in dev.

PR #3751, opened 2018-02-01
Redirect invalid email signups to root

chrisroos:
The motivation is to avoid the [`ActionController::ParameterMissing` exceptions being caught in Sentry][1].

This means that requesting https://www.gov.uk/government/email-signup/new will result in a redirect to www.gov.uk, instead of the current 400 Bad Request response. This mirrors the redirect we send in response to requests for https://www.gov.uk/government/email-signup (although that's handled in routes.rb).

I'm not sure how best to avoid these exceptions. It would seem reasonable to display some kind of useful information if someone were to request /government/email-signup/new but that's more work than I have time for. Another option is to ignore these `ActionController::ParameterMissing` exceptions in Sentry (there's precedent for this as they're [ignored by default by HoneyBadger][2]) but it feels like that might have other consequences I'm not aware of.

[1]: https://sentry.io/govuk/app-whitehall/issues/343728660/
[2]: https://docs.honeybadger.io/ruby/getting-started/ignoring-errors.html

### DIFF
--- a/app/controllers/email_signups_controller.rb
+++ b/app/controllers/email_signups_controller.rb
@@ -1,5 +1,10 @@
 class EmailSignupsController < PublicFacingController
   def new
+    unless params[:email_signup].present?
+      redirect_to '/'
+      return
+    end
+
     @email_signup = EmailSignup.new(email_signup_params)
   end
 
--- a/test/functional/email_signups_controller_test.rb
+++ b/test/functional/email_signups_controller_test.rb
@@ -5,6 +5,12 @@ class EmailSignupsControllerTest < ActionController::TestCase
   include FeedHelper
   include GdsApi::TestHelpers::EmailAlertApi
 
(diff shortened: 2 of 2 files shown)


tijmenb:
`ActionController::ParameterMissing` errors should probably not be reported to Sentry. We did an RFC on errors a while back and I think the error we're dealing here with here is "Bad request error":

https://github.com/alphagov/govuk-rfcs/blob/master/rfc-087-dealing-with-errors.md#bad-request-errors

The desired behaviour is "user gets feedback, error is not reported to Sentry". I think it would make sense to add it to Sentry's `excluded_exceptions`.

chrisroos:
Thanks @tijmenb! I'll take a look at adding `ActionController::ParameterMissing` to the list of excluded exceptions.

thomasleese:
I've [added a PR in `govuk_app_config` for this](https://github.com/alphagov/govuk_app_config/pull/41).

PR #3753, opened 2018-02-02
Move asset_migration queue to the lowest priority

chrisroos:
We're about to migrate about 1.5m attachments from Whitehall to Asset Manager and we think this might take a few days (partly because our Sidekiq config defines a concurrency of 1). This would block any `link_checks` jobs (currently scheduled at about 2am each day) from running until all the assets have migrated. Moving the asset_migration queue to the bottom of the list will ensure that Sidekiq will only process our jobs once the other queues are empty.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -11,8 +11,8 @@
   - email_alert_api_signup
   - bulk_republishing
   - sync_checks
-  - asset_migration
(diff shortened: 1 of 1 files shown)


chrisroos:
Thanks @floehopper. I'll get this merged and deployed.

PR #3755, opened 2018-02-02
[WIP] Spike update draft status of assets in Asset Manager

floehopper:
This is my first stab at how this might work. I've tested this on my development VM with the appropriate asset route setup for Whitehall attachments being served from Asset Manager and it appears to work:

1. Create draft publication
2. Add a PDF attachment
3. Preview publication on draft stack
4. Confirm that links to attachment and its thumbnail image require that the user is signed in
5. Publish publication
6. View publication on website
7. Confirm that links to attachment and its thumbnail image do not require that the user is signed in

I haven't tried anything funky like un-publishing / withdrawing, but I _think_ it should work.

/cc @chrislo, @chrisroos

### DIFF
--- a/app/services/service_listeners/attachment_draft_status_updater.rb
+++ b/app/services/service_listeners/attachment_draft_status_updater.rb
@@ -1,0 +1,24 @@
+module ServiceListeners
+  class AttachmentDraftStatusUpdater
+    attr_reader :edition
+
+    def initialize(edition)
+      @edition = edition
+    end
+
+    def update!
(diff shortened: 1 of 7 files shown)


floehopper:
I've fixed the lint errors and force-pushed.

floehopper:
I've switched the `AttachmentUploader` back to use the composite storage class - this change was causing lots of test failures and we don't need to make this change yet.

floehopper:
Closing in favour of #3765.

PR #3757, opened 2018-02-05
Migrate attachments in batches

chrislo:
See: https://github.com/alphagov/asset-manager/issues/441

There are around 1.5m assets under `attachment_data/file`. When we attempted to queue up jobs to migrate all of these assets we ran into issues with redis running out of memory. This PR allows us to specify a "batch" of assets to be migrated so that we hopefully avoid this problem again.

### DIFF
--- a/lib/migrate_assets_to_asset_manager.rb
+++ b/lib/migrate_assets_to_asset_manager.rb
@@ -1,10 +1,12 @@
 class MigrateAssetsToAssetManager
   include ActionView::Helpers::TextHelper
 
-  def self.migrate_attachments
+  def self.migrate_attachments(batch_start = 0, batch_end = Float::INFINITY)
     clean_uploads_root = Pathname.new(Whitehall.clean_uploads_root)
     attachments_parent_dir = clean_uploads_root.join('system', 'uploads', 'attachment_data', 'file')
     Pathname.glob(attachments_parent_dir.join('*')).each do |attachment_dir|
+      next unless (batch_start..batch_end).cover?(directory_number_from_attachment_dir(attachment_dir))
+
       relative_attachment_dir = attachment_dir.relative_path_from(clean_uploads_root)
       migrator = MigrateAssetsToAssetManager.new(relative_attachment_dir.to_s, true)
       migrator.perform
@@ -24,6 +26,10 @@ class MigrateAssetsToAssetManager
 
   def to_s
     "Migrating #{pluralize(@relative_file_paths.size, 'file')}"
+  end
+
+  def self.directory_number_from_attachment_dir(attachment_dir)
+    Integer(attachment_dir.each_filename.entries.last)
   end
 
   class Worker < WorkerBase
--- a/lib/tasks/asset_manager.rake
+++ b/lib/tasks/asset_manager.rake
@@ -7,8 +7,10 @@ namespace :asset_manager do
     migrator.perform
   end
 
-  task migrate_attachments: :environment do
-    MigrateAssetsToAssetManager.migrate_attachments
+  desc "Migrate assets under system/uploads/attachment_data/file to Asset Manager"
+  task :migrate_attachments, %i(batch_start batch_end) => :environment do |_, args|
+    abort(migrate_attachments_usage_string) unless args[:batch_start] && args[:batch_end]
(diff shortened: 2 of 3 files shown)


lib/migrate_assets_to_asset_manager.rb
```diff
@@ -26,6 +28,10 @@ def to_s
     "Migrating #{pluralize(@relative_file_paths.size, 'file')}"
   end
 
+  def self.directory_number_from_attachment_dir(attachment_dir)
+    attachment_dir.each_filename.entries.last.to_i
```
floehopper: Minor: Given we're confident that all directories are integers, I might use `Integer(attachment_dir.each_filename.entries.last)` rather than `String#to_i` so that this line will fail fast if the directory is not an integer. Note that `"abc".to_i # => 0`.

floehopper: I found this line `attachment_dir.each_filename.entries.last` hard to understand.

As I understand it, `attachment_dir` at this point would be something along the lines of `system/uploads/attachment_data/file/100/filename.pdf`. Is that right? I assume the `.each_filename.entries` bit is somehow obtaining `'100'` from this, but it's not at all obvious how this is working.

It's not obvious to me that this code is tested, so I think it would be useful to do one of the following: (a) add some tests; (b) extract explanatory local variables; or (c) explain how this works in the commit note.

lib/tasks/asset_manager.rake
```diff
@@ -7,8 +7,8 @@ namespace :asset_manager do
     migrator.perform
   end
 
-  task migrate_attachments: :environment do
-    MigrateAssetsToAssetManager.migrate_attachments
+  task :migrate_attachments, %i(batch_start batch_end) => :environment do |_, args|
+    MigrateAssetsToAssetManager.migrate_attachments(args[:batch_start].to_i, args[:batch_end].to_i)
```
floehopper: Minor: I think this implies the arguments are required. Is that right? If so, it would be nice to add a task description to explain this.

floehopper: Minor: See my previous concern about using `String#to_i`. However, I'm not so bothered about it here.

lib/migrate_assets_to_asset_manager.rb
chrislo: > As I understand it, attachment_dir at this point would be something along the lines of system/uploads/attachment_data/file/100/filename.pdf. Is that right?

No, i think it would be `system/uploads/attachment_data/file/100` due to the call to `glob('system/uploads/attachment_data/file/*')` in the block argument above. If the path was a string I'd do `.split('/').last` but `Pathname#split` doesn't work like `String#split` so I'd have to convert it to a String first, which is also confusing I think. I think the expectation test I added does test this method implicitly. I'll see if I can make it more obvious. 

The change to `Integer()` is a good idea - I'll do that.

test/unit/migrate_assets_to_asset_manager_test.rb
```diff
@@ -120,6 +120,14 @@ class MigrateAttachmentsToAssetManagerTest < ActiveSupport::TestCase
 
     MigrateAssetsToAssetManager.migrate_attachments
   end
+
+  test 'specifying a start and end range' do
+    MigrateAssetsToAssetManager.unstub(:new)
```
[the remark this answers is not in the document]
chrislo: I removed the call to `unstub` by expecting the code to `never` be called with the path ending in `100` when the batch arguments are used.

test/unit/migrate_assets_to_asset_manager_test.rb
```diff
@@ -120,6 +120,14 @@ class MigrateAttachmentsToAssetManagerTest < ActiveSupport::TestCase
...
+    MigrateAssetsToAssetManager.expects(:new).with('system/uploads/attachment_data/file/200', true).returns(@migrator)
```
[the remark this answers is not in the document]
chrislo: I feel like the expectations show that we are expecting another part of the code to be called only if the directory name is within the batch range. I've added an additional expectation and removed the `unstub` to clarify this.

lib/tasks/asset_manager.rake
chrislo: I've added a description and usage guidance if the arguments are not supplied.

chrislo:
Thanks @floehopper - rebasing and force pushing before merging.

PR #3759, opened 2018-02-06
Improvements to AssetManagerAndQuarantinedFileStorage

floehopper:
This PR contains a couple of changes which will make it easier to add functionality to update the draft status of existing attachment assets in Asset Manager:

1. Fix the return value for `AssetManagerAndQuarantinedFileStorage#store!`
  * Previously this was returning an instance of `CarrierWave::SanitizedFile` rather than `Whitehall::AssetManagerAndQuarantinedFileStorage::File`.
  * This was inconsistent and confusing.
  * It also caused problems in some tests I was writing where I wanted access to the relevant `File` instance from the `AttachmentUploader` after `#store!` had been called, but before `#retrieve!` had been called.

2. Make Asset Manager version of path available when using composite storage
  * I want to add some functionality to update the draft status of an attachment asset in Asset Manager.
  * This will need to operate outside the realms of `CarrierWave` and so I need to make this method available.
  * We should be able to remove it again once we switch the `AttachmentUploader` to use the `Whitehall::AssetManagerStorage` class instead of the `Whitehall::AssetManagerAndQuarantinedFileStorage` class.

I've also made some minor improvements to one of the tests.

### DIFF
--- a/lib/whitehall/asset_manager_and_quarantined_file_storage.rb
+++ b/lib/whitehall/asset_manager_and_quarantined_file_storage.rb
@@ -25,6 +25,10 @@ class Whitehall::AssetManagerAndQuarantinedFileStorage < CarrierWave::Storage::A
       @quarantined_file.url
     end
 
+    def asset_manager_path
+      @asset_manager_file.path
+    end
+
     def delete
       @quarantined_file.delete
       @asset_manager_file.delete
--- a/lib/whitehall/asset_manager_and_quarantined_file_storage.rb
+++ b/lib/whitehall/asset_manager_and_quarantined_file_storage.rb
@@ -1,7 +1,9 @@
 class Whitehall::AssetManagerAndQuarantinedFileStorage < CarrierWave::Storage::Abstract
   def store!(file)
-    Whitehall::AssetManagerStorage.new(uploader).store!(file)
-    Whitehall::QuarantinedFileStorage.new(uploader).store!(file)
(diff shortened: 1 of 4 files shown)


lib/whitehall/asset_manager_and_quarantined_file_storage.rb
```diff
@@ -25,6 +25,10 @@ def url
       @quarantined_file.url
     end
 
+    def asset_manager_path
+      @asset_manager_file.path
```
chrislo: I wonder if we should add an accessor for `legacy_url_path` to AssetManagerStorage and delegate to that instead? If I understand the commit message correctly, you will be eventually be looking up an asset in asset manager using the `legacy_url_path` passing in this method?

floehopper: I'm slightly unsure what you mean here. Do you mean something like this:

`AttachmentUploader#legacy_url_path` -> `Whitehall::AssetManagerAndQuarantinedFileStorage::File#legacy_url_path` -> `Whitehall::AssetManagerFileStorage::File#legacy_url_path`

If so, I'm not sure this is clearly better and in any case, since I expect these `asset_manager_path` methods to go away, so I think I'm going to merge this as it stands so I can make progress. Please feel free to add more explanatory comments here or elsewhere if you think it would be worth making a change.

floehopper:
@chrislo Thanks. I'm going to get this merged.

PR #3760, opened 2018-02-08
Handle external urls

cbaines:


app/models/html_attachment.rb
```diff
@@ -62,7 +62,7 @@ def url(options = {})
 
     if preview
       options[:preview] = id
-      options[:host] = Plek.find_uri("draft-origin").host
+      options[:host] = URI(Plek.new.external_url_for("draft-origin")).host
     else
       options[:host] = Plek.find_uri("www-origin").host
```
koetsier: might have to be external

app/workers/check_organisation_links_worker.rb
```diff
@@ -26,6 +26,6 @@ def public_editions(organisation)
   end
 
   def callback
-    Whitehall::UrlMaker.new(host: Plek.find('whitehall-admin')).admin_link_checker_api_callback_url
+    Whitehall::UrlMaker.new(host: Plek.new.external_url_for('whitehall-admin')).admin_link_checker_api_callback_url
```
koetsier: Could perhaps change AdminRequest in routes.rb to allow internal URLs as well?

app/models/html_attachment.rb
cbaines: Indeed, good spot, accessing www-origin is very odd, but I think I've found the two places where this will have an effect, and both look to be in URLs that go in to links:

1: https://github.com/alphagov/whitehall/blob/dcda55d77fd39ceb6d784aa7726c54d4dea47960/app/views/admin/attachments/_attachments.html.erb#L16
2: https://github.com/alphagov/whitehall/blob/78fbbe66dde427289147cea579dbba3ff4f2c0e2/app/views/admin/editions/_show_attachments.html.erb#L14

cbaines: I've changed this to be an external URL in 095989f

app/workers/check_organisation_links_worker.rb
cbaines: Yes, I think this is neater, I've done this in 30675a2

PR #3763, opened 2018-02-08
Refactor PublicUploadsController

chrislo:
See: https://github.com/alphagov/asset-manager/issues/213

We are now serving all of Whitehall's non-attachment assets from AssetManager. This means the PublicUploadsController has 3 remaining responsibilities:

1. Redirect all requests using relative URLs for non-attachment assets (e.g. '/government/uploads/system/uploads/person/image/person.png') to the asset host (and therefore Asset Manager)
2. Serve a small number of fixed HMRC assets in `/uploaded/hmrc`
3. Act as the super class for AttachmentController

This PR splits the functionality into 3 parts. First we extract the HMRC related functionality into a new controller. Then we break the inheritence between AttachmentsController and PublicUploadsController by moving methods from the latter into the former. We then simplify PublicUploadsController and HmrcAssetsController in that order.


chrisroos:
I'm going to review this now, @chrislo.

test/integration/upload_access_test.rb
```diff
@@ -47,27 +43,22 @@ def assert_sent_private_upload(_upload, content_type)
...
 
-  test 'allows everyone access to general uploads' do
+  test 'allows everyone to be redirected to general uploads' do
     upload = '/government/uploads/general-upload.jpg'
     create_uploaded_file(path_to_clean_upload(upload))
 
     get_via_nginx upload
 
-    assert_sent_public_upload upload, Mime[:jpg]
+    assert_redirected_to "/government/uploads/general-upload.jpg"
```
chrisroos: I think this assertion might be slightly clearer if the `assert_redirected_to` included the asset host. As it stands, the test implies that you're always redirected to the same path that was requested.

test/integration/upload_access_test.rb
```diff
@@ -47,27 +43,22 @@ def assert_sent_private_upload(_upload, content_type)
     host! asset_host
   end
 
-  test 'allows everyone access to general uploads' do
+  test 'allows everyone to be redirected to general uploads' do
```
chrisroos: I think this might be better named as something like "redirects all non-attachment, non-hmrc asset requests to the asset host"

app/controllers/asset_manager_redirect_controller.rb
```diff
@@ -0,0 +1,6 @@
+class AssetManagerRedirectController < ApplicationController
+  def show
```
chrisroos: I think it might be worth adding something about the possibility of an infinite redirect loop in the commit note for "PublicUploadsController only needs to redirect". Once this change has been deployed, and until we remove the [`/government/uploads` catch-all route from the assets-origin nginx config][1], it's going to be possible to request a made-up asset path (e.g. https://assets-origin.integration.publishing.service.gov.uk/government/uploads/foo/bar/baz.png) and have it result in an infinite redirect.

[1]: https://github.com/alphagov/govuk-puppet/blob/7e6954cda9f7976d552faa12b0019eb054d4772b/hieradata/common.yaml#L1643

chrislo:
Thanks @chrisroos. I've addressed your comments in 699d44d, rebased, force-pushed ready to merge.

PR #3764, opened 2018-02-09
Delete Number 10 virtual tour

### LINKED ISSUE #3632: No.10 page should not require / use Adobe Flash
As discovered by @joelanman, the No.10 history page appears to require Adobe Flash - with it blocked you get the below warning. It also has non standard navigation.

image_720

https://github.com/alphagov/whitehall/blob/a41ef2edaec2ce63facb2998d3d62c56252a70fe/app/views/histories/10_downing_street.html.erb#L326

As far as I'm aware, we avoid / don't allow flash on GOV.UK. Hoping this can be escalated.

fofr:
Delete the flash tour from the Number 10 page, linking instead to Google’s version: https://www.google.com/culturalinstitute/beta/u/0/partner/10-downing-street

Deletes:
* flash file
* JS, CSS, XML and image files
* content on Number 10 history page

Fixes #3632 Undoes https://github.com/alphagov/whitehall/pull/474


nickcolley:
Thanks so much for this Paul 👍 

Flash superhero running up a building

edwardhorsford:
Yay thanks @fofr!

joelanman:
Great work everyone, really appreciate theres a replacement

chrisroos:
> I think the images referred to in the xml.erb files live on the server (a direct upload I think, 'cos they're not in this repo) so we might want to make a note somewhere to remove those too.

These images have been migrated to Asset Manager so will need to be deleted from there if required.

chrislo:
See: https://github.com/alphagov/asset-manager/issues/268. Note these URLs are still valid so you'd have to put some redirects if you deleted the assets.

h-lame:
I think there aren't really any useful places we could redirect someone requesting one of those images.  For example: some of them are UI elements for the player, whereas others are the texture skins of the rooms for the player, but they're not flat photos so they need the player to interpret and deform them to be viewable properly.  Probably we'd just 410 them?

PR #3765, opened 2018-02-09
Update draft status of attachment assets

floehopper:
The Asset Manager assets for attachments are currently being created in the draft state by default. The changes in this pull request aim to update the draft status of attachment assets as their parent models change state. Hopefully the following gives a decent overview of the changes, but please see the individual commit notes for more detailed info.

There are three subclasses of the `Attachment` class, `FileAttachment`, `HtmlAttachment` & `ExternalAttachment`, but we are only concerned with `FileAttachment`, because they are the only ones stored in the Asset Manager.

I've used the existing `AttachmentVisibilty` class and its `#visible?` method to determine whether the public is allowed access to an attachment.

#### Editions

Models inheriting directly or indirectly from the `Edition` class may support attachments depending on whether they include the `Attachable` module. Whether or not a model supports attachments can be determined by calling the `#allows_attachments?` method.

I've used the existing `EditionServiceCoordinator` mechanism to hook a new service listener into `Edition` lifecycle events in `config/initializers/edition_services.rb`. `Admin::EditionWorkflowController` actions seem to use instances of `EditionService` to make all significant changes in `Edition` state and thus the new `ServiceListeners::AttachmentDraftStatusUpdater` should receive events for all these changes.

#### (Consultation) Responses

The terminology around these models is quite confusing. There is a base class, `Response`, with two subclasses, `ConsultationOutcome` & `ConsultationPublicFeedback` and these can both be (independently) associated with a `Consultation` via the `outcome` & `public_feedback` associations respectively. Anyway, the important thing for this pull request is that both subclasses of `Response` support attachments despite the fact they do not inherit from `Edition`.

The visibility of attachments on a `Response` is determined by the state of its `Consultation`. In order to keep the draft status of the corresponding assets up-to-date I've made the following changes:

1. Extended the logic in `config/initializers/edition_services.rb` to detect changes in the state of a `Consultation` and update the draft status of attachments on any associated `Response`.

2. Introduced a new `Whitehall.consultation_response_notifier` in order to detect the creation or updating of a `Response`. An instance of `ServiceListeners::AttachmentDraftStatusUpdater` is again used to handle events for this notifier and update the draft status of assets accordingly. This is needed, because e.g. an attachment can be added to a `Response` which is already associated with a published `Consultation`.

#### PolicyGroup

This model class also supports attachments despite the fact that it doesn't inherit from `Edition` either.

The visibility of attachments on a `PolicyGroup` is solely determined by their existence, i.e. a `PolicyGroup` is always public and so any attachments on it are also immediately public.

In order to keep the draft status of the associated assets up-to-date, I've introduced a new `Whitehall.policy_group_notifier` to detect the creation or updating of a `PolicyGroup`. An instance of  `ServiceListeners::AttachmentDraftStatusUpdater` is again used to handle events from this notifier and update the draft status of assets accordingly.

### Testing

I've added unit tests and an integration test, `AttachmentDraftStatusIntegrationTest`, but I decided not to add any Cucumber features, because the change in behaviour is not visible via the user interface. Also the integration test only covers a subset of the possible scenarios in order to avoid a combinatorial explosion - it's intended mainly to check that everything is wired up correctly.

### Risks

* It's possible that some important changes in `Edition` state happen without generating events from the `EditionServiceCoordinator` notifier. However, I'm pretty confident that they don't.
* Similarly it's possible that a new `Edition` state change might be added without making use of the `EditionServiceCoordinator` services.
* It's also possible that a new non-`Edition` model might be added with support for attachments.

However, in all these cases the consequence would be that the relevant attachments would always be in the draft state, i.e. their draft status would never be updated. Although this wouldn't be ideal, it would be considerably better than if attachments were inadvertently made public when they shouldn't be.

### DIFF
--- a/app/controllers/admin/policy_groups_controller.rb
+++ b/app/controllers/admin/policy_groups_controller.rb
@@ -12,6 +12,7 @@ class Admin::PolicyGroupsController < Admin::BaseController
   def create
     @policy_group = PolicyGroup.new(policy_group_params)
     if @policy_group.save
+      Whitehall.policy_group_notifier.publish('create', @policy_group)
       redirect_to admin_policy_groups_path, notice: %{"#{@policy_group.name}" created.}
     else
       render action: "new"
@@ -25,6 +26,7 @@ class Admin::PolicyGroupsController < Admin::BaseController
   def update
     @policy_group = PolicyGroup.friendly.find(params[:id])
     if @policy_group.update_attributes(policy_group_params)
+      Whitehall.policy_group_notifier.publish('update', @policy_group)
       redirect_to admin_policy_groups_path, notice: %{"#{@policy_group.name}" saved.}
     else
       render action: "edit"
--- a/app/controllers/admin/responses_controller.rb
+++ b/app/controllers/admin/responses_controller.rb
@@ -13,6 +13,7 @@ class Admin::ResponsesController < Admin::BaseController
     @response = response_class.new(response_params)
     @response.consultation = @edition
     if @response.save
+      Whitehall.consultation_response_notifier.publish('create', @response)
       redirect_to [:admin, @edition, @response.singular_routing_symbol], notice: "#{@response.friendly_name.capitalize} saved"
     else
       render :show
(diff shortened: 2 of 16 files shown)


chrisroos:
I'm going to review this now.

floehopper:
Rebasing against `master` and force-pushing.

floehopper:
We've realised there are some flaws in this approach. Closing.

floehopper:
Superseded by #3768.

PR #3766, opened 2018-02-09
Allow exclusion of england

dwhenry:
Currently, when adding a publication you can indicate it to not be relevant to Scotland, NI or Wales, it is always however relevant to England. For some Stats publications, the document is actually only relevant to Scotland, This PR makes it possible to make a document not relevant to England.

Add error validation to ensure at least one nation is applicable:

screenshot 2018-02-15 12 41 20

And highlight in the form:

screenshot 2018-02-15 12 41 30

This change allows documents to not be application to England.

Before:

screenshot 2018-02-15 11 14 13

After:

screenshot 2018-02-15 11 15 10

https://trello.com/c/ukIxMUVM/66-scotland-report-shows-this-applies-to-england-and-wales

### DIFF
--- a/app/models/edition/national_applicability.rb
+++ b/app/models/edition/national_applicability.rb
@@ -12,6 +12,7 @@ module Edition::NationalApplicability
   included do
     has_many :nation_inapplicabilities, foreign_key: :edition_id, dependent: :destroy, autosave: true
     validates_associated :nation_inapplicabilities
+    validates :nation_inapplicabilities, length: { maximum: Nation.all.count - 1, message: 'can not exclude all nations' }
 
     add_trait Trait
   end
--- a/app/models/nation.rb
+++ b/app/models/nation.rb
@@ -25,7 +25,7 @@ class Nation
   end
 
   def self.potentially_inapplicable
-    [Scotland, Wales, NorthernIreland]
+    [England, Scotland, Wales, NorthernIreland]
   end
 
   def self.find_by_name!(name)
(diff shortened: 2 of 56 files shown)


app/models/edition/national_applicability.rb
```diff
@@ -12,10 +12,16 @@ def process_associations_before_save(edition)
   included do
     has_many :nation_inapplicabilities, foreign_key: :edition_id, dependent: :destroy, autosave: true
     validates_associated :nation_inapplicabilities
+    alias_method :excluded_nations, :nation_inapplicabilities
+    validates :excluded_nations, length: { maximum: Nation.all.count - 1, message: 'can not exclude all nations' }
 
     add_trait Trait
   end
 
+  def has_nation_applicabilities
```
Davidslv: it sounds like a question `has_nation_applicabilities?`
Also this method seems to not be in use.

app/models/edition/national_applicability.rb
```diff
@@ -12,10 +12,16 @@ def process_associations_before_save(edition)
   included do
     has_many :nation_inapplicabilities, foreign_key: :edition_id, dependent: :destroy, autosave: true
     validates_associated :nation_inapplicabilities
+    alias_method :excluded_nations, :nation_inapplicabilities
+    validates :excluded_nations, length: { maximum: Nation.all.count - 1, message: 'can not exclude all nations' }
```
h-lame: If this validation isn't on the `nation_inapplicabilities` attribute is it going to be transferred to the view?

fofr:
What does the UI look like for this?
Can we indicate the need for this change in the PR too as well as the Trello ticket.

app/models/edition/national_applicability.rb
dwhenry: This adds an alert at the top of the page with the desired text, will add a screenshot for clarity

dwhenry: have changed this to use translations to get the wording I want

dwhenry:
@fofr update PR as requested.

PR #3767, opened 2018-02-14
Send access_limited data to Asset Manager

chrisroos:
Certain documents in Whitehall can be marked as access limited; which means that they're only accessible to users within the same organisation(s) that the document belongs to. Any attachments belonging to those documents are subject to the same restrictions in Whitehall so we need to mirror that behaviour when serving them from Asset Manager.

The changes in this branch mean that we send an array of user UIDs in the `access_limited` array when creating Whitehall assets in Asset Manager. The functionality to make use of this `access_limited` data was added to Asset Manager in https://github.com/alphagov/asset-manager/pull/471 and is based on similar functionality in the Content Store.

The access restrictions only apply to attachments (i.e. files modelled using `AttachmentData` and uploaded using the `AttachmentUploader`). Attachments can belong to an `Attachable` (one of `Edition`, `Response` or `PolicyGroup`) and the access restrictions are dependent on the type of `Attachable` they belong to:

* Attachments belonging to `Edition`s should respect the access restrictions of the `Edition`.
* Attachments belonging to `Response`s should respect the access restrictions of the `Response`'s `Edition`.
* Attachments belonging to `PolicyGroup`s should not be access limited.

It's not entirely trivial to get hold of the relevant `Attachable` object in the `AssetManagerCreateWhitehallAssetWorker#perform` method because there's no guarantee that an `Attachment` has been persisted to link the `Attachable` with the `AttachmentData` at the time the job runs. This is because the job is queued on the `after_save` of the `AttachmentData` and could feasibly run immediately. We considered failing fast in `AssetManagerCreateWhitehallAssetWorker#perform` if we determine that we don't have all the required information but that doesn't feel quite right. The approach I've gone with (after discussions with @floehopper), in the final commit of this branch, is to pass the `Attachable` object to the `AttachmentData` so that it can pass the `Attachable`'s class and ID at the point the job is queued up. This seems OK for now given that we only create attachments in the `Admin::AttachmentsController` but we should look to improve it in future given that this behaviour is somewhat surprising.

### DIFF
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -26,6 +26,9 @@ class Admin::AttachmentsController < Admin::BaseController
 
   def update
     attachment.attributes = attachment_params
+    if attachment.is_a?(FileAttachment)
+      attachment.attachment_data.attachable = attachable
+    end
     if save_attachment
       message = "Attachment '#{attachment.title}' updated"
       redirect_to attachable_attachments_path(attachable), notice: message
@@ -100,6 +103,7 @@ private
   def build_file_attachment
     FileAttachment.new(attachment_params).tap do |file_attachment|
       file_attachment.build_attachment_data unless file_attachment.attachment_data
+      file_attachment.attachment_data.attachable = attachable
     end
   end
 
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -48,6 +48,14 @@ class Attachment < ApplicationRecord
     end
   end
 
+  def attachable_is_access_limited?
+    attachable && attachable.access_limited?
+  end
+
+  def access_limited_object
+    attachable && attachable.access_limited_object
+  end
+
   def price
     if @price
       @price
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -19,18 +19,7 @@ class AttachmentData < ApplicationRecord
 
   OPENDOCUMENT_EXTENSIONS = %w(ODT ODP ODS).freeze
 
-  def attachable
-    return unless attachments.any?
-    attachments.order(:created_at).last.attachable
-  end
-
-  def attachable_is_access_limited?
-    attachable && attachable.access_limited?
-  end
-
-  def access_limited_object
-    attachable.access_limited_object
-  end
+  attr_accessor :attachable
 
   def filename
     url && File.basename(url)
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -19,6 +19,19 @@ class AttachmentData < ApplicationRecord
 
   OPENDOCUMENT_EXTENSIONS = %w(ODT ODP ODS).freeze
 
+  def attachable
+    return unless attachments.any?
+    attachments.order(:created_at).last.attachable
+  end
+
+  def attachable_is_access_limited?
+    attachable && attachable.access_limited?
+  end
+
+  def access_limited_object
+    attachable.access_limited_object
+  end
+
   def filename
     url && File.basename(url)
   end
--- a/app/models/policy_group.rb
+++ b/app/models/policy_group.rb
@@ -9,6 +9,14 @@ class PolicyGroup < ApplicationRecord
   validates_with SafeHtmlValidator
   validates_with NoFootnotesInGovspeakValidator, attribute: :description
 
+  def access_limited_object
+    nil
+  end
+
+  def access_limited?
+    false
+  end
+
   def published_policies
     Whitehall.search_client.search(
       filter_policy_groups: [slug],
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -6,9 +6,9 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
 
     if model_class && model_id
       model = model_class.constantize.find(model_id)
-      if model.respond_to?(:attachable_is_access_limited?)
-        if model.attachable_is_access_limited?
-          authorised_user_uids = AssetManagerAccessLimitation.for(model.access_limited_object)
+      if model.respond_to?(:access_limited?)
+        if model.access_limited?
+          authorised_user_uids = AssetManagerAccessLimitation.for(model)
           asset_options[:access_limited] = authorised_user_uids
         end
       end
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -1,8 +1,21 @@
 class AssetManagerCreateWhitehallAssetWorker < WorkerBase
-  def perform(file_path, legacy_url_path, draft = false)
+  def perform(file_path, legacy_url_path, draft = false, model_class = nil, model_id = nil)
     file = File.open(file_path)
     asset_options = { file: file, legacy_url_path: legacy_url_path }
     asset_options[:draft] = true if draft
+
+    if model_class && model_id
+      model = model_class.constantize.find(model_id)
+      if model.respond_to?(:attachable_is_access_limited?)
+        if model.attachable_is_access_limited?
+          object = model.access_limited_object
+          access_limited_hash = PublishingApi::PayloadBuilder::AccessLimitation.for(object)
+          authorised_user_uids = access_limited_hash[:access_limited][:users]
+          asset_options[:access_limited] = authorised_user_uids
+        end
+      end
+    end
+
     Services.asset_manager.create_whitehall_asset(asset_options)
     FileUtils.rm(file)
     FileUtils.rmdir(File.dirname(file))
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -8,9 +8,7 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
       model = model_class.constantize.find(model_id)
       if model.respond_to?(:attachable_is_access_limited?)
         if model.attachable_is_access_limited?
-          object = model.access_limited_object
-          access_limited_hash = PublishingApi::PayloadBuilder::AccessLimitation.for(object)
-          authorised_user_uids = access_limited_hash[:access_limited][:users]
+          authorised_user_uids = AssetManagerAccessLimitation.for(model.access_limited_object)
           asset_options[:access_limited] = authorised_user_uids
         end
       end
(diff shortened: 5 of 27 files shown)


test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -0,0 +1,71 @@
...
+  before do
+    login_as managing_editor
+  end
+
+  context 'given a draft document with file attachment' do
+    let(:edition) { create(:news_article, organisations: [organisation]) }
+
+    before do
+      setup_publishing_api_for(edition)
+      publishing_api_has_linkables([], document_type: "topic")
```
floehopper: Minor: I know that strictly speaking this isn't specific to an edition, but I might be tempted to move this into the `setup_publishing_api_for` method.

test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -0,0 +1,71 @@
...
+      VirusScanHelpers.simulate_virus_scan(include_versions: true)
+
+      stub_whitehall_asset('logo.png', id: 'asset-id', draft: true)
+    end
+
+    it 'marks attachment as access limited in Asset Manager when document is marked as access limited in Whitehall' do
+      visit edit_admin_news_article_path(edition)
+      check 'Limit access to producing organisations prior to publication'
+
+      Services.asset_manager.expects(:update_asset).with('asset-id', 'access_limited' => ['user-uid'])
```
floehopper: Minor: I think it might be a bit clearer if this expectation is moved to be the line before the call to drain the queue.

app/models/attachment.rb
```diff
@@ -48,6 +48,14 @@ def self.parliamentary_sessions
     end
   end
 
+  def attachable_is_access_limited?
+    attachable && attachable.access_limited?
```
floehopper: Minor: Again it feels as if there's some duplication here. Could the implementation not be `access_limited_object.present?` in all scenarios.

test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -0,0 +1,71 @@
...
+    )
+  end
+
+  def add_file_attachment(filename, to:)
+    to.attachments << FactoryBot.build(
+      :file_attachment,
+      attachable: to,
+      file: File.open(path_to_attachment(filename))
+    )
+  end
```
floehopper: Minor: I've made some changes to the other similar attachment integration tests which it would be good to adopt at some point, e.g. this method has been made redundant.

test/unit/services/service_listeners/attachment_access_limited_updater_test.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+    end
+
+    context 'when a different queue is specified' do
+      let(:queue) { 'alternative-queue' }
+      let(:worker) { stub('worker', perform_async: nil) }
+      let(:attachment) { FactoryBot.create(:file_attachment) }
+
+      it 'uses that queue' do
+        AssetManagerUpdateAssetWorker.expects(:set)
```
floehopper: Minor: I think this `expects` could be changed to be `stubs` thereby avoiding the need for the `at_least_once`.

app/controllers/admin/attachments_controller.rb
```diff
@@ -100,6 +103,7 @@ def build_external_attachment
   def build_file_attachment
     FileAttachment.new(attachment_params).tap do |file_attachment|
       file_attachment.build_attachment_data unless file_attachment.attachment_data
+      file_attachment.attachment_data.attachable = attachable
```
floehopper: I think we might need to do something similar for `Admin::BulkUploadsController` which also appears to create attachments.

chrislo:
@floehopper @chrisroos - as discussed I ran through some tests of this functionality on `integration`:

I created [a publication](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/publications/801065) with two organisations (GDS and NIHE).

I added [an attachment](https://whitehall-admin.integration.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/672653/test_pdf.pdf%0A) to it. I checked (in an \`asset-manager\` rails console on a \`backend\` machine in \`integration\`) to see if the uploaded asset was access limited

    asset = WhitehallAsset.where(legacy_url_path: '/government/uploads/system/uploads/attachment_data/file/672653/test_pdf.pdf').first
    
    irb(main):028:0> asset.access_limited.size
    => 279

The generated thumbnail was access limited in the same way

    tn = WhitehallAsset.where(legacy_url_path: '/government/uploads/system/uploads/attachment_data/file/672653/thumbnail_test_pdf.pdf.png').first
    
    tn.access_limited.size
    => 279

I then added "Cabinet Office" as an organisation to the document and saw that the `access_limited` list had more entries.

    irb(main):031:0> asset.reload.access_limited.size
    => 458

I removed the Cabinet Office and saved the document again

    asset = WhitehallAsset.where(legacy_url_path: '/government/uploads/system/uploads/attachment_data/file/672653/test_pdf.pdf').first
    asset.access_limited.size
    => 279

The `access_limited` returned to its previous size.

I then replaced the attachment PDF with [a different one](https://whitehall-admin.integration.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/672654/test_pdf_2.pdf%0A). It inherited the same access control list as its parent document (and the same as the original asset).

    asset1 = WhitehallAsset.where(legacy_url_path: '/government/uploads/system/uploads/attachment_data/file/672653/test_pdf.pdf').first
    asset2 = WhitehallAsset.where(legacy_url_path: '/government/uploads/system/uploads/attachment_data/file/672654/test_pdf_2.pdf').first
    
    asset1.access_limited.size
    => 279
    asset2.access_limited.size
    => 279

I also deleted `asset2` from the UI and confirmed that it was deleted from Asset Manager. Note, it is impossible to delete `asset1` via the UI once it has been replaced.

floehopper:
@chrislo Thanks for doing this! 👍

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -1,8 +1,21 @@
...
-  def perform(file_path, legacy_url_path, draft = false)
+  def perform(file_path, legacy_url_path, draft = false, model_class = nil, model_id = nil)
     file = File.open(file_path)
     asset_options = { file: file, legacy_url_path: legacy_url_path }
     asset_options[:draft] = true if draft
+
+    if model_class && model_id
+      model = model_class.constantize.find(model_id)
+      if model.respond_to?(:attachable_is_access_limited?)
+        if model.attachable_is_access_limited?
```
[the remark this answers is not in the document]
chrisroos: Do you mean that it'd be enough to check that the model responds to `attachable_is_access_limited?` and not worry about the value the method returns?

app/models/attachment_data.rb
```diff
@@ -19,6 +19,19 @@ class AttachmentData < ApplicationRecord
 
   OPENDOCUMENT_EXTENSIONS = %w(ODT ODP ODS).freeze
 
+  def attachable
+    return unless attachments.any?
+    attachments.order(:created_at).last.attachable
```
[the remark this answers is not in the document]
chrisroos: As I think you've seen, this implementation is changed in the final commit.

test/unit/attachment_uploader_test.rb
```diff
@@ -35,16 +35,14 @@ class AttachmentUploaderTest < ActiveSupport::TestCase
   end
 
   test "should store uploads in a directory that persists across deploys" do
-    model = stub("AR Model", id: 1)
```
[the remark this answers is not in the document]
chrisroos: Gotcha. I should've gone with "unnecessary" instead of "unused".

app/models/attachment_data.rb
```diff
@@ -19,6 +19,19 @@ class AttachmentData < ApplicationRecord
...
+  end
+
+  def attachable_is_access_limited?
```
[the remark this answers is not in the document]
chrisroos: I did consider naming it `access_limited?` but decided against it for reasons I no longer remember.

app/models/policy_group.rb
```diff
@@ -9,6 +9,14 @@ class PolicyGroup < ApplicationRecord
   validates_with SafeHtmlValidator
   validates_with NoFootnotesInGovspeakValidator, attribute: :description
 
+  def access_limited_object
+    nil
+  end
+
+  def access_limited?
+    false
+  end
```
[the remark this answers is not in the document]
chrisroos: Hmm, maybe. I'll look at that when I revisit the code.

app/controllers/admin/attachments_controller.rb
chrisroos: @danielroseman mentioned that it's possible the `BulkUploadsController` was only used for transition and is no longer necessary. I plan to investigate further.

app/models/attachment.rb
chrisroos: I'll try to remember to take a look at this when I revisit the code.

app/models/policy_group.rb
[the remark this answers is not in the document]
chrisroos: I'll try to remember to take a look at this when I revisit the code.

test/integration/attachment_access_limited_integration_test.rb
chrisroos: I'll try to remember to take a look at this when I revisit the code.

test/integration/attachment_access_limited_integration_test.rb
chrisroos: Good spot. I think this is leftover from the Sidekiq jobs running in `inline` mode which meant the expectation was satisfied at the point the 'Save' button was clicked.

test/integration/attachment_access_limited_integration_test.rb
chrisroos: Thanks for the heads up.

test/unit/services/service_listeners/attachment_access_limited_updater_test.rb
chrisroos: That makes sense. I'll try to remember to take a look at this when I revisit the code.

chrisroos:
> One more thought. Did you test whether this works when an access-limited document with an attachment is marked as not access-limited? i.e. the attachment should also be marked as not access-limited in Asset Manager. Did you consider adding an integration test for this scenario?

I think this was an oversight. I'll add an integration test for this scenario soon.

chrisroos:
Thanks for reviewing, merging and testing this all in my absence, @chrislo and @floehopper :-)

chrisroos:
>> One more thought. Did you test whether this works when an access-limited document with an attachment is marked as not access-limited? i.e. the attachment should also be marked as not access-limited in Asset Manager. Did you consider adding an integration test for this scenario?

> I think this was an oversight. I'll add an integration test for this scenario soon.

I've opened PR #3817 to add this integration test.

PR #3769, opened 2018-02-15
Update draft status of existing attachment assets

floehopper:
This is based on the changes in #3768 which keeps the draft status of assets for attachments up-to-date on an *on-going* basis, including new ones. This PR adds a Rake task to allow us to update the draft status of assets for all *existing* attachments.

Only the last two commits are unique to this branch - I plan to rebase this branch against the branch in #3768 before merging this PR.

The new `asset_manager:attachments:update_draft_status` Rake task iterates over a range of attachments and invokes `ServiceListeners::AttachmentDraftStatusUpdater#update!` to queue up jobs to update the draft status of the corresponding assets in Asset Manager. Note that I've tweaked this `AttachmentDraftStatusUpdater` class slightly to allow the Rake task to specify the (lower priority) `asset_migration` Sidekiq queue in order to avoid blocking up business-as-usual activity in the app.

Like some of the other asset migration Rake tasks, this new Rake task allows a range of attachment IDs to be specified so we can process the attachments in manageable batches.

We're probably going to need to do something similar for the following issues:

* https://github.com/alphagov/asset-manager/issues/467
* https://github.com/alphagov/asset-manager/issues/466
* https://github.com/alphagov/asset-manager/issues/469 (maybe)

I wonder what the rate-limiting step was when we migrated the attachment files, because if running the new Rake task for all attachments is going to take a similar amount of time, we might want to consider either (a) combining the work from all the above issues into a single migration job; or (b) investigating increasing the throughput of the Whitehall/Asset Manager queueing mechanism.

@chrisroos, @chrislo: Do you have any thoughts on the latter?

### DIFF
--- a/app/services/service_listeners/attachment_draft_status_updater.rb
+++ b/app/services/service_listeners/attachment_draft_status_updater.rb
@@ -1,9 +1,10 @@
 module ServiceListeners
   class AttachmentDraftStatusUpdater
-    attr_reader :attachment
+    attr_reader :attachment, :queue
 
-    def initialize(attachment)
+    def initialize(attachment, queue: nil)
       @attachment = attachment
+      @queue = queue
     end
 
     def update!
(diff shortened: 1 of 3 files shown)


floehopper:
Now that #3768 has been merged, I'm going to rebase this against `master` and force-push in preparation for merging.

PR #3781, opened 2018-02-16
Add superseeding & superseeded relationships to API

tijmenb:
This pulls in @frankieroberto's commit from https://github.com/alphagov/whitehall/pull/3775 and adds tests. New PR because we can't add to the original branch.

PR #3785, opened 2018-02-16
Expose govuk_closed_status in api

heathd:
Following on from #3781 this will give some valuable contextual
information in interpreting `superseded_organisations` and
`superseding_organisations`.

### DIFF
--- a/app/presenters/api/organisation_presenter.rb
+++ b/app/presenters/api/organisation_presenter.rb
@@ -14,6 +14,7 @@ class Api::OrganisationPresenter < Api::BasePresenter
         organisation_logo_type_class_name: model.organisation_logo_type.try(:class_name),
         closed_at: model.closed_at,
         govuk_status: model.govuk_status,
+        govuk_closed_status: model.govuk_closed_status,
         content_id: model.content_id,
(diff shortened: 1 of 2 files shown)


frankieroberto:
@heathd what's the difference between `govuk_closed_status` and the existing `govuk_status`, which can sometimes have the value `closed`, eg https://www.gov.uk/api/organisations/deputy-prime-ministers-office?

heathd:
@frankieroberto `govuk_closed_status` specifically describes the reason for the closure see eg. https://github.com/alphagov/whitehall/blob/master/app/models/organisation.rb#L157

frankieroberto:
@heathd aha! I did wonder where all the different messages on the closed org pages came from. I'd assumed it was free text.

I'd suggest renaming the field in the API to `closed_reason` or `closed_because` or similar, just to make it a bit clearer.

frankieroberto:
@heathd any chance you (or someone) could take a look at this, whenever you have a spare moment?  I'd like to do a _thing_ with the data...

frankieroberto:
@thomasleese 🎊  thank you!

PR #3792, opened 2018-02-19
Import HMCTS forms

suzannehamilton:
This picks up from the spike in #3746. It adds the remaining optional fields as well as a `DRY_RUN` parameter to let us test the data without creating any new drafts.

I haven't added tests because we're only going to run this once, and we'll review everything on integration before the final import.

https://trello.com/c/uEgpFGfV/71-2-hmcts-step-2-extend-functionality-for-bulk-import

Marked as "do not merge" until we sort out the final few TODOs.

### DIFF
--- a/lib/import/hmcts_csv_parser.rb
+++ b/lib/import/hmcts_csv_parser.rb
@@ -1,0 +1,71 @@
+module Import
+  class HmctsCsvParser
+    CSV_HEADERS = %i(
+      publication_page_id
+      artefact_code
+      category
+      artefact_file_name
+      artefact_type
+      artefact_title
+      artefact_link
+      artefact_url
+      publication_type
+      publication_title
+      publication_summary
+      publication_body
+      published_before
+      has_images
+      alternative_format_email
+      policy
+      policy_areas
+      lead_organisations
+      supporting_organisations
+      excluded_nations
+      access_limiting
+      schedule_publishing
+      attachment_accessible
+      new_publication_url
+      new_attachment_url
+    ).freeze
+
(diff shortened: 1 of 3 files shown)


lib/import/hmcts_csv_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+        end
+
+        yielder << publication_details
+      end
+    end
+
+    def self.attachment_details(row)
+      {
+        title: row[:artefact_title],
+        # TODO: Use "new artefact file name"
```
dwhenry: Is this still TODO?

lib/import/hmcts_importer.rb
```diff
@@ -0,0 +1,105 @@
...
+      HmctsCsvParser.publications(csv_path).each do |publication_data|
+        publication_ids = []
+
+        begin
+          publication = Publication.new
+          publication.publication_type = PublicationType.find_by_slug(publication_type_slug(publication_data[:publication_type]))
+          publication.title = publication_data[:title]
+          publication.summary = publication_data[:summary]
+          publication.body = publication_data[:body]
+          publication.topics = publication_data[:policy_areas].map { |policy_area| Topic.find_by(name: policy_area) }
```
dwhenry: should we be using `find_by!` heer so we get an error if the topic can't be found?

lib/import/hmcts_csv_parser.rb
suzannehamilton: Yep. We shouldn't merge this until we've handled them all. I've added a better explanation to each TODO in case someone else needs to pick up the changes.

PR #3793, opened 2018-02-20
Use Sidekiq's fake mode by default in test environment

chrisroos:
Sidekiq's [`fake` mode][1] is its default mode for testing. It appends jobs to a `jobs` array for inspection/processing instead of immediately executing the worker as with the `inline` mode. I think this is preferable because it forces us to be explicit in any tests that are relying on Sidekiq jobs to be processed. The tests requiring the Sidekiq jobs to be processed were updated in the previous commits in this branch.

The first commits in this branch explicitly enable Sidekiq's `inline` mode for those tests that rely on Sidekiq jobs being processed and the final commit contains the switch to using `fake` mode by default.

Note that I've chosen to clear all Sidekiq queues in the `teardown` block of `ActiveSupport::TestCase` in test_helper.rb. This isn't necessary for the tests to pass but the Sidekiq wiki suggests we do it to avoid jobs remaining in the queue between tests.

Note that the default mode for Cucumber tests is still `inline`.

Note that Sidekiq's `inline` mode was enabled in b3001c6e038dab9804d436a1971fcfdefb58bfd2 (PR #893) but it's not obvious why.

I think it's possible that we might see a speed improvement in the tests with this change as there'll be less code being executed.

[1]: https://github.com/mperham/sidekiq/wiki/Testing#testing-worker-queueing-fake

### DIFF
--- a/test/integration/request_tracing_test.rb
+++ b/test/integration/request_tracing_test.rb
@@ -2,10 +2,17 @@ require "test_helper"
 
 class RequestTracingTest < ActionDispatch::IntegrationTest
   setup do
+    @sidekiq_test_mode = Sidekiq::Testing.__test_mode
+    Sidekiq::Testing.inline!
+
     @govuk_request_id = "12345-67890"
     @draft_edition = create(:draft_publication)
     @presenter = PublishingApiPresenters.presenter_for(@draft_edition)
     login_as(create(:gds_admin))
+  end
+
+  teardown do
+    Sidekiq::Testing.__test_mode = @sidekiq_test_mode
   end
 
   def force_publish(edition, headers = {})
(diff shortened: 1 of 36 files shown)


test/integration/request_tracing_test.rb
```diff
@@ -2,12 +2,18 @@
...
+    Sidekiq::Testing.inline!
+
     @govuk_request_id = "12345-67890"
     @draft_edition = create(:draft_publication)
     @presenter = PublishingApiPresenters.presenter_for(@draft_edition)
     login_as(create(:gds_admin))
   end
 
+  teardown do
+    Sidekiq::Testing.fake!
```
floehopper: Minor: It might be more robust to use the construct I used in [this commit][1] here, i.e. to restore the Sidekiq test mode to what it was before it was changed to inline. 

[1]: https://github.com/alphagov/whitehall/commit/0eada72c6c29d123df316d96808f77a60d7232cf

chrisroos: I've adopted your approach and rewritten the commit. Thanks!

chrisroos:
Thanks @floehopper. I've rebased on master and force pushed in preparation for merging.

PR #3795, opened 2018-02-21
Update redirect URL of attachment assets in Asset Manager

floehopper:
This adds a new service listener which queues jobs to set the redirect URL for attachment assets in Asset Manager when the attachments parent document is unpublished/withdrawn and similarly unset the redirect URL when the parent document is re-published/un-withdrawn.

In combination with [these changes to Asset Manager][1], this will replace the behaviour [currently implemented in `AttachmentsController#fail`][2] when we start serving Whitehall attachment assets from Asset Manager.

I've enhanced `AttachmentRedirectDueToUnpublishingIntegrationTest` which I added in [a previous pull request][3] so that it now also sets expectations for the updates which need to be sent to the Asset Manager API. For the moment, I've left the assertions for the original behaviour, because I think it makes things a bit clearer. We'll need to remove those assertions before we remove the AttachmentsController when we start serving attachment assets from Asset Manager.

Note that I've also had to make a small change to `ServiceListeners::AttachmentDraftStatusUpdater` so that assets related to an unpublished document are no longer marked as draft in Asset Manager. In Asset Manager the draft status is used to decide whether to redirect requests via the `draft-assets` host which requires authentication. The fact that `AttachmentVisibility#visible?` returns `false` for attachments associated with an unpublished document is a bit of an aberration, because the attachment URL should in some sense be "visible", i.e. it should be accessible without authentication, it's just that requests to it will be redirected to the URL of the parent document which will have a public page explaining why the document has been unpublished.

See individual commit notes for more details.

[1]: https://github.com/alphagov/asset-manager/pull/481
[2]: https://github.com/alphagov/whitehall/blob/1db688fb3d795ae0e83d5490f0cb24202cc2e132/app/controllers/attachments_controller.rb#L43-L44
[3]: https://github.com/alphagov/whitehall/pull/3794

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -41,7 +41,7 @@ private
 
   def fail
     if (edition = attachment_visibility.unpublished_edition)
-      redirect_to public_document_path(edition, id: edition.unpublishing.slug)
+      redirect_to edition.unpublishing.document_path
     elsif (replacement = attachment_data.replaced_by)
       expires_headers
       redirect_to replacement.url, status: 301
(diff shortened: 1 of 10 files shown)


chrisroos:
I'm going to review this now.

PR #3800, opened 2018-02-22
Implement asset manager storage file method

chrislo:
See: https://github.com/alphagov/asset-manager/issues/482

We want to switch the AttachmentUploader to store files exclusively in Asset Manager by using the `AssetManagerStorage` engine instead of the `AssetManagerAndQuarantinedFileStorage` engine. The former doesn't currently implement `size` which makes the switch difficult. 

This PR implements a `size` method which fetches the value from the Asset Manager API (and returns `false` in error cases). 

Note that the size displayed to the public on pages with PDF attachments ([example](https://www.gov.uk/government/publications/technology-code-of-practice)) is actually stored in the whitehall database at the time the attachment is uploaded. I think this might end up being confusing as other asset types in Whitehall don't do this - eventually it would be good to use this new method to consistently fetch the size and other metadata. I think that can wait until after we've switched over the storage engine and have been able to simplify the code.

### DIFF
--- a/lib/whitehall/asset_manager_storage.rb
+++ b/lib/whitehall/asset_manager_storage.rb
@@ -52,7 +52,7 @@ class Whitehall::AssetManagerStorage < CarrierWave::Storage::Abstract
       else
         false
       end
-    rescue StandardError => e
+    rescue GdsApi::BaseError => e
       GovukError.notify(e)
       false
     end
--- a/lib/whitehall/asset_manager_storage.rb
+++ b/lib/whitehall/asset_manager_storage.rb
@@ -44,5 +44,17 @@ class Whitehall::AssetManagerStorage < CarrierWave::Storage::Abstract
     def content_type
       MIME::Types.type_for(filename).first.to_s
     end
+
+    def size
+      response = Services.asset_manager.whitehall_asset(path)
+      if response.has_key?('size') && !response['size'].nil?
+        Integer(response['size'])
+      else
+        false
+      end
+    rescue StandardError => e
+      GovukError.notify(e)
+      false
+    end
   end
 end
--- a/lib/whitehall/asset_manager_storage.rb
+++ b/lib/whitehall/asset_manager_storage.rb
@@ -47,14 +47,10 @@ class Whitehall::AssetManagerStorage < CarrierWave::Storage::Abstract
 
     def size
       response = Services.asset_manager.whitehall_asset(path)
-      if response.has_key?('size') && !response['size'].nil?
-        Integer(response['size'])
-      else
-        false
-      end
+      response['size'].to_i
     rescue GdsApi::BaseError => e
       GovukError.notify(e)
-      false
+      0
     end
   end
 end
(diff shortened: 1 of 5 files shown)


lib/whitehall/asset_manager_storage.rb
```diff
@@ -44,5 +44,17 @@ def path
...
     end
+
+    def size
+      response = Services.asset_manager.whitehall_asset(path)
+      if response.has_key?('size') && !response['size'].nil?
+        Integer(response['size'])
+      else
+        false
+      end
+    rescue StandardError => e
```
floehopper: Rescuing `StandardError` feels a bit dangerous. Would it be possible to be a bit more specific?

chrislo: Would rescuing `GdsApi` exceptions be better do you think?

floehopper: Yes, I think so. Is there a single base class exception you can rescue, or failing that a list of specific `GdsApi` exceptions...?

floehopper: When I first saw this, I wondered whether you were also trying to rescue `TypeError` exceptions from`Integer(nil)`, but then I realised you're checking for that in the `if` condition.

lib/whitehall/asset_manager_storage.rb
```diff
@@ -44,5 +44,17 @@ def path
...
       MIME::Types.type_for(filename).first.to_s
     end
+
+    def size
+      response = Services.asset_manager.whitehall_asset(path)
+      if response.has_key?('size') && !response['size'].nil?
+        Integer(response['size'])
+      else
+        false
+      end
```
floehopper: I wonder whether this could all be a lot simpler if we aim to return `0` if we can't obtain the size. Then I think this whole `if`/`else` block could be replaced with `response['size'].to_i`.

chrislo: I think that's probably a much better idea.

lib/whitehall/asset_manager_storage.rb
chrislo: Yes, its various API errors I'm mostly anticipating. Plus other things that might go wrong given that the schema of the reponse isn't strictly enforced.

chrislo: I'll change this to rescue GdsApi exceptions and their subclasses. Sentry will tell us if we've missed other types.

chrisroos:
What's the effect of this change as far as a user's concerned (particularly if/when the `#size` method returns something other than the size of the file)?

chrislo:
@chrisroos as far as I can tell, there's no user-visible change as the file size itself is not displayed to admin users (and the public see the file size stored in the database when the asset is first created). But there is some validation logic in other models that checks that the file size is greater than zero, and some code inside carrierwave assumes that storage engines have a file method. I'm hopeful that we can tidy some of this up when we've switched to use AssetManagerStorage.

chrislo:
I've added 2 commits (which I intend to rebase) to address @floehopper's suggestions.

PR #3801, opened 2018-02-22
Schedules link checker job an hour earlier

elenatanasoiu:
## Background

This job has begun taking longer and longer in the last week. It starts at 2AM every morning and is supposed to end at 8:30 AM.

However, we've seen it end later and later, going from 8:30 AM to 9:45 AM. 

Relevant grafana: 
screen shot 2018-02-22 at 15 47 45

## What 

This interferes with a number of scheduled publishing jobs that are due to go out at 9:30 every morning.

As a first step to solving the increased amount of 5xx we're getting in whitehall admin, we are scheduling this job to start an hour earlier.

For: https://trello.com/c/aqGcNwFT/90-2-whitehall-link-checker-pegs-mysql-at-400-cpu-every-morning

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -15,5 +15,5 @@
   - asset_migration
 :schedule:
   check_all_organisations_links_worker:
-    cron: '0 2 * * *' # Runs 2 a.m.
+    cron: '0 1 * * *' # Runs at 1 a.m every day
     class: CheckAllOrganisationsLinksWorker


config/sidekiq.yml
```diff
@@ -15,5 +15,5 @@
   - asset_migration
 :schedule:
   check_all_organisations_links_worker:
-    cron: '0 2 * * *' # Runs 2 a.m.
+    cron: '0 1 * * *' # Runs 2 a.m.
```
h-lame: We should update the comment too.

h-lame: Or remove it.

elenatanasoiu: updated

PR #3815, opened 2018-02-27
Add some logging and timing to link checker process

h-lame:
For: https://trello.com/c/aqGcNwFT/90-2-whitehall-link-checker-pegs-mysql-at-400-cpu-every-morning

We want to get some more data about how long various parts of the link
checking process are taking, and when they are triggered.  We use statsd
timing to get the length of time, and we log some start/end messages to
get when things are happening.  It may not be enough, but hopefully it'll
tell us something.

### DIFF
--- a/app/controllers/admin/link_checker_api_controller.rb
+++ b/app/controllers/admin/link_checker_api_controller.rb
@@ -3,12 +3,15 @@ class Admin::LinkCheckerApiController < ApplicationController
   before_action :verify_signature
 
   def callback
-    LinkCheckerApiReport.transaction do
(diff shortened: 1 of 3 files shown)


h-lame:
I plan on shipping this to staging as a branch to let me run some things and get timings, not planning on merging and deploying for real.

h-lame:
Now we've merged #3832 this extra logging and new stats collection actually seems like it might be worthwhile, so I've tidied it up to use the controller/sidekiq logger instances directly instead of `Rails.logger` which should mean things get shipped to kibana from the sidekiq jobs.

PR #3817, opened 2018-02-27
Add test for unsetting access limited state of assets

chrisroos:
I should've added this in PR #3767.


chrisroos:
Thanks for reviewing, @floehopper. I've rebased on master and force pushed in preparation for merging.

PR #3819, opened 2018-02-27
Extract CsvPreviewController

chrislo:
See: https://github.com/alphagov/asset-manager/issues/495

We are planning to remove `AttachmentsController` once we are serving attachments from Asset Manager. This controller is currently also responsible for creating previews of CSVs. It will be easier to make the change if we create a new controller just for previewing CSVs. This PR extracts the `preview` action from AttachmentsController into a new CsvPreviewController. 

It has introduced a small amount of duplication between the private methods of AttachmentsController and CsvPreviewController, but since the former will be removed shortly, I'm not too concerned. 

It's probably easier to review this PR using the combined diff, but I've left the individual commits I used to gradually extract the method to make it easier to follow if required.


chrisroos:
I'm going to review this now.

app/controllers/csv_preview_controller.rb
```diff
@@ -30,18 +30,7 @@ def attachment_visible?
...
-    elsif (replacement = attachment_data.replaced_by)
-      expires_headers
-      redirect_to replacement.url, status: 301
-    elsif image? upload_path
-      redirect_to view_context.path_to_image('thumbnail-placeholder.png')
-    elsif incoming_upload_exists? upload_path
-      redirect_to_placeholder
-    else
-      render plain: "Not found", status: :not_found
-    end
```
chrisroos: It's not obvious to me that it makes sense to remove this code. Although there aren't tests exercising these paths, it seems quite likely that they're actually being used. I think we should probably add the missing tests rather than remove the functionality.

chrislo:
@chrisroos - I've reinstated the branches I removed in `CsvPreviewController#fail` (with the exception of the check for whether the CSV was an image, which was redundant). I've added tests for those branches. 

I've rebased those changes - could you take another look?

chrisroos:
I'm looking at this again now, @chrislo.

chrislo:
Thanks @chrisroos, rebasing on master before merging.

PR #3822, opened 2018-02-28
Set access_limited state when bulk uploading attachments

chrisroos:
This ensures we're setting the access_limited state in Asset Manager when Whitehall attachments are created using the bulk upload functionality. The equivalent functionality for setting it when attachments are created was in https://github.com/alphagov/whitehall/pull/3767.

### DIFF
--- a/app/controllers/admin/bulk_uploads_controller.rb
+++ b/app/controllers/admin/bulk_uploads_controller.rb
@@ -23,6 +23,9 @@ class Admin::BulkUploadsController < Admin::BaseController
   def create
     @bulk_upload = BulkUpload.new(@edition)
     @bulk_upload.attachments_attributes = create_params[:attachments_attributes]
+    @bulk_upload.attachments.each do |attachment|
+      attachment.attachment_data.attachable = @edition
+    end
     if @bulk_upload.save_attachments
       redirect_to admin_edition_attachments_path(@edition)
     else
--- a/test/integration/attachment_access_limited_integration_test.rb
+++ b/test/integration/attachment_access_limited_integration_test.rb
@@ -62,6 +62,27 @@ class AttachmentAccessLimitedIntegrationTest < ActionDispatch::IntegrationTest
       click_button 'Save'
       AssetManagerCreateWhitehallAssetWorker.drain
     end
+
+    it 'marks attachment as access limited in Asset Manager when bulk uploaded to draft document' do
+      visit admin_news_article_path(edition)
+      click_link "Modify attachments"
+      click_link "Bulk upload from Zip file"
+      attach_file 'Zip file', path_to_attachment('sample_attachment.zip')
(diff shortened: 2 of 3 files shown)


app/controllers/admin/bulk_uploads_controller.rb
```diff
@@ -23,6 +23,9 @@ def upload_zip
   def create
     @bulk_upload = BulkUpload.new(@edition)
     @bulk_upload.attachments_attributes = create_params[:attachments_attributes]
+    @bulk_upload.attachments.each do |attachment|
+      attachment.attachment_data.attachable = @edition
+    end
```
floehopper: Minor: Did you consider adding a controller test for this behaviour? Don't worry if it's hard, but it might be good if you can.

test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -62,6 +62,27 @@ class AttachmentAccessLimitedIntegrationTest < ActionDispatch::IntegrationTest
...
+      click_link "Modify attachments"
+      click_link "Bulk upload from Zip file"
+      attach_file 'Zip file', path_to_attachment('sample_attachment.zip')
+      click_button 'Upload zip'
+      fill_in 'Title', with: 'file-title'
+      click_button 'Save'
+
+      Services.asset_manager.expects(:create_whitehall_asset).with do |params|
+        params[:legacy_url_path] =~ /greenpaper\.pdf/ &&
+          params[:access_limited] == ['user-uid']
```
floehopper: Minor: I think you might be able to do something like this here:

```ruby
Services.asset_manager.expects(:create_whitehall_asset).with(has_entries(
  legacy_url_path: regexp_matches(/greenpaper\.pdf/),
  access_limited: ['user-uid'])
)
```

If something like that does work, it might display a more useful error message if the expectation is not satisfied.

app/controllers/admin/bulk_uploads_controller.rb
chrisroos: This was an oversight. I'll add a controller test.

test/integration/attachment_access_limited_integration_test.rb
chrisroos: Thanks @floehopper! I've updated the test to use `has_entries` instead of the custom block matcher. I've left it as two expectations, though, as I think that makes it clearer that we're expecting the PDF and it's thumbnail to be access limited.

chrisroos:
Thanks for the review, @floehopper. I've addressed your feedback, rebased on master and force pushed in preparation for merging.

test/integration/attachment_access_limited_integration_test.rb
floehopper: @chrisroos Cool. I wasn't imagining combining the expectations. What you've done looks right to me!

PR #2202, opened 2015-05-28
Add rel link to parent document to attachments

bradleywright:
So that the cover sheet for an attachment is easily discoverable by
crawlers and other clients, we should surface the attachment edition's
public URL via HTTP headers.

It looks like this:

    Link: <http://www.gov.uk/government/publications/example>; rel="up"

This specific header construct gotten from:

http://tools.ietf.org/html/rfc5988#section-5

We're using `up` because it:

> Refers to a parent document in a hierarchy of documents.

http://tools.ietf.org/html/rfc5988#section-6.2.2

The user need for this came out of a meeting with the British Library's archiving team, but would also be useful for debugging attachments as it's currently non-trivial to reverse the parent document from a given attachment.

PR #3825, opened 2018-03-01
Set parent_document_url when updating attachments

chrisroos:
Send parent_document_url to AM on publication of docs

A `Link` header, containing the URL of an attachment's parent document,
was added to Whitehall attachments in
https://github.com/alphagov/whitehall/pull/2202, so that:

> the cover sheet for an attachment is easily discoverable by crawlers and
> other clients, we should surface the attachment edition's public URL via
> HTTP headers.

We need to send the value of the attachment's parent URL to to Asset
Manager so that we can continue setting the `Link` header once we're
serving Whitehall attachments from Asset Manager.

My implementation differs slightly from that in Whitehall. In Whitehall
the `Link` header will be set irrespective of whether a document (and
therefore its attachments) is draft or public. My interpretation of PR
 #2202 is that it's unnecessary to set it for draft documents as they're
not publicly accessible to clients such as web crawlers. As such, I've
decided to send the `parent_document_url` to Asset Manager at the time a
document is published in Whitehall.

It's not obvious that it's possible to change the URL of a published
document but if it is, and if the change in URL doesn't result in a
redirect from the old to the new, then we might need to add something to
handle that.

### DIFF
--- a/app/services/service_listeners/attachment_link_header_updater.rb
+++ b/app/services/service_listeners/attachment_link_header_updater.rb
@@ -1,0 +1,40 @@
+module ServiceListeners
+  class AttachmentLinkHeaderUpdater
+    attr_reader :attachment, :queue
+
+    def initialize(attachment, queue: nil)
+      @attachment = attachment
+      @queue = queue
+    end
+
+    def update!
+      return unless attachment.file?
+      attachment_data = attachment.attachment_data
+      return unless attachment_data.present?
+      return unless attachment.attachable.is_a?(Edition)
+
+      parent_document_url = Whitehall.url_maker.public_document_url(attachment.attachable)
+
+      enqueue_job(attachment_data.file, parent_document_url)
+      if attachment_data.pdf?
+        enqueue_job(attachment_data.file.thumbnail, parent_document_url)
+      end
+    end
+
+  private
+
+    def visibility_for(attachment_data)
+      AttachmentVisibility.new(attachment_data, _anonymous_user = nil)
+    end
+
+    def enqueue_job(uploader, parent_document_url)
+      legacy_url_path = uploader.asset_manager_path
+      worker.perform_async(legacy_url_path, parent_document_url: parent_document_url)
+    end
+
+    def worker
+      worker = AssetManagerUpdateAssetWorker
+      queue.present? ? worker.set(queue: queue) : worker
+    end
+  end
+end
--- a/config/initializers/edition_services.rb
+++ b/config/initializers/edition_services.rb
@@ -14,6 +14,11 @@ Whitehall.edition_services.tap do |coordinator|
       ServiceListeners::AttachmentRedirectUrlUpdater
         .new(attachment)
         .update!
+    end
+  end
+
+  coordinator.subscribe(/^(force_publish|publish)$/) do |_event, edition, options|
+    edition.attachables.flat_map(&:attachments).each do |attachment|
       ServiceListeners::AttachmentLinkHeaderUpdater
         .new(attachment)
         .update!
--- a/config/initializers/edition_services.rb
+++ b/config/initializers/edition_services.rb
@@ -14,6 +14,9 @@ Whitehall.edition_services.tap do |coordinator|
       ServiceListeners::AttachmentRedirectUrlUpdater
         .new(attachment)
         .update!
+      ServiceListeners::AttachmentLinkHeaderUpdater
+        .new(attachment)
+        .update!
     end
   end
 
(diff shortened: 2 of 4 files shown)


app/services/service_listeners/attachment_link_header_updater.rb
```diff
@@ -0,0 +1,46 @@
...
+      @queue = queue
+    end
+
+    def update!
+      return unless attachment.file?
+      attachment_data = attachment.attachment_data
+      return unless attachment_data.present?
+
+      # TODO: Fix the problem with this helper method appending .en to the URL
+      parent_document_url = public_document_url(attachment.attachable)
```
floehopper: Just in case you forgot about it, you might be able to use the `Whitehall::UrlMaker` here via `Whitehall.url_maker`.

chrisroos: Thanks! I remembered running into this problem but couldn't remember what the fix was.

chrisroos:
I think this is now ready for review. Can you take a look when you have a moment please, @floehopper.

app/services/service_listeners/attachment_link_header_updater.rb
```diff
@@ -0,0 +1,43 @@
+module ServiceListeners
+  class AttachmentLinkHeaderUpdater
+    include Rails.application.routes.url_helpers
+    include PublicDocumentRoutesHelper
```
floehopper: Minor: I'm not sure either of these `include` statements are needed are they? I have a feeling they've been unnecessarily copied & pasted into a number of the updater classes.

config/initializers/edition_services.rb
```diff
@@ -14,6 +14,9 @@
       ServiceListeners::AttachmentRedirectUrlUpdater
         .new(attachment)
         .update!
+      ServiceListeners::AttachmentLinkHeaderUpdater
+        .new(attachment)
+        .update!
```
floehopper: Minor: I wonder whether it might be worth being more selective about the events we do this for? However, this is potentially a problem for other updaters, so perhaps we should address it separately.

test/integration/attachment_link_header_test.rb
```diff
@@ -0,0 +1,104 @@
...
+        visit admin_news_article_path(edition)
+        force_publish_document
+
+        parent_document_url = Whitehall.url_maker.public_document_url(edition)
+
+        Services.asset_manager.expects(:update_asset)
+          .with(asset_id, 'parent_document_url' => parent_document_url)
+
+        AssetManagerUpdateAssetWorker.drain
+      end
```
floehopper: Minor: In [my most recent work on this stuff](https://github.com/alphagov/whitehall/blob/32bb13cf3ae3ae9ca71fa14bd4c620c631660419/test/integration/attachment_replacement_integration_test.rb#L30-L63), I've been moving the "when" bit of the test code into another nested context and then having two "it" tests where firstly the existing behaviour is tested (i.e. in this case requesting the asset and examining the response header) and, secondly, setting the expectation and draining the queue. The idea being to clearly illustrate the behaviour generated by `AttachmentsController#show` and that generated by calls to the Asset Manager API. It _might_ be worth considering doing the same here.

config/initializers/edition_services.rb
```diff
@@ -17,6 +17,14 @@
     end
   end
 
+  coordinator.subscribe(/^(force_publish|publish)$/) do |_event, edition, options|
+    edition.attachables.flat_map(&:attachments).each do |attachment|
+      ServiceListeners::AttachmentLinkHeaderUpdater
+        .new(attachment)
+        .update!
+    end
+  end
```
chrisroos: @floehopper: Is this what you meant by being more selective about the events we subscribe to?

app/services/service_listeners/attachment_link_header_updater.rb
chrisroos: Good spot. I've removed them.

config/initializers/edition_services.rb
floehopper: @chrisroos All I can see is this bit of the diff:

```diff
+        .new(attachment)
+        .update!
+    end
+  end
```

So I can't really tell, but I meant that the updater would only be subscribed to a subset of workflow events.

config/initializers/edition_services.rb
chrisroos: I've changed this in another commit in this branch. I'm now only subscribing to the `publish` and `force_publish` events.

test/integration/attachment_link_header_test.rb
chrisroos: I'm not going to do the same at the moment, although I can see/agree that it'd be good to make these tests consistent in future.

chrisroos:
Thanks @floehopper. I've addressed your feedback, rebased on master and merged.

PR #3826, opened 2018-03-01
Fetch CSV data for previews from public url

chrislo:
When we switch to storing all attachments exclusively in Asset Manager, the CsvPreview class will no longer be able to read CSV data from the local NFS mount to generate the previews presented to the
user[1].

Instead we can generate the CSV previews by fetching the CSV file served at the public host. Currently that will in effect by reading the file from the NFS mount, but when we do change over to serving from Asset Manager it will fetch the CSV file from there.

Some CSV file attachments are almost 200Mb[2] so it is not practical to fetch the entire file as part of the preview request. Fortunately the preview functionality is configured to present at most 1,000 rows
and 50 columns to the user.

I've configured the Range request to request the first 30,000 bytes of the file. This is somewhat arbitrary at the moment, and I could do some work to set this to a limit that ensures 1000 lines of every currently uploaded csv attachment is fetched.

[1] For example: https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/675936/Q4_2017_csv.csv/preview
[2]
https://www.gov.uk//government/uploads/system/uploads/attachment_data/file/399850/mappings-2015-01-27T10_26_13_00_00.csv/preview

### DIFF
--- a/test/unit/csv_file_from_public_host_test.rb
+++ b/test/unit/csv_file_from_public_host_test.rb
@@ -3,7 +3,7 @@ require "test_helper"
 class CsvFileFromPublicHostTest < ActiveSupport::TestCase
   def stub_csv_request(status: 206, body: '')
     stub_request(:get, "#{Whitehall.public_root}/some-path")
-      .with(headers: { 'Range' => 'bytes=0-30000' })
+      .with(headers: { 'Range' => 'bytes=0-300000' })
       .to_return(status: status, body: body)
   end
 
--- a/test/unit/csv_file_from_public_host_test.rb
+++ b/test/unit/csv_file_from_public_host_test.rb
@@ -29,4 +29,18 @@ class CsvFileFromPublicHostTest < ActiveSupport::TestCase
       assert_raises(CsvFileFromPublicHost::ConnectionError) { CsvFileFromPublicHost.new('some-path') }
     end
   end
+
+  test 'uses basic authentication if set in the environment' do
+    ENV.stubs(:[]).with('BASIC_AUTH_CREDENTIALS').returns('user:password')
+    ENV.stubs(:has_key?).with('BASIC_AUTH_CREDENTIALS').returns(true)
+    mock_response = mock('response')
+    mock_response.stubs(status: 206, body: '')
+    mock_connection = mock('connection')
+    mock_connection.stubs(get: mock_response)
+    Faraday.stubs(:new).returns(mock_connection)
+
+    mock_connection.expects(:basic_auth).at_least_once.with('user', 'password')
+
+    CsvFileFromPublicHost.new('some-path') {}
+  end
 end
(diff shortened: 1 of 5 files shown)


test/unit/csv_file_from_public_host_test.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  test '#new yields a temporary file that contains the contents of the request body' do
+    stub_csv_request(body: 'csv,file')
+
+    CsvFileFromPublicHost.new('some-path') do |file|
+      assert_equal 'csv,file', file.read
+    end
+  end
+
+  test '#new raises an exception if the request status is anything other than 206' do
```
chrisroos: Out of interest, do you still get a 206 if the file being requested is smaller than 30,000 bytes?

chrislo:
@chrisroos the CSV preview isn't available in admin, only to the public. For example, https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/675936/Q4_2017_csv.csv/preview. I'm hopeful that caching will mean we don't make too many requests.

test/unit/csv_file_from_public_host_test.rb
chrislo: Yes, it appears so. For example 

```
curl https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/675936/Q4_2017_csv.csv -H "Range: bytes=0-1000000" -v > /dev/null 
```

This file is 51798 bytes, but requesting the first 1000000 still returns a 206.

chrisroos:
> @chrisroos the CSV preview isn't available in admin, only to the public. For example, https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/675936/Q4_2017_csv.csv/preview. I'm hopeful that caching will mean we don't make too many requests.

Ah, OK. So to make sure I understand: a request for a CSV preview page will hit Whitehall, which will make a request to Asset Manager for a subset of the CSV file before sending it back to the client. The page sent back to the client will be cached by Fastly which will reduce the number of requests we're making to Asset Manager. Is that correct?

chrislo:
@chrisroos - it makes a request to the public URL for the asset which at the moment will hit Whitehall and fetch the file from NFS, but when we change over will hit Asset Manager. Does that make sense?

chrisroos:
> @chrisroos - it makes a request to the public URL for the asset which at the moment will hit Whitehall and fetch the file from NFS, but when we change over will hit Asset Manager. Does that make sense?

Yup, completely.

chrislo:
I've verified that this works on integration (I had to add a1ac03a to use basic auth in that environment). I've also set a value for MAXIMUM_RANGE_BYTES based on CSV files historically added to Whitehall. There's a small edge case remaining explained in the final commit, but I hope we can live with it to allow us to move on with migrating the assets over and come back to it shortly if we think it's necessary. 

@chrisroos - could you review when you get chance?

chrislo:
Thanks @chrisroos - as discussed, I'm going to rebase this on master and merge.

PR #3827, opened 2018-03-01
Update Asset Manager when attachment is replaced

floehopper:
The admin part of this app includes functionality to _replace_ an existing attachment. When an attachment is replaced, a request for the original attachment is redirected to the new attachment.

When we start serving Whitehall attachments from Asset Manager we'll want the attachment replacement functionality to continue to work. I added the ability to replace an asset in Asset Manager in https://github.com/alphagov/asset-manager/pull/499. This PR hooks into Whitehall at the point where an asset is replaced and queues up a job (or jobs) to update the relevant asset(s) in Asset Manager with the appropriate `replacement_id`.

I've chosen to hook into `AttachmentData#replace_with!` rather than in `AttachmentsController#update`, because it was more straightforward and I feel as if it's more robust against user interface changes.

The logic in `AttachmentVisibility` is really quite complicated and when considered in combination with the way `Attachment` instances are duplicated onto new editions, `AttachmentData` instances are shared between zero-many `Attachment` instances, and old editions are "quietly" marked as "superseded", and so I found it really hard to work out what was going on.

Consequently I'm slightly wary that `AttachmentVisibility#visible?` might return `false` for a replaced attachment in cases where you might expect it to return `true`. This would mean that `ServiceListeners::AttachmentDraftStatusUpdater#update!` might incorrectly try to change the draft status of an asset associated with a published, but replaced, attachment to `true`, which would mean that the redirect to the replacement asset would not be publicly available.

To protect against this problem, I've added some validation in Asset Manager in https://github.com/alphagov/asset-manager/pull/503. If we start seeing exceptions in Whitehall because of this we might need to revisit this PR.

I'm increasingly feeling as if we might be better merging the generic functionality in `AssetManagerUpdateAssetWorker` into the various attachment updater classes and converting those into more specialist workers, but that can wait for now.

I'm also starting to think we should've added new Asset Manager API endpoints for things like un-publishing/withdrawing an asset or, in this case, replacing an asset, rather than re-using the existing generic update endpoint, but again that can wait for now.

Note that the new updater class, `AttachmentReplacementIdUpdater`, operates on an instance of an `AttachmentData` rather than an instance of `Attachment`. This was because the former is actually what we care about in the updater and since we're hooking into `AttachmentData#replace_with!`, we already know the instance we want to operate on. We might want to consider changing all the updater classes to work the same way.

It might also make sense to combine the existing `Whitehall.attachment_notifier` with the `Whitehall.attachment_data_notifier` introduced in this PR, but that can also wait for now.

Note that I've left [the code which optimises out multiple redirects][1] in Whitehall for now. We might want to consider moving that to Asset Manager.

[1]: https://github.com/alphagov/whitehall/blob/c78b00f489276ba67411f688475329e3c4f9f449/app/models/attachment_data.rb#L101-L103

### DIFF
--- a/app/services/service_listeners/attachment_replacement_id_updater.rb
+++ b/app/services/service_listeners/attachment_replacement_id_updater.rb
@@ -18,7 +18,11 @@ module ServiceListeners
 
       enqueue_job(attachment_data.file, replacement.file)
       if attachment_data.pdf?
-        enqueue_job(attachment_data.file.thumbnail, replacement.file.thumbnail)
+        if replacement.pdf?
+          enqueue_job(attachment_data.file.thumbnail, replacement.file.thumbnail)
+        else
+          enqueue_job(attachment_data.file.thumbnail, replacement.file)
+        end
       end
     end
 
--- a/app/services/service_listeners/attachment_replacement_id_updater.rb
+++ b/app/services/service_listeners/attachment_replacement_id_updater.rb
@@ -1,0 +1,38 @@
+module ServiceListeners
+  class AttachmentReplacementIdUpdater
+    include Rails.application.routes.url_helpers
+    include PublicDocumentRoutesHelper
+
+    attr_reader :attachment_data, :queue
+
+    def initialize(attachment_data, queue: nil)
+      @attachment_data = attachment_data
+      @queue = queue
+    end
+
+    def update!
+      return unless attachment_data.present?
+
+      return unless attachment_data.replaced_by.present?
+      replacement = attachment_data.replaced_by
+
+      enqueue_job(attachment_data.file, replacement.file)
+      if attachment_data.pdf?
+        enqueue_job(attachment_data.file.thumbnail, replacement.file.thumbnail)
+      end
+    end
+
+  private
+
+    def enqueue_job(uploader, replacement_uploader)
+      legacy_url_path = uploader.asset_manager_path
+      replacement_legacy_url_path = replacement_uploader.asset_manager_path
+      worker.perform_async(legacy_url_path, replacement_legacy_url_path: replacement_legacy_url_path)
+    end
+
+    def worker
+      worker = AssetManagerUpdateAssetWorker
+      queue.present? ? worker.set(queue: queue) : worker
+    end
+  end
+end
--- a/app/workers/asset_manager_update_asset_worker.rb
+++ b/app/workers/asset_manager_update_asset_worker.rb
@@ -3,6 +3,11 @@ class AssetManagerUpdateAssetWorker < WorkerBase
 
   def perform(legacy_url_path, new_attributes = {})
     attributes = find_asset_by(legacy_url_path)
+
+    if (replacement_path = new_attributes.delete('replacement_legacy_url_path'))
+      new_attributes['replacement_id'] = find_asset_by(replacement_path)['id']
+    end
+
     keys = new_attributes.keys
     unless attributes.slice(*keys) == new_attributes.slice(*keys)
       asset_manager.update_asset(attributes['id'], new_attributes)
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -98,6 +98,7 @@ class AttachmentData < ApplicationRecord
     cant_be_replaced_by_self
     raise ActiveRecord::RecordInvalid, self if self.errors.any?
     self.update_column(:replaced_by_id, replacement.id)
+    Whitehall.attachment_data_notifier.publish('replace', self)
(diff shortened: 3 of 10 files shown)


chrisroos:
I'm going to take a look at this now.

app/services/service_listeners/attachment_replacement_id_updater.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def update!
+      return unless attachment_data.present?
+
+      return unless attachment_data.replaced_by.present?
+      replacement = attachment_data.replaced_by
+
+      enqueue_job(attachment_data.file, replacement.file)
+      if attachment_data.pdf?
+        enqueue_job(attachment_data.file.thumbnail, replacement.file.thumbnail)
```
chrisroos: Is it possible to replace a PDF with something other than a PDF? And if so, will this do something sensible?

app/workers/asset_manager_update_asset_worker.rb
```diff
@@ -3,6 +3,11 @@ class AssetManagerUpdateAssetWorker < WorkerBase
 
   def perform(legacy_url_path, new_attributes = {})
     attributes = find_asset_by(legacy_url_path)
+
+    if (replacement_path = new_attributes.delete('replacement_legacy_url_path'))
+      new_attributes['replacement_id'] = find_asset_by(replacement_path)['id']
```
chrisroos: Is it safe to assume that the replacement asset will exist in Asset Manager at this point?

floehopper: @chrisroos:

The replacement `AttachmentData` will definitely have been created in Whitehall, but you're right that there's no guarantee that the corresponding asset(s) will have been created in Asset Manager by this point. However, in this scenario, I'm confident that the update job will fail, because it will get a `404 Not Found` from Asset Manager when looking up the replacement asset ID (or possibly a `422 Unprocessable Entity` if it tries to do the update with a replacement asset ID which does not exist). The job will then be retried until the replacement asset is found and the update is executed successfully. While this isn't ideal, I think it will give the desired outcome.

Do you think this is OK? Can you think of a better way of doing it? Here are a couple of ideas:

1. Only have a single dedicated Sidekiq worker processing Asset Manager jobs so that they are always executed in the order they are queued. This doesn't feel great, because it will reduce the throughput of jobs.

2. In some ways I think the only problem with the solution in this PR is that exceptions will be reported in Sentry which are not actually real problems. Perhaps we could somehow allow a small number of retries for the update job if the asset lookup fails in very specific ways _without raising an exception_...?

What do you think?

chrisroos: I think that having the job fail and retry is absolutely fine - I don't think we should do anything more than that at the moment. I guess the main concern would be that the job succeeded but didn't actually set the replacement as expected.

app/services/service_listeners/attachment_replacement_id_updater.rb
floehopper: @chrisroos:

It does seem to be possible to to replace a PDF with a non-PDF. I've just tried this in integration and the original PDF thumbnail gets redirected to the replacement attachment URL (not its thumbnail, which doesn't exist anyway).

I think this is because once the PDF attachment has been replaced, and a request is made for the thumbnail, `AttachmentsController#show` will lookup the `AttachmentData` for the original PDF (which is the same for its thumbnail) and within `AttachmentsController#fail` it will redirect to the replacement attachment URL (not it's non-existent thumbnail). I suspect this scenario was not contemplated when this functionality was added, because the behaviour doesn't seem very sensible!

I'm now going to investigate what happens in this scenario with my changes in place.

Note that in doing the testing above in integration, I can see that the host is being set correctly (to `www-origin`) for the redirect - I had previously seen this being set to `whitehall-admin` vs `www` on my dev VM. So I assume that is only a problem on the dev VM.

floehopper: @chrisroos I've addressed this in a new commit: 385d7bdebe4c8558569b813281daa185525799c4.

app/workers/asset_manager_update_asset_worker.rb
floehopper: @chrisroos Cool. Thanks. 👍

floehopper: > I guess the main concern would be that the job succeeded but didn't actually set the replacement as expected.

I might be missing something, but I can't see how this could happen. On that basis, I'm happy to merge this PR. @chrisroos please say if you think I've missed something.

floehopper:
Rebasing against `master` and force-pushing in preparation for merging.

PR #3829, opened 2018-03-02
 Only show taxons with flag visible_to_departmental_editors

koetsier:
This change will show draft and published taxons in the 'edit editions tags'
interface, only when the 'visible_to_departmental_editors' flag has been set
in the level one taxons.

Previously all published taxons, regardless of the 'visible_to_departmental_editors' 
flag would be show.

Trello: https://trello.com/c/pzbqjfTU/159-spike-investigate-impact-on-education-world-and-transport-tagging-interfaces-in-whitehall

### DIFF
--- a/test/unit/taxonomy/publishing_api_adapter_test.rb
+++ b/test/unit/taxonomy/publishing_api_adapter_test.rb
@@ -6,72 +6,72 @@ class Taxonomy::PublishingApiAdapterTest < ActiveSupport::TestCase
   end
 
   test "#published_taxon_data" do
-    setup_published_taxons([published_taxon])
+    setup_taxons([visible_published_taxon, not_visible_published_taxon], with_drafts: false)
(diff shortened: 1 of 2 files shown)


test/unit/taxonomy/publishing_api_adapter_test.rb
```diff
@@ -6,72 +6,66 @@ def subject
...
     end
   end
 
-  def published_taxon
-    { "content_id" => "published" }
+  def visible_published_taxon
+    { "content_id" => "published_visible",
+      "details" => {
+        "visible_to_departmental_editors" => true
+      } }
```
cbaines: I'd prefer if the indentation of this was like other parts of the file, even if just for consistency https://github.com/alphagov/whitehall/blob/e0328f2497a59d8b0e8b4ad7a5fb4e21c1ca880c/test/unit/taxonomy/publishing_api_adapter_test.rb#L50-L55

PR #3833, opened 2018-03-04
Update Ruby to 2.4.4 patch

nicolasleger:


cbaines:
I think this would be good, but unfortunately with the current approach for installing Ruby versions, I don't think 2.4.3 is an available Ruby version where Whitehall is currently deployed for running on GOV.UK. I think the available versions are listed here [1], but I'm not quite sure.

1: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_rbenv/manifests/all.pp

cbaines:
Looks like as [1] has been merged, 2.4.4 is now available. If you're still interested in this change, then updating to 2.4.4 might not be possible?

1: https://github.com/alphagov/govuk-puppet/pull/7419

cbaines:
I've pushed your branch to this repository to allow the automated checks to run. Unfortunately, I think the Publishing E2E Tests check might fail, as there is a check in place to stop the Dockerfile getting out of sync with this `.ruby-version` file [1].

1: https://github.com/alphagov/whitehall/blob/master/Dockerfile#L1

nicolasleger:
Fixed but the `ruby:2.4.4` Docker image is not yet available.

cbaines:
This seems to pass the checks now.

I think the only remaining thing to do is tweak the commit messages, and do something about the commits changing the `.ruby-version` file. There isn't any value in keeping the commit upgrading to 2.4.3, so it would be good to remove that.

For the commit changing the Dockerfile, it would be good if the Dockerfile was mentioned in the commit message, at the moment it's just "Use patched Ruby 2.4.4" which doesn't really describe what is being changed.

nicolasleger:
Better to redo the change from scratch in a new branch if you want.

cbaines:
> Better to redo the change from scratch in a new branch if you want.

I don't mind either way, but feel free to replace the commits in this branch.

thomasleese:
Thank you for your contribution. We've recently upgraded all our apps to Ruby 2.4.4 due to potential security vulnerabilities and due to the timely nature around that I've decided to go ahead in #3952, so I will close this now.

PR #3835, opened 2018-03-05
Do not destroy instances of AttachmentData for attachments marked as deleted

floehopper:
There are two ways that `AttachmentsController#show` can respond with a `404 Not Found` for a deleted attachment:

A. If the `AttachmentData` does not exist the call to `AttachmentData.find` raises an `ActiveRecord::RecordNotFound` exception which is [translated into a `404 Not Found` in production][1].

B. If the `Attachment` is marked as deleted, its edition is not included in [`AttachmentVisibility#edition_ids`][2] and thus `AttachmentVisibility#visible?` will return `false` and [this line][3] in `AttachmentsController#fail` will be invoked.

Previously, for scenarios involving deleted attachments†, the application was only relying on mechanism A when a single attachment was deleted from a document which had never been published. Even in the very similar scenario when a document which has never been published is discarded, the AttachmentData instances associated with its attachments are not destroyed.

The changes in this commit mean that the application now relies on mechanism B for *all* scenarios involving deleted attachments, i.e. it makes things more consistent.

I did consider using mechanism A in all scenarios, but this would've meant a lot more changes and it would mean we were potentially deleting useful data.

In some ways this is a step backwards in terms of marking attachment assets as deleted in Asset Manager, but there was already work to do on that front and now it can start from a more consistent starting point.

† Mechanism A is still used if the `AttachmentData` ID in the URL path does not exist.

[1]:
http://guides.rubyonrails.org/v5.0.6/action_controller_overview.html#rescue
[2]:
https://github.com/alphagov/whitehall/blob/32bb13cf3ae3ae9ca71fa14bd4c620c631660419/app/models/attachment_visibility.rb#L116-L118
[3]:
https://github.com/alphagov/whitehall/blob/32bb13cf3ae3ae9ca71fa14bd4c620c631660419/app/controllers/attachments_controller.rb#L31

### DIFF
--- a/app/models/attachable.rb
+++ b/app/models/attachable.rb
@@ -124,7 +124,7 @@ module Attachable
   end
 
   def delete_all_attachments
-    attachments.each { |attachment| attachment.update(deleted: true) }
+    attachments.each(&:destroy)
   end
 
   def reorder_attachments(ordered_attachment_ids)
--- a/app/models/file_attachment.rb
+++ b/app/models/file_attachment.rb
@@ -6,8 +6,6 @@ class FileAttachment < Attachment
     :number_of_pages, :file, :filename, :filename_without_extension, :virus_status,
     to: :attachment_data
 
-  after_destroy :destroy_unused_attachment_data
-
   accepts_nested_attributes_for :attachment_data, reject_if: ->(attributes) { attributes[:file].blank? && attributes[:file_cache].blank? }
 
   validate :filename_is_unique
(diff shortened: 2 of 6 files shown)


floehopper:
@chrislo 

Thanks for reviewing.

> Does the oversight you mentioned in the second commit mean that there are assets in Whitehall / Asset Manager that should have been deleted but have not been? Do you think we should do anything about that (not as part of this PR, but later)?

I assume you are referring to this comment in that commit note:

> So although requests for the attachment URL will result in a `404 Not  Found`, because of the logic in `AttachmentVisibility#edition_ids` and `AttachmentsController#fail` mentioned in the previous commit, the corresponding asset in Asset Manager will not be deleted.

I plan to address this in https://github.com/alphagov/asset-manager/issues/509 and https://github.com/alphagov/asset-manager/issues/469. The idea of this commit is to make those changes easier, because they'll be able to start from a consistent set of behaviour. My plan is *not* to rely on the `AttachmentData` being destroyed in order to delete corresponding assets in Asset Manager, but instead to hook in to the relevant model/workflow lifecycle events and do queries involving the `Attachment#deleted` flag to determine whether the corresponding asset in Asset Manager should be deleted.

I hope that makes sense. I'm going to merge this now.

PR #3838, opened 2018-03-07
Handle CSV file encoding correctly when fetching from the public URL

boffbowsh:
We need to detect the encoding of the CSV before trying to write it to
a temporary file, so that we open the file with the correct encoding
mode.

The detection code here is borrowed from the `CsvPreview` class.


chrislo:
This makes sense to me @boffbowsh, thank you for fixing. To avoid confusion you may want to remove the reference to S3 from this commit - at the moment in production the CSV files are being fetched from Whitehall's NFS via "the front door" URL. The idea is that this won't have to change when we *do* switch over to Asset Manager / S3.

PR #3840, opened 2018-03-09
Respond with 404 Not Found for attachments on an edition which is both unpublished & deleted

floehopper:
Previously a request for an attachment on an edition which was both unpublished & deleted resulted in a redirect to the document page, which then responded with a `404 Not Found`.

This PR changes the behaviour so the attachment request itself responds with `404 Not Found`. I think this is more consistent and easier to understand. It should also help with some refactoring of `AttachmentVisibility` that I'm doing separately.

### DIFF
--- a/app/models/attachment_visibility.rb
+++ b/app/models/attachment_visibility.rb
@@ -43,7 +43,7 @@ class AttachmentVisibility
   def unpublished_edition
     attachable_ids = edition_ids + consultation_ids
     if (unpublishing = Unpublishing.where(edition_id: attachable_ids).first)
-      Edition.unscoped.find(unpublishing.edition_id)
+      Edition.find_by(id: unpublishing.edition_id)
     end
   end
 
--- a/app/models/unpublishing.rb
+++ b/app/models/unpublishing.rb
@@ -42,10 +42,8 @@ class Unpublishing < ApplicationRecord
     unpublishing_reason.as_sentence
   end
 
-  # Because the edition may have been deleted, we override the slug in case it
-  # has bee pre-fixed with 'deleted-'
   def document_path
-    Whitehall.url_maker.public_document_path(edition, id: slug)
+    Whitehall.url_maker.public_document_path(edition)
   end
 
   def document_url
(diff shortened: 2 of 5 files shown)


floehopper:
I've realised there was a slight flaw in the original version of this branch. I've addressed that concern and am about to force-push the latest changes.

floehopper:
@chrisroos @chrislo This is now ready for review again - sorry for the delay!

app/models/attachment_visibility.rb
```diff
@@ -43,7 +43,7 @@ def visible?
   def unpublished_edition
     attachable_ids = edition_ids + consultation_ids
     if (unpublishing = Unpublishing.where(edition_id: attachable_ids).first)
-      Edition.unscoped.find(unpublishing.edition_id)
+      Edition.find_by(id: unpublishing.edition_id)
```
chrisroos: I think this change means that `#unpublished_edition` will now return `nil` instead of raising an `ActiveRecord::RecordNotFound` exception in the case where the `unpublishing.edition_id` can't be found. Is that OK?

floehopper:
@chrisroos @chrislo Thanks. I'm going to rebase against `master` and force-push and then address your review comment and delete the test [mentioned in the comment](https://github.com/alphagov/whitehall/pull/3841#discussion_r173766237) on #3841.

app/models/attachment_visibility.rb
floehopper: @chrisroos:

Yes, that's OK. Now the condition in [this line of `BaseAttachmentsController#fail`][1] will not be met and execution will fall through to [this line][2] thus still resulting in a `404 Not Found`. I'll add something to this effect to the relevant commit note.

[1]: https://github.com/alphagov/whitehall/blob/d509099988dd800f37f9e817eabe461e743a6ccc/app/controllers/base_attachments_controller.rb#L9
[2]: https://github.com/alphagov/whitehall/blob/d509099988dd800f37f9e817eabe461e743a6ccc/app/controllers/base_attachments_controller.rb#L19

PR #3841, opened 2018-03-09
Set asset redirect_url to the URL of the parent document

chrisroos:
Instead of a relative path to the parent document.

We need to specify the full URL so that the redirect works as expected
when the asset is requested via either draft-assets or assets-origin.

I've added the `Unpublishing#document_url` method (to complement the
`Unpublishing#document_path` method) but it's not entirely clear that
this is better than calling `Whitehall.url_maker.public_document_url`
directly from within `AttachmentRedirectUrlUpdater`.

### DIFF
--- a/test/unit/unpublishing_test.rb
+++ b/test/unit/unpublishing_test.rb
@@ -121,7 +121,7 @@ class UnpublishingTest < ActiveSupport::TestCase
     assert_equal ['must be provided when withdrawing'], unpublishing.errors[:explanation]
   end
 
-  test '#document_path returns the URL for the unpublished edition' do
+  test '#document_path returns the URL path for the unpublished edition' do
     edition = create(:detailed_guide, :draft)
     original_path = Whitehall.url_maker.public_document_path(edition)
     unpublishing = create(:unpublishing, edition: edition,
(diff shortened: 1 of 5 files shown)


test/unit/unpublishing_test.rb
```diff
@@ -145,6 +145,29 @@ class UnpublishingTest < ActiveSupport::TestCase
...
+                          unpublishing_reason: UnpublishingReason::PublishedInError)
+
+    EditionDeleter.new(edition).perform!
+    # The default scope on Edition stops deleted editions being found when an
+    # unpublishing is loaded. To trigger the bug we need to reload.
+    unpublishing.reload
+
+    assert_equal original_url, unpublishing.document_url
+  end
+
```
floehopper: I've deleted the "path" version of the test for a "deleted unpublished edition" in https://github.com/alphagov/whitehall/pull/3840, although it's not yet merged. If this is merged after that PR, it would be good to delete the "URL" version of the same test before merging this PR.

chrisroos:
Thanks for the review, @floehopper. I've rebased on master and force pushed in preparation for merging.

test/unit/unpublishing_test.rb
chrisroos: @floehopper. I've merged this PR so I'll leave it to you to decide whether to delete this test as part of #3840.

floehopper: @chrisroos I have deleted this test in 88d1a0fd8d262e738ebb41ff90c05fe9e67ab7f6.

PR #3843, opened 2018-03-09
 Split HMCTS import into two steps: doc creation and draft creation

suzannehamilton:
When doing the initial HMCTS import, just create the documents and attachments in whitehall. Split out a separate step for running the draft updater, which sends the document to the draft content store.

When the tasks were combined, it slowed down whitehall publishing because these drafts were being added to the same queue as other publications. This delayed publications on staging by up to two hours.

Having a separate draft update step lets us tune the speed at which we create the drafts, and lets us kill the job without losing or duplicating data in whitehall.

https://trello.com/c/jDx2Cvqf/113-2-hmcts-step-3-import-data


lib/import/hmcts_draft_creator.rb
```diff
@@ -0,0 +1,18 @@
...
+        publication_id = row["publication whitehall ID"]
+        publication = Publication.find_by(id: publication_id)
+
+        puts "Creating draft preview of #{publication_id}: '#{publication.slug}'"
+
+        Whitehall.edition_services.draft_updater(publication).perform!
+
+        # Prevent the HMCTS import from overloading whitehall's publishing
+        # queue, blocking other documents from being published.
+        sleep(pause_seconds)
```
elenatanasoiu: You could provide a default for "pause_seconds", but it's optional.

elenatanasoiu: If it's a one time thing, then maybe it's not worth adding a default. But if it's going to be re-used, then it might be useful to type one less thing if the pause default to x seconds.

suzannehamilton: Yeah, it's just for this HMCTS import. We'll remove the code afterwards. I also don't know what the default should be yet. :)

suzannehamilton:
I tested it on integration and it doesn't solve the issue. It's the attachment uploading that blocks publishing, so we'll probably have to throttle it.

PR #3844, opened 2018-03-11
Reduce duplication in attachment-related controllers

floehopper:
I'm planning to do some significant refactoring of the `AttachmentsController`, but I didn't want to have to duplicate my efforts in the `CsvPreviewController`. This change will mean I can do the refactoring in one place.

I've also taken the opportunity to remove an unused method from `AttachmentsController`.


floehopper:
@chrisroos @chrislo Thanks!

PR #3850, opened 2018-03-12
Implement AttachmentData methods as alternative to AttachmentVisibility methods

floehopper:
`AttachmentVisibility#visible?` conflates a number of concepts, e.g. whether the attachment is deleted, draft, unpublished or replaced. Also the tests for AttachmentVisibility#visible? are patchy and not very logically organised.

The commits in this PR test-drive a *new* implementation of the logic in `AttachmentVisibility#visible` via lower level predicate methods. The tests in `AttachmentDataVisibilityTest` were used to drive out the implementation of the new methods in `AttachmentData`.

I felt that the implementation of `AttachmentVisibility` was overly focussed on performance versus readability. I've aimed to do the reverse with the new methods on `AttachmentData`, e.g. using model associations and polymorphism in preference to more sophisticated SQL queries.

This is one stage in a bigger refactoring where we're aiming to the same methods on `AttachmentData` for the logic in `AttachmentsController` and the various attachment-related service listener classes. We're also aiming to make the logic in `AttachmentsController` more similar to the related logic in Asset Manager's media controller `download` actions.

I'm hoping that eventually we'll be able to use the lower-level (and easier to understand) `AttachmentData` predicate methods in preference to `AttachmentVisibility#visible?`.

I should add that my original motivation for doing this was because I was struggling to understand the interplay between the deleted and visible status of an attachment when trying to implement code to delete Asset Manager assets for Whitehall attachments.

### DIFF
--- a/app/models/attachment_visibility.rb
+++ b/app/models/attachment_visibility.rb
@@ -37,83 +37,30 @@ class AttachmentVisibility
   end
 
   def visible?
-    visible_edition? || visible_policy_group? || visible_consultation_response?
+    attachment_data.visible_to?(user)
   end
 
   def unpublished_edition
-    attachable_ids = edition_ids + consultation_ids
-    if (unpublishing = Unpublishing.where(edition_id: attachable_ids).first)
-      Edition.find_by(id: unpublishing.edition_id)
-    end
+    attachment_data.unpublished_edition
   end
 
   def visible_attachment
-    if visible_attachable
-      (visible_attachable.attachments & attachment_data.attachments).first
-    end
+    attachment_data.visible_attachment_for(user)
   end
 
   def visible_attachable
-    visible_edition || visible_consultation_response || visible_policy_group
+    attachment_data.visible_attachable_for(user)
   end
 
   def visible_edition
-    visible_edition_scope.last
+    attachment_data.visible_edition_for(user)
   end
 
   def visible_consultation_response
-    if visible_consultation_response?
-      Response.where(edition_id: consultation_ids).last
-    end
+    attachment_data.visible_attachable_for(user)
   end
 
   def visible_policy_group
-    visible_policy_group_scope.last
-  end
-
-private
-
-  def id
-    attachment_data.id
-  end
-
-  def visible_edition?
-    visible_edition_scope.exists?
-  end
-
-  def visible_policy_group?
-    visible_policy_group_scope.exists?
-  end
-
-  def visible_consultation_response?
-    visible_consultation_scope.exists?
-  end
-
-  def visible_edition_scope
-    if user
-      Edition.accessible_to(user).where(id: edition_ids)
-    else
-      Edition.publicly_visible.where(id: edition_ids)
-    end
-  end
-
-  def visible_consultation_scope
-    if user
-      Edition.accessible_to(user).where(id: consultation_ids)
-    else
-      Edition.publicly_visible.where(id: consultation_ids)
-    end
-  end
-
-  def visible_policy_group_scope
-    PolicyGroup.joins(:attachments).where(attachments: { attachment_data_id: id })
-  end
-
-  def consultation_ids
-    @consultation_ids ||= Response.joins(:attachments).where(attachments: { attachment_data_id: id }).pluck(:edition_id)
-  end
-
-  def edition_ids
-    @edition_ids ||= Attachment.not_deleted.where(attachment_data_id: id).where(attachable_type: 'Edition').pluck(:attachable_id)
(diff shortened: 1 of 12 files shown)


test/unit/attachment_data_visibility_test.rb
```diff
@@ -0,0 +1,643 @@
...
+          before do
+            edition.major_change_published_at = Time.zone.now
+            edition.force_publish!
+          end
+
+          it 'is not deleted' do
+            refute attachment_data.reload.deleted?
+          end
+
+          it 'is draft' do
```
chrisroos: We discussed adding some more text to explain why this is draft - because it's the original `attachment_data` that's been replaced and has never been published.

chrisroos: We talked about adding something like `refute attachment.reload.attachment_data.draft?` so that we're illustrating the difference between the original and new `AttachmentData` instances.

test/unit/attachment_data_visibility_test.rb
```diff
@@ -0,0 +1,643 @@
...
+            it 'is not accessible to anonymous user' do
+              refute attachment_data.reload.accessible_to?(anonymous_user)
+            end
+
+            it 'is accessible to user in same organisation' do
+              assert attachment_data.reload.accessible_to?(user_in_same_organisation)
+            end
+
+            it 'is accessible to user in another organisation' do
+              assert attachment_data.reload.accessible_to?(user_in_another_organisation)
```
chrisroos: It took us a while to work out why this was accessible to a user in a different organisation. It's because the `AttachmentData` has been published and is therefore not draft, which means that it can't/shouldn't be access limited.

We discussed adding some more information to the commit note to explain this.

app/models/attachment_visibility.rb
```diff
@@ -37,83 +37,30 @@ def initialize(attachment_data, user)
...
   def visible_edition
-    visible_edition_scope.last
+    attachment_data.visible_edition_for(user)
   end
 
   def visible_consultation_response
-    if visible_consultation_response?
-      Response.where(edition_id: consultation_ids).last
-    end
+    attachment_data.visible_attachable_for(user)
```
chrisroos: We discussed explaining why this is OK in the commit note, i.e. why is it OK that `#visible_attachable_for` returns a `PolicyGroup` or `Edition` here?

floehopper: I've added the following paragraph to the commit note:

> Note that `AttachmentVisibility#visible_consultation_response` is only called directly from `AttachmentVisibilityTest` and so, as a bit of a cheat, I'm calling `AttachmentData#visible_attachable_for`, even though it could conceivably return something other than a consultation response, e.g. a policy group or an edition. We should be able to remove `AttachmentVisibility#visible_consultation_response` altogether in the
near future.

test/unit/attachment_data_visibility_test.rb
floehopper: I've added the following paragraph to the commit note:

> Note that all the assertions are against the *original* `AttachmentData`, rather than the replacement `AttachmentData`. This explains why, for example, when an attachment is added to a draft edition, the attachment is replaced, and the edition is published, then the original `AttachmentData` is still draft, because it was never published; whereas the replacement `AttachmentData` is not draft.

test/unit/attachment_data_visibility_test.rb
floehopper: I've added the following paragraph to the relevant commit note:

> Note that if an attachment is added to a draft edition, the edition is published, a new draft edition is created and marked as access-limited, then the attachment is accessible to a user in another organisation, because the published edition cannot be access-limited.

floehopper:
@chrislo @chrisroos Thanks for reviewing. I've addressed your comments by updating the relevant commit notes. I'm going to merge this now.

test/unit/attachment_data_visibility_test.rb
floehopper: It turns out I forgot to push up the changes to the commit note before I merged the PR, so the above comment will have to suffice - sorry!

test/unit/attachment_data_visibility_test.rb
floehopper: It turns out I forgot to push up the changes to the commit note before I merged the PR, so the above comment will have to suffice - sorry!

app/models/attachment_visibility.rb
floehopper: It turns out I forgot to push up the changes to the commit note before I merged the PR, so the above comment will have to suffice - sorry!

PR #3851, opened 2018-03-12
Inline AttachmentVisibility methods

floehopper:
This PR is the next stage in the larger refactoring which was started in #3850.

This inlines all the methods in `AttachmentVisibility` to use the methods in `AttachmentData` directly.

It also changes the various attachment-related updaters (and one a worker) to (a) all operate on an instance of `AttachmentData` rather than `Attachment`; and (b) to use the relevant methods on `AttachmentData` to highlight how they are equivalent to logic in the attachment-related controllers, particularly that in `BaseAttachmentsController#fail`.

Finally it removes the `AttachmentVisibility` class altogether.

The (hopefully) final stage will be to refactor the attachment-related controllers to make the logic more similar to that in Asset Manager's media controller `download` actions and to use lower-level (more understandable) methods on `AttachmentData`.

This should make it clearer how we are replacing Whitehall attachment functionality with a combination of code in Whitehall to update Asset Manager and new behaviour in Asset Manager to act on the data supplied by Whitehall.


floehopper:
@chrislo @chrisroos Thanks for reviewing! I'm going to rebase against `master` and force-push in preparation for merging.

PR #3852, opened 2018-03-13
Extract AttachmentRedirectUrlUpdater logic into worker

chrislo:
This PR extracts the logic from the AttachmentRedirectUrlUpdater service listener into a new worker. The motivation for this is to ensure that the data required to do the work is fetched at the time the work is done rather than when the job is added to the queue. This is best practice in general for asynchronous jobs and it also means that we can use the worker to help migrate the metadata for existing whitehall assets to asset manager (where we anticipate adding a lot of jobs to the queue).

### DIFF
--- a/app/workers/asset_manager_attachment_redirect_url_update_worker.rb
+++ b/app/workers/asset_manager_attachment_redirect_url_update_worker.rb
@@ -1,6 +1,7 @@
 class AssetManagerAttachmentRedirectUrlUpdateWorker < WorkerBase
   def perform(attachment_data_id)
-    attachment_data = AttachmentData.find(attachment_data_id)
+    attachment_data = AttachmentData.find_by(id: attachment_data_id)
+    return unless attachment_data.present?
     redirect_url = nil
     if attachment_data.unpublished?
       redirect_url = attachment_data.unpublished_edition.unpublishing.document_url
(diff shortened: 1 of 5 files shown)


app/workers/asset_manager_attachment_redirect_url_update_worker.rb
```diff
@@ -0,0 +1,20 @@
+class AssetManagerAttachmentRedirectUrlUpdateWorker < WorkerBase
+  def perform(attachment_data_id)
+    attachment_data = AttachmentData.find(attachment_data_id)
```
floehopper: Did you make a conscious decision to use `AttachmentData.find(attachment_data_id)` vs `AttachmentData.find_by(id: attachment_data_id)` here? This looks inconsistent with [`AssetManagerAttachmentReplacementIdUpdateWorker#perform`][1].

If we leave it like this and the AttachmentData is not found, an `ActiveRecord::RecordNotFound` exception will be raised and the job will be retried.

I don't know what the right approach is, but we should standardise on a single approach for all these attachment-related workers and if we adopt the approach in this commit, the choice should be explained in the commit note.

[1]: https://github.com/alphagov/whitehall/blob/bda339b6bd2b11a1f642780c6ed2772a5d675d41/app/workers/asset_manager_attachment_replacement_id_update_worker.rb#L3

chrislo: Good point. I've addressed this in c28675113.

chrislo:
Rebasing on master in preparation for merging.

PR #3853, opened 2018-03-13
Delete assets in Asset Manager when attachments are deleted

floehopper:
This wires up an instance of `ServiceListeners::AttachmentDeleter` so that the Asset Manager assets corresponding to an `AttachmentData` are deleted via a Sidekiq worker.

This will mark the relevant assets as deleted in Asset Manager. While there is functionality in the Asset Manager API to restore a deleted asset, there doesn't appear to be any such functionality in Whitehall.

Note that the Asset Manager assets will only be deleted if no other non-deleted Attachment instances are still referencing the `AttachmentData`, i.e. where `AttachmentData#deleted?` is `true`.

Also note that it looks as if when a (consultation) `Response` or a `PolicyGroup` is destroyed, the associated attachments are not destroyed. While it's not obvious that it's possible to delete a `Response` via the user interface, it does appear to be possible to delete a `PolicyGroup`.

While it doesn't feel appropriate to address this here, I'd recommend that for consistency and completeness, both of these problems should be fixed by calling `Attachable#delete_all_attachments` from a before/after destroy callback on `Response` and `PolicyGroup`. This would mean a 'destroy' event would be published to the attachment notifier for each attachment and the relevant assets would be deleted in Asset Manager.

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -8,6 +8,7 @@ class Attachment < ApplicationRecord
   before_save :set_ordering, if: -> { ordering.blank? }
   before_save :nilify_locale_if_blank
   before_save :prevent_saving_of_abstract_base_class
+  after_destroy :publish_destroy_event
 
   VALID_COMMAND_PAPER_NUMBER_PREFIXES = ['C.', 'Cd.', 'Cmd.', 'Cmnd.', 'Cm.'].freeze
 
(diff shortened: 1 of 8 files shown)


floehopper:
@chrislo:

Thanks for reviewing.

> Perhaps we should add an issue against Whitehall for this?

I think it would be good to capture it somewhere, but I'm not sure the repo issues are monitored by the GOV.UK team. I'll investigate and try to capture the issue somewhere, but in the meantime I'm going to merge this.

PR #3854, opened 2018-03-13
Fix AttachmentData#significant_attachment

floehopper:
@kevindew pointed out that `NoMethodError` exceptions like [this one][1] were occurring in the staging environment.

I discovered that this is possible if the attachment data belongs to a deleted edition which in turn has a previous edition. I've added this scenario to `AttachmentDataVisiblityTest` and called `AttachmentData#deleted?` to trigger the exception.

I've fixed the problem by using the existing `AttachmentData#last_attachable` method which uses the Null Object pattern to avoid ever returning `nil`. In hindsight, I should probably have done this in [this earlier commit][2].

While I was at it I looked through the other similar methods on `AttachmentData` to see if I could spot any possibility of a similar problem occurring, but it all looks OK to me.

[1]: https://sentry.io/govuk/app-whitehall/issues/491623681/
[2]:
https://github.com/alphagov/whitehall/commit/2256a57c603d61a0e71d69e4e7b167dd0ad1040f

### DIFF
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -201,7 +201,7 @@ private
   end
 
   def significant_attachment
-    if attachments.one? || last_attachment.attachable.publicly_visible?
+    if attachments.one? || last_attachable.publicly_visible?
       last_attachment
     else
       penultimate_attachment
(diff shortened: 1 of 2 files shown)


floehopper:
I tested this as follows:

1. Create draft publication
2. Add attachment
3. Save publication
4. Wait for attachment virus scanning
5. Force-publish
6. View on website
7. View attachment (and save its URL)
8. Create a new draft edition
9. Save new draft edition
10. Discard draft edition
11. Visit original attachment URL (from step 7)
12. See that there is no exception

I'm happy that this means the problem is fixed.

PR #3855, opened 2018-03-13
Use Unpublishing#slug in Unpublishing#document_path

chrisroos:
This partially reverts https://github.com/alphagov/whitehall/pull/3840/commits/7ea19016cd9c1366d1e6a306e4355c8d92b15a0d (PR #3840).

After deploying PR #3840 we saw a failure in the "Whitehall assets are served" scenario of the Whitehall feature in Smokey. The test was getting a 404 instead of a 200. This is because we redirect to the `Unpublishing#document_path` and the removal of the `id` option meant that this was now redirecting to a page that didn't exist. This is demonstrated in the curl requests below:

    # Before PR #3840
    $ curl -v \
      https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/32409/11-944-higher-education-students-at-heart-of-system.pdf \
      > /dev/null
    <snipped>
    < HTTP/2 302
    < location: https://www.gov.uk/government/publications/higher-education-students-at-the-heart-of-the-system--2

    $ curl -v \
      https://www.gov.uk/government/publications/higher-education-students-at-the-heart-of-the-system--2 \
      > /dev/null
    <snipped>
    < HTTP/2 301
    < location: /government/consultations/higher-education-white-paper-students-at-the-heart-of-the-system

    $ curl -v \
      https://www.gov.uk/government/consultations/higher-education-white-paper-students-at-the-heart-of-the-system \
      > /dev/null
    <snipped>
    < HTTP/2 200

    ---

    # After PR #3840
    $ curl -v \
      https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/32409/11-944-higher-education-students-at-heart-of-system.pdf \
      > /dev/null
    <snipped>
    < HTTP/1.1 302 Found
    < Location: https://www.gov.uk/government/publications/higher-education-students-at-the-heart-of-the-system

    $ curl -v \
      https://www.gov.uk/government/publications/higher-education-students-at-the-heart-of-the-system \
      > /dev/null
    <snipped>
    < HTTP/1.1 404 Not Found

I've checked the `Unpublishing` data in integration and can see that there are 4617 Unpublishing instances that are associated with an Edition with attachments where the Unpublishing has a different slug to the slug of its parent Document.

    > Unpublishing.count
    => 28083

    > mismatch_slug_count = 0; Unpublishing.all.each { |u|
        mismatch_slug_count += 1 if (u.slug != u.edition.document.slug)
      }; mismatch_slug_count
    => 5930

    > mismatch_slug_count = 0; Unpublishing.all.each { |u|
        mismatch_slug_count += 1 if (
          u.edition.respond_to?(:attachments) &&
          u.edition.attachments.any? &&
          u.slug != u.edition.document.slug
        )
      }; mismatch_slug_count
    => 4617

I think the change highlighted by the Smokey failure might also apply to these 4617 documents which is we I've chosen to revert the behaviour for now.

As an aside: the test failure in Smokey highlighted that it was previously returning a false positive response. [Smokey PR 345][1] attempts to avoid such false positives in future.

[1]: https://github.com/alphagov/smokey/pull/345

### DIFF
--- a/app/models/unpublishing.rb
+++ b/app/models/unpublishing.rb
@@ -43,7 +43,7 @@ class Unpublishing < ApplicationRecord
   end
 
   def document_path
-    Whitehall.url_maker.public_document_path(edition)
+    Whitehall.url_maker.public_document_path(edition, id: slug)
   end
 
   def document_url
--- a/test/unit/unpublishing_test.rb
+++ b/test/unit/unpublishing_test.rb
@@ -130,6 +130,15 @@ class UnpublishingTest < ActiveSupport::TestCase
     assert_equal original_path, unpublishing.document_path
   end
 
+  test '#document_path returns the URL path using the slug from the unpublishing' do
+    edition = create(:detailed_guide, :draft)
+    unpublishing = create(:unpublishing, edition: edition,
+                          unpublishing_reason: UnpublishingReason::PublishedInError)
+    unpublishing.update_attribute(:slug, 'a-different-slug')
+
+    assert_equal '/guidance/a-different-slug', unpublishing.document_path
+  end
+
   test '#document_url returns the URL for the unpublished edition' do
     edition = create(:detailed_guide, :draft)
     original_url = Whitehall.url_maker.public_document_url(edition)
@@ -137,6 +146,15 @@ class UnpublishingTest < ActiveSupport::TestCase
                           unpublishing_reason: UnpublishingReason::PublishedInError)
 
     assert_equal original_url, unpublishing.document_url
+  end
+
+  test '#document_url returns the URL using the slug from the unpublishing' do
+    edition = create(:detailed_guide, :draft)
+    unpublishing = create(:unpublishing, edition: edition,
+                          unpublishing_reason: UnpublishingReason::PublishedInError)
+    unpublishing.update_attribute(:slug, 'a-different-slug')
+
+    assert_match '/guidance/a-different-slug', unpublishing.document_url
   end
 
   test '#translated_locales is delegated to the edition' do


chrisroos:
> I'm still puzzled as to how `Unpublishing#slug` ends up not matching the `Document#slug`. Is it actually possible to change this via the UI?

I had a bit of a click around the interface and can't see that it's possible to change the slug. I wonder whether it was possible at some point in the past, though.

chrisroos:
Thanks for reviewing, @floehopper.

h-lame:
> I'm still puzzled as to how Unpublishing#slug ends up not matching the Document#slug. Is it actually possible to change this via the UI?

I encountered this recently and the following workflow is possible:

1. create a doc with a slug & publish it
2. create a new edition and change the title & publish it
3. unpublish that edition which resets it to a draft
4. publish that reset draft

At this point you'll have a document with editions and an unpublishing that have different slugs.

The intention here is that it allows you to correct an error where you published something too early or to the incorrect slug.  As far as I can tell slug is only hardened on first publish and I think resetting to a draft after an unpublishing effectively resets you to that "hasn't been first published" state and lets you change the slug again.  At least I think so, I didn't try this on a document with many editions.

floehopper:
@h-lame:

Hmm. Interesting. Doesn't the `Unpublishing` get destroyed in step 4 on [this line][1]?

[1]: https://github.com/alphagov/whitehall/blob/eea56932e62b7d74195281576c23ba8e88f49876/app/services/edition_publisher.rb#L54

h-lame:
Oh, well, I'm not sure I checked the unpublishing was still present after step 4 in my example.  But certainly on step 3 you can have a draft with a different slug to the unpublishing.

floehopper:
> But certainly on step 3 you can have a draft with a different slug to the unpublishing.

👍

PR #3856, opened 2018-03-13
Inline attachment link header updater logic into worker

chrislo:
The motivation for this change is to avoid any issues with data or code changing between the time link updater worker is queued and when the job is executed. As well as being good practice this is especially important as we want to use this worker to perform bulk update of data in Asset Manager from Whitehall and expect the individual jobs to wait on the queue for some time.

### DIFF
--- a/test/unit/services/service_listeners/attachment_link_header_updater_test.rb
+++ b/test/unit/services/service_listeners/attachment_link_header_updater_test.rb
@@ -22,7 +22,7 @@ module ServiceListeners
     end
 
     context "when attachment doesn't belong to an edition" do
-      let(:attachment) { FileAttachment.new(attachable: PolicyGroup.new) }
+      let(:attachment) { FactoryBot.create(:file_attachment) }
 
       it 'does not update draft status of any assets' do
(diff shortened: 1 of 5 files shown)


chrisroos:
I'll review this now.

test/unit/services/service_listeners/attachment_link_header_updater_test.rb
```diff
@@ -22,7 +22,7 @@ class AttachmentLinkHeaderUpdaterTest < ActiveSupport::TestCase
     end
 
     context "when attachment doesn't belong to an edition" do
-      let(:attachment) { FileAttachment.new(attachable: PolicyGroup.new) }
+      let(:attachment) { FactoryBot.create(:file_attachment) }
```
chrisroos: I'd prefer to explicitly set the `attachable` to something other than an Edition in this block. Not doing so makes it hard to see how this is working without digging into the `file_attachment` factory.

chrislo:
Thanks @chrisroos. Rebasing on master.

PR #3857, opened 2018-03-13
Fix NoMethodError in consultation Response methods

floehopper:
I noticed that exceptions like [this one][1] were occurring on staging.

I discovered that this is possible if the attachment data belongs to a consultation response which in turn belongs to a deleted consultation which in turn has a previous edition. I've added this scenario to `AttachmentDataVisiblityTest` and called AttachmentData#deleted? to trigger the exception.

In fixing the problem, I noticed a bunch more methods on `Response` which I've added recently are susceptible to the same problem. I've driven out fixes for them by adding tests to ResponseTest rather than making `AttachmentDataVisiblityTest` even longer.

I've reused `AttachmentData::NullAttachable` to avoid duplicating the default return values.

Note that `Response#access_limited_object` is not vulnerable to the same problem, because unlike the others, it doesn't call a method on the parent consultation. The same is true of all the equivalent methods on `PolicyGroup`.

I've also moved the following Null Object classes:

* `AttachmentData::NullAttachment` -> `Attachment::Null`
* `AttachmentData::NullAttachable` -> `Attachable::Null`

[1]: https://sentry.io/govuk/app-whitehall/issues/491627163/

### DIFF
--- a/app/models/attachable.rb
+++ b/app/models/attachable.rb
@@ -26,6 +26,32 @@ module Attachable
     end
   end
 
+  class Null
+    def publicly_visible?
+      false
+    end
+
+    def accessible_to?(_user)
+      false
+    end
+
+    def access_limited?
+      false
+    end
+
+    def access_limited_object
+      nil
+    end
+
+    def unpublished?
+      false
+    end
+
+    def unpublished_edition
+      nil
+    end
+  end
+
   def attachables
     [self]
   end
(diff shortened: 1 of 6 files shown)


floehopper:
I'm going to test this in integration now.

floehopper:
I tested this in integration as follows:

1. Create draft consultation with dates which means it's immediately closed
2. Force-publish consultation
3. Create a new draft edition
4. Add a public feedback to the new edition
5. Add an attachment to the new edition
6. Save the new edition
7. Force-publish new edition
8. View consultation on website
9. View attachment in public feedback (and save its URL)
10. Create a new draft edition
11. Save new draft edition
12. Discard draft edition
13. Visit original attachment URL (from step 9)
14. See that there is no exception

I'm happy that this means the problem is fixed.

PR #3859, opened 2018-03-14
Change the Whitehall tagging interface to respect Phases

koetsier:
The Whitehall tagging interface differentiated between Published and Draft taxons, to show that some were shown on GOV.UK and some (specifically Corporate Information) were not, while allowing publishers to tag to both types

Now that all taxons have been published and we're instead using Alpha, Beta and Live, the current interface should be updated to show which Taxons are Live and which ones are in Alpha/Beta

Trello: https://trello.com/c/lGAGierw/183-change-the-whitehall-tagging-interface-to-respect-phases


koetsier:
Use: `rake taxonomy:rebuild_cache` to populate Redis with the taxonomy trees

lib/taxonomy/tree.rb
```diff
@@ -10,30 +10,33 @@ def initialize(expanded_root_taxon_hash)
       @root_taxon = build_taxon(expanded_root_taxon_hash)
       root_taxon.children = parse_taxons(
         root_taxon,
-        expanded_root_taxon_hash['expanded_links_hash']
+          expanded_root_taxon_hash['expanded_links_hash']
```
cbaines: This indentation change looks unnecessary.

lib/taxonomy/taxon.rb
```diff
@@ -3,25 +3,27 @@
...
       @base_path = base_path
+      @phase = phase
+      @visible_to_departmental_editors = visible_to_departmental_editors
       @children = []
     end
 
-    def tree
-      children.each_with_object([self]) do |child, tree|
-        tree.concat(child.tree)
+    def taxon_list
```
cbaines: Tricky naming issue, but I think this is an improvement.

lib/taxonomy/tree.rb
```diff
@@ -10,30 +10,33 @@ def initialize(expanded_root_taxon_hash)
...
     end
 
-    def child_nodes(item_hash, key)
-      item_hash
-        .fetch(key, {})
-        .fetch('child_taxons', [])
-        .sort_by { |hsh| hsh.fetch('title') }
+    def child_nodes(item_hash)
+      children = item_hash.dig('links', 'child_taxons') ||
+        item_hash.dig('expanded_links', 'child_taxons') || []
```
cbaines: It looks like the previous `key` parameter was never used, so it's probably safe to assume that only the expanded links will ever exist.

PR #3861, opened 2018-03-14
Move AttachmentAccessLimitedUpdater logic to a worker

chrislo:
The motivation for this change is to avoid any issues with data or code changing between the time link updater worker is queued and when the job is executed. As well as being good practice this is especially important as we want to use this worker to perform bulk update of data in Asset Manager from Whitehall and expect the individual jobs to wait on the queue for some time.


chrislo:
Thanks @chrisroos - rebasing on master before merging.

PR #3863, opened 2018-03-14
Fix NoMethodError in PublishingApi::PayloadBuilder::AccessLimitation.for

floehopper:
The first three commits are just improvements to the `AttachmentAccessLimitedIntegrationTest`. The significant changes are all in the final commit.

I've added a new scenario to `AttachmentAccessLimitedIntegrationTest` where an attachment is added to an outcome on an access-limited consultation triggered. This triggered a `NoMethodError` exception as described in [this comment][1]. As mentioned there, this root cause of this is similar to [an exception that's been seen in production][2], but has since been partially/accidentally fixed.

My naïve fix is to allow `PublishingApi::PayloadBuilder::AccessLimitation.for` to obtain the organisations for a (consultation) `Response` by delegating to its parent `Consultation`.

I'm really not very happy with the overall implementation of this access-limiting behaviour; I feel as if the logic is rather scattered around the codebase. And I've only made that worse with this fix. However, I'll leave that for another PR.

Note that a number of `AssetManagerCreateWhitehallAssetWorker` jobs have have failed in production, which means that assets will not have been created in Asset Manager for the relevant attachments. However, assuming the Sidekiq workers are restarted after a deployment, if any of the jobs are still on the retry queue, they should succeed after this change has been deployed. Although there's a slight question in my mind about how long the files will have hung around in the `Whitehall.asset_manager_tmp_dir` directory.

[1]:
alphagov/asset-manager#522 (comment)
[2]: https://sentry.io/govuk/app-whitehall/issues/490263226/

### DIFF
--- a/test/integration/attachment_access_limited_integration_test.rb
+++ b/test/integration/attachment_access_limited_integration_test.rb
@@ -26,15 +26,18 @@ class AttachmentAccessLimitedIntegrationTest < ActionDispatch::IntegrationTest
       stub_whitehall_asset('logo.png', id: 'asset-id', draft: true)
     end
 
-    it 'marks attachment as access limited in Asset Manager when document is marked as access limited in Whitehall' do
-      visit edit_admin_news_article_path(edition)
-      check 'Limit access to producing organisations prior to publication'
+    context 'when document is marked as access limited in Whitehall' do
+      before do
+        visit edit_admin_news_article_path(edition)
+        check 'Limit access to producing organisations prior to publication'
+        click_button 'Save'
+      end
 
-      Services.asset_manager.expects(:update_asset).with('asset-id', 'access_limited' => ['user-uid'])
+      it 'marks attachment as access limited in Asset Manager' do
+        Services.asset_manager.expects(:update_asset).with('asset-id', 'access_limited' => ['user-uid'])
 
-      click_button 'Save'
-
-      AssetManagerAttachmentAccessLimitedWorker.drain
+        AssetManagerAttachmentAccessLimitedWorker.drain
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


chrisroos:
I'll review this now.

test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -162,6 +167,7 @@ def add_file_attachment(filename, to:)
     to.attachments << FactoryBot.build(
       :file_attachment,
       attachable: to,
+      title: filename,
```
chrisroos: Should this change be in a different commit? It's currently in "Assert setup complete in AttachmentAccessLimitedIntegrationTest" but doesn't seem to fit with the other changes.

floehopper: I think it was needed by the `find('.existing-attachments a', text: 'greenpaper.pdf')` that I added, otherwise the title was set to the default for the `file_attachment` factory which was a bit confusing. I'll add something to the commit note to explain.

floehopper: It turns out it was needed in the "when attachment is replaced" scenario, because the original attachment title is used in the flash notice which I want to assert against. I've added a paragraph to the commit note to explain this.

floehopper: While I was at it, I've changed the `refute_nil` to `assert` which seems better.

test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -81,6 +83,7 @@ class AttachmentAccessLimitedIntegrationTest < ActionDispatch::IntegrationTest
         click_button 'Upload zip'
         fill_in 'Title', with: 'file-title'
         click_button 'Save'
+        refute_nil find('.existing-attachments a', text: 'greenpaper.pdf')
```
floehopper: I've changed this `refute_nil` to `assert` which seems better.

floehopper:
I've addressed @chrisroos' comment and made one other very minor change via an interactive rebase. I've also rebased against `master` and am about to force-push in preparation for merging.

PR #3868, opened 2018-03-16
Fix AttachmentData#last_attachable bug

floehopper:
This is closely related to #3866.

Previously `AttachmentData#last_attachment` was always considering attachments with no attachable. The main (or only?) scenario where an attachment has no attachable is when the attachable is a deleted edition, thanks to [the `default_scope` on `Edition::Workflow`][1].

In this case, `Attachment#attachable` returns `nil` and so, given that `AttachmentData#last_attachment` was including the attachment, `AttachmentData#last_attachable` was falling back to returning `Attachable::Null`. In most scenarios it turns out that the methods on this return the correct values, but we felt it was happening more by accident than design.

We've added a realistic† scenario to `AttachmentDataVisibilityTest` which failed with the original implementation and forced us to drive out the new implementation.

We've added unit tests to AttachmentDataTest to give more complete coverage.

Note that we still need to consider attachments with no attachable in `AttachmentData#deleted?`, because of the scenario where an edition is deleted and marks all its attachments as deleted. Hence the optional keyword argument, `include_deleted_attachables`.

† Although the scenario is realistic, the incorrect return value from `AttachmentData#access_limited?` doesn't actually matter, because in this case the attachment is published and therefore access-limiting doesn't come into play. Having said that, I'm sure there are scenarios in which it (or other predicate methods on `AttachmentData`) might have returned a more consequential incorrect value before the fix.

[1]:
https://github.com/alphagov/whitehall/blob/fc62edcd5a9b1ba8bfb22911f69f128083535127/app/models/edition/workflow.rb#L21

### DIFF
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -105,7 +105,7 @@ class AttachmentData < ApplicationRecord
   end
 
   def deleted?
-    significant_attachment.deleted?
+    significant_attachment(include_deleted_attachables: true).deleted?
   end
 
   def draft?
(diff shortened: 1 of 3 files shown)


chrisroos:
I'm going to review this now.

PR #3880, opened 2018-03-19
Ignore missing files when updating metadata in Asset Manager

chrisroos:
When testing `AssetManagerAttachmentMetadataUpdater.update_range` in integration we ran into some problems where `AttachmentData` objects didn't have a corresponding file on the file system. The lack of file meant that the asset hadn't been created in Asset Manager (because we populated Asset Manager using the files on disk) and so we saw exceptions when trying to retrieve the asset from Asset Manager (in `AssetManagerWorkerHelper#find_asset_by`).

Some of these attachments have been replaced which means that, despite not having a file on disk, requests for them are redirected to their replacement. We've identified these attachments and are creating assets for them in Asset Manager as part of https://github.com/alphagov/asset-manager/issues/525. Once these records have been created we're confident that it's safe to ignore any remaining `AttachmentData` records that don't have files.

### DIFF
--- a/test/unit/asset_manager_attachment_metadata_updater_test.rb
+++ b/test/unit/asset_manager_attachment_metadata_updater_test.rb
@@ -19,6 +19,10 @@ class AssetManagerAttachmentMetadataUpdaterTest < ActiveSupport::TestCase
   ].each do |worker|
     let(:attachment_data) { FactoryBot.create(:attachment_data) }
 
+    before do
+      VirusScanHelpers.simulate_virus_scan(attachment_data.file)
+    end
+
     test "sets the #{worker} queue to 'asset_migration'" do
       worker.expects(:set).with(queue: 'asset_migration').returns(worker)
       AssetManagerAttachmentMetadataUpdater.update_range(attachment_data.id, attachment_data.id)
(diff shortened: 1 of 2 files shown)


test/unit/asset_manager_attachment_metadata_updater_test.rb
```diff
@@ -28,5 +32,11 @@ class AssetManagerAttachmentMetadataUpdaterTest < ActiveSupport::TestCase
       worker.expects(:perform_async).with(attachment_data.id)
       AssetManagerAttachmentMetadataUpdater.update_range(attachment_data.id, attachment_data.id)
     end
+
+    test "does not queue a job on the #{worker} if the file doesn't exist" do
+      FileUtils.rm(attachment_data.reload.file.path)
```
floehopper: Minor: I've used `VirusScanHelpers.erase_test_files` to achieve something similar in other tests, which might be worth considering.

chrisroos:
> When you say "When testing this branch in integration" in the PR description and commit note, which branch are you referring to? I might be missing something, but it would be good to clarify if not.

We've updated the commit note to make it clearer that we were testing the `AssetManagerAttachmentMetadataUpdater.update_range` method.

test/unit/asset_manager_attachment_metadata_updater_test.rb
chrisroos: We've decided to leave it as it is as we think it makes it clearer that we're deleting the specific file relating to the `AttachmentData`.

PR #3881, opened 2018-03-19
Refactor attachment-related controllers

floehopper:
## Motivations

* Simplify the logic in the two attachment-related controllers to make it easier to understand.
* Make the logic in the two attachment-related controllers more like that in the media controllers in Asset Manager. 

Both of the above should make things clearer when we switch over to serving Whitehall attachments from Asset Manager.

## Approach

* Re-write both controller tests from scratch to make tests easier to understand, to improve coverage, to reduce duplication. In hindsight, I wish I'd done this more incrementally so that the diff(s) would be easier to follow.
* Refactor both controllers to use multiple guard conditions instead of the more complex nested conditions.
* Add tests to specify the order of the guard conditions (based on the original implementation) and fix the ordering of the guard conditions to pass the tests. In hindsight I wish I'd added these tests at the beginning of the PR, but doing so now would require a lot of fairly unnecessary re-work.

See commit notes for more details.


chrisroos:
I'll review this now.

PR #3901, opened 2018-03-21
Decouple AttachmentData#draft? from AttachmentData#unpublished?

floehopper:
The fact that `AttachmentData#draft?` depended on `AttachmentData#unpublished?` was based on a faulty assumption, i.e. that `AttachmentVisibility#visible?` always returned `false` if the
`AttachmentData` was unpublished. The reason it always returned `false` for an `AttachmentData` on an unpublished edition was purely because its edition was not in a publicly visible state.

Another (faulty) reason for implementing `AttachmentData#draft?` this way was so that `AssetManagerAttachmentDraftStatusUpdateWorker#perform` would only need to call this predicate method to decide whether to set the corresponding Asset Manager assets to be draft or not.

As Asset Manager currently stands, we need to ensure that an asset with a redirect URL due to a document being unpublished is still marked as *not* draft even though its corresponding `AttachmentData` in Whitehall will be draft. This is so that the redirect is visible on the live stack and thus anonymous users are redirected. However, this is really an Asset Manager concern and not anything to do with Whitehall.

So I think it's less confusing if `AttachmentData#draft?` is independent of `AttachmentData#unpublished?` and the fact that an unpublished asset in Asset Manager needs to be marked as not draft is encapsulated in `AssetManagerAttachmentDraftStatusUpdateWorker#process`. This also brings `AttachmentData#visible_to?` into line with how `AttachmentVisibility#visible?` used to work.

Note that this is just a refactoring as evidenced by no changes being necessary in the integration tests.

I can imagine it might make sense to change Asset Manager so that an unpublished asset (i.e. one with a redirect URL) is always assumed to be not draft - that way we could remove the extra condition in
`AssetManagerAttachmentDraftStatusUpdateWorker#process`. However, I'm not going to address that in this PR.

### DIFF
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -109,7 +109,6 @@ class AttachmentData < ApplicationRecord
   end
 
   def draft?
-    return false if unpublished?
     !significant_attachable.publicly_visible?
   end
 
@@ -138,7 +137,7 @@ class AttachmentData < ApplicationRecord
   end
 
   def visible_to?(user)
-    !deleted? && !unpublished? && (!draft? || (draft? && accessible_to?(user)))
+    !deleted? && (!draft? || (draft? && accessible_to?(user)))
   end
 
   def visible_attachment_for(user)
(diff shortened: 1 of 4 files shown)


floehopper:
Closing to open again on a more appropriately named branch.

floehopper:
Superseded by https://github.com/alphagov/whitehall/pull/3902.

PR #3910, opened 2018-03-22
Re-write tests for attachment-related controllers

floehopper:
This PR focusses on re-writing the tests for the two attachment-related controllers, `AttachmentsController` & `CsvPreviewController`. It also makes a small change to `CsvPreviewController` to simplify it and make it easier to test, and makes a small change to the behaviour for infected image attachments.

My main motivation for doing this is to make it easier to effect the changes described in https://github.com/alphagov/asset-manager/issues/532.

Note that I've chosen to add many of the same tests to both controller tests thus increasing duplication (hopefully only temporarily), even though the implementation avoids much of this duplication by having logic extracted into `BaseAttachmentsController`. Soon we should be able to remove `AttachmentsController` entirely and inline `BaseAttachmentsController` into `CsvPreviewController`.

I re-wrote the tests from scratch by looking at the implementation and trying to achieve as full coverage as possible. I then also checked the previous implementation of the tests to see if there was anything I had missed.

I did this for the following reasons:

* I found many of these tests hard to read
* There was a lot of duplication
* Some tests were passing partly by accident
  * In many cases `BaseAttachmentsController#attachment_visible?` was returning `false` for the wrong reason
* Many combinatorial scenarios were not covered

The fact that it's a re-write rather than incremental changes means that unfortunately the diffs won't be very useful when reviewing the changes - it would probably be better to compare the before and after versions of the files.

I've taken the approach of stubbing `BaseAttachmentsController#attachment_data` and stubbing methods on `AttachmentData` rather than setting up all the relevant models in the database. These methods are well tested elsewhere and doing it this way made it easier to have a default set of stubs for the happy path which could be altered for the non-happy path. I think this makes the tests easier to read.

This approach means that I'm stubbing the `AttachmentData` predicate methods which are used in the implementation of `AttachmentData#visible_to?`.

While the approach means that the `#setup_stubs` method is quite complicated, I think it's worthwhile. I could've avoid this complication if I could've used `Minitest::Spec::DSL` style tests with nested contexts, but unfortunately this doesn't seem to work in Rails controller tests.

I've enabled CarrierWave processing for all the tests so that I can test the attachment thumbnail functionality more realistically - it was awkward to do it for just the tests which needed it and doing so doesn't seem to slow down the test appreciably.

I've chosen not to add tests for behaviour which I believe is unintended, for example:

* A Link header is added, even when the response is a `3XX` redirect or a `404 Not Found`
* A request for an infected image results in a `302 Found` redirect to the image placeholder, rather than a `404 Not Found` which is what happens if a non-image is infected


test/functional/attachments_controller_test.rb
```diff
@@ -1,188 +1,473 @@
...
-    attachment_data = unpublished_edition.attachments.first.attachment_data
+  # Not found
 
-    get_show attachment_data
+  test 'responds with 404 Not Found if attachment data does not exist' do
+    setup_stubs
+    controller.stubs(:attachment_data).raises(ActiveRecord::RecordNotFound)
 
-    assert_redirected_to publication_url(unpublished_edition.unpublishing.slug)
+    assert_raises(ActiveRecord::RecordNotFound) { get :show, params: params }
```
chrislo: Could you use `assert_response :not_found` here too?

floehopper: Unfortunately I don't think so. The problem is that the middleware (?) that converts a `ActiveRecord::RecordNotFound` into a `404 Not Found` is not used in the controller tests.

PR #3911, opened 2018-03-22
Avoid CsvPreviewController depending on attachments being on the filesystem

floehopper:
The motivation to do this is to make it easier to switch over to use `AssetManagerStorage` in `AttachmentUploader`.

This PR does the following:

* Simplifies/improves `CsvPreviewController` and its tests
* Splits `CsvFileFromPublicHost.new` into two methods so controller has access to HTTP response from range request
* Uses HTTP response from `CsvFileFromPublicHost` to implement `CsvPreviewController#incoming_upload_exists?` and `#upload_exists?` without accessing file system

I'm not very happy with how this leaves the implementation of `CsvFileFromPublicHost`, but I think it'll do for now.

Things still left to do:

* [x] Test on development VM and/or integration
* [x] The comparison of the placeholder URL in `CsvPreviewController#incoming_upload_exists?` might be wrong because of host differences

/cc @chrisroos @chrislo

### DIFF
--- a/lib/csv_file_from_public_host.rb
+++ b/lib/csv_file_from_public_host.rb
@@ -18,9 +18,6 @@ class CsvFileFromPublicHost
   def initialize(path)
     @path = path
 
-    temp_fn = CGI.escape(@path).truncate(50)
-    temp_dir = File.join(Rails.root, 'tmp')
-
     response = connection.get(@path) do |req|
       req.headers['Range'] = "bytes=0-#{MAXIMUM_RANGE_BYTES}"
     end
@@ -31,6 +28,9 @@ class CsvFileFromPublicHost
       set_encoding!(body)
       body
     end
+
+    temp_fn = CGI.escape(@path).truncate(50)
+    temp_dir = File.join(Rails.root, 'tmp')
 
     Tempfile.create(temp_fn, temp_dir, encoding: csv_file.encoding) do |tmp_file|
       tmp_file.write(csv_file)
--- a/lib/csv_file_from_public_host.rb
+++ b/lib/csv_file_from_public_host.rb
@@ -16,6 +16,9 @@ class CsvFileFromPublicHost
   end
 
   def initialize(path)
+    connection = Faraday.new(url: Whitehall.public_root)
+    connection.basic_auth(basic_auth_user, basic_auth_password) if ENV.has_key?("BASIC_AUTH_CREDENTIALS")
+
     response = connection.get(path) do |req|
       req.headers['Range'] = "bytes=0-#{MAXIMUM_RANGE_BYTES}"
     end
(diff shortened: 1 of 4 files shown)


chrislo:
I deployed this branch to integration and visited [a csv preview](https://www-origin.integration.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/603201/Postcode_level_gas_estimates_2015_experimental.csv/preview). I was able to preview the CSV file as expected.

chrislo:
The location returned from `@csv_response.headers['Location']` for a file that is being virus-scanned is `https://www-origin.integration.publishing.service.gov.uk/government/placeholder`. The value of `placeholder_url` is `https://whitehall-admin.integration.publishing.service.gov.uk/government/placeholder` so attempting to preview a pending csv file gives a 404 instead of redirecting to the placeholder. Perhaps we can fix this by comparing the paths rather than the full host?

chrislo:
I've addressed the issue in the previous comment in 4a9434d.

test/unit/csv_file_from_public_host_test.rb
```diff
@@ -11,73 +11,114 @@ def stub_csv_request(status: 206, body: '', path: 'some-path')
...
     stub_csv_request
 
-    CsvFileFromPublicHost.new('some-path') do |file|
+    response = CsvFileFromPublicHost.csv_response('some-path')
+    CsvFileFromPublicHost.new(response) do |file|
       assert File.exist?(file.path)
     end
   end
 
-  test '#new handles a very long filename' do
```
chrisroos: I thought you were going to keep this test, @chrislo. Did you change your mind?

Irrespective of whether we keep the test or not, I think we should update the commit note (titled "Avoid dependency on attachment path for tempfile prefix") to explain that using a string, instead of truncating the attachment path, for the temporary filename also fixes the problem that was previously fixed in 18bc4340b7ce0a2fd13a48261f6575a77e11119a.

lib/csv_file_from_public_host.rb
```diff
@@ -5,58 +5,50 @@ class FileEncodingError < ::EncodingError
...
-    conn
+    connection.get(path) do |req|
+      req.headers['Range'] = "bytes=0-#{MAXIMUM_RANGE_BYTES}"
+    end
   end
 
-  def response
-    connection.get(@path) do |req|
-      req.headers['Range'] = "bytes=0-#{MAXIMUM_RANGE_BYTES}"
+  def self.csv_preview_from(response)
```
chrisroos: It's very minor but it'd be good to remove the unnecessary "self" from the title of the commit that extracts this method. The commit note is currently "Extract CsvFileFromPublicHost.self.csv_preview_from".

test/unit/csv_file_from_public_host_test.rb
chrislo: Yes, I decided to remove it as it wasn't adding much value after the change to how we generate tempfile names, and it avoided having to change it to keep it passing with the changes to how stubs work. I'll make a note of that in the commit note.

lib/csv_file_from_public_host.rb
chrislo: Done.

chrislo:
Rebasing on master and force-pushing before merging.

floehopper:
@chrislo Thanks for testing, fixing and merging this 👍

PR #3923, opened 2018-03-27
Taxonomy Hide Legacy Form

benthorner:


h-lame:
It looks like this PR and #3922 are to merge commits from master into other long lived branches.  I don't know how long you are planning to keep these other branches around, but it'll make the git graph simpler when you do merge them into master if instead of these merge PRs you rebase against master & force push.

Obviously, I don't have the full context on your other branches or work so I could be wrong.  I'm just thinking about what'll make the master branch look cleanest when you do merge.  Particularly if we need to revert any of the commits you're merging in.

benthorner:
Hi Murray,

   Sorry about that - I'm used to Github pre-populating the base/compare
branches for me and got them the wrong way round :-|

Yours, Ben

PR #3924, opened 2018-03-27
Taxonomy new to old mapping

benthorner:


app/workers/publishing_api_legacy_tags_worker.rb
```diff
@@ -0,0 +1,49 @@
+class PublishingApiLegacyTagsWorker
+  API = Services.publishing_api
+
+  def perform(edition_id, taxon_ids)
```
cbaines: I'd be more comfortable if this worker didn't require state on the queue, as this leaves the possibility open for jobs to execute out of insertion order, and for old data to be persisted. I think it might be better to lookup the taxons for the edition as part of the job, as this means that this is less of a problem?

cbaines: They are stored in the Publishing API, and I think also storing them temporarily in the Whitehall Sidekiq queue could lead to quickly changing the taxons for some content causing a race condition in updating the legacy tagging information.

Say the job fails, as the Publishing API errors. I think Sidekiq will retry this later, which makes this scenario even more likely.

benthorner: Ah yes, sorry. I see - will do.

app/workers/publishing_api_legacy_tags_worker.rb
```diff
@@ -0,0 +1,72 @@
...
+  def perform(model_id, model_class)
+    model = model_class.constantize.find(model_id)
+    previous_version, taxon_ids = get_taxon_ids(model)
+    parent_ids = taxon_ids.flat_map { |id| expand_parents(id) }
+    legacy_ids = get_legacy_links(taxon_ids + parent_ids)
+
+    links = { policy_areas: [], policies: [], topics: [] }
+    update_topics(links, legacy_ids, model)
+    update_policies(links, legacy_ids, model)
+    update_policy_areas(links, legacy_ids, model)
```
koetsier: does the model need persisting? i.e. call model.save

app/workers/publishing_api_legacy_tags_worker.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    params = { previous_version: previous_version, links: links }
+    API.patch_links(model.content_id, params)
```
koetsier: Is the patch_links necessary? Does Whitehall already synchronise with the Publishing Api?

app/workers/publishing_api_legacy_tags_worker.rb
benthorner: None of these methods actually change the model, just relationships on it. So it's not necessary to save the model.

app/workers/publishing_api_legacy_tags_worker.rb
benthorner: So here I'm being consistent with the TaxonLinkPatcher. You're probably right that saving the model would eventually trigger patch_links, but as I haven't actually changed any of the actual model fields it seems more consistent to do it manually.

app/workers/publishing_api_legacy_tags_worker.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  def update_policies(links, legacy_ids, model)
+    return unless model.respond_to?(:policy_content_ids)
+    policies = Policy.all.map(&:content_id) & legacy_ids
+    links[:policies] += policies
+
+    model.policy_content_ids = policies
+
+    parent_policies = model.policy_content_ids - policies
+    links[:policy_areas] += parent_policies
```
cbaines: So, after much faffing, I've come to the opinion that this isn't quite right.

The assignment to `model.policy_content_ids` can silently add additional `content_id`s to the associated `policy_content_ids`. However, I don't think this will happen in the live environment, and I don't think the additional `content_id`s are actually policy areas, but instead will be other policies.

Taking the last point first, policies are published from Policy Publisher, and it does set the `policy_areas` links [1]. What it sets them to are not policy areas though, but instead other policies, which it refers to as parent policies (or just policies).

In reality though, there are not any sub policies. I'm not sure if this is intentional or not... all I know is that from Jan 2016 until Jan 2018, any attempt to edit a sub policy in Policy Publisher will have caused it to become a policy, and loose the parent information [2].

So, all in all. We may as well try and still cope with this behaviour, but I think this should look more like:
```ruby
model.policy_content_ids = policies
# It is possible that policies get added on assignment, so keep this behaviour here
links[:policies] = model.policy_content_ids
```

Now, there is a separate issue around the mapping from policies to policy areas that is maintained within Whitehall, as it uses that to add policy areas through JavaScript on the edit form. I'm still undecided what to do with that...

1: https://github.com/alphagov/policy-publisher/blob/master/app/presenters/links_presenter.rb#L17
2: https://github.com/alphagov/policy-publisher/pull/214

benthorner: Yuk. When I saw "policy_areas" in related_policies.rb I assumed it meant what it said. Lesson learnt, I guess - thanks for finding that out :-)

Your suggestion make sense, plus also re-using the logic from the admin/policies_controller.rb to do the mapping from policies to policy areas.

As the links hash would then be populated just once for each kind of link, it also means we can use the LinkPresenter like in the old-to-new code.

cbaines: I've put together a patch [1] to remove the use of the policy_areas link type from Policy Publisher. Once that is gone, maybe something can be done about the Whitehall side of this functionality as well.

1: https://github.com/alphagov/policy-publisher/pull/255

app/workers/publishing_api_legacy_tags_worker.rb
koetsier: I'm actually not sure if it does This may be necessary - have to dig through the innards of whitehall

cbaines:
I'm going to close this, and the other related Pull Requests for now. But I think it would be good to keep the branches around for now.

With the shift from Q4 (2017/2018) to Q1 (2018/2019), work is still continuing on this general area, but I'm hoping that this new time makes a more ambitious approach possible, which avoids some of the complexities involved with the mappings, and gets the Taxonomy tagging interface to publishers sooner.

PR #3925, opened 2018-03-27
Taxonomy hide legacy form

benthorner:


app/assets/javascripts/templates.js
```diff
@@ -0,0 +1,4 @@
+window.templates = {};
```
cbaines: I don't think this file should be comitted...?

benthorner: Thanks for spotting. It's populated by a rake task and found its way into a commit.

cbaines:
I'm going to close this, and the other related Pull Requests for now. But I think it would be good to keep the branches around for now.

With the shift from Q4 (2017/2018) to Q1 (2018/2019), work is still continuing on this general area, but I'm hoping that this new time makes a more ambitious approach possible, which avoids some of the complexities involved with the mappings, and gets the Taxonomy tagging interface to publishers sooner.

PR #3933, opened 2018-03-28
Fix redirect when filename contains multiple periods

chrisroos:
We received numerous support requests about attachments not being accessible after deploying [PR 3918][1] earlier today.

The problem was caused by the explicit `:extension` segment in the routes. Its presence meant that filenames containing multiple periods (e.g. 'filename.foo.bar') would be parsed and result in:

* params[:path] being set to 'filename',
* params[:extension] being set to 'foo'
* params[:format] being set to 'bar'

I don't fully understand why but calling `redirect_to host: asset_host` when params contains both an `extension` and `format` results in the `format` being dropped. You can see this in the example below (note that the .pdf has been dropped):

    curl -v "https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/637057/Rural_Land_and_Entitlements__RLE1__V1.0_2016.pdf"
    < HTTP/2 302
    < location: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/637057/Rural_Land_and_Entitlements__RLE1__V1.0_2016

I came across this previously when doing something similar with the non-attachment assets. That was fixed in [PR 3627][2].

I started out trying to write a functional test for the AttachmentsController but that wasn't easy because I have to specify the parameters individually. I've gone with an integration test as it allows me to request the path using a string.

[1]: https://github.com/alphagov/whitehall/pull/3918
[2]: https://github.com/alphagov/whitehall/pull/3627/commits/98bb15f5fed3c735f4f601227d331c0b41dcd924

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -19,6 +19,10 @@ class AttachmentsController < BaseAttachmentsController
 
 private
 
+  def file_with_extensions
+    [params[:file], params[:format]].compact.join('.')
+  end
+
   def link_rel_headers
     if (edition = attachment_data.visible_edition_for(current_user))
       response.headers['Link'] = "<#{public_document_url(edition)}>; rel=\"up\""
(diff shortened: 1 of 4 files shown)


chrisroos:
I deployed this branch to integration to test the fix and am happy that it's working as expected.

I visited two of the pages that were previously broken in production (https://www.gov.uk/government/publications/rdpe-growth-programme and https://www.gov.uk/government/publications/tell-the-rural-payments-agency-about-land-changes-and-entitlement-transfers) and confirmed that the PDF thumbnails and PDFs themselves were accessible.

chrisroos:
Thanks for reviewing, @chrislo.

PR #3936, opened 2018-03-29
WIP: Avoid redirecting to asset host for signed in users

chrisroos:
The redirection added in PR 3918 didn't take draft attachments into account; in that draft attachments aren't accessible via the asset host. This means that draft attachments (e.g. attachments added to a draft consultation) aren't visible from within whitehall-admin.

Rather than check the state of the attachment I've decided to avoid redirecting if the user is signed in. This means that if you're signed in you'll continue to view all attachments on whitehall-admin or draft-origin as you would have done previously.


chrisroos:
I've force pushed in an attempt to get the end to end tests to run.

PR #3957, opened 2018-04-12
Rename social housing "Regulating the Standards" guidance page

barrucadu:
The URL `/government/publications/social-housing-regulation-regulating-the-standards` doesn't conform to the normal style.

[Trello card](https://trello.com/c/vZL7QhCS/47-change-to-a-publications-url-social-housing-2)

---

When I tested this in dev, I got the same page accessible at both URLs.  Is this just an artefact of the dev VM, or does a 301 redirect need adding somewhere as well as this?


thomasleese:
I would expect the old URL to no longer be accessible on the dev VM! In terms of setting up a redirect, you can do something similar to https://github.com/alphagov/whitehall/blob/master/db/data_migration/20171114113603_redirect_deleted_teacher_supply_stats_announcement.rb

barrucadu:
I'm also a bit confused by the old URL still being accessible, as the slug is definitely gone:

```
vagrant@development:~$ cd /var/govuk/whitehall/
vagrant@development:/var/govuk/whitehall$ bundle exec rails console
I, [2018-04-12T09:46:24.395402 #8781]  INFO -- sentry: ** [Raven] Raven 2.7.2 configured not to capture errors: DSN not set
Loading development environment (Rails 5.0.6)
[1] pry(main)> Document.find_by(slug: "social-housing-regulation-regulating-the-standards")
  Document Load (0.2ms)  SELECT  `documents`.* FROM `documents` WHERE `documents`.`slug` = 'social-housing-regulation-regulating-the-standards' LIMIT 1
=> nil
```

This is after a restart of everything (including the VM)

barrucadu:
Since we want to redirect one URL to another, rather than a content ID to a URL, I copied the approach from https://github.com/alphagov/whitehall/commit/f6088835ca3371a3850de2427d5b2a6666348ba0#diff-4beba2c08746b9f02186f396b3cc1df0

PR #3966, opened 2018-04-17
Only send international delegations to Rummager

thomasleese:
We want Rummager to index the new world location taxons so we need to stop Whitehall from sending them to Rummager as they would show as duplicates in the search results. This is because Whitehall sends documents to the 'government' index while Rummager adds the taxons to the 'govuk' index.

[Trello Card](https://trello.com/c/M5sF6vyc/59-renamed-worldwide-taxonomy-pages-not-showing-correctly-in-search)

### DIFF
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -35,7 +35,7 @@ class WorldLocation < ApplicationRecord
   searchable title: :title,
              link: :search_link,
              description: :search_description,
-             only: :active,
+             only: :active_international_delegation,
              format: 'world_location',
              slug: :slug
   include PublishesToPublishingApi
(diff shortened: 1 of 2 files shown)


app/models/world_location.rb
```diff
@@ -35,7 +35,7 @@ class WorldLocation < ApplicationRecord
   searchable title: :title,
              link: :search_link,
              description: :search_description,
-             only: :active,
+             only: :active_and_international_delegation,
```
elliotcm: This is maybe just me, but would this be better as `active_international_delegations`?  At first glance it looks like you're sending active world locations and international delegation world locations.

PR #3967, opened 2018-04-17
Ensure features are deleted with their docs

elliotcm:
It's rare to delete a Document, and presumably likewise for offsite links and topical events, but it does happen from time to time and we should ensure that these ghost docs are no longer considered featured.

Adds dependent destroy from Document, OffsiteLink, and TopicalEvent, to Feature.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -37,7 +37,7 @@ class Document < ApplicationRecord
   has_many :document_collection_group_memberships, inverse_of: :document, dependent: :delete_all
   has_many :document_collection_groups, through: :document_collection_group_memberships
   has_many :document_collections, through: :document_collection_groups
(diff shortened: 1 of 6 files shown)


elliotcm:
@thomasleese I'm not against the idea but it's not part of mysql I've used before.  Is there prior art in the current DB?

thomasleese:
Hmm, actually, it looks like Whitehall doesn't really use foreign keys https://github.com/alphagov/whitehall/blob/3ad5e0fa6ba63ce694a203b650908aa329101fa3/db/schema.rb#L1207-L1208 and it's hard to add new ones without database downtime, so probably not worth it for this PR.

elliotcm:
@thomasleese That was my assumption, but assumptions are there to be challenged :)

PR #3969, opened 2018-04-17
Improve logging in rummager Sidekiq jobs

barrucadu:
Fixes the search index & world locations jobs using the Rails logger rather than the Sidekiq logger, and logs more details about the requests sent to Rummager.

### DIFF
--- a/lib/whitehall/rummageable.rb
+++ b/lib/whitehall/rummageable.rb
@@ -69,14 +69,22 @@ module Whitehall
         end
       end
 
-      def log_request(method, url, _payload = nil)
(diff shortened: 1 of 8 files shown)


lib/whitehall/rummageable.rb
```diff
@@ -69,14 +69,22 @@ def repeatedly
         end
       end
 
-      def log_request(method, url, _payload = nil)
-        @logger.info("Rummageable request: #{method.upcase} #{url}")
+      def log_request(method, url, payload = nil)
+        log("Rummeable request", method, url, payload)
```
thomasleese: Typo, should be `Rummageable`... not that it's a real word.

PR #3970, opened 2018-04-18
Send Organisation Logo Details to Rummager

peteglondon:
We want to display search results for organisation as logos in the collections app for the Trello card [Show the department logo in the Organisations list](https://trello.com/c/wSDRcUwW/27-show-the-department-logo-in-the-organisations-list).

This commit sends the following fields to rummager to allow this.

organisation_brand
organisation_crest
logo_formatted_title

There is a related [PR in rummager](https://github.com/alphagov/rummager/pull/1228)


app/models/organisation.rb
```diff
@@ -275,6 +278,10 @@ def organisation_logo_type=(organisation_logo_type)
     self.organisation_logo_type_id = organisation_logo_type && organisation_logo_type.id
   end
 
+  def organisation_crest
+    organisation_logo_type ? organisation_logo_type.class_name : nil
```
cbaines: Would `organisation_logo_type.try(:class_name)` do the same thing here? It's also a little shorter.

PR #3980, opened 2018-04-23
Handle pdf-reader errors more gracefully

bevanloon:
Pdf-readed was bumped to version 2.1.0
(https://github.com/alphagov/whitehall/pull/3890) which has changed the
error that is surfaced when a pdf has a password and `#page_count` is
sent.

Add OpenSSL::Cipher::CipherError to the rescue clasue so we continue to
allow these pdfs.

[Trello](https://trello.com/c/f2vfdrb9/102-fix-opensslcipher-error-on-whitehall-attachment-uploads)

### DIFF
--- a/test/unit/attachment_data_test.rb
+++ b/test/unit/attachment_data_test.rb
@@ -101,10 +101,16 @@ class AttachmentDataTest < ActiveSupport::TestCase
     assert_equal 3, attachment.number_of_pages
   end
 
-  test "should save attachment even if unable to count the number of pages" do
-    greenpaper_pdf = fixture_file_upload('greenpaper.pdf')
-    AttachmentData.any_instance.stubs(:`).raises
-    assert_nothing_raised { create(:attachment_data, file: greenpaper_pdf) }
+  test "should set number of pages to nil if pdf-reader cannot count the number of pages" do
+    greenpage_pdf = fixture_file_upload('greenpaper.pdf')
+
+    errors = %w(PDF::Reader::MalformedPDFError PDF::Reader::UnsupportedFeatureError OpenSSL::Cipher::CipherError)
+    errors.each do |err|
+      PDF::Reader.any_instance.stubs(:page_count).raises(err.constantize)
+      attachment = create(:attachment_data, file: greenpage_pdf)
+      attachment.reload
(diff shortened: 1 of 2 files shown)


test/unit/attachment_data_test.rb
```diff
@@ -101,10 +101,16 @@ class AttachmentDataTest < ActiveSupport::TestCase
...
   end
 
-  test "should save attachment even if unable to count the number of pages" do
-    greenpaper_pdf = fixture_file_upload('greenpaper.pdf')
-    AttachmentData.any_instance.stubs(:`).raises
-    assert_nothing_raised { create(:attachment_data, file: greenpaper_pdf) }
+  test "should set number of pages to nil if pdf-reader cannot count the number of pages" do
+    greenpage_pdf = fixture_file_upload('greenpaper.pdf')
+
+    errors = %w(PDF::Reader::MalformedPDFError PDF::Reader::UnsupportedFeatureError OpenSSL::Cipher::CipherError)
```
barrucadu: What is the goal of this test?  That pdf-reader exceptions are handled, or that exceptions types caught in `calculate_number_of_pages` are handled?  I don't think it does the former, as if a new version of pdf-reader adds a new type of exception the test won't flag that up; and given how simple `calculate_number_of_pages` is, I'm not sure there's much benefit in a test for the latter.

bevanloon: My thought is that there might be exceptions that pdf-reader throws that we do want to completely break on, either now, or in the future (without a complete audit of pdf-reader, however....). We do know about these three exceptions that we are happy with, though, so the test basically guards against my future self deleting these by mistake.

We are, of course, working against a moving target if we bump pdf-reader when new versions are released, but there we go.

barrucadu: Ok, that's reasonable.  It's a shame that a change in the exception type wasn't flagged up as a breaking API change (I'm assuming pdf-reader is using SemVer), or even mentioned in the changelog.

bevanloon: agreed - at a guess they probably considered it a bug that they hid the hid the OpenSSL error and it's probably rolled up into one of the "various bug fixes". I've not examined their code though so....

PR #3989, opened 2018-04-25
Link to ODF guidance document

### LINKED ISSUE #3962: Guidance on ODF should go to GOV.UK page
In user research, some users have been confused by the text which appears next to ODS, ODT, ODP files:

Screenshot of page on GOV.UK. Explanation text next to a file type.
> This file is in an [OpenDocument](https://en.wikipedia.org/wiki/OpenDocument_software) format

The link currently goes to Wikipedia - which doesn't help users open the files.

The link should go to our guidance page - https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software

This guidance is being updated this month to include better navigation and links to Android & iOS apps.

Ref https://github.com/alphagov/whitehall/blob/master/config/locales/en.yml#L446

bevanloon:
Fixes  #3962 
Supersedes https://github.com/alphagov/whitehall/pull/3963

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -443,7 +443,7 @@ en:
         version of this document in a more accessible format, please email %{email}.
         Please tell us what format you need. It will help us if you say what assistive technology you use.
     opendocument:
-      help_html: This file is in an <a rel="external" href="https://en.wikipedia.org/wiki/OpenDocument_software">OpenDocument</a> format
+      help_html: This file is in an <a href="https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software">OpenDocument</a> format
     see_more: See more information about this %{document_type}
   see_all:
     announcement: See all our announcements


edent:
Awesome! When will it be live on GOV.UK?

bevanloon:
@edent it's live. Thanks for the fix!
🎉

edent:
@bevanloon Are you sure? The link still goes to Wikipedia.

See https://www.gov.uk/government/publications/non-disclosure-agreements 
https://www.gov.uk/government/statistics/family-resources-survey-financial-year-201617

bevanloon:
@edent - oh yes of course - sorry; this is because of the way our publishing workflow is set up.
Published editions are served from what is essentially a static representation of the document at the time it was published. Documents published from here on will get the new link. We would need to do some additional work to find all documents which contain the relevant link and then re-publish them so the content store's representation is updated.

I'll put a card on our backlog to look into it.

edent:
All looking good in new publications - https://www.gov.uk/government/statistics/dbs-dataset-3-number-and-type-of-registered-and-umbrella-body-applications

Hope that older publications can also be updated.

Thank you so much for this!

deborahchua:
@edent the affected documents have now been re-published 🎉

PR #3990, opened 2018-04-25
Change slug to one that matches the page title

1pretz1:
Due to a zendesk ticket being opened asking the slug of
'digital-data-and-technology-job-roles-in-government' to be changed
to 'digital-data-and-technology-profession-capability-framework'
so that the slug reflects the changed title of the page.

https://govuk.zendesk.com/agent/tickets/2762067

### DIFF
--- a/db/data_migration/20180426092205_rename_digital_data_and_technology_job_roles_in_government_to_new_url.rb
+++ b/db/data_migration/20180426092205_rename_digital_data_and_technology_job_roles_in_government_to_new_url.rb
@@ -1,0 +1,14 @@
+old_slug = 'digital-data-and-technology-job-roles-in-government'
+new_slug = 'digital-data-and-technology-profession-capability-framework'
+
+document = Document.find_by(slug: old_slug)
+
+if document
+  # remove the most recent edition from the search index
+  edition = document.editions.published.last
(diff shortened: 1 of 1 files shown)


db/data_migration/20180425140942_change_slug_for_digital_data_and_technology_job_roles_in_government_to_diff_slug.rb
```diff
@@ -0,0 +1,30 @@
...
+  from_path = "/government/collections/#{old_slug}"
+  to_path = "/government/collections/#{new_slug}"
+  redirects = [
+    { path: from_path, type: "exact", destination: to_path }
+  ]
+  redirect = Whitehall::PublishingApi::Redirect.new(from_path, redirects)
+  content_id = SecureRandom.uuid
+
+  puts "Redirecting: #{from_path} to #{to_path} with a randomly generated content_id: #{content_id}"
+  Services.publishing_api.put_content(content_id, redirect.as_json)
```
cbaines: This shouldn't be necessary. The Publishing API should automatically create redirects when content moves.

It wasn't always like this, assuming you've copied a previous data migration, this might have been necessary at the time, but it shouldn't be necessary now.

PR #3991, opened 2018-04-26
Add a new report for all content published by an organisation

boffbowsh:
Generates a report for a given organisation for all content they
have published. To be given to P&E people to prepare reports for
the organisations so they can prepare for Brexit content changes.


lib/tasks/reporting.rake
```diff
@@ -73,6 +75,43 @@ namespace :reporting do
...
+        includes(:document, :translations).
+        where(
+          edition_organisations: { lead: 1, organisation_id: org_id },
+          state: %w{published}
+        )
+
+      progress_bar = ProgressBar.create(format: "%e [%b>%i] [%c/%C]", total: scope.count)
+
+      scope.find_each do |edition|
+        first_published = edition.document.ever_published_editions.order(:id).limit(1).pluck(:public_timestamp).first
```
bevanloon: I'm finding it a little tricky visualising the data coming back and wondering if we can get duplicate data?
i.e. can `scope` have multiple rows (essentially `editions`) per `document`? in which case is it possible that we end up with duplicate rows in the final csv?

boffbowsh: Good catch, I missed that between the version I used and the version I committed. Added a `distinct(:document_id)`.

PR #3997, opened 2018-04-27
Update db/schema.rb file

kevindew:
This makes the db/schema.rb consistent with what Rails produces when a
migration is run.

This is to resolve developers needing to cherry pick their changes into
the db/schema.rb file to avoid a large diff on a migration.

The [index](https://github.com/alphagov/whitehall/compare/sort-out-schema?expand=1#diff-1acd2e7e27a227829d5d14a91c863bb6L81) that has been removed is not present on the production db schema.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 20180424194943) do
 
-  create_table "about_pages", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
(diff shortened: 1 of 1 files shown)


db/schema.rb
```diff
@@ -1198,8 +1197,8 @@
   end
 
   create_table "worldwide_services", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
-    t.string   "name",            null: false
-    t.integer  "service_type_id", null: false
+    t.string   "name",            default: "", null: false
```
thomasleese: Do you know what this is? 😬

kevindew: Nope - seems to come up in staging and integration though so seems to be part of the db.

PR #3998, opened 2018-04-27
Add a rake task for reslugging a document

1pretz1:
Instead of running data_migration to reslug a document, now a rake task
can be ran by inputting the old slug and the new desired slug. This
reindexes the document in elasticsearch, republishes to pub API and
publishes a redirect content item to pub API.

### DIFF
--- a/lib/tasks/reslugging.rake
+++ b/lib/tasks/reslugging.rake
@@ -9,8 +9,7 @@ namespace :reslug do
   It performs the following steps:
   - changes the person's slug
   - reindexes the person for search
-  - republishes the person to Publishing API
-  - publishes a redirect content item to Publishing API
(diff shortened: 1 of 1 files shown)


lib/tasks/reslugging.rake
```diff
@@ -36,4 +36,24 @@ namespace :reslug do
 
     DataHygiene::RoleReslugger.new(role, new_slug).run!
   end
+
+  desc "Change a document's slug in whitehall (DANGER!).\n
+  It performs the following steps:
+  - changes the documents slug
+  - reindexes the document with it's new slug
+  - republishes to Publishing API
+  - publishes a redirect content item to Publishing API"
```
cbaines: This isn't quite right now, as the Publishing API automatically handles the redirect.

PR #4003, opened 2018-04-30
Log obfuscated email addresses for consultation reminders

elliotcm:
Logs elliot.crosby-mccullough@digital.cabinet-office.gov.uk as `el*****uk`.

Useful for monitoring and debugging.

### DIFF
--- a/app/models/consultation_reminder.rb
+++ b/app/models/consultation_reminder.rb
@@ -13,14 +13,26 @@ class ConsultationReminder
 
     def send_deadline_reminder(weeks_left:)
       Consultation.awaiting_response.closed_on((PUBLISH_DEADLINE - weeks_left).weeks.ago.to_date).each do |consultation|
+        log(consultation)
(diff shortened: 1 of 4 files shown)


app/models/consultation_reminder.rb
```diff
@@ -13,14 +13,26 @@ def send_all
...
       end
     end
+
+    def log(consultation)
+      Rails.logger.info("Sending reminder for consultation ##{consultation.id} '#{consultation.title}' to #{obfuscated_email_addresses(consultation)}")
+    end
+
+    def obfuscated_email_addresses(consultation)
+      consultation.authors.uniq.map { |author|
+        author.email.gsub(/^(.{2}).*(.{2})$/, '\1*****\2')
```
thomasleese: I think I would prefer that we do this in Ruby just because I find regular expressions to be unnecessarily confusing, but I don't really mind.

PR #4007, opened 2018-05-01
Republish organisation pages when information changes

rubenarakelyan:
This commit adds `after_save` and `after_destroy` callbacks to the following models that have their data included on an organisation homepage:

* Corporate information pages
* Features
* Featured links
* People
* Roles
* Role appointments
* Social media links

This republishes the content item for the relevant organisation homepage(s) each time one of these models is changed, so that the content item stays up-to-date. This is because this data is stored in the details hash of the relevant organisation(s).

These can be removed once they’ve been fully migrated and are represented as links from the organisation content item.

Models that can be defined as part of creating an organisation also check whether the organisation it
self has been saved yet, since saving the organisation will publish it anyway.

Trello: https://trello.com/c/V4xp9H9C/51-republish-organisation-home-pages-every-time-a-relevant-field-changes


sihugh:
Got some failures in CI :-(

PR #4010, opened 2018-05-01
Bulk slug change for 36 URL's

1pretz1:
Slug changes and redirects from old pages to new pages for 36 different
URL's.

As requested by;
https://govuk.zendesk.com/agent/tickets/2771132


vanitabarrett:
Tested on staging and all URLs look good and were redirecting to the new slugs.

PR #4013, opened 2018-05-02
Update PersonPresenter

Davidslv:
We want to pass to send two extra fields to publishing-api (`full_name` & `privy_counsellor`).

Uppon some research, we have noticed inconsistencies on how editors have been using the different fields (title, forename, surname and letters), some records had the title in the forename and others had the title, surname and letters into the surname.

To avoid future inconsistencies we have decided to have one field (full_name), this way we also are addressing the [Service Manual](https://www.gov.uk/service-manual/design/names#use-a-single-name-field-where-possible).

This is the beginning of that migration.

Part of: https://github.com/alphagov/govuk-content-schemas/pull/755
Trello: https://trello.com/c/32LO2xT8

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -91,6 +91,10 @@ class Person < ApplicationRecord
     name_as_words(("The Rt Hon" if privy_counsellor?), title, forename, surname, letters)
   end
 
+  def full_name
+    name_as_words(title, forename, surname, letters)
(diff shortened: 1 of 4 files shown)


sihugh:
I've merged the govuk-content-schema for this, but need to deploy before the tests will pass.

PR #4018, opened 2018-05-03
Add rake tasks for asset audit

barrucadu:
Adds rake tasks for auditing the assets:

- `asset_manager:dump`, gives (as a CSV) the URLs of every visible asset in Whitehall and its asset-manager path.
- `asset_manager:audit_check_status`, reads the CSV and prints a CSV of the status code, "location" header, "content-type" header, and "content-length" header.

[Trello Card](https://trello.com/c/0mGuSTyF/130-audit-whitehall-attachment-metadata-in-asset-manager-3)

### DIFF
--- a/lib/asset_audit.rb
+++ b/lib/asset_audit.rb
@@ -7,10 +7,9 @@ class AssetAudit
     whitehall_base_url = Plek.new.external_url_for('assets-origin')
     asset_manager_base_url = Plek.new.external_url_for('draft-assets')
 
-    CSV($stdout, headers: %i[whitehall asset_manager], write_headers: true) do |csv|
+    CSV.new($stdout, headers: %i[whitehall asset_manager], write_headers: true) do |csv|
       AttachmentData.find_each do |attachment_data|
-        visible = attachment_data.visible_to?(nil)
-        next unless visible
+        next unless attachment_data.visible_to?(nil)
 
         whitehall_path = attachment_data.file.url
         asset_manager_path = attachment_data.file.asset_manager_path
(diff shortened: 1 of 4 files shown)


lib/asset_audit.rb
```diff
@@ -0,0 +1,104 @@
...
+    "cache_bust=#{rand(1e8)}"
+  end
+
+  def self.dump
+    whitehall_base_url = Plek.new.external_url_for('assets-origin')
+    asset_manager_base_url = Plek.new.external_url_for('draft-assets')
+
+    CSV($stdout, headers: %i[whitehall asset_manager], write_headers: true) do |csv|
+      AttachmentData.find_each do |attachment_data|
+        visible = attachment_data.visible_to?(nil)
```
elliotcm: Might be worth putting some time into filtering invisible attachment data in SQL.  I had a bit of a look and couldn't find a quick solution, though.

lib/asset_audit.rb
```diff
@@ -0,0 +1,104 @@
...
+        next unless visible
```
elliotcm: Do we need this to be an extra variable?

lib/asset_audit.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  def self.check_status(email, urls_filename)
+    new.check_status(email, urls_filename)
+  end
+
+  def check_status(email, urls_filename)
+    mechanize = Mechanize.new
+    mechanize.agent.allowed_error_codes = %w(404)
+    # mechanize.log = Logger.new($stdout)
```
elliotcm: Dead comment

lib/asset_audit.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    sign_user_in(mechanize, email, ask_password)
+
+    csv = CSV.new(
```
elliotcm: Inconsistent use of `CSV(` vs `CSV.new(`.

lib/asset_audit.rb
barrucadu: We think this would be nice to have, but not really worth the effort.

thomasleese:
Thanks @elliotcm. I don't think there's any benefit to checking all the headers as there are bound to be many differences when it comes to version numbers and cache control headers and things like that, I think it's better to only check the headers we're checking now. I've addressed your other comments now.

thomasleese:
Thanks @elliotcm, I just had to make a quick change for 2FA access through Signon if you'd be able to take another look.

PR #4019, opened 2018-05-03
Move organisation index page links to details hash

rubenarakelyan:
This commit changes the organisations index page to have links to all organisations in the details hash rather than the links hash. This allows us to keep the details to a minimum and make the content item for this page smaller.

Trello: https://trello.com/c/kMgJ1gqF/56-link-to-all-organisations-from-the-organisation-list-content-item


app/models/organisation.rb
```diff
@@ -193,10 +193,8 @@ def foi_contacts
 
   before_destroy { |r| throw :abort unless r.destroyable? }
   after_save :ensure_analytics_identifier
-
-  # We only update links when a new organisation is created, since
-  # dependency resolution will take care of all updates and deletes.
-  after_create :update_organisations_index_page_links
+  after_save :update_organisations_index_page_links
+  after_destroy :update_organisations_index_page_links
```
sihugh: We should probably rename this method now if it's no longer links

test/unit/models/organisation_test.rb
```diff
@@ -995,7 +995,7 @@ class OrganisationTest < ActiveSupport::TestCase
     )
   end
 
-  test "#save triggers an update of the organisations index page links hash" do
+  test "#save triggers an update of the organisations index page links" do
```
sihugh: Can probably remove ` links` from the end of this

lib/tasks/publish_organisations_homepage.rake
```diff
@@ -0,0 +1,4 @@
+desc "Publish of the organisations homepage to the publishing api. This is called manually when needed."
+task publish_organisations_homepage: [:environment] do
```
sihugh: Good call to stick in a rake task 👍

app/models/organisation.rb
sihugh: Also, we should be consistent over whether it's the home page or the index page.  I think I prefer index page which means renaming `PublishOrganisationsHomepage`, but I'm not really fussed which way round we go.

lib/publish_organisations_homepage.rb
```diff
@@ -0,0 +1,105 @@
+class PublishOrganisationsHomepage
+  def publish
+    payload = present_for_publishing_api
+    Services.publishing_api.put_content(payload[:content_id], payload[:content])
+    Services.publishing_api.publish(payload[:content_id], nil, locale: "en")
+  end
+
+private
+
+  def base_path
```
tijmenb: `base_path` and `content_id` can be constants perhaps

rubenarakelyan: Yup, done

test/unit/models/organisation_test.rb
rubenarakelyan: Done

app/models/organisation.rb
rubenarakelyan: I've renamed everything to refer to the index page to reduce confusion with individual organisation homepages

PR #4026, opened 2018-05-09
Move tagging interfaces WIP [DO NOT MERGE]

peteglondon:
Moving the legacy tagging from the edition edit page to it's own 'legacy tagging' page.


app/assets/javascripts/templates.js
```diff
@@ -0,0 +1,4 @@
+window.templates = {};
```
tijmenb: This file shouldn't be committed.

PR #4027, opened 2018-05-10
Downgrade minitest to 5.10.3

peteglondon:
Rails 5.0.6 is incompatible with the latest version of minitest (5.11.3)

This causes the test run to crash on the first error.
We need to be able to run the tests so downgrade is necessary until
whitehall is upgraded to rails > 5.1.

The error is documented [in this rails PR](https://github.com/rails/rails/pull/31624)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -112,6 +112,8 @@ group :test do
   gem 'factory_bot'
   gem 'govuk-content-schema-test-helpers'
   gem 'maxitest'
+  # minitest 5.11.3 is incompatible with rails 5.0.6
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -112,6 +112,7 @@ group :test do
   gem 'factory_bot'
   gem 'govuk-content-schema-test-helpers'
   gem 'maxitest'
+  gem 'minitest', '5.10.3'
```
cbaines: A comment here would be ideal, as this can probably be removed once Rails has been updated. It will also help stop people from letting dependabot upgrade the version!

PR #4031, opened 2018-05-14
Update consultation reminder text

elliotcm:
This is the updated text from the content team, edited to accommodate text-only emails.

https://trello.com/c/OrQqkV8J/149-add-content-to-consultation-reminder-email-then-start-sending-these-emails-2


elliotcm:
@thomasleese It's conjured up in the same way controllers are, so I would imagine we'd have a similar error if we deleted the view.

I'll see about tests.

martinlugton:
Happy with these being full URLs. Thanks both.

elliotcm:
@thomasleese Test added

PR #4032, opened 2018-05-14
Present all roles to the publishing-api

Davidslv:
This commit replaces the MinisterialRolePresenter with a generic RolePresenter which presents all roles to the publishing-api. Roles other than MinisterialRoles don't have a base path, but presenting them to the publishing-api allows us to link to them from other documents.

It also adds a new RoleAppointmentPresenter which presents role appointments to the publishing-api for the first time. These model the relationships between people and roles, and allow us to have both current and previous roles represented as edition links.

Finally, it adds links between organisations and roles, and people and roles.

Trello: https://trello.com/c/EB6Dofoj/32-get-whitehall-to-publish-role-data-to-the-content-store


app/presenters/publishing_api/role_presenter.rb
```diff
@@ -0,0 +1,72 @@
...
+        whip_ordering: role.whip_ordering,
+        role_type: role.type.chomp("Role"),
+        seniority: role.seniority,
+        attends_cabinet_type: role.attends_cabinet_type,
+        role_payment_type: role.role_payment_type,
+        supports_historical_accounts: role.supports_historical_accounts,
+      }
+    end
+
+    def whip_organisation
```
Davidslv: WhipOrganisations is not really a model

https://github.com/alphagov/whitehall/blob/master/lib/whitehall/whip_organisation.rb#L17-L21

app/presenters/publishing_api/role_presenter.rb
```diff
@@ -0,0 +1,72 @@
...
+      content = BaseItemPresenter.new(
+        role,
+        title: role.name,
+        update_type: update_type,
+      ).base_attributes
+
+      content.merge!(
+        description: nil,
+        details: details,
+        document_type: role.class.name.underscore,
```
tijmenb: Can we make this `person`? I hope it won't ever change.

lib/tasks/publish_roles_to_publishing_api.rake
```diff
@@ -0,0 +1,10 @@
+namespace :publishing_api do
+  desc "Publishes all roles to Publishing API, async"
+  task publish_all_roles: :environment do
+    # There are 18 roles without a content_id
+    # I suspect that they never made it live, so will ignore those.
+    Role.where.not(content_id: nil).find_each do |role|
+      Whitehall::PublishingApi.publish_async(role)
```
Davidslv: This needs the RolePresenter to be passed on to `Whitehall::PublishingApi`

app/models/role.rb
```diff
@@ -1,6 +1,8 @@
 # @abstract
 class Role < ApplicationRecord
   include HasContentId
+  include PublishesToPublishingApi
```
Davidslv: There's a need to pass the correct RolePresenter for this to work properly

Davidslv: (this has been addressed here: https://github.com/alphagov/whitehall/pull/4032/files#diff-b11bd5eafd93758d3ba71cc8a8642f72R36)

app/presenters/publishing_api/role_appointment_presenter.rb
```diff
@@ -15,23 +15,31 @@ def content_id
     def content
       content = BaseItemPresenter.new(
         item,
-        title: item.name,
+        title: "",
```
sihugh: I guess we ignore this and use the name of the role/person now?

rubenarakelyan: Yeah it didn't make sense to have an appointment have a title since it's never shown and is just a link between two other items that have titles.

PR #4034, opened 2018-05-14
New legacy tagging interface [DO NOT MERGE]

peteglondon:
# What
Add a new page/route/views to edit the 'Policy', 'Policy Area', 'Primary Specialist Sector' and 'Secondary Specialist Sectors'.
The old edit controls have not been removed from the edition edit page by this commit, a new story has been added to do this.

# Why
This is to meet our need to be able to deliver a smooth transition from the current tagging situation in Whitehall, towards the sole use of the Topic Taxonomy.

Currently there is a disparity between the location for the legacy tagging interfaces and the form to tag to the Topic Taxonomy. The former sits within the Edit page in Whitehall, and the latter sits on a separate page.

This change would smooth the transition to the Topic Taxonomy tagging, by first moving the existing interface and functionality out of the edit form, making it similar to the current Topic Taxonomy tagging interface.

The trello card is [create new legacy tagging interfaces for Policies, Policy Areas and Specialist Topics out of the edit page in Whitehall](https://trello.com/c/bBTUraJ0/54-create-new-legacy-tagging-interfaces-for-policies-policy-areas-and-specialist-topics-out-of-the-edit-page-in-whitehall)


cbaines:
This doesn't seem quite right, what is the purpose of this Pull Request?

PR #4045, opened 2018-05-21
Alter links for providing feedback on topic taxonomy

1pretz1:
Remove vague links and replace with more specific ones, like,
'Suggest a new topic` takes the user to that actual form where they can
recommend a new topic. Previsously the feedback links would land the user
on the homepage of the support app with no real connection with what
the link text described.


cbaines:
> Remove vague links and replace with more specific ones, like,
> 'Suggest a new topic` takes the user to that actual form where they can
> recommend a new topic. Previsously the feedback links would land the user
> on the homeopage of the support app with no real connection with what
> the link text described.

:1st_place_medal:

PR #4052, opened 2018-05-23
Add rake tasks to delete attachments with an invalid attachable

barrucadu:
The audit we did was only against visible assets, so there's a task here to check all attachments.

~~I imagine the way we handle the full change (https://github.com/alphagov/asset-manager/pull/560) will be:~~

1. ~~Run `rake attachment:delete_where_attachable_invalid` in whitehall, which prints a list of legacy urls~~
2. ~~Run `rake govuk_assets:unreplace_whitehall_asset` in asset-manager for each one~~

I think we just need to run this on whitehall and wait for all the asset-manager jobs to complete.  Then we can run the audit again and see what happens.

---

[Trello card](https://trello.com/c/uIPJV2kI/150-fix-old-replaced-whitehall-attachments-on-discarded-editions-2)

### DIFF
--- a/lib/tasks/attachment.rake
+++ b/lib/tasks/attachment.rake
@@ -1,0 +1,20 @@
+namespace :attachment do
+  desc 'List and delete attachments with invalid attachables'
+  task delete_where_attachable_invalid: :environment do
+    # adding 'attachable: nil' to this 'where' doesn't work here:
+    # doing so only keeps those with a null attachable_id in the
+    # database; but if the attachable_id is invalid but non-null, the
+    # attachable will be nil:
+    #
+    # irb> Attachment.find(566845).attachable
+    # => nil
+    # irb> Attachment.find(566845).attachable_id
+    # => 330497
+    #
+    Attachment.where(deleted: false).find_each do |attachment|
+      next unless attachment.attachable == nil
+      puts attachment.attachment_data.file.asset_manager_path if attachment.attachment_data && attachment.attachment_data.file
+      attachment.destroy
+    end
+  end
+end


lib/tasks/edition.rake
```diff
@@ -0,0 +1,24 @@
...
+  task :delete_undeleted_attachments_of_deleted_edition, [:id] => :environment do |_, args|
+    begin
+      delete_edition_files(Edition.find(args[:id]))
+    rescue ActiveRecord::RecordNotFound
+      puts "edition #{id} not found"
+    end
+  end
+
+  desc 'Check if any deleted edition has undeleted attachments, and delete them'
+  task delete_undeleted_attachments_of_all_deleted_editions: :environment do
```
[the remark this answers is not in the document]
barrucadu: I'm not sure, because we want to look at the output of this, which feels more like a task to me.  I expect I'd only want to look at the output of a migration if something went wrong.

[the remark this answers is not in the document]
thomasleese: True, happy to keep it as a task!

lib/tasks/edition.rake
```diff
@@ -0,0 +1,12 @@
+namespace :edition do
+  desc 'Check if any deleted edition has undeleted attachments, and delete them'
+  task delete_edition_files: :environment do
+    Edition.where(state: :deleted).find_each do |edition|
+      edition.attachments.each do |attachment|
```
thomasleese: You could change the query here to be `edition.attachments.where(deleted: false)` and avoid the `next if`.

lib/tasks/edition.rake
```diff
@@ -0,0 +1,12 @@
...
+        next if attachment.deleted?
+        puts "deleted edition #{id} has undeleted attachment with url #{attachment.url}"
```
thomasleese: If we're outputting the attachment URL here, should be make https://github.com/alphagov/asset-manager/pull/560 accept the URL - or can we extract the UUID from the URL here?

barrucadu: I think I'll just change the task in that PR to take the legacy URL, rather than the UUID.  I don't think we can get the UUID in whitehall without talking to asset-manager, which would be a lot of overhead.

lib/tasks/edition.rake
```diff
@@ -0,0 +1,13 @@
+namespace :edition do
+  desc 'Check if any deleted edition has undeleted attachments, and delete them'
+  task delete_edition_files: :environment do
+    Edition.where(state: :deleted).find_each do |edition|
+      clear = false
+      edition.attachments.where(deleted: false).each do |attachment|
+        clear = true
+        puts "deleted edition #{id} has undeleted attachment with url #{attachment.url}"
```
thomasleese: Would it be better just to print the legacy ID to make it easier to copy this directly into the Rake task?

barrucadu:
Tried to run this task in integration and got:

```
I, [2018-05-29T15:16:25.579317 #12747]  INFO -- sentry: ** [Raven] Raven 2.7.3 configured not to capture errors: DSN not set
Rails Error: Unable to access log file. Please ensure that /data/vhost/whitehall-admin/releases/20180529151435/log/development.log exists and is writable (ie, make it writable for user and group: chmod 0664 /data/vhost/whitehall-admin/releases/20180529151435/log/development.log). The log level has been raised to WARN and the output directed to STDERR until the problem is fixed.
Using middleware GdsApi::GovukHeaderSniffer to sniff for GOVUK-Request-Id header
Using middleware GdsApi::GovukHeaderSniffer to sniff for GOVUK-Original-Url header
Using middleware GdsApi::GovukHeaderSniffer to sniff for X-GOVUK-Authenticated-User header
Using middleware GdsApi::GovukHeaderSniffer to sniff for GOVUK-Auth-Bypass-Id header
Raven 2.7.3 configured not to capture errors: DSN not set
cannot load such file -- teaspoon/suite_controller excluded from capture: DSN not set
rake aborted!
LoadError: cannot load such file -- teaspoon/suite_controller
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:293:in `require'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:293:in `block in require'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:259:in `load_dependency'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:293:in `require'
/data/vhost/whitehall-admin/releases/20180529151435/config/initializers/teaspoon.rb:2:in `<top (required)>'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:287:in `load'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:287:in `block in load'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:259:in `load_dependency'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:287:in `load'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/engine.rb:648:in `block in load_config_initializer'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/notifications.rb:166:in `instrument'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/engine.rb:647:in `load_config_initializer'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/engine.rb:612:in `block (2 levels) in <class:Engine>'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/engine.rb:611:in `each'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/engine.rb:611:in `block in <class:Engine>'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/initializable.rb:30:in `instance_exec'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/initializable.rb:30:in `run'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/initializable.rb:55:in `block in run_initializers'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/initializable.rb:44:in `each'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/initializable.rb:44:in `tsort_each_child'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/initializable.rb:54:in `run_initializers'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/application.rb:352:in `initialize!'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/railtie.rb:193:in `public_send'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/railtie.rb:193:in `method_missing'
/data/vhost/whitehall-admin/releases/20180529151435/config/environment.rb:5:in `<top (required)>'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:293:in `require'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:293:in `block in require'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:259:in `load_dependency'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/activesupport-5.0.6/lib/active_support/dependencies.rb:293:in `require'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/application.rb:328:in `require_environment!'
/data/vhost/whitehall-admin/shared/bundle/ruby/2.4.0/gems/railties-5.0.6/lib/rails/application.rb:448:in `block in run_tasks_blocks'
/data/apps/whitehall/shared/bundle/ruby/2.4.0/gems/rake-12.3.1/exe/rake:27:in `<top (required)>'
/usr/lib/rbenv/versions/2.4.4/bin/bundle:23:in `load'
/usr/lib/rbenv/versions/2.4.4/bin/bundle:23:in `<main>'
Tasks: TOP => edition:delete_edition_files => environment
(See full trace by running task with --trace)
```

That looks like an issue with a dependency?

barrucadu:
My first attempt at this didn't work, because it turns out that the `:deleted` state for editions just isn't used!  Maybe it is now, and nobody has deleted anything since the change was made.

```
irb(main):001:0> Edition.where(state: :deleted)
=> #<ActiveRecord::Relation []>
```

So now the task looks for attachments with an invalid attachable, which successfully finds the example from #543:

```
irb> Attachment.where(deleted: false).where(id: 566845).find_each do |attachment|
irb> next unless attachment.attachable == nil
irb> puts attachment.attachment_data.file.asset_manager_path
irb> end
/government/uploads/system/uploads/attachment_data/file/310014/DH_Consultation_on_NHS_Trust.pdf
```

barrucadu:
I ran the rake task on integration.  It took two runs to succeed, because there was one dodgy attachment which made it crash (PR amended):

- The first run (run-rake-task/679) took 13 minutes and found 23,364 attachments.
- The second run (run-rake-task/680) took 80 minutes and found 76,074 attachments.  It also found the invalid replacement from https://github.com/alphagov/asset-manager/issues/543  👍 

For context, whitehall has a total of about 2.2 million attachments. Overall there were 65,833 unique files.  The same file can appear in the output multiple times, as multiple `Attachment`s can share the same `AttachmentData`, so it's worth putting the file list through `| sort | uniq` before doing anything further.

---

I picked a random 6,500 of the unique files.  Of those, 3,629 404'd on production, but 2,871 are accessible files.

I looked at a few of the accessible files, and it turns out that only *some* of their `Attachment`s are invalid.  This wasn't a very thorough investigation, but I didn't find any accessible files where *all* their `Attachment`s were invalid.

---

Destroying an `Attachment` causes an [`asset_manager_attachment_delete_worker`](https://github.com/alphagov/whitehall/blob/master/app/workers/asset_manager_attachment_delete_worker.rb) job to be enqueued.  I suspect the asset-manager state will be fixed by just waiting for all these jobs to complete.

barrucadu:
Ran this on staging, it fixed some of the mismatched assets.

PR #4056, opened 2018-05-24
Tweak the Topic Taxonomy tagging message

cbaines:
I don't think this needs to be a warning, or so alarming, as Topic
Taxonomy tagging is done at the document level as the tagging should
be appropriate for any edition.

Therefore, use blue rather than red rather than colour, and update the
text to try to be more informative.

## Old
old-warning

## New
new

### DIFF
--- a/app/assets/stylesheets/admin/alerts.scss
+++ b/app/assets/stylesheets/admin/alerts.scss
@@ -69,6 +69,11 @@ p.warning {
   margin: 15px 0;
 }
 
+div.info {
+  font-size: 1.2em;
+  margin: 15px 0;
+}
+
 .required span {
   color: $orange;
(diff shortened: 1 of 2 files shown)


xodice:
https://signup.microsoft.com/Signup?OfferId=b8223c6c-5717-4564-8ed1-a46bf9b0e30a&dl=O365_BUSINESS_PREMIUM&culture=en-US&country=US&ali=1&origin=getbookings

sistephens:
Can we change this to:

> Topic changes are applied to all editions of a document, including any published editions.

Also, what do we mean by 'any published editions?' Are there multiple? Feel like this might be confusing for a publisher.

sistephens:
Hey @cbaines, are we still doing this?

PR #4057, opened 2018-05-24
Split the Worldwide tagging interface out from the Topic Taxonomy tagging interface [DO NOT MERGE]

leenagupte:
Trello: https://trello.com/c/iNqKZBEq

## What's changed

Create a separate interface to tag documents to the worldwide taxonomy.

The departments that could tag documents to the worldwide taxonomy previously can still tag to it, they just have to do it in a different place.

## Why

These changes should allow for future improvements to the topic tagging interface without affecting worldwide tagging. It also allows us to more easily restrict access to the worldwide tagging interface in the future.

### Before
screen shot 2018-05-24 at 17 35 15
screen shot 2018-05-24 at 17 47 46

### After
screen shot 2018-05-24 at 17 28 24
screen shot 2018-05-24 at 17 28 45
screen shot 2018-05-24 at 17 29 20
screen shot 2018-05-24 at 17 30 36

## Post-deployment tasks

Run `bundle exec rake taxonomy:rebuild_cache`

### DIFF
--- a/lib/taxonomy/world_taxonomy.rb
+++ b/lib/taxonomy/world_taxonomy.rb
@@ -1,0 +1,20 @@
+module Taxonomy
+  class WorldTaxonomy
+    def initialize(adapter: RedisCacheAdapter.new, tree_builder_class: Tree)
+      @adapter = adapter
+      @tree_builder_class = tree_builder_class
+    end
+
+    def all_world_taxons
+      @_world_taxon_list ||= world_taxon_branches
+    end
+
+  private
+
+    def world_taxon_branches
+      @_branches ||= @adapter.world_taxon_data.map do |world_taxon_hash|
+        @tree_builder_class.new(world_taxon_hash).root_taxon
+      end
+    end
+  end
+end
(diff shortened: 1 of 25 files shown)


lib/taxonomy/world_taxonomy.rb
```diff
@@ -0,0 +1,20 @@
...
+    def all_world_taxons
+      @_world_taxon_list ||= world_taxon_branches
+    end
+
+  private
+
+    def world_taxon_branches
+      @_branches ||= @adapter.world_taxon_data.map do |world_taxon_hash|
+        @tree_builder_class.new(world_taxon_hash).root_taxon
+      end
```
cbaines: This is a bit odd, as the concept of "branches" is a bit confused here.

With the Topic Taxonomy, branches mostly refer to level one taxons. As far as I can work out here, the code relies on the structure being the same as the Topic Taxonomy, treating the World taxon like it's a level one taxon, whereas really it's the root of the Worldwide Taxonomy (effectively level 0).

I don't think this is worth trying to unpick further now, but lets put something in the backlog.

cbaines:
I believe the consensus on deploying this is tomorrow (Wednesday) morning, so let's merge this towards the end of today, to avoid blocking Whitehall deployments. I've added `[DO NOT MERGE]` to the title as a reminder.

PR #4061, opened 2018-05-29
Hide legacy association section

peteglondon:
This PR implements the trello card [Hide the legacy associations section from the overview page when edition does not support any of them](https://trello.com/c/7xxP9Ij3/100-hide-the-legacy-associations-section-from-the-overview-page-when-edition-does-not-support-any-of-them)

## Before (when the edition hasn't been tagged)
empty_section

## After 
screen shot 2018-05-29 at 11 27 31

### DIFF
--- a/app/models/edition/specialist_sectors.rb
+++ b/app/models/edition/specialist_sectors.rb
@@ -9,15 +9,15 @@ module Edition::SpecialistSectors
   included do
     has_many :specialist_sectors, foreign_key: :edition_id, dependent: :destroy
     has_many :primary_specialist_sectors,
-             -> { where(primary: true) },
-             class_name: 'SpecialistSector',
-             foreign_key: :edition_id,
-             dependent: :destroy
+      -> { where(primary: true) },
+      class_name: 'SpecialistSector',
+      foreign_key: :edition_id,
+      dependent: :destroy
     has_many :secondary_specialist_sectors,
-             -> { where(primary: false) },
-             class_name: 'SpecialistSector',
-             foreign_key: :edition_id,
-             dependent: :destroy
+      -> { where(primary: false) },
+      class_name: 'SpecialistSector',
+      foreign_key: :edition_id,
+      dependent: :destroy
 
     add_trait do
       def process_associations_before_save(edition)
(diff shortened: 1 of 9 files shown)


app/models/edition/specialist_sectors.rb
```diff
@@ -48,6 +48,14 @@ def specialist_sector_tags
     specialist_sectors.order("specialist_sectors.primary DESC").map(&:topic_content_id)
   end
 
+  def has_primary_sector?
+    !primary_specialist_sector_tag.blank?
+  end
+
+  def has_secondary_sectors?
+    secondary_specialist_sector_tags.any?
+  end
```
cbaines: You could implement `has_specialist_sectors?`, as `specialist_sectors.any?` instead maybe?

app/views/admin/shared/tagging/_legacy_associations_box.html.erb
```diff
@@ -6,41 +6,39 @@
...
+      <% if edition.has_secondary_sectors? %>
+        <h4>Secondary Specialist Sectors</h4>
+        <ul class="secondary-specialist-sectors">
+          <% specialist_sector_names(edition.secondary_specialist_sector_tags).each do |name| %>
+            <li><%= name %></li>
+          <% end %>
+        </ul>
+      <% end %>
+    </div>
+  <% end %>
```
cbaines: It might be interesting to look at using the "No content" pattern here, as shown in the `govuk_admin_template` style guide [1]

1: https://github.com/alphagov/govuk_admin_template/blob/release-v6.6.0/app/views/govuk_admin_template/style_guide/index.html.erb#L544-L553

cbaines: I've just seen that you've done that here 19f2425c4548ca5f511cf278fded58510b7c7c95

app/models/edition/specialist_sectors.rb
cbaines: I've seen that both methods are used in the view, so maybe this shouldn't be consolidated here...

PR #4064, opened 2018-05-29
Separate topic and world tagging permissions

leenagupte:
Trello: https://trello.com/c/xcy8kV75

Related to PR #4057 

## What's changed

Add a new permission to determine whether an organisation can tag to the Worldwide taxonomy.

## Why

This allows us to allow more organisations to tag to the topic taxonomy, without also giving them permission to tag to the worldwide taxonomy.

### DIFF
--- a/app/models/edition/taggable_organisations.rb
+++ b/app/models/edition/taggable_organisations.rb
@@ -55,6 +55,6 @@ private
   end
 
   def education_taggable_organisation_ids
-    Whitehall.organisations_in_tagging_beta["education_related"].to_a
+    Whitehall.organisations_in_tagging_beta.to_a
   end
 end
--- a/app/models/edition/taggable_organisations.rb
+++ b/app/models/edition/taggable_organisations.rb
@@ -11,17 +11,21 @@ module Edition::TaggableOrganisations
   # Only Education based organisations have validation to prevent publishing if
   # they have not been tagged
   def must_be_tagged_to_taxonomy?
-    education_taggable?
+    topic_taxonomy_taggable?
   end
 
   def can_be_tagged_to_taxonomy?
-    education_taggable? || world_taggable?
+    topic_taxonomy_taggable?
+  end
+
+  def can_be_tagged_to_worldwide_taxonomy?
+    world_taggable?
   end
 
 private
 
-  def education_taggable?
-    organisations_in_education_tagging_beta?
+  def topic_taxonomy_taggable?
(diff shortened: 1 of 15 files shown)


app/models/edition/taggable_organisations.rb
```diff
@@ -11,17 +11,21 @@ module Edition::TaggableOrganisations
...
+
+  def can_be_tagged_to_worldwide_taxonomy?
+    world_taggable?
   end
 
 private
 
-  def education_taggable?
-    organisations_in_education_tagging_beta?
+  def topic_taggable?
```
cbaines: Might be confused with specialist topics, or topics (policy areas). Maybe say `topic_taxonomy_taggable?` to disambiguate?

config/organisations_in_tagging_beta.yml
```diff
@@ -2,24 +2,21 @@
 # that are allowed to have their documents tagged into the taxonomy
 # - This is part of the work performed by Finding Things team.
```
cbaines: Maybe this is the time to update the comment also... :)

config/worldwide_tagging_organisations.yml
```diff
@@ -0,0 +1,8 @@
+# This file has a list of organisation content_id's
+# that are allowed to have their documents tagged into the worldwide taxonomy
```
cbaines: Very subtle, but I think this could be misread, as this list controls access to the tagging interface.

We're not doing anything about checking the organisations of the content actually tagged to the Worldwide taxons, which is what I think the description here could infer?

leenagupte: Perhaps. So what change it to something like: 

```
# This file has a list of organisation content_id's
# that are allowed to access the web interface to tag their documents to the worldwide taxonomy
```

It's the same comment that's in `config/organisations_in_tagging_beta.yml` does that need updating too?

cbaines:
I believe this will have the affect of requiring an organisation like the FCO to tag to the Topic Taxonomy, where they haven't been required to before. I think we want to make that change, but maybe alongside the other changes in the feature branch?

config/worldwide_tagging_organisations.yml
cbaines: Yeah, that looks good. I think the "access the web interface to" could also be omitted, without loosing the important information.

As for the other file, yep, similar changes would be good.

PR #4068, opened 2018-05-30
Fix bug in localised articles

peteglondon:
When building the legacy tagging page, I should have use the
LocalisedModel class used by the original editions controller.

This introduced a problem where the edition was not localised
correctly.

It also broke the "Create a News article of type 'world news story'
in a non-English language" scenario in news-article.feature

Trello: https://trello.com/c/6ReeGM43/101-investigate-the-comment-in-features-stepdefinitions-newsarticlestepsrb120


cbaines:
There is a small typo in the commit message: `scenarion`

PR #4072, opened 2018-05-31
Fix a legacy associations test related to policies

cbaines:
This updates the expectation to match the behaviour in Whitehall. I
probably broke this when attempting to merge the addition of these
analytics related data attributes and the movement of the form
fields.


cbaines:
I think this has worked, but other tests are also broken, so I'm going to force merge this in to the feature branch.

PR #4076, opened 2018-06-01
first_published_at can't be set in the future

deborahchua:
There is an issue where if the `first_published_at` is set to any date after the current, it throws and error when trying to edit the edition again.

The `edition.goverment` relies on the `first_published_at` date to associate the edition with the right government and therefore should not be possible to set in the future.

Trello card: [Whitehall - documents can become uneditable if you change first published at date to future](https://trello.com/c/MzhjknzC/176-whitehall-documents-can-become-uneditable-if-you-change-first-published-at-date-to-future)

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -51,6 +51,7 @@ class Edition < ApplicationRecord
   validates :body, presence: true, if: :body_required?, length: { maximum: 16777215 }
   validates :summary, presence: true, if: :summary_required?, length: { maximum: 65535 }
   validates :first_published_at, recent_date: true, allow_blank: true
+  validates :first_published_at, previously_published: true
 
   UNMODIFIABLE_STATES = %w(scheduled published superseded deleted).freeze
   FROZEN_STATES = %w(superseded deleted).freeze
@@ -277,7 +278,6 @@ class Edition < ApplicationRecord
     public_timestamp: :public_timestamp,
     relevant_to_local_government: :relevant_to_local_government?,
     world_locations: nil,
-
     # DID YOU MEAN: Policy Area?
     # "Policy area" is the newer name for "topic"
     # (https://www.gov.uk/government/topics)
@@ -286,7 +286,6 @@ class Edition < ApplicationRecord
     # You can help improve this code by renaming all usages of this field to use
     # the new terminology.
     topics: nil,
-
     only: :search_only,
     index_after: [],
     unindex_after: [],
@@ -460,8 +459,12 @@ class Edition < ApplicationRecord
     unless published?
       raise "Cannot create new edition based on edition in the #{state} state"
     end
-    draft_attributes = attributes.except('id', 'type', 'state', 'created_at', 'updated_at', 'change_note', 'minor_change', 'force_published', 'scheduled_publication')
-    self.class.new(draft_attributes.merge('state' => 'draft', 'creator' => user)).tap do |draft|
+    ignorable_attribute_keys = %w(id type state created_at updated_at change_note
+                                  minor_change force_published scheduled_publication)
+    draft_attributes = attributes.except(*ignorable_attribute_keys)
+      .merge('state' => 'draft', 'creator' => user, 'previously_published' => previously_published)
+
+    self.class.new(draft_attributes).tap do |draft|
       traits.each { |t| t.process_associations_before_save(draft) }
       if draft.valid? || !draft.errors.keys.include?(:base)
         if draft.save(validate: false)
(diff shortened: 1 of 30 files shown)


deborahchua:
I'm not too happy with having to use nested `if` statements so any suggestions to improve it is welcome!

kevindew:
Hey @deborahchua I had a look into why this isn't working for you in the app. 

The problem seems to stem from https://github.com/alphagov/whitehall/blob/22080adb507ff7d8d05b55f191eb3ede78b64b9c/app/models/edition.rb#L637 which seems to stem from this: https://github.com/alphagov/whitehall/blob/22080adb507ff7d8d05b55f191eb3ede78b64b9c/app/controllers/admin/editions_controller.rb#L16

I'd suggest that since this validation should always be run that you add it as a one as per: https://github.com/alphagov/whitehall/blob/22080adb507ff7d8d05b55f191eb3ede78b64b9c/app/models/edition.rb#L53

app/models/edition.rb
```diff
@@ -639,7 +639,11 @@ def previously_published_documents_have_date
       errors[:base] << 'You must specify whether the document has been published before'
       @has_first_published_error = true
     elsif previously_published == 'true' # not a real field, so isn't converted to bool
-      errors.add(:first_published_at, "can't be blank") if first_published_at.blank?
+      if first_published_at.blank?
```
barrucadu: I think the nested `if` is fine.  The alternative is repeating the condition, which isn't great.

app/validators/previously_published_validator.rb
```diff
@@ -0,0 +1,14 @@
+class PreviouslyPublishedValidator < ActiveModel::Validator
+  def validate(record)
+    if record.previously_published.nil?
+      record.errors[:base] << "You must specify whether the document has been published before"
+      record.has_previously_published_error = true
```
kevindew: Do you want an inverse of this anywhere? I assume once you have this set to true and change state of the model there is nothing to set this back to false? (not sure that matters)

app/validators/previously_published_validator.rb
```diff
@@ -0,0 +1,14 @@
...
+    elsif record.previously_published
+      if record.first_published_at.blank?
+        record.errors.add(:first_published_at, "can't be blank")
+      elsif record.first_published_at > Time.zone.now
```
kevindew: so if I understand correct this means that:

all records that have been previously published cannot have a first_published_at in the future and thus the inverse that any record that has not been previously published can have a first_published_at in the future?

Would the bug not still occur when someone tries to associate with a government and has a first_published_at date in the future but the document is not previously published?

steventux: Both good points, I've attempted to address in the [following commit](https://github.com/alphagov/whitehall/pull/4076/commits/3adb1d2b560e2d77bf560b7b22cef10793a296be).
I think the UI behaviour would prevent the manual creation of a non-previously published doc with a future `first_published_at` date which is reason enough to make sure the validator covers this scenario. 
I've also falsified the `has_previously_published_error` var before validating.

app/validators/previously_published_validator.rb
```diff
@@ -0,0 +1,16 @@
+class PreviouslyPublishedValidator < ActiveModel::Validator
+  def validate(record)
+    record.has_previously_published_error = false
+
+    if record.previously_published.nil?
+      record.errors[:base] << "You must specify whether the document has been published before"
+      record.has_previously_published_error = true
+    elsif record.previously_published && record.first_published_at.blank?
+      record.errors.add(:first_published_at, "can't be blank")
```
barrucadu: Should this branch (and the `if` below) also set `has_previously_published_error`?  The `has_first_published_error` logic that this replaces does that: https://github.com/alphagov/whitehall/pull/4076/files#diff-d9c2d399fb6d507d2c0c4da96feece4eL647

test/support/admin_edition_controller_test_helpers.rb
```diff
@@ -293,7 +293,8 @@ def should_allow_editing_of(edition_type)
         put :update, params: {
           id: edition,
           edition: {
-            title: '   my title    '
+            title: '   my title    ',
+            previously_published: false
```
barrucadu: Should there be a trailing comma here (like https://github.com/alphagov/whitehall/pull/4076/files#diff-669351f665ea3e569b4809a078a4cf9dR16)?  I thought the linter enforced a common style for this.

app/validators/previously_published_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+    end
+
+    if record.first_published_at && Time.zone.now < record.first_published_at
+      record.errors.add(:first_published_at, "can't be set to a future date")
+    end
```
kevindew: Sorry - I'm sure you don't need another comment from me on this long running thing 😞 

Since this is pretty distinct from previously_published do we just want this as a separated validation?

app/validators/previously_published_validator.rb
binaryberry: The reason `has_previously_published_error` exists is to apply some frontend formatting when the user hasn't selected any option in the radio buttons - https://github.com/alphagov/whitehall/blob/3adb1d2b560e2d77bf560b7b22cef10793a296be/app/views/admin/editions/_first_published_at.html.erb#L2
I think the reason this is needed is because `previously_published` is a transient attribute, whereas`first_published_at` isn't, so when there is an error on it the generic bootstrap red formatting will be applied as per:
screen shot 2018-06-14 at 14 15 54

test/support/admin_edition_controller_test_helpers.rb
binaryberry: Yes there should! Thanks, amended now.

binaryberry: The linter only goes through the lines you're modifying I think! It'd be very noisy if not as it was introduced a few years into the life of the app :-)

cbaines: On this subject, I've got a PR to fix/ignore the few remaining linting issues, and then change the Jenkins configuration to check the entire codebase for any change: https://github.com/alphagov/whitehall/pull/4113

app/validators/previously_published_validator.rb
binaryberry: Thanks! Addressed in f17bcab. Seemed small enough to have directly in the validations, but happy to extract it in a method if not.

app/models/edition.rb
```diff
@@ -51,6 +51,10 @@ class Edition < ApplicationRecord
   validates :body, presence: true, if: :body_required?, length: { maximum: 16777215 }
   validates :summary, presence: true, if: :summary_required?, length: { maximum: 65535 }
   validates :first_published_at, recent_date: true, allow_blank: true
+  validates :first_published_at, previously_published: true
+  validates_each :first_published_at do |record, attr, value|
+    record.errors.add(attr, "can't be set to a future date") if Time.zone.now < value
```
kevindew: This looks good.

I was worried that this might catch people out as there are a bunch of editions with future publishing dates:

```
irb(main):005:0> Edition.where("first_published_at > ?", Time.zone.now).count
=> 38
```

However they are all superseded:

```
irb(main):006:0> Edition.where("first_published_at > ?", Time.zone.now).distinct.pluck(:state)
=> ["superseded"]
```

So I think we should probably go for this.

If it does transpire though that publishers have a valid reason to set first_published_at date into the future then we may have to add a conditional to this which only applies this check if `political` is true (which, as I understand, is the attribute that associates this record with a government)

binaryberry:
Deployed the branch to integration and tested in the UI the error messages for previously_published not selected, first_published_at can't be blank and first_published_at can't be set to a future date.
Also tried out in the UI the happy paths and all seems in good order.

PR #4077, opened 2018-06-01
Copy all content tagged to Brexit Policy to Brexit Taxon

1pretz1:
Every 30 minutes, a rake task will run which checks to see if there are
new content items tagged to the Brexit Policy taxon which are not tagged to
the Brexit Topic Taxonomy. If so, these content items will be tagged to
the new Brexit Topic Taxonomy, this will keep the new taxonomy up to
date with the latest information regarding Brexit.

### DIFF
--- a/lib/taxonomy/sync_brexit_policies.rb
+++ b/lib/taxonomy/sync_brexit_policies.rb
@@ -1,0 +1,51 @@
+module Taxonomy
+  class SyncBrexitPolicies
+    attr_reader :log
+
+    BREXIT_POLICY_ID = '2dcb5926-db6e-4347-b6d8-64fa9d5779a5'.freeze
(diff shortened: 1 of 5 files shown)


lib/taxonomy/sync_brexit_policies.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def links_for_brexit_policy_editions
+      Services.publishing_api.get_links_for_content_ids(brexit_policy_editions_query)
+    end
+
+    def patch_editions
+      links_for_brexit_policy_editions.each do |content_id, edition_links|
+        edition_taxons = edition_links.dig("links", "taxons") || []
+
+        if !edition_taxons.include?(BREXIT_TAXON_ID)
```
peteglondon: maybe use unless instead id if ! here

1pretz1: Agreed 👍

PR #4086, opened 2018-06-04
Alter legacy tagging page

1pretz1:
Add descriptive text on what legacy tagging is, and it's purpose. Change button text 'Save legacy associations' to 'Save' and redirect after clicking to the topic tagging page instead of the overview edit page.

### DIFF
--- a/app/views/admin/edition_legacy_associations/edit.html.erb
+++ b/app/views/admin/edition_legacy_associations/edit.html.erb
@@ -1,28 +1,20 @@
 <% page_title "Edit legacy associations: " + @edition.title %>
 <div class="row">
   <h1><%= @edition.title %></h1>
+  <p>
+    Policies, policy areas and specialist sectors will continue to be supported until the topic taxonomy replaces them.
+  </p>
 </div>
 
 <% if flash[:legacy_associations_changed] %>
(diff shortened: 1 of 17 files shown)


1pretz1:
## Original
screen shot 2018-06-04 at 15 57 02

## Updated

# Creating a new document with legacy tags auto generated 
screen shot 2018-06-04 at 15 55 24

# Editing existing document with tags already added
screen shot 2018-06-04 at 15 54 47

app/views/admin/edition_legacy_associations/edit.html.erb
```diff
@@ -1,28 +1,20 @@
 <% page_title "Edit legacy associations: " + @edition.title %>
 <div class="row">
   <h1><%= @edition.title %></h1>
+  <p>
+    Policies, policy areas and specialist sectors will continue to be supported until the topic taxonomy replaces them.
```
cbaines: Perhaps capitalise "Topic Taxonomy" here, as I see it as a proper noun. Same with "Policy Areas" and "Specialist Sectors" actually.

1pretz1: 👍

PR #4090, opened 2018-06-05
Add DH to taggle organisations for user research

1pretz1:
This needs to be removed after user research with the Department of Health.


cbaines:
This is just needed for the user research, so we don't need to merge this in, just deploy it to Integration.

cbaines:
@1pretz1 Hey, can we close this now? I think it's served it's purpose

PR #4094, opened 2018-06-06
Fix the delay of topic taxons appearing on the editions overview page

1pretz1:
https://trello.com/c/Ch9ufVJ4/139-bug-when-i-save-a-topic-selection-for-the-first-time-the-topics-dont-appear-on-the-overview-page

Currently, when rendering the overview edit page, after selecting the
Topic Taxons you want to tag the edition too, the taxons sometimes don't
show in the topic taxon display box.

This is due to the request getting the Publishing API's stored
data which might be out of date, adding `generate: true` into the
request requires publishing-api to generate the expanded links instead
of relying on its already stored expanded links. This should ensure that
the most up to date links are returned. Meaning there now should be no
delay in the displaying of taxon information.

This relies on https://github.com/alphagov/gds-api-adapters/pull/837 to be updated first.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -154,7 +154,7 @@ GEM
     fugit (1.1.1)
       et-orbi (~> 1.1, >= 1.1.1)
       raabro (~> 1.1)
-    gds-api-adapters (52.5.1)
+    gds-api-adapters (52.6.0)
       addressable
       link_header
       lrucache (~> 0.1.1)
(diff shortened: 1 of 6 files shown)


cbaines:
Good stuff, looks like the build is failing as the previously mocked requests don't quite match up any more.

> Currently, when rendering the overview edit page, after selecting the
> Topic Taxons you want to tag the edition too, the taxons sometimes don't
> show in the topic taxon display box.
> 
> This is due to the request trying to access the Publishing API's stored
> data which might be out of date,

This doesn't read quite right, Whitehall isn't "trying" to access the cached expanded links in the Publishing API, it's just what it gets at the moment.

>  adding generate: true into the
> request requires publishing-api to generate the expanded links instead
> of relying on its already stored expanded links. 

:+1: 

> This should ensure that
> the most up to date links are returned. Meaning their now should be no

Wrong `there/they're/their`, it should be "there" I think.

> delay in the displaying of taxon information.

PR #4099, opened 2018-06-07
Remove empty fields from role appointments presenter

rubenarakelyan:
This commit removes the empty fields from the role appointments presenter, and also provides a title field, which is required.

### DIFF
--- a/app/presenters/publishing_api/role_appointment_presenter.rb
+++ b/app/presenters/publishing_api/role_appointment_presenter.rb
@@ -13,22 +13,16 @@ module PublishingApi
     end
 
     def content
-      content = BaseItemPresenter.new(
-        item,
-        title: "",
-        update_type: update_type,
-      ).base_attributes
-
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/role_appointment_presenter.rb
```diff
@@ -41,5 +35,15 @@ def links
...
     end
+
+  private
+
+    def title
+      "#{item.person.name} - #{item.role.name}"
+    end
+
+    def locale
+      I18n.locale.to_s
```
tijmenb: Maybe it's clearer if this is always `en`? I don't think we'll ever publish non-English role appointments.

sihugh: I think roles and people can have translations, but not the appointment?  (There's no locale for role_appointment in the whitehall schema).  It won't hurt very much either way I think

PR #4103, opened 2018-06-07
Use correct form param for worldwide tagging

andrewgarner:
This fixes the controller to use the correct form param which can be seen in the body information of the Sentry alerts.

Fixes https://sentry.io/govuk/app-whitehall/issues/572623537/


cbaines:
Right, I've had a deeper look at this now, and I've got a suggested fix here https://github.com/alphagov/whitehall/pull/4105

PR #4104, opened 2018-06-07
Rename a regulatory judgement publication 

thomasleese:
A company name change means we should update the URL to match.

[Trello Card](https://trello.com/c/hp0YNMBB/210-whitehall-url-change)

### DIFF
--- a/db/data_migration/20180607142603_rename_regulatory_judgement_publication.rb
+++ b/db/data_migration/20180607142603_rename_regulatory_judgement_publication.rb
@@ -1,0 +1,4 @@
+DataHygiene::DocumentReslugger.call(
+  from: "regulatory-judgement-symphony-housing-group-ltd",
+  to: "regulatory-judgement-onwards-homes-limited",
+)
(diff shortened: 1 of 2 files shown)


db/data_migration/20180607142603_rename_regulatory_judgement_publication.rb
```diff
@@ -0,0 +1,4 @@
+DataHygiene::DocumentReslugger.call(
+  from: "regulatory-judgement-symphony-housing-group-ltd",
+  to: "regulatory-judgement-onwards-homes-limited",
+)
```
cbaines: Have you considered running the document reslugging rake task?
```
rake reslug:document[regulatory-judgement-symphony-housing-group-ltd,regulatory-judgement-onwards-homes-limited]
```

thomasleese: Well... I didn't even know that existed! It feels like these reslugging tasks should be better documented.

cbaines: Ha, probably :)

PR #4106, opened 2018-06-07
Always show the legacy associations box

cbaines:
Rather than making it show up only when you can't tag to the Worldwide
taxonomy, as this doesn't make sense.


1pretz1:
LGTM 👍

PR #4111, opened 2018-06-12
Add original primary publishing organisation

rubenarakelyan:
This commit adds a new `original_primary_publishing_organisation` field to the presented version of any document that is tagged to an organisation. This field contains the organisation that originally published the document, even if that has changed.

Trello: https://trello.com/c/u80rFvuW/152-retain-org-ownership-history-of-content

### DIFF
--- a/app/presenters/publishing_api/links_presenter.rb
+++ b/app/presenters/publishing_api/links_presenter.rb
@@ -3,6 +3,7 @@ module PublishingApi
     LINK_NAMES_TO_METHODS_MAP = {
       organisations: :organisation_ids,
       primary_publishing_organisation: :primary_publishing_organisation_id,
+      original_primary_publishing_organisation: :original_primary_publishing_organisation_id,
       policy_areas: :policy_area_ids,
       related_policies: :related_policy_ids,
       statistical_data_set_documents: :statistical_data_set_ids,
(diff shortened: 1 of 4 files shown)


cbaines:
> This commit adds a new original_primary_publishing_organisation field to the presented version of any document that is tagged to an organisation. This field contains the organisation that originally published the document, even if that has changed.

This seems a little odd, but then generally, the history of links isn't stored (ignoring editioned links). I can kind of see the use for the information though.

Would you see this changing if the organisation changes, .e.g is replaced by another organisation, or will this be unchanging, and link to the organisation, historical or not?

Also, is there any need to deal with multiple organisation changes, e.g. A is superseded by B is superseded by C? I guess the information on B would be lost if the fields used are `original_primary_publishing_organisation` and `primary_publishing_organisation`?

rubenarakelyan:
The intent is for this field to be a permanent record of the organisation that originally published a content item. We're not interested in intermediate changes of ownership at the moment - only original and current organisations. We're planning to display this in the metadata section of various content pages.

cbaines:
> The intent is for this field to be a permanent record of the organisation that originally published a content item. We're not interested in intermediate changes of ownership at the moment - only original and current organisations. We're planning to display this in the metadata section of various content pages.

Ok, thanks :)

PR #4112, opened 2018-06-13
Add tracking to taxon selections

peteglondon:
## Why
We want GA events for taxon selections. We want to know what users are tagging to, so we can analyse the amount of time a given publication takes against the type of topics a user selects.

## What
Send an event to GA with the selections the user has made, ideally once (on save) with the complete list of selected taxons.

We want to track the final selections our publishers make, but we do not want to track anything a user selects prior without saving as it gives us little insight. This would give us too much noise and not help gain insight into the decisions users are making prior to publishing. 

Trello: [Add GA tracking of the topic taxonomy selections in Whitehall](https://trello.com/c/Ex7vU71H/140-add-ga-tracking-of-the-topic-taxonomy-selections-in-whitehall)

### DIFF
--- a/app/assets/javascripts/admin/modules/track_selected_taxons.js
+++ b/app/assets/javascripts/admin/modules/track_selected_taxons.js
@@ -1,0 +1,25 @@
+(function (Modules) {
+    "use strict";
+
+    Modules.TrackSelectedTaxons = function () {
+        var extractBreadcrumb = function (element) {
+            return element.find('li').map(function () {
+                return $(this).text();
+            }).get().join(' > ');
+        };
+        this.start = function (container) {
+            var trackSelectedTaxons = function () {
+                var category = container.data("track-category"),
+                    label = container.data("track-label");
+
+                $('.taxon-breadcrumb').each(function (idx, element) {
+                    var action = extractBreadcrumb($(element));
+                    GOVUKAdmin.trackEvent(category, action, {label: label});
+                });
+            };
+
+            container.on("click", trackSelectedTaxons);
+        };
+    };
+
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/admin/modules/track_selected_taxons.js
```diff
@@ -0,0 +1,26 @@
...
+        };
+        this.start = function (container) {
+            var trackSelectedTaxons = function () {
+                var category = container.data("track-category"),
+                    label = container.data("track-label");
+
+                $('.taxon-breadcrumb').each(function (idx, element) {
+                    var action = extractBreadcrumb($(element));
+                    GOVUKAdmin.trackEvent(category, action, {label: label});
+
```
cbaines: You've got an odd blank line here.

app/views/admin/edition_tags/edit.html.erb
```diff
@@ -51,11 +51,21 @@
...
+          <%= form.submit('Save and review legacy tagging', name: "legacy_tags", class: "btn btn-lg btn-primary",
+              data: {
+                  module: 'track-selected-taxons',
+                  track_category: 'taxonSelection',
+                  track_label: request.path
+              }) %>
           <span class="or_cancel">
             or
-            <%= link_to('cancel', admin_edition_path(@edition)) %>
+            <%= link_to('cancel', edit_admin_edition_tags_path(@edition)) %>
```
cbaines: How come this is changing?

peteglondon: This is wrong, RubyMine getting a bit over-enthusiastic I think.

PR #4114, opened 2018-06-13
Make Worldwide taxons expand/collapse per country

1pretz1:
https://trello.com/c/O7if1mGu/147-improve-the-worldwide-tagging-interface-3

Currently, only `world` expands which displays a huge list of all
the countries and their sub taxons. This is not good for the user as its
an overload of information.

This change provides a link to expand/collapse per country, this new list
also displays everything alphabetically starting with all the taxons
that have no children.

## Before 
worldwide-tagging

## After
screen shot 2018-06-13 at 12 33 19

### DIFF
--- a/lib/taxonomy/world_taxonomy.rb
+++ b/lib/taxonomy/world_taxonomy.rb
@@ -6,10 +6,16 @@ module Taxonomy
     end
 
     def all_world_taxons
-      @_world_taxon_list ||= world_taxon_branches
+      @_all_world_taxons ||= sorted_world_taxons
     end
 
   private
+
+    def sorted_world_taxons
+      world_taxon_branches.sort_by do |world_taxon|
+        [world_taxon.children.count.zero? ? 0 : 1, world_taxon.name]
+      end
+    end
 
     def world_taxon_branches
       @_branches ||= @adapter.world_taxon_data.map do |world_taxon_hash|
--- a/test/integration/world_tagging_flow.rb
+++ b/test/integration/world_tagging_flow.rb
@@ -1,0 +1,66 @@
+require 'test_helper'
+require 'capybara/rails'
+
+class WorldTaggingFlow < ActionDispatch::IntegrationTest
+  extend Minitest::Spec::DSL
+  include Capybara::DSL
+  include TaxonomyHelper
+  include Rails.application.routes.url_helpers
+
+  setup do
+    login_as_admin
+  end
+
+  context 'given I want to tag to the WorldWide taxonomy' do
+    let(:world_organisation) do
+      create(:organisation, content_id: "9adfc4ed-9f6c-4976-a6d8-18d34356367c")
+    end
+    let(:world_edition) do
+      create(:publication, :guidance, organisations: [world_organisation])
+    end
+
+    before do
+      setup_world_tagging_page
+    end
+
+    it "asserts WorldWide tagging path" do
+      visit_world_tags_path
+      assert_equal(200, page.status_code)
+    end
+
+    it "asserts a selected checkbox' " do
+      visit_world_tags_path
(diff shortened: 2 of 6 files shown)


lib/taxonomy/world_taxonomy.rb
```diff
@@ -6,11 +6,17 @@ def initialize(adapter: RedisCacheAdapter.new, tree_builder_class: Tree)
     end
 
     def all_world_taxons
-      @_world_taxon_list ||= world_taxon_branches
+      @_all_world_taxons ||= sort_world_taxons
     end
 
   private
 
+    def sort_world_taxons
```
cbaines: I'd expect `sort_world_taxons` to take some taxons, and sort them. Maybe `sorted_world_taxons` would be a better name for the way this is being used?

test/unit/taxonomy/world_taxonomy_test.rb
```diff
@@ -7,14 +7,15 @@ def subject
...
+  test "#sort world_taxons alphabetically when children == 0, then alphabetically on world_taxons with children" do
     redis_cache_has_world_taxons(
       [
-        build(:taxon_hash, content_id: 'content_id_1'),
-        build(:taxon_hash, content_id: 'content_id_2'),
-        build(:taxon_hash, content_id: 'content_id_3')
+        build(:taxon_hash, title: 'France', children: [child_taxon]),
+        build(:taxon_hash, title: 'Australia', children: [child_taxon, child_taxon]),
+        build(:taxon_hash, title: 'Vietnam'),
+        build(:taxon_hash, title: 'Germany')
```
cbaines: I'd use more realistic data here, like a title of "News and Events", just to help people in the future make sense of the tests.

test/unit/taxonomy/world_taxonomy_test.rb
```diff
@@ -7,14 +7,15 @@ def subject
     Taxonomy::WorldTaxonomy.new
   end
 
-  test "#world_taxons" do
+  test "#sort world_taxons alphabetically when children == 0, then alphabetically on world_taxons with children" do
```
cbaines: This doesn't quite make sense to me, perhaps just say `test "#order of world taxons"`?

test/support/taxonomy_helper.rb
```diff
@@ -134,15 +134,15 @@ def draft_taxon_2
 
   def world_grandchild_taxon
     FactoryBot.build(:taxon_hash,
-                     title: "World Child Taxon",
+                     title: "World grand child taxon",
```
cbaines: grandchild is one word, not two.

cbaines:
It looks like you've made some changes related to comments I've made. Just a note, but it is useful to comment and say this has happened, just to make it clear what is going on.

cbaines:
Is this change going to require the cache for the taxon data to be repopulated from the Publishing API?

1pretz1:
Yes, the cache would have to be refreshed!

test/integration/world_tagging_flow.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    it "asserts WorldWide tagging path" do
+      visit_world_tags_path
+      assert_equal(200, page.status_code)
+    end
+
+    it "asserts a selected checkbox' " do
+      visit_world_tags_path
+      expand_and_check_checkbox
+      assert page.has_checked_field?('World grand child taxon')
```
cbaines: I'm a bit surprised this is still working, as the taxon has now been changed to "grandchild"...

1pretz1: Urmm, weird! I've ran it locally and it failed, i've corrected the test accordingly.

PR #4122, opened 2018-06-14
Separate DocumentListExportWorker to have own queue

Davidslv:
I would like to separate this worker, to have it's own queue, as I believe the emails are not being sent in production.

We currently have a few errors regarding the size of the attachment [being superior to 10MB ](https://sentry.io/govuk/app-whitehall/?query=is%3Aunresolved+Net%3A%3ASMTPFatalError).

I want to narrow down if the queue is too big in Production that the emails are not being triggered in a timely fashion way.

### DIFF
--- a/app/workers/document_list_export_worker.rb
+++ b/app/workers/document_list_export_worker.rb
@@ -1,4 +1,6 @@
 class DocumentListExportWorker < WorkerBase
+  sidekiq_options queue: 'export_documents_list'
+
   def perform(filter_options, user_id)
     user = User.find(user_id)
     filter = create_filter(filter_options, user)


cbaines:
> I want to narrow down if the queue is too big in Production that the emails are not being triggered in a timely fashion way.

When has this been happening? I ask, as the latency for the `default` queue, which I'm assuming is the current queue being used, hasn't got above 10 seconds this week.

Davidslv:
@cbaines I've received a [ticket yesterday](https://govuk.zendesk.com/agent/tickets/2847528), after some investigation I've realised the data is too big to be sent via email (it's above 10MB) - that error seems to be around sentry for 2 months now.

screen shot 2018-06-14 at 16 04 59

I thought that reducing the scope of the export would help the person receiving the report, although this morning I tried to send myself a CSV with 7 documents and did this for both Staging and Production, I've received the email from Staging, but the one from Production has never arrived... that's been several hours now, it shouldn't take that long to receive a small CSV.

cbaines:
> I thought that reducing the scope of the export would help the person receiving the report, although this morning I tried to send myself a CSV with 7 documents and did this for both Staging and Production, I've received the email from Staging, but the one from Production has never arrived... that's been several hours now, it shouldn't take that long to receive a small CSV.

Sounds like there could be a problem, but the data I can see on the `default` queue for the length and the latency, suggest that the problem isn't related to contention with other jobs, or delays on the `default` queue.

Davidslv:
@cbaines I'm going to close this, it seems the problem lies in some kind of email filtering preventing the CSV's to be delivered in production.

PR #4126, opened 2018-06-15
Add new fields to organisations index page

rubenarakelyan:
This commit adds new fields to the organisations index page content item, to be consumed by the new organisations API in collections.

Trello: https://trello.com/c/p7ywlbk1/130-migrate-organisations-api


rubenarakelyan:
Well it passed this: https://github.com/alphagov/whitehall/blob/master/test/unit/publish_organisations_index_page_test.rb#L12-L16 :)

PR #4128, opened 2018-06-18
Tweak formatting for organisations relationship text

rubenarakelyan:
This commit:

* Adds spacing before the organisation relationship text
* Only displays the organisation relationship text if the organisation has parents/children
* Adds a class to links in the organisations relationship text to ensure they have the correct colour

### DIFF
--- a/app/presenters/publishing_api/organisation_presenter.rb
+++ b/app/presenters/publishing_api/organisation_presenter.rb
@@ -90,7 +90,13 @@ module PublishingApi
     end
 
     def summary
-      "#{item.summary}#{organisation_display_name_including_parental_and_child_relationships(item)}"
+      "#{item.summary}#{parent_child_relationships_text}"
+    end
+
+    def parent_child_relationships_text
+      if item.parent_organisations.any? || item.supporting_bodies.any?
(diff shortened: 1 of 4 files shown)


app/presenters/publishing_api/organisation_presenter.rb
```diff
@@ -90,7 +90,13 @@ def acronym
     end
 
     def summary
-      "#{item.summary}#{organisation_display_name_including_parental_and_child_relationships(item)}"
+      "#{item.summary}#{parent_child_relationships_text}"
+    end
+
+    def parent_child_relationships_text
+      if (item.parent_organisations.any? || item.supporting_bodies.any?)
+        "\n\n#{organisation_display_name_including_parental_and_child_relationships(item)}"
```
sihugh: what's the spacing for?

andysellick: I think is the thing where sometimes there isn't a space between this sentence and the rest of the body? (see below towards the end of the paragraph)

screen shot 2018-06-18 at 09 45 29

sihugh: Ah, I just thought it was weird having new lines in an HTML string.

PR #4137, opened 2018-06-19
Redirect swaziland world news to eswatini world news

elliotcm:
https://trello.com/c/1tPr8NQ9/245-redirect-from-world-swaziland-news-to-world-eswatini-news


cbaines:
It would be good to use the unpublish action in the Publishing API if that's possible?

elliotcm:
@cbaines true

elliotcm:
@cbaines updated, thanks for the reminder that it existed

PR #4140, opened 2018-06-19
Use edition services to handle attachment replacement and deletion

barrucadu:
Drops the separate event streams for attachments/attachment data, and just re-uses the existing edition services.  Deleting and replacing attachments now works: I was able to successfully run through the checklist on integration.

I think deploying this requires a hard restart, as it changes an initialisation file.  We also need to run the `asset_manager:update_attachment_metadata` task after deploying this, to fix up any assets whose state has come out of sync.

---

There is one bit of residual weirdness, visiting a deleted asset via `draft-origin` serves it via whitehall successfully, even though it is deleted.  We couldn't figure out why.  The asset-manager correctly serves a 404 via `draft-assets` and `assets-origin`

---

[Trello card](https://trello.com/c/ukWjB3Lq/247-figure-out-fix-whitehall-not-syncing-asset-deletion-replacement-to-asset-manager)


config/initializers/edition_services.rb
```diff
@@ -15,6 +15,14 @@
         .new(attachment.attachment_data)
         .update!
     end
+    Attachment.where(attachable_id: edition.id).find_each do |attachment|
+      next unless attachment.attachment_data
+      AttachmentData.where(replaced_by_id: attachment.attachment_data.id).find_each do |attachment_data|
```
barrucadu: This nested query makes me uneasy.

thomasleese: I think you should be able to do something like:

```rb
AttachmentData.joins(:attachment).where(attachment: { attachable_id: edition.id }, replaced_by_id: attachment.attachment_data.id).find_each do |attachment_data|
  ...
```

app/controllers/admin/attachments_controller.rb
```diff
@@ -188,4 +192,9 @@ def save_attachment
       end
     end
   end
+
+  def run_updater
```
thomasleese: I would call this `update_attachments` and I reckon we don't need to memoise the variable we can just do:

```rb
ServiceListeners::AttachmentUpdater.new(attachable.reload).update!
```

app/services/service_listeners/attachment_updater.rb
```diff
@@ -0,0 +1,67 @@
...
+    def initialize(attachable)
+      @attachable = attachable
+    end
+
+    def update!
+      # if an attachment references an attachable, the attachable may
+      # not reference the attachment: for example, deleted attachments
+      # reference their attachable, but the attachable does not
+      # reference them.
+      @attachable.attachables.each do |attachable|
```
thomasleese: If you add an `attr_reader :attachable` just below the `private` you can access this without the `@`.

app/services/service_listeners/attachment_updater.rb
```diff
@@ -0,0 +1,67 @@
...
+        Attachment.where(attachable_id: attachable.id).find_each do |attachment|
```
thomasleese: I reckon you should be able to do `Attachment.where(attachable: attachable.attachables)` here and ActiveRecord will turn that into a SQL `attachable_id IN (...)` query.

PR #4143, opened 2018-06-20
Make all publishers tag to the topic taxonomy [DO NOT MERGE]

leenagupte:
Trello: https://trello.com/c/7k55NmyU

Run the `publishing_e2e_tests` against this PR: https://github.com/alphagov/publishing-e2e-tests/pull/241

## Changes
All publishers, regardless of organisation, are routed to the topic tagging page
So the flow for all publishers should be:

Edit page -> Topic tagging page -> Review page

OR

Edit page -> Topic tagging page -> Legacy Tagging page -> Review page

Tagging to Worldwide is unaffected


app/services/taxons_to_legacy_associations_tagging.rb
```diff
@@ -54,8 +54,20 @@ def legacy_mapping_for_taxons(selected_taxons)
...
+  def tagged_to_policies?(edition)
+    defined?(edition.policy_content_ids) && edition.policy_content_ids.any?
+  end
+
+  def tagged_to_policy_areas?(edition)
+    defined?(edition.topic_ids) && edition.topic_ids.any?
+  end
+
+  def tagged_to_specialist_sectors?(edition)
+    defined?(edition.specialist_sectors) && edition.specialist_sectors.any?
```
cbaines: Instead of splitting this out in to 3 methods, you could wrap the `any?` call in `.try(...)`, e.g. `edition.try(:specialist_sectors).try(:any?)`. I can't quite decide if it's better, but I thought I'd mention it.

app/views/admin/base/_policy_area_fields.html.erb
```diff
@@ -1,6 +1,6 @@
 <% cache_if edition.topic_ids.empty?, taggable_topics_cache_digest do %>
   <fieldset class="edition-topic-fields">
-    <%= form.label :topic_ids, 'Policy Areas', required: edition.must_be_tagged_to_policy_area? %>
```
cbaines: Haha, this will be the source of the infamous red *

leenagupte: Yes! I forgot I removed it in this story.

app/services/taxons_to_legacy_associations_tagging.rb
leenagupte: I think if there where's only one `try` I'd agree with you, but think that for readability I'd still split it into three methods.

cbaines: Sure :)

lib/taxonomy/publishing_api_adapter.rb
```diff
@@ -33,6 +31,8 @@ def level_one_taxons
 
     def expanded_links_hash(content_id)
       publishing_api_with_huge_timeout.get_expanded_links(content_id, with_drafts: false).to_h
+    rescue GdsApi::HTTPNotFound
```
1pretz1: Couldn't we add world taxon test data ? Like in this commit https://github.com/alphagov/whitehall/pull/4143/commits/2120c0f230f7af290b580441913e2a6883994efc

leenagupte: My first thought was to do that, but the worldwide taxonomy isn't being used in the e2e tests. Also it just felt better to make the code a bit more robust and not throw unexpected errors because another taxonomy isn't available.

1pretz1: Okay dokay 😃

cbaines:
I've added `[DO NOT MERGE]` to the title, as I think this is currently waiting for Q2, so hopefully should be deployed the week after next.

PR #4147, opened 2018-06-21
Add more details to organisations presenter

rubenarakelyan:
This commit makes the following changes to the organisations presenter:

* Adds `secondary_corporate_information_pages`, which is an optional string that contains translated sentences with links to other corporate information pages that aren’t included in the primary list
* Adds Govspeak rendering for the organisation summary
* Adds a link to the “about us” corporate information page for executive offices and the Civil Service only to mirror current behaviour
* Only adds links to corporate reports and transparency data if these have been published by the relevant organisation
* Fixes the image URL for promotional features

### DIFF
--- a/app/presenters/publishing_api/organisation_presenter.rb
+++ b/app/presenters/publishing_api/organisation_presenter.rb
@@ -70,6 +70,7 @@ module PublishingApi
         }.compact!,
         foi_exempt: foi_exempt,
         ordered_corporate_information_pages: corporate_information_pages,
+        secondary_corporate_information_pages: secondary_corporate_information_pages,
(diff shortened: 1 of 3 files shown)


app/presenters/publishing_api/organisation_presenter.rb
```diff
@@ -96,7 +97,7 @@ def summary
```
sihugh: I don't suppose you could Govspeak render the `summary` in this too could you?  We'll only need to republish :-(

rubenarakelyan: Done

PR #4149, opened 2018-06-22
Reslug Highways Agency road proposals

rubenarakelyan:
To update the name to Highways England.

Zendesk: https://govuk.zendesk.com/agent/tickets/2829919


cbaines:
Note that there are rake tasks in Whitehall to do this, which might be quicker.

PR #4150, opened 2018-06-22
Reslug “Exporting to Belgium”

rubenarakelyan:
To get rid of the `--2` suffix.

Zendesk: https://govuk.zendesk.com/agent/tickets/2854759


rubenarakelyan:
Using the rake task instead

PR #4152, opened 2018-06-22
Redirect draft asset requests to the draft assets host

barrucadu:
In https://github.com/alphagov/govuk-puppet/pull/7701, we're merging draft-whitehall-frontend with whitehall-admin, to avoid the special-case behaviour of serving asset files directly from NFS.  This means that the asset host redirects in whitehall now need to redirect to the draft-assets host if the request came from draft-origin.

---

[Trello card](https://trello.com/c/sMwEMvsv/265-redirect-draft-origin-assets-to-assets-origin)

### DIFF
--- a/app/controllers/asset_manager_redirect_controller.rb
+++ b/app/controllers/asset_manager_redirect_controller.rb
@@ -1,6 +1,10 @@
 class AssetManagerRedirectController < ApplicationController
   def show
-    asset_host = URI.parse(Plek.new.public_asset_host).host
-    redirect_to host: asset_host
+    asset_url = Plek.new.public_asset_host
+    if request.host.start_with?('draft-')
+      asset_url = Plek.new.external_url_for('draft-assets')
+    end
+
+    redirect_to host: URI.parse(asset_url).host
   end
 end
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -3,9 +3,15 @@ class AttachmentsController < BaseAttachmentsController
 
   def show
     unless user_signed_in?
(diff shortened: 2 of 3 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -3,9 +3,15 @@ class AttachmentsController < BaseAttachmentsController
 
   def show
     unless user_signed_in?
-      asset_host = URI.parse(Plek.new.public_asset_host).host
+      asset_url = if request.host.start_with?('draft-')
```
thomasleese: Could we extract this block of code into a common class or method as it's the same as the code in the `AssetManagerRedirectController` above and would be good to have it testable in one place?

app/controllers/asset_manager_redirect_controller.rb
```diff
@@ -1,6 +1,10 @@
 class AssetManagerRedirectController < ApplicationController
   def show
-    asset_host = URI.parse(Plek.new.public_asset_host).host
-    redirect_to host: asset_host
```
thomasleese: Does this mean currently that if you were trying to access a draft asset you would always get redirected to the live version?

barrucadu: If your request made it to this controller, yes; but this controller doesn't match whitehall assets yet: https://github.com/alphagov/whitehall/blob/f35fd6fb76e726fbd5e8e94037807e2af3f1dce8/config/routes.rb#L425-L429

app/controllers/attachments_controller.rb
thomasleese: Yeah true, probably not worth it.

PR #4167, opened 2018-06-28
Disable asset workers & block publishing an edition until all assets have uploaded to the asset-manager

barrucadu:
Doing attachment things to an edition (including publishing) triggers workers to run on all of its attachments, updating metadata.  However, if the asset has not yet made its way over to the asset-manager, then these workers will fail with an `AssetManagerWorkerHelper::AssetManagerAssetNotFound` exception.  This isn't a problem as such, because the workers will be retried, but it does lead to our logs and reporting filling up with these harmless errors.

This PR disables triggering those workers until assets have made it across.  To prevent this from being a problem, publishing is now blocked until this point.

**Review:** It would be good to deploy this to integration and do some asset stuff, to verify that the number of errors has been reduced.

**Downsides:** This is a new message in the UI, and it exposes an implementation detail to users.  We should be confident that this isn't going to confuse people - do we need to let the content community know?  On the other hand, uploading to the asset manager is really fast, it's just that the erroring workers are somewhat faster; so people usually won't even notice the delay.

---

[Trello card](https://trello.com/c/D5oLPmZR/266-add-a-pending-state-to-whitehall-attachmentdatas-while-state-is-copied-over-to-asset-manager)

### DIFF
--- a/app/models/attachable.rb
+++ b/app/models/attachable.rb
@@ -75,6 +75,14 @@ module Attachable
     true
   end
 
+  def uploaded_to_asset_manager?
+    attachments.each do |attachment|
+      next unless attachment.attachment_data
+      return false unless attachment.attachment_data.uploaded_to_asset_manager_at
+    end
+    true
+  end
+
   def allows_attachments?
     true
   end
--- a/app/models/attachable.rb
+++ b/app/models/attachable.rb
@@ -76,11 +76,10 @@ module Attachable
   end
 
   def uploaded_to_asset_manager?
-    attachments.each do |attachment|
-      next unless attachment.attachment_data
-      return false unless attachment.attachment_data.uploaded_to_asset_manager_at
-    end
-    true
+    attachments
+      .map(&:attachment_data)
+      .compact
+      .all?(&:uploaded_to_asset_manager?)
   end
 
   def allows_attachments?
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -108,6 +108,10 @@ class AttachmentData < ApplicationRecord
     ServiceListeners::AttachmentUpdater.update_attachment_data! self
   end
 
+  def uploaded_to_asset_manager?
+    uploaded_to_asset_manager_at.present?
+  end
+
   def deleted?
     significant_attachment(include_deleted_attachables: true).deleted?
   end
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -105,7 +105,7 @@ class AttachmentData < ApplicationRecord
 
   def uploaded_to_asset_manager!
     update!(uploaded_to_asset_manager_at: Time.zone.now)
-    ServiceListeners::AttachmentUpdater.update_attachment_data! self
+    ServiceListeners::AttachmentUpdater.call(attachment_data: self)
   end
 
   def uploaded_to_asset_manager?
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -103,6 +103,11 @@ class AttachmentData < ApplicationRecord
     end
   end
 
+  def uploaded_to_asset_manager!
+    update!(uploaded_to_asset_manager_at: Time.now)
+    ServiceListeners::AttachmentUpdater.update_attachment_data! self
+  end
+
   def deleted?
     significant_attachment(include_deleted_attachables: true).deleted?
   end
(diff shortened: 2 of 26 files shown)


app/models/attachable.rb
```diff
@@ -75,6 +75,14 @@ def valid_virus_state?
     true
   end
 
+  def uploaded_to_asset_manager?
+    attachments.each do |attachment|
+      next unless attachment.attachment_data
+      return false unless attachment.attachment_data.uploaded_to_asset_manager_at
```
thomasleese: Might be nice to have a `AttachmentData::uploaded_to_asset_manager?` method to use here for clarity as well, then you might be able to do something like:

```ruby
attachments
  .map { |attachment| attachment.attachment_data }
  .compact
  .all? { |attachment_data| attachment_data.uploaded_to_asset_manager? }
```

app/services/service_listeners/attachment_updater.rb
```diff
@@ -5,56 +5,57 @@ def initialize(attachable)
...
-
-        AttachmentData.where(replaced_by_id: attachment.attachment_data.id).find_each do |attachment_data|
-          replacement_id_updater attachment_data
-        end
+        self.class.update_attachment_data! attachment.attachment_data
       end
     end
 
-  private
+    def self.update_attachment_data!(attachment_data)
```
thomasleese: It might be nicer instead to have the interface for this class to work something like this:

```ruby
ServiceListeners::AttachmentUpdater.call(attachable: attachable)
ServiceListeners::AttachmentUpdater.call(attachment_data: attachment_data)
```

db/migrate/20180626121716_add_uploaded_to_asset_manager_at_time_to_attachment_data.rb
```diff
@@ -0,0 +1,5 @@
+class AddUploadedToAssetManagerAtTimeToAttachmentData < ActiveRecord::Migration[5.0]
+  def change
+    add_column :attachment_data, :uploaded_to_asset_manager_at, :datetime, default: -> { 'CURRENT_TIMESTAMP' }
```
thomasleese: Should this definitely have a default? This would mean that creating new instances of this object without specifying this would have this field set. It might be better to leave the default as null but include a second part of this migration which sets the time to now for all existing attachment datas.

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -21,6 +21,14 @@ def perform(file_path, legacy_url_path, draft = false, model_class = nil, model_
     end
 
     asset_manager.create_whitehall_asset(asset_options)
+
+    if model
+      Attachment.where(attachable: model.attachables).find_each do |attachment|
+        next unless attachment.attachment_data
```
thomasleese: Can we do `Attachment.where(attachable: model.attachables).where.not(attachment_data: nil)` in the query instead? Then we could have a scope that's like `.has_attachment_data`.

db/migrate/20180626121716_add_uploaded_to_asset_manager_at_time_to_attachment_data.rb
barrucadu: Yes, I think that's the better thing to do.  In fact, setting a default here is definitely wrong - but we do want all existing `AttachmentData`s to have a default.

app/models/attachment_data.rb
```diff
@@ -103,6 +103,15 @@ def replace_with!(replacement)
     end
   end
 
+  def uploaded_to_asset_manager!
+    update!(uploaded_to_asset_manager_at: Time.zone.now)
+    ServiceListeners::AttachmentUpdater.update_attachment_data! self
+  end
+
+  def uploaded_to_asset_manager?
+    uploaded_to_asset_manager_at != nil
```
thomasleese: Surprised the linter didn't make you do `uploaded_to_asset_manager_at.present?`.

app/services/service_listeners/attachment_updater.rb
```diff
@@ -1,60 +1,62 @@
...
   class AttachmentUpdater
-    def initialize(attachable)
-      @attachable = attachable
+    def self.call(attachable: nil, attachment_data: nil)
+      update_attachable! attachable if attachable
+      update_attachment_data! attachment_data if attachment_data
     end
 
-    def update!
+    def self.update_attachable!(attachable)
```
thomasleese: Can these be private now?

PR #4171, opened 2018-06-29
Revert "Disable asset workers & block publishing an edition until all assets have uploaded to the asset-manager"

thomasleese:
We've been unable to test this and it's blocking deploys of Whitehall.

Reverts alphagov/whitehall#4167


sihugh:
🚀 Thanks all 👏

PR #4182, opened 2018-07-02
 Make asset-manager attachment workers unique

thomasleese:
This should reduce the number of workers we're seeing in the queue.

[Trello Card](https://trello.com/c/R0eXp4Di/281-think-about-asset-sidekiq-stuff)


barrucadu:
We've had issues with this gem (I think) using up a lot of memory in Redis, will that be a problem here?

thomasleese:
@elliotcm Until we use Redis > 3.0.

elliotcm:
@thomasleese Even on our own fork?

thomasleese:
@elliotcm Our fork is the last sidekiq-unique-jobs 4.x release (which was the last one to support Redis < 3.0) with a cherry picked commit from a more recent version to help with clearing out the unique jobs key.

PR #4185, opened 2018-07-03
[WIP] Fix sending nil attribute changes to the asset-manager

barrucadu:
When we unpublish a thing, a redirect URL is set in the asset-manager.  This is all fine and dandy, but unpublished things can be re-published too.  In that case, we should *unset* the redirect URL, which we do by setting it to `nil`.

Unfortunately, the `asset_manager.update_asset` method uses multipart/form-data to send changes, and so it silently drops `nil` attributes.  So instead use the new `asset_manager.update_asset_attributes` which sends data as json instead.

---

[Trello card](https://trello.com/c/hoC9DL84/290-investigate-asset-manager-redirecturl-weirdness)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,9 +63,7 @@ gem 'validates_email_format_of'
 gem 'whenever', '~> 0.10.0', require: false
 
 # rubocop:disable Bundler/DuplicatedGem
-if ENV['GDS_API_ADAPTERS_DEV']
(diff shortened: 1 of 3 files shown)


barrucadu:
I was mistaken, this isn't a http request thing, it's an asset-manager thing: https://github.com/alphagov/asset-manager/commit/b9e3a6a3cae336069913959a67d928bcb3889beb

PR #4189, opened 2018-07-04
Revert " Make asset-manager attachment workers unique"

barrucadu:
Reverts alphagov/whitehall#4182

This is failing on master, even though the PR itself passed: https://ci.integration.publishing.service.gov.uk/job/whitehall/job/master/13592/console


barrucadu:
Not sure yet, I need to look into it.  It's definitely worrying that removing apparently-identical jobs causes a problem, though.

PR #4191, opened 2018-07-06
Revert "Do not do redundant Asset Manager jobs"

barrucadu:
Reverts alphagov/whitehall#4190

This is incorrect, as changing the draft state doesn't change the `updated_at`.


barrucadu:
Not sure why the tests didn't catch this, I'll look into that when this is merged.

PR #4193, opened 2018-07-09
Update govuk_sidekiq and gds-api-adapters

barrucadu:
govuk_sidekiq will hopefully fix the asset upload issue.  This PR also deletes an old test which the change broke:

> We believe this test was introduced at the same time as
> govuk_request_id to ensure that the change didn't break any
> pre-existing scheduled publication jobs (which would not have the
> parameter set).
>
> There aren't any scheduled publication jobs from 2016 still around, so
> the test isn't actually achieving anything useful any more.

---

[Trello card](https://trello.com/c/SRnVqyIm/289-investigate-asset-manager-edge-cases)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -154,7 +154,7 @@ GEM
     fugit (1.1.1)
       et-orbi (~> 1.1, >= 1.1.1)
       raabro (~> 1.1)
-    gds-api-adapters (52.5.1)
+    gds-api-adapters (52.6.0)
       addressable
       link_header
       lrucache (~> 0.1.1)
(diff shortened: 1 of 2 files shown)


barrucadu:
Yep, tomorrow

elliotcm:
@barrucadu Is the code that was testing also removed?

barrucadu:
@elliotcm The code that it's testing is, now, the logic in govuk_sidekiq which adds those headers if appropriate, which has its own tests.

barrucadu:
Given that govuk_sidekiq 3.0.2 is imminent, I'll close this and open a new PR.

PR #4203, opened 2018-07-12
Use govuk_app_config health checks

thomasleese:
This makes the health check code simpler to understand and adds new health checks on the database and Sidekiq connection.

[Trello Card](https://trello.com/c/iuPQ6P88/318-add-common-parameterised-healthcheck-classes)

### DIFF
--- a/app/models/healthcheck/scheduled_publishing.rb
+++ b/app/models/healthcheck/scheduled_publishing.rb
@@ -1,0 +1,39 @@
+module Healthcheck
+  class ScheduledPublishing
+    def name
+      :scheduled_queue
+    end
+
+    def status
+      queue_size_matches_edition_count? ? :ok : :warning
+    end
+
+    def details
+      {
+        queue_size: queue_size,
+        edition_count: edition_count,
+      }
+    end
+
+    def message
+      "#{edition_count} scheduled edition(s); #{queue_size} job(s) queued"
+    end
+
+  private
+
+    def queue_size_matches_edition_count?
+      queue_size == edition_count
+    end
+
+    def queue_size
+      @queue_size ||= ScheduledPublishingWorker.queue_size
+    end
+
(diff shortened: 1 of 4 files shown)


app/models/healthcheck/scheduled_publishing.rb
```diff
@@ -0,0 +1,39 @@
+module Healthcheck
+  class ScheduledPublishing
```
cbaines: This seems more like a service to me than a model, given it's not saved anywhere?

thomasleese: Yeah, we thought about this when we decided that all the health checks would go in models in email-alert-api. I would argue it's not a service because it doesn't "do anything". It may not get saved, but it's "data to be read" so I think it sits better as a model.

cbaines: > Yeah, we thought about this when we decided that all the health checks would go in models in email-alert-api. I would argue it's not a service because it doesn't "do anything". It may not get saved, but it's "data to be read" so I think it sits better as a model.

I guess if I was using the "data to be read" criteria, it would be hard to see any function that returns something not to be considered part of a model. I'm not sure if there are any existing models in Whitehall that don't inherit either directly or indirectly from ApplicationRecord, but I think it would be unlikely, and while I don't think there is much guidance on what goes in which directories, relationship to ApplicationRecord for models is a good criteria in my mind.

One example of a existing Whitehall service that provides some data is the EditionTaxonsFetcher. It's just the first I found, I don't know if its a good example, and while it doesn't "do anything" as in have some effect on some state, it encapsulates fetching some data.

cbaines: Anyway, I'm not too fussed where it ends up, it's all on the LOAD_PATH.

PR #4205, opened 2018-07-13
Set uploaded_to_asset_manager timestamp in a worker

barrucadu:
There exist a few `AttachmentData`s which do not have an `uploaded_to_asset_manager_at` timestamp, and which don't correspond to any `AssetManagerCreateWhitehallAssetWorker` exceptions in the logs.  The worker succeeded, but failed to set the timestamp.  This means that the `AttachmentData` for the given `legacy_url_path` was not found.

Having looked at a few of these assets, the `AttachmentData` in Whitehall and the `WhitehallAsset` in Asset Manager were created in the very same second.

Given that there doesn't appear to be any direct causal link between creating the `AttachmentData` and running the worker, I suspect there is a race condition between the two: if the worker runs to completion before the `AttachmentData` is created (which it could feasibly do for small files when the Sidekiq queue is otherwise empty), then the `AttachmentData` will not exist when the worker tries to set its `uploaded_to_asset_manager_at`.

This commit splits out the setting of the timestamp to a separate worker, which throws an exception (and so gets retried) if it fails to find the `AttachmentData`.  This is a bit of noise, which is sad, but
it does only appear to affect a very small proportion of uploads.

### DIFF
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -23,18 +23,10 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
     asset_manager.create_whitehall_asset(asset_options)
 
     if model
-      # sadly we can't just search for url, because it's a magic
-      # carrierwave thing not in our model
-      Attachment.where(attachable: model.attachables).where.not(attachment_data: nil).find_each do |attachment|
-        # 'attachment.attachment_data' can still be nil even with the
-        # check above, because if the 'attachment_data_id' is non-nil
-        # but invalid, the 'attachment_data' will be nil - and the
-        # generated SQL only checks if the 'attachment_data_id' is
-        # nil.
-        if attachment.attachment_data && attachment.attachment_data.url == legacy_url_path
-          attachment.attachment_data.uploaded_to_asset_manager!
-        end
-      end
+      # The AttachmentData we want to set the timestamp on may not
+      # exist yet, so create a worker to do it after a very short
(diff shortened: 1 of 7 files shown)


app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -23,18 +23,10 @@ def perform(file_path, legacy_url_path, draft = false, model_class = nil, model_
...
-        # generated SQL only checks if the 'attachment_data_id' is
-        # nil.
-        if attachment.attachment_data && attachment.attachment_data.url == legacy_url_path
-          attachment.attachment_data.uploaded_to_asset_manager!
-        end
-      end
+      # The AttachmentData we want to set the timestamp on may not
+      # exist yet, so create a worker to do it after a very short
+      # delay.  The worker will retry if it still doesn't exist.
+      SetUploadedToAssetManagerWorker.perform_in(0.5.seconds, model_class, model_id, legacy_url_path)
```
thomasleese: I wonder, rather than having this as a separate worker, can we call the `AssetManagerCreateWhitehallAssetWorker` with a 0.5 second delay?

barrucadu: And then have the `AssetManagerCreateWhitehallAssetWorker` throw an exception if it can't find the `AttachmentData`?  I feel like that's mixing too many concerns into the same worker.

thomasleese: Hmm, that's true, I hadn't considered that.

PR #4207, opened 2018-07-13
[DO NOT MERGE] Make asset-manager attachment workers unique 

thomasleese:
This should reduce the number of workers we're seeing in the queue.

This is a second attempt of #4182 which didn't work because it made the tests flaky. This should be solved by 9ef1cab.

[Trello Card](https://trello.com/c/1sd6Ve2a/292-investigate-and-reduce-daily-asset-job-spike)


barrucadu:
To clarify, `unique: :until_executing` means that if one `AssetManagerAttachmentDataWorker` is enqueued, and another is enqueued after the first starts executing but before it finishes, the second will still be enqueued?

I think that's the correct behaviour, because the workers look up the current state of the attachment when they run.  So you could imagine this sequence of events:

1. `AssetManagerAttachmentDataWorker(id)` is enqueued
2. starts running
3. It looks up the state of the attachment
4. The state changes!
5. Another `AssetManagerAttachmentDataWorker(id)` is enqueued
6. The first worker sets the state in the asset manager to the now-outdated state
7. At some later point the second worker sets the state in the asset manager to the now-correct state

If that second worker in step 5 were not enqueued, the asset manager would not get the updated state.

thomasleese:
@barrucadu Yes, that is my understanding of `until_executing` as well. We don't want to use `until_and_while_executing` as, with the example you give, it would mean we might miss out on state changes happening while a worker is running.

barrucadu:
Actually `until_and_while_executing` looks fine to me - it says that the same job can be pushed to the queue as soon as the first starts executing, but the second job won't be executed until after the first finishes.  I think it's `until_executed` and `until_expired` we don't want.

thomasleese:
@barrucadu Yep, you're right. I just spotted that too and edited my comment as you posted that!

thomasleese:
@barrucadu I'm actually thinking now that it would be better to use `until_and_while_executing` as it would solve this scenario:

1) A worker starts and loads the state of an asset
2) The state of the asset changes
3) A second worker starts and loads the new state of the asset
4) The first worker is delayed talking to the asset-manager for whatever reason (network issues?)
5) The second worker finishes, sending the new state to asset-manager
6) The first worker finishes, sending the old state to asset-manager

What do you think?

barrucadu:
Yep, that sounds reasonable to me.  Aren't distributed systems wonderful?

thomasleese:
This doesn't seem to be working properly.

PR #4216, opened 2018-07-17
Remove attachment metadata update worker

thomasleese:
This was used for asset migration but isn't required anymore.

We've already got a PR which removes this code but I kept getting confused with this worker and the very similar "data update worker", so I thought we could just remove it now.


config/sidekiq.yml
```diff
@@ -10,7 +10,6 @@
   - bulk_republishing
   - sync_checks
   - link_checks
-  - asset_migration
```
barrucadu: `lib/migrate_assets_to_asset_manager.r` still uses the `asset_migration` queue.  I think if the queue is being removed, that should be too.

thomasleese: Good catch, I've put the queue back though as I don't want to delete too much in one PR.

PR #4225, opened 2018-07-18
Group attachment updates together

thomasleese:
This is the final part after #4222. It refactors how the Attachment Update workers are designed to reduce the number of calls to Asset Manager.

Previously each subupdate of a larger asset update would required a database search and at least one request to the Asset Manager, for each part of the asset (access limiting, draft status, link header, replacement ID, redirect URL). Now there will only be at most two requests to Asset Manager for each asset in full.

[Trello Card](https://trello.com/c/1sd6Ve2a/292-investigate-and-reduce-daily-asset-job-spike)

### DIFF
--- a/app/services/asset_manager/attachment_updater.rb
+++ b/app/services/asset_manager/attachment_updater.rb
@@ -1,0 +1,34 @@
+class AssetManager::AttachmentUpdater
+  def self.call(
+    attachment_data,
+    access_limited: false,
+    draft_status: false,
+    link_header: false,
+    redirect_url: false,
+    replacement_id: false
+  )
+    updates = []
+
+    updates += AccessLimitedUpdates.call(attachment_data).to_a if access_limited
+    updates += DraftStatusUpdates.call(attachment_data).to_a if draft_status
+    updates += LinkHeaderUpdates.call(attachment_data).to_a if link_header
+    updates += RedirectUrlUpdates.call(attachment_data).to_a if redirect_url
+    updates += ReplacementIdUpdates.call(attachment_data).to_a if replacement_id
+
+    self.combined_updates(updates).each(&:call)
+  end
+
+  def self.combined_updates(updates)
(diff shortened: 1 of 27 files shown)


app/services/asset_manager/attachment_updater.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    self.combined_updates(updates).each(&:call)
+  end
+
+  def self.combined_updates(updates)
+    grouped_updates = updates.group_by do |update|
+      [update.attachment_data, update.legacy_url_path]
+    end
+
+    grouped_updates.map do |attachment_data_and_legacy_url_path, updates_to_combine|
```
elliotcm: A slightly neater thing here, you can deconstruct this param:
```
grouped_updates.map do |(attachment_data, legacy_url_path), updates_to_combine|
```

config/initializers/govuk_errors.rb
```diff
@@ -1,3 +0,0 @@
-GovukError.configure do |config|
-  config.excluded_exceptions << 'AssetManager::AttachmentReplacementIdUpdater::AssetNotFound'
```
elliotcm: Could you clarify in the commit message why we're no longer excluding this?

app/services/asset_manager/attachment_updater.rb
thomasleese: Good idea, have amended!

config/initializers/govuk_errors.rb
thomasleese: Done:

> I've also removed ignoring the old 
`AssetManager::AttachmentReplacementIdUpdater::AssetNotFound`
exception in Sentry as this exception has been merged together
with another and since we don't currently ignore the one it was
merged into it would be best to not start hiding more than we do
currently. If we decided that we don't need to see this exception
(that I suspect is the case) we can add it to the ignore list in
a later PR.

PR #4233, opened 2018-07-23
Add a link to taxonomy guidance in the tagging page

koetsier:
Trello: https://trello.com/c/ryxlf7gr/48-change-content-at-top-of-whp-tagging-interface-and-add-link-to-guidance

image

### DIFF
--- a/app/views/admin/edition_tags/edit.html.erb
+++ b/app/views/admin/edition_tags/edit.html.erb
@@ -6,9 +6,10 @@
 <div class="row">
   <div class="col-md-12">
     <h2>Topics (new taxonomy)</h2>
-    <p>Topics categorise content on GOV.UK by subject area.</p>
-    <p>Choose the topic(s) that best describe what this content is about.</p>
+    <p>Topics group content based on what it’s about.</p>
+    <p>Choose the topic or topics that best describe what this content is about.</p>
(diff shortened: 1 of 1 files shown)


app/views/admin/edition_tags/edit.html.erb
```diff
@@ -6,9 +6,10 @@
 <div class="row">
   <div class="col-md-12">
     <h2>Topics (new taxonomy)</h2>
-    <p>Topics categorise content on GOV.UK by subject area.</p>
-    <p>Choose the topic(s) that best describe what this content is about.</p>
+    <p>Topics group content based on what it’s about.</p>
+    <p>Choose the topic or topics that best describe what this content is about.</p>
     <p>You can use the whole taxonomy. There's no limit to the number of topics you can choose.</p>
+    <p><a href="https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#add-topic-tagging">Find out more about tagging to topics.</a></p>
```
cbaines: Link doesn't work, I assume that's because the guidance hasn't been published yet.

koetsier: That is why it can't be merged yet 👍

sistephens: Yup. I've just published the guidance though, so we can unblock this now 😃

PR #4237, opened 2018-07-26
Mark attachments as draft in Asset Manager when a document is unpublished

brucebolt:
Currently, after a document is unpublished, any attachments are marked as 'not draft' in Asset Manager.  This means it could be possible for the attachments to be accessed by anyone by guessing the URL.

This changes the logic is mark attachments as draft in Asset Manager once a document has been unpublished.

https://trello.com/c/PlPb1iNY/337-fix-attachments-on-unpublished

### DIFF
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -21,6 +21,8 @@ class AttachmentData < ApplicationRecord
 
   attr_accessor :attachable
 
+  attribute :present_at_unpublish, :boolean, default: false
+
   def filename
     url && File.basename(url)
   end
@@ -138,6 +140,10 @@ class AttachmentData < ApplicationRecord
 
   def unpublished_edition
     last_attachable.unpublished_edition
+  end
+
+  def present_at_unpublish?
+    self[:present_at_unpublish]
   end
 
   def replaced?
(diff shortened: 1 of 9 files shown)


app/models/attachment_data.rb
```diff
@@ -117,7 +117,11 @@ def deleted?
   end
 
   def draft?
-    !significant_attachable.publicly_visible?
+    if unpublished?
+      true
```
barrucadu: I don't think this is correct.

This would mark attachments which existed when the document was unpublished as draft, which then would send people to the draft-assets host when they tried to access them, rather than redirecting to the "sorry, this page was unpublished" page.

brucebolt:
I have now refactored the PR (quite considerably) to do two things:
1. Keep attachments added before unpublishing as not draft in Asset Manager and redirect to the information page,
2. Mark attachments added after unpublishing as draft in Asset Manager.

There is a new attribute on all attachments (`present_at_unpublish`) which is set to `true` on all attachments when a document is unpublished.  This is reverted to `false` when the document is published, this is to ensure the behaviour recurs if the document is unpublished for a second time.

brucebolt:
That's a good point and it does seem like a redundant change to be making.  I have updated the PR to remove the modification of this value to `false` during a publish event.

PR #4242, opened 2018-07-30
Render announcements on people page using rummager 

tijmenb:
This converts the "Announcements" section on people pages to use the search API. This means that content published by other applications (like upcoming content-publisher) will be surfaced on this page too.

The short/medium term plan for these pages is that they're moved to the collections app. To make this easier, I've chosen to use some components to rebuild the interface. This way I can avoid refactoring whitehall code to make use of the search data structure instead of the data from the database (the `feeds` partial is tightly coupled to the active record models). It hopefully makes the interface better as well.

## Before

screen shot 2018-08-01 at 12 43 34

## After 

screen shot 2018-08-01 at 12 42 22

(The data is different because this uses production data, not my local data).

https://trello.com/c/vYdbh3TQ

### DIFF
--- a/app/presenters/person_presenter.rb
+++ b/app/presenters/person_presenter.rb
@@ -30,10 +30,33 @@ class PersonPresenter < Whitehall::Decorators::Decorator
   end
 
   def announcements
-    announcements =
-      model.published_speeches.with_translations(I18n.locale).limit(10).map { |s| SpeechPresenter.new(s, context) } +
-      model.published_news_articles.with_translations(I18n.locale).limit(10).map { |na| NewsArticlePresenter.new(na, context) }
-    announcements.sort_by { |a| a.public_timestamp.to_datetime }.reverse[0..9]
+    search_results = Whitehall.search_client.search(
+      filter_people: model.slug,
+      count: 10,
+      order: "-public_timestamp",
+      reject_content_purpose_supergroup: "other",
+      fields: %w[title link content_store_document_type public_timestamp]
+    )["results"]
+
+    search_results.map do |item|
+      metadata = {}
+
+      if item["public_timestamp"]
+        metadata[:public_updated_at] = Date.parse(item["public_timestamp"])
+      end
+
+      if item["content_store_document_type"]
+        metadata[:document_type] = item["content_store_document_type"].humanize
+      end
+
+      {
+        link: {
+          text: item["title"],
+          path: item["link"]
+        },
+        metadata: metadata
+      }
+    end
   end
 
   def speeches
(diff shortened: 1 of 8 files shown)


app/presenters/person_presenter.rb
```diff
@@ -30,10 +30,37 @@ def has_policy_responsibilities?
   end
 
   def announcements
-    announcements =
-      model.published_speeches.with_translations(I18n.locale).limit(10).map { |s| SpeechPresenter.new(s, context) } +
-      model.published_news_articles.with_translations(I18n.locale).limit(10).map { |na| NewsArticlePresenter.new(na, context) }
-    announcements.sort_by { |a| a.public_timestamp.to_datetime }.reverse[0..9]
+    search_results = Whitehall.search_client.search(
```
benthorner: I can see the old code is filtering by locale and speeches/news, but I can't see the same here. Is it implicit somehow?

app/presenters/person_presenter.rb
```diff
@@ -30,10 +30,37 @@ def has_policy_responsibilities?
...
+      filter_people: model.slug,
+      count: 10,
+      order: "-public_timestamp",
+      reject_content_purpose_supergroup: "other",
+      fields: %w[title link content_store_document_type public_timestamp]
+    )["results"]
+
+    documents = []
+
+    search_results.each do |item|
```
benthorner: Could do a map here :-).

features/people.feature
```diff
@@ -13,11 +13,8 @@ Scenario: Viewing the person page for a person
 
 Scenario: Announcements by that person over their career are shown
   Given I am an editor
-  And "Don Deputy" is the "Deputy Prime Minister" for the "Cabinet Office"
   And "Harriet Home" is the "Home Secretary" for the "Cabinet Office"
```
benthorner: Is this step necessary anymore?

test/functional/people_controller_test.rb
```diff
@@ -78,18 +78,4 @@ def stub_role_appointment(role_type, options = {})
       assert_select "a[href='/government/policies/welfare-reform']", text: "Welfare reform"
     end
   end
-
```
benthorner: I think the link is still there - perhaps we don't care about testing it anymore?

features/people.feature
tijmenb: This creates the actual person record for "Harriet".

test/functional/people_controller_test.rb
tijmenb: Hmm, maybe we do. I've put it back.

app/presenters/person_presenter.rb
tijmenb: Yeah. I totes copy-pasted this from collections.

features/people.feature
benthorner: Ah yes. One more thing - could you remove the 'over the career' bit from the scenario? It doesn't make sense anymore.

tijmenb: Done!

PR #4243, opened 2018-07-30
Redirect people atom feed to rummager-backed equivalent

tijmenb:
This redirects the atom for people to an equivalent feed that is backed by rummager. This means pages by publishing apps other than whitehall (like content publisher).

https://trello.com/c/vYdbh3TQ

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -2,15 +2,17 @@ class PeopleController < PublicFacingController
   enable_request_formats show: [:atom]
 
   def show
-    @person = PersonPresenter.new(Person.friendly.find(params[:id]), view_context)
-
     respond_to do |format|
       format.html do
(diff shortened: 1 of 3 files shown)


app/controllers/people_controller.rb
```diff
@@ -2,15 +2,17 @@ class PeopleController < PublicFacingController
...
       format.html do
+        @person = PersonPresenter.new(Person.friendly.find(params[:id]), view_context)
         set_meta_description("Biography of #{@person.name}.")
         set_slimmer_organisations_header(@person.organisations)
         set_slimmer_page_owner_header(@person.organisations.first)
       end
-      format.atom
+
+      format.atom do
+        redirect_to "/government/announcements?people[]=#{params[:id]}"
```
kevindew: does this need to be announcements.atom ?

tijmenb: 🤕

PR #4252, opened 2018-08-03
Enable Brakeman

thomasleese:
This allows us to find security vulnerabilities. I've also had to ignore most of the current warnings and fix a few others, more details about each individual case in the commits.

[Trello Card](https://trello.com/c/JyrOAef3/363-fix-sql-injection-vulnerabilities-in-govuk-applications)


rubenarakelyan:
Good luck

PR #4255, opened 2018-08-03
Remove HMRC assets controller

barrucadu:
These files are now served by the Asset Manager.

This should not be merged until the puppet changes in https://github.com/alphagov/govuk-puppet/pull/7906 have propagated.

---

[Trello card](https://trello.com/c/SFeIQ9Dp/371-serve-hmrc-assets-from-the-asset-manager)


barrucadu:
Puppet change deployed, and seems to be working.  But I'll keep an eye on kibana for 404s for a bit before deploying this.

PR #4272, opened 2018-08-08
Republish document containing ODF attachment

deborahchua:
This data migration republishes a document that was missed in the [original query](https://github.com/alphagov/whitehall/blob/master/db/data_migration/20180803160419_republish_editions_with_opendocument_attachments.rb) to republish editions that contain open document attachments (https://www.gov.uk/government/consultations/making-open-data-real).

Trello card: [Republish whitehall editions that contain OpenDocument links](https://trello.com/c/Dgt3ta3y/369-republish-whitehall-editions-that-contain-opendocument-links)

### DIFF
--- a/db/data_migration/20180808140111_republish_odf_document.rb
+++ b/db/data_migration/20180808140111_republish_odf_document.rb
@@ -1,0 +1,1 @@
+PublishingApiDocumentRepublishingWorker.perform_async(77434)


barrucadu:
What was the problem with the original query?

steventux:
@barrucadu We think the joins in the original query omitted this record due to the attachment data being wrongly associated.

PR #4273, opened 2018-08-09
Use bootsnap

cbaines:
This should help with the startup time.

Some basic benchmarking:

  time rails runner "puts 1"

Without bootsnap ("real" time for 3 runs):
  0m2.430s 0m2.408s 0m2.382s

With bootsnap
  0m1.156s 0m1.140s 0m1.133s


1pretz1:
Will approve once the gem has been updated and the gem file has been amended to use the new version.

PR #4278, opened 2018-08-10
Handle drafts synchronously

cbaines:
Rather than offloading making the put_content request to the
Publishing API to Sidekiq, handle it during the request.

This makes it easier to provide feedback to the user. You know when
rendering the response if the request has failed or succeeded, so this
enables you to tell the user that. It also means that subsequent
actions like previewing the draft are more likely to succeed, as the
draft has definately been sent to the Publishing API.

In the longer term, removing some of the less necessary layers of
indirection within Whitehall should make the code easier to understand
and maintain.


cbaines:
This is still a work in progress, as handling for the possible errors needs adding.

cbaines:
> This is still a work in progress, as handling for the possible errors needs adding.

Actually, I've got ideas for improving the error handling, but this change enables it, rather than improved error handling being a prerequisite.

PR #4284, opened 2018-08-13
Change how organisations are republished

cbaines:
Currently the organisation pages have a number of dependencies that
are not handled by the Publishing API. Change how these are handled,
to make it clearer at least in the code that the organisation is being
"republished".


cbaines:
Thanks for taking a look @rubenarakelyan :+1:

PR #4288, opened 2018-08-15
Enable uploading of vCard files

rubenarakelyan:
This commit adds support for uploading vCard files with the `.vcf` extension, as per the guidance at https://www.gov.uk/government/publications/open-standards-for-government/exchange-of-contact-information.

This has been requested by the MoD so they can easily publish contact details in an open and shareable format.

### DIFF
--- a/app/helpers/document_helper.rb
+++ b/app/helpers/document_helper.rb
@@ -104,6 +104,7 @@ module DocumentHelper
       "rtf"  => file_abbr_tag('RTF', 'Rich Text Format'),
       "sch"  => file_abbr_tag('SCH', 'XML based Schematic'),
       "txt"  => "Plain text",
+      "vcf"  => "vCard file",
       "wsdl" => file_abbr_tag('WSDL', 'Web Services Description Language'),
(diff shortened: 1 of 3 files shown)


cbaines:
Could you mention in the commit message why this is being added. Also, if possible, avoid having very long lines in the commit message (the styleguide recommends 72 characters max https://github.com/alphagov/styleguides/blob/master/git.md#structure ).

app/helpers/document_helper.rb
```diff
@@ -103,7 +103,8 @@ def humanized_content_type(file_extension)
       "rdf"  => file_abbr_tag('RDF', 'Resource Description Framework'),
       "rtf"  => file_abbr_tag('RTF', 'Rich Text Format'),
       "sch"  => file_abbr_tag('SCH', 'XML based Schematic'),
-      "txt"  => "Plain text",
+      "txt"  => 'Plain text',
```
cbaines: If you made this quoting change to avoid issues with the linter, Whitehall no longer uses diff linting, so it shouldn't complain.

rubenarakelyan: Fixed

PR #4292, opened 2018-08-23
Reslug South Georgia and the South Sandwich Islands

rubenarakelyan:
This brings the slug in line with what government-frontend expects.

Zendesk: https://govuk.zendesk.com/agent/tickets/3044783

### DIFF
--- a/db/data_migration/20180823093420_reslug_south_georgia_and_the_south_sandwich_islands.rb
+++ b/db/data_migration/20180823093420_reslug_south_georgia_and_the_south_sandwich_islands.rb
@@ -1,0 +1,8 @@
+old_slug = "south-georgia-and-south-sandwich-islands"
+new_slug = "south-georgia-and-the-south-sandwich-islands"
+
+world_location = WorldLocation.where(slug: old_slug)
+
+Whitehall::SearchIndex.delete(world_location)
+
+world_location.update_all(slug: new_slug)


binaryberry:
Does search need to know about this? I understand Whitehall still speaks to Rummager directly, in addition of the Publishing API?

rubenarakelyan:
Yes it does, so I believe updating the slug should make the new one go into rummager.

binaryberry:
Should `Whitehall::SearchIndex` also be updated to remove the old slug, to avoid it trying to index the old slug in Rummager after the Publishing API has sent the new slug to Rummager?

db/data_migration/20180823093420_reslug_south_georgia_and_the_south_sandwich_islands.rb
```diff
@@ -0,0 +1,12 @@
...
+
+WorldLocation
+  .where(slug: old_slug)
+  .update_all(slug: new_slug)
+
+  Services.publishing_api.unpublish(
+    "ad4a059c-2cd7-4c49-9d0a-ba5f4b93bd72",
+    type: "redirect",
+    alternative_path: "/world/south-georgia-and-the-south-sandwich-islands/news",
+  )
```
cbaines: Moving this content should require a "put content" and "publish" in my mind, how come there is an "unpublish" request here?

rubenarakelyan: The unpublish redirects the old news page to the new one; this doesn't happen when you reslug the world location due to a bug. See https://github.com/alphagov/whitehall/pull/4137 for more details.

cbaines: I'm not sure it's a bug, but it's definitely something that can be improved, and I'd like to improve it before working around it here.

I've got one suggestion here, to store the content ids for the news pages in a similar way to the world locations themselves: https://github.com/alphagov/whitehall/pull/4293

cbaines: > I've got one suggestion here, to store the content ids for the news pages in a similar way to the world locations themselves: #4293

With help from Bevan, this is now merged. So hopefully once it's deployed, the unpublish command will be unnecessary.

rubenarakelyan: I've removed this

rubenarakelyan:
@binaryberry Yup, good spot, I've added this in.

PR #4312, opened 2018-08-31
Fix FactoryBot deprecation warnings

steventux:
Static attributes will be removed in FactoryBot 5.0 so update factories to use dynamic attributes.

```
DEPRECATION WARNING: Static attributes will be removed in FactoryBot 5.0. Please use dynamic
attributes instead by wrapping the attribute value in a block:

external_url { "http://www.google.com" }
```

### DIFF
--- a/test/factories/about_page.rb
+++ b/test/factories/about_page.rb
@@ -1,9 +1,9 @@
 FactoryBot.define do
   factory :about_page do
     sequence(:name) { |index| "about-page-#{index}" }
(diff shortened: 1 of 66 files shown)


cbaines:
@steventux Awesome stuff. Did you manage to automate this, or did you do it manually? I tried using Rubocop as the deprecation warning suggests, but it didn't work for me.

steventux:
@cbaines No it didn't work for me either, had to do it _The Roman Way_

PR #4315, opened 2018-09-03
Allow streaming export of news

benthorner:
https://trello.com/c/ORACsrIh/172-create-a-jenkins-job-to-update-integration-content-publisher-with-whitehall-content

This means we can avoid creating temporary export files on the whitehall
machines.

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -149,10 +149,9 @@ namespace :export do
     puts result.to_json
   end
 
-  desc "Export news documents to JSON format export:news_documents ORGS=org-slug FROM=2018-07-01 OUTPUT=/path/to/file"
+  desc "Export news documents to JSON format export:news_documents e.g. ORGS=org-slug FROM=2018-07-01"
   task news_documents: :environment do
     org_slugs = ENV.fetch("ORGS", "").split(',')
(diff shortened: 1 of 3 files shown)


lib/tasks/export.rake
```diff
@@ -165,12 +164,9 @@ namespace :export do
     scope = scope.where("editions.updated_at >= ?", ENV["FROM"]) if ENV["FROM"]
     total = scope.count
 
-    document_data = scope.find_each.with_index.map do |document, index|
-      puts "#{index + 1}/#{total} exported" if ((index + 1) % 100).zero?
-      ExportNewsDocument.new(document).as_json
+    scope.find_each.with_index.each do |document, index|
+      puts "#{index + 1}/#{total} exported", STDERR if ((index + 1) % 100).zero?
+      puts ExportNewsDocument.new(document).as_json.to_json
```
emmabeynon: Seems a bit odd to chain `as_json.to_json`?

benthorner: Yeah, it's because `as_json` creates a hash, which `to_json` then converts to actual JSON. The naming is confusing - I could change it to `JSON.dumps(... .as_json)` if you want?

emmabeynon: I'm not sure it's any better, so will stop blocking this PR and approve 👍

PR #4328, opened 2018-09-10
Remove references to Policies and Policy Areas

AlanGabbianelli:
When a Whitehall user creates a new document, we allow them to associate
it with the legacy associations. As we are in the process of
unpublishing/retiring all Policies and Policy Areas, we should remove
this possibility, once they are not present.

With this, we now show only the legacy associations that are still
present.

Once all the Policies and Policy Areas are unpublished, we can modify
this and remove the conditionals introduced.

### DIFF
--- a/app/views/admin/edition_legacy_associations/edit.html.erb
+++ b/app/views/admin/edition_legacy_associations/edit.html.erb
@@ -29,11 +29,15 @@
         <legend>Associations</legend>
 
         <div class="important-associations">
-          <% if @edition.can_be_related_to_policies? %>
+          <% if @edition.can_be_related_to_policies? && Policy.all.any? %>
             <%= render 'admin/shared/policy_fields', form: form %>
+          <% else %>
+            <p>There are no Policies to tag to.</p>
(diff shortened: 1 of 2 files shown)


app/views/admin/edition_legacy_associations/edit.html.erb
```diff
@@ -29,11 +29,15 @@
         <legend>Associations</legend>
 
         <div class="important-associations">
-          <% if @edition.can_be_related_to_policies? %>
+          <% if @edition.can_be_related_to_policies? && Policy.all.any? %>
             <%= render 'admin/shared/policy_fields', form: form %>
+          <% else %>
+            <p>There are no Policies to associate</p>
```
AlanGabbianelli: Suggestions on what to write here?

cbaines: I'd just change "associate" to "tag to" (as that makes a bit more sense to me), and then let's talk to Si and Roxanne before this appears to users.

PR #4334, opened 2018-09-12
Add taxon filter

koetsier:
Adds filtering of publications, statistics and announcements by first level taxons.

The following pages now have an added 'Topic Taxon' selection box that allows filtering on a first level taxon:
- https://www.gov.uk/government/publications
- https://www.gov.uk/government/statistics
- https://www.gov.uk/government/announcements

Trello: https://trello.com/c/drJGmEYF/176-xl-replace-the-policy-area-filtering-with-topic-based-filtering-with-the-level-1-topics 

image

### DIFF
--- a/lib/taxonomy/level_one_taxons_fetcher.rb
+++ b/lib/taxonomy/level_one_taxons_fetcher.rb
@@ -1,0 +1,15 @@
+class Taxonomy::LevelOneTaxonsFetcher
+  def self.fetch
+    root_taxon = Rails.cache.fetch('taxonomy.root_taxon', expires_in: 15.minutes) do
+      Whitehall.content_store.content_item('/')
+    end
+    level_one_taxon_hashes = root_taxon.dig('links', 'level_one_taxons') || []
+    taxons = level_one_taxon_hashes.map do |level_one_taxon_hash|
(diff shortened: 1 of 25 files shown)


sistephens:
Can we change the label to just 'Topic'?

lib/taxonomy/level_one_taxons_fetcher.rb
```diff
@@ -0,0 +1,13 @@
...
+    level_one_taxon_hashes = root_taxon.dig('links', 'level_one_taxons') || []
+    taxons = level_one_taxon_hashes.map do |level_one_taxon_hash|
+      Taxonomy::Taxon.from_taxon_hash(level_one_taxon_hash)
+    end
+    live_taxons = taxons.select { |t| t.phase == 'live' }
+    live_taxons.sort_by(&:name)
+  rescue GdsApi::ContentStore::ItemNotFound
+    {}
+  end
+end
```
cbaines: Could we cache this, with some sensible timeout (e.g. 5 to 10 minutes)? Just to avoid too many unnecessary requests?

koetsier: done

cbaines: :+1:

PR #4336, opened 2018-09-13
Avoid reporting EPIPE errors from export rake task

cbaines:
This rake task sometimes errors because the receiving end of the
pipeline has crashed. Rescue the exception to avoid it propagating up
to Sentry.


cbaines:
This should avoid errors reaching Sentry: https://sentry.io/govuk/app-whitehall/issues/672937126

PR #4349, opened 2018-09-19
Fix bio truncation in search results

elliotcm:
The biography of a person is currently truncated
to the first paragraph if that person is not in
an active role, however this was not being done
for bio text being sent to rummager.

https://trello.com/c/U3bfa2tZ/492-rummager-displaying-full-biography-in-search-results-instead-of-first-paragraph

### DIFF
--- a/test/unit/person_test.rb
+++ b/test/unit/person_test.rb
@@ -17,6 +17,25 @@ class PersonTest < ActiveSupport::TestCase
                   }, person.search_index)
   end
 
+  test "truncates the biography sent to rummager unless they currently hold a role" do
+    full_bio = <<~BIO
+      Dapibus Mattis Euismod
+
+      Sit Aenean Venenatis
+
+      Vulputate Euismod Vehicula
+    BIO
+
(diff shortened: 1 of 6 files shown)


test/unit/person_test.rb
```diff
@@ -17,6 +17,25 @@ class PersonTest < ActiveSupport::TestCase
                   }, person.search_index)
   end
 
+  test "truncates the biography sent to rummager if not currently holding a role" do
```
bevanloon: Minor quibble - feel free to ignore.
I think the test is for both cases - i.e. the assertions are checking both a person not in a role and a person in a role's biography in the search index. Perhaps a test name like "it sends the appropriate biography to rummager" or somesuch? might help make it more explicit for when someone's scanning the tests quickly.

bevanloon:
I like the new function names - easy to parse and understand.
Pity about the logic in the views causing CI to fail :(

elliotcm:
@bevanloon I did have the same thought when writing the name of that test.

Both issues should be fixed now.

PR #4353, opened 2018-09-20
Bump govuk_publishing_components 9.25.0

MahmudH:
- Adds hasPart schema to Article schema


cbaines:
Dependabot has already opened a near identical Pull Request https://github.com/alphagov/whitehall/pull/4352

Could you merge that one?

PR #4359, opened 2018-09-21
Calculate the percentage of closed consultations with published respo…

steventux:
…nses before and after a specific date

https://trello.com/c/EMMcgXqK/493-compare-the-of-closed-consultations-that-have-published-responses-before-and-after-17-may-2018

Could do with _#datascience_ review to make sure my maths (and SQL) is correct.

Results (for 2018-05-17):
```
Responses to closed consultations before 2018-05-17: 74.79%
Responses to closed consultations on or after 2018-05-17: 0%
```

Results (for 2017-01-01):
```
Responses to closed consultations before 2017-01-01: 79.77%
Responses to closed consultations on or after 2017-01-01: 31.58%
```


lib/consultation_reporter.rb
```diff
@@ -19,6 +20,24 @@ def all_consultations
     write_csv_from_scope(:all_consultations, scope)
   end
 
+  def response_percentages
+    scope = Consultation
+      .distinct('editions.id')
```
thomasleese: Might be nice to use `.distinct(editions: :id)`. (I think that's possible in ActiveRecord...)

lib/consultation_reporter.rb
```diff
@@ -19,6 +20,24 @@ def all_consultations
...
+      .joins('LEFT JOIN responses ON responses.edition_id = editions.id')
```
thomasleese: Similarly, might be worth seeing if we can do `.joins(:response)` here.

steventux:
Thanks @thomasleese your point about using ActiveRecord to make the join to `:responses` prompted a bit of investigation into why this wouldn't work and ultimately led to a simpler query as the presence of `public_feedback` is all we need to consider a consultation as having had a response. So the aggregate counts weren't necessary either. This made for a much simpler query.

PR #2612, opened 2016-05-31
HTML Publications republishing updates

boffbowsh:
- A performance improvement for presenting HTML Publications
- Handling edge cases where an HTML Publication has no govspeak content
- Check the parent exists and is correct in the sync check script
- Check the content_id is set correctly in the sync check script

PR #4369, opened 2018-09-25
Add HTML attachments to consultation responses

elliotcm:
We did want to revisit all the existing tests related to attachables
but it's too big a rabbithole, so we're bumping that to a future ticket.
 
https://trello.com/c/D7nzpDCa/494-allow-html-attachments-and-external-links-to-public-feedback-and-final-outcome-sections-in-consultation-responses

### New consultation with HTML attachment

screenshot 2018-09-26 11 54 43
screenshot 2018-09-26 11 54 50

### HTML attachment deleted

screenshot 2018-09-26 11 55 59

```
$ curl -I https://www-origin.integration.publishing.service.gov.uk/government/consultations/test-consult/test-attach
HTTP/2 301 
date: Wed, 26 Sep 2018 10:56:35 GMT
content-type: text/html; charset=utf-8
content-length: 73
server: nginx
cache-control: max-age=1800, public
expires: Wed, 26 Sep 2018 11:26:02 UTC
location: /government/consultations/test-consult
accept-ranges: bytes
x-varnish: 2011849466 2011849464
age: 33
via: 1.1 varnish
x-cache: HIT
strict-transport-security: max-age=31536000
```

### New HTML attachment added with the same name (and thus the same URL)

screenshot 2018-09-26 12 12 31
screenshot 2018-09-26 11 58 09

```
$ curl -I https://www-origin.integration.publishing.service.gov.uk/government/consultations/test-consult/test-attach?dsfs=23423
HTTP/2 200 
date: Wed, 26 Sep 2018 10:57:49 GMT
content-type: text/html; charset=utf-8
server: nginx
vary: Accept-Encoding
cache-control: max-age=1800, public
referrer-policy: strict-origin-when-cross-origin
strict-transport-security: max-age=31536000
vary: Accept-Encoding
via: 1.1 router
x-content-type-options: nosniff
x-download-options: noopen
x-frame-options: SAMEORIGIN
x-permitted-cross-domain-policies: none
x-request-id: ccf5cb42-cd94-45ac-b939-c5c7fd4b9ed5
x-xss-protection: 1; mode=block
x-varnish: 143851942
age: 0
via: 1.1 varnish
x-cache: MISS
```


app/presenters/publishing_api/html_attachment_presenter.rb
```diff
@@ -75,7 +75,7 @@ def parent
     end
 
     def parent_content_ids
-      ::Edition.joins(:document).where(editions: { id: item.attachable_id }).pluck(:content_id)
+      [parent.document.content_id]
```
bevanloon: 👍 this definitely makes the code more readable and contained....I notice this is sort of reverting a change made to improve performance in [PR 2612](https://github.com/alphagov/whitehall/pull/2612). Do you think memoising  `item.attachable` in `def parent` would be a good idea? That might help maintain performance but also keep the more readable code.

app/models/html_attachment.rb
```diff
@@ -102,10 +99,6 @@ def translated_locales
     [locale || I18n.default_locale.to_s]
   end
 
-  def generate_content_id
```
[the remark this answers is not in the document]
elliotcm: I would say that is reasonable

app/presenters/publishing_api/html_attachment_presenter.rb
elliotcm: 🤔 I'm not sure that would make much difference to do that; given that `attachable` is an activerecord association it'll be memoized anyway.

The note in the commit suggests it's done to avoid fully instantiating the associated `document`, which I think makes sense.  I'll do a few checks and see if it's still an issue, if so I'll find another way.

elliotcm: OK I've looked into this, and while this does add back an extra request for the document, there's no way to avoid this now that HTML attachments can be further than one step from an edition, without serious refactoring.

elliotcm: The document load is also very quick, so I don't think it's really a problem.  Avoiding loading all the organisations is the big gain, and we keep that here.

bevanloon: 👍awesome thanks for investigating

elliotcm:
@bevanloon I've split out the commits and added the results of testing the HTML attachment deletion behaviour to the original PR description.

PR #4372, opened 2018-09-25
Remove Policy Areas from Organisations pages

AlanGabbianelli:
Remove Policy Areas from Organisations pages

Trello cards:
trello.com/c/VzhuAWdA/229-remove-policy-areas-from-orgs-in-whp
trello.com/c/OKJzONoO/233-after-policy-areas-are-retired-rename-the-tagging-lists-for-orgs-in-whp

As part of the work for the new Single Taxonomy, we are in the process
of unpublishing all Policy Areas. Because of this, all references to
Policy Areas need to be removed.

This removes Policy Areas from the Organisations admin pages and only
leaves Topical Events.

### DIFF
--- a/app/views/admin/organisations/_form.html.erb
+++ b/app/views/admin/organisations/_form.html.erb
@@ -77,14 +77,14 @@
       <%= organisation_form.select :parent_organisation_ids, options_from_collection_for_select(Organisation.with_translations(:en) - [organisation_form.object], 'id', 'select_name', organisation.parent_organisation_ids), {}, multiple: true, class: 'chzn-select form-control', data: { placeholder: "Choose parent organisations…" } %>
     </p>
 
-    <h3>Policy Areas</h3>
-    <% organisation_form.object.organisation_classifications.each do |ot| %>
+    <h3>Topical events</h3>
+    <% organisation_form.object.organisation_classifications.each do |oc| %>
       <p>
-        <%= label_tag "organisation_topic_ids_#{ot.ordering}" do %>
-          Policy Area <%= ot.ordering + 1 %>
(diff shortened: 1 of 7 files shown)


app/views/admin/organisations/_form.html.erb
```diff
@@ -76,18 +76,6 @@
       <%= organisation_form.label :parent_organisation_ids, 'Sponsoring organisations' %>
       <%= organisation_form.select :parent_organisation_ids, options_from_collection_for_select(Organisation.with_translations(:en) - [organisation_form.object], 'id', 'select_name', organisation.parent_organisation_ids), {}, multiple: true, class: 'chzn-select form-control', data: { placeholder: "Choose parent organisations…" } %>
     </p>
-
-    <h3>Policy Areas</h3>
-    <% organisation_form.object.organisation_classifications.each do |ot| %>
```
1pretz1: This filter list is a weird mix of Topical events and Policy Areas so work will need to be done after Policy area's are retired to change the wording of the lists to 'Topical events'. And also test what happens when a policy area that an org is tagged to is retired.

db/schema.rb
```diff
@@ -41,6 +41,7 @@
     t.datetime "updated_at"
     t.integer "replaced_by_id"
     t.datetime "uploaded_to_asset_manager_at"
+    t.datetime "synchronised_with_asset_manager_at"
```
1pretz1: I think this has been accidentally added and should be removed? :)

cbaines: Looks to have been removed in this commit https://github.com/alphagov/whitehall/commit/5073041d082a9766a322ec79a90973d705804d05 but instead of adding another migration to remove the field, the original migration was removed.

This may have left your database in an inconsistent state with respect to this field.

1pretz1: 👍

PR #4373, opened 2018-09-26
Remove ability to access `Featured policies`

1pretz1:
As policies are being retired we need to remove code relating to them,
this is just a frontend solution by removing the route and the link to
get to the featured policies page.

Trello:
https://trello.com/c/HHnlPkM9/230-remove-featuredpolicies-from-editing-creating-orgs-in-whitehall-admin-publisher


cbaines:
Could this be removed as well? https://github.com/alphagov/whitehall/blob/support-subtopic-filtering-in-document-filter/app/views/admin/organisations/features.html.erb

1pretz1:
Nope, as`features` is a different part of the admin org page which doesn't include `featured_policies`. This file is referring to this page:
https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/organisations/home-office/features

1pretz1:
I've made all the changes to remove everything `featured_policy` related in another branch, just have to clean up the commits and open a pr

PR #4376, opened 2018-09-26
Change to the transaction Cucumber javascript_strategy

cbaines:
This will hopefully be faster than truncation, and hopefully won't
cause any issues. The Rails tests use transactions for isolation, as
well as separate databases.


cbaines:
After some rough testing, this seems to make the tests slower.... so lets not do this.

PR #4381, opened 2018-09-27
Remove policies component in whitehall admin

1pretz1:
Remove the excess code around `featured_policies`, in a previous pr the ability to access this feature was removed via removing the links and routes: https://github.com/alphagov/whitehall/pull/4373. 

However this pr completely removes all the code relating to `featured_policies` from whitehall as policies are soon to be retired.

Trello:
https://trello.com/c/HHnlPkM9/230-remove-featuredpolicies-from-editing-creating-orgs-in-whitehall-admin-publisher


db/schema.rb
```diff
@@ -683,6 +677,7 @@
     t.datetime "updated_at", null: false
     t.index ["batch_id"], name: "index_link_checker_api_reports_on_batch_id", unique: true
     t.index ["link_reportable_type", "link_reportable_id"], name: "index_link_checker_api_reportable"
+    t.index ["link_reportable_type", "link_reportable_id"], name: "index_old_link_checker_api_reportable"
```
cbaines: This seems to have crept in?

1pretz1:
Yep you were correct and i've removed the extra index! 👍

PR #4389, opened 2018-10-01
Add data migration to retag HCA content to RSH

barrucadu:
After migration the content is tagged to both organisations.

---

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/3229349)


barrucadu:
I've run this on integration, and there's a little lag in the change being visible, but otherwise it seemed to work fine.

db/data_migration/20181001113221_retag_hca_to_rsa.rb
```diff
@@ -0,0 +1,358 @@
...
+  'statistical-data-return-2015-to-2016',
+  'statistical-data-return-2016-to-2017',
+].each do |slug|
+  begin
+    document = Document.find_by(slug: slug)
+    edition = document.latest_edition
+
+    lead_organisations = edition.lead_organisations.to_a
+    next if lead_organisations.include? regulator_social_housing
+    lead_organisations << regulator_social_housing
```
thomasleese: Do we need to clear which lead organisation the document was previously tagged to? I notice the Zendesk ticket says "retag"?

barrucadu: It does, but the latest comment says

> Just to confirm that https://www.gov.uk/government/organisations/regulator-of-social-housing (https://www.gov.uk/government/organisations/regulator-of-social-housing) is now live and most things seem right.
>
> Once the publications are tagged on it (alongside HCA) that’d be great.

thomasleese: Ah missed that, okay!

PR #4393, opened 2018-10-02
Reslug document to fix URL typo

barrucadu:
[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/3245071)


thomasleese:
You don't actually have to create a migration for this, you can use the Rake task:

https://github.com/alphagov/whitehall/blob/master/lib/tasks/reslugging.rake#L43

PR #4394, opened 2018-10-02
Remove the 'Coming soon' related functionality

cbaines:
Working from the description in the ComingSoonPresenter, this was a
workaround for issues with caching in the wider publishing
platform. Now, the publishing platform handles cache invalidation, so
this shouldn't be a problem any more.

The functionality in Whitehall sits both around the scheduling, and
unscheduling, so I've removed the coming soon part of both actions.


cbaines:
I've tested this on staging, and it seems to work. I'm not sure how well tested scheduled publishing is in Whitehall, as some of the feature tests seem to be missing, I've found some that were removed, and I'm trying to add them back here https://github.com/alphagov/whitehall/pull/4395

PR #4396, opened 2018-10-02
Add data migration to move Natural England content to the RPA

barrucadu:
Unlike e4dcfa2, this is replacing the organisation, not just adding a
new one.

---

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/3115668)


barrucadu:
This is incorrect as it doesn't handle the case where one is the lead organisation and one is the supporting organisation, like this doc:

```
irb(main):026:0> Document.find_by!(slug: 'temporary-prescription-adjustments').latest_edition.lead_organisations.map(&:slug)
=> ["rural-payments-agency"]
irb(main):027:0> Document.find_by!(slug: 'temporary-prescription-adjustments').latest_edition.supporting_organisations.map(&:slug)
=> ["natural-england"]
```

PR #4411, opened 2018-10-04
Don't always send a value for first_published_at

cbaines:
Only send a value if the document has actually been
published. Otherwise, the obvious problem occurs, and the document
appears to have been published when it was created, not when it was
first published!

The FirstPublicAt class has the same problem as it uses the
FirstPublishedAt class, so apply the same fix there also.


cbaines:
Superseded by https://github.com/alphagov/whitehall/pull/4413

PR #4418, opened 2018-10-09
Ensure current_release_date is up-to-date

cbaines:
Due to how associations work in ActiveRecord, I think there could be
issues with changing dates for StatisticsAnnouncements, as the
association on the model doesn't know that the data it has is now
invalid if the say a new StatisticsAnnouncementDate is created.

Therefore, to avoid this, reload the association after a
StatisticsAnnouncementDate is saved.

### DIFF
--- a/app/models/statistics_announcement_date.rb
+++ b/app/models/statistics_announcement_date.rb
@@ -4,6 +4,8 @@ class StatisticsAnnouncementDate < ApplicationRecord
   belongs_to :statistics_announcement, touch: true
   belongs_to :creator, class_name: 'User'
 
+  after_save :update_statistics_announcement_current_release_date
+
   validates :release_date, presence: true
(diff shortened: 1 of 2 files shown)


app/models/statistics_announcement_date.rb
```diff
@@ -21,6 +23,10 @@ def display_date
 
 private
 
+  def update_statistics_announcement_current_release_date
+    statistics_announcement.try(:reload_current_release_date)
```
steventux: Is `reload_current_release_date` a valid method on `StatisticsAnnouncement`? Github search doesn't find it, is it railsmagic?

cbaines: Rails magic unfortunately, it's created by the `has_one` method: https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_one

PR #4431, opened 2018-10-18
Revert "Don't override notify! in ScheduledEditionPublisher"

vanitabarrett:
Reverts alphagov/whitehall#4428

We're having issues with scheduled publishing at the moment, and this is the only related commit that got deployed recently. As an attempt at a quick fix, we're going to try and revert this and see if it helps.


vanitabarrett:
Not needed.

PR #4432, opened 2018-10-18
Make scheduled publishing more reliable

cbaines:
For the majority of content for which being published by the
Publishing API is required for it to be "published", Whitehall still
doesn't consider the Publishing API interaction to be a critical
component.

This is a general problem, but specifically for scheduled publishing,
this is causing an issue because it can be a cause of extra load on
the Publishing API, to the degree that it causes timeouts. When this
happens, the ScheduledPublishingWorker jobs fail, but when they're
retried, they no-op, as the Edition has already been transitioned to
the published state.

Extending the transaction around the whole operation of the
ScheduledEditionPublisher service in this case, means that the
Publishing API interaction (along with all the other edition services
listening in) is part of the transaction, and thus if it fails, the
database won't be updated to say that the edition has been
successfully published.

This change could be made in the EditionService class, but as the
current problems are focused around scheduled publishing, making such
a wide reaching change seems too risky without enough benefit, when
making it here should have the same effect, but just for scheduled
publishing.

### DIFF
--- a/app/workers/scheduled_publishing_worker.rb
+++ b/app/workers/scheduled_publishing_worker.rb
@@ -47,8 +47,18 @@ class ScheduledPublishingWorker < WorkerBase
     # Call `ScheduledEditionPublisher` via the `EditionServiceCoordinator`.
     publisher = Whitehall.edition_services.scheduled_publisher(edition)
 
-    Edition::AuditTrail.acting_as(publishing_robot) do
-      publisher.perform! || raise(ScheduledPublishingFailure, publisher.failure_reason)
+    # The Publishing API part of publishing (which is critically
+    # important) isn't part of the transaction in the
+    # EditionService. Starting the transaction here ensures that it
+    # is, and thus the edition won't be marked as published in the
+    # event of the actual publishing failing.
+    #
+    # Hopefully the Publishing API interaction can be improved
+    # generally to the point where this can be removed.
+    ActiveRecord::Base.transaction do
+      Edition::AuditTrail.acting_as(publishing_robot) do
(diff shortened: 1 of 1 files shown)


cbaines:
Thanks for taking a look Bevan.

> I'm not sure how Rails handles the nested transactions - I think I'm right in saying that this transaction eventually wraps https://github.com/alphagov/whitehall/blob/master/app/services/edition_service.rb#L13?

Yep.

> I know this is possibly out of date but https://makandracards.com/makandra/42885-nested-activerecord-transaction-pitfalls might be relevant?

In this particular case, I don't think the pitfall it describes is directly applicable, as we're looking to handle exceptions from the outer transaction block, not the inner one. But I can't quite work out what the significance of the outer transaction block committing would be, if the inner transaction block has rolled back to a savepoint (which is what I think it should do).

> I know this work doesn't directly fall in line with your current mission work @cbaines so I've created a card for [Platform Health](https://trello.com/c/91gKydVP/679-investigate-a-pr-by-chris-baines-to-wrap-whitehall-scheduled-pubs-in-a-transaction) to follow up on this PR. Of course we are more than happy to have any of your help if you wish to stay involved :)

Good good. Now that it's not the evening during an incident, I've written up a more general change which will have a similar effect for scheduled publishing, but also Publishing API updates in general, and doesn't require nested transaction blocks [1]. While I think it's better to do something more general, I'll leave this Pull Request open for now.

1: https://github.com/alphagov/whitehall/pull/4440

cbaines:
As https://github.com/alphagov/whitehall/pull/4440 has been merged, this can be closed now :tada:

PR #4437, opened 2018-10-22
Don't require policy areas for stats announcements

cbaines:
This is a quick workaround for left over Policy Area (also know as
Topic) code affecting publishing statistical announcements.

Hopefully the remaining Policy Area code can be cleaned up shortly.


cbaines:
Duplicate of https://github.com/alphagov/whitehall/pull/4436

PR #4440, opened 2018-10-23
Handle the Publishing API within the EditionService

cbaines:
Rather than processing the Publishing API change after the change has
been saved in Whitehall, process it during the transaction. This means
that there is less opportunities for Whitehall's state to diverge from
the Publishing API, as the state will only be transitioned in
Whitehall when the Publishing API update succeeds.


barrucadu:
This makes sense to me.  When I'm done with my current publishing-api thing, I'll deploy this to staging and try publishing some stuff (normally and scheduled) with the publishing-api down to check it.

PR #4442, opened 2018-10-23
Migration to remove change note

MuriloDalRi:
Zendesk ticket: https://govuk.zendesk.com/agent/tickets/3409998


vanitabarrett:
@MuriloDalRi I think we should try and run this on integration to double check it does what we want first, before merging

MuriloDalRi:
We tested this on integration and it works

PR #4449, opened 2018-10-26
Change announcements finder rummager endpoint

emmabeynon:
For https://trello.com/c/5SNdtZIJ/341-change-announcements-finder-to-use-search-endpoint

Content Publisher will soon be publishing news content in production, which will be stored in the govuk index in Rummager.  Currently the announcements finder uses the 'advanced-search' Rummager endpoint which only queries the government index, so in this PR we switch the announcements finder to use the 'search' Rummager endpoint which queries both the government and govuk indices.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -9,7 +9,12 @@ class DocumentsController < PublicFacingController
 
 private
 
-  def build_document_filter
+  def build_document_filter(filter_type = nil)
+    search_backend = if filter_type == "announcements"
+                       Whitehall::DocumentFilter::SearchRummager
+                     else
+                       Whitehall::DocumentFilter::AdvancedSearchRummager
+                     end
     search_backend.new(cleaned_document_filter_params)
   end
 
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -10,12 +10,7 @@ class DocumentsController < PublicFacingController
 private
 
   def build_document_filter(filter_type = nil)
-    search_backend = if filter_type == "announcements"
-                       Whitehall::DocumentFilter::SearchRummager
-                     else
-                       Whitehall::DocumentFilter::AdvancedSearchRummager
-                     end
-    search_backend.new(cleaned_document_filter_params)
+    search_backend(filter_type).new(cleaned_document_filter_params)
   end
 
   def cleaned_document_filter_params
--- a/lib/whitehall/document_filter/search_result.rb
+++ b/lib/whitehall/document_filter/search_result.rb
@@ -56,21 +56,7 @@ module Whitehall::DocumentFilter
   private
 
     def fetch_from_cache(type, slug)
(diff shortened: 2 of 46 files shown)


lib/whitehall/document_filter/search_result.rb
```diff
@@ -0,0 +1,72 @@
...
+      @doc.fetch('document_collections', []).map { |document_collection| fetch_from_cache(:document_collection, document_collection.fetch('slug', nil)) }.compact
+    end
+
+    def operational_field
+      fetch_from_cache(:operational_field, @doc['operational_field'])
+    end
+
+  private
+
+    def fetch_from_cache(type, slug)
```
kevindew: Should we think about sharing this method with the Advanced Search Result?

lib/whitehall/not_quite_as_fake_search.rb
```diff
@@ -185,13 +185,15 @@ def filter_by_boolean_field(field, desired_field_value, document_hashes)
...
-            value = organisation_slugs(document_hash["organisations"])
-          else
-            value = document_hash.fetch(field, [])
-          end
+          value =
+            if field == "organisations" && announcements_search
+              organisation_slugs(document_hash.fetch("organisations", []))
+            else
+              document_hash.fetch(field, [])
+            end
```
kevindew: Was this conditional only applied on announcements searches before? I just notice that it will no longer ever return true if you filtered a publication by organisations for instance.

app/controllers/documents_controller.rb
```diff
@@ -9,7 +9,12 @@ class DocumentsController < PublicFacingController
 
 private
 
-  def build_document_filter
+  def build_document_filter(filter_type = nil)
+    search_backend = if filter_type == "announcements"
+                       Whitehall::DocumentFilter::SearchRummager
+                     else
+                       Whitehall::DocumentFilter::AdvancedSearchRummager
+                     end
```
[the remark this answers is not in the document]
emmabeynon: This indeed fixes my foreign language announcements problem :)

lib/whitehall/not_quite_as_fake_search.rb
emmabeynon: We need to treat organisations differently for announcements as the '/search' endpoint returns them as an array of hashes, while '/advanced-search' returns them as an array of slugs.  This logic should cater for non-announcement documents as `announcements_search` will be set to false in these cases and with fall through to the standard 'fetch' from the document hash.

lib/whitehall/not_quite_as_fake_search.rb
```diff
@@ -162,8 +195,25 @@ def filter_by_boolean_field(field, desired_field_value, document_hashes)
...
+
+          if value.is_a?(String)
+            [*desired_field_values].include?(value)
+          else
+            ([*desired_field_values] & value).any?
```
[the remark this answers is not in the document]
emmabeynon: Agreed - I had to trial it out to understand what was going on here, so I think `Array(desired_field_values)` would be easier to understand 👍

leenagupte:
Replaced with https://github.com/alphagov/whitehall/pull/4480

PR #4457, opened 2018-10-30
Use Rummager for related document lists on topical events

bilbof:
This change introduces RummagerDocumentPresenter, a wrapper for documents
returned by Rummager searches. It changes the source for related documents
on topical event pages from Whitehall to Rummager. For each related document
list on the topical event page, we make a rummager search query.

In addition it introduces additional views and test logic. This is necessary
because other pages will still have lists with documents sourced from Whitehall.

Hopefully RummagerDocumentPresenter will be extended and used more as we transition
away from Whitehall to using content store, publishing api and rummager to back
Finders and other kinds of list.

A side effect of this change is that the atom feed results for the topical events
will change. This may result in feed readers interpreting old documents as new,
since the IDs and other attributes will appear to have changed.

### Example

These are screenshots of a [topical event page](https://www.gov.uk/government/topical-events/western-balkans-summit-london-2018) after the changes.

End users shouldn't notice a difference after the changes have been deployed.

screen shot 2018-10-31 at 12 03 07
screen shot 2018-10-31 at 12 00 57

### DIFF
--- a/app/controllers/topical_events_controller.rb
+++ b/app/controllers/topical_events_controller.rb
@@ -4,9 +4,9 @@ class TopicalEventsController < ClassificationsController
   def show
     @classification = TopicalEvent.friendly.find(params[:id])
     @content_item = Whitehall.content_store.content_item(@classification.base_path)
-    @publications = fetch_associated(:published_publications, PublicationesquePresenter)
-    @consultations = fetch_associated(:published_consultations, PublicationesquePresenter)
-    @announcements = fetch_associated(:published_announcements, AnnouncementPresenter)
+    @publications = fetch_related_documents(filter_format: 'publication')
+    @consultations = fetch_related_documents(filter_format: 'consultation')
+    @announcements = fetch_related_documents(filter_format: 'news_article')
     @detailed_guides = @classification.published_detailed_guides.includes(:translations, :document).limit(5)
     @featurings = decorate_collection(@classification.classification_featurings.includes(:image, edition: :document).limit(5), ClassificationFeaturingPresenter)
 
@@ -17,22 +17,30 @@ class TopicalEventsController < ClassificationsController
     set_expiry 5.minutes
     respond_to do |format|
       format.html do
-        @recently_changed_documents = @classification.latest(3)
+        @recently_changed_documents = fetch_related_documents(count: 3)['results']
       end
       format.atom do
-        @recently_changed_documents = @classification.latest(10)
+        @recently_changed_documents = fetch_related_documents(count: 10)['results']
       end
     end
   end
 
 private
 
-  def fetch_associated(type, presenter_class)
-    editions = @classification
-      .send(type)
-      .in_reverse_chronological_order
-      .includes(:translations, :document)
-      .limit(3)
-    decorate_collection(editions, presenter_class)
+  def fetch_related_documents(additional_options = {})
+    options = default_search_options.merge(additional_options)
+    search_response = Whitehall.search_client.search(options)
+    search_response["results"].map! { |res| RummagerDocumentPresenter.new(res) }
+
+    search_response
+  end
+
+  def default_search_options
+    {
+      filter_topical_events: @classification.slug,
+      count: 3,
+      order: "-public_timestamp",
+      fields: %w[display_type title link public_timestamp format description content_id]
+    }
   end
 end
--- a/app/helpers/feed_helper.rb
+++ b/app/helpers/feed_helper.rb
@@ -13,7 +13,7 @@ module FeedHelper
     builder.updated feed_updated_timestamp
 
     documents.each do |document|
-      builder.entry(document, id: document_id(document, builder), url: public_document_url(document), published: document.first_public_at, updated: document.public_timestamp) do |_entry|
+      builder.entry(document, id: document_id(document, builder), url: public_document_url(document), published: document.try(:first_public_at), updated: document.public_timestamp) do |_entry|
         document_as_feed_entry(document, builder)
       end
     end
@@ -26,7 +26,7 @@ module FeedHelper
     # spec.
     # The interpolation logic is straight out of:
     # http://api.rubyonrails.org/classes/ActionView/Helpers/AtomFeedHelper/AtomFeedBuilder.html#method-i-entry
-    id = record.document ? record.document.id : record.id
+    id = record.try(:document) ? record.document.id : record.id
     "tag:#{host},#{schema_date(builder)}:#{record.class}/#{id}"
   end
 
@@ -55,6 +55,7 @@ module FeedHelper
   end
 
   def entry_content(document)
+    return '' if document.is_a?(RummagerDocumentPresenter)
     change_note = document.most_recent_change_note
     change_note = "<p><em>Updated:</em> #{change_note}</p>" if change_note
     "#{change_note}#{govspeak_edition_to_html(document)}"
--- a/app/helpers/public_document_routes_helper.rb
+++ b/app/helpers/public_document_routes_helper.rb
@@ -15,6 +15,8 @@ module PublicDocumentRoutesHelper
   end
 
   def document_url(edition, options = {}, _builder_options = {})
+    return edition.url if edition.is_a?(RummagerDocumentPresenter)
+
     if edition.non_english_edition?
       options[:locale] = edition.primary_locale
     elsif edition.translatable?
--- a/app/views/classifications/_document_list_from_rummager.html.erb
+++ b/app/views/classifications/_document_list_from_rummager.html.erb
@@ -1,0 +1,34 @@
+<%
+   heading ||= type.to_s.humanize
+   filter_option = (heading == "Publications") ? "all" : heading.downcase
+%>
+<section id="<%= heading.downcase %>" class="document-block">
+  <h1 class="label"><%= heading %></h1>
+  <div class="content">
+    <%= render partial: "shared/list_description_for_rummager_list", locals: { heading: heading, editions: documents, documents_count: documents_count, type: type } %>
+    <p class="see-all">
+    <%-
+      see_more_url = public_send(
+        "#{type}_filter_path",
+        @classification,
+        publication_filter_option: filter_option
+      )
+    %>
+    <%=
+      link_to(
+        "See all #{heading.downcase}",
+        see_more_url,
+        data: {
+          track_category: 'navPolicyAreaLinkClicked',
+          track_action: "#{heading}",
+          track_label: see_more_url,
+          track_options: {
+            dimension28: documents_count.to_s,
+            dimension29: "See all #{heading.downcase}"
+          }
+        }
+      )
+    %>
+    </p>
+  </div>
+</section>
--- a/app/views/shared/_list_description_for_rummager_list.html.erb
+++ b/app/views/shared/_list_description_for_rummager_list.html.erb
@@ -1,0 +1,27 @@
+<ol class="document-list">
+  <% editions.each_with_index do |edition, edition_index| %>
+    <li class="document-row" id="<%= "#{type}_#{edition.content_id}" %>">
+      <h2>
+        <%=
+          link_to(
+            edition.title,
+            edition.link,
+            data: {
+              track_category: 'navPolicyAreaLinkClicked',
+              track_action: "#{local_assigns.has_key?(:heading) ? heading : "Unknown"}.#{edition_index + 1}",
+              track_label: edition.link,
+              track_options: {
+                dimension28: documents_count.to_s,
+                dimension29: edition.title
+              }
+            }
+          )
+        %>
+      </h2>
+      <ul class="attributes">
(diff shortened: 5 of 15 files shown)


app/controllers/topical_events_controller.rb
```diff
@@ -4,9 +4,9 @@ class TopicalEventsController < ClassificationsController
   def show
     @classification = TopicalEvent.friendly.find(params[:id])
     @content_item = Whitehall.content_store.content_item(@classification.base_path)
-    @publications = fetch_associated(:published_publications, PublicationesquePresenter)
-    @consultations = fetch_associated(:published_consultations, PublicationesquePresenter)
-    @announcements = fetch_associated(:published_announcements, AnnouncementPresenter)
+    @publications = fetch_publications
+    @consultations = fetch_consultations
+    @announcements = fetch_announcements
```
bilbof: This is the meat of the change. Rather than looking up related documents in Whitehall, we're fetching them from Rummager.

app/views/shared/_list_description_for_rummager_list.html.erb
```diff
@@ -0,0 +1,27 @@
+<ol class="document-list">
+  <% editions.each_with_index do |edition, edition_index| %>
+    <li class="document-row" id="<%= "#{type}_#{edition.content_id}" %>">
+      <h2>
```
vanitabarrett: If it's possible, using the [document list component](https://govuk-publishing-components.herokuapp.com/component-guide/document_list) here would be ideal, so we can match document lists on other pages and in other applications. 

If not, we should at least remove the `h2` as the document title isn't really a heading in terms of being the heading for a section and it will create a lot of noise for screenreader users, especially those navigating by heading.

app/helpers/feed_helper.rb
```diff
@@ -13,7 +19,7 @@ def documents_as_feed_entries(documents, builder, feed_updated_timestamp = Time.
     builder.updated feed_updated_timestamp
 
     documents.each do |document|
-      builder.entry(document, id: document_id(document, builder), url: public_document_url(document), published: document.first_public_at, updated: document.public_timestamp) do |_entry|
+      builder.entry(document, id: document_id(document, builder), url: public_document_url(document), published: document.try(:first_public_at), updated: document.public_timestamp) do |_entry|
```
andrewgarner: Given that `first_public_at` doesn't appear to be a field in the Search API <sup>[1]</sup> could we either:

- fall back to `first_published_at` from the Search API _or_
- alias `first_public_at` to `first_published_at` in our models

[1]: https://docs.publishing.service.gov.uk/apis/search/fields.html

app/helpers/feed_helper.rb
```diff
@@ -4,6 +4,12 @@ def atom_feed_url_for(resource)
   end
 
   def documents_as_feed_entries(documents, builder, feed_updated_timestamp = Time.current)
+    # This is to support cases where the documents feed has been provided by
+    # another service (e.g. Rummager) rather than from Whitehall via ActiveRecord.
+    if documents.is_a?(GdsApi::Response) && documents['results']
+      documents = documents['results']
```
andrewgarner: Could we avoid this case if we returned the results in the topical_events_controller.rb `fetch_related_documents`?

app/views/classifications/_document_list_from_rummager.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        "#{type}_filter_path",
+        @classification,
+        publication_filter_option: filter_option
+      )
+    %>
+    <%=
+      link_to(
+        "See all #{heading.downcase}",
+        see_more_url,
+        data: {
```
vanitabarrett: For our data-tracking to take effect, we normally have to add `data-module="track-click"` to the element itself or a parent. Has this been included anywhere? I normally test the analytics using something like [Google Analytics Debugger](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) to double check the tracking events are firing.

features/latest-documents.feature
```diff
@@ -7,7 +7,7 @@ Feature: List of most recently published documents
   Scenario: Latest documents on topical event page
     Given a topical event with published documents
     When I view that topical event page
-    Then I can see some of the latest documents
+    Then I can see some of the latest documents from rummager
```
andrewgarner: Could we replace the implementation of the "Then I can see some of the latest documents" step definition? "from rummager" is an implementation detail, it's not something the user is aware of.

app/views/shared/_list_description_for_rummager_list.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        <%=
+          link_to(
+            edition.title,
+            edition.link,
+            data: {
+              track_category: 'navPolicyAreaLinkClicked',
+              track_action: "#{local_assigns.has_key?(:heading) ? heading : "Unknown"}.#{edition_index + 1}",
```
vanitabarrett: Is it likely that `heading` would ever be empty?

app/views/shared/_recently_updated_documents_from_rummager.html.erb
```diff
@@ -0,0 +1,13 @@
+<ol class="document-list">
```
vanitabarrett: Same applies here about the using document list component if possible. If not, removing the `h2`.

app/views/shared/_recently_updated_from_rummager.html.erb
```diff
@@ -0,0 +1,21 @@
+<% extra_class ||= '' %>
```
vanitabarrett: Question: what exactly is `extra_class` and in what situations is it provided? Is it possible to make this variable name a little clearer?

features/support/topical_events_helper.rb
```diff
@@ -30,6 +30,10 @@ def sample_document_types_and_titles
     }
   end
 
+  def rummager_response
+    File.read(Rails.root.join('features/fixtures/rummager_response.json'))
```
oscarwyatt: This is an interesting approach, I get that it's tidier and means you don't have 36 lines more in this file but to my mind so thats good, unless there is some super specific reason it has to be a separate file, I'd err away from it and move the resultant somewhat large method to the bottom  (feel free to disagree though)

app/views/classifications/_document_list_from_rummager.html.erb
```diff
@@ -0,0 +1,34 @@
+<%
```
oscarwyatt: It seems that this is trying to be the same as `document_list.html.erb` but with specific adaptions because Rummager results are slightly different(?) If so and you have the time, I'd suggest trying to find a way to make it so `document_list.html.erb` can present both - perhaps by either having a class that will take in the Rummager responses and respond in the same way as the original responses or by having some class that can take in either kind of response and act like a common interface for them. Otherwise, someone making a change to the document list will have to figure out that they are meant to be the same and make the same changes twice. This might also help with the problem we discussed about FeedHelper (app/helpers/feed_helper.rb line 9)
If they're different files because they fundamentally are meant to be different, ignore all of this!

app/helpers/public_document_routes_helper.rb
```diff
@@ -15,6 +15,8 @@ def preview_document_path(edition, options = {})
   end
 
   def document_url(edition, options = {}, _builder_options = {})
+    return edition.link if edition.is_a? RummagerDocumentPresenter
```
oscarwyatt: Similar to my comment on `document_list_from_rummager.html.erb`

app/views/classifications/_document_list_from_rummager.html.erb
bilbof: That's a valid question. Initially I had planned to update the `document_list` template so that it could take both Rummager and Whitehall documents.

Since the aim is to migrate document lists from Whitehall to Rummager, I thought it would be easier to manage two similar but separate templates rather than try to making one adjust to two models. Having this separation of concerns is intended to keep the migration and views simpler, and once we're done it'll let us just delete the Whitehall template rather than needing to untangle it from a combined view. You're right that having to manage two views that essentially have the same task is a bit odd, even if the models they're provided with are different. Let me know if having two views isn't simpler - I'll be happy to have another go at a common view that plays nice with both models.

oscarwyatt: Ah, right. That does make sense. In which case, I'd suggest making this super clear somewhere <waves hands indicating I'm not really sure where, maybe the PR notes, or a commit message> that this is the plan so other devs will know where you're coming from

features/support/topical_events_helper.rb
bilbof: I'm using this fixture in a couple of different places so I pulled it out into a JSON. I can put it into a new rummager test helper class, which might be a bit neater?

app/views/classifications/_document_list_from_rummager.html.erb
bilbof: Yeah that's a good idea, sorry this wasn't clear! I will update the commit message.

app/views/classifications/_document_list_from_rummager.html.erb
bilbof: This is included in the same place as the existing `_document_list`, so should behave in the same way (if the tracking on that template is broken this will be too).

app/helpers/feed_helper.rb
bilbof: Yeah good spot, thanks!

app/helpers/feed_helper.rb
bilbof: Neither first_published_at or first_public_at are reliably provided, so the option I went for was to default to not providing a `published` field in the feeds for rummager documents.

Do you think we should try getting `first_published_at`? I wasn't able to find any docs with this field.

app/controllers/topical_events_controller.rb
```diff
@@ -18,22 +18,46 @@ def show
...
+
+  def fetch_consultations
+    fetch_documents_for_format('consultation')
+  end
+
+  def fetch_announcements
+    fetch_documents_for_format('news_article')
+  end
+
+  def fetch_documents_for_format(filter_format)
```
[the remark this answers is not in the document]
bilbof: Sure! Changed this.

app/helpers/public_document_routes_helper.rb
bilbof: Thanks, fixed

app/controllers/topical_events_controller.rb
```diff
@@ -4,9 +4,9 @@ class TopicalEventsController < ClassificationsController
   def show
     @classification = TopicalEvent.friendly.find(params[:id])
     @content_item = Whitehall.content_store.content_item(@classification.base_path)
-    @publications = fetch_associated(:published_publications, PublicationesquePresenter)
-    @consultations = fetch_associated(:published_consultations, PublicationesquePresenter)
-    @announcements = fetch_associated(:published_announcements, AnnouncementPresenter)
+    @publications = fetch_related_documents(filter_format: 'publication')
+    @consultations = fetch_related_documents(filter_format: 'consultation')
+    @announcements = fetch_related_documents(filter_format: 'news_article')
     @detailed_guides = @classification.published_detailed_guides.includes(:translations, :document).limit(5)
```
emmabeynon: In our work to switch over the announcements finder to use Rummager we've decided to filter for announcements by `content_store_document_type` as according to the documentation `format` may be deprecated, and it allows us to be more specific with which documents are returned from Rummager.  We've got them listed out here: https://github.com/alphagov/whitehall/pull/4449/commits/dcdbd71bc2702945a4e8aab21aefd9f89b7511c7
For consistency I think it would be good to filter in the same way here so that the 3 announcements that a user sees on this page are the top 3 announcements on the announcements finder page. What do you think?

features/support/topical_events_helper.rb
koetsier: Never tried this before but there is this: https://api.rubyonrails.org/v5.1/classes/ActiveSupport/Testing/FileFixtures.html

app/controllers/topical_events_controller.rb
bilbof: That's a good idea. After or as part of #4449) we can add the following to the topical events controller.

```
  def fetch_announcements
    # This matches behaviour at https://github.com/alphagov/whitehall/commit/dcdbd71bc2702945a4e8aab21aefd9f89b7511c7
    announcement_doctypes = Whitehall::AnnouncementFilterOption.all.map(&:document_type).flatten
    fetch_related_documents(filter_content_store_document_type: announcement_doctypes)
  end
```

features/support/topical_events_helper.rb
bilbof: Do you know if it would work if the tests are in different directories? This is used in `/features` and `/test/unit`

koetsier: you could set it in a helper as:

```ActiveSupport::TestCase.file_fixture_path='features/fixtures'```

Anyway, not sure how much better it is - just came across this feature

PR #4465, opened 2018-11-02
Temporary fix for world location corp info pages

sihugh:
These are blocked at the moment by conflicts when communicating with publishing api.  This fix is to allow publishing to resume while we work on a better solution.

### DIFF
--- a/app/services/edition_service.rb
+++ b/app/services/edition_service.rb
@@ -13,7 +13,7 @@ class EditionService
       ActiveRecord::Base.transaction do
         prepare_edition
         fire_transition!
-        update_publishing_api!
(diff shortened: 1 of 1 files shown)


AlanGabbianelli:
More context on the issue and on the temporary fix can be found in [this google doc](https://docs.google.com/document/d/16c3kHSo0jWbPYRJiky7GpJq4TWE8dIEnbgvcMhrXX0k/edit#) and in [this trello card](https://trello.com/c/PlWSEch7/109-whitehall-worldwide-organisation-cips-and-normal-cips-are-different).
Just writing this here in case this PR gets referenced in the future (it would have been useful to me 😅).

PR #4467, opened 2018-11-05
Short circuit has_links for editions that are the length of a novel

bevanloon:
We have an edition that is over 200000 words long. That's longer than many novels.

This is causing timeouts when `LinkCheckerApiService#has_links?` tries to extract the links from the body of the edition. `LinkCheckerApiService#has_links?` is called indirectly from `editions/show.html.erb`. The results are used to decide whether or not to show the link checking sidebar (i.e. when no links are present in the edition's body, we don't bother to show the link check
sidebar).

This PR short circuits the logic for editions that are over 99999 words long and just displays the link check sidebar anyway. This does potentially introduce an inconsistency where a very long edition without links will show the link check sidebar regardless. However, this does also get around our timeout for now and will allow editors to actually get to their editions.

I've chosen 99999 words because [Wikipedia](https://en.wikipedia.org/wiki/Word_count#In_fiction) suggests a novel to be 100000 words.

[Trello](https://trello.com/c/KxLdqIhq/601-fix-whitehall-timeouts-for-novel-sized-editions)

### DIFF
--- a/app/helpers/admin/editions_helper.rb
+++ b/app/helpers/admin/editions_helper.rb
@@ -291,4 +291,23 @@ module Admin::EditionsHelper
   def show_similar_slugs_warning?(edition)
     !edition.document.published? && edition.document.similar_slug_exists?
   end
+
+  def edition_is_a_novel?(edition)
+    edition.body.split.size > 99999
+  end
+
+  def edition_has_links?(edition)
+    LinkCheckerApiService.has_links?(edition, convert_admin_links: false)
+  end
+
+  def show_link_check_report?(edition)
+    # There is an edition that is over 200000 words long.
+    # This causes timeouts when LinkCheckerApiService tries to extract links from the body.
+    # This is an exceptional case, but it stops publishers editing their editions.
+    # Short circuit the call to LinkCheckerApiService by testing for an edition being
+    # over 99999 words long. The number was chosen because Wikipedia suggests 100000 words is
+    # the lower length of a novel (https://en.wikipedia.org/wiki/Word_count#In_fiction).
+    # Returning true from the first half of the "or" means the second half doesn't get computed.
+    edition_is_a_novel?(edition) || edition_has_links?(edition)
+  end
 end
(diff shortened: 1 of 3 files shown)


app/helpers/admin/editions_helper.rb
```diff
@@ -291,4 +291,23 @@ def specialist_sector_name(sector_content_id)
...
+  def edition_has_links?(edition)
+    LinkCheckerApiService.has_links?(edition, convert_admin_links: false)
+  end
+
+  def show_link_check_report?(edition)
+    # There is an edition that is over 200000 words long.
+    # This causes timeouts when LinkCheckerApiService tries to extract links from the body.
+    # This is an exceptional case, but it stops publishers editing their editions.
+    # Short circuit the call to LinkCheckerApiService by testing for an edition being
+    # over 99999 words long (chosen because a novel Wikipedia suggest 100000 words is
```
barrucadu: Minor nitpick: should be "chosen because Wikipedia suggests 100000 words is"

bevanloon: 👍 fixed in https://github.com/alphagov/whitehall/pull/4467/commits/e6050a2c8f4f4de35f3735e729840d693625d988

bevanloon:
@kevindew thanks!

> Extracting the links seems to involve rendering the govspeak and then extracting the links from that - and I think that most of the time is the rendering. Are we at risk of trading this timeout for a different one, do things like preview govspeak and send to Publishing API synchronously still work without timing out?

Anecdotally, the onward flow seemed to work when I released the branch to staging. That's not scientific though. One step at a time I suppose :)

> Is the limit 99,999 the right point to not time out? On that basis I'm wondering if the 200,000 word document only just goes over the timeout or is miles over. Just on the basis that would an incredibly long document that is half the length not have problems.

The number is guess based on being curious as to how the size of the edition compares to works of fiction and having a quick peek at the data to find our largest editions. Our next largest is 91,000 which works just fine. Below that we're at 46,000, a couple in the 30,000s, a handful in the 20,000s and then in to more manageable numbers. Given we have somewhere in the region of 200,000 published editions in Whitehall, I thought finding the actual tipping point might be interesting, but probably not terribly urgent.

PR #4475, opened 2018-11-06
Improve Contact presentation to Publishing API

kevindew:
Trello: https://trello.com/c/gEAMmD0g/386-ability-to-embed-a-contact-via-markdown

This is marked as DNM as it requires this https://github.com/alphagov/govuk-content-schemas/pull/830 to be deployed.

This resolves a lot of empty data problems with Contacts data in the Publishing API by removing null and empty string values. It also adds in the iso2_country_code field to help with rendering contacts in govspeak correctly: https://github.com/alphagov/govspeak/pull/130

### DIFF
--- a/app/presenters/publishing_api/contact_presenter.rb
+++ b/app/presenters/publishing_api/contact_presenter.rb
@@ -14,6 +14,7 @@ module PublishingApi
     def content
       {
         title: title,
+        description: comments.presence,
         schema_name: schema_name,
         document_type: document_type,
         locale: locale,
@@ -58,10 +59,6 @@ module PublishingApi
 
     alias_method :document_type, :schema_name
 
-    def description
-      comments
-    end
-
     def locale
       I18n.locale.to_s
     end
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/contact_presenter.rb
```diff
@@ -71,34 +68,40 @@ def contact_type
     end
 
     def details
-      details = { description: description, contact_type: contact_type, title: title }
-      details[:contact_form_links] = [contact_form_links] if contact_form_url
-      details[:post_addresses] = [post_address] if postal_code
-      details[:email_addresses] = [email_address] if email
+      details = {
+        title: title,
```
benthorner: I think you changed the schema so these aren't required anymore - just wondering if you meant to remove these now?

app/presenters/publishing_api/contact_presenter.rb
```diff
@@ -71,34 +68,40 @@ def contact_type
...
+      details[:post_addresses] = post_addresses
+      details[:email_addresses] = [email_address] if email.present?
       details[:phone_numbers] = phone_numbers if contact_numbers.any?
 
       details
     end
 
     def contact_form_links
       {
-        title: title,
```
benthorner: Just a quick note-to-self this isn't used by govspeak when the form link is rendered.

app/presenters/publishing_api/contact_presenter.rb
kevindew: I'm not 100% sure what the consequences are of removing them, but am confident they shouldn't be used.

PR #4480, opened 2018-11-08
Change announcements finder rummager endpoint

leenagupte:
Supersedes PR #4449 
Trello: https://trello.com/c/5SNdtZIJ/341-change-announcements-finder-to-use-search-endpoint

Content Publisher will soon be publishing news content in production, which will be stored in the govuk index in Rummager. Currently the announcements finder uses the 'advanced-search' Rummager endpoint which only queries the government index, so in this PR we switch the announcements finder to use the 'search' Rummager endpoint which queries both the government and govuk indices.

### DIFF
--- a/lib/whitehall/document_filter/advanced_search_result.rb
+++ b/lib/whitehall/document_filter/advanced_search_result.rb
@@ -1,0 +1,72 @@
+module Whitehall::DocumentFilter
+  class AdvancedSearchResult
+    ACCESSORS = %w{title description indexable_content attachments
+                   format display_type link id search_format_types
+                   relevant_to_local_government}.freeze
+    ACCESSORS.each do |attribute_name|
+      define_method attribute_name.to_sym do
+        @doc[attribute_name.to_s]
+      end
+    end
+
+    def initialize(doc)
+      @doc = doc
+    end
+
+    def type
+      format
+    end
+
+    def search_government_name
+      @doc['government_name']
+    end
+
+    def historic?
+      @doc['is_historic']
+    end
+
+    def public_timestamp
+      Time.zone.parse(@doc['public_timestamp'])
+    end
+
+    def part_of_published_collection?
+      published_document_collections.any?
+    end
+
+    def organisations
+      @doc.fetch('organisations', []).map { |slug| fetch_from_cache(:organisation, slug) }.compact
+    end
+
+    def topics
+      @doc.fetch('topics', []).map { |slug| fetch_from_cache(:topic, slug) }.compact
+    end
+
+    def published_document_collections
+      @doc.fetch('document_collections', []).map { |slug| fetch_from_cache(:document_collection, slug) }.compact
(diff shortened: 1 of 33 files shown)


app/services/fetch_from_cache_service.rb
```diff
@@ -0,0 +1,26 @@
+class FetchFromCacheService
```
kevindew: I think this could use defining what it fetches from cache, just to differentiate it from say Rails.cache.fetch

Tricky thing is to define what they are, search associations?

lib/whitehall/document_filter/advanced_search_result.rb
```diff
@@ -52,21 +52,7 @@ def operational_field
...
-        when :document_collection
-          # Don't fall over if index refers a document which has had it's slug changed.
-          Document.find_by(slug: slug).try(:published_edition)
-        when :operational_field
-          OperationalField.find_by(slug: slug)
-        else
-          raise "Can't fetch '#{type}' -- unknown type"
-        end
-      end
+      FetchFromCacheService.new(type, slug).fetch
```
kevindew: I reckon rather than having this as a method you could just replace the method call with the service call.

test/unit/whitehall/document_filter/advanced_search_rummager_test.rb
```diff
@@ -0,0 +1,62 @@
...
+        .with(
+          has_entry(
+            search_format_types: format_types
+          )
+        )
+    end
+
+    def expect_search_by_taxonomy_tree(taxons)
+      Whitehall
+          .government_search_client
```
kevindew: indenting looks a little off here

lib/whitehall/document_filter/advanced_search_result.rb
leenagupte: I considered doing that, but thought the lines looked a bit long.

app/services/fetch_from_cache_service.rb
leenagupte: We ended up removing the `SearchResult` presenter (and therefore the code duplication), so we don't need this service anymore.

lib/whitehall/document_filter/advanced_search_result.rb
leenagupte: The FetchFromCacheService has been removed (as though it never existed) as we have instead removed the `SearchResult` presenter and replaced it with an updated `RummagerDocumentPresenter`.

test/unit/whitehall/document_filter/advanced_search_rummager_test.rb
leenagupte: Fixed and fixed-up.

leenagupte:
@kevindew @emmabeynon I'll delete the line we talked about on slack, and expand one of the commit messages and then fix-up.

PR #4481, opened 2018-11-08
Export contact lookup tables with news document

kevindew:
Trello: https://trello.com/c/gEAMmD0g/386-ability-to-embed-a-contact-via-markdown

This PR has a couple of small fixes and is then focused on updating ExportNewsDocument to include a lookup table of Whitehall contact id to content_ids so that the bodies of documents can be updated with the appropriate content_id.

Pairs with this PR for content publisher:

### DIFF
--- a/lib/export_news_document.rb
+++ b/lib/export_news_document.rb
@@ -27,7 +27,10 @@ class ExportNewsDocument
   end
 
   def call
-    document.as_json.merge("editions" => editions.as_json)
+    document.as_json.merge(
+      "editions" => editions.as_json,
+      "contacts" => contacts.as_json,
+    )
   end
 
 private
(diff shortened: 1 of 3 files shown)


kevindew:
@benthorner I toyed with that but I decided against as that is Whitehall expressly processing data for Content Publisher and the rest of this is very much a "here's all the data make best use of it" - so felt it took an inconsistent turn.

benthorner:
I think I understand your approach, although I can't imagine how anyone would find the meaningless numbers useful. But I also don't think it matters very much how we do it ;-).

kevindew:
@benthorner sure - if we reach a point when we're massaging this data specially for Content Publisher then it seems a good candidate but I just didn't want to introduce inconsistencies at this point.

PR #4489, opened 2018-11-13
Ensure email subsciptions use taxon_tree instead of taxons

koetsier:
If using the 'taxons' key when creating email signups, subscribers
will only receive emails tagged to that particular taxon.

What we want instead is that subcribers receive emails tagged to
the specified taxon _and_ all its children

This change uses the 'taxon_tree' tag instead of the 'taxons'
tag so find/create subscriber lists, which will enable this.

### DIFF
--- a/lib/url_to_subscriber_list_criteria.rb
+++ b/lib/url_to_subscriber_list_criteria.rb
@@ -11,20 +11,7 @@ class UrlToSubscriberListCriteria
   end
 
   def convert
-    @convert ||= begin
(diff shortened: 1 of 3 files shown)


sihugh:
Is there another thing to update existing taxon email subscriptions to do this?

koetsier:
> Is there another thing to update existing taxon email subscriptions to do this?

No. But that does need to be written - I thought in a new PR. fix this issue first and then the bad data.

PR #4504, opened 2018-11-20
Change world location news pages to use rummager

emmabeynon:
For https://trello.com/c/IILjvKUu/407-change-world-location-news-pages-to-use-rummager

This changes world location news pages and their accompanying atom feeds to use Rummager to  retrieve their documents instead of the Whitehall database when the locale is set to English. This allows content published by Content Publisher to surface on these pages. Currently, Content Publisher does not support translated content so we ensure that the announcements section continues to retrieve documents from the Whitehall database when the locale is not English.


app/controllers/world_location_news_controller.rb
```diff
@@ -30,4 +38,23 @@ def index
   def load_world_location
     @world_location = WorldLocation.with_translations(I18n.locale).find(params[:world_location_id])
   end
+
+  def fetch_documents(additional_options = {})
```
1pretz1: After i've merged my pr is the plan to refactor this code to use the SearchRummagerService?

emmabeynon: Yep that's right 😸

PR #4506, opened 2018-11-20
Use rummager for latest documents pages

1pretz1:
This change enables the latest documents page and atom feed to render results using rummager for topical events, world locations and organisations (latest orgs page has now been replaced by finder frontend pages although the Whitehall page is still accessible by url). 

We need to use Rummager results instead of Whitehall's local database as documents published by Content Publisher get indexed to the govuk index and would not be accessible from the database. 

As part of this work we investigated whether we needed to also retrieve translated documents from the Whitehall database in cases where the locale is not English (Rummager currently doesn't support locales). Our investigation concluded that there are no apparent cases where the locale is set to a foreign language; for example there don't seem to be any instances where a locale param is passed to the Latest Controller. A manual check on live GOV.UK also concluded that links from translated world location news pages (e.g. http://www.gov.uk/world/afghanistan/news.dr) to the latest finder also don't
pass any locale data, and so the results on the page are English only. Consequently we've decided to maintain this current behaviour and only provide English-language Rummager results.

Trello:
https://trello.com/c/aXQWMcmw/406-change-topical-events-latest-pages-to-use-rummager
https://trello.com/c/62hDsXHF/438-change-world-location-latest-pages-to-use-rummager

### DIFF
--- a/app/controllers/latest_controller.rb
+++ b/app/controllers/latest_controller.rb
@@ -7,24 +7,22 @@ class LatestController < PublicFacingController
 
 private
 
+  helper_method :subject, :documents
+
   def subject
     case subject_param
     when 'departments'
       Organisation.with_translations(I18n.locale).find(subject_id)
-    when 'topics'
-      Classification.find(subject_id)
+    when 'topical_events'
+      TopicalEvent.find(subject_id)
     when 'world_locations'
       WorldLocation.with_translations(I18n.locale).find(subject_id)
     end
   end
-  helper_method :subject
 
   def documents
-    Whitehall::Decorators::CollectionDecorator.new(filter.documents,
-                                                   LatestDocumentPresenter,
-                                                   view_context)
+    filter.documents
   end
-  helper_method :documents
 
   def filter
     @filter = LatestDocumentsFilter.for_subject(subject, page_params)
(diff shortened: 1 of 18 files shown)


test/unit/models/latest_documents_filter_test.rb
```diff
@@ -58,9 +55,57 @@ def documents_source
...
+  end
 end
 
 class OrganisationFilterTest < ActiveSupport::TestCase
+  include SearchRummagerHelper
+
+  test '#rummager_documents should return a list of documents for the organisation' do
+    filter = LatestDocumentsFilter::OrganisationFilter.new(organisation)
+
+
```
emmabeynon: Extra line here

app/controllers/latest_controller.rb
```diff
@@ -7,24 +7,22 @@ def index; end
 
 private
 
+  helper_method :subject, :documents
```
kevindew: Oh I didn't about these, they seem quite useful

emmabeynon:
@kevindew Yes to all of the above :)

kevindew:
@emmabeynon woop woop 🚢 :shipit: 🚀

PR #4522, opened 2018-11-26
Present default_news_image for Organisations to Publishing API

emmabeynon:
For https://trello.com/c/3aDALgY4/416-we-should-expose-default-organisation-images-to-the-publishing-api

When an organisation has a default news image, we present the URL for it and
its high res URL to the Publishing API.

### DIFF
--- a/app/presenters/publishing_api/organisation_presenter.rb
+++ b/app/presenters/publishing_api/organisation_presenter.rb
@@ -73,7 +73,7 @@ module PublishingApi
     end
 
     def details
-      {
+      details = {
         acronym: acronym,
         body: summary,
         brand: brand,
@@ -100,6 +100,8 @@ module PublishingApi
         organisation_type: organisation_type,
         social_media_links: social_media_links,
       }
+      details[:default_news_image] = default_news_image if default_news_image
+      details
     end
 
     def acronym
@@ -494,6 +496,25 @@ module PublishingApi
 
     def roles_links
       item.roles.distinct.pluck(:content_id)
+    end
+
+    def default_news_image
+      return unless item.default_news_image
+      return { url: default_news_image_url } if default_news_image_is_svg?
+
+      {
+        url: default_news_image_url(:s300),
+        high_resolution_url: default_news_image_url(:s960),
+      }
+    end
+
+    def default_news_image_url(size = nil)
+      size ? item.default_news_image.file.url(size) : item.default_news_image.file.url
+    end
+
+    def default_news_image_is_svg?
+      content_type = item.default_news_image.file.content_type
+      content_type && content_type =~ /svg/
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/organisation_presenter.rb
```diff
@@ -495,5 +497,13 @@ def high_profile_groups_links
     def roles_links
       item.roles.distinct.pluck(:content_id)
     end
+
+    def default_news_image
+      return unless item.default_news_image
+      {
+        url: item.default_news_image.file_url,
+        high_resolution_url: item.default_news_image.file.url(:s960),
```
kevindew: Do we need to wrap these in image_url helpers as per: https://github.com/alphagov/whitehall/blob/1311c133ee97c51e06c62a67139c68454222876d/app/presenters/publishing_api/news_article_presenter.rb#L129 ?

app/presenters/publishing_api/organisation_presenter.rb
```diff
@@ -495,5 +497,13 @@ def high_profile_groups_links
     def roles_links
       item.roles.distinct.pluck(:content_id)
     end
+
+    def default_news_image
+      return unless item.default_news_image
+      {
+        url: item.default_news_image.file_url,
```
kevindew: Tracing back through whitehall code I think we'll need to set a size on this too:

https://github.com/alphagov/whitehall/blob/1311c133ee97c51e06c62a67139c68454222876d/app/presenters/lead_image_presenter_helper.rb#L35-L42

app/presenters/publishing_api/organisation_presenter.rb
emmabeynon: I compared the output of the helpers and the file urls and they were exactly the same, so I felt the helpers were a bit redundant.

app/presenters/publishing_api/organisation_presenter.rb
emmabeynon: Ah thanks, I think I missed the bit of logic that checked if the content type was SVG.

app/presenters/publishing_api/organisation_presenter.rb
kevindew: Ah really, is it legacy since the asset manager change? 

Feel bold enough to kill it off in the NewsArticlePresenter?

emmabeynon: I'll test this out on integration; if it's all ok then I'll open a separate PR to sort out NewsArticlePresenter :)

app/presenters/publishing_api/organisation_presenter.rb
```diff
@@ -495,5 +497,24 @@ def high_profile_groups_links
...
+        url: default_news_image_url(:s300),
+        high_resolution_url: default_news_image_url(:s960),
+      }
+    end
+
+    def default_news_image_url(size = nil)
+      size ? item.default_news_image.file.url(size) : item.default_news_image.file.url
+    end
+
+    def svg?
```
kevindew: Minor, but I'd suggest this method name should either mention default news image or take in the image as an argument - just that it doesn't make too much sense to ask the presenter as a whole if it is an svg.

app/presenters/publishing_api/organisation_presenter.rb
kevindew: Yup separate sounds better

PR #4539, opened 2018-12-05
Reslug preparing-for-a-uk-trade-policy

alex-ju:
https://govuk.zendesk.com/agent/tickets/3490199


thomasleese:
There's actually a Rake task you can use here: https://github.com/alphagov/whitehall/blob/c2c1048bca075352028c93bd9306f46c210a206a/lib/tasks/reslugging.rake#L43

alex-ju:
@thomasleese thanks for pointing me to rake task and @elliotcm for helping to compose it.

PR #4546, opened 2018-12-10
Permit parameters in latest controller

barrucadu:
This will hopefully fix /government/latest?departments[0]=department-for-education just redirecting to /government/feed rather than giving a filtered feed.

---

[Trello card](https://trello.com/c/fQUnpTNh/678-from-load-testing-fix-government-latest-filtering)

### DIFF
--- a/app/controllers/latest_controller.rb
+++ b/app/controllers/latest_controller.rb
@@ -29,16 +29,16 @@ private
   end
 
   def page_params
-    { page: params.fetch(:page, 0) }
+    { page: filtered_params.fetch(:page, 0) }
   end
 
   def subject_id
-    params[subject_param].first
+    filtered_params[subject_param].first
   end
 
   def subject_param
     supported_subjects.find do |param_name|
-      params[param_name].present? && params[param_name].is_a?(Array)
+      filtered_params[param_name].present? && filtered_params[param_name].is_a?(Array)
     end
   end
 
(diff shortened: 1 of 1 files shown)


kevindew:
@barrucadu I don't think this is a great idea. I'll slack you.

kevindew:
Ah I may have missed you on Slack, but anyway since this is query string parameters this may cause unexpected 400 responses. 

This sets us to only accept "page, departments, topical_events, world_locations" query parameters but it's quite possible there could be other ones such as google analytics tracking as query string parameters, but they could have an error see:

```
[5] pry(main)> ActionController::Parameters.new("utm-content" => "this").permit("page")
ActionController::UnpermittedParameters: found unpermitted parameter: :utm-content
from /usr/lib/rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/actionpack-5.1.6/lib/action_controller/metal/strong_parameters.rb:863:in `unpermitted_parameters!'
```

So in general we shouldn't really use [permit](https://edgeapi.rubyonrails.org/classes/ActionController/Parameters.html#method-i-permit) unless we've already done a [require](https://edgeapi.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require)

To achieve the goal you're aiming for here it might be more appropriate to raise an exception if https://github.com/alphagov/whitehall/blob/becd111a979463e581435c0d7dbf0a8ea6e8614a/app/controllers/latest_controller.rb#L41 is not an array - maybe https://github.com/rails/rails/blob/bad1041b82df941d588ae2565f62424d88104933/actionpack/lib/action_controller/metal/strong_parameters.rb#L38 but I'm not sure

PR #4563, opened 2018-12-19
Rename invisible_draft_taxons to invisible_taxons

cbaines:
This was renamed elsewhere, but looks to have been missed in these places.


cbaines:
Merging without the E2E tests, as we want this change sooner rather than later.

PR #4567, opened 2018-12-20
Add summary for organisations

benthorner:
https://govuk.zendesk.com/agent/tickets/3536161

Previously we set the description to nil for all organisation, which
adversely affects the description metadata we use for search as it's
generated automatically by the engine. This set the description based on
the Govspeak-rendered summary of the organisation.


benthorner:
Superseded by https://github.com/alphagov/whitehall/pull/4569.

PR #4592, opened 2019-01-09
Add authentication to Rummager calls

karlbaker02:
This PR adds bearer token authentication to Rummager calls via the Rest client. Authentication was previously added to the Rummager GDS Api adapter, but Whitehall does not use this for all requests.

### DIFF
--- a/lib/whitehall/rummageable.rb
+++ b/lib/whitehall/rummageable.rb
@@ -92,7 +92,7 @@ module Whitehall
         response = nil
         log_request(method, *args)
         call_time = Benchmark.realtime do
-          response = RestClient.send(method, *args, content_type: :json, accept: :json)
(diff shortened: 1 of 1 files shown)


lib/whitehall/rummageable.rb
```diff
@@ -92,7 +92,7 @@ def make_request(method, *args)
         response = nil
         log_request(method, *args)
         call_time = Benchmark.realtime do
-          response = RestClient.send(method, *args, content_type: :json, accept: :json)
+          response = RestClient.send(method, *args, content_type: :json, accept: :json, authorization: "Bearer #{ENV['RUMMAGER_BEARER_TOKEN'] || 'example'}")
```
kevindew: might be slightly cleaner to do: `ENV.fetch('RUMMAGER_BEARER_TOKEN', 'example')`

karlbaker02: True, I think this was following the existing pattern elsewhere.

PR #4598, opened 2019-01-17
Add a data migration to unpublish old Policy Publisher content

cbaines:
This doesn't specifically relate to Whitehall, it's just an easy place
to put this code.

A single email signup page was left behind when Policy Publisher was
retired, so unpublish that and replace it with a redirect to the email
signup page for the replacement Topic in the Topic Taxonomy.

Policy Publisher also published a finder for policies, so unpublish
that, and replace it with a redirect to the homepage, as there isn't a
great alternative for that specific page.

https://trello.com/c/0jgKr3UT/105-policy-publisher-is-not-fully-retired


cbaines:
Sure, I've added a link to the description now.

PR #4610, opened 2019-01-24
Add a Rake Task to remove a change note

edwardkerry:
This commit adds a Rake task to allow developers to quickly remove a change note from a document.

It provides two interfaces to allow a dry run first, returning the change note that would be deleted. A real run will then update the document's edition to a minor version and the change note text to nil.

```
data_hygiene:remove_change_note:dry[content_id,locale,query]
data_hygiene:remove_change_note:real[content_id,locale,query]
```

Co-authored-by: Thomas Leese <thomas.leese@digital.cabinet-office.gov.uk>
[Trello](https://trello.com/c/eE4pp5Zm/721-rake-task-for-removing-changenotes-in-whitehall)
[Related PR in publishing-API](https://github.com/alphagov/publishing-api/pull/1426)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem 'deprecated_columns', '~> 0.1.1'
 gem 'equivalent-xml', '~> 0.6.0', require: false
 gem 'faraday'
 gem 'friendly_id', '~> 5.2.4'
+gem "fuzzy_match", "~> 2.1"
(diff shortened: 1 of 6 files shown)


kevindew:
I think this should also update the timestamps on the edition. I think each edition stores when it was major published and you'll need to update this to the previous major edition when you make an edition a minor change.

PR #4614, opened 2019-01-29
Adds data migration to change roles of DCMS ministers

oscarwyatt:
Changes the roles appointments for some ministers due to a naming issue around being "Secretary of State for Culture, Media and Sport", "Secretary of State for Culture, Olympics, Media and Sport" or "Secretary of State for Digital, Culture, Media and Sport"

There's a slight gotcha in that speeches, consulations etc are linked to a role appointment so those also have to be moved to the correct role appointments

Created in response to zendesk ticket
https://govuk.zendesk.com/agent/tickets/3570476

### DIFF
--- a/db/data_migration/20190128144735_change_dcms_roles.rb
+++ b/db/data_migration/20190128144735_change_dcms_roles.rb
@@ -1,0 +1,47 @@
+ActiveRecord::Base.transaction do
+  p "Changing role for Matthew Hancock"
+  matt_hancock = Person.where(forename: "Matt", surname: "Hancock").first
+  correct_role = Role.find_by(name: "Secretary of State for Digital, Culture, Media and Sport")
+  incorrect_role_one = Role.find_by(name: "Secretary of State for Culture, Media and Sport")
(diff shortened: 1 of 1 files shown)


db/data_migration/20190128144735_change_dcms_roles.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  correct_role_appointment = matt_hancock.role_appointments.where(role_id: correct_role.id).first
+
+  # Update consulations, publications etc. for old role appointments that must now go to the current role appointment
+  [incorrect_role_one, incorrect_role_two].each do |role|
+    incorrect_role_appointment = matt_hancock.role_appointments.where(role_id: role.id).first
+    EditionRoleAppointment.where(role_appointment_id: incorrect_role_appointment).each do |edition_role_appointment|
+      edition_role_appointment.update!(role_appointment_id: correct_role_appointment.id)
+    end
+    Edition.where(role_appointment_id: role.id).each do |edition|
```
bevanloon: This is very minor and I wouldn't let it hold up the merging, but would it be worth expressing this as `Speech.where` to make it explicit that it's only speeches here (also on line 41 below)?

PR #4643, opened 2019-02-15
Rename "Macedonia" to "North Macedonia"

deborahchua:
Trello card: https://trello.com/c/ySMq2bcm/786-slug-and-tag-changes-country-name-change-macedonia-to-north-macedonia


db/migrate/20190215171522_rename_macedonia.rb
```diff
@@ -0,0 +1,3 @@
+macedonia = WorldLocation.find_by(slug: "macedonia")
+macedonia.update(slug: "north-macedonia")
```
thomasleese: Might be worth using `update!` here so we get an exception if it fails for whatever reason.

AlanGabbianelli: Right

PR #4645, opened 2019-02-18
Replace announcements lists with news and comms lists on world location news pages

bilbof:
https://trello.com/c/u6doZAVR/373-replace-announcements-lists-with-news-and-comms-lists-on-world-location-news-pages

We have created a new finder within finder-frontend called [news and communications](https://www.gov.uk/news-and-communications).

This change ensures that the information under news and communications on the `/world/:slug/news` pages matches the content of the news and comms finder (they both filter results based on content_purpose_supergroup: news_and_communications).

The link to the announcements finder has also been replaced.

The goal is to deprecate the announcements finder and replace it with the news and communications finder; replacing the links is required before we can do this.

Before: https://www.gov.uk/world/jamaica/news

screen shot 2019-02-18 at 11 51 18

After:

screen shot 2019-02-18 at 11 51 13

### DIFF
--- a/app/helpers/filter_routes_helper.rb
+++ b/app/helpers/filter_routes_helper.rb
@@ -3,6 +3,10 @@ module FilterRoutesHelper
     announcements_path(path_arguments(objects))
   end
 
+  def news_and_communications_filter_path(*objects)
+    "/news-and-communications?#{path_arguments(objects).to_query}"
+  end
+
   def publications_filter_path(*objects)
     publications_path(path_arguments(objects))
   end
--- a/app/helpers/filter_routes_helper.rb
+++ b/app/helpers/filter_routes_helper.rb
@@ -3,10 +3,6 @@ module FilterRoutesHelper
     announcements_path(path_arguments(objects))
   end
 
-  def news_and_communications_filter_path(*objects)
-    "/news-and-communications?#{path_arguments(objects).to_query}"
-  end
-
   def publications_filter_path(*objects)
     publications_path(path_arguments(objects))
   end
--- a/test/functional/world_location_news_controller_test.rb
+++ b/test/functional/world_location_news_controller_test.rb
@@ -21,17 +21,17 @@ class WorldLocationNewsControllerTest < ActionController::TestCase
     }
   end
 
-  def announcement_search_options
+  def news_and_communications_search_options
     default_search_options.merge(
       count: 2,
-      filter_content_store_document_type: announcement_document_types
+      filter_content_purpose_supergroup: 'news_and_communications'
     )
   end
 
-  def announcement_document_types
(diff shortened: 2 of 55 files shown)


test/functional/world_location_news_controller_test.rb
```diff
@@ -21,17 +21,17 @@ def default_search_options
...
-      filter_content_store_document_type: announcement_document_types
+      filter_content_purpose_supergroup: 'news_and_communications'
     )
   end
 
-  def announcement_document_types
-    non_world_announcement_types = Whitehall::AnnouncementFilterOption.all.map(&:document_type).flatten
-    %w(world_location_news_article world_news_story).concat(non_world_announcement_types)
-  end
+  # def announcement_document_types
```
sihugh: Should these be removed?

app/helpers/filter_routes_helper.rb
```diff
@@ -3,6 +3,10 @@ def announcements_filter_path(*objects)
     announcements_path(path_arguments(objects))
   end
 
+  def news_and_communications_filter_path(*objects)
```
sihugh: is this used anywhere?  I haven't found it yet

sihugh:
I'm wondering whether we should use the existing translations for `Our announcements` until we can swap them out for updated translations.

test/functional/world_location_news_controller_test.rb
```diff
@@ -21,17 +21,17 @@ def default_search_options
     }
   end
 
-  def announcement_search_options
+  def news_and_communications_search_options
     default_search_options.merge(
       count: 2,
-      filter_content_store_document_type: announcement_document_types
+      filter_content_purpose_supergroup: 'news_and_communications'
```
karlbaker02: Do we also need to filter on document type as we did previously, to ensure that only world items are being shown?

app/helpers/filter_routes_helper.rb
Seraphiana: weirdly, it doesn't seem to show up anywhere in the code base, but when removed proceeds to break the tests. Which is pretty odd?

bilbof: The joy of dynamic programming. It's created here: app/views/classifications/_document_list.html.erb

```
      see_more_url = public_send(
        "#{type}_filter_path",
        @classification,
        publication_filter_option: filter_option
      )
```

test/functional/world_location_news_controller_test.rb
Seraphiana: apparently this shouldn't be an issue because the news and comms supergroup already has the world news type as part of the supergroup, as seen [here](https://github.com/alphagov/govuk_document_types/blob/master/data/supertypes.yml#L504)

test/functional/world_location_news_controller_test.rb
Seraphiana: removed

PR #4650, opened 2019-02-22
Add data migration to update document URLs

bilbof:
Requested here: https://govuk.zendesk.com/agent/tickets/3590864

Once merged and deployed, we'll need to run the rake task `db:data:migrate`.

I have run this on integration and it looks good. Rake task: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/24947/console.

### DIFF
--- a/db/data_migration/20190222130820_reslug_documents.rb
+++ b/db/data_migration/20190222130820_reslug_documents.rb
@@ -1,0 +1,56 @@
+puts "Reslugging documents..."
+
+documents = [
+  {
+    slug: "letter-to-providers-on-health-and-safety",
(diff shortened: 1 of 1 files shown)


db/data_migration/20190222130820_reslug_documents.rb
```diff
@@ -0,0 +1,56 @@
+puts "Reslugging documents..."
+
+documents = [
+  {
+    slug: "letter-to-providers-on-health-and-safety",
+    new_slug: "letters-from-the-regulator-of-social-housing-Grenfell",
```
kevindew: shall we lowercase this - since as you say it'll do it automatically? at least makes our expectations clearer.

PR #4660, opened 2019-03-04
Fix the Remove Policy Publisher Content data migration

cbaines:
A draft exists for the email-signup page, so set discard_drafts to
true for this item.


cbaines:
I've requested a review from Kevin, but I'm fine with anyone else reviewing this as well.

cbaines:
Thanks Kevin :+1:

PR #4665, opened 2019-03-08
Shortcut access to content_store_document_type

elliotcm:
Fleshing out the full content of a publishing API presenter is a hugely expensive operation, so let's skip straight to the end result which is a method called on this object anyway.

This represents at least a 50% reduction in loading time for the tag edit form, more so if the database is heavily loaded at the time.

https://trello.com/c/GQJ2AEsS/827-make-the-edit-tags-publishing-step-more-resilient

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -659,7 +659,7 @@ class Edition < ApplicationRecord
   end
 
   def content_store_document_type
-    PublishingApiPresenters.presenter_for(self).content.fetch(:document_type)
+    PublishingApiPresenters.presenter_for(self).document_type
   end
 
   def has_legacy_tags?
--- a/app/presenters/publishing_api/case_study_presenter.rb
+++ b/app/presenters/publishing_api/case_study_presenter.rb
@@ -19,7 +19,7 @@ module PublishingApi
       content.merge!(
         description: item.summary,
         details: details,
-        document_type: "case_study",
+        document_type: document_type,
         public_updated_at: item.public_timestamp || item.updated_at,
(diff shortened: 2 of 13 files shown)


app/models/edition.rb
```diff
@@ -659,7 +659,7 @@ def detailed_format
   end
 
   def content_store_document_type
-    PublishingApiPresenters.presenter_for(self).content.fetch(:document_type)
+    display_type_key
```
bevanloon: we also seem to have `content_store_document_type` which calls `display_type_key`. I don't know if that's neater or just yet another piece of obfuscation.

elliotcm: @bevanloon I'm confused about this, this is that method?

cbaines: How confident are you that this is always the case? Searching through the presenters module shows examples where document_type is set to display_type_key, and where it's not, but obviously it's not clear cut, as not all of the things in that module are relevant to editions...

elliotcm: I'll have a look

elliotcm: The only outlier was detailed guides (which for some reason is `detailed_guidance` in whitehall but `detailed_guide` in the publishing API`), so to avoid confusion I've put the code back to leaning on the presenters but without instantiating all of the content just to get at what is mostly a static string on the presenter.

bevanloon: bleach....yeah never mind I misread :( carry on.

cbaines: Ok, that looks good :+1:

PR #4667, opened 2019-03-08
Add rake task to report the number of draft...

deborahchua:
...publications by organisation within a given date range.

This rake task will generate a report to CSV as well as output in
the console.
Running this task weekly will mean we can monitor and see if there
is an increase in draft publications.

Trello card: https://trello.com/c/rLXkQ9Nj/835-find-out-how-many-drafts-exist-in-the-system-at-a-given-time

### DIFF
--- a/lib/tasks/reporting.rake
+++ b/lib/tasks/reporting.rake
@@ -80,6 +80,46 @@ namespace :reporting do
     end
   end
 
+  desc "A CSV report of the number of publications in draft state by organisation between the dates specified [Month Day Year 00:00:00]"
+  task :number_of_draft_publications_by_organisation_by_date_range, %i[start_date end_date] => :environment do |_t, args|
+    if args[:start_date].present? && args[:end_date].present?
+      start_date = Time.parse(args[:start_date])
+      end_date = Time.parse(args[:end_date])
+    end
+
+    date_range = start_date...end_date
+
+    path = "#{Rails.root}/tmp/number_of_draft_publications_by_organisation_between_#{start_date}_to_#{end_date}.csv"
+
+    CSV_HEADERS = ["Lead publishing organisation", "Number of draft publications"].freeze
+
+    CSV.open(path, "wb", headers: CSV_HEADERS, write_headers: true) do |csv|
+      Edition.include(Edition::Organisations)
+
+      puts "Searching for draft publications between #{start_date} and #{end_date}"
+
+      draft_publications = Edition.latest_edition
+      .joins(:edition_organisations)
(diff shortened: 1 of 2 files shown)


lib/tasks/reporting.rake
```diff
@@ -80,6 +80,47 @@ namespace :reporting do
...
+      start_date = Time.parse(args[:start_date])
+      end_date = Time.parse(args[:end_date])
+    end
+    date_range = start_date...end_date
+
+    path = "#{Rails.root}/tmp/number_of_draft_publications_by_organisation_between_#{start_date}_to_#{end_date}.csv"
+
+    CSV_HEADERS = ["Lead publishing organisation", "Number of draft publications"].freeze
+
+    CSV.open(path, 'wb') do |csv|
```
thomasleese: You can actually do `CSV.open(path, "wb", headers: CSV_HEADERS, write_headers: true)` here to avoid the line below.

lib/tasks/reporting.rake
```diff
@@ -80,6 +80,47 @@ namespace :reporting do
     end
   end
 
+  desc "A CSV report of the number of publications in draft state by organisation between the dates specified [year-month-day-hour-min]"
+  task :number_of_draft_publications_by_organisation_by_date_range, %i[start_date end_date] => :environment do |_t, args|
+    if args[:start_date].present? && args[:end_date].present?
+      start_date = Time.parse(args[:start_date])
+      end_date = Time.parse(args[:end_date])
+    end
```
thomasleese: Should we raise an error here if these arguments are missing? Perhaps like:

```ruby
raise "Missing start_time or end_time" unless args[:start_date].present? && args[:end_date].present?
```

and then we don't need the `if` block.

lib/tasks/reporting.rake
```diff
@@ -80,6 +80,47 @@ namespace :reporting do
...
+
+      draft_publications = Edition.latest_edition
+      .joins(:edition_organisations)
+      .where(state: "draft")
+      .where(type: "publication")
+      .where(updated_at: date_range)
+
+      organisations = Organisation.order(slug: :asc)
+
+      organisations.each do |organisation|
```
thomasleese: Although it might not make a difference here, `.find_each` can often be quicker as it avoids allocating a new object for each row in the database and reuses the same one each time.

deborahchua:
Thanks @thomasleese ! I have implemented your suggestions and moved the code into its own class in https://github.com/alphagov/whitehall/pull/4667/commits/601976abae5f51a11d22311c967609961a92ca8a

PR #4668, opened 2019-03-11
Optimise parts used in force publishing

thomasleese:
This optimises a few of the operations done in Whitehall, specifically aiming at improving the performance of force publishing.

See commits for individual details on optimisations.

Adding the index to `updated_at` in integration took less than 2 seconds, however it looks like MySQL allows adding of indexes without blocking the table: https://dev.mysql.com/doc/refman/5.6/en/innodb-online-ddl-operations.html

```
10:25:44 *** [err :: ip-10-1-4-230.eu-west-1.compute.internal] == 20190312101215 AddIndexToEditionUpdatedAt: migrating =======================
10:25:44 *** [err :: ip-10-1-4-230.eu-west-1.compute.internal] -- add_index(:editions, :updated_at)
10:25:46 *** [err :: ip-10-1-4-230.eu-west-1.compute.internal] -> 1.8123s
10:25:46 *** [err :: ip-10-1-4-230.eu-west-1.compute.internal] == 20190312101215 AddIndexToEditionUpdatedAt: migrated (1.8124s) ==============
```

[Trello Card](https://trello.com/c/H3MnHlyx/833-make-the-force-publish-step-more-resilient)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -90,7 +90,10 @@ class Document < ApplicationRecord
   end
 
   def published?
-    reload_published_edition.present?
+    Edition.exists?(
+      state: Edition::PUBLICLY_VISIBLE_STATES,
+      document_id: id,
+    )
   end
 
   def first_published_date
--- a/app/models/edition/publishing.rb
+++ b/app/models/edition/publishing.rb
@@ -42,7 +42,7 @@ module Edition::Publishing
     elsif draft? && new_record?
       false
     else
-      other_editions.published.any?
+      other_editions.published.exists?
     end
   end
 
--- a/app/services/edition_publisher.rb
+++ b/app/services/edition_publisher.rb
@@ -41,12 +41,18 @@ private
     delete_unpublishing!
   end
 
+  def previous_editions
+    Edition
+      .unscoped
+      .where(document_id: edition.document_id)
+      .where.not(id: edition.id)
+      .published
+  end
+
   def supersede_previous_editions!
-    edition.document.editions.published.each do |e|
-      if e != edition
-        e.supersede
-        e.save(validate: false)
-      end
+    previous_editions.each do |edition|
+      edition.supersede
+      edition.save(validate: false)
     end
   end
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -332,7 +332,7 @@ class Edition < ApplicationRecord
   end
 
   def creator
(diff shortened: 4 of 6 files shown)


app/models/document.rb
```diff
@@ -90,7 +90,10 @@ def update_slug_if_possible(new_title)
   end
 
   def published?
-    reload_published_edition.present?
+    Edition
+      .where(state: Edition::PUBLICLY_VISIBLE_STATES)
+      .where(document_id: id)
+      .exists?
```
elliotcm: You can combine all these together:
```
Edition.exists?(
  state: Edition::PUBLICLY_VISIBLE_STATES,
  document_id: id,
)
```

app/models/edition/publishing.rb
```diff
@@ -42,7 +42,7 @@ def change_note_required?
     elsif draft? && new_record?
       false
     else
-      other_editions.published.any?
```
elliotcm: This *should* be the same effect, what are you going for with this change?

app/services/edition_publisher.rb
```diff
@@ -41,12 +41,18 @@ def fire_transition!
     delete_unpublishing!
   end
 
+  def previous_editions
+    Edition
+      .unscoped
```
elliotcm: Are you sure you want to `unscoped` this?  That would include soft-deleted editions.

app/models/edition/publishing.rb
thomasleese: I've read in the past that `SELECT EXISTS` can be optimised by the DB more so than `SELECT 1 WHERE`, but it really is a minor/insignificant change at this point. I just think it reads better.

app/services/edition_publisher.rb
thomasleese: The `.published` means you get `.where(state: :published)` so it won't included soft-deleted editions.

app/models/document.rb
thomasleese: Nice! I've pushed up that change 👍

app/services/edition_publisher.rb
elliotcm: Gotcha.  What scope are we removing in that case?

thomasleese: `.where.not(state: :deleted)` - I imagine (or hope) it would get optimised away by the DB, but I  don't like using or relying on default scopes so I decided to include the `.unscoped`.

PR #4671, opened 2019-03-12
Make it possible to redirect static pages

bilbof:
https://trello.com/c/Xbz5SvfD/350-allow-publishing-of-whitehall-finders-routes-to-finder-frontend

This will enable us to run a rake task to remove pages
published via PublishStaticPages such as
/government/announcements.

The new rake task redirect_static_pages will replace
specific static pages with redirects to new URLs and
will remove them from the search index.

The logic of RedirectStaticPages is a reverse of
PublishStaticPages, so if you read both that should
help explain the intent of the new class.

### DIFF
--- a/lib/redirect_static_pages.rb
+++ b/lib/redirect_static_pages.rb
@@ -1,0 +1,24 @@
+# This enables us to redirect static pages published by PublishStaticPages
+# to specified URLs.
+
+class RedirectStaticPages
+  def redirect
+    pages.each do |page|
+      PublishingApiRedirectWorker.new.perform(page[:content_id], page[:target_path], "en")
+    rescue GdsApi::HTTPNotFound => e
+      # we can't unpublish something that doesn't exist
+      puts e
+      nil
+    end
+  end
+
+  def pages
+    [
+      {
+        base_path: "/government/announcements",
+        content_id: "88936763-df8a-441f-8b96-9ea0dc0758a1",
+        target_path: "/redirect/announcements",
+      },
+    ]
+  end
+end
--- a/lib/publish_static_pages.rb
+++ b/lib/publish_static_pages.rb
@@ -92,14 +92,6 @@ class PublishStaticPages
         locales: Locale.non_english.map(&:code),
       },
       {
-        content_id: "88936763-df8a-441f-8b96-9ea0dc0758a1",
-        title: "Government announcements",
-        document_type: "finder",
-        description: "Find news articles, speeches and statements from government organisations",
-        base_path: "/government/announcements",
-        locales: Locale.non_english.map(&:code),
-      },
-      {
         content_id: "324e4708-2285-40a0-b3aa-cb13af14ec5f",
         title: "Ministers",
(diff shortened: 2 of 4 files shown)


bilbof:
OK, I removed the locales logic for now.

lib/redirect_static_pages.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+private
+
+  def present_for_publishing_api(page)
+    {
+      content_id: page[:content_id],
+      content: {
+        base_path: page[:base_path],
+        document_type: "redirect",
```
cbaines: Explicitly sending a `redirect` to the Publishing API was somewhat superseded by the `unpublish` action, would it be possible to unpublish rather than publish a redirect here?

bilbof: @cbaines you are right. Thanks a lot, I wasn't aware of this. I amended the commit.

cbaines: Rummager has two different ways of updating it's database, I believe the norm for Whitehall content is for Whitehall to directly talk to Rummager to insert or delete things. Maybe call `SearchIndex.delete` [1]?
1: https://github.com/alphagov/whitehall/blob/master/lib/whitehall/search_index.rb#L16-L18

sihugh: Finders are [migrated](https://github.com/alphagov/rummager/blob/c4ee441b83d252f41531cda4abb0b31fc347b466/config/govuk_index/migrated_formats.yaml#L48) so direct communication with Rummager shouldn't be necessary

bilbof: Fantastic, thank for you for the context @cbaines @sihugh! I've removed the call to rummager from the commit.

lib/redirect_static_pages.rb
```diff
@@ -0,0 +1,24 @@
...
+      # we can't unpublish something that doesn't exist
+      puts e
+      nil
+    end
+  end
+
+  def pages
+    [
+      {
+        base_path: "/government/announcements",
```
koetsier: is base_path needed?

bilbof: Actually no, it just tells you what the current path is. It's not used, so we could remove it if you don't think this is helpful.

lib/redirect_static_pages.rb
```diff
@@ -0,0 +1,24 @@
+# This enables us to redirect static pages published by PublishStaticPages
+# to specified URLs.
+
+class RedirectStaticPages
+  def redirect
+    pages.each do |page|
+      PublishingApiRedirectWorker.new.perform(page[:content_id], page[:target_path], "en")
```
koetsier: I just looked at ```PublishingApiRedirectWorker``` and that only redirects withdrawn content - Is that ok here?

bilbof:
Closing as we've chosen a different approach: https://trello.com/c/nLPRCdKt/529-redirect-announcements-to-news-and-communications

PR #4682, opened 2019-03-15
Redirect announcements to news and communications finder

bilbof:
This will redirect the announcements finder to the news and communications finder.

This option was chosen over creating a redirect content item because:

1. We need to keep sending requests with locales to the old finder,
which is not currently possible with a redirect route
2. We can avoid putting legacy param handling code in finder-frontend

Trello: https://trello.com/c/nLPRCdKt/529-redirect-announcements-to-news-and-communications

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -9,6 +9,8 @@ class AnnouncementsController < DocumentsController
 
     respond_to do |format|
       format.html do
+        return redirect_to_news_and_communications if is_english_locale?
+
         @content_item = Whitehall
           .content_store
           .content_item("/government/announcements")
@@ -33,5 +35,31 @@ private
 
   def scheduled_announcements
     Announcement.scheduled.order("scheduled_publication asc")
+  end
+
+  def is_english_locale?
+    # We only want to redirect English locale requests for now,
+    # as translations for this finder haven't been moved to finder-frontend yet.
+    params[:locale].nil?
+  end
+
+  def redirect_to_news_and_communications
+    base_path = "#{Plek.new.website_root}/search/news-and-communications"
+    redirect_to("#{base_path}?#{news_and_communications_query_string}")
+  end
+
+  def news_and_communications_query_string
+    {
+      keywords: params['keywords'],
+      level_one_taxon: params['taxons'].try(:first),
+      level_two_taxon: params['subtaxons'].try(:first),
+      people: params['people'],
+      organisations: params['departments'],
+      world_locations: params['world_locations'],
+      public_timestamp: {
+        from: params['from_date'],
+        to: params['to_date']
+      }.compact.presence,
+    }.compact.to_query
   end
 end
--- a/features/announcements.feature
+++ b/features/announcements.feature
@@ -5,8 +5,8 @@ Background:
 
 @not-quite-as-fake-search
 Scenario: Viewing news articles and speeches together on the news and speeches page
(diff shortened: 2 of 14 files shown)


app/controllers/announcements_controller.rb
```diff
@@ -34,4 +36,28 @@ def index
   def scheduled_announcements
     Announcement.scheduled.order("scheduled_publication asc")
   end
+
+  def is_english_locale?
```
oscarwyatt: Unlike me to ever say something about comments (I've just read something, thinking of changing my ways). Anyway, might be worth a comment on why we're only redirecting english locales

app/controllers/announcements_controller.rb
```diff
@@ -34,4 +36,28 @@ def index
...
+    params[:locale].nil?
+  end
+
+  def redirect_to_news_and_communications
+    base_path = "#{Plek.new.website_root}/search/news-and-communications"
+    redirect_to("#{base_path}?#{news_and_communications_query_string}")
+  end
```
oscarwyatt: I _think_ we don't need the fully path so it could become
```suggestion
    redirect_to("/search/news-and-communications?#{news_and_communications_query_string}")
  end
```
But shan't to demur if there is a good reason

app/controllers/announcements_controller.rb
```diff
@@ -34,4 +36,28 @@ def index
...
+      level_one_taxon: params['taxons'].try(:first),
+      level_two_taxon: params['subtaxons'].try(:first),
+      people: params['people'],
+      organisations: params['departments'],
+      world_locations: params['world_locations'],
+      public_timestamp: {
+        from: params['from_date'],
+        to: params['to_date']
+      }.compact.presence,
+    }.to_query
```
oscarwyatt: Would be good to filter out any parameters that are empty, I tried going to `government/announcements` and it redirected me to `search/news-and-communications?keywords=&level_one_taxon=&level_two_taxon=&organisations=&people=&public_timestamp=&world_locations=` which is a bit ungainly

oscarwyatt:
Made a few pedantic comments were it to be more production-ized but the general approach is good (plus, tests!)

app/controllers/announcements_controller.rb
bilbof: I've put a note about this in the commit message. Would you also like a comment in the code?

app/controllers/announcements_controller.rb
bilbof: Good idea - done!

app/controllers/announcements_controller.rb
bilbof: Redirecting to the root over the path is more specific, so there's less chance it can go wrong. It'll also work in development, when running Whitehall in isolation.

app/controllers/announcements_controller.rb
```diff
@@ -9,6 +9,8 @@ def index
 
     respond_to do |format|
       format.html do
+        return redirect_to_news_and_communications if is_english_locale?
```
koetsier: I'm not sure about the 'return' here. I've seen ```redirect_path <path> and return``` or ```if <condition> redirect_to <path> else render <view>```

bilbof:
I made some changes to get the feature tests to pass. Where applicable I switched the tests to use a non-Whitehall Mysql db rather than the search-api to source non-English documents. Where not applicable I just removed the tests (e.g. signing up to emails from a French locale, or filtering docs in most ways isn't possible).

PR #4686, opened 2019-03-18
Improve loading attachments for publication

deborahchua:
As part of our load testing improvements, we want to make the
publication overview page more resilient.

There are three types of `Attachments`: `FileAttachment`, `HtmlAttachment`,
and `ExternalAttachment`. They all share the same view when loaded
into the publication overview and currently call the attachments of an
`edition` within the view. Each `attachment` requires it's `attachment_data`,
so including it here should optimise this query.

Co-authored-by: Alan Gabbianelli <alan.gabbianelli@digital.cabinet-office.gov.uk>

Trello card: https://trello.com/c/uHiHYiru/855-make-the-visit-publication-overview-step-more-resilient

### DIFF
--- a/app/views/admin/attachments/_attachments.html.erb
+++ b/app/views/admin/attachments/_attachments.html.erb
@@ -9,7 +9,7 @@
 
 <%= form_tag [:order, :admin, typecast_for_attachable_routing(attachable), Attachment], method: :put do %>
   <ol class="existing-attachments list-unstyled">
-    <% attachable.attachments.each do |attachment| %>
+    <% attachable.attachments.includes(:attachment_data).each do |attachment| %>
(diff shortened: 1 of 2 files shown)


app/models/attachment.rb
```diff
@@ -29,6 +29,8 @@ class Attachment < ApplicationRecord
     allow_blank: true, greater_than: 0
   }
 
+  default_scope { includes(:attachment_data) }
```
cbaines: What's your guess for how many places in the code would require changing if this wasn't done through a default scope?

I guess this makes sense if it's always useful, but is that definitely the case?

deborahchua: Hi @cbaines , thanks for this! We reviewed this again and found we can avoid using `default_scope`

cbaines: Awesome :+1:

PR #4716, opened 2019-04-01
Data migration for astana

bilbof:
Ticket: https://govuk.zendesk.com/agent/tickets/3645655

Adds a [data migration](https://github.com/alphagov/whitehall/tree/master/db/data_migration) to be run with `bundle exec rake db:data:migrate` on all environments (there's a jenkins job called `Run_Whitehall_Data_Migrations`).


thomasleese:
There's also a Rake task you can use to avoid having to write a migration: https://github.com/alphagov/whitehall/blob/648f5a5c8685cd3bc0db32783a6eef5a830735a1/lib/tasks/reslugging.rake#L113

bilbof:
Thanks, Tom, I used that instead.

PR #4725, opened 2019-04-08
[DO NOT MERGE] Use the internal name for taxons

barrucadu:
We want to use the internal name for taxons in the tagging interface, rather than the public name, because some taxons have an internal name which conveys more guidance than the public name (for example, saying that a taxon is for GDS use only).

---

[Trello card](https://trello.com/c/8PrfgmAS/582-make-whitehall-tagging-interface-look-at-internal-taxon-name)


sihugh:
@barrucadu, Si would love to have a look at this on integration prior to merging so he can check that it looks sane from a content designer's perspective.  Is that OK?

PR #4730, opened 2019-04-09
Fixes the paginator for empty collections

koetsier:
For empty collections, used for instance in end-to-end testing
the paginator breaks.

This fix addresses this.

### DIFF
--- a/app/presenters/api/page_presenter.rb
+++ b/app/presenters/api/page_presenter.rb
@@ -27,7 +27,7 @@ Api::PagePresenter = Struct.new(:page, :context) do
   end
 
   def next_page_url
-    unless page.last_page?
+    if !page.last_page? && !page.out_of_range?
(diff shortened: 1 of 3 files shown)


koetsier:
last_page? is a delegate to ```Kaminari::PageScopeMethods#last_page?``` which is in a gem.

It is implemented as:

```
    # Last page of the collection?
    def last_page?
      current_page == total_pages
    end
```

For an empty collection ```total_pages==0``` and ```current_page==1``` so it never reaches the last page

bilbof:
Thanks for explaining! Perhaps we could use `|| page.out_of_range?` here

https://www.rubydoc.info/github/amatsuda/kaminari/Kaminari/PageScopeMethods#out_of_range%3F-instance_method

koetsier:
Thanks @bilbof , Think that looks a bit nicer now!

PR #4731, opened 2019-04-11
Redirect html attachments

1pretz1:
For zendesk ticket - https://govuk.zendesk.com/agent/tickets/3655684.

There seemed to be a rake task for this sort of job
(`publishing_api:redirect_html_attachments`) however this finds the html
attachments for the last edition of a document. This is not wanted for
this ticket as the last edition has different attachments to the
attachments requested to be redirected.


1pretz1:
Tested in staging and works as expected!

PR #4732, opened 2019-04-12
Adds redirects for publications to finder frontend

SamJamCul:
This branch currently has the changes needed for the publications finder redirects. However, not all of the tests are passing. I have managed to get all the rails tests green, but the feature tests aren't something I've worked with before, and I haven't managed to get them to pass yet!

If anyone does pick this up, you might also be interested in this [PR](https://github.com/alphagov/whitehall/pull/4682) which does similar stuff to this branch.

https://trello.com/c/68uvW2mk/541-redirect-the-consultations-finder-to-policy-papers-and-consultations-but-in-whitehall


bilbof:
I removed the failing feature tests.

For the most part they are no longer needed, as they relate to filtering on the now removed English publications finder (filtering isn't possible on non-English finders). 

Where they are still relevant (e.g. English documents showing up on publications finder) these are best added to the end to end tests.

barrucadu:
The e2e tests are failing because they're checking that the application which renders the finder page is whitehall.

PR #4735, opened 2019-04-16
Publish default withdrawal notice for all translations of withdrawn edition

edwardkerry:
Currently only the default locale of a withdrawn edition displays a withdrawal notice. This can be confusing for translations, which have no indication that an edition is no longer valid.

This PR will allows all translations to display the given (untranslated) withdrawal notice to offer the user some indication that it is not valid.

### Published
Screen Shot 2019-04-16 at 16 29 26
----
Screen Shot 2019-04-16 at 16 29 36
----
Screen Shot 2019-04-16 at 16 29 48
----
Screen Shot 2019-04-16 at 16 29 57

### Withdrawn
Screen Shot 2019-04-16 at 16 43 45
----
Screen Shot 2019-04-16 at 16 43 54
----
Screen Shot 2019-04-16 at 16 44 05
----
Screen Shot 2019-04-16 at 16 44 12

### DIFF
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -33,3 +33,14 @@ namespace :data_hygiene do
     end
   end
 end
+
+namespace :data_hygiene do
+  desc "Add the withdrawal note to all translations of withdrawn editions."
+  task add_translation_withdrawal_notices: :environment do
+    Edition.where(state: 'withdrawn').find_each do |edition|
+      if edition.translations.count > 1
+        ServiceListeners::PublishingApiPusher.new(edition).push(event: "withdraw")
+      end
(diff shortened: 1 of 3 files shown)


lib/tasks/data_hygiene.rake
```diff
@@ -33,3 +33,14 @@ namespace :data_hygiene do
     end
   end
 end
+
+namespace :data_hygiene do
+  desc "Add the withdrawal note to all translations of withdrawn editions."
+  task add_translation_withdrawal_notices: :environment do
+    Edition.where(state: 'withdrawn').find_each do |edition|
```
AlanGabbianelli: Kudos for using `find_each` 👍

PR #4747, opened 2019-04-26
Add length validation for featured_link url

AgaDufrat:
The URL can have maximum of 255 characters as this is the default length of a
string in an ActiveRecord model. Users were getting a general “We're sorry, but
something went wrong” message when the URL they entered was too long.  This sets
the `maxlength` attribute on the edit form field and adds validation as well as
an error message in the model.

Zendesk ticket: https://govuk.zendesk.com/agent/tickets/3643823

### DIFF
--- a/app/models/featured_link.rb
+++ b/app/models/featured_link.rb
@@ -9,6 +9,7 @@ class FeaturedLink < ApplicationRecord
 
   validates :url, :title, presence: true
   validates :url, uri: true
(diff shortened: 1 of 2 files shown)


brucebolt:
I wonder if, rather than imposing a maximum length, we should update the schema to use a data type other than string that does not have a length limit (e.g. text)?  Is there anything that would stop us from doing this?

PR #4756, opened 2019-04-30
Add rake task to discard wrong draft about pages

AlanGabbianelli:
In certain cases, an about page (CorporateInformationPage) has the same
base_path of its WorldwideOrganisation. This shouldn't be the case and
prevents us from editing the WorldwideOrganisation.

This rake task will discard all the about pages in draft state that have
the same base_path as their WorldwideOrganisation.

These about pages were retrieved by running the following command in the
publishing_api console:
```
Edition
  .where(document_type: "about", state: 'draft')
  .select { |edition| edition.base_path =~ /\A\/world\/organisations\/[^\/]+\z/ }
  .map { |edition| [edition.document.content_id, edition.document.locale] }
  .to_yaml
```

More context can be found in:
  - [this trello card](https://trello.com/c/PlWSEch7/109-whitehall-worldwide-organisation-cips-and-normal-cips-are-different)
  - [this google doc](https://docs.google.com/document/d/16c3kHSo0jWbPYRJiky7GpJq4TWE8dIEnbgvcMhrXX0k/edit#)
  - [this pull request](https://github.com/alphagov/whitehall/pull/4465)
  - [this Zendesk ticket](https://govuk.zendesk.com/agent/tickets/3630111)


lib/tasks/publishing_api.rake
```diff
@@ -122,4 +122,13 @@ namespace :publishing_api do
       DataHygiene::PublishingApiHtmlAttachmentRedirector.call(args[:content_id], args[:destination], dry_run: false)
     end
   end
+
+  desc "Discard all draft about pages that have the same base_path as their WorldwideOrganisation"
+  task discard_draft_worldwide_organisation_about_pages: :environment do
+    about_pages = YAML.load_file(File.join(Rails.root, 'lib', 'tasks', 'about_pages.yml'))
```
brucebolt: Rather than dumping these as a YAML then re-loading them, would it be better to put the query here?  Then the YAML file wouldn't be needed and the rake task would become easily re-usable.
```suggestion
    about_pages = Edition
   .where(document_type: "about", state: 'draft')
   .select { |edition| edition.base_path =~ /\A\/world\/organisations\/[^\/]+\z/ }
   .map { |edition| [edition.document.content_id, edition.document.locale] }
```

AlanGabbianelli: That's not possible because that query has to be run in `publishing_api`.
`Edition` in `whitehall` is different from `Edition` in `publishing_api`:
```
irb(main):003:0> # This is the whitehall console
=> nil
irb(main):004:0> Edition.where(document_type: "about")
Traceback (most recent call last):
ActiveRecord::StatementInvalid (Mysql2::Error: Unknown column 'editions.document_type'...
```

brucebolt: Is it possible to access publishing-api using `GdsApi::PublishingApi`, e.g. https://github.com/alphagov/whitehall/blob/147b562eb6fd8f3acddcd9bdc1fa0186c2133782/lib/tasks/publishing_api.rake#L6

thomasleese: Since this doesn't modify any of the models in Whitehall, could we then move this Rake task into the publishing-api? That way it would be reusable and doesn't need the YAML file.

Alternatively, am I right in thinking that we only need to run this task once and then we can remove it? If that's the case then we may as well merge this is now, run it and then remove it afterwards.

AlanGabbianelli: Yes, we just need to run it once. I could have done it in the console directly, but @elliotcm and I decided that it would be a good idea to have it visible somewhere. We can definitely merge it, run it, and delete it.

PR #4757, opened 2019-04-30
Index public_timestamp for statistics announcements

hannako:
This is useful because it will let us sort on public_timestamp when looking at statistics and their corresponding announcements.

We are going to run rake rummager:index:goverment on all environments.

https://trello.com/c/TV5xhCBX/649-give-stats-announcement-doc-types-publictimestamp-so-date-sort-works-in-faceted-search


barrucadu:
> We are going to run rake rummager:index:goverment on all environments.

That's a pretty hefty task, can you just reindex the statistics announcements?

hannako:
@barrucadu new rake task added

PR #4759, opened 2019-05-02
Add primary publishing organisation to organisations

MahmudH:
Currently Organisations don't have primary publishing organisation set. This PR changes Whitehall to set a primary publishing organisation. It also adds a rake task to update links for all organisations.

### DIFF
--- a/app/presenters/publishing_api/organisation_presenter.rb
+++ b/app/presenters/publishing_api/organisation_presenter.rb
@@ -50,6 +50,7 @@ module PublishingApi
         ordered_successor_organisations: successor_organisation_links,
         ordered_high_profile_groups: high_profile_groups_links,
         ordered_roles: roles_links,
+        primary_publishing_organisation: [content_id]
       }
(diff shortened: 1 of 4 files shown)


MahmudH:
> Why are we doing this? It seems like unnecessary duplication without a bit more context.
> 
> It's the sort of thing (because orgs are so ubiquitous in the system), that in future people will stumble across this and wonder why orgs have a link back to themselves, so having the initial requirement in the commit message might help.

Currently, users of the content data app can't filter organisation pages by primary publishing organisation. Adding primary publishing organisation allows us to do this. [This commit message hints at this.](https://github.com/alphagov/whitehall/pull/4759/commits/1fef4b81b6f39f2b92e06fa67e49650700a254d0)

MahmudH:
@sihugh we would like to prevent adding custom logic to apps for specific document types. The primary organisation being in links is standard for most document types and we'd like to maintain this consistency.

PR #4782, opened 2019-05-14
Present primary pub org for html attachments

peteglondon:
# What

This commit sends primary_publishing_organisation to the publishing API.

The value is the first of the lead_organisations (the same logic other document types use).

# Why

html_attachment only sends `organisations` to the publishing API. We need `primary_publishing_organisation` to be available in the `content-data-api`

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -139,6 +139,24 @@ namespace :publishing_api do
     end
   end
 
+  desc "Patch links for html publications to include primary_publishing_organisation"
+  task patch_html_publication_links: :environment do
+    scope = Publicationesque.published
+    count = scope.count
+    attachment_count = 0
+    puts "# Sending HTML attachments for #{count} publications to Publishing API"
+    scope.each_with_index do |pub, i|
+      pub.html_attachments.each do |attachment|
+        Whitehall::PublishingApi.patch_links(attachment, bulk_publishing: true)
+        attachment_count += 1
+      rescue StandardError => e
+        puts e.inspect
+      end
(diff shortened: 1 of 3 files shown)


peteglondon:
[Add primary org to HTML Publications](https://trello.com/c/Tdmt6Gex/1389-add-primary-org-to-html-publications)

test/unit/presenters/publishing_api/html_attachment_presenter_test.rb
```diff
@@ -93,4 +93,15 @@ def present(record)
 
     assert_equal %w[abcdef], present(html_attachment).links[:organisations]
   end
+
+  test "HtmlAttachment presents primary_publishing_organisation" do
+    create(:publication, :with_html_attachment, :published)
+
+    html_attachment = HtmlAttachment.last
```
cbaines: Would this work?

```ruby
html_attachment = create(:publication, :with_html_attachment, :published)
```

It would be more concise.

peteglondon: Wouldn't that return the publication, not the attachment?

cbaines: Ah, yeah, good point.

lib/tasks/publishing_api.rake
```diff
@@ -139,6 +139,24 @@ namespace :publishing_api do
     end
   end
 
+  desc "Send HTML Attachments to publishing API"
```
cbaines: To be extra clear, maybe mention something about links, or patching links here?

cbaines:
Could you have a look at the commit messages? I think in the first one, it would be useful to at least mention the use case where looking at, using this data in Content Data. As for the second one, the rake task doesn't republish the HTML attachments, so that probably needs rephrasing.

PR #4787, opened 2019-05-15
Remove "all" from search parameters sent to finder-frontend

barrucadu:
finder-frontend doesn't support "all", so make sure it's not present.

~This PR doesn't fix all of the examples in [Fix issue with param `=all` in finder redirects](https://trello.com/c/M1yHmq8q/713-fix-issue-with-param-all-in-finder-redirects) because there's also a problem with the `level_one_taxon` parameter.~ We've decided that the `level_one_taxon` thing isn't an issue because it's only a problem for links to Whitehall finders which pre-date replacing policy areas with taxons.


koetsier:
Looks good! Can you add a unit test for this?

barrucadu:
I'll fix the typo (`stripes` => `strips`) after I've seen the tests have no other problems

PR #4792, opened 2019-05-16
Fix saving of HTML attachments

peteglondon:
Previously we were attempting to get the primary organisation
from the first of the parent's `lead_organisations`

The broke saving of attachments on some document types that
don't implement `lead_organisations`.

This commit makes the code more defensive, getting the primary
org from `organisations` if `lead_organisations` is not implemented.


peteglondon:
[(3) Add primary org to HTML Publications](https://trello.com/c/Tdmt6Gex/1389-3-add-primary-org-to-html-publications)

PR #4800, opened 2019-05-22
Link to content data from admin dashboard

peteglondon:
Adds a link to the `content-data` app in each environment.

Includes basic tracking.

Possibly there will be further PRs to improve the tracking.

# Before 
Screenshot 2019-05-22 at 15 54 34

# After
Screenshot 2019-05-22 at 15 53 56

Trello: https://trello.com/c/QQzBWx3s/1442-2-add-main-link-to-content-data-from-whitehall

PR #4801, opened 2019-05-23
Use Plek.current.external_url_for to get content-data

peteglondon:
Previously I was using the wrong Plek call to get the URL
for content-data, in [this PR](https://github.com/alphagov/whitehall/pull/4800)

This commit corrects that.


peteglondon:
[(2) Add main link to Content Data from Whitehall](https://trello.com/c/QQzBWx3s/1442-2-add-main-link-to-content-data-from-whitehall)

PR #4820, opened 2019-05-30
Link to content metrics

peteglondon:
# What

Add a link to `content-data` `/metrics` endpoint 
from the edition admin `show` page when the edition has been published.

Also adds tracking to the link.

There may be another PR to add taxonomy level 1 to the tracking dimensions.

# Why

We want to get more people to use `content-data`

Trello: https://trello.com/c/NagoCEK0/1439-3-add-button-to-content-data-from-whitehall

### DIFF
--- a/app/helpers/admin/content_data_routes_helper.rb
+++ b/app/helpers/admin/content_data_routes_helper.rb
@@ -1,5 +1,16 @@
 module Admin::ContentDataRoutesHelper
+  def content_data_home_url
+    @content_data_home_url ||= "#{content_data_base_url}/content"
+  end
+
+  def content_data_page_data_url(edition)
+    path = public_document_path(edition)
+    "#{content_data_base_url}/metrics#{path}"
+  end
+
+private
+
   def content_data_base_url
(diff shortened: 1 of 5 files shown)


app/helpers/admin/content_data_routes_helper.rb
```diff
@@ -1,5 +1,16 @@
 module Admin::ContentDataRoutesHelper
+  def content_data_home_url
+    @content_data_home_url ||= "#{content_data_base_url}/content"
+  end
+
+  def content_data_page_data_url(edition)
+    path = public_document_path(edition)
+    @content_data_page_data_url ||= "#{content_data_base_url}/metrics#{path}"
```
cbaines: I'm not sure what value the memoization adds here. `content_data_base_url` is memoized, and `path` is being computed every time, so the only computation being saved is building the combined string.

Currently, it looks to me like this function will only work for one edition, and after that, it'll return the same value regardless of the argument you provide.

peteglondon:
[(3) Add button to Content Data from Whitehall](https://trello.com/c/NagoCEK0/1439-3-add-button-to-content-data-from-whitehall)

PR #4831, opened 2019-06-05
Add a footer to emails sent from integration or staging

thomasleese:
This adds a clear note to emails stating if they have come from integration or staging, which should avoid anyone getting confused if they receive these emails.

From the card:

> We can't really disable emails from integration because there are many legitimate uses of integration - including content designers testing scheduled publications.

> We could stop the data sync from creating the jobs which schedule content but then 2nd line would see an alert about missing jobs, likely manually create them, and we'd have the same problem.

[Trello Card](https://trello.com/c/M9H2hnHe/1040-whitehall-sending-integration-emails-to-publishers-when-scheduling-publishing)

### DIFF
--- a/app/mailers/notifications.rb
+++ b/app/mailers/notifications.rb
@@ -88,6 +88,8 @@ class Notifications < ActionMailer::Base
       subject: "Consultation deadline breached"
   end
 
+  helper_method :production?
+
   def production?
     GovukAdminTemplate.environment_style == "production"
   end
--- a/app/mailers/notifications.rb
+++ b/app/mailers/notifications.rb
@@ -88,15 +88,24 @@ class Notifications < ActionMailer::Base
       subject: "Consultation deadline breached"
   end
 
+  def production?
+    GovukAdminTemplate.environment_style == "production"
+  end
+
 private
 
   def no_reply_email_address
     name = "GOV.UK publishing"
-    if !/production/i.match?(GovukAdminTemplate.environment_label)
+    unless production?
       name.prepend("[GOV.UK #{GovukAdminTemplate.environment_label}] ")
     end
 
-    address = Mail::Address.new("inside-government@digital.cabinet-office.gov.uk")
+    email_address = "inside-government@digital.cabinet-office.gov.uk"
+    unless production?
+      email_address = "inside-government+#{GovukAdminTemplate.environment_style}@digital.cabinet-office.gov.uk"
+    end
+
+    address = Mail::Address.new(email_address)
     address.display_name = name
     address.format
   end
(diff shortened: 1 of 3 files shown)


app/views/layouts/notifications.text.erb
```diff
@@ -0,0 +1,5 @@
+<%= yield %>
+
+<% if !/production/i.match?(GovukAdminTemplate.environment_label) -%>
```
elliotcm: "calls to action" would be more correct

app/mailers/notifications.rb
```diff
@@ -90,13 +90,22 @@ def consultation_deadline_passed(consultation)
 
 private
 
+  def is_production?
+    /production/i.match?(GovukAdminTemplate.environment_label)
```
elliotcm: If `environment_style` is the more "computery" version of this text, should we not do a simpler string comparison against that instead of titleizing it and then doing a case insensitive regex match against it?

app/views/layouts/notifications.text.erb
elliotcm: Do we also have access to `is_production?` here?

thomasleese: That's a good point, I didn't think we did, but maybe we do...

app/mailers/notifications.rb
elliotcm: Also wondering whether `is_production?` provides any more useful a name than `production?` would.

app/views/layouts/notifications.text.erb
elliotcm: I don't remember the exact behaviour of actionmailer but it's heavily modelled on controllers/views so I suspect if it's not already available as a helper method we might be able to make it one.

thomasleese:
Thanks @elliotcm, this should be ready for a re-review.

app/mailers/notifications.rb
```diff
@@ -88,15 +88,26 @@ def consultation_deadline_passed(consultation)
       subject: "Consultation deadline breached"
   end
 
+  helper_method :production?
+
+  def production?
+    /production/.match?(GovukAdminTemplate.environment_style)
```
elliotcm: Can this just be
```suggestion
    GovukAdminTemplate.environment_style == "production"
```
now?

thomasleese: Yeah, I thought about that. I wasn’t sure if there was a good reason for it being a regex other than the case insensitive comparison with a human readable string so decided to leave it, but perhaps it’s worth giving it a go.

thomasleese: I've amended the PR with this change now and all the tests still seem to be passing!

PR #4832, opened 2019-06-05
Remove org page part of organisations/index action

sihugh:
The [organisations index page](https://www.gov.uk/government/organisations) is now provided by the [collections app](https://github.com/alphagov/collections/blob/4cdfd64e29923fbcdcd0e3e93f4d4a0e93021b2f/app/controllers/organisations_controller.rb#L6) (and has been for quite a while).

The only bit of it that is still relevant here is for [the courts and tribunals](https://www.gov.uk/courts-tribunals) index page, so we can remove the other bits.

Unpicking the organisation/show pages is trickier, and is more easily done after we've properly migrated the courts and tribunals pages over.


sihugh:
Thanks both 👍

PR #4838, opened 2019-06-07
Add a Rake task for manually unpublishing content

cbaines:
Whitehall deletes some content when it unpublishes it, this makes it
very difficult to change the details of the unpublished
content (e.g. change from gone to redirect, or change the redirect).

This Rake task provides a workaround intended for these cases where
you want to manually send an unpublish request to the Publishing API.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -128,6 +128,46 @@ namespace :publishing_api do
     PublishingApiDocumentRepublishingWorker.new.perform(document.id)
   end
 
+  def ensure_document_with_content_id_doesnt_exist(content_id)
+    document = Document.find_by(content_id: content_id)
+
+    return unless document
+
+    puts "Document with this content ID exists: #{document}"
+    abort 'This Rake task is only for unknown content'
+  end
+
+  desc "Manually unpublish content with a redirect"
+  # This task is for unpublishing Whitehall managed content where
+  # Whitehall has forgotten it is managing the content (as it often
+  # does). Do not use this Rake task for content which Whitehall still
+  # has a record of.
+  namespace :unpublish_with_redirect do
+    task :dry_run, %i[content_id alternative_path locale] => :environment do |_, args|
+      args.with_defaults(locale: 'en')
+
+      ensure_document_with_content_id_doesnt_exist(args[:content_id])
+
+      puts "Would send an unpublish request to the Publishing API for #{args[:content_id]} with:"
+      puts "  type 'redirect', locale: #{args[:locale]} and alternative_path #{args[:alternative_path].strip}"
+    end
+
+    task :real, %i[content_id alternative_path locale] => :environment do |_, args|
+      args.with_defaults(locale: 'en')
+
+      ensure_document_with_content_id_doesnt_exist(args[:content_id])
+
+      response = Services.publishing_api.unpublish(
+        args[:content_id],
+        type: 'redirect',
+        locale: args[:locale],
+        alternative_path: args[:alternative_path].strip
+      )
+
+      puts response
+    end
+  end
+
   desc "Redirect HTML Attachments to a given URL"
   namespace :redirect_html_attachments do
     task :dry, %i[content_id destination] => :environment do |_, args|


lib/tasks/publishing_api.rake
```diff
@@ -128,6 +128,29 @@ namespace :publishing_api do
...
+  desc "Manually unpublish content with a redirect"
+  # This task is for unpublishing Whitehall managed content, where
+  # Whitehall has forgotten it is managing the content (as it often
+  # does).
+  task :unpublish_with_redirect, %i[content_id alternative_path locale] => :environment do |_, args|
+    args.with_defaults(locale: 'en')
+
+    document = Document.find_by(content_id: args[:content_id])
+    if document
+      puts "Document with this contnet ID exists: #{document}"
```
ChrisBAshton:
```suggestion
      puts "Document with this content ID exists: #{document}"
```

lib/tasks/publishing_api.rake
```diff
@@ -128,6 +128,29 @@ namespace :publishing_api do
     PublishingApiDocumentRepublishingWorker.new.perform(document.id)
   end
 
+  desc "Manually unpublish content with a redirect"
+  # This task is for unpublishing Whitehall managed content, where
+  # Whitehall has forgotten it is managing the content (as it often
+  # does).
```
ChrisBAshton: Could you rewrite the comment to say when you expect to use this? eg

```ruby
# This task is for unpublishing and redirecting Whitehall content. Normally,
# unpublishing deletes the content, making it difficult to set up a redirect.
# Do not use this task if the content has already been deleted - you'll need to
# do something else for that (TBC).
```

lib/tasks/publishing_api.rake
```diff
@@ -128,6 +128,29 @@ namespace :publishing_api do
...
+      puts "Document with this contnet ID exists: #{document}"
+      abort 'This Rake task is only for unknown content'
+    end
+
+    response = Services.publishing_api.unpublish(
+      args[:content_id],
+      type: 'redirect',
+      locale: args[:locale],
+      alternative_path: args[:alternative_path].strip
+    )
```
ChrisBAshton: Can we have a `:dry` version too?

```ruby
task :dry, %i[content_id alternative_path locale] => :environment do |_, args|
  # ...
  puts "This will unpublish #{args[:content_id]} and redirect it to #{args[:alternative_path]}" # or whatever
end
```

lib/tasks/publishing_api.rake
cbaines: To me, that is when Whitehall has forgotten it's managing the content in question.

Luckily, this isn't a normal situation, for most content managed by Whitehall, it isn't deleted from the database when you unpublish it.

I've changed this now though, is this any better?

```
  # This task is for unpublishing Whitehall managed content where
  # Whitehall has forgotten it is managing the content (as it often
  # does). Do not use this Rake task for content which Whitehall still
  # has a record of.
```

lib/tasks/publishing_api.rake
cbaines: Sure, I've added a `:dry_run` variant now, as well as the `:real` one.

lib/tasks/publishing_api.rake
cbaines: Thanks, I've fixed this now.

lib/tasks/publishing_api.rake
```diff
@@ -128,6 +128,45 @@ namespace :publishing_api do
...
+    end
+
+    task :real, %i[content_id alternative_path locale] => :environment do |_, args|
+      args.with_defaults(locale: 'en')
+
+      document = Document.find_by(content_id: args[:content_id])
+      if document
+        puts "Document with this content ID exists: #{document}"
+        abort 'This Rake task is only for unknown content'
+      end
```
ChrisBAshton: Could you reduce the duplication between the `:real` and `:dry_run` versions by extracting this out?

```ruby
def ensure_not_exists(content_id)
      document = Document.find_by(content_id: args[:content_id])
      if document
        puts "Document with this content ID exists: #{document}"
        abort 'This Rake task is only for unknown content'
      end
end

task :real, %i[content_id alternative_path locale] => :environment do |_, args|
    args.with_defaults(locale: 'en')
    ensure_not_exists(args[:content_id])
    # etc
end
```

cbaines: Yeah, I'll put that in.

lib/tasks/publishing_api.rake
```diff
@@ -128,6 +128,46 @@ namespace :publishing_api do
     PublishingApiDocumentRepublishingWorker.new.perform(document.id)
   end
 
+  def ensure_document_with_content_id_doesnt_exist(content_id)
+    document = Document.find_by(content_id: content_id)
+
+    return unless document
+
+    puts "Document with this content ID exists: #{document}"
+    abort 'This Rake task is only for unknown content'
```
ChrisBAshton: Don't know if rake lint picks this up, but this seems a funny mix of single and double quotes?

ChrisBAshton:
(PS: do we document this rake task anywhere once it's merged?)

lib/tasks/publishing_api.rake
cbaines: You can't use double quotes for the first string, as that's what enables using `#{ }`. You also don't need to use double quotes for the second string, as it's just a simple string.

I'm not sure there are any linting rules about using single vs double quotes...

cbaines:
> (PS: do we document this rake task anywhere once it's merged?)

Hmm, it might be worth trying to start documenting how unpublishing/redirecting different types of Whitehall content is managed, but I think that would be a new bit of documentation. I'm not sure if there's something existing where this would fit in and be a useful addition.

ChrisBAshton:
Presumably here would be a good place? https://docs.publishing.service.gov.uk/manual/howto-redirect-html-attachment-urls-from-whitehall.html

PR #4855, opened 2019-06-18
Adds a redirect for announcements atom feed

SamJamCul:
This change introduces a redirect in the announcements finder
controller. When going to an announcements finder atom feed,
this checks to see if the locale selected is English, and if so it will
redirect to the news and comms finder atom feed instead.

Tests have also been removed/added to account for this, but there might be some missing feature tests that should be included. 

This is part of [this card](https://trello.com/c/E5QJ99L5/805-redirect-english-announcements-atom-feeds), and thanks to @koetsier for helping!

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -26,6 +26,8 @@ class AnnouncementsController < DocumentsController
         )
       end
       format.atom do
+        return redirect_to_news_and_communications(".atom") if is_english_locale?
+
         @announcements = @filter.documents
       end
     end
@@ -43,8 +45,8 @@ private
     params[:locale].nil?
   end
 
-  def redirect_to_news_and_communications
-    base_path = "#{Plek.new.website_root}/search/news-and-communications"
+  def redirect_to_news_and_communications(format = "")
+    base_path = "#{Plek.new.website_root}/search/news-and-communications#{format}"
     redirect_to("#{base_path}?#{news_and_communications_query_string}")
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/announcements_controller.rb
```diff
@@ -43,8 +45,8 @@ def is_english_locale?
     params[:locale].nil?
   end
 
-  def redirect_to_news_and_communications
-    base_path = "#{Plek.new.website_root}/search/news-and-communications"
+  def redirect_to_news_and_communications(atom = "")
```
koetsier: maybe 'atom' isn't such a great name for this, because it would work if you called it with 'json' for instance.

test/functional/announcements_controller_test.rb
```diff
@@ -231,55 +260,6 @@ def refute_select(hash)
...
-    with_stubbed_rummager(@rummager, true) do
-      create(:published_news_story, first_published_at: 1.week.ago)
-
-      @rummager.expects(:search).returns(search_results_from_rummager)
-
-      processed_rummager_documents = search_results_from_rummager['results'].map do |result|
-        RummagerDocumentPresenter.new(result)
-      end
-
-      get :index, params: { announcement_type_option: 'news-stories' }, format: :atom
```
bilbof: What happens if you set the `locale: :fr` here?

SamJamCul: It doesn't seem to like it, I've passed in the `locale: :fr` as one of the params and the test seems to fail

app/controllers/announcements_controller.rb
SamJamCul: Changing it to `format` so it's a more general parameter name

test/functional/announcements_controller_test.rb
SamJamCul: I think this is because the localised finders don't really have any parameters to set, so they can't inherit any which is what this test is looking at.

PR #4870, opened 2019-06-24
Permit malformed array params when redirecting

bilbof:
When redirecting old finders publications/announcements
to the new finders served by finder-frontend, we don't
support malformed arrays in the query string. This fixes
that by fixing the malformed arrays before we redirect.

Example

?departments%5B0%5D=department-for-education
is parsed by rails as
{departments=>{0=>department-for-education}}
but really it should be
{departments=>[department-for-education]}

This uses the existing Whitehall::DocumentFilter::CleanedParams
class (which contains the method clean_malformed_array_params)
to fix this in a consistent manner.

Trello: https://trello.com/c/eTCe4Vun/825.

### DIFF
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -57,18 +57,19 @@ private
   end
 
   def publications_query_string
-    level_one_taxon = params['taxons'].try(:first) || params['topics'].try(:first)
-    level_two_taxon = params['subtaxons'].try(:first)
+    allowed_params = cleaned_document_filter_params
+    level_one_taxon = allowed_params['taxons'].try(:first) || allowed_params['topics'].try(:first)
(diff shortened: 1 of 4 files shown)


app/controllers/publications_controller.rb
```diff
@@ -57,18 +57,19 @@ def special_params
   end
 
   def publications_query_string
-    level_one_taxon = params['taxons'].try(:first) || params['topics'].try(:first)
-    level_two_taxon = params['subtaxons'].try(:first)
+    allowed_params = cleaned_document_filter_params
```
barrucadu: Can this be called `filters_params` - or `filters_params` called `allowed_params`?

bilbof: I shall make it so.

PR #4882, opened 2019-07-01
Make Rails initialiser idempotent

tijmenb:
`mkdir_p` will crash if the directory already exists, which happens in Docker. This adds a check first.

https://trello.com/c/jRbKkgHd

### DIFF
--- a/config/initializers/bulk_upload_zip_file.rb
+++ b/config/initializers/bulk_upload_zip_file.rb
@@ -4,4 +4,8 @@ BULK_UPLOAD_ZIPFILE_DEFAULT_ROOT_DIRECTORY = if Rails.env.test?
                                                Rails.root.join('bulk-upload-zip-file-tmp')
                                              end
 
-FileUtils.mkdir_p(BULK_UPLOAD_ZIPFILE_DEFAULT_ROOT_DIRECTORY)
+begin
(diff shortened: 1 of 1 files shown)


config/initializers/bulk_upload_zip_file.rb
```diff
@@ -4,4 +4,8 @@
                                                Rails.root.join('bulk-upload-zip-file-tmp')
                                              end
 
-FileUtils.mkdir_p(BULK_UPLOAD_ZIPFILE_DEFAULT_ROOT_DIRECTORY)
+begin
+  FileUtils.mkdir_p(BULK_UPLOAD_ZIPFILE_DEFAULT_ROOT_DIRECTORY)
```
thomasleese: I think the issue is actually that when you run the dev VM, [these directories end up being created as symlinks](https://github.com/alphagov/govuk-puppet/blob/8bcdde471ed62a61ba48fb72af420323fdf5d19b/modules/govuk/manifests/apps/whitehall.pp#L392-L396) which on your local machine ends up linking to a place that doesn't exist.

`mkdir_p` should already ignore existing directories I believe: https://apidock.com/ruby/FileUtils/mkdir_p

Ideally, we would delete the directory if it already exists as a symlink and then create it...

PR #4893, opened 2019-07-10
Reslug preventing-sexual-violence-initiative

koetsier:
to reflect the name change to Preventing Sexual Violence in Conflict Initiative

see Zendesk: https://govuk.zendesk.com/agent/tickets/3740145


MuriloDalRi:
I recently found out that this doesn't redirect the atom feed, it just gets unpublished, so if you care about that this is how I fixed it: https://github.com/alphagov/whitehall/blob/master/db/data_migration/20190709094308_redirect_department_health_atom.rb

koetsier:
Awesome thanks! I'll write a new PR after merging this one

PR #4901, opened 2019-07-16
Add endpoint to export documents information

1pretz1:
This endpoint takes two arguments, `lead_organisation` and
`type` as well as pagination options like `items_per_page` and `page_number`.

`lead_organisation` is needed to find all the documents where
the `lead_organisation` is present on the latest edition of each
document. `type` is used to specify the whitehall edition type of the documents
you want returning, e.g NewsArticle or Publication.

The endpoint returns the following information for every document:
  - the document_id
  - all the locales the document has had
  - all the subtypes the document has had (news_article_type, speech_type, publication_type, corporate_information_page_type)
  - all the lead_organisations of the most recent edition

Trello:
https://trello.com/c/Tmq0PbYg/1037-create-an-endpoint-for-getting-a-list-of-documents-from-a-publisher

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -5,4 +5,38 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
     @document = Document.find(params[:id])
     respond_with DocumentExportPresenter.new(@document)
   end
+
+  def index
+    result_set = Whitehall::Exporters::DocumentsInfoExporter.new(paginated_document_ids).call
+
+    respond_with(
+      documents: result_set,
+      page_number: page_number,
+      page_count: result_set.count
+    )
+  end
+
+  private
+
+  def paginated_document_ids
+    Edition
+      .joins("INNER JOIN edition_organisations eo ON eo.edition_id = editions.id")
+      .joins("INNER JOIN organisations o ON o.id = eo.organisation_id")
+      .where(o: { content_id: params.require(:lead_organisation) })
+      .where(eo: { lead: true })
+      .where(type: params.require(:type))
+      .latest_edition
+      .order(:document_id)
+      .page(page_number)
+      .per(items_per_page)
+      .pluck(:document_id)
+  end
+
+  def page_number
+    params.fetch(:page_number, 1)
+  end
+
+  def items_per_page
+    params.fetch(:page_count, 100)
+  end
 end
--- a/test/unit/whitehall/exporters/documents_info_exporter_test.rb
+++ b/test/unit/whitehall/exporters/documents_info_exporter_test.rb
@@ -1,0 +1,85 @@
+require 'test_helper'
+
+module Whitehall::Exporters
+  class DocumentsInfoExporterTest < ActiveSupport::TestCase
+    test 'returns all subtype information of a document' do
+      org = create(:organisation)
+      document = create(:document)
+      create(
+        :news_article,
+        :superseded,
+        organisations: [org],
+        document: document,
+        news_article_type: NewsArticleType::NewsStory,
+      )
+      create(
+        :news_article,
+        :published,
+        organisations: [org],
+        document: document,
+        news_article_type: NewsArticleType::PressRelease,
+      )
+
+      documents_info_exporter = DocumentsInfoExporter.new(
+        [document.id]
+      )
+
+      assert_equal(
+        %w[news_story press_release],
+        documents_info_exporter.call.first[:document_information][:subtypes]
+      )
+    end
+
+
+    test 'returns all the locales of a document' do
+      org = create(:organisation)
+      news_story = create(
+        :news_article,
+        :published,
+        :translated, primary_locale: 'en', translated_into: %w(ms ar cy),
+        organisations: [org],
+        news_article_type: NewsArticleType::NewsStory,
+      )
+
+      documents_info_exporter = DocumentsInfoExporter.new(
+        [news_story.document_id]
+      )
+
+      assert_same_elements(
+        %w[ar cy en ms],
+        documents_info_exporter.call.first[:document_information][:locales]
+      )
+    end
+
+    test 'returns the lead organisations of the latest edition of a document' do
+      published_edition_org = create(:organisation)
+      published_edition_org_2 = create(:organisation)
+      superseded_edition_org = create(:organisation)
+      document = create(:document)
+      create(
+        :news_article,
+        :superseded,
+        document: document,
+        organisations: [superseded_edition_org],
+        news_article_type: NewsArticleType::NewsStory,
+      )
+
+      create(
+        :news_article,
+        :published,
+        document: document,
+        organisations: [published_edition_org, published_edition_org_2],
+        news_article_type: NewsArticleType::NewsStory,
+      )
+
+      documents_info_exporter = DocumentsInfoExporter.new(
+        [document.id]
+      )
+
+      assert_same_elements(
+        [published_edition_org.content_id, published_edition_org_2.content_id],
+        documents_info_exporter.call.first[:document_information][:lead_organisations]
+      )
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -184,7 +184,7 @@ Whitehall::Application.routes.draw do
         root to: 'dashboard#index', via: :get
 
         namespace 'export' do
-          resources :document, only: [:show], defaults: { format: :json }
+          resources :document, only: %i[show index], defaults: { format: :json }
         end
 
         get 'find-in-admin-bookmarklet' => 'find_in_admin_bookmarklet#index', as: :find_in_admin_bookmarklet_instructions_index
--- a/lib/whitehall/exporters/documents_info_exporter.rb
+++ b/lib/whitehall/exporters/documents_info_exporter.rb
@@ -1,0 +1,64 @@
+class Whitehall::Exporters::DocumentsInfoExporter
+  attr_reader :document_ids
+
+  def initialize(document_ids)
+    @document_ids = document_ids
+  end
+
+  def call
+    document_ids.map do |doc_id|
+      {
+        document_id: doc_id,
+        document_information: {
+          locales: locales_hash[doc_id],
+          subtypes: subtypes_hash[doc_id],
+          lead_organisations: lead_orgs_for_latest_edition_hash[doc_id]
+        }
+      }
+    end
+  end
+
+  private
+
+  def locales_hash
+    @locales_hash ||= Edition
+      .joins(:translations)
+      .where(document_id: document_ids)
+      .group(:document_id)
+      .pluck(:document_id, "GROUP_CONCAT(DISTINCT(edition_translations.locale))")
+      .each_with_object({}) { |(k, v), memo| memo[k] = v.split(",") }
+  end
+
+  def subtypes_hash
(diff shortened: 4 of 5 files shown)


app/presenters/documents_info_export_presenter.rb
```diff
@@ -0,0 +1,86 @@
...
+    lead_orgs_hash[doc_id].split(',')
+  end
+
+  def locales_hash
+    @locales_hash ||= Edition
+      .joins(:translations)
+      .where(document_id: document_ids)
+      .group(:document_id)
+      .pluck(:document_id, "GROUP_CONCAT(DISTINCT(edition_translations.locale))")
+      .to_h
```
kevindew: rather than doing `.to_h` it might be better to do `.each_with_object({}) { |(k, v), memo| memo[k] = v.split(",") }` as that will mean you don't have to worry about the array split later.

app/presenters/documents_info_export_presenter.rb
```diff
@@ -0,0 +1,86 @@
+class DocumentsInfoExportPresenter
+  attr_reader :lead_organisation, :type
+
+  def initialize(lead_organisation:, type:)
+    @lead_organisation = lead_organisation
+    @type = type
+  end
+
+  def call
+    document_ids.sort.map do |id|
```
kevindew: I'd suggest this sort would be more intuitive as part of the data query (and hopefully can be part of the pagination)

app/presenters/documents_info_export_presenter.rb
```diff
@@ -0,0 +1,86 @@
...
+      .group(:document_id)
+      .pluck(:document_id,
+             "GROUP_CONCAT(DISTINCT(editions.news_article_type_id))",
+             "GROUP_CONCAT(DISTINCT(editions.speech_type_id))",
+             "GROUP_CONCAT(DISTINCT(editions.publication_type_id))",
+             "GROUP_CONCAT(DISTINCT(editions.corporate_information_page_type_id))")
+      .map(&:compact).to_h
+  end
+
+  def lead_orgs_hash
```
kevindew: For lead orgs I think you might want to update the method name to reflect it's only for latest edition or add a comment. It's a nuance that I don't think comes across easily in the code at the moment.

app/presenters/documents_info_export_presenter.rb
```diff
@@ -0,0 +1,86 @@
...
+  def locales(doc_id)
+    locales_hash[doc_id].split(',')
+  end
+
+  def subtypes(doc_id)
+    if type_has_subtype?
+      subtypes_hash[doc_id].split(',').map do |subtype_id|
+        (type + 'Type').constantize.find_by_id(subtype_id.to_i).singular_name
+      end
+    end
```
kevindew: This seems to be the bit where we have the greatest intersection between the type input. I'm wondering if we can avoid needing it by just treating everything as having any one of the types and expect that the query sorts this out for us - since we already query in that way.

So something along the lines of:

```
Edition
      .where(document_id: document_ids)
      .group(:document_id)
      .pluck(:document_id,
             "GROUP_CONCAT(DISTINCT(editions.news_article_type_id))",
             "GROUP_CONCAT(DISTINCT(editions.speech_type_id))",
             "GROUP_CONCAT(DISTINCT(editions.publication_type_id))",
             "GROUP_CONCAT(DISTINCT(editions.corporate_information_page_type_id))")
      .each_with_object do |(document_id, news, speeches, publications, corporate), memo|
         news_article_types = news.split(",").map { |id| NewsArticleType.find_by_id(id)&.key }
         speech_types = speech.split(",").map { |id| SpeechType.find_by_id(id)&.key }
         ...
         memo[document_id] = [news_article_types, speeches_types, publications_types, corporate_types].flatten
      end
```

Though I appreciate my suggestion here is pretty heavy in a method chain so could be broken up a bit.

Also couple of things I've suggested here: use key rather than singular name - seems to be consistent with https://github.com/alphagov/whitehall/pull/4963, and we should consider that the type class doesn't exist (example below)

```
[21] pry(main)> Publication.where.not(publication_type_id: PublicationType.all.map(&:id)).count
Creating scope :open. Overwriting existing method Consultation.open.
   (57.1ms)  SELECT COUNT(*) FROM `editions` WHERE `editions`.`type` IN ('Publication') AND (`editions`.`state` != 'deleted') AND (`editions`.`publication_type_id` NOT IN (1, 2, 3, 19, 4, 5, 15, 6, 7, 17, 10, 12, 14, 18, 20, 21, 8, 13, 22, 999, 1000))
=> 6
```

app/presenters/documents_info_export_presenter.rb
```diff
@@ -0,0 +1,86 @@
+class DocumentsInfoExportPresenter
```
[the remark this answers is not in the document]
kevindew: Reflecting on this comment, my suggestion that this acts like a model is incorrect as I realise it's actually an array of document_ids rather than a single one. So rather than acting like a model that's looked up it's more a class that is used to perform a transformation.

app/controllers/admin/export/document_controller.rb
```diff
@@ -5,4 +5,32 @@ def show
...
+
+  def index
+    documents_information = DocumentsInfoExportPresenter.new(
+      lead_organisation: params[:lead_organisation],
+      type: params[:type],
+    ).call
+
+    page_result = Kaminari.paginate_array(documents_information)
+                          .page(page_number)
+                          .per(items_per_page)
```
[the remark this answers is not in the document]
1pretz1: Done!

app/controllers/admin/export/document_controller.rb
```diff
@@ -5,4 +5,32 @@ def show
     @document = Document.find(params[:id])
     respond_with DocumentExportPresenter.new(@document)
   end
+
+  def index
+    documents_information = DocumentsInfoExportPresenter.new(
+      lead_organisation: params[:lead_organisation],
+      type: params[:type],
```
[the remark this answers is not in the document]
1pretz1: Yup and done!

app/controllers/admin/export/document_controller.rb
```diff
@@ -5,4 +5,32 @@ def show
...
+      page_number: page_number,
+      page_count: page_result.count,
+      total: documents_information.count
+    )
+  end
+
+  private
+
+  def page_number
+    params[:page_number] || 1
```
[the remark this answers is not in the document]
1pretz1: Sure

app/presenters/documents_info_export_presenter.rb
```diff
@@ -0,0 +1,86 @@
...
+          subtypes: subtypes(id),
+          lead_organisations: lead_orgs(id)
+        } }
+    end
+  end
+
+  private
+
+  def document_ids
+    @document_ids ||= type.constantize
```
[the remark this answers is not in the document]
1pretz1: Yup good point, I hadn't thought of the security implications. I've re-written the code to use `Edition` instead of the `Edition` type, it does mean however there is some sql in the joins as edition doesn't define a join with tables like organisations.

app/presenters/documents_info_export_presenter.rb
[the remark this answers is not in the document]
1pretz1: As per our conversation f2f, we decided the best place was to put this is in the whitehall/exporters directory.

app/presenters/documents_info_export_presenter.rb
1pretz1: Yup sure

app/presenters/documents_info_export_presenter.rb
1pretz1: Have made the name more descriptive.

app/presenters/documents_info_export_presenter.rb
1pretz1: Done, have split the method up as suggested.

app/presenters/documents_info_export_presenter.rb
1pretz1: Yup done.

1pretz1:
@kevindew I've made the changes you've requested, hopefully all the tests pass as the VPN doesn't seem to work on the train which means I can't see CI and my whitehall tests don't run properly locally! 😆

app/controllers/admin/export/document_controller.rb
```diff
@@ -1,8 +1,44 @@
+require "#{Rails.root}/lib/whitehall/exporters/documents_info_exporter"
```
kevindew: Do we need to require this? it seems to be in the eager load path: https://github.com/alphagov/whitehall/blob/26e2e9eecd862af329cd85184e93cfe9c1932789/config/application.rb#L18-L20

If we did need to require it we'd have lib in the include path so just `require "whitehall/exporters/documents_info_exporter"` would work.

app/controllers/admin/export/document_controller.rb
```diff
@@ -1,8 +1,44 @@
...
+      documents: result_set,
+      page_number: page_number,
+      page_count: result_set.count
+    )
+  end
+
+  private
+
+  def paginated_document_ids
+    org_id = Organisation.find_by(content_id: params.require(:lead_organisation)).id
```
kevindew: Is this the only usage of `org_id`? if so we can probably just have this as part of the join.

test/unit/whitehall/exporters/documents_info_exporter_test.rb
```diff
@@ -0,0 +1,155 @@
+require 'test_helper'
+require "#{Rails.root}/lib/whitehall/exporters/documents_info_exporter"
+
+class DocumentsInfoExporterTest < ActiveSupport::TestCase
```
kevindew: I think these tests could be a bit more focused on what we're testing.

We should have tests that cover: locales, subtypes, and organisations.

locales seems to be covered by: https://github.com/alphagov/whitehall/pull/4901/files#diff-ed16559bfa573dce66e35a458a78c64cR105
subtypes is covered by: https://github.com/alphagov/whitehall/pull/4901/files#diff-ed16559bfa573dce66e35a458a78c64cR76
and orgs is covered by: https://github.com/alphagov/whitehall/pull/4901/files#diff-ed16559bfa573dce66e35a458a78c64cR125

I'm not really sure we need the tests at https://github.com/alphagov/whitehall/pull/4901/files#diff-ed16559bfa573dce66e35a458a78c64cR5 and https://github.com/alphagov/whitehall/pull/4901/files#diff-ed16559bfa573dce66e35a458a78c64cR47 as I'm not sure what functionality that's being tested.

app/controllers/admin/export/document_controller.rb
1pretz1: We don't actually need require as it turns out, I just needed to add the correct namespaces to the class / class calls.

app/controllers/admin/export/document_controller.rb
1pretz1: Yep it's only used once and so I've added it to the main query.

test/unit/whitehall/exporters/documents_info_exporter_test.rb
1pretz1: I've removed the two tests you've highlighted, these were needed originally when the exporter class was concerned about the different document types, this logic now resides in the controller so these tests as you've pointed out are pretty redundant.

1pretz1:
@kevindew PR is ready for another review

PR #4903, opened 2019-07-17
Add locked attribute to Document

emmabeynon:
For https://trello.com/c/FAiIY6Eu/1016-add-flag-to-indicate-a-document-is-going-to-be-migrated-from-whitehall

This is in preparation for locking a document for editing before importing it
into Content Publisher.

### DIFF
--- a/db/migrate/20190711112443_add_locked_flag_to_document.rb
+++ b/db/migrate/20190711112443_add_locked_flag_to_document.rb
@@ -1,0 +1,5 @@
+class AddLockedFlagToDocument < ActiveRecord::Migration[5.1]
+  def change
+    add_column :documents, :locked, :boolean, default: false, null: false
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190426132424) do
+ActiveRecord::Schema.define(version: 20190711112443) do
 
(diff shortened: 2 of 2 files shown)


db/migrate/20190711112443_add_locked_flag_to_document.rb
```diff
@@ -0,0 +1,5 @@
+class AddLockedFlagToDocument < ActiveRecord::Migration[5.1]
+  def change
+    add_column :documents, :locked, :boolean, default: false, null: false
```
kevindew: Because this has a default it might be a rather slow migration to run. I've not worked with MySQL in a while but doing this on Postgres is slow and locking. Some googling led me to this: https://docs.gitlab.com/ee/development/migration_style_guide.html which seems to imply it can be a problem

Does it take long at all on your VM? I assume there you have basically a full set of data so is probably a reasonable approximation. There probably isn't actually that many document records so might be virtually pain free.

emmabeynon: Oops sorry, must have commented and merged at the same time 😬 I've got 278.5k documents locally and it took 7.68 seconds to run the migration.

kevindew: No worries. 

Ok that doesn't sound too bad a time, often the migrations can take longer on prod with all the db activity so you probably brace yourself that it could potentially cause downtime in the region of about 30s so you shouldn't supervise it's deploy.

Obviously now it's merged there isn't a chance to update commit with more details but could still update PR description if you agree that'd be useful.

PR #4904, opened 2019-07-17
Tags social media links as written in english

injms:
## Why?
Pages that have been translated appear to be both in the translated language _and_ in english. This causes issues with screenreaders as they try and read one language with the 'voice' of another language. This should help fix that by making sure that social media links, such as the ones on the page for the [British Embassy in Seoul](https://www.gov.uk/world/organisations/british-embassy-seoul.ko), are marked as being written in english.

## What
 - Social media links were not being translated into languages other than English, despite being on a translated page. This change tags the links as being written in English.
 - Whitehall will now tag the `main` element with the locale language, unless the language is English which is already set on the `html` element.

### DIFF
--- a/app/views/layouts/frontend.html.erb
+++ b/app/views/layouts/frontend.html.erb
@@ -10,7 +10,7 @@
       <li><a href="/news-and-communications">News and communications</a></li>
     </ul>
   </nav>
-  <main id="content" role="main" class="whitehall-content"<%= " lang=#{locale.to_s}" unless locale == :en %>>
+  <main id="content" role="main" class="whitehall-content"<%= " lang=#{I18n.locale.to_s}" unless I18n.locale == :en %>>
     <div id="page" class="<%= content_for?(:page_class) ? yield(:page_class) : '' %>" >
       <%= yield %>
       <%= render_mustache_templates if Rails.env.development? %>
--- a/app/views/layouts/frontend.html.erb
+++ b/app/views/layouts/frontend.html.erb
@@ -10,7 +10,7 @@
       <li><a href="/news-and-communications">News and communications</a></li>
     </ul>
   </nav>
-  <main id="content" role="main" class="whitehall-content">
+  <main id="content" role="main" class="whitehall-content"<%= " lang=#{locale.to_s}" unless locale == :en %>>
     <div id="page" class="<%= content_for?(:page_class) ? yield(:page_class) : '' %>" >
       <%= yield %>
       <%= render_mustache_templates if Rails.env.development? %>
--- a/app/views/shared/_social_media_accounts.html.erb
+++ b/app/views/shared/_social_media_accounts.html.erb
@@ -1,14 +1,14 @@
 <% followus ||= false %>
 <% if socialable.social_media_accounts.any? %>
-  <section>
+  <section lang="en">
     <% unless followus %>
-      <h2 class="label"><%= t('social_media.follow_us') %></h2>
+      <h2 class="label">Follow on social media</h2>
     <% end %>
     <ul class="social-media-accounts js-hide-extra-social-media">
       <% socialable.social_media_accounts.each do |account| %>
         <%= content_tag_for(:li, account) do %>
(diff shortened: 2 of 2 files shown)


app/views/shared/_social_media_accounts.html.erb
```diff
@@ -8,8 +8,8 @@
       <% socialable.social_media_accounts.each do |account| %>
         <%= content_tag_for(:li, account) do %>
           <%= link_to account.url,
-                      class: "social-media-link #{account.service_name.parameterize}" do %>
-            <span class="connect">Connect with <%= account.socialable.display_name %> on</span> <%= account.display_name %>
+                      class: "social-media-link #{account.service_name.parameterize}", lang: 'en' do %>
```
maxgds: I think we should apply the language tag to the container - the section element that opens at line 3. There would be a case for doing it this way if each one might be a different language, but since we're declaring them all to be English less repetition would be good.

app/views/layouts/frontend.html.erb
```diff
@@ -10,7 +10,7 @@
       <li><a href="/news-and-communications">News and communications</a></li>
     </ul>
   </nav>
-  <main id="content" role="main" class="whitehall-content">
+  <main id="content" role="main" class="whitehall-content"<%= " lang=#{locale.to_s}" unless locale == :en %>>
     <div id="page" class="<%= content_for?(:page_class) ? yield(:page_class) : '' %>" >
```
maxgds: I noticed this differs slightly from government-frontend which uses I18n.locale but I'm not clear on why that would be better and I'm sure you've tested this is inserting the correct codes so probably no change to make here.

app/views/shared/_social_media_accounts.html.erb
injms: Good idea. I'll need to untranslate the heading, which might be for the best as then the entire `section` is in one language.

app/views/layouts/frontend.html.erb
injms: Thanks for sharing - I've changed it to use `I18n.locale`. My thinking is that `I18n.locale` is only about the translation, whereas `locale` could be hijacked to something more general.

injms:
[Mark up share links as English](https://trello.com/c/sET78e3z/14-mark-up-share-links-as-english)

injms:
[Mark up share links as English](https://trello.com/c/sET78e3z/14-mark-up-share-links-as-english)

injms:
@vanitabarrett  As Max is on holiday, can you take a look please?

app/views/shared/_social_media_accounts.html.erb
```diff
@@ -1,15 +1,15 @@
...
-      <h2 class="label"><%= t('social_media.follow_us') %></h2>
+      <h2 class="label">Follow on social media</h2>
     <% end %>
     <ul class="social-media-accounts js-hide-extra-social-media">
       <% socialable.social_media_accounts.each do |account| %>
         <%= content_tag_for(:li, account) do %>
           <%= link_to account.url,
                       class: "social-media-link #{account.service_name.parameterize}" do %>
-            <span class="connect">Connect with <%= account.socialable.display_name %> on</span> <%= account.display_name %>
+            <span class="connect"></span><%= account.display_name %>
```
vanitabarrett: Is there a reason why we're removing the "Connect with" text?

injms: This is screen reader text that could be in two languages, so shouldn't be tagged as being in English. For example, on the [British Embassy in Seoul page](https://www.gov.uk/world/organisations/british-embassy-seoul.ko) it has "Connect with 주한영국대사관 on" because the `account.socialable.display_name` is somehow translated before it arrives at the template, but the rest of the sentence is not. This was a more consistent fix than trying to tag the display name in one language and the rest of the sentence in another.

vanitabarrett: That makes sense 👍

PR #4910, opened 2019-07-18
Add API endpoint to export JSON representation of a document

brucebolt:
Creates a new API endpoint (`/government/admin/export/document/:id`) which permits all data about a document to be     exported from Whitehall and will be used during the migration to Content Publisher.

We grab all associations for each edition of the document and get all fields for each of these, resulting in a JSON blob giving the entire history of a document and all it's associated data.

Additionally, this endpoint is access protected through the `Export data` permission in Signon.  Any controllers for further endpoints which require the same access restriction can inherit from `Admin::Export::BaseController`.

An example of the output (with editor personal information redacted) is attached to the Trello card.  This example comes from the Whitehall news article which has received the most edits.

Trello card: https://trello.com/c/gtyLTQ8P/985-work-out-how-to-get-data-out-of-whitehall

### DIFF
--- a/app/controllers/admin/export/base_controller.rb
+++ b/app/controllers/admin/export/base_controller.rb
@@ -1,0 +1,14 @@
+class Admin::Export::BaseController < Admin::BaseController
+  before_action :require_export_data_permission!
+  respond_to :json
+
+private
+
+  def require_export_data_permission!
+    forbidden! unless current_user.can_export_data?
+  end
+
+  def forbidden!
+    respond_with Hash.new, status: :forbidden
+  end
+end
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -1,0 +1,31 @@
+class Admin::Export::DocumentController < Admin::BaseController
+  respond_to :json
+
+  self.responder = Api::Responder
+
+  def show
+    @document = Document.find(params[:id])
+    output = {
+               "document": @document,
+               "editions": []
+             }
+    @document.editions.each do |edition|
+      output[:editions].push(edition_associations(edition))
+    end
+    respond_with output
+  end
+
+private
+
+  def edition_associations(edition)
+    output = {
+               "edition": edition
+             }
+    output[:associations] = {}
+    associations = edition.class.reflect_on_all_associations.map(&:name)
+    associations.each do |association|
+      output[:associations][association] = edition.public_send(association)
+    end
+    output
+  end
+end
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -1,6 +1,4 @@
-class Admin::Export::DocumentController < Admin::BaseController
-  respond_to :json
-
+class Admin::Export::DocumentController < Admin::Export::BaseController
   self.responder = Api::Responder
 
   def show
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -26,6 +26,7 @@ class User < ApplicationRecord
     WORLD_EDITOR = 'World Editor'.freeze
     FORCE_PUBLISH_ANYTHING = 'Force publish anything'.freeze
     GDS_ADMIN = 'GDS Admin'.freeze
+    EXPORT_DATA = 'Export data'.freeze
   end
 
   def role
@@ -60,6 +61,10 @@ class User < ApplicationRecord
 
   def gds_admin?
     has_permission?(Permissions::GDS_ADMIN)
+  end
+
+  def can_export_data?
+    has_permission?(Permissions::EXPORT_DATA)
   end
 
   def scheduled_publishing_robot?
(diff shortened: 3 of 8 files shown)


app/controllers/export/base_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Export::BaseController < PublicFacingController
```
kevindew: I wonder if we'd benefit more from subclassing from Admin::BaseController since that requires authentication and it looks like that does things we might not need: `https://github.com/alphagov/whitehall/blob/master/app/controllers/public_facing_controller.rb#L6-L8`

app/controllers/export/base_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  skip_before_action :set_cache_control_headers
+  skip_before_action :restrict_request_formats
+  before_action :require_export_data_permission!
+  before_action :set_api_cache_control_headers
+  before_action :set_api_access_control_allow_origin_headers
```
kevindew: I don't think we'll need this since we're not requesting this client side, can probably also skip the cache control headers as I doubt we'll be using them.

app/controllers/export/document_controller.rb
```diff
@@ -0,0 +1,39 @@
+class Export::DocumentController < Export::BaseController
+  self.responder = Api::Responder
+
+  def show
+    @document = Document.find_by(id: params[:id])
```
kevindew: I assume this could just be `Document.find(params[:id])` that could save the need for the 404 handling as Rails hopefully covers that for us. Also are we expecting the interface to this to be the Whitehall private id or the content id?

app/controllers/export/document_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def respond_with_not_found
+    respond_with Hash.new, status: :not_found
+  end
+
+  def edition_associations(edition)
+    output = {
+               "edition": edition
+             }
+    if edition.class.methods.include? :reflect_on_all_associations
```
kevindew: Do we think there's much chance this would ever be false and we'd still want to proceed?

app/controllers/export/document_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      output[:associations] = {}
+      associations = edition.class.reflect_on_all_associations.map(&:name)
+      associations.each do |association|
+        output[:associations][association] = edition.send(association)
```
kevindew: minor, but good practice to use `public_send` to respect method privacy.

app/controllers/export/base_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  respond_to :json
+
+private
+
+  def require_export_data_permission!
+    forbidden! unless current_user.can_export_data?
```
kevindew: I'm a bit confused (I probably missed something) but I don't think this controller authenticates so wondering whether this blows up if current_user is nil or if this represents a different object when not authenticated.

app/controllers/export/document_controller.rb
brucebolt: My original plan was to iterate over the association's associations, which sometimes didn't have a `reflect_on_all_associations` method.  I'll remove this `if` statement since it's no longer needed (should've removed it before).

app/controllers/export/document_controller.rb
brucebolt: I did originally use `find` here, but it raises a `ActiveRecord::RecordNotFound` exception, which Rails does not handle

app/controllers/export/base_controller.rb
brucebolt: Ok, I'll remove these

app/controllers/export/base_controller.rb
brucebolt: This seems like a sensible thing to do 👍

app/controllers/export/base_controller.rb
brucebolt: I've changed `Export::BaseController` to `Admin::Export::BaseController`, which will inherit `Admin::BaseController` which has authentication.

Will leave the check for the export data permission in place.

app/controllers/export/document_controller.rb
kevindew: When you say Rails isn't handling it do you mean in the controller tests or when the application is running?

I'd expect that it's raised in controller tests but that it is actually converted to a 404 when the application is running.

brucebolt: There's no 404 returned when running the application using `find`, in place of `find_by`.  There's a 500 error with `ActiveRecord::RecordNotFound in Admin::Export::DocumentController#show`.

It appears this way of handling record not found errors is consistent across all the existing API controllers, e.g. https://github.com/alphagov/whitehall/blob/master/app/controllers/api/governments_controller.rb#L16-L20.

In this case, in addition to the 404 response code, you get a JSON response:
```
{
  "_response_info": {
    "status": "not found"
  }
}
```

PR #4911, opened 2019-07-18
Add rake task to reindex a collection of Documents

AlanGabbianelli:
This rake task reindexes all `Document`s associated with `Edition`s that
have a specific `world_location`. In order to do this, the
`WorldLocation`'s `slug` needs to be passed as an argument to the rake
task.

Trello card: https://trello.com/c/DauneS2z/1122-burma-myanmar-name-change-outstanding-issues

Co-authored-by: Mahmud Hussain <mahmud.hussain@digital.cabinet-office.gov.uk>
Co-authored-by: Alan Gabbianelli <alan.gabbianelli@digital.cabinet-office.gov.uk>

### DIFF
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -4,6 +4,21 @@ namespace :search do
     Document.find_by(content_id: args[:content_id]).published_edition.update_in_search_index
   end
 
+  desc "Re-index a collection of Documents with specified world location. Takes a `world_location_slug` as argument."
+  task :resend_documents_in_world_location, [:world_location_slug] => [:environment] do |_, args|
+    world_location = WorldLocation.find_by(slug: args[:world_location_slug])
+
+    if world_location
+      content_ids = Edition.in_world_location(world_location.id).map(&:content_id)
+
+      Document.where(content_id: content_ids).each do |document|
+        document.published_edition.try(:update_in_search_index)
+      end
+    else
+      puts "World location for #{args[:world_location_slug]} not found"
+    end
+  end
+
   desc "indexes all published searchable whitehall content"
   task index: ['search:index:detailed', 'search:index:government']
 


lib/tasks/search.rake
```diff
@@ -4,6 +4,21 @@ namespace :search do
...
 
+  desc "Re-index a collection of Documents with specified world location. Takes a `world_location_slug` as argument."
+  task :resend_document, [:world_location_slug] => [:environment] do |_, args|
+    world_location = WorldLocation.find_by(slug: args[:world_location_slug])
+
+    if world_location
+      content_ids = Edition.in_world_location(world_location.id).map(&:content_id)
+
+      Document.where(content_id: content_ids).each do |document|
+        document.published_edition.update_in_search_index
```
thomasleese: To avoid having to do `published_edition` here (and reduce the number of queries), I wonder if you could change the query to something like this:

```rb
documents = Edition.in_world_location(world_location.id).where(state: Edition::PUBLICLY_VISIBLE_STATES).map(&:document)
documents.each do { |d| d.update_in_search_index }
```

lib/tasks/search.rake
```diff
@@ -4,6 +4,21 @@ namespace :search do
     Document.find_by(content_id: args[:content_id]).published_edition.update_in_search_index
   end
 
+  desc "Re-index a collection of Documents with specified world location. Takes a `world_location_slug` as argument."
+  task :resend_document, [:world_location_slug] => [:environment] do |_, args|
```
thomasleese:
```suggestion
  task :resend_documents_in_world_location, [:world_location_slug] => [:environment] do |_, args|
```

I think this might be clearer for what the task does.

lib/tasks/search.rake
AlanGabbianelli: That is not exactly the same:
```
irb(main):009:0> Document.where(content_id: Edition.in_world_location(25).map(&:content_id)).each(&:published_edition).count
=> 424
irb(main):010:0> Edition.in_world_location(25).where(state: Edition::PUBLICLY_VISIBLE_STATES).map(&:document).count
=> 413
```

PR #4913, opened 2019-07-19
Bump to Govspeak 6.3

kevindew:
Trello: https://trello.com/c/LFamxCVY/954-investigating-relaxing-or-removing-sanitisation-from-govspeak-6-so-we-can-upgrade-whitehall

A variety of fixes have been done to hopefully allow for Whitehall to
upgrade to this without problems 🤞 see
https://github.com/alphagov/govspeak/pull/159 and
https://github.com/alphagov/govspeak/pull/164 for details.

As it has been over a year since Govspeak [was last updated](https://github.com/alphagov/whitehall/pull/3798) there is a risk here that this may introduce some surprises, so I'll be planning to brief 2nd-line to be alert for any tickets.


Gemfile
```diff
@@ -75,9 +75,7 @@ end
 if ENV['GOVSPEAK_DEV']
   gem 'govspeak', path: '../govspeak'
 else
-  # 5.5.0 only because sanitize 4.6.4 is stricter on govspeak, which makes attachments invalid
-  # See https://trello.com/c/90AjvFzy/77-whitehall-isnt-able-to-use-latest-govspeak-version
-  gem 'govspeak', '~> 5.5.0'
+  gem 'govspeak', '~> 6'
```
brucebolt: Does this need to be `~> 6.3` (i.e. `>=6.3` and `<7`) if we need specific features from 6.3?

kevindew: That's a good point. I'll update.

PR #4917, opened 2019-07-19
Fix contact update not republishing org

benthorner:
https://govuk.zendesk.com/agent/tickets/3749668

Previously the organisation content item would not update if the data
for one of its contacts changed. This ensures any update to a contact
behaves similarly to their creation and destruction by republishing the
organisation content.

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -16,6 +16,7 @@ class Contact < ApplicationRecord
 
   after_update :republish_dependent_editions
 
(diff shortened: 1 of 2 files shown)


benthorner:
Closing as we believe this should be handled by dependency resolution, although the behaviour for new/deleted contacts seems to be unclear...

https://github.com/alphagov/whitehall/commit/5c49bc99ae5f042ba258fc212acb0abf746deead

PR #4920, opened 2019-07-22
Exclude locked documents from featuring

emmabeynon:
For https://trello.com/c/XeW2zCKr/1017-migrated-documents-appear-and-dont-appear-in-the-right-places-in-whitehall

This ensures that locked documents cannot be featured on organisation, world location and topical event pages, avoiding broken links to content that will eventually be managed in Content Publisher.

### DIFF
--- a/app/models/admin/edition_filter.rb
+++ b/app/models/admin/edition_filter.rb
@@ -117,7 +117,7 @@ module Admin
       editions = editions.from_date(from_date) if from_date
       editions = editions.to_date(to_date) if to_date
       editions = editions.only_broken_links if only_broken_links
-      editions = editions.without_locked_documents
+      editions = editions.without_locked_documents unless include_locked_documents?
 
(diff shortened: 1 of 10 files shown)


app/models/admin/edition_filter.rb
```diff
@@ -117,6 +117,7 @@ def unpaginated_editions
       editions = editions.from_date(from_date) if from_date
       editions = editions.to_date(to_date) if to_date
       editions = editions.only_broken_links if only_broken_links
+      editions = editions.without_locked_documents
```
brucebolt: Will this also exclude locked documents from the index page?  If this is the case, then it may be worth doing the part of the card to maintain this functionality as part of this PR, so functionality is not lost.

emmabeynon: Sure, was planning on raising a separate PR, but can add to this one 👍

emmabeynon: Now added :)

PR #4926, opened 2019-07-23
Append image file_url to image response hash

1pretz1:
This commit appends the image url to each image in the response hash,
this is needed so we know where the live file lives on asset manager.

Trello:
https://trello.com/c/QKdmIBcQ/1035-add-image-data-to-document-export


app/controllers/admin/export/document_controller.rb
```diff
@@ -21,9 +21,21 @@ def edition_associations(edition)
...
+      else
+        output[:associations][association] = edition.public_send(association)
+      end
     end
     output
   end
+
+  def complete_images_hash(edition_images)
+    edition_images.map do |image|
+      image.attributes.merge("file_url" => image.image_data.file_url)
```
kevindew: Btw you can call [`.url`] on an image object to avoid the need to access image_data: https://github.com/alphagov/whitehall/blob/2f580a2ded3964394d5de9d08a28cda12fecd9d2/app/models/image.rb#L14-L16

Also It seems a bit weird to be merging into the attributes of a model here. It might be better to use as_json to convert it into a hash. So this could be `edition_images.map { |i| i.as_json(methods: :url) }` and may mean you avoid the need for an extra method.

1pretz1: Ah ah yep that is better & I didn't know `as_json` could do such things... done!

1pretz1:
@kevindew I'm going to wait until Bruce has merged his PR here - https://github.com/alphagov/whitehall/pull/4927/files then adjust mine accordingly to use the new export class.

Also brakeman is failing for all builds of whitehall but there is a PR up to fix this.

brucebolt:
I've merged https://github.com/alphagov/whitehall/pull/4927, so this should now be ready for rebasing

PR #4927, opened 2019-07-24
Include public URL of embedded Whitehall admin links in document export API

brucebolt:
Some publishers use Whitehall's admin URLs in their content, instead of public URLs.  Whitehall replaces these with public URLs when the document is published.

Since Content Publisher has no knowledge of the Whitehall document structure, these URLs need to be resolved during the export of content from Whitehall.

The resolved links are added to each edition, e.g.:
```
"whitehall_admin_links": [
  {
    "whitehall_admin_url": "/government/admin/publications/973699",
    "public_url": "http://www.dev.gov.uk/government/publications/pacer-competition-application-form-and-guidelines"
  },
  {
    "whitehall_admin_url": "/government/admin/news/958818",
    "public_url": "http://www.dev.gov.uk/government/news/new-trains-to-be-rolled-out-across-the-country-alongside-48-billion-investment-to-upgrade-tracks"
  },
  {
    "whitehall_admin_url": "/government/admin/news/815912",
    "public_url": "http://www.dev.gov.uk/government/news/grayling-progress-on-transpennine-route-upgrade-and-crossrail-2-to-advance-in-lockstep"
  }
]
```

Trello card: https://trello.com/c/Azgd0U80

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -1,4 +1,7 @@
 class Admin::Export::DocumentController < Admin::Export::BaseController
+  include GovspeakHelper
+  include PublicDocumentRoutesHelper
+
   self.responder = Api::Responder
 
   def show
@@ -17,13 +20,31 @@ private
 
   def edition_associations(edition)
     output = {
-               "edition": edition
+               "edition": edition,
+               "associations": {},
+               "whitehall_admin_links": []
              }
-    output[:associations] = {}
     associations = edition.class.reflect_on_all_associations.map(&:name)
     associations.each do |association|
       output[:associations][association] = edition.public_send(association)
     end
+    output[:whitehall_admin_links].concat(resolve_whitehall_admin_links(edition.body))
+    if edition.withdrawn?
+      output[:whitehall_admin_links].concat(resolve_whitehall_admin_links(edition.unpublishing.explanation))
+    end
     output
+  end
+
+  def resolve_whitehall_admin_links(body)
+    whitehall_admin_links(body).map { |link|
+      { "whitehall_admin_url": link, "public_url": public_url_for_admin_link(link) }
+    }
+  end
+
+  def public_url_for_admin_link(url)
+    edition = Whitehall::AdminLinkLookup.find_edition(url)
+    if edition.present?
+      public_document_url(edition)
+    end
   end
 end
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -1,50 +1,8 @@
 class Admin::Export::DocumentController < Admin::Export::BaseController
-  include GovspeakHelper
-  include PublicDocumentRoutesHelper
-
   self.responder = Api::Responder
 
   def show
     @document = Document.find(params[:id])
-    output = {
-               "document": @document,
-               "editions": []
-             }
-    @document.editions.each do |edition|
-      output[:editions].push(edition_associations(edition))
-    end
-    respond_with output
-  end
-
-private
-
-  def edition_associations(edition)
-    output = {
-               "edition": edition,
-               "associations": {},
-               "whitehall_admin_links": []
-             }
-    associations = edition.class.reflect_on_all_associations.map(&:name)
-    associations.each do |association|
-      output[:associations][association] = edition.public_send(association)
-    end
-    output[:whitehall_admin_links].concat(resolve_whitehall_admin_links(edition.body))
-    if edition.withdrawn?
-      output[:whitehall_admin_links].concat(resolve_whitehall_admin_links(edition.unpublishing.explanation))
-    end
-    output
-  end
-
-  def resolve_whitehall_admin_links(body)
-    whitehall_admin_links(body).map { |link|
-      { "whitehall_admin_url": link, "public_url": public_url_for_admin_link(link) }
-    }
-  end
-
-  def public_url_for_admin_link(url)
-    edition = Whitehall::AdminLinkLookup.find_edition(url)
-    if edition.present?
-      public_document_url(edition)
-    end
+    respond_with DocumentExportPresenter.new(@document)
   end
 end
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -103,6 +103,12 @@ module GovspeakHelper
     { heading_numbering: numbering_method, contact_heading_tag: 'h4' }
   end
 
+  def whitehall_admin_links(body)
+    govspeak = build_govspeak_document(body)
+    links = Govspeak::LinkExtractor.new(govspeak).call
+    links.select { |link| DataHygiene::GovspeakLinkValidator::is_internal_admin_link?(link) }
+  end
+
 private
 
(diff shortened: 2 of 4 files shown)


app/controllers/admin/export/document_controller.rb
```diff
@@ -24,6 +27,24 @@ def edition_associations(edition)
     associations.each do |association|
       output[:associations][association] = edition.public_send(association)
     end
+    output[:whitehall_admin_links] = resolve_whitehall_admin_links(edition.body)
```
kevindew: Do you think we should start moving this code out into a presenter soon?

Also, I'm hoping this is the case but is there risk these could be in other fields than edition.body? I feel like they might be something that can be part of a withdrawn notice and maybe an unpublishing - hopefully no others 😬

app/controllers/admin/export/document_controller.rb
```diff
@@ -24,6 +27,24 @@ def edition_associations(edition)
...
+    whitehall_links.each do |link|
+      public_url = public_url_for_admin_link(link)
+      public_urls.push("whitehall_admin_url": link, "public_url": public_url)
+    end
+    public_urls
+  end
+
+  def public_url_for_admin_link(url)
+    edition = Whitehall::AdminLinkLookup.find_edition(url)
+    if edition.present? && edition.linkable?
```
kevindew: What does linkable mean in this context? I ask as I'm wondering if we want to link even if it's not linkable so we have something to fill in and may be a 404 or a redirect but that seems to reveal more than if we have a nil response

app/controllers/admin/export/document_controller.rb
```diff
@@ -24,6 +27,24 @@ def edition_associations(edition)
...
     output
   end
+
+  def resolve_whitehall_admin_links(body)
+    whitehall_links = whitehall_admin_links(body)
+    public_urls = []
+    whitehall_links.each do |link|
+      public_url = public_url_for_admin_link(link)
+      public_urls.push("whitehall_admin_url": link, "public_url": public_url)
+    end
```
kevindew: This looks like it'd more idiomatic as a map rather than an each iteration. 

Could get the whole method down to `whitehall_admin_links(body).map { |link| { whitehall_admin_url: link, public_url: public_url_for_admin_link(link) }`

app/helpers/govspeak_helper.rb
```diff
@@ -103,6 +103,18 @@ def govspeak_options_for_html_attachment(attachment)
...
 
+  def whitehall_admin_links(body)
+    govspeak = build_govspeak_document(body)
+    links = Govspeak::LinkExtractor.new(govspeak).call
+    urls = []
+    links.each do |link|
+      if DataHygiene::GovspeakLinkValidator::is_internal_admin_link?(link)
+        urls.push(link)
+      end
+    end
```
kevindew: This iteration should probably be a select rather than an each.

app/controllers/admin/export/document_controller.rb
brucebolt: It means the document has been marked as published.  In this case, we probably don't want to have this check, so we still get the forecast public URL for documents that are still in draft.

app/controllers/admin/export/document_controller.rb
brucebolt: I've refactored the controller to use a presenter, which seems like a very sensible change.

I guess we'll need to audit all the places these links are included.  I'll have to think about a way of doing that.

brucebolt: As far as I can see, it is only the edition body, attachments and the alternative format contact email that are processed as Govspeak (and therefore Whitehall URLs replaced).

app/presenters/admin/export/document_presenter.rb
```diff
@@ -0,0 +1,44 @@
...
+    associations = edition.class.reflect_on_all_associations.map(&:name)
+    associations.each do |association|
+      output[:associations][association] = edition.public_send(association)
+    end
+    output[:whitehall_admin_links] = resolve_whitehall_admin_links(edition.body)
+    output
+  end
+
+  def resolve_whitehall_admin_links(body)
+    whitehall_admin_links(body).map { |link|
```
kevindew: As this is a multi-line block it'd be more conventional to use do notation: https://rubystyle.guide/#single-line-blocks

app/presenters/admin/export/document_presenter.rb
```diff
@@ -0,0 +1,44 @@
...
+      { "whitehall_admin_url": link, "public_url": public_url_for_admin_link(link) }
+    }
+  end
+
+  def public_url_for_admin_link(url)
+    edition = Whitehall::AdminLinkLookup.find_edition(url)
+    if edition.present?
+      Whitehall.url_maker.public_document_url(edition)
```
kevindew: Could put the conditional inline with this

app/controllers/admin/export/document_controller.rb
kevindew: Withdrawn notices are done by Govspeak as well atleast:

```
app/workers/publishing_api_withdrawal_worker.rb
14:      explanation: Whitehall::GovspeakRenderer.new.govspeak_to_html(explanation),

app/workers/publishing_api_gone_worker.rb
4:      rendered_explanation = Whitehall::GovspeakRenderer
```

kevindew: I can see withdrawals and gones are too:

```
app/workers/publishing_api_withdrawal_worker.rb
14:      explanation: Whitehall::GovspeakRenderer.new.govspeak_to_html(explanation),

app/workers/publishing_api_gone_worker.rb
4:      rendered_explanation = Whitehall::GovspeakRenderer
```

there's also HTML attachments, which I guess we're probably pulling out with this data.

brucebolt: Turns out we're not exporting the body of HTML attachments.  I'll write up a card for that (the Whitehall links can be dealt with at the same time).

I've addressed the withdrawal explanation in this PR.

app/presenters/admin/export/document_presenter.rb
```diff
@@ -0,0 +1,44 @@
+class Admin::Export::DocumentPresenter < Whitehall::Decorators::Decorator
```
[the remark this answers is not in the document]
brucebolt: Hmm, if I move the presenter out of the Admin namespace, but leave the controller in the `Admin` namespace, then I'll have to require the DocumentPresenter in the controller.  Is this what you mean?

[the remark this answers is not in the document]
kevindew: No sorry that's not what I meant. 

I meant that many of the presenters that lack an admin namespace are already used in an admin context so giving them this namespace would be a potential point of confusion and shouldn't add any value since the Export namespace should suffice.

Although looking at the code the admin controllers only seem to use one presenter:

```
➜  admin git:(add-document-export-api) ✗ ag Presenter
roles_controller.rb
68:    RoleTypePresenter.role_attributes_from(params[:role][:role_type])[:type]
```

you shouldn't have to do anything with requiring or scoping class names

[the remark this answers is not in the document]
brucebolt: I see, thanks for the explanation 👍.  Made the change and pushed the update.

PR #4928, opened 2019-07-24
Set lang attribute correctly on people page headings

injms:
When translated, the roles, policy, and announcements on a person page can have a mix of two different languages. 

This patch sets the headings to have the appropriate `lang` attribute depending on what language is being used. This is limited to content that comes from the internationalisation files because the language can be detected.

Other content doesn't seem to be have information about which language the content is written in so the `lang` attribute can't be set correctly. An example of this is the previous role - the string 'British High Commissioner to Singapore' hasn't been translated so defaults back to English.

These examples are from [Scott Wightman's page](https://www.gov.uk/government/people/scott-wightman.ko).

## Previous roles before
Screen Shot 2019-07-29 at 16 35 38
Screen Shot 2019-07-29 at 16 36 31

## Previous roles after
Screen Shot 2019-08-02 at 12 32 43
Screen Shot 2019-08-02 at 12 34 27

## Biography heading before
Screen Shot 2019-08-02 at 12 41 10

Screen Shot 2019-08-02 at 12 36 05

## Biography heading after
Screen Shot 2019-08-02 at 12 40 49

Screen Shot 2019-08-02 at 12 35 37

## Page navigation before
Screen Shot 2019-08-02 at 12 38 38

Screen Shot 2019-08-02 at 12 36 58

## Page navigation after
Screen Shot 2019-08-02 at 12 39 37

Screen Shot 2019-08-02 at 12 38 00

### DIFF
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -1,7 +1,7 @@
 <% page_title @person.name %>
-<% page_class "people-show biographical-page" %>
+<% page_class 'people-show biographical-page' %>
 
-<%= content_tag_for :div, @person, class: "two-column-page" do %>
+<%= content_tag_for :div, @person, class: 'two-column-page' do %>
   <header class="block headings-block">
     <div class="inner-block floated-children">
       <%= render partial: 'shared/heading',
@@ -26,8 +26,8 @@
             </figure>
           </div>
         <% end %>
-        <h1 <%= t_lang('document.contents') %>><%= t('document.contents') %></h1>
-        <nav role="navigation">
+        <nav role="navigation" aria-labelledby="document-contents-heading">
+          <h2 id="document-contents-heading" <%= t_lang('document.contents') %>><%= t('document.contents') %></h2>
           <ul>
             <li <%= t_lang('people.biography') %> >
               <%= link_to t('people.biography'), '#biography' %>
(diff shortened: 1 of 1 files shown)


app/views/people/show.html.erb
```diff
@@ -128,10 +127,11 @@
...
 
           <div class="read-more">
-            <%= link_to t('announcements.view_all'), announcements_path(people: [@person.slug]) %>
+            <%= link_to t('announcements.view_all', locale: :en), announcements_path(people: [@person.slug]) %>
           </div>
         </section>
       <% end %>
     </div>
   </div>
+  <% binding.pry %>
```
sihugh: this line might need to come out 😉

injms: Sorry, should've attached a `do no review` tag to this. Since Whitehall is refusing to run locally, I'm having to use integration at the moment - hence the use of pry 😄 Good catch though!

app/views/people/show.html.erb
```diff
@@ -26,14 +26,14 @@
...
-        <h1 <%= t_lang('document.contents') %>><%= t('document.contents') %></h1>
-        <nav role="navigation">
+        <nav role="navigation" aria-labelledby="document-contents-heading">
+          <h2 id="document-contents-heading" <%= t_lang('document.contents') %>><%= t('document.contents') %></h2>
           <ul>
             <li <%= t_lang('people.biography') %> >
               <%= link_to t('people.biography'), '#biography' %>
             </li>
             <% if @person.currently_in_a_role? %>
-              <li <%= t_lang('roles.heading', count: @person.current_role_appointments.count) %> >
```
vanitabarrett: I think we might still need to pass `count: @person.current_role_appointments.count` as an option to `t_lang`?

injms: Ah, sorry! I've reverted that line now. Thanks for spotting that 👍

injms:
[(3) Mark mixed languages on people pages](https://trello.com/c/Uq7gAzCu/13-3-mark-mixed-languages-on-people-pages)

injms:
[(3) Mark mixed languages on people pages](https://trello.com/c/Uq7gAzCu/13-3-mark-mixed-languages-on-people-pages)

PR #4929, opened 2019-07-24
Add lang attribute to contents list

vanitabarrett:
## What
Add a function to Whitehall `TranslationHelper` which detects whether a translation is using the current locale or falling back to the default locale. Use this method to add the `lang` attribute to contents list items when appropriate.

## Why
We have some cases where a translated page contains links that are still in English, because we do not have the translations for that particular langauge ([example here](https://www.gov.uk/government/people/scott-wightman.ko)). These currently don't have the `lang` attribute, meaning a screenreader thinks the text is in the same language as the page. This PR adds`lang="en" when the text is falling back to English.

## Before (no lang present)
Screen Shot 2019-07-24 at 13 21 19
Screen Shot 2019-07-24 at 13 26 35

## After
Screen Shot 2019-07-24 at 13 21 19
Screen Shot 2019-07-24 at 13 26 21

## Notes
There doesn't seem to be a built-in I18n method for detecting which locale is serving the translation. The closest they have is the `I18n.exists?` method. This doesn't work for our use case as it takes fallbacks into account, so it will return true for `I18n.exists?(:de, "people.biography")` even if that translation doesn't exist, as long as an English translation exists.

Other approaches considered:

- Removing the ability to translate these strings and always marking the section as english. This means we wouldn't have mixed languages on the page, but ideally we want to make the most of the translations we have, especially if we may add more in the future.

Trello: https://trello.com/c/Uq7gAzCu/13-3-mark-language-as-english-on-subscription-links

### DIFF
--- a/app/helpers/translation_helper.rb
+++ b/app/helpers/translation_helper.rb
@@ -29,6 +29,25 @@ module TranslationHelper
     I18n.t("document.speech.#{speech_type.published_externally_key}")
   end
 
+  def t_lang(key, options = {})
+    fallback = t_fallback(key, options)
+    if fallback && fallback != I18n.locale
+      "lang=#{fallback}"
+    end
+  end
+
+  def t_fallback(key, options = {})
+    translation = I18n.t(key, options, locale: I18n.locale, fallback: false, default: "fallback")
+
+    if !translation || translation.eql?("fallback")
+      I18n.default_locale
+    elsif translation.is_a? Hash
+      translation.values.all?(&:nil?) ? I18n.default_locale : false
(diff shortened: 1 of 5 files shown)


vanitabarrett:
[(3) Mark language as English on subscription links](https://trello.com/c/Uq7gAzCu/13-3-mark-language-as-english-on-subscription-links)

vanitabarrett:
[Mark contents list for mixed language pages](https://trello.com/c/FghnzvsI/37-mark-contents-list-for-mixed-language-pages)

PR #4930, opened 2019-07-24
Prevent external API calls for locked documents

emmabeynon:
For https://trello.com/c/QaZakMWe/1018-disable-whitehall-external-api-calls-on-a-migrated-document

This attempts to find places where Publishing API calls are made and raise an error if the document that is being changed is locked (pending migration to Content Publisher).  The initial plan was to focus on raising an error in the PublishingApiPusher as it seemed like a central place for handling Publishing API calls for editions, however it seems there are various workers being called from all sorts of different places in Whitehall, so it feels prudent to add checks for locked documents to these as well.

There are some controllers and services making calls directly to the Publishing API but it feels excessive at this point to add checks to all of them.  Instead, I propose adding in checks where necessary as we come across these calls during the course of the [work to lock documents in Whitehall](https://trello.com/c/tWSO1M1R/1020-lock-a-document-in-whitehall).

We also add a check before calling the SearchIndexAddWorker to prevent locked documents from being added to the search index via the Search API.  We've decided not to add this check before deleting the document as it's probably no bad thing to remove locked documents from the government index, as once they're in Content Publisher they get indexed into the govuk index via the Publishing API anyway.

API calls to Asset Manager are quite confusing in this application but we believe that the most important one to prevent in this context is for updating attachments.

### DIFF
--- a/app/concerns/locked_document_concern.rb
+++ b/app/concerns/locked_document_concern.rb
@@ -3,11 +3,12 @@ module LockedDocumentConcern
 
   class LockedDocumentError < RuntimeError; end
 
-  def check_if_locked_document(content_id: nil, document: nil)
+  def check_if_locked_document(content_id: nil, document: nil, edition: nil)
     content_id_locked = content_id && Document.exists?(locked: true, content_id: content_id)
     document_locked = document && document.locked?
+    edition_locked = edition && edition.locked?
 
-    if content_id_locked || document_locked
+    if content_id_locked || document_locked || edition_locked
       raise LockedDocumentError, "Cannot perform this operation on a locked document"
     end
   end
--- a/app/concerns/locked_document_concern.rb
+++ b/app/concerns/locked_document_concern.rb
@@ -1,0 +1,13 @@
+module LockedDocumentConcern
+  extend ActiveSupport::Concern
+
+  class LockedDocumentError < RuntimeError; end
+
+  def check_if_locked_document(content_id: nil)
+    content_id_locked = content_id && Document.exists?(locked: true, content_id: content_id)
+
+    if content_id_locked
+      raise LockedDocumentError, "Cannot perform this operation on a locked document"
+    end
+  end
+end
--- a/app/concerns/locked_document_concern.rb
+++ b/app/concerns/locked_document_concern.rb
@@ -3,10 +3,11 @@ module LockedDocumentConcern
 
   class LockedDocumentError < RuntimeError; end
 
-  def check_if_locked_document(content_id: nil)
+  def check_if_locked_document(content_id: nil, document: nil)
     content_id_locked = content_id && Document.exists?(locked: true, content_id: content_id)
+    document_locked = document && document.locked?
 
-    if content_id_locked
+    if content_id_locked || document_locked
       raise LockedDocumentError, "Cannot perform this operation on a locked document"
     end
   end
--- a/app/services/service_listeners/attachment_updater.rb
+++ b/app/services/service_listeners/attachment_updater.rb
@@ -1,6 +1,12 @@
 module ServiceListeners
   class AttachmentUpdater
+    extend LockedDocumentConcern
+
     def self.call(attachable: nil, attachment_data: nil)
+      if attachable.is_a?(Edition)
+        check_if_locked_document(edition: attachable)
+      end
+
       update_attachable! attachable if attachable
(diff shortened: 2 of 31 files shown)


app/services/service_listeners/publishing_api_pusher.rb
```diff
@@ -10,6 +10,8 @@ def push(event:, options: {})
       # This is done synchronously before the rest of the publishing.
       # Currently (02/11/2016) publishing-api links
       # are not recalculated on parent documents when their translations are unpublished.
```
cbaines: I think this comment specifically relates to the `handle_translations` lines.

emmabeynon: Ah, I will move the raise above it 👍

app/services/service_listeners/publishing_api_pusher.rb
```diff
@@ -7,6 +7,10 @@ def initialize(edition)
     end
 
     def push(event:, options: {})
+      if edition.locked?
+        raise RuntimeError, "Cannot send a locked document to the Publishing API"
```
kevindew: RuntimeError is actually superfluous here as a raise will default to RuntimeError.

I do wonder if this should/could be a specific exception.

emmabeynon:
@kevindew Thanks Kevin, for some reason I thought concerns were only for models but if we can use them elsewhere in the app then it makes sense to go down that route.  We could also create a specific LockedDocumentError or similar?

test/functional/admin/attachments_controller_test.rb
```diff
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class Admin::AttachmentsControllerTest < ActionController::TestCase
+  include LockedDocumentConcern
```
kevindew: Is this used to have the exception in scope? If so you can avoid the need for this by referencing the exception as `LockedDocumentConcern::LockedDocumentError` as including the module seems to apply we're using methods from the module.

app/concerns/locked_document_concern.rb
```diff
@@ -0,0 +1,31 @@
...
+    return unless document.present?
+
+    raise_locked_document_error if document.locked?
+  end
+
+  def check_if_locked_document_with_edition(edition)
+    return unless edition
+
+    raise_locked_document_error if edition.locked?
+  end
```
kevindew: Seeing these with the `with_` suffix makes me think these are looking more like candidates for keyword arguments:

for example:

```
def check_if_locked_document(content_id: nil, edition: nil, document: nil)
  document_locked = document && document.locked?
  edition_locked = edition && edition.document.locked?
  content_id_locked = content_id && Document.exists?(locked: true, content_id: content_id)

  if document_locked || edition_locked || content_id_locked
    raise LockedDocumentError, "Cannot perform this operation on a locked document"
  end 
end
```

I'm also suggesting here switching to use a document model rather than a document_id for consistency with edition - I haven't found a case where this wouldn't meet.

app/services/service_listeners/attachment_updater.rb
```diff
@@ -1,6 +1,13 @@
 module ServiceListeners
   class AttachmentUpdater
+    extend LockedDocumentConcern
+
     def self.call(attachable: nil, attachment_data: nil)
+      if attachable.is_a?(Edition)
+        document = attachable.document
+        check_if_locked_document_with_document_id(document.id)
```
kevindew: Since we're checking the item is an edition it might be slightly simpler to follow if we pass the edition into check.

app/concerns/locked_document_concern.rb
```diff
@@ -0,0 +1,31 @@
+module LockedDocumentConcern
+  extend ActiveSupport::Concern
+
+  class LockedDocumentError < RuntimeError; end
```
kevindew: Specific error looks good 👍

emmabeynon:
Thanks @kevindew - your comments should be addressed now.

test/unit/workers/publishing_api_discard_draft_worker_test.rb
```diff
@@ -1,7 +1,7 @@
 require 'test_helper'
 require 'gds_api/test_helpers/publishing_api_v2'
 
-class PublishingApDiscardDraftiWorkerTest < ActiveSupport::TestCase
```
kevindew: Good catch 👍

PR #4932, opened 2019-07-24
Switch cookie serializer to json

chao-xian:
Brakeman is reporting this as a fix, otherwise Whitehall's builds aren't passing.

### DIFF
--- a/config/initializers/cookies_serializer.rb
+++ b/config/initializers/cookies_serializer.rb
@@ -2,4 +2,4 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :marshal
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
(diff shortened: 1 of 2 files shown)


config/initializers/cookies_serializer.rb
```diff
@@ -2,4 +2,4 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :marshal
+Rails.application.config.action_dispatch.cookies_serializer = :json
```
thomasleese:
```suggestion
Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```

From this conversation it sounds better to first switch to `:hybrid` and then switch to `:json` a bit later: https://gds.slack.com/archives/CAB4Q3QBW/p1564042299039500

https://blog.bigbinary.com/2014/12/23/migrating-existing-session-cookies-while-upgrading-to-rails-4-1-and-above.html

thomasleese:
You'll likely need to use [`brakeman -I` to ignore the failure here](https://docs.publishing.service.gov.uk/manual/brakeman.html#dealing-with-false-positives).

chao-xian:
[Fix Smokey errors with youtube-nocookie](https://trello.com/c/YvdU9D39/33-fix-smokey-errors-with-youtube-nocookie)

PR #4934, opened 2019-07-25
Add government information to each edition

1pretz1:
This PR adds which government was associated with each edition of a document. The political boolean doesn't need to be added as it's already present on the data exported from the edition.

Trello:
https://trello.com/c/1uhpv9pA/1038-add-government-field-to-document-export

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -32,7 +32,9 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
         output[:associations][association] = edition.public_send(association)
       end
     end
-    output[:whitehall_admin_links].concat(resolve_whitehall_admin_links(edition.body))
+
+    output[:government] = edition.government
+    output[:whitehall_admin_links] = resolve_whitehall_admin_links(edition.body)
     if edition.withdrawn?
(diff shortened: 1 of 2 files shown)


app/presenters/document_export_presenter.rb
```diff
@@ -55,4 +58,22 @@ def public_url_for_admin_link(url)
...
+  end
+
+  def government(edition)
+    gov = edition.government
+    return nil unless gov
+
+    {
+      title: gov.name,
+      slug: gov.slug,
+      current: gov.current?
```
brucebolt: Instead of a boolean for this being the current government, would the start/end dates of the government be a better value to include?  This would then allow Content Publisher to determine itself whether the government is current.  Since we haven't yet implemented history mode in Content Publisher, it might be better to have more data than less.

With that in mind, would it be possible just to include the entire `Government` object here since it contains all this data?

1pretz1: Yes sure I'll include the full `Government` object, I initially went with this approach as it's all publishing-api needs to represent the document as political, but more data is better than less so... changes incoming 💯

1pretz1:
@brucebolt I'm now only exporting the `Government` object in this PR as we don't actually need the `political` flag as it's stored on the `edition` part of the export, now content-publisher will have all the information itself and can decipher if the edition is political / from a previous government etc.

PR #4936, opened 2019-07-26
Disable Whitehall DB operations on a document when locked

Tetrino:
When a document has been migrated it is locked on Whitehall side.
This commit takes steps to ensure that when locked, the user
cannot edit the document at a database level even if UI functionality
is exposed to allow it.

This commit:
- Adds a locking check on Feature, Document, Edition and
DocumentCollectionGroupMembership as a before_* step based on
the behaviour that could occur in that class (validation or update).
- Adds model tests for the above that ensures the expected behaviour.
- Modifies the way editions are created for existing document lock
tests such that they work with the new behaviour for edition lock checks.

This commit uses the new LockedDocumentConcern created for this purpose,
but relies on untidy logic to work around the issues with class vs
instance methods when passing the required objects to the concern.

Trello card: https://trello.com/c/iSEv8w6b/1027-disable-whitehall-db-operations-on-a-document-when-locked

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -4,6 +4,7 @@ class Document < ApplicationRecord
   extend FriendlyId
 
   include Document::Needs
+  include LockedDocumentConcern
 
   friendly_id :sluggable_string, use: :scoped, scope: :document_type
 
@@ -38,6 +39,8 @@ class Document < ApplicationRecord
   has_many :document_collection_groups, through: :document_collection_group_memberships
   has_many :document_collections, through: :document_collection_groups
   has_many :features, inverse_of: :document, dependent: :destroy
+
+  before_save { check_if_locked_document(document: self) unless locked_changed? }
 
   validates_presence_of :content_id
 
--- a/app/models/document_collection_group_membership.rb
+++ b/app/models/document_collection_group_membership.rb
@@ -1,10 +1,13 @@
 class DocumentCollectionGroupMembership < ApplicationRecord
   BANNED_DOCUMENT_TYPES = %w[DocumentCollection].freeze
+  include LockedDocumentConcern
 
   belongs_to :document, inverse_of: :document_collection_group_memberships
   belongs_to :document_collection_group, inverse_of: :memberships
 
   before_create :assign_ordering
+
+  before_save { check_if_locked_document(document: self.document) }
 
   validates :document, presence: true
   validates :document_collection_group, presence: true
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -31,6 +31,7 @@ class Edition < ApplicationRecord
   extend Edition::FindableByWorldwideOrganisation
 
   include Searchable
+  include LockedDocumentConcern
 
   has_many :editorial_remarks, dependent: :destroy
   has_many :edition_authors, dependent: :destroy
@@ -92,6 +93,7 @@ class Edition < ApplicationRecord
 
   # @!group Callbacks
   before_save :set_public_timestamp
+  before_save { check_if_locked_document(edition: self) }
   # @!endgroup
 
   class UnmodifiableValidator < ActiveModel::Validator
--- a/app/models/feature.rb
+++ b/app/models/feature.rb
@@ -1,8 +1,11 @@
 class Feature < ApplicationRecord
+  include LockedDocumentConcern
   belongs_to :document
   belongs_to :topical_event
   belongs_to :offsite_link
   belongs_to :feature_list
+
+  before_save { check_if_locked_document(document: self.document) }
 
   after_save :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
--- a/features/step_definitions/featuring_locked_documents_steps.rb
+++ b/features/step_definitions/featuring_locked_documents_steps.rb
@@ -1,10 +1,8 @@
-Given(/^a locked document titled "([^"]*)"$/) do |title|
+Given(/^a document titled "([^"]*)"$/) do |title|
   @edition = create(:published_news_article, title: title)
-  @edition.document.locked = true
-  @edition.document.save
 end
 
-And(/^the locked document is tagged to organisation "([^"]*)"$/) do |organisation_name|
+And(/^the document is tagged to organisation "([^"]*)"$/) do |organisation_name|
   organisation = Organisation.find_by(name: organisation_name)
   @edition.organisations = [organisation]
   @edition.save
@@ -22,7 +20,7 @@ Then(/^I cannot see the document in the list of featurable documents$/) do
   end
 end
 
-And(/^the locked document is tagged to the world location "([^"]*)"$/) do |world_location_name|
+And(/^the document is tagged to the world location "([^"]*)"$/) do |world_location_name|
   world_location = WorldLocation.find_by!(name: world_location_name)
   @edition.world_locations = [world_location]
   @edition.save
@@ -33,7 +31,7 @@ And(/^I visit the world location admin page for "([^"]*)"$/) do |world_location_
   visit admin_world_location_path(world_location)
 end
 
-And(/^the locked document is tagged to the topical event "([^"]*)"$/) do |topical_event_name|
+And(/^the document is tagged to the topical event "([^"]*)"$/) do |topical_event_name|
   topical_event = TopicalEvent.find_by(name: topical_event_name)
   @edition.topical_events = [topical_event]
   @edition.save
@@ -42,4 +40,9 @@ end
 And(/^I visit the topical event admin page for "([^"]*)"$/) do |topical_event_name|
   topical_event = TopicalEvent.find_by!(name: topical_event_name)
   visit admin_topical_event_classification_featurings_path(topical_event)
+end
+
+And(/^the document is locked/) do
+  @edition.document.locked = true
+  @edition.document.save
 end
--- a/test/factories/editions.rb
+++ b/test/factories/editions.rb
@@ -161,6 +161,12 @@ FactoryBot.define do
         edition.unpublishing = build(:withdrawn_unpublishing, edition: edition)
       end
     end
+
+    trait(:with_locked_document) do
+      after(:create) do |edition|
+        edition.document.update(locked: true)
+      end
+    end
   end
 
   factory :announcement, parent: :edition, class: Announcement, traits: %i[with_organisations with_topics]
--- a/test/functional/admin/attachments_controller_test.rb
+++ b/test/functional/admin/attachments_controller_test.rb
@@ -250,7 +250,7 @@ class Admin::AttachmentsControllerTest < ActionController::TestCase
   end
 
   test "POST :create with an edition belonging to a locked document does not save the attachment" do
-    edition = create(:edition, document: build(:document, locked: true))
+    edition = create(:unpublished_edition, :with_locked_document)
     assert_raise LockedDocumentConcern::LockedDocumentError, "Cannot perform this operation on a locked document" do
       post :create, params: { edition_id: edition, attachment: valid_file_attachment_params }
     end
--- a/test/unit/document_test.rb
+++ b/test/unit/document_test.rb
@@ -238,4 +238,12 @@ class DocumentTest < ActiveSupport::TestCase
 
     assert_equal document.patch_meets_user_needs_links, "Links updated"
   end
+
+  test 'should raise an exception when attempting to modify a locked document' do
+    document = create(:document, locked: true)
+    document.id = 42
+    assert_raises LockedDocumentConcern::LockedDocumentError, "Cannot perform this operation on a locked document" do
+      document.save
+    end
+  end
 end
--- a/test/unit/edition/validation_test.rb
+++ b/test/unit/edition/validation_test.rb
@@ -161,4 +161,19 @@ class Edition::ValidationTest < ActiveSupport::TestCase
     edition.supporting_organisations = [organisation_1]
     assert edition.valid?
   end
+
+  test 'should raise an exception when attempting to modify an edition of a locked document' do
+    edition = create(:unpublished_edition, :with_locked_document)
+    assert_raises LockedDocumentConcern::LockedDocumentError, "Cannot perform this operation on a locked document" do
+      edition.title = "hello world"
+      edition.save
+    end
+  end
+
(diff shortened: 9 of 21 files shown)


Tetrino:
This commit can't be merged until Emma's change makes it to mainline due to reliance on her locked method in Edition. https://github.com/alphagov/whitehall/pull/4930/

emmabeynon:
Thanks for putting this together Jon.  Could you add a unit test for each of the models, as you have done already for Edition?

test/unit/edition/validation_test.rb
```diff
@@ -161,4 +161,15 @@ class Edition::ValidationTest < ActiveSupport::TestCase
     edition.supporting_organisations = [organisation_1]
     assert edition.valid?
   end
+
+  test 'should raise an exception when attempting to modify an edition of a locked document' do
+    edition = create(:edition)
+    edition.document.locked = true
+    edition.document.save
+
+    edition.title = 'another title'
```
brucebolt: Is there any need to set the title on the edition?  If would seem the edition factory sets a title of `edition-title-n` (where `n` is the index of the edition).

app/models/feature.rb
```diff
@@ -4,6 +4,8 @@ class Feature < ApplicationRecord
   belongs_to :offsite_link
   belongs_to :feature_list
 
+  before_save { raise "Cannot modify a locked document" if document.locked }
```
brucebolt:
```suggestion
  before_save { raise "Cannot feature a locked document" if document.locked }
```

app/models/edition.rb
```diff
@@ -42,6 +42,8 @@ class Edition < ApplicationRecord
   has_many :depended_upon_contacts, through: :edition_dependencies, source: :dependable, source_type: 'Contact'
   has_many :depended_upon_editions, through: :edition_dependencies, source: :dependable, source_type: 'Edition'
 
+  before_save { raise "Cannot modify a locked document" if document.locked }
```
brucebolt:
```suggestion
  before_save { raise "Cannot modify an edition of a locked document" if document.locked }
```

Tetrino:
I don't like how I've had to handle the checks with the new concern, they feel a bit messy. Based on experimentation with Peter we found that we need to make sure we call an instance method to get the correct detail in the before_ calls.

Also don't like what I had to do to the tests after my changes so that they are still valid. Lots of c&p code to flip the document states in the feature, for instance.

app/models/document_collection_group_membership.rb
```diff
@@ -1,17 +1,24 @@
...
 
+  before_validation :check_if_locked_document_internal
+
   validates :document, presence: true
   validates :document_collection_group, presence: true
   validate :document_is_of_allowed_type, if: -> { document.present? }
 
 private
 
+  def check_if_locked_document_internal
```
emmabeynon: Similarly would suggest renaming this.

app/models/edition.rb
```diff
@@ -681,11 +684,15 @@ def has_policy_areas?
   end
 
   def locked?
-    document.locked?
+    document&.locked?
   end
 
 private
 
+  def check_if_locked_document_internal
```
emmabeynon: Similarly would suggest renaming this.

app/models/feature.rb
```diff
@@ -58,6 +61,10 @@ def locale
 
 private
 
+  def check_if_locked_document_internal
```
emmabeynon: Similarly would suggest renaming this.

app/models/feature.rb
```diff
@@ -1,9 +1,12 @@
 class Feature < ApplicationRecord
+  include LockedDocumentConcern
   belongs_to :document
   belongs_to :topical_event
   belongs_to :offsite_link
   belongs_to :feature_list
 
+  before_update :check_if_locked_document_internal
```
emmabeynon: I think `before_save` might be more appropriate here as we don't want a feature to be created that is associated with a locked document.

test/unit/whitehall/search_index_test.rb
```diff
@@ -16,7 +16,10 @@ class ::SearchableClass
     end
 
     test 'SearchIndex.add does not queue a search index add job if a document is locked' do
-      edition = create(:edition, document: build(:document, locked: true))
+      document = build(:document)
+      edition = create(:unpublished_edition, document: document)
```
emmabeynon: Same comment as above.

app/models/document.rb
```diff
@@ -39,6 +40,9 @@ class Document < ApplicationRecord
   has_many :document_collections, through: :document_collection_groups
   has_many :features, inverse_of: :document, dependent: :destroy
 
+  #binding.pry
+  before_update :check_if_locked_document_internal
```
brucebolt: I'm not too keen on the name here (perhaps consider removing the `_internal` part and giving it a unique name)?

test/unit/document_test.rb
```diff
@@ -238,4 +238,16 @@ class DocumentTest < ActiveSupport::TestCase
 
     assert_equal document.patch_meets_user_needs_links, "Links updated"
   end
+
+  test 'should raise an exception when attempting to modify a locked document' do
+    document = create(:document)
+    #Check set to locked does not raise error
+    document.locked = true
+    document.save
+    #Then...
```
brucebolt: I assume this line wasn't meant to be committed?

features/step_definitions/featuring_locked_documents_steps.rb
```diff
@@ -7,7 +7,11 @@
 And(/^the locked document is tagged to organisation "([^"]*)"$/) do |organisation_name|
   organisation = Organisation.find_by(name: organisation_name)
   @edition.organisations = [organisation]
+  @edition.document.locked = false
```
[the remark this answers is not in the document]
kevindew: @emmabeynon that sounds good, but if that seems impractical with the amount of background steps you could reserve the locking for a step in the features too e.g:

```
Background:
  Given a document titled "Quidditch World Cups comes to Hogsmeade"

Scenario: Featuring a locked document on an organisation page
  Given the organisation "Department of Magical Games and Sports" exists
  And the document is tagged to organisation "Department of Magical Games and Sports"
  And the document is locked
  When I visit the organisation admin page for "Department of Magical Games and Sports"
  And I search for a document titled "Quidditch World Cups comes to Hogsmeade" in the list of featurable documents
  Then I cannot see the document in the list of featurable documents
```

test/functional/admin/attachments_controller_test.rb
```diff
@@ -250,7 +250,10 @@ def self.supported_attachable_types
   end
 
   test "POST :create with an edition belonging to a locked document does not save the attachment" do
-    edition = create(:edition, document: build(:document, locked: true))
+    document = build(:document)
+    edition = create(:unpublished_edition, document: document)
```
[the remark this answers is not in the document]
kevindew: I'd suggest it'd be good to pass in a locked one so we could avoid the need to modify the document. For example: `create(:unpublished_edition, document: build(:document, locked: true))` - but do agree it should be avoided if there are no args.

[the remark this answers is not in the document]
Tetrino: My rationale for this is a hiccup with the locked logic. the original create line now throws a locked document error, as the document is created first, saved, then applied to the edition. The edition lock check then fires on the save step of the edition creation.

I also think this is a bad way to do things, but for the purposes of the test it may be the only way. I'll try the one-line approach again to see if there is a workaround.

app/models/edition.rb
```diff
@@ -681,11 +684,15 @@ def has_policy_areas?
   end
 
   def locked?
-    document.locked?
+    document&.locked?
```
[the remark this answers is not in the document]
Tetrino: This breaks several test cases that create editions without documents, I thought it a safe shorthand response to this that didn't involve editing cases to include information they didn't explicitly need.

features/step_definitions/featuring_locked_documents_steps.rb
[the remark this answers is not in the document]
Tetrino: This feels like a more sensible option, as it's more scalable for future tests.

app/models/edition.rb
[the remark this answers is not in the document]
Tetrino: Just putting it down so it's on the record - there is some question over how Whitehall requires documents - especially as the test factory doesn't actually add them without a trait - so for now this is going up as is, and a change should come later.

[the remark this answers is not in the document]
emmabeynon: Is that a change being made in this PR or a separate PR?  I feel a bit uneasy about making a change to code just to make a test pass rather than configuring tests to work with the code.  The tests tell us that a document is required to create an edition: https://github.com/alphagov/whitehall/blob/master/test/unit/edition/validation_test.rb#L21-L22

test/unit/edition/validation_test.rb
```diff
@@ -161,4 +161,12 @@ class Edition::ValidationTest < ActiveSupport::TestCase
     edition.supporting_organisations = [organisation_1]
     assert edition.valid?
   end
+
+  test 'should raise an exception when attempting to modify an edition of a locked document' do
```
emmabeynon: The description of this test indicates that we are testing that an edition belonging to a locked document is being modified, whereas in the test itself we're creating an edition then changing its document's `locked` status to true and then saving it, which isn't quite right.  I think we should create an edition belonging to a locked document and then attempt to modify the edition itself.

test/unit/edition_test.rb
```diff
@@ -936,4 +936,12 @@ class EditionTest < ActiveSupport::TestCase
     assert edition.has_secondary_sectors?
     assert edition.has_legacy_tags?
   end
+
+  test 'should raise an exception when attempting to modify an edition of a locked document' do
```
emmabeynon: The division between this test file and `test/unit/edition/validation_test.rb` is a bit muddled to say the least, but on further investigation I think this file is more about testing methods on Edition rather than validation, so I think we should move this test into the validation test rile and change the description to reflect that an exception is raised when attempting to create a new edition for a locked document.

test/factories/editions.rb
```diff
@@ -161,6 +161,12 @@
         edition.unpublishing = build(:withdrawn_unpublishing, edition: edition)
       end
     end
+
+    trait(:locked) do
```
emmabeynon: I'm a bit unsure about this being a trait for edition because it's the document that's locked, not the edition. @brucebolt do you have any thoughts on this?

app/models/edition.rb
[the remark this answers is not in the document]
Tetrino: I also feel uneasy about this. Intent is to make a separate PR but I've not gotten around to it yet (meetings, issues with local environment). If you'd rather it'd be in this PR I can solve it but it will delay it until tomorrow.

test/factories/editions.rb
brucebolt: I agree with @emmabeynon.  Adding a locked trait to the edition reads like you're created a locked edition, with is not the case.  I think I'd rather the edition were created, then the document locked separately after that.

Tetrino: This was down to @kevindew based on the fact that creating the edition is a two step problem and we really need to save the document lock after the edition is saved, else we cannot save said edition.

Tetrino: > I agree with @emmabeynon. Adding a locked trait to the edition reads like you're created a locked edition, with is not the case. I think I'd rather the edition were created, then the document locked separately after that.

Yeah that was my original intent.

kevindew: Yup, this was my contribution. Tada! 😁 

We could rename this trait to be `create(:edition, :with_locked_document)` which is consistent with: https://github.com/alphagov/whitehall/blob/master/test/factories/editions.rb#L130-L132

Though given we have a method of `edition.locked?` that is already implying we treat the edition as locked rather than the document. So I see `create(:edition, :locked)` to be pretty much analogous to `edition.locked?`.

I don't mind too much though if `:with_locked_document` or something different satisfies concerns. I do think think a trait is preferable to a multi step process in each test though.

kevindew: Incidentally I tried to get rid of the `with_document` trait as that seemed rather unnecessary since documents are created automatically: https://github.com/alphagov/whitehall/commit/972c949742864c817277bd638d43dd47609172ef - however the amount of test failures that would need to be resolved discouraged me from pursuing it further.

app/models/edition.rb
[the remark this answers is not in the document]
kevindew: I looked into this with Jon for a bit. It's a bit of a confusing beast - edition doesn't require a document at a db level (which may just be a mistake) but one is always created as part of the [`before_validation`](https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/models/edition/identifiable.rb#L7) callback and [some](https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/models/edition/identifiable.rb#L8) ([but not all](https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/models/edition/identifiable.rb#L9)) callback methods expect a document to exist.

What I guess makes this situation here a bit confusing is for this to be a problem the `before_validation` callback needs to have not run and the `before_save` one has to have run. I ran a [quick isolated version](https://ci.integration.publishing.service.gov.uk/job/whitehall/job/locked-edition/1/console) and it looks like we only have the one failing test for this: https://github.com/alphagov/whitehall/blob/e34f3c1b263201838394375b600952a9ad9fc56b/test/unit/edition/images_test.rb#L82-L104

I think we could resolve this by putting a document in the [hash of attributes](https://github.com/alphagov/whitehall/blob/e34f3c1b263201838394375b600952a9ad9fc56b/test/unit/edition/images_test.rb#L109-L119) or in the [valid_edition_attributes](https://github.com/alphagov/whitehall/blob/e34f3c1b263201838394375b600952a9ad9fc56b/test/unit/edition/images_test.rb#L142-L149) method.

test/factories/editions.rb
emmabeynon: I think `:with_locked_document` would be clearer to me.  I understand what you mean about having `edition.locked?` already, though it was really to avoid having to repeat `edition.document.locked?` all the time.

Tetrino: Will be in next update.

PR #4945, opened 2019-07-29
Align previous chancellors list

AlanGabbianelli:
This will display them in lines of three.

Trello card: https://trello.com/c/f12zlzdi/1144-update-former-foreign-secretaries-page-content-requests

# Before

Screen Shot 2019-07-29 at 16 07 23

# After

Screen Shot 2019-07-29 at 16 42 50


AlanGabbianelli:
Agreed. We should change the way we define how they are displayed to not be manual.

PR #4947, opened 2019-07-30
Fix broken link and outdated reference to Rummager

sengi:


sengi:
Thanks for the quick review!

PR #4956, opened 2019-08-02
Add more attachment data to document export

Tetrino:
To facilitate the export of documents from Whitehall to Content
Publisher we need as much detail as relevant in our export.

This commit

- Ensures we are exporting all the data relevant to any given
attachment assigned to an edition by adding the export via association
of both the base attachment and its own relevant data object where
applicable (govspeak for HTML, attachment_data chunk for others).

- Includes tests to that affect.

https://trello.com/c/KOiNK5J2/1070-add-more-attachment-data-to-document-export

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -27,6 +27,9 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
       if association == :images
         edition_images = edition.public_send(association)
         output[:associations][association] = complete_images_hash(edition_images)
+      elsif association == :attachments
+        edition_attachments = edition.public_send(association)
+        output[:associations][association] = complete_attachments_hash(edition_attachments)
       else
         output[:associations][association] = edition.public_send(association)
       end
@@ -43,6 +46,14 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   def complete_images_hash(edition_images)
     edition_images.map do |image|
       image.as_json(methods: :url)
+    end
+  end
+
+  def complete_attachments_hash(edition_attachments)
+    edition_attachments.map do |attachment|
+      attachment.as_json(include: :attachment_data, methods: %i[url type]).tap do |json|
+        json.merge!("govspeak_content" => attachment.govspeak_content.as_json) if attachment.respond_to?(:govspeak_content)
+      end
     end
   end
 
--- a/test/unit/presenters/document_export_presenter_test.rb
+++ b/test/unit/presenters/document_export_presenter_test.rb
@@ -49,6 +49,37 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
     assert_equal image.image_data.file_url, images.first['url']
   end
 
+  test "appends expected attachment data to the file attachment response hash" do
+    publication_file = create(:publication, :with_command_paper)
+
+    result = DocumentExportPresenter.new(publication_file.document).as_json
+    attachments = result[:editions].first[:associations][:attachments]
+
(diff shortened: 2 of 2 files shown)


kevindew:
This looks like it'll cover the relevant data for a [`FileAttachment`](https://github.com/alphagov/whitehall/blob/04edd3bc88a715ff2c764e0892e9c4bb4733a979/app/models/file_attachment.rb) subclass of [`Attachment`](https://github.com/alphagov/whitehall/blob/04edd3bc88a715ff2c764e0892e9c4bb4733a979/app/models/attachment.rb) but I don't think it'll pull out enough data for [`HtmlAttachment`](https://github.com/alphagov/whitehall/blob/04edd3bc88a715ff2c764e0892e9c4bb4733a979/app/models/html_attachment.rb) (which is referenced in the previous implementation on the card) and have no idea for [`ExternalAttachment`](https://github.com/alphagov/whitehall/blob/04edd3bc88a715ff2c764e0892e9c4bb4733a979/app/models/external_attachment.rb) - of which the `edition.attachments` association can contain all 3.

I see the card only references data relevant to `FileAttachment`. @brucebolt are we expecting this card to be only file attachment data or all types of attachment?

brucebolt:
@kevindew The intention of this card was for file attachments only.  We have a separate card for HTML attachments: https://trello.com/c/i90aWPC3/1123-include-html-attachment-body-in-whitehall-document-export (this is only a placeholder for now, but will be expanded to include things other than the attachment body).

kevindew:
> @kevindew The intention of this card was for file attachments only. We have a separate card for HTML attachments: https://trello.com/c/i90aWPC3/1123-include-html-attachment-body-in-whitehall-document-export (this is only a placeholder for now, but will be expanded to include things other than the attachment body).

Cool cool - should we be nuking if the attachment class is not a `FileAttachment`?

brucebolt:
Yes, that feels like something that should be done until we complete the work to export HTML attachments

Tetrino:
@kevindew @brucebolt added the check for now, will need to update the HTML card when it comes to writing to mention this thing is here.

brucebolt:
Looks like there's a linting error that's stopping the CI build from completing.  Do you want to fix this before I review?

app/presenters/document_export_presenter.rb
```diff
@@ -46,6 +49,14 @@ def complete_images_hash(edition_images)
     end
   end
 
+  def complete_attachments_hash(edition_attachments)
+    edition_attachments.map do |attachment|
+      if attachment.is_a?(FileAttachment)
+        attachment.as_json(include: :attachment_data, methods: :url)
+      end
```
kevindew: This is going to leave nil values in the array of attachments if items are not file attachments. 

When I suggested nuke I was thinking more like an exception but that feels a bit fishy given we're now unlikely to have the next part of the work done anytime soon, and it shouldn't be too much work to get it working for all 3 attachments.

I think we can get this resolved with:

```
edition_attachments.map do |attachment|
  attachment.as_json(include: :attachment_data, methods: %i[url type]).tap do |json|
     json.merge!(govspeak_content: attachment.govspeak_content.as_json) if attachment.respond_to?(:govspeak_content)
  end
end
```

The type in there is so that it returns "FileAttachment", "ExternalAttachment", or "HtmlAttachment" - govspeak content is an association only HtmlAttachments has.

and then we'd just need tests that cover each attachment type.

kevindew: @Tetrino I see you marked this as resolved, but you may have missed the bit about adding tests for other attachment types.

Tetrino: @kevindew oops, you're right I apologise. Head's not been screwed on.

PR #4962, opened 2019-08-06
Update mysql set up instructions in readme

hannako:
The commands to set up the whitehall user are out of date. The mysql documentation for v 8.0 states that you must first CREATE USER, and then GRANT access.

Ref: https://dev.mysql.com/doc/refman/8.0/en/grant.html

### DIFF
--- a/docs/detailed_setup_guide.md
+++ b/docs/detailed_setup_guide.md
@@ -25,8 +25,9 @@ The database.yml for this project is checked into source control so
 you'll need a local user with credentials that match those in
 database.yml.
 
-```
-mysql> grant all on `whitehall\_%`.* to whitehall@localhost identified by 'whitehall';
+```sql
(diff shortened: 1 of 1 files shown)


injms:
Do you think it's worth mentioning this in the [instructions for setting up the Dev VM](https://docs.publishing.service.gov.uk/manual/install-development-vm.html)? The VM is deprecated, but Whitehall isn't fully docker-ised yet.

docs/detailed_setup_guide.md
```diff
@@ -26,7 +26,8 @@ you'll need a local user with credentials that match those in
 database.yml.
 
 ```
```
thomasleese:
```suggestion
 ```sql
```

Small thing but if you change this you should get some nice syntax highlighting in the block below :smile:

hannako: oh lovely

PR #4963, opened 2019-08-06
Add document type to Whitehall document export

Tetrino:
In order to correctly handle migration we need to ensure that we get
the mapped value for a document type from the API export instead of
just the type ID, as Content Publisher has no knowledge of the
mappings used by Whitehall.

This commit
- Introduces a check to append the required data for corporate
information, speeches, publications and news stories to the API
export for each edition where relevant.
- Introduces a test to this affect
- Fixes an exporter test that was not failing, but was passing
with the wrong comparison (it was comparing two nils, which are
still equal but not what was intended by the test).

https://trello.com/c/Z1yHfuJw/1092-add-document-type-to-whitehall-document-export

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -32,6 +32,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
       end
     end
 
+    provide_doctype_information(edition, output)
     output[:government] = edition.government
     output[:whitehall_admin_links] = resolve_whitehall_admin_links(edition.body)
     if edition.withdrawn?
@@ -43,6 +44,12 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   def complete_images_hash(edition_images)
     edition_images.map do |image|
       image.as_json(methods: :url)
+    end
+  end
+
+  def provide_doctype_information(edition, output)
+    %i[news_article_type publication_type corporate_information_page_type speech_type].each do |type|
+      output[type] = edition.public_send(type)&.key if edition.respond_to?(type)
     end
   end
 
--- a/test/unit/presenters/document_export_presenter_test.rb
+++ b/test/unit/presenters/document_export_presenter_test.rb
@@ -49,9 +49,31 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
     assert_equal image.image_data.file_url, images.first['url']
   end
 
+  test "appends the editions document type key to the response" do
+    news = create(:news_article)
+    publication = create(:publication)
+    speech = create(:speech)
+    corp_info = create(:corporate_information_page)
+
+    news_result = DocumentExportPresenter.new(news.document).as_json
+    doctype_data = news_result[:editions].first[:news_article_type]
+    assert_equal news.news_article_type.key, doctype_data
+
+    pub_result = DocumentExportPresenter.new(publication.document).as_json
+    doctype_data = pub_result[:editions].first[:publication_type]
+    assert_equal publication.publication_type.key, doctype_data
+
+    speech_result = DocumentExportPresenter.new(speech.document).as_json
+    doctype_data = speech_result[:editions].first[:speech_type]
+    assert_equal speech.speech_type.key, doctype_data
+
+    corp_result = DocumentExportPresenter.new(corp_info.document).as_json
+    doctype_data = corp_result[:editions].first[:corporate_information_page_type]
+    assert_equal corp_info.corporate_information_page_type.key, doctype_data
+  end
+
   test "returns government information about an edition" do
     edition = create(:edition)
-
     result = DocumentExportPresenter.new(edition.document).as_json
     assert_equal edition.government, result[:editions].first[:government]
   end
--- a/test/unit/presenters/document_export_presenter_test.rb
+++ b/test/unit/presenters/document_export_presenter_test.rb
@@ -73,8 +73,10 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
   end
 
   test "returns government information about an edition" do
+    current_government = create(:current_government)
     edition = create(:edition)
     result = DocumentExportPresenter.new(edition.document).as_json
     assert_equal edition.government, result[:editions].first[:government]
+    assert_equal current_government, result[:editions].first[:government]
   end
 end


1pretz1:
Hey @Tetrino, looks good just got some small suggestions -

You could potentially dry up the `provide_doctype_information` method with the snippet below and maybe rename it to just doctype_information as the other assignments don't have a verb prefix, lastly I think just returning the key is fine instead of the whole ~Type as a lot of the data in this object is irrelevant. -

```
def doctype_information(edition)
  if edition.respond_to?("#{edition.class.to_s.underscore}_type")
    edition.public_send("#{edition.class.to_s.underscore}_type").key
  end
end
```

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -49,9 +49,32 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
+    doctype_data = speechresult[:editions].first[:document_type_information]
+    assert_equal speech.speech_type.as_json, doctype_data
+
+    corpresult = DocumentExportPresenter.new(corpinfo.document).as_json
+    doctype_data = corpresult[:editions].first[:document_type_information]
+    assert_equal corpinfo.corporate_information_page_type.as_json, doctype_data
+  end
+
   test "returns government information about an edition" do
+    create(:current_government)
```
kevindew: Is this meant to change?

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -49,9 +49,32 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
+    doctype_data = pubresult[:editions].first[:document_type_information]
+    assert_equal publication.publication_type.as_json, doctype_data
+
+    speechresult = DocumentExportPresenter.new(speech.document).as_json
+    doctype_data = speechresult[:editions].first[:document_type_information]
+    assert_equal speech.speech_type.as_json, doctype_data
+
+    corpresult = DocumentExportPresenter.new(corpinfo.document).as_json
+    doctype_data = corpresult[:editions].first[:document_type_information]
+    assert_equal corpinfo.corporate_information_page_type.as_json, doctype_data
```
kevindew: You seem to have stopped separating words with underscores here?

app/presenters/document_export_presenter.rb
```diff
@@ -46,6 +47,13 @@ def complete_images_hash(edition_images)
     end
   end
 
+  def provide_doctype_information(edition, output)
+    output[:document_type_information] = edition.public_send(:news_article_type).as_json if edition.respond_to?(:news_article_type)
+    output[:document_type_information] = edition.public_send(:publication_type).as_json if edition.respond_to?(:publication_type)
+    output[:document_type_information] = edition.public_send(:corporate_information_page_type).as_json if edition.respond_to?(:corporate_information_page_type)
+    output[:document_type_information] = edition.public_send(:speech_type).as_json if edition.respond_to?(:speech_type)
```
kevindew: I think it'd be more intuitive for these to have a key that reflects their name - as these aren't actually document type information (which is a field on document).

So I'd have these set as `:news_article_type`, `publication_type`, etc.

app/presenters/document_export_presenter.rb
```diff
@@ -46,6 +47,13 @@ def complete_images_hash(edition_images)
     end
   end
 
+  def provide_doctype_information(edition, output)
+    output[:document_type_information] = edition.public_send(:news_article_type).as_json if edition.respond_to?(:news_article_type)
```
kevindew: This doesn't need to use `public_send`

test/unit/presenters/document_export_presenter_test.rb
Tetrino: Tired error.

test/unit/presenters/document_export_presenter_test.rb
Tetrino: Yes. It appears creating an edition doesn't create any form of government for it in the backend, and there appears to be no means to do so built into the edition class.

Previously this test was "passing" only because we were comparing nil to nil - no government was assigned, rather than actually comparing the government data. Creating a government at the start of the test in this way allows for the test to run as intended, and matches patterns I've seen elsewhere (typically in a before step).

app/presenters/document_export_presenter.rb
Tetrino: This makes more sense when providing all the type data, however when providing the key only as discussed above then yes I'll be tidying these up.

app/presenters/document_export_presenter.rb
Tetrino: Yet you do above in the example snippet you gave above:

`output[type] = edition.public_send(type).key if edition.respond_to?(type)`

May I ask why not here? Should I be using .send()? I mean this is being replaced anyway but I'll need to learn somehow.

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -49,9 +49,32 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
     assert_equal image.image_data.file_url, images.first['url']
   end
 
+  test "appends detailed document type information about an edition" do
```
kevindew: This should probably be rephrased to reflect that it's sub type rather than document type information, and since it's just a key we wouldn't describe it as detailed.

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -49,9 +49,32 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
+    doctype_data = pub_result[:editions].first[:publication_type]
+    assert_equal publication.publication_type.key.as_json, doctype_data
+
+    speech_result = DocumentExportPresenter.new(speech.document).as_json
+    doctype_data = speech_result[:editions].first[:speech_type]
+    assert_equal speech.speech_type.key.as_json, doctype_data
+
+    corp_result = DocumentExportPresenter.new(corp_info.document).as_json
+    doctype_data = corp_result[:editions].first[:corporate_information_page_type]
+    assert_equal corp_info.corporate_information_page_type.key.as_json, doctype_data
```
kevindew: Note the `as_json` on the keys are redundant since these are strings.

app/presenters/document_export_presenter.rb
kevindew: [#public_send](https://ruby-doc.org/core-2.5.2/Object.html#method-i-public_send) is used to call a particular method on the object. Thus it makes sense to use if we are iterating through a list of methods to call or constructing the method name dynamically.

In your code it was being called without any dynamism which meant it was equivalent to just calling the method. For example: `edition.public_send(:news_article_type).as_json ==  edition.news_article_type.as_json`.

test/unit/presenters/document_export_presenter_test.rb
Tetrino: @kevindew to clarify, a separate commit on this PR? Easy enough.

app/presenters/document_export_presenter.rb
Tetrino: Ah understood, thanks.

test/unit/presenters/document_export_presenter_test.rb
kevindew: @Tetrino yup, exactly that.

app/presenters/document_export_presenter.rb
```diff
@@ -46,6 +47,12 @@ def complete_images_hash(edition_images)
     end
   end
 
+  def provide_doctype_information(edition, output)
+    %i[news_article_type publication_type corporate_information_page_type speech_type].each do |type|
+      output[type] = edition.public_send(type).key if edition.respond_to?(type)
```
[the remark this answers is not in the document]
Tetrino: That's on me, I should be paying closer attention.

test/unit/presenters/document_export_presenter_test.rb
Tetrino: Aye, remnant. Same with the as_json below.

PR #4964, opened 2019-08-07
Display alternative show page for locked documents

emmabeynon:
For https://trello.com/c/QmVdXHrM/1026-disable-edit-in-the-whitehall-ui

We don't want users to be able to edit locked documents, so when they navigate
to a show page for a locked document we display a stripped-back version without
navigation to pages where the document could be edited. Instead we display only
key metadata, a notice explaining that the document has been moved to Content
Publisher, and a link to the document in Content Publisher.

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -2,7 +2,7 @@ class Admin::EditionsController < Admin::BaseController
   before_action :remove_blank_parameters
   before_action :clean_edition_parameters, only: %i[create update]
   before_action :clear_scheduled_publication_if_not_activated, only: %i[create update]
-  before_action :find_edition, only: %i[show edit update submit revise diff reject destroy]
+  before_action :find_edition, only: %i[show show_locked edit update submit revise diff reject destroy]
   before_action :prevent_modification_of_unmodifiable_edition, only: %i[edit update]
   before_action :delete_absent_edition_organisations, only: %i[create update]
   before_action :build_edition, only: %i[new create]
@@ -27,7 +27,7 @@ class Admin::EditionsController < Admin::BaseController
     case action_name
     when 'index', 'topics'
       enforce_permission!(:see, edition_class || Edition)
-    when 'show'
+    when 'show', 'show_locked'
       enforce_permission!(:see, @edition)
     when 'new'
       enforce_permission!(:create, edition_class || Edition)
@@ -74,6 +74,11 @@ class Admin::EditionsController < Admin::BaseController
   end
 
   def show
+    if @edition.locked?
+      redirect_to show_locked_admin_edition_path(@edition)
+      return
+    end
+
     fetch_version_and_remark_trails
 
     @edition_taxons = EditionTaxonsFetcher.new(@edition.content_id).fetch
@@ -82,6 +87,8 @@ class Admin::EditionsController < Admin::BaseController
       @edition_world_taxons = EditionTaxonsFetcher.new(@edition.content_id).fetch_world_taxons
     end
   end
+
+  def show_locked; end
 
   def new; end
 
--- a/app/helpers/admin/content_publisher_routes_helper.rb
+++ b/app/helpers/admin/content_publisher_routes_helper.rb
@@ -1,0 +1,14 @@
+module Admin::ContentPublisherRoutesHelper
+  def content_publisher_document_summary_url(edition)
+    content_id = edition.content_id
+    locale = edition.primary_locale
+
+    "#{content_publisher_base_url}/documents/#{content_id}:#{locale}"
+  end
+
+private
+
+  def content_publisher_base_url
+    Plek.current.external_url_for('content-publisher')
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -309,6 +309,7 @@ Whitehall::Application.routes.draw do
             post :unschedule, to: 'edition_workflow#unschedule'
             post :convert_to_draft, to: 'edition_workflow#convert_to_draft'
             get  :audit_trail, to: 'edition_audit_trail#index'
+            get  :show_locked, to: 'editions#show_locked'
           end
           resources :link_check_reports
           resource :unpublishing, controller: 'edition_unpublishing', only: %i[edit update]
--- a/app/assets/stylesheets/admin/base.scss
+++ b/app/assets/stylesheets/admin/base.scss
@@ -61,6 +61,7 @@
 // they relate to
 @import "views/_dashboard";
 @import "views/_edition_form";
+@import "views/_edition_show_locked";
 @import "views/_edition_show";
 @import "views/_editions-index";
 @import "views/_edition_edit_help";
--- a/app/assets/stylesheets/admin/views/_edition_show_locked.scss
+++ b/app/assets/stylesheets/admin/views/_edition_show_locked.scss
@@ -1,0 +1,10 @@
+.edition-show-locked {
+
+  .jumbotron-header {
+    font-size: 48px;
+  }
+
+  .jumbotron {
+    padding: 40px;
+  }
+}
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -43,7 +43,7 @@ class Admin::BaseController < ApplicationController
 
   def forbid_editing_of_locked_documents
     if @edition.locked?
-      redirect_to [:admin, @edition],
+      redirect_to show_locked_admin_edition_path(@edition),
                   alert: %{This document is locked and cannot be edited}
     end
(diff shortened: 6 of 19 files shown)


app/controllers/admin/editions_controller.rb
```diff
@@ -74,6 +74,10 @@ def confirm_export
   end
 
   def show
+    if @edition.locked?
+      render :locked
+    end
```
kevindew: you probably want to return here as well to avoid executing the rest of the code. I would wonder if you want to have a separate action for this `show_locked` and redirect the show to this to keep them fully distinct.

app/helpers/admin/content_publisher_routes_helper.rb
```diff
@@ -0,0 +1,18 @@
+module Admin::ContentPublisherRoutesHelper
+  def content_publisher_home_url
+    @content_publisher_home_url ||= "#{content_publisher_base_url}/content"
```
kevindew: We don't have a url at `/content` - it doesn't look like this method is used anywhere too.

app/views/admin/editions/_show_document_metadata.html.erb
```diff
@@ -0,0 +1,24 @@
...
+         status_text << "unpublished #{time_ago_in_words(@edition.unpublishing.created_at)} ago" if @edition.unpublishing.present? %>
+      <%= status_text.join(', ') %>
+    </dd>
+    <dt>Change type:</dt>
+    <dd><%= @edition.minor_change? ? 'Minor' : 'Major' %></dd>
+    <% if @edition.respond_to?(:organisations) %>
+      <dt>Organisations:</dt>
+      <dd><%= joined_list(@edition.organisations.map { |o| o.name }) %></dd>
+    <% end %>
+  </dl>
```
kevindew: Sorry, I missed this when the design was going on. 

I don't think we want to show status or change_type since that would reflect data that Content Publisher is now authoritative on. 

I'd suggest rather than making this partial it'd also be slightly cleaner to have this distinct (despite the code duplication) as we'd anticipate that any changes to this metadata for live editions probably is not what locked documents needs so could cause issues. So I think it'd be best to copy and paste the the title, summary, document and organisations parts into the locked view. I figure that should come out at about ~10 lines of code so probably not a painful amount of duplication.

app/views/admin/editions/locked.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <div class="col-md-8">
+    <div class="jumbotron">
+      <h1 class="jumbotron-header">Moved to Content Publisher</h1>
+      <p>This document has been moved from Whitehall to Content Publisher.</p>
+      <p>Information about it in Whitehall may now be out of date.</p>
+      <p>You can now edit this document in Content Publisher.</p>
+      <%= content_publisher_button(@edition) %>
+    </div>
+  </div>
+  <div class="<%= @edition.imported? ? "col-md-6" : "col-md-8" %> edition-metadata">
```
kevindew: I assume we don't care about this imported question in this view?

app/views/admin/editions/locked.html.erb
```diff
@@ -0,0 +1,24 @@
+<% page_title @edition.title, @edition.format_name %>
+<% page_class "edition-show" %>
+<% initialise_script "GOVUK.adminEditionShow" %>
+
+<div class="row" data-module="track-button-click" data-track-category="button-pressed" data-track-action="<%= @edition.format_name.parameterize(separator: '-') %>-button">
```
kevindew: Do we know if this tracking logic still applies? I'm not sure which button it's supposed to track

app/views/admin/editions/locked.html.erb
```diff
@@ -0,0 +1,24 @@
+<% page_title @edition.title, @edition.format_name %>
+<% page_class "edition-show" %>
+<% initialise_script "GOVUK.adminEditionShow" %>
```
kevindew: Looking at the code for this it doesn't seem to be something we'll need: https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/assets/javascripts/admin/views/editions/show.js

app/views/admin/editions/locked.html.erb
```diff
@@ -0,0 +1,24 @@
+<% page_title @edition.title, @edition.format_name %>
+<% page_class "edition-show" %>
```
kevindew: Depending on how much CSS is actually shared with this, it might make sense for it to be a separate scss file just for locked.

app/helpers/admin/content_publisher_routes_helper.rb
emmabeynon: Whoops, forgot to remove this before committing.

app/views/admin/editions/locked.html.erb
emmabeynon: I'm not really sure about what imported means but I figured I should keep it this code as close to the original as possible in case it's important for a future migrated document.

app/views/admin/editions/locked.html.erb
emmabeynon: Similar to above, it wraps around the "show_metadata" partial in the show view (https://github.com/alphagov/whitehall/blob/374ee2ec2ed0964e93243924ff1f76488ba94cf8/app/views/admin/editions/show.html.erb#L5) so I've kept it wrapped around similar metadata for the locked view in case it's still needed

app/views/admin/editions/locked.html.erb
emmabeynon: We were using `.edition-view-edit-buttons` from the edition-show CSS file but it doesn't seem to have any impact if it's removed so will create a new file for locked CSS.

app/controllers/admin/editions_controller.rb
emmabeynon: I've got this to call a `show_locked` method which renders the view instead of creating a separate route for it, does that sound ok?

app/views/admin/editions/_show_document_metadata.html.erb
emmabeynon: I've removed the partial and added only the necessary metadata to the `show_locked` view.

app/controllers/admin/editions_controller.rb
kevindew: Hmm looking at that I think that ends up creating a method that you'd assume is an action.

If we're to go down the non-redirect route I think this would actually be more consistent with existing behaviours to have a before_action handle this conditional.

app/views/admin/editions/locked.html.erb
kevindew: Sure, sounds reasonable.

app/controllers/admin/editions_controller.rb
emmabeynon: I think I've got the redirect working properly now 👍

app/controllers/admin/editions_controller.rb
```diff
@@ -83,6 +87,10 @@ def show
     end
   end
 
+  def show_locked
+    render :show_locked
+  end
```
kevindew: Minor but you can reduce this to `def show_locked; end`

Thinking about it you may not even need to define the method, but since this is the precedent with `new` below we should probably be consistent.

config/routes.rb
```diff
@@ -327,6 +327,7 @@ def external_redirect(path_prefix, target)
         end
 
         get "/editions/:id" => "editions#show"
+        get "/editions-locked/:id" => "editions#show_locked", as: :show_locked_edition
```
kevindew: I think this would be more consistent to be flagged as a member as per: https://github.com/alphagov/whitehall/blob/301072565a9dcdcf2d55c81e664497afc3654880/config/routes.rb#L311

If you defined this as `get :show_locked` you'd then get a route method of `show_locked_admin_edition_path` defined for free.

emmabeynon: Thanks for this - I found the routes a bit of a minefield 😱

app/controllers/admin/editions_controller.rb
emmabeynon: Done 👍

emmabeynon:
@kevindew This should be ready for another review 👍

app/controllers/admin/editions_controller.rb
```diff
@@ -74,6 +74,10 @@ def confirm_export
   end
 
   def show
+    if @edition.locked?
+      redirect_to show_locked_admin_edition_path(@edition)
```
kevindew: I assume this needs a return too.

emmabeynon: Yes - I was having trouble suffixing `&& return` to this line but seems adding a new line with `return` works.

PR #4966, opened 2019-08-07
Indicate locked documents on index page

emmabeynon:
For https://trello.com/c/QmVdXHrM/1026-disable-edit-in-the-whitehall-ui

We add a label to locked documents on the document index page to indicate that
it has been moved to Content Publisher. Once the document has been moved it
won't make sense to display the "Updated by", "Updated" and "State" values as
it won't be in sync with what is in Content Publisher, so we replace these with
another message about having been moved to Content Publisher.

Screen Shot 2019-08-16 at 11 40 58

### DIFF
--- a/app/views/admin/editions/_search_results.html.erb
+++ b/app/views/admin/editions/_search_results.html.erb
@@ -66,6 +66,9 @@
             <% if edition.access_limited? %>
               <span class="access_limited label label-danger">limited access</span>
             <% end %>
+            <% if edition.locked? %>
+              <span class="label label-info">Moved to Content Publisher</span>
+            <% end %>
             <% if edition.link_check_reports.any? && edition.link_check_reports.last.completed? %>
               <% if edition.link_check_reports.last.broken_links.any? %>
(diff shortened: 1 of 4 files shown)


app/views/admin/editions/_search_results.html.erb
```diff
@@ -74,10 +77,14 @@
               <% end %>
             <% end %>
           </td>
-          <td class="author"><%= linked_author(edition.last_author) %></td>
-          <td class="updated"><span title="<%= edition.updated_at %>"><%= time_ago_in_words edition.updated_at %> ago</span></td>
-          <% if viewing_all_active_editions? %>
-            <td class="state"><%= admin_edition_state_text(edition) %></td>
+          <% if edition.locked? %>
+            <td colspan="3">Moved to Content Publisher</td>
```
kevindew: I'd suggest doing this with `text-align: center` (assuming that's cool with Sonia)

emmabeynon: Sonia doesn't mind either way so will change.

PR #4967, opened 2019-08-08
[WIP] Support non-Whitehall links in document collections

kevindew:
Trello: https://trello.com/c/VdQtNZeV/1122-spike-making-content-publisher-documents-show-up-in-whitehall-document-collections

This is rather rapid attempt at trying to make the document collections feature in Whitehall support links to GOV.UK content that does not originate from Whitehall. Ideally this whole feature should be powered by Publishing API data so that content from all publishing applications is treated equally.

To do this a `DocumentCollectionGroupMembership` has been changed from requiring an association to a `Document` to supporting an association to either a `Document` or a `DocumentCollectionNonWhitehallLink`. A `DocumentCollectionNonWhitehallLink` contains data that can be populated from the Publishing API: `base_path`, `title`, `publishing_app` and `content_id`. `content_id` is the only value that is needed for presenting the `DocumentCollection` to the Publishing API (and this code caters for this) the other fields are intended for the user interface. There is an expected risk that the data in these fields may fall out of date.

I envision this as eventually allowing a user to enter a GOV.UK URL in a progressively disclosed form to add to the group. For example:

Screen Shot 2019-08-08 at 10 06 30

The next steps I see for this are:

## 1. Allow showing, moving and removing a non-Whitehall link in the UI for a document collection.

This focuses on altering the rendering of the document collection view to show memberships of groups rather than documents. A view will need to be created to allow showing a non-Whitehall link. The form parameters will need to be altered from working with [documents](https://github.com/alphagov/whitehall/blob/374ee2ec2ed0964e93243924ff1f76488ba94cf8/app/views/admin/document_collection_groups/_collection_document.html.erb) to working with memberships so that [set_document_ids_in_order](https://github.com/alphagov/whitehall/blob/374ee2ec2ed0964e93243924ff1f76488ba94cf8/app/controllers/admin/document_collection_groups_controller.rb#L43) can be replaced with [set_membership_ids_in_order](https://github.com/alphagov/whitehall/compare/document-collections?expand=1#diff-f5360a4fd8e851e1448380058b2dae6eR32). 

I anticipate the view looking something like this: 

Screen Shot 2019-08-08 at 10 01 43

Note this does not include adding the document to the collection which is a bigger task.

## 2. Provide means of a Rake task to create these 

I'd anticipate there being a rake task that would allow adding items to a group as a faster means of allowing this feature to ship, meaning this could be done as a developer task initially until we have availability to build the UI.

I'd expect the task to be something like: `document_collection:add_non_whitehall_link_to_group[<content_id>, <document_collection_group_id>]`.

The task should look up the `DocumentCollectionGroup` for the id and check whether that is part of an editable document collection, if not abort.

It should then look up the [content item in the Publishing API](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2content) for the content_id and, assuming one exists, create a `DocumentCollectionNonWhitehallLink` with the data and save that to the `DocumentCollectionGroup`.

## 3.Allow ability to add to collection via Whitehall UI

This would allow a progressively disclosed form that allows adding a GOV.UK URL to a document collection group.

This would need means to validate the URL. It should determine the user input is a valid URL and that this a GOV.UK URL. Assuming those checks pass it should extract the path of the URL and use [Publishing API lookup-by-base-path](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#post-lookup-by-base-path) to get the content_id and [Publishing API get content](https://github.com/alphagov/publishing-api/blob/master/doc/api.md#get-v2content) to look up the data. This would then be used to create the `DocumentCollectionNonWhitehallLink`.

What is unclear to me at this point is how we would render validate errors to the user and what class/model would be used to perform the validation.

/cc @Tetrino @JonathanHallam

### DIFF
--- a/app/assets/javascripts/admin/document_group_ordering.js
+++ b/app/assets/javascripts/admin/document_group_ordering.js
@@ -49,7 +49,7 @@
     for(var i=0; i<this.documentGroups.length; i++) {
       postData.groups.push({
         id: this.documentGroups[i].groupID(),
-        document_ids: this.documentGroups[i].documentIDs(),
+        membership_ids: this.documentGroups[i].membershipIDs(),
         order: i
       });
     }
@@ -84,8 +84,8 @@
       return documentList.data('group-id');
     };
 
-    this.documentIDs = function documentIDs() {
-      return documentList.find("input[name='documents[]']").map(function(i, input) {
+    this.membershipIDs = function membershipIDs() {
+      return documentList.find("input[name='memberships[]']").map(function(i, input) {
         return input.value;
       }).toArray();
     };
--- a/app/controllers/admin/document_collection_group_memberships_controller.rb
+++ b/app/controllers/admin/document_collection_group_memberships_controller.rb
@@ -15,12 +15,11 @@ class Admin::DocumentCollectionGroupMembershipsController < Admin::BaseControlle
   end
 
   def destroy
-    document_ids = params.fetch(:documents, []).map(&:to_i)
-    if document_ids.present?
-      delete_from_old_group(document_ids)
-      move_to_new_group(document_ids) if moving?
+    membership_ids = params.fetch(:memberships, []).map(&:to_i)
+    if membership_ids.present?
+      moving? ? move_to_new_group(membership_ids) : delete_from_old_group(membership_ids)
       redirect_to admin_document_collection_groups_path(@collection),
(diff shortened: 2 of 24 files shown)


kevindew:
Superseded by: https://github.com/alphagov/whitehall/pull/4980

PR #4969, opened 2019-08-08
Change Organisation Type to "Tribunal"

AlanGabbianelli:
Not all organisations that are of type 'Tribunal non-departmental public
body' are actually non-departmental. Some of them are in fact
departmental. Changing this type to be just 'Tribunal' will reflect this
more accurately.

Trello card: https://trello.com/c/KHXlOF6V/1191-change-organisation-type-tribunal-non-departmental-public-body-to-tribunal

### DIFF
--- a/app/models/organisation_type.rb
+++ b/app/models/organisation_type.rb
@@ -6,7 +6,7 @@ class OrganisationType
     executive_agency:            { name: "Executive agency",                       analytics_prefix: "EA", agency_or_public_body: true,  non_departmental_public_body: false, allowed_promotional: false },
     executive_ndpb:              { name: "Executive non-departmental public body", analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: true, allowed_promotional: false },
     advisory_ndpb:               { name: "Advisory non-departmental public body",  analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: true, allowed_promotional: false },
-    tribunal_ndpb:               { name: "Tribunal non-departmental public body",  analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: true, allowed_promotional: false },
+    tribunal:                    { name: "Tribunal",                               analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: false, allowed_promotional: false },
     public_corporation:          { name: "Public corporation",                     analytics_prefix: "PC", agency_or_public_body: false, non_departmental_public_body: false, allowed_promotional: false },
     independent_monitoring_body: { name: "Independent monitoring body",            analytics_prefix: "IM", agency_or_public_body: true,  non_departmental_public_body: false, allowed_promotional: false },
     adhoc_advisory_group:        { name: "Ad-hoc advisory group",                  analytics_prefix: "AG", agency_or_public_body: true,  non_departmental_public_body: false, allowed_promotional: false },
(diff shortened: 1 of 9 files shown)


app/models/organisation_type.rb
```diff
@@ -6,7 +6,7 @@ class OrganisationType
     executive_agency:            { name: "Executive agency",                       analytics_prefix: "EA", agency_or_public_body: true,  non_departmental_public_body: false, allowed_promotional: false },
     executive_ndpb:              { name: "Executive non-departmental public body", analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: true, allowed_promotional: false },
     advisory_ndpb:               { name: "Advisory non-departmental public body",  analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: true, allowed_promotional: false },
-    tribunal_ndpb:               { name: "Tribunal non-departmental public body",  analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: true, allowed_promotional: false },
+    tribunal:                    { name: "Tribunal",                               analytics_prefix: "PB", agency_or_public_body: true,  non_departmental_public_body: false, allowed_promotional: false },
```
cbaines: What are the implications of changing the `non_departmental_public_body` value to `false` here?

MuriloDalRi: I'm not sure. Do you reckon it would break anything?

MuriloDalRi:
This now fails:
`You have 1 pending migration:
  20190813142450 ChangeTribunalOrgType`

cbaines:
> This now fails:
> `You have 1 pending migration: 20190813142450 ChangeTribunalOrgType`

Probably because the `schema.rb` file hasn't been updated. You'll need to run the migration locally, and commit the change to the version in the `schema.rb` file.

app/models/organisation_type.rb
cbaines: I think it's worth investigating further at least, as it would be good to avoid breaking anything (and that seems possible with this kind of change).

MuriloDalRi: The only place I found that would be affected by this change is this: https://github.com/alphagov/whitehall/blob/b54f7e2524d42679cda986fa8b552bca5aa3c1dd/app/views/admin/organisations/_form.html.erb#L123

Now I'm thinking maybe we should leave it as true?

cbaines: > Now I'm thinking maybe we should leave it as true?

Good investigating :+1: . That seems reasonable to me, otherwise those form elements would disappear from some organisations to which I guess they could be relevant.

db/schema.rb
```diff
@@ -1006,6 +1006,15 @@
...
 
+  create_table "sync_check_results", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+    t.string "check_class"
+    t.integer "item_id"
+    t.text "failures"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["check_class", "item_id"], name: "index_sync_check_results_on_check_class_and_item_id", unique: true
+  end
+
```
cbaines: This looks like a spurious change. I'd suggest removing it, unless you're sure this change should be made to the schema.rb file.

MuriloDalRi: Ah yeah for some reason there was another migration that ran, I've removed it now

PR #4973, opened 2019-08-09
Use membership ID's 

JonathanHallam:
Part of https://trello.com/c/VdQtNZeV/1122-spike-making-content-publisher-documents-show-up-in-whitehall-document-collections

Moves to using membership ID's rather than document ID's as a means of tracking the group of a document. This will make it easier to show content publisher documents in whitehall. 

Whilst this appears to work testing manually, as it stands a few tests are failing, I suspect that a couple just need some variables renaming but that one may be a more major change.


kevindew:
Superseded by https://github.com/alphagov/whitehall/pull/4980

PR #4975, opened 2019-08-12
Migration to change tribunals org type

MuriloDalRi:
**What**

On the Organisations page on Signon, any organisation that has a tribunal type is labelled 'tribunal non-departmental public body'.

However, many organisations are tribunal bodies but are departmental public bodies (such as HMCTS). There is only one type tag available (non-DPB), so we should rename this type as 'Tribunal' to cover both types of organisations.

**Why**

Not all organisations fit into the current type tagging. The current tag is very specific - we should make it easier for publishers by changing the type name to a more generic 'Tribunal'.

[Trello](https://trello.com/c/KHXlOF6V/1191-change-organisation-type-tribunal-non-departmental-public-body-to-tribunal)

### DIFF
--- a/db/data_migration/20190812093415_change_tribunal_org_type.rb
+++ b/db/data_migration/20190812093415_change_tribunal_org_type.rb
@@ -1,0 +1,9 @@
+orgs = Organisation.where(organisation_type_key: :tribunal_ndpb)
+
+orgs.each do |org|
+  puts "Changing #{org.name} old org type #{org.organisation_type_key}"
+
+  org.update_attributes!(organisation_type_key: :tribunal)
+
+  puts "#{org.name} new org type: #{org.organisation_type_key}"
+end


MuriloDalRi:
> The commit message is a bit short, and doesn't mention why the change is being made. Could you expand it a bit. Maybe just work in some of the information you've put in the Pull Request description.

Good point, have updated it now!

db/data_migration/20190812093415_change_tribunal_org_type.rb
```diff
@@ -0,0 +1,9 @@
+orgs = Organisation.where(organisation_type_key: :tribunal_ndpb)
+
+orgs.each do |org|
+  puts "Changing #{org.name} old org type #{org.organisation_type_key}"
+
+  org.update_attributes!(organisation_type_key: :tribunal)
```
cbaines: `tribunal` isn't a value listed here https://github.com/alphagov/whitehall/blob/master/app/models/organisation_type.rb#L2-L18

Does that matter?

cbaines:
> Good point, have updated it now!

Great, thanks.

I noticed from the Trello card that there was already a Pull Request open relating to seemingly the same issue: https://github.com/alphagov/whitehall/pull/4969 How does the change here fit in with those proposed changes?

db/data_migration/20190812093415_change_tribunal_org_type.rb
MuriloDalRi: There's another PR https://github.com/alphagov/whitehall/pull/4969/files

MuriloDalRi:
> > Good point, have updated it now!
> 
> Great, thanks.
> 
> I noticed from the Trello card that there was already a Pull Request open relating to seemingly the same issue: #4969 How does the change here fit in with those proposed changes?

So I think that PR needs to go in first and then we do this migration

cbaines:
> So I think that PR needs to go in first and then we do this migration

It looks to me like the Pull Requests are mutually dependent, you wouldn't want to apply either commit without the other?

If that is the case, it would be better if this was changed to a normal migration (rather than a data migration, as it's effectively the schema that's being changed), and merged in to the commit in the other Pull Request.

MuriloDalRi:
> > So I think that PR needs to go in first and then we do this migration
> 
> It looks to me like the Pull Requests are mutually dependent, you wouldn't want to apply either commit without the other?
> 
> If that is the case, it would be better if this was changed to a normal migration (rather than a data migration, as it's effectively the schema that's being changed), and merged in to the commit in the other Pull Request.

I'll close this PR and add a migration to the other one

PR #4976, opened 2019-08-12
Fix current role h2 on people pages

sihugh:
The styles on this look a bit broken at present.  Adding "gem-c-heading" as per the other h2's on the page make it look a bit better. This is on people pages like https://www.gov.uk/government/people/michael-gove#current-roles

## Before
Screenshot 2019-08-12 13 15 42

## After
Screenshot 2019-08-12 13 15 31

### DIFF
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -68,7 +68,7 @@
       <div class="current-roles" id="current-roles">
         <% @person.current_role_appointments.each do |appointment| %>
           <%= content_tag_for :section, appointment, class: "role" do %>
-            <h2 id="<%= appointment.role.name.parameterize%>"><%= appointment.role.name %></h2>
+            <h2 class="gem-c-heading" id="<%= appointment.role.name.parameterize%>"><%= appointment.role.name %></h2>
             <%= appointment.role.responsibilities %>
             <div class="read-more">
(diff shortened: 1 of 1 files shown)


sihugh:
Yeah, I know that @andysellick but it's been used in this file in a few places already and it looks like the accessibility team are all over it...

sihugh:
I think it's been used outside the component because locale isn't supported as a parameter on the heading component yet. The other uses of this class have calls to the [t_lang helper](https://github.com/alphagov/whitehall/blob/e5a44fc993d1f05e92a5acc96f136bb188793f03/app/helpers/translation_helper.rb#L36-L41) which inserts `lang=cy` or equivalent into the tag.

chao-xian:
Agree that it should be using the helper but fixing consistency as an interim step is ok for now.

injms:
What Kelvin said.

The page headings with the `t_lang` helper detect what language any string requested from the internationalisation files. 

Because the content of this heading ('Chancellor of the Duchy of Lancaster') comes from the page content, rather than the internationalisation files, we can't detect what language it is (to a degree that we were happy with). 

So it could be a heading, but it probably won't matter that it's not since adding the `lang` option for components is on the long list of things to do - then all of the headings could be components on this page.

PR #4978, opened 2019-08-13
Update docs for retired finders

sihugh:
We've retired some finders but left some code behind. We should explain why.

We should also deprecate publishing finders from Whitehall because there is
a better place to do this.

https://trello.com/c/BE2ooOcY/934-redirect-whitehall-statistics-finder-to-finder-frontend

### DIFF
--- a/docs/finders.md
+++ b/docs/finders.md
@@ -1,4 +1,27 @@
-# How to publish a finder in whitehall
+# Finders
+
+## Retired finders
+
+There used to be finders for announcements, publications and statistics in this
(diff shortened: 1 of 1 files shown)


docs/finders.md
```diff
@@ -1,4 +1,27 @@
-# How to publish a finder in whitehall
+#Finders
```
bilbof:
```suggestion
# Finders
```

sihugh: Thanks @bilbof 👍

PR #4979, opened 2019-08-13
Lazy load profile images for past prime ministers

alex-ju:
### What
This PR updates the `person` presenter to lazy load profile images.

### Why
Chrome shipped lazy loading support in version 76. The 'Past Prime Ministers' page is a good use case for this feature having 76 profile images that could be saved from being downloaded straight away. This attribute doesn't harm any browser and improves experience for users of Chrome 76 and further. The standard is awaiting implementation from other major players so this change will hopefully have more reach in future.

### DIFF
--- a/app/presenters/person_presenter.rb
+++ b/app/presenters/person_presenter.rb
@@ -80,7 +80,7 @@ class PersonPresenter < Whitehall::Decorators::Decorator
 
   def image
     if (img = image_url(:s216))
-      context.image_tag img, alt: name
+      context.image_tag img, alt: name, loading: "lazy"
     end
   end
 end


Nooshu:
Excellent work Alex. This simple change makes quite a difference to the user-centric performance metrics for this page due to the large number of images:

| Metric                 | Without LL (ms) | With LL (ms) | diff (%) |
|------------------------|-----------------|--------------|----------|
| First Contentful Paint | 3271            | 2520         | -22      |
| First Meaningful Paint | 3393            | 2578         | -24      |
| Time to Interactive    | 3325            | 2840         | -15      |
| Speed Index            | 3564            | 3240         | -9       |
| Fully Loaded           | 7280            | 5821         | -20      |

[WPT comparison](https://www.webpagetest.org/video/compare.php?tests=190820_2Q_d9c0b6d6543174427b062ef38bf06c6f%2C190820_EE_777c483684b175f06ec4d929e9108c22&thumbSize=200&ival=100&end=visual) on a 3G Fast connection, Chrome version 78.0.3880.4.

PR #4980, opened 2019-08-14
Support non-Whitehall links in document collections

kevindew:
Trello: https://trello.com/c/VdQtNZeV/1122-spike-making-content-publisher-documents-show-up-in-whitehall-document-collections

This adds the ability to add content created in other publishing applications to a Whitehall document collection. Which resolves a problem Content Publisher publishers have faced. The solution to this is a rather quick and dirty one that allows a document collection group to have a membership with either a Whitehall document or a GOV.UK content_id, had more time been available it'd have been good to replace the document collection approach in Whitehall to be one powered entirely by the Publishing API.

The user interface for this is like so:

Aug-14-2019 12-12-27

And will appear on GOV.UK like this:

Screen Shot 2019-08-14 at 12 09 58

More details in the individual commits.

Despite the commit history this also represents works by @1pretz1, @Tetrino and @JonathanHallam

### DIFF
--- a/test/functional/admin/document_collection_group_memberships_controller_test.rb
+++ b/test/functional/admin/document_collection_group_memberships_controller_test.rb
@@ -40,11 +40,12 @@ class Admin::DocumentCollectionGroupMembershipsControllerTest < ActionController
     id_params.merge(commit: 'Move')
   end
 
-  view_test 'DELETE #destroy removes documents and redirects when Remove clicked' do
-    documents = [create(:publication), create(:publication)].map(&:document)
-    @group.documents << documents
-    assert_difference '@group.reload.documents.size', -1 do
-      delete :destroy, params: remove_params.merge(documents: [documents.first.id])
+  view_test 'DELETE #destroy removes memberships and redirects when Remove clicked' do
+    memberships = [create(:document_collection_group_membership),
+                   create(:document_collection_group_membership)]
+    @group.memberships << memberships
+    assert_difference '@group.reload.memberships.size', -1 do
+      delete :destroy, params: remove_params.merge(memberships: [memberships.first.id])
     end
     assert_redirected_to admin_document_collection_groups_path(@collection)
     assert_match %r[1 document removed], flash[:notice]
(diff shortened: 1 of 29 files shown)


test/functional/admin/document_collection_group_memberships_controller_test.rb
```diff
@@ -56,14 +57,15 @@ def move_params
   end
 
   test 'DELETE #destroy moves documents and redirects when Move clicked' do
```
emmabeynon: Should we change references to 'documents' to 'memberships'?

test/functional/admin/document_collection_groups_controller_test.rb
```diff
@@ -34,14 +34,14 @@ class Admin::DocumentCollectionGroupsControllerTest < ActionController::TestCase
   end
 
   view_test 'GET #index shows a warning if > 50 documents in collection' do
```
emmabeynon: Should we change references to 'documents' to 'memberships'?

test/unit/models/document_collection_group_test.rb
```diff
@@ -74,4 +74,38 @@ class DocumentSeriesGroupTest < ActiveSupport::TestCase
...
+    non_whitehall_link = build(:document_collection_non_whitehall_link)
+    group = create(:document_collection_group, memberships: [
+      build(:document_collection_group_membership, document: document),
+      build(:document_collection_group_membership,
+            document: nil,
+            non_whitehall_link:
+            non_whitehall_link),
+    ])
+
+    group.memberships[0].update(ordering: 2)
```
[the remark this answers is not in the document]
kevindew: Unfortunately not, the model overwrites it: https://github.com/alphagov/whitehall/blob/296109ffb4d2d5ec26ac941c5649fe214103ce9d/app/models/document_collection_group.rb#L18

test/functional/admin/document_collection_group_memberships_controller_test.rb
kevindew: Good point 👍

kevindew:
Thanks @emmabeynon all comments should be addressed, just 🤞 for CI now.

PR #4988, opened 2019-08-16
Revert "Change Organisation Type to "Tribunal""

MuriloDalRi:
Reverts alphagov/whitehall#4969

### DIFF
--- a/db/migrate/20190813142450_change_tribunal_org_type.rb
+++ b/db/migrate/20190813142450_change_tribunal_org_type.rb
@@ -1,7 +1,0 @@
-class ChangeTribunalOrgType < ActiveRecord::Migration[5.1]
-  def change
-    Organisation
-      .where(organisation_type_key: :tribunal_ndpb)
-      .update_all(organisation_type_key: :tribunal)
-  end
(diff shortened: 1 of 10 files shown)


db/migrate/20190813142450_change_tribunal_org_type.rb
```diff
@@ -1,7 +0,0 @@
-class ChangeTribunalOrgType < ActiveRecord::Migration[5.1]
```
sihugh: This has already run on integration and staging.  It might be OK as long as the data sync from production runs and overwrites those databases, though it might be better to write another migration that reverses this one. That would ensure that the data is in sync with the reverted code.

MuriloDalRi: Makes sense, I'll do that

MuriloDalRi: @sihugh does it look better now?

sihugh: Thanks, that's great 👍

PR #4992, opened 2019-08-20
[WIP]Add tribunal org type

MuriloDalRi:
Not all organisations that are of type 'Tribunal non-departmental public body' are actually non-departmental. Some of them are in fact
departmental. Changing this type to be just 'Tribunal' will reflect this
more accurately.

Trello card:
https://trello.com/c/KHXlOF6V/1191-change-organisation-type-tribunal-non-departmental-public-body-to-tribunal

### DIFF
--- a/app/models/organisation/organisation_type_concern.rb
+++ b/app/models/organisation/organisation_type_concern.rb
@@ -31,8 +31,8 @@ module Organisation::OrganisationTypeConcern
     scope :hmcts_tribunals, -> {
       hmcts_id = Organisation.unscoped.where(slug: "hm-courts-and-tribunals-service").ids.first
       joins(:parent_organisational_relationships).
-        where(organisation_type_key: :tribunal_ndpb).
-        where("organisational_relationships.parent_organisation_id" => hmcts_id)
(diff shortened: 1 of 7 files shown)


app/models/organisation/organisation_type_concern.rb
```diff
@@ -31,8 +31,10 @@ module Organisation::OrganisationTypeConcern
     scope :hmcts_tribunals, -> {
       hmcts_id = Organisation.unscoped.where(slug: "hm-courts-and-tribunals-service").ids.first
       joins(:parent_organisational_relationships).
-        where(organisation_type_key: :tribunal_ndpb).
-        where("organisational_relationships.parent_organisation_id" => hmcts_id)
+      where("organisations.organisation_type_key = ? OR
```
thomasleese: I think you might be able to do something like this to simplify the query.

```suggestion
      where(organisations: { organisation_type_key: %w(tribunal_ndpb tribunal) })
```

MuriloDalRi: I'll see if that works but it doesn't matter much as I'll delete all references to `tribunal_ndpb` after merging all my PRs and migrating orgs

PR #4996, opened 2019-08-22
Redirect email signup journeys to email-alert-frontend

1pretz1:
Do not merge until https://github.com/alphagov/email-alert-frontend/pull/546 is deployed.

This PR changes where users get taken when they click on a email-signup link in whitehall-frontend. 

The signup-links on the following pages have been changed to user email-alert-frontend: 
* people
* ministers
* latest pages (organisations, topical events and world locations)
* topical_events
* international delegations
* finders (although all these have been migrated to finder frontend,
unless it's a translated finder in which case emails don't work anyway)
* topics (actually policy areas but have already been retired but dead code everywhere)
* organisations (migrated to collections)

Redirects have also been put in place for all signup paths, see commits for further information.

Trello:
https://trello.com/c/tgWUR1dx/25-switch-whitehalls-email-signup-pages-to-use-email-alert-frontends


app/controllers/email_signups_controller.rb
```diff
@@ -1,32 +1,38 @@
...
-      redirect_to @email_signup.signup_url
-    else
-      render action: 'new'
-    end
+  def non_finder_url
+    feed_url.match?(%r{/government/organisations/(.*)\.atom$}) ||
+      feed_url.match?(%r{/government/ministers/(.*)\.atom$}) ||
+      feed_url.match?(%r{/government/people/(.*)\.atom$}) ||
+      feed_url.match?(%r{/government/topical-events/(.*)\.atom$}) ||
+      feed_url.match?(%r{/world/(.*)\.atom$})
```
kevindew: We can probably reduce to two conditionals: `.match?(%r{/government/(organisations|ministers|people|topical-events)/.*\.atom$})`

app/controllers/email_signups_controller.rb
```diff
@@ -1,32 +1,38 @@
...
-    if feed_url.match?(%r{/government/organisations/(.*)\.atom$})
-      redirect_to email_alert_frontend_signup(feed_url)
+    if non_finder_url
+      redirect_to email_alert_frontend_signup
+    elsif feed_url.match?(%r{/government/publications\.atom(.*)$})
+      redirect_to feed_url.sub('.atom', '')
+    elsif feed_url.match?(%r{/government/announcements\.atom(.*)$})
+      redirect_to feed_url.sub('.atom', '')
+    elsif feed_url.match?(%r{/government/statistics\.atom(.*)$})
+      redirect_to feed_url.sub('.atom', '')
```
kevindew: We can probably reduce these 3 conditionals to a single one: `%r{/government/(publications|announcements|statistics)\.atom}` - we shouldn't need to worry about after the atom extension.

app/controllers/email_signups_controller.rb
```diff
@@ -1,32 +1,38 @@
...
+      feed_url.match?(%r{/government/ministers/(.*)\.atom$}) ||
+      feed_url.match?(%r{/government/people/(.*)\.atom$}) ||
+      feed_url.match?(%r{/government/topical-events/(.*)\.atom$}) ||
+      feed_url.match?(%r{/world/(.*)\.atom$})
   end
 
-private
-
   def email_signup_params
     params.require(:email_signup).permit(:feed)
```
kevindew: might want to require feed, eg `params.require(:email_signup).require(:feed)`. Then we can probably avoid the feed_url method.

app/controllers/email_signups_controller.rb
1pretz1: Yup this is a good shout, will remove some clutter!

app/controllers/email_signups_controller.rb
1pretz1: Yup good point no2

app/controllers/email_signups_controller.rb
1pretz1: Sure

PR #4997, opened 2019-08-22
Allow a single HTML Attachment to be redirected

emmabeynon:
This adapts the PublishingApiHtmlRedirector to accept a single HtmlAttachment
and redirect it. This is useful in cases where an HtmlAttachment has become
orphaned from an edition and so cannot be redirected using the rake task in its
current form.

### DIFF
--- a/lib/data_hygiene/publishing_api_html_attachment_redirector.rb
+++ b/lib/data_hygiene/publishing_api_html_attachment_redirector.rb
@@ -7,7 +7,9 @@ module DataHygiene
     end
 
     def call
-      raise DataHygiene::EditionNotUnpublished.new unless unpublished_or_withdrawn
+      if document
+        raise DataHygiene::EditionNotUnpublished.new unless unpublished_or_withdrawn
+      end
       raise DataHygiene::HTMLAttachmentsNotFound.new unless html_attachments.any?
       return dry_run_results if dry_run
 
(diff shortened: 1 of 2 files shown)


lib/data_hygiene/publishing_api_html_attachment_redirector.rb
```diff
@@ -37,7 +39,11 @@ def unpublished_or_withdrawn
     end
 
     def html_attachments
-      @html_attachments ||= last_edition.html_attachments
+      @html_attachments ||= document ? last_edition.html_attachments : [html_attachment]
+    end
+
+    def html_attachment
+      @html_attachment ||= HtmlAttachment.find_by(content_id: content_id)
```
cbaines: It seems a bit awkward to me to try and either use the content_id to find the document, or the HTML attachment.

Does this worker change the state of the HTML attachment in the Whitehall database?

emmabeynon: No it doesn't change the state.  It just calls the Publishing API.  I started writing a separate rake task for individual HTML Attachments but the code was so similar to this that it made sense to me to just include it here.  Is there a particular reason why we shouldn't use the content id to find an HTML Attachment?  We do have precedent in the app for checking if a content id is for a Document or HTML Attachment https://github.com/alphagov/whitehall/blob/master/app/controllers/admin/documents_controller.rb#L2

cbaines: Only that it seems awkward to use the content_id to either find a document, or a HTML attachment.

Saying that, I was thinking this was part of the app, but it's actually just used in a Rake task right? In that case, I don't think it matters as much.

PR #4998, opened 2019-08-22
[WIP]Update Rails from 5.1.7 to 5.2.4.1

AlanGabbianelli:
This updates both gems and configs (by running `rails app:update`).

Trello card: 
https://trello.com/c/m0r07ZNR/1216-upgrade-whitehall-to-rails-6


kevindew:
@laurentqro I've seen from alerts that you're doing a fair bit of wrestling to get libsass support rather than ruby sass (which is awesome that you're in the fight, as we need to sort this one day).

I believe all other GOV.UK Rails apps are using ruby-sass still and once you install the sass linter you basically have a hard time switching to libsass. You may find that for Whitehall it's easier to keep this with Ruby Sass and have this be upgraded to libsass as part of a wider piece of GOV.UK work.

There was a thread on the problems in Slack yesterday: https://gds.slack.com/archives/CARGH33JS/p1582626595005800

laurentqro:
Thanks for the info @kevindew, very helpful.

I've gone back to this for now:

```
  # Rather than use a CSS compressor, use the SASS style to perform compression
  config.sass.style = :compressed	  config.sass.style = :compressed
```

Focusing on getting the Rails upgrade across the finish line, and we can tackle libsass later.

test/parallel_test_runner.rb
```diff
@@ -1,7 +1,5 @@
 class ParallelTestRunner < TestQueue::Runner::MiniTest
   def after_fork(number)
-    super
```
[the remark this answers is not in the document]
laurentqro: I upgraded the test-queue gem to the latest version and then current documentation does not show the necessity of this `super`, so I removed it. https://github.com/tmm1/test-queue#usage

test/unit/data_hygiene/organisation_reslugger_test.rb
```diff
@@ -43,7 +43,7 @@ def teardown
 
       stub_any_publishing_api_call
       Sidekiq::Testing.inline! do
-        DataHygiene::OrganisationReslugger.new(@organisation, "corrected-slug").run!
+        DataHygiene::OrganisationReslugger.new(@organisation.reload, "corrected-slug").run!
```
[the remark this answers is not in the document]
laurentqro: This one stumped me. The `after_commit` callbacks in https://github.com/alphagov/whitehall/blob/master/lib/publishes_to_publishing_api.rb#L6 only gets triggered if I reload here for some reason.

Interestingly, replacing `after_commit` with `after_update` makes it work, so there's something happening inside the database transaction after the save that rolls back the entire thing.

Interested to hear your thoughts about this one.

app/services/search_rummager_service.rb
```diff
@@ -2,11 +2,9 @@ class SearchRummagerService
   def fetch_related_documents(filter_params = {})
     options = default_search_options.merge(filter_params)
 
-    Rails.cache.fetch(options, expires_in: 5.minutes) do
-      search_response = Whitehall.search_client.search(options)
-      search_response["results"].map! { |res| RummagerDocumentPresenter.new(res) }
-      search_response
-    end
```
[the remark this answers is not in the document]
laurentqro: So I removed caching because it caused this error: `TypeError: no _dump_data is defined for class Thread::Mutex`.

I suspect this is because we're trying to serialise a mutex.

From the [Ruby docs](https://ruby-doc.org/core-2.7.0/Marshal.html): 

```
Some objects cannot be dumped: if the objects to be dumped include bindings, procedure or method objects, instances of class IO, or singleton objects, a TypeError will be raised.
```

I am investigating how to deal with this. Any suggestions warmly welcomed :)

config/environments/production.rb
```diff
@@ -24,7 +24,11 @@
 
   # Compress JavaScripts and CSS.
   config.assets.js_compressor = :uglifier
-  config.assets.css_compressor = :sass
+  # config.assets.css_compressor = :sass
+
+  # Rather than use a CSS compressor, use the SASS style to perform compression
+  config.sass.style = :compressed
+  config.sass.line_comments = false
```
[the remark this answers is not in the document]
laurentqro: Looks like it yeah!

issyl0:
Needs a rebase on master.

app/services/search_rummager_service.rb
```diff
@@ -1,17 +1,21 @@
...
   def fetch_related_documents(filter_params = {})
-    options = default_search_options.merge(filter_params)
+    search_params = default_search_params.merge(filter_params)
 
-    Rails.cache.fetch(options, expires_in: 5.minutes) do
-      search_response = Whitehall.search_client.search(options)
-      search_response["results"].map! { |res| RummagerDocumentPresenter.new(res) }
-      search_response
-    end
+    return search_response(search_params) if Rails.env.test?
```
cbaines: Do you know why this is useful? Ideally the code paths wouldn't differ based on the Rails environment.

laurentqro: Yeah I am not a fan of this either.

This is an attempt to get around: `LatestControllerTest#test_GET_:index_should_accept_pagination_parameters:
TypeError: no _dump_data is defined for class Thread::Mutex` (https://ci.integration.publishing.service.gov.uk/job/whitehall/job/upgrade-rails-to-version-5.2.3/68/console)

The alternative would be to use dependency injection to use an object in our test that does not use caching.

laurentqro: I would have expected `config.action_controller.perform_caching = false` in  `config/environments/test.rb` to prevent this, but no.

laurentqro: @cbaines  `SearchRummagerService`' is a thin wrapper around the search client, whose main function seems to be: 

* cache results returned bu the search client
* merge default search options into the the provided filter params

So isn't [this test](https://github.com/alphagov/whitehall/blob/master/test/unit/services/search_rummager_service_test.rb#L13) testing the search client itself rather than the object under test?

cbaines: I can see this has been addressed in https://github.com/alphagov/whitehall/pull/4998/commits/4314f729feb1b357faca93878009cb306cdb2056 now, thanks :)

config/application.rb
```diff
@@ -14,6 +14,9 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
 
+    # Initialize configuration defaults for originally generated Rails version.
+    config.load_defaults 5.0
```
AlanGabbianelli: Do we need 5.2 here?
```suggestion
    config.load_defaults 5.2
```

laurentqro: yes totally!

AlanGabbianelli:
Closing this as work has been reprioritized 🤷‍♂️

PR #5001, opened 2019-08-27
Stop image link in attachment block from being tab-able

injms:
The attachment component had the same link for the document thumbnail and for the document title. The thumbnail link focus state is non-existent, so people that use a keyboard to navigate around the page couldn't tell where the focus was. 

To fix this the `tabindex` on the thumbnail link has been set to `-1` to prevent keyboard navigation of duplicate link.

Trello card: https://trello.com/c/W3UJPgdZ/62-2-fix-focus-state-for-attachment-icons

image

Code before:
```html
<div class="attachment-thumb">
  <a aria-hidden="true" class="thumbnail" href="https://example.com/document.pdf">
    <img alt="" src="https://example.com/document.png">
  </a>
</div>
```

Code after:
```html
<div class="attachment-thumb">
  <a aria-hidden="true" class="thumbnail" tabindex="-1" href="https://example.com/document.pdf">
    <img alt="" src="https://example.com/document.png">
  </a>
</div>
```

### DIFF
--- a/app/views/documents/_attachment.html.erb
+++ b/app/views/documents/_attachment.html.erb
@@ -2,7 +2,7 @@
   published_on ||= ""
   help_block_id = "attachment-#{attachment.id}-accessibility-help"
   help_block_toggle_id = "attachment-#{attachment.id}-accessibility-request"
-  thumbnail_link_options = { "aria-hidden" => true, class: 'thumbnail' }
+  thumbnail_link_options = { "aria-hidden" => true, class: 'thumbnail', tabindex: '-1'}
 
   title_link_options = {}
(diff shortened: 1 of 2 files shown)


nickcolley:
Could we consider removing the link from the image entirely?

injms:
We could, but I think we should maintain feature parity with the current [attachment component](https://components.publishing.service.gov.uk/component-guide/attachment) in the gem which has a link on both the thumbnail and the heading.

nickcolley:
Generally it's bad practice to have duplicate links, as it'll be shown in overall lists of links.

Agreed the components should be the same :+1:

PR #5029, opened 2019-09-13
Unlink people translations from roles

vanitabarrett:
The current logic in Whitehall is to only show the available translations where: 1) translations for the person are present AND 2) translations for every role they have are present.

For reference, [this is the commit where the code was added for translations](https://github.com/alphagov/whitehall/commit/4dd466918202d848827377911b1bb16cebf0c75e), which states: "the translated name & responsibilities will show up on the locale-specific version of the public-facing person page, as long as both the person and *all* their current roles have translations into a given locale."

I'm not sure why the two need to be linked, and it's confusing when translations are present in Whitehall but don't appear on the live site.

## Before
Screen Shot 2019-09-13 at 17 03 24

## After
Screen Shot 2019-09-13 at 17 07 38

Example pages: 
 - /government/people/kirsty-cooper.cy
 - /government/people/alun-cairns-mp.cy

### DIFF
--- a/app/views/people/show.html.erb
+++ b/app/views/people/show.html.erb
@@ -68,18 +68,16 @@
       <div class="current-roles" id="current-roles">
         <% @person.current_role_appointments.each do |appointment| %>
           <%= content_tag_for :section, appointment, class: "role" do %>
-            <h2 class="gem-c-heading" id="<%= appointment.role.name.parameterize%>"><%= appointment.role.name %></h2>
-            <%= appointment.role.responsibilities %>
+            <h2 class="gem-c-heading" id="<%= appointment.role.name.parameterize%>" <%= t_lang_translated_locales(appointment.role) %> ><%= appointment.role.name %></h2>
+            <div <%= t_lang_translated_locales(appointment.role) %> ><%= appointment.role.responsibilities %></div>
             <div class="read-more">
               <% if appointment.role.ministerial? %>
                 <%= link_to t('roles.read_more'), appointment.role %>
               <% end %>
               <% if appointment.role.worldwide? %>
-                <%= render partial: 'organisations/organisations_name_list',
-                           locals: { organisations: appointment.role.worldwide_organisations } %>
+                <%= render partial: 'organisations/organisations_name_list', locals: { organisations: appointment.role.worldwide_organisations } %>
               <% else %>
-                <%= render  partial: 'organisations/organisations_name_list',
-                            locals: { organisations: appointment.role.organisations } %>
+                <%= render  partial: 'organisations/organisations_name_list', locals: { organisations: appointment.role.organisations } %>
               <% end %>
             </div>
           <% end %>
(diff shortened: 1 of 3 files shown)


cbaines:
> I'm not sure why the two need to be linked, and it's confusing when translations are present in Whitehall but don't appear on the live site.

I'd guess that either now, or in the past, data from the role and person showed up on the same page. One way of avoiding a Welsh page showing up with some English text would be to restrict the available locales to ones where both the role data, and person data is available in that locale.

vanitabarrett:
@cbaines Thanks, that makes sense and I hadn't thought of that. It seems quite a strict rule to just never display the translated page because parts of it might not be available in the translated language. 

I've followed the pattern the Frontend Accessibility team have implemented elsewhere in Whitehall, where we mark up specific sections of the page with `lang=en` when the translation is not available, e.g:

Screen Shot 2019-09-16 at 12 38 23

app/views/people/show.html.erb
```diff
@@ -68,15 +68,15 @@
...
+            <h2 class="gem-c-heading" id="<%= appointment.role.name.parameterize%>" <%= t_lang_translated_locales(appointment.role) %> ><%= appointment.role.name %></h2>
+            <div <%= t_lang_translated_locales(appointment.role) %> ><%= appointment.role.responsibilities %></div>
             <div class="read-more">
               <% if appointment.role.ministerial? %>
                 <%= link_to t('roles.read_more'), appointment.role %>
               <% end %>
               <% if appointment.role.worldwide? %>
                 <%= render partial: 'organisations/organisations_name_list',
-                           locals: { organisations: appointment.role.worldwide_organisations } %>
+                          locals: { organisations: appointment.role.worldwide_organisations } %>
```
cbaines: This looks like an unintentional whitespace change.

test/unit/presenters/person_presenter_test.rb
```diff
@@ -60,8 +60,8 @@ class PersonPresenterTest < ActionView::TestCase
     @person.stubs(:current_role_appointments).returns([role_appointment_1, role_appointment_2])
     @person.stubs(:translated_locales).returns(%i[en es])
 
-    assert_equal [:en], @presenter.translated_locales
-    refute @presenter.available_in_multiple_languages?
+    assert_equal %i(en es), @presenter.translated_locales
+    assert @presenter.available_in_multiple_languages?
   end
 
   test "is available in multiple languages if person and all current roles are available in the same multiple languages" do
```
cbaines: This test description looks like it could be updated.

cbaines: Also, given that the behaviour is now simpler, you might be able to reduce the number of tests.

cbaines:
> @cbaines Thanks, that makes sense and I hadn't thought of that. It seems quite a strict rule to just never display the translated page because parts of it might not be available in the translated language.

It's just a guess, so I wouldn't read too much in to it.

> I've followed the pattern the Frontend Accessibility team have implemented elsewhere in Whitehall, where we mark up specific sections of the page with `lang=en` when the translation is not available, e.g:

That sounds good :+1:

PR #5037, opened 2019-09-18
Update table guidance

chao-xian:
### Before

Screenshot at Sep 18 6-16-59 pm

### After

Screenshot at Sep 19 1-01-30 pm

- [x] Update examples to have narrower text


chao-xian:
[[Stefan] Document and convey how to use new accessible table header in govspeak](https://trello.com/c/6jVt8xr5/35-stefan-document-and-convey-how-to-use-new-accessible-table-header-in-govspeak)

PR #5040, opened 2019-09-20
Remove organisations API from Whitehall

sihugh:
[This is now provided by the collections app](https://github.com/alphagov/collections/blob/master/docs/api.md).

The organisation page references are fine to remove because org pages aren't served by Whitehall any more either.  They're used in a lot of Whitehall tests though so are a little tricky to unpick cleanly.


app/views/organisations/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <% page_title @organisation.name %>
 <% page_class "organisations-show" %>
 <% atom_discovery_link_tag atom_feed_url_for(@organisation), "Latest activity" %>
-<%= api_link_tag api_organisation_url(@organisation) %>
```
cbaines: Isn't this still relevant? The public URL should be the same, maybe the magic rails function could be replaced with one that just generates the right URL?

sihugh: That view's not actually rendered any more - the live one is [over in collections](https://github.com/alphagov/collections/blob/master/app/views/organisations/_meta.html.erb#L5). Removing all the org page stuff in its entirety is a larger task

cbaines: Right, ok. I was thinking that maybe this was still used for some pages, but I think I'm actually thinking of some other type of content... :+1:

PR #5055, opened 2019-09-30
Send unwithdrawn editions to the Publishing API

thomasleese:
Currently we only mark the edition as re-published in Whitehall without actually sending the edition to the Publishing API. This means that when users pressed "unwithdraw" they would have to manually publish the edition by pressing "force publish".

[Trello Card](https://trello.com/c/3jYLdax3/1293-5-investigate-being-unable-to-unwithdraw-whitehall-content)


thomasleese:
@edwardkerry Yep, good idea, I've added a test.

PR #5056, opened 2019-10-01
Run organisations link checker task at 3am

thomasleese:
We're seeing a problem where the link checker task means that Link Checker API ends up trying to talk to Whitehall while the data sync is in process, which means we see a lot of errors in Sentry.

By moving the task to 3am, the data sync will have finished by then, and the link checking task itself doesn't take very long.

[Trello Card](https://trello.com/c/nT3bi1q3/1317-5-investigate-link-checker-api-talking-to-whitehall-overnight)

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -12,5 +12,5 @@
   - asset_migration
 :schedule:
   check_all_organisations_links_worker:
-    cron: '0 1 * * *' # Runs at 1 a.m every day
+    cron: '0 3 * * *' # Runs at 3 a.m every day
     class: CheckAllOrganisationsLinksWorker


bevanloon:
Do we collect any data about how long this normally takes?

thomasleese:
Not really - it's complicated because the task runs over multiple apps (Whitehall and Link Checker API) and Link Checker API is checking links for more than just Whitehall.

However, on the Whitehall side it looks like it only takes about 30 minutes: https://grafana.publishing.service.gov.uk/dashboard/file/sidekiq.json?refresh=1m&orgId=1&var-Application=whitehall&var-Queues=All

PR #5063, opened 2019-10-02
Update frontend templates to latest publishing components gem

injms:
## What
Updating Whitehall to use the latest version of GOV.UK Publishing Components, which brings in version 3 of GOV.UK Frontend so it can meet WCAG 2.1 focus state rules. This pull request only updates the gem, makes sure that the global link styles don't override `govuk-link` from Frontend, and sets the compatibility flags needed for colour and text size.

## Why
Using the updated gem means that the work to upgrade Whitehall to the latest link focus state styles can begin. Rather than having a long drawn out pull request and a complicated rebase at the end the work has been chunked up to make it more manageable - so this pull request only covers the gem upgrade and work to prevent the clash of global link styles.

## Visual differences
None, since the components used in Whitehall haven't been affected by the gem upgrade.


chao-xian:
Minor things to note in future: keep commit message length under 70 chars.

PR #5081, opened 2019-10-15
Update how government works page

injms:
* Updates the links to have the new focus state
* Updates Sass to use variables from GOV.UK Frontend (via GOV.UK Publishing Components gem)
* Switches app components to gem components
* Replaces `<a href...` with `<%= link_to...` where possible
* Updates ministers circle feature to not break on medium sized screens
* Adds context to the links in the ministers circle feature so they're more readable by assistive tech
* Remove mysterious leading zero from the Prime Minister circle
* Updates the department numbers to stop it overlapping with the rest of the page
* Makes sure that the links in the department numbers make sense on their own so they're more readable

Live page: https://www.gov.uk/government/how-government-works

# Visual differences

## Ministers circle feature before

Showing layout glitch:
Screen Shot 2019-10-10 at 11 42 23

Small screens with focus state, hover state, no state, no state:
Screen Shot 2019-10-10 at 11 43 06

Focus state, hover state, no state, no state (total has different styles):
Screen Shot 2019-10-10 at 11 46 26

## Ministers circle feature after

Focus state, hover state, no state, no state (total has different styles):
Screen Shot 2019-10-10 at 11 48 39

Small screens with focus state, hover state, no state, no state:
Screen Shot 2019-10-10 at 11 42 57

## Department numbers before
Layout glitch and hover state:
Screen Shot 2019-10-10 at 11 41 41

## Department numbers after
`300+` reduced in size, and new hover state:
Screen Shot 2019-10-10 at 13 03 12

Part of https://trello.com/c/KJo6tZXo/58-5-update-whitehall-to-govukpublishingcomponents-v21x-with-compatibility-mode-on

### DIFF
--- a/app/assets/stylesheets/frontend/base.scss
+++ b/app/assets/stylesheets/frontend/base.scss
@@ -135,19 +135,21 @@ $govuk-typography-use-rem: false;
   @import "layouts/two_column_page";
 }
 
+
 // TODO: This 👇 `a.govuk-link:focus` overrides the global `a:link:focus` style
 // coming from static and the global styles set in `global/_links.scss`. This
 // can be removed when the style from static and Whitehall removed.
 #whitehall-wrapper {
-  a.govuk-link {
+  a.govuk-link{
+    text-decoration: underline;
     &:focus {
-      text-decoration: none;
       color: $govuk-focus-text-colour;
+      text-decoration: none;
     }
   }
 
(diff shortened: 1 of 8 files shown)


injms:
> as the goal isn't to replace toolkit, 🚢

Not yet - but that will come soon!

PR #5082, opened 2019-10-15
Update focus states for Get Involved page

injms:
- Adds govuk-link class to links
- Overrides colours where necessary
- Removes repetitive links from tab order and assistive tech
- Reworks large links to consultations to contextualise them

Part of https://trello.com/c/KJo6tZXo/58-5-update-whitehall-to-govukpublishingcomponents-v21x-with-compatibility-mode-on

### DIFF
--- a/app/assets/stylesheets/frontend/views/_get-involved.scss
+++ b/app/assets/stylesheets/frontend/views/_get-involved.scss
@@ -49,6 +49,27 @@
       }
     }
 
+    .count {
+
+      @include govuk-media-query($from: desktop) {
+        width: $half;
+      }
+
+      display: block;
+      margin: $gutter-half;
+      @include heading-80;
+      font-weight: bold;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/frontend/views/_get-involved.scss
```diff
@@ -120,6 +99,10 @@
           padding: $gutter-half $gutter-half 0;
           a {
             color: $white;
+
+            &:focus {
```
kr8n3r: could you do `:link:focus` to avoid `important`?

injms: That gives a specificity of 1, 6, 2 - unfortunately the thing that this is trying to override is 1, 6, 3. So close! I've shuffled it down the page to nest it under `.attributes`, which then gives it 1, 8, 3.

PR #5088, opened 2019-10-18
Fix url whitespace bug

1pretz1:
Previously if there was any whitespace in the URL for the related
content fields whitehall would crash throwing the standard 'Something
went wrong' message. This change strips all the whitespace from the URL
before parsing.

This problem became apparent when a zendesk came in which was related to
extra whitespace unknowingly added to the end of a URL.

Zendesk:
https://govuk.zendesk.com/agent/tickets/3831838


thomasleese:
That's funny, we had the same problem come up in Publisher very recently: https://github.com/alphagov/publisher/pull/1125

PR #5091, opened 2019-10-22
Remove organisations from preloaded DocumentCollectionGroups

bilbof:
CorporateInformationPages don't have organisations as an association, so this throws an ActiveRecord error if one of the documents is a corp info page.

Examples: https://sentry.io/organizations/govuk/issues/1272469573

The fix is to not use `organisations` in the includes query, necessitating an additional query for organisations.

I tested on integration and it seems to work. The downside is that page load may be a little slower (though it wasn't noticeable on integration).

https://govuk.zendesk.com/agent/tickets/3834746

### DIFF
--- a/app/controllers/admin/document_collection_groups_controller.rb
+++ b/app/controllers/admin/document_collection_groups_controller.rb
@@ -5,7 +5,7 @@ class Admin::DocumentCollectionGroupsController < Admin::BaseController
   def index
     @groups = @collection.groups.includes(
       memberships: [
-        { document: { latest_edition: %i[organisations translations] } },
+        { document: { latest_edition: %i[translations] } },
         :non_whitehall_link,
       ],
     )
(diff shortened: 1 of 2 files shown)


app/controllers/admin/document_collection_groups_controller.rb
```diff
@@ -5,7 +5,7 @@ class Admin::DocumentCollectionGroupsController < Admin::BaseController
   def index
     @groups = @collection.groups.includes(
       memberships: [
-        { document: { latest_edition: %i[organisations translations] } },
+        { document: { latest_edition: %i[translations] } },
```
barrucadu: > The fix is to not use organisations in the includes query, necessitating an additional query for organisations.

Should there be another query here, or does that happen elsewhere?

bilbof: It happens automatically elsewhere when `.organisations` is called.

PR #5093, opened 2019-10-28
Re-structure and reduce data output for document export

1pretz1:
Trello: https://trello.com/c/at9jVJVu/1036-review-contents-of-document-export-and-remove-unnecessary-fields

This restructures the data of a document export from Whitehall so that it structured as a JSON dump of a document with editions inside that and edition relations inside the edition. This is instead of the previous system where we had document and editions distinct and then edition associations distinct too. A consequence of this is that Content Publisher will have to adapt its importing logic.

This reduces the amount of data output substantially be reducing most associations to just the bare ids. This is to only export data that we have control over and anticipate not to change. This has the benefit of reducing the amount of data exported massively too.

Btw @kevindew edited this empty PR description in-case writing this confuses future Peter

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -24,6 +24,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
       alternative_format_provider_content_id: edition.try(:alternative_format_provider)&.content_id,
       attachments: present_attachments(edition),
       authors: edition.authors.map { |u| present_user(u) },
+      contacts: slice_association(edition, :depended_upon_contacts, %i[id content_id]),
       edition_policies: slice_association(edition, :edition_policies, %i[id policy_content_id]),
       editorial_remarks: present_editorial_remarks(edition),
       fact_check_requests: present_fact_check_requests(edition),
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -17,7 +17,8 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   private
 
   def present_edition(edition)
-    except = DOCUMENT_SUB_TYPES.map { |type| "#{type}_id".to_sym }
+    document_subtype_ids = DOCUMENT_SUB_TYPES.map { |type| "#{type}_id" }
+    except = %w[title summary body] + document_subtype_ids
 
     edition.as_json(except: except).merge(
       alternative_format_provider_content_id: edition.try(:alternative_format_provider)&.content_id,
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -17,7 +17,9 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   private
 
   def present_edition(edition)
-    edition.as_json.merge(
+    except = DOCUMENT_SUB_TYPES.map { |type| "#{type}_id".to_sym }
+
+    edition.as_json(except: except).merge(
       alternative_format_provider_content_id: edition.try(:alternative_format_provider)&.content_id,
       attachments: present_attachments(edition),
       authors: edition.authors.map { |u| present_user(u) },
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -138,7 +138,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
     end
 
     def call(edition)
-      links = admin_links(edition.body)
+      links = edition.translations.flat_map { |t| admin_links(t.body) }
 
       if edition.unpublishing&.explanation
         links.concat(admin_links(edition.unpublishing.explanation))
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -9,8 +9,10 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   ].freeze
 
   def as_json
+    presented_editions = model.editions.map { |e| present_edition(e) }
+    presented_users = users.map { |u| present_user(u) }
     model.as_json
-         .merge(editions: model.editions.map { |e| present_edition(e) })
+         .merge(editions: presented_editions, users: presented_users)
          .deep_symbolize_keys
   end
 
@@ -23,7 +25,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
     edition.as_json(except: except).merge(
       alternative_format_provider_content_id: edition.try(:alternative_format_provider)&.content_id,
       attachments: present_attachments(edition),
-      authors: edition.authors.map { |u| present_user(u) },
+      authors: edition.authors.pluck(:id),
       contacts: slice_association(edition, :depended_upon_contacts, %i[id content_id]),
       edition_policies: slice_association(edition, :edition_policies, %i[id policy_content_id]),
       editorial_remarks: present_editorial_remarks(edition),
@@ -31,7 +33,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
       government: edition.government&.as_json,
       revision_history: present_revision_history(edition),
       images: present_images(edition),
-      last_author: present_user(edition.last_author),
+      last_author: edition.last_author&.id,
       organisations: present_organisations(edition),
       role_appointments: slice_association(edition, %i[role_appointment role_appointments], %i[id content_id]),
       specialist_sectors: slice_association(edition, :specialist_sectors, %i[id topic_content_id primary]),
@@ -75,25 +77,19 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
 
     edition.editorial_remarks.map do |remark|
       remark.as_json(only: %i[id body created_at author_id])
-            .merge(author: present_user(remark.author))
     end
   end
 
   def present_fact_check_requests(edition)
     return [] unless edition.try(:fact_check_requests)
 
-    edition.fact_check_requests.map do |request|
-      request.as_json(except: "requestor_id")
-             .merge(requestor: present_user(request.requestor))
-    end
+    edition.fact_check_requests.map(&:as_json)
   end
 
   def present_revision_history(edition)
-    user_ids = edition.versions.map(&:whodunnit)
-    users = User.where(id: user_ids).index_by(&:id)
     edition.versions.map do |version|
-      version.as_json(only: %w[created_at event state whodunnit])
-             .merge(user: present_user(users[version.whodunnit.to_i]))
+      version.as_json(only: %w[created_at event state])
+             .merge(whodunnit: version.whodunnit.to_i)
     end
   end
 
@@ -126,9 +122,27 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   end
 
   def present_user(user)
-    return {} unless user
+    {
+      id: user.id,
+      uid: user.uid,
+      name: user.name,
+      email: user.email,
+      organisation_slug: user.organisation_slug,
+      organisation_content_id: user.organisation_content_id,
+    }
+  end
 
-    { id: user.id, uid: user.uid }
+  def users
+    user_ids = model.editions.inject([]) do |memo, edition|
+      whodunnits = edition.versions.pluck(:whodunnit).map(&:to_i)
+      remarkers = edition.editorial_remarks.pluck(:author_id)
+      requestors = edition.try(:fact_check_requests)&.pluck(:requestor_id).to_a
+      authors = edition.authors.pluck(:id)
+
+      memo + whodunnits + remarkers + requestors + authors
+    end
+
+    User.where(id: user_ids).order(:id)
   end
 
   class AdminLinkResolver
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -33,7 +33,6 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
       government: edition.government&.as_json,
       revision_history: present_revision_history(edition),
       images: present_images(edition),
-      last_author: edition.last_author&.id,
       organisations: present_organisations(edition),
       role_appointments: slice_association(edition, %i[role_appointment role_appointments], %i[id content_id]),
       specialist_sectors: slice_association(edition, :specialist_sectors, %i[id topic_content_id primary]),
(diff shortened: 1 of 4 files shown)


app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +12,127 @@ def as_json
...
         json.merge!("govspeak_content" => attachment.govspeak_content.as_json) if attachment.respond_to?(:govspeak_content)
       end
     end
   end
 
-  def provide_doctype_information(edition, output)
-    %i[news_article_type publication_type corporate_information_page_type speech_type].each do |type|
-      output[type] = edition.public_send(type)&.key if edition.respond_to?(type)
+  def provide_doctype_information(edition)
+    %i[news_article_type publication_type corporate_information_page_type speech_type].each_with_object({}) do |type, memo|
```
ChrisBAshton: I'd suggest moving these out so they can be read and configured more easily:

```ruby
def doctypes
  %i[
    news_article_type
    publication_type
    corporate_information_page_type
    speech_type
  ]
end
```

test/factories/specialist_sector.rb
```diff
@@ -1,4 +1,5 @@
 FactoryBot.define do
   factory :specialist_sector do
+    topic_content_id factory: :topic
```
ChrisBAshton: Did you figure out why this empty factory existed in the end? 😁

app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +12,127 @@ def as_json
...
   end
 
-  def complete_images_hash(edition_images)
-    edition_images.map do |image|
+  def stripped_authors(edition)
+    authors = edition.try(:authors)
+    return [] unless authors
+
+    authors.map do |a|
+      { id: a.id, uid: a.uid }
```
ChrisBAshton: General suggestion for all these methods: avoid abbreviating the variables. This is longer but scans more easily (less thinking involved):

```suggestion
    authors.map do |author|
      { id: author.id, uid: author.uid }
```

app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +12,127 @@ def as_json
...
-    output[:government] = edition.government
-    output[:whitehall_admin_links] = resolve_whitehall_admin_links(edition.body)
-    if edition.withdrawn?
-      output[:whitehall_admin_links].concat(resolve_whitehall_admin_links(edition.unpublishing.explanation))
+  def complete_translations(edition)
+    translations = edition.try(:translations)
+    return [] unless translations
+
+    translations.map do |translation|
+      translation.attributes.except("edition_id").merge(
```
ChrisBAshton: Nice use of `except` - you're right, we don't need that field 👍However, think you can use the `:edition_id` symbol instead here.

app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +12,127 @@ def as_json
...
+        government: edition.government,
+        whitehall_admin_links: resolve_whitehall_admin_links(edition),
+        associations: {
+          attachments: complete_attachments(edition),
+          authors: stripped_authors(edition),
+          edition_organisations: complete_organisations(edition),
+          edition_policies: stripped_edition_policies(edition),
+          images: complete_images(edition),
+          last_author: stripped_last_author(edition),
+          topical_events: strippped_topical_events(edition),
```
ChrisBAshton: This has got three `p`s.

app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +12,127 @@ def as_json
...
-        edition_attachments = edition.public_send(association)
-        output[:associations][association] = complete_attachments_hash(edition_attachments)
-      else
-        output[:associations][association] = edition.public_send(association)
-      end
+  def stripped_last_author(edition)
+    author = edition.try(:last_author)
+    return {} unless author
+
+    { id: author.id, uid: author.uid }
```
ChrisBAshton: Do we need both the `id` and the `uid`? That feels strange to me 🤔

app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +12,127 @@ def as_json
 
   def editions
     model.editions.map do |edition|
-      edition_associations(edition)
+      {
+        edition: edition,
+        government: edition.government,
```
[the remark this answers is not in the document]
1pretz1: Good question, the reason being is that a government isn't directly related to an edition. This is worked out on the fly from when an edition is created (or published or something, can't quite remember) against which government (table in whitehall) was in power at the time.

So it's not really an association as there is no connection via a foreign key like there is for all the others. I'm not sure what you mean by duplicated though?

app/presenters/document_export_presenter.rb
1pretz1: Ah damn it, too eager with the p key, good spot!

app/presenters/document_export_presenter.rb
1pretz1: `id` is the field local to whitehall's db and uid is what's shared between applications. In signon the uid is used to identify a user however I agree that having the `id` may be overkill as probably don't need it. (however according to Kevins checklist having the id seems a required field).

app/presenters/document_export_presenter.rb
1pretz1: Hmm don't think so as `translation.attributes` returns string keys, you could if you called something like hash_with_indifferent_access on it though - https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html.

app/presenters/document_export_presenter.rb
1pretz1: Sure, i'll un-abbreviate (don't think thats a word) all the things.

app/presenters/document_export_presenter.rb
1pretz1: Not my code 😢But I agree with the premiss of your suggestion with a slight tweak, I'll make the list into constant array as this seems more ruby'esq.

test/factories/specialist_sector.rb
1pretz1: Apparently you don't need to define anything in the factory itself, as you can just define all the fields on creation e.g `create(:specialist_sector, name: 'something, ...)` however this is just laborious as you at least want default associations.

1pretz1:
Thanks for the review @ChrisBAshton, sorry it's not well presented commit-wise, was going to do this at the end due to the nature of the work! 

I also agree with you about the `uid` refactoring, it was one of things on my to-do list!

test/factories/specialist_sector.rb
ChrisBAshton: Ah I suppose it was done for readability, even if it does little under the hood.

app/presenters/document_export_presenter.rb
ChrisBAshton: Yeah I'm using that method on my PR. Hmm on second thoughts that would just make this weightier with little added value. Fine to ignore!

app/presenters/document_export_presenter.rb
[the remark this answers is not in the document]
1pretz1: I'm not sure I know what you mean, the `government` field is for the edition to see what government it was published under not it's associations?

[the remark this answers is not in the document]
kevindew: I feel like we need to do something about this field structure overall. It seems to be getting progressively stranger with edition and associations being distinct and then these two new ones of government and whitehall_admin_links.

I wonder if we can just merge the association data into an edition so they all sit at the same top level as fields such as content_id and such.

app/presenters/document_export_presenter.rb
kevindew: I think there are going to be some places where we're just getting user ids not uids so may need a mechanism to look them up - the versions table comes to mind. 

It could well be that some of the local Whitehall ids are superfluous but I think in general they offer quite a lot of look-up information that is not present on some of these global ids.

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -221,22 +221,38 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
-                 summary: "Ça va",
-                 body: "ça va bien merci",
-                 created_at: french_translation.created_at,
-                 updated_at: french_translation.updated_at,
-                 base_path: "/government/news/news-title.fr" }
-    assert_equal expected,
-                 result.dig(:editions, 0, :translations, 1)
+    expected = [
+      { id: english_translation.id,
+        locale: :en,
```
ChrisBAshton: Yes, odd that this comes out as a symbol! Have looked at the presenter and can't see an obvious reason for it.
But not an issue as it comes out as a JSON string in the end anyway.

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -149,6 +149,31 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
+    edition = create(:edition)
+
+    user_2 = create(:user)
+    Edition::AuditTrail.whodunnit = user_2
+    edition.update!(change_note: "changed")
+
+    result = DocumentExportPresenter.new(edition.document).as_json
+    expected = [
+      { event: "create",
+        whodunnit: user_1.id.to_s,
```
ChrisBAshton: Did we agree to convert this to `int`?

app/presenters/document_export_presenter.rb
```diff
@@ -1,171 +1,142 @@
...
-        "public_base_path" => Whitehall.url_maker.public_document_path(edition, locale: translation.locale),
-      )
+    edition.attachments.map do |attachment|
+      govspeak_content = if attachment.respond_to?(:govspeak_content)
+                           { govspeak_content: attachment.govspeak_content.as_json }
+                         else
+                           {}
+                         end
+      attachment.as_json(include: :attachment_data, methods: %i[url type])
+                .merge(govspeak_content)
```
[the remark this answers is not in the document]
kevindew: I believe this is a separate association that isn't included. I don't believe it can be included as attachment is a superclass and govspeak_content is only available on the HtmlAttachment subclass (hence the respond_to?)

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -1,6 +1,23 @@
...
 class DocumentExportPresenterTest < ActiveSupport::TestCase
+  test "includes basic document and edition information" do
+    document = create(:document)
+    edition = create(:edition, document: document)
+    result = DocumentExportPresenter.new(document).as_json
+
+    assert_equal document.content_id, result[:content_id]
+    assert_equal document.slug, result[:slug]
+    assert_equal edition.id, result.dig(:editions, 0, :id)
+  end
```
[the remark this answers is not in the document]
kevindew: You're right - we seemed to miss just a plain "it works" test

test/unit/presenters/document_export_presenter_test.rb
kevindew: Yes - though we need to work out when the user changes are done.

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -159,6 +159,17 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
 
+  test "includes contacts" do
+    contact = create(:contact)
+    edition = create(:news_article,
+                     body: "[Contact:#{contact.id}]",
+                     depended_upon_contacts: [contact])
+
+    result = DocumentExportPresenter.new(edition.document).as_json
+    assert_equal ([{ id: contact.id, content_id: contact.content_id }]), result.dig(:editions, 0, :contacts)
+    assert_nil result.dig(:editions, 0, :depended_upon_contacts)
```
ChrisBAshton: I'm not quite sure how this this is `nil` as we don't seem to be making any effort to remove it 🤔

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -159,6 +159,17 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
                  result.dig(:editions, 0, :authors, 0)
   end
 
+  test "includes contacts" do
+    contact = create(:contact)
+    edition = create(:news_article,
+                     body: "[Contact:#{contact.id}]",
```
ChrisBAshton: This line probably isn't needed - what's used is the `depended_upon_contacts` array, which is populated [as per edition_dependencies_populator_test.rb](https://github.com/alphagov/whitehall/blob/master/test/unit/edition_dependencies_populator_test.rb#L5-L13). I did consider writing a test for when `depended_upon_contacts` isn't yet populated but `body` contains a contact - but don't think this is a valid scenario.

kevindew: Yeah you're right, I removed it from the test since it doesn't seem to be doing anything.

test/unit/presenters/document_export_presenter_test.rb
kevindew: We don't need to remove anything as it's not automatically populated. I've removed this test condition as I don't think there's a particular reason we're bothered about it's presence or not.

app/presenters/document_export_presenter.rb
```diff
@@ -12,60 +19,56 @@ def as_json
 
   def editions
     model.editions.map do |edition|
-      edition_associations(edition)
+      {
+        edition: edition,
+        government: edition.government,
```
leenagupte: Is the value used in history mode?

kevindew: Yup

kevindew:
@ChrisBAshton I've put the user stuff into this @brucebolt do you want to check if this is good for you? I ended up adding the organisation details as I realised without it we have no idea which organisation a user is from until they sign in.

brucebolt:
The users commits look great, thank you @kevindew.  I'm just reviewing the remainder of the PR.

app/presenters/document_export_presenter.rb
```diff
@@ -27,6 +27,7 @@ def present_edition(edition)
       edition_policies: slice_association(edition, :edition_policies, %i[id policy_content_id]),
       fact_check_requests: present_fact_check_requests(edition),
       government: edition.government&.as_json,
+      history: present_history(edition),
```
brucebolt: I wonder if the name `history` could cause confusion with history mode?  Would something like `change_log` or `revision_history` be clearer?

ChrisBAshton: Have now changed this to `revision_history` and updated the method name and commit message.

PR #5099, opened 2019-10-31
Stop locked documents from appearing in CSV export

JonathanHallam:
Trello - https://trello.com/c/qgeWWHYa/1150-create-card-to-exclude-locked-doc-from-csv-export

As it stands when a user exports a CSV of documents from whitehall this includes
locked Documents that are either in the process of or has finished migrating to
content publisher. This should not be the case as once it has been migrated,
information that Whitehall has on the document may now be out of date. This PR
ensures that locked documents are no longer included in the CSV.

We've manually tested this in integration and it successfully changes the URL at
the bottom of the search results and stops Editions from appearing in the CSV report

Co-authored-by: Bruce Bolt <bruce.bolt@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/workers/document_list_export_worker.rb
+++ b/app/workers/document_list_export_worker.rb
@@ -17,7 +17,14 @@ private
   end
 
   def create_filter(filter_options, user)
-    Admin::EditionFilter.new(Edition, user, filter_options.symbolize_keys.merge(include_unpublishing: true))
+    Admin::EditionFilter.new(
+      Edition,
+      user,
+      filter_options.symbolize_keys.merge(
+        include_unpublishing: true,
+        include_locked_documents: false,
+      ),
+    )
   end
 
   def generate_csv(filter, csv_file)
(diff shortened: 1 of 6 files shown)


app/workers/document_list_export_worker.rb
```diff
@@ -17,7 +17,10 @@ def send_mail(csv, user, filter)
   end
 
   def create_filter(filter_options, user)
-    Admin::EditionFilter.new(Edition, user, filter_options.symbolize_keys.merge(include_unpublishing: true))
+    Admin::EditionFilter.new(Edition, user, filter_options.symbolize_keys.merge(
+                                              include_unpublishing: true,
+                                              include_locked_documents: false,
+    ))
```
ChrisBAshton: The formatting of this one looks odd to me - might be neater with each argument on its own line?

JonathanHallam: Yeah the linter complained a lot about this, never would have put stuff here on my own. Will mutliline this now :)

PR #5100, opened 2019-11-01
Export variants of image sizes and file attachments

ChrisBAshton:
When we import content from Whitehall, we need to ensure that we
transfer all the variants of image sizes and attachments too, in case they are
used by frontend applications. This is particularly true of image thumbnails.
There is no easy way to query these variants from outside of Whitehall, so this
work is best done as part of the export.

Trello card: https://trello.com/c/XOJ64DPo/1151-export-variants-of-images-and-attachments

Screenshot of new data structure

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -95,7 +95,16 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
   def present_images(edition)
     return [] unless edition.try(:images)
 
-    edition.images.map { |image| image.as_json(methods: :url) }
+    edition.images.map do |image|
+      image.as_json(methods: :url)
+           .merge(variants: image_variants(image))
+    end
+  end
+
+  def image_variants(image)
+    image.image_data.file.versions.each_with_object({}) do |(variant, details), memo|
+      memo[variant] = details.url
+    end
   end
 
   def present_organisations(edition)
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -68,6 +68,18 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
                          end
       attachment.as_json(include: :attachment_data, methods: %i[url type])
                 .merge(govspeak_content)
+                .merge(variants: attachment_variants(attachment))
+    end
+  end
+
+  def attachment_variants(attachment)
+    return {} unless attachment.try(:attachment_data)
+
+    attachment.attachment_data.file.versions.each_with_object({}) do |(variant, details), memo|
+      memo[variant] = {
+        content_type: details.content_type,
+        url: details.url,
+      }
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/image.rb
```diff
@@ -15,6 +15,14 @@ def url(*args)
     image_data.file_url(*args)
   end
 
+  def variants
+    sizes = {}
+    image_data.file.versions.each do |variant, details|
+      sizes[variant] = details.url
+    end
+    sizes
```
kevindew: Minor, you might want to look at using each_with_object for this.

app/models/image.rb
```diff
@@ -15,6 +15,14 @@ def url(*args)
     image_data.file_url(*args)
   end
 
+  def variants
+    sizes = {}
+    image_data.file.versions.each do |variant, details|
+      sizes[variant] = details.url
```
kevindew: Nice one digging in and finding this out 👍

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -128,6 +148,25 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
     assert_equal "FileAttachment", attachment[:type]
     assert_equal publication_file.attachments.first.attachment_data.as_json.symbolize_keys,
                  attachment[:attachment_data]
+    # TODO - do titles vary per attachment variant?
+    assert_match(/^file\-attachment\-title\-[0-9]$/, attachment[:title])
```
ChrisBAshton: I'd like to double check that we don't expect each variant of FileAttachment to have its own title (e.g. "File attachment PDF" vs "File attachment thumbnail").

Currently I believe we just have `:title` on the FileAttachment root, which in this hypothetical scenario would be "File attachment".

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -145,6 +145,7 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
     attachment = result.dig(:editions, 0, :attachments, 0)
 
     assert_equal publication_file.attachments.first.url, attachment[:url]
+    assert_match(/^file\-attachment\-title\-[0-9]$/, attachment[:title])
```
ChrisBAshton: I'd like to double check that we don't expect each variant of FileAttachment to have its own title (e.g. "File attachment PDF" vs "File attachment thumbnail").

Currently I believe we just have `:title` on the FileAttachment root, which in this hypothetical scenario would be "File attachment".

app/presenters/document_export_presenter.rb
```diff
@@ -68,6 +68,18 @@ def present_attachments(edition)
                          end
       attachment.as_json(include: :attachment_data, methods: %i[url type])
                 .merge(govspeak_content)
+                .merge(variants: attachment_variants(attachment))
+    end
+  end
+
+  def attachment_variants(attachment)
+    return [] unless attachment.try(:attachment_data)
```
kevindew: Should this return a hash to be consistent with the next statement?

ChrisBAshton: Whoops - good spot - it should!

PR #5103, opened 2019-11-04
Update focus state for Past Prime Ministers page

maxgds:
Trello: https://trello.com/c/6H95jFL2

- Adds govuk-link class where needed
- Uses title component to inherit correct focus state
- Add tabindex to prevent images from getting focus (links are duplicated below in text)

Live page: https://www.gov.uk/government/history/past-prime-ministers

### Before

Screen Shot 2019-11-04 at 14 29 59
Screen Shot 2019-11-04 at 14 30 24
Screen Shot 2019-11-04 at 14 30 33

### After
Screen Shot 2019-11-04 at 14 31 44
Screen Shot 2019-11-04 at 14 31 55

### DIFF
--- a/app/views/historic_appointments/_role_appointment.html.erb
+++ b/app/views/historic_appointments/_role_appointment.html.erb
@@ -1,9 +1,9 @@
 <%= content_tag_for :li, role_appointment, class: 'person person-excerpt' do %>
   <div class="inner">
     <div class="image-holder">
-      <%= link_to_if role_appointment.has_historical_account?, role_appointment.person.image, historic_appointment_path(role.historic_param, role_appointment.person), class: 'img' %>
+      <%= link_to_if role_appointment.has_historical_account?, role_appointment.person.image, historic_appointment_path(role.historic_param, role_appointment.person), class: "img", aria: {hidden: true}, tabindex: "-1" %>
     </div>
-    <h3 class="name"><%= link_to_if role_appointment.has_historical_account?, role_appointment.person.name, historic_appointment_path(role.historic_param, role_appointment.person) %></h3>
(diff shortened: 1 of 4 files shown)


app/views/historic_appointments/_role_appointment.html.erb
```diff
@@ -1,9 +1,9 @@
 <%= content_tag_for :li, role_appointment, class: 'person person-excerpt' do %>
   <div class="inner">
     <div class="image-holder">
-      <%= link_to_if role_appointment.has_historical_account?, role_appointment.person.image, historic_appointment_path(role.historic_param, role_appointment.person), class: 'img' %>
+      <%= link_to_if role_appointment.has_historical_account?, role_appointment.person.image, historic_appointment_path(role.historic_param, role_appointment.person), class: "img", aria: {hidden: true}, tabindex: "-1" %>
     </div>
-    <h3 class="name"><%= link_to_if role_appointment.has_historical_account?, role_appointment.person.name, historic_appointment_path(role.historic_param, role_appointment.person) %></h3>
+    <h3 class="name"><%= link_to_if role_appointment.has_historical_account?, role_appointment.person.name, historic_appointment_path(role.historic_param, role_appointment.person), class: "govuk-link" %></h3>
```
injms: Can we adjust the line height to avoid the focus lower border being partially obscured?

A `line-height` of 1.2 gives this result in Safari:
Screenshot 2019-11-05 at 10 45 27
 
This'll need checking in other browsers. I found a [similar thing happened for related links](https://github.com/alphagov/govuk_publishing_components/pull/1045) and the behaviour varied from browser to browser. Unfortunately those numbers don't _quite_ work here.

maxgds: Changed to 1.16 where it works in Chrome, Firefox, IE11 and Safari!

injms: Great work!

PR #5119, opened 2019-11-06
Add unpublishing details to document export

leenagupte:
Trello: https://trello.com/c/8U7JPIHT

## What's changed?
Adds unpublishing details to the export.

## Why?
So the that information it contains, for example the reason the document was unpublished, the user facing explanation, and where users may be being redirected to are not lost when the document is exported.

## Expected changes
Screenshot 2019-11-06 at 12 21 30

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -38,6 +38,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
       specialist_sectors: slice_association(edition, :specialist_sectors, %i[id topic_content_id primary]),
       topical_events: slice_association(edition, :topical_events, %i[id content_id]),
       translations: present_translations(edition),
+      unpublishing: present_unpublishing(edition),
       whitehall_admin_links: AdminLinkResolver.call(edition),
       world_locations: slice_association(edition, :world_locations, %i[id content_id]),
       worldwide_organisations: slice_association(edition, %i[worldwide_organisation worldwide_organisations], %i[id content_id]),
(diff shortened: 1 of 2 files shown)


app/presenters/document_export_presenter.rb
```diff
@@ -141,6 +142,20 @@ def present_translations(edition)
...
 
+  def present_unpublishing(edition)
+    edition
+      .unpublishing
+      .as_json(
+        only: %i[
+          id
+          explanation
+          alternative_url
+          redirect created_at
```
ChrisBAshton: Looks like these two slipped in on the same line.

test/unit/presenters/document_export_presenter_test.rb
```diff
@@ -393,6 +393,22 @@ class DocumentExportPresenterTest < ActiveSupport::TestCase
...
+      id: edition.unpublishing.id,
+      explanation: edition.unpublishing.explanation,
+      alternative_url: nil,
+      created_at: edition.unpublishing.created_at,
+      updated_at: edition.unpublishing.updated_at,
+      redirect: false,
+    }
+
+    assert_equal expected, result.dig(:editions, 0, :unpublishing)
+  end
```
ChrisBAshton: Is it worth writing another test for unpublish-and-redirect just to make sure `redirect` and `alternative_url` are set?

app/presenters/document_export_presenter.rb
```diff
@@ -143,17 +143,22 @@ def present_translations(edition)
...
+        .as_json(
+          only: %i[
+            id
+            explanation
+            alternative_url
+            redirect created_at
+            updated_at
+          ],
+        )
+        .tap do |unpublishing|
```
ChrisBAshton: This feels cleaner than tapping:

```ruby
.merge(reason: edition.unpublishing.unpublishing_reason.name)
```

leenagupte:
@ChrisBAshton Thanks for your comments, I've addressed them and fixed up.

> I wonder if explanation and reason are confusingly similar? AFAIK the explanation is user facing, and the reason is an internal note. Perhaps user_facing_explanation and unpublishing_reason are better names?

I've updated `reason` to be `unpublishing_reason` as suggested, but I've kept `explanation` as both fields now match the original column names.

PR #5122, opened 2019-11-06
Kill off advanced search

barrucadu:
search-api has three ways of doing search:

- Normal search (what you're familiar with from /api/seach.json)
- Batch search (used by some finders)
- Advanced search (used by whitehall)

Whitehall doesn't just use advanced search, it also uses the normal search endpoint for some things.

We've got a card on the tech debt board about killing off this whitehall-specific search endpoint.  We killed off all the (English) whitehall finders, and all of the non-English whitehall finders use mysql as a search backend.  So if the search-based but non-findery parts of whitehall work with the normal search endpoint, we can remove advanced search!

---

[Trello card](https://trello.com/c/kWUWnvcz/1120-spike-kill-off-search-apis-advanced-search-endpoint)


config/initializers/search_backend.rb
```diff
@@ -1,5 +1 @@
-Whitehall.search_backend = if Rails.env.test?
```
sihugh: I hope this is OK?  I'm not sure how much confidence I have in the tests catching whether this will still work afterwards given the liberal use of fake things.

barrucadu: Yeah, I didn't have *too* much confidence in the tests, but the e2e tests ran fine.  Smokey and traffic replay should help too.

PR #5127, opened 2019-11-08
Remove redundant code and standardise return values

1pretz1:
The two returns aren't needed as an empty array is returned anyway when
calling `Array(...)`.

I've also updated attachment_variants to return nil instead of an empty
hash to follow the pattern of other associations.

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -53,8 +53,6 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
 
   def slice_association(edition, associations, fields)
     association_data = Array(associations).flat_map { |a| edition.try(a) }.compact
-    return [] unless association_data
-
     association_data.map { |item| item.as_json(only: fields) }
   end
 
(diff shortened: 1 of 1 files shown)


app/presenters/document_export_presenter.rb
```diff
@@ -143,21 +140,18 @@ def present_translations(edition)
...
-            explanation
-            alternative_url
-            redirect
-            created_at
-            updated_at
-          ],
-        )
-        .merge(unpublishing_reason: edition.unpublishing.unpublishing_reason.name)
-    end
+    return {} unless edition.unpublishing
```
leenagupte: I talked to @kevindew before doing this, and he suggested returning `nil` rather than an empty object as that is the current pattern. e.g. `government` returns `nil` if the document isn't assigned to a government.

1pretz1: Very true, didn't see that example! Instead I've changed the associations returning an empty hash to `nil` so now the default for everything is either an empty array or nil! 🎉

1pretz1:
Sorry for being pedantic @leenagupte! 😆I'll go back in my hole now.

PR #5131, opened 2019-11-12
Export dimensions of main image

ChrisBAshton:
This PR:

- exports `width` and `height` attributes of the main image, so that Content Publisher can know
  whether or not it conforms to the required 960x640 size.
- does not attempt to do this for SVGs
- removes redundant ID fields which have no use outside of Whitehall

I've decided not to implement `width`/`height` detection for image `variants` as we have no immediate need for this in Content Publisher. It can be calculated in Content Publisher later if needed, or even derived from the aspect ratio of the main image divided by the size of the variant (e.g. `960x640` => `s720 variant` => `720x480`). Deriving image dimensions is slow because of the network requests involved, and this way we can avoid making 6 out of 7 requests, making the export much faster than it would otherwise be.

Trello card: https://trello.com/c/VFSu0ClH/1179-export-exact-image-sizes-in-the-whitehall-export

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'dalli', '~> 2.7'
 gem 'deprecated_columns', '~> 0.1.1'
 gem 'equivalent-xml', '~> 0.6.0', require: false
 gem 'faraday'
+gem 'fastimage'
 gem 'friendly_id', '~> 5.2.4'
 gem "fuzzy_match", "~> 2.1"
 gem 'gds-sso', '~> 14.2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -138,6 +138,7 @@ GEM
       activesupport (>= 4.2.0)
     faraday (0.17.0)
       multipart-post (>= 1.2, < 3)
+    fastimage (2.1.7)
     ffi (1.11.1)
     friendly_id (5.2.4)
       activerecord (>= 4.0.0)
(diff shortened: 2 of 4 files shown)


1pretz1:
Nice work @ChrisBAshton , fastimage is a cool little gem! I wonder if we need this though due to whitehall restricting all uploaded images to be of a set dimension of `960x640`. 

Do we have any example's where this is not the case?

ChrisBAshton:
I don't personally have any examples of Whitehall images that aren't 960x640, no. But this check was asked for in the [original Trello card](https://trello.com/c/jzV3scso/1093-import-images-into-content-publisher) by Kevin. I can only assume he has some historical context and that there are some images that differ.

1pretz1:
Hmm true, who knows whats inside whitehall really! 😆

brucebolt:
It looks like image dimensions have been enforced for at least 6 years (https://github.com/alphagov/whitehall/blame/13ccb069d3df45371c7c7390c292968d052c7ea5/app/models/image_data.rb).  Maybe we should identify if any images don't meet this criteria before exporting the dimensions?

ChrisBAshton:
After talking through the new image approach with Kevin today, sounds like we'll be downloading & reuploading images as part of the import script in Content Publisher, in which case it'll be relatively trivial to check the dimensions locally. Closing.

PR #5140, opened 2019-11-13
Add placeholder endpoint to mark a document as migrated

brucebolt:
Once a document has been migrated into Content Publisher, we will need to perform a number of post-migration processes in Whitehall.  This endpoint will be used to trigger such processes.

Trello card: https://trello.com/c/YIyJBmHN

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -29,6 +29,7 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
 
   def migrated
     document = Document.find(params[:id])
+    head :bad_request unless document.locked?
   end
 
   private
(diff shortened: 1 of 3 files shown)


app/controllers/admin/export/document_controller.rb
```diff
@@ -27,6 +27,13 @@ def unlock
     document.update!(locked: false)
   end
 
+  def migrated
+    document = Document.find(params[:id])
+    unless document.locked?
+      respond_with Hash.new, status: 422
```
kevindew: It seems a bit unusual to use `Hash.new` rather than `{}`. Though I'd wonder if we can just do a head method rather than needing a respond with and have it return empty content.

I'd suggest doing `head :unprocessable_entity` instead. 

From [a little reading](https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data) I'm also wondering if :bad_request might be more suited. There isn't any issues with the syntax concerns here which lead it slightly more towards 422 but 400 does make it quite distinct from Rails validation where my concern lies there.

PR #5143, opened 2019-11-14
Replace internal links in documents that link to a migrated document

brucebolt:
Whitehall allows publishers to link to other Whitehall documents using an internal URL (rather than the public one).  On migration of a document out of Whitehall, links to this document will no resolve.  Therefore, we need to replace internal links to this document in other documents with the public URL.

E.g.

```
a [link](/government/admin/policies/101010)
```
in a document would be changed to (when edition 101010 is migrated):
```
a [link](https://www.gov.uk/news/test)
```

Trello card: https://trello.com/c/jSKNSjtA

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -29,7 +29,11 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
 
   def migrated
     document = Document.find(params[:id])
-    head :bad_request unless document.locked?
+    return head :bad_request unless document.locked?
+
+    document.editions.each do |edition|
+      Whitehall::InternalLinkUpdater.new(edition).call
+    end
   end
 
   private
--- a/lib/whitehall/internal_link_updater.rb
+++ b/lib/whitehall/internal_link_updater.rb
@@ -1,0 +1,53 @@
+class Whitehall::InternalLinkUpdater
+  include GovspeakHelper
+
+  attr_reader :linked_to_edition
+
+  def initialize(linked_to_edition)
+    @linked_to_edition = linked_to_edition
+  end
+
+  def call
+    update_all_linking_editions(linked_to_edition)
+  end
+
+  private
+
+  def update_all_linking_editions(linked_to_edition)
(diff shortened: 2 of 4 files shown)


app/controllers/admin/export/document_controller.rb
```diff
@@ -55,4 +59,34 @@ def page_number
...
+    linked_from_editions(edition).each do |linked_from_edition|
+      linked_from_edition.translations.each do |translation|
+        whitehall_admin_links(translation.body).each do |whitehall_admin_link|
+          if whitehall_admin_link_edition(whitehall_admin_link).document_id == edition.document_id
+            translation.body.sub!(whitehall_admin_link, edition_public_url(edition))
+          end
+        end
+        translation.save!
+      end
+    end
```
kevindew: We may want to think about refactoring this as 3 nestings of loop is pretty tough for comprehension.

app/controllers/admin/export/document_controller.rb
```diff
@@ -55,4 +59,34 @@ def page_number
...
+
+  def update_linking_documents(edition)
+    linked_from_editions(edition).each do |linked_from_edition|
+      linked_from_edition.translations.each do |translation|
+        whitehall_admin_links(translation.body).each do |whitehall_admin_link|
+          if whitehall_admin_link_edition(whitehall_admin_link).document_id == edition.document_id
+            translation.body.sub!(whitehall_admin_link, edition_public_url(edition))
+          end
+        end
+        translation.save!
```
kevindew: Does this cause any callbacks? I know edition has a bunch but I don't think this does.

brucebolt: Doesn't look like it.

Although I noticed this updates the `updated_at` time on the translation, which we probably don't want, so I'll use `translation.save(touch: false)` to stop that happening.

brucebolt:
I've refactored this out of the controller, removed the nested loop and stopped the translation's `updated_at` from being changed.

brucebolt:
Good point.  I've made a small change to work through all editions for a document.

PR #5144, opened 2019-11-15
[DO NOT MERGE] Adjust rules about marking content as political

cbaines:
Don't merge yet, as we're looking in to how these changes will be communicated.


cbaines:
There was uncertainty about making this change, so I'll close this.

PR #5146, opened 2019-11-15
Update previous role appointment for Sarah Albon

emmabeynon:
Zendesk: https://govuk.zendesk.com/agent/tickets/3853039

It seems the Person Sarah Albon was assigned an incorrect role appointment. The
correct role appointment has now been created, however we also need to update
all content tagged to Sarah and the incorrect role appointment.


vanitabarrett:
Tested on integration and it seems to work as expected.

## Before
Screen Shot 2019-11-19 at 13 54 41

## After
Screen Shot 2019-11-19 at 13 54 11

PR #5161, opened 2019-11-19
Stop sending role appointment links

thomasleese:
This changes the links we send to the Publishing API to make use of the new reverse links that have been added.

See https://github.com/alphagov/publishing-api/pull/1645 for more details.

The main benefit to doing this is to avoid having to maintain four sets of links for each role appointment (person -> appointment, appointment -> person, role -> appointment, appointment -> role) and instead we only have two links (appointment -> person, appointment -> role) and we can use the Publishing API to correctly present the reverse links back to the appointment on the person and the role.

This means we don't need to worry about the links ending up out of sync with each other which in practice would've meant that we had added some `after_save` hooks on role appointments which republish the person and role to the Publishing API.

[Trello Card](https://trello.com/c/roHfuPUV/1362-8-use-reverse-links-for-role-appointments)

### DIFF
--- a/app/presenters/publishing_api/person_presenter.rb
+++ b/app/presenters/publishing_api/person_presenter.rb
@@ -32,8 +32,8 @@ module PublishingApi
 
     def links
       {
-        ordered_current_appointments: item.current_role_appointments.pluck(:content_id).compact,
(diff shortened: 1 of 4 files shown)


thomasleese:
Once everything is republished without these links, we can remove this code from Whitehall.

PR #5162, opened 2019-11-19
New focus states on ministers page

injms:
Updated the link focus state on the ministers page.

 - [Live page](https://www.gov.uk/government/ministers)
 - [Local dev page](http://whitehall-admin.dev.gov.uk/government/ministers)

[Trello ticket](https://trello.com/c/oYiSsLBg/).

## Notice before
To see this, you need to set `@is_during_reshuffle`in `ministerial_roles_controller.rb` to `true` .

Screenshot 2019-11-11 at 17 13 57

## Notice after
Screenshot 2019-11-11 at 16 52 18

---

## Introduction before
image

## Introduction after
image

---
## Person partial before

With an image:

image

and without an image:

image

## Person partial after

With an image (image not loaded because development environment):

image

Without an image:

image

---

## Organisation logo before

image

## Organisation logo after

image


app/assets/stylesheets/frontend/global/_aspect-ratio.scss
```diff
@@ -0,0 +1,20 @@
+// This class should be used on an element that wraps an image to prevent jank
+// (aka that thing where the page jumps around because the image has loaded and
+// now takes up space). For much more information (and a couple of other
+// techniques) take a look at https://css-tricks.com/aspect-ratio-boxes/.
+.aspect-ratio-3\:2 {
+  background-color: govuk-colour("mid-grey", #f3f2f1);
```
maxgds: Query - why a grey background? Most of (all?) our backgrounds are white, so if an image fails to load will the alt text have enough contrast?

Gemfile
```diff
@@ -19,15 +19,14 @@ gem "globalize", "~> 5"
 gem "govuk_ab_testing", "~> 2.4x"
 gem "govuk_admin_template", "~> 6.7"
 gem "govuk_app_config", "~> 2.0"
-gem "govuk_publishing_components", "~> 21.11.0"
+gem "govuk_publishing_components", "~> 21.3.0"
```
injms: It needs at least [`21.11.0`](https://github.com/alphagov/govuk_publishing_components/blob/master/CHANGELOG.md#21130), as the organisation logos need to be wrapped in a heading rather than a `div`

app/assets/stylesheets/frontend/global/_aspect-ratio.scss
injms: It's a grey background because I was making sure that the image took up the correct amount of space (since images weren't working on the dev environment), and mistakenly left it in. Good catch - thanks!

PR #5175, opened 2019-11-22
Add content ids to governments

cbaines:
In preparation for publishing them to the Publishing API.


cbaines:
> Should we add a default to this field, or will Whitehall be changed to set a `content_id` when a new government is created?

I haven't quite figured out the how, but I'll look next at ensuring new governments get content ids, as well as making the field NON NULL in the database.

PR #5176, opened 2019-11-22
Update features when document migrated away from Whitehall

brucebolt:
When a document is marked as being migrated away from Whitehall, we need to update all featured document links to point to the public GOV.UK URL rather than referencing a document in Whitehall.

Trello card: https://trello.com/c/Fyt9he0f

## Featured document before migration
before-screenshot-whitehall-admin integration publishing service gov uk-2019 11 29-11-14-26

## Feature document after migration
after-screenshot-whitehall-admin integration publishing service gov uk-2019 11 29-11-21-56

```
#<OffsiteLink:0x00007f8d308937e8
 id: 2902,
 title: "Kylo Ren, monster and murderer?",
 summary:
  "The life and times of Kylo Ren the new Supreme Leader of the First Order",
 url:
  "https://www.integration.publishing.service.gov.uk/government/news/kylo-ren-monster-and-murderer",
 link_type: "content_publisher_news_story",
 parent_id: 1056,
 parent_type: "Organisation",
 date: Fri, 29 Nov 2019 11:21:30 GMT +00:00,
 created_at: Fri, 29 Nov 2019 11:21:30 GMT +00:00,
 updated_at: Fri, 29 Nov 2019 11:21:30 GMT +00:00>
```

### DIFF
--- a/lib/content_publisher/featured_document_migrator.rb
+++ b/lib/content_publisher/featured_document_migrator.rb
@@ -1,0 +1,31 @@
+module ContentPublisher
+  class FeaturedDocumentMigrator
+    attr_reader :document
+
+    def initialize(document)
+      @document = document
+    end
+
+    def call
+      edition = document.published_edition.present? ? document.published_edition : document.latest_edition
+
+      public_url = Whitehall.url_maker.public_document_url(edition)
+      public_updated_at = (edition.public_timestamp || edition.updated_at)
+
+      Feature.includes(:feature_list).where(document_id: document.id).each do |feature|
+        offsite_link = OffsiteLink.create!(title: edition.title,
+                                          summary: edition.summary,
+                                          link_type: "content_publisher_#{link_type(edition)}",
+                                          url: public_url,
+                                          date: public_updated_at,
+                                          parent_type: feature.feature_list.featurable_type,
+                                          parent_id: feature.feature_list.featurable_id)
+        feature.update!(document_id: nil, offsite_link_id: offsite_link.id)
+      end
+    end
+
+    def link_type(edition)
+      edition.news_article_type.key
+    end
+  end
+end
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -34,6 +34,8 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
     document.editions.each do |edition|
       Whitehall::InternalLinkUpdater.new(edition).call
     end
+
+    ContentPublisher::FeaturedDocumentMigrator.new(document).call
   end
 
   private
--- a/test/functional/admin/export/document_controller_test.rb
+++ b/test/functional/admin/export/document_controller_test.rb
@@ -146,4 +146,19 @@ class Admin::Export::DocumentControllerTest < ActionController::TestCase
     mock = MiniTest::Mock.new
     mock.expect :call, nil, [Whitehall::InternalLinkUpdater]
   end
+
+  test "calls FeaturedDocumentMigrator when document is marked as migrated" do
+    edition = create(:published_news_article)
+    create(:feature, document: edition.document, feature_list: create(:feature_list, locale: "en"))
+    edition.document.update(slug: "some-document", locked: true)
+
+    login_as :export_data_user
+
+    assert_difference -> { OffsiteLink.count } do
+      post :migrated, params: { id: edition.document.id }, format: "json"
+    end
+
+    assert_response :success
+    assert_equal "content_publisher_press_release", OffsiteLink.last.link_type
+  end
 end
--- a/test/unit/content_publisher/featured_document_migrator_test.rb
+++ b/test/unit/content_publisher/featured_document_migrator_test.rb
@@ -1,0 +1,60 @@
+require "test_helper"
+require "content_publisher/featured_document_migrator"
+
+module ContentPublisher
+  class FeaturedDocumentMigratorTest < ActiveSupport::TestCase
+    test "updates feature with offsite url" do
+      edition = create(:published_news_article)
+      feature = create(:feature, document: edition.document, feature_list: feature_list)
+      edition.document.update(locked: true)
+
+      ContentPublisher::FeaturedDocumentMigrator.new(edition.document).call
+      feature.reload
+
+      assert_nil feature.document_id
+      assert_equal "https://www.test.gov.uk/government/news/news-title", feature.offsite_link.url
+      assert_equal edition.public_timestamp, feature.offsite_link.date
+    end
+
+    test "uses the latest edition if document is not published" do
+      edition = create(:news_article)
+      feature = create(:feature, document: edition.document, feature_list: feature_list)
+      edition.document.update(locked: true)
+
+      ContentPublisher::FeaturedDocumentMigrator.new(edition.document).call
+      feature.reload
+
+      assert_nil feature.document_id
+      assert_equal "https://www.test.gov.uk/government/news/news-title", feature.offsite_link.url
+      assert_equal edition.updated_at, feature.offsite_link.date
+    end
+
+    test "sets the correct link type for press_release" do
+      edition = create(:published_news_article)
+      feature = create(:feature, document: edition.document, feature_list: feature_list)
+      edition.document.update(locked: true)
+
+      ContentPublisher::FeaturedDocumentMigrator.new(edition.document).call
+      feature.reload
+
+      assert_nil feature.document_id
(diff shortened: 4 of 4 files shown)


lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+      return unless valid_link_type?(latest_published_edition)
+
+      features.each do |feature|
+        offsite_link = OffsiteLink.create(title: latest_published_edition.title,
+                                          summary: latest_published_edition.summary,
+                                          link_type: "content_publisher_#{link_type(latest_published_edition)}",
+                                          url: public_url,
+                                          date: Time.zone.now,
+                                          parent_type: feature.feature_list.featurable_type,
+                                          parent_id: feature.feature_list.featurable_id)
```
kevindew: As this is a polymorphic association it might be easier to read as parent: `feature.feature_list.featurable` though this could come at expense of an additional DB query.

lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+    def update_all_features(document)
+      features = Feature.where(document_id: document.id)
+      latest_published_edition = document.editions.latest_published_edition.first
+      return unless latest_published_edition
+
+      public_url = Whitehall.url_maker.public_document_url(latest_published_edition)
+
+      return unless valid_link_type?(latest_published_edition)
+
+      features.each do |feature|
```
kevindew: Should we be doing a join on feature_list here to avoid an n+1 on feature_list?

lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+        offsite_link = OffsiteLink.create(title: latest_published_edition.title,
+                                          summary: latest_published_edition.summary,
+                                          link_type: "content_publisher_#{link_type(latest_published_edition)}",
+                                          url: public_url,
+                                          date: Time.zone.now,
```
kevindew: Do we know what this date is? It seems a bit strange to be setting it to current time so wonder if there's an existing date we should use?

lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def call
+      update_all_features(document)
+    end
+
+  private
+
+    def update_all_features(document)
+      features = Feature.where(document_id: document.id)
+      latest_published_edition = document.editions.latest_published_edition.first
```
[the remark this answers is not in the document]
leenagupte: Don't we need to use the published edition so that we can be sure that the external link will work?

lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+      return unless latest_published_edition
+
+      public_url = Whitehall.url_maker.public_document_url(latest_published_edition)
+
+      return unless valid_link_type?(latest_published_edition)
```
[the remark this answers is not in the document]
leenagupte: `valid_link_type?` checks a sub-set of the link types. At the moment only News Article types are allowed.

lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+                                          url: public_url,
+                                          date: Time.zone.now,
+                                          parent_type: feature.feature_list.featurable_type,
+                                          parent_id: feature.feature_list.featurable_id)
+        feature.update(document_id: nil, offsite_link_id: offsite_link.id)
+      end
+    end
+
+    def valid_link_type?(edition)
+      edition.news_article_type_id && [1, 2].include?(edition.news_article_type_id)
```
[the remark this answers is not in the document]
leenagupte: Wasn't sure how else to limit the check to only allow news_story and press_release in the allowed link types.

lib/whitehall/featured_document_updater.rb
leenagupte: Hmmm. Not sure, I'll take a look.

lib/whitehall/featured_document_updater.rb
kevindew: Looking at an organisations page this seems to be the public updated at of the content.

lib/whitehall/featured_document_updater.rb
```diff
@@ -0,0 +1,44 @@
...
+    def call
+      update_all_features(document)
+    end
+
+  private
+
+    def update_all_features(document)
+      features = Feature.where(document_id: document.id)
+      latest_published_edition = document.editions.latest_published_edition.first
+      return unless latest_published_edition
```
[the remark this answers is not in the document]
leenagupte: I've changed the check to use the `latest_edition` if the document has never been published so this guard isn't needed anymore.

leenagupte:
@kevindew Thanks for your comments. I've addressed them and they should be ready for a re-review. I've moved the class from `Whitehall::FeaturedDocumentUpdater` to `ContentPublisher::FeaturedDocumentMigrator` and added an integration test 🤞

lib/content_publisher/featured_document_migrator.rb
```diff
@@ -0,0 +1,32 @@
...
+      @document = document
+    end
+
+    def call
+      edition = document.published_edition.present? ? document.published_edition : document.latest_edition
+
+      public_url = Whitehall.url_maker.public_document_url(edition)
+      public_updated_at = (edition.public_timestamp || edition.updated_at)
+
+      features = Feature.includes(:feature_list).where(document_id: document.id)
```
kevindew: Might want to use `includes(feature_list: :featurable)` as we probably still have an n+1 going on for the parent.

Or we could switch back to parent_type and parent_id fields passed in, as they won't need a query.

lib/content_publisher/featured_document_migrator.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      features.each do |feature|
+        offsite_link = OffsiteLink.create(title: edition.title,
+                                          summary: edition.summary,
+                                          link_type: "content_publisher_#{link_type(edition)}",
+                                          url: public_url,
+                                          date: public_updated_at,
+                                          parent: feature.feature_list.featurable)
+        feature.update!(document_id: nil, offsite_link_id: offsite_link.id)
```
kevindew: Should this be `document: nil, offsite_link: offsite_link` rather than individual ids? - I don't mind too much so long as it's consistent with the parent call.

test/integration/content_publisher_migration_integration_test.rb
```diff
@@ -0,0 +1,19 @@
...
+    user = create(:export_data_user)
+    login_as(user)
+
+    assert_difference -> { OffsiteLink.count } do
+      post migrated_admin_export_document_path(id: edition.document.id)
+    end
+
+    assert_response :success
+    assert_equal "content_publisher_press_release", OffsiteLink.last.link_type
+  end
```
kevindew: Can we move this test into the controller one to replace the mock?

leenagupte:
@kevindew Thanks for the comments. I've addressed them and fixed-up.

lib/content_publisher/featured_document_migrator.rb
```diff
@@ -0,0 +1,33 @@
...
+    def call
+      edition = document.published_edition.present? ? document.published_edition : document.latest_edition
+
+      public_url = Whitehall.url_maker.public_document_url(edition)
+      public_updated_at = (edition.public_timestamp || edition.updated_at)
+
+      features = Feature.includes(:feature_list).where(document_id: document.id)
+
+      features.each do |feature|
+        offsite_link = OffsiteLink.create(title: edition.title,
```
kevindew: oh you missed the bang here

lib/content_publisher/featured_document_migrator.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def call
+      edition = document.published_edition.present? ? document.published_edition : document.latest_edition
+
+      public_url = Whitehall.url_maker.public_document_url(edition)
+      public_updated_at = (edition.public_timestamp || edition.updated_at)
+
+      features = Feature.includes(:feature_list).where(document_id: document.id)
+
+      features.each do |feature|
```
kevindew: And if you're editing this you might find this is nicer as `Feature.includes(:feature_list).where(document_id: document.id).each` to save a line - but no biggie.

lib/content_publisher/featured_document_migrator.rb
leenagupte: Argh!

leenagupte:
@kevindew Updated and fixed up 🤞

PR #5178, opened 2019-11-22
Present governments to the Publishing API

cbaines:
This is in preparation for linking content to these governments in the Publishing API, and then using this information in Government Frontend.


cbaines:
This depends on https://github.com/alphagov/govuk-content-schemas/pull/935

PR #5179, opened 2019-11-22
Ensure governments have content ids

cbaines:
The previous data migration generates content_id's for all governments, but these changes ensure that governments have content_id's going forward.


cbaines:
I'm hoping the tests will pass here once https://github.com/alphagov/whitehall/pull/5178 is merged.

PR #5183, opened 2019-11-27
Create rake task to link editions to a government

benjamineskola:
Iterates over all published and drafted editions (superseded ones shouldn't be seen), then creates a presenter and uses that to patch the edition's links in the publishing API.

https://trello.com/c/7yyMiJwi/1574-link-all-relevant-whitehall-content-to-the-appropriate-governments-in-the-publishing-api

### DIFF
--- a/lib/tasks/link_governments.rake
+++ b/lib/tasks/link_governments.rake
@@ -1,0 +1,24 @@
+namespace :governments do
+  task relink: :environment do
+    STDOUT.sync = true
+    Edition.where.not(state: %w[deleted superseded]).find_each do |edition|
+      presenter = PublishingApiPresenters.presenter_for(edition)
+      begin
+        links = presenter.links
+
+        if links[:government].present?
+          print "."
+
+          Services.publishing_api.patch_links(
+            presenter.content_id,
+            links: links.slice(:government),
+            bulk_publishing: true,
+          )
+        end
+      rescue StandardError => e
+        puts "\nFAIL #{e.class} (#{e.message}) for document: #{presenter.content_id}"
+      end
+    end
+    puts
+  end
+end


lib/tasks/link_governments.rake
```diff
@@ -0,0 +1,13 @@
+namespace :governments do
+  task relink: :environment do
+    Edition.where(state: %w[draft published]).each do |edition|
```
thomasleese:
```suggestion
    Edition.where(state: %w[draft published]).find_each do |edition|
```

Using `find_each` can improve performance as it avoids needing to instance a new object for each record (the same instance is re-used multiple times).

lib/tasks/link_governments.rake
```diff
@@ -0,0 +1,13 @@
+namespace :governments do
```
thomasleese: It might be good to have this task in the `publishing_api` namespace as we have a few tasks in there already which interact with the Publishing API.

lib/tasks/link_governments.rake
```diff
@@ -0,0 +1,13 @@
...
+  task relink: :environment do
+    Edition.where(state: %w[draft published]).each do |edition|
+      presenter = PublishingApiPresenters.presenter_for(edition)
+      links = presenter.links
+      if links[:government].present?
+        print '.'
+        Services.publishing_api.patch_links(presenter.content_id, links: links.slice(:government), bulk_publishing: true)
```
thomasleese: I would suggest splitting this line up into multiple lines just because it's a bit long, something like:

```rb
Services.publishing_api.patch_links(
  presenter.content_id,
  links: links.slice(:government),
  bulk_publishing: true
)
```

PR #5185, opened 2019-11-27
Add Brexit CTA

emmabeynon:
For https://trello.com/c/RosX23YC/199-add-brexit-callout-element-to-whitehall

This adds support to convert `$BrexitCTA` 'markdown' to a CTA with static messaging.  The intention is that when the messaging needs to change, a developer with update the static messaging and republish content which contains the markdown (to be covered in a future PR).

**Body input**
brexit-cta-body-input

**Brexit CTA body preview**
brexit-cta-preview

_See https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/detailed-guides/1016488/edit for comparison_

**Frontend representation**
brexit-cta-frontend

_See https://www.gov.uk/guidance/driving-in-the-eu-after-brexit for comparison_

**Govspeak help sidebar**
2019-11-28-09-41-whitehall-admin dev gov uk

### DIFF
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -126,6 +126,7 @@ private
     govspeak = render_embedded_fractions(govspeak)
     govspeak = set_classes_for_charts(govspeak)
     govspeak = set_classes_for_sortable_tables(govspeak)
+    govspeak = convert_brexit_cta(govspeak)
 
     markup_to_nokogiri_doc(govspeak, images)
       .tap { |nokogiri_doc|
@@ -260,6 +261,14 @@ private
       else
         ""
       end
+    end
+  end
+
+  def convert_brexit_cta(govspeak)
+    return govspeak if govspeak.blank?
+
+    govspeak.gsub(/\$BrexitCTA/) do
+      render(partial: "documents/brexit_cta", formats: :text)
     end
   end
 
--- a/app/views/admin/editions/_govspeak_help.html.erb
+++ b/app/views/admin/editions/_govspeak_help.html.erb
@@ -198,6 +198,11 @@
   $CTA</pre>
   </div>
 
+  <h3><a data-toggle="collapse" href="#brexit-cta" aria-expanded="false">Brexit call to action</a></h3>
+  <div class="collapse" id="brexit-cta">
+    <pre>$BrexitCTA</pre>
+  </div>
+
   <h3><a data-toggle="collapse" href="#govspeak-acronyms" aria-expanded="false">Abbreviations and acronyms</a></h3>
   <div class="collapse" id="govspeak-acronyms">
     <p>The first time you use an abbreviation or acroynm, write it out in full.</p>
--- a/test/unit/helpers/govspeak_helper_test.rb
+++ b/test/unit/helpers/govspeak_helper_test.rb
@@ -475,4 +475,19 @@ class GovspeakHelperTest < ActionView::TestCase
     html = govspeak_with_attachments_to_html(body, attachments, "batman@wayne.technology")
     assert html.include? ">batman@wayne.technology</a>"
   end
+
+  test "converts Brexit CTA govspeak to HTML" do
+    body = "$BrexitCTA\nSome other text"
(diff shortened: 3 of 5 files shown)


app/views/admin/editions/_govspeak_help.html.erb
```diff
@@ -198,6 +198,11 @@
   $CTA</pre>
   </div>
 
+  <h3><a data-toggle="collapse" href="#brexit-cta" aria-expanded="false">Brexit call to action</a></h3>
```
huwd: @vanitabarrett thinking about the stuff we had with the details component. Is there any other tracking stuff we'd expect to have bundled with this?

vanitabarrett: Look at the other dropdowns, it doesn't look like we're currently tracking these links so it should be fine as is.

huwd: Ace!

vanitabarrett:
@emmabeynon This change looks good to me 👍 Do we need to add any kind of tracking to the links in the callout itself? I know there was previously some talk about that.

emmabeynon:
@vanitabarrett Unsure at the moment.  Do you happen to know if we have any precedent for using tracking with Govspeak?

vanitabarrett:
@emmabeynon I don't think we can add tracking attributes using the normal markdown syntax for links, but we can use HTML for the link instead and add data-attributes there for the tracking.

E.g:

```
<a href="/email-signup?topic=%2Fbrexit" data-module="track-click" data-track-action="blah">Sign up for email alerts</a>
```

app/helpers/govspeak_helper.rb
```diff
@@ -263,6 +264,18 @@ def govspeak_with_attachments_and_alt_format_information(govspeak, attachments =
     end
   end
 
+  def convert_brexit_cta(govspeak)
+    return govspeak if govspeak.blank?
+
+    govspeak.gsub(/\$BrexitCTA/) do
+      "$CTA\n" +
```
koetsier: Could we put the html content in a separate file? That would make it easier to find and edit at a later date. 

Not sure where to put it - I was thinking maybe we could render it as normal view template? Let me know what you think

test/unit/helpers/govspeak_helper_test.rb
```diff
@@ -475,4 +475,19 @@ class GovspeakHelperTest < ActionView::TestCase
     html = govspeak_with_attachments_to_html(body, attachments, "batman@wayne.technology")
     assert html.include? ">batman@wayne.technology</a>"
   end
+
+  test "converts Brexit CTA govspeak to HTML" do
+    body = "$BrexitCTA\nSome other text"
+    output = govspeak_to_html(body)
+    paragraph_text = "The UK is leaving the EU. This page tells you how to "\
```
koetsier: This is quite coupled to the callout box as it is now. If that is changed this test breaks. If we extract the html in a file, we could check if the govspeak gets replaced by the contents of that file

app/helpers/govspeak_helper.rb
emmabeynon: Sure - there is some precedent for the helpers to render a HTML view.  We could probably create a markdown view?

test/unit/helpers/govspeak_helper_test.rb
emmabeynon: This is a tricky one, because the output of `govspeak_to_html` is HTML, whereas as the content of the file would be markdown, so I think we will need to check for some HTML.  If the content of the CTA does change I don't think it would be too much effort to update the test.

app/helpers/govspeak_helper.rb
koetsier: ah yes of course its markdown - hadn't had my coffee yet. Maybe a markdown view? If you can think of a better place to put it then that would be great

test/unit/helpers/govspeak_helper_test.rb
koetsier: true, it was a minor thing - hopefully the line you're checking won't change 👍

PR #5187, opened 2019-11-29
Add rake task to send political content to search index

bevanloon:
If the government changes, we currently republish all political content
so that the banner stating that the content was published under a
previous government is displayed appropriately.

According to alphagov/search-api@ee2f8d8,  we also need to send 
these editions to search so they can be re-indexed.

### DIFF
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -104,6 +104,13 @@ namespace :search do
     end
   end
 
+  desc "Update search index for all political content"
+  task political: :environment do
+    Edition
+      .where(political: true, state: "published")
+      .find_each { |edition| edition&.update_in_search_index }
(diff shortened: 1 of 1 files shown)


lib/tasks/search.rake
```diff
@@ -104,6 +104,13 @@ namespace :search do
     end
   end
 
+  desc "Update search index for all political content"
+  task political: :environment do
+    Edition
+      .where(political: true, state: "published")
+      .find_each { |edition| edition&.update_in_search_index }
```
thomasleese: I found it confusing that the `&.` operator applies to the right hand side as well as the left hand side! But since it's just `edition.try(:update_in_search_index)` I suppose that makes sense.

PR #5188, opened 2019-11-29
Enable rails-specific rules for rubocop-govuk

benjamineskola:
trello: https://trello.com/c/Lijki8aj/1584-3-enable-rails-rubocop-rules-via-rubocop-govuk-on-all-our-rails-apps


issyl0:
This has conflicts.

PR #5190, opened 2019-11-29
Use non whitehall link for migrated docs

leenagupte:
Trello: https://trello.com/c/9jaHpTwx

## What's changed and why?

When a document is marked as being migrated in Whitehall we need to change content that has pointers to the Whitehall local data and switch this to be something Whitehall agnostic.

One of the places we have to do this is document collections. If a document collection links to a document that has been migrated, change this link to be non-Whitehall link.

### Before
before-screenshot-whitehall-admin integration publishing service gov uk-2019 12 02-11-47-00

### After
after-screenshot-whitehall-admin integration publishing service gov uk-2019 12 02-11-50-08

### DIFF
--- a/lib/content_publisher/document_collection_group_membership_migrator.rb
+++ b/lib/content_publisher/document_collection_group_membership_migrator.rb
@@ -9,6 +9,8 @@ module ContentPublisher
     def call
       edition = document.published_edition || document.latest_edition
 
+      return unless DocumentCollectionGroupMembership.exists?(document_id: document.id)
+
       non_whitehall_link = DocumentCollectionNonWhitehallLink.create!(
         content_id: document.content_id,
         title: edition.title,
--- a/lib/content_publisher/document_collection_group_membership_migrator.rb
+++ b/lib/content_publisher/document_collection_group_membership_migrator.rb
@@ -1,0 +1,27 @@
+module ContentPublisher
+  class DocumentCollectionGroupMembershipMigrator
+    attr_reader :document
+
+    def initialize(document)
+      @document = document
+    end
+
+    def call
+      edition = document.published_edition || document.latest_edition
+
+      non_whitehall_link = DocumentCollectionNonWhitehallLink.create!(
+        content_id: document.content_id,
+        title: edition.title,
+        base_path: Whitehall.url_maker.public_document_path(edition),
+        publishing_app: "content-publisher",
+      )
+
+      DocumentCollectionGroupMembership
+        .where(document_id: document.id)
+        .update_all(
+          document_id: nil,
+          non_whitehall_link_id: non_whitehall_link.id,
+        )
+    end
+  end
+end
--- a/test/functional/admin/export/document_controller_test.rb
+++ b/test/functional/admin/export/document_controller_test.rb
@@ -161,4 +161,19 @@ class Admin::Export::DocumentControllerTest < ActionController::TestCase
     assert_response :success
     assert_equal "content_publisher_press_release", OffsiteLink.last.link_type
   end
+
+  test "calls DocumentCollectionGroupMembershipMigrator when document is marked as migrated" do
+    edition = create(:published_news_article)
+    create(:document_collection_group_membership, document: edition.document)
+    edition.document.update(slug: "some-document", locked: true)
+
+    login_as :export_data_user
+
+    assert_difference -> { DocumentCollectionNonWhitehallLink.count } do
+      post :migrated, params: { id: edition.document.id }, format: "json"
+    end
+
+    assert_response :success
+    assert_equal "content-publisher", DocumentCollectionNonWhitehallLink.last.publishing_app
+  end
 end
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -36,6 +36,7 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
     end
 
     ContentPublisher::FeaturedDocumentMigrator.new(document).call
+    ContentPublisher::DocumentCollectionGroupMembershipMigrator.new(document).call
   end
 
   private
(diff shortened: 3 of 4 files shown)


lib/whitehall/document_collection_group_membership_updater.rb
```diff
@@ -0,0 +1,28 @@
+module Whitehall
+  class DocumentCollectionGroupMembershipUpdater
+    attr_reader :document
+
+    def initialize(document)
+      @document = document
+    end
+
+    def call
+      return unless document.published_edition
```
kevindew: I think we should see what we can do to avoid aborting on this if there isn't a live edition.

lib/whitehall/document_collection_group_membership_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      document_collection_group_memberships = DocumentCollectionGroupMembership.where(document_id: document.id)
+
+      return unless document_collection_group_memberships.any?
+
+      non_whitehall_link = DocumentCollectionNonWhitehallLink.create(
```
kevindew: Should use the bang equivalent here

lib/whitehall/document_collection_group_membership_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+        content_id: document.content_id,
+        title: document.published_edition.title,
+        base_path: Whitehall.url_maker.public_document_path(document.published_edition),
+        publishing_app: "content-publisher",
```
kevindew: Should we have this app passed in as a parameter? It feels to be the only place where we refer directly to content publisher and seems a surprising part of this class.

lib/whitehall/document_collection_group_membership_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+      )
+
+      document_collection_group_memberships.each do |membership|
+        membership.update!(document_id: nil, non_whitehall_link_id: non_whitehall_link.id)
+      end
```
kevindew: Seems like a clever move to re-use the same non_whitehall_link model many times. Rather than doing a loop we can probably do an update_all?

Do you know if there are any risks with us re-using the same non_whitehall_link model. I don't think there is myself as I don't think there are any cases where we update the data in it, but wasn't sure if it gets auto deleted when a memebership is removed or anything.

lib/whitehall/document_collection_group_membership_updater.rb
```diff
@@ -0,0 +1,28 @@
+module Whitehall
+  class DocumentCollectionGroupMembershipUpdater
```
kevindew: I don't know if I'm opening a can of worms here but I wonder if the class name here captures what we're doing properly. I wonder if we want to cover that this is for a migration and/or one to content publisher. 

Maybe we should have some of these classes under a ContentPublisherMigration namespace? What do you think?

test/functional/admin/export/document_controller_test.rb
```diff
@@ -146,4 +146,16 @@ class Admin::Export::DocumentControllerTest < ActionController::TestCase
...
+  test "calls DocumentCollectionGroupMembershipUpdater when document is marked as migrated" do
+    edition_linked_to = create(:edition_with_document, body: "Some document being migrated to Content Publisher")
+    edition_linked_to.document.update(slug: "some-document", locked: true)
+    create(:edition_with_document)
+    login_as :export_data_user
+
+    post :migrated, params: { id: edition_linked_to.document.id }, format: "json"
+
+    mock = MiniTest::Mock.new
+    mock.expect :call, nil, [Whitehall::DocumentCollectionGroupMembershipUpdater]
```
kevindew: Sorry I know this is a precedent set in a previous test that you're following.

It feels a bit strange to be checking a mock when we're in a functional test here. Should we be using this to do an end-to-end test?

lib/whitehall/document_collection_group_membership_updater.rb
leenagupte: We talked about yesterday. It looks like nothing happens to the non whitehall link if the document collection group membership is deleted, but all associated document collection group memberships are deleted if non whitehall link is deleted.

leenagupte:
@kevindew Thanks for your comments. I've updated this PR so that it matches PR #5176, and addressed the other comments.

lib/content_publisher/document_collection_group_membership_migrator.rb
```diff
@@ -0,0 +1,28 @@
...
+    def initialize(document)
+      @document = document
+    end
+
+    def call
+      edition = document.published_edition.present? ? document.published_edition : document.latest_edition
+
+      document_collection_group_memberships = DocumentCollectionGroupMembership.where(document_id: document.id)
+
+      return unless document_collection_group_memberships.any?
```
kevindew:
```suggestion
      return unless DocumentCollectionGroupMembership.exists?(document_id: document.id)
```

You can probably lose the check via variable and call an exists directly on the model class.

lib/content_publisher/document_collection_group_membership_migrator.rb
```diff
@@ -0,0 +1,28 @@
...
+      non_whitehall_link = DocumentCollectionNonWhitehallLink.create!(
+        content_id: document.content_id,
+        title: edition.title,
+        base_path: Whitehall.url_maker.public_document_path(edition),
+        publishing_app: "content-publisher",
+      )
+
+      document_collection_group_memberships.each do |membership|
+        membership.update!(document_id: nil, non_whitehall_link_id: non_whitehall_link.id)
+      end
```
kevindew:
```suggestion
DocumentCollectionGroupMembership.where(document_id: document.id).update_all(document_id: nil, non_whitehall_link_id: non_whitehall_link.id)
```
We can do these lots of queries in a single query.

lib/content_publisher/document_collection_group_membership_migrator.rb
```diff
@@ -0,0 +1,28 @@
+module ContentPublisher
+  class DocumentCollectionGroupMembershipMigrator
+    attr_reader :document
+
+    def initialize(document)
+      @document = document
+    end
+
+    def call
+      edition = document.published_edition.present? ? document.published_edition : document.latest_edition
```
kevindew: minor but this could be simplified to `edition = document.published_edition || document.latest_edition`

PR #5201, opened 2019-12-10
Enable rails-specific rules for rubocop-govuk

benjamineskola:


issyl0:
I can't push this branch fresh to `origin` because there's already a branch with the same name. I'll cherry-pick the commits instead, I guess, so that CI continues to run?

benjamineskola:
Oh, I thought it was just going to be the same branch, but I'd forgotten I'd amended the commit. Not sure what the easiest solution is, then; it might be worth just waiting until I can push to the original branch/PR directly.

benjamineskola:
Duplicate of #5188

PR #5207, opened 2019-12-16
Remove migrated content from Whitehall search index

gclssvglx:
## User story

**As a** publisher
**I want** my content to be removed from Whitehall search index when it is migrated
**So that** the content doesn't appear in search twice

## What
We are adding a step to the migrated endpoint to remove the content for this document from the government index.

## Why

When content has been migrated from Whitehall to Content Publisher we need it to no longer exist in the Search API index for Whitehall content.

[Trello](https://trello.com/c/nJARKkob/1231-remove-content-from-whitehall-search-index)

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -33,6 +33,7 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
 
     document.editions.each do |edition|
       Whitehall::InternalLinkUpdater.new(edition).call
+      Whitehall::SearchIndex.delete(edition)
     end
 
     ContentPublisher::FeaturedDocumentMigrator.new(document).call
--- a/test/functional/admin/export/document_controller_test.rb
+++ b/test/functional/admin/export/document_controller_test.rb
@@ -126,6 +126,18 @@ class Admin::Export::DocumentControllerTest < ActionController::TestCase
     assert_response :no_content
   end
 
+  test "removes migrated document from search index" do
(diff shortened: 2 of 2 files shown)


app/controllers/admin/export/document_controller.rb
```diff
@@ -33,6 +33,7 @@ def migrated
 
     document.editions.each do |edition|
       Whitehall::InternalLinkUpdater.new(edition).call
+      Whitehall::SearchIndex.delete(edition)
```
brucebolt: It's likely only the most recent edition of the document will be present in the search index.  Since this is looping through all editions, do you know how `Whitehall::SearchIndex.delete` will behave for an edition that is not present in the search index?

.gitignore
```diff
@@ -24,3 +24,4 @@
 /doc/
 /failures
 .DS_Store
+.byebug_history
```
brucebolt: Could this be added into a separate commit to ensure the commit message ties up with the code change?

gclssvglx: Created a separate commit for this.

app/controllers/admin/export/document_controller.rb
gclssvglx: Hi Bruce,

We've tested this (via console rails c) and it appears to behave exactly the same if the edition exists or not.  The method call delegates out to a worker job (SearchIndexDeleteWorker) that is queued.

When run the worker job calls `Whitehall::SearchIndex.for(index.to_sym, logger: logger).delete(link)` which will do nothing (return nil) if the edition is not in the index.

Think we're safe!

PR #5209, opened 2019-12-18
Add h1 to stylesheets for worldwide organisation pages

JonathanHallam:
https://github.com/alphagov/whitehall/pull/5193/files#diff-bb26fcf8c43892d384807e5e8cae09b4L179
Stopped users from being able to use h1 tags within their text bodies. While this shouldn't be happening because it's not great from an accessibility front, there are too many pages that this effects for 2nd line to make changes to. We've created a card (below) that outlines the need to get content creators to remove h1 tags from text bodies. When that has been actioned these changes should be reverted.

Trello - https://trello.com/c/kzfGMNdH/800-look-into-replacing-h1s-with-h2s-on-worldwide-org-pages

### DIFF
--- a/app/assets/stylesheets/frontend/helpers/_govspeak.scss
+++ b/app/assets/stylesheets/frontend/helpers/_govspeak.scss
@@ -1,4 +1,5 @@
 .govspeak {
+  h1:first-child,
   h2:first-child,
   h3:first-child,
   h4:first-child,
@@ -6,6 +7,7 @@
     margin-top: 0;
   }
 
+  h1,
   h2 {
     @include ig-core-27;
     font-weight: bold;


kr8n3r:
this unfortunately won't have the desired affect as affected h1's don't sit in `.corporate-information` where code has been placed
suggest to add h1 styles here https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/helpers/_govspeak.scss

JonathanHallam:
thanks @kr8n3r and @alex-ju, will adjust.

injms:
I worry that even allowing this temporarily will give people the idea that this is okay, as it won't highlight that it's an issue for new pages and goes contrary to the guidance given for content.

alex-ju:
> I worry that even allowing this temporarily will give people the idea that this is okay, as it won't highlight that it's an issue for new pages and goes contrary to the guidance given for content.

We all agree with you @injms, we shouldn't have h1s in the content-store. The guidance tells editors not to use h1s and by not styling it properly means that if they do they will be forced to update to h2.

However, we have a huge amount of legacy documents that are using h1 incorrectly and until the content is updated to structure them properly (which is recorded and will be actioned in the card John mentioned) people will have a bad 'visual experience'.

I will track this down and make sure we're removing it as soon as the content is updated and we'll also raise this with Content Publisher to see what can be done to prevent people from using h1 from now on.

PR #5221, opened 2019-12-20
Stop sending people and roles to Search API

thomasleese:
These formats have been migrated so Search API is getting updates on the Publishing API message queue.

[Trello Card](https://trello.com/c/yYhQcqrQ/1651-stop-whitehall-sending-people-and-roles-to-search-api)


cbaines:
Adding Do not merge as it's probably wise to wait a little bit before merging.

thomasleese:
I've checked again now and roles and people seem to all be coming from `"index": "govuk"` so this is good to go I think!

PR #5224, opened 2019-12-23
Export content for the first lead organisation

AgaDufrat:
### User story
**As a** migrated department
**I want** to have all the documents that I own/publish in Content Publisher
**So that** I can easily find my documents. 

Some documents in Whitehall have multiple lead organisations while Content Publisher only allows one lead organisation. We need to migrate all documents where the target organisation is the first or only lead organisation. 

This change allows to export content for the first lead organisation. 

[Trello card](https://trello.com/c/zRBVbotR/1240-import-content-with-multiple-lead-organisations)

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -44,7 +44,14 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
 
   def paginated_document_ids
     Edition
-      .joins("INNER JOIN edition_organisations eo ON eo.edition_id = editions.id")
+      .joins("INNER JOIN edition_organisations eo ON eo.edition_id = editions.id
+        AND eo.organisation_id = (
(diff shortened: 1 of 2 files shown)


app/controllers/admin/export/document_controller.rb
```diff
@@ -44,7 +44,14 @@ def migrated
 
   def paginated_document_ids
     Edition
-      .joins("INNER JOIN edition_organisations eo ON eo.edition_id = editions.id")
+      .joins("INNER JOIN edition_organisations eo ON eo.edition_id = editions.id
+        AND eo.organisation_id = (
+          SELECT organisation_id FROM edition_organisations
+          WHERE lead = true AND
+                edition_organisations.edition_id = editions.id
+          ORDER BY lead_ordering ASC
```
brucebolt: Nice!

PR #5226, opened 2019-12-24
Work around an odd issue with atom feeds

cbaines:
Somehow, with recent versions of the govuk_publishing_components,
`.try(:document)` is calling a function defined in the
_document_list.html.erb partial in the govuk_publishing_components...

I don't quite understand why, so switch the code around to avoid
calling try on objects that don't respond to .document. In the case of
english responses, the record will be a RummagerDocumentPresenter,
which doesn't respond to .document, so just return it's id. In the
case of non english responses, the record is a presented edition,
which does respond to .document.

### DIFF
--- a/app/helpers/feed_helper.rb
+++ b/app/helpers/feed_helper.rb
@@ -26,10 +26,13 @@ module FeedHelper
     # spec.
     # The interpolation logic is straight out of:
     # http://api.rubyonrails.org/classes/ActionView/Helpers/AtomFeedHelper/AtomFeedBuilder.html#method-i-entry
-    id = record.try(:document) ? record.document.id : record.id
-    return id if record.is_a?(RummagerDocumentPresenter)
+    if record.is_a?(RummagerDocumentPresenter)
+      record.id
+    else
+      id = record.try(:document) ? record.document.id : record.id
 
-    "tag:#{host},#{schema_date(builder)}:#{record.class}/#{id}"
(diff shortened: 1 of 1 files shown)


cbaines:
So this Pull Request removed the function that was causing problems, but I still think this change is wise, as it'll prevent similar problems in the future: https://github.com/alphagov/govuk_publishing_components/pull/1244

app/helpers/feed_helper.rb
```diff
@@ -26,10 +26,13 @@ def document_id(record, builder)
     # spec.
     # The interpolation logic is straight out of:
     # http://api.rubyonrails.org/classes/ActionView/Helpers/AtomFeedHelper/AtomFeedBuilder.html#method-i-entry
-    id = record.try(:document) ? record.document.id : record.id
-    return id if record.is_a?(RummagerDocumentPresenter)
+    if record.is_a?(RummagerDocumentPresenter)
+      record.id
+    else
+      id = record.try(:document) ? record.document.id : record.id
```
thomasleese:
```suggestion
      id = record&.document&.id || record.id
```

I wonder if this might be easier to read.

cbaines: It's definitely shorter. As with all Ruby, I'm not entirely sure what the semantics are to begin with, let alone if they're equivalent, but let's go with it.

PR #5237, opened 2020-01-07
Fix inline attachment rendering

vanitabarrett:
## What
Removes trailing newlines from inline attachment view. There were two things causing the additional space:

- The closing span tag being on a newline
- The inline_attachment file having a newline at the end of the file (part of our standards)

## Why
- Inline attachments were rendering with an additional space after the inline attachment.
- Inline attachments in numbered lists displayed on a new line and showed a closing `</span>` tag on the page

## Before
Screenshot 2020-01-07 at 15 30 11

## After
Screenshot 2020-01-07 at 15 29 57

## Before
Screenshot 2020-01-07 at 15 52 18

## After
Screenshot 2020-01-07 at 15 45 57

### DIFF
--- a/app/views/documents/_inline_attachment.html.erb
+++ b/app/views/documents/_inline_attachment.html.erb
@@ -1,4 +1,1 @@
-<span id="attachment_<%= attachment.id %>" class="attachment-inline">
-  <%= link_to attachment.title, attachment.url %>
-  (<%= attachment_attributes(attachment) %>)
-</span>
+<span id="attachment_<%= attachment.id %>" class="attachment-inline"><%= link_to attachment.title, attachment.url %> (<%= attachment_attributes(attachment) %>)</span>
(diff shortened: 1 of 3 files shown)


app/views/documents/_inline_attachment.html.erb
```diff
@@ -1,4 +1 @@
-<span id="attachment_<%= attachment.id %>" class="attachment-inline">
-  <%= link_to attachment.title, attachment.url %>
-  (<%= attachment_attributes(attachment) %>)
-</span>
+<span id="attachment_<%= attachment.id %>" class="attachment-inline"><%= link_to attachment.title, attachment.url %> (<%= attachment_attributes(attachment) %>)</span>
```
huwd: In some ways this is a bit of a shame to put this on one line.
I find it less readable than what was there previously.

But i guess you need it this way as spans preserve line breaks?
🤔 and a helper seems OTT...

vanitabarrett: Yeah, I tried just using `chomp` and putting it back over multiple lines, but both contribute to the additional space, and it's such a small file that anything else seemed like overkill.

PR #5239, opened 2020-01-08
Redirect UKNCP page to new location

karlbaker02:
This PR redirects the UKNCP page (https://www.gov.uk/government/groups/uk-national-contact-point-for-the-organisation-for-economic-co-operation-and-development-guidelines) to the new location (https://www.gov.uk/government/organisations/uk-national-contact-point), as part of Zendesk ticket 3885814.

### DIFF
--- a/db/data_migration/20200108171705_redirect_ukncp_organisation_page.rb
+++ b/db/data_migration/20200108171705_redirect_ukncp_organisation_page.rb
@@ -1,0 +1,6 @@
+# Redirect UKNCP to new location
+
+content_id = "1bc90138-9b23-46ca-8a8f-57c6cc50c9b1"
+destination = "/government/organisations/uk-national-contact-point"
+
+PublishingApiRedirectWorker.new.perform(content_id, destination, "en")


db/data_migration/20200108171705_redirect_ukncp_organisation_page.rb
```diff
@@ -0,0 +1,6 @@
+# Redirect UKNCP to new location
+
+content_id = "1bc90138-9b23-46ca-8a8f-57c6cc50c9b1"
+destination = "/government/organisations/uk-national-contact-point"
+
+PublishingApiRedirectWorker.new.perform(content_id, destination, I18n.default_locale.to_s)
```
ChrisBAshton: Oops, I hit "review" too early 😢 

One thing I would say is we should hardcode the locale to "en", as we know the content is in English. This is how it has been done in [other](https://github.com/alphagov/whitehall/blob/master/db/data_migration/20190819114211_redirect_the_high_way.rb) [examples](https://github.com/alphagov/whitehall/blob/master/db/data_migration/20191001120013_redirect_right_to_rent_3_step.rb), and protects us if our default locale changes in the future. Not a blocking comment though.

karlbaker02:
Thank you @ChrisBAshton

PR #5241, opened 2020-01-10
Add a scope for filtering on link report links

thomasleese:
This allows us to filter on link report links which can be deleted. The link table has exceeded 10 GB and is the biggest Whitehall table so we need to find a way of reducing the size. We've decided that links with a status of `ok` which were created over 6 months ago can be deleted.

Once this is deployed I will manually delete these links by doing `Link.deletable.delete_all` and then after that set up some sort of automated process to do this regularly.

[Trello Card](https://trello.com/c/akZkqlUE/1669-5-investigate-why-linkcheckerapireportlinks-table-in-the-whitehall-database-is-very-large)

### DIFF
--- a/app/models/link_checker_api_report/link.rb
+++ b/app/models/link_checker_api_report/link.rb
@@ -4,6 +4,10 @@ class LinkCheckerApiReport::Link < ApplicationRecord
 
   belongs_to :report, class_name: "LinkCheckerApiReport"
 
+  scope :status_ok, -> { where(status: "ok") }
+  scope :created_six_months_ago, -> { where("created_at < ?", 6.months.ago) }
+  scope :deletable, -> { self.status_ok.merge(self.created_six_months_ago) }
+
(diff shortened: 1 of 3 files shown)


app/models/link_checker_api_report/link.rb
```diff
@@ -4,6 +4,10 @@ class LinkCheckerApiReport::Link < ApplicationRecord
 
   belongs_to :report, class_name: "LinkCheckerApiReport"
 
+  scope :status_ok, -> { where(status: "ok") }
+  scope :created_six_months_ago, -> { where("created_at < ?", 6.months.ago) }
```
issyl0: Is six months enough? I'd go even shorter: three.

thomasleese: I wanted to be cautious at the beginning to see what sort of space savings we get with six months. We could certainly reduce it to three in the future if we feel that six months is too long.

PR #5243, opened 2020-01-10
Upgrade Ruby to 2.7.0

issyl0:
Upgrade Ruby to 2.7.0, see commits for more details.

            https://trello.com/c/XoJF36Pn/1663-8-upgrade-to-ruby-27


issyl0:
Closing. Sorry for the spam across all the apps!

It turns out that there's [a key deprecation](ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) which is causing a load of errors in gems that we use. The gem authors (including [Rails](https://github.com/rails/rails/pulls?utf8=%E2%9C%93&q=is%3Apr+2.7)) are working hard to fix their software, but there's no point us using it until they've done that or we'll get a tonne of warnings (and turning them off is bad practice). Also, it's possible that Rails may only put their fixes into Rails 6 - and in most of the cases we're still running Rails 5.

Platform Health is going back to the drawing board on this one, because it may be that we want to upgrade every app to Rails 6 first.

PR #5247, opened 2020-01-14
Add a rake task for deleting old report links

thomasleese:
This table has the potential to grow indefinitely, so we will configure this task to run periodically so it remains a good size. It also always us to delete the links using Jenkins so we have better visibility on the job.

This follows on from https://github.com/alphagov/whitehall/pull/5241.

[Trello Card](https://trello.com/c/akZkqlUE/1669-5-investigate-why-linkcheckerapireportlinks-table-in-the-whitehall-database-is-very-large)

### DIFF
--- a/lib/tasks/link_checker.rake
+++ b/lib/tasks/link_checker.rake
@@ -1,0 +1,6 @@
+namespace :link_checker do
+  task delete_old_report_links: :environment do
+    count = LinkCheckerApiReport::Link.deletable.delete_all
+    puts "Deleted #{count} old report links."
+  end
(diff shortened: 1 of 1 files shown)


AlanGabbianelli:
Are we thinking of setting a cron job for this or so we have to run it manually?

thomasleese:
@AlanGabbianelli Yes, I'm going to add it to [`config/schedule.rb`](https://github.com/alphagov/whitehall/blob/master/config/schedule.rb) once I've run the Rake task one more time in production.

PR #5252, opened 2020-01-15
Update colours

maxgds:
https://trello.com/c/t9vZxNXX/124-switch-to-new-colours-in-whitehall

This PR mops up a few outstanding issues related to focus, and switches to the new colours. 

The branch is currently deployed to integration as it won't run locally, so worth looking there (although another deploy might be needed if anything else goes up meanwhile)

Some links on https://www.gov.uk/world/embassies weren't getting the right focus state as they were missing the link class, this was only visible with the colour change:
### Before
Screenshot 2020-01-15 at 11 09 54
### After
Screenshot 2020-01-15 at 11 09 24

World locations pages had a dropdown that was hard to read generally and didn't take the right focus state:
### Before
Screenshot 2020-01-15 at 11 12 38

### After
Screenshot 2020-01-15 at 11 12 13

World Locations list page had an input box that didn't take the correct focus state
https://www.gov.uk/world

### Before
Screenshot 2020-01-15 at 11 15 36

### After

Screenshot 2020-01-15 at 11 15 17

### DIFF
--- a/app/assets/stylesheets/frontend/base.scss
+++ b/app/assets/stylesheets/frontend/base.scss
@@ -1,6 +1,6 @@
 // BASE STYLESHEET COMPILER
 
-$govuk-use-legacy-palette: true;
+$govuk-use-legacy-palette: false;
 $govuk-typography-use-rem: false;
 
 @import "govuk_publishing_components/all_components";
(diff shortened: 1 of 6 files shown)


injms:
On `/world`, the link takes up the full width of the list:

Screenshot 2020-01-15 at 15 51 13

I think it should display inline-block, so it looks like this:

Screenshot 2020-01-15 at 15 50 33

It's more consistent, since the large invisible clickable area is removed. What do you think?

maxgds:
> I think it should display inline-block, so it looks like this:

Makes sense, I've updated in a new commit. Hopefully this is good to go now?

PR #5253, opened 2020-01-15
Update to accomodate carrierwave 2.1.0

1pretz1:
- Updated carrierwave from 1.3.1 to 2.1.0
- As per this change
(carrierwaveuploader/carrierwave@629afec#diff-a2fae088fb0c57593e3c4c6bd69bc382R189),
carrierwave forced you to explicitly set what type of cache storage you
wanted, the default was file previsouly so it has been set to that.
- Convert carrierwave 'remove_#{field}!' to '#{field}.remove!'
- Add callback to store image after save on take part page

Trello:
https://trello.com/c/rM6VLd6U/1301-audit-dependabot-for-whitehall
https://trello.com/c/31CIOZdx/1358-carrierwave-gem-bum-to-202-in-whitehall

### DIFF
--- a/app/models/take_part_page.rb
+++ b/app/models/take_part_page.rb
@@ -10,6 +10,8 @@ class TakePartPage < ApplicationRecord
   extend FriendlyId
   friendly_id :title
 
+  include PublishesToPublishingApi
+
   mount_uploader :image, ImageUploader, mount_on: :carrierwave_image
 
   validates :image, presence: true, on: :create
@@ -22,8 +24,6 @@ class TakePartPage < ApplicationRecord
              content: :body_without_markup,
              description: :summary,
              format: "take_part"
-
-  include PublishesToPublishingApi
 
   def search_link
     Whitehall.url_maker.take_part_page_path(self.slug)
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem "addressable", ">= 2.3.7"
 gem "babosa", "1.0.3"
 gem "bootsnap", require: false
 gem "bootstrap-kaminari-views", "0.0.5"
-gem "carrierwave", "~> 2.0.2"
+gem "carrierwave", "~> 2.1.0"
 gem "carrierwave-i18n"
 gem "chronic"
 gem "dalli", "~> 2.7"
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem "addressable", ">= 2.3.7"
 gem "babosa", "1.0.3"
 gem "bootsnap", require: false
 gem "bootstrap-kaminari-views", "0.0.5"
-gem "carrierwave", "~> 1.3.1"
+gem "carrierwave", "~> 2.0.2"
 gem "carrierwave-i18n"
 gem "chronic"
 gem "dalli", "~> 2.7"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -77,7 +77,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (~> 1.5)
       xpath (~> 3.2)
-    carrierwave (2.0.2)
+    carrierwave (2.1.0)
       activemodel (>= 5.0.0)
       activesupport (>= 5.0.0)
       addressable (~> 2.6)
@@ -581,7 +581,7 @@ DEPENDENCIES
   binding_of_caller
   bootsnap
   bootstrap-kaminari-views (= 0.0.5)
-  carrierwave (~> 2.0.2)
+  carrierwave (~> 2.1.0)
   carrierwave-i18n
   chronic
   ci_reporter_minitest
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -77,10 +77,13 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (~> 1.5)
       xpath (~> 3.2)
-    carrierwave (1.3.1)
-      activemodel (>= 4.0.0)
-      activesupport (>= 4.0.0)
-      mime-types (>= 1.16)
+    carrierwave (2.0.2)
+      activemodel (>= 5.0.0)
+      activesupport (>= 5.0.0)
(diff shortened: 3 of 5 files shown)


ChrisBAshton:
Currently have one remaining unit test failure:

```
$ govuk-docker-run bin/rails test test/integration/take_part_page_test.rb:10

...

Failure:
TakePartPageTest#test_TakePartPage_is_published_to_the_Publishing_API_on_save [/govuk/whitehall/test/integration/take_part_page_test.rb:21]:
The request PUT https://publishing-api.test.gov.uk/v2/content/5a582560-eb6d-4fc5-bcf1-7b092dd2961f with given block was expected to execute 1 time but it executed 0 times

The following requests were made:

PATCH https://publishing-api.test.gov.uk/v2/links/5a582560-eb6d-4fc5-bcf1-7b092dd2961f with body '{"links":{"policy_areas":[]}}' with headers {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer example', 'Content-Length'=>'29', 'Content-Type'=>'application/json', 'Host'=>'publishing-api.test.gov.uk', 'User-Agent'=>'gds-api-adapters/61.0.0 ()'} was made 1 time
PUT https://publishing-api.test.gov.uk/v2/content/5a582560-eb6d-4fc5-bcf1-7b092dd2961f with body '{"title":"A take part page title","locale":"en","publishing_app":"whitehall","redirects":[],"update_type":"major","description":"Summary text","details":{"body":"\u003cdiv class=\"govspeak\"\u003e\u003cp\u003eSome govspeak body text\u003c/p\u003e\n\u003c/div\u003e","image":{"url":"/govuk/whitehall/carrierwave-tmp/1321009871-334495616722338-0001-1208/s300_minister-of-funk.960x640.jpg","alt_text":"Image alt text"}},"document_type":"take_part","public_updated_at":"2011-11-11T11:11:11.000+00:00","rendering_app":"government-frontend","schema_name":"take_part","base_path":"/government/get-involved/take-part/a-take-part-page-title","routes":[{"path":"/government/get-involved/take-part/a-take-part-page-title","type":"exact"}]}' with headers {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer example', 'Content-Length'=>'728', 'Content-Type'=>'application/json', 'Host'=>'publishing-api.test.gov.uk', 'User-Agent'=>'gds-api-adapters/61.0.0 ()'} was made 1 time
POST https://publishing-api.test.gov.uk/v2/content/5a582560-eb6d-4fc5-bcf1-7b092dd2961f/publish with body '{"update_type":null,"locale":"en"}' with headers {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Authorization'=>'Bearer example', 'Content-Length'=>'34', 'Content-Type'=>'application/json', 'Host'=>'publishing-api.test.gov.uk', 'User-Agent'=>'gds-api-adapters/61.0.0 ()'} was made 1 time
```

...the key bit being `"image":{"url":"/govuk/whitehall/carrierwave-tmp/1321009871-334495616722338-0001-1208/s300_minister-of-funk.960x640.jpg"` (temporary filepath rather than the AWS filepath. Same issue as reported here, I believe: https://github.com/carrierwaveuploader/carrierwave/issues/2421

brucebolt:
I've bumped this to 2.1.0 (which is now the most recent release) and added a commit that fixes the outstanding issue.  It turns out carrierwave now stores the file through `after_commit`, whilst we publish the document in `after_save`.  `after_commit` is executed after `after_save`, so the image is not stored at the point we attempt to publish the document.  Since the `after_commit` is a callback only on create and update (mimicking `after_save`) I've therefore removed the `after_commit` and added a `after_save` callback instead.

As part of this, I found the documentation for carrierwave was out of sync with the gem code, so have raised a PR to update this too: https://github.com/carrierwaveuploader/carrierwave/pull/2449.

app/models/take_part_page.rb
```diff
@@ -12,6 +12,9 @@ class TakePartPage < ApplicationRecord
 
   mount_uploader :image, ImageUploader, mount_on: :carrierwave_image
 
+  skip_callback :commit, :after, :store_image!
+  after_save :store_image!
+
```
kevindew: Nice work digging into this. In the [commit message](https://github.com/alphagov/whitehall/pull/5253/commits/4c658210ed2d76ac94fe7f72d6721c315ce2aefc) we say that we do our publishing to the Publishing API through an `after_save` - though as far as I can tell we do [this in an after_commit](https://github.com/alphagov/whitehall/blob/4c658210ed2d76ac94fe7f72d6721c315ce2aefc/lib/publishes_to_publishing_api.rb#L6)? So I'm wondering if there's another callback I'm missing or if we maybe have a callback ordering problem? https://github.com/alphagov/whitehall/blob/4c658210ed2d76ac94fe7f72d6721c315ce2aefc/lib/publishes_to_publishing_api.rb#L6

brucebolt: Ah, yes, think I was getting confused.  I've updated the commit message to better reflect reality.

I think this is definitely down to ordering, but I cannot see anything incorrect with the ordering as specified in the model.

kevindew: @brucebolt cool - I see what you mean. I had a quick look myself and can see that there is something odd with the after_commit hooks. I did experiment and we can resolve this by re-ordering the file rather than swapping the hook: https://github.com/alphagov/whitehall/commit/0288a45827180adf52b0ef577ab5c29bf90980bc

brucebolt: This peculiarity now explains why I couldn't understand the issue initially.  I've updated this commit to move the include of `PublishesToPublishingApi`, which has resolved the issue.

kevindew: @brucebolt super!

PR #5242, opened 2020-01-10
Bump friendly_id from 5.2.4 to 5.3.0

gclssvglx:
For: https://trello.com/c/rM6VLd6U/1301-audit-dependabot-for-whitehall

Added `extends FriendlyId` to `Attachment` and all classes that inherit from `Attachment` - as recommended [here](https://github.com/norman/friendly_id/blob/9cd4ce2b5d0425aa704872e1334b442cab03c29c/lib/friendly_id/base.rb#L21)...

> A note about single table inheritance (STI): you must extend FriendlyId in all classes that participate in STI, both your parent classes and their children.

Pull `friendly_id :title, use: :scoped, scope: :attachable` from `HtmlAttachment` up into the parent `Attachment` class.

PR #5256, opened 2020-01-16
Revert "Bump friendly_id from 5.2.4 to 5.3.0"

hannako:
Reverts alphagov/whitehall#5242

Adding the friendly id logic to the parent class seems to have had unexpected consequences ie users have been able to create attachments with identical URLs attachment.

See thread on this ticket: https://govuk.zendesk.com/agent/tickets/3893279 for further information.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem "dalli", "~> 2.7"
 gem "deprecated_columns", "~> 0.1.1"
 gem "equivalent-xml", "~> 0.6.0", require: false
 gem "faraday"
-gem "friendly_id", "~> 5.3.0"
+gem "friendly_id", "~> 5.2.4"
 gem "fuzzy_match", "~> 2.1"
 gem "gds-sso", "~> 14.2"
 gem "globalize", "~> 5"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -138,7 +138,7 @@ GEM
     faraday (0.17.1)
       multipart-post (>= 1.2, < 3)
     ffi (1.11.3)
-    friendly_id (5.3.0)
+    friendly_id (5.2.4)
       activerecord (>= 4.0.0)
     fugit (1.3.3)
       et-orbi (~> 1.1, >= 1.1.8)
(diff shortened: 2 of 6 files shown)


kevindew:
Oh no - how disappointing. Have we captured how to replicate the problem users experienced? It sounds like something we should have a unit test for.

hannako:
Thanks for merging @1pretz1 !

@kevindew I wasn't able to replicate the problem. I tried adding an HTML attachment and a PDF attachment of the same name, but they were correctly incremented. To fix the problem for the users, I forced friendly ID to reassign their slugs as per the docs http://norman.github.io/friendly_id/

```
HtmlAttachment.friendly_id # title-of-document
HtmlAttachment.slug = nil
HtmlAttachment.save!
HtmlAttachment.friendly_id # title-of-document--2
```

kevindew:
Thanks @hannako - that's a really bummer that we've not been able to replicate it 🤔 

@1pretz1 do you have any idea what's happened? Do we know that this happened a reasonable amount of time after our deploy to not be a wrinkle in deploy process?

1pretz1:
Im not sure @kevindew, I haven't looked into it! I merged the PR after confirming with Jess that this was the change that caused the issues as seen from zendesk. I can try and look into exactly what went wrong when I have some free time.

kevindew:
> Im not sure @kevindew, I haven't looked into it! I merged the PR after confirming with Jess that this was the change that caused the issues as seen from zendesk. I can try and look into exactly what went wrong when I have some free time.

Yeah be good to do some learning, otherwise we'll lose all the value of time spent into trying to fix this in the first place.

1pretz1:
Sure, I threw together a card in our doing board to make sure we didn't forget about the issue when Jess originally opened this PR [here](https://trello.com/c/EiQSm4yx/1359-look-into-friendlyid-bump-and-why-it-caused-unexpected-issues).

hannako:
@kevindew I think the original PR was merged on the 13th, and the 4 attachments with issues were added on the 14th. Not sure when the release was deployed though, maybe you're right and the deploy coincided with the problems occurring. We tried to replicate the bug by adding two new attachments (one of which was an html) with identical titles. This failed to show the bug. My next step was to find a publication with an existing pdf and to add an html attachment of the same title, to see if that replicated the issue. If that doesn't replicate it either, maybe it was a wrinkle after all! Hope that helps, and apologies to @1pretz1 if I reverted a perfectly good PR!!

PR #5258, opened 2020-01-17
Extract GOV.UK link validation logic to validator

laurentqro:
This refactor will allow us to use this custom validator wherever needed, potentially for the no-deal content notice links (https://github.com/alphagov/whitehall/pull/5249).

It extracts validation logic from `app/models/document_collection_non_whitehall_link/govuk_url.rb` into a custom validator, at `app/validators/gov_uk_url_validator.rb`.

### DIFF
--- a/app/models/document_collection_non_whitehall_link/govuk_url.rb
+++ b/app/models/document_collection_non_whitehall_link/govuk_url.rb
@@ -1,11 +1,11 @@
 class DocumentCollectionNonWhitehallLink::GovukUrl
   include ActiveModel::Validations
 
-  attr_reader :url, :document_collection_group, :content_item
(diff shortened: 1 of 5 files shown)


app/models/document_collection_non_whitehall_link/govuk_url.rb
```diff
@@ -26,31 +26,17 @@ def title
     content_item["title"]
   end
 
-private
+  private
```
huwd: Is our styleguide to have `private` indented with the class/end or with the methods?

I've seen both enforced in rubocop at various points?

laurentqro:
Fixed the linting issue but it turns out, the validator was being used elsewhere, so I have to make a few tweaks.

PR #5268, opened 2020-01-24
Bulk change organisation

thomasleese:
This class takes a path to a CSV file and performs a bulk update of the lead and supporting organisations associated with documents. This is useful during machinery of government changes.

The CSV file must contain at least the following columns:
- 'Slug'
- 'New lead organisations'
- 'New supporting organisations'

Multiple organisations may be split by a comma.

[Trello Card](https://trello.com/c/tDJ5oeIu/1702-add-rake-tasks-to-handle-machinery-of-government-changes)

### DIFF
--- a/lib/data_hygiene/bulk_organisation_updater.rb
+++ b/lib/data_hygiene/bulk_organisation_updater.rb
@@ -1,0 +1,70 @@
+module DataHygiene
+  class BulkOrganisationUpdater
+    def initialize(filename)
+      @filename = filename
+    end
+
+    def call
+      CSV.foreach(filename, **CSV_OPTIONS) do |row|
+        process_row(row)
+      end
+    end
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+  private
+
+    attr_reader :filename
+
+    CSV_OPTIONS = {
+      headers: true,
+    }.freeze
+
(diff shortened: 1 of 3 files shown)


lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def update_document(document, new_lead_organisations, new_supporting_organisations)
+      edition = document.latest_edition
+
+      edition.update(
+        lead_organisations: new_lead_organisations,
+        supporting_organisations: new_supporting_organisations,
+      )
+
+      PublishingApiDocumentRepublishingWorker.perform_async(document.id)
```
cbaines: Checking if a change has been made would be good around this point, as if no change has been made, you don't need to republish the document.

I can imagine this coming in useful if you process a file, find an error part way through, then process the fixed file.

thomasleese: That's a good idea! Will add that 👍

thomasleese: I've amended the commit with that and added a test.

cbaines: :+1:

PR #5272, opened 2020-01-24
Ensure we build all edition dependencies on validation error

laurentqro:
This fixes a bug where not all three pairs of fields would get built for no deal notice links when there has been a validation error.

This PR also reduces leakage of Brexit no deal notice logic into the app, by moving most of it into `Edition::BrexitNoDealContentNoticeLinks` module.

### Before

Screenshot 2020-01-24 at 16 15 47

### After

Screenshot 2020-01-24 at 16 16 29


koetsier:
Optionally, you could add a test case showing this is fixed?

laurentqro:
> Optionally, you could add a test case showing this is fixed?

@koetsier added a test 👌

PR #5276, opened 2020-01-28
Export image blank caption as nil

brucebolt:
When presenting a Whitehall image to Publishing API, we replace empty strings with a nil value: 
https://github.com/alphagov/whitehall/blob/9396ec5a69750bd41c106fa7517d8c3476988d31/app/presenters/lead_image_presenter_helper.rb#L26-L31

In the export API, we are including a blank string, so the Content Publisher integrity check fails as the caption is not the same.  Therefore the blank captions should be replaced with a nil value.

This affects six of NDA's news articles.

Trello card: https://trello.com/c/3AV7mZTG

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -107,7 +107,7 @@ class DocumentExportPresenter < Whitehall::Decorators::Decorator
 
     edition.images.map do |image|
       image["alt_text"].squish! if image["alt_text"]
-      image["caption"].strip! if image["caption"]
+      image["caption"] = image["caption"].presence&.strip if image["caption"]
       image.as_json(methods: :url)
            .merge(variants: image_variants(image))
     end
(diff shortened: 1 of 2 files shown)


app/presenters/document_export_presenter.rb
```diff
@@ -108,6 +108,7 @@ def present_images(edition)
     edition.images.map do |image|
       image["alt_text"].squish! if image["alt_text"]
       image["caption"].strip! if image["caption"]
+      image["caption"] = nil if image["caption"] && image["caption"].empty?
```
kevindew: Might want to handle both of these things on the same line: e.g. `image["caption"] = image["caption"].presence&.strip`

Do we need to be doing the same thing with alt_text too? 

On the other hand should Content Publisher just set caption and alt_text to empty texts? I think that's how they end up in Content Publisher through the UI if someone doesn't fill them in.

brucebolt: I'll update this to do both on the same line, since that's cleaner.

It's not an issue with alt_text:
https://github.com/alphagov/whitehall/blob/9396ec5a69750bd41c106fa7517d8c3476988d31/app/presenters/lead_image_presenter_helper.rb#L18-L24

I think it would be nice if Content Publisher could have these set as empty strings, however it would cause issues with the integrity check and would involve including an exception.

brucebolt:
Closing in favour of https://github.com/alphagov/content-publisher/pull/1704.

PR #5282, opened 2020-01-28
Some linting TLC

kevindew:
In https://github.com/alphagov/whitehall/pull/5278 I noticed that the feature step files had various linting issues. To resolve this I've now updated the Jenkinsfile configuration to lint all directories and resolved the linting errors in this and other ignored directories.

I also did a review of things ignored in the rubocop.yml file and determined that a bunch of these could be resolved.


features/support/retrospective_stub.rb
```diff
@@ -8,6 +8,8 @@ def initialize
     @calls = []
   end
 
+  # rubocop:disable Style/MethodMissingSuper
+  # rubocop:disable Style/MissingRespondToMissing
```
ChrisBAshton: You can comma-separate these:

```ruby
  # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
```

Same suggestion below.

.rubocop.yml
```diff
@@ -41,14 +41,6 @@ Style/FormatStringToken:
     - 'lib/migrate_assets_to_asset_manager.rb'
     - 'lib/tasks/scheduled_publishing.rake'
 
-# Cops disabled, due to false positives and infinite loops
-
-Layout/AccessModifierIndentation:
-  Enabled: false
-
-Style/AccessModifierDeclarations:
-  Enabled: false
```
ChrisBAshton: I understand these changes based on the commit message, but there are a lot of other little changes in c3888444c022567ac6209ae3e503962b7d5965e4, could they be moved to a 'general tidy up' commit?

features/support/retrospective_stub.rb
kevindew: Oh nice, didn't realise that. I've applied.

.rubocop.yml
kevindew: Ah - all of those changes are because of those cops being re-enabled. I've amended the commit message to be clearer.

ChrisBAshton: Excellent 👍

PR #5302, opened 2020-01-30
Update history of UK government template

injms:
## What

The new focus states weren't being used on the [History of the UK Government](https://www.gov.uk/government/history) page. This updates the links and lists to use the styles from GOV.UK Frontend. 

The images have had lazy loading attributes added to improve page load performance, and had a element added around them to stop any jank once an image has loaded.

## Why

To improve the accessibility of the page, to keep visual consistency with other pages, and to improve page load performance.

## Visual differences

Changes only made on `/government/history`.

### Focus states updated

Before:

image

After:

image

---

### Bullet points updated

Before:

image

After:

image

### DIFF
--- a/app/views/histories/index.html.erb
+++ b/app/views/histories/index.html.erb
@@ -31,7 +31,9 @@
     </div>
     <div class="column image">
       <figure class="content">
-        <%= image_tag 'history/horse-guards-vintage.jpg', alt: 'Introduction to history of government' %>
+        <span class="aspect-ratio-3:2 govuk-!-margin-bottom-3">
+          <%= image_tag "history/horse-guards-vintage.jpg", alt: "An old painting of the Horse Guards building.", loading: "lazy" %>
+        </span>
         <figcaption>Thomas Shotter Boys &ndash; A View of the Horse Guards from Whitehall. Government Art Collection.</figcaption>
       </figure>
     </div>
@@ -43,7 +45,9 @@
     <hr class="keyline">
     <div class="column">
       <div class="content">
-        <%= image_tag 'how-gov-works/Churchill_01.jpg', alt: 'History of government' %>
+        <span class="aspect-ratio-3:2">
+          <%= image_tag 'how-gov-works/Churchill_01.jpg', alt: "", loading: "lazy" %>
+        </span>
       </div>
     </div>
     <div class="column">
@@ -113,7 +117,9 @@
     </div>
     <div class="column">
       <div class="content">
-        <%= image_tag 'history/government-buildings.jpg', alt: 'Government buildings' %>
+        <span class="aspect-ratio-3:2">
+          <%= image_tag 'history/government-buildings.jpg', alt: "", loading: "lazy" %>
+        <span>
       </div>
     </div>
   </div>
@@ -124,7 +130,9 @@
     <hr class="keyline">
     <div class="column">
       <div class="content">
-        <%= image_tag 'history/historical-documents.jpg', alt: 'Historical documents' %>
+        <span class="aspect-ratio-3:2">
+          <%= image_tag 'history/historical-documents.jpg', alt: "", loading: "lazy" %>
+        <span>
       </div>
     </div>
     <div class="column">
--- a/app/views/histories/index.html.erb
+++ b/app/views/histories/index.html.erb
@@ -20,9 +20,13 @@
     <hr class="keyline">
     <div class="column copy">
       <div class="content">
-        <h2>Why is history important?</h2>
-        <p class="intro">Governments are remembered for their leaders and the course they set for their country. The British government has a long and fascinating history, and exploring its past can help us understand how it is run today.</p>
-        <p>The information here provides a starting point for research. It includes objective factual content and research carried out by independent and civil service historians.</p>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Why is history important?",
+          id: "why-is-history-important",
+          margin_bottom: 3,
+        } %>
+        <p class="govuk-body intro">Governments are remembered for their leaders and the course they set for their country. The British government has a long and fascinating history, and exploring its past can help us understand how it is run today.</p>
+        <p class="govuk-body">The information here provides a starting point for research. It includes objective factual content and research carried out by independent and civil service historians.</p>
       </div>
     </div>
     <div class="column image">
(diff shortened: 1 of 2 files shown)


app/views/histories/index.html.erb
```diff
@@ -57,23 +83,46 @@
...
-          <li><a href="/government/history/11-downing-street">11 Downing Street</a></li>
-          <li><a href="/government/history/king-charles-street">King Charles Street (FCO)</a></li>
-          <li><a href="/government/history/lancaster-house">Lancaster House (FCO)</a></li>
-          <li><a href="/government/publications/history-of-the-ministry-of-defence">Ministry of Defence, Whitehall</a></li>
-          <li><a href="/government/history/1-horse-guards-road">1 Horse Guards Road</a></li>
-          <li><a href="/hillsborough-castle">Hillsborough Castle</a></li>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Government buildings",
+          id: "government-buildings",
+          heading_level: 2,
```
andysellick: Component is a h2 by default.

app/views/histories/index.html.erb
```diff
@@ -84,17 +133,27 @@
...
       <div class="content">
-        <h2 id="documents-records">Documents and records</h2>
-        <p>The National Archives is preserving UK government information from letters and treaties to web pages Cabinet papers. You can search the archive, or browse the themed collections.</p>
-        <p>You can also explore the Government Art Collection online, which contains over 13,500 works of art, mainly from British artists.</p>
-        <ul>
-          <li><a href="http://www.nationalarchives.gov.uk/webarchive/" rel="external">The National Archives: UK Government Web Archive</a></li>
-          <li><a href="http://www.number10.gov.uk/history-and-tour/government-art-collection-at-number-10/" rel="external">Government Art Collection</a></li>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Documents and records",
+          heading_level: 2,
```
andysellick: H2 by default.

app/views/histories/index.html.erb
```diff
@@ -16,14 +20,21 @@
     <hr class="keyline">
     <div class="column copy">
```
andysellick: If we're introducing govuk-frontend's grid elsewhere in this PR, is there a reason not to do it here as well? (sorry, don't want to make more work for you unnecessarily)

app/views/histories/index.html.erb
```diff
@@ -106,19 +165,30 @@
     <div class="column">
       <hr class="keyline">
       <div class="content" id="articles-and-blog-posts">
-        <h2>Articles and blog posts</h2>
-        <p>Some government departments, including Number 10 and the Foreign and Commonwealth Office, write or commission articles and research to improve our knowledge of the history of the British government.</p>
-        <p>Find out more on the <a href="http://history.blog.gov.uk">history of government blog</a>.</p>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Articles and blog posts",
+          heading_level: 2,
```
andysellick: H2 by default.

app/views/histories/index.html.erb
```diff
@@ -106,19 +165,30 @@
...
       <hr class="keyline">
       <div class="content">
-        <h2>More sources</h2>
-        <p>Many institutions in the UK, including universities, museums and libraries, hold information about the history of our government. The following websites may be useful for further research:</p>
-        <ul>
-          <li><a href="http://www.parliament.uk/about/living-heritage/" rel="external">Parliament - Living Heritage</a></li>
-          <li><a href="http://www.bl.uk/" rel="external">The British Library</a></li>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "More sources",
+          heading_level: 2,
```
andysellick: H2 by default.

app/views/histories/index.html.erb
injms: You've nerdsniped me into doing this.

app/views/histories/index.html.erb
```diff
@@ -2,125 +2,179 @@
...
-        <%= image_tag 'history/horse-guards-vintage.jpg', alt: 'Introduction to history of government' %>
-        <figcaption>Thomas Shotter Boys &ndash; A View of the Horse Guards from Whitehall. Government Art Collection.</figcaption>
-      </figure>
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Why is history important?",
+        id: "why-is-history-important",
+        margin_bottom: 3,
+      } %>
+      <p class="govuk-body intro">Governments are remembered for their leaders and the course they set for their country. The British government has a long and fascinating history, and exploring its past can help us understand how it is run today.</p>
```
andysellick: I wonder if we could get rid of the `intro` class and use something from govuk-frontend instead?  The [typography](https://design-system.service.gov.uk/styles/typography/) page seems to have a few options.

app/views/histories/index.html.erb
```diff
@@ -2,125 +2,179 @@
...
-          <li><a href="/government/history/king-charles-street">King Charles Street (FCO)</a></li>
-          <li><a href="/government/history/lancaster-house">Lancaster House (FCO)</a></li>
-          <li><a href="/government/publications/history-of-the-ministry-of-defence">Ministry of Defence, Whitehall</a></li>
-          <li><a href="/government/history/1-horse-guards-road">1 Horse Guards Road</a></li>
-          <li><a href="/hillsborough-castle">Hillsborough Castle</a></li>
-        </ul>
-      </div>
+  <div class="govuk-grid-row government-buildings">
+    <div class="govuk-grid-column-full">
+      <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--l">
```
andysellick: I don't know how this bit of govuk-frontend works, but does this HR need to be in a grid column? Could it be moved a few lines up to between the wrapping govuk-grid-row and the previous one?

injms: The `hr` doesn't have any left or right margin - so is wider than the full width column. I could add the margin with the override classes - but that's not a good path to take. Doing what you propose means it will overhang each side:

image

My thought was either adding more CSS to add a margin to the left and right of the `hr`; or adding more HTML to wrap it in a full width column. 

Not much between either option to be honest - so I chose adding the column in the template because I think that's clearer in showing what's being done, and doesn't add anything new.

app/views/histories/index.html.erb
injms: Good spot - replaced with `govuk-body-l`.

app/views/histories/index.html.erb
andysellick: Cool, that's fine. Tricky reviewing when you can't see it, eh. Happy to look over this when you're deploying if you like.

PR #5309, opened 2020-01-31
Republish worldwide organisation editions when reslugging

bilbof:
When worldwide organisations are re-slugged with the rake task `reslug:worldwide_organisation`, we want their editions to be updated, so we can filter on them using their new slug.


bilbof:
Though this may be necessary, it's currently non-functional since `WorldwideOrganisation` doesn't have an `editions` method.

Something like this should work, though it'd be good to update the world locations rake task to also republish docs:

```
WorldwideOrganisation.find_by(slug: slug).edition_worldwide_organisations.each { |org_edition|
    edition = org_edition.edition
    edition.update_in_search_index if edition && edition.published?
}
```

PR #2792, opened 2016-10-24
Add a rake task to publish links for withdrawn items

mobaig:
A data hygiene task to ensure that the publishing api has the correct
links present for withdrawn content items, bringing it line with the
content-store and whitehall. We've detected multiple instances in
production where this is not the case.

https://trello.com/c/jVPQzU1S/232-republish-the-links-of-withdrawn-whitehall-content

Note: needs testing against a sensible data setup, which my VM is currently lacking in a big way. Will get round to that shortly but this can be reviewed in the meantime, particularly as the code is nearly identical to the existing published items task in the same file.

PR #5317, opened 2020-02-03
Send draft item links to Publishing Api

Rosa-Fox:
[Trello](https://trello.com/c/V1uLtsQV/1409-fix-mismatch-in-publishing-organisations-on-migration)

- links information is missing in the publishing-api for draft content.
- This is causing errors when importing data into Content Publisher as draft documents are missing `primary_publishing_organisation` within `links`.
- This task will patch the links information for draft documents. It looks like it was previously only run for published and withdrawn content.
- It seems like we may be encountering this issue for similar reasons to: https://github.com/alphagov/whitehall/pull/2792#issuecomment-255812139 - only in our case for draft rather than withdrawn content.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -142,6 +142,21 @@ namespace :publishing_api do
     puts "Finished queuing items for Publishing API"
   end
 
+  desc "Send draft item links to Publishing API."
+  task patch_draft_item_links: :environment do
+    editions = Edition.draft
+    count = editions.count
+    puts "# Sending #{count} draft editions to Publishing API"
+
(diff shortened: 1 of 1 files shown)


emmabeynon:
Thanks for putting this together Rosa.  Have we looked into this was a legacy issue for old draft NDA content, and whether patching links occurs automatically if an organisation creates new content now?  My concern is we would need to remember to run this right before kicking off the import process.

Could you also add a link to the Trello card in the description?  A link to this PR will then automatically get added to the card.

Rosa-Fox:
@emmabeynon I tested on Whitehall integration and drafts do have the links data when created, so it should be a case of fixing the existing draft documents.

Will link the PR :)

PR #5321, opened 2020-02-05
Bump friendly

1pretz1:
This PR bumps friendly but adds some code so it behaves exactly the same as it did before, huraah...

Further details in the commits: ->

Trello:
https://trello.com/c/EiQSm4yx/1359-bump-friendlyid-to-version-530

### DIFF
--- a/db/data_migration/20200205160940_remove_slugs_for_file_and_external_attachments.rb
+++ b/db/data_migration/20200205160940_remove_slugs_for_file_and_external_attachments.rb
@@ -1,0 +1,4 @@
+Attachment
+  .where.not(type: "HtmlAttachment")
+  .where.not(slug: nil)
+  .update_all(slug: nil)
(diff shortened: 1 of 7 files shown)


db/data_migration/20200205160940_remove_slugs_for_file_and_external_attachments.rb
```diff
@@ -0,0 +1,4 @@
+Attachment
+  .where.not(type: "HtmlAttachment")
+  .where.not(slug: nil)
```
kevindew: Minor - this can be where.not(type: "HtmlAttachment", slug: nil)

1pretz1: Nice, I normally like to put my where clauses on separate lines to be super explicit as I think it reads better but thats just personal preference!

1pretz1:
Tested on integration, all working as expected so going to merge! 🚀

PR #5345, opened 2020-02-12
Update to remove .button class

andysellick:
This PR is part of the work to remove govuk_frontend_toolkit from static, specifically to remove the `.button` class. Work has previously been done in `govspeak` to change it to use the `govuk_publishing_components` gem to render buttons (removing the reliance on the `.button` class there) and a new version of the components gem produced to account for this change in the govspeak component.

This PR updates whitehall to have both of these changes. It also fixes a test that expected the old class. While whitehall does make use of the `.button` class, it appears to be relying on its own styles for this, not the styles from static.

Trello card: https://trello.com/c/d8D7JTJE/203-remove-design-patterns-buttons-from-static

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -173,7 +173,7 @@ GEM
       activemodel (>= 4.2, < 6.1)
       activerecord (>= 4.2, < 6.1)
       request_store (~> 1.0)
-    govspeak (6.5.2)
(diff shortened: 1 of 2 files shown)


chao-xian:
https://github.com/alphagov/govspeak/pull/176

PR #5349, opened 2020-02-13
[DNM cos reshuffle] Export contacts for all documents

1pretz1:
Previously contacts were only exported for documents that had been
force_published, published or unwithdrawn. This was because we were
relying on an edition association - depended_upon_contacts which is only
populated when a document transitions to one of the states stated above.

Consequently draft editions never had any contacts information exported
and so when importing one in content-publisher an error was thrown
regarding missing contacts. The change in this commit now looks through
the body of text of the edition to pull out the relevant contacts.

Trello:
https://trello.com/c/OzdqkkW2/1455-fix-documents-that-failed-the-import-with-an-array-error

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -26,7 +26,7 @@ private
       alternative_format_provider_content_id: edition.try(:alternative_format_provider)&.content_id,
       attachments: present_attachments(edition),
       authors: edition.authors.pluck(:id).uniq,
-      contacts: slice_association(edition, :depended_upon_contacts, %i[id content_id]),
+      contacts: present_contacts(edition),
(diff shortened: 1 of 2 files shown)


kevindew:
Nice detective work Pete 👍 

I feel like we also need to make an amend in Content Publisher so that we're resilient if a contact isn't found: https://github.com/alphagov/content-publisher/blob/9865b9e8a5da1b9a29e16110d803b7d093865a78/lib/whitehall_importer/embed_body_references.rb#L29-L30 as we have no guarantees that all the Whitehall ones entered are real things.

1pretz1:
Good point @kevindew, I've added an AC to the card linked on this PR to update content-publisher too.

PR #5351, opened 2020-02-13
Expand the unpublishing hack in the HTML attachments worker

cbaines:
The updating of edition.unpublishing is apparently broken, and this
worker works around that by also attempting to access the unpublishing
by id.

Do this in a couple of more cases, so the behaviour is hopefully
broken less often.


cbaines:
I've raised a Pull Request for a much better change now, so this can probably be ignored.

cbaines:
https://github.com/alphagov/whitehall/pull/5354

cbaines:
Closing, as the better change has been merged now.

PR #5358, opened 2020-02-14
Wrap edition#create_draft in a transaction to ensure all traits are fully persisted

edwardkerry:
Wrapping this method in a transaction will ensure that the new draft is fully saved in the database and contains all the expected properties.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -469,16 +469,18 @@ class Edition < ApplicationRecord
       raise "Cannot create new edition based on edition in the #{state} state"
     end
 
-    ignorable_attribute_keys = %w(id type state created_at updated_at change_note
-                                  minor_change force_published scheduled_publication)
-    draft_attributes = attributes.except(*ignorable_attribute_keys)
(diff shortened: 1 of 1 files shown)


app/models/edition.rb
```diff
@@ -465,20 +465,22 @@ def included_in_statistics_feed?
...
-    end
-
-    ignorable_attribute_keys = %w(id type state created_at updated_at change_note
-                                  minor_change force_published scheduled_publication)
-    draft_attributes = attributes.except(*ignorable_attribute_keys)
-      .merge("state" => "draft", "creator" => user, "previously_published" => previously_published)
+    ActiveRecord::Base.transaction do
+      unless published?
+        raise "Cannot create new edition based on edition in the #{state} state"
+      end
```
cbaines: You could do this prior to the transaction, as it's not modifying any state.

That would also avoid beginning the transaction, then rolling back if the state is wrong.

cbaines:
The first line of your commit message is a little long. Perhaps just move the "to ensure all traits are fully persisted" bit to the body.

PR #5364, opened 2020-02-17
Remove link between about pages and publications

ChrisBAshton:
Removes logic that republishes Organisation/About pages when certain document types (FOI, Corporate Report and Transparency Data) are published.

Post-merge, we'll need to run the (existing) [`publishing_api:republish_all_organisations`](https://github.com/alphagov/whitehall/blob/8f1b71d7faa130547a3fca621542b6f9f865034b/lib/tasks/publishing_api.rake#L100-L103) rake task and the (new) `publishing_api:republish_all_about_pages` rake task.

# Tested on Integration

Tested using `publishing_api:republish_organisation["government-digital-service"]` and `publishing_api:republish_about_page["government-digital-service"]`

## Organisation page

https://www.integration.publishing.service.gov.uk/government/organisations/government-digital-service

Before:

Corporate reports and Transparency data links showing

After:

No corporate reports or Transparency data links

## About page

https://www.integration.publishing.service.gov.uk/government/organisations/government-digital-service/about

Before:

Corporate reports and Transparency data links showing

After:

No corporate reports or Transparency data links

---

Trello card: https://trello.com/c/i90DA1XW/1396-remove-link-between-about-pages-and-publications-in-whitehall


lib/tasks/publishing_api.rake
```diff
@@ -102,6 +102,19 @@ namespace :publishing_api do
...
+  task :republish_about_page, [:org_slug] => :environment do |_, args|
+    about_us_page = Organisation.find_by!(slug: args[:org_slug]).about_us
+    PublishingApiDocumentRepublishingWorker.new.perform(about_us_page.document_id)
+  end
+
+  desc "Republish all About pages"
+  task republish_all_about_pages: :environment do
+    about_us_pages = Organisation.all.map(&:about_us).compact
+    about_us_pages.map(&:document_id)
+      .each(&PublishingApiDocumentRepublishingWorker.method(:perform_async))
```
1pretz1: Cool use of `each`, didn't know you could use it this way!

benthorner: Should we do this synchronously, so we know if all the jobs completed successfully?

benthorner: Also, I notice the implementation of this task is different to [(re)publishing for organisations](https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/models/about_page.rb). The technique we're using here seems better, since it should also update the draft stack. Are we confident we can republish the new org pages on the draft stack too?

lib/tasks/publishing_api.rake
```diff
@@ -102,6 +102,19 @@ namespace :publishing_api do
     Organisation.find_each(&:publish_to_publishing_api)
   end
 
+  desc "Republish an organisation's About page"
```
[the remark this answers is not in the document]
ChrisBAshton: This particular rake task was just to test it on integration. The other rake task is the one I envisaged us keeping (there are other rake tasks to do with mass-republishing). I agree there's probably little value in keeping them afterwards, though these things do come in handy on 2nd line sometimes.

lib/tasks/publishing_api.rake
ChrisBAshton: Hi Ben - I have just re-tested this on Integration and running `publishing_api:republish_organisation["government-digital-service"]` updated both the live and draft stacks.

ChrisBAshton: > Should we do this synchronously, so we know if all the jobs completed successfully?

Performing it asynchronously follows the convention of the other rake tasks. I've now added some logging, which helps to provide at least some visibility of what the rake task is doing - again this is a convention as per other rake tasks. See `patch_published_item_links` for a similar example.

lib/tasks/publishing_api.rake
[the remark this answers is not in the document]
benthorner: OK. Can we delete it if it's not necessary for this PR? If 2ndline do need it, they can always add it themselves.

lib/tasks/publishing_api.rake
benthorner: We seem to have a mixture of sync and async. For example, the [task to republish all orgs](https://github.com/alphagov/whitehall/blob/795d103153f516ee665f0dc9008025be422cfc1b/lib/tasks/publishing_api.rake#L102) appears to be synchronous. How will we verify the about pages were all republished successfully? The synchronous approach should make this clear - worth noting Whitehall has historical problems with the failing async jobs going unnoticed.

ChrisBAshton: Have now changed to synchronous 👍

cbaines:
I've read the commit messages, but I can't quite join up what they're saying with the removal of the links I see in the screenshots. What's the motivation behind removing the links?

ChrisBAshton:
> What's the motivation behind removing the links?

@cbaines the links were problematic because they were not updated dynamically. As [Ben's Trello comment](https://trello.com/c/i90DA1XW/1396-remove-link-between-about-pages-and-publications-in-whitehall#comment-5e31bb28d27591133cd0bef8) states:

> Pick an organisation that doesn't have "Corporate reports" or "Transparency data". I chose "Internal Audit Profession".
> Publish a new "Corporate report" or "Transparency data" document under that organisation.
> **Note that the organisation home page does not update.**
> Manually republish the organisation home page to see the update.

We could try fixing the behaviour to update automatically, but this would actually worsen the situation for documents published in Content Publisher, as we'd then find that publishing a corporate report in Content Publisher has the unexpected side-effect of republishing an Organisation page in Whitehall.

The links we've removed ("Transparency data" and "Corporate reports") are subsets of "Transparency and freedom of information releases", which is a link that is shown automatically anyway. After doing performance analysis and finding the two subset links are rarely used, it was decided to simply drop the links altogether.

cbaines:
> > What's the motivation behind removing the links?
> 
> @cbaines the links were problematic because they were not updated dynamically. As [Ben's Trello comment](https://trello.com/c/i90DA1XW/1396-remove-link-between-about-pages-and-publications-in-whitehall#comment-5e31bb28d27591133cd0bef8) states:
> 
> > Pick an organisation that doesn't have "Corporate reports" or "Transparency data". I chose "Internal Audit Profession".
> > Publish a new "Corporate report" or "Transparency data" document under that organisation.
> > **Note that the organisation home page does not update.**
> > Manually republish the organisation home page to see the update.

Thanks for sharing, I don't have access to the Trello card.

> We could try fixing the behaviour to update automatically, but this would actually worsen the situation for documents published in Content Publisher, as we'd then find that publishing a corporate report in Content Publisher has the unexpected side-effect of republishing an Organisation page in Whitehall.

My guess is that the best technical way of supporting this behaviour is through link expansion and dependency resolution in the Publishing API, and having the rendering logic on the frontend use that. If this is the case, I think it just hasn't been implemented due to time limitations when the rendering for organisation pages was migrated to the collections app.

> The links we've removed ("Transparency data" and "Corporate reports") are subsets of "Transparency and freedom of information releases", which is a link that is shown automatically anyway. After doing performance analysis and finding the two subset links are rarely used, it was decided to simply drop the links altogether.

It's good to know that there's still a way of getting to the data. Removing the links still makes me a bit uncomfortable, as I'm unsure about the impact. Did you consider just including the links for all organisations, regardless of the content they have?

ChrisBAshton:
> Did you consider just including the links for all organisations, regardless of the content they have?

That would be a confusing user experience. To take the Internal Audit Profession example, a link to "[Transparency Data](https://www.gov.uk/search/transparency-and-freedom-of-information-releases?content_store_document_type=transparency&organisations%5B%5D=civil-service-internal-audit-profession)" would look like this:

Screenshot 2020-02-25 at 14 21 13

PR #5365, opened 2020-02-17
Refactor the ReplacementIdUpdates class for readability

cbaines:
Format to avoid the lines which were too long. Also reduce the code
duplication by using a variable.

### DIFF
--- a/app/services/asset_manager/attachment_updater/replacement_id_updates.rb
+++ b/app/services/asset_manager/attachment_updater/replacement_id_updates.rb
@@ -5,20 +5,27 @@ class AssetManager::AttachmentUpdater::ReplacementIdUpdates
     replacement = attachment_data.replaced_by
 
     Enumerator.new do |enum|
-      enum.yield AssetManager::AttachmentUpdater::Update.new(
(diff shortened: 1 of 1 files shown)


app/services/asset_manager/attachment_updater/replacement_id_updates.rb
```diff
@@ -6,19 +6,24 @@ def self.call(attachment_data)
...
+                   replacement_legacy_url_path: replacement.file.asset_manager_path
+                 )
 
       if attachment_data.pdf?
-        if replacement.pdf?
-          enum.yield AssetManager::AttachmentUpdater::Update.new(
-            attachment_data, attachment_data.file.thumbnail, replacement_legacy_url_path: replacement.file.thumbnail.asset_manager_path
+        replacement_legacy_url_path = if replacement.pdf?
+                                        replacement.file.thumbnail.asset_manager_path
+                                      else
```
issyl0: You too are a fan of hanging indents, I see. 😇

cbaines: No really... it's just that the linter won't accept anything else!

PR #5368, opened 2020-02-17
Fix potential issues when handling attachments during unpublishing

cbaines:
Previously, when an edition was unpublished,
AttachmentPresentAtUnpublishUpdater service listener would run to set
the attachments as being present at unpublish. However, there was no
guarantee that this would happen prior to the AttachmentUpdater
running, enqueuing a AssetManagerAttachmentMetadataWorker job, and
then this Sidekiq job using the present_at_unpublish value.

If the Sidekiq job happened to run after the present_at_unpublish
value had been set to true, then I think this is fine. However, if it
runs before the value is updated, I think the draft property of the
attachments could mistakenly be set to true.

Assets for an unpublished edition being set as drafts is problematic
because if the assets are the replacements for other assets, the
redirect setup for the assets for the unpublished document won't
work. Asset Manager won't redirect a request for an asset to a
replacement if that replacement is a draft, so instead of redirecting
the request, the old replaced asset becomes accessible.

For assets affected by this problem, the issue can be rectified by
running the AssetManagerAttachmentMetadataWorker for the draft asset,
as that should correctly determine the asset not to be a draft, and
inform Asset Manager.


cbaines:
One note, if we do merge this, we should look at all WhitehallAsset's in Asset Manager that are both draft, and have a redirect_url. Because Whitehall specifies the redirect_url when the edition related to the asset is unpublished, it's likely that the asset shouldn't be a draft.

PR #5370, opened 2020-02-18
Data migration: remove BrexitCTA from withdrawn documents

laurentqro:
Run this data migration before merging https://github.com/alphagov/whitehall/pull/5367.

### DIFF
--- a/db/data_migration/20200218104101_remove_brexit_cta_tag_from_withdrawn_docs.rb
+++ b/db/data_migration/20200218104101_remove_brexit_cta_tag_from_withdrawn_docs.rb
@@ -1,0 +1,11 @@
+withdrawn_brexit_cta_editions = Edition.in_default_locale
+  .includes(:document)
+  .where("edition_translations.body LIKE ?", "%$BrexitCTA%")
+  .withdrawn
+
+withdrawn_brexit_cta_editions.each do |edition|
+  edition.update(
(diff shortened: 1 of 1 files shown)


db/data_migration/20200218104101_remove_brexit_cta_tag_from_withdrawn_docs.rb
```diff
@@ -0,0 +1,11 @@
+withdrawn_brexit_cta_editions = Edition.in_default_locale
```
huwd: Will this allow us to 💥 `BrexitCTA` that are in other languages, like welsh?
Are there any?

laurentqro: So after some digging, turns out `Edition.joins(:translations).includes(:document).where("edition_translations.body LIKE ?", "%$BrexitCTA%")` and `Edition.in_default_locale.includes(:document).where("edition_translations.body LIKE ?", "%$BrexitCTA%")` return the same list.

In other words, all `$BrexitCTA` tags are in the English translations, which happens to be the default locale. So we're good.

huwd: ace, thanks for chasing that!

PR #5391, opened 2020-02-25
Fix vertical alignment on past PMs page

chao-xian:
## What
The alignment of the heading on https://www.gov.uk/government/history/past-prime-ministers is broken (bleeding into the gutter).

https://trello.com/c/hwc5fDgR

### Desktop

image

### Mobile

image (1)

### Desktop

whitehall-admin dev gov uk_government_history_past-prime-ministers (1)

### Mobile

whitehall-admin dev gov uk_government_history_past-prime-ministers

### DIFF
--- a/app/assets/stylesheets/frontend/views/_historic-appointments.scss
+++ b/app/assets/stylesheets/frontend/views/_historic-appointments.scss
@@ -1,3 +1,15 @@
+.historic-appointments:not(.historical-past-chancellors) .featured-profiles .year-block {
+  @include media(mobile) {
+    h2.profiles {
+      margin: 0 0 $gutter-one-sixth;
+    }
+
+    .inner {
+      padding-left: 0;
+      padding-right: 0;
+    }
+  }
+}
 .historic-appointments {
   .historic-people-list {
     @include media(tablet) {
@@ -6,7 +18,7 @@
     }
     h3 {
       @include core-19;
-      margin-bottom: $gutter;
+      margin: $gutter 0;
 
       span {
         @include core-80;
--- a/app/assets/stylesheets/frontend/views/_historic-appointments.scss
+++ b/app/assets/stylesheets/frontend/views/_historic-appointments.scss
@@ -1,7 +1,7 @@
 .historic-appointments:not(.historical-past-chancellors) .featured-profiles .year-block {
-  @include media(mobile) {
+  @include govuk-media-query($until: tablet) {
     h2.profiles {
-      margin: 0 0 $gutter-one-sixth;
+      margin: 0 0 govuk-spacing(1) 0;
     }
 
     .inner {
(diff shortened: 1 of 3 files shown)


app/views/historic_appointments/_role_appointments_list.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="block historic-people-list">
-  <div class="inner-block">
+  <div class="inner-block govuk-!-padding-0">
```
injms: We try not to use these override classes where possible - they use `!important`, so can quickly lead to CSS-soup.

chao-xian: @injms thanks - any suggestions on alternatives?

injms: Unfortunately it might be rebuilding that layout using GOVUK Frontend's `govuk-grid-` or creating a new CSS class in the app. I'd go for using Frontend if it were me - more short term work - but work that we'd do anyway - and clearer in the longer term.

injms: Hmm, after our conversation I think this opens a much bigger rebuild of the whole page. In this case I think it's acceptable to use the override class, maybe with a `TODO` comment near it, since it should be compontenised in the near future anyways.

kr8n3r:
ideally we should start using govuk-spacing
gutter-one-sixth = govuk-spacing(1)

app/assets/stylesheets/frontend/views/_historic-appointments.scss
```diff
@@ -1,3 +1,15 @@
+.historic-appointments:not(.historical-past-chancellors) .featured-profiles .year-block {
+  @include media(mobile) {
```
injms: I think that this is clearer:

```suggestion
  @include govuk-media-query($until: tablet) {
```

app/assets/stylesheets/frontend/views/_historic-appointments.scss
```diff
@@ -1,3 +1,15 @@
...
+    h2.profiles {
+      margin: 0 0 $gutter-one-sixth;
```
injms: On reflection, I agree with @kr8n3r on this:

```suggestion
      margin: 0 0 govuk-spacing(1) 0;
```

app/assets/stylesheets/frontend/views/_historic-appointments.scss
```diff
@@ -6,7 +18,7 @@
     }
     h3 {
       @include core-19;
```
injms:
```suggestion
      @include govuk-font(19);
```

app/assets/stylesheets/frontend/views/_historic-appointments.scss
```diff
@@ -6,7 +18,7 @@
...
-      margin-bottom: $gutter;
+      margin: $gutter 0;
```
injms:
```suggestion
      margin: govuk-spacing(6) 0;
```

kr8n3r:
yay 🎆

PR #5407, opened 2020-03-02
Do not check for document being locked if no changes are being made

brucebolt:
This has the effect of allowing us to lock a locked document without getting an error.  In these cases, we currently return a 500 error, but this change means an OK response will be returned instead.

Trello card: https://trello.com/c/9ULodMA9

### DIFF
--- a/app/controllers/admin/export/document_controller.rb
+++ b/app/controllers/admin/export/document_controller.rb
@@ -19,7 +19,7 @@ class Admin::Export::DocumentController < Admin::Export::BaseController
 
   def lock
     document = Document.find(params[:id])
-    document.update!(locked: true)
+    document.update!(locked: true) unless document.locked?
   end
 
   def unlock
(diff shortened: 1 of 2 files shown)


app/models/document.rb
```diff
@@ -40,7 +40,7 @@ class Document < ApplicationRecord
   has_many :document_collections, through: :document_collection_groups
   has_many :features, inverse_of: :document, dependent: :destroy
 
-  before_save { check_if_locked_document(document: self) unless locked_changed? }
+  before_save { check_if_locked_document(document: self) unless locked_changed? || !changed? }
```
kevindew: Rather than changing this - which increases complexity in this callback could we add a conditional in the controller: https://github.com/alphagov/whitehall/blob/master/app/controllers/admin/export/document_controller.rb#L22 e..g `unless document.locked?`

brucebolt: I'm not sure how we missed that 😊

PR #5409, opened 2020-03-02
Add rake task to migrate speeches between roles

benjamineskola:
Move the speeches from one role to another by setting its `role_appointment` to the latest appointment of the specified new role (i.e., current or most recent holder).

https://trello.com/c/t7fbOpqB/1779-3-migrate-content-from-old-joint-mco-paymaster-general-role-to-mco-role


lib/tasks/migrate_role_content.rake
```diff
@@ -0,0 +1,16 @@
...
+    new_role_app = RoleAppointment.find_by!(id: args[:new_role_appointment])
+
+    old_role_app.edition_role_appointments.each do |era|
+      era.update(role_appointment: new_role_app)
+    end
+
+    old_role_app.speeches.each do |speech|
+      speech.role_appointment = new_role_app
+      speech.save!
+    end
```
thomasleese:
```suggestion
    old_role_app.speeches.update_all(role_appointment: new_role_app)
```

Similar comment as above.

test/unit/tasks/migrate_role_content_test.rb
```diff
@@ -0,0 +1,26 @@
+require "test_helper"
+require "rake"
+class MigrateRoleContentTest < ActiveSupport::TestCase
+  setup do
+    Rake.application.rake_require "tasks/migrate_role_content"
+    Rake::Task.define_task(:environment)
+    Rake::Task["migrate:role_content"].reenable
+  end
+
+  test "what" do
```
thomasleese: I think this might need a better name 😆

test/unit/tasks/migrate_role_content_test.rb
```diff
@@ -0,0 +1,26 @@
+require "test_helper"
+require "rake"
+class MigrateRoleContentTest < ActiveSupport::TestCase
```
thomasleese: Nice! We don't often unit test Rake tasks, this is pretty nice to see.

lib/tasks/migrate_role_content.rake
```diff
@@ -0,0 +1,16 @@
+namespace :migrate do
```
thomasleese: I think the convention is that usually the name of the namespace matches the name of the file. I also wonder if this could fit in [the existing `data_hygiene` namespace](https://github.com/alphagov/whitehall/blob/master/lib/tasks/data_hygiene.rake)?

test/unit/tasks/migrate_role_content_test.rb
benjamineskola: I spent half of Friday trying to manually test it and then realized with a unit test I can just use the factories to do it instead. 🙂

lib/tasks/migrate_role_content.rake
```diff
@@ -0,0 +1,16 @@
...
+  desc "Move content from one role to another (DANGER!)."
+  task :role_content, %i[old_role_appointment new_role_appointment] => :environment do |_task, args|
+    old_role_app = RoleAppointment.find_by!(id: args[:old_role_appointment])
+    new_role_app = RoleAppointment.find_by!(id: args[:new_role_appointment])
+
+    old_role_app.edition_role_appointments.each do |era|
+      era.update(role_appointment: new_role_app)
+    end
```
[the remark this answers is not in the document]
benjamineskola: That's what I'd initially intended to do, but it seems not to work:
```
Unknown column 'edition_role_appointments.role_appointment' in 'field list': UPDATE `edition_role_appointments` SET `edition_role_appointments`.`role_appointment` = 18 WHERE `edition_role_appointments`.`role_appointment_id` = 17
```
(and similar for the other instance below).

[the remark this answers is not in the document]
thomasleese: That's disappointing!

test/unit/tasks/migrate_role_content_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    Rake::Task.define_task(:environment)
+    Rake::Task["migrate:role_content"].reenable
+  end
+
+  test "what" do
+    speech = create(:speech)
+    # edition = create(:published_publication, speech: speech)
+
+    person = create(:person, forename: "Sluggy", surname: "McSlugson")
+    old_role = create(:ministerial_role, slug: "head-slug-and-chief-snail-executive", name: "Head Slug and Chief Snail Executive")
```
issyl0: :joy:

lib/tasks/migrate_role_content.rake
```diff
@@ -0,0 +1,16 @@
+namespace :data_hygiene do
```
thomasleese: I think this can go in the existing file for `data_hygiene` tasks as it's relatively small and the file itself is currently not very long: https://github.com/alphagov/whitehall/blob/master/lib/tasks/data_hygiene.rake#L10

PR #5417, opened 2020-03-03
Export content_id from documents_info

Rosa-Fox:
[Trello](https://trello.com/c/ACUVDeXD/1523-export-contentid-from-whitehall-into-content-publisher)
- This will enable us to populate the document_imports#content_id field in content-publisher
- This in turn will allow us to link to view the imported document in Whitehall, from the content-publisher migration UI
- Makes it easier to find the relationship between a content-publisher document_import and the original Whitehall document


lib/whitehall/exporters/documents_info_exporter.rb
```diff
@@ -7,10 +7,12 @@ def initialize(document_ids)
 
   def call
     document_ids.map do |doc_id|
+      document = Document.find(doc_id)
```
brucebolt: Instead of looking up each Document from its ID here, would it be nicer to update `paginated_document_ids` in `app/controllers/admin/export/document_controller.rb` to return the documents (instead of the IDs)?

E.g. replace `pluck(:document_id)` with `.map(&:document)` at https://github.com/alphagov/whitehall/blob/d86695f2a60b6f204ab92bb98ec9968599d5f66c/app/controllers/admin/export/document_controller.rb#L63

This will reduce the number of database queries we need to do.  Although you would need to rename a few methods.

This method would then become:
```
  def call
    documents.map do |document|
      {
        document_id: document.id,
        document_information: {
          locales: locales_hash[doc_id],
          content_id: document.content_id,
        },
      }
    end
  end
```

kevindew:
You could also expand the pluck to be `pluck(:document_id, :content_id)`.

Since we don't actually make any use of the locale information in the presentation of `Whitehall::Exporters::DocumentsInfoExporter` we might want to just get rid of that exporter and do the presentation in the view. I'd guess we'd be happy with responses like:

```
[
{ document_id: 12312, content_id: "123123-12312..." },
{ document_id: 12313, content_id: "123123-12313..." }
]
```

---

In other, grander, out-of-scope thoughts I wonder if we could actually abstract away the need for a Whitehall document id at all and do everything through a content_id.

Rosa-Fox:
@brucebolt @kevindew Thanks for your comments. I used a combination of the suggestions as it wasn't possible to access the `content_id` of the document using `pluck(:document_id, :content_id`. However I did remove the `Whitehall::Exporters::DocumentsInfoExporter`. I have also updated the corresponding [PR in content-publisher](https://github.com/alphagov/content-publisher/pull/1861).

brucebolt:
This is great!

The `pluck(:document_id, :content_id)` can be done on the document, instead of the edition, e.g.
```
...
.per(items_per_page)
.map(&:document)
.pluck(:document_id, :content_id)
```

You can then rename `paginated_documents` to `result_set` and remove the `result_set` method.

Rosa-Fox:
> This is great!
> 
> The `pluck(:document_id, :content_id)` can be done on the document, instead of the edition, e.g.
> 
> ```
> ...
> .per(items_per_page)
> .map(&:document)
> .pluck(:document_id, :content_id)
> ```
> 
> You can then rename `paginated_documents` to `result_set` and remove the `result_set` method.

Thanks! In the end I amended with Kevin to just pluck out the ids in order to avoid `.map(&:document)`. Let me know if all looks ok.

PR #5418, opened 2020-03-04
Remove order url and price

JonathanHallam:
Trello - https://trello.com/c/MZ6js01b/1441-simplify-order-a-copy-implementation-in-whitehall

Removal of Price and Order URL from attachments

The order URL should be the same and the prices are often out of date. This PR removes the option for inputting a price or order URL and shows a link on official documents that lets users know how to purchase them. 

There will be a further PR to remove the fields and data from the database

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -23,9 +23,6 @@ class Attachment < ApplicationRecord
     allow_blank: true,
     message: "is invalid. The number must start with one of #{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}",
   }
-  validates :price, numericality: {
-    allow_blank: true, greater_than: 0
-  }
   validates :unique_reference, length: { maximum: 255 }, allow_blank: true
 
   scope :with_filename, ->(basename) {
@@ -55,19 +52,6 @@ class Attachment < ApplicationRecord
       ends = Date.new(year + 1).strftime("%y") # %y gives last two digits of year
       "#{starts}-#{ends}"
     end
-  end
-
-  def price
-    if @price
-      @price
-    elsif price_in_pence
-      price_in_pence / 100.0
-    end
-  end
-
-  def price=(price_in_pounds)
-    @price = price_in_pounds
-    store_price_in_pence
   end
 
   def is_command_paper?
@@ -165,14 +149,6 @@ class Attachment < ApplicationRecord
   end
 
 private
-
-  def store_price_in_pence
-    self.price_in_pence = if price && price.to_s.empty?
-                            nil
-                          elsif price
-                            price.to_f * 100
-                          end
-  end
 
   def set_ordering
     self.ordering = attachable.next_ordering
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -23,11 +23,6 @@ class Attachment < ApplicationRecord
     allow_blank: true,
     message: "is invalid. The number must start with one of #{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}",
   }
-  validates :order_url, uri: true, allow_blank: true
-  validates :order_url, presence: {
-    message: "must be entered as you've entered a price",
-    if: ->(publication) { publication.price.present? },
-  }
   validates :price, numericality: {
     allow_blank: true, greater_than: 0
   }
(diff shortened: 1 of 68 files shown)


config/locales/et.yml
```diff
@@ -217,7 +217,6 @@ et:
         command_paper_number:
         hoc_paper_number: 'Parlamendi alamkoja dokumendi nr. '
         isbn: 'Manuse ISBN '
-        order_url: Manuse tellimise link
```
leenagupte: Is this something that's used in the frontend or in whitehall-admin? It looks like user facing content, so perhaps it should be kept in?

config/locales/en.yml
```diff
@@ -68,7 +68,6 @@ en:
       command_paper_number: Attachment command paper number
       hoc_paper_number: House of Commons paper number
       isbn: Attachment ISBN
-      price: Attachment price
```
leenagupte: Here too. Is this something that's used in the frontend or whitehall-admin?

leenagupte: I think that by removing this, we are losing the internationalisation on the frontend.

config/locales/et.yml
JonathanHallam: It is user facing but the idea is that it's not being used any more. There's now just a link to generic information that doesn't have a heading. Also from what I can tell the internationalization wasn't being used for order_url or price_in_pence so removing it doesn't really matter.

config/locales/en.yml
JonathanHallam: I don't think so, I've removed all mention of order_url and I didn't take out any internationalisation, I think it's probably just in here to mirror the model.

JonathanHallam: actually scratch that, I've just spotted something that I missed I think I need to change something around a bit. I think it's safe to remove `order_url` and `price` in the way that I have because they're only used when inputting but I need to use `order_a_copy_full` when presenting the link.

leenagupte: Whilst we're here, can you move the I18n removals to separate commits? There's a lot of them, so you have to scroll a long way to see which tests have been removed.

app/views/documents/_attachment.html.erb
```diff
@@ -56,6 +56,11 @@
         <%= attachment_attributes(attachment) %>
       <% end %>
     </p>
+    <% if attachment.is_official_document? %>
```
leenagupte: Should this have a test?

app/models/attachment.rb
```diff
@@ -54,6 +54,10 @@ def self.parliamentary_sessions
     end
   end
 
+  def is_official_document?
+    is_command_paper? || is_act_paper?
+  end
```
leenagupte: Should this have a test?

JonathanHallam: I don't think so, both `is_command_paper` and `is_act_paper` have tests and this is super simple logic. If you take a look at https://github.com/alphagov/whitehall/blob/master/test/unit/attachable_test.rb#L209 you can see a similar pattern; a `has_official_document` method exists that does the same thing but isn't tested because `has_command_paper` and `has_act_paper` are.

Happy to write one if you'd like though :)

app/views/documents/_attachment.html.erb
JonathanHallam: I'm happy that this is tested further down the line which is why i've not written one but will write one if you think it needs it :)

app/views/documents/_attachment.html.erb
```diff
@@ -56,14 +56,9 @@
         <%= attachment_attributes(attachment) %>
       <% end %>
     </p>
-    <% if attachment.order_url.present? %>
-      <p>
-        <%= link_to t('attachment.headings.order_a_copy'), attachment.order_url,
+    <% if attachment.is_official_document? %>
+      <%= link_to t('attachment.headings.order_a_copy'), "https://www.gov.uk/guidance/how-to-buy-printed-copies-of-official-documents",
```
kevindew: Sorry ignorant drive-by comment. Should this always be www.gov.uk or should we use Plek to have this be environment specific?

leenagupte: Good spot! 🤦‍♂

JonathanHallam: poop, yeah fair play, I'll raise a pr to fix this

PR #5419, opened 2020-03-04
Remove order url and price from attachments

JonathanHallam:
https://github.com/alphagov/whitehall/pull/5418 removed order url and price from the frontend. This removes them from the database.

Trello - https://trello.com/c/MZ6js01b/1441-simplify-order-a-copy-implementation-in-whitehall

This has had a [DO NOT MERGE FLAG] for about a month. This was to give us time to talk to publishers and make them aware of the changes that we've made. No one has raised any issues so this is now being merged and the data removed.

### DIFF
--- a/db/migrate/20200304165045_remove_order_url_and_price_in_pence_from_attachments.rb
+++ b/db/migrate/20200304165045_remove_order_url_and_price_in_pence_from_attachments.rb
@@ -1,0 +1,6 @@
+class RemoveOrderUrlAndPriceInPenceFromAttachments < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :attachments, :order_url, :string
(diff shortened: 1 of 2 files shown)


db/migrate/20200304113719_remove_price_in_pence_from_attachments.rb
```diff
@@ -0,0 +1,5 @@
+class RemovePriceInPenceFromAttachments < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :attachments, :price_in_pence, :integer
```
kevindew: Can we combine these two things into the same migration?

JonathanHallam: 👍

PR #5428, opened 2020-03-05
Add task to republish all editions with at least one official attachment

JonathanHallam:
https://github.com/alphagov/whitehall/pull/5418 removed order url and price from attachments. In order to see these changes the editions that they're attachments of need to be republished. This is a rake task to republish all of those editions. It will only need to be run once on each environment and then swiftly deleted.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -188,6 +188,21 @@ namespace :publishing_api do
     puts "Finished queuing items for Publishing API"
   end
 
+  desc "Republish all documents with official attachments"
+  task republish_documents_with_official_attachments: :environment do
+    attachment_scope = Attachment.where("unnumbered_command_paper = true OR command_paper_number != '' OR unnumbered_hoc_paper = true OR hoc_paper_number != ''")
+    edition_ids = attachment_scope.where(attachable_type: "Edition").distinct.pluck(:attachable_id)
+    response_ids = attachment_scope.where(attachable_type: "Response").distinct.pluck(:attachable_id)
+    edition_ids += Response.where(id: response_ids).pluck(:edition_id)
+    document_ids = Document.joins(:editions).where("editions.id": edition_ids).pluck(:document_id).uniq
+
+    puts "# Sending #{document_ids.count} documents to Publishing API"
+
+    document_ids.each do |doc_id|
+      PublishingApiDocumentRepublishingWorker.perform_async(doc_id)
+    end
+  end
+
   desc "Republish a document to the Publishing API"
   task :republish_document, [:slug] => :environment do |_, args|
     document = Document.find_by!(slug: args[:slug])


lib/tasks/publishing_api.rake
```diff
@@ -188,6 +188,26 @@ namespace :publishing_api do
     puts "Finished queuing items for Publishing API"
   end
 
+  desc "Republish all editions with official attachments"
+  task republish_editions_with_official_attachments: :environment do
+    editions = Edition.where(
+      id: (
+        (
+          Attachment.where.not(unnumbered_hoc_paper: [nil, false]).pluck(:attachable_id) + Attachment.where.not(unnumbered_command_paper: [nil, false]).pluck(:attachable_id) + Attachment.where.not(command_paper_number: nil).pluck(:attachable_id) + Attachment.where.not(hoc_paper_number: nil).pluck(:attachable_id)
```
cbaines: This line is rather long, could it be formatted differently?

lib/tasks/publishing_api.rake
```diff
@@ -188,6 +188,26 @@ namespace :publishing_api do
...
+        ).uniq
+      ),
+    )
+    puts "# Sending #{editions.count} editions to Publishing API"
+
+    editions.pluck(:id).each_with_index do |item_id, i|
+      PublishingApiLinksWorker.perform_async(item_id)
```
cbaines: This worker doesn't republish the edition, it just patches the links for the relevant document. Is that sufficient for what you want to change?

lib/tasks/publishing_api.rake
JonathanHallam: yeah, it also didn't do what I wanted it to. This has been shortened some and fixed, it's still quite long though.

lib/tasks/publishing_api.rake
JonathanHallam: I hadn't realised this, I've changed it so it now invokes the document republishing worker.

lib/tasks/publishing_api.rake
```diff
@@ -188,6 +188,18 @@ namespace :publishing_api do
     puts "Finished queuing items for Publishing API"
   end
 
+  desc "Republish all documents with official attachments"
+  task republish_documents_with_official_attachments: :environment do
+    attachment_ids = Attachment.where("unnumbered_command_paper = true OR command_paper_number != '' OR unnumbered_hoc_paper = true OR hoc_paper_number != ''").pluck(:attachable_id).uniq
+    document_ids = Document.joins(:editions).where("editions.id": attachment_ids).pluck(:document_id)
```
cbaines: Could this list have duplicate document ids in it?

lib/tasks/publishing_api.rake
```diff
@@ -188,6 +188,18 @@ namespace :publishing_api do
     puts "Finished queuing items for Publishing API"
   end
 
+  desc "Republish all documents with official attachments"
+  task republish_documents_with_official_attachments: :environment do
+    attachment_ids = Attachment.where("unnumbered_command_paper = true OR command_paper_number != '' OR unnumbered_hoc_paper = true OR hoc_paper_number != ''").pluck(:attachable_id).uniq
```
cbaines: The internet suggests it might be possible to write this as:
```ruby
Attachment
  .where(unnumbered_command_paper: true)
  .or("command_paper_number != ''")
  .or(unnumbered_hoc_paper: true)
  .or("hoc_paper_number != ''")
  .pluck(:attachable_id)
  .uniq
```

lib/tasks/publishing_api.rake
JonathanHallam: yeah, almost certainly good spot, i'll put `.uniq` on it

JonathanHallam: just checked on integration and that's cut 4000 documents, thanks!

lib/tasks/publishing_api.rake
JonathanHallam: Just tried this in integration and it doesn't like giving a string to `or` like this. I can mess around with something like `Attachment.where(unnumbered_command_paper: true).or(Attachment.where("command_paper_number != ''")).......` but I don't think that's better

cbaines: Ok, no problem. If this was more longer lived code, maybe it would be worth putting up more of a fight against ActiveRecord, but seen as it's a temporary Rake task, I think it's wise to leave it be now.

kevindew: Minor - these aren't attachment_ids but are attachable_ids. If we want to be sure these are edition ids we could do:

```
edition_ids = Attachment.where("unnumbered_command_paper = true OR command_paper_number != '' OR unnumbered_hoc_paper = true OR hoc_paper_number != ''")
                        .where(attachable_type: "Edition")
                        .distinct
                        .pluck(:attachable_id)
```
Which would also save having to do the uniq after a query.

PR #5432, opened 2020-03-09
Remove government from the political details presenter

cbaines:
The association between content and a government is now handled
through links in the Publishing API. This makes it easier to handle
transitions between governments, as the Publishing API will take care
of updating content where the details of the government has
changed (it's no longer current for example).

To avoid having the problem of updating the details of a document when
the government changes for new content in the future, stop sending
information about the government in the details for the content.


cbaines:
Depends on: https://github.com/alphagov/govuk-content-schemas/pull/968

PR #5433, opened 2020-03-09
Send attachment IDs in featured attachments list

barrucadu:
Frontend apps will render these by looking up the ID in the
attachments list and chucking the metadata into a component.

This change looks a bit strange, because all the attachments IDs get
included, so a reasonable question is "what's the point? why not just
render 'attachments' as featured attachments if it's a publication or
consultation?"

This was addressed in the RFC but, in summary, there's no reason why a
format or document couldn't have both featured and non-featured
attachments, and there's no guarantee that the current situation will
remain the case.  So by adopting this approach, we don't need to
change things in the future.  You could say "YAGNI", but it's not like
it adds a lot of complication.

---

This is safe to deploy because [government-frontend isn't rendering this field at the moment](https://github.com/alphagov/government-frontend/pull/1676).

[Trello card](https://trello.com/c/TJfKNCSn/1430-do-featuredattachments-metadata-properly)

### DIFF
--- a/app/presenters/publishing_api/consultation_presenter.rb
+++ b/app/presenters/publishing_api/consultation_presenter.rb
@@ -143,7 +143,7 @@ module PublishingApi
       end
 
       def featured_attachments
-        []
+        consultation.attachments.map { |a| a.publishing_api_details[:id] }
       end
     end
 
@@ -208,7 +208,7 @@ module PublishingApi
       end
 
       def final_outcome_attachments
-        []
+        outcome.attachments.map { |a| a.publishing_api_details[:id] }
       end
     end
 
@@ -278,7 +278,7 @@ module PublishingApi
       end
 
       def attachments
-        []
+        public_feedback.attachments.map { |a| a.publishing_api_details[:id] }
       end
 
       def publication_date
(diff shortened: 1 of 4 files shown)


kevindew:
Sounds good - are these attachment collections implicitly ordered?

One of the expected differences between this and the attachments collection is that the featured one is explicitly ordered whereas there isn't ordering implied in the attachments collection.

barrucadu:
Yes, it looks like they're implicitly ordered.  The current `documents` value is just the HTML for all the attachments, in the same order as the `edition.attachments`:

https://github.com/alphagov/whitehall/blob/c1068b2cf8d9d3483d8532b8665bfa968a551d81/app/presenters/publishing_api/consultation_presenter.rb#L138-L143

https://github.com/alphagov/whitehall/blob/c1068b2cf8d9d3483d8532b8665bfa968a551d81/lib/whitehall/govspeak_renderer.rb#L4-L9

https://github.com/alphagov/whitehall/blob/c1068b2cf8d9d3483d8532b8665bfa968a551d81/app/helpers/attachments_helper.rb#L16-L30

PR #5450, opened 2020-03-12
Handle StandardError when republishing whitehall docs

barrucadu:
Some old docs are so invalid trying to do things with them throws
ActiveRecord validation exceptions!

I've used `start` so we don't do repeated work.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -6,8 +6,8 @@ namespace :publishing_api do
   task republish_publications: :environment do
     all_publications = Document.where(document_type: "Publication")
     count = all_publications.count
-    all_publications.find_each.with_index do |d, i|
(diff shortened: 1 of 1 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -6,8 +6,8 @@ namespace :publishing_api do
   task republish_publications: :environment do
     all_publications = Document.where(document_type: "Publication")
     count = all_publications.count
-    all_publications.find_each.with_index do |d, i|
-      puts "#{i.fdiv(count) * 100}%: #{d.id}"
+    all_publications.find_each(start: 128025).with_index do |d, i|
+      puts "#{i.fdiv(count) * 100 + 9.788097334775}%: #{d.id}"
```
sihugh: 😂

PR #5455, opened 2020-03-12
Export alternative_path instead of alternative_url

emmabeynon:
When checking an imported unpublishing alternative URL against the alternative
path present in Publishing API, we have found that Whitehall sends the path,
not the URL, to Publishing API. As such, the check was failing. This changes
the Document Export Presenter to send the unpublishing alternative path instead
so that these conflicts can be avoided.

### DIFF
--- a/app/presenters/document_export_presenter.rb
+++ b/app/presenters/document_export_presenter.rb
@@ -148,19 +148,23 @@ private
   def present_unpublishing(edition)
     return unless edition.unpublishing
 
+    unpublishing_data = {
+      unpublishing_reason: edition.unpublishing.unpublishing_reason.name,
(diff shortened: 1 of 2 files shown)


kevindew:
Oh just a thought - are all Whitehall alternative_url GOV.UK pages? or do some redirect to external sites?

emmabeynon:
@kevindew Oh, that's disappointing 😞 I thought [this](https://github.com/alphagov/whitehall/blob/master/app/models/unpublishing.rb#L10https://github.com/alphagov/whitehall/blob/master/app/models/unpublishing.rb#L10) meant it had to be in a GOV.UK URL format.  I'll look into this further.

PR #5456, opened 2020-03-12
Fix buggy command paper number prefixes

ChrisBAshton:
This PR:

- fixes the command paper number regex, which previously allowed `.`, which allows any character in a regex. It is now escaped (`\.`), therefore only a period is allowed at that point in the regex
- tightens up the spacing and capitalisation validation, as many variations exist in the wild
- adds support for the new `CP` prefix, which differs from the others as it should not have a `.` period after it
- adds validation for suffixes (only capitalised Roman numerals are allowed)
- updates the validation message accordingly
- adds some hint text to the form, covering the more stringent validation checks
- adds a rake task to run which fixes up the existing command paper numbers and provides detailed logs.

Trello: https://trello.com/c/yG4gXFxI/1544-fix-buggy-command-paper-number-prefixes-in-whitehall

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -18,11 +18,6 @@ class Attachment < ApplicationRecord
   validates :attachable, presence: true
   validates :title, presence: true, length: { maximum: 255 }
   validates :isbn, isbn_format: true, allow_blank: true
-  validates :command_paper_number, format: {
-    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.map { |prefix| Regexp.escape(prefix) }.join('|')}) \d+/,
-    allow_blank: true,
-    message: "is invalid. The number must start with one of #{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}, followed by a space",
-  }
   validates :unique_reference, length: { maximum: 255 }, allow_blank: true
 
   scope :with_filename, ->(basename) {
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -12,7 +12,7 @@ class Attachment < ApplicationRecord
   before_save :nilify_locale_if_blank
   before_save :prevent_saving_of_abstract_base_class
 
-  VALID_COMMAND_PAPER_NUMBER_PREFIXES = ["C.", "Cd.", "Cmd.", "Cmnd.", "Cm."].freeze
+  VALID_COMMAND_PAPER_NUMBER_PREFIXES = ["CP", "C.", "Cd.", "Cmd.", "Cmnd.", "Cm."].freeze
 
   validates_with AttachmentValidator
   validates :attachable, presence: true
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -19,9 +19,9 @@ class Attachment < ApplicationRecord
   validates :title, presence: true, length: { maximum: 255 }
   validates :isbn, isbn_format: true, allow_blank: true
   validates :command_paper_number, format: {
-    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.map { |prefix| Regexp.escape(prefix) }.join('|')}) ?\d+/,
+    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.map { |prefix| Regexp.escape(prefix) }.join('|')}) \d+/,
     allow_blank: true,
-    message: "is invalid. The number must start with one of #{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}",
+    message: "is invalid. The number must start with one of #{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}, followed by a space",
   }
   validates :unique_reference, length: { maximum: 255 }, allow_blank: true
 
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -19,7 +19,7 @@ class Attachment < ApplicationRecord
   validates :title, presence: true, length: { maximum: 255 }
   validates :isbn, isbn_format: true, allow_blank: true
   validates :command_paper_number, format: {
-    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join('|')}) ?\d+/,
+    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.map { |prefix| Regexp.escape(prefix) }.join('|')}) ?\d+/,
     allow_blank: true,
     message: "is invalid. The number must start with one of #{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}",
   }
--- a/lib/data_hygiene/attachment_attribute_updater.rb
+++ b/lib/data_hygiene/attachment_attribute_updater.rb
@@ -1,0 +1,43 @@
+module DataHygiene
+  class AttachmentAttributeUpdater
+    def initialize(attachment, dry_run:)
+      @attachment = attachment
+      @dry_run = dry_run
+    end
+
+    def call
+      old_number = attachment.command_paper_number
+      valid_number = fix_command_paper_number(old_number)
+
+      unless AttachmentValidator.new.command_paper_number_valid?(valid_number)
+        raise DataHygiene::AttachmentAttributeNotFixable.new
+      end
+
+      return valid_number if dry_run || old_number == valid_number
+
+      attachment.update!(command_paper_number: valid_number)
+      valid_number
+    end
+
+    def self.call(*args)
+      new(*args).call
+    end
+
+    private_class_method :new
+
+  private
+
+    attr_reader :attachment, :dry_run
+
+    def fix_command_paper_number(original_number)
+      # for example, "CM.123-iv "
+      original_number.tr(". ", "") # remove periods and spaces: "CM123-iv"
+        .sub(/(\d+)/) { |number| ". #{number}" } # inject period & space before number: "CM. 123-iv"
+        .capitalize # "Cm. 123-iv"
+        .sub(/^Cp\./, "CP") # fix the special case "CP", which should be all caps and no period
+        .sub(/\d+-(.+)$/, &:upcase) # make suffix uppercase: "Cm. 123-IV"
+    end
+  end
+
+  class AttachmentAttributeNotFixable < StandardError; end
+end
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -53,9 +53,10 @@ namespace :data_hygiene do
     end
   end
 
-  desc "Make attachments' command paper numbers valid"
+  desc "Make attachments' command paper numbers valid and republish affected documents"
   namespace :make_command_paper_numbers_valid do
     def call_attachment_attribute_updater(dry_run:)
+      affected_attachments = []
       Attachment
         .where.not(command_paper_number: nil)
         .where.not(command_paper_number: "")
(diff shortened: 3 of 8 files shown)


test/unit/attachment_test.rb
```diff
@@ -58,11 +58,16 @@ class AttachmentTest < ActiveSupport::TestCase
...
       attachment = build(:file_attachment, command_paper_number: "#{prefix} 1234")
       assert_not attachment.valid?
-      expected_message = "is invalid. The number must start with one of #{Attachment::VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}"
+      expected_message = "is invalid. The number must start with one of #{Attachment::VALID_COMMAND_PAPER_NUMBER_PREFIXES.join(', ')}, followed by a space"
       assert attachment.errors[:command_paper_number].include?(expected_message)
     end
   end
 
+  test "should be invalid when the command paper number has no space after the prefix" do
+    attachment = build(:file_attachment, command_paper_number: "C1234")
```
ChrisBAshton: To avoid hard-coding, I could change this to:

```suggestion
    attachment = build(:file_attachment, command_paper_number: "#{Attachment::VALID_COMMAND_PAPER_NUMBER_PREFIXES.first}1234")
```

But it feels quite heavy handed and less readable. I invite feedback!

app/models/attachment.rb
```diff
@@ -12,16 +12,16 @@ class Attachment < ApplicationRecord
...
+  VALID_COMMAND_PAPER_NUMBER_PREFIXES = ["C.", "Cd.", "Cmd.", "Cmnd.", "Cm.", "CP"].freeze
 
   validates_with AttachmentValidator
   validates :attachable, presence: true
   validates :title, presence: true, length: { maximum: 255 }
   validates :isbn, isbn_format: true, allow_blank: true
   validates :command_paper_number, format: {
-    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join('|')}) ?\d+/,
+    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.map { |prefix| prefix.gsub(/\./, "\\.") }.join('|')}) \d+/,
     allow_blank: true,
```
GDSNewt: Just curious as to what allow_blank does?

ChrisBAshton: It allows `command_paper_number` to be an empty string (""), which would otherwise fail the regex on line 22.

test/unit/attachment_validator_test.rb
```diff
@@ -68,4 +68,43 @@ def assert_error_message(expectation, errors)
...
+      @validator.validate(attachment)
+      assert_not attachment.errors[:command_paper_number].empty?
+    end
+  end
+
+  test "should be invalid when the command paper number has no space after the prefix" do
+    attachment = build(:file_attachment, command_paper_number: "C1234")
+    @validator.validate(attachment)
+    assert_not attachment.errors[:command_paper_number].empty?
+  end
```
GDSNewt: What is the rationale behind two sets of tests? It seems that it testing the validation you are effective running the tests in _attachment-test_ again, albeit with an additional step. Would it not be sufficient to just have _attachment_validator_test.rb_?

test/unit/attachment_test.rb
GDSNewt: I agree it's less readable. I'd say hard coding like this is acceptable for a test, I suppose it depends on how likely the prefixes are to change. Also C1234 is an invalid prefix, not only because the lack of space but the lack of a period no? I accept this is the nit pickiest thing in the universe 😄

test/unit/attachment_validator_test.rb
ChrisBAshton: Whoops, good spot! This was a rebase that went wrong. I'll fix it up now 👍 thanks

test/unit/attachment_test.rb
ChrisBAshton: Thanks - updated.

app/models/attachment.rb
```diff
@@ -19,7 +19,7 @@ class Attachment < ApplicationRecord
   validates :title, presence: true, length: { maximum: 255 }
   validates :isbn, isbn_format: true, allow_blank: true
   validates :command_paper_number, format: {
-    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.join('|')}) ?\d+/,
+    with: /\A(#{VALID_COMMAND_PAPER_NUMBER_PREFIXES.map { |prefix| prefix.gsub(/\./, "\\.") }.join('|')}) ?\d+/,
```
benthorner: Can we use `Regexp.escape` here instead?

https://ruby-doc.org/core-2.7.0/Regexp.html#method-c-escape

db/data_migration/20200311133826_make_command_paper_numbers_valid.rb
```diff
@@ -0,0 +1,12 @@
+Attachment
+  .where.not(command_paper_number: nil)
+  .where.not(command_paper_number: "")
+  .each do |attachment|
+  valid_number = attachment.command_paper_number # for example, "CM.123-iv "
+    .tr(". ", "") # remove periods and spaces: "CM123-iv"
+    .gsub(/(\d+)/) { |number| ". #{number}" } # inject period & space before number: "CM. 123-iv"
```
benthorner: I think we may need `sub` specifically here, as this could also match subsequent numbers e.g. `CM 123-1`.

db/data_migration/20200311133826_make_command_paper_numbers_valid.rb
```diff
@@ -0,0 +1,12 @@
...
+    .capitalize # "Cm. 123-iv"
+    .gsub(/Cp\./, "CP") # fix the special case "CP", which should be all caps and no period
```
benthorner: Minor: we could be more specific here with `sub(/^Cp/)`, as it's one substitution at the beginning of the string.

db/data_migration/20200311133826_make_command_paper_numbers_valid.rb
```diff
@@ -0,0 +1,12 @@
...
+    .gsub(/\d+-(.+)/, &:upcase) # make suffix uppercase: "Cm. 123-IV"
```
benthorner: Are we sure this will match the whole suffix? I'm wondering if we need a `$` or `\z` constraint.

db/data_migration/20200311133826_make_command_paper_numbers_valid.rb
ChrisBAshton: Good point. I think there's only one case where we have a number suffix, which content designers will fix manually, but still better to be safe here.

db/data_migration/20200311133826_make_command_paper_numbers_valid.rb
ChrisBAshton: It's not necessary but it does help to signify intent better 👍

app/models/attachment.rb
ChrisBAshton: Nice! 👍

lib/data_hygiene/attachment_attribute_not_fixable.rb
```diff
@@ -0,0 +1,3 @@
+module DataHygiene
```
benthorner: Minor: could we make this a subclass of `AttachmentAttributeUpdater`? This keeps the number of classes/files down, and makes it clear it's only used in this context.

lib/data_hygiene/attachment_attribute_updater.rb
```diff
@@ -0,0 +1,40 @@
...
+    def call
+      valid_number = fix_command_paper_number(attachment.command_paper_number)
+
+      unless AttachmentValidator.new.command_paper_number_valid?(valid_number)
+        raise DataHygiene::AttachmentAttributeNotFixable.new
+      end
+
+      return valid_number if dry_run
+
+      attachment.update(command_paper_number: valid_number)
```
benthorner: I think we should use `update!` here, just in case there's a problem with the save operation.

lib/tasks/data_hygiene.rake
```diff
@@ -52,4 +52,36 @@ namespace :data_hygiene do
       speech.save(validate: false)
     end
   end
+
+  desc "Make attachments' command paper numbers valid"
```
benthorner: Minor: we should have something somewhere to remove (all) this when we're done, as it won't be used again.

lib/tasks/data_hygiene.rake
```diff
@@ -52,4 +52,36 @@ namespace :data_hygiene do
...
+  namespace :make_command_paper_numbers_valid do
+    def call_attachment_attribute_updater(dry_run:)
+      Attachment
+        .where.not(command_paper_number: nil)
+        .where.not(command_paper_number: "")
+        .each do |attachment|
+        old_attr = attachment.command_paper_number
+        begin
+          new_attr = DataHygiene::AttachmentAttributeUpdater.call(attachment, dry_run: dry_run)
+          if old_attr == new_attr
```
benthorner: This isn't quite right, since the updater will have done an `update` already, unless you're relying on [ActiveRecord::Dirty](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html)?

test/unit/tasks/data_hygiene/attachment_attribute_updater_test.rb
```diff
@@ -0,0 +1,66 @@
+require "test_helper"
+
+class AttachmentAttributeUpdaterTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  describe DataHygiene::AttachmentAttributeUpdater do
+    let(:attachment) { build(:file_attachment, attachment_data: build(:attachment_data), command_paper_number: original_command_paper_number) }
+    let(:original_command_paper_number) { "CM123" }
```
benthorner: Minor: in general it can be confusing to have a top-level `let` that gets overridden further down. In this case, we could move it into the individual `context`s, and the first `it` for a real run.

lib/data_hygiene/attachment_attribute_updater.rb
benthorner: If something does go wrong, we'll want this to be in a transaction (that'll have to go in the rake task), in case we lose sight of the ones we've changed.

lib/tasks/data_hygiene.rake
ChrisBAshton: Ah yes, good spot. At one point this was true, when I was handling the `update` call in the rake task itself. I'll fix it 👍

lib/tasks/data_hygiene.rake
ChrisBAshton: You mean just a note in the description? Like this?

```suggestion
  desc "Make attachments' command paper numbers valid (temporary. Will delete task after use)"
```

benthorner: Hmm...I was thinking more of a follow-up task, or a 'DO NOT MERGE' PR to remove it.

ChrisBAshton: Have now added this as a To-do item on the card.

ChrisBAshton:
Hi @benthorner - I've now added functionality for republishing the affected documents, as part of the rake task (see second-to-last commit). Here is a successful job (dry run) on Integration: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/130260/console

I haven't added tests for it, but it seems there's a precedent for not testing this kind of thing in rake tasks. I think the effort taken to test it would not be worth it, especially given we're deleting the code afterwards, we're recycling prior art, and I've manually tested it on Integration.

With that in mind, could you give it another look please? 😁

lib/tasks/data_hygiene.rake
```diff
@@ -65,15 +66,44 @@ namespace :data_hygiene do
...
+      puts "Attachments updated: #{affected_attachments.map(&:id).join(', ')}"
+
+      edition_ids = affected_attachments.map do |attachment|
+        case attachment.attachable_type
+        when "Edition"
+          attachment.attachable_id
+        when "Response"
+          Response.where(id: attachment.attachable_id).pluck(:edition_id).first
+        else
+          raise "Unknown attachable type #{attachment.attachable_type}"
```
benthorner: This isn't ideal, as if this happens during the real run, we could lose all the data about attachments already changed, unless we'd previously done a dry run and saved it. As long as we're doing that, I guess it doesn't matter. We could always put this whole task in a transaction, as a quick fix for that.

benthorner: I see L82 also accounts for that somewhat, as well 👍

lib/tasks/data_hygiene.rake
```diff
@@ -65,15 +66,44 @@ namespace :data_hygiene do
...
           end
         rescue DataHygiene::AttachmentAttributeNotFixable
           puts "Attachment ID #{attachment.id} command paper number cannot be fixed automatically: #{old_attr}"
         end
       end
+
+      puts "Attachments updated: #{affected_attachments.map(&:id).join(', ')}"
+
+      edition_ids = affected_attachments.map do |attachment|
+        case attachment.attachable_type
```
benthorner: I think you can just do `attachment.attachable` instead of this.

lib/tasks/data_hygiene.rake
```diff
@@ -65,15 +66,44 @@ namespace :data_hygiene do
...
+      document_ids = Document.joins(:editions).where("editions.id": edition_ids).pluck(:document_id).uniq
+
+      puts "Documents affected: #{document_ids.join(', ')}"
+
+      if dry_run
+        puts "Dry run. Skipping sending #{document_ids.count} documents to Publishing API"
+      else
+        puts "Sending #{document_ids.count} documents to Publishing API"
+        document_ids.each do |doc_id|
+          PublishingApiDocumentRepublishingWorker.perform_async(doc_id)
```
benthorner: Could we do this synchronously, just to be sure?

lib/tasks/data_hygiene.rake
ChrisBAshton: That's fair 👍 L82 wouldn't be reached if the exception is raised, and we wouldn't know which attachments we've updated if we hadn't run a dry run first. I'll look at putting all this in a transaction.

PR #5457, opened 2020-03-13
Remove reference to withStatsdTiming

barrucadu:
This was removed in https://github.com/alphagov/govuk-jenkinslib/pull/69


thomasleese:
Too slow :sad:

PR #5464, opened 2020-03-16
Revert world location signup

1pretz1:
We want to revert some of the earlier work done to reduce the amount of email alert signup paths. 

This worked for most signup journeys (people, topical events etc) but actually broke for world locations. The change that lead to this bug came about when we moved the creation of the subscriber list from whitehall to email-alert-frontend as it turns out it's only possible to create a valid subscriber list for a world location from the information locally to whitehall for these world locations - world news, international delegations, world latest. The reason we can't signup users outside of whitehall is because the world locations previously listed don't really exist outside of whitehall, as in they don't have a content item in content-store and so all the data about them is only stored in whitehall (content-id, title etc).

### Whitehall email signup url
Screenshot 2020-03-16 at 11 28 56

### Whitehall email signup page
Screenshot 2020-03-16 at 11 29 07

### Emails being generated for world location Spain (aws staging)
Screenshot 2020-03-16 at 12 57 14

Further details on the exact cause can be found within the incident document created for this bug - https://docs.google.com/document/d/1fL5mQjRX2ApoVd4rOIxrnwjhrC-w4pgNgNXaA93KeM8/edit#heading=h.p99426yo0rbv.

Reverting some of this PR:
https://github.com/alphagov/whitehall/pull/4996

Trello:
https://trello.com/c/lkD6aJ9v/1568-fix-world-locations-email-signup-bug

### DIFF
--- a/app/helpers/email_signup_helper.rb
+++ b/app/helpers/email_signup_helper.rb
@@ -1,10 +1,16 @@
 module EmailSignupHelper
   def email_signup_path(atom_feed_url)
+    feed_uri = URI.parse(atom_feed_url)
+
     if EmailSignupPagesFinder.exists_for_atom_feed?(atom_feed_url)
       organisation_email_signup_information_path_from_atom_feed(atom_feed_url)
+    elsif feed_uri.path.match(%r{^/world/(.*)\.atom$})
+      new_email_signups_path(email_signup: { feed: atom_feed_url })
     else
-      email_alert_frontend_signup_url(atom_feed_url)
+      email_alert_frontend_signup_url(feed_uri)
     end
+  rescue URI::InvalidURIError
+    "/"
   end
 
 private
@@ -17,8 +23,7 @@ private
     /\/([\w-]+).atom$/.match(atom_feed_url)[1]
   end
 
-  def email_alert_frontend_signup_url(feed_url)
-    base_path = URI.parse(feed_url).path.chomp(".atom")
-    "#{Plek.new.website_root}/email-signup?link=#{base_path}"
+  def email_alert_frontend_signup_url(feed_uri)
+    "#{Plek.new.website_root}/email-signup?link=#{feed_uri.path.chomp('.atom')}"
   end
 end
--- a/app/models/world_location_email_signup.rb
+++ b/app/models/world_location_email_signup.rb
@@ -1,0 +1,48 @@
+class WorldLocationEmailSignup
+  attr_reader :feed
+
+  def initialize(feed)
+    @feed = feed
+  end
+
+  def signup_url
+    subscriber_list = Services.email_alert_api.find_or_create_subscriber_list(criteria)
+    subscriber_list["subscriber_list"]["subscription_url"]
+  end
+
+  def name
+    world_location.name
+  end
+
+  def valid?
+    uri && world_location_slug && world_location.present?
+  end
+
+private
+
+  def criteria
+    {
+      "links" => {
+        "world_locations" => [
+          world_location.content_id,
+        ],
+      },
+    }.merge("title" => world_location.name)
+  end
+
+  def world_location
+    @world_location ||= WorldLocation.find_by(slug: uri.path.match(%r{^/world/(.*)\.atom$})[1])
+  end
+
+  def uri
+    @uri ||= begin
+               URI.parse(feed)
+             rescue URI::InvalidURIError
+               nil
+             end
+  end
+
+  def world_location_slug
+    uri.path.match(%r{^/world/(.*)\.atom$}).try(:[], 1)
+  end
+end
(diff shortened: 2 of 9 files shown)


app/helpers/email_signup_helper.rb
```diff
@@ -2,6 +2,8 @@ module EmailSignupHelper
   def email_signup_path(atom_feed_url)
     if EmailSignupPagesFinder.exists_for_atom_feed?(atom_feed_url)
       organisation_email_signup_information_path_from_atom_feed(atom_feed_url)
+    elsif URI.parse(atom_feed_url).path.match(%r{^/world/(.*)\.atom$})
```
kevindew: Do we need to rescue a `URI::InvalidURIError` like we do in WorldLocationEmailSignup model? not sure if this is user input or not.

app/models/world_location_email_signup.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  def valid?
+    uri && recognised_url? && world_feed_type? && world_location.present?
+  end
+
+private
+
+  def ensure_subscriber_list_exists
+    @ensure_subscriber_list_exists ||= Services.email_alert_api.find_or_create_subscriber_list(criteria)
```
kevindew: This looks pretty whacky. I know you didn't write it originally

I'd suggest you change this method to be called subscriber_list

and then change the save method to be:

```
def save
    return unless valid?

    subscriber_list
  end
```

not too sure why it was previously set to booleans as it doesn't really need them

app/models/world_location_email_signup.rb
```diff
@@ -0,0 +1,80 @@
...
+  def topic_id
+    ensure_subscriber_list_exists["subscriber_list"]["gov_delivery_id"]
+  end
+
+  def name
+    world_location.name
+  end
+
+  def valid?
+    uri && recognised_url? && world_feed_type? && world_location.present?
```
kevindew: Could we do more logic in this method or something to reduce the spread of this across so many methods. I know you didn't write it originally.

thinking it could be like:

```
def valid?
  return false unless uri
  return false unless uri.host == Whitehall.public_host && uri.scheme == Whitehall.public_protocol

  recognised_path = Rails.application.routes.recognize_path(uri.path)
  valid_extension = File.extname(uri.path) == ".atom"   
  world_feed_type = uri.path.split("/")[1] == "world"

  recognised_path && valid_extension && world_feed_type && world_location.present?
rescue ActionController::RoutingError
  false
end
```

I'm also not sure we need this much validation? a lot of it feels superfluous. We could probably replace most of this logic with say:

```
def valid?
  uri && world_location_slug && world_location.present?
end

def world_location_slug
  uri.path.match(%r{^/world/(.*)\.atom$})[1]
end

def world_location
  @world_location ||= WorldLocation.find_by(slug: world_location_slug)
end
```

app/models/world_location_email_signup.rb
```diff
@@ -0,0 +1,80 @@
...
+    ensure_subscriber_list_exists
+    true
+  end
+
+  def signup_url
+    ensure_subscriber_list_exists["subscriber_list"]["subscription_url"]
+  end
+
+  def topic_id
+    ensure_subscriber_list_exists["subscriber_list"]["gov_delivery_id"]
```
kevindew: Is there a more modern field we can use instead of gov_delivery_id it feels rather legacy?

app/helpers/email_signup_helper.rb
1pretz1: Hmmm I don't think so as the atom_feed_url in most cases (where I've looked) comes from a method call on an instance from the db and most of the time it would bomb before getting to this point however because this method is used indirectly in quite a few places I've added the rescue which points the user to the homepage if it's invalid as a precaution.

app/models/world_location_email_signup.rb
1pretz1: True, I've gone one step further and removed some of these methods as they aren't actually needed.

app/models/world_location_email_signup.rb
1pretz1: Good shout, I didn't know how much of the validating I should pull across so I've gone with your suggestion.

app/models/world_location_email_signup.rb
1pretz1: Tbh I've just removed this bit of the code as it's not used and I'm not sure why I pulled it across from the original PR.

1pretz1:
Thanks for the review @kevindew, I've addressed your comments and chopped some more of the code down which isn't really needed. What do you think?

app/models/world_location_email_signup.rb
```diff
@@ -0,0 +1,52 @@
...
+               nil
+             end
+  end
+
+  def world_location_slug
+    world_path = uri.path.match(%r{^/world/(.*)\.atom$})
+
+    return unless world_path
+
+    world_path[1]
```
kevindew: Minor - you could get this down to a one liner with `uri.path.match(%r{^/world/(.*)\.atom$}).try(:[], 1)`

1pretz1: Nice one liner, saved 4 lines! 🔻

PR #5468, opened 2020-03-18
Revert "Fix buggy command paper number prefixes"

ChrisBAshton:
Note: the commits look horrid here, I had to do some intense rebasing and cherry-picking.
My attempt at flattening the commits led to there being [no changes against master](https://github.com/alphagov/whitehall/compare/tmp-revert?expand=1), so I propose we use the "Squash and merge" button on this PR.

---

Unfortunately this had an unintended side effect, whereby
attempting to create a new edition of a Publication that contains
an attachment that fails validation, fails to copy the attachments
across to the new edition.

This wouldn't be too much of a problem ordinarily, as we would run
the rake task that fixes the invalid attachment properties, but
we identified a few that would need manually fixing and they cannot
be manually fixed if we cannot create a new edition in which to
fix them.

The fix will be to follow the same approach as editions:
https://github.com/alphagov/whitehall/commit/3561dfb5ab650cd579dba3c59252f7b5d11dbb3a#diff-d9c2d399fb6d507d2c0c4da96feece4eR481

...but for when deep cloning attachments in Attachable model:
https://github.com/alphagov/whitehall/commit/7890625f1810df4398e66c5486ff71155f4a87d8#diff-ffbae7ee351ead6d9f0e5780363c0820R12


lib/tasks/publishing_api.rake
```diff
@@ -1,6 +1,28 @@
 require "gds_api/publishing_api/special_route_publisher"
 
 namespace :publishing_api do
+  # this will be removed after being run once
```
benthorner: Should this be in here?

ChrisBAshton: Whoops - good spot.

ChrisBAshton:
Argh I clicked merge by mistake 😫 so frustrating!!

benthorner:
Oops. Now everyone will know 😆

PR #5476, opened 2020-03-20
Use created_at of the unpublishing for unpublished_at date

thomasleese:
This changes how the `unpublished_at` date of a withdrawn document is found. Previously we were using the `updated_at` of the edition, which has the potential to change (most obviously through a bulk changing data migration). By using the `created_at` of the `unpublishing`, this date remains stable even if the editions change.

See individual commits for more details. Once this is merged, I will republish all the withdrawn documents where the `updated_at` doesn't match `unpublishing.created_at`.

[Trello Card](https://trello.com/c/wyvPYlI6/1837-5-investigate-why-some-withdrawn-documents-have-stray-drafts) ・ [Zendesk Ticket](https://govuk.zendesk.com/agent/tickets/3929420)

### DIFF
--- a/app/workers/publishing_api_withdrawal_worker.rb
+++ b/app/workers/publishing_api_withdrawal_worker.rb
@@ -1,13 +1,10 @@
 class PublishingApiWithdrawalWorker < PublishingApiWorker
-  def perform(content_id, explanation, locale, allow_draft = false)
+  # `explanation` and `unpublished_at` come from the unpublishing object. Rather than
+  # performing a database query here to look up the `unpublishing` linked to the most
+  # recent edition, we pass it in directly because the `unpublishing` isn't always
+  # saved in the database yet when this worker runs.
+  def perform(content_id, explanation, locale, allow_draft = false, unpublished_at = nil)
     check_if_locked_document(content_id: content_id)
-
-    unpublished_at = Edition
-      .joins(:document)
-      .where(documents: { content_id: content_id })
-      .where(state: "withdrawn")
-      .pluck(:updated_at)
-      .first
 
     Services.publishing_api.unpublish(
       content_id,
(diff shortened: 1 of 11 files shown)


thomasleese:
> I think we should at a bit more documentation around `given_unpublishing_at`, as it's unclear why this is introduced here, instead of just looking up the unpublishing, or passing the ID of it. From memory, I think this is because the unpublishing hasn't always been saved when this code is called (can't remember from where, though). I think it would help someone else reading this.

Actually the main reason I did it like this was to avoid an unnecessary database query when in all cases the caller of this worker (either directly or through Sidekiq) has already looked up (or just created) the unpublishing instance. Although I think you're right that sometimes the unpublishing object isn't yet in the database. It's not necessarily how I would design the worker, but it does follow the existing pattern with the `explanation` variable that again just comes direct from the `unpublishing` object.

Anyway, I've amended the commit with a comment explaining where the arguments come from - is that what you were thinking?

app/workers/publishing_api_withdrawal_worker.rb
```diff
@@ -1,24 +1,40 @@
...
   end
+
+private
+
+  def find_unpublished_at(content_id, given_unpublishing_at)
+    if given_unpublishing_at
+      # We call this job both directly and via Sidekiq. When called by Sidekiq, the date gets turned
+      # into a string (because jobs must be JSON serialisable) and `Services.publishing_api.unpublish`
+      # rejects it.
+      given_unpublishing_at.respond_to?(:utc) ? given_unpublishing_at : Time.zone.parse(given_unpublishing_at)
```
benthorner: Minor: we could make this slightly neater with `Time.zone.parse(given_unpublishing_at.to_s)`. Also, I think you mean `unpublish*ed*_at`.

PR #5478, opened 2020-03-20
Use attachment ID as publishing-api ID

barrucadu:
Using filename for file attachments isn't safe, as a document could
have two attachments with the same filename.  Using "{id}-{filename}"
would work but, in that case, why not just use the id directly?

There's no need to republish docs after this, as the old IDs are still
valid.


barrucadu:
@cbaines How about this?

cbaines:
> @cbaines How about this?

Yeah, that's great. Thanks :)

PR #5479, opened 2020-03-20
Fix the only_broken_links Edition method

cbaines:
I think the purpose of this is to limit the scope of the query to
editions with broken links. There can be more than one report for a
single edition though, which makes sense. This query however makes no
sense, as it doesn't just look at the latest report, but any
report. So it's more like asking what editions have ever had a broken
link!

I was looking in to this because of reports of it not working due to
timeouts.

This commit changes the query to join on just the latest reports for
each edition, and then only select editions where a broken link exists
for that latest report. I don't know much about MySQL, let alone
ActiveRecord, hence the big blob of SQL.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -207,10 +207,25 @@ class Edition < ApplicationRecord
   end
 
   def self.only_broken_links
-    joins("INNER JOIN link_checker_api_reports ON link_checker_api_reports.link_reportable_id = editions.id")
-    .joins("INNER JOIN link_checker_api_report_links ON link_checker_api_report_links.link_checker_api_report_id = link_checker_api_reports.id")
-    .where("link_checker_api_reports.link_reportable_type = 'Edition' AND link_checker_api_report_links.status != 'ok'")
-    .distinct
+    joins(
+      "
+LEFT JOIN (
+  SELECT id, link_reportable_type, link_reportable_id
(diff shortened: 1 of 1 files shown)


app/models/edition.rb
```diff
@@ -207,10 +207,25 @@ def self.to_date(date)
   end
 
   def self.only_broken_links
-    joins("INNER JOIN link_checker_api_reports ON link_checker_api_reports.link_reportable_id = editions.id")
-    .joins("INNER JOIN link_checker_api_report_links ON link_checker_api_report_links.link_checker_api_report_id = link_checker_api_reports.id")
-    .where("link_checker_api_reports.link_reportable_type = 'Edition' AND link_checker_api_report_links.status != 'ok'")
```
ChrisBAshton: I _think_ you could accomplish this with an additional `where` rather than a full rewrite:

```suggestion
    .where("link_checker_api_reports.id = (SELECT MAX(id) FROM link_checker_api_reports WHERE link_checker_api_reports. link_reportable_id = editions.id)")`
    .where("link_checker_api_reports.link_reportable_type = 'Edition' AND link_checker_api_report_links.status != 'ok'")`
```

cbaines: Yeah, that looks like it might work, although I think the `GROUP BY` approach is a little neater.

PR #5492, opened 2020-03-24
Use sidekiq for deleting old link checker data

cbaines:


config/sidekiq.yml
```diff
@@ -14,3 +14,5 @@
   check_all_organisations_links_worker:
     cron: '0 4 * * *' # Runs at 4 a.m every day
     class: CheckAllOrganisationsLinksWorker
+  LinkCheckerDeleteOldDataWorker:
+    cron: '30 3 * * *' # Runs at 4 a.m every day
```
thomasleese:
```suggestion
    cron: '30 3 * * *' # Runs at 3:30 a.m every day
```

cbaines: Good spot :+1:

PR #5539, opened 2020-04-14
Remove organisations from whitehall frontend

barrucadu:
I found this old branch of whitehall.  It changed the rendering app for org pages from

```
    def rendering_app
      if court_or_tribunal?
        Whitehall::RenderingApp::WHITEHALL_FRONTEND
      else
        Whitehall::RenderingApp::COLLECTIONS_FRONTEND
      end
    end
```

To just `Whitehall::RenderingApp::COLLECTIONS_FRONTEND`, and then deleted a bunch of unused frontend code.  The branch was never merged.

Since then, the rendering app change was made in #4864, but the frontend code was never deleted.  I've rebased it and removed a bunch of failing feature tests.

---

[Trello card](https://trello.com/c/ZMCkw9FQ/1628-remove-organisation-rendering-from-whitehall-frontend)


barrucadu:
@sihugh Is any of this frontend code still needed?  Or was removing it just missed in the other PR?

barrucadu:
It seems there are still a fair few requests hitting these pages because of the prefix routing, eg:

- `/government/organisations/ministry-of-justice/about.html` (different page to `/government/organisations/ministry-of-justice/about`)
- `/government/organisations/uk-visas-and-immigration%20` (different page to `/government/organisations/uk-visas-and-immigration`)

barrucadu:
@thomasleese Do we want to put something in place to redirect the bad URLs, or do we just accept that they're going to start 404ing?  We're not ensuring feature parity between collections and whitehall-frontend, so they're already broken in a sense.

I've checked that paegs like https://www-origin.integration.publishing.service.gov.uk/government/organisations/nuclear-decommissioning-authority/about.cy and https://www-origin.integration.publishing.service.gov.uk/government/organisations/medicines-and-healthcare-products-regulatory-agency/email-signup still work with this branch

thomasleese:
I don't think we need to put in redirects.

This feels like something we're going to hit more and more as we move things over to Collections, and staying on top of it seems like an impossible task. Getting rid of the `/government` prefix route is a long term goal of this migration, and adding redirections for this will make that harder.

sihugh:
There were a bunch of other page types that were using org partials.  The /courts-tribunals page is still rendered from whitehall.  
I wanted to remove all this when I could get rid of everything, but that time has never quite arrived.  There's also a bunch of assets that might be removable.

barrucadu:
I don't think the deleted partials are still in use anywhere, and `/courts-tribunals` works with this branch

barrucadu:
I'll merge this today and leave it on staging for a little bit when I get to deploying it as a final check

PR #5557, opened 2020-04-23
Update PublishingApiVanishWorker to not take strong args

1pretz1:
This commit stop's using strong arguments when calling the vanish
worker. According to sidekiq you can't use key / value arguments when
calling perform_async but you apparently can when calling perform
synchronously (why... :facepalm:). This led to the tests not catching
the issue and only finding out when trying to run the associated data
migration. I've also updated the tests to reflect the actual use cases.

See the issue here -> https://github.com/mperham/sidekiq/issues/2372.

```
PublishingApiVanishWorker.perform_async('s', 's', discard_drafts: "S")
Sidekiq::RetrySet.new.select { |x| x }.last.item["error_message"]
-> "wrong number of arguments (given 3, expected 2)"
```

Trello:
https://trello.com/c/ygOqmhWw/1921-remove-legacy-placeholderworkinggroup-code-data-from-govuks-codebase

### DIFF
--- a/test/unit/workers/publishing_api_vanish_worker_test.rb
+++ b/test/unit/workers/publishing_api_vanish_worker_test.rb
@@ -8,16 +8,15 @@ class PublishingApiVanishWorkerTest < ActiveSupport::TestCase
     publication = create(:withdrawn_publication)
 
     request = stub_publishing_api_unpublish(
-      publication.document.content_id,
+      publication.content_id,
       body: {
         type: "vanish",
         locale: "en",
       },
     )
 
-    PublishingApiVanishWorker.new.perform(
-      publication.document.content_id, "en"
-    )
+    PublishingApiVanishWorker.perform_async(publication.content_id, "en")
+    PublishingApiVanishWorker.drain
 
     assert_requested request
   end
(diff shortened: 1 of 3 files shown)


test/unit/workers/publishing_api_vanish_worker_test.rb
```diff
@@ -34,9 +33,8 @@ class PublishingApiVanishWorkerTest < ActiveSupport::TestCase
       },
     )
 
-    PublishingApiVanishWorker.new.perform(
-      publication.document.content_id, "en", discard_drafts: true
-    )
+    PublishingApiVanishWorker.perform_async(publication.content_id, "en", true)
```
ChrisBAshton, line 36: This should be:

```suggestion
    PublishingApiVanishWorker.perform_async(publication.document.content_id, "en", true)
```

I don't think it makes a difference, mind.

1pretz1: You can fetch the content_id directly from the edition (publication) so it seems superfluous to explicitly call the document. However your point highlighted that the test file wasn't consistent which has now been addressed, good spot!

PR #5567, opened 2020-04-28
Do not present "ministers" links to pub-api

barrucadu:
This link type was [deprecated in 2018](https://github.com/alphagov/govuk-content-schemas/commit/100723ddca309a65d80d03493bf61ba7e91d5366), and doesn't seem to be used any
more.

---

[Trello card](https://trello.com/c/i1ffrXiZ/1913-5-link-ministers-to-the-government-ministers-content-item)


thomasleese:
Are we able to get rid of [the `key` argument](https://github.com/alphagov/whitehall/blob/6a00e10c54470b4d3e1c15782c5b5931be19afdf/app/presenters/publishing_api/payload_builder/people.rb#L12) to the payload builder?

PR #5585, opened 2020-05-07
[DO NOT MERGE] Noindex withdrawn content for crawlers

sihugh:
** This has been paused until we sort out the proposition/comms **

We publish a sitemap with down-weighting for withdrawn content, however this does not seem to affect the ranking of external search results as much as we might hope.

By no-indexing here, we hope to surface current information more readily in places like google.

Should users wish to find withdrawn content, then it'll be findable via site search or navigation.

https://trello.com/c/YcXJ8OFD/258-noindex-withdrawn-content

### DIFF
--- a/app/views/documents/_header.html.erb
+++ b/app/views/documents/_header.html.erb
@@ -17,6 +17,9 @@
   )
 %>
 
+<%# This will be hoisted into <head> by Slimmer %>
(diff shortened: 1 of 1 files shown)


DilwoarH:
Look good - do we need any tests?

sihugh:
We've seen that this is only affecting a few pages, so we're not going to do this now

PR #5595, opened 2020-05-11
Present internal taxon name to publishers

sihugh:
This will help give more context to areas of the taxonomy that are mixed, such as the new coronavirus branch.  This has areas that collide with others (such as education or business topics), so additional information in the publishing tool is useful.


sihugh:
Not doing this for now

PR #5604, opened 2020-05-14
Update change note form text

edwardkerry:
This PR updates the text on change note radio buttons, and removes the placeholder text from the text box. 

It uses JS to append some text between a form label and text box, as there is not currently a straightforward way of implementing this using our custom FormBuilder.

Before:
before

After:
Screenshot 2020-05-18 at 17 02 49

[Trello](https://trello.com/c/XwabeJac/1966-2-changes-to-change-note-box-in-whitehall)

### DIFF
--- a/app/assets/javascripts/admin/views/editions/_form.js
+++ b/app/assets/javascripts/admin/views/editions/_form.js
@@ -31,7 +31,7 @@
       function appendHintToFormLabel() {
         var $label = $($change_notes_section).find('label');
         var example_text = "For example, \"Edited chapter 6 - centres in Cardiff and Aberystwyth have closed.\""
-        var info_text = "The text will be published on the page and emailed to users signed up to alerts."
+        var info_text = "A change note will be published on the page and emailed to users signed up to alerts."
         $label.append('<p class="hint"; style="font-weight: normal">' + example_text + '<br>' + info_text + '</p>');
       }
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -23,10 +23,18 @@
...
+      var $change_notes_label        = $('.js-change-notes-section label');
 
       $radio_buttons.change(showOrHideChangeNotes);
+      appendHintToFormLabel();
       showOrHideChangeNotes();
 
+      function appendHintToFormLabel() {
+        var example_text = "For example, \"Edited chapter 6 - centres in Cardiff and Aberystwyth have closed.\""
+        var info_text = "The text will be published on the page and emailed to users signed up to alerts."
+        $change_notes_label.append(`<p class="hint"; style='font-weight: normal'>${example_text}<br>${info_text}<p>`);
```
richardTowers, line 35: Does this JS get compiled into ES5? I think backticks are going to trip up old browsers if not.

richardTowers: https://caniuse.com/#feat=template-literals

If you want to be safe, I think:

```suggestion
        $change_notes_label.append('<p class="hint"; style="font-weight: normal">' +example_text + '<br>' + info_text + '<p>');
```

Is not much worse.

edwardkerry: ah, very good point, thank you. I'll make the change

ChrisBAshton:
What's the plausibility of us pulling in the [Form Textarea component](https://components.publishing.service.gov.uk/component-guide/textarea/with_hint) from the design system? It's got native support for the hint text.

edwardkerry:
> What's the plausibility of us pulling in the [Form Textarea component](https://components.publishing.service.gov.uk/component-guide/textarea/with_hint) from the design system? It's got native support for the hint text.

I think this is probably (definitely) the way to go in the long term but I'm not sure how simple it would be to introduce as part of this change. We're not making much use of `govuk-publishing-components` in Whitehall in general and I _think_ it might be quite tricky to introduce to the forms as a one-off.

PR #5610, opened 2020-05-15
Simplify deleting document collection groups

cbaines:
Rather than having the user delete the group items if there are any,
just handle that as part of deleting the group.

This also resolves an issue where a document associated with the group
with just one edition, which happens to be deleted, would prevent the
group from being deleted, as it's not visible to the user to remove
from the group.


cbaines:
> Looks good - but I would also link to our current Trello, and also link to the commit that introduced the feature in the first place and didn't explain why they thought it was needed.

Sorry, I completely forgot to do this prior to merging :(

PR #5623, opened 2020-05-20
Invoke mailer in a loop

benjamineskola:
Notify doesn't support sending to multiple recipients in a single API call, so we need to iterate over the list of recipients and invoke a separate instance of the mailer for each before we're able to migrate to that.

This is the technique we used in [short-url-manager](https://github.com/alphagov/short-url-manager/pull/451) and [publisher](https://github.com/alphagov/publisher/pull/1243).

https://trello.com/c/Q7xeFxCI/1993-8-whitehall-send-to-multiple-addresses-with-separate-mailer-instances-to-support-notify

### DIFF
--- a/app/mailers/multi_notifications.rb
+++ b/app/mailers/multi_notifications.rb
@@ -1,0 +1,15 @@
+class MultiNotifications < ApplicationMailer
+  def self.consultation_deadline_upcoming(consultation, weeks_left:, mailer: Notifications)
+    addresses = consultation.authors.pluck(:email).uniq
+    addresses.map do |address|
+      mailer.consultation_deadline_upcoming(consultation, weeks_left: weeks_left, recipient_address: address)
+    end
+  end
+
+  def self.consultation_deadline_passed(consultation, mailer: Notifications)
+    addresses = consultation.authors.pluck(:email).uniq
+    addresses.map do |address|
+      mailer.consultation_deadline_passed(consultation, recipient_address: address)
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/mailers/multi_notifications.rb
```diff
@@ -0,0 +1,15 @@
+class MultiNotifications < ApplicationMailer
+  def self.consultation_deadline_upcoming(consultation, weeks_left:, mailer: Notifications)
+    addresses = consultation.authors.uniq.map(&:email)
```
1pretz1, line 3: You could do a `consultation.authors.pluck(:email).uniq` here which would be more efficient

app/mailers/multi_notifications.rb
```diff
@@ -0,0 +1,15 @@
...
+    addresses.map do |address|
+      mailer.consultation_deadline_upcoming(consultation, weeks_left: weeks_left, recipient_address: address)
+    end
+  end
+
+  def self.consultation_deadline_passed(consultation, mailer: Notifications)
+    addresses = consultation.authors.uniq.map(&:email)
```
1pretz1, line 10: Same comment as the one above ^

test/functional/notifications_consultation_reminders_test.rb
```diff
@@ -3,15 +3,17 @@
...
     @consultation = build(:consultation)
+    author = build(:author)
+    @consultation.update(authors: [author, author])
   end
 
   test "reminder emails should contain the title text and weeks remaining" do
-    @email = Notifications.consultation_deadline_upcoming(@consultation, weeks_left: 2)
-    assert_includes @email.body.to_s, %(Publish the government response for "#{@consultation.title}" within 2 weeks)
+    @email = MultiNotifications.consultation_deadline_upcoming(@consultation, weeks_left: 2)
+    assert_includes @email.first.body.to_s, %(Publish the government response for "#{@consultation.title}" within 2 weeks)
```
1pretz1, line 12: Maybe also add an assert that checks only one email was sent - this would effectively check that one author doesn't receive two emails, what do you think?

PR #5640, opened 2020-05-28
Tweak DocumentListExportPresenterTest test to not depend on timezone

cbaines:


cbaines:
This is blocked by https://github.com/alphagov/whitehall/pull/5639

PR #5655, opened 2020-06-02
Prevent making call to Asset Manager if missing file information

ChrisBAshton:
We're seeing [issues in Sentry][sentry-issue] as follows:

> URL: https://asset-manager.production.govuk-internal.digital/assets/5b115d34ed915d2cc681ab6f
> Response body:
> {"_response_info":{"status":["File can't be blank"]}}

An attempt was made in [#5581][log-attempt] to catch this error
and provide more detail to help make the error actionable.
Unfortunately, the error condition coded in that PR doesn't
ever appear to happen, so we're still seeing the more obtuse
Sentry logs.

It is evident that `new_attributes` ISN'T empty, but the file
is considered 'blank' when attempting to send to Asset Manager.
This commit has an explicit check that the `attachment_data.file`
property exists before trying to update the asset.

This is a short/medium term approach which will give us a slightly
more meaningful error message, and prevent us from making the
faulty Asset Manager call in the first place. In the long term
we will still need to figure out how some assets get into this
state and prevent that from happening.

Trello: https://trello.com/c/FDu35RHS/1977-investigate-whitehall-sending-blank-file-when-updating-assets

[log-attempt]: https://github.com/alphagov/whitehall/pull/5581
[sentry-issue]: https://sentry.io/organizations/govuk/issues/1650510012/?project=202259&query=blank&statsPeriod=14d

### DIFF
--- a/app/services/asset_manager/asset_updater.rb
+++ b/app/services/asset_manager/asset_updater.rb
@@ -13,11 +13,21 @@ class AssetManager::AssetUpdater
     end
   end
 
+  class AssetFileDataEmpty < StandardError
+    def initialize(attachment_data_id, legacy_url_path)
+      super("Attempting to update '#{legacy_url_path}' for Attachment Data #{attachment_data_id} with blank file")
+    end
+  end
+
   def self.call(*args)
     new.call(*args)
   end
 
   def call(attachment_data, legacy_url_path, new_attributes = {})
+    if attachment_data.file.blank?
+      raise AssetFileDataEmpty.new(attachment_data.id, legacy_url_path)
+    end
+
     attributes = find_asset_by(legacy_url_path)
     asset_deleted = attributes["deleted"]
 
--- a/test/unit/services/asset_manager/asset_updater_test.rb
+++ b/test/unit/services/asset_manager/asset_updater_test.rb
@@ -42,6 +42,21 @@ class AssetManager::AssetUpdaterTest < ActiveSupport::TestCase
     end
   end
 
+  test "does not update asset if blank file is provided" do
+    @worker.stubs(:find_asset_by).with(@legacy_url_path)
+      .returns("id" => @asset_id, "draft" => true)
+    id = SecureRandom.random_number(1000)
+    attachment_data_without_file = FactoryBot.build(:attachment_data, id: id, file: nil)
+
+    err = assert_raises(AssetManager::AssetUpdater::AssetFileDataEmpty) do
+      @worker.call(attachment_data_without_file, @legacy_url_path, "draft" => false)
+    end
+    assert_match(
+      /Attempting to update '#{@legacy_url_path}' for Attachment Data #{id} with blank file/,
+      err.message,
+    )
+  end
+
   test "marks draft asset as published" do
     @worker.stubs(:find_asset_by).with(@legacy_url_path)
       .returns("id" => @asset_id, "draft" => true)


kevindew:
I seem to remember digging into this error a little bit with Ed and the idea we came up with was that potentially it wasn't Whitehall that was doing anything wrong here and instead it could be that Asset Manager had data that was already invalid. Since AssetManager is using the ActiveModel approach of validating the model rather than the input this can result in validation errors being returned despite the input itself being valid (which is most frustrating).

I just popped onto an Asset Manager console to try verify with the Sentry link you provided:

```
irb(main):012:0> asset = Asset.find("5b115d34ed915d2cc681ab6f"); nil
=> nil
irb(main):013:0> asset.valid?
=> false
irb(main):014:0> asset.errors.to_a
=> ["File can't be blank"]
```

So I think you might find the approach to end this error is to look into Asset Manager to see how/why these assets are invalid.

To add a bit of extra confusion:

```
irb(main):019:0> asset.file.present?
=> false
irb(main):020:0> asset.file
=> #<AssetUploader:0x00007f871aa26360 @model=#<WhitehallAsset _id: 5b115d34ed915d2cc681ab6f, created_at: "2018-06-01T14:50:28.153Z", updated_at: "2020-02-14T12:54:48.712Z", replacement_id: nil, state: "clean", filename_history: [], uuid: "b69c56f6-eacd-4ac1-a970-b062729fa2c3", draft: false, redirect_url: nil, etag: "5b115d34-3d797", last_modified: "2018-06-01T14:50:28.150Z", md5_hexdigest: "ae028b7c4f20624a39dc763a45193d79", size: 251799, access_limited: [], access_limited_organisation_ids: [], auth_bypass_ids: [], parent_document_url: "https://www.gov.uk/guidance/applications-and-decisions-for-the-west-midlands", deleted_at: nil, file: "ad-2789-1-June-2018.pdf.pdf", _type: "WhitehallAsset", legacy_url_path: "/government/uploads/system/uploads/attachment_data/file/713227/ad-2789-1-June-2018.pdf.pdf", legacy_etag: nil, legacy_last_modified: nil>, @mounted_as=:file, @storage=#<CarrierWave::Storage::File:0x00007f871aa26248 @uploader=#<AssetUploader:0x00007f871aa26360 ...>>, @file=#<CarrierWave::SanitizedFile:0x00007f871aa2f398 @file="/var/apps/asset-manager/uploads/assets/11/5d/5b115d34ed915d2cc681ab6f/ad-2789-1-June-2018.pdf.pdf", @original_filename=nil, @content_type=nil>, @versions={}>
```

ChrisBAshton:
Thanks for the pointer, @kevindew! The conclusion I've come to is this:

The reason `bad_asset.valid?` is false is because of [this line](https://github.com/alphagov/asset-manager/blob/4ac799cfdbae57d2ccdea00e11a207e2d18ca371/app/models/asset.rb#L45) (`validates :file, presence: true, unless: :uploaded?`).
Its file is neither present*, nor is its state uploaded (`bad_asset.state => "clean"`), so it fails validation.

\* I checked the S3 bucket on staging and it has not been uploaded. I also SSH'd into all Production `backend` machines to `ls /var/apps/asset-manager/uploads/assets/11/5d/5b115d34ed915d2cc681ab6f/ad-2789-1-June-2018.pdf.pdf` but it didn't exist on any of them.

Given that it has a state of clean (and not [unscanned, which is the first state assets receive](https://github.com/alphagov/asset-manager/blob/4ac799cfdbae57d2ccdea00e11a207e2d18ca371/app/models/asset.rb#L69)), it is reasonable to assume that the file was successfully uploaded to Asset Manager and was successfully scanned. It then somehow got deleted before it could be uploaded to S3.

The two ideas I'm investigating at the moment are:

1) A logic error in the code caused the file to get deleted before it was uploaded.
2) The machine the file was locally hosted on got restarted in the time after virus scanning but before deploying. The temporary file got lost in the process.

PR #5662, opened 2020-06-04
Remove WorldLocationNewsArticle

kevindew:
Trello: https://trello.com/c/oNEtjIVp/99-serve-static-assets-from-www-hostname

Back in 2017 this model was replaced by a new subtype of news article,
world news story. Since then this code has lay dormant in this repo.

My motivation for removing this was relatively modest compared to the
scale of the removal. I had to make an edit to asset handling and I
noticed that world location news article would be affected, this seemed
odd since it was replaced. Then I started pulling on the string of what
could be removed and all of this came out.

Removing this has a couple of small user noticable changes:

1) A redirect of /government/world-location-news to
   /government/announcements will go. I did some looking in CDN logs and
   the only access I can find of this redirect is from me testing
   what it does. It's been redirecting since 2017 so it's likely fine to
   let go
2) Admin users can no longer choose "World Location News Article" as a
   document type to filter by. Since there are 0 documents this seems to
   be inconsequential.


kevindew:
Thanks @thomasleese. I'll merge it in and happily take full blame for any unforseen problems 🤞

PR #5663, opened 2020-06-04
Remove Newsesque

thomasleese:
We no longer need this class following f0844e4e54a62f3226aeb5793baa57421557972d as only `NewsArticle` inherits from `Newsesque`.


thomasleese:
@kevindew Thanks for the review! I won't merge this in until your change is deployed in case there's a problem.

kevindew:
> @kevindew Thanks for the review! I won't merge this in until your change is deployed in case there's a problem.

Cool - I've deployed that to prod now, so I guess just wait and hope it doesn't catch anyone out in some unexpected way.

PR #5666, opened 2020-06-06
Policy area (also known as Topics) code cleanup

cbaines:
These are some changes that should have been made long ago, these changes form part of the great code churn relating to categorising GOV.UK content.

The Topic Taxonomy was designed to replace Policies, Policy Areas and Specialist Sectors, in fact I believe Policy Areas and Specialist Sectors were called Topics at one point or another, however, the drive to build new things, and continue to use the term Topic was too strong.

There's probably some follow up work to drop some unused tables from the database, but that should happen separately. This Pull Request deals with removing the code.


app/presenters/publishing_api/consultation_presenter.rb
```diff
@@ -40,7 +40,7 @@ def links
     def edition_links
       LinksPresenter
         .new(consultation)
-        .extract(%i[organisations parent policy_areas topics government])
+        .extract(%i[organisations parent topics government])
```
thomasleese, line 43: We might want to remove all these links from the Publishing API database as well - because they won't go away unless an explicit `policy_areas: []` is passed in when doing patch links.

cbaines: Yeah, maybe keeping some of this code around would be helpful if we were going to do that from Whitehall... I'm usually a proponent of using the Publishing APIs API, but in this case I think it may be better to do the cleanup from the Publishing API itself. Do you have any thoughts on this?

thomasleese: Yeah, I think in this case I'd be tempted to write some sort of data migration against the Publishing API database directly.

PR #5685, opened 2020-06-11
Change assets path and host

kevindew:
Trello: https://trello.com/c/oNEtjIVp/99-serve-static-assets-from-www-hostname

This changes the path that Whitehall uses for hosting assets from the unconventional "/government/assets" to a more conventional "/assets/whitehall". This then changes Whitehall to serve assets without a host by default. This is done to complete the implementation of [RFC-115](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-115-enabling-http2-on-govuk.md) where all GOV.UK frontend assets are served off the www hostname rather than the assets one (there is more backstory on https://github.com/alphagov/govuk-puppet/pull/10360).

Anyway - doing this change for other apps was [pretty simple](https://github.com/alphagov/government-frontend/pull/1767) and doing this for Whitehall has been rather thorny and difficult, hence the large number of commits (sorry), it does at least have a net negative code reduction.

Things that I've done here:

- Switch the places where we need an absolute host for assets to use Whitehall.public_root (this is done when we employ the very dubious practice of embedding links to fingerprinted assets into content items)
- Removed the proc used to determine asset host, this seems redundant since moving assets to Asset Manager
- Swapped the show production images option to use plek rather than changing asset host.

To test this I've tried in integration: running admin app and frontend app to check they serve pages and assets ok; I've tried publishing documents with fractions to check they still work; tried publishing documents without images to see placeholder image comes through. Any other scenarios to test would be useful.

One thing I was worried about was whether this would break serving images that weren't uploaded to asset manager due to a form validation error. I.e. go to Organisation that supports an image (Student Loan Company is one), change the image and remove a required field and hit submit. It turns out that showing this temporary image has been long broken before this change with us serving the path of the operating system. I've no idea how this can be fixed without re-introducing NFS so seems like something for me to put on a Platform Health backlog (sorry @thomasleese).

Once I've got this change done I plan to: 

- Add a redirect from assets.publishing.service.gov.uk/government/assets/* to www.gov.uk/assets/whitehall/* to continue supporting the assets embedded in content items
- Remove the rather strange use of [GOVUK_ASSET_ROOT for Whitehall](https://github.com/alphagov/govuk-puppet/blob/b1eeb3904ef806a9dbdf19c9385c350d363d9342/modules/govuk/manifests/apps/whitehall.pp#L356) and then remove GOVUK_ASSET_HOST from Whitehall (and hopefully whole stack) so we don't have two nearly identically named env vars.


kevindew:
Thanks @thomasleese - I'll wait until Monday to give this a deploy in case it has any surprises.

The temporary file issue looks pretty grim 😬

PR #5707, opened 2020-06-22
Add ministers index presenter

callumknights:
### Create a new presenter for the ministers index page.

This is part of the work to migrate rendering of the ministers index page from Whitehall to collections. The first 
step in this was to present a content item to the publishing-api that could be later rendered by collections.

I've added necessary tests to check that the content item contains what it should do, and that it abides by the schema
set out in govuk-content-schemas (https://github.com/alphagov/govuk-content-schemas/commit/66ebaf0e2d0454719c6f3a96f1b5ae22317a50a4#L9-L10)

https://trello.com/c/I92GWawi/2027-5-write-ministers-index-presenter

### DIFF
--- a/app/presenters/publishing_api/ministers_index_presenter.rb
+++ b/app/presenters/publishing_api/ministers_index_presenter.rb
@@ -1,0 +1,47 @@
+module PublishingApi
+  class MinistersIndexPresenter
+    attr_accessor :update_type
+
+    def initialize(update_type: nil)
+      self.update_type = update_type || "major"
+    end
+
+    def content_id
+      "324e4708-2285-40a0-b3aa-cb13af14ec5f"
+    end
+
+    def content
+      content = BaseItemPresenter.new(
+        nil,
+        title: "ministers_index",
+        update_type: update_type,
+      ).base_attributes
+
+      content.merge!(
+        base_path: base_path,
+        details: details,
+        document_type: "ministers_index",
+        rendering_app: Whitehall::RenderingApp::WHITEHALL_FRONTEND,
+        schema_name: "ministers_index",
+      )
+
+      content.merge!(PayloadBuilder::Routes.for(base_path))
+    end
+
+  private
+
+    def details
+      setting = SitewideSetting.find_by(key: :minister_reshuffle_mode)
+
+      return {} unless setting.on
+
+      { reshuffle: { message: setting.govspeak } }
+    end
+
+    def base_path
+      return "/government/ministers" if I18n.locale == :en
+
+      "/government/ministers.#{I18n.locale}"
+    end
+  end
+end
--- a/test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
@@ -1,0 +1,83 @@
+require "test_helper"
+
+class PublishingApi::MinistersIndexPresenterTest < ActionView::TestCase
+  def presented_item
+    PublishingApi::MinistersIndexPresenter.new
+  end
+
+  test "presenter is valid against ministers index schema" do
+    I18n.with_locale(:en) do
+      create(:sitewide_setting, key: :minister_reshuffle_mode, on: true)
+
+      assert_valid_against_schema(presented_item.content, "ministers_index")
+    end
+  end
+
+  test "presents ministers index page ready for the publishing-api in english" do
+    I18n.with_locale(:en) do
+      create(:sitewide_setting, key: :minister_reshuffle_mode, on: false)
+
+      expected_hash = {
+        title: "ministers_index",
+        locale: "en",
+        publishing_app: "whitehall",
+        redirects: [],
+        update_type: "major",
+        base_path: "/government/ministers",
+        details: {},
+        document_type: "ministers_index",
+        rendering_app: "whitehall-frontend",
+        schema_name: "ministers_index",
+        routes: [
+          {
+            path: "/government/ministers",
+            type: "exact",
+          },
+        ],
+      }
+
+      assert_equal expected_hash, presented_item.content
+    end
+  end
+
+  test "presents ministers index page ready for the publishing-api with correct reshuffle message" do
+    I18n.with_locale(:en) do
+      create(:sitewide_setting, key: :minister_reshuffle_mode, on: true)
+
+      expected_details = {
+        reshuffle: {
+          message: "example text",
+        },
+      }
+
+      assert_equal expected_details, presented_item.content[:details]
+    end
+  end
+
+  test "presents ministers index page ready for the publishing-api in welsh" do
+    I18n.with_locale(:cy) do
+      create(:sitewide_setting, key: :minister_reshuffle_mode, on: false)
+
+      expected_hash = {
+        title: "ministers_index",
+        locale: "cy",
+        publishing_app: "whitehall",
+        redirects: [],
+        update_type: "major",
+        base_path: "/government/ministers.cy",
+        details: {},
+        document_type: "ministers_index",
+        rendering_app: "whitehall-frontend",
+        schema_name: "ministers_index",
+        routes: [
+          {
+            path: "/government/ministers.cy",
+            type: "exact",
+          },
+        ],
+      }
+
+      assert_equal expected_hash, presented_item.content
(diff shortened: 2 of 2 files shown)


app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
...
+  class MinistersIndexPresenter
+    attr_accessor :item
+    attr_accessor :update_type
+
+    def initialize(item: nil, update_type: nil)
+      self.item = nil
+      self.update_type = update_type || "major"
+    end
+
+    delegate :content_id, to: :item
```
thomasleese, line 11: Since `item` is `nil`, it won't have a `content_id` method. Instead we'll want to [hard code the `content_id` for this page to `324e4708-2285-40a0-b3aa-cb13af14ec5f`](https://github.com/alphagov/whitehall/blob/52aff8f61a29b3999054b5b5c94875a5534eaf9a/lib/publish_static_pages.rb#L59).

```suggestion
    def content_id
      "324e4708-2285-40a0-b3aa-cb13af14ec5f"
    end
```

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
+module PublishingApi
+  class MinistersIndexPresenter
+    attr_accessor :item
+    attr_accessor :update_type
+
+    def initialize(item: nil, update_type: nil)
```
thomasleese, line 6: I don't think we even need to accept an `item` parameter because it will never be set to anything meaningful. Then we can remove the `item` `attr_accessor`.

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
...
+      self.item = nil
+      self.update_type = update_type || "major"
+    end
+
+    delegate :content_id, to: :item
+
+    def content
+      content = BaseItemPresenter.new(
+        item,
```
thomasleese, line 15: Once `item` has gone, we can hard code this to `nil`.

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
...
+        title: 'ministers_index',
+        update_type: update_type,
+      ).base_attributes
+
+      content.merge!(
+        base_path: base_path,
+        details: details,
+        document_type: 'ministers_index',
```
thomasleese, line 23: Best to be consistent with `'` or `"`. I believe the linter will want `"`.

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
...
+            :message => setting.on ? setting.govspeak : nil
+          }.compact
+        } 
+      end
+      details
+    end
+
+    def base_path
+      "/government/ministers"
+    end
```
thomasleese, line 58:
```suggestion
    def base_path
      return "/government/ministers" if I18n.locale == :en
      "/government/ministers.#{I18n.locale}"
    end
```

This ensures that we get a `/government/ministers` path with the English locale, and for every other locale the two letter code is appended to the end.

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      if I18n.locale.to_s == "cy"
+        content.merge!(
+          PayloadBuilder::Routes.for(base_path, additional_routes: ["cy"])
+        )
+      else
+        content.merge!(
+          PayloadBuilder::Routes.for(base_path)
+        )
+      end
```
thomasleese, line 36: If we follow the suggestions in the `base_path` method, this code can be simplified quite a bit:

```suggestion
      content.merge!(PayloadBuilder::Routes.for(base_path))
```

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -0,0 +1,61 @@
...
+      setting = SitewideSetting.find_by(key: :minister_reshuffle_mode)
+
+      if setting 
+        details = {
+          :reshuffle => { 
+            :message => setting.on ? setting.govspeak : nil
+          }.compact
+        } 
+      end
+      details
```
thomasleese, line 53:
```suggestion
      return {} unless setting.on

      { reshuffle: { message: setting.govspeak } }
```

This simplifies the code quite a bit and I think makes it easier to read.

cbaines: Personally, I think there is some value in having some indication in the Content Item that there's something going on with the reshuffle mode when it's not on.

I know there's the JSON schema, but you might not think to look there if you look at the content item, and don't see anything related to the reshuffle (because it's not toggled on currently).

test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
```diff
@@ -0,0 +1,102 @@
...
+  end  
+  
+  test "presenter is valid against ministers index schema" do
+    create(:sitewide_setting, key: :minister_reshuffle_mode, on: true)
+
+    presented_item = present()
+
+    print presented_item.content
+    assert_valid_against_schema(presented_item.content, "ministers_index")
+  end
```
AlanGabbianelli, line 15: Since the `present` method returns an instance instead of performing an action, it'd be better to call it with a noun rather than a verb. That's more accurate and clearer to the reader.
You could call it `presented_item` and avoid creating the variable.

test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
```diff
@@ -0,0 +1,102 @@
...
+      ]
+    }
+
+    assert_equal expected_hash, presented_item.content
+  end
+
+  test "presents ministers index page ready for the publishing-api in english with reshuffle mode" do
+    create(:sitewide_setting, key: :minister_reshuffle_mode, on: true)
+    I18n.locale = :en
+    presented_item = present()
```
AlanGabbianelli, line 47:
```suggestion
    presented_item = present
```
https://rubystyle.guide/#method-invocation-parens-no-args

test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
```diff
@@ -0,0 +1,95 @@
...
+        },
+      ],
+    }
+
+    assert_equal expected_hash, presented_item.content
+  end
+
+  test "presents ministers index page ready for the publishing-api in welsh" do
+    create(:sitewide_setting, key: :minister_reshuffle_mode, on: false)
+    I18n.locale = :cy
```
cbaines, line 72: I wonder if this could affect other tests... it might be less risky to use `with_locale` rather than setting it.

test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
```diff
@@ -0,0 +1,95 @@
...
+      document_type: "ministers_index",
+      rendering_app: "whitehall-frontend",
+      schema_name: "ministers_index",
+      routes: [
+        {
+          path: "/government/ministers",
+          type: "exact",
+        },
+      ],
+    }
```
cbaines, line 65: I think you could make this test simpler and less brittle by just testing the bit of the hash relevant to the test, so just the reshuffle bit.

test/unit/presenters/publishing_api/ministers_index_presenter_test.rb
callumknights: I'll try that, as I did have an issue with this in that if the locale is set this way it does persist across tests, hence each test it's set individually. I'll give your suggestion a go 👍

app/presenters/publishing_api/ministers_index_presenter.rb
thomasleese: I don't mind either way, I do like the idea of the `reshuffle` field always being present. We should probably [change the schema to make `message` the optional part](https://github.com/alphagov/govuk-content-schemas/blob/327a3ae972381106a401825ab79bece8184fc1c5/formats/ministers_index.jsonnet#L7-L12) if we decide to do this.

PR #5709, opened 2020-06-22
Upload csv exports to s3 and allow requesting them

benjamineskola:
This uses a similar approach to that in specialist-publisher (alphagov/specialist-publisher#1576 and alphagov/specialist-publisher#1625).

https://trello.com/c/5xKHcam8/2012-5-whitehall-replace-attachments-in-documentlistexport-so-that-notify-can-be-used


Gemfile
```diff
@@ -10,6 +10,7 @@ gem "carrierwave-i18n"
 gem "chronic"
 gem "dalli", "~> 2.7"
 gem "faraday"
+gem "fog-aws", "~> 3.5"
```
ChrisBAshton, line 13: It would be good to link to the specialist publisher PRs in the commit message itself.

app/workers/document_list_export_worker.rb
```diff
@@ -6,14 +6,28 @@ def perform(filter_options, user_id)
...
+
+  def upload_file(csv, type)
+    export_id = SecureRandom.uuid
+    document_type_slug = if type.present?
+                           type.downcase.gsub(%r{[^a-z0-9]+}, "_")
+                         else
+                           "documents"
+                         end
+
+    filename = "document_list_#{document_type_slug}_#{export_id}.csv"
```
ChrisBAshton, line 28: It seems a shame / risky to duplicate this logic here and in the `show` method of the controller - could it be pulled out?

test/unit/workers/document_list_export_worker_test.rb
```diff
@@ -4,6 +4,25 @@ class DocumentListExportWorkerTest < ActiveSupport::TestCase
...
+    ENV["AWS_S3_BUCKET_NAME"] = "test-bucket"
+
+    # Create an S3 bucket so the code being tested can find it
+    connection = Fog::Storage.new(
+      provider: "AWS",
+      region: ENV["AWS_REGION"],
+      aws_access_key_id: ENV["AWS_ACCESS_KEY_ID"],
+      aws_secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"],
+    )
+    @directory = connection.directories.get(ENV["AWS_S3_BUCKET_NAME"]) || connection.directories.create(key: ENV["AWS_S3_BUCKET_NAME"]) # rubocop:disable Rails/SaveBang
```
ChrisBAshton, line 21: Is it worth extracting this setup stuff into a helper?

EDIT - I see this is also copied in the third commit, so three times total? Definitely worth moving into a helper IMO 👍

test/unit/workers/document_list_export_worker_test.rb
```diff
@@ -22,12 +41,41 @@ class DocumentListExportWorkerTest < ActiveSupport::TestCase
...
+    assert_match %r{^document_list_documents_[a-f0-9-]{36}\.csv$}, file.key
+    assert_equal "this|is|my|csv", file.body
+  end
+
+  test "email body includes url" do
+    csv = "this|is|my|csv"
+    title = "Everyone's editions"
+    @worker.stubs(:create_filter).returns(stub(page_title: title))
+    @worker.stubs(generate_csv: csv)
+    # @worker.stubs(:send_mail)
```
ChrisBAshton: Stray comment here

test/unit/workers/document_list_export_worker_test.rb
```diff
@@ -22,12 +41,41 @@ class DocumentListExportWorkerTest < ActiveSupport::TestCase
...
+    assert_match %r{\bhttps://whitehall\.test\.gov\.uk/export/documents/[a-f0-9-]{36}\b}, message.body.raw_source
+  end
+
+  test "sends mail with CSV link to user" do
     csv = "this|is|my|csv"
     title = "Everyone's editions"
     @worker.stubs(:create_filter).returns(stub(page_title: title))
     @worker.stubs(generate_csv: csv)
-    Notifications.expects(:document_list).with(csv, @user.email, title).returns(stub(deliver_now: nil))
+    Notifications.expects(:document_list).with(regexp_matches(%r{^https://whitehall\.test\.gov\.uk/export/documents/[a-f0-9-]{36}$}), @user.email, title).returns(stub(deliver_now: nil))
```
ChrisBAshton, line 65: I'd be tempted to extract this regex into a variable/method so we avoid repeating it twice.

test/unit/workers/document_list_export_worker_test.rb
```diff
@@ -57,19 +57,6 @@ class DocumentListExportWorkerTest < ActiveSupport::TestCase
     assert_equal "this|is|my|csv", file.body
   end
 
-  test "email body includes url" do
```
ChrisBAshton: The "Fixes tests" commit should be fixed up to the first commit.

test/functional/document_list_export_request_controller_test.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  test "returns an error if there is no request" do
+    login_as :gds_editor
+
+    get :show, params: { document_type_slug: "asylum-support-decisions", export_id: "aaaa-bbbb" }
+    assert_equal 404, response.status
+  end
+
+  test "returns an error if not logged in" do
```
ChrisBAshton, line 42: I'd be more specific, e.g.

```suggestion
  test "returns a 401 Unauthorized error if not logged in" do
```

...saves me having to look it up 😄  Same for the test above it.

test/unit/workers/document_list_export_worker_test.rb
benjamineskola: Yep — wanted to keep it around temporarily because I wasn't sure about removing one of the tests, but I won't merge as-is.

app/workers/document_list_export_worker.rb
benjamineskola: I'm not sure where the best place to move it to would be: neither the controller nor the worker seem right, and there isn't a model for this — or would creating one simply for that make sense?

ChrisBAshton: Perhaps in `DocumentListExportPresenter`?

test/unit/workers/document_list_export_worker_test.rb
benjamineskola: One of the instances has been removed but it's still used in a couple of different places — it could go in `test_helper.rb` but it feels a little bit too specific for that.

lib/s3_file_uploader.rb
```diff
@@ -0,0 +1,18 @@
...
+    connection = Fog::Storage.new(
+      provider: "AWS",
+      region: ENV["AWS_REGION"],
+      aws_access_key_id: ENV["AWS_ACCESS_KEY_ID"],
+      aws_secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"],
+    )
+
+    directory = connection.directories.get(ENV["AWS_S3_BUCKET_NAME"])
+
+    directory.files.create( # rubocop:disable Rails/SaveBang
```
ChrisBAshton, line 12: Can you explain in the commit why this is disabled? I can't see this being used in the two PRs you linked to.
And would it be possible to improve the code to fix the linting error rather than ignore it?

config/routes.rb
```diff
@@ -429,6 +429,8 @@ def external_redirect(path_prefix, target)
 
   get "/government/uploads/system/uploads/attachment_data/file/:id/*file.:extension/preview" => "csv_preview#show", as: :csv_preview
 
+  resources :document_list_export_request, path: "/export/:document_type_slug", param: :export_id, only: [:show]
```
ChrisBAshton, line 432: This adds a new route, but I wouldn't have guessed a new route was being added by the commit message contents. Could you spend a bit of time explaining what's being added and why?

test/test_helper.rb
```diff
@@ -178,6 +178,23 @@ def force_publish(edition)
...
+    ENV["AWS_S3_BUCKET_NAME"] = "test-bucket"
+
+    # Create an S3 bucket so the code being tested can find it
+    connection = Fog::Storage.new(
+      provider: "AWS",
+      region: ENV["AWS_REGION"],
+      aws_access_key_id: ENV["AWS_ACCESS_KEY_ID"],
+      aws_secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"],
+    )
+    @directory = connection.directories.get(ENV["AWS_S3_BUCKET_NAME"]) || connection.directories.create(key: ENV["AWS_S3_BUCKET_NAME"]) # rubocop:disable Rails/SaveBang
```
ChrisBAshton, line 196: Could this be simplified to this?

```ruby
    @directory ||= connection.directories.create(key: ENV["AWS_S3_BUCKET_NAME"])
```

lib/s3_file_uploader.rb
benjamineskola: The problem is that `rubocop-rails` is full of false positives: it assumes that a `create` method is always on an ActiveRecord object and has an associated `create!` method, and that the latter should be used if the return value isn't needed. The problem is that it has no way of telling when the object in question is actually an ActiveRecord model.

This particular one seems to get disabled a lot in whitehall:

```
$ rg 'rubocop:disable Rails/SaveBang' | wc -l
      48
$ rg 'rubocop:disable Rails' | wc -l
      58
```

I'm not completely sure why it didn't matter in specialist-publisher, for example, but it's possible this check wasn't enabled at the time.

test/test_helper.rb
benjamineskola: Looks like this might be some over-cautious coding, since in production the bucked will always exist and in tests it will always have to be created. `||=` won't work because it will be `nil` initially either way, but the `get` can be dropped, I think.

lib/s3_file_uploader.rb
benjamineskola: I've just double-checked and the equivalent lines in specialist-publisher now also have these warnings.

ChrisBAshton: Ah ok, thanks for the explanation. I wonder if that's something worth fixing upstream in https://github.com/alphagov/rubocop-govuk? Out of scope for this PR though 👍

config/routes.rb
benjamineskola: I've added an explanation in the relevant commit message.

test/test_helper.rb
```diff
@@ -193,7 +193,7 @@ def setup_fog_mock
       aws_access_key_id: ENV["AWS_ACCESS_KEY_ID"],
       aws_secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"],
     )
-    @directory = connection.directories.get(ENV["AWS_S3_BUCKET_NAME"]) || connection.directories.create(key: ENV["AWS_S3_BUCKET_NAME"]) # rubocop:disable Rails/SaveBang
+    @directory = connection.directories.create(key: ENV["AWS_S3_BUCKET_NAME"]) # rubocop:disable Rails/SaveBang
```
ChrisBAshton, line 196: This last commit could be squashed into the penultimate commit (there seems little value in separating it out).

config/routes.rb
ChrisBAshton: I'm still not seeing an explanation in https://github.com/alphagov/whitehall/pull/5709/commits/dfe36a60b5c355e7a64005a192ab3f5300d1df5f - did you add it to the right commit?

benjamineskola: It's in d9af9e765afe068edca914566feed0b47711b56b — that's where the controller itself is added. Actually the route should have been added there too, separating it out then committing it at the same time as the tests wasn't really what I had in mind.

ChrisBAshton: From what you're saying, it sounds like this line should be moved to the first commit, but that it might be more work than it's worth. I'll leave it up to you to decide.

app/controllers/document_list_export_request_controller.rb
```diff
@@ -15,12 +15,17 @@ def show
   end
 
   def get_csv_file_from_s3(filename)
-    connection = Fog::Storage.new(
-      provider: "AWS",
+    params = if ENV.key? "AWS_ACCESS_KEY_ID"
+               { aws_access_key_id: ENV["AWS_ACCESS_KEY_ID"],
+                 aws_secret_access_key: ENV["AWS_SECRET_ACCESS_KEY"] }
+             else
+               { use_instance_profile: true }
```
ChrisBAshton, line 22: Where did this field come from?

Via [this issue](https://github.com/fog/fog-aws/issues/423), it looks like what you want is [use_iam_profile](https://github.com/fog/fog-aws/search?q=use_iam_profile&unscoped_q=use_iam_profile) (I can't see [use_instance_profile](https://github.com/fog/fog-aws/search?q=use_instance_profile&unscoped_q=use_instance_profile) anywhere)

benjamineskola: I found it on stackoverflow but upon doublechecking it seems like it was `use_iam_profile` there too, oops.

The conditional still seems to be necessary though, which is annoying.

lib/s3_file_handler.rb
```diff
@@ -1,5 +1,25 @@
-class S3FileUploader
+class S3FileHandler
+  def self.get_csv_file_from_s3(filename)
```
ChrisBAshton, line 2: This could be renamed to `get_file_from_s3` (there's nothing CSV specific in its implementation).

test/lib/s3_file_handler_test.rb
```diff
@@ -10,7 +10,7 @@ class S3FileUploaderTest < ActiveSupport::TestCase
   end
 
   test "has the expected filename and content" do
-    s3_file = S3FileUploader.save_file_to_s3("test_file_name.txt", "hello, world\n")
+    s3_file = S3FileHandler.save_file_to_s3("test_file_name.txt", "hello, world\n")
```
ChrisBAshton, line 13: This consolidation of code is a nice idea 👍  shame there aren't any tests to do with getting the file from s3 - could you write some?

test/lib/s3_file_handler_test.rb
```diff
@@ -9,11 +9,17 @@ class S3FileHandlerTest < ActiveSupport::TestCase
...
     assert_equal "test_file_name.txt", s3_file.key
     file = @directory.files.get("test_file_name.txt")
     assert_not_nil file
     assert_equal "hello, world\n", file.body
   end
+
+  test "download fetches the right content" do
+    S3FileHandler.save_file_to_s3("test_file_name.txt", "hello, world\n")
+    file = S3FileHandler.get_file_from_s3("test_file_name.txt")
+    assert_equal "hello, world\n", file
```
ChrisBAshton, line 23: Nice 🎉

PR #5727, opened 2020-07-08
Fail validation if URL contains whitespace

1pretz1:
A user managed to save a URL with whitespace which lead to
publishing-api rejecting the unpublishing request but whitehall saving
the consolidated URL leaving the edition is a funky state. This PR
prevents this from happening in the future.

This also switches over to using ruby's URI instead of Addressable. This
seems to of been added originally due to limitations in our URL
validation however from my testing using plain old ruby's URI is fine.
[Original PR](5b31351)

Now the class validating the URL matches what is present in
[publishing-api](https://github.com/alphagov/publishing-api/blob/93605514d7c32ddcb93bf793529de00b64cc1c39/app/validators/routes_and_redirects_validator.rb#L171) so we should no longer get a disparity in validating URL's between the
two apps.

This validation is used [here](https://github.com/alphagov/whitehall/blob/master/app/models/unpublishing.rb#L9) on the unpublishing model.

---

Zendesk:
https://govuk.zendesk.com/agent/tickets/4144535

### DIFF
--- a/app/validators/uri_validator.rb
+++ b/app/validators/uri_validator.rb
@@ -4,12 +4,16 @@ require "addressable/uri"
 # Accepts options[:message] and options[:allowed_protocols]
 class UriValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    if !(uri = Addressable::URI.parse(value))
+    return if value.nil?
+
+    uri = URI.parse(value)
+
+    if uri.blank?
       record.errors[attribute] << failure_message
-    elsif !allowed_protocols.include?(uri.scheme)
+    elsif allowed_protocols.exclude?(uri.scheme)
       record.errors[attribute] << "is not valid. Make sure it starts with http(s)"
     end
-  rescue Addressable::URI::InvalidURIError
+  rescue URI::Error
     record.errors[attribute] << failure_message
   end
 
--- a/test/unit/validators/uri_validator_test.rb
+++ b/test/unit/validators/uri_validator_test.rb
@@ -5,6 +5,11 @@ class UriValidatorTest < ActiveSupport::TestCase
     @validator = UriValidator.new(attributes: [:url])
   end
 
+  test "validates nil urls" do
+    feature_link = validate(PromotionalFeatureLink.new(url: nil))
+    assert feature_link.errors.empty?
+  end
+
   test "validates http urls" do
     feature_link = validate(PromotionalFeatureLink.new(url: "http://example.com"))
     assert feature_link.errors.empty?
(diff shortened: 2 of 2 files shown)


benthorner:
Nice one finding this bug! The code looks fine to me.

We need to be careful with the rollout of this change: this could cause a blocking issue for any code/tasks that does a mass update on records where the URL is invalid. Example of this issue: https://github.com/alphagov/whitehall/pull/5487.

Before this goes in, can we run all the data through the modified validator, for models that use it?

kevindew:
If this is failing validation for spaces do we know what other URI validation is missing? I'm a bit concerned that we've just added a case for one of the many ways a URI might not be valid when I guess the problem is that [Addressable](https://github.com/alphagov/whitehall/blob/163d65828f191a421406cdf5202df818a36e8d3c/app/validators/uri_validator.rb#L7) and JSON schema disagree on what a valid URI is (it looks like Addressable is rather permissive) and may mean having to add more edge cases later. 

For example Ruby's URI class will reject things Addressable does not:

```
irb(main):025:0> Addressable::URI.parse("http://abc/optio😬n%3E")
=> #<Addressable::URI:0x3fb8d37228e4 URI:http://abc/optio😬n%3E>
irb(main):026:0> URI.parse("http://abc/optio😬n%3E")
Traceback (most recent call last):
        1: from (irb):26
URI::InvalidURIError (URI must be ascii only "http://abc/optio\u{1F62C}n%3E")
irb(main):027:0> Addressable::URI.parse(" ")
=> #<Addressable::URI:0x3fb8d372edc4 URI: >
irb(main):028:0> URI.parse(" "))
Traceback (most recent call last):
        1: from (irb):28
URI::InvalidURIError (bad URI(is not URI?): " ")
```

Addressable seems to be rather clever so could be they have different modes or complexities, and maybe there was a particular motivation for using it in the first place. However you may just find using Rubys' URI provides a more compatible parser to the one used in Publishing API.

1pretz1:
I've taken your recommendation @kevindew and mimicked the validation that occurs in publishing-api by using ruby's own MRI class so there shouldn't be a disparity between these two apps now.

The only difference between them that would potentially cause issues is the handling of nils which occurs frequently in whitehall's db which I've taken care of and I've added an extra rescue class as certain formats of URL's will cause an different URI error. E.g `irb(main):046:0> URI.parse('mailto://hello.com')` will raise a `URI::InvalidComponentError (missing opaque part for mailto URL)`. Maybe we should update publishing-api to include this rescue too.

As per your request @benthorner - I've looked through all the class that use the validator and the logic should be fine with all existing records apart from 1 which has been deleted anyway! 😄

app/validators/uri_validator.rb
```diff
@@ -4,12 +4,16 @@
...
+    uri = URI.parse(value)
+
+    if uri.blank?
       record.errors[attribute] << failure_message
-    elsif !allowed_protocols.include?(uri.scheme)
+    elsif allowed_protocols.exclude?(uri.scheme)
       record.errors[attribute] << "is not valid. Make sure it starts with http(s)"
     end
-  rescue Addressable::URI::InvalidURIError
+  rescue URI::InvalidComponentError, URI::InvalidURIError
```
kevindew, line 16: Rather than catching two descendent exceptions you could catch just their superclass of `URI::Error`.

PR #5730, opened 2020-07-09
Sync enforce style for Naming/VariableNumber

benthorner:
https://github.com/alphagov/rubocop-govuk/issues/73

There is no reason for this to be different to RuboCop GOV.UK.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,9 +7,6 @@ inherit_mode:
   merge:
     - Exclude
 
-Naming/VariableNumber:
-  EnforcedStyle: snake_case
-
 # This has special infrastructure for one-time data
 # migration tasks, for which it's common to write to
 # STDOUT using 'puts'.
--- a/features/step_definitions/consultation_steps.rb
+++ b/features/step_definitions/consultation_steps.rb
@@ -21,9 +21,9 @@ When(/^I draft a new consultation "([^"]*)"$/) do |title|
   click_button "Save"
 end
 
-Then(/^I can see links to the consultations "([^"]*)" and "([^"]*)"$/) do |title_1, title_2|
-  assert has_css?(".consultation a", text: title_1)
-  assert has_css?(".consultation a", text: title_2)
+Then(/^I can see links to the consultations "([^"]*)" and "([^"]*)"$/) do |title1, title2|
+  assert has_css?(".consultation a", text: title1)
+  assert has_css?(".consultation a", text: title2)
 end
 
 When(/^I add an outcome to the consultation$/) do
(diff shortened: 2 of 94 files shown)


cbaines:
For this change especially, I'm doubtful if the code churn is really worth it, especially given I think the changes here make code more difficult to read, even less correct in places.

Things like:
```
log_error_and_render_500 -> log_error_and_render500
closed_at_or_within_24_hours_of -> closed_at_or_within24_hours_of
windows_1252_encoding -> windows1252_encoding
https://public.govdelivery.com/accounts/UKMHRA/subscriber/new?topic_id=UKMHRA_0044 -> https://public.govdelivery.com/accounts/UKMHRA/subscriber/new?topic_id=UKMHRA0044
...
```

I realise plenty of changes have already been made that could be worse along these same lines, and the Git history is already pretty unreadable because of other sweeping linting changes, but... for some reason I decided to comment on this change.

benthorner:
This is now ready for review. 

@cbaines all the changes in `app/` and `lib/ were accidental. I've removed them now. Given that we're now very close to having all repos in sync with our styleguide, I think it's worth pursuing this still.

cbaines:
Ok, I think that's not changed the things I mentioned above, one gem I spotted though was `edition_linked_to_2 -> edition_linked_to2`.

I guess I still prefer the readability that the consistent use of underscores in variable names offers.

benthorner:
@cbaines just wondering, are you intending to review this PR? It follows our styleguide, and given this is one of the last apps to bring in-sync, I'd suggest raising any general issues with the styleguide [over there](https://github.com/alphagov/rubocop-govuk).

What I'm looking for in a review of this PR:

- Any mistakes I've made that have slipped past the tests (unlikely, since all of the changes _are_ the tests)
- Any places where we need to deviate from the styleguide

cbaines:
> @cbaines just wondering, are you intending to review this PR? It follows our styleguide, and given this is one of the last apps to bring in-sync, I'd suggest raising any general issues with the styleguide [over there](https://github.com/alphagov/rubocop-govuk).

I'm not aware that any of the styleguides for GOV.UK code have gone in to this level of detail, specifying rules around variable names containing numbers. Admittedly, I didn't follow the transition to the GDS Way, and I wasn't really aware that the location of the styleguide had moved again. Anyway, I've checked the rubocop-govuk repository, and I can't see it saying anything about this either.

> What I'm looking for in a review of this PR:
> 
>     * Any mistakes I've made that have slipped past the tests (unlikely, since all of the changes _are_ the tests)
> 
>     * Any places where we need to deviate from the styleguide

What I've been thinking about when reviewing these changes is just whether it makes the code objectively better, or worse. I'm not that fussed if you find someone else to approve these changes, but I at least wanted to put forward the point that these changes are making the code worse in my opinion.

benthorner:
@cbaines in moving to govuk-lint, and subsequently rubocop-govuk, we have adopted RuboCop's defaults as part of our styleguide. This makes sense, because over time we continue to follow the community styleguide, which RuboCop is based upon. We also appear to have explicitly embraced this Cop [[1](https://github.com/alphagov/rubocop-govuk/commit/3e8f7164f536fcb5b59182147b89242f2663b597)].

We shouldn't debate the applicability of the styleguide in individual repos. This is the main benefit of having a styleguide: to minimise and centralise debate, so we can be consistent in our objectivity. If this Cop does objectively make things worse, then I'm willing to review a PR to reconfigure it for all our repos.

For now, I'll ask someone else to review.

PR #5735, opened 2020-07-15
Deindex publications if the type has changed

thomasleese:
In some of these cases the `base_path` will have changed and if we don't first remove the previous edition, two entries will appear in the search index. This is because the search index uses the `base_path` as the unique index for each edition.

This branch is currently deployed to integration for testing.

[Trello Card](https://trello.com/c/r4tnFu0g/2038-investigate-search-results-showing-republished-document-twice-timebox-2-days)

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -105,6 +105,10 @@ class Publication < Publicationesque
     self.publication_type_id
   end
 
+  def has_changed_publication_type?
+    previous_edition && previous_edition.publication_type != publication_type
+  end
+
   def national_statistic?
     publication_type == PublicationType::NationalStatistics
   end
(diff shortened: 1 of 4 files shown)


thomasleese:
> LGTM 👍 My only thought is could/should this apply to non-publications?

Yeah, I had that thought too. I also thought about whether we could make it generic enough that it just compares the `base_path`. Unfortunately doing this makes the code a bit ugly because it couples the code to the presenters for the search index, which ends up making it a bit circular (we generate the indexable content to decide whether to generate the indexable content and send it to the Search API). I did a brief check and as far as I can tell only publications have the possibility of changing the `base_path`.

I also think that really this should be something that is fixed in the Search API. It seems a bit problematic to rely on the `base_path` to uniquely represent a document rather than using the `content_id`. I imagine we're going to hit a problem as we continue to migrate things so that they're picked off the Publishing API queue because at that point the Search API isn't going to have any knowledge of the previous edition... so this PR doesn't really fix the underlying problem, but fixing the Search API would be a very substantial change.

ChrisBAshton:
Thanks for the explanation 😄

PR #5740, opened 2020-07-20
Enable Rails/HelperInstanceVariable Cop

benthorner:
Please see the commits for more details.


app/helpers/admin/sidebar_helper.rb
```diff
@@ -14,16 +14,16 @@ def simple_formatting_sidebar(options = {})
...
     options = { editing: false, history_count: 0, remarks_count: 0 }.merge(options)
     {}.tap do |tabs|
       if options[:editing]
         tabs[:govspeak_help] = "Help"
       end
       tabs[:notes] = ["Notes", options[:remarks_count]]
       tabs[:history] = ["History", options[:history_count]]
-      if @edition.can_be_fact_checked?
-        tabs[:fact_checking] = ["Fact checking", @edition.all_completed_fact_check_requests.count]
+      if edition.can_be_fact_checked?
```
ChrisBAshton, line 25: Curious! Looks like this snuck in 8 years ago: https://github.com/alphagov/whitehall/commit/186e61b1a35f83ece10b5d91bb977b1544d4dd23#diff-5ede572e8c3d99712827002de6d600eeR10-R17. Interesting how we haven't been referencing the argument `edition` that entire time 😧

app/views/admin/document_collection_groups/index.html.erb
```diff
@@ -79,10 +79,10 @@
                 <li class="remove">
                   <%= submit_tag 'Remove', class: 'btn btn-sm btn-default' %>
                 </li>
-                <% if other_groups(group).size > 0 %>
+                <% if (@groups - [group]).size > 0 %>
```
ChrisBAshton, line 82: I'd be tempted to precede this with a `other_groups = @groups - [group]` to make the code slightly more self documenting and dry. Does the job of killing off a redundant module while keeping the benefits it was trying to encapsulate.

app/views/home/how_government_works.html.erb
```diff
@@ -1,6 +1,5 @@
 <% page_title "How government works" %>
 <% page_class "inside_gov_how-gov-works" %>
-<% meta_description "In the UK, the Prime Minister leads the government with the support of the Cabinet and ministers. You can find out who runs government and how government is run, as well as learning about the history of government." %>
```
ChrisBAshton, line 3: Hmm, the commit message is a little misleading, as it does appear that the meta description is [making it through to the rendered page](https://www.gov.uk/government/how-government-works) (via [application_helper](https://github.com/alphagov/whitehall/blob/master/app/helpers/application_helper.rb#L12-L14) and [frontend base](https://github.com/alphagov/whitehall/blob/cf70ac9f379ca638cb735b83e0ea02f585ee5d4b/app/views/layouts/_frontend_base.html.erb#L11)):

```html
<meta name="description" content="In the UK, the Prime Minister leads the government with the support of the Cabinet and ministers. You can find out who runs government and how government is run, as well as learning about the history of government.">
```

However, it makes sense to use a consistent approach throughout Whitehall.

app/views/admin/document_collection_groups/index.html.erb
benthorner: Sure, I've added a view variable here.

app/views/home/how_government_works.html.erb
benthorner: Oops, I've amended the commit. (I actually thought those pages weren't being used anymore, as well! Thanks for correcting me there - wouldn't want to delete how government works.)

PR #5744, opened 2020-07-21
Use individual component assets

andysellick:
Update whitehall to import the sass and Javascript for only the components it is using, instead of all components, as [documented here](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#import-sass-for-individual-components) and [here](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#import-javascript-for-individual-components).

See [this sheet](https://docs.google.com/spreadsheets/d/1no3HnEayRn2YDsaS0jCmZzTfllDzUt_yliXhE-wNMoE/edit#gid=0) for some example pages to test.

## Filesize comparison

CSS before: **1.6 MB** 

whitehall-css-before-government-how-government-works

CSS after: **1.1 MB** 

whitehall-css-after

JS before: **1.5 MB** 

whitehall-js-before-government-how-government-works

JS after: **1.1 MB** 

Screenshot 2020-07-21 at 16 49 06

### HTML publications

`html-publication.css` is used on [pages like this](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/214962/passport-impact-indicat.csv/preview), which I can't get up and running locally to test, so here's some 'before' numbers for checking later.

CSS before:

Screenshot 2020-07-21 at 16 17 59

JS before:

Screenshot 2020-07-21 at 16 18 10

Trello card: https://trello.com/c/GWttlBxs/228-upgrade-applications-to-use-individual-component-sass


app/assets/javascripts/application.js
```diff
@@ -4,6 +4,11 @@
...
 // the compiled file.
 //
+//= require govuk_publishing_components/lib
+//= require govuk_publishing_components/components/button
+//= require govuk_publishing_components/components/feedback
+//= require govuk_publishing_components/components/govspeak
+//
 //= require jquery_ujs
 //= require jquery.ui.all
 //= require vendor/jquery/magna-charta.min
```
injms, line 14: I _think_ that the `govspeak` component already includes the Magna Charta library - let me check.

alex-ju: [magna-charta](https://github.com/alphagov/govuk_publishing_components/blob/9531ad85ded5c74fa47b8d32c54a587edb39e44c/app/assets/javascripts/govuk_publishing_components/lib/govspeak/barchart-enhancement.js) is already in `govuk_publishing_components/lib` indeed

andysellick: Great spot, thanks, have removed!

andysellick: Oooh that brought the JS size down a little more, excellent. Have updated the PR description.

alex-ju:
> I haven't been able to verify that the chart part of govspeak is working correctly. I'm trying to find a real world example - but when I add a table that should be converted into a chart into a page template it doesn't render correctly.

It seemed fine in my tests. Can you share the govspeak code you've tried with? maybe I'm missing something.
Screenshot 2020-07-24 at 17 40 19

andysellick:
Thanks for the thorough testing @injms @alex-ju appreciate you taking the time. Last one done!

PR #5748, opened 2020-07-22
Switch from SES to Notify in ActionMailer

benjamineskola:
Rebasing #5627 on top of #5709

https://trello.com/c/1dROdOVn/1972-8-whitehall-use-notify-instead-of-amazon-ses-%F0%9F%8D%90


benjamineskola:
Tested successfully in integration.

PR #5762, opened 2020-08-11
Update the Brexit no-deal notice for transition

sihugh:
This content is out of date and does not match the current intent of the notice box.

Guidance for publishers will be communicated on Basecamp.

https://trello.com/c/ULzLyIBv/366-update-language-used-for-brexit-no-deal-content-notice-in-whitehall-publisher


sihugh:
Not to merge until @Weeza gives the nod

PR #5763, opened 2020-08-14
[DO NOT MERGE] WIP Edit excluded nation form

edwardkerry:
This WIP PR adds an additional checkbox, `all_nation_applicability`, to the Excluded Nations fieldset in `/admin/publications/new`.

It adds validation to ensure that either this new checkbox or one of the existing Excluded Nations checkboxes is selected, and also that they are not both selected.

If `all_nation_applicability` is selected any previous `nation_inapplicabilities` will be discarded, similarly to submitting the form with all Excluded Nations empty.


edwardkerry:
superceded by https://github.com/alphagov/whitehall/pull/5766

PR #5764, opened 2020-08-14
Add base tag to html attachments view

karlbaker02:
This PR adds a `base` tag to the `html_attachments` view. This view is used by the `csv_preview` controller and links currently rendered in the footer on this page currently point to the wrong URL (assets rather than GOV.UK); as a result, they return 404s. The footer itself is obtained through Slimmer, using the `chromeless` view that's part of `Static` where most URLs in the footer are relative.

By adding a `base` tag, all relative URLs on the page will be prefixed with the `href` attribute specified, which is the GOV.UK website root. The `base` tag can only appear in the `head` of a page, which is the reason for adding it into the partial `frontend_base`.

### DIFF
--- a/app/controllers/csv_preview_controller.rb
+++ b/app/controllers/csv_preview_controller.rb
@@ -8,6 +8,7 @@ class CsvPreviewController < ApplicationController
           @edition = visible_edition
           @attachment = attachment_data.visible_attachment_for(current_user)
           @csv_preview = CsvFileFromPublicHost.csv_preview_from(@csv_response)
+          @page_base_href = Plek.new.website_root
           render layout: "html_attachments"
         else
           fail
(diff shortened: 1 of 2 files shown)


kevindew:
Super, this looks like a good approach. Nice one working this out. Have you tried it on integration? Might also be worth checking it works ok on draft stack, but I assume Plek is fine for that.

Implementation wise I'd suggest moving this from setting it on all html_publications (as the vast majority of them don't need it) and localising this to just csv_preview since that is the quirky bit of the stack. You could do that by setting `@page_base_href` in the [controller](https://github.com/alphagov/whitehall/blob/6e9838d6feff7b94595d6a59fa68bb968103f257/app/controllers/csv_preview_controller.rb) and modifying your base tag accordingly which will cut html attachments out of the picture

karlbaker02:
Makes sense, thanks @kevindew. Haven't tested on integration yet as I'm waiting on the tests to pass before deploying the branch.

I wasn't sure how wide to scope this change originally, which is why I added it into `local_assigns` and only passed that in from the `html_attachments` view. I think your recommendation is the way forward, so I'll pick that up now.

kevindew:
cool cool - probably worth giving it a spin on integration as is anyway just to verify it does fix the problem

PR #5766, opened 2020-08-18
Ensure editions have accurate UK Nation applicability

edwardkerry:
When creating or updating an edition of a `Publication`, `Consultation` or `DetailedGuide`, the user has the option to exclude one or more UK Nation if the document is not applicable. This feature is not well used or understood, leading to some editions being created without relevant excluded nations being applied.

This PR adds a new checkbox to the form, to explicitly state that the edition should apply to all UK nations. 

Validations have been added to ensure that
- either the new checkbox is selected or at least one nation is excluded
- only one of these selections is allowed

As the validation means the newly added `all_nation_applicability` parameter is mandatory for creating or updating, we have added a column to the DB schema. A data migration will check all relevant existing edition types and, if they have excluded nations present, set this `all_nation_applicability` to `false`. 

See individual commit messages for further details.

[Passing Publishing-E2E tests](https://github.com/alphagov/publishing-e2e-tests/pull/394)

[trello](https://trello.com/c/GBdunyhP/2082-3-change-excluded-nation-fieldset-in-whitehall)

### DIFF
--- a/app/models/edition/national_applicability.rb
+++ b/app/models/edition/national_applicability.rb
@@ -20,10 +20,10 @@ module Edition::NationalApplicability
 
   def applicability_options
     if !all_nation_applicability && nation_inapplicabilities.none?(&:excluded?)
-      errors.add(:nation_inapplicabilities, "must either allow all nations or exclude at least one nation")
+      errors.add(:nation_inapplicabilities, "- you must select whether this content applies to all UK nations or which ones it excludes")
     end
     if all_nation_applicability && nation_inapplicabilities.any?(&:excluded?)
-      errors.add(:nation_inapplicabilities, "cannot allow all nations and exclude nations")
+      errors.add(:nation_inapplicabilities, "- you cannot select all UK nations and also exclude nations")
     end
   end
 
--- a/app/models/edition/national_applicability.rb
+++ b/app/models/edition/national_applicability.rb
@@ -10,11 +10,21 @@ module Edition::NationalApplicability
   end
 
   included do
+    validate :applicability_options
     has_many :nation_inapplicabilities, foreign_key: :edition_id, dependent: :destroy, autosave: true
     validates_associated :nation_inapplicabilities
     validates :nation_inapplicabilities, length: { maximum: Nation.all.count - 1, message: "can not exclude all nations" }
 
     add_trait Trait
+  end
+
+  def applicability_options
+    if !all_nation_applicability && nation_inapplicabilities.none?(&:excluded?)
+      errors.add(:nation_inapplicabilities, "must either allow all nations or exclude at least one nation")
+    end
+    if all_nation_applicability && nation_inapplicabilities.any?(&:excluded?)
+      errors.add(:nation_inapplicabilities, "cannot allow all nations and exclude nations")
+    end
   end
 
   def nation_inapplicabilities_attributes=(attributes)
--- a/db/data_migration/20200817081009_add_all_nation_applicability_to_editions.rb
+++ b/db/data_migration/20200817081009_add_all_nation_applicability_to_editions.rb
@@ -1,0 +1,12 @@
+PUBLISHED_AND_PUBLISHABLE_STATES = %w[published draft archived submitted rejected scheduled].freeze
+EDITIONS_WITH_NATIONAL_APPLICABILITY = ['DetailedGuide', 'Publication', 'Consultation'].freeze
+
+edition_scope = Edition.where(type: EDITIONS_WITH_NATIONAL_APPLICABILITY).where(state: PUBLISHED_AND_PUBLISHABLE_STATES)
+
+def update_all_nation_applicability(edition)
+  if edition.nation_inapplicabilities.any?
+    edition.update_column(:all_nation_applicability, false)
+  end
+end
+
+edition_scope.find_each {|edition| update_all_nation_applicability(edition)}
--- a/db/migrate/20200814153802_add_all_nation_applicability_to_edition.rb
+++ b/db/migrate/20200814153802_add_all_nation_applicability_to_edition.rb
@@ -1,0 +1,5 @@
+class AddAllNationApplicabilityToEdition < ActiveRecord::Migration[5.1]
+  def change
+    add_column :editions, :all_nation_applicability, :boolean, default: true
+  end
+end
--- a/test/support/tests_for_national_applicability.rb
+++ b/test/support/tests_for_national_applicability.rb
@@ -88,7 +88,7 @@ module TestsForNationalApplicability
 
       assert_nil Edition.last
 
-      assert_page_has_error("Excluded nations must either allow all nations or exclude at least one nation")
+      assert_page_has_error("Excluded nations - you must select whether this content applies to all UK nations or which ones it excludes")
     end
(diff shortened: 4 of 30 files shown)


db/migrate/20200814153802_add_all_nation_applicability_to_edition.rb
```diff
@@ -0,0 +1,5 @@
+class AddAllNationApplicabilityToEdition < ActiveRecord::Migration[5.1]
+  def change
+    add_column :editions, :all_nation_applicability, :boolean
```
thomasleese, line 3: Can this have a `default: false`? (Or are we likely to see validation errors with older editions?)

edwardkerry: Whoops! 

I think that should be OK provided the data migration runs and sets any older editions without any inapplicabilities to `all_nation_applicability: true`.

Perhaps the default should be `true` though as that is the current behaviour unless the user opts out, and also for the other Edition types without the concept of `nation_inapplicability` that will now have this (unused) attribute.

thomasleese: Hmm yes, I think `default: true` does make more sense! Do we know how long the data migration runs for?

edwardkerry: Locally it took about 7 minutes with 156,849 records.
Integration has 165,733 and staging 165,777.

thomasleese: I guess this means during that time, there's the potential for validation errors on the existing editions that haven't been updated yet (although users can fix it manually by ticking or unticking the box?)?.

edwardkerry: I think that is correct - any validations could be user corrected. 

I also think this is going to be a managed release - there will be some comms around it so we could possibly ask users not to update for a half-hour window.

thomasleese: Cool, that makes sense 👍

db/data_migration/20200817081009_add_all_nation_applicability_to_editions.rb
```diff
@@ -0,0 +1,14 @@
+PUBLISHED_AND_PUBLISHABLE_STATES = %w[published draft archived submitted rejected scheduled].freeze
+EDITIONS_WITH_NATIONAL_APPLICABILITY = ['DetailedGuide', 'Publication', 'Consultation'].freeze
+
+edition_scope = Edition.where(type: EDITIONS_WITH_NATIONAL_APPLICABILITY).where(state: PUBLISHED_AND_PUBLISHABLE_STATES)
+
+def update_all_nation_applicability(edition)
+  if edition.nation_inapplicabilities.any?
+    edition.update_column(:all_nation_applicability, false)
+  else
+    edition.update_column(:all_nation_applicability, true)
```
thomasleese, line 10: I think we could skip this as we have `default: true` and that might save some time on this migration?

db/migrate/20200814153802_add_all_nation_applicability_to_edition.rb
edwardkerry: i can also now remove updating the record to `true` in the migration to save time.

edwardkerry: it seems that an unchecked check box will not send a "false" value, but merely nothing...meaning that an unchecked box will now default to true.

app/models/edition/national_applicability.rb
```diff
@@ -10,13 +10,23 @@ def process_associations_before_save(edition)
...
+    validate :applicability_options
     has_many :nation_inapplicabilities, foreign_key: :edition_id, dependent: :destroy, autosave: true
     validates_associated :nation_inapplicabilities
     validates :nation_inapplicabilities, length: { maximum: Nation.all.count - 1, message: "can not exclude all nations" }
 
     add_trait Trait
   end
 
+  def applicability_options
+    if !all_nation_applicability && nation_inapplicabilities.none? {|ni| ni.excluded?}
```
AlanGabbianelli, line 22:
```suggestion
    if !all_nation_applicability && nation_inapplicabilities.none?(&:excluded?)
```

I think this is the preferred style: https://www.rubydoc.info/gems/rubocop/0.26.1/RuboCop/Cop/Style/SymbolProc

test/support/tests_for_national_applicability.rb
```diff
@@ -21,6 +21,72 @@ module TestsForNationalApplicability
...
+        },
+        scotland: {
+          label: "Scotland",
+          applicable: true,
+        },
+        wales: {
+          label: "Wales",
+          applicable: true,
+        },
+      }
```
AlanGabbianelli, line 43: Since `expected_national_applicability` is exactly the same for both tests, can we pull it out and have it only once? It would avoid repetition and I think it would make the tests more readable.

test/support/tests_for_national_applicability.rb
```diff
@@ -123,8 +111,49 @@ module TestsForNationalApplicability
...
 
+    view_test "creating with all nations excluded should fail validation" do
+      create(:government)
+      attributes = attributes_for_edition(all_nation_applicability: "0")
+
+      all_nations = nation_inapplicabilities_attributes_for({
+        Nation.england => "http://www.england.com/",
+        Nation.scotland => "http://www.scotland.com/",
+        Nation.wales => "http://www.wales.com/",
+        Nation.northern_ireland => "http://www.ni.com/"
```
AlanGabbianelli, line 122: Are these the official websites? If not, it would be better to use the standard url used for examples: `http://www.example.com/england`, `http://www.example.com/scotland`, etc.

test/support/tests_for_national_applicability.rb
```diff
@@ -256,6 +289,10 @@ def create_edition(attributes = {})
     create(edition_class.name.underscore, attributes)
   end
 
+  def build_edition(attributes = {})
+    build(edition_class.name.underscore, attributes)
+  end
```
AlanGabbianelli, line 294: I don't see this being used, do we need it?

test/support/tests_for_national_applicability.rb
edwardkerry: good idea

test/support/tests_for_national_applicability.rb
edwardkerry: thanks, use was removed but this slipped through.

test/support/tests_for_national_applicability.rb
edwardkerry: good idea

app/models/edition/national_applicability.rb
edwardkerry: 👍

AlanGabbianelli (approved):

AlanGabbianelli:
It might be worth mentioning that the failing tests have been fixed here: https://github.com/alphagov/publishing-e2e-tests/pull/394

thomasleese (approved):

PR #5768, opened 2020-08-18
Send broken_links_report via S3 instead of email attachment

benjamineskola:
https://trello.com/c/BIes5aYv/2080-2-fix-brokenlinksreport-in-whitehall

### DIFF
--- a/app/controllers/broken_links_export_request_controller.rb
+++ b/app/controllers/broken_links_export_request_controller.rb
@@ -1,0 +1,16 @@
+class BrokenLinksExportRequestController < ApplicationController
+  def show
+    if user_signed_in?
+      filename = "broken-link-reports-#{params[:export_id]}.zip"
+
+      begin
+        file = S3FileHandler.get_file_from_s3(filename)
+        send_data(file, filename: filename)
+      rescue Fog::AWS::Storage::NotFound
+        head :not_found
+      end
+    else
+      head :unauthorized
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


app/controllers/broken_links_export_request_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  def show
+    if user_signed_in?
+      filename = "broken-link-reports-#{params[:export_id]}.zip"
+
+      begin
+        file = S3FileHandler.get_file_from_s3(filename)
+        send_data(file, filename: filename)
+      rescue Fog::AWS::Storage::NotFound
+        head :not_found
+      end
```
AlanGabbianelli, line 11: We can remove the begin block: https://www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Style/RedundantBegin

I was also wondering if it would be better to use a conditional rather than an exception for flow control here. 🤔 Something like `if file... else...`.

benjamineskola: I think the conditional was tried in one of the other places this was used (`publisher` perhaps) — I think the slight issue is that it requires an extra call to S3 in order to do so.

But perhaps it would make sense for `S3FileHandler` to handle that exception and return `nil` if there's no file, so that we can have a conditional here at least.

thomasleese (approved):

app/controllers/broken_links_export_request_controller.rb
AlanGabbianelli: > In fact `S3FileHandler` raises the exception if `file` is `nil`.
Ah, that's a shame. Oh, well! 🤷‍♂️

AlanGabbianelli (approved):

app/controllers/broken_links_export_request_controller.rb
benjamineskola: @AlanGabbianelli no, that's good :) It's our code so easily changed.

PR #5778, opened 2020-08-26
Add seniority to ministerial roles

callumknights:
Part of the work to migrate the rendering of /government/ministers
from whitehall to collections. Seniority will be used by collections
to sort the list of ministers so needed to be added to the content
item to collections had this information via link expansion.

More information here: https://trello.com/c/T6u4LjxS/2062-3-add-sorting-information-to-ministers-content-item

and here: https://trello.com/c/loAlkIp7/1715-8-migrate-government-ministers-to-collections

Relevant PR's:
https://github.com/alphagov/govuk-content-schemas/pull/999
https://github.com/alphagov/publishing-api/pull/1843

### DIFF
--- a/app/presenters/publishing_api/role_presenter.rb
+++ b/app/presenters/publishing_api/role_presenter.rb
@@ -58,7 +58,12 @@ module PublishingApi
         attends_cabinet_type: item.attends_cabinet_type&.name,
         role_payment_type: item.role_payment_type&.name,
         supports_historical_accounts: item.supports_historical_accounts,
-      }
+      }.merge(ministerial_role_details)
+    end
+
+    def ministerial_role_details
+      return {} unless item.is_a?(MinisterialRole)
+      { seniority: item.seniority }
     end
 
     def body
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/role_presenter.rb
```diff
@@ -58,7 +58,7 @@ def details
         attends_cabinet_type: item.attends_cabinet_type&.name,
         role_payment_type: item.role_payment_type&.name,
         supports_historical_accounts: item.supports_historical_accounts,
-      }
+      }.merge!( item.is_a?(MinisterialRole) ? {seniority: item.seniority} : {} )
```
thomasleese, line 61: I'd suggest splitting the logic here into a method which might make it easier to read. Perhaps something like this:

```suggestion
      }.merge(ministerial_role_details)
    end
    
    def ministerial_role_details
      return {} unless item.is_a?(MinisterialRole)
      { seniority: item.seniority }
```

thomasleese (commented):
Looks good, I've added a comment about making a code a bit cleaner, but the logic looks good!

app/presenters/publishing_api/role_presenter.rb
callumknights: This was my initial thinking, I was very unsure about the way I had done it but couldn't immediately think of a tidy alternative. Thanks!

thomasleese (approved):

PR #5780, opened 2020-08-27
Bulk organisation updater improvements

cbaines:


lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -35,10 +35,14 @@ def process_row(row)
 
     def find_document(row)
       slug = row.fetch("Slug")
+
+      return if slug.nil?
+      return if slug.strip.empty?
```
AlanGabbianelli, line 40: Since `"  ".blank?` is `true`, we can do
```suggestion
      return if slug.blank?
```

cbaines: Sure, do you want to make the change? I don't know how to use the "suggest change" thing on GitHub.

AlanGabbianelli: If I make the change then I can't approve the PR. If the change looks good to you, you can just edit the commit and force push.

cbaines: Ok, I've incorporated this in to the first commit now.

AlanGabbianelli (approved):

PR #5782, opened 2020-08-28
Handle document types in the bulk organisation updater

cbaines:
As two documents can have the same slug, if they're different types,
previously whichever document was updated was just decided by which
one was returned by the database.

These changes add a new column to the spreadsheet that can optionally
contain a document type.

### DIFF
--- a/lib/data_hygiene/bulk_organisation_updater.rb
+++ b/lib/data_hygiene/bulk_organisation_updater.rb
@@ -34,11 +34,22 @@ module DataHygiene
     end
 
     def find_document(row)
-      slug = row.fetch("Slug")
+      slug = row.fetch("Slug")&.strip
+      document_type = row.fetch("Document type")&.strip
 
       return if slug.blank?
 
-      document = Document.find_by(slug: slug)
+      if document_type.blank?
+        documents = Document.where(slug: slug.strip).to_a
+
+        if documents.length > 1
+          puts "error: ambiguous slug: #{slug} (document_types: #{documents.map(&:document_type)})"
+        else
+          document = documents.first
+        end
+      else
+        document = Document.find_by(slug: slug, document_type: document_type)
+      end
 
       if document.nil?
         puts "error: #{slug}: could not find document"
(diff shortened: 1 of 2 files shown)


lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -34,11 +34,22 @@ def process_row(row)
...
+      document_type = row.fetch("Document type")&.strip
 
       return if slug.blank?
 
-      document = Document.find_by(slug: slug)
+      if document_type.blank?
+        documents = Document.where(slug: slug.strip).to_a
+
+        if documents.length > 1
+          puts "error: ambiguous slug: #{slug} (document_types: #{documents.map(&:document_type).uniq})"
```
ChrisBAshton, line 46: I'd leave off the `.uniq` - if we end up in a situation where a slug is applied to two documents of the same type, this error will highlight the issue, whereas with `uniq` it would not.

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -34,11 +34,22 @@ def process_row(row)
...
+      slug = row.fetch("Slug")&.strip
+      document_type = row.fetch("Document type")&.strip
 
       return if slug.blank?
 
-      document = Document.find_by(slug: slug)
+      if document_type.blank?
+        documents = Document.where(slug: slug.strip).to_a
+
+        if documents.length > 1
```
ChrisBAshton, line 45: It feels like this condition is missing a test - can we add one that checks that two document types with the same slug raises the error you expect?

ChrisBAshton (commented):
Couple of suggestions below:

lib/data_hygiene/bulk_organisation_updater.rb
cbaines: That should be impossible https://github.com/alphagov/whitehall/blob/master/db/schema.rb#L290 and generally I trust database constraints.

cbaines: Which I guess means the `.uniq` is redundant, so I'll remove it anyway.

lib/data_hygiene/bulk_organisation_updater.rb
cbaines: Sure, I've included one now.

ChrisBAshton (approved):
Nice 🎉

PR #5783, opened 2020-08-28
Use edition links for most DetailedGuide links

cbaines:
For everything other than topics, I think the expectation is for this
to work on the level of the edition. Currently, users are changing the
organisations for drafts, and this is effecting the published
content. Using edition links will prevent this.


kevindew (approved):
Seems good to me, this is consistent with Content Publisher that also does edition links for all but the topic taxonomy. 

Though not sure this won't have any surprising consequences. I seem to remember Whitehall didn't send links when a draft already existed to get around the consequences of this issue.

cbaines:
This requires https://github.com/alphagov/whitehall/pull/5783

PR #5785, opened 2020-09-02
Remove heading

maxgds:
https://trello.com/c/awxxKXMf/330-attachments-confusing-accessibility-warning

Whitehall creates a block of text related to inaccessible file attachments that currently contains a spurious H2. This is bad for accessibility as it's both repetitive and confusing where not contextualised. This PR switches the H2 for a P tag. The contents of the H2 come from a translation file that also referred to it as a heading so that has also been updated. As a result of changing from an H2 to a paragraph the font size has changed, but it seems reasonable to not update that as there's no clear reason why it is significantly smaller than surrounding text currently.

This snippet of content is part of the published page so a republish will be needed on all relevant content to see the effects of this change when it goes live. To test and capture the "after" screenshot below I pushed the branch to integration and edited/previewed a content item there.

Screenshots of https://www.gov.uk/government/publications/our-plan-to-rebuild-the-uk-governments-covid-19-recovery-strategy

### Before
Screenshot 2020-09-02 at 15 29 25

### After
Screenshot 2020-09-02 at 15 28 46


owenatgov (approved):
Look a ok to me

PR #5787, opened 2020-09-03
Respect user defined organisation order when presenting links

brucebolt:
Users can put lead organisations into a specified order, however these were not being presented to Publishing API in that order. Instead, they were presented ordered by the database's organisation ID.

This had the effect of organisations appearing in a seemingly random order on GOV.UK.  While [government-frontend has code to put "emphasised organisations" first](https://github.com/alphagov/government-frontend/blob/2a09d4dde8c2caf1399f3eb65d94261f85a35488/app/presenters/content_item/linkable.rb#L43-L48), it doesn't know which order to put organisations in if multiple are considered emphasised.

Zendesk ticket: https://govuk.zendesk.com/agent/tickets/4200472

### DIFF
--- a/app/presenters/publishing_api/links_presenter.rb
+++ b/app/presenters/publishing_api/links_presenter.rb
@@ -37,7 +37,19 @@ module PublishingApi
     end
 
     def organisation_ids
-      (item.try(:organisations) || []).map(&:content_id)
+      if item.try(:edition_organisations)
+        item
+          .try(:edition_organisations)
+          .sort_by { |organisation| [organisation.lead_ordering ? 0 : 1, organisation.lead_ordering] }
+          .map(&:organisation)
+          .map(&:content_id)
+      elsif item.try(:organisations)
+        item
+          .organisations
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/links_presenter.rb
```diff
@@ -37,7 +37,19 @@ def statistical_data_set_ids
...
 
     def organisation_ids
-      (item.try(:organisations) || []).map(&:content_id)
+      if item.try(:edition_organisations)
+        item
+          .try(:edition_organisations)
+          .sort_by { |organisation| [organisation.lead_ordering ? 0 : 1, organisation.lead_ordering] }
+          .map(&:organisation)
+          .map(&:content_id)
+      elsif item.try(:organisations)
```
brucebolt, line 46: This is required because statistics announcements don't have `edition_organisations` (they have `statistics_announcement_organisations` instead).

cbaines (approved):
Sounds good to me :+1:

PR #5788, opened 2020-09-03
Re-work title of world page

owenatgov:
## What
* Alters the markup of the title on the world page [the world page](https://www.gov.uk/world).
* Downgrades other headings on the page eg: h2 to h3, h1 to h2.
* Remove styling for `.js-filter-form`
* Replace other instances of h1 tags within a `.page-header` element across whitehall

## Why
This page currently has a confusing heading structure, there are several h1s and the first h1 doesn't explicitly describe what the page is ("worldwide"). This replaces the title with a much clearer "Help and services around the world" as well as moving the filter input out of the title for a bit more clarity.

Some additional housekeeing why's:

* Reworking the heading structure of the rest of the page means it makes more sense. There now aren't multiple h1s and headings within sections are suitably deemphasised.
* Removing the form class styles allows the input element to make use of default govuk styling as opposed to the bespoke styling that's currently in place
* Replacing h1s across whitehall is a small quality of life change to get rid of some unsightly CSS around h1s within `.page-header`

### Before
Screenshot 2020-09-03 at 14 17 47

### After
Screenshot 2020-09-04 at 14 05 43

**Card:** https://trello.com/c/ok9ltp3q/339-problematic-heading-structure

### DIFF
--- a/app/views/world_locations/index.html.erb
+++ b/app/views/world_locations/index.html.erb
@@ -30,7 +30,7 @@
     <% @world_locations.each do |type, locations| %>
       <section id="<%= type.name.pluralize.parameterize %>" class="world-location-type js-filter-block">
         <header class="type-heading">
-          <h1><%= type.name.pluralize %></h1>
+          <h2 class="world-locations__heading"><%= type.name.pluralize %></h2>
           <p><span class="js-filter-count"><%= locations.length %></span> <span class="visuallyhidden">locations</span></p>
         </header>
         <div class="content">
--- a/app/views/world_locations/index.html.erb
+++ b/app/views/world_locations/index.html.erb
@@ -3,14 +3,25 @@
 
 <div class="block">
   <div class="inner-block">
-    <header class="page-header js-filter-list">
-      <h1 class="title">Worldwide</h1>
-      <form class="js-filter-form">
-        <label for="help-and-services-in" class="label">
-          Help and services in <div><input name="help-and-services-in" id="help-and-services-in" placeholder="Example: France" class="govuk-input"></div>
-        </label>
-      </form>
-    </header>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <header class="js-filter-list">
+          <%= render "govuk_publishing_components/components/title", {
+            title: sanitize("Help and services around the&nbsp;world")
+          } %>
+
+          <form class="js-filter-form">
+            <%= render "govuk_publishing_components/components/input", {
+              label: {
+                text: "Search by country or territory",
+              },
+              name: "help-and-services-in",
+              id: "help-and-services-in",
+            } %>
+          </form>  
+        </header>
+      </div>
+    </div>
   </div>
 </div>
 
--- a/app/views/world_locations/index.html.erb
+++ b/app/views/world_locations/index.html.erb
@@ -37,7 +37,7 @@
           <% if type === WorldLocationType::WorldLocation %>
             <% group_and_sort(locations).each do |letter, locations| %>
               <div class="js-filter-block">
-                <h2 class="letter"><%= letter %></h2>
+                <h3 class="letter"><%= letter %></h3>
                 <ol class="location-list">
                   <% locations.each do |location| %>
                     <%= render partial: 'world_location',
--- a/app/assets/stylesheets/frontend/helpers/_index-list.scss
+++ b/app/assets/stylesheets/frontend/helpers/_index-list.scss
@@ -40,30 +40,6 @@
 
   .js-filter-form {
     display: none;
-
-    label {
-      display: inline-block;
-    }
-    input {
-      display: inline-block;
-      margin: 7px $gutter-one-third 3px;
-      width: $half;
-      @include media(tablet) {
-        width: 30%;
-      }
-    }
-    div {
-      display: block;
-      input {
-        margin-left: 0;
-      }
-      @include media(tablet) {
-        display: inline;
-        input {
-          margin-left: $gutter-one-third;
-        }
-      }
-    }
   }
 
 }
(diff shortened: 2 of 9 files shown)


app/views/world_locations/index.html.erb
```diff
@@ -4,12 +4,26 @@
...
+        <%= render "govuk_publishing_components/components/input", {
+          label: {
+            text: "Search by country or territory",
+          },
+          name: "help-and-services-in",
+          id: "help-and-services-in",
+        } %>
+      </form>
+
+      <!-- <form class="js-filter-form">
```
danacotoran, line 21: Could this HTML comment be removed?

app/views/world_locations/index.html.erb
```diff
@@ -4,12 +4,26 @@
 <div class="block">
   <div class="inner-block">
     <header class="page-header js-filter-list">
-      <h1 class="title">Worldwide</h1>
+      <%= render "govuk_publishing_components/components/title", {
```
danacotoran, line 7: If we are using the proper title component, can we not remove all the whitehall level `h1` styling and leave it as the default `h1` styles instead?

**With whitehall styles**

Screenshot 2020-09-03 at 18 08 31


**With component styles** (just checked off the `#whitehall-wrapper .page-header h1 { }` rules in devtools)

Screenshot 2020-09-03 at 18 08 50


If it's not too much work, it might be good to look into how many instances of `.page-header` with a hard coded `h1` there are? 
If those instances can be easily replaced with the title component, we can remove those unsightly `#whitehall-wrapper .page-header h1 { }` rules. 🤷🏻‍♀️

danacotoran (commented):
Could we ask for Mia's input on the layout? 
The implementation looks slightly different from the mockup (for instance, the input field looks smaller in the mockup to my eyes) – it would be good to have a designer sign it off!

app/views/world_locations/index.html.erb
owenatgov: There are 202 instances of hard coded h1s 😬  I also don't yet have means to test everything thanks to AWS permissions. I agree that I'd like this to be fixed but are you happy to de-scope it for now?

owenatgov:
> Could we ask for Mia's input on the layout?
> The implementation looks slightly different from the mockup (for instance, the input field looks smaller in the mockup to my eyes) – it would be good to have a designer sign it off!

@danacotoran I've tried to squeeze it a bit further without dropping in any new styling, added new screenshot for reference. I'll run this by Mia just in case.

app/views/world_locations/index.html.erb
danacotoran: @owenatgov yes there are many hard coded `h1`s in Whitehall.
However the rules I highlighted only apply to hard coded `h1` inside a `page-header`, of which there are only 4 instances that I can see from doing a quick search.
It's definitely not within the scope of this PR though, so I will leave it to your discretion whether to do this or not.

app/views/world_locations/index.html.erb
```diff
@@ -19,7 +30,7 @@
     <% @world_locations.each do |type, locations| %>
       <section id="<%= type.name.pluralize.parameterize %>" class="world-location-type js-filter-block">
         <header class="type-heading">
-          <h1><%= type.name.pluralize %></h1>
+          <h2><%= type.name.pluralize %></h2>
```
injms, line 33: Could we use a heading component here?

injms: Related, but I can't comment on [line 40](https://github.com/alphagov/whitehall/pull/5788/files#diff-1d953526c50fc8bd376b4dae739632efR40).

I think it's worth updating the heading level for the letters to a level 3 heading - at the moment they're also level 2, so the hierarchy looks like this:



With a level 3 heading the letters would be nested underneath "World locations" - do you think that makes more sense?

owenatgov: I'm not sure a heading component is the right choice as this element isn't styled like a heading. It's the 2 text elements on (this page)[https://www.gov.uk/world] that read "world locations" and "international delegations". They don't look like headings for design/space reasons (I assume?) behave like headings for the purpose of screen reader navigation. Do we think this not looking like a heading is problematic and if so, should this be something we change as part of this PR?

injms: Apologies, I meant to swap the `h2` element for a heading component:

```suggestion
        <%= render "govuk_publishing_components/components/heading", {
          text: type.name.pluralize
        } %>
```

owenatgov: My comment holds up my dude. That `type.name.pluralize` element in the world page resolves twice to "world locations" and "international delegations". Maybe linking to live is a bit misleading as they're currently h1s with this PR swapping them out for h2s. Here's an example of the styling on my local:

Screenshot 2020-09-04 at 14 30 23

You can see that the h2 is being styled by whitehall to be tiny. The question again is do we think that these tiny h2s are problematic? I personally think they are problematic and we should style these like titles but there's the consideration of if making a design change like this is within scope. I guess we could?

owenatgov: I'll make the h3 change no problem though, good spot 👍

injms: Ah, okaydokey - thanks for being patient. 

Could we change the CSS to target the heading directly, as it's not clear that this is a custom heading:

```scss
.world-locations {
  h2 {
```

to be something like:
```scss
.app-world-locations-heading {
```

and 
```html
<h2 class="app-world-locations-heading">...
```

Or even remove the need for a custom CSS class altogether?

```html
<h2 class="govuk-body-l">World locations</h2>
```

injms (approved):
I'm happy with this if @danacotoran's happy 👍

PR #5795, opened 2020-09-09
Republish Organisation when an OrganisationRole is removed from a Role

edwardkerry:
When an Organisation is unlinked from a Role, this PR ensures that the Organisation will be republished to the PublishingAPI. It uses an Association callback,`after_remove`, to identify when an OrganisatonRole has been removed from a Role, and republish the associated Organisation.

[trello](https://trello.com/c/Nqwy3H8W/2106-5-unlinking-a-role-from-an-organisation-in-whitehall-does-not-republish-the-organisation-page)

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -25,7 +25,8 @@ class Role < ApplicationRecord
   has_many :current_people, class_name: "Person", through: :current_role_appointments, source: :person
 
   has_many :organisation_roles, inverse_of: :role
-  has_many :organisations, through: :organisation_roles
+  has_many :organisations, through: :organisation_roles,
+            after_remove: :republish_organisation_to_publishing_api
 
   has_many :worldwide_organisation_roles, inverse_of: :role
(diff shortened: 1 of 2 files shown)


app/models/role.rb
```diff
@@ -50,20 +51,24 @@ def self.columns
...
-  def republish_organisation_to_publishing_api
+  def republish_organisations_to_publishing_api
     organisations.each do |organisation|
-      Whitehall::PublishingApi.republish_async(organisation)
+      republish_organisation_to_publishing_api(organisation)
     end
   end
 
+  def republish_organisation_to_publishing_api(organisation)
+    Whitehall::PublishingApi.republish_async(organisation)
```
cbaines, line 69: `.bulk_republish_async` would better replicate how the Publishing API would prioritise this, if it was being done "properly" through dependency resolution. It doesn't really matter though, this is fine as well I think.

cbaines (approved):

thomasleese (approved):

PR #5799, opened 2020-09-10
Rescue bad request error and display a message

benjamineskola:
This is currently only a partial fix, as document exports are done from a worker and therefore can't feed back to the user (or at least not easily). One solution to that (and a nicer fix in general) would be if we could query the Notify API for a list of team members (or confirm that an address is in the team) before sending, rather than only upon sending. 

https://trello.com/c/l1qn22OR/2122-5-handle-not-in-team-notify-error-in-whitehall

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,6 +17,8 @@ class ApplicationController < ActionController::Base
   layout "frontend"
   after_action :set_slimmer_template
 
+  rescue_from Notifications::Client::BadRequestError, with: :notify_bad_request
+
 private
 
   def set_audit_trail_whodunnit
@@ -86,5 +88,9 @@ private
     if current_user && GdsApi::GovukHeaders.headers[:x_govuk_authenticated_user].nil?
       GdsApi::GovukHeaders.set_header(:x_govuk_authenticated_user, current_user.uid)
     end
+  end
+
+  def notify_bad_request(_exception)
+    render plain: "Error: One or more recipients not in GOV.UK Notify team (code: 400)", status: :bad_request
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -87,4 +89,8 @@ def set_authenticated_user_header
       GdsApi::GovukHeaders.set_header(:x_govuk_authenticated_user, current_user.uid)
     end
   end
+
+  def notify_bad_request(_exception)
+    render plain: "Error: One or more recipients not in GOV.UK Notify team (code: 400)", status: :bad_request
```
cbaines, line 94: It would probably be good to check the exception before giving such a specific message, as there's probably other things that can lead to a bad request I'm guessing?

ChrisBAshton: Probably worth writing a test for it too 🤔

benjamineskola: > there's probably other things that can lead to a bad request I'm guessing?

If I'm reading [their documentation](https://docs.notifications.service.gov.uk/ruby.html#error-codes) correctly, the only other thing that would raise a `BadRequestError` is if the service was in trial mode and couldn't send messages at all. I'm not actually sure in what situation that would occur that wasn't already covered by the not-in-team error, since rate limiting is a separate error code.

There are other reasons a `BadRequestError` might occur but they mostly relate to sending files etc, or on other API calls like retrieving a template, so they don't apply here — on the other hand, perhaps if there's even the possibility we might use file attachments in future it's better to be careful with this.

ChrisBAshton (approved):
Thanks - LGTM 👍

PR #5800, opened 2020-09-14
Allow mp4 uploads

oscarwyatt:
Allow mp4 uploads via whitehall
This is meant to be a temporary deploy in response to this ticket: https://govuk.zendesk.com/agent/tickets/4182931
Other approaches to uploading mp4 files are possible but are likely to have problems - this is by far the simplest and most reliable

### DIFF
--- a/app/uploaders/attachment_uploader.rb
+++ b/app/uploaders/attachment_uploader.rb
@@ -4,7 +4,7 @@ class AttachmentUploader < WhitehallUploader
 
   THUMBNAIL_GENERATION_TIMEOUT = 10.seconds
   FALLBACK_PDF_THUMBNAIL = File.expand_path("../assets/images/pub-cover.png", __dir__)
-  EXTENSION_WHITELIST = %w[chm csv diff doc docx dot dxf eps gif gml ics jpg kml odp ods odt pdf png ppt pptx ps rdf ris rtf sch txt vcf wsdl xls xlsm xlsx xlt xml xsd xslt zip].freeze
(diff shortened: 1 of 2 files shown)


app/uploaders/attachment_uploader.rb
```diff
@@ -4,7 +4,7 @@ class AttachmentUploader < WhitehallUploader
 
   THUMBNAIL_GENERATION_TIMEOUT = 10.seconds
   FALLBACK_PDF_THUMBNAIL = File.expand_path("../assets/images/pub-cover.png", __dir__)
-  EXTENSION_WHITELIST = %w[chm csv diff doc docx dot dxf eps gif gml ics jpg kml odp ods odt pdf png ppt pptx ps rdf ris rtf sch txt vcf wsdl xls xlsm xlsx xlt xml xsd xslt zip].freeze
+  EXTENSION_BLOCKLIST = %w[chm csv diff doc docx dot dxf eps gif gml ics jpg kml mp4 odp ods odt pdf png ppt pptx ps rdf ris rtf sch txt vcf wsdl xls xlsm xlsx xlt xml xsd xslt zip].freeze
```
cbaines, line 7: You're changing two things in one commit, it would be better to have two separate commits, or at least describe the second change as well.

I don't think blocklist is a suitable substitute for whitelist, as the semantics are reversed.

oscarwyatt: Good points

cbaines:
The commit message is a bit light on information, but I think the diff is fine now.

leenagupte (approved):

PR #5805, opened 2020-09-15
Make case study images optional

deborahchua:
We currently don't have the ability of not showing a main image on a case study. 
If an image is not added, the case study will use the default image (the default organisation image), but in some 
cases an image is not required, and does not feel relevant to the case study. This has been a specific request from 
a department (https://govuk.zendesk.com/agent/tickets/4105207)

Paired with @beccapearce 

Trello card: https://trello.com/c/HYtRDzzY/2093-make-case-study-images-optional-and-dont-use-default-image

### DIFF
--- a/app/assets/javascripts/admin/views/editions/_form.js
+++ b/app/assets/javascripts/admin/views/editions/_form.js
@@ -14,6 +14,7 @@
       this.toggleWorldNewsStoryVisibility();
       this.resetWorldNewsStoryFields();
       this.toggleFirstPublishedDate();
+      this.showImageUploaderIfCustomImage();
 
       GOVUK.formChangeProtection.init($('#edit_edition'), 'You have unsaved changes that will be lost if you leave this page.');
     },
@@ -265,6 +266,22 @@
         $(this).parent().hide().next().removeClass('if-js-hide');
         e.preventDefault();
       });
+    },
+
+    showImageUploaderIfCustomImage: function showImageUploaderIfCustomImage() {
+      var $imageUploader = $('.image_section .js-show-image-uploader');
+      var $customImageRadioButton = $('#edition_image_display_option_custom_image');
+      var $radioButtons = $('.image_section input[type=radio]');
+
+      function showOrHideImageUploader() {
+        if ($customImageRadioButton.prop('checked')){
+          $imageUploader.show();
+        } else {
+          $imageUploader.hide();
+        }
+      }
+      $radioButtons.on('change', showOrHideImageUploader);
+      showOrHideImageUploader();
     }
   }
 
--- a/app/presenters/publishing_api/case_study_presenter.rb
+++ b/app/presenters/publishing_api/case_study_presenter.rb
@@ -54,7 +54,11 @@ module PublishingApi
         first_public_at: first_public_at,
         format_display_type: item.display_type_key,
       }
-      details_hash[:image] = image_details if image_available?
+      details_hash[:image] = if image_available? && image_required?
+                               image_details
+                             else
+                               { url: "", caption: nil, alt_text: "" }
+                             end
       details_hash.merge!(PayloadBuilder::TagDetails.for(item))
       details_hash.merge!(PayloadBuilder::BrexitNoDealContent.for(item))
     end
@@ -79,6 +83,10 @@ module PublishingApi
 
     def image_available?
       item.images.any? || emphasised_organisation_default_image_available?
+    end
+
+    def image_required?
+      item.image_display_option != "no_image"
     end
 
     def emphasised_organisation_default_image_available?
--- a/app/views/admin/editions/_image_fields_case_studies.html.erb
+++ b/app/views/admin/editions/_image_fields_case_studies.html.erb
@@ -1,0 +1,64 @@
+<fieldset class="image_section">
+  <div class="radio">
+    <label>
+      <%= form.radio_button(:image_display_option, "no_image") %>
+      <strong>Don't use an image</strong>
+      <p class="hint">No image will be shown on page</p>
+    </label>
+  </div>
+  <div class="radio">
+    <label>
+      <%= form.radio_button(:image_display_option, "organisation_image") %>
+      <strong>Use the default organisation image</strong>
+      <p class="hint">The image set in the organisation page will be shown</p>
+    </label>
+  </div>
+  <div class="radio">
+    <label>
+      <%= form.radio_button(:image_display_option, "custom_image") %>
+      <strong>Use a custom image</strong>
+      <p class="hint">Upload a custom image</p>
+    <label>
+  </div>
+  <div class="js-show-image-uploader show-image-uploader">
+    <%= content_tag :fieldset, id: "image_fields", class: "images multiple_file_uploads#{' right-to-left' if edition.rtl?}" do %>
+      <% i = 0 %>
+        <%= form.fields_for :images do |images_fields| %>
+          <% if images_fields.object.new_record? %>
+            <div class="file_upload well">
+              <h3 class="remove-top-margin">New image</h3>
+              <p>Images must be 960px wide and 640px tall.</p>
+              <%= images_fields.fields_for :image_data do |image_data_fields| %>
+                <%= image_data_fields.upload :file, required: false, class: 'js-upload-image-input' %>
+              <% end %>
+              <%= images_fields.text_field :alt_text, label_text: 'Alt text' %>
+              <%= images_fields.text_area :caption, rows: 2 %>
+            </div>
+          <% else %>
+            <% i += 1 %>
+            <div class="row image">
+              <%= images_fields.hidden_field :id %>
+              <div class="col-md-6 fields-container">
+                <%= images_fields.text_field :alt_text, label_text: 'Alt text' %>
+                <%= images_fields.text_area :caption, rows: 2 %>
+                <% checkbox_args = nested_attribute_destroy_checkbox_options(images_fields, label_text: "(uncheck to remove)") %>
+                <%= images_fields.check_box :_destroy, *checkbox_args  %>
+                <% if edition.image_disallowed_in_body_text?(i) %>
+                  <p>This image is shown automatically, and is not available for use inline in the body.</p>
+                <% else %>
+                  <p>Markdown to use:
+                    <input type="text" readonly="readonly" value="!!<%= i %>" />
+                  </p>
+                <% end %>
+              </div>
+              <div class="col-md-6">
+                <a href="<%= images_fields.object.url %>">
+                  <%= image_tag(images_fields.object.url, class: "img-responsive") %>
+                </a>
+              </div>
+            </div>
+          <% end %>
+        <% end %>
+      <% end %>
+  </div>
+</fieldset>
--- a/db/data_migration/20200924134112_update_image_display_option_for_editions.rb
+++ b/db/data_migration/20200924134112_update_image_display_option_for_editions.rb
@@ -1,0 +1,11 @@
+case_studies_with_images = Edition.find_by_sql(["SELECT e.* FROM editions e LEFT OUTER JOIN images ON e.id = images.edition_id WHERE type in ('CaseStudy') AND images.id IS NOT NULL"])
+
+case_studies_with_images.each do |case_study|
+  case_study.update_columns(image_display_option: "custom_image")
+end
+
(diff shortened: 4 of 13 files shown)


db/migrate/20200909141146_add_image_display_option_to_editions.rb
```diff
@@ -0,0 +1,15 @@
+class AddImageDisplayOptionToEditions < ActiveRecord::Migration[5.1]
+  EDITIONS_TYPES_WITH_NO_IMAGES = ["CorporateInformationPage", "DocumentCollection", "StatisticalDataSet"].freeze
+
+  def change
+    add_column :editions, :image_display_option, :string, default: "organisation_image"
+
+    editions_with_images = Edition.find_by_sql(["SELECT e.* FROM editions e LEFT OUTER JOIN images ON e.id = images.edition_id WHERE type in ('CaseStudy', 'Consultation', 'DetailedGuide', 'FatalityNotice', 'NewsArticle', 'Publication', 'Speech') AND images.id IS NOT NULL"])
```
ChrisBAshton: It would be nice to see this use the same approach as you've used on line 2, so that we can see at a glance which edition types have images and which don't. I.e.

```ruby
EDITIONS_TYPES_WITH_IMAGES = ['CaseStudy', 'Consultation', 'DetailedGuide', 'FatalityNotice', 'NewsArticle', 'Publication', 'Speech'].freeze
# ...
    editions_with_images = Edition.find_by_sql(["SELECT e.* FROM editions e LEFT OUTER JOIN images ON e.id = images.edition_id WHERE type in (#{EDITIONS_TYPES_WITH_IMAGES.to_s}) AND images.id IS NOT NULL"])
```

app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -265,6 +266,27 @@
         $(this).parent().hide().next().removeClass('if-js-hide');
         e.preventDefault();
       });
+    },
+
+    showImageUploaderIfCustomImage: function showImageUploaderIfCustomImage() {
+      var $imageUploader = $('.image_section .js-show-image-uploader');
+      var $customImage = $('#edition_image_display_option_custom_image');
```
ChrisBAshton, line 273: I'd rename this to `$customImageRadioButton` so it's a little clearer what it's referring to.

app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -265,6 +266,27 @@
...
+          $imageUploader.show();
+        } else {
+          $imageUploader.hide();
+        }
+      }
+      $radioButtons.on('change', showOrHideImageUploader);
+      showOrHideImageUploader();
+
+      $('.image_fields a').on('click', function(e) {
+        $(this).parent().hide().next().removeClass('if-js-hide');
```
ChrisBAshton, line 287: This anonymous function isn't particularly self-documenting, unlike the rest of the (excellent!) JS code above it. Could you use some well named functions/variables to describe what's going on here?

app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -265,6 +266,27 @@
...
+        if ($customImage.prop('checked')){
+          $imageUploader.show();
+        } else {
+          $imageUploader.hide();
+        }
+      }
+      $radioButtons.on('change', showOrHideImageUploader);
+      showOrHideImageUploader();
+
+      $('.image_fields a').on('click', function(e) {
```
ChrisBAshton, line 286: I think this will only work on click, which wouldn't be accessible to people who need to use a keyboard. `on('submit'` is usually a better choice, but I'm not sure it'll work with links (it's usually used for buttons). May need some experimentation to get working for keyboard users 🤔

app/presenters/publishing_api/case_study_presenter.rb
```diff
@@ -54,7 +54,11 @@ def details
         first_public_at: first_public_at,
         format_display_type: item.display_type_key,
       }
-      details_hash[:image] = image_details if image_available?
+      if image_available? && image_required?
+        details_hash[:image] = image_details
+      else
+        details_hash[:image] = { url: "", caption: nil, alt_text: "", }
```
ChrisBAshton, line 60: Does this actually set the image as `null` in content store, or does it set it to this JSON?

If it's the latter, is that what we want? It seems an odd thing to set in the content store, though I don't know much about it.

db/data_migration/20200924134112_update_image_display_option_for_editions.rb
```diff
@@ -0,0 +1,7 @@
+editions_with_images = Edition.find_by_sql(["SELECT e.* FROM editions e LEFT OUTER JOIN images ON e.id = images.edition_id WHERE type in ('CaseStudy', 'Consultation', 'DetailedGuide', 'FatalityNotice', 'NewsArticle', 'Publication', 'Speech') AND images.id IS NOT NULL"])
```
ChrisBAshton: This migration isn't referenced in the commit - I assume it was accidentally committed here and should be removed?

ChrisBAshton (commented):
Have had a quick look, well done for dealing with both JS and Ruby in one PR! 👍 
A few questions / suggestions below.

app/assets/javascripts/admin/views/editions/_form.js
beccapearce: Just looked into this and I think the `if-js-hide` class has been removed so I think I can remove the method. It was meant to be for some text that says only to fill in the field if you wanted to add an image but since it isn't marked as required and users can leave the field blank on the form typically we've taken the text out. I'll remove it now.

Sorry I missed it! 🙈

app/presenters/publishing_api/case_study_presenter.rb
beccapearce: My mistake that commit message is very misleading! We do set it to this JSON because the content-schema requires url and alt_text to be a string, so we couldn't set it as nil. 

If we don't set the image at all then it picks up on the organisation image, this was our way of stopping it from grabbing any organisation image that exists

db/data_migration/20200924134112_update_image_display_option_for_editions.rb
beccapearce: Think it's made it's way into the wrong commit, I've moved it to the `Fix up migrations for image_display_option` where I talk about adding a data_migration

db/migrate/20200909141146_add_image_display_option_to_editions.rb
beccapearce: I think this is replaced with another migration later. But I can add it in if it'll be useful for people going forward :)

app/views/admin/editions/_image_fields_case_studies.html.erb
```diff
@@ -32,7 +32,7 @@
                 <%= image_data_fields.upload :file, required: false, class: 'js-upload-image-input' %>
               <% end %>
               <%= images_fields.text_field :alt_text, label_text: 'Alt text' %>
-              <%= images_fields.text_area :caption, rows: 2 %>
+              <%= images_fields.text_field :caption %>
```
ChrisBAshton, line 35: Hmmm I'm not sure why this change is necessary. You shouldn't need to change the input type to get something showing or to get a test to pass 😕  It's introduced some conditionals in `test/support/admin_edition_controller_test_helpers.rb ` that adds complexity.

I'd be tempted to drop this commit and find another fix for the issue - would you like to pair on that? Or are there some steps I can follow to reproduce the issue myself?

db/migrate/20200909141146_add_image_display_option_to_editions.rb
ChrisBAshton: Fair enough - I've made a suggestion in another comment to fix up the migration to this commit so that this file goes away completely! 👍

ChrisBAshton (commented):
Sorry I forgot about this yesterday! The PR is shaping up nicely now, most of my comments are to do with commit history.

I do have one query about the textarea/input replacement and the motivation behind that - sounds like that's worth spending a bit more time on and fixing the issue rather than finding a workaround.

Please click the "Request a re-review" button in GitHub when this is ready for review next - that should stop me forgetting! 😊

app/assets/javascripts/admin/views/editions/_form.js
beccapearce: I think the `on('click'` part has been removed with the removal of the anonymous function

app/views/admin/editions/_image_fields_case_studies.html.erb
beccapearce: This was a really weird issue. So if the caption field is set to a text_area it doesn't appear when the rest of the image_fields appear.

Screenshot 2020-10-13 at 12 59 55
^^ This screenshot is when the field is set to be a text_area. It's on the page to create a new case_study after clicking on the `Use a custom image` radio button.

The text area method comes from [here](https://github.com/alphagov/whitehall/blob/774eb78cd4c75622e786bdb18762850cd0b91840/lib/whitehall/form_builder.rb#L88-L97), I wasn't sure if it was running in a peculiar order after adding in the javascript, I tried some debugging in this text_area method and in the new showHide function but can't spot anything obvious as to why it isn't showing.

I would love a hand with it if you have some time :)

app/assets/javascripts/admin/views/editions/_form.js
ChrisBAshton: Oops, of course! Thanks. I'll resolve this convo 👍

app/views/admin/editions/_image_fields_case_studies.html.erb
```diff
@@ -0,0 +1,64 @@
+<fieldset class="image_section">
+  <div class="radio">
+    <%= form.label :image_display_option, (form.object.image_display_option), class: 'radio' do %>
```
ChrisBAshton, line 3: Think this needs to be a [`labelled_radio_button`](https://github.com/alphagov/whitehall/blob/774eb78cd4c75622e786bdb18762850cd0b91840/lib/whitehall/form_builder.rb#L17) instead - although replacing the entire line with `<label>` works as well and is more readable (line 7 needs to be `</label>` in that case).

This fixes the accessibility such that you can click anywhere on the text to select the radio (you don't have to click within the radio circle).

app/views/admin/editions/_image_fields_case_studies.html.erb
ChrisBAshton: Argh, it looks like a [training feature](https://github.com/alphagov/whitehall/blob/41b611f3d8c092706b3a27f7063b6d014a0efb88/config/initializers/enable_words_to_avoid_highlighting.rb#L7-L9) is interfering with the output of the textarea :( I've had a quick play and can't get this `highlightTextarea` thing to go away (even if I set that const to false). Turning off JavaScript makes the textarea render as it should.

Screenshot 2020-10-13 at 15 40 28

I have to leave right now but I'll have another play when I come back - it's worth investigating if that feature is even still being used!

ChrisBAshton (commented):
Have had a quick look locally and yes, looks like the textarea thing is a pain!

A quick suggestion about fixing the radio buttons so the labels are accessible:

app/views/admin/editions/_image_fields_case_studies.html.erb
ChrisBAshton: I've had another look and found a fix - details in the commit: https://github.com/alphagov/whitehall/commit/7a455d094b27ad966c161f5af7b01123d2330ae1

beccapearce: WOOO Awesome!!

beccapearce: I've cherry-picked you commit over and made a small change to apply it to the summary text-area as well as the body to get the tests passing

app/views/admin/editions/_image_fields_case_studies.html.erb
ChrisBAshton: Cool, so this works but you've got a redundant label in there at the moment - if you inspect the HTML (Right Click -> View Source) you'll see something like:

```html
<label class="radio" for="bla">
  <label>
    <input type="radio" />
    ...etc
  </label>
</label>
```

The outer label can be removed, i.e. lines 3 and 9 of this file (and the same lines in the subsequent radio sections below).

ChrisBAshton (commented):
This is coming along great - just a couple of very minor things now 👍

app/views/admin/editions/_image_fields_case_studies.html.erb
ChrisBAshton: Nice - just tested it locally and it works 👍

ChrisBAshton (approved):
Looks great 🎉

PR #5807, opened 2020-09-16
Rename whitelist to allowlist

oscarwyatt:


mammykins (approved):
this has underscores whereas the previous word didn't; is there a standard for this?

PR #5809, opened 2020-09-17
Add whips information to role presenter

callumknights:
Adding the information about the given whip organisation for ministerial
roles that are whips.

Part of the work to migrate rendering of the ministers index page
to collections.

https://trello.com/c/loAlkIp7/2063-8-migrate-government-ministers-to-collections

The purpose of these changes is to allow collections to sort the list
of whips at a rendering level. This will eventually allow some of the
function of the `ministerial_roles_controller` to be removed out to
collections.

https://github.com/alphagov/whitehall/blob/1512033670b30512872b56fbf4801cb438be55ec/app/controllers/ministerial_roles_controller.rb#L43-L60

### DIFF
--- a/app/presenters/publishing_api/role_presenter.rb
+++ b/app/presenters/publishing_api/role_presenter.rb
@@ -52,12 +52,24 @@ module PublishingApi
       end
     end
 
+    def whip
+      return {} if item.whip_organisation_id == nil
+
+      whip_organisation = Whitehall::WhipOrganisation.find_by_id(item.whip_organisation_id) 
+
+      {
+        sort_order: whip_organisation.sort_order,
+        label: whip_organisation.label,
+      }
+    end
+
     def details
       {
         body: body,
         attends_cabinet_type: item.attends_cabinet_type&.name,
         role_payment_type: item.role_payment_type&.name,
         supports_historical_accounts: item.supports_historical_accounts,
+        whip_organisation: whip,
       }.merge(ministerial_role_details)
     end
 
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/role_presenter.rb
```diff
@@ -52,12 +52,24 @@ def polymorphic_path
       end
     end
 
+    def whip
+      return {} unless item.whip_organisation_id != nil
```
cbaines, line 56: There's a double negative here, and the code would probably be easier to read if this was avoided.

app/presenters/publishing_api/role_presenter.rb
```diff
@@ -52,12 +52,24 @@ def polymorphic_path
...
+
+      organisation = Whitehall::WhipOrganisation.find_by_id(item.whip_organisation_id) 
+
+      {
+        sort_order: organisation.sort_order,
+        label: organisation.label,
+      }
```
cbaines, line 63: Is the label here in another content item? If so, it would probably be better to use links, rather than the details of this content item, to avoid things getting out of sync.

test/unit/presenters/publishing_api/role_presenter_test.rb
```diff
@@ -59,6 +60,67 @@ def present(model_instance, options = {})
...
+    }
+
+    presented_item = present(role)
+
+    assert_equal expected_hash, presented_item.content
+    assert_hash_includes presented_item.links, expected_links
+    assert_equal "major", presented_item.update_type
+    assert_equal role.content_id, presented_item.content_id
+
+    print presented_item.content
```
cbaines, line 119: This looks like left over debugging code.

callumknights: Cleared, thanks for picking that up 👍

app/presenters/publishing_api/role_presenter.rb
callumknights: Renamed for clarity 👍

cbaines (approved):

PR #5812, opened 2020-09-18
Add temp Rake task to update NewsArticles with inaccessible alt text

edwardkerry:
New content recommendations related to accessibility advise to leave alt text blank and to describe the image in the content of the page. (https://www.gov.uk/guidance/content-design/images).

Previously Whitehall would not allow images to be created without alt text. This forced users into entering placeholder text, which would create a tag which screenreaders would read e.g: 
text: `alt=""` => tag: `alt="alt="""`
text: `*` => tag: `alt="*"`

To improve accessibility, this PR adds a one-off Rake task that will remove the alt text from all identified instances of NewsArticles and republish the documents, to ensure empty alt text will display correctly as `alt=""`.

[trello](https://trello.com/c/GmO8zKt7/2142-3-update-certain-newsarticle-image-alt-texts-and-republish-newsarticles)


lib/tasks/remove_image_alt_text.rake
```diff
@@ -0,0 +1,23 @@
+require 'csv'
+
+namespace :news_article do
+  desc "Remove inaccessible alt text"
+  task remove_inaccessible_alt_text: :environment do
+
+    ids = CSV.read(File.join(Rails.root, 'lib', 'tasks', 'newsarticle_alt_text_ids.csv'), :headers=>true)
```
thomasleese, line 7:
```suggestion
    ids = CSV.read(Rails.root.join('lib', 'tasks', 'newsarticle_alt_text_ids.csv'), :headers=>true)
```

I think the linter might complain if you don't do this.

thomasleese (approved):

lib/tasks/remove_image_alt_text.rake
edwardkerry: seem to have got away with it, but will change it as it's much nicer, thanks!

thomasleese (approved):

PR #5814, opened 2020-09-18
Allow indexing AccessibleDocumentsPolicy documents of non-live organisations

thomasleese:
We want to make it possible for corporate information pages of type `AccessibleDocumentsPolicy` to be indexed even when the parent organisation is in a different state to `live`.

This is part of the GOV.UK Accessibility work - this type of document needs to be available for all organisations that publish on GOV.UK.

[Trello Card](https://trello.com/c/T9yzMvXF/2141-5-make-accessible-documents-policies-from-non-live-organisations-available-to-users)

### DIFF
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -53,12 +53,18 @@ class CorporateInformationPage < Edition
   def self.search_only
     live_govuk_status = super.with_organisation_govuk_status("live")
 
-    accessible_govuk_status = super
+    accessible_other_govuk_status = super
       .accessible_documents_policy
       .with_organisation_govuk_status(%w[joining exempt transitioning])
 
+    accessible_devolved_govuk_status = super
+      .accessible_documents_policy
+      .with_organisation_govuk_status("closed")
+      .where(organisations: { govuk_closed_status: "devolved" })
+
     live_govuk_status
-      .or(accessible_govuk_status)
+      .or(accessible_other_govuk_status)
+      .or(accessible_devolved_govuk_status)
   end
 
   def title_required?
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -27,6 +27,8 @@ class CorporateInformationPage < Edition
   validates :corporate_information_page_type_id, presence: true
   validate :only_one_organisation_or_worldwide_organisation
 
+  scope :with_organisation_govuk_status, ->(status) { joins(:organisation).where(organisations: { govuk_status: status }) }
+
   def republish_organisation_to_publishing_api
     Whitehall::PublishingApi.republish_async(owning_organisation) if owning_organisation.is_a?(Organisation)
   end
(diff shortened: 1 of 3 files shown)


edwardkerry (approved):

app/models/corporate_information_page.rb
```diff
@@ -51,8 +51,16 @@ def search_index
...
+    live_govuk_status = super.with_organisation_govuk_status("live")
+
+    accessible_other_govuk_status = super.accessible_documents_policy
+      .with_organisation_govuk_status(%w[joining exempt transitioning])
+
+    accessible_devolved_govuk_status = super.accessible_documents_policy
+      .with_organisation_govuk_status("closed")
+      .where(organisations: { govuk_closed_status: "devolved" })
+
+    live_govuk_status.or(accessible_other_govuk_status).or(accessible_devolved_govuk_status)
```
ChrisBAshton, line 63: This could be broken onto multiple lines, I think? Would read better:

```suggestion
    live_govuk_status
      .or(accessible_other_govuk_status)
      .or(accessible_devolved_govuk_status)
```

test/unit/corporate_information_page_test.rb
```diff
@@ -207,6 +207,24 @@ def self.should_be_invalid_without(type, attribute_name)
     corporate_information_page4.update_in_search_index
   end
 
+  test "will be indexed if the org it belongs to is not live on gov.uk and the document is an accessible policy" do
```
ChrisBAshton, line 210: We could make the intent a little clearer:

```suggestion
  test "will index accessible policy document even if the org it belongs to is not live on gov.uk" do
```

app/models/corporate_information_page.rb
```diff
@@ -51,8 +51,16 @@ def search_index
...
-    # Ensure only CIPs associated with a live Organisation are indexed in search.
-    super.with_organisation_govuk_status("live")
+    live_govuk_status = super.with_organisation_govuk_status("live")
+
+    accessible_other_govuk_status = super.accessible_documents_policy
+      .with_organisation_govuk_status(%w[joining exempt transitioning])
+
+    accessible_devolved_govuk_status = super.accessible_documents_policy
+      .with_organisation_govuk_status("closed")
+      .where(organisations: { govuk_closed_status: "devolved" })
```
ChrisBAshton, line 61: I'm having a bit of trouble understanding all the states here - it might be worth splitting this into a couple of commits, e.g. starting with just `live_govuk_status.or(accessible_other_govuk_status)` here and then working your way towards the `accessible_devolved_govuk_status`, which could be its own test case (and commit message).

ChrisBAshton (commented):
LGTM from an outsider's perspective, but would like to be a little more confident I know what's happening in the final commit, so just tentatively commenting for now 😅

app/models/corporate_information_page.rb
thomasleese: Yep, that's a good idea!

test/unit/corporate_information_page_test.rb
thomasleese: Yeah, that's worded much better than what I had written...

app/models/corporate_information_page.rb
thomasleese: I've split up the commits and hopefully that makes things clearer. What do you think?

ChrisBAshton (approved):
Much better - thanks! ⭐

PR #5817, opened 2020-09-21
Add aria label to site header navigation

danacotoran:
The site header links are a nav element, often appearing on pages which have other nav elements. When there are multiple navigation landmarks on a page, they should be labelled in order to help assistive tech users differentiate between them.

This adds `aria-label` to the site header nav. The suggested name was "Departments and policy navigation" because in the footer the section containing these links is called 'Departments and policy'.

No visual changes should occur as a result of this.

Screenshot 2020-09-18 at 13 03 21

https://trello.com/c/XFKhZwEI

### DIFF
--- a/app/views/layouts/frontend.html.erb
+++ b/app/views/layouts/frontend.html.erb
@@ -1,5 +1,5 @@
 <%= render :layout => 'layouts/frontend_base' do %>
-  <nav id="proposition-menu" class="no-proposition-name">
+  <nav id="proposition-menu" class="no-proposition-name" aria-label="Departments and policy navigation">
     <ul id="proposition-links">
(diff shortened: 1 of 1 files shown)


injms (approved):
🚀

owenatgov (approved):
Works fine on my local, just one question:

I was going to suggest that this whole component be replaced by a [layout header](https://components.publishing.service.gov.uk/component-guide/layout_header) component but then noticed that said component doesn't actually have any means to pass an `aria-label` and has a different `aria-label` setup: no `aria-label` on the `nav` and instead one on the `ul` which reads `Top Level Navigation`. How come the work wasn't done at the component level so that it could be piped in here (if it's even possible to get it into whitehall)?

PR #5818, opened 2020-09-22
Use SimpleCov to monitor code coverage

benthorner:
https://trello.com/c/epNWudCR/176-write-an-rfc-for-continuous-deployment-and-seek-review

We should always monitor the amount of code that is exercised when
we run our tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ group :test do
   gem "mocha"
   gem "rack-test"
   gem "rails-controller-testing"
+  gem "simplecov"
   gem "timecop"
   gem "webmock", require: false
 end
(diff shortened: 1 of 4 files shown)


benthorner:
Closing as this is proving to be very difficult to integrate with the multiple layers of processes and parallel testing in this app [[1](https://github.com/alphagov/whitehall/blob/master/script/test_queue)] [[2](https://github.com/alphagov/whitehall/blob/96190ad6a3622cfb92d72512356c1eb82a7b302a/test/parallel_test_runner.rb#L1)]. It's unclear why the tests in this repo are run in such a complex way.

The final code of this PR does produce a usable coverage result (around 90%) by running `bundle exec rake test cucumber` after fixing a couple of other issues with coverage for this repo:

- The automatic detection of the test suite is faulty for MiniTest specs, so needs hard-coding.
- The duration of the tests causes coverage results to be lost unless the merge timeout is increased.

PR #5819, opened 2020-09-22
Use translated organisation name in organisation logo if available

andysellick:
Use the translated organisation name in the organisation logo on organisation pages, if it's not the English version of the organisation page. Organisation.

- organisation names on non-English pages seem to all be in English
- this change returns the translated organisation name instead of logo_formatted_name for the organisation logo, if the page language is not English
- shouldn't interfere with the existing code for organisation_logo_name, which is already in use in csv_preview, operational fields and topical_events (thanks @sihugh)

Example pages:

- https://www.gov.uk/world/organisations/british-consulate-general-montreal.fr
- https://www.gov.uk/world/organisations/british-consulate-general-guangzhou.zh

Before | After
------ | ------
Screenshot 2020-09-22 at 13 55 15 | Screenshot 2020-09-22 at 13 54 56

Trello card: https://trello.com/c/bQn51uw9/392-update-language-attribute-of-current-organisation-name-on-world-organisation-embassy-pages

### DIFF
--- a/app/helpers/logo_helper.rb
+++ b/app/helpers/logo_helper.rb
@@ -30,4 +30,12 @@ module LogoHelper
       tag.span(class: css_classes) { tag.span { linked_logo } }
     end
   end
+
+  def translated_organisation_logo_name(organisation)
+    if I18n.locale == "en"
+      format_with_html_line_breaks(ERB::Util.html_escape(organisation.logo_formatted_name))
+    else
+      organisation.name
+    end
+  end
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Looks fine to me code wise but I'd recommend getting a 🥈 👀 on this, mainly because I can't run this reliably on my local because of AWS stuff.

sihugh (approved):

PR #5816, opened 2020-09-21
Fix worldwide organisation corporate information template

injms:
## What

 - Replaced page navigation with [content lists component](https://components.publishing.service.gov.uk/component-guide/contents_list)
 - Fixed wobbly margins by using the GOV.UK Frontend grid - it was already being used in the header partial inside this template, but not in the template itself.
- Fixed the page hierarchy to no longer have multiple `h1`s on the page
 - Removed CSS that now is no longer needed
 - Changed organisation logo to be inline to avoid humungously large link area
- Update focus state for '+X others' link
- Linted hide other links JavaScript file

## Why

The focus states for the internal page links and the '+X others' links on this template needed updated to use the latest styles from GOV.UK Frontend.

There were multiple `h1`s on the page - this has been patched, but really needs the page and content to be rejigged.

## Visual differences

Before:

image

After:

image

Before:
<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td>

image

</td><td>

image

</td></tr>
<tr><td>

image

</td><td>

image

</td></tr>
</table>

Heading structure

<table><tr><th>Before</th><th>After</th></tr>
<tr><td>

image

</td><td>

image

</td></tr></table>

PR #5823, opened 2020-09-22
Add lang attribute to page contents title

danacotoran:
In some instances, on translated world location pages, the "Contents" title will be in English and not marked as such using an appropriate lang attribute.

image

This adds `lang="en"` where "Contents" is in English, to ensure that screen readers read this word in an understandable fashion.

Example of a page where this problem occurs: https://www.gov.uk/world/organisations/british-embassy-moscow/office/british-embassy-moscow-main-contact.ru

### DIFF
--- a/app/views/corporate_information_pages/show_worldwide_organisation.html.erb
+++ b/app/views/corporate_information_pages/show_worldwide_organisation.html.erb
@@ -8,7 +8,9 @@
     <% if headers.any? %>
       <nav>
         <div class="content">
-          <h1><%= t('worldwide_organisation.headings.contents', default: 'Contents') %></h1>
+          <h1 <%= "lang=#{t_fallback('worldwide_organisation.headings.contents')}" if t_fallback('worldwide_organisation.headings.contents') %>>
+            <%= t('worldwide_organisation.headings.contents', default: 'Contents') %>
+          </h1>
           <ol>
             <% headers.each do |header| %>
               <li><%= link_to header.text, "##{header.id}" %></li>
--- a/app/views/worldwide_offices/show.html.erb
+++ b/app/views/worldwide_offices/show.html.erb
@@ -8,7 +8,7 @@
     <% if headers.any? %>
       <nav>
         <div class="content">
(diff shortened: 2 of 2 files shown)


app/views/corporate_information_pages/show_worldwide_organisation.html.erb
```diff
@@ -8,7 +8,9 @@
     <% if headers.any? %>
       <nav>
         <div class="content">
-          <h1><%= t('worldwide_organisation.headings.contents', default: 'Contents') %></h1>
+          <h1 <%= "lang=#{t_fallback('worldwide_organisation.headings.contents')}" if t_fallback('worldwide_organisation.headings.contents') %>>
```
injms, line 11: We could use a heading component here - but that shouldn't block this fix.

app/views/worldwide_offices/show.html.erb
```diff
@@ -8,7 +8,7 @@
     <% if headers.any? %>
       <nav>
         <div class="content">
-          <h1>Contents</h1>
+          <h1 lang="en">Contents</h1>
           <ol>
```
injms, line 12: We could use a heading component and a [contents list component](https://components.publishing.service.gov.uk/component-guide/contents_list) here.

injms (approved):

injms:
Actually, is this a duplicate of https://github.com/alphagov/whitehall/pull/5816?

app/views/corporate_information_pages/show_worldwide_organisation.html.erb
owenatgov, line 11: I think the `t_lang` helper does what you need it to here. If it doesn't, I'd recommend putting `t_fallback('worldwide_organisation.headings.contents')` into a variable for brevity.

owenatgov (commented):
Single thing

danacotoran:
Closing this PR as I have opened https://github.com/alphagov/whitehall/pull/5831 instead which should do the same thing but... better.

PR #5830, opened 2020-09-23
Move image position in source

maxgds:
https://trello.com/c/997pjz76/370-misleading-source-order-image-before-heading

What:
This change affects: https://www.gov.uk/government/how-government-works
It moves the images for the PM section and Cabinet Office section and places them below the heading and text that they are related to in the source code. CSS is then used to float them into the original positions at tablet and desktop views, and to switch the order at mobile view. 

Why:
For accessibility reasons it is desirable to have images appearing below the heading they are related to. This offers better support for anyone using a screenreader or linearising the page.

No screenshots as there are no visual changes.

### DIFF
--- a/app/assets/stylesheets/frontend/views/_how-gov-works.scss
+++ b/app/assets/stylesheets/frontend/views/_how-gov-works.scss
@@ -213,13 +213,31 @@
     border-width: 0 0 1px;
   }
 
-  @media (min-width: 640px) and (max-width: 900px) {
-    .prime-minister,
+  .prime-minister,
+  .cabinet {
+    .one-half {
+      &:first-child {
+        float: right;
+      }
+
+    }
+  }
+
+  @include govuk-media-query($from: tablet, $until: desktop) {
     .history {
       .one-half {
         width: $two-thirds;
         &:first-child {
           width: $one-third;
+        }
+      }
+    }
+    .prime-minister {
+      .one-half {
+        width: $one-third;
+
+        &:first-child {
+          width: $two-thirds;
         }
       }
     }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/frontend/views/_how-gov-works.scss
```diff
@@ -213,8 +213,20 @@
...
+  .prime-minister,
+  .cabinet {
+    .one-half {
+      &:first-child {
+        float: right;
+      }
+
+      &:nth-child(2) {
+        float: left;
+      }
```
injms, line 225: I really like that you've used (good old-fashioned) floats here as it works all the way back to older versions of Internet Explorer. 

Since IE 8 doesn't support the `nth-child()` selector and this still works, I wonder if the `nth-child` selector is needed? Removing the `float:left` doesn't cause any change in layout using the latest versions of Firefox, Chrome, or Safari.
```suggestion
      &:first-child {
        float: right;
      }
```

app/assets/stylesheets/frontend/views/_how-gov-works.scss
```diff
@@ -223,6 +235,18 @@
         }
       }
     }
+    .prime-minister {
+      .one-half {
+        width: $one-third;
+        &:nth-child(2) {
+          width: $one-third;
+        }
```
injms, line 243: I think that the `nth-child(2)` selector isn't needed, as it inherits from `.one-half`. Toggling it off in dev tools in FF, Chrome, and Safari doesn't affect the layout.
```suggestion
        width: $one-third;
```

app/assets/stylesheets/frontend/views/_how-gov-works.scss
```diff
@@ -256,6 +280,16 @@
     }
   }
 
+  @include govuk-media-query($from: mobile, $until: tablet) {
+    .prime-minister,
+    .cabinet {
+      display: flex;
+      display: -ms-flexbox;
+      flex-direction: column-reverse;
+      -ms-flex-direction: column-reverse;
```
injms, line 289: ◀️ 👍

maxgds:
@injms Thanks for the review, actioned the suggested changes, you were right! Could you take another look?

injms (approved):
Super! 🚢

PR #5836, opened 2020-10-05
World news story - fix up translations when changing primary locale

ChrisBAshton:
A number of published world news stories have a `primary_locale` of `en`
when they are in fact in a different language. 

When a publisher changes the primary locale of a world news story to the
correct value, we end up in a strange state whereby the draft edition has two
translations: one with the old locale, and one with the new. When the publisher
attempts to publish the draft, Publishing API raises an error:
"Cannot publish an already published edition".

This happens because both translations are sent to publishing API,
and the translation with the old locale has already been published,
so cannot be published again. The simplest solution would be to
remove the redundant translation, so that it is automatically
[removed at time of publish](https://github.com/alphagov/whitehall/blob/277d3216a1009b69cd04b63a1185a7b19e7a4c26/lib/translatable_model.rb#L29-L31).

This PR adds some tests that evaluate the code added in #5832 (the PR designed
to enable the changing of locales for world news stories), and also recreates the issue
we're seeing on production. It then adds the code that removes the non-primary
translations when the primary translation is no longer English, as hinted by the UI:
"Warning: News stories without an English version cannot have other translations."

Finally, it also removes some unused step definitions I came across while writing the test.

Trello: https://trello.com/c/y9H3BLeG/2168-new-editions-of-foreign-language-only-world-news-stories-cant-be-published

### DIFF
--- a/app/models/edition/translatable.rb
+++ b/app/models/edition/translatable.rb
@@ -69,7 +69,7 @@ private
   end
 
   def remove_other_translations_if_primary_locale_no_longer_english
-    if translations.first.locale != :en
+    if translations.count > 1 && translations.first.locale != :en
       translations[1..].each(&:destroy)
     end
   end
--- a/app/models/edition/translatable.rb
+++ b/app/models/edition/translatable.rb
@@ -24,6 +24,10 @@ module Edition::Translatable
       :change_translations_locale_if_primary_locale_changed,
       on: :update,
     )
+    before_validation(
+      :remove_other_translations_if_primary_locale_no_longer_english,
+      on: :update,
+    )
 
     scope :in_default_locale, -> { joins(:translations).where("edition_translations.locale" => I18n.default_locale) }
     validate :locale_is_valid
@@ -61,6 +65,12 @@ private
   def change_translations_locale_if_primary_locale_changed
     if primary_locale_changed? && translations.count == 1
       translations.first.update(locale: primary_locale)
+    end
+  end
+
+  def remove_other_translations_if_primary_locale_no_longer_english
+    if translations.first.locale != :en
+      (translations - [translations.first]).each(&:destroy)
     end
   end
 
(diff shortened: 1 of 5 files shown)


test/unit/edition/translatable_test.rb
```diff
@@ -1,4 +1,4 @@
-require "test_helper"
+require_relative "../../test_helper"#require "test_helper"
```
ChrisBAshton, line 1: This is annoyingly temporarily needed so I can run this test file in isolation:

```bash
govuk-docker-run bundle exec rake mtest ./test/unit/edition/translatable_test.rb
```

MuriloDalRi (approved):

app/models/edition/translatable.rb
```diff
@@ -64,6 +68,12 @@ def change_translations_locale_if_primary_locale_changed
     end
   end
 
+  def remove_other_translations_if_primary_locale_no_longer_english
```
thomasleese, line 71: Does this mean it's not possible for a document to have a "primary" locale of Welsh and a secondary locale of English? Is that a scenario that could exist? (I forget exactly but I suspect not because Whitehall assumes default language is English?)

thomasleese: I see now that the scenario I described shouldn't happen. I think we could add a `primary_locale_changed?` conditional on this to make it slightly safer though. I'll have a go at committing a change that does that.

thomasleese: For some reason I couldn't get `primary_locale_changed?` to work. But I don't think my concerns are valid because:

```
irb(main):007:0> Edition.publicly_visible.where.not(primary_locale: :en).select { |e| e.translations.count > 1 }.count
=> 0
```

thomasleese (approved):
I've been playing around with this in integration and I think it makes sense. There's a lot of confusing bits of code with translations in Whitehall which makes this quite hard to reason about!

ChrisBAshton (commented):
I've checked your code changes and they look good 👍  thanks.
And thanks for reviewing, good to know we're reasonably confident in this change!

PR #5837, opened 2020-10-05
Reflect strong text in page semantics

maxgds:
An info graphic on the How Government Works page has the numbers of ministers called out in bold text:

Screenshot 2020-10-05 at 15 55 30
The semantics of the page don't reflect this, anybody using custom styles will not get recognise the highlighting as captured here: https://trello.com/c/DiiPEdbd/409-ministers-infographic-looks-like-a-heading-but-isnt-one

This changes a span tag to a strong tag to fix this:

Before without styling:
Screenshot 2020-10-05 at 15 53 30

After without styling:
Screenshot 2020-10-05 at 15 55 15


injms (approved):
🧮 👍

PR #5843, opened 2020-10-15
Change customised call to contents-list component

maxgds:
The contents list component is being updated to remove the ability for hiding the title or customising it. It will detect the page language and apply the "lang=en" parameter if the translation isn't available, meaning there is no need to pass these values in from the calling app.

https://trello.com/c/LbVmpaK6/441-inconsistent-naming-of-table-of-contents

There are no visual changes intended as a result of this PR but if it is merged ahead of the gem update then pages such as www.gov.uk/world/organisations/british-embassy-paris/about/social-media-use will temporarily lose the translated title on the contents list component.

### DIFF
--- a/app/views/corporate_information_pages/show_worldwide_organisation.html.erb
+++ b/app/views/corporate_information_pages/show_worldwide_organisation.html.erb
@@ -19,8 +19,6 @@
               }
             end,
             underline_links: true,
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
💪

maxgds:
In draft state to prevent merge ahead of component.

PR #5844, opened 2020-10-15
Use Selenium to run teaspoon tests

kevindew:
This replaces the usage of phantomjs (and potentially GOV.UK's last
usage - yay) with selenium webdriver. The config is mostly just following
the guidelines in https://github.com/jejacks0n/teaspoon/wiki/Using-Selenium-WebDriver

However I did have a very miserable time trying to get these tests to
run in govuk-docker and I shall now recount that tale and how it was
resolved.

So once the tests were running in CI and on my local machine I tried to
run them on GOV.UK docker and received the following output:

```
docker-compose -f [...] run whitehall-lite bundle exec teaspoon
Creating govuk-docker_whitehall-lite_run ... done
Starting the Teaspoon server...
Teaspoon running default suite at http://127.0.0.1:33721/teaspoon/default

Finished in 0.01500 seconds
0 examples, 0 failures
```

This was peculiar as there should have been some tests. I was able to
verify the tests did indeed exist by running the webserver and
navigating to http://whitehall-admin.dev.gov.uk/teaspoon/default where I
could see them passing.

To try work out what was going on I put a byebug in the teaspoon code at
[1]. I then used the method `driver.manage.logs.get("browser")` to see
the console logs, where I saw these alerts:

```
[#<Selenium::WebDriver::LogEntry:0x00007f84eacbd650 @level="SEVERE", @timestamp=1602707509799, @message="http://127.0.0.1:39159/assets/whitehall/admin/views/attachments/_form.self-f222986c7f66eb85b1b4f53fde444f276da1c6adf03325178c964fbd4ffabc29.js?body=1?body=1 - Failed to load resource: net::ERR_INSUFFICIENT_RESOURCES">, #<Selenium::WebDriver::LogEntry:0x00007f84eacbd628 @level="SEVERE", @timestamp=1602707509918, @message="http://127.0.0.1:39159/assets/whitehall/admin/views/edition_workflow/confirm_unpublish.self-2e8e91bb70df0da422d9dd90db0e74e9a247c90abbbda2c46545480d7b51a33e.js?body=1?body=1 - Failed to load resource: net::ERR_INSUFFICIENT_RESOURCES">, #<Selenium::WebDriver::LogEntry:0x00007f84eacbd600 @level="SEVERE", @timestamp=1602707509958, @message="http://127.0.0.1:39159/assets/whitehall/admin/views/editions/_filter_options.self-d8e9280ee88680b58ceb7cac23ca8c4f403aa018aefd7e6cf6c2659d7ba76be9.js?body=1?body=1 - Failed to load resource: net::ERR_INSUFFICIENT_RESOURCES">, #<Selenium::WebDriver::LogEntry:0x00007f84eacbd5d8 @level="SEVERE", @timestamp=1602707510052, @message="http://127.0.0.1:39159/assets/whitehall/admin/views/editions/_form.self-431d309c5420d9913adf265526ed9b6963d4ab0074927d15236273b9cd84e404.js?body=1?body=1 - Failed to load resource: net::ERR_INSUFFICIENT_RESOURCES">, #<Selenium::WebDriver::LogEntry:0x00007f84eacbd5b0 @level="SEVERE", @timestamp=1602707510064, @message="http://127.0.0.1:39159/assets/whitehall/admin/views/editions/diff.self-8b446f7fe3e3fd92fd96e371f14e106054a0f3b996487e09e1aa1e1f2d8234d1.js?body=1?body=1 - Failed to load resource: net::ERR_INSUFFICIENT_RESOURCES">, #<Selenium::WebDriver::LogEntry:0x00007f84eacbd588 @level="SEVERE", @timestamp=1602707510087, @message="http://127.0.0.1:39159/assets/whitehall/admin/views/editions/show.self-5a18be47fbd621db51e19fe6dc65d0dd0e7857fde3ba8f441dc31137e2959839.js?body=1?body=1 - Failed to load resource: net::ERR_INSUFFICIENT_RESOURCES">,...
```

So the outcome seemed to be that enough teaspoon files were loading that
teaspoon could talk to the test runner, but not enough files had loaded
for there to be any actual tests (hence the 0 examples).

Googling seemed to suggest that this was likely caused by Chrome in
docker lacking sufficient memory for the page (as it only has 64MB in
/dev/shm). I then learnt this could be resolved by setting the
--disable-dev-shm flag.

[1]: https://github.com/jejacks0n/teaspoon/blob/e952dbc35fb08528a6802eb24b806e2361b7927e/lib/teaspoon/driver/selenium.rb#L31


1pretz1 (approved):
Woo, no more phantom!

Also nice investigation - it was an interesting read!

benthorner:
Ditto!

I thought I'd try this out locally, but I'm getting a timeout error (after quite some time!). Did you need to do anything special for the tests to run?

kevindew:
> Ditto!
> 
> I thought I'd try this out locally, but I'm getting a timeout error (after quite some time!). Did you need to do anything special for the tests to run?

Oh interesting. I did experience that at an early stage in the process actually on my local machine  - after a long wait do you get connection timeouts to port 9515 (Errno::ECONNREFUSED I think)? I resolved that with a `tmp:clear` and put it down to my tmp directory being in a muddle being shared between govuk-docker and my host machine (I didn't experience it again when using the docker version with a tmp mount). You can try run the web app and navigate to `/teaspoon/default` to see if that takes a holy age to load - if that's the case it's likely the source of the timeout.

kevindew:
> Thanks for suggestion @kevindew. It turned out to be the slowness of CSS compilation in Docker, and running the tests in the browser meant it could finish the compilation to the point where they passed. Then I could run them in the console as well 👍.
> 
> (Clearing `tmp/` was actually a step backwards, as it caused errors due to missing mustache templates.)

Oh that's interesting - I wouldn't have thought the CSS was slow as it shouldn't be using the app one and thus be free of publishing components and govuk-frontend.

Oh bummer about tmp. I was using the `test:javascript` rake task which generates that mustache file for you before the tests - it's super annoying that the mustache template is even a thing.

benthorner:
> Oh that's interesting - I wouldn't have thought the CSS was slow as it shouldn't be using the app one and thus be free of publishing components and govuk-frontend.

That's just my assumption. I think it's a reasonable one, since we use JS tests to look for things like visibility, which need the actual stylesheets in place for the functionality to manifest.

PR #5845, opened 2020-10-16
Add ministers_index page republishing logic

callumknights:
Added logic to republish the ministers index page whenever reshuffle
mode is toggled on or off. This is needed to display the reshuffle
message on the page while a minister reshuffle is happening.

Part of the work to migrate the ministers index page to Collections.

https://trello.com/c/loAlkIp7/2063-8-migrate-government-ministers-to-collections

### DIFF
--- a/app/models/sitewide_setting.rb
+++ b/app/models/sitewide_setting.rb
@@ -4,11 +4,22 @@ class SitewideSetting < ApplicationRecord
   validates :key, uniqueness: true # rubocop:disable Rails/UniqueValidationWithoutIndex
   validates_with SafeHtmlValidator
 
+  after_save :republish_ministers_if_reshuffle
+
   def human_status
     on ? "On" : "Off"
   end
 
   def name
     key.humanize
+  end
+
+  def republish_ministers_if_reshuffle 
+    return unless key == "minister_reshuffle_mode"
+
+    payload = PublishingApi::MinistersIndexPresenter.new
+
+    Services.publishing_api.put_content(payload.content_id, payload.content)
+    Services.publishing_api.publish(payload.content_id, nil, locale: "en")
   end
 end
--- a/test/unit/models/sitewide_setting_test.rb
+++ b/test/unit/models/sitewide_setting_test.rb
@@ -1,0 +1,18 @@
+require "test_helper"
+
+class SitewideSettingTest < ActiveSupport::TestCase
+ 
+  test "toggling reshuffle mode republished the ministers index page" do
+    payload = PublishingApi::MinistersIndexPresenter.new
+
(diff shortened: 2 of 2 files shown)


app/models/sitewide_setting.rb
```diff
@@ -4,11 +4,22 @@ class SitewideSetting < ApplicationRecord
...
     key.humanize
   end
+
+  def republish_ministers_if_reshuffle 
+    return unless key == "minister_reshuffle_mode"
+
+    payload = PublishingApi::MinistersIndexPresenter.new
+
+    Services.publishing_api.put_content(payload.content_id, payload.content)
+    Services.publishing_api.publish(payload.content_id, nil, locale: "en")
```
thomasleese, line 23:
```suggestion
    Services.publishing_api.publish(payload.content_id, locale: "en")
```

I think you can avoid the `nil` [because it's `update_type = nil`](https://github.com/alphagov/gds-api-adapters/blob/8c5224f9c331cf6a22d6065ef7f5de9fcbbbed20/lib/gds_api/publishing_api.rb#L123).

test/unit/models/sitewide_setting_test.rb
```diff
@@ -0,0 +1,12 @@
+require "test_helper"
+
+class SitewideSettingTest < ActiveSupport::TestCase
+ 
+  test "toggling reshuffle mode republished the ministers index page" do
+    payload = PublishingApi::MinistersIndexPresenter.new
+    
+    Services.publishing_api.expects(:publish).with(payload.content_id, nil, locale: "en").once
```
thomasleese, line 8: I'd suggest using the [stub methods here](https://github.com/alphagov/gds-api-adapters/blob/8c5224f9c331cf6a22d6065ef7f5de9fcbbbed20/lib/gds_api/test_helpers/publishing_api.rb#L71-L84) instead of `expects` because it means we're mocking at a "high level", i.e. the HTTP level, so we have more confidence in the code.

They're used in a similar way and here's an example: https://github.com/alphagov/whitehall/blob/da56426b9ae1a8b430185bf3f6da26fc4d0e68af/test/unit/whitehall/publishing_api_test.rb#L25

thomasleese (commented):
This looks good to me, I've just left a few small comments!

app/models/sitewide_setting.rb
callumknights: 👍

test/unit/models/sitewide_setting_test.rb
callumknights: 👍

thomasleese (approved):

PR #5849, opened 2020-10-21
Add constraint for publication_id in statistics announcements

benjamineskola:
Adds a unique index and a model validation to ensure that no two statistics announcements share the same publication

Does not currently ensure the existing records are unique (a rough check suggests there are about 150 affected) so the migration cannot be applied.

https://trello.com/c/he3gFWkI/2163-5-only-allow-one-statistics-announcement-to-be-linked-to-a-publication

### DIFF
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -46,6 +46,7 @@ class StatisticsAnnouncement < ApplicationRecord
               in: PublicationType.statistical.map(&:id),
               message: "must be a statistical type",
             }
+  validates :publication_id, uniqueness: true, if: -> { publication_id.present? }
 
   accepts_nested_attributes_for :current_release_date, reject_if: :persisted?
 
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -22,7 +22,7 @@ class StatisticsAnnouncement < ApplicationRecord
 
   belongs_to :creator, class_name: "User"
   belongs_to :cancelled_by, class_name: "User"
-  belongs_to :publication
+  belongs_to :publication, optional: true
 
   has_one  :current_release_date,
            -> { order("created_at DESC") },
--- a/db/migrate/20201021152850_add_index_to_statistics_announcement_publication_id.rb
+++ b/db/migrate/20201021152850_add_index_to_statistics_announcement_publication_id.rb
@@ -1,8 +1,8 @@
 class AddIndexToStatisticsAnnouncementPublicationId < ActiveRecord::Migration[5.1]
   def change
-    # rubocop:disable Rails/BulkChangeTable
-    remove_index :statistics_announcements, name: "index_statistics_announcements_on_publication_id", column: :publication_id
-    add_index :statistics_announcements, :publication_id, unique: true
-    # rubocop:enable Rails/BulkChangeTable
+    change_table :statistics_announcements, bulk: true do |t|
+      t.remove_index name: "index_statistics_announcements_on_publication_id", column: :publication_id
+      t.index :publication_id, unique: true
+    end
   end
 end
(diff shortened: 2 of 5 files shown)


app/models/statistics_announcement.rb
```diff
@@ -46,6 +46,7 @@ def can_publish_to_publishing_api?
               in: PublicationType.statistical.map(&:id),
               message: "must be a statistical type",
             }
+  validates :publication_id, uniqueness: true, if: -> { publication_id.present? }
```
AlanGabbianelli, line 49: Should it be possible for a `publication_id` to not be present? Or should we enforce its presence?

If `StatisticsAnnouncement belongs_to :publication` and `Publication has_one :statistics_announcement`, why would we allow `publication_id` to be `nil`/`NULL`?

benjamineskola: It seems like it's quite common for statistics announcements not to be assigned to any publication; this is from integration:

```
irb(main):002:0> StatisticsAnnouncement.where(publication_id: nil).count
=> 2889
```

AlanGabbianelli: Ok, then if we're ok with there being `nil` values, I'd suggest being explicit about it by changing `belongs_to :publication` to `belongs_to :publication, optional: true`.

db/migrate/20201021152850_add_index_to_statistics_announcement_publication_id.rb
```diff
@@ -0,0 +1,8 @@
+class AddIndexToStatisticsAnnouncementPublicationId < ActiveRecord::Migration[5.1]
+  def change
+    # rubocop:disable Rails/BulkChangeTable
+    remove_index :statistics_announcements, name: "index_statistics_announcements_on_publication_id", column: :publication_id
+    add_index :statistics_announcements, :publication_id, unique: true
+    # rubocop:enable Rails/BulkChangeTable
```
AlanGabbianelli, line 6: I think rubocop is suggesting this:

```suggestion
    change_table :statistics_announcements do |t|
      t.remove_index name: "index_statistics_announcements_on_publication_id", column: :publication_id
      t.index :publication_id, unique: true
    end
```

benjamineskola: Ah, that's what I was looking for! I was confused by the fact that the normal methods are `remove_index` and `add_index`, in the context of `change_table` there's `remove_index` but no `add_index`... 🤔

lib/tasks/data_hygiene.rake
```diff
@@ -52,4 +52,14 @@ namespace :data_hygiene do
       puts "#{sponsorship.worldwide_organisation.name} updated from FCO to FCDO."
     end
   end
+
+  desc "Ensure there is only one statistics announcement per publication"
+  task ensure_statistics_announcement_unique: :environment do
+    announcements_by_id = StatisticsAnnouncement.where.not(publication_id: nil).group_by(&:publication_id).filter { |_s, p| p.count > 1 }
+    announcements_by_id.each_value do |announcements|
+      announcements.sort_by(&:created_at)[1,].each do |announcement|
```
AlanGabbianelli, line 60: Don't we want all but the first here?
```suggestion
      announcements.sort_by(&:created_at)[1..-1].each do |announcement|
```

```
irb(main):021:0> ary=[:a,:b,:c,:d]
=> [:a, :b, :c, :d]
irb(main):022:0> ary[1,]
=> :b
irb(main):023:0> ary[1..-1]
=> [:b, :c, :d]
```

db/migrate/20201021152850_add_index_to_statistics_announcement_publication_id.rb
```diff
@@ -0,0 +1,8 @@
+class AddIndexToStatisticsAnnouncementPublicationId < ActiveRecord::Migration[5.1]
+  def change
+    change_table :statistics_announcements do |t|
```
AlanGabbianelli, line 3: Sorry, I forgot this:

```suggestion
    change_table :statistics_announcements, bulk: true do |t|
```

AlanGabbianelli (approved):

AlanGabbianelli (changes requested):
As discussed, we should deploy the rake on its own and run it before merging this.

nicholsj:
Closing because we're not going to proceed with this

PR #5854, opened 2020-10-23
Github Action to combine Dependabot PRs

NeilvB:
GOV.UK is investigating how to improve management of frequent Dependabot PR updates.

This PR adds a custom Github Actions workflow to group commits together based on a common prefix.

It was taken from https://github.com/hrvey/combine-prs-workflow.

This approach is being trialled in a few repositories with a backlog of dependabot updates.


thomasleese:
Unfortunately we've noticed a problem with this because we don't get have approval to use GitHub Actions with secrets, closing for now.

PR #5868, opened 2020-11-02
Change role appointment ordering

callumknights:
Change ordering of role appointments in the admin interface in whitehall,
so when tagging a role appointment to a new publication, the appointments
are presented in descending order. This is that so when a user searches a
minister name the appointments that minister currently holds will appear
first, then previous appointments in descending order by start date.

https://trello.com/c/dFGjneqv/2190-change-ordering-when-selecting-a-minister-in-whitehall-content

image


thomasleese (commented):
I think it would be good to squash the commits together, but the code itself looks good to me (except it looks like the tests don't quite pass yet)!

thomasleese (approved):

PR #5877, opened 2020-11-10
Don't update attachment metadata for already-deleted attachments

ChrisBAshton:
We're seeing lots of `AssetManager::ServiceHelper::AssetNotFound`
errors in Sentry (here's an [example][]), originating from
`AssetManagerAttachmentMetadataWorker`. Looking at the assets in
question, their respective 'attachment' always seems to be
`deleted: true`, which explains why the asset cannot be found.
It also begs the question: why are we trying to apply any workers
to them in the first place?

When an attachment is deleted, it shouldn't be touched again.
Once this commit is merged, we expect to see these kinds of errors
reduce by half. We still have to apply a similar fix to errors
originating from `AssetManagerAttachmentRedirectUrlUpdateWorker`,
which account for the other half.

Trello: https://trello.com/c/RethzUSW/2088-5-fix-assetmanagerservicehelperassetnotfound-errors-on-staging-integration

[example]: https://sentry.io/organizations/govuk/issues/1668643818/events/f1bf7ff83ede4822b9a781459d749760/?environment=staging&project=202259&query=error.type%3AAssetManager%3A%3AServiceHelper%3A%3AAssetNotFound+transaction%3ASidekiq%2FAssetManagerAttachmentMetadataWorker&statsPeriod=90d

### DIFF
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -205,6 +205,6 @@ private
   end
 
   def attachment_updater
-    ServiceListeners::AttachmentUpdater.call(attachable: attachable.reload)
+    ServiceListeners::AttachmentUpdater.update_all_attachment_data_for(attachable.reload)
   end
 end
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -81,7 +81,7 @@ class AttachmentData < ApplicationRecord
 
   def uploaded_to_asset_manager!
     update!(uploaded_to_asset_manager_at: Time.zone.now)
-    ServiceListeners::AttachmentUpdater.call(attachment_data: self)
+    ServiceListeners::AttachmentUpdater.update_attachment_data(self)
   end
 
   def uploaded_to_asset_manager?
--- a/app/services/service_listeners/attachment_updater.rb
+++ b/app/services/service_listeners/attachment_updater.rb
@@ -1,25 +1,19 @@
 module ServiceListeners
   class AttachmentUpdater
     extend LockedDocumentConcern
-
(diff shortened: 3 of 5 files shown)


ChrisBAshton:
I deployed this PR to both Integration & Staging around 5:30pm last night, expecting to see no more `AssetManagerAttachmentMetadataWorker` errors.

When I checked Sentry, I did see an error, that happened at 2:47am:
https://sentry.io/organizations/govuk/issues/1668643818/events/ec3ead98958a4223ae9f5c8554cb556d/?project=202259

However, the job has a `created_at` value of `1605000605`, which corresponds to 9:30 on 10th November. It has a `retried_at` value of `1605041859`, which corresponds to 8:57pm. Given that the error is for 2:47am, and there is a retry_count of 12, it retried on average every half hour or so (though in practice uses [exponential backoff for retry times](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry)).

Point being, the code change I've made should prevent _new_ invocations of the `AssetManagerAttachmentMetadataWorker`, but it doesn't stop the old ones from retrying.

I've now run `Sidekiq::RetrySet.new.filter { |job| job.klass == "AssetManagerAttachmentMetadataWorker" }.map(&:delete)` on Integration and Staging, to [clear the retry queue](https://github.com/mperham/sidekiq/wiki/API#retries). I'm now monitoring for more errors - I expect to see _no_ `AssetManagerAttachmentMetadataWorker` errors for the rest of the day.

When I'm confident this has worked I'll open the PR up for technical review. I also need to work on a similar fix for the `AssetManagerAttachmentRedirectUrlUpdateWorker`.

ChrisBAshton:
Superseded by #5882.

PR #5885, opened 2020-11-16
Update attachment heading level to h3

danacotoran:
Document headings on [this page](https://www.gov.uk/government/publications/our-plan-to-rebuild-the-uk-governments-covid-19-recovery-strategy
) all have the same heading level as their parent section.

Screenshot 2020-11-16 at 17 30 37

Although the page is rendered by `government-frontend`, the code behind these attachments lives in `whitehall` and is consumed by `government-frontend` via the content API as [legacy_pre_rendered_documents](https://github.com/alphagov/government-frontend/blob/e5dd4c8bf39fb840d7923bf8abdaa8e58785d7b5/app/views/content_items/_attachments.html.erb#L11). 😄 

This updates the code in `whitehall`  to use `h2` instead of `h3` for attachment headings.
Next step would probably be to replace this old code with the [attachment component](https://components.publishing.service.gov.uk/component-guide/attachment) – the reason I didn't do this as part of this work is that the attachment component does not currently have as many features as the Whitehall prerendered stuff. The component would have to be extended before it can be used for Whitehall attachment rendering.

https://trello.com/c/KzyicaBh


app/views/documents/_attachment.html.erb
```diff
@@ -21,7 +21,7 @@
     <% end %>
   </div>
   <div class="attachment-details">
-    <h2 class="title"><%= link_to_unless previewable?(attachment), attachment.title, attachment.url(preview: preview), title_link_options %></h2>
+    <h3 class="title govuk-!-font-size-27 govuk-!-font-weight-regular"><%= link_to_unless previewable?(attachment), attachment.title, attachment.url(preview: preview), title_link_options %></h3>
```
chris-gds, line 24: Taking at look a the temp [integration link](https://www.integration.publishing.service.gov.uk/government/publications/our-plan-to-rebuild-the-uk-governments-covid-19-recovery-strategy) this is as expected. ✅

---

🤔  Wondering if this diverges from the Design System a little and should just use default h3 styles. However out of scope

image

chris-gds (approved):

app/views/documents/_attachment.html.erb
danacotoran: That's a good point, imo it even looks nicer with the default styles but it's not within the scope of this piece of work to change it. 
We do have some exceptions from the default styling but I do wonder why this was one as I don't really see a benefit in it, in this particular case.

PR #5890, opened 2020-11-20
Switch from Standard to Standardx for JS linting

kevindew:
Standard 16 introduced a rule that disallowed the use of `var`, instead
preferring `let` or `const` [1]. This conflicts with the GOV.UK approach
where we tend to not embrace features that we know will break old
browsers even if they're not necessarily supported [2], disallowing var
will mean that < IE 11 will be unable to run any of the JS.

In order to customise standard rules this project has switched to using
standardx [3] which allows us to disallow rules. I've used this so we
can disallow the 'no-var' rule.

[1]: https://github.com/standard/standard/issues/633
[2]: https://github.com/alphagov/govuk_publishing_components/pull/1611#discussion_r456493365
[3]: https://github.com/standard/standardx

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,7 +40,7 @@ updates:
       interval: daily
     allow:
       # Framework packages
-      - dependency-name: standard
+      - dependency-name: standardx
         dependency-type: direct
 
   # Ruby needs to be upgraded manually in multiple places, so cannot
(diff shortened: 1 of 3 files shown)


ChrisBAshton (approved):
standardx sounds like a good solution to the problem :+1: I was hesitant as I hadn’t heard of standardx, but it's an officially supported alternative under the [standard](https://github.com/standard) org, and switching to it will benefit a large number of GOV.UK users. It should also be quite simple to switch back to Standard if/when we do eventually drop JS support on IE.

PR #5894, opened 2020-11-24
DO NOT MERGE Brexit current state notice

hannako:
## What

Allow publishers to add a Brexit current state call out box to relevant pages on gov.uk.

Screenshot 2020-11-30 at 12 40 12

## Context

- We currently allow publishers to add a call out box - referred to in the product domain as the "post transition call out box." An example can be seen on [this page](https://www.gov.uk/guidance/driving-in-the-eu-from-1-january-2021). In the code this is called a `brexit-no-deal-content-notice`.

- We want publishers to be able to add an alternative call out box, which gives information on the current state of the guidance. Referred to as a "current state notice"

##  Roll out

- The form will add the new fields to the payload for publishing api, which will be present in the content item as:
```
"brexit_current_state_notice": [
  {
    "href": "/interesting-stuff",
    "title": "Interesting Stuff"
  }
 ]
```
- We should deploy the [changes to government frontend](https://github.com/alphagov/government-frontend/pull/1924) that will pick up this new field and display the current state content notice before we allow publishers to add it. 

## TO DO

- Add the js to show/hide the form based on the radio button selection.
- Content to be finalised
- Update content-schema so the new field is allowed to be included in the details hash

---

https://trello.com/c/IwXzb5FM/589-design-whitehall-publisher-functionality-to-generate-current-state-notice

### DIFF
--- a/app/presenters/publishing_api/payload_builder/brexit_content_notices.rb
+++ b/app/presenters/publishing_api/payload_builder/brexit_content_notices.rb
@@ -14,16 +14,16 @@ module PublishingApi
       def call
         return {} if no_brexit_content_notices?
 
-        show_no_deal_notice? ? brexit_no_deal_notice_payload : brexit_current_state_notice_payload
+        show_no_deal_notice? ? no_deal_notice_payload : current_state_notice_payload
       end
 
     private
 
-      def brexit_no_deal_notice_payload
-        { brexit_no_deal_notice: links }
+      def no_deal_notice_payload
+        { brexit_no_deal_notice: no_deal_links }
       end
 
-      def brexit_current_state_notice_payload
+      def current_state_notice_payload
         { brexit_current_state_notice: current_state_links }
       end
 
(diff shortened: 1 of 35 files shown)


sihugh:
Ugh, conflicts, sorry

alex-ju:
I just sneaked in a recent design update (see preview below)
Screenshot 2020-12-03 at 11 29 57

hannako:
> Ugh, conflicts, sorry

Ah it's just because of this https://github.com/alphagov/whitehall/pull/5901/files I'll rebase.

app/presenters/publishing_api/payload_builder/brexit_content_notices.rb
```diff
@@ -14,49 +14,46 @@ def initialize(item)
...
+      def to_hash(link)
+        {
+          title: link.title,
+          href: href(link),
+        }
+      end
+
+      def non_blank_links(links)
+        links.reject do |link|
           link.title.blank? && link.url.blank?
```
sihugh, line 57: Wondering if this should be an `||` ? If either the title or url are blank it's probably bad?

sihugh (approved):
I appreciate building this so that it's easily removable 👍 . Can we give this a good try out on integration?

hannako:
The current state notice is no longer required.

PR #5895, opened 2020-11-24
Update email signup with new design

benthorner:
https://trello.com/c/mk5M2aKD/619-general-design-and-stylistic-updates-to-signup-workflow

Please see the commits for more details.

## Before

Screenshot 2020-11-24 at 17 13 50
Screenshot 2020-11-24 at 17 14 20

## After

Screenshot 2020-11-24 at 17 13 05
Screenshot 2020-11-24 at 17 13 16


app/views/email_signups/new.html.erb
```diff
@@ -9,10 +9,6 @@
 
 <div class="block">
   <div class="inner-block">
-    <% if @email_signup.valid? %>
-      <%= render 'form', email_signup: @email_signup %>
-    <% else %>
-      <p>Sorry, we could not find a valid email alerts feed for that.</p>
-    <% end %>
```
kevindew, line 16: Wow - how strange.

app/views/email_signups/new.html.erb
```diff
@@ -1,14 +1,29 @@
...
 
 <div class="block">
   <div class="inner-block">
-    <%= render 'form', email_signup: @email_signup %>
+    <p class="govuk-body">
+      <%= t("email_signups.new.description") %>
+    </p>
+
+    <p class="govuk-body govuk-!-margin-bottom-8">
+      <strong style="font-weight: bold"><%= @email_signup.name %></strong>
```
kevindew, line 17: Similar concern to https://github.com/alphagov/finder-frontend/pull/2315#pullrequestreview-537788983 regarding the inline style.

app/views/email_signups/new.html.erb
```diff
@@ -1,14 +1,29 @@
...
 <header class="block email-header">
   <div class="inner-block">
-    <h1>Email alert subscription</h1>
+    <h1 class="govuk-heading-l"><%= t("email_signups.new.title") %></h1>
   </div>
 </header>
 
 <div class="block">
   <div class="inner-block">
-    <%= render 'form', email_signup: @email_signup %>
```
kevindew: Can this partial be deleted?

config/locales/ar.yml
```diff
@@ -496,6 +496,10 @@ ar:
     world_locations:
       all: كافة المواقع
       label: مواقع في العالم
+  email_signups:
+    new:
+      description:
+      title:
```
kevindew, line 502: Wow, this is a beast of a step.

kevindew (commented):
Looks good to me, just a couple of minor things.

The page looks a million times nicer with this new design.

app/views/email_signups/new.html.erb
benthorner: Yup - same problem too! I've updated the commit message to explain.

app/views/email_signups/new.html.erb
benthorner: Oops - yes, I've done that in the second commit 👍.

benthorner:
Thanks for looking at @kevindew - I agree it's less of a visual relic now! I think I've addressed your comments too, if you can take another look?

kevindew (approved):

app/views/email_signups/new.html.erb
gavinwye: And the same solution ;p

config/locales/ar.yml
gavinwye: The ordering here is different from the other files. It may make more sense for people in the future if they are consistent.

gavinwye:
I'm confused by the after 404 is a user ever going to see that?

config/locales/ar.yml
benthorner: I think that's a minor concern - I doubt it will cause an issue, because we don't rely on the ordering for anything. Whitehall has significant issues with its use of locales, but we're not going to fix any of them here, unfortunately.

benthorner:
> I'm confused by the after 404 is a user ever going to see that?

@gavinwye no, it's OK - they'll see the GOV.UK standard 404 page. In general, we should avoid custom 404 pages unless it's a predictable scenario where we could actually help the user. In this case, we've got no evidence of that, and the previous page wasn't doing anything to help the user, so there's no harm in removing the snowflake.

gavinwye:
> @gavinwye no, it's OK - they'll see the GOV.UK standard 404 page. In general, we should avoid custom 404 pages unless it's a predictable scenario where we could actually help the user. In this case, we've got no evidence of that, and the previous page wasn't doing anything to help the user, so there's no harm in removing the snowflake.

Okay, I didn't understand that was what was happening here. 

All looks good to me.

PR #5897, opened 2020-11-26
Catch errors sending via Notify and warn users they may occur

benjamineskola:
https://trello.com/c/9Xfhhj2B/2222-investigation-fix-notificationsclientbadrequesterror-error-timebox-2-days

### DIFF
--- a/app/views/admin/editions/confirm_export.html.erb
+++ b/app/views/admin/editions/confirm_export.html.erb
@@ -7,6 +7,9 @@
     <ul>
       <li><%= @filter.page_title %></li>
     </ul>
+    <% if Whitehall.integration_or_staging? %>
+      <p class="warning">Warning: in staging and integration environments, if you are not on the permitted recipients list for GOV.UK Notify, this email will not be exported.</p>
+    <% end %>
     <%= form_tag "#{export_admin_editions_path}?#{@filter.options_for_export.to_param}" %>
       <%= submit_tag "Export as CSV", class: 'btn btn-primary' %>
 
--- a/app/workers/document_list_export_worker.rb
+++ b/app/workers/document_list_export_worker.rb
@@ -15,6 +15,13 @@ private
 
   def send_mail(csv_filename, user, filter)
     MailNotifications.document_list(csv_filename, user.email, filter.page_title).deliver_now
+  rescue Notifications::Client::BadRequestError => e
+    if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
+      # in production we care about all errors
+      # in staging and integration the team-only error is unrecoverable when running asynchronously
+      # (team-only error is unrecoverable in production too, but almost certainly impossible)
+      raise
+    end
   end
 
   def upload_file(csv, type)
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,19 +1,19 @@
+$LOAD_PATH << "."
+require "active_support"
+require "lib/whitehall"
+
 # default cron env is "/usr/bin:/bin" which is not sufficient as govuk_env is in /usr/local/bin
 env :PATH, "/usr/local/bin:/usr/bin:/bin"
 
 # We need Rake to use our own environment
 job_type :rake, "cd :path && govuk_setenv whitehall bundle exec rake :task --silent :output"
 
-def integration_or_staging?
-  ENV.fetch("GOVUK_WEBSITE_ROOT") =~ /integration|staging/
-end
-
 every :day, at: ["3am", "12:45pm"], roles: [:backend] do
   rake "export:mappings"
 end
 
 def search_index_consultations_cron_rule
-  if integration_or_staging?
+  if Whitehall.integration_or_staging?
     # Don't run near midnight, as this is when the data sync will
     # likely happen, and the task will error
     "0 2-22 * * *"
@@ -27,7 +27,7 @@ every search_index_consultations_cron_rule, roles: [:backend] do
 end
 
 def taxonomy_cron_rules
-  if integration_or_staging?
+  if Whitehall.integration_or_staging?
     # at every 10th minute past the hour between 7am and 8pm
     "*/10 7-20 * * *"
   else
--- a/lib/whitehall.rb
+++ b/lib/whitehall.rb
@@ -210,4 +210,8 @@ module Whitehall
     Rails.root + "config" + "whitehall_secrets.yml"
   end
   private_class_method :secrets_path
+
+  def self.integration_or_staging?
+    (ENV.fetch("GOVUK_WEBSITE_ROOT", "") =~ /integration|staging/).present?
+  end
 end
--- a/test/unit/workers/document_list_export_worker_test.rb
+++ b/test/unit/workers/document_list_export_worker_test.rb
@@ -52,4 +52,49 @@ class DocumentListExportWorkerTest < ActiveSupport::TestCase
     MailNotifications.expects(:document_list).with(regexp_matches(%r{^https://whitehall-admin\.test\.gov\.uk/export/documents/[a-f0-9-]{36}$}), @user.email, title).returns(stub(deliver_now: nil))
     @worker.perform({ "state" => "draft" }, @user.id)
   end
+
+  test "does not raise an error if an email cannot be sent via notify" do
+    raises_exception = lambda { |_url, _address, _title|
+      response = MiniTest::Mock.new
+      ENV["SENTRY_CURRENT_ENV"] = "integration"
+      response.expect :code, 400
+      response.expect :body, "Can't send to this recipient using a team-only API key"
+      raise Notifications::Client::BadRequestError, response
+    }
+
+    MailNotifications.stub(:document_list, raises_exception) do
+      ActionMailer::Base.deliveries.clear
+
+      csv = "this|is|my|csv"
+      title = "Everyone's editions"
+      @worker.stubs(:create_filter).returns(stub(page_title: title))
+      @worker.stubs(generate_csv: csv)
+      assert_nothing_raised do
+        @worker.perform({ "state" => "draft" }, @user.id)
+      end
+    end
+  end
+
+  test "raises an error if an email cannot be sent via notify in production" do
+    raises_exception = lambda { |_url, _address, _title|
+      response = MiniTest::Mock.new
+      ENV["SENTRY_CURRENT_ENV"] = "production"
+      response.expect :code, 400
+      response.expect :body, "Can't send to this recipient using a team-only API key"
+      raise Notifications::Client::BadRequestError, response
+    }
+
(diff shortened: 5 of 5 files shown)


app/views/admin/editions/confirm_export.html.erb
```diff
@@ -7,6 +7,9 @@
     <ul>
       <li><%= @filter.page_title %></li>
     </ul>
+    <% if ENV.fetch("GOVUK_WEBSITE_ROOT") =~ /integration|staging/ %>
```
benjamineskola, line 10: This is based on logic in `config/schedule.rb`, but I wonder if there might be a better way to do it?

app/workers/document_list_export_worker.rb
```diff
@@ -15,6 +15,8 @@ def perform(filter_options, user_id)
 
   def send_mail(csv_filename, user, filter)
     MailNotifications.document_list(csv_filename, user.email, filter.page_title).deliver_now
+  rescue Notifications::Client::BadRequestError
```
thomasleese, line 18: I wonder if we could check if the error is the not in team error (regex match on the exception message maybe?) and if not send it to Sentry. Because if it was something else (incorrect API key for example) in production, it might be difficult to debug without these exeptions.

ChrisBAshton: I wonder if you could look at the [`SENTRY_CURRENT_ENV` ENV variable](https://github.com/alphagov/govuk_app_config#automatic-error-reporting) from within the worker, and only call `GovukError.notify` if it's not 'integration' or 'staging'? (I deliberately suggest that rather than "only if it's 'production'" in case the variable goes missing one day and we lose visibility on production errors).

thomasleese: Yeah, that sounds like a good idea.

I also think that actually in production we should simply `raise` rather than using `GovukError.notify` directly so we retain Sidekiq's retry behaviour if it's just a transient issue.

benjamineskola: Just to clarify the intended logic here: do you mean if it's not the not-in-team error _and_ it's production, or do you mean if it's not the not-in-team error _or_ it's production?

i.e., we don't care about not-in-team errors in any environment; we care about all errors in production; but do we care about other bad-request errors in integration/staging?

thomasleese: My thinking was:

- In production: we care about _all_ errors, including the not-in-team ones (although why they would appear in production... I'm not sure... only a misconfiguration I imagine, which would still be a useful thing to know about)
- In the other environments: we don't care about the not-in-team errors, but we do care about everything else

app/views/admin/editions/confirm_export.html.erb
thomasleese: I think this method is okay, but wonder if we can put it somewhere where it can shared across the various places. I wonder if we could put it here for example: https://github.com/alphagov/whitehall/blob/master/lib/whitehall.rb

And then it's `Whitehall.integration_or_staging?` I guess.

ChrisBAshton: You could then write a test for it too (although I appreciate it's a very simple regex).

app/workers/document_list_export_worker.rb
ChrisBAshton: I'd ideally remove this change from the first commit, making it just about warning the user. Then in your next commit you have the exception handling change.

app/workers/document_list_export_worker.rb
```diff
@@ -15,8 +15,13 @@ def perform(filter_options, user_id)
 
   def send_mail(csv_filename, user, filter)
     MailNotifications.document_list(csv_filename, user.email, filter.page_title).deliver_now
-  rescue Notifications::Client::BadRequestError
-    # unrecoverable when running asynchronously
+  rescue Notifications::Client::BadRequestError => e
+    if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
```
ChrisBAshton, line 19: Could you write some corresponding tests in [DocumentListExportWorkerTest](https://github.com/alphagov/whitehall/blob/4e2df36b2ba62a11986fa17fa5ce9983d8a25bc0/test%2Funit%2Fworkers%2Fdocument_list_export_worker_test.rb)?

app/workers/document_list_export_worker.rb
```diff
@@ -15,8 +15,13 @@ def perform(filter_options, user_id)
...
+      # in production we care about all errors
+      # in staging and integration the team-only error is unrecoverable when running asynchronously
+      # (team-only error is unrecoverable in production too, but almost certainly impossible)
+      raise
```
ChrisBAshton, line 23: We should probably re-raise the exception we caught, rather than raise with no arguments.

```suggestion
raise e
```

thomasleese: Actually [if you call `raise` without any arguments it will re-raise the last error](https://stackoverflow.com/a/23771227), which I think can help in avoiding an extra unnecessary line in the stacktrace.

ChrisBAshton: Oh, cool! Apologies, I should have checked first.

test/unit/workers/document_list_export_worker_test.rb
```diff
@@ -52,4 +52,24 @@ class DocumentListExportWorkerTest < ActiveSupport::TestCase
...
   end
+
+  test "does not raise an error if an email cannot be sent via notify" do
+    raises_exception = lambda { |_url, _address, _title|
+      response = MiniTest::Mock.new
+      ENV["SENTRY_CURRENT_ENV"] = "integration"
+      response.expect :code, 400
+      response.expect :body, "Can't send to this recipient using a team-only API key"
+      raise Notifications::Client::BadRequestError, response
+    }
```
ChrisBAshton, line 63: This test took me a few reads 🤔  so it looks like you're stubbing that `MailNotifications.document_list` raises an exception, and you're asserting that the status code and body associated with the response are what you expect.

But it looks like we're missing an assertion that the error doesn't get re-raised... unless I'm missing something?

For comparison, we should also have a test where we set `SENTRY_CURRENT_ENV` to "production".

ChrisBAshton (commented):
This is shaping up nicely, Ben - well done for wrangling with the Whitehall internals like this.

One comment below, and also a suggestion that commits 1 and 4 should be combined, and commits 2, 3 and 5 should be combined too (so two commits in total). These commits represent the two discrete pieces of functionality you're adding to the system.

config/schedule.rb
```diff
@@ -4,16 +4,12 @@
...
-  ENV.fetch("GOVUK_WEBSITE_ROOT") =~ /integration|staging/
-end
-
 every :day, at: ["3am", "12:45pm"], roles: [:backend] do
   rake "export:mappings"
 end
 
 def search_index_consultations_cron_rule
-  if integration_or_staging?
+  if Whitehall.integration_or_staging?
```
thomasleese, line 12: We should double check this on integration by deploying the branch, but I have a feeling that the `Whitehall` module might not be available when this file is set up on deploy...

test/unit/workers/document_list_export_worker_test.rb
benjamineskola: I had to copy it from the previous test I wrote (in test/functional/admin/fact_check_requests_controller_test.rb) because I no longer remembered how testing an exception in minitest works.

I _think_ that because the test finishes at all, it indicates that the error is not raised (or is caught and handled) — changing `SENTRY_CURRENT_ENV` causes the test to fail, for example. But it could probably be written in a way that explicitly asserts that no error is raised (I seem to recall a discussion about that recently).

I'll add another with the env as "production" where it asserts that an exception _is_ raised.

ChrisBAshton:
It's shaping up nicely, Ben 👍  Not sure where my original comment went, it seems to have disappeared, but the point still stands - could you squash up the commit history a little?

Looks to me like you can combine commits 1 and 4, and also combine commits 2, 3, 5, 6 and 7 (so there should be two commits in total). These would each represent the self-contained, logical code changes that have been made.

benjamineskola:
> It's shaping up nicely, Ben 👍 Not sure where my original comment went, it seems to have disappeared, but the point still stands - could you squash up the commit history a little?

Ah yeah, I'd intended to do that last to make reviewing the additions easier and then almost forgot.

ChrisBAshton (approved):
This LGTM now. Nice one 👍

config/schedule.rb
benjamineskola:
```
NameError: uninitialized constant #<Class:#<Whenever::JobList:0x00007f31ce8a6608>>::Whitehall
```

Seems like you're right. Will importing it work or is that not possible from here?

lib/whitehall.rb
```diff
@@ -212,6 +212,6 @@ def self.secrets_path
   private_class_method :secrets_path
 
   def self.integration_or_staging?
-    ENV.fetch("GOVUK_WEBSITE_ROOT") =~ /integration|staging/
+    ENV.fetch("GOVUK_WEBSITE_ROOT", "") =~ /integration|staging/
```
ChrisBAshton, line 215: I'm guessing this defaults to "" if `GOVUK_WEBSITE_ROOT` isn't defined.

That's probably a good idea 👍   just wondering if you added this because it fell over when deployed? If so, do we know that it actually works on integration?

ChrisBAshton (approved):
LGTM 👍  just one question below, which may be worth addressing in the commit message itself.

lib/whitehall.rb
benjamineskola: I _think_ it was actually that it falls over when testing it locally. There were no issues specifically with this when I deployed, it was just that I couldn't test the actual change without defining this.

benjamineskola:
```
irb(main):001:0> Whitehall.integration_or_staging?
=> 12
```

An integer was not what I was expecting here, but an integer is truthy, I suppose.

Just to triple-check:

```
irb(main):003:0> ENV.fetch("GOVUK_WEBSITE_ROOT", "")
=> "https://www.integration.publishing.service.gov.uk"
…
irb(main):008:0>  ENV.fetch("NOT_A_REAL_VARIABLE", "") =~ /integration|staging/
=> nil
```

ChrisBAshton: Ha, I was surprised by that too! It really ought to return a boolean, as that would follow the `method_name_ending_with_question_mark?` convention. We could do so with:

```
    !(ENV.fetch("GOVUK_WEBSITE_ROOT", "") =~ /integration|staging/).nil?
```

benjamineskola: Yeah, presumably it's always going to be interpreted as a boolean anyway but it would be nicer to do so explicitly.

ChrisBAshton: `.present?` is nicer 👍

PR #5899, opened 2020-11-30
Add rake tasks for editions and attachments

danacotoran:
Recent fixes around Whitehall attachments and publications containing
attachments have meant that there has been an increasing need for an
easier way to republish HTML attachments and/or editions containing
attachments.


thomasleese (approved):
Looks good to me!

PR #5900, opened 2020-12-01
Fix zoom overflow

maxgds:
### What & why:

https://trello.com/c/sH0fdbRR/547-text-overlapping-with-bigger-font-size-on-how-government-works
Affected page: https://www.gov.uk/government/how-government-works

The big numbers affected in this commit currently overflow into the surrounding
region when a user increases text size but doesn't otherwise zoom. This corrects
that by reducing the sizes of the big numbers as they run out of space, based on
em sizes. This will not make them illegible as they are still enormous in
comparison to the copy text.

Firefox is the easiest browser to test this in, go to View->zoom -> zoom text only and then adjust the zoom level as required.

#### 150%
Screenshot 2020-12-01 at 09 53 43

#### 170%
Screenshot 2020-12-01 at 09 54 01

#### 200%
Screenshot 2020-12-01 at 09 54 14

#### 150%
Screenshot 2020-12-01 at 09 58 13

#### 170%
Screenshot 2020-12-01 at 09 58 27

#### 200%
Screenshot 2020-12-01 at 09 58 38

### DIFF
--- a/app/assets/stylesheets/frontend/views/_how-gov-works.scss
+++ b/app/assets/stylesheets/frontend/views/_how-gov-works.scss
@@ -73,6 +73,10 @@
       line-height: 1;
     }
 
+    @media (max-width: 65em) {
+      font-size: .75em;
+    }
+
     @include govuk-media-query($from: tablet, $until: 815px) {
       &--small .feature-value__count {
         font-size: 3.85em;


chris-gds:
I’m wondering - if you have a user that can’t see this 300+ well and they try to zoom in (150 > 200 > 300), does this solution prevent it this text getting bigger for them?

maxgds:
See the screenshots and the explanation. Yes it prevents it getting as big as it would otherwise, that's the point, there isn't room for it. BUT it's still huge in comparison to the rest of the text and if the user can read that then yes, they can read the text that says "300+".  The size of the 300 at the start is bigger than the browser supports for the body text and it will still grow, just not as much as previously, when zoomed.

chris-gds:
Hmm I'm not explaining myself well - the solution makes sense for the context of the overlap, this resolves this issue well.  

I was thinking about the case of a user that can't see 300 at 100% very well and then tries to zoom into to 300%  however it sounds like this is not valid / already been factored in

chris-gds (approved):

PR #5904, opened 2020-12-02
Add rake task to reslug a `TopicalEvent`

AlanGabbianelli:
We had a request to reslug a topical_event page in 2nd line. The
functionality wasn't present so this adds it, following the pattern of
similar rake tasks to reslug Person and Role.


edwardkerry (approved):
This looks good to me, and clearly tested

PR #5905, opened 2020-12-02
Add task to republish withdrawn HTML attachments

chao-xian:
This task ensures we handle the state of the documents when we
republish.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -240,10 +240,14 @@ namespace :publishing_api do
     end
   end
 
-  desc "Republish all HTML attachments to the Publishing API"
+  desc "Republish all documents with HTML attachments to the Publishing API"
   task republish_html_attachments: :environment do
-    HtmlAttachment.where(attachable_type: "Edition", attachable_id: Edition.publicly_visible.select("id")).each do |a|
-      Whitehall::PublishingApi.bulk_republish_async(a)
+    document_ids = Edition
+      .publicly_visible
+      .where(id: HtmlAttachment.where(attachable_type: "Edition").select(:attachable_id))
+      .pluck(:document_id)
+    document_ids.each do |document_id|
+      PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
     end
   end
 


lib/tasks/publishing_api.rake
```diff
@@ -241,9 +241,14 @@ namespace :publishing_api do
   end
 
   desc "Republish all HTML attachments to the Publishing API"
-  task republish_html_attachments: :environment do
-    HtmlAttachment.where(attachable_type: "Edition", attachable_id: Edition.publicly_visible.select("id")).each do |a|
-      Whitehall::PublishingApi.bulk_republish_async(a)
+  task :republish_html_attachments, [:state] => :environment do |_, args|
+    args.with_defaults(state: "withdrawn")
```
kevindew, line 245: Oh cool, I didn't know you could do this - looks super handy.

Should there be a default state for this task though? the task wouldn't imply that you were dealing with withdrawn as the default.

lib/tasks/publishing_api.rake
```diff
@@ -241,9 +241,14 @@ namespace :publishing_api do
...
+    publications = Publications.where(state: args[:state])
+    publications.find_each do |publication|
+      if !publication.html_attachments.nil? || publication.html_attachments.count.positive?
+        document = Document.find_by(id: publication.document_id)
```
kevindew, line 249: You could do `publication.document` here. Thought I don't think you need to do this line at all and can instead change the below line to have `publication.document_id`.

lib/tasks/publishing_api.rake
```diff
@@ -241,9 +241,14 @@ namespace :publishing_api do
...
 
   desc "Republish all HTML attachments to the Publishing API"
-  task republish_html_attachments: :environment do
-    HtmlAttachment.where(attachable_type: "Edition", attachable_id: Edition.publicly_visible.select("id")).each do |a|
-      Whitehall::PublishingApi.bulk_republish_async(a)
+  task :republish_html_attachments, [:state] => :environment do |_, args|
+    args.with_defaults(state: "withdrawn")
+    publications = Publications.where(state: args[:state])
+    publications.find_each do |publication|
+      if !publication.html_attachments.nil? || publication.html_attachments.count.positive?
```
kevindew, line 248: I imagine you can do this in one conditional? `publication.html_attachments&.count&.positive?`

Though I do wonder if this conditional could be done as part of the Publications query to only pull out publications with HTML attachments.

lib/tasks/publishing_api.rake
```diff
@@ -241,9 +241,14 @@ namespace :publishing_api do
   end
 
   desc "Republish all HTML attachments to the Publishing API"
```
kevindew, line 243: This description should be updated given it would republish the whole document.

lib/tasks/publishing_api.rake
```diff
@@ -241,9 +241,14 @@ namespace :publishing_api do
...
-  task republish_html_attachments: :environment do
-    HtmlAttachment.where(attachable_type: "Edition", attachable_id: Edition.publicly_visible.select("id")).each do |a|
-      Whitehall::PublishingApi.bulk_republish_async(a)
+  task :republish_html_attachments, [:state] => :environment do |_, args|
+    args.with_defaults(state: "withdrawn")
+    publications = Publications.where(state: args[:state])
```
kevindew, line 246: It looks like this lacks the `publicly_visible` part of the previous iteration. Does it need this? I'm not sure if this restriction is needed prior to going to republish worker

kevindew (commented):
Thanks for digging into this @chao-xian - it seems a rather thorny part of Whitehall republishing.

I've left a few comments, some are somewhat reaching into the wider purpose of the task. I did wonder whether the goal here was to have something quick to fix the immediate problem at hand rather than come up with a lasting approach (my hunch for that was the state of withdrawn). If that is the case it might be better to have a temporary task or a snippet of code that can be run directly from the console, rather than amend the current one?

If possible, it'd be also be good to have a basic test for this too. 

@thomasleese may have some thoughts/ideas too. I'm not sure if there are any other HtmlAttachment gotchas I'm not considering, or anything that we'd lose from the original task with this change, that he's better placed to advise on.

thomasleese:
@kevindew Thanks for your review, @chao-xian is unable to look at this right now, so I've had a go myself. I think I've addressed everything you've mentioned. I've also removed the `state` filter as you say I think this PR is for the long term fix, and to solve the immediate problem I'll run a query in production which does an extra filter on only withdrawn editions.

lib/tasks/publishing_api.rake
```diff
@@ -240,10 +240,14 @@ namespace :publishing_api do
...
 
-  desc "Republish all HTML attachments to the Publishing API"
-  task republish_html_attachments: :environment do
-    HtmlAttachment.where(attachable_type: "Edition", attachable_id: Edition.publicly_visible.select("id")).each do |a|
-      Whitehall::PublishingApi.bulk_republish_async(a)
+  desc "Republish all documents with HTML attachments to the Publishing API"
+  task republish_html_attachments: :environment do |_, args|
+    document_ids = Edition
+      .publicly_visible
+      .where(id: HtmlAttachment.where(attachable_type: "Edition").select(:attachable_id))
```
kevindew, line 247: Oh neat - I've never used this approach

kevindew (approved):
Nice one

kevindew:
Oh - could still use a test? unless we have no precedent for those in Whitehall rake tasks?

thomasleese:
> Oh - could still use a test? unless we have no precedent for those in Whitehall rake tasks?

There [doesn't seem to be a precedent](https://github.com/alphagov/whitehall/tree/master/test/unit/tasks) for tests... I don't mind there being one, but I won't be able to look at this anymore today.

kevindew:
Cool not to worry - ship may have long sailed on this with Whitehall. Feel free to merge.

chao-xian:
@thomasleese thank you so much for picking this up and for the great review @kevindew 🙇

PR #5916, opened 2020-12-10
Distinguish between latin script and non latin script languages

AlanGabbianelli:
See commit messages for details.

Trello:
- https://trello.com/c/u1FYvrIZ/2241-5-allow-primarylocale-of-consultations-to-be-changed
- https://trello.com/c/3MrqbQae/2200-5-allow-primarylocale-of-documentcollections-to-be-changed


lib/whitehall.rb
```diff
@@ -32,70 +32,70 @@ def self.support_url
...
-      :tk,
-      :tr,
-      :uk,
-      :ur,
-      :uz,
-      :vi,
-      :zh,
-      "zh-hk",
-      "zh-tw",
+    %i[
```
ChrisBAshton, line 35: Love this first commit. Alphabetising, making type consistent, and testing that the files are in sync. Well done 👏

test/unit/locale_test.rb
```diff
@@ -63,4 +63,13 @@ class LocaleTest < ActiveSupport::TestCase
...
     assert_equal english, Locale.coerce(english)
   end
+
+  test "knows if languages use a Latin-script writing system" do
+    non_latin_script_locale_codes = %i[ar be bg bn dr el fa gu he hi hy ja ka ko pa ps ru si ta th uk ur zh zh-hk zh-tw]
+    non_latin_script_locales = non_latin_script_locale_codes.map { |code| Locale.new(code) }
+    latin_script_locales = Locale.all - non_latin_script_locales
+
+    assert non_latin_script_locales.none?(&:latin_script?)
+    assert latin_script_locales.all?(&:latin_script?)
```
ChrisBAshton, line 73: Great, idiomatic code 👍

ChrisBAshton (approved):
Can't fault this 👍 

My only thought was whether it might be better to have string values such as `script_type: "latin"`. You could still expose these as booleans with `#latin_script?`, but it would give a bit more flexibility as I can see us perhaps introducing more variations like this in future, e.g. `cyrillic_script?: true`. We wouldn't want to allow the system to, say, set both `latin_script?: true` _and_ `cyrillic_script?: true`, and a string value system would enforce just one value.

That could be refactored in later if there is a need, though.

AlanGabbianelli:
That's a good point @ChrisBAshton and I initially thought about it. The reason I did it this way is that knowing what specific script a language uses is not relevant in this context. We are only interested in knowing whether it's a Latin-script. Also, that approach would mean finding out the name of potentially many scripts - a work that I'd want to do only if/when really needed! 😄

PR #5918, opened 2020-12-11
Allow document collections to change locale

edwardkerry:
This PR allows a DocumentCollection to be published with a custom locale. Currently we have issues where a collection of Welsh documents, with a Welsh language DocumentCollection, will still have a locale of `:en`.

This PR adapts logic used for `NewsArticleType::WorldNewsStory`, which are the only existing Edition type that may have a primary locale that is not `:en`. We can now:

* Change the primary_locale of a DocumentCollection.
* Display the `locale_fields` locale selection partial on the DocumentCollection admin page
* Ensure that a non-`:en` localed DocumentCollection will use their title for their slug (logic otherwise exists to use the id)
* Set the `base_path` to append `.:locale` where needed when sending to the PublishingAPI.

[Trello](https://trello.com/c/3MrqbQae/2200-5-allow-primarylocale-of-documentcollections-to-be-changed)

### DIFF
--- a/app/views/admin/editions/_locale_fields.html.erb
+++ b/app/views/admin/editions/_locale_fields.html.erb
@@ -2,7 +2,7 @@
   <fieldset class="foreign-language">
     <div class="checkbox">
       <%= label_tag :create_foreign_language_only, nil, class: [:checkbox, "js-toggle-scheduled-publication-date-picker"] do %>
-        <%= check_box_tag :create_foreign_language_only, "1", form.object.primary_locale != "en" %> Create a foreign language only news article
+        <%= check_box_tag :create_foreign_language_only, "1", form.object.primary_locale != "en" %> <%="Create a foreign language only #{edition.model_name.human.downcase}"%>
       <% end %>
     </div>
     <div class="form-group foreign-language-select js-hidden">
@@ -10,7 +10,9 @@
       <div class="form-inline add-label-margin">
         <%= form.select :primary_locale, options_for_foreign_language_locale(edition), {}, {class: 'form-control input-md-6'} %>
       </div>
-      <p class="warning">Warning: News stories without an English version cannot have other translations.</p>
+      <% if edition.is_a?(NewsArticle) %>
+        <p class="warning">Warning: News stories without an English version cannot have other translations.</p>
+      <% end %>
     </div>
   </fieldset>
 <% end %>
--- a/test/unit/edition/translatable_test.rb
+++ b/test/unit/edition/translatable_test.rb
@@ -36,8 +36,14 @@ class Edition::TranslatableTest < ActiveSupport::TestCase
     assert_not create(:news_article).locale_can_be_changed?
   end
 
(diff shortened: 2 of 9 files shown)


test/unit/edition/translatable_test.rb
```diff
@@ -36,8 +36,17 @@ class Edition::TranslatableTest < ActiveSupport::TestCase
...
   end
 
+  test "locale_can_be_changed? returns true for a new DocumentCollection" do
+    assert DocumentCollection.new.locale_can_be_changed?
+  end
+
+  test "locale_can_be_changed? returns true for an existing DocumentCollection" do
+    doc_collection = create(:document_collection)
+    assert doc_collection.locale_can_be_changed?
+  end
```
AlanGabbianelli, line 46: These could become one test:
```
test "locale_can_be_changed? returns true for existing and new DocumentCollections" do
  new_doc_collection = DocumentCollection.new
  existing_doc_collection = create(:document_collection)
  assert [new_doc_collection, existing_doc_collection].all?(&:locale_can_be_changed?)
end
```
The benefit would be that when we quickly scan the code we'd be less likely to skip one of the two. Also, it makes it extremely clear that we expect the same behaviour for both new and existing document collections.

test/unit/models/document_collection_test.rb
```diff
@@ -33,6 +33,17 @@ class DocumentCollectionTest < ActiveSupport::TestCase
     assert_valid build(:document_collection, body: nil)
   end
 
+  test "it should be able to change locale" do
+    doc_collection = create(:document_collection, groups: [])
+    assert doc_collection.locale_can_be_changed?
+  end
```
AlanGabbianelli, line 39: We're already testing this in `test/unit/edition/translatable_test.rb`. Do we need the same test here?

app/views/admin/editions/_locale_fields.html.erb
```diff
@@ -2,15 +2,17 @@
...
     </div>
     <div class="form-group foreign-language-select js-hidden">
       <%= form.label :primary_locale, 'Document language' %>
       <div class="form-inline add-label-margin">
         <%= form.select :primary_locale, options_for_foreign_language_locale(edition), {}, {class: 'form-control input-md-6'} %>
       </div>
-      <p class="warning">Warning: News stories without an English version cannot have other translations.</p>
+      <% if edition.is_a?(NewsArticle) %>
+        <p class="warning">Warning: News stories without an English version cannot have other translations.</p>
+      <% end %>
```
AlanGabbianelli, line 15: Should this be the case for document collections too?

edwardkerry: I don't believe it is possible for DocumentCollections to have a translation at all, so I don't think we need to warn.

test/unit/models/document_collection_test.rb
edwardkerry: no we don't, thank you

AlanGabbianelli (approved):

PR #5931, opened 2021-01-05
Allow Consultations to change their primary_locale

AlanGabbianelli:
This allows a Consultation to be published with a custom locale.

This follows the same pattern as https://github.com/alphagov/whitehall/pull/5918

* Change the primary_locale of a Consultation.
* Display the locale_fields locale selection partial on the Consultation admin page
* Ensure that a non-:en localed Consultation will use their title for their slug (logic otherwise exists to use the id)
* Set the base_path to append .:locale where needed when sending to the PublishingAPI.

Trello card: https://trello.com/c/u1FYvrIZ/2241-5-allow-primarylocale-of-consultations-to-be-changed

### DIFF
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -180,6 +180,10 @@ class Consultation < Publicationesque
     newly_created ? false : all_nation_applicability
   end
 
+  def locale_can_be_changed?
+    true
+  end
+
 private
 
   def validate_closes_after_opens
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -184,6 +184,10 @@ class Consultation < Publicationesque
     true
   end
 
+  def string_for_slug
+    title
+  end
+
 private
 
   def validate_closes_after_opens
--- a/features/step_definitions/consultation_steps.rb
+++ b/features/step_definitions/consultation_steps.rb
@@ -6,8 +6,10 @@ Given(/^an unopened consultation exists$/) do
   create(:unopened_consultation)
 end
 
-When(/^I draft a new consultation "([^"]*)"$/) do |title|
-  begin_drafting_document type: "consultation", title: title, summary: "consultation-summary", alternative_format_provider: create(:alternative_format_provider), all_nation_applicablity: false
+When(/^I draft a new "(.*?)" language consultation "(.*?)"$/) do |locale, title|
+  document_options = { type: "consultation", title: title, summary: "consultation-summary", alternative_format_provider: create(:alternative_format_provider), all_nation_applicablity: false }
+  document_options.merge!(locale: locale) unless locale == "English"
+  begin_drafting_document document_options
   fill_in "Link URL", with: "http://participate.com"
   fill_in "Email", with: "participate@gov.uk"
   select_date 1.day.ago.to_s, from: "Opening Date"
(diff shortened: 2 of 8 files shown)


features/step_definitions/consultation_steps.rb
```diff
@@ -21,6 +21,21 @@
   click_button "Save"
```
ChrisBAshton, line 21: This step definition can now be simplified to one line, reusing the new step definition you've created:

```ruby
step "I draft a new "English" language consultation \"#{title}\""
```

test/unit/presenters/publishing_api/consultation_presenter_test.rb
```diff
@@ -180,6 +180,26 @@ class BasicConsultationTest < TestCase
...
+
+    test "it presents the default global process wide locale as the locale of the consultation" do
+      assert_equal "en", presented_content[:locale]
+    end
+
+    test "it presents the selected global process wide locale as the locale of the consultation" do
+      with_locale("it") do
+        assert_equal "it", presented_content[:locale]
+      end
+    end
```
ChrisBAshton, line 202: This set of tests would probably read better as two contexts:

```ruby
context "with default locale" do
  test "it presents the default global process wide locale as the locale of the consultation" do
  # 
  # etc
end

context "with overridden locale" do
  with_locale("it") do
    test "it presents ..."
    # etc
```

app/models/consultation.rb
```diff
@@ -184,6 +184,10 @@ def locale_can_be_changed?
     true
   end
 
+  def string_for_slug
+    title
```
ChrisBAshton, line 188: I wasn't sure why this change was made, until I read it in the PR description:

> Ensure that a non-:en localed Consultation will use their title for their slug (logic otherwise exists to use the id)

Could this info be added to the commit too?

ChrisBAshton (commented):
This is looking good 👍 just a couple of suggestions below.

app/models/consultation.rb
edwardkerry: good point - this was included in the "template" PR but neglected here.

features/step_definitions/consultation_steps.rb
edwardkerry: good idea, thanks!

test/unit/presenters/publishing_api/consultation_presenter_test.rb
edwardkerry: Unfortunately it seems that we do not have easy access to `context` blocks in this test file without introducing a `describe` block also, and the objects the tests assert against are method calls in the test's parent class (`BasicConsultationTest < TestCase`) which we then lose access to. 

The setup and running of these tests is relatively complex!

ChrisBAshton: Oh, I had no idea! Fair enough :D

features/step_definitions/consultation_steps.rb
```diff
@@ -6,8 +6,10 @@
   create(:unopened_consultation)
 end
 
-When(/^I draft a new consultation "([^"]*)"$/) do |title|
-  begin_drafting_document type: "consultation", title: title, summary: "consultation-summary", alternative_format_provider: create(:alternative_format_provider), all_nation_applicablity: false
+When(/^I draft a new "(.*?)" language consultation "(.*?)"$/) do |locale, title|
+  document_options = { type: "consultation", title: title, summary: "consultation-summary", alternative_format_provider: create(:alternative_format_provider), all_nation_applicablity: false }
+  document_options.merge!(locale: locale) unless locale == "English"
```
ChrisBAshton, line 11: Looks like you don't need the `unless` anymore, as you've specified the locale in all the feature tests.

ChrisBAshton (approved):
Approved, with one non-blocking comment :) 👍

thomasleese (approved):

thomasleese (approved):

PR #5936, opened 2021-01-07
Fix email signup for world locations

brucebolt:
Signing up for email alerts to world locations currently gives a 500 error. This is because we are attempting to redirect to `nil`. The reason for this is `subscription_url` was removed in https://github.com/alphagov/email-alert-api/commit/75137c2f40b1902a2bf1da481a3a835744dc8e13.

The changes here redirect the user into email-alert-frontend, which handles all parts of the signup journey, meaning we don't need to have a view here in which the user has to unnecessarily click a button, bringing these signups in line with other email signup journeys.

Trello card: https://trello.com/c/qr574f7A


AlanGabbianelli (approved):

benthorner:
Just saw this. Sorry about the bug - I thought I checked it at the time, but apparently not. 

Thanks for fixing ⭐.

PR #5942, opened 2021-01-11
Re-slug HTML attachments if they are unpublished

brucebolt:
It appears we have long standing logic (added in c45922d) that will do the following with HTML attachments:
- if a document has never been published before, the URL slug will be updated automatically each time the attachment title is changed
- if a document has previously been published, the URL slug for new (or modified existing) attachments will never be updated after the title has been entered for the first time

This behaviour was added to ensure links to previously published HTML attachments continue functioning even if the attachment's title is changed in a subsequent edition.

However, this causes confusion if a publisher adds a new attachment in a later edition and finds the slug does not get updated (e.g. from a placeholder). Therefore we need to handle both scenarios correctly.

This adds an additional scenario, so placeholder titles can be used in subsequent editions:
- if a document has never been published before, the URL slug for attachments will be updated automatically each time the attachment title is changed
- if a document has previously been published, the URL slug for existing attachments will never be updated after the title has been entered for the first time (to ensure users who have bookmarked the attachment can still access it)
- if a document has previously been published, the URL slug for new attachments will be updated automatically each time the attachment title is changed until the attachment has been published

Zendesk ticket: https://govuk.zendesk.com/agent/tickets/4420527

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -108,7 +108,9 @@ class Attachment < ApplicationRecord
   end
 
   def deep_clone
-    dup
+    dup.tap do |clone|
+      clone.safely_resluggable = false
+    end
   end
 
   def external?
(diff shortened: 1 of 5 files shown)


AlanGabbianelli (approved):
LGTM and thanks for clarifying in the commit message why we need this extra column. ⭐

PR #5945, opened 2021-01-13
Exclude Details elements from HTML sanitization for inline attachments

edwardkerry:
Inline attachments are wrapped in a `<details>` HTML element which is
being sanitized and removed by Govspeak. This is preventing a JS module
from attaching and results in the accessible format request block
always being expanded.

This change allows editions with inline attachments to request
that Govspeak excludes the `details` element from sanitization
by submitting it as a member of the optional
`allowed_elements` array.

This was introduced to [Govspeak in release 6.6.0](https://github.com/alphagov/govspeak/pull/203)

[trello](https://trello.com/c/uQpgMD83/2289-5-accessible-format-request-show-hide-not-working-on-detailed-guide-pages)


Gemfile.lock
```diff
@@ -221,7 +221,7 @@ GEM
       activemodel (>= 4.2, < 7.0)
       activerecord (>= 4.2, < 7.0)
       request_store (~> 1.0)
-    govspeak (6.5.10)
+    govspeak (6.6.0)
```
ChrisBAshton, line 224: Might be better to make this version an explicit dependency in `Gemfile` too. Otherwise another dependency could in theory require an older version of govspeak, and there'd be no compatibility clash, so Gemfile.lock would merrily downgrade and then the `<details>` element would break again.

ChrisBAshton (commented):
Looks great, and nice easy-to-follow commits 👍 Just one suggestion below.

Gemfile.lock
ChrisBAshton: Ooh, interesting! Yes I think our reasoning still holds up and is worth applying here, where we know that a lower version would break Whitehall.

kevindew: I wouldn't expect it's worth the long term overhead this will bring. I assume the test would fail if the version was lowered?

Putting a pin is advised more for when you need to stay behind the latest version.

edwardkerry: the tests will fail, and following the docs and this advice I've removed the constraint once more 👍

thomasleese (approved):

ChrisBAshton (approved):

PR #5962, opened 2021-01-21
Fix Statistics pages' HTML attachment base paths

AlanGabbianelli:
Statistics are available at "/government/statistics/:slug", but associated
HTMLAttachments are given the base path "/government/publication/:slug/:attachment".

This is due to the HTMLAttachment class using the Attachable's class name to form
the url, and OfficialStatistics and NationalStatistics are subtypes of Publication.

This PR adds an `html_attachment_type` method to Publication, allowing the class
to differentiate between different subtypes and allow HTMLAttachment to
set the  basepath to "/government/statistics".

If this method is not available to a class, HTMLAttachment will continue to
fallback to the Attachable's class name.

[Trello](https://trello.com/c/l7Q4urHs/2295-5-fix-statistics-pages-html-attachment-base-paths-%F0%9F%8D%90)

### DIFF
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -66,7 +66,7 @@ class HtmlAttachment < Attachment
       options[:host] = URI(Plek.new.external_url_for("www-origin")).host
     end
 
-    type = attachable.class.name.underscore
+    type = attachable.try(:html_attachment_type) || attachable.class.name.underscore
     path_or_url = full_url ? "url" : "path"
     path_helper = "#{type}_html_attachment_#{path_or_url}"
 
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -66,7 +66,7 @@ class HtmlAttachment < Attachment
       options[:host] = URI(Plek.new.external_url_for("www-origin")).host
     end
 
-    type = attachable.try(:html_attachment_type) || attachable.class.name.underscore
+    type = attachable.path_name
     path_or_url = full_url ? "url" : "path"
     path_helper = "#{type}_html_attachment_#{path_or_url}"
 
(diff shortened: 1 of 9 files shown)


app/models/html_attachment.rb
```diff
@@ -66,7 +66,7 @@ def url(options = {})
       options[:host] = URI(Plek.new.external_url_for("www-origin")).host
     end
 
-    type = attachable.class.name.underscore
+    type = attachable.try(:html_attachment_type) || attachable.class.name.underscore
```
thomasleese, line 69: I'm just thinking about how we ended up in a situation where the path to the HTML attachment didn't match the parent document and I think it's because the code on this line ends up "out of sync" with this code here:

https://github.com/alphagov/whitehall/blob/5d76209841f3e25ebe2c7c8d0b944ce934a12c80/app/helpers/public_document_routes_helper.rb#L32-L36

I was wondering if there was a way we could have both of these call a shared function so the logic always stays the same.

thomasleese (commented):
I've just got one comment about the code changes, but it makes sense to me.

I'm wondering if you had any thoughts on how best to redirect and republish the existing statistics announcements? I think it should just be a case of republishing them all and the Publishing API will automatically put in the redirect.

app/models/html_attachment.rb
edwardkerry: Thanks for this, I hadn't realised that. I have added a `base_path` method to the classes with HTML Attachments. As ConsultationOutcomes also allow HTML Attachments but are not editions it's been added to a few places but hopefully breaks the reliance and places the knowledge back in the owning class.

Though I'm not sure of the method name.

edwardkerry:
> I've just got one comment about the code changes, but it makes sense to me.
> 
> I'm wondering if you had any thoughts on how best to redirect and republish the existing statistics announcements? I think it should just be a case of republishing them all and the Publishing API will automatically put in the redirect.

I think that should be fine having manually tested a few in integration - perhaps we can add a temporary rake task following this being deployed.

ChrisBAshton (approved):
Changes look good to me. 👍 Good job!

PR #5966, opened 2021-01-28
Catch Notify BadErrorRequest in AuthorNotifierService

edwardkerry:
The AuthorNotifierWorker is currently raising `Notifications::Client::BadRequestError` in staging and integration environments. This is expected behaviour as our Notify service has restricted e-mail recipient lists in these environments, so it is safe to rescue from these errors.
 
Based on https://github.com/alphagov/whitehall/pull/5897

[trello](https://trello.com/c/8YKMuGwk/2320-2-fix-notificationsclientbadrequesterror-in-whitehall)

### DIFF
--- a/app/services/author_notifier_service.rb
+++ b/app/services/author_notifier_service.rb
@@ -18,6 +18,11 @@ class AuthorNotifierService
         edition_admin_url,
         public_document_url,
       ).deliver_now
+    rescue Notifications::Client::BadRequestError => e
+      # in production we care about all errors
+      # in staging and integration the team-only error is unrecoverable when running asynchronously
+      # (team-only error is unrecoverable in production too, but almost certainly impossible)
+      raise if ENV["SENTRY_CURRENT_ENV"] !~ /integration|staging/ || e.message !~ /team-only API key/
     end
   end
 
--- a/app/workers/document_list_export_worker.rb
+++ b/app/workers/document_list_export_worker.rb
@@ -16,12 +16,10 @@ private
   def send_mail(csv_filename, user, filter)
     MailNotifications.document_list(csv_filename, user.email, filter.page_title).deliver_now
   rescue Notifications::Client::BadRequestError => e
-    if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
-      # in production we care about all errors
-      # in staging and integration the team-only error is unrecoverable when running asynchronously
-      # (team-only error is unrecoverable in production too, but almost certainly impossible)
-      raise
-    end
+    # in production we care about all errors
+    # in staging and integration the team-only error is unrecoverable when running asynchronously
+    # (team-only error is unrecoverable in production too, but almost certainly impossible)
+    raise if ENV["SENTRY_CURRENT_ENV"] !~ /integration|staging/ || e.message !~ /team-only API key/
   end
 
   def upload_file(csv, type)
--- a/test/unit/services/author_notifier_service.rb
+++ b/test/unit/services/author_notifier_service.rb
@@ -82,4 +82,40 @@ class AuthorNotifierTest < ActiveSupport::TestCase
     production_notification = ActionMailer::Base.deliveries.third
     assert_no_match pattern, production_notification.to_s
   end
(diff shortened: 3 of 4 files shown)


app/services/author_notifier_service.rb
```diff
@@ -18,6 +18,13 @@ def notify!
         edition_admin_url,
         public_document_url,
       ).deliver_now
+    rescue Notifications::Client::BadRequestError => e
+      if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
+        # in production we care about all errors
+        # in staging and integration the team-only error is unrecoverable when running asynchronously
+        # (team-only error is unrecoverable in production too, but almost certainly impossible)
+        raise
+      end
```
brucebolt, line 27:
```suggestion
    rescue Notifications::Client::BadRequestError => e
      # in production we care about all errors
      # in staging and integration the team-only error is unrecoverable when running asynchronously
      # (team-only error is unrecoverable in production too, but almost certainly impossible)
      raise if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
```

brucebolt (approved):
Looks good!

I've left a non-blocking suggestion that will make the if-raise more succinct by removing the block, but feel free to ignore if you prefer.

app/services/author_notifier_service.rb
```diff
@@ -18,6 +18,13 @@ def notify!
         edition_admin_url,
         public_document_url,
       ).deliver_now
+    rescue Notifications::Client::BadRequestError => e
+      if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
```
1pretz1, line 22: I don't think this logic is quite correct (at least for Whitehall) as in integration `SENTRY_CURRENT_ENV` equals `"integration-blue-aws"`. For staging it is fine though. 👍 

Side note: we could use `exclude?` here instead of inverting `includes?` boolean.

edwardkerry: thanks for noticing that environment @1pretz1!

benjamineskola: Elsewhere we use `ENV.fetch("GOVUK_WEBSITE_ROOT") =~ /integration|staging/`; I think for this we'd still want to check the sentry environment but we could still do `=~` instead of `include?`.

app/services/author_notifier_service.rb
edwardkerry: I was in two minds about this as I thought it drew more attention to the comment, but on balance 👍

app/services/author_notifier_service.rb
edwardkerry: How about we instead check it's production
`raise if ENV["SENTRY_CURRENT_ENV"] =~ /production/`?

benjamineskola: I think @thomasleese suggested doing it this way around because if `SENTRY_CURRENT_ENV` is unset we'd want it to default to the production behaviour.

app/workers/document_list_export_worker.rb
```diff
@@ -16,12 +16,10 @@ def perform(filter_options, user_id)
   def send_mail(csv_filename, user, filter)
     MailNotifications.document_list(csv_filename, user.email, filter.page_title).deliver_now
   rescue Notifications::Client::BadRequestError => e
-    if !%w[integration staging].include?(ENV["SENTRY_CURRENT_ENV"]) || e.message !~ /team-only API key/
-      # in production we care about all errors
-      # in staging and integration the team-only error is unrecoverable when running asynchronously
-      # (team-only error is unrecoverable in production too, but almost certainly impossible)
-      raise
-    end
+    raise if ENV["SENTRY_CURRENT_ENV"] =~ /production/ || e.message !~ /team-only API key/
```
ChrisBAshton, line 19: Nitpick: the `raise` is above the comment in this file, but below it in the other. It should be consistent (below the comment, IMO).

Otherwise looks good - thanks Ed! 👍

benjamineskola (approved):

PR #5969, opened 2021-02-01
Add standard document type

1pretz1:
Add Standard document type.

I've tagged the document type as 'less common' which means it doesn't feature at the top of the publications list when creating a publication. This can be changed to 'primary' if the new type is deemed sufficiently important. 

Screenshot 2021-03-23 at 15 48 59

Trello:
https://trello.com/c/ugeapNY0/2318-5-create-new-standard-document-sub-type-in-whitehall


brucebolt (commented):
Changes look fine, pending the content update.

The `PublicationType` model looks horrendous to work with!  Good work figuring out how this fits together.

Tetrino (approved):
Seems complete, good job!

PR #5972, opened 2021-02-04
Permit withdrawn redirects to allow listed external URLs

brucebolt:
Energy Technology List and Courts & Tribunals Judiciary have been given exemption from publishing on GOV.UK, so we need the ability to unpublish documents and redirect to their own domains.

Requires https://github.com/alphagov/publishing-api/pull/1898 to be merged.

Trello card: https://trello.com/c/138P3vIk

### DIFF
--- a/app/models/unpublishing.rb
+++ b/app/models/unpublishing.rb
@@ -61,6 +61,8 @@ class Unpublishing < ApplicationRecord
   def alternative_path
     return if alternative_uri.nil?
 
+    return alternative_uri.to_s unless GovUkUrlFormatValidator.matches_gov_uk?(alternative_uri)
+
     path = alternative_uri.path
     path << "##{alternative_uri.fragment}" if alternative_uri.fragment.present?
     path
--- a/app/validators/gov_uk_url_format_validator.rb
+++ b/app/validators/gov_uk_url_format_validator.rb
@@ -1,19 +1,19 @@
 # Accepts options[:message] and options[:allowed_protocols]
 class GovUkUrlFormatValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    unless matches_gov_uk?(value) || matches_allow_list?(value)
+    unless self.class.matches_gov_uk?(value) || matches_allow_list?(value)
       record.errors[attribute] << failure_message
     end
+  end
+
+  def self.matches_gov_uk?(value)
+    %r{\A#{Whitehall.public_protocol}://#{Whitehall.public_host}/}.match?(value)
   end
 
 private
 
   def failure_message
     options[:message] || "must be in the form of #{Whitehall.public_protocol}://#{Whitehall.public_host}/example"
-  end
-
-  def matches_gov_uk?(value)
-    %r{\A#{Whitehall.public_protocol}://#{Whitehall.public_host}/}.match?(value)
   end
 
   def matches_allow_list?(value)
--- a/app/validators/gov_uk_url_format_validator.rb
+++ b/app/validators/gov_uk_url_format_validator.rb
@@ -1,7 +1,7 @@
 # Accepts options[:message] and options[:allowed_protocols]
 class GovUkUrlFormatValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    unless %r{\A#{Whitehall.public_protocol}://#{Whitehall.public_host}/}.match?(value)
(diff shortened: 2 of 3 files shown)


ChrisBAshton (approved):
LGTM 👍 nice one.

brucebolt:
Thanks @chrisbashton!  I've updated one of the commits to use `end_with?` with multiple arguments, as suggested by @1pretz1 in https://github.com/alphagov/publishing-api/pull/1898#discussion_r570296506.

app/validators/gov_uk_url_format_validator.rb
```diff
@@ -11,4 +11,15 @@ def validate_each(record, attribute, value)
...
     options[:message] || "must be in the form of #{Whitehall.public_protocol}://#{Whitehall.public_host}/example"
   end
+
+  def matches_gov_uk?(value)
+    %r{\A#{Whitehall.public_protocol}://#{Whitehall.public_host}/}.match?(value)
+  end
+
+  def matches_allow_list?(value)
+    uri = URI.parse(value)
+    uri.host&.end_with?(".gov.uk", ".judiciary.uk", "etl.beis.gov.uk")
```
ChrisBAshton, line 21: This is currently a little confusing, as I'd expect `.gov.uk` to only be checked in `matches_gov_uk?`, and `matches_allow_list?` to only contain the special domains.

```suggestion
    uri.host&.end_with?(".judiciary.uk", "etl.beis.gov.uk")
```

brucebolt: Good spot!  I just copied this from the `publishing-api` PR 🤦

app/models/unpublishing.rb
```diff
@@ -89,4 +91,8 @@ def redirect_not_circular
   def ensure_presence_of_content_id
     self.content_id ||= SecureRandom.uuid
   end
+
+  def alternative_uri_matches_gov_uk?
```
ChrisBAshton, line 95: Would be great if we didn't duplicate this, and call `GovUkUrlFormatValidator.matches_gov_uk?` instead. Would just need to move it above the `private` and prefix it with a `self.`

brucebolt: Thanks. I've removed the duplication here.

app/validators/gov_uk_url_format_validator.rb
```diff
@@ -1,14 +1,25 @@
 # Accepts options[:message] and options[:allowed_protocols]
 class GovUkUrlFormatValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    unless %r{\A#{Whitehall.public_protocol}://#{Whitehall.public_host}/}.match?(value)
+    unless GovUkUrlFormatValidator.matches_gov_uk?(value) || matches_allow_list?(value)
```
ChrisBAshton, line 4: I think we can remove the `GovUkUrlFormatValidator` bit, as we're already inside the class.

ChrisBAshton (approved):
LGTM 👍 one non-blocking nitpick below:

app/validators/gov_uk_url_format_validator.rb
brucebolt: Unfortunately not.  I get a `NoMethodError: undefined method `matches_gov_uk?' for #<GovUkUrlFormatValidator:0x0000562d6d9863e0>` error without the class name.

ChrisBAshton: Ah, apologies. Hmm, how about `self.matches_gov_uk?`?

brucebolt: It's the same error in that case too

ChrisBAshton: Looks like the [approach](https://stackoverflow.com/questions/2527830/ruby-calling-class-method-from-instance) is `self.class.matches_govuk_uk?`. Shame, not as terse as I thought it would be, but probably slightly better than repeating the class name, as:

1. it is quicker to establish that it's referring to itself (using the class name looks like we're calling out to another class, at first glance)
2. it's one less thing to replace if we ever rename the class

Very nitpicky of me though 😄

brucebolt: That's a good point.  I've updated to use `self.class`.

ChrisBAshton (approved):

PR #5973, opened 2021-02-04
Temporary rake task to republish all statistics with HTML Attachments

edwardkerry:
https://github.com/alphagov/whitehall/pull/5962 introduces a change to ensure that HTMLAttachments for statistics pages are located at `/government/statistics` rather than `/government/publications`.

This PR adds two temporary rake tasks to republish all statistics publications with HTML Attachments to update their path, and then reindex so the HTMLAttachments appear appended in search results.

[trello](https://trello.com/c/l7Q4urHs/2295-5-fix-statistics-pages-html-attachment-base-paths-%F0%9F%8D%90)

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -251,6 +251,17 @@ namespace :publishing_api do
     end
   end
 
+  desc "Republish all statistics which have HTML attachments to the Publishing API"
+  task republish_stats_with_html_attachments: :environment do
+    document_ids = Publication
+      .statistical_publications
+      .where(id: HtmlAttachment.where(attachable_type: "Edition").select(:attachable_id))
+      .pluck(:document_id)
+    document_ids.each do |document_id|
+      PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
+    end
+  end
+
   desc "Bulk republishing"
   namespace :bulk_republish do
     desc "Republish all documents of a given type, eg 'NewsArticle'"
--- a/lib/tasks/search.rake
+++ b/lib/tasks/search.rake
@@ -53,6 +53,16 @@ namespace :search do
       index.commit
     end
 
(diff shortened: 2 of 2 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -251,6 +251,18 @@ namespace :publishing_api do
     end
   end
 
+  desc "Republish all statistics which have HTML attachments to the Publishing API"
+  task republish_stats_with_html_attachments: :environment do
+    document_ids = Publication
+      .publicly_visible
```
1pretz1, line 257: Do we not want to republish draft HTML attachments too?

1pretz1 (commented):
Looks good Ed, just one question.

lib/tasks/publishing_api.rake
edwardkerry: thanks, we probably do! there's also a scope already set up for `statistical_documents` so i'll redraft this shortly.

lib/tasks/publishing_api.rake
```diff
@@ -251,6 +251,17 @@ namespace :publishing_api do
     end
   end
 
+  desc "Republish all statistics which have HTML attachments to the Publishing API"
+  task republish_stats_with_html_attachments: :environment do
```
thomasleese, line 255: Do we need to republish the HTML attachments directly as well, or does that happen as part of republishing the publication?

edwardkerry: The attachments are also republished, the PublishingAPIDocumentRepublishingWorker uses the PublishingApiHtmlAttachments service listener to take care of them.
https://github.com/alphagov/whitehall/blob/2fead7d3d01ac1240d564153ace564809d7a6b72/app/workers/publishing_api_document_republishing_worker.rb#L150

1pretz1 (approved):
Nice work! 🍏

PR #5980, opened 2021-02-09
Revert "Temporary rake task to republish all statistics with HTML Attachments"

edwardkerry:
Reverts alphagov/whitehall#5973
These tasks have now run and can be removed

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -251,17 +251,6 @@ namespace :publishing_api do
     end
   end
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM 👍 would be good to get the "These have now been run and can be removed" into the commit message, and to get the first line of the message <50 chars so that it displays nicely on GitHub.

ChrisBAshton:
Merging now as doing a deployment, and devs can find their way back to this PR for more context if needed.

PR #5981, opened 2021-02-10
Adds locale to social media links

MuriloDalRi:
Adds the option to select a language when adding or editing social media
links. These links are only shown when the locale of the document
matches the locale of the social media account.

[Trello](https://trello.com/c/RRvGDDmD/2243-5-make-whitehall-social-media-links-translatable)

### DIFF
--- a/app/presenters/publishing_api/worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/worldwide_organisation_presenter.rb
@@ -36,5 +36,15 @@ module PublishingApi
     def description
       item.summary
     end
+
+    def social_media_links
+      item.social_media_accounts.where(locale: I18n.locale).map do |account|
+        {
+          service_type: account.service_name.parameterize,
+          title: account.display_name,
+          href: account.url,
+        }
+      end
+    end
   end
 end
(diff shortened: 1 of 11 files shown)


thomasleese (approved):
These changes look good to me. I guess there's a plan to update the rendering of non-worldwide orgs [that happens in Collections as well](https://github.com/alphagov/collections/blob/9ae0fbf8214117c52bbe2bff03645495ab1598bd/app/views/organisations/_what_we_do.html.erb#L30-L43)?

app/presenters/publishing_api/worldwide_organisation_presenter.rb
```diff
@@ -36,5 +36,16 @@ def links
     def description
       item.summary
     end
+
+    def social_media_links
+      item.social_media_accounts.map do |account|
```
thomasleese, line 41: I think we can do the same here with `.where(locale: I18n.locale)`. I realise it's less important because these pages are fully rendered by Whitehall still, but I think worth being consistent.

MuriloDalRi: Yes, I'd missed that one!

thomasleese (approved):

thomasleese:
I'll close this for now as we're going for a different approach, although it will likely take some code from this.

PR #5995, opened 2021-02-19
Fix up withdrawn documents which erroneously have draft editions

ChrisBAshton:
For years, we've been incorrectly saving a draft edition in
Publishing API after informing it of Withdrawn documents in
Whitehall. This was fixed in #5992, but we still have a number of
existing documents this affects which would currently be unable
to have their withdrawal explanations updated.

This commit adds a rake task to resync all withdrawn documents
with Publishing API, explicitly removing any draft editions that
exist. This should solve the issue once and for all.

I've had to change the `PublishingApiUnpublishingWorker` to
inherit from `WorkerBase` in order to get the
`perform_async_in_queue` method. I see no disadvantage to doing
so, and it should free up the main queue for more important
publishing.

I also had to remove some lines from `ResluggingTest`, which are
no longer needed now that we're loading all Rake tasks in
test_helper.rb.

Trello: https://trello.com/c/Oicf9IVG/2354-5-investigate-fix-withdrawn-explanations-not-updating-in-whitehall

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -274,6 +274,28 @@ namespace :publishing_api do
       end
       puts "Finished enqueueing items for Publishing API"
     end
+
+    desc "Remove any draft edition associated with Withdrawn editions and republish"
+    task fix_withdrawn_documents: :environment do
+      puts "Finding withdrawn editions that need fixing..."
+      unpublishings = Unpublishing.unscoped.
+        # we only want documents that were withdrawn
+        where(unpublishing_reason_id: UnpublishingReason::Withdrawn.id).filter do |unpublishing|
+          # the associated edition should be in a `withdrawn` state
+          unpublishing.edition.state == "withdrawn" &&
+            # the associated edition should be the latest edition
+            unpublishing.edition.id == unpublishing.edition.document.latest_edition.id
+        end
+      editions = unpublishings.map(&:edition)
+
+      editions.each do |edition|
+        edition.translations.pluck(:locale).each do |locale|
+          puts "Fixing edition #{edition.content_id}, locale #{locale}"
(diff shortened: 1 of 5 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -292,6 +292,9 @@ namespace :publishing_api do
         edition.translations.pluck(:locale).each do |locale|
           puts "Fixing edition #{edition.content_id}, locale #{locale}"
           Services.publishing_api.discard_draft(edition.content_id, locale: locale)
+        rescue GdsApi::HTTPNotFound, GdsApi::HTTPUnprocessableEntity
```
brucebolt, line 295: You might want to update the commit message to include error code 422 (in addition to 404).

I don't know if there's any possibility we'll ever get this error, but publishing-api can [return a 409](https://github.com/alphagov/publishing-api/blob/8b9b753866a013d012093f93a51c0809e2a74f96/app/commands/base_command.rb#L84), in addition to [404/422](https://github.com/alphagov/publishing-api/blob/8b9b753866a013d012093f93a51c0809e2a74f96/app/commands/v2/discard_draft.rb#L27).

brucebolt (approved):
Looks good to me.  I've just added a really minor comment about the error handling.

lib/tasks/publishing_api.rake
ChrisBAshton: Thanks for finding that line of code. I did wonder why I was seeing a 404 on my first Integration test, and a 422 on the subsequent test 😅 I'll split that out into separate unit tests.

I won't worry about a 409 unless I come across it in the full Integration run here: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/220338/console

PR #6001, opened 2021-02-23
Add healthcheck endpoint for scheduled editions without a job

brucebolt:
This adds an endpoint `/healthcheck/unenqueued_scheduled_editions` to check for the number of scheduled editions that have not been queued in Sidekiq.

We already have this data included in the `/healthcheck` endpoint, but this is shown in Icinga for every host (rather than once per app). Once this change has been deployed and the alert set up, the old system can be removed.

Trello card: https://trello.com/c/UAbbL13x

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -5,4 +5,13 @@ class HealthcheckController < ActionController::API
     logger.error "HealthcheckController#overdue: #{e.message}"
     render json: { overdue: nil }, status: :service_unavailable
   end
+
+  def unenqueued_scheduled_editions
+    render json: {
+      unenqueued_scheduled_editions: Healthcheck::ScheduledPublishing.new.unenqueued_editions,
+    }
+  rescue ActiveRecord::StatementInvalid => e
+    logger.error "HealthcheckController#unenqueued_scheduled_editions: #{e.message}"
+    render json: { unenqueued_scheduled_editions: nil }, status: :service_unavailable
+  end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/healthcheck_controller.rb
```diff
@@ -5,4 +5,13 @@ def overdue
     logger.error "HealthcheckController#overdue: #{e.message}"
     render json: { overdue: nil }, status: :service_unavailable
   end
+
+  def unenqueued_scheduled_editions
+    render json: {
+      unenqueued_scheduled_editions: Healthcheck::ScheduledPublishing.new.unenqueued_editions,
+    }
+  rescue ActiveRecord::StatementInvalid => e
```
thomasleese, line 13: Since both controller methods catch this, is there a way we can do a [`rescue_from`](https://apidock.com/rails/ActiveSupport/Rescuable/ClassMethods/rescue_from) to reduce the duplication?

thomasleese (approved):
I've left one comment, not major enough for me not to approve the PR though.

app/controllers/healthcheck_controller.rb
brucebolt: I chose not to `rescue_from` here as the JSON is slightly different in the two methods.  Do you know if there's a way round this?

thomasleese: Hmm that's true, I wonder if the `overdue` or `unenqueued_scheduled_editions` field is superfluous since it's already in the path and we could just have a `value` field. However, this feels a bit out of scope for this given that it requires changing the other one too, so we can leave it like this for now.

PR #6002, opened 2021-02-23
Remove scheduled job check

brucebolt:
A new endpoint for unenqueued scheduled editions was added in #6001, so we can now safely remove it from `/healthcheck`.

Not to be merged until https://github.com/alphagov/whitehall/pull/6001 and https://github.com/alphagov/govuk-puppet/pull/10963 are merged and deployed.

Trello card: https://trello.com/c/meKsxLti


app/controllers/healthcheck_controller.rb
```diff
@@ -5,4 +5,13 @@ def overdue
     logger.error "HealthcheckController#overdue: #{e.message}"
     render json: { overdue: nil }, status: :service_unavailable
   end
+
+  def unenqueued_scheduled_editions
+    render json: {
+      unenqueued_scheduled_editions: Edition.future_scheduled_editions.count - ScheduledPublishingWorker.queue_size,
+    }
+  rescue ActiveRecord::StatementInvalid => e
```
thomasleese, line 13: Since both controller methods catch this, is there a way we can do a [`rescue_from`](https://apidock.com/rails/ActiveSupport/Rescuable/ClassMethods/rescue_from) to reduce the duplication?

thomasleese (approved):
I've left one comment, not major enough for me not to approve the PR though.

app/controllers/healthcheck_controller.rb
thomasleese: (Can ignore, as this is more relevant to #6001)

PR #6003, opened 2021-02-25
Remove support for pa-ur locale

edwardkerry:
This PR removes support for the `pa-ur: Punjabi Shahmukhi` locale which was accidentally introduced with an incorrect tag.

It also includes a data migration to update translations published with the inaccurate locale.

[trello](https://trello.com/c/devOCero/2369-3-change-punjabi-pakistan-language-tag)


MuriloDalRi (approved):

db/data_migration/20210225103359_update_pa_ur_translations_to_pa_pk.rb
```diff
@@ -0,0 +1,18 @@
+content_ids = %w[
```
1pretz1, line 1: Is this list of content ids guaranteed to be the only documents effected? If not, we could query the incorrect locale documents on runtime which may be more comprehensive than using a hardcoded list.

1pretz1 (commented):
Changes look good Ed! I did have one question which I can now see you've kind of answered in the PR description. It may be worth using a query in that case to potentially save a second data migration.

db/data_migration/20210225103359_update_pa_ur_translations_to_pa_pk.rb
edwardkerry: This suggestion is certainly preferable, but unfortunately having experimented and taken advice there doesn't seem to be a simple/performant way of retrieving content-ids of editions with particular locales from within Whitehall itself. This list has actually been taken from a Publishing-API  query. 

As always, very happy to swap this if there is an alternative approach :)

1pretz1: Ah, nice to know this approach has already been thought of. 

Hmm, I don't think you necessarily need the content_id's of the documents as you can get away with just passing round the doc id. I _think_ something like whats below would suffice: 

```
translations = Edition::Translation.where(locale: "pa-ur")

translations.update_all!(...)

document_ids = Edition.distinct.where(id: translations.select(:edition_id)).pluck(:document_id)

document_ids.each do |id|
  PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", id)
end
```

This would boil down to one update and one select query which seems to be relatively performant. Let me know if this is viable?

edwardkerry: this looks sensible and correct - the thing I forgot to mention was that this approach will not find Documents outside the Edition structure - such as Organisations etc. But currently it returns everything in Publishing-API and is certainly an improvement!

1pretz1 (approved):
LGTM! 🍏

MuriloDalRi (approved):

PR #6004, opened 2021-02-25
Use image card component

owenatgov:
## What
Replace bespoke components with the [image card component](https://components.publishing.service.gov.uk/component-guide/image_card) where appropriate.

## Why

There are several places where this component could've theoretically been used but I made the choice to leave it alone. These reasons include:

- Strict data structure that doesn't lend itself to the flat data structure of the publishing component and would be too large a job to unpick as part of this PR. A perfect example is the [Person class](https://github.com/alphagov/whitehall/blob/master/app/presenters/person_presenter.rb) which is a candidate for the image card but is spitting out html in its calls and is used heavily across the app.
- Inability to replicate locally. There are several locations where image cards could be used but are very difficult to replicate locally due to a laundry list of specific needs. [Example page](https://www.gov.uk/world/france/news). I've made the choice after banging my head against these pages for a day or so to leave them be in the interest of progressing this work.

### Pages tested against
- https://www.gov.uk/government/history/past-foreign-secretaries
- https://www.gov.uk/world/france/news
- https://www.gov.uk/world/uk-mission-to-the-wto-un-and-other-international-organisations-geneva

## Visual changes
Visual changes here are negligible and principally boil down to slight spacing difference. The example below is illustrative of all locations where this change has taken place.

### Before
Screenshot 2021-02-23 at 16 57 50
Screenshot 2021-03-04 at 16 04 28

### After
Screenshot 2021-02-23 at 16 57 32
Screenshot 2021-03-04 at 16 04 18

### DIFF
--- a/app/views/past_foreign_secretaries/index.html.erb
+++ b/app/views/past_foreign_secretaries/index.html.erb
@@ -19,152 +19,112 @@
     <ol>
       <li class="person person-excerpt">
         <div class="inner">
-          <div class="image-holder govuk-!-margin-bottom-1">
-            <a href="/government/history/past-foreign-secretaries/edward-wood" class="aspect-ratio-3:2" tabindex="-1" aria-hidden="true">
-              <%= image_tag 'history/past-foreign-secretaries/viscount-halifax.jpg', alt: 'Edward Frederick Lindley Wood, Viscount Halifax', loading: "lazy" %>
-            </a>
-          </div>
-          <p class="govuk-heading-s govuk-!-margin-bottom-1">
-            <a href="/government/history/past-foreign-secretaries/edward-wood" class="govuk-link">
-              Edward Frederick Lindley Wood, Viscount Halifax
-            </a>
-          </p>
-          <p class="govuk-body-s govuk-!-margin-bottom-0 term">1938 to 1940</p>
+          <%= render "govuk_publishing_components/components/image_card", {
+            href: "/government/history/past-foreign-secretaries/edward-wood",
+            image_src: image_path("history/past-foreign-secretaries/viscount-halifax.jpg"),
+            image_alt: "Edward Frederick Lindley Wood, Viscount Halifax",
+            heading_text: "Edward Frederick Lindley Wood, Viscount Halifax",
+            description: "1938 to 1940",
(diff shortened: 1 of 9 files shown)


app/views/past_foreign_secretaries/index.html.erb
```diff
@@ -19,152 +19,112 @@
...
-              <%= image_tag 'history/past-foreign-secretaries/viscount-halifax.jpg', alt: 'Edward Frederick Lindley Wood, Viscount Halifax', loading: "lazy" %>
-            </a>
-          </div>
-          <p class="govuk-heading-s govuk-!-margin-bottom-1">
-            <a href="/government/history/past-foreign-secretaries/edward-wood" class="govuk-link">
-              Edward Frederick Lindley Wood, Viscount Halifax
-            </a>
-          </p>
-          <p class="govuk-body-s govuk-!-margin-bottom-0 term">1938 to 1940</p>
+          <%= render "govuk_publishing_components/components/image_card", {
```
chris-gds, line 22: Are we ok with this change on mobile? 🤔  

Screenshot 2021-03-04 at 10 56 36

owenatgov: Good spot! I personally think that's fine... It's the standard behaviour coming from the image card. [Live usage](https://www.gov.uk/government/organisations/attorney-generals-office). I'm happy to get a design eye on this if we want?

chris-gds: They do seem a bit small, esp on like a 4-inch device.  We don't use images much nor does the image card support the previous layout.

Design comment might be good thought just in case.

However, if you're happy, I'm good -  👍🏼   the default behaviour angle I get.

chris-gds (approved):

app/views/past_foreign_secretaries/index.html.erb
owenatgov: I'll run this past Mia just for safety's sake.

owenatgov: We've agreed that we're gonna merge this for now and I might drop the component past Mia at a later date.

PR #6008, opened 2021-03-01
Use the title component on email signup information pages

owenatgov:
## What
Replaces the title on the email signup info view ([example](https://www.gov.uk/government/organisations/medicines-and-healthcare-products-regulatory-agency/email-signup)) with the [title component](https://components.publishing.service.gov.uk/component-guide/title).

## Why

### Before
Screenshot 2021-03-01 at 15 52 16

### After
Screenshot 2021-03-01 at 15 52 08

### DIFF
--- a/app/assets/stylesheets/frontend/views/_email-signup-information.scss
+++ b/app/assets/stylesheets/frontend/views/_email-signup-information.scss
@@ -1,14 +1,4 @@
 .email-signup-information-show {
-  h1 {
-    @include heading-48;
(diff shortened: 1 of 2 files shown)


chris-gds (approved):

chris-gds:
Note padding change but default coming from gem

image

owenatgov:
@chris-gds I've updated the spacing on this to be a bit less vast and closer to the live instance.

PR #6017, opened 2021-03-04
Migrate Attachments with pa-ur locale to pa-pk

edwardkerry:
This PR adds a data migration to update Attachments with the old `pa-ur` locale to the updated `pa-pk`

[trello](https://trello.com/c/devOCero/2369-3-change-punjabi-pakistan-language-tag)


1pretz1 (approved):
Code looks v familiar 😄 

Do `FileAttachment`'s need to be updated in Asset Manager as part of this? Can't see any locale field in Asset Manager so I assume not. Approving as I think I've answered my own question. 💯

PR #6024, opened 2021-03-10
Update README

chris-gds:
## What

Adding additional documentation

## Why

So starting the application is clearer

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,24 @@
 
 whitehall is a Ruby on Rails content management application for content published by government departments and agencies.
 
+## Running the Application
+
+Startup [using govuk-docker](https://github.com/alphagov/govuk-docker). 
+
+There are two different views within this App, a *Publishing view* and a *Site view*. Once running, whitehall does not have an index, some suggested starting pages are below:
+
+Publishing view:
+- <http://whitehall-admin.dev.gov.uk/government/admin/news/new>
+
+Site view:
+- <http://whitehall-frontend.dev.gov.uk/government/get-involved>
+
+Some pages will need data locally to display, whitehall uses mySQL. You'll need to gain relevant permissions to access data from AWS
+
+- [Get setup with AWS access](https://docs.publishing.service.gov.uk/manual/get-started.html)
+
+- Once completed [a guide to install local data on whitehall can be found here](https://github.com/alphagov/govuk-docker/blob/master/docs/how-tos.md#how-to-replicate-data-locally)
+
 ## Nomenclature
 
 - *Govspeak* A variation of [Markdown](https://daringfireball.net/projects/markdown) used throughout whitehall as the general publishing format


README.md
```diff
@@ -2,6 +2,18 @@
 
 whitehall is a Ruby on Rails content management application for content published by government departments and agencies.
 
+## Running the Application
+
+Startup [using govuk-docker](https://github.com/alphagov/govuk-docker). Once running, `Whitehall` does not have an index, some suggested starting pages are below:
+
+- [`http://whitehall-admin.dev.gov.uk/government/admin/news/new`](http://whitehall-admin.dev.gov.uk/government/admin/news/new)
```
owenatgov, line 9: I wonder if it might be worth differentiating these 2 links between "publishing view" and "site view" or something similar.

Alternatively, you don't need the square brackets followed by the standard brackets for URLs. You can instead do:

```
<http://whitehall-admin.dev.gov.uk/government/admin/news/new>
```

That's unless you still want the back ticks for code view.

README.md
```diff
@@ -2,6 +2,18 @@
...
+- [`http://whitehall-frontend.dev.gov.uk/government/get-involved`](http://whitehall-frontend.dev.gov.uk/government/get-involved)
+
+Some pages will need local data to display
+
+- [Setup with credentials](https://docs.publishing.service.gov.uk/manual/get-started.html)
+- [Installing local data on `Whitehall`](https://github.com/alphagov/govuk-docker/blob/master/docs/how-tos.md#how-to-replicate-data-locally)
```
owenatgov, line 15: It's worth noting somewhere that:
- whitehall uses mysql
- in order to get local data you'll need permission to access AWS data, which _I think_ requires production clearance which itself requires 2 second line shifts
- maaaaybe explicitly clarifying that the 2nd point is key to getting the local data but you need 1 to do it

owenatgov (commented):
Couple of little comments. Really happy to see this going in.

README.md
chris-gds: Good points 🙌🏼  
I'll update to include all these - 🤔  debating if displaying link should formatting _as_ code or even if I should include this `http:` element

README.md
chris-gds: Updated 😁
Ah so I've been added as a `power-user` and can access the data in this context. There might be more nuance to this, i'll note _relevant_ permissions

README.md
```diff
@@ -2,6 +2,25 @@
...
+Site view:
+- <http://whitehall-frontend.dev.gov.uk/government/get-involved>
+
+
+Some pages will need data locally to display, whitehall uses mySQL. You'll need to gain relevant permissions to access data from AWS
+
+- [Get setup with AWS access](https://docs.publishing.service.gov.uk/manual/get-started.html)
+
+- Once completed [a guide to install local data on whitehall can be found here](https://github.com/alphagov/govuk-docker/blob/master/docs/how-tos.md#how-to-replicate-data-locally)
 ## Nomenclature
```
chris-gds, line 24: Noticed space, will amend / squash once happy with Content

README.md
```diff
@@ -2,6 +2,25 @@
...
 whitehall is a Ruby on Rails content management application for content published by government departments and agencies.
 
+## Running the Application
+
+Startup [using govuk-docker](https://github.com/alphagov/govuk-docker). 
+
+There are two different views within this App, a *Publishing view* and a *Site view*. Once running, whitehall does not have an index, some suggested starting pages are below:
+
+Publishing view:
+
```
owenatgov, line 12: nitpick hours: could we get rid of this space for consistency?

owenatgov (approved):
One final nitpick. Otherwise big thumbs up 👍

PR #6026, opened 2021-03-11
Replace the prev/next and document list patterns with components gem on "latest" view

owenatgov:
## What
Adds 2 components to the "latest" view:

- [The prev/next component](https://components.publishing.service.gov.uk/component-guide/previous_and_next_navigation)
- [The document list component](https://components.publishing.service.gov.uk/component-guide/document_list)

## Why

[Page developed against](https://www.gov.uk/government/latest?page=11&world_locations%5B%5D=china)

### Before
Screenshot 2021-03-11 at 16 00 09
Screenshot 2021-03-11 at 16 00 48

### After
Screenshot 2021-03-11 at 16 00 31
Screenshot 2021-03-11 at 16 01 06

### DIFF
--- a/app/views/kaminari/_paginator.html.erb
+++ b/app/views/kaminari/_paginator.html.erb
@@ -6,11 +6,25 @@
     remote:        data-remote
     paginator:     the paginator that renders the pagination tags inside
 -%>
-<%= paginator.render do -%>
-  <nav id="show-more-documents" role="navigation">
-    <ul class="previous-next-navigation">
-      <%= prev_page_tag unless current_page.first? %>
-      <%= next_page_tag unless current_page.last? %>
-    </ul>
-  </nav>
-<% end -%>
+
+<%
+  nav = {}
+
+  unless current_page.last?
+    nav[:next_page] = {
+      url: paginator.next_page_tag.url,
+      title: "Next page",
+      label: "#{current_page + 1} of #{total_pages}"
+    }
+  end
+
+  unless current_page.first?
+    nav[:previous_page] = {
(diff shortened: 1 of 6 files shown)


chris-gds:
Just need an integration link for this review

app/views/kaminari/_paginator.html.erb
```diff
@@ -6,11 +6,25 @@
...
-  </nav>
-<% end -%>
+
+<%
+  nav = {}
+
+  unless current_page.last?
+    nav[:next_page] = {
+      url: paginator.next_page_tag.url,
+      title: raw("Next <span class='visuallyhidden'>page</span>"),
```
chris-gds, line 16: 🤔 Based on gem use I'm going to suggest
`Next page` rather than  `Next <span class='visuallyhidden'>page</span>`

I'm not sure there's value of hiding this, I think maintaining the mismatch I think is unhelpful for sighted + AT users.

Also, I noticed this blink-and-you-miss-it extra hover style text-underline next to the arrow. That shouldn't be a blocker but would be good to remove as well if possible.

Mar-15-2021 18-54-43

app/views/kaminari/_paginator.html.erb
```diff
@@ -6,11 +6,25 @@
...
-  </nav>
-<% end -%>
+
+<%
+  nav = {}
+
+  unless current_page.last?
+    nav[:next_page] = {
+      url: paginator.next_page_tag.url,
+      title: "Next page",
```
chris-gds, line 16: 🙌🏼

chris-gds (approved):

PR #6027, opened 2021-03-11
Add edition attachments to fact check request

edwardkerry:
Editions sent to fact check only displayed the main document.

This change adds the edition's attachment partial so that any
attachments can also be easily fact-checked.

### Before
Screenshot 2021-03-11 at 17 26 52

-----
### After

Screenshot 2021-03-11 at 17 07 08
----
Screenshot 2021-03-11 at 17 07 47

[trello](https://trello.com/c/PJaC8aYE/2409-investigate-making-it-possible-to-preview-attachments-in-whitehall-publisher-timebox-2-days)


app/views/admin/fact_check_requests/show.html.erb
```diff
@@ -4,6 +4,7 @@
   <div class="col-md-8">
     <section class="document_page" id="<%= dom_id(@edition) %>">
       <%= render partial: 'admin/editions/edition', locals: { edition: @edition } %>
+      <%= render partial: 'admin/editions/show_attachments', locals: { edition: @edition } %>
```
ChrisBAshton, line 7: I'm not sure what effect this change has. Looking at the screenshots and description in the PR, I can see the tab changes in `edit.html.erb` make sense, but I'm not sure what _this_ change is. Could you clarify?

edwardkerry: I'll update the PR with a screenshot etc, but also answer here for clarity. On submitting a fact-check the user is presented with this partial with a message thanking them for submitting their review. It also displays the the document so it seemed appropriate to also present the attachments once more.

ChrisBAshton: Understood. It doesn't necessarily need a screenshot, but would be good to add that context to the commit message.

ChrisBAshton (approved):

ChrisBAshton (approved):

PR #6044, opened 2021-03-31
Update csv preview template to use components

owenatgov:
## What
Does the following to the CSV preview template ([example page](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/459824/LargeEmployers1415R02.csv/preview)):

- Applies the [govuk frontend grid system](https://design-system.service.gov.uk/styles/layout/#grid-system)
- Applies the [organisation logos component](https://components.publishing.service.gov.uk/component-guide/organisation_logo)
- Applies the [page title component](https://components.publishing.service.gov.uk/component-guide/title)
- Applies the [table component](https://components.publishing.service.gov.uk/component-guide/table)
- Makes the table navigable to keyboard and screenreader users (see [this PR](https://github.com/alphagov/datagovuk_find/pull/847) for why we would want to do this)
- Adds `govuk-body` and `govuk-link` to page elements where appropriate
- Tidies up the markup, styling and code

## Why
Part of ongoing work by the govuk accessibility team to ensure that we are consuming our components gem in as many places as possible to reduce the risk of hidden tech debt across our frontend apps.

[Card](https://trello.com/c/8uj38k8J/617-update-csv-preview-page)

## Visual changes
| Before | After |
| --- | --- |
| Screenshot 2021-03-31 at 10 28 23 | Screenshot 2021-03-31 at 10 28 33 |

### Tab navigation example
https://user-images.githubusercontent.com/64783893/113127557-cbb99000-9210-11eb-9efe-5b3b9831b165.mov

## Testing
There currently isn't an easy way to test this locally so I've been deploying to integration. Let me know when you want to do a test pass and I'll deploy to integration for you.


app/views/csv_preview/show.html.erb
```diff
@@ -1,70 +1,68 @@
...
-            <tr>
-              <% row.each do |data| %>
-                <td><%= data %></td>
-              <% end %>
-            </tr>
+<main role="main" id="content">
+  <header class="publication-header">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full">
+        <%= render "govuk_publishing_components/components/title", {
```
chris-gds, line 30: 👋🏼 Stumbled across this Whitehall page as part my heading update:
I'm thinking the the [Inverse Header](https://components.publishing.service.gov.uk/component-guide/inverse_header) component might be more apt here, however I don't think this "Download CSV" secondary link exists  on the current component iteration so the gem might need to be extended  (also, the font looks off on the gem page atm..)

owenatgov: I don't know if we'd necessarily need to add a link option like that. It's both very specific and mostly very basic markup (just a paragraph with a link inside and some extra styling for the situation). I can't think of any other places where this would be used. The inverse header though I will deffo sort out, good spot!

app/assets/stylesheets/frontend/views/attachments/_csv-preview.scss
```diff
@@ -0,0 +1,15 @@
+.csv-preview {
+  border: govuk-spacing(1) solid govuk-colour("light-grey");
```
chris-gds, line 2: Do we need this? Taking a look at the gem for tables + the Design System I'm thinking this is something bespoke and therefore questioning if it's needed.

chris-gds (approved):

chris-gds:
This [current link](https://assets.integration.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/459824/LargeEmployers1415R02.csv/preview) has reverted to the old version, I'm seeing the darker blue and bolded copy.  
 
The gem swaps are active and working - it's nice a few files of sass have been able to be removed as a result.

app/assets/stylesheets/frontend/views/attachments/_csv-preview.scss
```diff
@@ -0,0 +1,15 @@
...
+  margin-bottom: govuk-spacing(6);
+}
+
+.csv-preview__inner {
+  border: 1px solid govuk-colour("mid-grey");
+  padding: govuk-spacing(2) govuk-spacing(4) 0;
+  overflow: auto;
+
+  &:focus {
+    border: 3px solid govuk-colour("black");
```
chris-gds, line 12: could `@include govuk-focused-text;`be used?

app/assets/stylesheets/frontend/views/attachments/_csv-preview.scss
owenatgov: We don't need it necessarily. I primarily left it there to keep it inline with the existing design. Here's how it looks if I remove it:
Screenshot 2021-04-13 at 09 47 23
I personally prefer it being there as a little visual enhancement so my instinct is to keep it, but you're right that it's bespoke so it's a small risk.

app/assets/stylesheets/frontend/views/attachments/_csv-preview.scss
owenatgov: Not quite. Looks like this:
Screenshot 2021-04-13 at 10 19 45
I think the focus text include is intended for text elements and the like. I feel like it's a bit too intense for this use case.

For what it's worth, I copied this implementation from the design system docs website on their code blocks.

PR #6045, opened 2021-04-06
Add republish all rake task

maxgds:
This adds an extra rake task to bulk republish all whitehall documents.

I tested it on integration, output is here: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/232545/console


owenatgov (approved):
Look alright to me 👍  I'm not super familiar with our rake setup but it works in integration and I don't see any probs with the code.

PR #6056, opened 2021-04-08
Internationalise display_type methods

benjamineskola:
Use I18n for display_type methods, and fix tests to used translated texts as appropriate.

https://trello.com/c/8Y98ExuW/2456-normalise-how-the-document-type-is-presented-to-publishing-api-for-organisation-pages-5

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -153,12 +153,24 @@ en:
       draft_text:
         one: Draft text
         other: Draft texts
+      edition:
+        one: Edition
+        other: Editions
+      edition_with_appointment:
+        one: Edition with appointment
+        other: Editions with appointment
+      fatality_notice:
+        one: Fatality notice
+        other: Fatality notices
       foi_release:
         one: FOI release
         other: FOI releases
       form:
         one: Form
         other: Forms
+      generic_edition:
+        one: Generic edition
+        other: Generic editions
       guidance:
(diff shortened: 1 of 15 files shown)


config/locales/en.yml
```diff
@@ -153,12 +153,24 @@ en:
       draft_text:
         one: Draft text
         other: Draft texts
+      edition:
+        one: Edition
+        other: Editions
+      edition_with_appointment:
```
1pretz1, line 159: Do we need to add translations for these new locales?

1pretz1 (commented):
Looks good @benjamineskola, just a few small comments and one question.

config/locales/en.yml
benjamineskola: The changes to this file are all because those formats previously relied on `Edition`'s default behaviour; they now need a name explicitly declared. As I understand it, it'll fall back to the English translation if others don't exist, and so they should be no worse off than prior to this change (where the English name would have been returned in all cases).

Adding translated versions of these names will be part of [this card](https://trello.com/c/kYXvexWo/2455-add-missing-keys-to-locale-files-3).

1pretz1: Awesome, I was just checking the translations for these new locales would be picked up at some point. 👍

1pretz1 (approved):

PR #6057, opened 2021-04-09
Update contacts partial to use govuk components and styling

owenatgov:
## What
Updates the contacts partial on world organisation pages with the following changes:

- Replaces Whitehall's default grid styling with the [Design System grid](https://design-system.service.gov.uk/styles/layout/#grid-system)
- Replaces headings with the [heading component](https://components.publishing.service.gov.uk/component-guide/heading)
- Generally cleans up the markup and design to be closer in line with [other contact sections on govuk](https://www.gov.uk/government/organisations/department-of-health-and-social-care)

## Why
Part of ongoing work by the govuk accessibility team to ensure that we are consuming our components gem in as many places as possible to reduce the risk of hidden tech debt across our frontend apps.

[Card](https://trello.com/c/8uj38k8J/617-update-csv-preview-page)

[Page developed against](https://www.gov.uk/world/organisations/british-embassy-kabul)

## Visual changes
| Before | After |
| --- | --- |
| Screenshot 2021-04-09 at 13 57 32 | Screenshot 2021-04-09 at 13 57 50 |

### DIFF
--- a/app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
+++ b/app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
@@ -115,80 +115,6 @@
     }
   }
 
-  .contact-us {
-    .contact {
-      min-width: 0; // to overide something from static
-      @include core-16;
-      margin-bottom: $gutter;
-
-      @include media(desktop) {
-        float: left;
-        width: $one-third;
-      }
-
-      &.clear-contact {
-        clear: both;
-      }
-
-      .content {
-        padding: 0 $gutter-half;
-      }
-
-      h2 {
-        @include core-19;
-        font-weight: bold;
-      }
-
-      .comments {
-        margin: $gutter-one-third 0;
-      }
-
-      .adr {
-        margin: $gutter-one-third 0;
-      }
-
-      .email-url-number {
-        p {
-          margin: $gutter-one-third 0;
-        }
-
-        .type {
-          display: block;
-        }
-      }
-    }
-
-    .contact.main {
-
-      @include media(desktop) {
-        width: $full-width;
-        margin: 0 0 $gutter;
-      }
-
-      .content {
-        margin: 0 $gutter-half;
-        @extend %contain-floats;
-        padding: $gutter 0;
-        border-bottom: 1px solid $border-colour;
-
-        .comments {
-          @include media(desktop) {
-            float: right;
-            width: $one-third;
-          }
-        }
-
-        .adr,
-        .email-url-number {
-          @include media(desktop) {
-            float: left;
-            width: $one-third;
-          }
-        }
-      }
-    }
-  }
-
   .contact-us,
   .corporate-information {
     .keyline-header {
@@ -214,4 +140,13 @@
       }
     }
   }
+}
+
+.contact-section {
+  border-top: 1px solid $govuk-border-colour;
+  padding: govuk-spacing(6) 0;
+}
+
+.contact-section__address {
+  font-style: normal;
 }
--- a/app/views/contacts/_contact.html.erb
+++ b/app/views/contacts/_contact.html.erb
@@ -1,40 +1,39 @@
-<%
-  extra_class = []
-  if local_assigns.include?(:contact_counter) and contact_counter % 3 == 0
-    extra_class << "clear-contact"
-  end
-  extra_class << 'main' if local_assigns[:is_main]
-  extra_class << 'contact' unless contact.is_a? Contact
-  extra_class << 'postal-address' if contact.has_postal_address?
-%>
-<%= content_tag_for(:div, contact, class: extra_class.join(' ')) do %>
-  <div class="content" <%= lang if local_assigns[:lang] %> >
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
```diff
@@ -215,3 +141,12 @@
     }
   }
 }
+
+.contact-section {
+  border-top: 1px solid govuk-colour("mid-grey");
```
chris-gds, line 146: Can we use `$govuk-border-colour` here?

app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
```diff
@@ -215,3 +141,12 @@
...
+  padding: govuk-spacing(6) 0;
```
chris-gds, line 147: Ponder: you could use `govuk-!-padding-6` on the class - has me wondering which technique would be desirable first _as a rule_ 🤔

app/views/contacts/_contact.html.erb
```diff
@@ -1,59 +1,57 @@
...
+    <% end %>
 
-    <div class="vcard contact-inner">
-      <% unless local_assigns[:hide_title] %>
-      <p><%= contact.title %></p>
-      <% end %>
+    <div class="govuk-grid-row">
       <% if contact.has_postal_address? %>
-        <%= render_hcard_address(contact) %>
+        <address class="govuk-grid-column-one-third contact-section__address">
```
chris-gds, line 13: Nice `<address>` FTW.

app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
owenatgov: I'd rather put it here because it feels neater to me to output this as a shorthand style rule hidden in the CSS than the classes necessary to replicate it (I think `govuk-!-padding-6-bottom` and `govuk-!-padding-6-top`). It also feels more inline with the component concept to isolate the style here.

For what it's worth, the way I treat the helper classes is if I need one specific style or one specific bit of markup like a container div with some bottom margin. The same way might use a `strong` tag for example.

app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
owenatgov: Good idea. Done 👍

app/views/worldwide_organisations/show.html.erb
```diff
@@ -68,10 +68,12 @@
 <% end %>
 
 <% if @main_office %>
-  <section class="govuk-grid-row contact-us" id="contact-us">
-    <h2 class="keyline-header">
-      <%= t("worldwide_organisation.headings.contact_us" ) %>
-    </h2>
+  <section class="ontact-us" id="contact-us">
```
chris-gds, line 71: tiny typo - missing "c"  `class="ontact-us`

app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
```diff
@@ -215,3 +141,12 @@
     }
   }
 }
+
+.contact-section {
+  border-top: 1px solid $govuk-border-colour;
```
chris-gds, line 146: [`<hr class="govuk-section-break govuk-section-break--visible">`](https://design-system.service.gov.uk/styles/typography/) seems more apt here, if at all actually, this line is pretty random on this page. No action required, just a note.

app/views/contacts/_contact.html.erb
```diff
@@ -1,59 +1,57 @@
...
       <% end %>
     </div>
-  </div>
+
+    <% if contact.is_a?(WorldwideOffice) && contact.access_and_opening_times_body.present? %>
+      <%
+        fallback = t_fallback('contact.access_and_opening_times')
+        lang = fallback if fallback && fallback != I18n.locale
+      %>
+      <%= link_to t('contact.access_and_opening_times'), [contact.worldwide_organisation, contact], class: "govuk-link", lang: lang %>
```
chris-gds, line 55: I feel like [`<time>`](http://whitehall-frontend.dev.gov.uk/world/organisations/british-embassy-kabul) could be used on this section for the opening hours. 
However this is probably true everywhere on the site...

Capturing this for later.

chris-gds (approved):
The whole page in general needs work, but this update looks good eps on mobile. Once tiny typo to resolve. Captured the `<time>` comment elsewhere

app/assets/stylesheets/frontend/views/_worldwide-organisations.scss
owenatgov: I went for the boarder because this is the convention we normally follow on pages like the collections example in the PR description. However I think you make a good point that this feels like more suitable markup. I've made a note to start a thread on the govuk-frontenders channel at the very least to get input on this.

app/views/contacts/_contact.html.erb
owenatgov: Something worth noting is that this content isn't the literal opening times but the link to opening times so the time tag wouldn't be appropriate here. All the same I'd forgotten about this tag so cheers for flagging!

PR #6060, opened 2021-04-12
Fix header section for topical events

alex-ju:
- Use components for rendering title and metadata. The current template renders `<span class="translation_missing" title="translation missing: en.document.headings.organisations, count: 3">Organisations</span>`
 - Remove `<aside>` as it must not be contained in another landmark (`<main>` in this case)
 - Refactor `<header>` layout to use `govuk-frontend` grid

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

topical-events-before

</td><td valign="top">

topical-events-after

</td></tr>
</table>

### DIFF
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -17,17 +17,11 @@
         </div>
       <% end %>
       <% if @topical_event.organisations.any? %>
-        <aside class="meta metadata-list">
-          <div class="inner-heading">
-            <dl>
-              <dt><%= t('document.headings.organisations', count: @topical_event.organisations.length) %>:</dt>
-              <dd>
-                <%= render  partial: 'organisations/organisations_name_list',
-                            locals: { organisations: @topical_event.organisations } %>
-              </dd>
-            </dl>
-          </div>
-        </aside>
+        <%= render "govuk_publishing_components/components/metadata", {
+          other: {
+            "Organisations": sanitize(array_of_links_to_organisations(@topical_event.organisations).to_sentence)
+          }
+        } %>
       <% end %>
     </div>
   </header>
--- a/app/views/topical_events/show.html.erb
+++ b/app/views/topical_events/show.html.erb
@@ -4,25 +4,29 @@
 
 <%= content_tag_for(:div, @topical_event, class: "classification #{@topical_event.class.name.underscore}") do %>
 
-  <header class="block headings-block">
-    <div class="inner-block floated-children">
-      <%= render partial: 'shared/heading',
-                locals: { heading: "#{h(@topical_event.name)}#{' <span class="archived">(Archived)</span>' if @topical_event.archived?}".html_safe,
-                          big: true, extra: true } %>
-      <% if @topical_event.logo_url.present? %>
-        <div class="heading-extra topical-event-logo">
-          <div class="heading-inner">
-            <%= image_tag(@topical_event.logo_url(:s300), alt: @topical_event.logo_alt_text) if @topical_event.logo_url %>
-          </div>
+  <header class="headings-block">
+    <div class="inner-block">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <%= render "govuk_publishing_components/components/title", {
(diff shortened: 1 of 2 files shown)


app/views/topical_events/show.html.erb
```diff
@@ -4,31 +4,29 @@
...
-      <%= render partial: 'shared/heading',
-                locals: { heading: "#{h(@topical_event.name)}#{' <span class="archived">(Archived)</span>' if @topical_event.archived?}".html_safe,
-                          big: true, extra: true } %>
-      <% if @topical_event.logo_url.present? %>
-        <div class="heading-extra topical-event-logo">
-          <div class="heading-inner">
-            <%= image_tag(@topical_event.logo_url(:s300), alt: @topical_event.logo_alt_text) if @topical_event.logo_url %>
-          </div>
+  <header class="headings-block">
+    <div class="inner-block">
```
owenatgov, line 8: This is part of whitehall's gross custom grid system. Is it still necessary? Could it be replaced with `govuk-width-container` or removed altogether?

app/views/topical_events/show.html.erb
```diff
@@ -4,31 +4,29 @@
...
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <%= render "govuk_publishing_components/components/title", {
+            title: "#{h(@topical_event.name)}#{' <span class="archived">(Archived)</span>' if @topical_event.archived?}".html_safe
```
owenatgov, line 12: What is that `archived` span doing?

app/views/topical_events/show.html.erb
```diff
@@ -4,31 +4,29 @@
...
-                <%= render  partial: 'organisations/organisations_name_list',
-                            locals: { organisations: @topical_event.organisations } %>
-              </dd>
-            </dl>
-          </div>
-        </aside>
-      <% end %>
+        <div class="govuk-grid-column-one-third">
+          <% if @topical_event.logo_url.present? %>
+            <div class="topical-event-logo">
```
owenatgov, line 24: I can't see this so I may be making annoying assumptions, but could this be replaced with the [org logo component](https://components.publishing.service.gov.uk/component-guide/organisation_logo)?

owenatgov (commented):
A couple of Q's about classes. Cheers for doing this Alex!

app/views/topical_events/show.html.erb
alex-ju: We still need these until the main container with `class="whitehall-content"` will be replaced with `class="govuk-width-container"`. It's sort of a hybrid approach that would help us transition to the new layout system, otherwise it will require all layouts to be changed at once.

app/views/topical_events/show.html.erb
alex-ju: It seems to be a thing we have in Whitehall to label archived events: https://www.gov.uk/government/topical-events/budget-2019

app/views/topical_events/show.html.erb
alex-ju: The logos here more of an [event-related featured image](https://www.gov.uk/government/topical-events/cop26) rather than an organisation logo

alex-ju: The class itself sets some margin at the top, to align with the page title and makes sure it doesn't overflow on narrow screens.

app/views/topical_events/show.html.erb
owenatgov: I should've looked up the styling before commenting! It makes it grey I see, that's alright then, feel free to resolve.

app/views/topical_events/show.html.erb
owenatgov: Gotcha, feel free to resolve.

app/views/topical_events/show.html.erb
owenatgov: Makes sense, bloody whitehall. Resolvable as well.

owenatgov (approved):
My comments have been addressed so I'm happy to approve 👍

PR #6061, opened 2021-04-14
Adds temp rake task to migrate Whitehall docs

1pretz1:
Migration of a number of documents from GDS to CDDO. The attached CSV
was provided externally and contains the slug, lead orgs, supporting
orgs and 'document type' for each document. The ordering of the
organisations is important as it dictates the order in the frontend and
so this has been taken into consideration.

The rake task turned out to be a little bespoke for the given CSV
unfortunately due to a few different edge cases but as it's only a
temporary task I think this is fine.

Test run:
https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/234442/console

Trello:
https://trello.com/c/pdTmWCJ2/2464-content-tagging-for-gds-mog-change-cddo-whitehall-content


1pretz1:
Closed in favour of: https://github.com/alphagov/whitehall/pull/6063

PR #6064, opened 2021-04-14
Remove the coming soon template and references to it

owenatgov:
## What
Removes the coming soon template, which was originally intended to display when content was scheduled but not yet live, and the reference to it in the document controller.

## Why
This functionality is no longer supported by whitehall. It is suspected that this was missed in a project to remove this functionality in 2018:

- https://github.com/alphagov/whitehall/pull/4394
- https://github.com/alphagov/whitehall/pull/4851

[Card](https://trello.com/c/z3ZEAEll/619-update-coming-soon-template)

No visual changes


maxgds:
What is here looks fine to me, but I'm unclear on whether we should also remove these references to scheduled_for_publication_as
image

Can you get a steer from a backend dev to confirm if this is also as useful deletion or would break things?

sihugh (approved):
Seems fine 👍 . If you're worried, you could always put it on staging and use traffic replay to see if anything starts erroring?

sihugh:
> Can you get a steer from a backend dev to confirm if this is also as useful deletion or would break things?

I think that's used for scheduled publication, which is still a thing 👍

PR #6065, opened 2021-04-14
Update unpublished template

chris-gds:
## What?

Updating an unpublished template to more closely align with gem and Design System

## Why?

Align architecture, improve consistency, improve a11y.

## Visuals
(Before > After)

Screenshot 2021-04-14 at 11 39 25

Screenshot 2021-04-14 at 11 38 50

## How to test
[`government/organisations/foreign-Commonwealth-office/about/procurement/`](https://www.gov.uk/government/organisations/foreign-Commonwealth-office/about/procurement)

## Anything else?

There is a wider refactor on the parent container of `Whitehall` that is needed to [include structure](https://design-system.service.gov.uk/styles/page-template/) like: `.govuk-main-wrapper`  that has been omitted (but is needed) from this PR as it's out of scope / has wider implications as it touches slimmer

Based on PR conversations the `govspeak (gem)` accepting Markdown has an established pattern that's been adopted for the context of this PR however a deeper Frontend conversation about the ideal pattern could help for future as it appears in multiple places in various repos.


app/views/documents/unpublished.html.erb
```diff
@@ -1,19 +1,18 @@
-<% page_title "No longer available" %>
+<% page_title "The page you're looking for is no longer available" %>
```
owenatgov, line 1: Nice oooooone

app/views/documents/unpublished.html.erb
```diff
@@ -1,19 +1,18 @@
...
+      } %>
+      <p class="govuk-body">
+        The information on this page has been removed because <%= @unpublishing.reason_as_sentence %>.
+      </p>
       <% if @unpublishing.explanation.present? %>
         <%= govspeak_to_html @unpublishing.explanation %>
       <% end %>
-
       <% if @unpublishing.alternative_url.present? %>
         <p>Visit: <%= link_to @unpublishing.alternative_url, @unpublishing.alternative_url %></p>
```
owenatgov, line 17: Nitpick: could we also put `govuk-body` here?
```suggestion
        <p class="govuk-body">Visit: <%= link_to @unpublishing.alternative_url, @unpublishing.alternative_url %></p>
```

app/views/documents/unpublished.html.erb
```diff
@@ -1,19 +1,18 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/title", {
+        title: "The page you're looking for is no longer available",
+        margin_bottom: 4
+      } %>
+      <p class="govuk-body">
+        The information on this page has been removed because <%= @unpublishing.reason_as_sentence %>.
+      </p>
       <% if @unpublishing.explanation.present? %>
         <%= govspeak_to_html @unpublishing.explanation %>
```
owenatgov, line 14: I can't tell if this is govspeak or not. Would you mind having a go at applying the govspeak component and see how it turns out?
```suggestion
        <%= render "govuk_publishing_components/components/govspeak", {} do %>
            @unpublishing.explanation
        <% end %>
```

owenatgov (commented):
One weeny comment and one medium comment.

app/views/documents/unpublished.html.erb
chris-gds: So, this outputs pure markdown (below) - the [govspeak component](https://components.publishing.service.gov.uk/component-guide/govspeak) accepts HTML. There to be a [helper functions](
https://github.com/alphagov/collections-publisher/blob/a4307156e4ad8f98624b50813c9fde75894e4074/app/helpers/application_helper.rb#L34) in use in other repo's called `markdown_to_html` & `render_markdown` that does what I expect, but in a [different way](https://github.com/alphagov/collections-publisher/blob/e3a39576729ec197cc8886ec347fa53a3839aff6/app/views/step_by_step_pages/guidance.html.erb#L42).

I think  this is a good case to extend the gem to move this helper functions so that govspeak can accept markdown (unless I misunderstand this and it already does?)

What do you think?

Using this existing setup (adding a new helper function to Whitehall), this works:

  ``` 
  <p class="govuk-body">
    <%= render_markdown(@unpublishing.explanation) %>
  </p>
 ```

However ideally you want something like:

```  
<%= render "govuk_publishing_components/components/govspeak", {
} do %>
  render_markdown: @unpublishing.explanation
<% end %>
```

----
Markdown Output:
`The Foreign and Commonwealth Office (FCO) was replaced by the Foreign, Commonwealth & Development Office (FCDO) in September 2020. This FCO page is no longer current. Instead, see [FCDO procurement](https://www.gov.uk/government/organisations/foreign-commonwealth-development-office/about/procurement).`

chris-gds: 👋🏼 @DilwoarH I understand that you might have raised something similar about extending `govspeak` previously? Do you remember the outcome of the conversation?

(Other considerations re this idea is that impact on performance that needs investigation)

Cc @alex-ju

owenatgov: Before we get into this, I don't know if it's worth stalling this work to discuss extending our govspeak component. My loose understanding is that said govspeak component applies design system styling to the mostly plain html spat out by govspeak, so extending it would involve importing the govspeak compiler into the gem which feels quite weighty to me, especially if there's only this use case (?). I'd like to discuss it at the next govuk frontenders catchup if that's ok. Having said that, if you still have some thoughts r Alex/Dilwoar you want to weigh in I'm not against talking about it on the PR.

chris-gds: 100% to all of this, it's out of scope of this ticket and worth separating out and having a wider chat.

So for context - I was speaking to @injms yesterday and it was mentioned this idea was brought up previously, so was curious to understand a little more of the outcome of this.

My preferred approach in context of this ticket would be to leave as is or use the existing implementation found in other repos and simply capture the `govspeak` idea for a later date.

kevindew: Wading in... I'm afraid govuk_publishing_components cannot depend on govspeak as this creates a cyclic dependency - govspeak depends on govuk_publishing_components as some of govspeak's output is components.

A common convention in apps is to have a helper function like: https://github.com/alphagov/content-publisher/blob/5df3739a75fea7dd00d0afbf415135dfb3800318/app/helpers/application_helper.rb#L6-L10 for rendering govspeak (a note that in that example I gave the function uses `sanitize: false` as it was intended for trusted input, you'd not have trusted input here).

chris-gds: Thanks for this @kevindew, 🤔  I'm not sure I entirely follow why the cyclic dependency prevents the gem from accepting markdown as an input.  However that snippet you shared works as expected for the `govspeak (gem)` component to accept markdown without any additional changes. 

I'm wondering if this is more a convention convo, as I would expect this to be present in the view

```  
<%= render "govuk_publishing_components/components/govspeak", {
  content: raw(Govspeak::Document.new(@unpublishing.explanation, sanitize: true).to_html)
  } do %>
<% end %>
```

I'm thinking if this is OK? 👆🏼 🤔  then it's just updating the `govspeak (gem)` docs to reference if the input is markdown.

owenatgov (approved):
Noyce

PR #6072, opened 2021-04-19
Remove manual redirect and statistics markup

owenatgov:
## What
- Remove the conditional in the manual redirect written into the statistics controller
- remove all markup templates associated with statistics in whitehall

## Why
The statistics routes have been redirected however locales have not. Eg:

- https://www.gov.uk/government/statistics.uk
- https://www.gov.uk/government/statistics.bg
- https://www.gov.uk/government/statistics.es-419

These locale pages are behaving a bit weirdly when following the [usual redirect method](https://docs.publishing.service.gov.uk/manual/redirect-routes.html#manually-apply-a-redirect-from-any-publishing-app) so removing the locale conditional from the controller to ensure that the stats page and all its locales redirect to [the research and statistics search page](https://www.gov.uk/search/research-and-statistics) gets the work done at least whilst we work out why these redirects are being weird.

### Why are we redirecting non-English pages to an English only template?
Almost all these translated stats pages have no statistics on them and there are also almost no avenues for users to actually reach them. I believe Welsh is the one exception but there are still very few views on this template (20 views in the last 2 weeks, including myself and team members).

The additional points here are that non-English finders are due to be retired anyway and there's a much wider issue around how non-English speaking users find non-English content on govuk which isn't going to be solved by us not redirecting these templates.

[Card](https://trello.com/c/Qa5BazvX/439-inconsistently-placed-translation-navigation)

No visual changes.


thomasleese (commented):
Code changes seem fine, very happy to see this sort of code out of Whitehall ([original tech debt card for the English finders](https://trello.com/c/xu0u9Igw/97-migrate-whitehall-finders)).

It's probably worth adding a bit more information to the commit messages, specifically around why we're making the change and the justification for it.

The finder that we'll be redirecting the users to only supports English content (because it's powered by [Search API rather than Whitehall](https://trello.com/c/jJnJp1t7/127-rummager-accepts-content-that-is-not-english)), so by removing this redirect, we are taking away the ability for users to see research and statistics published in other languages. I assume this product decision was acceptable hence the PR, but I think it's worth making a point of that.

owenatgov:
@thomasleese I've updated the commits with some context.

Regarding the comment on a non-English page going to an English page, I have some product insight here. Almost all these translated stats pages have no statistics on them and there are also almost no ways to actually reach these pages anyway. I believe Welsh is the one exception but there are still very few views on this template (20 views in the last 2 weeks, including myself and team members looking at the pages).

The additional points here are that non-English finders are due to be retired anyway and there's a much wider issue around how non-English speaking users find non-English content on govuk which isn't going to be solved by us not redirecting these templates.

I'll updated the PR description with the above.

thomasleese (approved):

PR #6075, opened 2021-04-22
Bump mail-notify gem

ChrisBAshton:
The bump of this gem removes a constraint on Rails versions that
prevents dependabot from upgrading Rails. [1]

[1]: https://github.com/dxw/mail-notify/pull/26


kevindew:
Unfortunately Tom beat you to this: https://github.com/alphagov/whitehall/pull/6074

PR #6087, opened 2021-04-27
Remove unused files

chris-gds:
## What?

Removing unused files from Whitehall 

⚠️ Please leave on staging to ensure that testing / logs indicates that this confident to deploy.

## Why?

Part on-going component swaps to align architecture / improve a11y.

## Anything else?

Some files under `/documents` are not in use anymore so have been removed.


README.md
```diff
@@ -20,6 +20,16 @@ Some pages will need data locally to display, whitehall uses mySQL. You'll need
 
 - Once completed [a guide to install local data on whitehall can be found here](https://github.com/alphagov/govuk-docker/blob/master/docs/how-tos.md#how-to-replicate-data-locally)
 
+## Testing on intergration
```
maxgds, line 23: This is useful information but it's not part of removing the old files so it should go in separately. Also, I think it's best to add the instructions on deploying a branch to integration to the tech docs, assuming they aren't already there, because they could be useful for other apps/reasons. From Whitehall I'd reduce this down to a line about some pages not rendering locally, eg "Some pages do not render in the dev environment, in which case it is best to deploy changes to view and test on Integration (link to tech docs)"

maxgds (changes requested):
Deletions are good to go but I recommend reworking the readme add addressing it separately.

README.md
chris-gds: Salient point, this should live in tech docs. I've dropped this from this PR to update separately.

owenatgov:
All these deletions are fine by me but just a note on `classifications/_document_list`, I was able to delete it fine here https://github.com/alphagov/whitehall/pull/6090. I feel pretty confident about this  Does it cause test failures when it's deleted on this PR? Is there a concern that it's being secretly used somewhere?

chris-gds:
Thanks for raising [#6090](https://github.com/alphagov/whitehall/pull/6090) 🙌🏼

> I feel pretty confident about this Does it cause test failures when it's deleted on this PR? 

Ace, not as far as I'm aware.

>Is there a concern that it's being secretly used somewhere?

Yes, this was an initially a concern it was being as it was highlighted as in use however the code didn't reflect this nor could I see the suggested pages rendered.  It was suggested was the logging work would offer insight - however equally #6090 is another approach.

maxgds (approved):

PR #6091, opened 2021-04-30
Replace shared heading

chris-gds:
## What?

- Replace shared heading template on  `fields-of-operation` &  `world-location`
- Refactor  `fields-of-operation` pages to use gem
  - Headings
  - Page Title
  - Org logo
  - Content List
  
## Why?

As part of on-going work to align architecture to gem to improve consistency and in turn a11y.  This template is doing the same function as the [title component](https://components.publishing.service.gov.uk/component-guide/title).  

## Visuals 
(Before > After)

Screenshot 2021-05-11 at 17 40 39

Screenshot 2021-05-11 at 17 40 06

Screenshot 2021-05-11 at 17 41 31

Screenshot 2021-05-10 at 10 10 21

Outdated:
Screenshot 2021-04-30 at 17 50 42

Screenshot 2021-04-30 at 17 50 20

## How to test:

http://www.gov.uk/world/australia/news
http://www.gov.uk/government/fields-of-operation/
http://www.gov.uk/government/fields-of-operation/iraq

## Anything else?

The _Fatalities_ section originally was marked as a Headings but these did not visually match headings causing an a11y issue.
These have been changed been changed to `p` as this more closely aligns with the Content. Also, as the outer elements are list this allows navigation for screen readers.  This section could be resigned.

image

Design System has been used instead of 
- List 
- Content List.

To use the [List Component](https://components.publishing.service.gov.uk/component-guide/list) in this context requires to pass an ID to the `<li>` element. Whitehall uses the IDs in routing and this within the tests.  The Gem does not support this at present so implementation is keeping the existing but applying the Design System when possible.

[Draft PR for extending list](https://github.com/alphagov/govuk_publishing_components/pull/2059)

This also applied to using the [Contents List](https://components.publishing.service.gov.uk/component-guide/contents_list) on the the index page on `fields-of-operation`

This is what the desirable fix was originally until this was reverted to what is seen above due to the same issue of passing an ID to a `<li>` element.

(Before > Proposed After - not implemented)
Screenshot 2021-05-08 at 19 15 24


app/views/operational_fields/show.html.erb
```diff
@@ -1,28 +1,25 @@
...
-          <%= content_tag_for(:div, @organisation,
-            class: organisation_brand_colour_class(@organisation)) do %>
-            <%= link_to @organisation,
-            class: logo_classes(organisation: @organisation, stacked: true, use_identity: true) do %>
-              <span><%= organisation_logo_name(@organisation) %></span>
-            <% end %>
-          <% end %>
-        </div>
-      </div>
+<div class="govuk-width-container">
```
owenatgov, line 4: I think this element could still be the template header.

app/views/operational_fields/show.html.erb
```diff
@@ -1,28 +1,25 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= render "govuk_publishing_components/components/title", {
+        context: "British fatalities",
+        title: "Operations in #{@operational_field.name}",
+      } %>
+    </div>
+    <div class="govuk-grid-column-one-third">
+      <%= content_tag_for(:div, @organisation,
```
owenatgov, line 13: Out of scope for the ticket but this could possibly be an [org logo component](https://components.publishing.service.gov.uk/component-guide/organisation_logo). It's a bit of a tricky component so no worries if it turns into a bother.

app/views/world_location_news/index.html.erb
```diff
@@ -3,21 +3,27 @@
...
+        <%= render "govuk_publishing_components/components/title", {
+          context: "World location news",
+          title: @world_location.title,
+        } %>
+      </div>
+      <div class="govuk-grid-column-one-third">
+        <div class="govuk-!-margin-top-8 govuk-!-margin-bottom-8">
+          <aside class="heading-extra">
+            <div class="inner-heading">
+              <%= render 'shared/available_languages', object: @world_location %>
```
owenatgov, line 18: Another out of scope comment: It looks like this is rendering a custom translation nav as well as scoping the decision to render it to that component, which I personally don't like as it makes it harder to understand at a glance if the component is rendering or not. Could we replace this with a [translation nav component](https://components.publishing.service.gov.uk/component-guide/translation_nav)?

app/views/world_location_news/index.html.erb
```diff
@@ -3,21 +3,27 @@
...
+      <div class="govuk-grid-column-two-thirds">
+        <%= render "govuk_publishing_components/components/title", {
+          context: "World location news",
+          title: @world_location.title,
+        } %>
+      </div>
+      <div class="govuk-grid-column-one-third">
+        <div class="govuk-!-margin-top-8 govuk-!-margin-bottom-8">
+          <aside class="heading-extra">
+            <div class="inner-heading">
```
owenatgov, line 17: Can we get rid of this?

app/views/world_locations/show.html.erb
```diff
@@ -7,20 +7,24 @@
...
-                locals: { type: t_world_location(@world_location),
-                          heading: @world_location.title,
-                          extra: true, big: true } %>
-      <aside class="heading-extra">
-        <div class="inner-heading">
-          <%= render 'shared/available_languages', object: @world_location %>
-          <%= render 'shared/featured_links', links: @world_location.featured_links.only_the_initial_set %>
-        </div>
-      </aside>
+  <div class="govuk-width-container">
```
owenatgov, line 10: Header again

app/views/world_locations/show.html.erb
```diff
@@ -7,20 +7,24 @@
...
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <%= render "govuk_publishing_components/components/title", {
+          context: t_world_location(@world_location),
+          title: @world_location.title,
+        } %>
+      </div>
+      <div class="govuk-grid-column-one-third govuk-!-margin-top-8 govuk-!-margin-bottom-8">
+        <aside class="heading-extra">
```
owenatgov, line 19: Gunky classes again

owenatgov (commented):
Few little comments and queries, some can be rebuffed as they might be out of scope.

app/views/operational_fields/show.html.erb
chris-gds: Sorry 🙈    what do you mean? 

This element was seemingly not ideal as a `<header>`&  it's not really reflected anywhere else (that I can see on GOV.UK) and having multiple headers in this context when there is a `<header role="banner" id="global-header"` element didn't really align with the use of this element?

app/views/operational_fields/show.html.erb
chris-gds: Agreed. I'm mindful to keep this in scope - I have captured though.

app/views/world_locations/show.html.erb
chris-gds: I do agree the sentiment, it's actively keeping the existing for scope.🤔 

Let me take these on board and expand the scope of this ticket a bit to adjust for org logo and a wider refactor

app/views/operational_fields/show.html.erb
owenatgov: Doh, you're right we already have a bloody heading! Feel free to resolve this and the other heading comments.

app/views/world_locations/show.html.erb
chris-gds: Again, this world location has an `API error` - leaving this one for now.

owenatgov (approved):
Good job 👍  I reckon these pages need some more general attention but that's a topic for another day. Sorry you had to duck and dive through whitehall spaghetti for my sake.

chris-gds:
> Good job 👍 I reckon these pages need some more general attention but that's a topic for another day.

I think these might even be incoming as part of redesign.

> Sorry you had to duck and dive through whitehall spaghetti for my sake.

Not at all - it's all part of the same umbrella in the context of this epic

PR #6105, opened 2021-05-12
Update README

chris-gds:
## What

Adding some troubleshooting steps should a user face the same issues.

## Why

I've run into these issues a couple of times while working on [this PR](https://github.com/alphagov/whitehall/pull/6147) - felt it would be valuable to add to README

## Anything else

Open to feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,17 @@
 # whitehall
 
+### Running the Application
+
+**Use GOV.UK Docker to run any commands that follow.**
+
 Whitehall is deployed in two modes:
 
-- 'admin' for publishers to create and manage content; and
-- 'frontend' for rendering some content under https://www.gov.uk/government and https://www.gov.uk/world.
+- 'admin' for publishers to create and manage content (e.g. <http://whitehall-admin.dev.gov.uk/government/admin/news/new>)
+- 'frontend' for rendering content (eg: <http://www.gov.uk/government/get-involved>, <https://www.gov.uk/world>)
+
+Traditionally, the two sides of Whitehall are available on different domains in development, which reflect their counterparts in production:
+
+While this usually results in different routing behaviour, in development [all routes can be accessed from either domain](https://github.com/alphagov/whitehall/blob/main/config/routes.rb#L3-L5), although [the redirect behaviour may differ](https://github.com/alphagov/whitehall/blob/main/config/routes.rb#L25-L28).
 
 ## Nomenclature
 
@@ -14,19 +22,6 @@ Whitehall is deployed in two modes:
 This is a Ruby on Rails app, and should follow [our Rails app conventions](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html).
 
 You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Follow [the usage instructions](https://github.com/alphagov/govuk-docker#usage) to get started.
-
-**Use GOV.UK Docker to run any commands that follow.**
-
-### Running the Application
-
-Traditionally, the two sides of Whitehall are available on different domains in development, which reflect their counterparts in production:
-
-- admin side e.g. <http://whitehall-admin.dev.gov.uk/government/admin/news/new>
-
-- frontend side e.g. <http://whitehall-frontend.dev.gov.uk/government/get-involved>
-
-While this usually results in different routing behaviour, in development [all routes can be accessed from either domain](https://github.com/alphagov/whitehall/blob/530abc13018145a6efe6ab4a19f6210254e2e304/config/routes.rb#L3-L5), although [the redirect behaviour may differ](https://github.com/alphagov/whitehall/blob/530abc13018145a6efe6ab4a19f6210254e2e304/config/routes.rb#L25-L28).
-
 ### Shared mustache templates
 
 The shared mustache templates must be compiled for JavaScript and functional tests to pass.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,9 +1,17 @@
...
 
+### Running the Application
+
+**Use GOV.UK Docker to run any commands that follow.**
+
 Whitehall is deployed in two modes:
 
-- 'admin' for publishers to create and manage content; and
-- 'frontend' for rendering some content under https://www.gov.uk/government and https://www.gov.uk/world.
+- 'admin' for publishers to create and manage content (e.g. <http://whitehall-admin.dev.gov.uk/government/admin/news/new>) and
```
owenatgov, line 9: Nitpick time: I know you didn't add this but I have a personal vendetta against that "and". It's unnecessary in a list. Would you mind getting rid of it?

README.md
```diff
@@ -1,9 +1,17 @@
...
+- 'frontend' for rendering some content under https://www.gov.uk/government, https://www.gov.uk/world & http://www.gov.uk/government/get-involved
```
owenatgov, line 10: Bit of a personal prefs this here:

```suggestion
- 'frontend' for rendering some content under https://www.gov.uk/government (eg: <http://www.gov.uk/government/get-involved>) and https://www.gov.uk/world
```

README.md
```diff
@@ -1,9 +1,17 @@
...
+
+Traditionally, the two sides of Whitehall are available on different domains in development, which reflect their counterparts in production:
+
+While this usually results in different routing behaviour, in development [all routes can be accessed from either domain](https://github.com/alphagov/whitehall/blob/530abc13018145a6efe6ab4a19f6210254e2e304/config/routes.rb#L3-L5), although [the redirect behaviour may differ](https://github.com/alphagov/whitehall/blob/530abc13018145a6efe6ab4a19f6210254e2e304/config/routes.rb#L25-L28).
```
owenatgov, line 14: These links tie these lines to a specific git sha which cretes risk of those files change downstream. Could we change this to reference `main` instead? Eg: https://github.com/alphagov/whitehall/blob/main/config/routes.rb#L3-L5

docs/troubleshooting.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+Mysql2::Error: Access denied for user 'whitehall'@'localhost' (using password: YES)
+```
+
+Solution:
+Login, create user, give access & quit
+
+```
+mysql -u root -p
```
owenatgov, line 15: I'm assuming you'd run these on the docker image? Could we at least reference instructions for how to get into the image at the top?

docs/troubleshooting.md
```diff
@@ -0,0 +1,59 @@
...
+`gem update --system`
+
+Problem:
+
+```
+API Error
+```
+
+Solution:
+[`search-api`](https://github.com/alphagov/search-api), [`static`](https://github.com/alphagov/static) are helpful to have to running in parallel (depending on what page you are on) via [govuk-docker](https://github.com/alphagov/govuk-docker)
```
owenatgov, line 48: I know what you're referring to here but could you unpack it for newbies please? Is there a page or a line in the govuk-docker docs which unpack this? All you'd really need is something that says "make sure the referenced apps are up to date as well".

owenatgov (commented):
A few errant comments. Looking alright :)

docs/troubleshooting.md
chris-gds: Good shout

docs/troubleshooting.md
```diff
@@ -0,0 +1,59 @@
+## Whitelhall troubleshooting
```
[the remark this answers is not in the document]
chris-gds: Course! Sorted - bonus points please.

docs/troubleshooting.md
chris-gds: Added the key sample commands

README.md
chris-gds: Makes sense, added

docs/troubleshooting.md
owenatgov: So... I'm expecting there to be some sort of `exec`ing into the docker image or maybe a `govuk-docker-run` or something as part of these commands. Surely running `mysql` locally is something that we don't want to encourage since it's all in the docker container?

README.md
owenatgov: It's still there!

chris-gds: 🤔  Ah there was two "and"s
I've removed on both lines and also rewritten that line a little 

```- 'frontend' for rendering some content under https://www.gov.uk/government and https://www.gov.uk/world.```

`/government` doesn't appear to be an active URL (it directs to the home page), also the `some` felt redundant.

```- 'frontend' for rendering content (eg: <http://www.gov.uk/government/get-involved>, <https://www.gov.uk/world>)```

chris-gds:
@owenatgov ah I think I need a re-review.
(I'll look at the test a bit later, not sure why this would be failing)

chris-gds:
@owenatgov this ok to go in?

owenatgov (approved):
Yes!

chris-gds:
`You’re not authorized to merge this pull request.` 😮‍💨  @maxgds do the honours?

kevindew:
@chris-gds @maxgds It looks like this has drifted this README away from the conventional format: https://docs.publishing.service.gov.uk/manual/readmes.html and has introduced a h3 (Running the Application) directly below a h1 (whitehall).

Would you be able to amend this so that it complies with the convention? A bunch of work has gone into making these [READMEs consistent](https://github.com/alphagov/whitehall/pull/6115) so it'd be nice to try keep them that way. You can chat in #govuk-dev-tools if any of the conventions don't make sense.

chris-gds:
>Would you be able to amend this so that it complies with the convention? 

Sure - i'll take a re-look and add raise a change

PR #6106, opened 2021-05-13
Add styles for `govuk-select` class used in world location

alex-ju:
Using the component class at a time where all style components were imported in all applications resulted later in an unstyled element.


alex-ju:
This was addressed in the meantime.

PR #6107, opened 2021-05-14
Add missing keys

1pretz1:
Normalises and adds missing locales.

More info in the commits ->

---

Trello:
https://trello.com/c/kYXvexWo/2455-add-missing-keys-to-locale-files-3


config/locales/ar.yml
```diff
@@ -212,6 +227,18 @@ ar:
         other: مسودة النصوص
         two:
         zero:
+      edition:
```
brucebolt, line 230: I think the pluralisation needs updating here, to reflect the plurals used in each language (e.g. Arabic has `one`, `many`, `few`, `other`, `two` and `zero` that will need placeholder keys).

brucebolt (commented):
Great work!

I've added a comment about plurals: we need to add the correct keys for each language (i.e. to keep in line with the work done in [this card](https://trello.com/c/lkVB57ri/488-3-standardise-keys-for-plural-forms-across-locale-files)).  Other than that, everything else looks really good.

config/locales/ar.yml
1pretz1: Ah yes you're right, good spot! I have added / removed pluralisations for the other apps but Whitehall must've slipped through the net.

1pretz1:
Thanks for the review @brucebolt, I believe the correct pluralisations per language have now been added. Would you mind giving it another gander?

brucebolt (approved):
Thanks for making that change.  Everything looks good now 🎉.

There's a test failure, but it looks completely unrelated 😕, maybe a flaky test?

1pretz1:
Hmm, yeah it does seem to be a flakey test. I reran the tests and they passed. 

I've had a little dig into it and can't see anything obvious to fix and can't reproduce the failure so I'm going to stop there before I end up in a rabbit hole 🐰

PR #6110, opened 2021-05-17
Upgrade Rails to 6.0.3.7

thomasleese:
This is to resolve an issue I was experiencing installing the dependencies related to `mimemagic`. I imagine this isn't currently affecting production because we run on Ubuntu which will have those files.

```
Fetching mimemagic 0.4.3
Installing mimemagic 0.4.3 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

rake aborted!
Could not find MIME type database in the following locations:
["/usr/local/share/mime/packages/freedesktop.org.xml",
"/opt/homebrew/share/mime/packages/freedesktop.org.xml",
"/opt/local/share/mime/packages/freedesktop.org.xml", "/usr/share/mime/packages/freedesktop.org.xml"]
```

It turned into a bigger upgrade than I was expecting for a patch update!


Tetrino (approved):

Tetrino:
Seems okay to me, unsure on why the change from whitelist to allowlist but I'm assuming that's just a requirements hiccup.

thomasleese:
> Seems okay to me, unsure on why the change from whitelist to allowlist but I'm assuming that's just a requirements hiccup.

It was due to this change https://github.com/carrierwaveuploader/carrierwave/commit/4c3cac75f3a473e941045c23ebb781f61af67d79 that we started getting deprecation warnings, so I thought I may as well fix them.

PR #6119, opened 2021-05-26
Refresh markup on world embassies index page

owenatgov:
## What
Updates the [world embassies index page](https://www.gov.uk/world/embassies) to use design system classes and reduce the number of override classes to adhere to the design of the page.

## Why
Part of ongoing work by the govuk accessibility team to ensure we are consuming our frontend tooling (the components gem and the design system) as often as possible to reduce the risk of legacy frontend in our apps.

## Visual changes
| Before | After |
| --- | --- |
| Screenshot 2021-05-26 at 09 46 00 | Screenshot 2021-05-26 at 09 45 50 |

### DIFF
--- a/app/views/embassies/index.html.erb
+++ b/app/views/embassies/index.html.erb
@@ -1,11 +1,9 @@
 <% page_title "Find a British embassy, High Commission or Consulate" %>
 <% page_class "worldwide-organisations embassies-index govuk-width-container" %>
-<header class="block headings-block">
-  <%= render "govuk_publishing_components/components/title", {
-    context: "Worldwide",
-    title: "Find a British embassy, high commission or consulate",
-  } %>
-</header>
+<%= render "govuk_publishing_components/components/title", {
+  context: "Worldwide",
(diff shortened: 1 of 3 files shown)


app/views/embassies/index.html.erb
```diff
@@ -1,11 +1,9 @@
 <% page_title "Find a British embassy, High Commission or Consulate" %>
 <% page_class "worldwide-organisations embassies-index govuk-width-container" %>
-<header class="block headings-block">
```
chris-gds, line 3: removing this makes sense, i'm seeing these extra headers in a few places.
https://github.com/alphagov/frontend/pull/2775

chris-gds (approved):

app/views/embassies/index.html.erb
owenatgov: I reckon it was just how govuk used to be structured and was forgotten about in Whitehall. As we discussed on a previous PR, we already have a header in our default template so removing this one makes sense.

PR #6120, opened 2021-05-26
Remove link from heading context and replace with breadcrumb

maxgds:
https://trello.com/c/3ENWcE93/729-confusing-contextual-link-on-some-pages

This removes a link from the contextual text that sits above a header, and replaces it with breadcrumbs. This is to preserve the links back to parent content but to make the link more obvious and usable. We will later remove the ability to link contextual text as it is too difficult for users to recognise as a link.

To deal with the breadcrumbs I've had to bring in the CSS and do some overrides of greedy CSS selectors elsewhere. 

Sample URLs:
https://www.gov.uk/government/latest?topical_events%5B%5D=cop26
https://www.gov.uk/government/latest?departments%5B%5D=department-for-education&t=3
https://www.gov.uk/government/history/past-foreign-secretaries
https://www.gov.uk/government/history/past-foreign-secretaries/robert-cecil

I'm particularly interested to know if there's a better way to detect whether the latests documents (app/views/latest/index.html.erb) screen is showing topical events content or organisational content as the "subject" didn't seem to share much commonality.

## Before
Screenshot 2021-07-05 at 13 55 58
## After
Screenshot 2021-07-05 at 13 53 07

### DIFF
--- a/app/views/past_foreign_secretaries/index.html.erb
+++ b/app/views/past_foreign_secretaries/index.html.erb
@@ -1,13 +1,21 @@
 <% page_title "Past Foreign Secretaries" %>
 <% page_class "historic-people-index govuk-width-container" %>
-
+<%= render "govuk_publishing_components/components/breadcrumbs", {
+  breadcrumbs: [
+    {
+      title: "Home",
+      url: "/"
+    },
+    {
+      title: "History of the UK Government",
+      url: "/government/history"
+    }
+  ]
+} %>
 <header class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     <%= render "govuk_publishing_components/components/title", {
-      context: {
-        text: "History",
-        href: "/government/history",
-      },
+      context:"History",
       title: "Past Foreign Secretaries",
     } %>
   </div>
--- a/app/assets/stylesheets/frontend/base.scss
+++ b/app/assets/stylesheets/frontend/base.scss
@@ -10,6 +10,7 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/component_support';
 
 @import 'govuk_publishing_components/components/button';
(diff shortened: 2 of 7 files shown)


chris-gds:
This URL appears to still have the original link as it's located in [this file](https://github.com/alphagov/whitehall/blob/78dfffc389e7c0838b74c36e4de9ab751cc90771/app/views/past_foreign_secretaries/index.html.erb#L5).

`/government/history/past-foreign-secretaries`

chris-gds:
I just popped this integration to test the rest of pages below:

`/latest?topical_events%5B%5D=cop26`
`/government/latest?departments%5B%5D=department-for-education&t=3`
`/government/history/past-foreign-secretaries`
`/government/history/past-foreign-secretaries/robert-cecil`

I couldn't seem to view the other pages locally to test, I'm pretty sure it's known data thing. I do wonder if there's a way to get these working locally - the rest of the changes are all good and present. ✅

(I did wonder if the breadcrumb had lost their hover state - however it's just because Whitehall doesn't have the link underline updates yet)

maxgds:
> /government/history/past-foreign-secretaries

Good spot, now fixed.

app/views/past_foreign_secretaries/index.html.erb
```diff
@@ -4,10 +4,7 @@
 <header class="govuk-grid-row">
```
chris-gds, line 4: I've a feeling the breadcrumbs should go above this?  However i'm wondering if this was intentional is the top level and therefore we don't need it

maxgds: You're right, it should  link back to  history like other pages. Soon it will!

maxgds: Fixed.

chris-gds (approved):

PR #6123, opened 2021-05-28
Remove Brexit callout code

sihugh:
We've removed the code that renders the Brexit callout from government-frontend already, and the Whitehall publishing app no longer allows people to add config. This is the last stage which is to remove the underlying model and db entries.

https://trello.com/c/xnVMpmH8/1526-actually-remove-brexit-callout-for-24-may


hannako (approved):
👍

PR #6134, opened 2021-06-07
Add support for Nepali (ne) language

MuriloDalRi:
https://trello.com/c/cSuhVSSq/2550-add-nepalese-translation-option-to-whitehall


config/locales/ne.yml
```diff
@@ -0,0 +1,683 @@
...
+        minister:
+        speaker:
+      delivered_on:
+      delivery_title:
+        minister:
+        speaker:
+      written_on:
+    type:
+      about:
+        few:
```
brucebolt, line 125: I _think_ (based on the [Unicode rules](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html#ne)), that Nepali only uses `one` and `other` for it's pluralisation.

brucebolt (commented):
This looks good @MuriloDalRi.  I've added one comment as I don't think we need all the plural forms for Nepali.

config/locales/ne.yml
MuriloDalRi: Does it look ok now?

brucebolt: Perfect 🎉

brucebolt (approved):

PR #6143, opened 2021-06-10
Bump to govuk_app_config 4.0.0.pre.4

ChrisBAshton:
This should stop datasync-related issues from being logged in Sentry.

Trello: https://trello.com/c/4JOuje6O/2546-fix-broken-data-sync-related-sentry-errors-not-being-ignored


Gemfile
```diff
@@ -21,7 +21,7 @@ gem "globalize"
 gem "govspeak"
 gem "govuk_ab_testing"
 gem "govuk_admin_template"
-gem "govuk_app_config"
+gem "govuk_app_config", "4.0.0.pre.4"
```
brucebolt, line 24: Should we be specifying this as a minimum version, rather than fixing it on this particular release?

ChrisBAshton: I don't think it matters hugely either way - this is just to test out the pre-release, and will need a further PR to remove this constraint once v4 is officially released. I suppose it probably ought to have a minimum version, that way Dependabot would raise patch PRs if we release any more hotfixes (and I'm hoping we don't need to!).

Perhaps if we find we do need a `.pre.5` version, that would be the time to set all of these to minimum versions?

brucebolt: That sounds like a good approach :+1:

PR #6144, opened 2021-06-10
Publish routes to some hardcoded pages

sihugh:
(It's basically a copy of the rake task directly above, to limit the impact)

This will make them findable in search, and should also allow them to be tagged to recent post holders that exist as "people" pages in this site.

The pages don't have descriptions or leading paragraphs that are appropriate for use here.


hannako (approved):
👍

PR #6147, opened 2021-06-11
[DO NOT MERGE] Update World Location News and Topical Events templates

chris-gds:
## What

Update World location news template to use components / the [Design System](https://design-system.service.gov.uk/)

## Why

On-going a11y work, to swap components, align architecture and, in turn, improve accessibility.

## Visuals

Note (local "after" screenshots have missing images due to local data - this won't be an exact replica)

| Before  | After |
|---|---|
| www gov uk_world_france_news(iPhone 6_7_8) (2)  | www integration publishing service gov uk_world_france_news(iPhone 6_7_8) (1) |
| www gov uk_world_france_news (1) | www integration publishing service gov uk_world_france_news |

## How to test

Only some pages have "statistics documents" one example can be found below:
[/world/falkland-islands/news](https://www.gov.uk/world/falkland-islands/news)

Other example URLs:
[/world/france/news](https://www.gov.uk/world/france/news)
[/world/new-zealand/news](https://www.gov.uk/world/new-zealand/news)

## Anything else

[DO NOT MERGE]
There are multiple files with `/app/views/shared` that require updating across all templates in use
eg: `recently_updated` appears on [/world/uk-and-the-commonwealth
](http://www.gov.uk/world/uk-and-the-commonwealth)
Swapping [`document_list`](https://components.publishing.service.gov.uk/component-guide/document_list) and [`translation_nav`](https://components.publishing.service.gov.uk/component-guide/translation_nav) proved problematic, this has been omitted from this PR to be raised separately. 

There are known logical heading order issues on these pages.  This should be addressed in the `document_list` swap

During this work I faced some local development issues, I've [updated the README](https://github.com/alphagov/whitehall/pull/6105) with some troubleshooting notes.


app/assets/stylesheets/frontend/helpers/_featured-links.scss
```diff
@@ -1,12 +1,8 @@
 .featured-links {
   list-style: none;
 
-  @include media(tablet) {
-    padding-left: $gutter-two-thirds;
-  }
-
   li {
-    @include core-16;
+    @include core-19;
```
owenatgov, line 5: Some poking around suggests that this `core-*` include is a whitehall-only feature, or even worse a feature from the previous govuk design system implementation 😱 Some poking around locally suggests as well that it's unnecessary. Could you instead try adding `govuk-link` to the parent `ul` and only using this selector to add the bold and play with the spacing? I'd also suggest checking out the components docs for lists as I think it has options for bold and adjusted spacing?

app/views/shared/_recently_updated.html.erb
```diff
@@ -1,23 +1,20 @@
...
+    text: t('latest_feed.title'),
+  } %>
+  <% if recently_updated.any? %>
+    <%= render partial: defined?(documents_source) && documents_source == 'rummager' ?
+      "shared/recently_updated_documents_from_rummager" :
+      "shared/recently_updated_documents",
+      locals: { recently_updated: recently_updated } 
+    %>
+  <% else %>
+    <p><%= t('latest_feed.no_updates') %></p>
```
owenatgov, line 12: `govuk-body`?

app/views/shared/_recently_updated.html.erb
```diff
@@ -1,23 +1,20 @@
...
+  <% end %>
+  <% if local_assigns[:see_all_link] %>
+    <p class="govuk-!-margin-top-3 govuk-!-margin-bottom-6" <%= t_lang('document.see_all') %> >
```
owenatgov, line 15: `govuk-body` again? I also feel a bit sus about these override classes here.

app/views/world_location_news/index.html.erb
```diff
@@ -21,78 +21,103 @@
...
-        <% end %>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full">
+        <hr class="govuk-section-break govuk-section-break--visible world-locations__section-break">
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Featured",
+          font_size: "l",
+          margin_bottom: 4,
+        } %>
+        <section class="featured-news items-<%= @feature_list.current_feature_count %>" id="featured-documents govuk-!-margin-bottom-8">
```
owenatgov, line 33: `govuk-!-margin-bottom-8` ain't an id my dude

app/views/world_location_news/index.html.erb
```diff
@@ -21,78 +21,103 @@
...
-  </div>
 
-  <% if (@non_statistics_publications + @announcements + @statistics_publications).any? %>
-    <div class="block documents-grid">
-      <div class="inner-block">
-        <h1 class="block-title"><%= t('world_location.headings.documents') %></h1>
+    <% if (@non_statistics_publications + @announcements + @statistics_publications).any? %>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">    
+        <hr class="govuk-section-break govuk-section-break--visible world-locations__section-break govuk-!-margin-top-9 govuk-!-margin-bottom-3">
```
owenatgov, line 76: If these `hr`'s are using override classes then there's no value in the bespoke class (`world-locations__section-break`). I'd recommend committing to a single ideology here: either use the class for additional spacing or just use the overrides.

app/views/world_location_news/index.html.erb
```diff
@@ -21,78 +21,103 @@
...
-              </p>
-            </div>
+            <hr class="govuk-section-break govuk-section-break--visible world-locations__section-break govuk-!-margin-top-9 govuk-!-margin-bottom-3">
+            <%= render "govuk_publishing_components/components/heading", {
+              text: t('world_location.headings.publications'),
+              heading_level: 3,
+              margin_bottom: 7,
+            } %>  
+            <%= render partial: "shared/list_description", locals: { editions: @non_statistics_publications } %>
+            <p class="see-all">
```
owenatgov, line 100: `govuk-body` again again?

app/views/world_locations/show.html.erb
```diff
@@ -81,7 +81,7 @@
           </section>
         <% end %>
         <% if @statistics_publications.any? %>
-          <section id="statistics-publications" class="document-block documents-<%= document_block_counter %>">
+          <section id="statistics-publications" class="document-block documents-<%= document_block_counter %> govuk-!-margin-top-3">
             <h1><%= t('world_location.headings.statistics') %></h1>
```
owenatgov, line 85: Where is this page rendered? I'm not sure about this `h1` hanging around on the page. Could it be a title component?

test/unit/helpers/document_helper_test.rb
```diff
@@ -81,7 +81,7 @@ class DocumentHelperTest < ActionView::TestCase
     controller.stubs(:url_options).returns(
       action: "show",
       controller: "world_locations",
-      locale: "it",
+      locale: "de",
```
owenatgov, line 84: Not major but how come this changed?

app/assets/stylesheets/frontend/helpers/_featured-links.scss
chris-gds: Good spot, I didn't even clock, I thought this _was_ the Design System but of course it's prefixed differently.  I'll sort.

test/unit/helpers/document_helper_test.rb
chris-gds: Tests were seemingly failing, further down you can see it's testing for a German locale but it's passing Italian

app/views/world_locations/show.html.erb
chris-gds: Ah it appears here: <https://www.gov.uk/world/falkland-islands/news> - but yes h1 needs to go!

app/views/world_location_news/index.html.erb
chris-gds: 🙈

app/views/world_locations/show.html.erb
owenatgov: Meu deus that section has a lot of h1s! If you feel like cleaning that up that'd be well good of you 😁

test/unit/helpers/document_helper_test.rb
owenatgov: How weird! Feel free to resolve my dude.

app/views/world_location_news/index.html.erb
chris-gds: So, in theory I do agree however 😁
The `<hr>` [exists within the Design System](https://design-system.service.gov.uk/styles/typography/#section-break)

However this thicker black bar doesn't exist anywhere (by _anywhere_ the gem or the Design system - ideally it should)

I thought a nice halfway was to use the Design System as much as possible _then_ layer on top bespoke styling for the thicker black bar (as it then gets all the benefits of any release)

However I do take your point, Whitehall uses this quite a lot it might be better to make this style more global within this repo - what do you think?

image

app/views/world_locations/show.html.erb
chris-gds: Tell a lie - this page appears here: <https://www.gov.uk/world/uk-and-the-commonwealth>
i'm going to remove this change this shouldn't have bene touched.

app/views/shared/_recently_updated.html.erb
chris-gds: >I also feel a bit sus about these override classes here.

How come? I thought they are preferred approach to avoid extra CSS

image

owenatgov: Basically, I don't really understand why we need it. The space difference feels negligible from a design perspective and muddies the convention of consistent spacing across govuk.

app/views/world_location_news/index.html.erb
owenatgov: In that case could we use the convention that the rest of the page uses and use a border on top of the section instead of the `hr`? I feel uncomfortable about this bespoke flavour of the section-break hanging around in whitehall.

chris-gds: >I feel uncomfortable about this bespoke flavour of the section-break hanging around in whitehall.

I know what you mean. I feel like my preference this instance is to keep the `<hr>` and adjust the Design so it's inline with the Design System (make it thinner on this page)[1]

Opening this can up a tiny bit, it's a real mix across the site. Eg if you look on the footer or other pages sometimes it's a border, other times it's a `<hr>` i'm not convinced there's a clear convention.

https://www.gov.uk/coronavirus < Mix
https://www.gov.uk/access-to-elected-office-fund < `border`
https://www.gov.uk/ < Footer: mix of `border` and `<hr>` 
https://www.gov.uk/government/consultations/taking-action-on-climate-risk-improving-governance-and-reporting-by-occupational-pension-schemes-response-and-consultation-on-regulations < `border`

> In that case could we use the convention that the rest of the page uses and use a border on top of the section
 
I'm not sure that's 100% right either as this _is_ a `<hr>` semantically.
> You can use the `govuk-section-break` classes on an `<hr>`  element to create a thematic break between sections of content. 

Will check quickly with Design and decide: - make thin = `hr`, 
Keep thick = use border.

---
[1]
image

app/views/shared/_recently_updated.html.erb
chris-gds: > The space difference feels negligible from a design perspective

It's subtle enough, removed.

app/views/world_location_news/index.html.erb
chris-gds: So @owenatgov this has been resolved in a better way - i've realised that [headings](https://components.publishing.service.gov.uk/component-guide/heading) on the gem have a border property. 🙌🏼

app/views/world_location_news/index.html.erb
```diff
@@ -21,78 +21,107 @@
...
-  </div>
 
-  <div class="block our-mission">
-    <div class="inner-block">
-      <section class="article-section">
-        <h1 class="keyline-header"><%= t('world_location.headings.mission') %></h1>
-        <p>
-          <span class="mission_statement"><%= govspeak_to_html(@world_location.mission_statement) %></span>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds world-locations__our-mission govuk-!-margin-bottom-3">
```
owenatgov, line 60: You might've already explained this to me but why do we have override classes littered throughout `grid-row`'s on this page? My broad impression is that it's to adhere to the design/the current layout of the page? If that's the case I'm pushing back, we have a consistent spacing matrix for the entire site and I think it's unwise to bypass that for the sake of this one page.

chris-gds: >My broad impression is that it's to adhere to the design/the current layout of the page?   

So! It’s the opposite - it’s to more closely adhere to the current re-design:
There’s a lot more breathing space on this between the two sections.  On re-review I’m more tempted to add _more_ space to this.  

Redesign (Left) > Update (Right)
Screenshot-2021-07-23-at-20 47 54

> we have a consistent spacing matrix for the entire site 

I’m not sure if this responsive scale / grid system has been overlaid on these designs in perhaps the way that’s expected?
 In this context I felt it aligned with the design system ideology:

>Occasionally, you might need to make minor adjustments like adding or removing spacing to elements of your design. You can use the responsive spacing override classes for this.\

Unless there's another technique that's more apt for adding spacing for a Design change?

owenatgov: Aha, that makes sense! I have a feeling that whitehall's own bothersome styling matrix probably messes with this a little bit. I can't think of an alternative right now so I think your approach makes sense.

One last little press though is that when I test this on int, the spacing for the documents section looks a little whacky:

Screenshot 2021-07-28 at 10 27 38

chris-gds: Ace! that's been kept and tweaked.

> One last little press though is that when I test this on int, the spacing for the documents section looks a little whacky:

Missing local data pretty much means "Dev via integration" is the way to go - updated 

image

owenatgov: Ok, one final nitpick and then I promise that's it! The space between the headings in that documents section are pretty colossal. You've got this outrageous gap between "Documents" and "Our announcements" then under all the other headings there's that same big gap. Could we give the subheadings more uniform spacing, like that under the first one (Our announcements) and then squeeze the spacing under the main section heading? Ideally in line with other headings on the page eg: Our mission.

chris-gds: Hey ya - I'm happy to update these but I've a feeling these are Design changes rather than Dev issues - I've sent a link to the [final Design](https://www.figma.com/file/6oVdr466OnwiHjdBmqvZAX/Untitled?node-id=0%3A1). Do these exist within this?

owenatgov: They do sort of but the design feels more consistent space-wise. At the very least could we align the spacing on all the subheadings in that section please?

chris-gds: Hey @owenatgov I've updated to try find a nice halfway to address these comments and maintain the Design.  Let me know what you think (sharing with Design as well)

owenatgov (dismissed):
We're DONE

jon-kirwan:
https://trello.com/c/B2jO8w52/626-update-topical-events-template

##What##
To be added

##Why##
To be added

##Visuals##
To be added

app/views/topical_events/show.html.erb
```diff
@@ -51,75 +61,107 @@
...
+          padding: true,
+          border_top: 2,
+        } %>
+
+        <%= render(partial: 'shared/document_list_from_rummager', locals: { documents: @publications['results'], type: :publications, documents_count: @publications['total'] }) if @publications['results'].any? %>
+        <%= render(partial: 'shared/document_list_from_rummager', locals: { documents: @consultations['results'], type: :consultations, documents_count: @consultations['total'] }) if @consultations['results'].any? %>
+        <%= render(partial: 'shared/document_list_from_rummager', locals: { documents: @announcements['results'], type: :announcements, documents_count: @announcements['total'] }) if @announcements['results'].any? %>
+
+        <% if @detailed_guides.any? %>
+          <div id="detailed-guidance" class="documents-<%= document_block_counter %>">
```
owenatgov, line 110: Another mystery ID

app/views/topical_events/show.html.erb
```diff
@@ -51,75 +61,107 @@
...
+            <%= render "govuk_publishing_components/components/heading", {
+              text: "Detailed guidance",
```
owenatgov, line 112: There are a few instances of free text floating around here. Could we get this into the translation file and pull it that way?

owenatgov (commented):
A few comments. Glad to see this out the door finally!

app/views/classifications/_classification_featuring.html.erb
```diff
@@ -1,16 +1,20 @@
-<%= content_tag_for(:article, classification_featuring, nil, class: "feature item-#{classification_featuring_counter}") do %>
-
-  <%
-    link_attributes ||= { class: 'govuk-link' }
-    link_attributes[:rel] = "external" if is_external?(classification_featuring.url)
-  %>
-  <div class="content">
+<%= content_tag(:div, nil, class: 'govuk-grid-column-one-third') do %>
```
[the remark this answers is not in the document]
jon-kirwan: Good catch, yep i've updated all occurrences in that file. As a side note I've found it difficult to set up a lint tool that'll pick up (and automatically fix) these types of issues. I've found that the Whitehall lint tasks don't seem to pick up these types of issues locally and only some things when committed / checked on Jenkins. 

I'd used 'Prettier' https://prettier.io/ in the past but unfortunately that doesn't seem to work with .erb files.

app/views/classifications/_classification_featuring.html.erb
```diff
@@ -1,16 +1,20 @@
...
       heading_text: classification_featuring.title,
       description: truncate(classification_featuring.summary, length: 160, separator: ' '),
     } %>
-  </div>
+  <% end %>
 <% end %>
+
+<% if (classification_featuring_counter.to_i + 1) % 3 == 0 %>
+  <%= content_tag(:div, nil, class: 'govuk-clearfix') %>
+<% end %>
```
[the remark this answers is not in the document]
jon-kirwan: Thanks, fixed and yep should keep it consistent.

app/views/topical_events/show.html.erb
```diff
@@ -51,75 +61,107 @@
...
+          <div id="organisations" class="organisations">
+            <%= render "govuk_publishing_components/components/heading", {
+              text: "Who\'s involved",
+                font_size: "l",
+                margin_bottom: 4,
+                padding: true,
+                border_top: 2,
+            } %>
+
             <ul>
```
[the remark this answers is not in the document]
jon-kirwan: Unfortunately, I don't think I can use it here right now. The content in the list is a collection of logos but has a very non-linear layout achieved with custom styles e.g. -

```
#whitehall-wrapper .topical-events-show .organisations ul li {
    float: left;
    padding: 0 15px 15px;
    min-height: 90px;
}
```

You can see an example here: https://www.integration.publishing.service.gov.uk/government/topical-events/first-world-war-centenary - under 'Who’s involved'.

app/views/topical_events/show.html.erb
jon-kirwan: I think you're right. I got a bit confused thinking the `detailed-guidance` id was used in a test elsewhere but it's not so have now removed (page still looks the same and no test fails)!

As a side note also removed: 
- the `travel-advice` id. I don't think it's needed either and
- `class="documents-<%= document_block_counter %>">` from both elements and
- each containing div since it's also no longer needed

app/views/topical_events/show.html.erb
jon-kirwan: Thanks, i've added the following headings in `config/locales/en.yml` -
```
  topical_event:
    headings:
      detailed_guidance: Detailed guidance
      documents: Documents
      featured: Featured
      organisations: Who\'s involved
      travel_advice: Travel advice
```
and referenced each in the view

jon-kirwan: ... and also have generated translations for each locale.

app/views/topical_events/show.html.erb
```diff
@@ -51,75 +61,107 @@
...
+                  }
+                }
+              end
+            } %>
           </div>
-          <div class="content">
+        <% end %>
+
+        <% if @topical_event.lead_organisations.any? %>
+          <div id="organisations" class="organisations">
```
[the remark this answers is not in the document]
jon-kirwan: Looks like it's not used anywhere. I've removed it.

app/views/topical_events/show.html.erb
```diff
@@ -3,43 +3,53 @@
...
 
-  <header class="headings-block">
-    <div class="inner-block">
-      <div class="govuk-grid-row">
-        <div class="govuk-grid-column-two-thirds">
-          <%= render "govuk_publishing_components/components/title", {
-            title: "#{h(@topical_event.name)}#{' <span class="archived">(Archived)</span>' if @topical_event.archived?}".html_safe
-          } %>
-          <% if @topical_event.organisations.any? %>
+        <div class="page-description">
```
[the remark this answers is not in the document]
jon-kirwan: Okay the `description` and `page-description` selectors were previously used to format govspeak content (in this case the remaining use was to format the lead paragraph copy).

Changes now include:
- removed both selectors (and associated styles) for `description` and `page-description`
- added a lead paragraph component which accepts text that has been processed by the `govspeak_to_html` and `strip_tags` helpers to remove any html (code sample below).

https://github.com/alphagov/whitehall/blob/cf0fc29bcd9d9c1da56cee82be7bc3eb7fda47ae/app/views/topical_events/show.html.erb#L14-L18

The reason for these changes is because some pages have heading markdown, in govspeak, where the lead paragraph copy should be e.g. https://www.gov.uk/government/topical-events/g7-uk-2021

```
##The UK hosted the G7 Summit as part of its 2021 G7 Presidency

The G7 Summit was held in Carbis Bay, Cornwall on 11-13 June 2021.
```

How does all this seems to you? It ties in in with some of the changes also carried out on this PR to separate out the page description in to a lead paragraph and a secondary description (detailed under _Anything else_)

app/assets/stylesheets/frontend/views/_topical-events.scss
```diff
@@ -1,60 +1,29 @@
...
     }
   }
 
   .organisations {
-    padding-top: $gutter;
-
     ul {
       list-style: none;
 
       li {
```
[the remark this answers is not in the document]
jon-kirwan: I think I may have some trouble swapping out the tag styles here as they are used to format a list of logos (also added more detailed below).

app/views/topical_events/show.html.erb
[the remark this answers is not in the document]
owenatgov: Addressed this in an above comment

owenatgov (commented):
Some final comments. We're real close 👍

app/views/topical_events/show.html.erb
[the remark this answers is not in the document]
owenatgov: I think you've done the best you can with the wobbly data you've had to work with. Resolving 👍

app/assets/stylesheets/frontend/views/_topical-events.scss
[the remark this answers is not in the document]
jon-kirwan: Sounds good to me - i've updated like so:

https://github.com/alphagov/whitehall/blob/24c899ca48d408d1341ba7a9d5cf9a04179799d1/app/assets/stylesheets/frontend/views/_topical-events.scss#L9-L15

and

https://github.com/alphagov/whitehall/blob/24c899ca48d408d1341ba7a9d5cf9a04179799d1/app/views/topical_events/show.html.erb#L156-L168

owenatgov (approved):
🚀

jon-kirwan:
> 🚀

Cheers Owen! Could I ask you to hit the merge button for me too 👍

PR #6148, opened 2021-06-14
Add a rake task to create attachment report

brucebolt:
This adds a rake task that produces a CSV file containing a list of all published attachments and some related metadata.

Trello card: https://trello.com/c/BSt0Ya5M

### DIFF
--- a/lib/reports/published_attachments_report.rb
+++ b/lib/reports/published_attachments_report.rb
@@ -1,0 +1,40 @@
+module Reports
+  class PublishedAttachmentsReport
+    def report
+      path = Rails.root.join("tmp/attachments_#{Time.zone.now.strftime('%d-%m-%Y_%H-%M')}.csv")
+      csv_headers = ["Organisation", "Filename", "Filetype", "Published Date"]
+
+      attachments = Attachment.find_by_sql([
+        "SELECT a.*
+          FROM attachments a
+          WHERE a.attachable_type = 'Edition'
+          AND a.attachment_data_id IS NOT NULL
+          AND EXISTS(
+            SELECT e.id
+            FROM editions e
+            WHERE e.state = 'published'
+            AND a.attachable_id = e.id
+          )
+          AND EXISTS(
+            SELECT ad.id
+            FROM attachment_data ad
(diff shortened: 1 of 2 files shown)


lib/reports/published_attachments_report.rb
```diff
@@ -0,0 +1,40 @@
+module Reports
+  class PublishedAttachmentsReport
+    def report
+      path = Rails.root.join("tmp/attachments_#{Time.zone.now.strftime('%d-%m-%Y_%H-%M')}.csv")
+      csv_headers = ["Organisation", "Filename", "Filetype", "Published Date"]
+
+      attachments = Attachment.find_by_sql([
```
brucebolt, line 7: We can't do this query using ActiveRecord as the relation between Attachment and Edition (via. `attachable`) is polymorphic, so such as query would involve multiple steps which exceeded the machine's memory.

ChrisBAshton (approved):
Would be nice to have a test for this, that exercises that HTMLAttachments aren't included in the results. But appreciate that might be too ambitious for a Whitehall rake task PR, where the code is self-contained and the output can be checked manually :)

brucebolt:
> Would be nice to have a test for this, that exercises that HTMLAttachments aren't included in the results. But appreciate that might be too ambitious for a Whitehall rake task PR, where the code is self-contained and the output can be checked manually :)

I did start writing a test, but found it became too messy (e.g. dealing with where the CSV file gets output etc) for something that is likely to only ever be run once (and whose output can be verified manually).

1pretz1:
I agree with your points both. However, during my test coverage work I realised it was probably worthwhile to at least include a basic test for rake tasks as the [DraftPublicationsReport task](https://github.com/alphagov/whitehall/blob/7d1a0acc65045556b44d1b0d798d3208df7fd5e6/lib/tasks/reporting.rake#L8) was actually broken and even a simple test would have caught that the code returned an error.

If this task is going to be removed shortly then I think not having a test is fine. If not, I think we should have at least a simple one. I got round the CSV file output dilemma with `Timecop`: https://github.com/alphagov/whitehall/compare/wip-report?expand=1

PR #6162, opened 2021-06-18
Normalize locale files

brucebolt:
Normalizes the locale files in a way that is consistent with other applications.

[Trello card](https://trello.com/c/kB1a8jJE)


1pretz1 (approved):
Glad we've settled on a line length setting as it feels like we've been jumping between long and short lines every time we run normalize 😄

brucebolt:
> Glad we've settled on a line length setting as it feels like we've been jumping between long and short lines every time we run normalize 😄

I just took the plunge and decided the entire value on a single line looked neater and allows for easier commit comparisons.  It may be possible to put this into `govuk_app_config` so it's configured once everywhere ... I'll look into that.

PR #6166, opened 2021-06-23
Update check_organisation_links_worker.rb

abnsy:
**Bad binary expression operand order**
In a binary expression, the literal should preferably be on the right side of the expression. Yoda expressions, which are binary expressions spelled in reverse to how we normally speak, affect code readability & understandibility. This can be fixed by simply shifting the literal to the right and the variable to the left.


brucebolt (changes requested):
Hi @abnsy, thank you for raising this PR.

Looking at the code change, it would appear that the change affects the resulting value as the `**4` part (raising the value to a power of 4) has now been removed.

For example, with a `retry_count` of 4, the code at the moment would give a retry time of 9,375 (`((0.5 * (4 + 1)^4) * 30)`) but will now give a retry time of 75 (`(((4 + 1) * 0.5) * 30)`).

PR #6175, opened 2021-06-29
Remove unused code

1pretz1:
This PR focuses on removing a load of junk from Whitehall. All these unused classes and methods distorted the actual test coverage of the application when using [SimpleCov][1]. We needed accurate information regarding coverage and so unused code had to be removed. 

On the plus side Whitehall is a bit leaner.

**This PR will need to be carefully deployed to make sure that the removed code doesn't cause any issues**

---

Trello:

[1]: https://github.com/alphagov/whitehall/pull/6174


app/helpers/document_helper.rb
```diff
@@ -1,16 +1,9 @@
...
-  include MinisterialRolesHelper
-  include PolicyGroupsHelper
-  include RoleAppointmentsHelper
-  include TopicsHelper
-  include TranslationHelper
 
-  def edition_page_title(edition)
-    edition.withdrawn? ? "[Withdrawn] #{edition.title}" : edition.title
-  end
+  MS_WORD_DOCUMENT_HUMANIZED_CONTENT_TYPE = "MS Word Document".freeze
```
brucebolt, line 4: Could these be moved into the locale files?

brucebolt (approved):
Great work!  I can't believe how much unused code there is in this application 🤯.

app/helpers/document_helper.rb
1pretz1: Ooh yes they should, good spot. Same could be said about the [other humanised content types in the helper](https://github.com/alphagov/whitehall/blob/984764296ac058ac673e58652be79ef23fb6b184/app/helpers/document_helper.rb#L23-L62). 

Will do this in a separate PR!

PR #6177, opened 2021-06-29
Add tests for reports

1pretz1:
This PR adds tests for some (of the many) rake tasks. I focused mainly on the Publishing API rake tasks as they were the least tested and high value. There has also been some refactoring to try make navigating and adding new publishing api tasks a bit easier.

---

Trello:


test/unit/reports/organisation_attachments_report_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    Timecop.freeze do
+      path = Rails.root.join("tmp/#{organisation.slug}-attachments_#{Time.zone.now.strftime('%d-%m-%Y_%H-%M')}.csv")
+
+      Reports::OrganisationAttachmentsReport.new(organisation.slug).report
+    end
+
+    assert_equal Reports::OrganisationAttachmentsReport::CSV_HEADERS, CSV.read(path)[0]
+    assert_equal 1, CSV.read(path, headers: true).count
+
+    File.delete(path) if File.exist?(path)
```
ChrisBAshton, line 18: We can probably remove the conditional here.

test/unit/reports/organisation_attachments_report_test.rb
```diff
@@ -0,0 +1,42 @@
+require "test_helper"
+
+class Reports::OrganisationAttachmentsReportTest < ActiveSupport::TestCase
+  test "returns a report containing unaccessible attachments for an organisation" do
+    publication = create(:publication, :published, :with_file_attachment)
+    organisation = publication.organisations.last
+
+    path = nil
+    Timecop.freeze do
```
ChrisBAshton, line 9: Could we move all of the test contents inside this block, and remove the slightly hacky `path = nil` initialisation? Same applies to the other tests.

test/unit/reports/published_attachments_report_test.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    assert_equal Reports::PublishedAttachmentsReport::CSV_HEADERS, CSV.read(path)[0]
+    assert_equal 1, CSV.read(path, headers: true).count
+
+    File.delete(path) if File.exist?(path)
+  end
+
+  test "returns blank report if there are no eligible file attachments" do
+    create(:publication, :with_file_attachment)
```
ChrisBAshton, line 21: So a publication can't have a file attachment? 🤔 Or it can, but we just don't want it included in the report?

app/models/redirect_route.rb
```diff
@@ -0,0 +1,11 @@
...
+  def self.all
+    [
+      {
+        base_path: "/government/world",
+        destination: "/world",
+        content_id: "36620cf9-00b3-4d51-afab-e05e457061e3",
+      },
+    ]
+  end
+end
```
ChrisBAshton, line 11: Nice 👍

lib/tasks/publishing_api.rake
```diff
@@ -100,13 +61,7 @@ namespace :publishing_api do
...
   desc "Publish redirect routes (eg /government/world)"
   task publish_redirect_routes: :environment do
-    [
-      {
-        base_path: "/government/world",
-        destination: "/world",
-        content_id: "36620cf9-00b3-4d51-afab-e05e457061e3",
-      },
-    ].each do |route|
+    RedirectRoute.all.each do |route|
```
ChrisBAshton, line 64: Yep, much nicer to keep config away from logic.

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
...
+      tasks = [
+        { name: "organisation_by_slug", model: "Organisation" },
+        { name: "person_by_slug", model: "Person" },
+        { name: "role_by_slug", model: "Role" },
+      ]
+
+      test "republishes record by slug" do
+        tasks.each do |task|
+          record = create(task[:model].underscore) # rubocop:disable Rails/SaveBang
+          model = task[:model].constantize
```
ChrisBAshton, line 81: The more conventional approach to this kind of DRYing up of tests is to use [shared example groups](https://relishapp.com/rspec/rspec-core/v/2-0/docs/example-groups/shared-example-group).

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
...
+
+          model.any_instance.expects(:publish_to_publishing_api).at_least_once
```
ChrisBAshton, line 83: The `at_least_once` threw me slightly - do we know how many times it is called? Should it just be once?

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
+require "test_helper"
+
+class PublishingApiRake < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  teardown do
+    task.reenable if defined?(task) # without this, calling `invoke` does nothing after first test
```
ChrisBAshton, line 7: It's a shame this doesn't get applied to the `republish` set of tests, though it could if we defined a `let(:task)` in there. Can we do that?

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
...
+            true,
+          )
+          task.invoke
+        end
+      end
+
+      describe "#document_type" do
+        let(:task) { Rake::Task["publishing_api:bulk_republish:document_type"] }
+
+        test "republishes all documents from document type" do
```
ChrisBAshton, line 233:
```suggestion
        test "republishes all documents of the specified document type" do
```

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
...
+          PublishingApiDocumentRepublishingWorker.expects(:perform_async_in_queue).with(
+            "bulk_republishing",
+            edition.document.id,
+            true,
+          )
+
+          task.invoke(org.slug)
+        end
+
+        test "Ignores documents owned by other organisation" do
```
ChrisBAshton, line 261:
```suggestion
        test "Ignores documents owned by other organisations" do
```

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
...
+
+          PublishingApiDocumentRepublishingWorker.expects(:perform_async_in_queue).with(
+            "bulk_republishing",
+            edition.document.id,
+            true,
+          )
+
+          task.invoke(filename)
+
+          File.delete("lib/tasks/#{filename}.csv") if File.exist?("lib/tasks/#{filename}.csv")
```
ChrisBAshton, line 312: I think we can remove this conditional too.

test/unit/tasks/publishing_api_test.rb
```diff
@@ -0,0 +1,384 @@
...
+
+        test "Republishes documents by content ids from csv" do
+          edition = create(:publication)
+          filename = "content_ids_#{Time.zone.now.to_i}"
+
+          File.open("lib/tasks/#{filename}.csv", "w+")
+          CSV.open("lib/tasks/#{filename}.csv", "wb") do |csv|
+            csv << %w[content_id]
+            csv << [edition.content_id]
+          end
```
ChrisBAshton, line 302: Might be nicer to use a string for this, something like:

```ruby
File.write("lib/tasks/#{filename}.csv", "content_id\n#{edition.content_id}")
```

lib/tasks/publishing_api.rake
```diff
@@ -84,6 +84,12 @@ namespace :publishing_api do
     end
   end
 
+  # TODO: Remove task after https://github.com/alphagov/publishing-e2e-tests/pull/422 has been merged
+  desc "Republish all organisations"
+  task republish_all_organisations: :environment do
+    Organisation.find_each(&:publish_to_publishing_api)
+  end
```
ChrisBAshton, line 91: Nice 👍

ChrisBAshton (commented):
Really good job bringing the test coverage up, Peter 👍 I have a few questions and suggestions below.

test/unit/reports/organisation_attachments_report_test.rb
1pretz1: Sure thing

test/unit/reports/published_attachments_report_test.rb
1pretz1: It can but it won't get returned in the report unless the document is published. I've matched the factories document types (both now detailed guides) which should make the difference in test setup more apparent.

test/unit/tasks/publishing_api_test.rb
1pretz1: I wish this was RSpec world, unfortunately we're currently faking RSpec features in Minitest which doesn't include shared examples, I don't think.

test/unit/tasks/publishing_api_test.rb
1pretz1: Yes it should be just once - `at _least_once` was copied over from the test below. Turns out that `publish_to_publishing_api` was being called more than once due to the Role factory creating an Organisation (and attempting to publish to pub api) on initialisation. Altered the test config slightly to mitigate this. Good spot!!

test/unit/tasks/publishing_api_test.rb
1pretz1: Agreed so I've tried to refactor using `let`, with not much success. I'm open to suggestions though 😄

test/unit/tasks/publishing_api_test.rb
1pretz1: Nice suggestion, done!

1pretz1:
Thanks for the review @ChrisBAshton! Think I've addressed everything.

I was getting some strange behaviour when I was running the rake tests locally (tasks running twice, when only invoked once) and have narrowed it down to the tasks themselves being initialised more than once. I've updated this in the last commit.

test/unit/reports/published_attachments_report_test.rb
ChrisBAshton: Makes sense, thanks 👍

test/unit/tasks/publishing_api_test.rb
ChrisBAshton: Ah yes, I'd missed that this is Minitest.

test/unit/tasks/publishing_api_test.rb
ChrisBAshton: I've had a go here: https://github.com/alphagov/whitehall/commit/fe6d249da76630a37862f1d8a33bf442e7352d5d
It does lead to _slightly_ more lines of code, but I think this is worth it for the increased readability.

The rest of the PR looks great @1pretz1, but think it's worth making this one last tweak before merge ✅

1pretz1: Sure thing, I've split them into their own separate tests. I was in two minds when writing them as it did seem like a compromise on consistency when trying to DRY up the code.

ChrisBAshton (approved):
Great work, @1pretz1 👏

PR #6173, opened 2021-06-29
Make social media links translatable

brucebolt:
Makes the fields on the social media links in Whitehall translatable.  If a translation is not available for a link, the link from the default locale is used.

Users can add/edit translations for social media links in any language that the socialable (i.e. Organisation or Worldwide Organisation) page is available in.

We do not need to make any changes in Publishing API or Collections, as Whitehall will handle the translation and only present the relevant links to Publishing API for the language of the content it is publishing (using the `globalize` gem).

All existing social media links will be labelled as English and will need to be manually edited by the departmental editors if they wish to add translations.

#### Organisation

Editing an Organisation's social media links where the organisation has an English and Welsh translation:
Screenshot 2021-06-30 at 11 30 42

Editing the English social media link for an Organisation:
Screenshot 2021-06-30 at 11 30 50

Editing the Welsh translation of the social media link from above:
Screenshot 2021-06-30 at 11 31 04

English version of Organisation page:
Screenshot 2021-06-30 at 09 25 38

Welsh version of Organisation page:
Screenshot 2021-06-30 at 09 25 45

#### Worldwide Organisation

Editing a Worldwide Organisation's social media links where the organisation has an English and Greek translation:
Screenshot 2021-06-30 at 11 28 07

Editing the English social media link for a Worldwide Organisation:
Screenshot 2021-06-30 at 11 29 51

Editing the Greek translation of the social media link from above:
Screenshot 2021-06-30 at 11 30 01

English version of Worldwide Organisation page:
Screenshot 2021-06-30 at 11 28 32

Greek version of Worldwide Organisation page:
Screenshot 2021-06-30 at 11 28 38

[Trello card](https://trello.com/c/RRvGDDmD)

PR #6180, opened 2021-06-30
Add missing republish_organisation_to_publishing_api method

brucebolt:
This method was missed from https://github.com/alphagov/whitehall/pull/6173.

[Trello card](https://trello.com/c/RRvGDDmD)

### DIFF
--- a/app/models/social_media_account_translation.rb
+++ b/app/models/social_media_account_translation.rb
@@ -1,4 +1,12 @@
 class SocialMediaAccountTranslation < ApplicationRecord
+  belongs_to :social_media_account
+
   after_save :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
+
+  def republish_organisation_to_publishing_api
+    if social_media_account.socialable_type == "Organisation" && social_media_account.socialable.persisted?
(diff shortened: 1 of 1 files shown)


app/models/social_media_account_translation.rb
```diff
@@ -1,4 +1,12 @@
 class SocialMediaAccountTranslation < ApplicationRecord
+  belongs_to :social_media_account
+
   after_save :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
+
+  def republish_organisation_to_publishing_api
```
1pretz1, line 7: Is it worth adding a test for this method?

brucebolt: A very similar method is included in another 7 models with no test :grimacing:.

We should probably look at DRYing this up and adding a test ... but for now the missing method is blocking a migration on deployment.

1pretz1: Oh 🙈 

> We should probably look at DRYing this up and adding a test ... but for now the missing method is blocking a migration on deployment.

Sure thing, I'll give it a tick!

1pretz1 (approved):

PR #6181, opened 2021-06-30
Remove html-publications-with-asides.csv

richardTowers:
This wasn't needed in the end - we published all HTML attachments
instead (which caused an incident - there are a lot of HTML attachments).

We can remove this file now.

https://trello.com/c/rRruBvQN/706-govspeaks-statistic-headlines-use-wrong-semantics


maxgds (approved):
Begone!

PR #6202, opened 2021-07-19
Use new link styles

andysellick:
## What

Apply the new `govuk-frontend` link styles to whitehall frontend pages.

Pages checked:

- https://www.gov.uk/world
- https://www.gov.uk/government/get-involved
- https://www.gov.uk/world/organisations/british-embassy-kabul
- https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/882529/National_Nominal_Roll_for_England_and_Wales_V.csv/preview
- https://www.gov.uk/government/how-government-works
- https://www.gov.uk/government/history/past-prime-ministers
- https://www.gov.uk/government/history/past-prime-ministers/robert-gascoyne-cecil
- https://www.gov.uk/government/history/past-foreign-secretaries
- https://www.gov.uk/government/history/past-foreign-secretaries/edward-wood
- https://www.gov.uk/world/organisations/british-antarctic-territory
- https://www.gov.uk/world/british-antarctic-territory/news
- https://www.gov.uk/world/embassies
- https://www.gov.uk/courts-tribunals

I went through the documentation and whitehall's routes.rb file but if anyone knows of any pages I've missed please let me know.

## Why

Part of a wider effort to update link styles across GOV.UK.

## Visual changes

Increase spacing in metadata links (top right) to prevent focus state overlap on embassy pages. Also link styles on social media links (it's a component, so defaulting back to normal styles). Showing normal, hover and focus states.

Before | After
------ | ------
Screenshot 2021-07-19 at 16 03 49 | Screenshot 2021-07-19 at 16 04 12

Change to link style for big numbers on left of get involved page. Showing normal and hover states.

Before | After
------ | ------
Screenshot 2021-07-19 at 15 29 07 | Screenshot 2021-07-19 at 15 29 16

Trello card: https://trello.com/c/jK8M47sa/761-update-whitehall-to-use-new-link-styles


owenatgov (commented):
This all looks fine to me. I have one little comment:

On the navigation header in whitehall (found in `app/views/layouts/frontend.html.erb`) the links are still on the old link styles. You won't be able to fix this 100% here as the `proposition-links` style rule seems to sit in static but I reckon this can be halfway solved by adding `govuk-link` to the links themselves. This will at the very least create space between the text and the underline. You could alternatively try applying the [government nav component](https://components.publishing.service.gov.uk/component-guide/government_navigation).

andysellick:
> On the navigation header in whitehall (found in `app/views/layouts/frontend.html.erb`) the links are still on the old link styles.

I've applied some `govuk-link` styles in an extra commit, so the links at least stay the same. This isn't perfect, but at least it doesn't get any worse. The styles are actually coming from `govuk_template`, which is unfortunately pulled in as a complete CSS file directly into static's page template, so the only option for fixing these links properly is to add more gnarly overrides, which I'm not prepared to do.

owenatgov (approved):
Looks good to me 🔗 💅

PR #6203, opened 2021-07-21
Add Pact and use RSpec matchers

1pretz1:
We need to write some Pact tests as part of enabling continuous deployment for Whitehall. As a result of adding Pact as a dependency, RSpec is also bundled with the app. This causes a number of naming clashes between Minitest and RSpec, causing test failures in the Cucumber tests.

The simplest approach to rectify these clashes is to change the failing Minitest assertions to RSpec matchers across the entire feature test suite. The approach I took was to convert all assertions to use RSpec so at least in the feature tests, a consistent test framework is used. 

As a result, the unit tests `test/` will be using a different test framework to the feature tests, which isn't really avoidable and follows the same approach as what was implemented for [Imminence](https://github.com/alphagov/imminence/pull/644/commits/7c332c4fd69470eb6ec764a14f7a46f4d74d0615). 

Trello:


features/step_definitions/admin_statistical_data_sets_steps.rb
```diff
@@ -3,6 +3,6 @@
 
   data_sets = @new_edition.statistical_data_sets
 
-  assert_equal 1, data_sets.count
-  assert_equal data_set_name, data_sets.first.title
+  expect(1).to eq(data_sets.count)
```
brucebolt, line 6: A really minor point, but I wonder if this would read better the other way round (e.g. `expect(data_sets.count).to eq(1)`)?

features/step_definitions/sitewide_settings_steps.rb
```diff
@@ -16,20 +16,20 @@
...
+Then(/^I should not see the minister counts$/) do
+  expect(page).to_not have_selector(".feature-ministers")
+end
+
+Then(/^I should see a reshuffle warning message$/) do
+  expect(page).to have_content("Test minister reshuffle message")
+end
+
+Then(/^I should not see a reshuffle warning message$/) do
+  expect(page).to_not have_content("Test minister <a rel=\"external\" href=\"http://example.com\">reshuffle</a> message")
```
brucebolt, line 32: I'm not sure this assertion is correct as the affirmative situation above doesn't contain the HTML for the link.  Also, I'm pretty sure `have_content` [only matches the text, not the HTML](https://www.rubydoc.info/gems/capybara/Capybara%2FRSpecMatchers:have_content).

brucebolt (approved):
This looks like a very pragmatic solution to a gnarly problem!

I've spotted a potential problem with one of the existing step definitions (nothing blocking, but would be good to fix).

features/step_definitions/admin_statistical_data_sets_steps.rb
1pretz1: Ah, good spot! Have switched the args around.

features/step_definitions/sitewide_settings_steps.rb
1pretz1: You're correct, good catch! `have_content` only matches text, not HTML and so the first statement passes even though it does contain a link in the HTML. The second step was passing but not checking the right content, due to looking for raw HTML. Have updated this.

brucebolt (approved):
Great work, everything looks good to me :+1:

1pretz1:
Thanks for the review Bruce!! ⭐

PR #6204, opened 2021-07-21
Setup Pact for GDS API Adapters 

1pretz1:
Adds Pact setup configuration for testing Whitehall's world API endpoints. The consumer tests are located in [GDS API Adapters](https://github.com/alphagov/gds-api-adapters/pull/1099).  

Depends on:

- [x] https://github.com/alphagov/whitehall/pull/6203

---

Trello:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,7 +8,13 @@ ENV["LOG_LEVEL"] = "warn"
 require File.expand_path("config/application", __dir__)
 require "ci/reporter/rake/minitest" if Rails.env.test?
 
+begin
+  require "pact/tasks"
+rescue LoadError
+  # Pact isn't available in all environments
+end
+
 Whitehall::Application.load_tasks
 
 Rake::Task[:default].clear if Rake::Task.task_defined?(:default)
-task default: %i[lint test:in_parallel]
+task default: %i[lint test:in_parallel pact:verify]
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@
 /failures
 /node_modules
 /yarn-error.log
+
+spec/reports/pacts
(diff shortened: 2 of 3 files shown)


Rakefile
```diff
@@ -10,5 +10,11 @@ require "ci/reporter/rake/minitest" if Rails.env.test?
...
 
+begin
+  require "pact/tasks"
+rescue LoadError
+  # Pact isn't available in all environments
+end
+
 Rake::Task[:default].clear if Rake::Task.task_defined?(:default)
-task default: %i[lint test:in_parallel]
+task default: %i[lint test:in_parallel pact:verify]
```
brucebolt, line 20: Would this result in an error when `pact/tasks` hasn't been included (where we rescue `LoadError` above)?

1pretz1: Yes it would (this was just a copy / paste from other Pact test config). After seeing your comment I moved the logic into the rescue with something like:

```
begin
  require "pact/tasks"
  task default: %i[lint test:in_parallel pact:verify]
rescue LoadError
  # Pact isn't available in all environments
  task default: %i[lint test:in_parallel] # rubocop:disable Rake/DuplicateTask
end
```

However, after looking at some of our other apps it seems like we don't cater for this situation when requiring, probably due to us never actually running Rake out of the test / dev environment. e.g [Content Store](https://github.com/alphagov/content-store/blob/main/Rakefile#L1-L16), [Link Checker API](https://github.com/alphagov/link-checker-api/blob/main/Rakefile#L8-L19) etc which makes me think that the setup, as is, works ok. What do you think?

brucebolt: I can't think of a situation where we would run the default rake task outside of test or dev, so this looks reasonable to leave.

brucebolt (approved):

PR #6208, opened 2021-07-22
Fix 500s due to type not being available

maxgds:
This is a fix for a number of errors Sentry is reporting. It changes the way the document types are detected and adds support for World Locations which had been overlooked initially.

Tested with branch on integration:
https://www.integration.publishing.service.gov.uk/government/latest?departments%5B%5D=department-for-education
https://www.integration.publishing.service.gov.uk/government/latest?world_locations%5B%5D=germany
https://www.integration.publishing.service.gov.uk/government/latest?topical_events%5B%5D=budget-2021

image

### DIFF
--- a/app/views/latest/index.html.erb
+++ b/app/views/latest/index.html.erb
@@ -21,7 +21,7 @@
           },
         ]
       } %>
-    <% elsif subject&.type == "TopicalEvent" %>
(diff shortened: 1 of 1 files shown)


owenatgov (approved):
Much better than my one. Cheers Max!

PR #6210, opened 2021-07-23
Redesign historic appointments templates

owenatgov:
## What
Apply redesigns to the [past foreign secretaries](https://www.gov.uk/government/history/past-foreign-secretaries/william-grenville) and [past prime ministers](https://www.gov.uk/government/history/past-prime-ministers/william-pitt) templates.

## Why
To ensure that these pages are consistent with similar informative pages on govuk and to improve the readability and thereby accessibility of the page.

[Card](https://trello.com/c/Ei02GZeX/586-update-historic-appointments-pages)

## Visual changes
| Before | After |
| --- | --- |
| www gov uk_government_history_past-foreign-secretaries_william-grenville (1) | www integration publishing service gov uk_government_history_past-foreign-secretaries_william-grenville (1) |

## Anything else

The image is decorative in this context as there is already a title / h1 outlining who this page is about.  As a result I don't think an `image_card` is suitable for this element an `image_tag` has been used instead.  However there might be a case to add a variation to the `image_card` incase changes that need to impact "all images" need to be reflected.  Will discuss outside this PR


app/assets/stylesheets/frontend/views/_person-profile.scss
```diff
@@ -0,0 +1,18 @@
+.person-profile {
+  border-top: 1px solid $govuk-border-colour;
+  padding-top: govuk-spacing(4);
+  margin-bottom: govuk-spacing(6);
+}
+
+.person-profile__image {
```
chris-gds, line 7: Is there a variation of the image card that can be used here instead

app/views/past_foreign_secretaries/_show_person.html.erb
```diff
@@ -12,28 +12,83 @@
...
-      <p class="govuk-body-s">
-        <a href="/government/history/past-foreign-secretaries" class="govuk-link">See all past Foreign Secretaries</a>
-      </p>
+  <div class="govuk-grid-column-two-thirds">
+    <div class="person-profile">
+      <div class="person-profile__image">
+        <%= image_tag image, alt: "", loading: "lazy" %>
+      </div>
+      <div class="person-profile__text">
+        <%= image_alt %>
```
chris-gds, line 36: I'm not sure this is ideal, AT will ignore the image and then hear "Portrait of..." without context.  However if it was on the actual alt it would be "image, Portrait of..." that would make more sense.

I was taking a look at the image card on the gem there's use of `<figure>` and other patterns

app/views/historic_appointments/show.html.erb
```diff
@@ -1,57 +1,74 @@
...
+
+    <%= govspeak_to_html @historical_account.body %>
   </div>
-  <div class="person-detail">
-    <div class="inner">
-      <p class="govuk-body-l"><%= @historical_account.summary %></p>
 
-      <%= govspeak_to_html @historical_account.body %>
-    </div>
+  <div class="govuk-grid-column-one-third">
```
chris-gds, line 64: i'm guessing a design change was to remove the "big numbers" entirely?

app/assets/stylesheets/frontend/views/_history-people.scss
```diff
@@ -67,39 +67,6 @@
       width: $one-quarter;
     }
 
-    .name-title,
```
chris-gds, line 70: I feel like once these Whitehall swaps are done more of an intentional sweep of removing redundant styles would be good.

app/views/past_foreign_secretaries/_show_person.html.erb
```diff
@@ -1,4 +1,4 @@
-<header class="govuk-grid-row">
+<div class="govuk-grid-row">
```
chris-gds, line 1: nice catch

app/views/historic_appointments/show.html.erb
```diff
@@ -1,57 +1,74 @@
...
     </div>
+
+    <% @historical_account.appointment_info_array.each do |info| %>
+      <% unless info[:text].blank? %>
+        <%= render "govuk_publishing_components/components/heading", {
+          text: info[:title],
+          font_size: "m",
+          margin_bottom: 1,
+        } %>
+        <p class="govuk-body"><%= info[:text] %></p>
```
chris-gds, line 51: 👀

app/controllers/historic_appointments_controller.rb
```diff
@@ -33,4 +33,17 @@ def previous_appointments
   def previous_appointments_with_unique_people
     previous_appointments.distinct(&:person)
   end
+
+  def previous_appointments_list
```
chris-gds, line 37: 👀

chris-gds:
Luckily the test issue looks pretty straight forward:

https://github.com/alphagov/whitehall/blob/d20e358a14fddb31f17d17722b82a02515a1b575/test/functional/past_foreign_secretaries_controller_test.rb#L26-L27

app/assets/stylesheets/frontend/views/_person-profile.scss
owenatgov: There isn't. I'm not sure about extending an existing component for the sake of this one as I don't think there's necessarily harm in a tiny pattern that requires zero customisation existing in the only app it's being used in. This pattern exists for `app-c` components across our other frontend apps ( I can try `app-c`ing this if you'd like). What do you think?

app/views/past_foreign_secretaries/_show_person.html.erb
owenatgov: This choice is based of [our own guidance on image publishing](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos). The specific line:

> Describe your image in the body text, and leave the ‘Alt text’ field blank. Only write a caption if you need one.

A quick test myself does suggest that this might be a little weird for screen reader users as it's as you say, just the caption itself without any context. I'm wondering if it might make more sense to just remove the caption altogether. That way screen readers would just skip over this purely decorative section. Opinions? @nnagewad?

app/views/historic_appointments/show.html.erb
owenatgov: Yup. In summary:

- With whitehall in it's at-the-time messy link state it was a huge hassle to make it look nice
- Nikin pointed out that the numbers on the past prime ministers page are misleading as they account for multiple terms as PM

app/views/historic_appointments/show.html.erb
owenatgov: What do these eyes mean?

chris-gds: Vague emoji use 😁  - at the time of selecting I believe they meant: I like this swap technique - let me take a closer look.

app/assets/stylesheets/frontend/views/_person-profile.scss
chris-gds: Yeah I think `app-c`-ing is a good shout, however in order of preference:

- Use a current version of image card here (if possible change the design to match)
- `app-c`-ing as this types of image use are Whitehall-centric from what I can tell
- Add a new variation to cater for this new Design so it can be reused elsewhere.

app/views/past_foreign_secretaries/_show_person.html.erb
chris-gds: > I'm wondering if it might make more sense to just remove the caption altogether. That way screen readers would just skip over this purely decorative section. 

This might work but in a different way - if you change the Content of the caption to remove "Portrait of" this would qualify for [example 3](https://www.w3.org/WAI/tutorials/images/decorative/) as you have an _image with adjacent text alternative_.

owenatgov: Updating this content is pretty easy as I'm injecting "portrait of"/"picture of"/"sculpture of" into this. I'm however now a little skeptical as we've then got 2 bits of duplicate content: The h1 and this caption. I don't know if we have other examples where the context of duplicate content makes it ok?

app/assets/stylesheets/frontend/views/_person-profile.scss
chris-gds: Reviewing this, there isn't a good `image_card` that would sit here.  I considered extended this the component itself.

However, taking a step back and looking at the wider page context, this image is decorative, the main page title tells the user what this is.  In this context i'm proposing a good solution is just to use an image tag with an empty alt

chris-gds:
Note: current testing errors appears unrelated

jon-kirwan:
There seems to be a layout issue, only on mobile, I think, and which seems to exist on the current live pages too. This is probably due to the nested row classes 
e.g. -

```
<div id="page" class="govuk-width-container">
  <div class="govuk-grid-row">
    <div class="govuk-grid-column-full">
      <div class="gem-c-breadcrumbs govuk-breadcrumbs" data-module="gem-track-click">...</div>
    </div>
    <div class="govuk-grid-row">
      <div class="govuk-grid-column-two-thirds">
        <div class="gem-c-title govuk-!-margin-top-8 govuk-!-margin-bottom-8">
          <span class="govuk-caption-xl gem-c-title__context">Past Foreign Secretaries</span>
            ...
```
whitehall-frontend dev gov uk_government_history_past-foreign-secretaries_william-grenville(Moto G4)

jon-kirwan:
One other possible change is the profile images have an inline style of 100% -

```
<img class="govuk-!-margin-bottom-4" alt="" loading="lazy" style="width:100%" src="/assets/whitehall/history/past-foreign-secretaries/lord-grenville-b69e820993c591d031aabad4d1c080953d575ca9a70ee7f227ae4d3c909b09ba.jpg">
```
... and I wonder if we can replace this with `.govuk-\!-width-full` from the design system?

Otherwise all looks good 👍

jon-kirwan (changes requested):

app/views/past_foreign_secretaries/_show_person.html.erb
```diff
@@ -12,26 +12,84 @@
...
-  <div class="historic-people-list govuk-grid-column-one-quarter govuk-!-padding-left-3 govuk-!-padding-right-3">
-    <a href="#contents" class="js-showhide contents-toggle govuk-!-margin-bottom-3">Show past Foreign Secretaries</a>
-    <div id="contents">
-      <ul class="govuk-list past-foreign-secretaries-menu govuk-!-margin-bottom-6">
-        <%= past_foreign_secretary_nav(current_person) %>
-      </ul>
-      <p class="govuk-body-s">
-        <a href="/government/history/past-foreign-secretaries" class="govuk-link">See all past Foreign Secretaries</a>
-      </p>
+  <div class="govuk-grid-row">
```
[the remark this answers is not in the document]
chris-gds: Ah good spot! Nesting wasn't quite right - i've updated

app/views/historic_appointments/show.html.erb
```diff
@@ -1,55 +1,76 @@
...
-        <%= historical_fact('Major acts', format_with_html_line_breaks(@historical_account.major_acts)) %>
-        <%= historical_fact('Interesting facts', format_with_html_line_breaks(@historical_account.interesting_facts)) %>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <hr class="govuk-section-break govuk-section-break--s govuk-section-break--visible govuk-!-margin-bottom-4">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-one-third">
+      <%= image_tag @person.image_url, class: "govuk-!-margin-bottom-4", alt: "", loading: "lazy", style: "width:100%" %>
```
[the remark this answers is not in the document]
chris-gds: Nice, I was looking for _something_ like this. Added

jon-kirwan (approved):
Looks good to me!

chris-gds:
@maxgds could you merge when happy

chris-gds:
Noted this could be reverted once merged 
https://github.com/alphagov/whitehall/pull/6268

chris-gds:
Hey @maxgds can you merge if happy

PR #6214, opened 2021-07-26
Always present telephone numbers as left-to-right

brucebolt:
When a telephone number is included in the contact details of an organisation page's translation for a right-to-left language, we are incorrectly applying this to the phone number too.

This resulted in `+972 (0) 2 541 4100` being presented as `4100 541 2 (0) 972+`, which is incorrect.

By always including the telephone number as left-to-right, we will ensure the correct number is displayed.

The alternative was to replaces the spaces in the phone number with non-breaking spaces, but this still results in issues where parentheses and dashes are used.

There are multiple other issues with English content being presented as right-to-left on worldwide organisation pages, but these can be dealt with separately, as the phone number is the more pressing issue (i.e. the number is completely wrong).

Before:
Screenshot 2021-07-26 at 17 24 18

After:
Screenshot 2021-07-26 at 17 25 52

[Trello card](https://trello.com/c/9VmNiMU7)

### DIFF
--- a/app/views/contacts/_organisation_contact.html.erb
+++ b/app/views/contacts/_organisation_contact.html.erb
@@ -34,7 +34,7 @@
         <% contact.contact_numbers.each do |number| %>
           <p class="govuk-!-margin-bottom-4">
             <%= number.label %> <br />
-            <%= number.number %>
+            <span dir="ltr"><%= number.number %></span>
           </p>
         <% end %>
       </div>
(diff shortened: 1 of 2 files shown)


1pretz1 (approved):
Nice work Bruce! 🍏 

> There are multiple other issues with English content being presented as right-to-left on worldwide organisation pages, but these can be dealt with separately, as the phone number is the more pressing issue (i.e. the number is completely wrong).

Yeah, we'll focus on the actual formatting in another piece of work.

PR #6218, opened 2021-08-05
Exclude Details elements from HTML sanitisation for documents with attachments

edwardkerry:
Inline attachments are wrapped in a `<details>` HTML element which is
being sanitized and removed by Govspeak. This is preventing a JS module
from attaching and results in the accessible format request block
always being expanded.

This change allows editions with inline attachments to request
that Govspeak excludes the `details` element from sanitization
by submitting it as a string in the optional
`allowed_elements` array


brucebolt (approved):
Good find :+1:

PR #6221, opened 2021-08-06
Add more healthchecks

1pretz1:
These healthchecks (Memcached and S3) were missing from Whitehall, as
originally outlined in the CD compatibility document [1].

We've had to create a bespoke healthcheck for the connectivity to S3 [2]
which just opens a connection and fetches a bucket. If any errors occur
during this process, the healthcheck fails.

There is also a conditional to skip the S3 healthcheck if being run from the E2E tests, 
see the 2nd commit for more details. 

Dependant on: 

- [x] https://github.com/alphagov/publishing-e2e-tests/pull/424

[1]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128/compatibility.md
[2]: https://github.com/alphagov/whitehall/blob/4d1de9a6818041994f555ae40164b08cd85f043e/lib/s3_file_handler.rb

## Healthcheck endpoint request

Screenshot 2021-08-06 at 15 00 22

### DIFF
--- a/lib/healthcheck/s3.rb
+++ b/lib/healthcheck/s3.rb
@@ -5,8 +5,10 @@ module Healthcheck
     end
 
     def status
-      connection = S3FileHandler.connection
-      connection.directories.get(ENV["AWS_S3_BUCKET_NAME"])
+      if ENV["SKIP_S3_HEALTHCHECK_FOR_PUBLISHING_E2E_TESTS"].blank?
+        connection = S3FileHandler.connection
+        connection.directories.get(ENV["AWS_S3_BUCKET_NAME"])
+      end
 
       GovukHealthcheck::OK
     rescue StandardError
--- a/lib/healthcheck/s3.rb
+++ b/lib/healthcheck/s3.rb
@@ -1,0 +1,16 @@
+module Healthcheck
+  class S3
+    def name
+      :s3
+    end
+
+    def status
+      connection = S3FileHandler.connection
+      connection.directories.get(ENV["AWS_S3_BUCKET_NAME"])
+
+      GovukHealthcheck::OK
+    rescue StandardError
+      GovukHealthcheck::CRITICAL
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


benjamineskola (approved):

lib/healthcheck/s3.rb
```diff
@@ -0,0 +1,16 @@
+module Healthcheck
```
brucebolt, line 1: Is there a reason this isn't namespaced `GovukHealthcheck`?

brucebolt (approved):
Looks good to me.  I've just asked a non-blocking question about the naming of the new healthcheck, just to help me understand.

ChrisBAshton (approved):
LGTM 👍 as we'll eventually be deleting the e2e's anyway, it doesn't matter too much that we're introducing some slightly unappealing logic, and as you say, none of the e2e's exercise the S3 code at all.

I echo Bruce's question about the namespacing, but otherwise, good job! 👏

lib/healthcheck/s3.rb
1pretz1: Not from a technical standpoint. However, having separate namespaces does give the benefit of clearly defining where the healthcheck lives, e.g. `Govuk` prefix means the code resides in [govuk-app-config](https://github.com/alphagov/govuk_app_config/tree/main/lib/govuk_app_config/govuk_healthcheck), whereas no prefix indicates it's a local healthcheck. I remember seeing this pattern in some of our apps, e.g. [Content Publisher](https://github.com/alphagov/content-publisher/blob/main/config/routes.rb#L118).

Do you think it's clearer if we just go with the `GovukHealthcheck` for all healthchecks? cc @ChrisBAshton

ChrisBAshton: Thanks for the explanation - that rings a bell now! Happy to leave as-is 👍

brucebolt: Thanks for explaining.  Leaving it like this is good with me.

PR #6225, opened 2021-08-10
Return AB response headers with all variants

sihugh:
At present we're only returning Vary response headers when we get a B variant. We want to do this all the time or it's confusing for Fastly to know how to cache.

I've added a couple of tests for the other variants to make sure it's working.


injms (approved):
Super duper - thanks Simon!

PR #6229, opened 2021-08-11
Fix context hashes

maxgds:
In the pre-emptive work to fix calls to context that used the href parameter in the title component, one item was overlooked and the syntax left spurious hashes lying around in the rest. This fixes that issue and will fix the currently broken Past Prime Ministers title on live:

image


chris-gds (approved):
image

LGTM

PR #6241, opened 2021-08-17
Update past chancellor

chris-gds:
## What 

Update [Past Chancellor page](https://www.gov.uk/government/history/past-chancellors) to apply [new Design](https://www.figma.com/file/UjCeRTxle1bLQv3VFNz9Wt/Redesign-of-history-pages?node-id=0%3A1) + Design system + use Components 

## Why 

On-going work to align architecture, have pages benefit from Accessibility improvements

## Visuals 

| Before      | After |
| ----------- | ----------- |
| www gov uk_government_history_past-chancellors(iPhone 6_7_8) | mobile-after |
| ----------- | ----------- |
| www gov uk_government_history_past-chancellors | desktop-after |

## Anything else?

[Design Ref](https://www.figma.com/file/UjCeRTxle1bLQv3VFNz9Wt/Redesign-of-history-pages)

### Grouped related PRs
https://github.com/alphagov/whitehall/pull/6236
https://github.com/alphagov/whitehall/pull/6243
https://github.com/alphagov/whitehall/pull/6241

## Done When

- [x] Design Review
- [ ] QA Content

PR #6236, opened 2021-08-13
Update past prime ministers

chris-gds:
## What 

Update Past Prime Minster's page to apply [new Design](https://www.figma.com/file/UjCeRTxle1bLQv3VFNz9Wt/Redesign-of-history-pages?node-id=11%3A138) + Design system + use Components 

## Why 

On-going work to align architecture, have pages benefit from Accessibility improvements

## Visuals 

| Before      | After* |
| ----------- | ----------- |
| www gov uk_government_history_past-prime-ministers(iPhone 6_7_8 Plus) | 1www gov uk_government_history_past-prime-ministers(iPhone-6_7_8-Plus) |
|  |
| www gov uk_government_history_past-prime-ministers (1)  | screenshot-www integration publishing service gov uk-2021 08 31-12_55_10 |

## Anything else

* `image_card` was extended to factor in a [new Design](https://github.com/alphagov/govuk_publishing_components/pull/2286) and 
* [lazy loading](https://github.com/alphagov/govuk_publishing_components/pull/2270) was added to the `image_card` as this maintains existing performance benefits.

[Discussion around Design System layout + lists can be found here](https://github.com/alphagov/whitehall/pull/6236#discussion_r695518441), [an issue capturing the concerns can be found here](https://github.com/alphagov/govuk-frontend/issues/2328)

### Grouped related PRs
https://github.com/alphagov/whitehall/pull/6236
https://github.com/alphagov/whitehall/pull/6243
https://github.com/alphagov/whitehall/pull/6241

## Done when

- [x] Confirm IA of breadcrumbs
- [x] Confirm [not WCAG failure of 2.4.5](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-mult-loc.html)
- [x] Add `loading="lazy"` to image card within `govuk_publishing_gem`
PR: https://github.com/alphagov/govuk_publishing_components/pull/2270
- [x] Gem [released](https://github.com/alphagov/govuk_publishing_components/pull/2271) with `loading="lazy"` update
- [x] Design reviewed
- [x] Extend [`image_card`](https://components.publishing.service.gov.uk/component-guide/image_card) to give more options to match Design
PR: https://github.com/alphagov/govuk_publishing_components/pull/2286
- [x] [Gem released](https://github.com/alphagov/whitehall/pull/6253)
- [x] QA Content

PR #6243, opened 2021-08-19
Update past foreign secretaries

chris-gds:
## What

Update [Past Foreign Secretaries Page](https://www.gov.uk/government/history/past-foreign-secretaries) to apply new Design + Design system + use Components

## Why

On-going work to align architecture, have pages benefit from Accessibility improvements

## Visuals

| Before | After |
| ----------- | ----------- |
| www gov uk_government_history_past-foreign-secretaries(iPhone-6_7_8-Plus) | www gov uk_government_history_past-foreign-secretaries(Moto G4) |
|  |
| www gov uk_government_history_past-foreign-secretaries | www gov uk_government_history_past-foreign-secretaries (1) | 

### Done when 

- [x] [Release merged](https://github.com/alphagov/whitehall/pull/6242)
- [x] Design Reviewed
- [ ] QA Content

### Grouped related PRs
https://github.com/alphagov/whitehall/pull/6236
https://github.com/alphagov/whitehall/pull/6243
https://github.com/alphagov/whitehall/pull/6241

## Anything else?

Add link to image card heading - this is not dependant
https://github.com/alphagov/govuk_publishing_components/pull/2277

[Implementing a semantic list](https://github.com/alphagov/whitehall/pull/6243/commits/992f5a1bb28dd8c20912c03f583803273bf3fcc8) in this context + the [Design System layout](https://design-system.service.gov.uk/styles/layout/) could be done various different ways.  This might prompt a change during review.

### DIFF
--- a/app/assets/stylesheets/frontend/views/_history-people.scss
+++ b/app/assets/stylesheets/frontend/views/_history-people.scss
@@ -1,68 +1,27 @@
-.historic-people-index {
-  .clear-person {
-    clear: both;
+.historic-people-index__section-row-header {
+  @include govuk-media-query($from: tablet) {
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(4);
   }
+}
 
-  .featured-profiles {
-    margin-bottom: $gutter;
+.historic-people-index__section-row {
+  .govuk-body {
+    color: $govuk-secondary-text-colour;
 
-    .profiles {
-      @include heading-36;
-      border-bottom: $gutter-one-sixth solid govuk-colour("black");
-      margin: 0 $gutter-half $gutter-half;
+    // Targets all of the years / periods of service but the last one.
+    &:nth-last-child(n+2) {
+      margin-bottom: 0;
     }
 
-    .name {
-      @include core-19;
-      font-weight: bold;
-    }
-
-    .term {
-      color: $govuk-secondary-text-colour;
-      display: block;
-    }
-
-    .person-excerpt {
-      .inner {
-        .image-holder {
-          @include govuk-media-query($until: tablet) {
-            margin-right: govuk-spacing(2);
-          }
-
-          img {
-            max-width: $full-width;
-          }
-        }
-      }
-    }
-
-    @include media(tablet) {
-      margin-bottom: $gutter * 2;
-    }
-  }
-
-  .foreign-secretaries {
-    .heading {
-      h3 {
-        @include heading-36;
-        font-weight: bold;
-        padding: 0 $gutter-half;
-
-        span {
-          @include core-19;
-          display: block;
-        }
-        @include media(tablet) {
-          padding-top: 0;
-        }
-      }
-    }
+    // Targets the last set of years in the list
+    @include govuk-responsive-margin(7, "bottom");
   }
 }
 
 .historic-people-show {
   .person-info {
-    @include media(tablet) {
+    @include govuk-media-query($from: tablet) {
       float: left;
       width: $one-quarter;
     }
@@ -115,7 +74,7 @@
       }
     }
 
-    @include media(tablet) {
+    @include govuk-media-query($from: tablet) {
       position: relative;
       width: 75%;
 
--- a/app/views/past_foreign_secretaries/index.html.erb
+++ b/app/views/past_foreign_secretaries/index.html.erb
@@ -12,736 +12,103 @@
     }
   ]
 } %>
-<header class="govuk-grid-row">
+<div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     <%= render "govuk_publishing_components/components/title", {
       context:"History",
       title: "Past Foreign Secretaries",
     } %>
   </div>
-</header>
+</div>
 
-<div class="govuk-grid-row featured-profiles">
-  <div class="govuk-grid-column-full govuk-!-padding-0">
-    <h2 class="profiles">Selection of profiles</h2>
-    <ol>
-      <li class="person person-excerpt">
-        <div class="inner">
-          <%= render "govuk_publishing_components/components/image_card", {
-            href: "/government/history/past-foreign-secretaries/edward-wood",
-            image_src: image_path("history/past-foreign-secretaries/viscount-halifax.jpg"),
-            image_alt: "Edward Frederick Lindley Wood, Viscount Halifax",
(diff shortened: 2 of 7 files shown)


jon-kirwan:
All looks good to me. Just one small comment to add with relation to -

https://github.com/alphagov/whitehall/pull/6243/files#diff-4c36183021f514ff3edc854264a1c1c1d3b5a954d499f383bd9a56cd5729cfeaR21

```
  .govuk-heading-s--foreign-secretary-profile {
    margin-bottom: 0;
  }
```
I wondered if an alternative approach could be to replace this CSS with the govuk-!-margin-bottom-0 class?

chris-gds:
> I wondered if an alternative approach could be to replace this CSS with the govuk-!-margin-bottom-0 class?

Good call, should work the same, i'll update and remove those extra lines.

chris-gds:
@jon-kirwan updated! Tests _were_ failing but needed a rebuild.

jon-kirwan (approved):

app/views/past_foreign_secretaries/index.html.erb
```diff
@@ -12,736 +12,769 @@
...
+<div class="historic-people-index__section-row" role="list">
+  <div class="govuk-grid-row">
+    <div role="listitem" class="govuk-grid-column-one-quarter">
+      <%# Custom heading as links to external page %>
+      <h3 class="govuk-heading-s govuk-!-margin-bottom-0">
+        <a class="govuk-link" href="/government/people/dominic-raab/">Dominic Raab</a>
+      </h3>
+      <p class="govuk-body">2019 to present</p>
+    </div>
+    <div role="listitem" class="govuk-grid-column-one-quarter">
```
maxgds, line 177: This (and the other PRs) all seem very repetitive. I think there would be value in making hashes of the data then looping through them and using a single partial to populate with the data along the lines of this:

```
<h3 class="govuk-heading-s govuk-!-margin-bottom-0">
      <% if link.present? %>  
         <a class="govuk-link" href="*variable here*">*variable here*</a>
     <% else %>
          *variable here*
    <% end %>
      </h3>
<div role="listitem" class="govuk-grid-column-one-quarter">
        text: *variable here*,
        heading_level: 3,
        font_size: "s",
      } %>
      <p class="govuk-body">*variable here*</p>
    </div>
```
You could potentially expand it out to the image card examples as well? I don't think it's common enough to be worth making a component, but it's not at all DRY to be seeing repeated HTML markup like this.

maxgds (changes requested):
While the output is sound, I have reservations around the div vs li markup and am concerned about the level of repetition that exists. I think the 3 pages dealt with in the one trello ticket could be sharing more code because of their similarities and would like you to explore that a bit more rather than merging as-is.

app/views/past_foreign_secretaries/index.html.erb
```diff
@@ -12,736 +12,769 @@
...
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "21st century",
+      padding: true,
+      border_top: 2,
+      font_size: "l"
+    } %>
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+  </div>
+</div>
+<div class="historic-people-index__section-row" role="list">
```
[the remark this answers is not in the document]
chris-gds: Yeah it's the same problem across 3 - if I markup as a list I lose the Design system, if I use the Design System I lose the markup but replicate with aria.  Once a consensus is reached i'll update across all

app/views/past_foreign_secretaries/index.html.erb
chris-gds: As this is hardcode - do you have a good example to ref to use for this context, is this worth adding to `content_item` or locale .[yml](https://github.com/alphagov/whitehall/blob/175fec283eab3efa5e61698d4eb90cf47cefe869/config/locales/en.yml) file?

maxgds: I don't think it fits in a locale file as the strings are not really translatable (names/dates) and they're definitely not page furniture and content items are complicated. I'd keep it in situ as arrays of hashes like: 
```
C21 = [
{name:"Boris Johnson", dates: "2020-2021"}, 
{name:"Theresa May", dates: "xxxx-xxxx", link: "/government/whatever"}, 
]

C21.each do
	call partial
end```
Excuse any bad syntax, but you get the drift?

Comment (chris-gds) [reply app/views/past_foreign_secretaries/index.html.erb]: So! This has been split out, added to the controller and the partial iterates over.

File: app/assets/stylesheets/frontend/views/_history-people.scss [hunk 3]
```diff
@@ -1,68 +1,37 @@
...
+.historic-people-index__section-row-header {
+  @include govuk-media-query($from: tablet) {
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(4);
   }
 
-  .featured-profiles {
-    margin-bottom: $gutter;
+  // unique image card styles for "past-foreign-secretaries" & "past-past-chancellor"
+  .gem-c-image-card__list-item-link {
```
maxgds, line 8: This doesn't feel great, overriding component styles. Willing to let it pass if we absolutely have to but I'd suggest considering either/or 1) Checking with design that the default isn't good enough, 2) making this style the default for the "unlinked links".

maxgds (changes requested):
This is looking much better from a DRY perspective, but I think I haven't communicated very well what I was aiming for. The three pages on this one trello card are using the same patterns throughout, there seems to be value in sharing not just the css, but the partial here. I  think this is a good first step towards uniting them like that, but if that's the end game the person and image partial names and locations should be more generic. I know the underlying code that builds each page differs but do you think this is achievable?

app/assets/stylesheets/frontend/views/_history-people.scss
andysellick: We shouldn't be overriding component styles in applications - it breaks component isolation - e.g. if we were to change the name of this class in the component, this override would suddenly stop working and we wouldn't know about it.

If the link styles definitely need to be different in this one situation, please add it as an option to the component. Do they definitely need to be different?

chris-gds: Agreed. 🤔  
Let me do the reverse and reduce the font everywhere.

chris-gds:
Moving into Draft (aim to close) as combined PR raised: https://github.com/alphagov/whitehall/pull/6264

PR #6259, opened 2021-08-31
Adds Take Part Pages patch link to Get Involved item

Tetrino:
The migration away from Whitehall for the Get Involved page needed more
detail from the Take Part Pages than what was supplied via API calls.
This is the most effective way to bring that information across.

This branch relies on other changes to be in place:
Publishing-API change: https://github.com/alphagov/publishing-api/pull/1969
Schema change: https://github.com/alphagov/govuk-content-schemas/pull/1069

Trello: https://trello.com/c/z6H3HPNw/2716-add-take-part-pages-as-links-to-government-get-involveds-content-item

### DIFF
--- a/test/integration/take_part_page_test.rb
+++ b/test/integration/take_part_page_test.rb
@@ -70,4 +70,107 @@ class TakePartPageTest < ActiveSupport::TestCase
       )
     end
   end
+
+  test "TakePartPage patches links in the correct order in the Get Involved Page when created" do
+    Sidekiq::Testing.inline! do
+      # Working with a specific ID
+      get_involved_content_id = "dbe329f1-359c-43f7-8944-580d4742aa91"
+
+      # Build a couple Take Part pages with different ordering
+      @take_part_page.attributes = { title: "First Page", ordering: 1 }
+      @take_part_page.save!
+
+      third_take_part_page = build(:take_part_page)
+      third_take_part_page.attributes = { title: "Third Page", ordering: 3 }
+      third_take_part_page.save!
+
+      assert_publishing_api_patch_links(
+        get_involved_content_id,
+        { links: {
+          take_part_pages: [@take_part_page.content_id, third_take_part_page.content_id],
+        } },
+      )
+
+      second_take_part_page = build(:take_part_page)
+      second_take_part_page.attributes = { title: "Second Page", ordering: 2 }
+      second_take_part_page.save!
+
+      assert_publishing_api_patch_links(
+        get_involved_content_id,
+        { links: {
+          take_part_pages: [@take_part_page.content_id, second_take_part_page.content_id, third_take_part_page.content_id],
+        } },
+      )
+    end
+  end
+
+  test "TakePartPage patches links in the correct order in the Get Involved Page when modified" do
+    Sidekiq::Testing.inline! do
+      # Working with a specific ID
+      get_involved_content_id = "dbe329f1-359c-43f7-8944-580d4742aa91"
+
+      # Build a couple Take Part pages with different ordering
+      @take_part_page.attributes = { title: "First Page", ordering: 2 }
+      @take_part_page.save!
+
+      third_take_part_page = build(:take_part_page)
+      third_take_part_page.attributes = { title: "Third Page", ordering: 3 }
+      third_take_part_page.save!
+
+      assert_publishing_api_patch_links(
+        get_involved_content_id,
+        { links: {
+          take_part_pages: [@take_part_page.content_id, third_take_part_page.content_id],
+        } },
+      )
+
+      # Easiest way to check a patch on edit is to change the ordering
+      @take_part_page.attributes = { title: "First Page", ordering: 4 }
+      @take_part_page.save!
+
+      assert_publishing_api_patch_links(
+        get_involved_content_id,
+        { links: {
+          take_part_pages: [third_take_part_page.content_id, @take_part_page.content_id],
+        } },
+      )
+    end
+  end
+
+  test "TakePartPage patches links in the correct order in the Get Involved Page when deleted" do
+    Sidekiq::Testing.inline! do
+      Services.asset_manager.stubs(:whitehall_asset).returns("id" => "http://asset-manager/assets/asset-id")
+      # Working with a specific ID
+      get_involved_content_id = "dbe329f1-359c-43f7-8944-580d4742aa91"
+
+      # Build a couple Take Part pages with different ordering
+      @take_part_page.attributes = { title: "First Page", ordering: 1 }
+      @take_part_page.save!
+
+      second_take_part_page = build(:take_part_page)
+      second_take_part_page.attributes = { title: "Second Page", ordering: 2 }
+      second_take_part_page.save!
+
+      third_take_part_page = build(:take_part_page)
+      third_take_part_page.attributes = { title: "Third Page", ordering: 3 }
+      third_take_part_page.save!
+
+      assert_publishing_api_patch_links(
+        get_involved_content_id,
+        { links: {
+          take_part_pages: [@take_part_page.content_id, second_take_part_page.content_id, third_take_part_page.content_id],
+        } },
+      )
+
+      # Destroy the middle item
+      second_take_part_page.destroy!
+
+      assert_publishing_api_patch_links(
+        get_involved_content_id,
+        { links: {
+          take_part_pages: [@take_part_page.content_id, third_take_part_page.content_id],
+        } },
+      )
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/admin/take_part_pages_controller.rb
```diff
@@ -53,4 +55,20 @@ def take_part_page_params
       :title, :summary, :body, :image, :image_alt_text, :image_cache
     )
   end
+
+  def patch_getinvolved_page_links
```
brucebolt, line 59: At the moment, this method isn't being called when a TakePartPage is created or deleted, which means the links will become out of sync unless another one is updated or reordered.

It might be an idea to move the `patch_getinvolved_page_links` method to be a callback (e.g. [`after_commit`](https://api.rubyonrails.org/v6.1.4/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_commit)) on the `TakePartPage` model.  Then it'll also cover the create and destroy methods, and you won't need to call it at all in the controller.

app/controllers/admin/take_part_pages_controller.rb
```diff
@@ -53,4 +55,20 @@ def take_part_page_params
...
+    get_involved_cid = "dbe329f1-359c-43f7-8944-580d4742aa91"
+    pages = []
+
+    TakePartPage.in_order.each do |page|
+      pages << page.content_id
+    end
```
brucebolt, line 65: As @ChrisBAshton suggested on Slack, this can be done in one line using the `map` method.  This also means you won't need to initialise the `pages` array, as `map` always returns an array.
```suggestion
    pages = TakePartPage.in_order.map(&:content_id)
```

app/controllers/admin/take_part_pages_controller.rb
```diff
@@ -53,4 +55,20 @@ def take_part_page_params
       :title, :summary, :body, :image, :image_alt_text, :image_cache
     )
   end
+
+  def patch_getinvolved_page_links
+    get_involved_cid = "dbe329f1-359c-43f7-8944-580d4742aa91"
```
brucebolt, line 60: A little nitpick, but could `cid` be fully spelt out as `content_id` to keep the naming consistent with other places.
```suggestion
    get_involved_content_id = "dbe329f1-359c-43f7-8944-580d4742aa91"
```

brucebolt (commented):
This is looking good @tetrino.

I've added some comments where I think there's missing scenarios (new `TakePartPage` being created, or an existing one deleted) where the links won't get updated.  For this reason, I think it's better to use a callback on the model to patch the links, rather than calling the method explicitly in each controller method.  There are [several examples](https://github.com/alphagov/whitehall/blob/5cba22ae73efabd6e93f8da75b28a2ae88ffdec7/app/models/topical_event.rb#L13) of this already in Whitehall.

Also, would you be able to add some tests to ensure that when a model is updated/reordered/created/deleted, the patched links are correctly sent to Publishing API in the correct order?

Tetrino:
@brucebolt I moved the respective code into the actual model as I was a fool and put it in the wrong place. All other changes are as suggested.

app/controllers/admin/take_part_pages_controller.rb
Tetrino: Added this as an "after_action" call in the Controller as this was the most logical place to put it that ensured everything was being updated correctly - the model didn't want to note certain changes.

brucebolt: I don't think `after_action` is necessarily the right callback to use here, as that will execute after every action (including `index` and `show`, where a view is being rendered and no data being changed).

What was the issue with adding this to the model?  I think that would be a nicer solution, as it would ensure the links get updated regardless of which controller makes the changes.

Tetrino: The model doesn't fire the `after_commit` upon updating orders, for one. You can actually define which actions trigger the execution with [an `on:` definition](https://api.rubyonrails.org/v6.1.4/classes/ActiveRecord/Transactions/ClassMethods.html#method-i-after_commit) which for me may be the best solution here? I meant to put it in place but I wanted to see your thoughts.

brucebolt: The `after_commit` doesn't get called when the pages are redordered because there is a `update_column` [here](https://github.com/alphagov/whitehall/commit/1100ced7622cecbab9989a86a8c769bafb6cafa9#diff-4553ff371557c14602c2d83baea491d31490b06f0be595b350cf26fad5f6e392R48) which [skips callbacks](https://apidock.com/rails/v3.2.8/ActiveRecord/Persistence/update_column).

If you change that line to `page.update(ordering: ids_in_new_ordering.index(page.id.to_s) + 1)`, you should find the callback won't be skipped.

As the reordering happens in a transaction, the `after_commit` will only get called once at the end of the commit, after the items have been correctly placed in order in the database.  I tried this locally on a Rails console (adding a `puts "patching"` statement into `patch_getinvolved_page_links` method to see when this actually occurs) and you can see the patching only takes place after all the reordering has taken place:
```
[21] pry(main)> TakePartPage.reorder!([3,1,2])
  TRANSACTION (0.4ms)  BEGIN
  TakePartPage Load (0.5ms)  SELECT `take_part_pages`.* FROM `take_part_pages` WHERE `take_part_pages`.`id` IN (3, 1, 2) ORDER BY `take_part_pages`.`id` ASC LIMIT 1000
  TakePartPage Update (0.4ms)  UPDATE `take_part_pages` SET `take_part_pages`.`ordering` = 2, `take_part_pages`.`updated_at` = '2021-09-01 14:38:14' WHERE `take_part_pages`.`id` = 1
  TakePartPage Load (0.4ms)  SELECT `take_part_pages`.* FROM `take_part_pages` WHERE `take_part_pages`.`id` = 1 LIMIT 1
  TakePartPage Update (0.4ms)  UPDATE `take_part_pages` SET `take_part_pages`.`ordering` = 3, `take_part_pages`.`updated_at` = '2021-09-01 14:38:14' WHERE `take_part_pages`.`id` = 2
  TakePartPage Load (0.4ms)  SELECT `take_part_pages`.* FROM `take_part_pages` WHERE `take_part_pages`.`id` = 2 LIMIT 1
  TakePartPage Update (0.4ms)  UPDATE `take_part_pages` SET `take_part_pages`.`ordering` = 1, `take_part_pages`.`updated_at` = '2021-09-01 14:38:14' WHERE `take_part_pages`.`id` = 3
  TakePartPage Load (0.5ms)  SELECT `take_part_pages`.* FROM `take_part_pages` WHERE `take_part_pages`.`id` = 3 LIMIT 1
  TakePartPage Update All (0.4ms)  UPDATE `take_part_pages` SET `take_part_pages`.`ordering` = 4 WHERE (id NOT IN ('3','1','2'))
  TRANSACTION (1.3ms)  COMMIT
patching
  TakePartPage Load (0.5ms)  SELECT `take_part_pages`.* FROM `take_part_pages` ORDER BY `take_part_pages`.`ordering` ASC
```

Tetrino: Oooh perfect, thank you, I'll throw this in.

Tetrino: Right this now works as designed again - now including edits, deletions and additions, the only issue is [within the rake task](https://github.com/alphagov/whitehall/blob/main/lib/tasks/publishing_api.rake#L98) - it doesn't update the patching. That is a separate piece of work I think and can be approached later.

brucebolt: In theory, re-publishing the Take Part page to the Publishing API should result in link expansion updating everywhere the document is linked to, so hopefully nothing for us to worry about for now!

brucebolt (commented):
Thanks for making the changes @Tetrino.  The code looks good, but would you be able to add some tests to ensure that when a model is updated/reordered/created/deleted, the patched links are correctly sent to Publishing API in the correct order?

Tetrino:
Well ahead of you @brucebolt that's currently what I'm working on.

brucebolt (approved):

test/integration/take_part_page_test.rb
```diff
@@ -70,4 +70,107 @@ class TakePartPageTest < ActiveSupport::TestCase
       )
     end
   end
+
+  test "TakePartPage patches links in the correct order in the Get Involved Page when created" do
+    Sidekiq::Testing.inline! do
+      # Working with a specific ID
+      get_involved_content_id = "dbe329f1-359c-43f7-8944-580d4742aa91"
```
ChrisBAshton, line 77: I'd be tempted to move this into a shared variable in the `setup` block, i.e. `@get_involved_content_id`, to remove the duplication of this ID throughout the file.

ChrisBAshton (commented):
Code LGTM, just one nitpick comment below 👍

PR #6282, opened 2021-09-14
Fix race condition deleting test files prematurely

richardTowers:
In 154a173 we switched to running tests in parallel.

One edge case this has thrown up is that the LinkReporterCsvServiceTest
class creates files in the setup method and cleans them up in the
teardown. When this is running in parallel there's a race condition
where one of the parallel workers runs the teardown method before all of
the workers are finished with the files.

I've fixed this by giving the test class a reference to
`@@worker_number`. This means they can create separate files for
separate workers, and there's no risk of them accidentally deleting
files being used by another worker when they're done.

The `parallelize_setup` method is provided by ActiveSupport::TestCase
(https://api.rubyonrails.org/v6.1.4/classes/ActiveSupport/TestCase.html).
It runs in the context of the class (not an instance), hence the use of
a `@@class_variable` instead of an `@instance_variable`. I also
considered using a `$global_variable` (which would be fine since each
test worker is its own entire process) but it's not needed, and I'm
pretty sure I couldn't have snuck it through code review.

### DIFF
--- a/test/unit/services/link_reporter_csv_service_test.rb
+++ b/test/unit/services/link_reporter_csv_service_test.rb
@@ -1,13 +1,26 @@
 require "test_helper"
 
 class LinkReporterCsvServiceTest < ActiveSupport::TestCase
+  # Tests are run in parallel. Any shared resources (like files) which the
+  # tests use need to be managed carefully. For example the setup method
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👍  nice find!

PR #6287, opened 2021-09-16
Add bespoke section topical events page to link out to 2022 events orgs

maxgds:
This adds a bespoke area to the topical events page for 2022 events. This is because there's a request via zendesk to replicate the behaviour seen on the war centenary topical event. https://govuk.zendesk.com/agent/tickets/4662915

Two new logos have been added and are displayed via css updates. The text is held in translation files which have all been updated (in a separate commit to reduce noise) as per the centenary code.

Currently the Birmingham org is linked via whitehall due to being listed as a lead org for the event, that has been changed on integration where this branch was deployed to resulting in the screenshot below. The demotion from "lead" would need to be reflected on live before merge/deploy to avoid a repeating link and bad layout.

https://www.gov.uk/government/topical-events/2022-events-platinum-jubilee-commonwealth-games-festival-uk?3333

### Screenshot
Screenshot 2021-09-16 at 12 34 03


app/views/topical_events/show.html.erb
```diff
@@ -155,6 +155,15 @@
                   <a href="http://www.warmemorials.org"><%= t("topical_event.lead_organisations.war_memorials_trust") %></a>
                 </li>
               <% end %>
+
+              <% if @topical_event.slug == '2022-events-platinum-jubilee-commonwealth-games-festival-uk' %>
+                <li class="non-govuk birmingham_2022_commonwealth_games">
+                  <a href=""><%= t("topical_event.lead_organisations.birmingham_2022_commonwealth_games") %></a>
```
ChrisBAshton, line 161: Is this the bit RE "Currently the Birmingham org is linked via whitehall due to being listed as a lead org for the event"? I don't understand how Whitehall populates this `href` (if I've understood correctly).

The rest of the PR looks good to me 👍

maxgds: Whitehall has them listed as a lead org but that doesn't allow them to display a logo, only "live" government departments can do that, hence the need to hard code. I will go and adjust so they're listed as involved but no longer lead to prevent a duplicate link.

maxgds: Ah, but I've failed to hard code the link, I'll fix that now. Thanks!

ChrisBAshton (approved):

PR #6292, opened 2021-09-16
Fix incorrect parameter names on image card components

huwd:
This change fixes two issues:
- A Prime Minister went missing from the historic Prime Ministers page
- Details were missing from the past foreign secretaries page

Around a week ago we modernised the views on these pages to use [image card](https://components.publishing.service.gov.uk/component-guide/image_card), After which govuk publishing components release v27, which changes the parameters that are required here.

The combination of the missing parameters and a lack of href on one PM caused them to go missing.

By updating the parameter names to those required by v27 of the publishing componenets we see the past PM return, and past foreign secretaries get their details back


ChrisBAshton (approved):

chris-gds:
Related PR https://github.com/alphagov/govuk_publishing_components/pull/2300

PR #6304, opened 2021-09-27
Make featured links translatable

brucebolt:
Makes the fields on the featured links in Whitehall translatable.  If a translation is not available for a link, the link from the default locale is used.

Users can add/edit translations for featured links in any language that the linkable (i.e. Organisation or Worldwide Organisation) page is available in.

We do not need to make any changes in Publishing API or Collections, as Whitehall will handle the translation and only present the relevant links to Publishing API for the language of the content it is publishing (using the `globalize` gem).

All existing featured links will be labelled as English and will need to be manually edited by the departmental editors if they wish to add translations.

This PR was inspired by https://github.com/alphagov/whitehall/pull/6173.

### Screenshots

[Trello card](https://trello.com/c/XOB88Gcp)


app/views/admin/shared/_featured_link_fields.html.erb
```diff
@@ -15,8 +15,13 @@
 
   <%= form.fields_for(:featured_links, featured_links) do |featured_links_form| %>
     <div class="js-duplicate-fields-set well">
-      <%= featured_links_form.text_field :title, class: 'input-md-4' %>
-      <%= featured_links_form.text_field :url, label_text: 'URL' %>
+      <% if @translation_locale && @translation_locale.code != I18n.default_locale %>
```
ChrisBAshton, line 18: This conditional feels like it could do with a view test.

ChrisBAshton (commented):
LGTM, I do wonder if there's a bit of a coverage gap in the view though, would it be possible to add one more test?

db/migrate/20210927095227_move_featured_links_to_featured_link_translations.rb
```diff
@@ -0,0 +1,22 @@
...
+        featured_link_id: link[0],
+        title: link[1],
+        url: link[2],
+        locale: "en",
+      )
+    end
+
+    change_table :featured_links do |t|
+      t.remove :url, :title
+    end
```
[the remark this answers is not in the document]
brucebolt: I believe (as this migration includes a schema change), the [entire migration will already run in a transaction](https://edgeguides.rubyonrails.org/active_record_migrations.html#migration-overview).

app/views/admin/shared/_featured_link_fields.html.erb
brucebolt: I've added a feature test for adding translations to organisations, which covers the conditional in this partial.

ChrisBAshton (approved):
Excellent stuff ⭐ Thanks for adding that feature test.

PR #6329, opened 2021-10-14
Add big number component to the get involved page

owenatgov:
## What
Applies the [big number component](https://components.publishing.service.gov.uk/component-guide/big_number) on the [get involved page](https://www.gov.uk/government/get-involved).

## Why
So that we're using components in as many places as we possibly can, reducing tech debt and risk of repeated implementation.

## Visual changes
| Before | After |
| --- | --- |
| Screenshot 2021-10-14 at 16 23 31 | Screenshot 2021-10-14 at 16 37 10 |

### DIFF
--- a/app/assets/stylesheets/frontend/views/_get-involved.scss
+++ b/app/assets/stylesheets/frontend/views/_get-involved.scss
@@ -50,39 +50,6 @@
       }
     }
 
-    .count {
-      @include govuk-media-query($from: desktop) {
-        width: $half;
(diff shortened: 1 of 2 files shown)


Tetrino:
I highly recommend _not_ merging this in. I am currently wrapping up the migration of this page to the Government Frontend repo and this change will simply not be applied. I will manually do the same on this branch. https://github.com/alphagov/government-frontend/pull/1939

PR #6363, opened 2021-11-15
Use HTML attachment content-id in URLs

benjamineskola:

https://trello.com/c/rmphfmD6/2760-change-whitehall-html-attachments-to-use-content-ids-in-url-5

### DIFF
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -70,7 +70,12 @@ class HtmlAttachment < Attachment
     path_or_url = full_url ? "url" : "path"
     path_helper = "#{type}_html_attachment_#{path_or_url}"
 
-    Whitehall.url_maker.public_send(path_helper, attachable.slug, self, options)
+    # This depends on the rails url helpers to construct a url based on config/routes.rb:
+    # composed of document type, slug for the parent document, and identifier for the attachment;
+    # for non-english attachments the identifier is the content_id, for english
+    # attachments it is self i.e. the slug
+    identifier = sluggable_locale? ? self : content_id
+    Whitehall.url_maker.public_send(path_helper, attachable.slug, identifier, options)
   end
 
   def should_generate_new_friendly_id?
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -74,7 +74,11 @@ class HtmlAttachment < Attachment
     # composed of document type, slug for the parent document, and identifier for the attachment;
     # for non-english attachments the identifier is the content_id, for english
     # attachments it is self i.e. the slug
-    identifier = sluggable_locale? ? self : content_id
+    identifier = if attachable.updated_at < Time.zone.local(2021, 11, 29)
+                   self
+                 else
+                   sluggable_locale? ? self : content_id
+                 end
     Whitehall.url_maker.public_send(path_helper, attachable.slug, identifier, options)
   end
 
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -68,17 +68,18 @@ class HtmlAttachment < Attachment
 
     type = attachable.path_name
     path_or_url = full_url ? "url" : "path"
-    path_helper = "#{type}_html_attachment_#{path_or_url}"
 
     # This depends on the rails url helpers to construct a url based on config/routes.rb:
     # composed of document type, slug for the parent document, and identifier for the attachment;
     # for non-english attachments the identifier is the content_id, for english
     # attachments it is self i.e. the slug
-    identifier = if attachable.updated_at < Time.zone.local(2021, 11, 29)
-                   self
-                 else
-                   sluggable_locale? ? self : content_id
-                 end
+    if sluggable_locale? || attachable.updated_at < Time.zone.local(2021, 11, 29)
+      identifier = self
+      path_helper = "#{type}_html_attachment_#{path_or_url}"
+    else
+      identifier = content_id
+      path_helper = "#{type}_html_attachment_by_content_id_#{path_or_url}"
+    end
     Whitehall.url_maker.public_send(path_helper, attachable.slug, identifier, options)
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -106,7 +106,8 @@ Whitehall::Application.routes.draw do
 
     get "/publications(.:locale)", as: "publications", to: "publications#index", constraints: { locale: valid_locales_regex }
     get "/publications/:id(.:locale)", as: "publication", to: "_#_", constraints: { locale: valid_locales_regex }
-    get "/publications/:publication_id/:content_id" => "_#_", as: "publication_html_attachment"
+    get "/publications/:publication_id/:content_id" => "_#_", as: "publication_html_attachment_by_content_id"
+    get "/publications/:publication_id/:id" => "_#_", as: "publication_html_attachment"
 
     # TODO: Remove when paths can be generated without a routes entry
     get "/case-studies/:id(.:locale)", as: "case_study", to: "case_studies#show", constraints: { locale: valid_locales_regex }
@@ -122,7 +123,8 @@ Whitehall::Application.routes.draw do
     resources :statistics_announcements, path: "statistics/announcements", only: %i[index show]
     get "/statistics(.:locale)", as: "statistics", to: "statistics#index", constraints: { locale: valid_locales_regex }
     get "/statistics/:id(.:locale)", as: "statistic", to: "_#_", constraints: { locale: valid_locales_regex }
-    get "/statistics/:statistics_id/:content_id" => "_#_", as: "statistic_html_attachment"
+    get "/statistics/:statistics_id/:content_id" => "_#_", as: "statistic_html_attachment_by_content_id"
+    get "/statistics/:statistics_id/:id" => "_#_", as: "statistic_html_attachment"
 
     get "/consultations/:id(.:locale)", as: "consultation", to: "consultations#show", constraints: { locale: valid_locales_regex }
     resources :consultations, only: %i[index] do
@@ -133,8 +135,10 @@ Whitehall::Application.routes.draw do
       end
     end
     get "/consultations/:consultation_id/:id" => "_#_", as: "consultation_html_attachment"
+    get "/consultations/:consultation_id/outcome/:content_id" => "_#_", as: "consultation_outcome_html_attachment_by_content_id"
     get "/consultations/:consultation_id/outcome/:id" => "_#_", as: "consultation_outcome_html_attachment"
-    get "/consultations/:consultation_id/public-feedback/:content_id" => "_#_", as: "consultation_public_feedback_html_attachment"
+    get "/consultations/:consultation_id/public-feedback/:content_id" => "_#_", as: "consultation_public_feedback_html_attachment_by_content_id"
+    get "/consultations/:consultation_id/public-feedback/:id" => "_#_", as: "consultation_public_feedback_html_attachment"
 
     resources :topical_events, path: "topical-events", only: [:show] do
       # Controller removed. Whitehall frontend no longer serves these
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -106,7 +106,7 @@ Whitehall::Application.routes.draw do
 
     get "/publications(.:locale)", as: "publications", to: "publications#index", constraints: { locale: valid_locales_regex }
     get "/publications/:id(.:locale)", as: "publication", to: "_#_", constraints: { locale: valid_locales_regex }
-    get "/publications/:publication_id/:id" => "_#_", as: "publication_html_attachment"
+    get "/publications/:publication_id/:content_id" => "_#_", as: "publication_html_attachment"
 
     # TODO: Remove when paths can be generated without a routes entry
     get "/case-studies/:id(.:locale)", as: "case_study", to: "case_studies#show", constraints: { locale: valid_locales_regex }
@@ -122,7 +122,7 @@ Whitehall::Application.routes.draw do
     resources :statistics_announcements, path: "statistics/announcements", only: %i[index show]
     get "/statistics(.:locale)", as: "statistics", to: "statistics#index", constraints: { locale: valid_locales_regex }
     get "/statistics/:id(.:locale)", as: "statistic", to: "_#_", constraints: { locale: valid_locales_regex }
-    get "/statistics/:statistics_id/:id" => "_#_", as: "statistic_html_attachment"
+    get "/statistics/:statistics_id/:content_id" => "_#_", as: "statistic_html_attachment"
 
     get "/consultations/:id(.:locale)", as: "consultation", to: "consultations#show", constraints: { locale: valid_locales_regex }
     resources :consultations, only: %i[index] do
@@ -134,7 +134,7 @@ Whitehall::Application.routes.draw do
     end
     get "/consultations/:consultation_id/:id" => "_#_", as: "consultation_html_attachment"
     get "/consultations/:consultation_id/outcome/:id" => "_#_", as: "consultation_outcome_html_attachment"
-    get "/consultations/:consultation_id/public-feedback/:id" => "_#_", as: "consultation_public_feedback_html_attachment"
+    get "/consultations/:consultation_id/public-feedback/:content_id" => "_#_", as: "consultation_public_feedback_html_attachment"
 
     resources :topical_events, path: "topical-events", only: [:show] do
       # Controller removed. Whitehall frontend no longer serves these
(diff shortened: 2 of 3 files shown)


test/unit/html_attachment_test.rb
```diff
@@ -116,6 +116,7 @@ class HtmlAttachmentTest < ActiveSupport::TestCase
     attachment.title = "a-new-title"
     attachment.save!
     attachment.reload
+    puts attachment.url
```
ChrisBAshton, line 119: This can be removed.

test/unit/presenters/publishing_api/html_attachment_presenter_test.rb
```diff
@@ -21,7 +21,7 @@ def present(record)
     html_attachment = HtmlAttachment.last
 
     expected_hash = {
-      base_path: "/government/publications/#{edition.document.slug}/#{html_attachment.slug}",
+      base_path: "/government/publications/#{edition.document.slug}/#{html_attachment.content_id}",
```
ChrisBAshton, line 24: The English HTML attachments should retain the title-to-slug behaviour. We only want to apply the `content_id` to non-English HTML attachments (in place of their edition ID).

config/routes.rb
```diff
@@ -134,7 +134,7 @@ def external_redirect(path_prefix, target)
     end
     get "/consultations/:consultation_id/:id" => "_#_", as: "consultation_html_attachment"
     get "/consultations/:consultation_id/outcome/:id" => "_#_", as: "consultation_outcome_html_attachment"
-    get "/consultations/:consultation_id/public-feedback/:id" => "_#_", as: "consultation_public_feedback_html_attachment"
+    get "/consultations/:consultation_id/public-feedback/:content_id" => "_#_", as: "consultation_public_feedback_html_attachment"
```
ChrisBAshton, line 137: These changes all look good on the surface (without knowing too much about all the different HTML attachment types!). But I suspect we'll want to keep the old routes for the time being, so that existing HTML attachment URLs continue to work. We can then remove the old routes in https://trello.com/c/3BeDlugo/2761-republish-whitehall-documents-containing-non-english-html-attachments-5 (please add to that card as an acceptance criteria once this PR is merged!)

app/models/html_attachment.rb
```diff
@@ -70,7 +70,7 @@ def url(options = {})
     path_or_url = full_url ? "url" : "path"
     path_helper = "#{type}_html_attachment_#{path_or_url}"
 
-    Whitehall.url_maker.public_send(path_helper, attachable.slug, self, options)
+    Whitehall.url_maker.public_send(path_helper, attachable.slug, content_id, options)
```
ChrisBAshton, line 73: I tried to find the class definition for UrlMaker to figure out what parameters the constructor takes, but all I can see is [whitehall/url_maker.rb](https://github.com/alphagov/whitehall/blob/main/lib/whitehall/url_maker.rb), which pulls in behaviour from other places. I'm not really sure what the `self` was doing before this change. Dug up a couple of links (https://github.com/alphagov/whitehall/blob/5cba22ae73efabd6e93f8da75b28a2ae88ffdec7/test/support/url_helpers.rb#L13, https://github.com/rails/rails/issues/29992) but still a bit nonplussed).

Could you explain in a comment in the code, what this is doing? E.g. what is the relationship between `attachable.slug` and `content_id`?

ChrisBAshton (commented):
These changes are looking promising 👍 some comments/questions below.

app/models/html_attachment.rb
benjamineskola: This bit of the code was particularly horrible, this is what I think is going on (which I'll add in a comment if that seems to be right).

First I had assumed that `public_send` was a method specific to `url_maker` but when it clicked that it was a form of `send` it clarified that it was constructing a method name as a string (`path_helper`) and then calling it. Unfortunately these are all dynamically-generated methods based (I think) on Rails' routing. So Rails is creating methods like `publication_html_attachment_path` with two parameters `publication_id` and `id`, the latter of which I've changed to `content_id` both in the routes file and this method.

However my knowledge of this is somewhat tenuous, and e.g. the tests still passed when I didn't change the routes file even though I _think_ it's necessary, so I'd like to test in integration before going any further.

And the need to support the old URL style still also makes me wonder if the method still needs to _generate_ old URLs for older content still, which will add another level of complexity

ChrisBAshton: 🤮 Argh yes, I hadn't considered we'd need to maintain two sets of URL-generating methods! I'd look into doing that in the first instance.

If it's a _total_ nightmare, we could think about just getting this PR in a state where it's good to go (with the new URL style only), then work on the rake task in [the next card](https://trello.com/c/3BeDlugo/2761-republish-whitehall-documents-containing-non-english-html-attachments-5), and finally merge & deploy both PRs, and run the rake task. I can't think of a way of doing that without some temporary downtime on all non-English HTML attachments, though, so would be a candidate for out-of-hours work, as well as being high-stakes / big bang release. So I'd certainly feel more comfortable if we could get both styles of URL working simultaneously.

benjamineskola: I _think_ how it would end up looking is something like this:

```ruby
if locale == "en"
  # keep the old code here
else
  # new code
end
```

But then for the time being the new code should probably actually be something like:

```ruby
if publication_date < switchover_date
  # old code
else
  # new code
end
```

And the routes might take some work to disambiguate too. Since the locale isn't included in the URL, we can't use that but if the format of the content ID and edition ID is guaranteed to be different then we might be able to use a constraint:

```ruby
get "/publications/:publication_id/:content_id" => "_#_", as: "publication_html_attachment", constraints: { content_id: valid_content_id_regex }
get "/publications/:publication_id/:id" => "_#_", as: "publication_html_attachment"
```

test/unit/html_attachment_test.rb
ChrisBAshton: Still needs removing (from the first commit).

test/unit/presenters/publishing_api/publication_presenter_test.rb
```diff
@@ -39,7 +39,7 @@ def present(edition)
...
+        documents: ["<section class=\"attachment embedded\" id=\"attachment_#{publication.attachments.first.id}\">\n  <div class=\"attachment-thumb\">\n      <a aria-hidden=\"true\" class=\"thumbnail\" tabindex=\"-1\" href=\"/government/publications/publication-title/#{publication.attachments.first.title}\"><img alt=\"\" src=\"https://www.test.gov.uk/assets/whitehall/pub-cover-html.png\" /></a>\n  </div>\n  <div class=\"attachment-details\">\n    <h3 class=\"title govuk-!-font-size-27 govuk-!-font-weight-regular\"><a href=\"/government/publications/publication-title/#{publication.attachments.first.title}\">#{publication.attachments.first.title}</a></h3>\n    <p class=\"metadata\">\n        <span class=\"type\">HTML</span>\n    </p>\n\n\n  </div>\n</section>"],
```
ChrisBAshton: As this change reverses some of the changes you made in the previous commit, I'd be tempted to squash these middle two commits together. I.e. make one super commit of "use content IDs for non-English HTML attachments".

test/unit/html_attachment_test.rb
```diff
@@ -48,6 +48,7 @@ class HtmlAttachmentTest < ActiveSupport::TestCase
     edition = create(:draft_publication, :with_html_attachment)
     attachment = edition.attachments.first
     attachment.update!(locale: "fi")
+    Timecop.travel(11.years) { edition.update!(updated_at: Time.zone.now) }
```
ChrisBAshton, line 51: The `11.years` is an arbitrary choice, repeated throughout - could you extract it to a variable somewhere and then just refer to that?

```ruby
let(:to_the_future) { 11.years }

Timecop.travel(to_the_future) { ...
```

test/unit/html_attachment_test.rb
```diff
@@ -125,10 +132,12 @@ class HtmlAttachmentTest < ActiveSupport::TestCase
...
-    consultation = create(:consultation_with_public_feedback_html_attachment)
-    attachment = consultation.public_feedback.attachments.first
-    attachment.update!(locale: "fi")
-    assert_equal "/government/consultations/#{consultation.slug}/public-feedback/#{attachment.content_id}", attachment.url
+    Timecop.travel(11.years) do
+      consultation = create(:consultation_with_public_feedback_html_attachment)
+      attachment = consultation.public_feedback.attachments.first
+      attachment.update!(locale: "fi")
+      assert_equal "/government/consultations/#{consultation.slug}/public-feedback/#{attachment.content_id}", attachment.url
+    end
```
ChrisBAshton, line 140: Also worth adding a test that editions published before 29th November get the old behaviour.

app/models/html_attachment.rb
```diff
@@ -74,7 +74,11 @@ def url(options = {})
     # composed of document type, slug for the parent document, and identifier for the attachment;
     # for non-english attachments the identifier is the content_id, for english
     # attachments it is self i.e. the slug
-    identifier = sluggable_locale? ? self : content_id
+    identifier = if attachable.updated_at < Time.zone.local(2021, 11, 29)
```
ChrisBAshton: Worth adding a time in here, so we can time the deploy. E.g. set the new behaviour to apply from 7am, and arrange to do the deploy in the 30 minutes before that. We'll need to have a think about what time works best (I suspect we'll need to arrange an out-of-hours deployment to minimise risk here).

This code could also do with a big comment above it, explaining how this is a temporary measure while we patch up the old editions, and linking to that second Trello card.

config/routes.rb
```diff
@@ -106,7 +106,8 @@ def external_redirect(path_prefix, target)
 
     get "/publications(.:locale)", as: "publications", to: "publications#index", constraints: { locale: valid_locales_regex }
     get "/publications/:id(.:locale)", as: "publication", to: "_#_", constraints: { locale: valid_locales_regex }
-    get "/publications/:publication_id/:content_id" => "_#_", as: "publication_html_attachment"
+    get "/publications/:publication_id/:content_id" => "_#_", as: "publication_html_attachment_by_content_id"
+    get "/publications/:publication_id/:id" => "_#_", as: "publication_html_attachment"
```
ChrisBAshton, line 110: I'd drop the first commit and just do all the route editing in this final commit.

As above, I'd want to put a comment above each 'legacy' route, explaining when and where it will be removed. We should then add a 'remove old routes' acceptance criteria to the 2nd card.

app/models/html_attachment.rb
```diff
@@ -68,9 +68,19 @@ def url(options = {})
...
+    # for non-english attachments the identifier is the content_id, for english
+    # attachments it is self i.e. the slug
+    if sluggable_locale? || attachable.updated_at < Time.zone.local(2021, 11, 29)
+      identifier = self
+      path_helper = "#{type}_html_attachment_#{path_or_url}"
+    else
+      identifier = content_id
+      path_helper = "#{type}_html_attachment_by_content_id_#{path_or_url}"
+    end
+    Whitehall.url_maker.public_send(path_helper, attachable.slug, identifier, options)
```
ChrisBAshton, line 83: My final comment is a bit of a spanner in the works, but hopefully could lead to a much simpler solution. Do we need to make the distinction between the different path helpers? Could we not do something like this instead?

routes.rb:

```ruby
get "/consultations/:consultation_id/public-feedback/:edition_or_content_id" => "_#_", as: "consultation_public_feedback_html_attachment"
```

html_attachment.rb:

```ruby
identifier = sluggable_locale? ? self : content_id
Whitehall.url_maker.public_send("consultation_public_feedback_html_attachment", attachable.slug, identifier, options)
```

And then the missing piece of the puzzle in the controller would be something like:

```ruby
edition = Edition.find_by(id: edition_or_content_id) || Edition.find_by(content_id: edition_or_content_id) # and maybe even || Edition.find_by(slug: edition_or_content_id)
```

That way, we wouldn't need to have an out-of-hours deploy, a timing based cut-off point, or maintain two different route styles in routes.rb. We'd be encapsulating the different ID types in the controller and handling it there. Does that make sense?

ChrisBAshton (commented):
Thanks for working through this Ben - it's definitely more complicated a problem than I first imagined! I have a suggestion at the end which might help simplify things though.

test/unit/html_attachment_test.rb
benjamineskola: I probably could explicitly set this to be after 29 November 2021 to be honest — the 11 years was just based on the date at which the tests were already frozen at. But also, these tests would be expected to be removed once the migration goes ahead.

app/models/html_attachment.rb
benjamineskola: I wondered about doing the deploy in-hours on a day before the configured time, and then letting it take effect at midnight — but also there's nothing to lose by setting a time as well. (Though I assume it's unlikely that there'd be much publishing going on overnight anyway.)

config/routes.rb
benjamineskola: I _believe_ that two routes will continue to be necessary, since English attachments will continue to be identified by the `:id` not the `:content_id`. And if there's a way to make that work with only one route I think we can do so straight away, I don't think the time-limited effect (to be reverted) is relevant to the routes.

app/models/html_attachment.rb
benjamineskola: This certainly seems preferable yeah, I just had trouble making sense of the routes and how they bound to controller methods.

It didn't occur to me that this would allow us to avoid the timing-based cut-off, though: I thought that the point was that editions published before the cutover date should continue to have old-style URLs.

If we don't mind the new-style URLs appearing immediately, then I think we can just drop the commit adding that distinction. Having both the old-style route and the new-style route in place would still achieve that. (To be clear we should _also_ make this change to avoid having multiple routes — I just think the two things can be decoupled.)

ChrisBAshton: I think with the pseudocode snippets above, the old-style URLs should still work.

So for example, https://www.gov.uk/government/publications/universal-credit-and-rented-housing--2/5654420 would continue to resolve. But with this change, the same HTML attachment would also be accessible at https://www.gov.uk/government/publications/universal-credit-and-rented-housing--2/2c9002ee-19fd-4202-9720-b70aa8042126; the controller would just attempt to find by Edition ID first, fail to do that, and then try to find by Content ID.

**Subsequent publishes** of the parent edition would cause a switch to the new-style URL, because that is when the `consultation_public_feedback_html_attachment` helper would get used. But that's intentional behaviour (and something we'll be relying on in the 2nd card when we batch-republish all applicable documents). Does that make sense?

I should caveat this with "this code is untested" - I _think_ the approach would work, but we'd want to spike it first. If we took it in this direction, I'd suggest probably attempting it in a separate branch & PR.

benjamineskola: Presumably we could leave the double routes in place and then figure out if we can tidy that up as part of reverting the unnecessary parts after the migration?

On the other hand, if I can get it working now, then there'd be no need to revert it anyway. So I think I'll try to figure out if I can make sense of how the routes call the controllers (`_#_` is unhelpfully difficult to google for).

ChrisBAshton: From Andrew White, via Slack:

> If those route actually get hit they’ll look for a `Controller` class with an action `_` - that implies they’re being used to generate custom url helpers. There’s a relatively new method for defining things like that called `direct` - you can probably switch to that for these helpers https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/CustomUrls.html#method-i-direct

ChrisBAshton: Ah, I think what I'd somehow missed/forgotten is that these HTML attachments are rendered by government-frontend. 

So, given an [example non-English HTML attachment](https://www.gov.uk/government/publications/universal-credit-and-rented-housing--2/universal-credit-and-rented-housing-guide-for-landlords), if we look at its [content item](https://www.gov.uk/api/content/government/publications/universal-credit-and-rented-housing--2/universal-credit-and-rented-housing-guide-for-landlords), we can see `"rendering_app": "government-frontend"`.

As far as I can gather, government-frontend passes most requests to `ContentItemsController` with a [catch-all route](https://github.com/alphagov/government-frontend/blob/b986beb486d8de29703449d5b2f0d70f12c70020/config/routes.rb#L21-L24). This [fetches the content item via the request path](https://github.com/alphagov/government-frontend/blob/7e3397b98d091259c286cc9a52156e0338518f0e/app/controllers/content_items_controller.rb#L90-L101), which basically does a `Document.find_by(path: THE_PATH)` on Content Store. So we could publish _any_ route, so long as we tell Publishing API that the rendering app is Government Frontend.

I've taken your first commit, removed any routes.rb changes (as they're not needed), and raised #6377. I've tested this on Integration and can confirm that the Welsh HTML attachment's URL changes to content-ID based URL, a redirect is applied from the old attachment to the new, and English attachments are unaffected.

ChrisBAshton:
Superseded by #6377. See https://github.com/alphagov/whitehall/pull/6363#discussion_r761288330.

PR #6369, opened 2021-11-19
Add translations and sync the locales

1pretz1:
Imports translations from the supplier (ASP). As part of this work, the translations have also been audited and synced.

More info in the commits ->

Trello:
https://trello.com/c/Xz78vNR4/2757-import-newly-sourced-translations-into-locale-files-5


config/locales/fr.yml
```diff
@@ -247,7 +247,7 @@ fr:
       delivery_title:
         minister: Ministre
         speaker: Orateur
-      written_on: 'Rédiger le: '
+      written_on: 'Rédiger le :'
```
MuriloDalRi: This seems weird, is it a special case?

MuriloDalRi (commented):
This looks good, I'm just not sure why we need to remove unused keys, is that step necessary? Don't we want to keep those translations?

1pretz1:
Thanks for the review @MuriloDalRi! 

> I'm just not sure why we need to remove unused keys, is that step necessary? Don't we want to keep those translations?

It's not strictly necessary to remove unused keys as part of the import, but I thought it would be useful to remove them, rather than add keys that are not actually used in the application. If we were to add them then this would create more work for us when we come to validate the locales. It is sad that the locale files have changed since we sent the files off and quite a few of them are no longer needed.

config/locales/fr.yml
MuriloDalRi: Oh wow, I've learned something new today! :D

MuriloDalRi (approved):

config/locales/fr.yml
ChrisBAshton: +1!

PR #6372, opened 2021-11-25
Fix new corporate information pages not showing up on "about us"

barrucadu:
The "about us" corporate information page is special in that it
displays a list of an organisation's other corporate information pages
at the bottom.  This list comes from the links hash, which has the
content IDs of all the other pages to show.

If a new page is created, the about page needs republishing, so that
the content ID of the new page will be in its links hash.  Currently
this doesn't happen, so new corporate information pages don't show up
on the about page until it gets updated.

As a simple fix, just republish the about page when another page is
updated.  After deploying this, we should republish all the about
pages (there's a rake task for this), to ensure we don't have any
missing links.

---

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4784973)

### DIFF
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -3,6 +3,7 @@ class CorporateInformationPage < Edition
   include Searchable
 
   after_commit :republish_organisation_to_publishing_api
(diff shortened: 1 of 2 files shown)


hannako (approved):
LGTM

PR #6377, opened 2021-12-02
Use content-id in URLs for non-English HTML attachments

ChrisBAshton:
Every time you make an edit to a piece of guidance, it creates a
new edition of its associated HTML attachments. This works fine
in most cases.

However, for non-English HTML attachments, the edition ID is used
in the attachment URL. (English HTML attachments have a path
generated from its title, but fears around non-Latin characters
mean we can't apply the same technique to non-English attachments.
See #5963 for an investigation into this).

The result is that non-English HTML attachments get a new URL
every time the parent document is published. The previous URL is
set to redirect to the new URL. Over time, this causes a long
chain of redirects, which causes a `ERR_TOO_MANY_REDIRECTS` in
browsers once the chain reaches a certain length.

We've decided to switch from using edition IDs (which change
constantly) to content IDs (which are constant). This will solve
the browser chain redirect issue. English HTML attachments will
continue to have a path based on their title.

Supersedes #6363.

Trello: https://trello.com/c/rmphfmD6/2760-change-whitehall-html-attachments-to-use-content-ids-in-url-5

---

### DIFF
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -70,7 +70,12 @@ class HtmlAttachment < Attachment
     path_or_url = full_url ? "url" : "path"
     path_helper = "#{type}_html_attachment_#{path_or_url}"
 
-    Whitehall.url_maker.public_send(path_helper, attachable.slug, self, options)
+    # This depends on the rails url helpers to construct a url based on config/routes.rb:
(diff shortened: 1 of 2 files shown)


deborahchua (approved):
LGTM!

PR #6380, opened 2021-12-06
Fix failing Pact test by using correct branch

ollietreend:
This change configures the Pact tests to default to reading the `branch-master` version from the Pact Broker. Previously it was using a temporary branch which no longer exists, which was resulting in 404 Not Found responses from the broker.


kevindew (approved):
Nice one

PR #6383, opened 2021-12-08
Remove statistics_announcement_search_client 

brucebolt:
We have a client named `statistics_announcement_search_client` that is not being used in the application, so it can be removed.

Additionally, we had a class named `DevelopmentModeStubs::FakeRummagerApiForStatisticsAnnouncements` which is used to mock Search API (previously known as Rummager) in tests as part of this client.

However it appears this mock is not being used for anything in any of the tests (except the test for the mock itself), so it can be removed.  The mock also contains SQL queries that are not compatible with MySQL 8, so are blocking us from upgrading.

[Trello card](https://trello.com/c/TUzTsg9z/252-fix-whitehall-issues-for-db-upgrade)


beccapearce (approved):
Nice

PR #6391, opened 2021-12-14
Add a rake task to republish all documents with non-english attachments

benjamineskola:
https://trello.com/c/3BeDlugo/2761-republish-whitehall-documents-containing-non-english-html-attachments-5

### DIFF
--- a/lib/tasks/republish.rake
+++ b/lib/tasks/republish.rake
@@ -1,0 +1,9 @@
+desc "Republish all documents with non-english html attachments"
+task republish_non_english_html_attachments: :environment do
+  HtmlAttachment.where.not(deleted: true)
+  .where.not(locale: "en")
+  .where.not(locale: nil) # HtmlAttachment#sluggable_locale? treats nil locales the same as English
+  .each do |attachment|
+    PublishingApiDocumentRepublishingWorker.perform_async(attachment.attachable.document.id)
+  end
+end


lib/tasks/republish.rake
```diff
@@ -0,0 +1,8 @@
+desc "Republish all documents with non-english html attachments"
+task :republish_non_english_html_attachments do
+  Edition.where(state: %[published draft]).where.not(attachables: []).each do |edition|
```
ChrisBAshton, line 3: Trying this out in the app console, looks like you need a `%w` instead of a `%`.

```suggestion
  Edition.where(state: %w[published draft]).where.not(attachables: []).each do |edition|
```

Also, `Edition.where(attachables: [])` returns an empty array, and `Edition.where(attachables: nil)` gives `ActiveRecord::StatementInvalid (Mysql2::Error: Unknown column 'editions.attachables' in 'where clause')`. I suspect some sort of [joins](https://guides.rubyonrails.org/active_record_querying.html#joins) method will be needed before we can query this,  but because of the [weird way we save `attachments` in Whitehall](https://github.com/alphagov/whitehall/blob/7f0b913e2f3753864701d71f2c74c8f59bf225cb/app/models/attachable.rb#L18-L26), I'm not quite sure how to go about that.

lib/tasks/republish.rake
```diff
@@ -0,0 +1,8 @@
...
+    if edition.attachables.any? { |attachment| attachment.primary_local != "en"}
+        PublishingApiDocumentRepublishingWorker.perform_async(edition.id)
```
ChrisBAshton, line 5: This is indented too much.

ChrisBAshton (commented):
Hi Ben 👋 Quick drive-by review:

lib/tasks/republish.rake
benjamineskola: The other option would be to select everything and skip them in the ruby code — presumably a bit slower but I wonder if it would be significant, when it's a one-off task (and compared to the time needed for the republications themselves).

benjamineskola: I think I can do something like this:

```ruby
HtmlAttachment.where(attachable_type: 'Edition').where.not(deleted: true).joins("JOIN editions ON editions.id = attachments.attachable_id").where.not(editions: {primary_locale: 'en'})
```

But that returns only six results, which seems doubtful.

ChrisBAshton (commented):
The [example I keep coming back to](https://www.gov.uk/government/publications/universal-credit-and-rented-housing--2/5654420) has an ID of `5654420`. I tested that this is included by:

```
HtmlAttachment.where.not(deleted: true).where.not(locale: "en").where.not(locale: nil).pluck(:id).include?(5654420)
=> true
```

I also checked that [the equivalent English attachment](https://www.gov.uk/government/publications/universal-credit-and-rented-housing--2/universal-credit-and-rented-housing-guide-for-landlords) is not included by:

```
HtmlAttachment.where.not(deleted: true).where.not(locale: "en").where.not(locale: nil).pluck(:id).include?(1655467)
=> false
```

I'd like to spend a bit longer with a few more examples before approving, but looking good so far ✅

lib/tasks/republish.rake
```diff
@@ -0,0 +1,9 @@
+desc "Republish all documents with non-english html attachments"
+task republish_non_english_html_attachments: :environment do
+  HtmlAttachment.where.not(deleted: true)
+  .where.not(locale: "en")
+  .where.not(locale: nil) # HtmlAttachment#sluggable_locale? treats nil locales the same as English
+  .each do |attachment|
+    PublishingApiDocumentRepublishingWorker.perform_async(attachment.attachable.document.id)
```
ChrisBAshton, line 7: I've just tried this with that example:

```
PublishingApiDocumentRepublishingWorker.perform_async(HtmlAttachment.find(5654420).attachable.document.id)
=> "9c733a97d538128d594d1a2d"
```

...it worked! 🎉 

So now the last thing I want to try is testing what happens when there is a draft edition (i.e. we don't want to accidentally publish something that's in draft).

benjamineskola: Ah yeah, I think at some point I was filtering based on status but I think that got lost along the way.

(I think because I previously also wanted to filter out the `deleted` state but then found I could check the attachment's `deleted` property and in fact _had_ to check that property. Hopefully it won't require too much to restore the `state` check, though it might need the join to be reintroduced.)

ChrisBAshton: Just tried the command above on a HtmlAttachment that was part of a draft publication.

It didn't publish the publication or the attachment 👍

ChrisBAshton (approved):
This LGTM ✅  great job!

And good that it's running async, although there aren't _too_ many documents affected in the grand scheme of things:

```
HtmlAttachment.where.not(deleted: true).where.not(locale: "en").where.not(locale: nil).count
=> 10264
```

baisa (approved):

ChrisBAshton:
Gives rise to #6397.

PR #6398, opened 2021-12-21
Add rule for vip_editor permission

sihugh:
We want to limit who can edit certain people. This is a functional start that may be expanded to be more configurable through the UI.

### DIFF
--- a/lib/whitehall/authority/rules/person_rules.rb
+++ b/lib/whitehall/authority/rules/person_rules.rb
@@ -1,0 +1,11 @@
+module Whitehall::Authority::Rules
+  PersonRules = Struct.new(:actor, :subject) do
+    def can?(_action)
+      if subject.slug == "boris-johnson"
+        actor.vip_editor? || actor.gds_admin?
+      else
+        true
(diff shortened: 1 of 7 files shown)


lib/whitehall/authority/rules/person_rules.rb
```diff
@@ -0,0 +1,11 @@
+module Whitehall::Authority::Rules
+  PersonRules = Struct.new(:actor, :subject) do
+    def can?(_action)
+      if subject.slug == "boris-johnson"
+        actor.vip_editor?
```
brucebolt, line 5: Should we also add GDS Admins here, as GDS's Content Designers often change people/roles during reshuffles?

brucebolt (approved):
As discussed on Slack, a VIP status on the `Person` model would be good, but that can come in later work.

lib/whitehall/authority/rules/person_rules.rb
sihugh: Yep, sounds good 👍

PR #6402, opened 2021-12-29
Fix return in email signup controller

brucebolt:
Where a user attempts to signup to an invalid World Location (e.g. because they have manually changed the URL), we return a not found response.
    
However because the `return` was missing from this line, we then also attempted to redirect them. This redirect called a method that then attempted to retrieve a content ID from a non-existent World Location, resulting in a `NoMethodError` being reported to Sentry.
    
Adding the `return` here solves that problem.  A test has been added for this scenario, which failed before the fix was implemented.


brucebolt:
Sentry issue: [APP-WHITEHALL-8NC](https://sentry.io/organizations/govuk/issues/2874429616/?referrer=github_plugin)

beccapearce (approved):
Nice

davidgisbey (approved):
lgtm!

PR #6410, opened 2022-01-12
Updating Dockerfile with multistage build and streamlining.

nsabri1:
This commit streamlines the Whitehall image by installing dependencies etc in a new intermediate builder stage, before copying back the build outputs to the output image.
Please see previous commit for more details: https://github.com/alphagov/publisher/pull/1519

A few  values have been removed such as, Healthcheck, PORT and DATABASE_URL - These are not needed as part of the Build stage.

Adding Puma as part of the commit, given that we want to run Puma in production, we are instead configuring it via govuk_app_config

NodeJS is needed on runtime thus be included.

Enable JSON-style logging has been removed from production.rb, this seems to be old code and unused. Looking in Integration Whitehall EC2 this log file is empty.

https://trello.com/c/ZCL2pDyO/739-package-whitehall

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,36 @@
-FROM ruby:2.6.6
-RUN apt-get update -qq && apt-get upgrade -y
-RUN apt-get install -y build-essential nodejs && apt-get clean
-RUN gem install foreman
+# TODO: make this default to govuk-ruby once it's being pushed somewhere public
+# (unless we decide to use Bitnami instead)
+ARG base_image=ruby:2.6.6
 
+FROM $base_image AS builder
 # This image is only intended to be able to run this app in a production RAILS_ENV
-ENV RAILS_ENV production
-
-ENV GOVUK_APP_NAME whitehall
-ENV GOVUK_CONTENT_SCHEMAS_PATH /govuk-content-schemas
-ENV PORT 3020
-ENV DATABASE_URL mysql2://root:root@mysql/whitehall_development
-
-ENV APP_HOME /app
-RUN mkdir $APP_HOME
-
-WORKDIR $APP_HOME
-ADD Gemfile* $APP_HOME/
-ADD .ruby-version $APP_HOME/
-RUN bundle config set deployment 'true'
-RUN bundle config set without 'development test'
-RUN bundle install --jobs 4
-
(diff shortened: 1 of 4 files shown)


sengi (approved):
LGTM. I'd bung a link in the description to alphagov/publisher#1519 which is essentially the template for the Dockerfile changes.

Thanks for fixing the logging issue 👍

Dockerfile
```diff
@@ -1,34 +1,36 @@
...
-HEALTHCHECK CMD curl --silent --fail localhost:$PORT/healthcheck/ready || exit 1
-
-CMD foreman run web
+FROM $base_image
+ENV RAILS_ENV=production GOVUK_APP_NAME=whitehall
+# TODO: include nodejs in the base image (govuk-ruby).
+# TODO: apt-get upgrade in the base image
+RUN apt-get update -qy && \
+    apt-get upgrade -y && \
+    apt-get install -y nodejs
```
kevindew, line 32: Do you know if we definitely need nodejs at runtime? I anticipated it might only be needed to generate the assets.

config/environments/production.rb
```diff
@@ -119,11 +119,6 @@
     config.logger = ActiveSupport::TaggedLogging.new(logger)
   end
 
-  # Enable JSON-style logging
-  config.logstasher.enabled = true
-  config.logstasher.logger = Logger.new(Rails.root.join("log/#{Rails.env}.json.log"))
-  config.logstasher.suppress_app_log = false
-
```
kevindew, line 126: I've given this a try on integration and all seems to be working fine 👍

kevindew (commented):
Looks good to me, just a question about NodeJS

Dockerfile
nsabri1: This is a good question. The short answer to that is i dont know for sure. There is yarn.lock and it was included in the original Dockerfile - which from what i can see as you mention is used to generate the assets. If i can be sure somehow that its not needed happy to omit it from this stage of the build.

config/environments/production.rb
nsabri1: Thank you. :)

kevindew (approved):

PR #6413, opened 2022-01-17
DRY-up the database configuration file

ollietreend:
This change moves common config keys into a 'default' block which is reused across the development, test and production database configurations.

This makes it easier to spot configuration that remains the same between different environments, and avoids having to maintain multiple copies of it.

### Rationale for change

I'm soon going to need to apply a new variable to the database connection settings to [fix a bug with MySQL 8.0.27](https://trello.com/c/LNKQ5rmj/). This should only be a temporary workaround which can be removed once the MySQL bug is fixed (scheduled in the next patch release of MySQL).

Ideally I'd be able to configure it in one place and have it apply to all Rails environments.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,12 +1,15 @@
-development:
+default: &default
   encoding: utf8
   adapter: mysql2
+  variables:
+    sql_mode: TRADITIONAL
+
+development:
+  <<: *default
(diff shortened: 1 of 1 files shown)


davidgisbey (approved):
lgtm

PR #6415, opened 2022-01-17
Validate locale files as part of a test run

benjamineskola:
Adds a test that calls the RTM `LocaleChecker`, and aligns the locale files with the pluralisation rules in order to make this test past.

Where the pluralisation rule says only plural should exist, and the    singular and plural are identical, the singular has been removed; if    this is incorrect for a given language it would require a translator to    fix anyway: `id`, `ja`, `ko`, `ms`, `th`, `zh`.

However where the singular and plural forms are different, to err on the    side of caution the pluralisation rules have been updated rather than    deleting translations in order to make the tests pass: `az`, `fa`, `ka`, `tr`.

For Vietnamese, though most of the singulars are identical to the    plurals, a small number are different; so again, I've chosen to err on    the side of caution and update the pluralisation rules.

For some blank translations, it's also been necessary to add some extra blank entries (for languages that have more than just `:one` and    `:other` forms): `ar`, `be`, `cs`, `cy`, `gd`, `hr`, `lt`, `mt`, `pl`, `ro`, `ru`, `sk`, `sl`, `sr`, `uk`.

A small number of singulars have been deleted that appeared to be either    errors, or differing only by capitalisation from the plural.

https://trello.com/c/pPw5e9me/2828-apply-locale-file-validation-to-whitehall

### DIFF
--- a/config/locales/plurals.rb
+++ b/config/locales/plurals.rb
@@ -1,1 +1,9 @@
-GovukI18n.plurals
+GovukI18n.plurals.merge(
+  {
+    az: { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+    fa: { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+    ka: { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+    tr: { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+    vi: { i18n: { plural: { keys: %i[one other], rule: ->(n) { n == 1 ? :one : :other } } } },
+  },
+)
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -390,8 +390,12 @@ ar:
         two:
         zero:
       modern_slavery_statement:
+        few:
+        many:
         one:
         other:
+        two:
+        zero:
       national_statistics:
         few:
         many:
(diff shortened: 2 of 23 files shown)


config/locales/plurals.rb
```diff
@@ -1 +1,9 @@
-GovukI18n.plurals
+GovukI18n.plurals.merge(
```
MuriloDalRi, line 1: Would it be worth adding these to https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/govuk_i18n.rb ?

MuriloDalRi (approved):

config/locales/plurals.rb
brucebolt: We've previously been using [this unicode document](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html) as a definitive definition of the pluralisation rules.

I'd suggest we should update `govuk_app_config` only (so we don't lose the benefits of previous work that moved everything there, rather than having different fragmented rules across various apps) and if we think the translations in other apps are wrong (based on the unicode definitions), then we should go back to the translations firm to query why they didn't include these plurals.

benjamineskola: That makes sense — starting work on government-frontend I've noticed a similar set of issues, so it looks like it's not limited to Whitehall and I'm going to be extracting this change.

However I think what's happened in most cases is actually that we have _extra_ translations (singular and plural for a language where the rule says only plural), and that sometimes these are different in ways that I'm not entirely comfortable deleting (e.g., more than just capitalisation).

benjamineskola: Based on that document, all of the languages I've added here ought to be changed except for Vietnamese — which as noted above is mostly just duplicated except for a few (about 4) singulars which differ from the plural.

PR #6427, opened 2022-01-28
Double the sort buffer size (again)

ollietreend:
There's a bug with MySQL 8.0.27 which has reared its head in Sentry issue [APP-WHITEHALL-8S8](https://sentry.io/organizations/govuk/issues/2970128422/?project=202259).

We previously addressed this issue in PR #6414, but have since found another record which requires the memory to be increased again.

Until MySQL 8.0.28 is available in AWS RDS, we need to make sure this value is sufficiently large enough to avoid query errors that result in `500 Internal Server Error` messages to users.

We've doubled the memory limit again, so it's now 4x the default – sitting at ~1MB rather than the default ~256KB.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -308,7 +308,7 @@ GEM
     httparty (0.18.1)
       mime-types (~> 3.0)
       multi_xml (>= 0.5.2)
-    i18n (1.9.0)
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Super, thanks for this 👍

PR #6432, opened 2022-02-01
Upgrade Ruby to 2.6.9

ChrisBAshton:
Upgrades Ruby to 2.6.9. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-# TODO: make this default to govuk-ruby once it's being pushed somewhere public
+FROM ruby:2.6.9
```
ChrisBAshton, line 1: Buggy output from [upgrade-ruby-version](https://github.com/alphagov/upgrade-ruby-version). Works for most apps, but Whitehall is a special case!

benjamineskola (approved):

PR #6437, opened 2022-02-08
Escape special characters in filename when previewing CSVs

ollietreend:
This fixes a bug where CSVs couldn't be previewed if their filenames included non-ASCII/special characters. This bug was observed in production as Sentry issue [APP-WHITEHALL-8SQ][sentry].

Special characters in file names were causing the following error:
```
URI::InvalidURIError: URI must be ascii only "/government/uploads/system/uploads/attachment_data/file/729119/page_60_details_of_closed_cases_over_u\u0301300_000.csv"
```

This was because the app was attempting to download the raw CSV file over HTTP, but without escaping the CSV filename first to make it URL safe.

[sentry]: https://sentry.io/organizations/govuk/issues/2986708524/events/a5b647ed215e44ae9d4807f6debe47bb/?project=202259

## Demonstration of the problem

The publication [Cabinet Office data sets: annual report and accounts 2017 to 2018](https://www.gov.uk/government/publications/cabinet-office-data-sets-annual-report-and-accounts-2017-to-2018) includes a CSV attachment called 'Page 60 Details of closed cases over £300,000'.

The CSV filename includes the special character `ú`:

```
page_60_details_of_closed_cases_over_ú300_000.csv
```

Whilst it's possible to download the raw CSV file, it's not possible to preview it.

The preview URL:
https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/729119/page_60_details_of_closed_cases_over_u%25CC%2581300_000.csv/preview

Triggers an error in Sentry, and returns an error response to the end user.

This PR should fix the issue and allow the CSV to be previewed at the above URL.

---

Trello: https://trello.com/c/rhR7l3dj/458-whitehall-cant-preview-csv-files-with-special-characters-in-their-filename

### DIFF
--- a/lib/csv_file_from_public_host.rb
+++ b/lib/csv_file_from_public_host.rb
@@ -16,7 +16,7 @@ class CsvFileFromPublicHost
       connection.basic_auth(basic_auth_user, basic_auth_password)
     end
 
-    connection.get(path) do |req|
+    connection.get(Addressable::URI.escape(path)) do |req|
       req.headers["Range"] = "bytes=0-#{MAXIMUM_RANGE_BYTES}"
     end
   end
--- a/test/unit/csv_file_from_public_host_test.rb
+++ b/test/unit/csv_file_from_public_host_test.rb
@@ -114,11 +114,21 @@ class CsvFileFromPublicHostTest < ActiveSupport::TestCase
     assert_nil CsvFileFromPublicHost.csv_preview_from(response)
   end
 
-  test "#csv_response uses basic authentication if set in the environment" do
+  test ".csv_response uses basic authentication if set in the environment" do
     stub_csv_request.with(basic_auth: %w[user password])
     env = { "BASIC_AUTH_CREDENTIALS" => "user:password" }
 
     response = CsvFileFromPublicHost.csv_response("some-path", env: env)
+    assert_equal 206, response.status
+  end
+
+  test ".csv_response escapes non-ASCII characters in the path" do
+    raw_path = "uploads/path/with-špëçîål-characters.csv"
+    url_encoded_path = "uploads/path/with-%C5%A1p%C3%AB%C3%A7%C3%AE%C3%A5l-characters.csv"
+
+    stub_csv_request(path: url_encoded_path)
+
+    response = CsvFileFromPublicHost.csv_response(raw_path)
     assert_equal 206, response.status
   end
 end


test/unit/csv_file_from_public_host_test.rb
```diff
@@ -114,11 +114,21 @@ def stub_csv_request(status: 206, body: "", path: "some-path")
     assert_nil CsvFileFromPublicHost.csv_preview_from(response)
   end
 
-  test "#csv_response uses basic authentication if set in the environment" do
+  test ".csv_response uses basic authentication if set in the environment" do
```
kevindew, line 117: Nice catch on the `#` / `.` mistake

lib/csv_file_from_public_host.rb
```diff
@@ -16,7 +16,7 @@ def self.csv_response(path, env: ENV)
       connection.basic_auth(basic_auth_user, basic_auth_password)
     end
 
-    connection.get(path) do |req|
+    connection.get(Addressable::URI.escape(path)) do |req|
```
kevindew, line 19: Are we using Addressable here for a somewhat magic URI escape rather than say CGI.escape ? I'm guessing we're wanting backslashes to not be separated and not expecting other characters?

Just wondered, if my assumption is correct, if a) we need to document that and b) if we're risking any surprises - e.g. potential for weird user input to do anything fishy. 

I also wondered if already have the information escaped at all on the way in from Rails Request, but can't remember how much unescaping goes on - I assume a bit.

kevindew (commented):
This looks good - thanks for digging into it 🎉 

Just as an FYI on this area, there is an RFC on moving this CSV functionality out of Whitehall but the work on it was paused when the Publishing team was paused for Covid - we may want to resurrect it at some point: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-122-csv-preview.md

lib/csv_file_from_public_host.rb
ollietreend: @kevindew Good point. A more realistic test example would be a path containing slashes – because that's what happens in the real world.

I've used Addressable here because there's already a precedent for it being used to escape URLs:
https://github.com/alphagov/whitehall/blob/7a94656ccaaf030d7f2db590a77abda627f712f3/lib/whitehall/asset_manager_storage.rb#L36-L38

I did initially try with `URI.escape` but got warnings that it's an obsolete method.

Do you think I should be using `CGI.escape` instead?

ollietreend: In fact, I've just found some prior art for this which I'll reference in the commit message for posterity.

kevindew: Yup good shout on not using `URI.escape`. 

`CGI.escape` wouldn't be a drop replacement as it will escape slashes. E.g: 

```
[1] pry(main)> CGI.escape("test/item-with-special-chars-%://.pdf")
=> "test%2Fitem-with-special-chars-%25%3A%2F%2F.pdf"
```

I always find myself a bit unsure if having Addressable do it smart introduces a fresh risk as it, presumably, has to guess a little as to what a special character means - this is me being a bit Captain Edgecase - so wondered if it was absolutely necessary. 

For what it's worth, even if I'm right and there's a potential edge case it'd probably not surface in actual use case.

ollietreend: Oh yes, definitely. I've not introduced Addressable to the codebase – it looks like it's [already in use in Whitehall][whitehall], as well as [Asset Manager] and [GDS API adapters]. So I guess if there's an edge case, at least everything will fail consistently! 😬 

[whitehall]: https://github.com/alphagov/whitehall/blob/7a94656ccaaf030d7f2db590a77abda627f712f3/lib/whitehall/asset_manager_storage.rb#L36-L38
[Asset Manager]: https://github.com/alphagov/asset-manager/pull/396
[GDS API adapters]: https://github.com/alphagov/gds-api-adapters/pull/794/

ollietreend: @kevindew would you mind taking a look again now, given this conversation and the extra detail I've added to the commit message? 🙏🏻

kevindew (approved):
Great stuff - have you tried it out on integration?

ollietreend:
@kevindew I've just tested this in integration and it seems to solve the problem.

You should now be able to preview this 'problem file':
https://assets.integration.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/729119/page_60_details_of_closed_cases_over_ú300_000.csv/preview

Whereas it returns an error in other environments.

PR #6445, opened 2022-02-18
Add link to accessible format request form for pilot organisations

edwardkerry:
A pilot scheme for a limited number of organisations is soon to run,
allowing users to submit accessible format requests via a form in
the Feedback app rather than sending an email.

This commit adds a link to attachments owned by organisations in the
pilot, replacing the current toggled "send an email" text.

As not all document types making use of the attachments partial have
an owning, or alternative_format_provider, organisation we have added
the method to PolicyGroups (also known as WorkingGroups) and for Responses
delegated this method to their associated Consultation.

[Trello](https://trello.com/c/EDLOlTnt/892-create-an-allowed-list-of-departments-for-the-pilot)

Before
Screenshot 2022-02-21 at 10 11 47

After
Screenshot 2022-02-21 at 10 11 56

### DIFF
--- a/app/helpers/attachments_helper.rb
+++ b/app/helpers/attachments_helper.rb
@@ -13,6 +13,13 @@ module AttachmentsHelper
     csv_preview_path(id: attachment.attachment_data.id, file: attachment.filename_without_extension, extension: attachment.file_extension)
   end
 
+  def participating_in_accessible_format_request_pilot?(contact_email)
+    # A small number of organisations are taking part in a pilot scheme for accessible
+    # format requests to be submitted by a form rather than a direct email
+    pilot_addresses = GovukPublishingComponents::Presenters::Attachment::EMAILS_IN_ACCESSIBLE_FORMAT_REQUEST_PILOT
+    pilot_addresses.include?(contact_email)
+  end
+
   def block_attachments(attachments = [],
                         alternative_format_contact_email = nil,
                         published_on = nil)
--- a/app/views/documents/_attachment.html.erb
+++ b/app/views/documents/_attachment.html.erb
@@ -69,23 +69,27 @@
     <% end %>
 
     <% unless attachment.accessible? %>
-      <div data-module="toggle" class="accessibility-warning" id="<%= help_block_id %>">
-        <p class="govuk-body"><%= t('attachment.accessibility.intro') %></p>
-        <%= render "govuk_publishing_components/components/details", {
-          title: t('attachment.accessibility.request_a_different_format'),
-          data_attributes: {
-            track_category: "Accessible Format Request",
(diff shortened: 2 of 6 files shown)


app/helpers/attachments_helper.rb
```diff
@@ -1,4 +1,8 @@
 module AttachmentsHelper
+  # The following departments are taking place in a pilot scheme to use the accessible format
+  # request form. DFE(6), DWP(10), DHSC(12), HMRC(25), DVSA(570), UKHSA(1328)
+  ACCESSIBLE_FORMAT_REQUEST_PILOT_ORGANISATION_IDS = [6, 10, 12, 25, 570, 1328].freeze
```
1pretz1, line 4: We could fetch the email addresses from GOVUK Publishing Components and then use them to query the organisation, which would reduce the need to maintain two lists. We'd just need to pass the email instead of org_id from the view:

```
address = attachment.attachable.alternative_format_provider.alternative_format_contact_email

addresses = GovukPublishingComponents::Presenters::Attachment::EMAILS_IN_ACCESSIBLE_FORMAT_REQUEST_PILOT

addresses.include?(address)
```

1pretz1 (commented):
Looks good Ed! Spiked into whether we could remove this legacy way of displaying the accessible format section, but it's probably out of scope of this work. Although, it would be nice to not have two different implementations.

app/views/documents/_attachment.html.erb
```diff
@@ -69,23 +69,27 @@
...
-          }
-        } do %>
-          <%= t('attachment.accessibility.full_help_html',
-          email: alternative_format_order_link(attachment, alternative_format_contact_email),
-          title: attachment.title,
-          references: attachment_references(attachment)) %>
-        <% end %>
-      </div>
+      <% if attachment.attachable.alternative_format_provider.present? && organisation_in_accessible_format_request_pilot?(attachment.attachable.alternative_format_provider.id) %>
+        <%= link_to "Request an accessible format of this document", "/contact/govuk/request-accessible-format?content_id=#{attachment.attachable.content_id}&attachment_id=#{attachment.id}" %>
```
1pretz1, line 73: From https://github.com/alphagov/govuk_publishing_components/pull/2636/files#r813954690: 
```suggestion
        <%= link_to "Request an accessible format of this document", "/contact/govuk/request-accessible-format?content_id=#{attachment.attachable.content_id}&attachment_id=#{attachment.id}", class: "govuk-link"  %>
```

app/helpers/attachments_helper.rb
edwardkerry: good idea, implemented this now the publishing component have been updated

PR #6447, opened 2022-02-23
Fix login after editing an orphaned Corporate Information Page

brucebolt:
There may be cases when a `CorporateInformationPage` has no organisations (i.e. it is orphaned).  At the moment, users who have a draft CorporateInformationPage with no organisation(s) assigned cannot access the Whitehall homepage as the 'My Documents' section returns a 500 error.

The error is caused by the helper that forms the URL for the edit page, which requires the organisation's slug.  No with assigned organisation, the URL cannot be formed.

This removes the orphaned corporate information pages from appearing on the user's dashboard.


KludgeKML (approved):
Makes sense.

PR #6449, opened 2022-02-24
Update to Rails 7.0.x

benjamineskola:
Updating to Rails 7: https://trello.com/c/n6D8yVSh/253-upgrade-whitehall-from-rails-61x-to-70x


Gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rails", "6.1.5"
+gem "rails", "~> 7.0.2"
```
brucebolt, line 3: We should fix this at a specific release, otherwise we run a risk of rails being downgraded if a dependency becomes incompatible.

brucebolt (commented):
Great work!

I've added one comment, but otherwise everything else looks good.

Gemfile
benjamineskola: Yep — had done this before I saw the discussion on Slack. Will update.

app/models/corporate_information_page.rb
```diff
@@ -134,7 +134,7 @@ def title_prefix_organisation_name
   end
 
   def title(_locale = :en)
-    corporate_information_page_type.title(owning_organisation)
+    corporate_information_page_type&.title(owning_organisation)
```
kevindew, line 137: It'd be good to explain why we've added this.

config/application.rb
```diff
@@ -23,7 +23,13 @@ class Application < Rails::Application
...
     require "whitehall"
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.1
+    config.load_defaults 7.0
+
+    # Disable redirect protection. This is turned on by default in Rails 7.0.
+    config.action_controller.raise_on_open_redirects = false
+
+    # Disable rails 7.0+ button_to behaviour
+    config.action_view.button_to_generates_button_tag = false
```
kevindew, line 32: It'd be good to explain why these are switched off? Do we plan to have them switched off forever or is this just temporary?

config/application.rb
```diff
@@ -23,7 +23,13 @@ class Application < Rails::Application
 
     require "whitehall"
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.1
+    config.load_defaults 7.0
```
kevindew, line 26: I think to switch to this default we're going to need some transitory code for cookie rotation.

As per: https://github.com/alphagov/content-tagger/commit/a54df45b37efee1f070855927952465e041d7cd6

test/unit/role_test.rb
```diff
@@ -126,7 +126,7 @@ class RoleTest < ActiveSupport::TestCase
   end
 
   test "should not be destroyable when it has appointments" do
-    role = create(:role, role_appointments: [create(:role_appointment)])
+    role = create(:role, :occupied)
```
kevindew, line 129: Is this functionally the same? It's a little unclear to me whether this fixes an error or just something where we thought an alternative syntax would be neater?

app/views/admin/worldwide_offices/_worldwide_office.html.erb
```diff
@@ -46,9 +46,9 @@
   <% unless worldwide_organisation.is_main_office?(worldwide_office) %>
     <%= button_to 'Set as main office', set_main_office_admin_worldwide_organisation_path(worldwide_organisation, worldwide_organisation: {main_office_id: worldwide_office.id}), method: :put, class: 'btn btn-info' %>
     <% if worldwide_organisation.office_shown_on_home_page?(worldwide_office) %>
-      <%= button_to "Remove from home page", [:remove_from_home_page, :admin, worldwide_organisation, worldwide_office], class: 'btn btn-default' %>
+      <%= button_to "Remove from home page", [:remove_from_home_page, :admin, worldwide_organisation, worldwide_office], class: 'btn btn-default', method: :post %>
     <% else %>
-      <%= button_to "Add to home page", [:add_to_home_page, :admin, worldwide_organisation, worldwide_office], class: 'btn btn-default' %>
+      <%= button_to "Add to home page", [:add_to_home_page, :admin, worldwide_organisation, worldwide_office], class: 'btn btn-default', method: :post %>
```
kevindew, line 51: I'm a bit confused by the reasoning behind this change. How does this end up as a PATCH form? 

Looking at the source code it looks like it should only be get or post https://github.com/rails/rails/blame/0169d15bc7ec4557971d6ac6120e48b2cac9c407/actionview/lib/action_view/helpers/url_helper.rb#L342 unless I'm missing something 🤔

features/support/document_helper.rb
```diff
@@ -185,6 +185,31 @@ def preview_document_path(edition, options = {})
...
+
+    find(:xpath, ".//select[@id='#{base_dom_id}_4i']").select(time.hour.to_s.rjust(2, "0"))
+    find(:xpath, ".//select[@id='#{base_dom_id}_5i']").select(time.min.to_s.rjust(2,  "0"))
+  end
+
+  def select_datetime(datetime, options)
+    # this overrides the capybara-provided select_datetime which wrongly assumes we're using html5 input type="datetime".
+    select_date(datetime, options)
+    select_time(datetime, options)
+  end
```
kevindew, line 212: Nice work sussing this out. It looks like there is now a fix for this merged into cucumber-rails but not yet released: https://github.com/cucumber/cucumber-rails/blob/main/CHANGELOG.md

If it's not released by Monday we may perhaps want to pin the cucumber-rails dependency to the git hash of the fix until a new release is made.

kevindew (commented):
Great job working through all of this, it hasn't looked an easy application to update.

I've flagged a few bits where I'm a bit confused as to why you've done something. It'd be great if you could add more info into the commits on things like this where you explain what was breaking and why your code change fixed it.

test/unit/role_test.rb
benjamineskola: I think this relates to the issue you spotted a few days ago: `:occupied` specifically adds a `role_appointment` in a hook (and sets the dates appropriately), so at the very least it's a neater way of writing the same thing; but it was also leading to a validation error that looked the same as elsewhere and needed adjusting the hooks, so I think it's just incorrect in tests to create a role in the original way.

config/application.rb
benjamineskola: To be honest the latter would be simple enough to turn on, it would just mean updating tests to match the new generated output.

I'm not sure I understand enough about the former to know why it was necessary but it seems like it would be worth digging into, either as part of this card or a follow-up one.

benjamineskola: `raise_on_open_redirects` turns out to be a false positive. We're redirecting to an absolute URL generated with `Plek.new.website_root` which Rails detects as a third-party redirect (because it's absolute).

kevindew: Does we use those redirects everywhere or just in a few places? The preferential option would be to use the `allow_other_host` argument on the redirects where we expect this, but not practical if it's all of them.

benjamineskola: Yep, I've just pushed changes to address that and the other overridden default.

benjamineskola: OK, `button_to_generates_button_tag` seems to be more complex than I thought; it seems like there's somewhere in the code still generating `<input>` rather than `<button>`, and initially I'd thought maybe some could be left as-is but now I'm at a point where changing the test helpers breaks some tests, but leaving them breaks others. It may be possible to unpick it further though if I can figure out where the old-style buttons are coming from.

app/views/admin/worldwide_offices/_worldwide_office.html.erb
benjamineskola: Weirdly I can no longer reproduce this, so I suppose I'll just remove this change. 🤔

Gemfile
```diff
@@ -100,7 +100,7 @@ end
 
 group :cucumber, :test do
   gem "cucumber"
-  gem "cucumber-rails", require: false
+  gem "cucumber-rails", require: false, git: "https://github.com/cucumber/cucumber-rails.git", ref: "5bfb4219abc50a5d1c8b2fa0a04c441bbc6e5288"
```
kevindew, line 103: Can we add a comment to explain this (as per: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#gemfile-organisation)

config/initializers/patches.rb
```diff
@@ -1,3 +1,5 @@
 Dir[Rails.root.join("lib/patches/*.rb")].sort.each do |patch|
   require patch
 end
+
+Rails.autoloaders.main.ignore("lib/patches")
```
kevindew, line 5: It'd be good to add either a comment above this or information to the commit message to just state that we don't files in lib/patches to follow standard Zeitwerk conventions and they should all be required manually.

app/controllers/announcements_controller.rb
```diff
@@ -53,7 +53,7 @@ def scheduled_announcements
 
   def redirect_to_news_and_communications(format = "")
     base_path = "#{Plek.new.website_root}/search/news-and-communications#{format}"
-    redirect_to("#{base_path}?#{news_and_communications_query_string}")
+    redirect_to("#{base_path}?#{news_and_communications_query_string}", allow_other_host: true)
```
kevindew, line 56: Not something to fix now, but I wonder why these URLs have a host name in them when they could just be a redirect to a path. It's somewhat confusing how base_path is a full URL rather than a path 🤔

kevindew (approved):
Nice job. I've just a couple of suggestions for a little more info but I don't think you'll need another review.

app/controllers/announcements_controller.rb
benjamineskola: It feels like maybe `website_root` ought to return `/` when it's being asked for the same instance rather than a different environment. For that matter I wonder if we even need to invoke Plek at all for this, unless there are cases I'm overlooking.

kevindew: Yeah I think we just don't need to involve Plek here. I can see it's done a few times though so I'm wondering if there was a reason why someone did it - commit message doesn't seem to explain why :-/

benjamineskola: I can only think that if `GOVUK_WEBSITE_ROOT` was ever anything but a hostname the distinction would matter (i.e., `http://www.dev.gov.uk/some-prefix`. But that doesn't seem to be the case anywhere I can find.

PR #6454, opened 2022-03-04
Add temporary Rake task to republish all editions with attachments fo…

edwardkerry:
This PR adds a temporary Rake task to republish all live editions with attachments that are part of a pilot scheme to use a new accessible format request form. And editions with attachments that have an alternative format contact email will be republished to the publishing-api in order to present a link to the new form rather than a mailto email link.

This is dependent on https://github.com/alphagov/whitehall/pull/6445 first being deployed in order to have the desired effect.

Once this has been run, this task can be deleted.

[trello](https://trello.com/c/RZcLEIGR/1107-accessible-format-request-form-add-rake-task-to-whitehall-to-republish-all-pilot-documents-post-launch)


lib/tasks/republish_accessible_format_request_pilot_documents.rake
```diff
@@ -0,0 +1,24 @@
...
+  organisations = Organisation.where(alternative_format_contact_email: [pilot_emails])
+
+  organisations.each do |org|
+    all_editions_for_org = Edition.latest_edition.in_organisation(org)
+    puts "Total editions for #{org.slug}: #{all_editions_for_org.count}"
+    editions_with_attachments = all_editions_for_org.publicly_visible.where(
+      id: Attachment.where(accessible: false, attachable_type: "Edition").select("attachable_id"),
+    )
+    puts "Enqueueing #{editions_with_attachments.count} editions with attachments for #{org.slug}"
+    editions_with_attachments.joins(:document).distinct.pluck("documents.id").each do |document_id|
```
1pretz1, line 19: Potential edge case: if a draft is the latest edition which has removed the attachment that's present on the published edition, will the published edition get skipped from being republished? We may have to query for all draft / published editions which have an attachment instead of grabbing the latest edition

1pretz1: Awesome, thanks Ed!

1pretz1 (approved):
LGTM! 🍏

PR #6463, opened 2022-03-10
Allow CorporateInformationPages to respond to alternative_format_provider methods

edwardkerry:
This will allow CorporateInformationPages to use links to the accessible format request form introduced in
https://github.com/alphagov/whitehall/pull/6445 for attachments that are inaccessible.


KludgeKML (approved):
Looks good to me.

PR #6466, opened 2022-03-11
Add jQuery to whitehall frontend

andysellick:

## What / why
Give whitehall frontend its own copy of jQuery, instead of getting it from the components gem via static.

The reason for this is that we're removing jQuery from GOV.UK and there's far too much of it in whitehall to convert it to vanilla JS right now. Instead, the approach we're taking is to give some applications their own copy of jQuery, allowing us to remove it from the rest of GOV.UK and deal with those problematic applications later.

This PR doesn't change whitehall admin, but that seems to already have its own jQuery somehow (I don't really understand how, but it does). Have tested this against a branch of the components gem without jQuery, and it appears to function normally.

I've already checked what happens if you include two copies of jQuery inside an application and it seems to be fine, but I don't think this should be merged until we're ready to also remove jQuery from the components gem, as it will increase the asset size unnecessarily. Better to have only a short time when whitehall has two jQuerys.

## Visual changes
None.

Trello card: https://trello.com/c/DUaKxWvl/101-remove-jquery-from-govuk-javascript

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -8,6 +8,7 @@
 //= require govuk_publishing_components/components/button
 //= require govuk_publishing_components/components/govspeak
 //
+//= require jquery/dist/jquery
 //= require jquery_ujs
 //= require jquery.ui.all
 //= require vendor/jquery/jquery.player.min.js
--- a/package.json
+++ b/package.json
@@ -30,6 +30,9 @@
       "no-var": 0
     }
   },
+  "dependencies": {
+    "jquery": "1.12.4"
+  },
   "devDependencies": {
     "standardx": "^7.0.0",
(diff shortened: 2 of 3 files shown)


ollietreend:
@andysellick

> This PR doesn't change whitehall admin, but that seems to already have its own jQuery somehow (I don't really understand how, but it does).

I think jQuery in the admin interface will be coming from the [govuk_admin_template](https://github.com/alphagov/govuk_admin_template) gem, which pulls in its own copy of jQuery via a gem dependency:

https://github.com/alphagov/govuk_admin_template/blob/f00266ff97ae882831dc27c9e8a8b6b202b65767/govuk_admin_template.gemspec#L20

kevindew (approved):
This looks good to me 👍

PR #6478, opened 2022-03-23
Remove emails from Rake task for departments not in pilot

edwardkerry:
Currently only HMRC are taking part in the accessible request format pilot scheme. This change ensures only their documents will be republished to display the link to the form when the associated change in govuk_publishing_components is released.

https://github.com/alphagov/govuk_publishing_components/pull/2695

[trello](https://trello.com/c/7vyvANlE/1172-accessible-format-request-add-links-to-form-for-hmrc-documents-ready-for-user-testing)


KludgeKML (approved):
Looks good to me.

PR #6479, opened 2022-03-23
Revert "Update to Rails 7.0.x"

benjamineskola:
Reverting while investigating production issue 😢 

https://trello.com/c/n6D8yVSh/253-upgrade-whitehall-from-rails-61x-to-70x


kevindew (approved):
Thanks, I do agree with the 😢 

It does seem a bit of a pain to work out what is up with it.

I wonder if we can push in some of the changes from this PR as small separate PRs so the upgrade doesn't have to change so much at once.

PR #6483, opened 2022-03-24
Update to Rails 7.0.x

benjamineskola:
Upgrading to Rails 7

(Expecting this PR to shrink as it's split into smaller pieces)

https://trello.com/c/n6D8yVSh/253-upgrade-whitehall-from-rails-61x-to-70x


app/models/edition.rb
```diff
@@ -699,6 +699,10 @@ def has_legacy_tags?
 
   delegate :locked?, to: :document
 
+  def attribute_names
+    @attributes.keys
+  end
```
kevindew, line 704: Can we test if we can remove https://github.com/alphagov/whitehall/pull/6483/commits/c20279853ddef4ff69581b27880cd1febef2360a thanks to this?

benjamineskola: Ah, yes we can. I'll test 49b2aa3 and ad1c9bf too.

benjamineskola: Seems like issue with `Role`/`RoleAppointment` is unrelated, still occurs even with this change in place (they don't inherit from `Edition`, but moving the definition to `ApplicationRecord` or adding it to `Role` etc doesn't help either).

app/assets/javascripts/application.js
```diff
@@ -8,7 +8,7 @@
 //= require govuk_publishing_components/components/button
 //= require govuk_publishing_components/components/govspeak
 //
-//= require jquery/dist/jquery
+//= require jquery
```
kevindew, line 11: Hmm I think this is caused by Yarn not running. It looks like Rails 7 may have removed the coupling of assets:precompile and yarn install (https://github.com/rails/rails/pull/43641) but haven't seem to have documented in the update guide. I'll see if I can push something up quickly that sorts it.

kevindew (commented):
Nice work persisting with this 👍 

I think I've found a few places that could do valid external redirects. These are:

https://github.com/alphagov/whitehall/blob/17fc8fccab7eaada0e251a54b73e467371bd0047/app/controllers/documents_controller.rb#L37
https://github.com/alphagov/whitehall/blob/17fc8fccab7eaada0e251a54b73e467371bd0047/app/controllers/email_signups_controller.rb#L5-L6
https://github.com/alphagov/whitehall/blob/17fc8fccab7eaada0e251a54b73e467371bd0047/app/controllers/csv_preview_controller.rb#L29-L30

kevindew (approved):
Super 🤞

PR #6490, opened 2022-03-25
Avoid redirecting to absolute URLS

benjamineskola:
In all of these cases we're redirecting to the same domain, so prefixing with the fully-qualified domain gets us nothing. In Rails 7 this will be an error and needs to be specifically permitted; better to rewrite them instead.

(These are all the cases where the existing tests raise exceptions when run under Rails 7, e.g. in #6449. There may be others that are untested.)

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -52,7 +52,7 @@ private
   end
 
   def redirect_to_news_and_communications(format = "")
-    base_path = "#{Plek.new.website_root}/search/news-and-communications#{format}"
+    base_path = "/search/news-and-communications#{format}"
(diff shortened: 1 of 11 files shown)


kevindew (commented):
Nice one, this looks great.

Have you given it a spin on integration? It'd be good to test it there just incase we learn that this was set-up to deal with any weird reverse proxy misconfiguration - but hopefully not.

benjamineskola:
I've tested a few and they seem to work fine, yeah — I can't actually figure out which route triggers the redirect in `EmailSignupsController` but I don't think there's any reason to expect it would be any different.

kevindew (approved):
Awesome, this will make the Rails 7 upgrade much less painful

PR #6491, opened 2022-03-25
Update page title to better describe the page

MartinJJones:
## What

Update the title of the "Help and services around the world" page to better describe the content - [https://www.gov.uk/world](https://www.gov.uk/world
)

## Why

The current page title of "UK and the world" is not descriptive enough to help users identify their current location without requiring users to read or interpret page content.

Further information - [WCAG - Understanding Success Criterion 2.4.2: Page Titled](https://www.w3.org/WAI/WCAG21/Understanding/page-titled.html)

### DIFF
--- a/app/views/world_locations/index.html.erb
+++ b/app/views/world_locations/index.html.erb
@@ -1,4 +1,4 @@
-<% page_title "UK and the world" %>
+<% page_title "Help and services around the world" %>
 <% page_class "govuk-width-container world-locations-index index-list-page" %>
 
 <div class="govuk-grid-row">
--- a/app/views/world_locations/index.html.erb
+++ b/app/views/world_locations/index.html.erb
@@ -1,11 +1,11 @@
-<% page_title "Help and services around the world" %>
+<% page_title sanitize(t("world_location.content.heading")) %>
 <% page_class "govuk-width-container world-locations-index index-list-page" %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <header class="js-filter-list">
       <%= render "govuk_publishing_components/components/title", {
-        title: sanitize("Help and services around the&nbsp;world")
+        title: sanitize(t("world_location.content.heading"))
       } %>
 
       <p class="govuk-body"><%= t('world_location.content.find_out') %></p>
--- a/app/views/world_locations/index.html.erb
+++ b/app/views/world_locations/index.html.erb
@@ -1,4 +1,4 @@
-<% page_title sanitize(t("world_location.content.heading")) %>
+<% page_title sanitize(t("world_location.content.heading").gsub!(/&nbsp;/, " ")) %>
 <% page_class "govuk-width-container world-locations-index index-list-page" %>
 
 <div class="govuk-grid-row">
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -868,6 +868,7 @@ ar:
     content:
       complete_list:
       find_out:
+      heading:
       list_link:
     headings:
       announcements: إعلاناتنا
--- a/config/locales/az.yml
+++ b/config/locales/az.yml
@@ -567,6 +567,7 @@ az:
     content:
       complete_list:
       find_out:
+      heading:
       list_link:
     headings:
       announcements: Elanlarımız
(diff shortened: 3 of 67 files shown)


app/views/world_locations/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<% page_title "UK and the world" %>
+<% page_title "Help and services around the world" %>
```
danacotoran, line 1: Good detective work figuring out where this update needs to take place 🕵️‍♂️ 

My one tiny suggestion: we can see that "Help and services around the world" is repeated now on this line and on L8. So I wonder if the string `Help and services around the&nbsp;world` could be moved to the [locale file](https://github.com/alphagov/whitehall/blob/4ecfa1b4f3400a6de3cdece2148d3a5dd5a8e340/config/locales/en.yml#L577) under `world_location.content` like the rest of the content in this partial?

It could then be referenced in this partial as something like:
```suggestion
<% page_title sanitize(t("world_location.content.heading")) %>
```


For reference: `sanitize` is a method which sanitises html input in rails. I think it's being used on L8 because of the `&nbsp;` entity (they probably used `&nbsp;` here to prevent the word "world" being alone on a new line on certain breakpoints)

The `t()` bit is a translation function – it's basically a feature of the I18n API in Rails that you can use to look up translations. So let's say there if was a Welsh translation for "Help and services around the world" in the [Welsh locale file](https://github.com/alphagov/whitehall/blob/4ecfa1b4f3400a6de3cdece2148d3a5dd5a8e340/config/locales/cy.yml), then `t()` would return that translated string when the locale of the page is Welsh.

MartinJJones: Thanks Dana,

The HTML `&nbsp` entity is still included in the Title tag even after using the `sanitize` method.

I looked into this further and it looks like the `sanitize` method is only applied to HTML elements and attributes.

The HTML still renders the white space correctly when used in the title, I cannot see an issue with keeping the `&nbsp` entity in the title, from either an Accessibility or SEO perspective.

I tested using VoiceOver and the title was read out correctly, I also checked the guidelines with regards to whitespace and it would only fail if the whitespace was used in a word.

https://www.w3.org/TR/WCAG20-TECHS/F32.html

MartinJJones:
Translations log updated to include the new key:

https://docs.google.com/spreadsheets/d/1Wips8w48mUoLE3UqY-TIeoHJcW9l371qUuRyCqbl5J0/edit?pli=1#gid=0

maxgds (changes requested):
I don't think we can put it out with the non breaking space feeding into the title. The title tag is displayed to users both in search results like google, and in the tab on their browser. While it's clearly there to add a little visual sugar I don't think it should go into the translation file that way. Check with design whether they are happy to live with the last word wrapping by itself. If not, pull the string from the translation in the template and manipulate it using ruby code to switch out the last whitespace in the string for a nonbreaking space. (Nothing is ever simple...)

MartinJJones:
Thanks @maxgds,

I checked with design and it was decided to keep the `&nbsp` in the translation file so the last word does not wrap in the heading.

I made the change suggested so that the `&nbsp;` entity is no longer include in the title, but remains in the h1.

** Title **

```html
<title>Help and services around the world - GOV.UK</title>
```

** Heading 1 **

```html
<h1 class="gem-c-title__text govuk-heading-xl">
    Help and services around the&nbsp;world
 </h1>
```

maxgds:
> I checked with design and it was decided to keep the `&nbsp` in the translation file so the last word does not wrap in the heading.

I'm not going to ask you to change this at this point but if this sort of thing comes up again we need to have a think about the best way to handle it, it's got a bit of code smell about it. When you look at GOV.UK as a whole, the majority of H1s will be dynamic - we won't know what publishers are putting on pages, so making a special exception for one hard coded instance is a bit pointless in the grand scheme of things, there are going to be lots of orphan single words on newlines., especially when you consider different screen sizes and zoom levels. 

Further, I'd rather have kept the translated text pure and applied the manipulation at the point where it was needed (ie inserting the space character in the template rather than removing it), rather than having the manipulation presented to everything and needing to be overridden. This is something that we could apply more widely, even to dynamic content, if we wanted to (I'm not convinced we do!). But I guess to an extent that's just personal preference, and what you've done works so it's cool.

PR #6492, opened 2022-03-28
Update Find in admin bookmarklet

AgaDufrat:
The Find in admin bookmark stopped working in Chrome: https://govuk.zendesk.com/agent/tickets/4923446

`document_page_id` was being returned as null and calling `length` on it was causing an error.

This also updates content_id_from_meta_tag() to fix `Uncaught TypeError: $(...).attr is not a function`

Tested in Chrome developer console in integration [on this page](https://www.integration.publishing.service.gov.uk/guidance/apply-for-an-advance-tariff-ruling). 
Line 64 replaced with:
 `admin_path = "https://whitehall-admin.integration.publishing.service.gov.uk/government/admin";`

---

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6497, opened 2022-03-29
Fix find in admin bookmarklet

andysellick:

## What/why

- in whitehall admin at '/government/admin/find-in-admin-bookmarklet/' there is a feature that allows users to drag a bookmarklet to their toolbar, that will then open a GOV.UK public page in whitehall publisher
- this broke recently because it relied upon jQuery and we removed jQuery from GOV.UK last week
- this fix removes the jQuery dependency from this script
- unfortunately users with the existing bookmarklet will need to delete and recreate it

## Visual changes
None.

### DIFF
--- a/app/views/admin/find_in_admin_bookmarklet/_bookmarklet.erb
+++ b/app/views/admin/find_in_admin_bookmarklet/_bookmarklet.erb
@@ -64,9 +64,9 @@ function nav_to_backend() {
   if (mapping){
     admin_path = "<%= Whitehall.admin_root + Whitehall.router_prefix %>/admin";
     if(mapping.id_finder()){
-      window.location = admin_path + "/" + mapping.path_builder(mapping.id_finder());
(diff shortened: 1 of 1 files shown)


kevindew:
Thanks for putting this together Andy. 

As per: https://github.com/alphagov/whitehall/pull/6492#issuecomment-1080593563 it'd be good if we could get some tracking on this, especially as people will need to recreate it, so we can find out about usage (as this really feels like I feature we should sunset)

Do you want me to add a commit for this?

Or if you're able to do it yourself it'd be good to have a query string of utm_content=find-in-admin-bookmarklet

andysellick:
@kevindew yes, will definitely add some tracking. Just about to test that this fix actually works, and then I'll look into that, thanks.

kevindew:
@andysellick that's great thank you

AgaDufrat (approved):
Tested in integration with different types of documents and seems to work ok.

PR #6508, opened 2022-04-08
Add filtering for browse page derived topics

Tetrino:
While migrating Mainstream Browse pages to topics we need to prevent
the migrated clones from being visible to the user. Here we're adding
filtering to prevent items being show if the relevant field doesn't
exist or is true.

https://trello.com/c/m13ffUVc/292-update-publishing-apps-that-know-of-specialist-topics-to-hide-new-synced-ones


kevindew:
I'm assuming this isn't needed anymore. Can always be reopened if so.

PR #6514, opened 2022-04-14
Auth bypass ids for editions

baisa:
Ensure all newly created Whitehall editions are stored with auth bypass ids.

[Trello card](https://trello.com/c/MBi8s6Lh/374-ensure-all-newly-created-whitehall-editions-are-stored-with-auth-bypass-ids)

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance]
(https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️


db/schema.rb
```diff
@@ -10,50 +10,49 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[6.1].define(version: 2021_10_19_095600) do
```
brucebolt, line 13: There seem to be quite a few unrelated changes to the `schema.rb` file that appear related to the Rails 7 upgrade.  Maybe we missed a step during the upgrades?  In any case, this probably ought to be in another commit (or ideally PR) to keep those changes separate.

app/presenters/publishing_api/base_item_presenter.rb
```diff
@@ -18,6 +18,7 @@ def base_attributes
         publishing_app: "whitehall",
         redirects: [],
         update_type: update_type,
+        auth_bypass_ids: [item.auth_bypass_id]
```
brucebolt, line 21:
```suggestion
        auth_bypass_ids: [item.auth_bypass_id],
```
You'll need a comma here to appease rubocop.

brucebolt (commented):
Changes to add the `auth_bypass_id` look good to me 👍 , although there's some Rails 7 stuff that ideally should be another PR or commit.

test/factories/editions.rb
```diff
@@ -8,6 +8,7 @@
     change_note { "change-note" }
     summary { "edition-summary" }
     previously_published { false }
+    auth_bypass_id { "52db85fc-0f30-42a6-afdd-c2b31ecc6a67" }
```
brucebolt, line 11: Should this also be a random value?

db/schema.rb
baisa: Good point, thank you @brucebolt

app/models/edition.rb
```diff
@@ -643,6 +644,10 @@ def set_public_timestamp
                             end
   end
 
+  def set_auth_bypass_id
+    self.auth_bypass_id = SecureRandom.uuid
```
kevindew, line 648: Just a quick comment as I know you're still working on this, we shouldn't have this set on before_save as it'll mean every edit to the document changes the auth_bypass_id and thus invalidates the token.

kevindew, line 648: Just a quick comment as I know you're still working on this, we shouldn't have this set on before_save as it'll mean every edit to the document changes the auth_bypass_id and thus invalidates the token.

app/presenters/publishing_api/case_study_presenter.rb
```diff
@@ -24,6 +24,7 @@ def content
           public_updated_at: item.public_timestamp || item.updated_at,
           rendering_app: Whitehall::RenderingApp::GOVERNMENT_FRONTEND,
           schema_name: "case_study",
+          auth_bypass_ids: [item.auth_bypass_id],
```
kevindew, line 27: @baisa, just a thought based off our chat last week. Code changes like this will cause problems until we've backfilled all the past editions with auth_bypass_ids. 

You might need to break this up into a few PRs:

1) create auth_bypass_ids on new editions
2) backfill old editions with auth_bypass_ids
3) set auth_bypass_ids to non null
4) send them to Publishing API

BeckaL: Hey Kevin, I can't see anywhere in publishing api that suggests it doesn't accept arrays with nil, and e.g. this [test data](https://github.com/alphagov/publishing-api/blob/dd8dd9232d3cbf33b8945fdd898ebe80d7dcfcf6/spec/support/request_helpers/mocks.rb#L53) suggests that it can. (Also just wrote a Command test to see, but don't know what it's mocking out).

Am I missing something here? (Imagine this is v probable :-) ) Original thinking was that it'd be nice to be able to deploy a fully end to end thing, but more than happy to split it out (and obviously need to do it if it's not safe).

PR #6526, opened 2022-04-25
Add auth bypass ids field to editions

baisa:
Ensure all newly created Whitehall editions are stored with auth bypass ids.

[Trello card](https://trello.com/c/MBi8s6Lh/374-ensure-all-newly-created-whitehall-editions-are-stored-with-auth-bypass-ids)

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -96,6 +96,7 @@ class Edition < ApplicationRecord
   scope :future_scheduled_editions,     -> { scheduled.where(Edition.arel_table[:scheduled_publication].gteq(Time.zone.now)) }
 
   # @!group Callbacks
+  before_create :set_auth_bypass_id
   before_save :set_public_timestamp
   before_save { check_if_locked_document(edition: self) }
   # @!endgroup
@@ -641,6 +642,10 @@ EXISTS (
                             else
                               major_change_published_at
                             end
+  end
+
+  def set_auth_bypass_id
+    self.auth_bypass_id = SecureRandom.uuid
   end
 
   def title_required?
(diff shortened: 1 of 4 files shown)


app/models/edition.rb
```diff
@@ -96,6 +96,7 @@ class Edition < ApplicationRecord
   scope :future_scheduled_editions,     -> { scheduled.where(Edition.arel_table[:scheduled_publication].gteq(Time.zone.now)) }
 
   # @!group Callbacks
+  before_create :set_auth_bypass_id
```
brucebolt, line 99: I wonder if it would be possible to add a test for this?  Maybe in `EditionTest` you could create a new edition and check the `auth_bypass_id` field is not nil.

baisa: Thnak you @brucebolt We are going to add not NULL restraint as a separate PR (see conclusions here https://github.com/alphagov/whitehall/pull/6514/files#r857406916), and I was thinking to add a test then?

brucebolt: I guess you could add a test now to ensure that newly created editions have a `auth_bypass_id`.  You could then extend this to check for the NULL constraint in the next PR?

test/unit/edition_test.rb
```diff
@@ -21,6 +21,11 @@ class EditionTest < ActiveSupport::TestCase
     assert_kind_of Document, edition.document
   end
 
+  test "adds auth bypass id to a newly created edition" do
+    edition = build(:edition)
+    assert_not_nil edition.set_auth_bypass_id
```
brucebolt, line 26:
```suggestion
    assert_not_nil set_auth_bypass_id
```
It looks like this is currently testing the method only, rather than ensuring the `before_create` is triggered correctly and the value saved into the database.

brucebolt (approved):
Thanks for adding the test.

All looks good to me 👍 .

PR #6531, opened 2022-04-26
Add auth_bypass_id to the presenter for each edition type

baisa:
Auth_bypass_id is added to presenters for each edition type to send through to the Publishing Api.

[Trello card](https://trello.com/c/MBi8s6Lh/374-ensure-all-newly-created-whitehall-editions-are-stored-with-auth-bypass-ids)

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6536, opened 2022-04-27
Add a rake command to republish all drafts

ollietreend:
This extends the existing collection of Publishing API rake tasks to include one which will bulk republish all documents which have an active draft edition.

This will be needed once the auth_bypass_id fields have been backfilled for existing Edition records. The changes need to be sent to the Publishing API so that shared preview links can be supported for exising drafts.

Trello: https://trello.com/c/PAZNzo4z/382-ensure-existing-whitehall-draft-editions-are-stored-with-authbypassids

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -234,6 +234,18 @@ namespace :publishing_api do
       puts "Finished queuing items for Publishing API"
     end
 
+    desc "Republish all documents with draft editions"
+    task all_drafts: :environment do
+      editions = Edition.in_pre_publication_state.includes(:document)
+
+      puts "Enqueueing #{editions.count} documents"
+      editions.find_each do |edition|
+        document_id = edition.document.id
+        PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
+      end
+      puts "Finished enqueueing items for Publishing API"
+    end
+
     desc "Republish all editions which have attachments to the Publishing API"
     task editions_with_attachments: :environment do
       editions = Edition.publicly_visible.where(
--- a/test/unit/tasks/publishing_api_test.rb
+++ b/test/unit/tasks/publishing_api_test.rb
@@ -230,6 +230,46 @@ class PublishingApiRake < ActiveSupport::TestCase
       end
     end
 
+    describe "#all_drafts" do
+      let(:task) { Rake::Task["publishing_api:bulk_republish:all_drafts"] }
+
+      test "republishes all draft editions" do
+        # A draft document which hasn't been published yet
+        draft_only = create(:draft_detailed_guide).document
+
+        # A published document with no draft edition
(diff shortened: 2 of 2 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -234,6 +234,18 @@ namespace :publishing_api do
       puts "Finished queuing items for Publishing API"
     end
 
+    desc "Republish all documents with draft editions"
+    task all_drafts: :environment do
+      editions = Edition.draft.includes(:document)
```
brucebolt, line 239:
```suggestion
      editions = Edition.in_pre_publication_state.includes(:document)
```
There are a few document states (e.g. `submitted`) which are not draft but where a `auth_bypass_id` will be present.  The [`in_pre_publication_state` scope](https://github.com/alphagov/whitehall/blob/9eb0010f6472346b2a99bcddb70c63f874ed981d/app/models/edition.rb#L83) should cover those.

lib/tasks/publishing_api.rake
```diff
@@ -234,6 +234,18 @@ namespace :publishing_api do
...
+
+      puts "Enqueueing #{editions.count} documents"
+      editions.find_each do |edition|
+        document_id = edition.document.id
+        PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
```
brucebolt, line 244: I wonder if we actually need to queue this in Whitehall?  Publishing API has its own queues, including a low priority queue for bulk publishing.

If we don't need to queue here, then something like `Whitehall::PublishingApi.save_draft(edition, bulk_publishing: true)` will republish only the draft edition of the document and we won't need the `.includes(:document)` on line 239 either.

brucebolt:
It's also worth noting that this rake task will be dependent on https://github.com/alphagov/whitehall/pull/6531 being merged.

BeckaL:
As per @brucebolt's comment above, it might also be worth waiting for the assets PR to be merged, otherwise we would have to run it again to ensure all the assets have auth bypass ids too.

(Note that we'll still need to run it again once the html attachments changes are made, I think I'm OK with this as it'll affect a subset of documents)

lib/tasks/publishing_api.rake
ollietreend: My concern with directly calling the Publishing API (i.e. `Whitehall::PublishingApi.save_draft`) is that we'll be processing tens of thousands of draft editions. If the rake tasks fails or times out mid-way though execution (e.g. if Publishing API throws an exception), we'll need to start re-publishing again from the beginning.

Processing them async with Sidekiq allows us to protect ourselves from that possibility as each edition can be retried independently.

ollietreend: We've talked it through and agreed to keep the async worker approach, given the number of documents involved.

Marking this thread as resolved.

baisa (approved):

PR #6537, opened 2022-04-27
Correct backfill migration to migrate only for pre-publication states

BeckaL:

The migration as previously written was failing because it did not allow modifications to editions not in pre-publication states (e.g. already published). This modifies the migration to only migrate content in pre-publication state.

### DIFF
--- a/db/migrate/20220426132125_backfill_auth_bypass_id.rb
+++ b/db/migrate/20220426132125_backfill_auth_bypass_id.rb
@@ -4,7 +4,7 @@ class BackfillAuthBypassId < ActiveRecord::Migration[7.0]
   def up
     to_update = Edition.where(auth_bypass_id: nil)
     to_update.find_each do |edition|
-      edition.update!(auth_bypass_id: SecureRandom.uuid)
+      edition.update_column(:auth_bypass_id, SecureRandom.uuid)
     end
   end
 end


db/migrate/20220426132125_backfill_auth_bypass_id.rb
```diff
@@ -2,7 +2,7 @@ class BackfillAuthBypassId < ActiveRecord::Migration[7.0]
   disable_ddl_transaction!
 
   def up
-    to_update = Edition.where(auth_bypass_id: nil)
+    to_update = Edition.in_pre_publication_state.where(auth_bypass_id: nil)
```
kevindew, line 5: You could skip validation instead with something like:

```
edition.auth_bypass_id = SecureRandom.uuid
edition.save!(validate: false)
```

kevindew: Just thinking that you'll lose the ability to set the column to not-null with this approach.

BeckaL: Ah OK. Do you have a view on what's better? It would be nice to have non-null columns, but also auth_bypass_id doesn't really make sense for non-draft content.

BeckaL: So maybe it does make sense to e.g. in presenters only send it when it exists

kevindew: Yup that makes sense. Are we at risk of sending it when in a post draft state?

I sort of thought it just needed to be something set at the point of creation and then didn't matter if it was still around. However if it's beneficial to not have it on places where it doesn't make sense then I think we should set the column to null at the point of publishing. 

Happy with you to make the call here, only situation I think we should avoid is it living forever on an Edition yet some old ones not being backfilled.

kevindew (approved):
Looks good to me 👍

brucebolt (approved):

PR #6538, opened 2022-04-28
Make auth bypass ids non null

BeckaL:

Similar to [this migration](https://github.com/alphagov/content-publisher/blob/main/db/migrate/20200701131534_auth_bypass_id_not_null.rb) in content publisher


beccapearce (commented):
Should the db/schema have `null: false` added to the `auth_bypass_id` field?
if you run the migrations does it add it in?

ollietreend (approved):

PR #6539, opened 2022-04-29
Auth bypass ids for assets

BeckaL:
Trello: https://trello.com/c/udy9k9H4/364-ensure-that-assets-in-whitehall-for-draft-editions-are-stored-with-authbypassids

This PR adds functionality to send auth bypass ids to asset manager from their respective editions. This is so that an edition's assets can be viewed as part of the shareable preview.

### DIFF
--- a/app/models/image_data.rb
+++ b/app/models/image_data.rb
@@ -7,4 +7,12 @@ class ImageData < ApplicationRecord
 
   validates :file, presence: true
   validates_with ImageValidator, size: [960, 640]
+
+  def auth_bypass_ids
+    images
+      .joins(:edition)
+      .where("editions.state in (?)", Edition::PRE_PUBLICATION_STATES)
+      .map { |e| e.edition.auth_bypass_id }
+      .uniq
+  end
 end
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -3,29 +3,29 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
 
   sidekiq_options queue: "asset_manager"
 
-  def perform(file_path, legacy_url_path, draft = false, model_class = nil, model_id = nil)
+  def perform(file_path, legacy_url_path, draft = false, attachable_model_class = nil, attachable_model_id = nil, auth_bypass_ids = [])
     return unless File.exist?(file_path)
 
     file = File.open(file_path)
-    asset_options = { file: file, legacy_url_path: legacy_url_path }
+    asset_options = { file: file, legacy_url_path: legacy_url_path, auth_bypass_ids: auth_bypass_ids }
     asset_options[:draft] = true if draft
 
-    if model_class && model_id
-      model = model_class.constantize.find(model_id)
-      if model.respond_to?(:access_limited?) && model.access_limited?
-        authorised_user_uids = AssetManagerAccessLimitation.for(model)
+    if attachable_model_class && attachable_model_id
+      attachable_model = attachable_model_class.constantize.find(attachable_model_id)
+      if attachable_model.respond_to?(:access_limited?) && attachable_model.access_limited?
+        authorised_user_uids = AssetManagerAccessLimitation.for(attachable_model)
         asset_options[:access_limited] = authorised_user_uids
       end
     end
 
     asset_manager.create_whitehall_asset(asset_options)
 
-    if model
+    if attachable_model
       # The AttachmentData we want to set the timestamp on may not
       # exist yet, so create a worker to do it after a very short
       # delay.  The worker will retry if it still doesn't exist.
       AssetManagerAttachmentSetUploadedToWorker.perform_in(
-        0.5.seconds, model_class, model_id, legacy_url_path
+        0.5.seconds, attachable_model_class, attachable_model_id, legacy_url_path
       )
     end
 
--- a/lib/whitehall/asset_manager_storage.rb
+++ b/lib/whitehall/asset_manager_storage.rb
@@ -11,11 +11,15 @@ class Whitehall::AssetManagerStorage < CarrierWave::Storage::Abstract
     FileUtils.cp(original_file, temporary_location)
     legacy_url_path = ::File.join("/government/uploads", uploader.store_path)
     draft = uploader.assets_protected?
+
     if uploader.model.respond_to?(:attachable)
-      model_class = uploader.model.attachable && uploader.model.attachable.class.to_s
-      model_id = uploader.model.attachable && uploader.model.attachable.id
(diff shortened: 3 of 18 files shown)


app/assets/javascripts/templates.js
```diff
@@ -0,0 +1,4 @@
+window.templates = {};
```
ollietreend, line 1: We shouldn't commit this file. (It's weird)

lib/whitehall/asset_manager_storage.rb
```diff
@@ -11,12 +11,23 @@ def store!(file)
...
+  end
+
+  def get_model_class_and_model_id
     if uploader.model.respond_to?(:attachable)
       model_class = uploader.model.attachable && uploader.model.attachable.class.to_s
       model_id = uploader.model.attachable && uploader.model.attachable.id
+    elsif uploader.model.respond_to?(:images)
+      model_class = uploader.model.class.to_s
+      model_id = uploader.model.id
     end
```
ollietreend, line 29: I think we can tidy up this implementation.

Let's make the worker responsible for this bit instead. So this class doesn't behave differently depending on `:attachable` behaviour.

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -12,15 +12,12 @@ def perform(file_path, legacy_url_path, draft = false, model_class = nil, model_
...
-        authorised_user_uids = AssetManagerAccessLimitation.for(model)
-        asset_options[:access_limited] = authorised_user_uids
-      end
+      asset_options.merge!(access_options(model))
     end
 
     asset_manager.create_whitehall_asset(asset_options)
 
-    if model
+    if is_attachment
```
ollietreend, line 20: We can determine `is_attachment` within this worker, rather than relying on the storage class to work it out for us.

test/integration/attachment_access_limited_integration_test.rb
```diff
@@ -40,14 +40,67 @@ class AttachmentAccessLimitedIntegrationTest < ActionDispatch::IntegrationTest
...
+        stub_whitehall_asset("logo.png", id: "asset-id", draft: true)
+
+        visit admin_case_study_path(edition)
+        click_link "Edit draft"
+        choose "Use a custom image"
+        attach_file "File", path_to_attachment("minister-of-funk.960x640.jpg")
+        click_button "Save"
+      end
+
+      it "sends the attachment to asset manager with the case study's auth_bypass_id" do
```
ollietreend, line 96: Would be good to clarify that this is a test for an image rather than an attachment.

app/assets/javascripts/templates.js
BeckaL: Removed 👍 (well left it as an empty file as before)

lib/whitehall/asset_manager_storage.rb
BeckaL: Resolving because we've gone for a different approach now

app/workers/asset_manager_create_whitehall_asset_worker.rb
BeckaL: this now reverted back to the previous way of deriving this in the worker (albeit with some hopefully better naming) so resolving this comment

BeckaL:
A couple of points on this:
1) From working on this section of the codebase, we've seen that access limiting is currently only applied to assets uploaded as 'Attachables' and not e.g. via the ImageUploader or the ConsultationResponseFormUploader. I've considered this to be out of the scope of the current ticket, but we could change this at the same time.

2) As noted by @ollietreend, the way we derive auth_bypass_ids for images uploaded via the image uploader is inefficient (need to get all images associated with an ImageData and look up each of their editions). So simplifying this would be good.

3) We ended up putting a lot of the tests into integration tests, might be worth putting some of them in unit tests for asset_manager_storage

lib/whitehall/asset_manager_storage.rb
```diff
@@ -11,14 +11,31 @@ def store!(file)
...
+         .filter_map { |image| image.edition.auth_bypass_id if image.edition.pre_publication? }
+         .uniq
+    elsif uploader.model.respond_to?(:attachable) && uploader.model.attachable.respond_to?(:auth_bypass_id)
+      [uploader.model.attachable.auth_bypass_id]
+    elsif uploader.model.respond_to?(:consultation_response_form) && !uploader.model.consultation_response_form.nil?
+      [uploader.model.consultation_response_form.consultation_participation.consultation.auth_bypass_id]
+    else
+      []
+    end
+  end
```
ollietreend, line 37: Refactor:

- Create methods on each model – i.e. `ImageData#auth_bypass_ids`, `AttachmentData#auth_bypass_ids`, etc.
- Each model becomes responsible for knowing how to get bypass IDs given its own associations
- In this `AssetManagerStorage` class, we can now call `uploader.model.auth_bypass_ids` without having to know about how the model finds those bypass IDs. We can lose the `if ... else` statement, and probably even the entire `AssetManagerStorage#auth_bypass_ids` method.

ollietreend (commented):
@BeckaL as we discussed

lib/whitehall/asset_manager_storage.rb
BeckaL: 👍 this should now be addressed in the below commit

BeckaL:
> I wonder maybe we could squash some of the commits, especially those responsible for refactoring/cleaning up.

Yeah agreed @baisa my thoughts was just to squash it all into one at the end but was just leaving them up for review in case it made it easier for e.g. OIlie to follow where we've been discussing throughout

app/models/image_data.rb
```diff
@@ -7,4 +7,11 @@ class ImageData < ApplicationRecord
 
   validates :file, presence: true
   validates_with ImageValidator, size: [960, 640]
+
+  def auth_bypass_ids
+    images
+      .filter { |image| image.respond_to?(:edition) }
+      .filter_map { |image| image.edition.auth_bypass_id if image.edition.pre_publication? }
```
baisa, line 14: Do we need line 13 considering image belongs to edition?
Maybe worth testing something like `images.joins(:edition).where("edition.states in (?)", Edition::PRE_PUBLICATION_STATES).uniq`

app/models/image_data.rb
```diff
@@ -9,6 +9,10 @@ class ImageData < ApplicationRecord
   validates_with ImageValidator, size: [960, 640]
 
   def auth_bypass_ids
-    images.filter_map { |image| image.edition.auth_bypass_id if image.edition.pre_publication? }.uniq
+    images
+      .joins(:edition)
+      .where("editions.state in (?)", Edition::PRE_PUBLICATION_STATES)
+      .map { |e| e.edition.auth_bypass_id }
+      .uniq
```
baisa, line 16: Looks good!

baisa (approved):

PR #6550, opened 2022-05-09
Use SVG icons for attachment thumbnails instead of PNGs

benjamineskola:
For attachment types with static thumbnails (ODT, PNG, etc, not PDF) the thumbnail URL is generated based on the website root (i.e., `www.`); however, if the thumbnail is being displayed on a draft preview, this fails because the wrong host is used. This could be solved by correctly choosing the host, but a better solution would be to embed SVG icons instead, which will be sent with the page (and match what is used elsewhere).

Further work:
- use the `govuk_publishing_components` attachment component in its entirety (but I don't think this supports PDF preview thumbnails)
- use an SVG icon as the fallback for PDFs when thumbnail generation fails

https://trello.com/c/BZUpCC7s/398-fix-preview-issue-with-attached-assets


app/helpers/document_helper.rb
```diff
@@ -78,16 +78,18 @@ def attachment_reference(attachment)
   end
 
   def attachment_thumbnail(attachment)
+    image_host = attachment.attachable.draft? ? URI(Plek.new.external_url_for("draft-origin")).host : Whitehall.public_root
```
ollietreend, line 81: I think that the `attachment` variable here is an `AttachmentData` object.

In which case, it's worth noting that `attachment.attachable` is an `attr_accessor` which is `nil` unless explicitly set:

```ruby
AttachmentData.last.attachable
=> nil
```

Is that okay? I haven't looked too deeply into what sets that attribute or when, but I thought I'd point out that this change introduces the assumption that it's been set before `attachment_thumbnail` is called.

kevindew:
Thanks for looking into this. As I understand this is just an issue for people on integration and not on the VPN? 

I'd guess this is only a partial fix as this would mean anything that is in a live state and on the draft content store would have this problem still.

There is a better approach that could be taken, which is to switch over to use the component for these: https://components.publishing.service.gov.uk/component-guide/attachment - which uses SVGs rather than having to have files that are downloaded.

As an interim approach you could switch these file paths to be the same svgs that the component gem uses (as per: https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/attachment/_thumbnail_document.html.erb but may need a couple of modernisations) - which I imagine could be a nice upgrade. One of the things that's a massive smell about the current approach is it relies on Whitehall assets to exist which couples the results of `image_url` to the deployment of Whitehall, so if Whitehall deleted https://github.com/alphagov/whitehall/blob/main/app/assets/images/pub-cover-html.png this would break government-frontend rendering pages

app/helpers/document_helper.rb
benjamineskola: Just double-checked the relations here between Attachment, AttachmentData, and Attachable, and I think in principle `attachable` could be null but I'm not sure if that actually makes sense in practice. That would seem to imply an attachment that isn't actually attached to anything — but if it wasn't attached to anything, we couldn't be trying to render it.

On the other hand it can't hurt to be cautious and I think if I use a safe-navigation operator it'll end up defaulting to `Whitehall.public_root` anyway.

benjamineskola:
> Thanks for looking into this. As I understand this is just an issue for people on integration and not on the VPN?

I believe it could apply to any draft, including in production; but the further issue of prompting for basic auth for the assets would only apply in integration. Worst-case scenario in production is a missing-image placeholder.

>  anything that is in a live state and on the draft content store would have this problem still.

I think I've overlooked this scenario, yes; I'd conflated 'in draft state' and 'on draft content store'. I'm not sure if there'd be a better way to determine whether something's on the draft content store.

> …  switch over to use the component for these … [or]  switch these file paths to be the same svgs that the component gem uses …

I'll look into these options today.

app/helpers/document_helper.rb
```diff
@@ -78,18 +78,17 @@ def attachment_reference(attachment)
...
-                elsif %w[xls xlsx ods csv].include? attachment.file_extension.downcase
-                  image_url("pub-cover-spreadsheet.png", host: Whitehall.public_root)
-                else
-                  image_url("pub-cover.png", host: Whitehall.public_root)
-                end
-    image_tag(image_url, alt: "")
+    if attachment.pdf?
+      image_tag(attachment.file.thumbnail.url, alt: "")
+    elsif attachment.html?
+      render "govuk_publishing_components/components/attachment/thumbnail_document"
```
kevindew, line 84: Ooh do we need to create one for HTML attachment?

app/helpers/document_helper.rb
```diff
@@ -78,18 +78,17 @@ def attachment_reference(attachment)
...
+    elsif %w[doc docx odt].include? attachment.file_extension
+      render "govuk_publishing_components/components/attachment/thumbnail_document"
+    elsif %w[xls xlsx ods csv].include? attachment.file_extension.downcase
+      render "govuk_publishing_components/components/attachment/thumbnail_spreadsheet"
+    else
+      render "govuk_publishing_components/components/attachment/thumbnail_generic"
+    end
```
kevindew, line 91: I wonder if it's a bit risky to be using these pulled in from publishing_components - they feel a bit like a private API - and should move copies of these SVGs into Whitehall

kevindew (commented):
Nice job with this progress - this will be a nice improvement.

There's also a cool gem https://github.com/jamesmartin/inline_svg that we should probably use in embedding the SVGs but maybe worth talking about that when we're a bit further.

app/helpers/document_helper.rb
benjamineskola: I sort of wondered about this. Ideally I'd use the entire component, but that would mean going one level up and replacing a lot more code which I'm not entirely sure I'm comfortable with. (In particular I think it would mean losing PDF previews, which since they're currently working perfectly feels counterproductive.)

kevindew: Yup, switching to that component seems definitely out of scope. One day!

app/views/documents/attachment/_thumbnail_document.html.erb
```diff
@@ -0,0 +1,3 @@
+<svg class="gem-c-attachment__thumbnail-image" version="1.1" viewBox="0 0 99 140" width="99" height="140" aria-hidden="true">
```
ollietreend, line 1: It might be nice to add a `<%# comment %>` to the beginning of these thumbnail files (which should be possible since they're `.erb` rather than `.svg`) just to note where the SVGs were originally sourced from.

e.g. something like:
```suggestion
<%# Source: https://github.com/alphagov/govuk_publishing_components/blob/1facbe66c93fcfc15f77cbcde00d90b20874765c/app/views/govuk_publishing_components/components/attachment/_thumbnail_document.html.erb %>
<svg class="gem-c-attachment__thumbnail-image" version="1.1" viewBox="0 0 99 140" width="99" height="140" aria-hidden="true">
```

ollietreend (commented):
This looks really nice ✨ Just one comment from me.

benjamineskola:
For reference, if we prefer to use the lighter shade of grey matching the original PNGs rather than the darker that the publishing components use, that should be (approximately) `#dee0e2` according to the macOS colour meter.

davidgisbey (approved):
LGTM. I've manually tested that each of the attachment types renders as expected. Looks great!

PR #6562, opened 2022-05-16
Add UI for shareable preview

baisa:
This is to present the link for the draft document preview.


app/views/admin/editions/_edition_view_edit_buttons.html.erb
```diff
@@ -18,6 +18,35 @@
       <%= content_data_button(@edition) %>
   <% end %>
 </section>
+
+<% if @edition.state == "draft" %>
```
BeckaL, line 22: We need a couple more conditions here to display it:
* only in the case of the edition types that we're enabling it for (so not publications etc that have html attachments or links to other documents - details on the ticket)
* only when a user has the permission - Becca P and I looked into this today, so can talk you through it, shouldn't be too bad

Maybe best to put this logic in the controller under a single 'can view preview?' method or something?

config/application.rb
```diff
@@ -90,5 +90,9 @@ class Application < Rails::Application
     # allow overriding the asset host with an enironment variable, useful for
     # when router is proxying to this app but asset proxying isn't set up.
     config.asset_host = ENV["ASSET_HOST"]
+
+    # unless Rails.application.secrets.jwt_auth_secret
+    #   raise "JWT auth secret is not configured. See config/secrets.yml"
+    # end
```
BeckaL, line 96: we should uncomment this

app/helpers/public_document_routes_helper.rb
```diff
@@ -47,6 +47,16 @@ def preview_document_url(edition, options = {})
     document_url(edition, options)
   end
 
+  def preview_document_url_with_auth_bypass_token(edition, options = {})
```
BeckaL, line 50: Do we need the `options` param here? Are we ever going to use it?

app/views/admin/editions/_edition_view_edit_buttons.html.erb
```diff
@@ -18,6 +18,35 @@
...
+            Share document preview
+            </a>
+          </h4>
+        </div>
+        <div id="collapseOne" class="panel-collapse collapse" role="tabpanel" aria-labelledby="headingOne">
+          <div class="panel-body">
+            <p>Send the preview link to someone so they can see a preview of how the document will appear on GOV.UK. No password or GOV.UK Signon account is needed.</p>
+            <p><strong>Select all and copy the link</strong></p>
+            <textarea class="form-control" rows="3"><%= utm_uri %></textarea>
+          </div>
```
BeckaL, line 43: I think we should also add the text in here from [Nikin's design](https://www.figma.com/file/f8z3zGgXCEorFKHZVfP6gb/Sharable-previews?node-id=80%3A152) about when the link expires - so "The preview link will expire on [day month year] or when the document is published."

app/models/edition.rb
```diff
@@ -702,6 +702,25 @@ def has_legacy_tags?
     has_primary_sector? || has_secondary_sectors?
   end
 
+  # For more info, see https://docs.publishing.service.gov.uk/manual/content-preview.html#authentication
```
BeckaL, line 705: We should add tests for both the methods added in here to the edition spec. For the auth_bypass_token, you might be able to take some inspiration from [these tests](https://github.com/alphagov/content-publisher/blob/8df63907dab486a54894105b829ab9ea02936b67/spec/models/edition_spec.rb#L203) in content publisher

app/models/user.rb
```diff
@@ -26,6 +26,7 @@ module Permissions
     FORCE_PUBLISH_ANYTHING = "Force publish anything".freeze
     GDS_ADMIN = "GDS Admin".freeze
     EXPORT_DATA = "Export data".freeze
+    CAN_SHARE_PREVIEWS = "can share previews".freeze
```
BeckaL, line 29: Let's ask someone with access to add this to production. Maybe we should also capitalise it for consistency? (We can remove the non capitalised one from integration and add it again.

app/helpers/public_document_routes_helper.rb
BeckaL: Also, worth adding a test for this?

app/views/admin/editions/_edition_view_edit_buttons.html.erb
baisa: We have agreed with Nikin not to include it at this time.

test/unit/edition_test.rb
```diff
@@ -26,6 +26,65 @@ class EditionTest < ActiveSupport::TestCase
...
+  # test below will be removed after enabling shareable preview for this doccument type
+  test "edition has shareable preview disabled if it has document collection type" do
+    edition = create(:draft_document_collection)
+    user = create(:gds_editor)
+    user.permissions << "can share previews"
+    assert_equal edition.has_enabled_shareable_preview?(user), false
+  end
+
+  # def has_enabled_shareable_preview?(user)
+  #   state == "draft" && user.can_share_previews? && type != "Consultation" && type != "Publication" && type != "DocumentCollection"
```
BeckaL, line 85: we can remove this now

BeckaL (approved):
Thanks for all your work on this! LGTM. As discussed, would be good if we could also make the text change to the preview button (assuming it's easy e.g. no publishing e2e tests break) as part of this to save another round of reviews and merging :-) Nikin's design has it in, might also just be worth showing it to him one more time once it's in to double check that everything looks good.

PR #6570, opened 2022-05-20
Auth bypass ids for existing assets

BeckaL:

This PR is part of work for shareable previews to ensure that a draft edition's assets are viewable via the preview functionality. We have already implemented this for assets linked to new editions, but for assets for existing drafts (i.e. ones that were created prior to us deploying the last PR) we need to resend these to asset-manager with their edition's auth_bypass_id.

There are two elements to this PR:

1) The introduction of a generic "UpdateWhitehallAsset" worker (and an associated queue) that resends an asset to asset manager with any associated attributes. This is currently only used by the rake task but the aim of this is to introduce something that might be able to be more widely used in the codebase later (currently updates to assets are done synchronously, unlike creating assets)

2) Rake tasks to republish various draft assets with their edition's auth_bypass_id

Trello: https://trello.com/c/Lm9RyCuP/397-ensure-existing-draft-assets-have-authbypassids

### DIFF
--- a/app/workers/asset_manager_update_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_update_whitehall_asset_worker.rb
@@ -1,0 +1,19 @@
+class AssetManagerUpdateWhitehallAssetWorker < WorkerBase
+  sidekiq_options queue: "asset_manager_updater"
+
+  def perform(attachment_data_global_id, attributes)
+    asset_data = GlobalID::Locator.locate(attachment_data_global_id)
+    path = asset_data.file.asset_manager_path
+    AssetManager::AssetUpdater.call(asset_data, path, attributes)
+
+    # Update any other versions of the file (e.g. PDF thumbnail, or resized versions of an image)
+    asset_versions = asset_data.file.versions.values.select(&:present?)
+    asset_versions.each do |version|
+      AssetManager::AssetUpdater.call(asset_data, version.asset_manager_path, attributes)
+    end
+  rescue AssetManager::ServiceHelper::AssetNotFound,
+         AssetManager::AssetUpdater::AssetAlreadyDeleted,
+         ActiveRecord::RecordNotFound => e
+    logger.error "AssetManagerUpdateWhitehallAssetWorker: #{e.message}"
+  end
+end
--- a/lib/tasks/asset_manager.rake
+++ b/lib/tasks/asset_manager.rake
@@ -1,0 +1,61 @@
+namespace :asset_manager do
+  desc "Update draft file attachments with auth_bypass_ids"
+  task update_attachments_with_auth_bypass_ids: :environment do
+    # republish assets who's editions are draft with an auth bypass id
+    latest_draft_editions = Edition.in_pre_publication_state.latest_edition
+
+    latest_draft_editions.find_each do |edition|
+      next unless edition.respond_to?(:attachments)
+
+      edition.attachments.files.each do |file_attachment|
+        new_attributes = { auth_bypass_ids: [edition.auth_bypass_id] }
+        attachment_data_id = file_attachment.attachment_data.to_global_id
+        AssetManagerUpdateWhitehallAssetWorker.perform_async_in_queue("asset_manager_updater", attachment_data_id, new_attributes)
+      end
+    end
+  end
+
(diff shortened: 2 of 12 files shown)


ollietreend (commented):
This looks excellent ✨ 

Just a few small comments from me, which I'll address and push up!

ollietreend (approved):

lib/tasks/asset_manager.rake
```diff
@@ -0,0 +1,61 @@
+namespace :asset_manager do
+  desc "Update draft file attachments with auth_bypass_ids"
+  task update_attachments_with_auth_bypass_ids: :environment do
+    # republish assets who's editions are draft with an auth bypass id
+    latest_draft_editions = Edition.in_pre_publication_state.latest_edition
+
+    latest_draft_editions.each do |edition|
```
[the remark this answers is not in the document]
BeckaL: Oh nice, v useful :-)

app/workers/asset_manager_update_whitehall_asset_worker.rb
```diff
@@ -0,0 +1,15 @@
+class AssetManagerUpdateWhitehallAssetWorker < WorkerBase
+  sidekiq_options queue: "asset_manager_updater"
+
+  def perform(attachment_data_global_id, attributes)
+    asset_data = GlobalID::Locator.locate(attachment_data_global_id)
```
kevindew, line 5: Do we think there's risk that this may raise an exception? could an asset be attachment data be deleted during the execution of this?

app/workers/asset_manager_update_whitehall_asset_worker.rb
```diff
@@ -0,0 +1,15 @@
...
+    path = asset_data.file.asset_manager_path
+    AssetManager::AssetUpdater.call(asset_data, path, attributes)
+
+    # Update PDF thumbnail
+    if asset_data.try(:pdf?)
```
kevindew, line 10: rather minor but do we want to be checking `asset_data.file.thumbnail` (or maybe `asset_data.file.try(:thumbnail)` for existence? I'm wondering whether there's some edge cases (e.g. slow to generate preview) where you don't have a thumbnail image.

app/workers/asset_manager_update_whitehall_asset_worker.rb
```diff
@@ -0,0 +1,15 @@
+class AssetManagerUpdateWhitehallAssetWorker < WorkerBase
+  sidekiq_options queue: "asset_manager_updater"
+
+  def perform(attachment_data_global_id, attributes)
+    asset_data = GlobalID::Locator.locate(attachment_data_global_id)
+    path = asset_data.file.asset_manager_path
+    AssetManager::AssetUpdater.call(asset_data, path, attributes)
```
kevindew, line 7: I assume this method can (and will) raise exceptions on some assets that don't exist. It'd probably be prudent to handle them rather than letting a unresolvable retry cycle occur.

kevindew (commented):
This seems good to me, I've put in a few comments. I'm anticipating that this is temporary code that we'll remove after running so you can take my comments with a pinch of salt if you don't mind dealing with errors manually.

app/workers/asset_manager_update_whitehall_asset_worker.rb
ollietreend: ~Annoyingly, every `AttachmentData#file` (which returns an `AttachmentUploader`) will respond to `:thumbnail`, even if it doesn't have a thumbnail preview.~

~Thumbnails are [only generated for PDFs](https://github.com/alphagov/whitehall/blob/d5c9121d1bf4109a7ddc72c76417d81c1e81b47f/app/uploaders/attachment_uploader.rb#L26-L41), but the `#thumbnail` method always exists & returns an object regardless.~

You've actually sent me down a helpful path – thanks @kevindew! It turns out `ImageData` images get resized into various different versions – but we were only applying the auth_bypass_id to the main file.

I think there'll be a way for us to update additional CarrierWave file 'versions' so that they get auth_bypass_ids too.

ollietreend: @kevindew take a look at 70781e1cab699a9d623fe927775b4bef9c2d872e because I think it'll resolve this comment.

benjamineskola (approved):

PR #6575, opened 2022-05-24
Allow Responses to access their owning Consultation's content_id

edwardkerry:
The accessible format request form requires a content_id.

Response objects, and models that inherit from Response (ConsultationOutcome and ConsultationPublicFeedback) do not have content_ids as they are closely tied and owned by a Consultation which is a "standard" Publicationesque Edition.

This PR delegates the Response content_id to the owning consultation, and removes a guard clause preventing documents without a content id from using the Accessible Format Request form.

As a result all Whitehall documents with inaccessible attachments will be able to link to the new Accessible Format Request form.

[trello](https://trello.com/c/EotfMMsy/1165-allow-consultationoutcomes-to-use-the-new-accessible-format-request-form-whitehall)

### DIFF
--- a/app/models/response.rb
+++ b/app/models/response.rb
@@ -41,7 +41,7 @@ class Response < ApplicationRecord
     to_model.class.name.underscore
   end
 
-  delegate :public_timestamp, :first_published_version?, :slug, :document, :images, to: :consultation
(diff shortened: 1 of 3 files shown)


KludgeKML (approved):
Looks good to me.

PR #6577, opened 2022-05-24
Make Docker image run as non-root.

sengi:
Not using the govuk-ruby common base/builder images just yet, but that's the next step once this lands.

- Use Debian 11 (bullseye) because it makes it way easier to install compatible versions of Node.js and Yarn that are still supported (at least for now).
- Make the path for uploaded temp files configurable via an environment variable instead of hardcoding them to paths inside the app's source directory. Defaults remain the same so there's no change to the old EC2 deployments.
    - This affects the `asset-manager-tmp`, `bulk-upload-zip-file-tmp`, `carrierwave-tmp` dirs. There are a bunch of others (`attachment-cache`, `clean`, `incoming` etc.) but those are all disused since 2018.
- Fix a bunch of lints (found using hadolint, which is too rough around the edges to enable as a pre-commit/pre-merge but somewhat useful if taken with a pinch of salt).
- Use the right version of bundler. Turns out the feature where it tries to reinstall the right version of itself doesn't always do the right thing. Otherwise the app fails at runtime because the default bundler is still the wrong version.
- Avoid installing perl during build. (We don't need it.)

Also contains all the same fixes as alphagov/feedback#1432:

- Add some files to `.dockerignore`.
- Run as an unprivileged user by default (uid 1001, which is fairly common and consistent with Bitnami images because might as well).
- Set `RAILS_LOG_TO_STDOUT` because we always want that when running in a container.
- Symlink `/app/tmp` to `/tmp` so that third-party gems don't try to write to the (read-only) image filesystem. (Ruby/Rails doesn't seem to have a consistent convention for telling gems where to write temp files, but some default to `tmp` under the Rails root.)
- Run `rails assets:precompile` in the build stage.
- Avoid installing "suggested" and "recommended" dependencies of Debian packages.
- Avoid installing gem documentation when building the container image.
- Avoid including the Debian APT cache in the container image.
- Increase build parallelism (saves a couple of minutes on a 2019 MBP).

https://trello.com/c/Jef1t4xU

Tested: `docker build -t whitehall . && docker run -it --rm -e GOVUK_APP_DOMAIN=x -e GOVUK_WEBSITE_ROOT=x -e JWT_AUTH_SECRET=x whitehall`. Also tested manually in the integration environment; created a publication, uploaded images and attachments etc.

Rollout: this app is push-on-green and test coverage is sketchy, so look out when merging. Requires https://github.com/alphagov/publishing-e2e-tests/pull/494 to be merged first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,44 +1,91 @@
-# (unless we decide to use Bitnami instead)
-ARG base_image=ruby:3.0.4
+ARG ruby_version=3.0.4
+ARG base_image=ruby:$ruby_version-slim-bullseye
+ARG gem_home=/usr/local/bundle
 
 FROM $base_image AS builder
-# This image is only intended to be able to run this app in a production RAILS_ENV
-ENV RAILS_ENV=production
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-# TODO: have a separate build image which already contains the build-only deps.
+ARG gem_home
+ENV RAILS_ENV=production \
+    RAILS_LOG_TO_STDOUT=1 \
+    NODE_ENV=production \
+    GEM_HOME=$gem_home \
+    BUNDLE_PATH=$gem_home \
+    BUNDLE_BIN=$gem_home/bin \
+    PATH=$gem_home/bin:$PATH \
+    BUNDLE_WITHOUT="development test cucumber" \
+    BOOTSNAP_CACHE_DIR=/var/cache/bootsnap
 
-# Add yarn to apt sources
-RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
-RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+# TODO: set these in the builder image.
+ENV BUNDLE_IGNORE_MESSAGES=1 \
+    BUNDLE_SILENCE_ROOT_WARNING=1 \
+    BUNDLE_JOBS=12 \
+    MAKEFLAGS=-j12
 
+ENV ASSETS_PREFIX=/assets/whitehall \
+    GOVUK_APP_DOMAIN=unused \
+    GOVUK_WEBSITE_ROOT=unused
+
+# TODO: have an up-to-date builder image and stop running apt-get upgrade.
+# TODO: have a separate builder image which already contains the build-only deps.
 RUN apt-get update -qy && \
     apt-get upgrade -y && \
-    apt-get install -y build-essential nodejs yarn && \
-    apt-get clean
-RUN mkdir /app
+    apt-get install -y --no-install-recommends \
+        g++ make libc-dev curl yarnpkg libmariadb-dev-compat && \
+    ln -s /usr/bin/yarnpkg /usr/bin/yarn
+
+RUN mkdir -p /app && ln -fs /tmp /app/tmp && ln -fs /tmp /home/app
 WORKDIR /app
-COPY Gemfile Gemfile.lock .ruby-version package.json yarn.lock /app/
-RUN bundle config set deployment 'true' && \
-    bundle config set without 'development test' && \
-    bundle install --jobs 4 --retry=2
-RUN yarn install --production --frozen-lockfile
+RUN echo 'install: --no-document' >> /etc/gemrc && gem update --system --silent && gem cleanup
+COPY Gemfile Gemfile.lock .ruby-version /app/
+# Make the installed version of bundler match the one that wrote Gemfile.lock.
+RUN gem install bundler \
+        --silent \
+        --version "$(sed -e '1,/BUNDLED WITH/d' Gemfile.lock | grep -Eo '[0-9.]+')" && \
+    bundle install
+COPY package.json yarn.lock /app/
+RUN yarnpkg install --production --frozen-lockfile --non-interactive --link-duplicates
 COPY . /app
-# TODO: We probably don't want assets in the image; remove this once we have a proper deployment process which uploads to (e.g.) S3.
-RUN GOVUK_ASSET_ROOT=https://assets.publishing.service.gov.uk \
-  GOVUK_APP_DOMAIN=www.gov.uk \
-  GOVUK_WEBSITE_ROOT=www.gov.uk \
-  GOVUK_APP_DOMAIN_EXTERNAL=www.gov.uk \
-  JWT_AUTH_SECRET=secret \
-  bundle exec rake assets:precompile
+RUN bundle exec bootsnap precompile --gemfile .
+RUN bundle exec rails assets:precompile && rm -fr log bulk-upload-zip-file-tmp
+
 
 FROM $base_image
-ENV RAILS_ENV=production GOVUK_APP_NAME=whitehall
-# TODO: include nodejs in the base image (govuk-ruby).
-# TODO: apt-get upgrade in the base image
+
+# TODO: set these in the base image.
+ARG gem_home
+ENV RAILS_ENV=production \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,43 +1,90 @@
...
 RUN apt-get update -qy && \
     apt-get upgrade -y && \
-    apt-get install -y build-essential nodejs yarn && \
-    apt-get clean
-RUN mkdir /app
+    apt-get install -y --no-install-recommends \
+        g++ make libc-dev curl yarnpkg libmariadb-dev-compat && \
+    ln -s /usr/bin/yarnpkg /usr/bin/yarn
+
+RUN mkdir -p /app && ln -fs /tmp /app/tmp && ln -fs /tmp /home/app
```
nsabri1, line 37: why not make this part of the above RUN

nsabri1 (approved):
LGTM

Dockerfile
sengi: Could do, but it doesn't really belong there logically. The stuff above is about installing build-time dependencies via package managers, whereas this stanza is about building/installing the app.

nsabri1: thats fair point

sengi:
Still haven't been able to figure out why publishing-e2e-tests `Uploading an attachment on Whitehall Uploading an attachment on Whitehall` is failing. Can't get it to log anything useful.

sengi:
The test failure looks like this, but I'm pretty sure it's failing because something's going wrong with the file upload. Damned if I can get it to log anything useful from either the worker or the -admin process.

```
  1) Uploading an attachment on Whitehall Uploading an attachment on Whitehall
     Failure/Error: check "Test taxon"
     
     Capybara::ElementNotFound:
       Unable to find checkbox "Test taxon" that is not disabled
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/finders.rb:303:in `block in synced_resolve'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/base.rb:77:in `synchronize'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/finders.rb:292:in `synced_resolve'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/finders.rb:53:in `find'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/actions.rb:367:in `block in _check_with_label'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/base.rb:83:in `synchronize'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/actions.rb:366:in `_check_with_label'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/node/actions.rb:150:in `check'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/session.rb:762:in `check'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/dsl.rb:53:in `call'
     # /usr/local/bundle/gems/capybara-3.35.3/lib/capybara/dsl.rb:53:in `check'
     # ./spec/whitehall/uploading_attachment_spec.rb:30:in `given_i_have_a_draft_document_with_attachment'
     # ./spec/whitehall/uploading_attachment_spec.rb:9:in `block (2 levels) in <top (required)>'
```

nsabri1:
> I think it's very likely that it's the removal of this line which "fixes" the e2e tests and the base image is probably a red herring.

I suppose it could be, but on multiple tries and tests - both User and Image was required to be updated for the test to pass.

sengi:
So the e2e test fails with the full-fat image when running as non-root. The failure is the same as before. Unfortunately that means there's a real bug in this PR; we can't just keep the e2e test running as root and pretend there's no problem, because then we'll break production.

Next step is to (continue to) figure out what's really failing when it's running as non-root. Which is easier said than done, but it's very likely to be some filesystem permissions / path issue relating to uploading assets.

sengi:
I don't know why the other tests are failing since the rebase, but I'm pretty sure that's unrelated. Also seemed kinda flaky with a couple of different tests failing each time, but hasn't completely passed at all in four runs so who knows.

sengi:
At least the other testsuite is now passing again (didn't change anything - must have been something else that broke it).

I'm starting to suspect that the e2e test failures are specific to the e2e environment after all, since I haven't been able to get asset upload to fail when running this branch under govuk-docker. I'm able to manually follow along with the failing e2e test scenario but under govuk-docker it successfully creates the edition and uploads the asset (and I see the page containing the `taxonomy tags` string that the test looks for.

Also the error in the `docker.log` from the e2e test artefacts contains this, which kinda sounds like `/tmp` isn't writable in the e2e whitehall_admin container, somehow (I've no idea why that should be yet though) :

```
[35mwhitehall-admin_1                       |[0m 2022-07-11T15:45:14.703245782Z system temporary path is not writable: /tmp
[35mwhitehall-admin_1                       |[0m 2022-07-11T15:45:14.703270483Z /tmp is not writable: /tmp
[35mwhitehall-admin_1                       |[0m 2022-07-11T15:45:14.703274315Z . is not writable: /app
```

fredericfran-gds:
looks good, I don't know if you want a ruby developer to have a look too. Great work.

fredericfran-gds (approved):
looks good to me

sengi:
+ @theseanything could you have a quick look over my Ruby changes here, just in case I'm doing anything whacky? thanks! 🙏

app/models/bulk_upload.rb
```diff
@@ -107,7 +107,7 @@ def initialize(zip_file = nil)
     end
 
     def temp_dir
-      @temp_dir ||= Dir.mktmpdir(nil, BULK_UPLOAD_ZIPFILE_DEFAULT_ROOT_DIRECTORY)
+      @temp_dir ||= Dir.mktmpdir(nil, Whitehall.bulk_upload_tmp_dir)
```
theseanything, line 110: Do we need call `Dir.mktmpdir`, if we're already creating the directory in the `Whitehall.bulk_upload_tmp_dir` function with `FileUtils.mkdir_p`?

config/initializers/carrierwave.rb
```diff
@@ -1,11 +1,17 @@
+require "pathname"
 require "whitehall/asset_manager_storage"
 require "carrier_wave/sanitized_file"
 
 CarrierWave.configure do |config|
   config.storage_engines[:asset_manager] = "Whitehall::AssetManagerStorage"
   config.storage Whitehall::AssetManagerStorage
   config.enable_processing = false if Rails.env.test?
-  config.cache_dir = Rails.root.join "carrierwave-tmp"
+  uploads_root = if !ENV.fetch("GOVUK_UPLOADS_ROOT", "").empty?
```
theseanything, line 9: Nit: 

Not 💯 off the top of my head whether the Rails helper methods are available in the initialisers - but you should be able to simplify this to:
```suggestion
  uploads_root = if ENV["GOVUK_UPLOADS_ROOT"].present?
```
Avoids having to think through a double negative.

app/models/bulk_upload.rb
theseanything: I.e. should we just have `Whitehall.bulk_upload_tmp_dir` return the path and call FileUtils.mkdir_p here?

config/initializers/carrierwave.rb
sengi: Thanks, that reads much better!

app/models/bulk_upload.rb
theseanything: Ah righto! This makes much more sense now - we're passing a directory prefix, in which we make temporary directories within. Thanks for clarification!

theseanything (approved):

sengi:
Rebased against origin/main to update and to clean up the commit history.

PR #6579, opened 2022-05-24
Remove DVSA from accessible format request republish task

edwardkerry:
DVSA are opting out of the pilot so can be removed from the republish task


1pretz1 (approved):
😢

PR #6589, opened 2022-05-31
Enable CI using GitHub Actions

theseanything:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,18 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  test:
+    name: Test
+    uses: alphagov/govuk-infrastructure/.github/workflows/test-rails.yaml@main
+    with:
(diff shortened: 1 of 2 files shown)


aldavidson:
We've decided to stop pursuing it for now. Reasons include:

- there is still quirkiness & tech debt around how Whitehall sets up for the tests (requiring asset precompilation, shared_mustache precompilation, etc) which would need uncomfortable workarounds to force into the shared workflow.
- there are other important steps in the Jenkins jobs which we would not be able to port over at the moment (running the end to end tests,  releasing to integration, etc). We understand that eventually we will move to Argo CD for deployment when we move to the new K8s platform, but we don't know when that K8s move is likely to become imminent.
- if we are still going to be tied to Jenkins for the majority of the test/build/release process, the team would rather stick with just one CI / CD tool for the intervening time, until we're ready to port the whole pipeline over to the new platform & tools.

We'll keep trying to remove some of the unique quirks in how Whitehall sets up and runs tests, as they'll be useful improvements whatever happens

PR #6590, opened 2022-06-01
fix test coverage reporting when running with parallelization

aldavidson:

When running tests with multiple parallel workers (as in C.I. and locally by default), SimpleCov was not correctly merging the coverage reports from each individual worker. As a result, our test coverage was being mis-reported as 63%.

The issue dates back to [commit 979437](https://github.com/alphagov/whitehall/pull/6169/commits/9794376efb8bff59dbcd96fba586c60bd601c212) which removed the parallel_tests gem, which [SimpleCov detects automatically and works well alongside](https://github.com/simplecov-ruby/simplecov#user-content-notes-on-specific-frameworks-and-test-utilities), with Rails 6's native test parallelization, which [SimpleCov has known issues with](https://github.com/simplecov-ruby/simplecov/issues/718).

This PR adds a few lines of extra config on parallelization setup/teardown to fix SimpleCov's naming of the individual coverage reports, and to correctly surface the coverage output from each worker

Before:

```
$ govuk-docker run whitehall-lite rake test cucumber
(...) 12464 / 19675 LOC (63.35%) covered
```

After:

```
$ govuk-docker run whitehall-lite rake test cucumber
(...)
 16269 / 17333 LOC (93.86%) covered.
```


leenagupte (commented):
Al, this looks really good. Amazing bit of sleuthing!

Just a couple things to make the change history easier to follow for future developers: 1. Can you fix up the linting fixes to previous commit and 2. the PR description is great, could you add that to the commit message. Thanks!

aldavidson:
> 1. Can you fix up the linting fixes to previous commit and 2. the PR description is great, could you add that to the commit message. 

Done!

ollietreend (approved):
Great work @aldavidson ⭐ 

It's odd that your test run in the PR description seemed to produce 93.86% coverage, whereas Jenkins claims it to be 94.55%. I wonder if it's the difference between running MiniTest & Cucumber vs just Cucumber on its own, since [Jenkins does both](https://github.com/alphagov/whitehall/blob/f57384eb5f87c95038b68b7037032d9e0ee9a4b3/Jenkinsfile#L32-L35) but I think you only ran Cucumber.

Either way, looks great. Thanks for this!

leenagupte (approved):
Thanks for updating the commit message. 🎉 

You don't need the line about "rubocop fixes" though! 😄

aldavidson:
> It's odd that your test run in the PR description seemed to produce 93.86% coverage, whereas Jenkins claims it to be 94.55%.

That may also be because I rebased from main after writing the description?

aldavidson:
> You don't need the line about "rubocop fixes" though! smile

Eep! I'm rusty :) amended now, thanks

PR #6591, opened 2022-06-01
Compact the government_id for html attachments

davidgisbey:
## Description 

This is causing issues downstream at the moment as it passes nil within a blank array which causes the publishing API to blow up downstream.

## Sentry error

Example sentry error: https://gds.slack.com/archives/C02L13S214K/p1654082398471929


ollietreend (approved):
👍🏻 would be good to follow up with tests after we've fixed the immediate issue

PR #6594, opened 2022-06-07
Make default rake task run all tests

aldavidson:
This will make Whitehall more consistent with the majority of other Rails apps on GOV.UK, and allow us to use the new [shared workflow](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/test-rails.yaml) to run our CI suite in Github Actions, rather than Jenkins.

Full details in the [Trello card](https://trello.com/c/U9duqlgV/82-make-sure-both-jenkins-and-the-default-rake-task-run-the-same-tests-the-same-way)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -21,4 +21,4 @@ end
 Whitehall::Application.load_tasks
 
 Rake::Task[:default].clear if Rake::Task.task_defined?(:default)
-task default: %i[lint test jasmine pact:verify]
+task default: %i[lint test shared_mustache:compile cucumber jasmine pact:verify]


aldavidson:
> seems to be missing all those odd extra directories

I'm able to run this rake task locally via docker without those directories, so I didn't think they were required. Looks to me like they're only needed for the publishing-e2e-tests, so maybe we can port that over separately after the 'standard' tests?

kevindew:
> > seems to be missing all those odd extra directories
> 
> I'm able to run this rake task locally via docker without those directories, so I didn't think they were required. Looks to me like they're only needed for the publishing-e2e-tests, so maybe we can port that over separately after the 'standard' tests?

They won't actually do anything for the e2e tests as that runs off a [different jenkinsfile](https://github.com/alphagov/publishing-e2e-tests/blob/main/Jenkinsfile) - perhaps that step in the Jenkinsfile is redundant. 

I have those directories myself locally and didn't create them manually so I guess something creates them (aside: they're pretty odd would love to get rid of them).

PR #6605, opened 2022-06-10
Refactor Edition audit trail methods

ollietreend:
This PR refactors methods on Edition related to 'audit trails' (a.k.a. version history). In doing so, we've been able to improve code readability and performance/efficiency.

This refactor has come out of some [investigatory work][1] that I've been carrying out to try and reduce the performance impact of documents with a large edit history. Part of that work involves reducing our dependence on the [`AuditEntry` classes][1] which have thus far made it difficult to perform more efficient pagination of large document edit histories.

Special thanks to @kevindew for his determination to have `Edition#submitted_by` perform just one database query. 🧙🏻‍♂️

[1]: https://github.com/alphagov/whitehall/compare/limit-document-history
[2]: https://github.com/alphagov/whitehall/blob/1bc8e03611294427e0eb1c3c500f8c2272e8556a/app/models/edition/audit_trail.rb#L125-L206

Trello: https://trello.com/c/JTaGC5v0

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -40,6 +40,8 @@ class Document < ApplicationRecord
   has_many :document_collections, through: :document_collection_groups
   has_many :features, inverse_of: :document, dependent: :destroy
 
+  has_many :edition_versions, through: :editions, source: :versions
+
   before_save { check_if_locked_document(document: self) unless locked_changed? }
 
   validates :content_id, presence: true
@@ -104,6 +106,10 @@ class Document < ApplicationRecord
 
   def first_published_date
     published_edition.first_public_at if published?
+  end
+
+  def first_published_on_govuk
+    edition_versions.where(state: "published").pick(:created_at)
   end
 
   def change_history
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -523,7 +523,24 @@ EXISTS (
   end
 
   def submitted_by
-    most_recent_submission_audit_entry.try(:actor)
+    # Find this edition's most recent state change from non-submitted to submitted.
+    # This will tell us when it was most recently submitted, even if there were subsequent
+    # changes from other users while the document remained in a "submitted" state.
+
+    latest_submitted_version = versions_desc.select("created_at, id")
+                                            .where(state: :submitted)
+                                            .limit(1)
+
+    pre_submitted_version = versions_desc.select("created_at, id")
+                                         .where.not(state: "submitted")
+                                         .where("(created_at, id) < (:submitted_version)", submitted_version: latest_submitted_version)
+                                         .limit(1)
+
+    first_submitted_version = versions_asc.where(state: "submitted")
+                                          .where("(created_at, id) > (:pre_submitted_version)", pre_submitted_version: pre_submitted_version)
+                                          .first
+
+    first_submitted_version.try(:user)
   end
 
   def title_with_state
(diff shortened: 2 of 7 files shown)


davidgisbey (approved):
Makes sense to me 👍

app/models/document.rb
```diff
@@ -126,6 +130,11 @@ def humanized_document_type
     document_type.underscore.tr("_", " ")
   end
 
+  def edition_versions
+    Version.where(item_type: "Edition", item_id: editions.select(:id))
+           .order(created_at: :asc, id: :asc)
```
davidgisbey, line 135: not a blocker and i'm guessing the reason, but why do we need to order by both of these? Would we ever get a record where the created at and id are out of sync?

ollietreend: This replicates the same sorting as is already defined on `Edition#versions` here:

https://github.com/alphagov/whitehall/blob/d4f946000a26adbf375e95e6e2bc771cdcf5622d/app/models/edition/audit_trail.rb#L17

Versions will primarily be sorted by their `created_at` timestamp. However the `id` is used as a tie-breaker, which is possible because it's an auto-incrementing number so newer ones will always be higher. So the secondary sort on `:id` just protects against that edge case. Without it, we could end up with an indeterminate sort order which could cause history to sort unpredictably in those cases.

I believe the resolution of the `created_at` timestamp field is one second, which is the [default for MySQL `DATETIME` fields](https://dev.mysql.com/doc/refman/8.0/en/fractional-seconds.html). It's not likely that the edition will be updated multiple times within the same second, but it's not impossible. There could be a race condition between users hitting 'save', or even a script which loops and updates things more than once, etc.

In practice, I don't think this is likely to happen. But it's actually pretty common in our test suite. We regularly call `Timecop.freeze` to stop time, and then sequentially create new versions of an edition to manipulate it into the desired state.

For example, here's what `publication.versions` looks like [in this test](https://github.com/alphagov/whitehall/blob/d4f946000a26adbf375e95e6e2bc771cdcf5622d/test/unit/edition_test.rb#L291-L299):

Screenshot 2022-06-13 at 17 15 36

Note that the `created_at` timestamps are all identical. So we've implicitly come to depend upon on that secondary sort acting as a tie-breaker.

app/models/document.rb
```diff
@@ -126,6 +130,11 @@ def humanized_document_type
     document_type.underscore.tr("_", " ")
   end
 
+  def edition_versions
+    Version.where(item_type: "Edition", item_id: editions.select(:id))
```
kevindew, line 134: I think there's a neater way you can do this association. I've got it in my draft PR: https://github.com/alphagov/whitehall/pull/6607/files#diff-893a2b609f3fc4aade494f624a20993eb6c32598512b5e70d9ad39866b8eb51dR43

app/models/document.rb
```diff
@@ -106,6 +106,10 @@ def first_published_date
     published_edition.first_public_at if published?
   end
 
+  def first_published_on_govuk
+    edition_versions.where(state: "published").first.try(:created_at)
```
kevindew, line 110: Nit: probably nicer to just pluck the date here instead of getting the model and discarding it. 

There was a new(ish) method added, pick, that'd do a good job: https://sikac.hu/relation-pick-in-rails-6-0-87b2112f6097

kevindew (commented):
Awesome job, just a couple of minor tweaks suggested.

app/models/document.rb
ollietreend: Oh, nice! Thanks.

kevindew (approved):

PR #6606, opened 2022-06-13
Add national applicability data to html attachments

davidgisbey:
## Description

We're doing some work to expose national applicability data for html attachments. If a consultation or Publication only relates to some of the devolved nations, then it also renders for html attachments that belong to it.

This PR does the work to ensure that when applicable the devolved nations details are passed down in the details hash. These will then be exposed on the frontend.

## Related PRs 

GOVUK Content Schema - https://github.com/alphagov/govuk-content-schemas/pull/1100

Governement Frontend - https://github.com/alphagov/government-frontend/pull/2460

## Trello card

https://trello.com/c/pLnsviDN/402-implement-devolved-nations-national-applicability-banner-on-html-attachments

### DIFF
--- a/app/presenters/publishing_api/html_attachment_presenter.rb
+++ b/app/presenters/publishing_api/html_attachment_presenter.rb
@@ -56,12 +56,20 @@ module PublishingApi
     end
 
     def details
-      {
+      details_hash = {
         body: body,
         public_timestamp: public_timestamp,
         first_published_version: first_published_version?,
         political: political?,
       }
+
+      maybe_add_national_applicability(details_hash)
+    end
+
+    def maybe_add_national_applicability(details_hash)
+      return details_hash unless item.attachable.try(:nation_inapplicabilities)&.any?
+
+      details_hash.merge(national_applicability: item.attachable.national_applicability)
     end
 
     def body
--- a/test/unit/presenters/publishing_api/html_attachment_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/html_attachment_presenter_test.rb
@@ -126,4 +126,46 @@ class PublishingApi::HtmlAttachmentPresenterTest < ActiveSupport::TestCase
     assert_equal [html_attachment.attachable.organisations.first.content_id],
                  presenter.links[:primary_publishing_organisation]
   end
+
+  test "HtmlAttachments parent object has national_applicability exclusions" do
+    scotland_nation_inapplicability = create(
+      :nation_inapplicability,
+      nation: Nation.scotland,
+      alternative_url: "http://scotland.com",
+    )
+    consultation = create(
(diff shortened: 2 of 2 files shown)


app/presenters/publishing_api/html_attachment_presenter.rb
```diff
@@ -56,12 +56,20 @@ def base_path
...
         body: body,
         public_timestamp: public_timestamp,
         first_published_version: first_published_version?,
         political: political?,
       }
+
+      maybe_add_national_applicability(details_hash)
+    end
+
+    def maybe_add_national_applicability(content)
```
ollietreend, line 69: It might be nice to name this parameter `details_hash` rather than `content`, for the sake of consistency.

```suggestion
    def maybe_add_national_applicability(details_hash)
```

app/presenters/publishing_api/html_attachment_presenter.rb
```diff
@@ -56,12 +56,20 @@ def base_path
...
+      return content unless item.attachable.try(:nation_inapplicabilities)&.any?
```
ollietreend, line 70: I think we should be able to use the `all_nation_applicability_selected?` method here, which might be nicer for readability rather than checking for the presence of associated `nation_inapplicabilities`.

`all_nation_applicability_selected?` [should be safe](https://github.com/alphagov/whitehall/blob/41acceca56b077a028873e3e8ca8fb545d52ffcd/app/models/edition.rb#L569-L571) to call on every type of Edition. It's [used in the edit form](https://github.com/alphagov/whitehall/blob/41acceca56b077a028873e3e8ca8fb545d52ffcd/app/views/admin/editions/_nation_fields.html.erb#L5) to drive that checkbox, and should always be correct according to [the validation rules](https://github.com/alphagov/whitehall/blob/d33a442db01b8c705f87c52b4ad8395dbc7b306a/app/models/edition/national_applicability.rb#L21-L28).

```suggestion
      return content if item.attachable.all_nation_applicability_selected?
```

ollietreend (approved):
Looks good to me 👍🏻 

A couple of stylistic opinions from me, but nothing blocking.

app/presenters/publishing_api/html_attachment_presenter.rb
davidgisbey: It's not always editions tried in so we will still need try but this is def nicer 👍

ollietreend: Ah – this might not work then... 🤔

If a non-Edition model is passed in, I guess it'll skip past that guard clause now that it's checking for the presence of `all_nation_applicability_selected?`. So it'll hit `item.attachable.national_applicability` in the line below.

PR #6607, opened 2022-06-13
Spike - demonstration of paginated history/remarks on Admin::Edition#show

kevindew:
Trello: https://trello.com/c/JTaGC5v0/486-royal-courts-of-justice-incident

This is a prototype proof-of-concept of changing the approach to loading history/remarks when showing the show page of an Edition. 

The previous approach was to load all remarks and history from the database when displaying the page. For a page with a long history this would add a significant performance overhead (for example: https://www.gov.uk/government/publications/royal-courts-of-justice-cause-list has 1000's of updates).

This offers a like-for-like replacement of the history aspect - this was achievable as that already has pagination, albeit an inefficient approach.

For remarks there was not previously a pagination approach and on the live version there are semantic rendering issues on the presentation of which edition they are associated with. Therefore this offers an MVP style approach of this with pagination. 

More details in the commits.

### DIFF
--- a/app/controllers/admin/edition_audit_trail_controller.rb
+++ b/app/controllers/admin/edition_audit_trail_controller.rb
@@ -3,6 +3,6 @@ class Admin::EditionAuditTrailController < Admin::EditionsController
 
   def index
     @edition = Edition.find(params[:id])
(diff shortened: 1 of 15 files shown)


kevindew:
This is no longer needed

PR #6611, opened 2022-06-13
Add rake task for republishing docs with HMTL attachments & nation_inapplicability

davidgisbey:
## Description

When a document is only applicable for specific countries HTML attachments that belong to it should also render the devolved nations component.

This adds a rake task which republishes documents which have specific nation applicability and an associated HTML attachment.

## Trello card 

https://trello.com/c/pLnsviDN/402-implement-devolved-nations-national-applicability-banner-on-html-attachments


lib/tasks/republish_nation_applicable_html_attachments.rake
```diff
@@ -0,0 +1,16 @@
+desc "Republish all documents with html attachments and national inapplicabilites"
+task republish_nation_inapplicable_html_attachments: :environment do
+  document_ids = Edition
+    .publicly_visible
+    .where(id: NationInapplicability.joins(:edition).distinct(:edition_id).pluck(:edition_id))
```
ollietreend, line 5: We **should** be able to rely on the `all_nation_applicability` field here.

```suggestion
    .where(all_nation_applicability: false)
```

ollietreend (approved):
Similar to [my comment](https://github.com/alphagov/whitehall/pull/6606/files#r895933473) on #6606 – I think we should be able to rely in that `all_nation_applicability` field to aid code readability.

But as it is, this still looks good 👍🏻

PR #6613, opened 2022-06-14
Improve mustache templates compilation config

jyoung-gds:
[Trello ticket link](https://trello.com/c/ISXwzUZE/475-make-shared-mustache-templates-less-annoying)

- [Commit 1:](https://github.com/alphagov/whitehall/commit/659a517423171db7c79e07cc8c2f094c8c4c414e) Adds app/assets/javascripts/templates.js to .gitignore
- [Commit 2:](https://github.com/alphagov/whitehall/commit/8662edbb4a9bc0b086ba09975e23acac300762fd) Extends the asset:precompile rake task to also run the shared_mustache:compile rake task and adds a test file to check shared_mustache:compile is invoked when asset:precompile is invoked.
- [Commit 3:](https://github.com/alphagov/whitehall/commit/d0db0c58df4264963062c851ebfb4875d5193e88) Removes all direct use of shared_mustache:compile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@
 /yarn-error.log
 
 spec/reports/pacts
+
+/app/assets/javascripts/templates.js
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ RUN GOVUK_ASSET_ROOT=https://assets.publishing.service.gov.uk \
   GOVUK_WEBSITE_ROOT=www.gov.uk \
   GOVUK_APP_DOMAIN_EXTERNAL=www.gov.uk \
   JWT_AUTH_SECRET=secret \
-  bundle exec rake shared_mustache:compile assets:precompile
+  bundle exec rake assets:precompile
 
 FROM $base_image
 ENV RAILS_ENV=production GOVUK_APP_NAME=whitehall
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,6 @@ node {
           govuk.runRakeTask("test:publishing_schemas --trace")
         } else {
           govuk.runRakeTask("ci:setup:minitest test --trace")
-          govuk.runRakeTask("shared_mustache:compile")
           sh("bundle exec cucumber")
           govuk.runRakeTask("jasmine")
         }
--- a/Rakefile
+++ b/Rakefile
@@ -21,4 +21,4 @@ end
 Whitehall::Application.load_tasks
 
 Rake::Task[:default].clear if Rake::Task.task_defined?(:default)
-task default: %i[lint test shared_mustache:compile cucumber jasmine pact:verify]
+task default: %i[lint test assets:precompile cucumber jasmine pact:verify]
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -1,2 +1,5 @@
 # Maintain Rails < 7 behaviour of running yarn:install before assets:precompile
 Rake::Task["assets:precompile"].enhance(["yarn:install"])
+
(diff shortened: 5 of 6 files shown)


jyoung-gds:
I wasn't sure about removing shared_mustache:compile from the default rake tasks in Rakefile. It was called explicitly before, but was also being called in jasmine:prepare when the jasmine rake task runs. I'm not sure if the assets:precompile in jasmine:prepare means we can remove shared_mustache:compile from the default rake tasks. Lmk.

jyoung-gds:
I also wasn't sure about removing shared_mustache:compile from the [Jenkinsfile](https://github.com/alphagov/whitehall/blob/2ca0bad4f9f409cf046757472783846158ad3f7b/Jenkinsfile#L33), and documentation (i.e. [README](https://github.com/alphagov/whitehall/blob/main/README.md#shared-mustache-templates) and [this error message](https://github.com/alphagov/whitehall/blob/2ca0bad4f9f409cf046757472783846158ad3f7b/features/support/missing_mustache_templates_helper.rb#L12))

.gitignore
```diff
@@ -30,3 +30,5 @@
 /yarn-error.log
 
 spec/reports/pacts
+
+/app/assets/javascripts/templates.js
```
kevindew, line 34: Looks like you're missing a new line character on this file.

test/unit/tasks/assets_test.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    test "shared mustache templates get compiled" do
+      mustache.reenable
+      task.reenable
+      mustache.expects(:execute)
+
+      task.invoke
+    end
+  end
+end
```
kevindew, line 18: I don't think there's much value in having this test since we're really just testing rake enhance works. And if it's not set-up we'd expect to have problems with tests.

kevindew:
> I wasn't sure about removing shared_mustache:compile from the default rake tasks in Rakefile. It was [called explicitly before](https://github.com/alphagov/whitehall/blob/2ca0bad4f9f409cf046757472783846158ad3f7b/Rakefile#L24), but was also being called in [jasmine:prepare](https://github.com/alphagov/whitehall/blob/2ca0bad4f9f409cf046757472783846158ad3f7b/package.json#L11) when the jasmine rake task runs. I'm not sure if the assets:precompile in jasmine:prepare means we can remove shared_mustache:compile from the default rake tasks. Lmk.

We should probably replace it with assets:precompile so the tests still run successfully.

> I also wasn't sure about removing shared_mustache:compile from the [Jenkinsfile](https://github.com/alphagov/whitehall/blob/2ca0bad4f9f409cf046757472783846158ad3f7b/Jenkinsfile#L33), and documentation (i.e. [README](https://github.com/alphagov/whitehall/blob/main/README.md#shared-mustache-templates) and [this error message](https://github.com/alphagov/whitehall/blob/2ca0bad4f9f409cf046757472783846158ad3f7b/features/support/missing_mustache_templates_helper.rb#L12))

We can remove it from the Jenkinsfile as assets:precompile will run before handle it.

I think it's ok to leave those bits in. I've got a prototype of the code to remove shared_mustache from this app completely that I hope to open in the next few days, so hopefully a lot of the bits here are quite short lived.

.gitignore
jyoung-gds: I've added a new line here but for some reason it's not showing in files changed. It is there though I promise!

test/unit/tasks/assets_test.rb
jyoung-gds: Removed it, thanks

kevindew (approved):
Great stuff. All looks good to me.

PR #6618, opened 2022-06-16
Add rake tasks for amending a change note

brucebolt:
The [documentation](https://docs.publishing.service.gov.uk/manual/howto-modify-change-note.html#whitehall) currently expects developers to log into a production Rails console, perform a fuzzy match then update the attribute within the console 😬.

This is _very_ bad practice. There is also no audit trail of who made the change or what was changed.

Adding a rake task will allow us to avoid using the console for this task, until a user interface is built for this regular task.

Usage:
1. Get a list of change notes for the given document:
```
rake change_note:list[content_id]
```

2. From the output, get the edition ID for the edition whose change note
needs amending, then amend the change note:
```
rake change_note:amend[edition_id, "New change note", email]
```

The email of the person running the rake task is required. This creates an editorial remark in Whitehall, so there is an audit log of the change that was made and who made it.

We need to skip validations in updating the change note as Whitehall does not permit changes to published editions.

### DIFF
--- a/lib/tasks/change_note.rake
+++ b/lib/tasks/change_note.rake
@@ -1,0 +1,48 @@
+namespace :change_note do
+  desc "List change notes for a content ID"
+  task :list, %i[content_id] => :environment do |_, args|
+    change_history = Document
+      .find_by(content_id: args[:content_id])
+      .editions
+      .where
+      .not(major_change_published_at: nil)
+      .pluck(:id, :major_change_published_at, :change_note)
+
+    if change_history.empty?
+      puts "No change notes found"
+      next
+    end
+
+    change_history.each do |change|
+      puts change.join("\t")
+    end
+  end
+
+  desc "Amend a change note for an edition"
+  task :amend, %i[edition_id new_change_note email] => :environment do |_, args|
+    edition = Edition.find(args[:edition_id])
+    old_change_note = edition.change_note
+
+    user = User.find_by(email: args[:email])
+    unless user
+      puts "User with email #{args[:email]} not found"
+      next
+    end
+
+    # rubocop:disable Rails/SkipsModelValidations
+    edition.update_attribute(:change_note, args[:new_change_note])
+    # rubocop:enable Rails/SkipsModelValidations
+
+    EditorialRemark.create!(
+      edition: edition,
+      body: "Updated change note from #{old_change_note} to #{args[:new_change_note]}",
+      author: user,
+      created_at: Time.zone.now,
+      updated_at: Time.zone.now,
+    )
+
+    PublishingApiDocumentRepublishingWorker.perform_async(edition.document.id)
+
+    puts "Updated change note from #{old_change_note} to #{args[:new_change_note]}"
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/tasks/change_note.rake
```diff
@@ -0,0 +1,37 @@
...
+
+    change_history.each do |change|
+      puts change.join("\t")
+    end
+  end
+
+  desc "Amend a change note for an edition"
+  task :amend, %i[edition_id new_change_note email] => :environment do |_, args|
+    edition = Edition.find(args[:edition_id])
+    user_id = User.find_by(email: args[:email]).id
```
hannako, line 18: I guess there's a small chance that we might end up calling `.id` on `nil` if for example the task is run by someone new to gov.uk who hasn't set up signon access for Whitehall yet?

lib/tasks/change_note.rake
```diff
@@ -0,0 +1,37 @@
+namespace :change_note do
+  desc "List change notes for a content ID"
+  task :list, %i[content_id] => :environment do |_, args|
+    change_history = Document
+      .find_by(content_id: args[:content_id])
+      .editions
+      .where(minor_change: false)
+      .pluck(:id, :major_change_published_at, :change_note)
+
+    change_history.each do |change|
```
hannako, line 10: v minor comment:
If you put in the wrong `content_id` and get an empty array for `change_history`, there won't be any output right? Which I guess could be a bit confusing. Maybe  `if change_history.empty? puts "no change notes found" ` kind of thing might be nice?

lib/tasks/change_note.rake
```diff
@@ -0,0 +1,37 @@
...
+
+    EditorialRemark.create!(
+      edition: edition,
+      body: "Updated change note from #{old_change_note} to #{args[:new_change_note]}",
+      author_id: user_id,
+      created_at: Time.zone.now,
+      updated_at: Time.zone.now,
+    )
+
+    PublishingApiDocumentRepublishingWorker.perform_async_in_queue(edition.document.id)
```
hannako, line 33: Do we not need to specify a `queue` here? I'm not very familiar with publishing pipeline.

lib/tasks/change_note.rake
brucebolt: That's a good point.  I'll add a guard to ensure the user exists before we continue any further.

lib/tasks/change_note.rake
brucebolt: I'll add an output for when there's no history.  That should make this clearer.

lib/tasks/change_note.rake
brucebolt: I don't think we do, unless we are bulk republishing.

hannako (approved):
Looks good to me! 2nd line devs of the future salute you 🖖

lib/tasks/change_note.rake
brucebolt: Actually, I think a queue name is needed with `perform_async_in_queue`.  I'll change this to `perform_async` so it uses the default publishing queue.

PR #6620, opened 2022-06-16
Port documents filter to no longer use shared_mustache

kevindew:
Trello: https://trello.com/c/ZuM77dVF/136-sharedmustache-is-a-thing

This PR ends up being a bit involved so is best reviewed commit by commit. It finishes paying down a tech debt that we've had since 2019.

The main focus is the removal of the last usage of shared_mustache, which are the non-english finders (such as https://www.gov.uk/government/publications.de). It removes the use of shared_mustache by considering that the finders are much simpler than the ones the initial code was wrote for and replacing a complex AJAX process for world_locations filtering with a `form.submit()` call when the select field changes value.

However, to do that I did need to actually fix the form which has been broken for the last 8 months.

The PR then proceeds with the removal of shared_mustache and related accommodations. I did need to leave a basic no-op rake task of `shared_mustache:compile` so that deployments can work with and without this, which will need a future amend.


app/views/documents/_filter_table.html.erb
```diff
@@ -0,0 +1,61 @@
+<% if filter[:results].any? %>
```
davidgisbey, line 1: Looks good! Something like this would be a prime candidate to unit test if we every decide to use something like view components though! There's a lot going on

davidgisbey:
Have gone through the code and deployed it out to int to manually test it. It looks good to me!

app/views/documents/_filter_table.html.erb
kevindew: Yup, it was quite a delicate operation to check it was all outputting correctly.

davidgisbey (approved):

davidgisbey:
@kevindew i did sorry!

PR #6635, opened 2022-06-27
Add emphasised_organisation to topical events presenter

callumknights:
Topical event rendering is being migrated to collections, so
these organisations need to be in the content item.

### DIFF
--- a/test/unit/presenters/publishing_api/topical_event_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/topical_event_presenter_test.rb
@@ -11,6 +11,12 @@ class PublishingApi::TopicalEventPresenterTest < ActiveSupport::TestCase
       logo_alt_text: "Alternative text",
     )
     create(:topical_event_about_page, topical_event: topical_event, read_more_link_text: "Read more about this event")
+
+    first_lead_org = create(:organisation)
+    first_lead_org.organisation_classifications.create!(classification_id: topical_event.id, lead: true, lead_ordering: 1)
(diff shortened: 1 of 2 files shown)


brucebolt (approved):

test/unit/presenters/publishing_api/topical_event_presenter_test.rb
```diff
@@ -41,6 +41,7 @@ class PublishingApi::TopicalEventPresenterTest < ActiveSupport::TestCase
       details: {
         about_page_link_text: topical_event.about_page.read_more_link_text,
         body: Whitehall::GovspeakRenderer.new.govspeak_to_html(topical_event.description),
+        emphasised_organisations: topical_event.lead_organisations.map(&:content_id),
```
BeckaL, line 44: This topical event actually has any lead organisations set, this test also passes if you replace `topical_event.lead_organisations.map(&:content_id)` with `[]`. It would be nice to test it with some actual data. 

I'm also not a fan of creating the test assertion with the same code as the implementation when you can avoid it as (I know it's done in the above two lines too but maybe harder with govspeak). If there's something off with your implementation you wouldn't spot it - so I'd prefer just doing the dumb thing of putting the array of content ids in here (you could create the lead organisations from a variable set to the array of content ids, and reuse that here, so it's clear what the link is).

JonathanHallam: have sorted this, going to merge!

PR #6639, opened 2022-06-29
Show attachment changes in change note box

beccapearce:
Update the change note box:
- Improve content around choosing major or minor changes
- Include a list of attachments that have been changed

Before: 
Screenshot 2022-07-12 at 17 42 16

After :
image

Trello card: https://trello.com/c/isH3BtNG/419-display-attachment-changes-in-edition-change-note

### DIFF
--- a/app/assets/javascripts/admin/views/editions/_form.js
+++ b/app/assets/javascripts/admin/views/editions/_form.js
@@ -26,14 +26,7 @@
       var $changeNotesSection = $('.js-change-notes-section', $fieldset)
 
       $radioButtons.change(showOrHideChangeNotes)
-      appendHintToFormLabel()
       showOrHideChangeNotes()
-
-      function appendHintToFormLabel () {
-        var $label = $($changeNotesSection).find('label')
-        var exampleText = 'For example, "Edited chapter 6 - centres in Cardiff and Aberystwyth have closed."'
-        $label.append('<p class="hint"; style="font-weight: normal">' + exampleText + '</p>')
-      }
 
       function showOrHideChangeNotes () {
         if ($minorChangeRadioButton.prop('checked')) {
--- a/app/models/attachable.rb
+++ b/app/models/attachable.rb
@@ -1,6 +1,6 @@
 module Attachable
   extend ActiveSupport::Concern
-  EDITION_CREATE_GRACE_PERIOD = 30.seconds
+  EDITION_CREATE_GRACE_PERIOD = 20.seconds
 
   included do
     has_many :attachments,
@@ -77,7 +77,7 @@ module Attachable
     # Using a 30 second grace period because GovspeakContent#render_govspeak! historically used to change the updated_at timestamp
     # when running in an async worker to convert body to HTML. This meant that all HTML attachments ended up with
     # (updated_at > created_at) so would show as 'changed'.
-    updated_html_attachments = html_attachments.joins(:govspeak_content).where("govspeak_contents.updated_at > DATE_ADD(govspeak_contents.created_at,  INTERVAL 30 SECOND)")
+    updated_html_attachments = html_attachments.joins(:govspeak_content).where("govspeak_contents.updated_at > DATE_ADD(govspeak_contents.created_at,  INTERVAL 20 SECOND)")
     updated = (updated_attachments + updated_html_attachments) - created
 
     created.map { |a| ChangedAttachment.new(a, :created) } +
--- a/app/models/attachable.rb
+++ b/app/models/attachable.rb
@@ -1,11 +1,13 @@
 module Attachable
   extend ActiveSupport::Concern
+  EDITION_CREATE_GRACE_PERIOD = 30.seconds
 
   included do
     has_many :attachments,
              -> { not_deleted.order("attachments.ordering, attachments.id") },
              as: :attachable,
              inverse_of: :attachable
+
     has_many :html_attachments,
              -> { not_deleted.order("attachments.ordering, attachments.id") },
              as: :attachable
@@ -13,6 +15,11 @@ module Attachable
     has_many :deleted_html_attachments,
              -> { deleted },
              class_name: "HtmlAttachment",
+             as: :attachable
+
+    has_many :deleted_attachments,
+             -> { deleted },
+             class_name: "Attachment",
              as: :attachable
 
     if respond_to?(:add_trait)
@@ -58,6 +65,24 @@ module Attachable
 
   def attachables
     [self]
+  end
+
+  ChangedAttachment = Struct.new(:attachment, :status)
+
+  def changed_attachments
+    created = attachments.where("created_at > ?", created_at + EDITION_CREATE_GRACE_PERIOD)
+
+    updated_attachments = attachments.where("updated_at > created_at")
+
+    # Using a 30 second grace period because GovspeakContent#render_govspeak! historically used to change the updated_at timestamp
+    # when running in an async worker to convert body to HTML. This meant that all HTML attachments ended up with
+    # (updated_at > created_at) so would show as 'changed'.
+    updated_html_attachments = html_attachments.joins(:govspeak_content).where("govspeak_contents.updated_at > DATE_ADD(govspeak_contents.created_at,  INTERVAL 30 SECOND)")
+    updated = (updated_attachments + updated_html_attachments) - created
+
+    created.map { |a| ChangedAttachment.new(a, :created) } +
+      updated.map { |a| ChangedAttachment.new(a, :updated) } +
+      deleted_attachments.map { |a| ChangedAttachment.new(a, :deleted) }
   end
 
   def uploaded_to_asset_manager?
--- a/app/views/admin/editions/_change_notes.html.erb
+++ b/app/views/admin/editions/_change_notes.html.erb
@@ -1,24 +1,41 @@
 <fieldset class="js-change-notes">
-  <h3 class="remove-top-margin">What sort of change are you making?</h3>
-  
-  <div class="radio">
-    <%= form.label :minor_change_true do %>
-      <%= form.radio_button(:minor_change, true) %>
-      <strong>Fixing a typo or broken link, a style change or similar</strong>
-      <p class="hint">Users signed up to email alerts won't be notified</p>
-    <% end %>
-  </div>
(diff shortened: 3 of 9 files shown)


app/views/admin/editions/_change_notes.html.erb
```diff
@@ -1,24 +1,47 @@
...
+        <% edition.updated_attachments.each do |attachment| %>
+          <li><%= "#{attachment.title} - #{attachment.readable_type} attachment" %> <span class="label label-warning">Updated</span></li>
+        <% end %>
+      <% end %>
+
+      <% if edition.has_deleted_attachments? %>
+        <% edition.deleted_attachments.each do |attachment| %>
+          <li><%= "#{attachment.title} - #{attachment.readable_type} attachment" %> <span class="label label-danger">Deleted</span></li>
+        <% end %>
+      <% end %>
```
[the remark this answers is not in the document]
ollietreend: For a more mild DRY-up, we should be able to drop these `if` statements because `#each`ing through an empty array will simply do nothing anyway.

```suggestion
      <% edition.created_attachments.each do |attachment| %>
        <li><%= "#{attachment.title} - #{attachment.readable_type} attachment" %> <span class="label label-success">Created</span></li>
      <% end %>
      <% edition.updated_attachments.each do |attachment| %>
        <li><%= "#{attachment.title} - #{attachment.readable_type} attachment" %> <span class="label label-warning">Updated</span></li>
      <% end %>
      <% edition.deleted_attachments.each do |attachment| %>
        <li><%= "#{attachment.title} - #{attachment.readable_type} attachment" %> <span class="label label-danger">Deleted</span></li>
      <% end %>
```

app/assets/javascripts/admin/views/editions/_form.js
```diff
@@ -31,7 +31,7 @@
 
       function appendHintToFormLabel () {
         var $label = $($changeNotesSection).find('label')
-        var exampleText = 'For example, "Edited chapter 6 - centres in Cardiff and Aberystwyth have closed."'
+        var exampleText = 'Tell users what has changed, where and why. Write in full sentences, leading with the most important words. For example, "College A has been removed from the registered sponsors list because its licence has been suspended." <br> Read the <a href="https://www.gov.uk/guidance/content-design/writing-for-gov-uk#change-notes" target="_blank">guidance about change notes</a> (opens in a new tab).'
```
ollietreend, line 34: This is as 'yuck' as you think it is! It doesn't feel right to be injecting hardcoded HTML into the page using JavaScript here.

I haven't delved too far down that 🐰🕳️ , but I'd hope there's a good reason why this is being done rather than it being rendered in the server-side HTML view.

If this is simply to work around the fact that the use of the `form.text_area` helper:

https://github.com/alphagov/whitehall/blob/963d8779decf00a8206aa7ffc403f08a2d5ba29b/app/views/admin/editions/_change_notes.html.erb#L38

doesn't support a `hint:` parameter, then I'd be tempted to manually add the the label and hint text rather than using the helper to create it. That might mean we could do away with the JS here, too.

Not sure if I've explained that very well – happy to talk it through.

ollietreend (commented):
A few comments! 🤓

app/models/attachable.rb
```diff
@@ -6,6 +6,7 @@ module Attachable
              -> { not_deleted.order("attachments.ordering, attachments.id") },
              as: :attachable,
              inverse_of: :attachable
+
```
ollietreend, line 9: Should we be implementing this behaviour in `Attachable`, or should it be implemented directly on `Edition`? 🤔

I don't know the answer, but want to ask the question because I'm not 100% sure it feels right in `Attachable`.

Other non-Edition models can use `Attachable` (such as [Response] and [PolicyGroup]), so we can't necessarily assume that we're operating on a model which follows the [Edition Workflow].

This is reflected in the fact that we've only implemented the UI side of this in the Edition views – because this functionality is specific to Edition change notes.

[Response]: https://github.com/alphagov/whitehall/blob/da1a57c95f323e5d25577b4de231589c09cab3e9/app/models/response.rb#L2
[PolicyGroup]: https://github.com/alphagov/whitehall/blob/86bbd48cfc106ff05960b407fa04f6d7bb234338/app/models/policy_group.rb#L3
[Edition Workflow]: https://github.com/alphagov/whitehall/blob/6a9dab86a24e1fdbed6d508dfbaff96469f19200/app/models/edition/workflow.rb

ollietreend (commented):
One more thing which I forgot to add earlier!

app/models/attachable.rb
```diff
@@ -60,6 +66,35 @@ def attachables
...
+  end
+
+  def updated_attachments
+    updated_attachments = attachments.select do |attachment|
+      attachment.updated_at > attachment.created_at ||
+        (attachment.respond_to?(:govspeak_content) &&
+        attachment.govspeak_content.updated_at > attachment.govspeak_content.created_at)
+    end
+    updated_attachments - created_attachments
+  end
```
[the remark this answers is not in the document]
ollietreend: Also: I feel like we should probably do a couple of things with that `5.seconds`:

1. Sense check to make sure this is a reasonable delay to expect when new Editions are created. We have some documents with _hundreds_ of attachments, so I'd be interested to see if this `5.seconds` value works there.
2. Document it somehow to explain what this 'magic value' represents, why it's needed, etc. to avoid any future WTFs. An inline comment might do the trick. It might even be worth naming it as a named constant – although that might be too much 🤷🏻

app/models/attachable.rb
beccapearce: Hmmm, I'm in two minds about this. I think attachable makes sense since there are editions that don't have attachments that don't need these methods. 

But totally agree with what you're saying that the edition.created_at stuff is going to mess up a bit. Not sure what the best place is either, happy to move it to edition, or any other suggestions? New file of some sort or is that adding to the complexity further?

ollietreend: I'm sort of on the fence, too. This is probably fine as it is, because Attachable does seem to have some implicit knowledge of Edition (e.g. the `Attachable::Null` class, which seems to emulate an Edition).

I'm tempted to just mark this as resolved and accept it as is.

app/models/attachable.rb
```diff
@@ -60,6 +67,21 @@ def attachables
...
+  def created_attachments
+    attachments.where("created_at > ?", created_at + CASE_CREATION_TIME)
+  end
+
+  def updated_attachments
+    updated_attachments = attachments.where("updated_at > ?", created_at) ||
+      attachments.join(:govspeak_content).where("govspeak_content.updated_at > ?", govspeak_content.created_at)
+
+    updated_attachments - created_attachments
+  end
```
ollietreend, line 83: I'm going to have a quick think about this one – I have a feeling there's going to be a nicer way to do it 🤔

In particular, I'm not sure the `||` is helpful here. What about if we have one `FileAttachment` which has changed name (`attachments.where("updated_at > ?", created_at)`) and also a `HtmlAttachment` which has some updated Govspeak content? I think we'll only get one OR the other right now.

app/models/attachable.rb
```diff
@@ -1,11 +1,13 @@
 module Attachable
   extend ActiveSupport::Concern
+  CASE_CREATION_TIME = 5.seconds
```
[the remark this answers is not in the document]
ollietreend: On second thoughts – it'd be great if we could avoid the 'time' issue altogether.

I'm going to keep thinking about the `created_attachments` and `updated_attachments` queries. It'd be lovely if there was another way to detect a new attachment which didn't exist on the last edition.

test/unit/attachable_test.rb
```diff
@@ -322,6 +322,91 @@ def build_edition_with_three_attachments
...
+    changed_title.update!(title: "This attachment's title has been changed")
+
+    # Change body of a HTML attachment
+    changed_html.govspeak_content.update!(body: "This HTML attachment's body has been changed")
+
+    # Delete an attachment
+    deleted_html.destroy!
+  
+    # Ensure there are no duplicates
+    changed_ids = publication.changed_attachments.select { |attachment| attachment.attachment.id }
```
ollietreend, line 397: Don't we want to use `map` here? 🤔 I don't think `select` is going to give us the right result.

```suggestion
    changed_ids = publication.changed_attachments.map { |attachment| attachment.attachment.id }
```

test/unit/attachable_test.rb
```diff
@@ -322,6 +322,91 @@ def build_edition_with_three_attachments
...
+    assert_equal changed_ids.uniq, changed_ids
+
+    # Check the changed attachments are correct
+    created_attachments = publication.changed_attachments.select { |attachment| attachment.status == :created }
+    updated_attachments = publication.changed_attachments.select { |attachment| attachment.status == :updated }
+    deleted_attachments = publication.changed_attachments.select { |attachment| attachment.status == :deleted }
```
ollietreend, line 403: I think you'll additionally want to use `map` here, so you get the actual `attachment` object.

```suggestion
    created_attachments = publication.changed_attachments.select { |attachment| attachment.status == :created }.map(&:attachment)
    updated_attachments = publication.changed_attachments.select { |attachment| attachment.status == :updated }.map(&:attachment)
    deleted_attachments = publication.changed_attachments.select { |attachment| attachment.status == :deleted }.map(&:attachment)
```

test/unit/whitehall/form_builder_test.rb
```diff
@@ -71,4 +71,12 @@ def removal_check_box(label_text = "Check to remove image")
...
     assert_dom_equal expected_html, @builder.upload(:image, label_text: "Image upload")
   end
+
+  test "Whitehall::FormBuilder#text_area can produce no label" do
+    expected_html = '<div class="form-group">' \
+                      '<textarea class="form-control" name="promotional_feature_item[summary]" id="promotional_feature_item_summary">
+</textarea></div>'
+
+    assert_dom_equal expected_html, @builder.text_area(:summary, label: false)
+  end
```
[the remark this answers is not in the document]
ollietreend: I don't remember where we landed on this in the end 😅 is it resolved?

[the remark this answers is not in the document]
beccapearce: It all displays correctly now so we can delete that nasty js method

ollietreend (approved):
Looks great ✨ nice work @beccapearce!

PR #6655, opened 2022-07-06
Tidy up stylesheet and JavaScript imports

injms:
## What

- Remove stylesheets already included in Static
- Add missing component stylesheets and JavaScript

Note - this is only for the base CSS file. The CSS file for HTML publications has been left alone for this pull request to avoid extra complication as the rendering of the HTML publications appears to be shared between `whitehall` and `government-frontend` - [for example](https://www.gov.uk/goernment/publications/coronavirus-covid-19-antibody-tests/coronavirus-covid-19-antibody-tests).

## Why

Removing the stylesheets means there should be no duplication of CSS between the stylesheet served by `whitehall` and the stylesheet served by `static`.

There were a couple of uses of components without the required stylesheet and JavaScript - this would lead to a broken component. Adding them back in should fix things.

## Visual changes

None.

---

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -6,6 +6,8 @@
 //
 //= require govuk_publishing_components/lib
 //= require govuk_publishing_components/components/button
+//= require govuk_publishing_components/components/details
+//= require govuk_publishing_components/components/metadata
 //= require govuk_publishing_components/components/govspeak
 //
 //= require jquery/dist/jquery
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -5,7 +5,6 @@
 // the compiled file.
 //
 //= require govuk_publishing_components/lib
-//= require govuk_publishing_components/components/button
 //= require govuk_publishing_components/components/details
 //= require govuk_publishing_components/components/metadata
 //= require govuk_publishing_components/components/govspeak
--- a/app/assets/stylesheets/frontend/base.scss
+++ b/app/assets/stylesheets/frontend/base.scss
@@ -9,18 +9,11 @@ $govuk-new-link-styles: true;
 
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/components/big-number";
-@import "govuk_publishing_components/components/button";
-@import "govuk_publishing_components/components/breadcrumbs";
 @import "govuk_publishing_components/components/contents-list";
 @import "govuk_publishing_components/components/document-list";
-@import "govuk_publishing_components/components/error-message";
-@import "govuk_publishing_components/components/feedback";
+@import "govuk_publishing_components/components/details";
 @import "govuk_publishing_components/components/govspeak";
-@import "govuk_publishing_components/components/heading";
-@import "govuk_publishing_components/components/hint";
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/frontend/base.scss
```diff
@@ -9,18 +9,11 @@ $govuk-new-link-styles: true;
 
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/components/big-number";
-@import "govuk_publishing_components/components/button";
```
andysellick, line 12: Button is also referenced in `html-publication.scss`.

app/assets/javascripts/application.js
```diff
@@ -6,6 +6,8 @@
 //
 //= require govuk_publishing_components/lib
 //= require govuk_publishing_components/components/button
```
andysellick, line 8: Looks like we're removing button from the CSS but leaving it in the JS. Button is used in this application but although the button CSS is included in static it doesn't look like the button JS is - which means this current configuration works, but is arguably incorrect - we should include button JS in static and remove it from here.

app/assets/stylesheets/frontend/base.scss
```diff
@@ -9,18 +9,11 @@ $govuk-new-link-styles: true;
...
-@import "govuk_publishing_components/components/breadcrumbs";
 @import "govuk_publishing_components/components/contents-list";
 @import "govuk_publishing_components/components/document-list";
-@import "govuk_publishing_components/components/error-message";
-@import "govuk_publishing_components/components/feedback";
+@import "govuk_publishing_components/components/details";
 @import "govuk_publishing_components/components/govspeak";
-@import "govuk_publishing_components/components/heading";
```
andysellick, line 19: Heading is also referenced in `html-publication.scss`. Actually a whole load of things are still referenced there so I'll stop pointing them out individually and let you look into it.

app/assets/stylesheets/frontend/base.scss
```diff
@@ -9,18 +9,11 @@ $govuk-new-link-styles: true;
 
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/components/big-number";
-@import "govuk_publishing_components/components/button";
-@import "govuk_publishing_components/components/breadcrumbs";
 @import "govuk_publishing_components/components/contents-list";
 @import "govuk_publishing_components/components/document-list";
-@import "govuk_publishing_components/components/error-message";
-@import "govuk_publishing_components/components/feedback";
```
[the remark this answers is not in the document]
injms: I've left `html-publications.scss` alone for the time being, as it started to get complicated as they appear to be published using `whitehall` and rendered by `government-frontend`. I've updated the pull request description to clarify.

app/assets/javascripts/application.js
injms: I've raised a pull request to add it to `static`: https://github.com/alphagov/static/pull/2820

app/assets/stylesheets/frontend/base.scss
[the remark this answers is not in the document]
andysellick: Hmm, interesting. Fair enough.

andysellick (approved):

PR #6663, opened 2022-07-08
Add script to generate codebase metrics

aldavidson:
For the last couple of months, we've been tracking some metrics about the codebase, to help us check progress on our mission to make it easier to work in. So far, we've retrieved the metrics manually. This is error-prone, and vulnerable to differences in method or OS shells on different dev machines, mis-recollections of what commands to use, etc.

To mitigate this, I've added this script that will execute as many of the metrics as possible, via cloning a temporary copy of the repo, and removing the `.git` directory.

By default it will use the `main` branch, but particular branches or revisions can be specified with the `REVISION` env var

Example for `main`:

```
~/govuk/whitehall$ ./script/codebase_metrics
(...)
Codebase metrics as of 2022-07-08T09:42:41+01:00, current revision f9967fbc91b32ae6765b6ad234f5aad55c619222
-----------------------------------------------------------------------------------------------------------

Total size:                 26728354 bytes
Total size (MB):            25.49MB
Number of items in Github:  3619
Number of models:           199
Number of views:            522
Number of controllers:      119
Number of presenters:       92
Number of routes:           440

You can retrieve "Test coverage %" and "Time to build (s)" from the latest
build against "main" in Jenkins, at:
  https:/(...removed...)/job/whitehall/job/main/lastSuccessfulBuild/console

Developer rating is an average score from 1 to 10 to the question:
"Overall, how difficult is it to make changes to Whitehall publisher?"
where 1 = terrible, 10 = very easy
It"s best asked of @publishing-experience-devs on Slack.

```

Example for named revision:

```
~/govuk/whitehall$ REVISION=40561aea88c40b5a53b3983b9d9f8fcb12065c20  ./script/codebase_metrics
(...)
Codebase metrics for revision 40561aea88c40b5a53b3983b9d9f8fcb12065c20
-----------------------------------------------------------------------------------------------------------

Total size:                 26806752 bytes
Total size (MB):            25.56MB
Number of items in Github:  3654
Number of models:           200
Number of views:            534
Number of controllers:      121
Number of presenters:       93
Number of routes:           N/A unless on main

To get "Number of routes", run this line in a Rails console against this revision:
  Rails.application.routes.routes.map { |r| {path: r.path.spec.to_s}}.uniq.count

  (Note, this gets less and less likely to work for older revisions, due to
  gem version locks & deprecations, etc)


To get "Test coverage %" and "Time to build (s)", run:
  git checkout 40561aea88c40b5a53b3983b9d9f8fcb12065c20
  git push -u origin (some unique branch name)

You can then retrieve those figures from the console output of that build in Jenkins, at:
  https://(...removed...)/job/whitehall/job/40561aea88c40b5a53b3983b9d9f8fcb12065c20/lastSuccessfulBuild/console

(...etc...)

```


brucebolt (commented):
The metrics retrieval looks fine to me, but I'm wondering if this might be better as a shell script, rather than a rake task since all of the methods perform system commands?

aldavidson:
> The metrics retrieval looks fine to me, but I'm wondering if this might be better as a shell script, rather than a rake task since all of the methods perform system commands?

Yeah, that makes sense. I originally put it as a Rake task because I wanted to instantiate the Rails environment to count the routes - but I couldn't find a way to get that working reliably against previous revisions, so took it out. Happy to port it to a shell script if you'd prefer that

aldavidson:
I've ported it to a bash script. This works fine on my Ubuntu 21.10 laptop, it will need testing on GDS-issue macbooks too

script/codebase_metrics
```diff
@@ -0,0 +1,115 @@
...
+  if [ $REVISION=="main" ]; then 
+    ROUTES=$(get_number_of_routes)
+  else
+    ROUTES="N/A unless on main"
+  fi
+}
+
+
+# main flow control here
+mkdir $REPO_DIR
```
brucebolt, line 110:
```suggestion
mkdir -p $REPO_DIR
```
This needs a `-p` option added to make the intermediate `/tmp/codebase-metrics` directory if it doesn't already exist.

brucebolt (approved):
This looks good to me.

I had to add an option to `mkdir` as I didn't have the intermediate directories already created.

Also, I couldn't run this directly on my Mac (as I didn't have the right Ruby version installed or any of the required gems).  Instead I ran it in govuk-docker.  It's probably worth adding the command somewhere (`govuk-docker-run script/codebase_metrics`).

aldavidson:
> This looks good to me.
> 
> I had to add an option to `mkdir` as I didn't have the intermediate directories already created.
> 
> Also, I couldn't run this directly on my Mac (...) Instead I ran it in govuk-docker. 

If it runs OK in the docker container, then I think that's probably enough to tick the "runs repeatably on other machines" box. 👍

PR #6667, opened 2022-07-11
Add limit to topical events displayed featured documents

callumknights:
This limit was originally chosen (as far as I'm aware) fairly arbitrarily several years ago. This commit basically mirrors the functionality that's currently implemented on the frontend in the backend. This will ensure that the limit is now set on publication so no more than 5 featured documents will show up once we fully migrate topical events to collections. It has the added side efect of reducing the size of the content item as the limit's being applied earlier.

Trello: https://trello.com/c/gBGGw6T3/147-front-end-tweaks-to-documents-section

### DIFF
--- a/app/presenters/publishing_api/topical_event_presenter.rb
+++ b/app/presenters/publishing_api/topical_event_presenter.rb
@@ -58,17 +58,21 @@ module PublishingApi
     end
 
     def ordered_featured_documents
-      item.classification_featurings.includes(:image, edition: :document).map do |feature|
-        {
-          title: feature.title,
-          href: feature.url,
-          image: {
-            url: feature.image.file.url(:s465),
-            alt_text: feature.alt_text,
-          },
-          summary: feature.summary,
-        }
-      end
+      item
+        .classification_featurings
+        .includes(:image, edition: :document)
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/topical_event_presenter.rb
```diff
@@ -41,7 +41,7 @@ def details
         details[:image] = image if item.logo_url
         details[:start_date] = item.start_date.rfc3339 if item.start_date
         details[:end_date] = item.end_date.rfc3339 if item.end_date
-        details[:ordered_featured_documents] = ordered_featured_documents
+        details[:ordered_featured_documents] = ordered_featured_documents.limit(5)
```
kevindew, line 44: Could we use the [`FeaturedLink::DEFAULT_SET_SIZE`](ihttps://github.com/alphagov/whitehall/blob/f9967fbc91b32ae6765b6ad234f5aad55c619222/app/models/organisation.rb#L504) constant for this? it'll make it less arbitrary.

callumknights: No problem 😄

PR #6671, opened 2022-07-12
Allow document withdrawals to be backdated

ollietreend:
### Context

This PR changes the way withdrawal dates are populated on the GOV.UK website:

[Example of a document withdrawal notice](https://www.gov.uk/government/news/wear-and-tear-repairs-for-m53-link-road)

This is [represented in the Content Item](https://www.gov.uk/api/content/government/news/wear-and-tear-repairs-for-m53-link-road) as:

```
"withdrawn_notice": {
  "explanation": "<div class=\"govspeak\"><p>This document is no longer current</p>\n</div>",
  "withdrawn_at": "2017-10-12T20:52:31Z"
}
```

### How it works currently (before this PR)

When an Edition is withdrawn in Whitehall, the details are kept in an `Unpublishing` record.

The `"withdrawn_notice"` JSON object is populated by:

| JSON key | Ruby value |
| --- | --- |
| `"explanation"` | `Unpublishing#explanation` |
| `"withdrawn_at"` | `Unpublishing#created_at` |

However, we now have a need to backdate the withdrawal date under particular circumstances. This means that the `"withdrawn_at"` field might need to be populated with a value which is different to the `Unpublishing#created_at` date.

### What this PR changes

This PR adds a new field to the `Unpublishing` model called `unpublished_at`.

This new field is now used to populate the `"withdrawn_notice"` JSON object:

| JSON key | Ruby value |
| --- | --- |
| `"explanation"` | `Unpublishing#explanation` |
| `"withdrawn_at"` | ~`Unpublishing#created_at`~ <br> **`Unpublishing#unpublished_at`** |

A [database migration](https://github.com/alphagov/whitehall/blob/add-unpublished-at-timestamp/db/migrate/20220711144922_add_unpublished_at_to_unpublishings.rb) adds this new field to the `unpublishings` table. Since all previous withdrawals used the `created_at` date, the migration backfills all `unpublished_at` fields to be the same as `created_at`.

For new `Unpublishing` records, the `unpublished_at` field is [automatically populated with the current time](https://github.com/alphagov/whitehall/blob/add-unpublished-at-timestamp/test/unit/models/unpublishing_test.rb#L4-L7) – which means it's functionally the same as the previous `created_at` timestamp. This change is therefore backwards compatible with the previous implementation.

However, it's now possible to [manually set the `unpublished_at` value](https://github.com/alphagov/whitehall/blob/add-unpublished-at-timestamp/test/unit/models/unpublishing_test.rb#L9-L13) if you want it to be something different. This can be used to backdate the withdrawal date that gets published on GOV.UK.

For now, nothing will be manually setting this new field. However, some proposed upcoming changes to the UI will make it possible for users to re-use an old withdrawal date if the document is being re-withdrawn (i.e. it was withdrawn, unwithdrawn, and is now being withdrawn again.)

---

Trello ticket: https://trello.com/c/lQVUQh0s/507-spike-reverting-withdrawn-date-when-a-document-if-its-not-the-first-withdrawal

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_05_26_090755) do
+ActiveRecord::Schema[7.0].define(version: 2022_07_11_144922) do
   create_table "about_pages", id: :integer, charset: "utf8mb3", collation: "utf8_unicode_ci", force: :cascade do |t|
     t.integer "topical_event_id"
     t.string "name"
@@ -416,7 +416,7 @@ ActiveRecord::Schema[7.0].define(version: 2022_05_26_090755) do
     t.boolean "political", default: false
     t.string "logo_url"
     t.boolean "read_consultation_principles", default: false
-    t.boolean "all_nation_applicability"
+    t.boolean "all_nation_applicability", default: true
     t.string "image_display_option"
     t.string "auth_bypass_id", null: false
     t.index ["alternative_format_provider_id"], name: "index_editions_on_alternative_format_provider_id"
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -416,7 +416,7 @@
     t.boolean "political", default: false
     t.string "logo_url"
     t.boolean "read_consultation_principles", default: false
-    t.boolean "all_nation_applicability"
+    t.boolean "all_nation_applicability", default: true
```
davidgisbey, line 419: It looks like some unrelated changes have snuck in?

davidgisbey:
left one comment, but other than that it LGTM!

davidgisbey (approved):
lgtm!

PR #6676, opened 2022-07-13
Fix topical event ordering

brucebolt:
There was a bug where new topical event (aka classification) features get incorrectly numbered when the list has previously been sorted. The new feature always gets numbered as `1`, meaning it is in the same position as another feature. The rendered ordering then varies (maybe somewhat randomly).

The problem was caused as the `order` method did not override the default scope, which orders the features in ascending order. Because of this, the last item was actually the first and was ordered 0.

Changing to the `maximum` method means Rails will ignore the default scope.

Additionally, there was an anomaly in ordering of topical event features.

If you add a new feature to a new topical event, the ordering starts from 1 and continues upwards. But if you reorder existing features on a topical event, the ordering starts from 0 and continues upwards.

Updating the behaviour so both situations start from 0.

The tests have been added in a separate commit so they deliberately fail and it would be possible to see why the ordering was incorrect.

### DIFF
--- a/app/models/classification.rb
+++ b/app/models/classification.rb
@@ -133,7 +133,7 @@ class Classification < ApplicationRecord
 
   def next_ordering
     last = classification_featurings.order("ordering desc").limit(1).last
-    last ? last.ordering + 1 : 1
+    last ? last.ordering + 1 : 0
   end
 
   def to_s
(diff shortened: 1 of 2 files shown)


BeckaL (approved):
Nice spot, code looks super neat and implementation makes a lot more sense now 👍

PR #6656, opened 2022-07-07
Switch Topical Event rendering to use Collections

brucebolt:
The rendering of Topical Events has been built into the Collections app, as part of the work to remove frontend rendering from Whitehall.

This will allow us to republish the content items for all topical events to switch the rendering to a different application, by running the following rake task: `publishing_api:republish:all_topical_events`.

**Not to be merged until we have confirmed the Collections rendered page is complete.**

[Trello card](https://trello.com/c/KcNZgjIA)

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6680, opened 2022-07-15
Remove topical event frontend

BeckaL:
Do not merge yet - dependent on:

* Final PRs for collections rendering going in (afaik just https://github.com/alphagov/collections/pull/2822 and https://github.com/alphagov/collections/pull/2853)
https://github.com/alphagov/whitehall/pull/6656 being made to topical events in whitehall
* Rake task being run to republish all topical events
* Double check to ensure that the rake task has been successful, i.e. all topical events in content store have their rendering app set to collections, not whitehall

This PR removes rendering of topical events from Whitehall, as we are moving these over to being rendered by collections.

[Trello](https://trello.com/c/almDgQWm/130-clean-up-whitehall-after-topical-event-move)


config/locales/de.yml
```diff
@@ -542,26 +542,6 @@ de:
...
     formats:
       long_ordinal: "%e %B %Y %l:%M%P"
-  topical_event:
-    headings:
-      detailed_guidance:
-      documents:
-      featured:
-      organisations:
-      travel_advice:
-    latest_activity: Jüngste Aktivität bei %{event}
```
BeckaL, line 552: We need to check that we are not missing this 'latest_activity' label anywhere in collections - was being used for atom feed here, but we haven't added this or any translations to collections atm.

app/views/topical_events/show.html.erb
```diff
@@ -161,49 +161,6 @@
                 </div>
               <% end %>
             </div>
-
-            <% if @topical_event.slug == 'first-world-war-centenary' %>
```
brucebolt, line 171: Could the commit message here be made slightly clearer, as someone coming back later might not realise what 'snowflake pages' are or what they are related to.

app/controllers/classifications_controller.rb
```diff
@@ -1,5 +0,0 @@
-class ClassificationsController < PublicFacingController
```
brucebolt, line 1: With this commit message, could it be made slightly clearer that it is the classification and topical event controllers/tests/routes that are being removed.

config/routes.rb
```diff
@@ -132,6 +132,7 @@ def external_redirect(path_prefix, target)
     get "/consultations/:consultation_id/outcome/:id" => "_#_", as: "consultation_outcome_html_attachment"
     get "/consultations/:consultation_id/public-feedback/:id" => "_#_", as: "consultation_public_feedback_html_attachment"
 
+    # TODO: can we remove?
```
brucebolt, line 135: Could this comment and the commit message be updated to explain why it is still there, e.g. for `/get-involved` we say:
```
# Controller removed. Whitehall frontend no longer serves these
# pages however the route is needed to generate path and urlgenerate path and url
# helper methods.
```

config/routes.rb
```diff
@@ -247,6 +248,7 @@ def external_redirect(path_prefix, target)
           end
           resources :offsite_links
         end
+        # TODO: can this be removed?
```
brucebolt, line 251: Ditto here regarding the comment

features/step_definitions/topical_event_steps.rb
```diff
@@ -16,122 +11,12 @@
...
-  expect(page).to have_selector(".govuk-caption-xl.gem-c-title__context", text: "Archived")
-end
-
 Then(/^I should see the topical event "([^"]*)" in the admin interface$/) do |topical_event_name|
   topical_event = TopicalEvent.find_by!(name: topical_event_name)
   visit admin_topical_events_path(topical_event)
   expect(page).to have_selector(record_css_selector(topical_event))
 end
 
-Then(/^I should see the topical event "([^"]*)" on the frontend$/) do |topical_event_name|
```
brucebolt, line 31: Could these step definition removals be squashed into the previous commit, where the steps are actually being deleted?

brucebolt (commented):
This looks great ⭐ .

We've just added a few small comments about making the commit messages clearer so the history is easier to follow and a comment about comments in the route file.

brucebolt (commented):
~I just stumbled upon another file which can be deleted.  There's a [JSON file](https://github.com/alphagov/whitehall/blob/main/lib/finders/topical_events.json) representing the old Whitehall rendered finder for topical events.~

Looks like this publishes the finder that is rendered by Finder Frontend, so we actually need to keep this file.

BeckaL:
Hey @brucebolt and @callumknights - thanks for your review! Was thinking that before merging this we'd just squash it all into a single commit, as I don't think the history is going to be useful for anyone looking at the codebase. What do you reckon? Was about to start rebasing commits but might be simpler to go the other way and put it all into one. Happy to do the rebasing though if you think it's worth having separate commits.

brucebolt:
Squashing into one commit sounds good to me.  There's a couple of comment in the routes file that could do with updating, but otherwise everything else looks good 👍.

brucebolt (approved):
Looks good to me, especially the reduction in lines of code by 2,127!

config/routes.rb
```diff
@@ -247,6 +250,9 @@ def external_redirect(path_prefix, target)
           end
           resources :offsite_links
         end
+        # Controller removed. Whitehall frontend no longer serves these
+        # pages however the route is needed to generate path and urlgenerate path and url
+        # helper methods.
```
brucebolt, line 255: These comments are in the `/admin` part of the routes file, so don't need to be here as the admin controller still exists.

BeckaL: Good point, have removed

PR #6681, opened 2022-07-19
Consolidate `Classification` into `TopicalEvent`

brucebolt:
The `Classification` model is currently only inherited by `TopicalEvent`.  Previously, this was shared between `Topic` (known in the UI as 'Policy areas') and `TopicalEvent`, but [the former has previously been removed](https://github.com/alphagov/whitehall/pull/5666).

This split caused confusion when doing work on Topical Events recently and made the code more difficult to work with.

Within this PR, we consolidate `Classification` into `TopicalEvent` and update the associated code.

[Trello card](https://trello.com/c/r2f4CMfF)


app/controllers/admin/topical_event_featurings_controller.rb
```diff
@@ -0,0 +1,100 @@
+class Admin::TopicalEventFeaturingsController < Admin::BaseController
+  before_action :load_topical_event
```
BeckaL, line 2: So for organisations and world location news, they use 'Feature Lists' and a 'Feature Lists controller', by using `include Featureable`. So just wondering whether we could rationalise this even further by removing this altogether and using the same - I don't think there's a particular reason why classifications should have their own version of this, but maybe I'm missing something. But seems a bit fishy to me.

db/migrate/20220726111449_drop_classification_policies_table.rb
```diff
@@ -0,0 +1,5 @@
+class DropClassificationPoliciesTable < ActiveRecord::Migration[7.0]
```
JonathanHallam, line 1: Super minor but there's a link missing in the commit message for this one

app/controllers/admin/topical_event_featurings_controller.rb
BeckaL: Maybe we could do this as a separate PR to this but I think it'd be worth exploring to simplify things, as it's another bit of weirdness in Whitehall that we've encountered

JonathanHallam: This is a really good idea for a spike or mob. Excellent spot

JonathanHallam (approved):
Am happy with the code changes, this is a really gnarly bit of work, massive kudos!

brucebolt:
I've cherry picked the `classification_policies` and `classification_relations` table dropping into a separate PR: https://github.com/alphagov/whitehall/pull/6711

JonathanHallam (approved):
Nice one, thanks for this, super bit of work. Happy now that the more destructive bits of this have been moved elsewhere

PR #6687, opened 2022-07-21
Configure Pact to use the test database

ollietreend:
Prior to this, Pact tests would run against the development database by default. This had the unfortunate side effect of truncating tables and populating them with test data.

Whilst that wasn't noticable in the CI environment, it did mean that development databases were cleared when running Pact tests on local development environments (e.g. govuk-docker). This behaviour resulted in confusion as test suites are usually expected to run against the test database.

It seems pretty common for `ENV['RAILS_ENV'] = "test"` to be at the beginning of test helper files – for example, that's what `test/test_helper.rb` does. So I've taken the same approach here.

### DIFF
--- a/spec/service_consumers/pact_helper.rb
+++ b/spec/service_consumers/pact_helper.rb
@@ -1,3 +1,5 @@
+ENV["RAILS_ENV"] = "test"
+
 require "pact/provider/rspec"
 require "webmock/rspec"
 require "factory_bot"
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Looks good. While you here you might want to add the do not track env var too. Then it'll match Publishing API: https://github.com/alphagov/publishing-api/blob/5e79d9c01a58a9677bb6e5559f202ed494212a1c/spec/service_consumers/pact_helper.rb#L1-L2

PR #6688, opened 2022-07-21
Allow publishers to reuse previous withdrawal notices

ollietreend:
This PR adds functionality for users who need to reuse a previous document withdrawal.

### First withdrawal of a document

The first time a document is withdrawn, publishers will be asked to provide a public explanation for the withdrawal. This preserves the existing UI, where the 'public explanation' field appears once the user has selected the 'Withdraw' radio.

First withdrawal of a document

### Subsequent withdrawals

If a document has been withdrawn, and subsequently un-withdrawn, this is when the UI changes. The next time a publisher chooses to withdraw the document, they will be presented with the dates and public explanations of previous withdrawals on the document.

The publisher must choose to either re-use a previous withdrawal date or withdraw using today's date. If they choose a previous date, the withdrawal notice will be backdated to show the previous date and public explanation. Alternatively, choosing today's date will make the 'public explanation' field appear for user input.

#### Using a previous withdrawal date

Withdraw using a previous withdrawal date

#### Using today's date

Withdrawing using today's date

## Effect on frontend

This affects the date and explanation shown to end users on the GOV.UK frontend.

Withdrawal notice on GOV.UK

## Code changes

Code changes for this feature are relatively self contained.

### Edition Workflow controller

Handles form input – including the new radios which are used to select between creating a new withdrawal vs re-using a previous withdrawal. A validation error will be shown if no radio is selected.

The controller sets `@previous_withdrawals` for use by the view. This is a list of previous withdrawals that have been performed on the document, but with an arbitrary limit of 50. This limit is purely there to ensure it's not a limitless query which could cause page load performance issues on documents with extreme numbers of withdrawals.

### Edition Withdrawer service

I've adapted the `EditionWithdrawer` service to accept an additional parameter `previous_withdrawal`. This parameter is **optional**. When provided, it should be the previous `Unpublishing` object that is to be re-used.

### Document model

I've added `Document#withdrawals` as a new association on the Document model.

This makes it possible to retrieve previous withdrawals on Editions associated with the current Document.

---

Trello: https://trello.com/c/lQVUQh0s

### DIFF
--- a/app/assets/javascripts/admin/views/edition_workflow/confirm_unpublish.js
+++ b/app/assets/javascripts/admin/views/edition_workflow/confirm_unpublish.js
@@ -8,9 +8,11 @@
 
       this.revealCorrectForm()
       this.hideExplanationIfRedirecting()
+      this.revealNewWithdrawalFields()
 
       $("input[name='unpublishing_reason_id']").change(this.revealCorrectForm)
       $('#unpublishing_redirect').change(this.hideExplanationIfRedirecting)
+      $("input[name='previous_withdrawal_id']").change(this.revealNewWithdrawalFields)
     },
 
     revealCorrectForm: function () {
@@ -39,6 +41,14 @@
         $('#published_in_error_explanation').val('').closest('fieldset').hide()
       } else {
         $('#published_in_error_explanation').closest('fieldset').show()
+      }
+    },
+
+    revealNewWithdrawalFields: function () {
+      var withdrawalRadios = $("input[name='previous_withdrawal_id']")
+      if (withdrawalRadios.length > 0) {
+        var selected = withdrawalRadios.filter(':checked').val()
+        $('#new-withdrawal').toggle(selected === 'new')
       }
     }
   }
--- a/app/controllers/admin/edition_workflow_controller.rb
+++ b/app/controllers/admin/edition_workflow_controller.rb
@@ -12,6 +12,7 @@ class Admin::EditionWorkflowController < Admin::BaseController
   before_action :set_minor_change_flag
   before_action :ensure_reason_given_for_force_publishing, only: :force_publish
   before_action { check_if_locked_document(edition: @edition) }
+  before_action :set_previous_withdrawals, only: %i[confirm_unpublish unpublish]
 
   rescue_from ActiveRecord::StaleObjectError do
     redirect_to admin_edition_path(@edition), alert: "This document has been edited since you viewed it; you are now viewing the latest version"
@@ -98,17 +99,17 @@ class Admin::EditionWorkflowController < Admin::BaseController
   end
 
   def confirm_unpublish
-    @unpublishing = @edition.build_unpublishing(unpublishing_reason_id: UnpublishingReason::Withdrawn.id)
+    @unpublishing = @edition.build_unpublishing
   end
 
   def unpublish
-    @service_object = withdrawer_or_unpublisher_for_edition
+    success, message = withdrawing? ? withdraw_edition : unpublish_edition
 
-    if @service_object.perform!
-      redirect_to admin_edition_path(@edition), notice: unpublish_success_notice
+    if success
+      redirect_to admin_edition_path(@edition), notice: message
     else
-      @unpublishing = @edition.unpublishing
-      flash.now[:alert] = @service_object.failure_reason
+      @unpublishing = @edition.unpublishing || @edition.build_unpublishing(unpublishing_params)
+      flash.now[:alert] = message
       render :confirm_unpublish
     end
   end
@@ -181,12 +182,52 @@ private
     end
   end
 
-  def withdrawer_or_unpublisher_for_edition
-    if withdrawing?
-      Whitehall.edition_services.withdrawer(@edition, user: current_user, remark: "Withdrawn", unpublishing: unpublishing_params)
+  def set_previous_withdrawals
+    # The limit here is entirely arbitrary with the purpose of avoiding page-load performance issues on
+    # documents with lots of withdrawals. This scenario is unlikely, but a limit is better than a timeout.
+    @previous_withdrawals = @edition.document.withdrawals.last(50)
+  end
+
+  def withdraw_edition
+    if new_withdrawal? || previous_withdrawal.present?
+      withdrawer = Whitehall.edition_services.withdrawer(
+        @edition,
+        user: current_user,
+        remark: "Withdrawn",
+        unpublishing: unpublishing_params,
+        previous_withdrawal: previous_withdrawal,
+      )
+
+      success = withdrawer.perform!
+      message = if success
+                  "This document has been marked as withdrawn"
+                else
+                  withdrawer.failure_reason
+                end
     else
-      Whitehall.edition_services.unpublisher(@edition, user: current_user, remark: "Reset to draft", unpublishing: unpublishing_params)
+      success = false
+      message = "Select which withdrawal date you want to use"
     end
+
+    [success, message]
+  end
+
+  def unpublish_edition
+    unpublisher = Whitehall.edition_services.unpublisher(
+      @edition,
+      user: current_user,
+      remark: "Reset to draft",
+      unpublishing: unpublishing_params,
+    )
+
+    success = unpublisher.perform!
+    message = if success
+                "This document has been unpublished and will no longer appear on the public website"
+              else
+                unpublisher.failure_reason
+              end
+
+    [success, message]
   end
 
   def unpublishing_params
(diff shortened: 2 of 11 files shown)


test/unit/document_test.rb
```diff
@@ -267,4 +267,31 @@ class DocumentTest < ActiveSupport::TestCase
...
   end
+
+  test "#withdrawals returns withdrawals that have happened on editions of the document" do
+    # Create & publish a document
+    document = create(:document)
+    edition = create(:edition, document: document)
+    user = edition.creator
+    EditionForcePublisher.new(edition).perform!
+
+    3.times do
```
davidgisbey, line 278: Not necessarily a blocker, but this is quite an expensive unit test. Can we just create the objects in the desired state?

ollietreend: Yeah – fair point! I wasn't sure which approach was best.

I'll take a look at both approaches and see what the performance impact is.

ollietreend: So, in terms of performance, the test takes:

- **~1.5 seconds** to run when using the `EditionForcePublisher`, `EditionWithdrawer` and `EditionUnwithdrawer` services
- **~0.2 seconds** to run when creating objects directly

Quite a difference! 😅

I think your point about the test feeling more integration(ish) is completely valid. My initial concern was that just creating objects directly doesn't realistically reproduce changes that'd happen in related models (e.g. older editions becoming "superseded" rather than remaining "withdrawn"). So I wanted to setup a more realistic state.

But you're absolutely right. Since this is a unit test, the other side effects aren't really as important. We can setup everything this method needs without the overhead of using those higher level services.

ollietreend: I've just pushed up that change. Thanks for suggesting it @davidgisbey 🤓

baisa (approved):
LGTM 👍

app/controllers/admin/edition_workflow_controller.rb
```diff
@@ -98,16 +98,20 @@ def force_publish
   end
 
   def confirm_unpublish
-    @unpublishing = @edition.build_unpublishing(unpublishing_reason_id: UnpublishingReason::Withdrawn.id)
+    @unpublishing = @edition.build_unpublishing
+    @previous_withdrawals = @edition.document.withdrawals
```
kevindew, line 102: I'm a bit worried that this is boundless. Do we think it's ok for someone to flip back multiple withdrawals?

As a way to try not make this to big an issue I'd perhaps suggest a compromise of a limit of 5. I can't really imagine why anyone would need more than 1 or 2 or why there'd be a lot - but we've seen the dangers of limitless queries.

Also should it have an ordering? I'm guessing it's relying on implicit ordering at the mo

app/services/edition_withdrawer.rb
```diff
@@ -16,6 +33,25 @@ def past_participle
...
+    @edition.unpublishing.explanation = previous_withdrawal.explanation
+    @edition.unpublishing.unpublished_at = previous_withdrawal.unpublished_at
+  end
+
+  def creating_new_record?
+    @options[:unpublishing].present?
+  end
+
+  def use_previous_withdrawal?
+    @use_previous_withdrawal.present? && @use_previous_withdrawal != "new"
```
kevindew, line 46: Is this `use_previous_withdrawal != "new"` part of user input? if so I wonder if we want to deal with that at the controller level and just have a boolean in. I had assumed thus far that we were considering @use_previous_withdrawal as a boolean rather than a string or an id. Might be best to separate them into just a previous_withdrawal_id that is an integer or nil, rather than having user input go deeper into the classes.

app/services/edition_withdrawer.rb
```diff
@@ -16,6 +33,25 @@ def past_participle
...
+  end
+
+  def previous_withdrawal
+    if use_previous_withdrawal?
+      @previous_withdrawal ||= @edition.document.withdrawals.find_by(id: @use_previous_withdrawal)
+    end
```
kevindew, line 52: Nit, might be cleaner as

```
return unless use_previous_withdrawal?

@previous_withdrawal ||= @edition.document.withdrawals.find_by(id: @use_previous_withdrawal)
```

app/views/admin/edition_workflow/confirm_unpublish.html.erb
```diff
@@ -46,10 +46,43 @@
...
+
+        <div class="radio">
+          <%= label_tag "use_previous_withdrawal_new" do %>
+            <% checked = (@use_previous_withdrawal == 'new') %>
+            <%= radio_button_tag 'use_previous_withdrawal', "new", checked, id: "use_previous_withdrawal_new" %>
+            <strong>This is a new withdrawal</strong>
+          <% end %>
+        </div>
+      <% else %>
+        <%= hidden_field_tag 'use_previous_withdrawal', 'new' %>
```
kevindew, line 77: Do we actually need to pass this? I wonder if we can infer that the lack of this parameter (or it being empty) is a sign of a new withdrawal.

test/unit/services/edition_withdrawer_test.rb
```diff
@@ -90,9 +90,62 @@ class EditionWithdrawerTest < ActiveSupport::TestCase
...
+
+    invalid_values.each do |use_previous_withdrawal|
+      unpublisher = EditionWithdrawer.new(edition,
+                                          use_previous_withdrawal: use_previous_withdrawal,
+                                          unpublishing: unpublishing_params)
+
+      assert_not unpublisher.can_perform?
+      assert_equal "That withdrawal could not be found. Select one from the list.", unpublisher.failure_reason
+    end
+  end
```
kevindew, line 144: These look good 👍

features/unpublishing-published-documents.feature
```diff
@@ -1,43 +1,40 @@
 Feature: Unpublishing published documents
-  As a GDS Editor
+  As a managing editor
   I want to be able to revert published documents back to the draft state
   So that I can remove documents from the site that were published in error
 
+  Background:
+    Given I am a managing editor
```
kevindew, line 7: Nice 👍

kevindew (commented):
Looks great - just a few comments on unbounded queries and user input parameters.

app/views/admin/edition_workflow/confirm_unpublish.html.erb
ollietreend: The presence of the parameter, but it being empty, would indicate that the user was shown radio inputs but _didn't choose one_ – i.e. it's a validation error.

I suppose we could infer it server side though by checking for `@previous_withdrawals.any?` – in other words, just repeat the same conditional check that causes this to appear in the view.

app/assets/javascripts/admin/views/edition_workflow/confirm_unpublish.js
```diff
@@ -8,9 +8,11 @@
 
       this.revealCorrectForm()
       this.hideExplanationIfRedirecting()
+      this.revealNewWithdrawalFields()
 
       $("input[name='unpublishing_reason_id']").change(this.revealCorrectForm)
       $('#unpublishing_redirect').change(this.hideExplanationIfRedirecting)
+      $("input[type=radio][name='use_previous_withdrawal']").change(this.revealNewWithdrawalFields)
```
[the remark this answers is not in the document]
ollietreend: Not necessarily on this line. However we do need it in the `revealNewWithdrawalFields` function:

https://github.com/alphagov/whitehall/blob/65deb36c7055170eae02be1c59d7dc3c77c01b5e/app/assets/javascripts/admin/views/edition_workflow/confirm_unpublish.js#L48-L52

Because we go on to `.filter(':checked')` to find the checked radio.

However, if the Document does not have previous withdrawals, the input with `name="use_previous_withdrawal"` is actually `type="hidden"`. Although I note that you've commented on that too – so may be able to ditch both (see other comment thread).

app/controllers/admin/edition_workflow_controller.rb
ollietreend: Having looked across Whitehall, the highest number of previous withdrawals on a document currently is 4 or 5 (can't remember off the top of my head) so I'm not too worried about this scenario.

Originally we'd decided just to go for _the most recent_ previous withdrawal (singular) but realised that this would be problematic if publishers accidentally withdraw as 'today' and then need to undo it and reuse an older withdrawal instead. So we allowed for multiple to support that use case.

kevindew: Sure, I think if we're not worried about the scenario just stick an arbitrary limit on it to make it not boundless. Then we can hear if users have a problem where they expected more choice rather than risk a crash one day.

I think you missed the question about ordering btw.

app/views/admin/edition_workflow/confirm_unpublish.html.erb
kevindew: Sorry I should have been a bit more verbose. I wondered about losing this one and having the one that is currently 'new' above be blank.

I hadn't clocked too much about the validation error though. However could validate that it's not nil rather than blank. 

Using server side check seems better for the validation given there's not anything a user can do with the validation error if they have no withdrawals to choose from.

PR #6695, opened 2022-07-26
Change link on world locations admin panel to point to live site

BeckaL:
Currently, the 'view on website' link on the world locations admin panel, features and translation tabs point to the whitehall admin link for the content (e.g. https://whitehall-admin.integration.publishing.service.gov.uk/world/some-news-url/news instead of https://www.integration.publishing.service.gov.uk/world/some-news-url/news). This will no longer work when we remove rendering of world location news from Whitehall, so this link is updated to point to the live site.

[This](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/world_locations/france) is an example page where you can currently see the link to whitehall admin (including on the features and translations tab)
[Trello](https://trello.com/c/Qet0mRxb/167-fix-link-from-whitehall-to-world-location-news-on-live-site) 

### DIFF
--- a/test/functional/admin/world_location_translations_controller_test.rb
+++ b/test/functional/admin/world_location_translations_controller_test.rb
@@ -76,6 +76,26 @@ class Admin::WorldLocationTranslationsControllerTest < ActionController::TestCas
     end
   end
 
+  view_test "the 'View on website' link on the show page goes to the news page" do
+    location = create(:world_location, translated_into: [:fr], name: "France")
+    get :index, params: { world_location_id: location }
+
+    assert_select "a" do |links|
+      view_links = links.select { |link| link.text =~ /View on website/ }
+      assert_match(/#{Regexp.escape("https://www.test.gov.uk/world/france/news")}/, view_links.first["href"])
+    end
+  end
+
+  view_test "the view buttons for translations link to the new page on the live site" do
+    location = create(:world_location, translated_into: [:fr], name: "France")
+    get :index, params: { world_location_id: location }
+
+    assert_select "a" do |links|
+      view_links = links.select { |link| link.text =~ /view/ }
+      assert_match(/#{Regexp.escape("https://www.test.gov.uk/world/france/news.fr")}/, view_links.first["href"])
+    end
+  end
+
   view_test "update updates translation and redirects back to the index" do
     put :update,
         params: { world_location_id: @location,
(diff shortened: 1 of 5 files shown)


test/functional/admin/world_location_translations_controller_test.rb
```diff
@@ -76,6 +76,26 @@ class Admin::WorldLocationTranslationsControllerTest < ActionController::TestCas
...
+    end
+  end
+
+  view_test "the view buttons for translations link to the new page on the live site" do
+    location = create(:world_location, translated_into: [:fr], name: "France")
+    get :index, params: { world_location_id: location }
+
+    assert_select "a" do |links|
+      view_links = links.select { |link| link.text =~ /view/ }
+      assert_match(/https:\/\/www\.test\.gov\.uk\/world\/france\/news\.fr/, view_links.first["href"])
```
brucebolt, line 95: I think you can use `assert_equal` here (as we're matching the entire string, rather than just part of it) which will mean the escaping won't be needed.
```suggestion
      assert_equal("https://www.test.gov.uk/world/france/news.fr", view_links.first["href"])
```

brucebolt (commented):
Looks good, especially the new tests.

I think we can use a different matcher in the tests, which will mean the URLs won't need to be escaped.

test/functional/admin/world_location_translations_controller_test.rb
BeckaL: Thanks for review @brucebolt! So the problem the match gets around is that there is a random cachebust string added as a query param. Can you think of a good way of just matching the beginning of a string? I agree that being able to specify the string as per yours is nice and readable.

BeckaL: What do you think of this instead?

`assert_match(/#{Regexp.escape("https://www.test.gov.uk/world/france/news.fr")}/, view_links.first["href"])`

BeckaL: I've updated them to the above ^^ as it's the neatest way I can think of still using the string, but let me know if you manage to come up with a nicer way

brucebolt: Oh, I forgot about the cachebust string.

This looks like a nice alternative.

brucebolt (approved):

PR #6696, opened 2022-07-26
Persist latest_edition_id and published_edition_id on document

davidgisbey:
## Document 

https://docs.google.com/document/d/1WQ5u0JmFHYu1Hv0TX9YGEfPLWyhMZblVvHkSlObxMJM/edit

## Trello card 

https://trello.com/c/iWcORsD5/549-investigate-if-we-can-get-the-latest-edition-of-a-document-more-efficiently

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -14,7 +14,7 @@ class Document < ApplicationRecord
   has_many :edition_relations, dependent: :destroy, inverse_of: :document
 
   has_one  :published_edition,
-           -> { where(state: Edition::PUBLICLY_VISIBLE_STATES) },
+           -> { joins(:document).where("editions.id = documents.live_edition_id") },
            class_name: "Edition",
            inverse_of: :document
   has_one  :pre_publication_edition,
@@ -23,14 +23,7 @@ class Document < ApplicationRecord
            inverse_of: :document
 
   has_one  :latest_edition,
-           lambda {
-             where(%(
-               NOT EXISTS (
-               SELECT 1 FROM editions e2
-               WHERE e2.document_id = editions.document_id
-               AND e2.id > editions.id
-               AND e2.state <> 'deleted')))
-           },
+           -> { joins(:document).where("editions.id = documents.latest_edition_id") },
            class_name: "Edition",
            inverse_of: :document
 
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -13,7 +13,7 @@ class Document < ApplicationRecord
   has_many :editions, inverse_of: :document
   has_many :edition_relations, dependent: :destroy, inverse_of: :document
 
-  has_one  :published_edition,
+  has_one  :live_edition,
            -> { joins(:document).where("editions.id = documents.live_edition_id") },
            class_name: "Edition",
            inverse_of: :document
@@ -44,7 +44,7 @@ class Document < ApplicationRecord
   attr_accessor :sluggable_string
 
   def self.published
-    joins(:published_edition)
+    joins(:editions).where(editions: { state: "published" })
   end
 
   def self.at_slug(document_types, slug)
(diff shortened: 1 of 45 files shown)


kevindew:
Hey @davidgisbey great that you're working on this. Before you do too much implementation that is hard to come back from it'd be great to have a chat about the DB architecture behind it. There's various ways we can skin this cat with various pros and cons.

/cc @ollietreend

davidgisbey:
Sure there's some context on this card https://trello.com/c/iWcORsD5/549-investigate-if-we-can-get-the-latest-edition-of-a-document-more-efficiently

test/factories/editions.rb
```diff
@@ -10,6 +10,12 @@
     previously_published { false }
     auth_bypass_id { SecureRandom.uuid }
 
+    after :create do |edition|
+      document = edition.document
+      document.update!(published_edition_id: edition.id) if edition.state.in?(Edition::PUBLICLY_VISIBLE_STATES)
+      document.update!(latest_edition_id: edition.id) if !edition.deleted? && (document.latest_edition_id.blank? || edition.id == document.editions.where(state: [Edition::PRE_PUBLICATION_STATES]).order(id: :desc).first&.id)
```
davidgisbey, line 16: There's some complexity here around making sure the correct values are set for this when objects are created in the factory. This is just me hacking around and should go into it's own service where it's unit tests

```
SetPublishedAndLatestEditionIDs.new(edition).call 
```

or something similar

app/models/edition.rb
```diff
@@ -244,21 +244,11 @@ def self.without_locked_documents
...
   end
 
   def self.latest_published_edition
-    published.where("NOT EXISTS (
-      SELECT 1
-        FROM editions e2
-       WHERE e2.document_id = editions.document_id
-         AND e2.id > editions.id
-         AND e2.state = 'published')")
+    joins(:document).where("editions.id = documents.published_edition_id")
```
davidgisbey, line 251: The behaviour here has subtly changed. PReviously it would now have passed back an edition that is withdrawn, only published. However, i've been through the code and this class method isn't actually being used anywhere. I've left it in as it might be useful for scoping etc. when querying the db.

It could potentially be pulled out though.

app/models/document.rb
```diff
@@ -13,7 +13,7 @@ class Document < ApplicationRecord
   has_many :editions, inverse_of: :document
   has_many :edition_relations, dependent: :destroy, inverse_of: :document
 
-  has_one  :published_edition,
+  has_one  :live_edition,
```
davidgisbey, line 16: If we decide we want to rename this (i think we should), then this work can be pulled out and completed in a separate PR, prior to adding the new columns to the documents table.

app/models/edition.rb
```diff
@@ -247,8 +247,8 @@ def self.latest_edition
     joins(:document).where("editions.id = documents.latest_edition_id")
   end
 
-  def self.latest_published_edition
-    joins(:document).where("editions.id = documents.live_edition_id AND state = 'published'")
+  def self.latest_live_edition
```
davidgisbey, line 250: So if we want to retain the current behaviour the correct thing to do here is update this to:

```
joins(:document).where("editions.id = documents.latest_edition_id and state = 'published'")
```

Since all our production code is scoped to a single document there might even be an argument for removing this?

app/models/document.rb
ollietreend: I'm going to pull this out into a separate PR, as you said. It seems like a reasonable change.

Annoyingly I can't just cherry-pick this commit into a new branch because it follows all the other changes 😢 

So just wrapping my head around it so I can figure out how to re-create it from `main`.

app/models/document.rb
```diff
@@ -44,7 +44,7 @@ class Document < ApplicationRecord
   attr_accessor :sluggable_string
 
   def self.published
-    joins(:published_edition)
+    joins(:editions).where(editions: { state: "published" })
```
ollietreend, line 47: I think this changes the behaviour. The previous `joins(:published_edition)` included both "published" and "withdrawn", where it's now limited just to "published" editions.

So I'm thinking we should probably rename this method to `self.live` and maintain the existing behaviour. Wdyt?

davidgisbey:
Closing this as it proved the concept and works been completed in other PRs

PR #6699, opened 2022-07-26
Turn off font

jon-kirwan:
## What

Prevent the font from being included in `whitehall`'s stylesheet.

## Why

We should be serving the font files from `static` so they're cached across the entirety of GOVUK - so the font shouldn't be included in `whitehall`'s stylesheet.

With the font being served by each application, users need to re-download it every time they went from a page rendered in one application to a page rendered in another application.

Serving it from `static` means that the same font URI is used in each application, allowing the fonts to be cached and reused.

## Visual changes

None - the same font is being served from `static`, instead of `whitehall`.

Before (the font paths contains `whitehall`):

After (the font path contains `static`):

---

## Search page examples to sanity check:

- TBC


injms (approved):
🚀 👍

PR #6701, opened 2022-07-27
Add info to world location news presenter

BeckaL:
This PR adds the following fields to the details presented to publishing-api for world location news items:

* featured links
* featured documents
* mission statement

and switches to using the newly created schema for world location news, since world location news was previously using a placeholder schema. 

Other changes include:

* Renaming the world location news presenter for consistency
* Moving some organisations presenter code into a shared helper since there is similar functionality across the organisations and world_location_news content types.

Depends on [the sister content schemas change](https://github.com/alphagov/govuk-content-schemas/pull/1110) - CI reported below as passing has been run against that branch. 

[Trello](https://trello.com/c/FUNRgTWI/168-get-world-news-content-into-content-item)

### DIFF
--- a/app/presenters/publishing_api/featured_documents_presenter.rb
+++ b/app/presenters/publishing_api/featured_documents_presenter.rb
@@ -1,6 +1,6 @@
 module PublishingApi
   module FeaturedDocumentsPresenter
-    def featured_documents(featurable_item, document_limit = 6)
+    def featured_documents(featurable_item, document_limit)
       featurable_item.feature_list_for_locale(I18n.locale).current.limit(document_limit).map do |feature|
         if feature.document
           featured_documents_editioned(feature)
--- a/app/presenters/publishing_api/featured_documents_presenter.rb
+++ b/app/presenters/publishing_api/featured_documents_presenter.rb
@@ -1,0 +1,65 @@
+module PublishingApi
+  module FeaturedDocumentsPresenter
+    def featured_documents(featurable_item, document_limit = 6)
+      featurable_item.feature_list_for_locale(I18n.locale).current.limit(document_limit).map do |feature|
+        if feature.document
+          featured_documents_editioned(feature)
+        elsif feature.topical_event
+          featured_documents_topical_event(feature)
+        elsif feature.offsite_link
+          featured_documents_offsite_link(feature)
+        end
+      end
+    end
+
+  private
+
+    def featured_documents_editioned(feature)
+      # Editioned formats (like news) that have been featured
+      edition = feature.document.published_edition
+      {
+        title: edition.title,
+        href: Whitehall.url_maker.public_document_path(edition),
+        image: {
+          url: feature.image.url,
+          alt_text: feature.alt_text,
+        },
+        summary: Whitehall::GovspeakRenderer.new.govspeak_to_html(edition.summary),
+        public_updated_at: edition.public_timestamp,
+        document_type: edition.display_type,
+      }
+    end
+
+    def featured_documents_topical_event(feature)
+      # Topical events that have been featured
+      topical_event = feature.topical_event
+      {
+        title: topical_event.name,
+        href: Whitehall.url_maker.polymorphic_path(topical_event),
+        image: {
+          url: feature.image.url,
+          alt_text: feature.alt_text,
+        },
+        summary: Whitehall::GovspeakRenderer.new.govspeak_to_html(topical_event.description),
+        public_updated_at: topical_event.start_date,
+        document_type: nil, # We don't want a type for topical events
+      }
+    end
+
+    def featured_documents_offsite_link(feature)
+      # Offsite links that have been featured
+      offsite_link = feature.offsite_link
+      {
+        title: offsite_link.title,
+        href: offsite_link.url,
+        image: {
+          url: feature.image.url,
+          alt_text: feature.alt_text,
+        },
+        summary: Whitehall::GovspeakRenderer.new.govspeak_to_html(offsite_link.summary),
+        public_updated_at: offsite_link.date,
+        document_type: offsite_link.display_type,
+      }
+    end
+  end
+end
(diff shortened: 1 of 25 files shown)


test/unit/presenters/publishing_api/world_location_news_page_presenter_test.rb
```diff
@@ -30,6 +30,7 @@ def world_location
             href: newer_link.url,
           },
         ],
+        mission_statement: "This is a great mission", #TODO I don't think this ever actually can be nil in practice (although it's nillable in db schema. How to handle)
```
BeckaL: I will remove this comment as decided to handle nil cases. I could only find two in the database, and I think they are old. Ideally it'd be nice for schemas to have fewer nullable fields (I think the current UI allows for blank mission statements, but these are represented as empty strings in the database) but that's beyond the scope of this.

app/presenters/publishing_api/world_location_news_page_presenter.rb
```diff
@@ -17,10 +17,10 @@ def content
       content.merge!(
         description: description,
         details: {},
-        document_type: "placeholder_world_location_news_page",
+        document_type: "placeholder_world_location_news_page", # TODO: do we also want to change this document type? What are the implications
```
BeckaL, line 26: I'll remove this comment - decided to split out updating the document type into a different ticket, as it requires a bit more work than just changing this string.

app/presenters/publishing_api/featured_documents_presenter.rb
```diff
@@ -0,0 +1,65 @@
+module PublishingApi
+  module FeaturedDocumentsPresenter
+    def featured_documents(featurable_item)
+      featurable_item.feature_list_for_locale(I18n.locale).current.limit(6).map do |feature|
```
brucebolt, line 4: As far as I can see, we don't specify a limit for featured documents in the Whitehall Admin UI.  Although we do actually cap at 5 (e.g. on [this page](https://www.gov.uk/world/afghanistan/news) which actually has 8 featured documents) 🤯.

There's also a constant specified [here](https://github.com/alphagov/whitehall/blob/27b40a8f9dac2d2f432ad4b2ecf4726bc2ecb480/app/models/featured_link.rb#L4), but that does't appear to be used everywhere (e.g. for organisation pages) 🤯.

Maybe we should add a message to the UI and consider standardising the limit across all of these pages?

app/presenters/publishing_api/world_location_news_page_presenter.rb
brucebolt: 👍

brucebolt (commented):
This is looking good 👍.

It appears the featured document limit isn't consistent across document types (that doesn't surprise me about Whitehall at all).  I've added a comment next to the relevant line of code, so maybe we could look into this further.

app/presenters/publishing_api/featured_documents_presenter.rb
BeckaL: Ah, a very good point. 6 feels like a sensible limit given the column width (although part of me feels like we should just send them all and let the rendering code decide how many it wants to display). Do you think it would make sense in this PR to leave this as is, add the warning on the UI as you say, and then when we render we can decide whether or not to render 5 (as current) or 6 (as per organisations).

Had a double take on that constant - does not surprise me that there are inconsistently applied limits. But actually I think unless I'm missing something, the constant that you're linking to is actually unrelated to the featured documents - is about the featurable links, not documents, so is applied to organisation pages as far as I can see.

brucebolt: It appears we used this constant incorrectly [in the topical event presenter](https://github.com/alphagov/whitehall/blob/27b40a8f9dac2d2f432ad4b2ecf4726bc2ecb480/app/presenters/publishing_api/topical_event_presenter.rb#L64), so that'll probably need fixing.

My preference would be for us to actually limit the number of featured documents that can be added to a page in the first place, so publishers can't add more than the permitted amount and see an error if they do.  That could be done with a simple Rails validation on one of the models.  Then we wouldn't need any limiting at all when presenting the content.  It would also mean publishers won't assume content is appearing as featured, when it actually isn't.

However, I recall we had a discussion about this with topical events and decided to keep the limit at the point of presentation (I don't seem to find the PR or Slack thread though and can't remember the reasoning).

On that basis, I'd suggest keeping the limit in the publishing application so the the value is set as a constant and used in both the logic and message displayed to users.  Otherwise I fear the limit in the publishing and rendering apps could diverge and become different.

BeckaL: OK cool, that all makes sense, will have a look at adding constants. Can also raise a pr to fix for topical events.

As a side note, while I was having a look at this I was wondering why topical events had their own special stuff for featured documents. Looks like it was because of the classifications stuff, I [created a card](https://trello.com/c/6duIhmSQ/192-spike-using-shared-featurable-code-for-topical-events-instead-of-custom-topicaleventfeaturingscontroller) to spike just using the same shared featureable stuff

db/data_migration/20170814083356_republish_world_location_news_pages.rb
```diff
@@ -1,4 +1,4 @@
 WorldLocation.where(world_location_type_id: 1).each do |world_location|
   puts "Republishing news page for #{world_location.name}"
-  WorldLocationNewsPageWorker.new.perform(world_location.id)
+  WorldLocationNewsWorker.new.perform(world_location.id)
```
brucebolt, line 3: I don't know whether we should change this in historic migrations or not?

It does raise a question about how long migrations should exist for after running in production, before we delete them 🤔.

brucebolt (approved):
Looks good to me, thanks for making the changes 👍

db/data_migration/20170814083356_republish_world_location_news_pages.rb
BeckaL: Yeah I wasn't sure about this so had a chat with Ben about this last week, we came down on the side of they should be changed since if we needed to run all migrations from scratch it wouldn't work.

PR #6706, opened 2022-07-29
Improve paginated history performance

baisa:
[Trello card](https://trello.com/c/UUAOsryV/537-use-query-based-pagination-for-the-document-history-tab)

This PR changes the approach to loading document history when rendering the show page of an Edition.

The previous approach was to load all document history from the database when displaying the page. For a page with a long history this would add a significant performance overhead (for example: https://www.gov.uk/government/publications/royal-courts-of-justice-cause-list has 1000's of updates). 

Co-authored-by: Johnny Young <jonathan.young@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/models/document/paginated_history.rb
+++ b/app/models/document/paginated_history.rb
@@ -1,0 +1,64 @@
+class Document::PaginatedHistory
+  PER_PAGE = 30
+
+  attr_reader :document, :query
+
+  delegate :total_count, to: :query
+
+  def initialize(document, page)
+    @document = document
+    @query = document.edition_versions
+                     .includes(:user)
+                     .where.not(state: :superseded)
+                     .reorder(created_at: :desc, id: :desc)
+                     .page(page)
+                     .per(PER_PAGE)
+  end
+
+  def audit_trail
+    first_edition_id = document.editions.first&.id
+    query.map.with_index do |version, index|
+      AuditTrailEntry.new(version,
+                          is_first_edition: version.item_id == first_edition_id,
+                          previous_version: query[index + 1])
+    end
+  end
+
+  class AuditTrailEntry
+    extend ActiveModel::Naming
+
+    def self.model_name
+      ActiveModel::Name.new(Version, nil)
+    end
+
+    attr_reader :version
+
+    delegate :created_at, :to_key, to: :version
+
+    def initialize(version, is_first_edition:, previous_version: nil)
+      @version = version
+      @is_first_edition = is_first_edition
+      @preloaded_previous_version = previous_version
+    end
+
+    def actor
+      version.user
+    end
+
+    def action
+      case version.event
+      when "create"
+        @is_first_edition ? "created" : "editioned"
+      else
+        previous_version&.state != version.state ? version.state : "updated"
+      end
+    end
+
+  private
+
+    def previous_version
+      # we can avoid n+1 queries by using our preloaded_prev_version
+      @previous_version ||= @preloaded_previous_version || version.previous
+    end
+  end
+end
--- a/app/controllers/admin/edition_audit_trail_controller.rb
+++ b/app/controllers/admin/edition_audit_trail_controller.rb
@@ -3,6 +3,6 @@ class Admin::EditionAuditTrailController < Admin::EditionsController
 
   def index
     @edition = Edition.find(params[:id])
-    @edition_history = Kaminari.paginate_array(@edition.document_version_trail(superseded: false).reverse).page(params[:page]).per(30)
+    @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
   end
 end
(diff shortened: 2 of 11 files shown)


app/models/document/paginated_history.rb
```diff
@@ -0,0 +1,56 @@
...
+    query.map.with_index do |version, index|
+      # [index - 1] returns next version, as array has newest first
+      AuditTrailEntry.new(version,
+                          is_first_edition: version.item_id == document.first_edition&.id,
+                          previous_version: query[index + 1])
+    end
+  end
+
+  class AuditTrailEntry
+    attr_reader :version, :is_first_edition, :preloaded_previous_version
```
ollietreend, line 28: I think we could probably do without `preloaded_previous_version` being part of the public interface for this class.

In other words: I don't think this needs to be an `attr_reader` for `:preloaded_previous_version`. The only place we read its value is from within the class itself, so we could just as easily refer directly to the `@preloaded_previous_version` instance variable.

It'd also mean we can do without the associated test for that attribute. I think it's more of an internal implementation detail.

app/models/document/paginated_history.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def actor
+      version.user
+    end
+
+    def previous_version
+      # we can avoid n+1 queries by using our preloaded_prev_version
+      @previous_version ||= preloaded_previous_version || version.previous
+    end
```
ollietreend, line 45: Likewise for this method. This method is only used internally within this class (to determine what the `action` should be).

Therefore this could become a private method. The only thing we really care about testing is whether the `action` is correct or not – the implementation details of `previous_version` and `preloaded_previous_version` aren't of interest to external consumers of this class.

app/models/document/paginated_history.rb
ollietreend: In fact, I think we could also do without `is_first_edition` being on the public interface. That also looks like it's only used internally.

app/models/document.rb
```diff
@@ -132,6 +132,10 @@ def humanized_document_type
     document_type.underscore.tr("_", " ")
   end
 
+  def first_edition
+    editions.order(id: :asc).first
+  end
```
ollietreend, line 137: I think we should consider adding a sort order to the [existing `Document#editions` association](https://github.com/alphagov/whitehall/blob/4e2a50caba4224ed16689f4fc08157dd4cb24504/app/models/document.rb#L13).

For example:

```ruby
has_many :editions,
         -> { order(created_at: :asc, id: :asc) },
         inverse_of: :document
```

In which case, it'll be safe to call `document.editions.first` and we won't need this `first_edition` method.

For what it's worth: I think editions are already implicitly ordered like this, so I wouldn't expect this to break any tests 🤞🏻

ollietreend (commented):
A couple more things I've just noticed

app/models/document/paginated_history.rb
```diff
@@ -0,0 +1,56 @@
...
+                     .reorder(created_at: :desc, id: :desc)
+                     .page(page)
+                     .per(PER_PAGE)
+  end
+
+  def audit_trail
+    query.map.with_index do |version, index|
+      # [index - 1] returns next version, as array has newest first
+      AuditTrailEntry.new(version,
+                          is_first_edition: version.item_id == document.first_edition&.id,
```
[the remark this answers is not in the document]
jyoung-gds: Done

app/models/document/paginated_history.rb
jyoung-gds: Done and done. I had to get rid of the test for `is_first_edition` as well but this shouldn't matter because the behaviour is tested with the test of action determination, which is the only place the `is_first_edition` is used.

app/models/document/paginated_history.rb
jyoung-gds: Done

app/models/document/paginated_history.rb
```diff
@@ -0,0 +1,65 @@
...
+                     .where.not(state: :superseded)
+                     .reorder(created_at: :desc, id: :desc)
+                     .page(page)
+                     .per(PER_PAGE)
+  end
+
+  def audit_trail
+    first_edition_id = document.first_edition&.id
+    query.map.with_index do |version, index|
+      # [index - 1] returns next version, as array has newest first
```
ollietreend, line 21: I don't think this comment is helpful any more. I think we should probably just remove it to avoid future confusion. _(e.g. "Why does the comment say `index - 1` when the code does `index + 1`? 🤔")_

```suggestion
```

app/models/document.rb
ollietreend: @jyoung-gds I've just merged in #6717 so it should now be safe to use `document.editions.first` and drop this method.

ollietreend (commented):
Really nice work on this @jyoung-gds ✨

Just a couple more things which would be nice to tidy up, and then this will be good to go.

ollietreend (approved):
Nice one 🎉 

Thanks for all your effort on this @baisa, @jyoung-gds & @beccapearce! This is going to make a huge difference to page performance for publishers. ✨

PR #6713, opened 2022-08-02
Add Design system scaffolding Whitehall

DilwoarH:
## What

- Add the design system layout to the Whitehall app.
- Adds bootstrap styling scope

## Why

This will help build up the base layout for all future design system pages. Currently, we have three layouts, admin, frontend, and notification - all of which use bootstrap. We want to introduce another one which has the design system included as well as bootstrap (but within a scope). This will allow us to use components like the bootstrap navbar while still being able to use the design system components without being overridden by the default bootstrap resets.

## Example PRs
- https://github.com/alphagov/travel-advice-publisher/pull/1286

## Trello ref
https://trello.com/c/IeB7CEE9/608-add-design-system-scaffolding-whitehall

## Screenshots
Screenshot 2022-08-02 at 4 53 08 pm

## How to use design system layout

Example controller code:

```ruby
class Admin::DesignSystemController < Admin::BaseController
  layout "design_system"

  def index; end
end
```

## Other potential useful things to consider

If you need a full screen layout, we will need to extend the `.govuk-width-container` class with something like this:

```scss
.govuk-width-container--full {
  max-width: 100%;
  margin: 0 govuk-spacing(4);
}
```


davidtrussler (approved):
This looks great to me. Good stuff! 👍

ollietreend (commented):
Minor nitpick: would you mind renaming the `javascripts/legacy.js` file to `javascripts/frontend.js` instead?

I think this will be clearer to indicate that the JS file is used on the frontend rather than the (legacy) Bootstrap admin interface.

ollietreend (approved):

PR #6717, opened 2022-08-04
Add a sort to Document#editions so they're in order of creation

jyoung-gds:
Added an ascending sort order on `created_at` and `id`. This needs to be investigated a bit more to see why the test `test/unit/tasks/data_hygiene/change_note_remover_test.rb:67` is failing.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -10,7 +10,9 @@ class Document < ApplicationRecord
 
   after_destroy :destroy_all_editions
 
-  has_many :editions, inverse_of: :document
+  has_many :editions,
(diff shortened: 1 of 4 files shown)


ollietreend:
@jyoung-gds the tests should pass on this now.

See 9ba336d892774d13097624b7cf2d02bcc8612865 for the fix. It turns out the test setup wasn't setting the correct `created_at` timestamps.

I've also added a test for the sort order on `Edition#documents` just to make it really clear that this is expected behaviour. We want something really obvious to break if that sort is ever removed.

beccapearce (approved):
Nice

PR #6722, opened 2022-08-09
Split World Location News out of World Locations

brucebolt:
The `WorldLocation` model currently contains the data for two published pages: the World Location itself and the World Location News page.

This has led to some unusual code needing to be written specifically for the news page as publishing the `WorldLocation` model using `PublishesToPublishingApi` and `Searchable` will lead to only the World Location page being published.

By splitting the model into two parts, we will be able to remove this custom code and re-use existing methods for publishing the news page.

[Trello card](https://trello.com/b/NHxZVLsg)


BeckaL:
For this todo (you won't be able to see it in this code once I remove the commit that adds the TODO)

TODO 1/3:

```
# TODO: what do we actually need to send to search for world location
  searchable title: :title,
             link: :search_link,
             description: :search_description,
             only: :active_international_delegation,
             format: "world_location",
             slug: :slug
```

I think I understand things very slightly more having looked into international delegations and their news pages. We only really publish world locations that are of the type 'international delegation', hence the code in this stuff we send to search. I think we should do a similiar refactoring to what we've done here for international delegations, splitting them out of world locations (and dropping their news pages which I don't think are useful - pending a product decisions). So I'm voting to keep this in for now, and address as part of addressing international delegations. Anybody feel free to shout if they think we _should_ do something as part of this.

BeckaL:
TODOs 2 and 3 / 3:

```
  # TODO: what do we actually need to send to publishing api for world location

  # TODO: why does this override the method in PublishesToPublishingApi
  def publish_to_publishing_api
    # WorldLocations no longer support translations
    # but the current world news pages use the world location object
    # to build their featured articles lists so we need to keep
    # the translations but not publish them.
    run_callbacks :published do
      PublishingApiWorker.perform_async(
        self.class.name,
        id,
        nil,
        "en",
      )
    end
  end
```
I'm also proposing we leave these as they are and address as part of future work to rationalise world locations / international delegations.

I basically think we should be able to split out international delegations and remove world locations as a model entirely. But not addressing them as part of this.

BeckaL:
Have just broken the tests again, I think there's one last bit of search behaviour that's off in that we're sending translated news pages to search now, will look at this tomorrow. But otherwise think this _should_ be a passing build

app/models/searchable.rb
```diff
@@ -148,7 +148,16 @@ def search_index
     end
   end
 
+  # TODO: I don't think this is a very nice way of achieving this behaviour
```
brucebolt, line 151: I wonder how this works for organisations, as these have translations but aren't included in search?

brucebolt: I've come up with a solution that doesn't seem to affect any of the other document types.  I'm not sure why we're not filtering for other documents 😕 .

test/integration/world_location_integration_test.rb
```diff
@@ -37,6 +37,51 @@ def put_content_hash_containing(locale, title, mission_statement)
...
+      Services.publishing_api.expects(:put_content).once.with("id-123", put_content_hash_containing("en",  world_location_without_translations.title, new_mission_statement))
+      Services.publishing_api.expects(:put_content).once.with(world_location_without_translations.content_id, has_entries(document_type: "world_location"))
+      Whitehall::FakeRummageableIndex.any_instance.expects(:add).with(presenter.content_for_rummager("id-123"))
+      Services.publishing_api.expects(:publish).at_least_once
+
+      click_on "Save"
+    end
+  end
+
+  test "when updating an inactive location, does not make any calls to search api" do
```
JonathanHallam, line 82: totally not the place for this chat, but what's an inactive location?

brucebolt: From what I can see, we've imported the list of World Locations from somewhere.  For some of these (e.g. the UK), we don't want a world location page, so they are marked as 'inactive'.  Although we weirdly still have a location page published, presumably for legacy reasons 😕.

app/controllers/admin/world_location_news_controller.rb
```diff
@@ -46,16 +46,17 @@ def optional_filter_params
...
+    @world_location_news = @world_location.world_location_news
   end
 
-  def world_location_params
-    params.require(:world_location).permit(
-      :world_location_type_id,
-      :title,
-      :active,
+  def world_location_news_params
+    Rails.logger.debug params.inspect
```
JonathanHallam, line 54: is this meant to still be here?

app/models/world_location.rb
```diff
@@ -13,23 +13,18 @@ class WorldLocation < ApplicationRecord
            source: :document
   has_many :worldwide_organisation_world_locations, dependent: :destroy
   has_many :worldwide_organisations, through: :worldwide_organisation_world_locations
-  has_many :offsite_links, as: :parent
 
-  has_many :featured_links, -> { order(:created_at) }, as: :linkable, dependent: :destroy
-  accepts_nested_attributes_for :featured_links, reject_if: ->(attributes) { attributes["url"].blank? }, allow_destroy: true
-
-  include Featurable
+  has_one :world_location_news
```
JonathanHallam, line 17: Really like how this cuts down the complexity of the model. I'm assuming the featured/offsite links were only used in the previously coupled world location news bit of this

app/presenters/publishing_api/world_location_news_presenter.rb
```diff
@@ -2,16 +2,19 @@ module PublishingApi
   class WorldLocationNewsPresenter
     include FeaturedDocumentsPresenter
 
-    attr_accessor :world_location, :update_type
+    attr_accessor :world_location, :world_location_news, :update_type
 
-    def initialize(world_location, update_type: nil)
-      self.world_location = world_location
+    def initialize(world_location_news, update_type: nil)
```
JonathanHallam, line 7: suuuuper minor and I know it was done this way already, but what's the advantage of passing in `nil` here rather than `"major"`, would mean line 10 could be `self.update_type = update_type`

app/views/admin/world_location_news/edit.html.erb
```diff
@@ -0,0 +1,30 @@
+<% page_title "Edit " + @world_location_news.name %>
+
+<div class="row">
+  <section class="col-md-8">
+    <h1>Edit world location</h1>
+  <p class="warning">Warning: changes to worldwide locations appear instantly on the live site.</p>
```
JonathanHallam, line 6: Should this be "changes to world location news articles appear instantly...." or do we not care about publishers knowing the difference?

app/controllers/admin/world_location_news_controller.rb
brucebolt: Nope.  I'll remove and rebase.

app/models/world_location.rb
brucebolt: Yes, they were only used for the news part.  The rest of the `WorldLocation` didn't use them at all.

app/views/admin/world_location_news/edit.html.erb
brucebolt: We've kept [the previous wording](https://github.com/alphagov/whitehall/blob/26f54ba8ac13ef5c2da8ba4c14a6dd1b610ded0d/app/views/admin/world_locations/edit.html.erb#L6) as the content being published is exactly the same as before.

The whole interface around world locations probably needs revisiting from a design perspective, so maybe we could ask Publishing Experience to put this on their backlog?

test/integration/world_location_integration_test.rb
```diff
@@ -37,6 +37,51 @@ def put_content_hash_containing(locale, title, mission_statement)
...
+      world_location_without_translations = create(:world_location, news_page_content_id: "id-123")
+      visit edit_admin_world_location_path(world_location_without_translations)
+      new_mission_statement = "a different mission"
+      fill_in "world_location_mission_statement", with: new_mission_statement
+
+      presenter = PublishingApi::WorldLocationNewsPresenter.new(world_location_without_translations)
+
+      Services.publishing_api.expects(:put_content).once.with("id-123", put_content_hash_containing("en",  world_location_without_translations.title, new_mission_statement))
+      Services.publishing_api.expects(:put_content).once.with(world_location_without_translations.content_id, has_entries(document_type: "world_location"))
+      Whitehall::FakeRummageableIndex.any_instance.expects(:add).with(presenter.content_for_rummager("id-123"))
```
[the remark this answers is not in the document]
brucebolt: Hopefully we can look at [this Trello card](https://trello.com/c/rC4EjFOV/125-rename-references-to-rummager-in-whitehall-to-search-api) sometime soon, so all the references can be finally removed.

app/presenters/publishing_api/world_location_news_presenter.rb
brucebolt: We looked into this offline and found it isn't possible.  The `update_type` will always be `major`, as this is non-editionable content, so we think further work should be done (separately) to refactor the non-editionable presenters so this value isn't needed to be defined at all.

config/routes.rb
```diff
@@ -361,13 +361,22 @@ def external_redirect(path_prefix, target)
           resources :role_appointments, only: %i[new create edit update destroy], shallow: true
           resources :translations, controller: "role_translations"
         end
-        resources :world_locations, only: %i[index edit update show] do
+        # resources :world_locations, only: %i[index edit update show] do
+        #   member do
+        #     get "/features(.:locale)", as: "features", to: "world_locations#features", constraints: { locale: valid_locales_regex }
```
JonathanHallam, line 366: suspect this comment should be removed

db/migrate/20220825093553_split_world_location_news_page.rb
```diff
@@ -0,0 +1,40 @@
...
+    remove_column :world_location_translations, :title, :string # rubocop:disable Rails/BulkChangeTable
+    remove_column :world_location_translations, :mission_statement, :text
+
+    OffsiteLink.where(parent_type: "WorldLocation").update_all(parent_type: "WorldLocationNews")
+    FeaturedLink.where(linkable_type: "WorldLocation").update_all(linkable_type: "WorldLocationNews")
+    FeatureList.where(featurable_type: "WorldLocation").update_all(featurable_type: "WorldLocationNews")
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
JonathanHallam, line 38: minor - maybe add a line in the commit message on why this change is irreversible

test/functional/admin/world_location_news_controller_test.rb
```diff
@@ -27,59 +27,60 @@ class Admin::WorldLocationsControllerTest < ActionController::TestCase
...
 
   test "updating should be able to create a new featured link" do
     world_location = create(:world_location)
 
     post :update,
          params: { id: world_location,
-                   world_location: {
-                     featured_links_attributes: { "0" => {
+                   world_location_news: {
+                     featured_links_attributes: [{
```
JonathanHallam, line 59: 😬

test/integration/world_location_integration_test.rb
```diff
@@ -20,47 +20,72 @@ class WorldLocationIntegrationTest < ActionDispatch::IntegrationTest
...
-                             mission_statement: @original_english_mission_statement,
-                             news_page_content_id: "id-123",
-                             title: @original_english_title,
                              translated_into:
-                               { fr: {
-                                 name: "La France",
-                                 title: @original_french_title,
-                                 mission_statement: @original_french_mission_statement,
-                               } })
+                               { fr: { name: "La France" } },
```
JonathanHallam, line 33: 🇫🇷

config/routes.rb
brucebolt: Ah, yes, it should've been.  I'll remove it now.

db/migrate/20220825093553_split_world_location_news_page.rb
brucebolt: I've split the migration out into a separate commit and added an explanation as to why it's irreversible.  It probably would be possible to make it reversible, but the migration would become very complex (since we'd need to somehow recombine the translations from two different tables).

test/support/govspeak_validation_test_helper.rb
```diff
@@ -1,12 +1,17 @@
 module GovspeakValidationTestHelper
-  def should_protect_against_xss_and_content_attacks_on(*attributes)
+  def should_protect_against_xss_and_content_attacks_on(*attributes, supplied_factory_name)
```
JonathanHallam: why have changes been made to this file?

brucebolt: The method previously made an assumption that you wanted to run the XSS tests against the factory whose name matched the class being tested (e.g. the `:world_location_news` factory would be used for the `WorldLocationNews` model).  However we want to test against the `:world_location_news_with_stubbed_slug` factory instead, so can't use this assumption anymore.

We shifted this to always be the first argument to the method as it's no longer optional and would've had to be utilised by appending `supplied_factory_name: :foo` to the method calls, because of the usage of the splat operator prior to this.  This change means we can just add it as the first argument.

config/locales/lt.yml
```diff
@@ -45,7 +45,7 @@ lt:
     accessibility:
       full_help_html: 'Jei norėtumėte gauti šį dokumentą kitokiu formatu (brailio, audio ar pan.), siųskite el. paštą %{email}, parašydami savo adresą, telefoną bei leidinio pavadinimą .Prašome pasakyti, kokio formato jums reikia. Mums padės, jei pasakysite, kokią pagalbinę technologiją naudojate.
 
-'
+        '
```
JonathanHallam, line 48: extra whitespace

config/locales/pa-pk.yml
```diff
@@ -45,7 +45,7 @@ pa-pk:
     accessibility:
       full_help_html: 'اگر تُسی مدد گارٹیکنالوجی نوں استعمال کردے او ( جیویں کہ اک سکرین ریڈر) تے تہانوں ایس دستاویز دے ترجمے دی لوڑ اے جہیڑا اک بوہت پہنچ والی شکل وچ ہوئے گا، برائے مہربانی ای میل کریں {email}% برائے مہربانی سانوں ایہ دسو کہ تہانوں کس شکل دی لوڑ اے۔ ایس توں سانوں مدد ملے گی کہ تسی کس طراں دی مددگار ٹیکنالوجی استعمال کردے او۔
 
-'
+        '
```
JonathanHallam, line 48: whitespace

config/locales/ja.yml
```diff
@@ -45,7 +45,7 @@ ja:
     accessibility:
       full_help_html: 'お客様が支援技術（スクリーンリーダーなど）を使用していて、このドキュメントよりアクセスしやすい形式のバージョンが必要な場合は、必要なフォーマットや使用している支援技術を添え％{email}にEメールをお送くりください。
 
-'
+        '
```
JonathanHallam, line 48: whitespace

JonathanHallam (approved):
Approving this, I'm adding a do not merge flag to this so that we can do some testing on integration immediately before merging

config/locales/ja.yml
brucebolt: Looks like this was added in when I normalized the locale files.  I'll remove the whitespace here and add the normalization as either another PR or another commit.

config/locales/pa-pk.yml
brucebolt: 👍

PR #6724, opened 2022-08-10
Limit world location featured documents at 5

brucebolt:
When World Location News pages are rendered by Whitehall, we limit the number of featured links at 5, regardless of the number entered into the UI.

This is consistent with the message displayed to publishers (even though we don't actually limit the number they can enter).

Updating the behaviour here so the content item also only contains thefirst 5 featured links, so we present the same content through Collections when rendering is switched over.

[Trello card](https://trello.com/c/i577upcN)


BeckaL (approved):
Ah missed this, thanks for spotting ⭐

PR #6727, opened 2022-08-11
Port in accessible autocomplete from content publisher

DilwoarH:
This PR ports in accessible autocomplete from the content publisher

Preview: http://whitehall-admin.dev.gov.uk/component-guide/autocomplete

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -3,5 +3,6 @@
 //= require govuk-admin-template
 //= require admin
 //= require components/miller-columns
+//= require components/autocomplete
 
 window.GOVUK.navBarHelper.init()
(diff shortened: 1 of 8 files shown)


DilwoarH:
> Is it worth considering putting this into the gem instead?

I'd like to do so but not this version, this version isn't well supported. This one: https://github.com/alphagov/accessible-autocomplete - is worth putting into the GEM

andysellick:
In terms of actual components, there's also this one in smart answers, which is a very trimmed down simpler version, again based on the same original code: https://github.com/alphagov/smart-answers/blob/main/app/views/components/_autocomplete.html.erb

davidgisbey (approved):
awesome we need for the edit page 👍

DilwoarH:
> In terms of actual components, there's also this one in smart answers, which is a very trimmed down simpler version, again based on the same original code: https://github.com/alphagov/smart-answers/blob/main/app/views/components/_autocomplete.html.erb

That's interesting (didn't know this), I'll work on getting these into the components gem

PR #6728, opened 2022-08-11
Remove extra parameter in deploy workflow

theseanything:
This removes the workflowTrigger parameter from the deploy workflow, which is no longer used. This should fix the call to the reusable workflow.


nsabri1 (approved):
LGTM

PR #6730, opened 2022-08-12
Fix world location news translations

BeckaL:
This fixes a bug with the content items sent for world location news pages, whereby editing a translatable field, such as the mission statement, sent that field through in the content item for all locales. So this means that editing the French content for a world location news item would have resulted in French content appearing on the English page. The same would happen for editing the English content, with English content then appearing on the French page.

The fix in this commit is to switch the after_update hook to an after_commit hook for world location news pages. This addresses a known issue - see https://github.com/globalize/globalize/issues/659 - with the Globalize gem, whereby in an after_update or after_save hook does not respect the I18n locale that it is acting in the context of. After_commit does not have the same issue.

To note that we are adding the publishing to publishing api behaviour now when the world location page is saved, as well as when it is updated. Previously, this behaviour only happened on update. We have decided to introduce this behaviour now since when we switch to using the `PublishesToPublishingApi` trait, we will introduce this behaviour anyway. If this results in undesired behaviour (at worst, some empty world location / world location news pages published to the website), the fix for this should be to make these pages properly editionable and creatable via the UI, rather than continuing to hack around world location news pages.

[Trello](https://trello.com/c/BcmT8Ley/205-fix-presentation-of-translated-world-location-news-pages-to-publishing-api)

### DIFF
--- a/test/unit/workers/world_location_news_page_worker_test.rb
+++ b/test/unit/workers/world_location_news_page_worker_test.rb
@@ -33,6 +33,7 @@ class WorldLocationNewsPageWorkerTest < ActiveSupport::TestCase
     world_location = FactoryBot.create(
       :world_location,
       name: "France",
+      title: "UK and France",
       news_page_content_id: "id-123",
       translated_into: [:fr],
     )
@@ -42,8 +43,12 @@ class WorldLocationNewsPageWorkerTest < ActiveSupport::TestCase
     end
 
     I18n.with_locale(:fr) do
+      world_location.update!(title: "Le Royaume-Uni et la France")
       @french = PublishingApi::WorldLocationNewsPresenter.new(world_location).content
     end
+
+    assert_equal("UK and France", @english[:title])
+    assert_equal("Le Royaume-Uni et la France", @french[:title])
 
     Services.publishing_api.expects(:put_content).with("id-123", @english)
     Services.publishing_api.expects(:publish).with("id-123", nil, locale: :en)
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -23,7 +23,7 @@ class WorldLocation < ApplicationRecord
   accepts_nested_attributes_for :edition_world_locations
   accepts_nested_attributes_for :offsite_links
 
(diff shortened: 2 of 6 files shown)


app/workers/world_location_news_worker.rb
```diff
@@ -2,7 +2,7 @@ class WorldLocationNewsWorker < WorkerBase
   attr_accessor :world_location
 
   def perform(world_location_id)
-    self.world_location = WorldLocation.find(world_location_id)
+    self.world_location = WorldLocation.find(world_location_id).reload
```
brucebolt, line 5: Do we still need the `reload` here?  I think we only added it to test something.

test/unit/workers/world_location_news_page_worker_test.rb
```diff
@@ -43,7 +43,7 @@ class WorldLocationNewsPageWorkerTest < ActiveSupport::TestCase
     end
 
     I18n.with_locale(:fr) do
-      world_location.update(title: "Le Royaume-Uni et la France")
+      world_location.update!(title: "Le Royaume-Uni et la France")
```
brucebolt, line 46: Could this line be moved into the previous commit, as that's where it was first added?

app/workers/world_location_news_worker.rb
BeckaL: Oh yeah good shout, will remove

test/unit/workers/world_location_news_page_worker_test.rb
BeckaL: Think this one might also actually be an unnecessary change as well from when I was playing around, so will see if just taking this change out affects the test (I think not)

BeckaL: Taken out now

brucebolt: Yeah, I was just about to say that rubocop complains if you don't use the bang method and don't ever use the method's output.

brucebolt (approved):
Looks good to me 👍

brucebolt (approved):

PR #6733, opened 2022-08-15
Make it possible to find the latest and live editions of a document more efficiently

ollietreend:
This PR introduces more efficient 'has one' associations on the Document model to find its most recent Edition, and the currently live Edition (published or withdrawn).

The interface remains unchanged – the latest edition can be found with `Document#latest_edition` and the currently live edition is at `Document#live_edition`.

## Database changes

New columns `latest_edition_id` and `live_edition_id` have been added to the `documents` table.

These fields can be `null` because it's possible for a document _not to have_ a live edition, or any editions at all.

[Foreign key constraints](https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html) have been configured to to maintain referential integrity between Documents and Editions.

If an Edition is deleted ('hard' deleted), any associated Documents will be updated with a `null` reference. This way, the Document won't be left referencing an Edition that no longer exists. And if an Edition has its ID changed, associated Documents will be updated with the new ID too, so that the existing reference isn't broken. In practice, these scenarios are unlikely to ever happen because we only ever 'soft delete' Editions and never change their IDs. But it should give us extra confidence that we can always depend upon the linked `latest_edition_id` and `live_edition_id` to exist.

### Do these foreign key constraints cause test failures?

In short: no.

In the spike PR (#6696) which implemented foreign key constraints, tests needed to be changed to accommodate for the new constraints (211eb42e639da6e38a4c4a18a4376b93ea215f83). In that branch, the ['on delete'](https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-key-referential-actions) behaviour hadn't been explicitly configured, meaning MySQL defaulted to the 'restrict' behaviour. That meant an Edition couldn't be deleted if a Document was still referencing it.

Instead, this PR configures the foreign key constraint with the 'set null' option. This means Editions can be deleted, and any Document references will be reset to `null`. As a result, no workarounds needed to be added to tests for compatibility with these new constraints.

## Data migration

There are a large number of Editions and Documents which will need migrating to this new schema. Due to the scale of this migration, we'll want to take a ['slow but safe'](https://www.userlike.com/en/blog/database-schema-migrations) approach:

1. Add new columns to schema
2. Forwards populate new records
3. Backfill existing records
4. Deploy code changes to use the new columns

This PR covers steps 1–3 of the above. Any newly created or published Editions will populate the Document's `latest_edition_id` and `live_edition_id` fields immediately (this is the 'forwards populate' step). The [data migration](https://github.com/alphagov/whitehall/blob/811ff58d063e7306fbf6d4a25f00f134c61a4c37/db/data_migration/README.md) can then be run which will populate those fields for existing records (the 'backfill' step). In testing, I found the backfill to take ~1 hour to run across the entire Whitehall dataset.

Once the data migration has completed, another PR can be raised which switches the app over to use those new columns.

## Rationale for using a Rails model callback

Rails model callbacks [aren't always a good idea](https://samuelmullen.com/2013/05/the-problem-with-rails-callbacks/). When callbacks trigger external behaviours (for example, an after save callback which sends an email) then it becomes difficult to unit test the model in isolation without them firing (for example, attempting to fire off a bunch of emails). It then usually becomes necessary to stub various methods so external behaviour doesn't interfere with unit tests.

However, a generally accepted rule for using callbacks is:
> “Use a callback only when the logic refers to state internal to the object.”

In this case, the Edition will update the references on the linked Document. This slightly bends the above rule, because it's not technically _internal state_. However since every Edition has a Document, and this behaviour is about updating the association between the two, I still think could be seen as an acceptable use case. There is already a coupling between these two models.

Thankfully the introduction of this callback didn't cause tests to break or require any new stubs. Hopefully this can be seen as a sign that we're not introducing a code smell.

One of the advantages of this approach is that we don't need to change the various areas of the codebase which create and update Editions. The responsibility for maintaining that ActiveRecord association lives in one place, directly within the model itself.

---

Trello: https://trello.com/c/iWcORsD5

### DIFF
--- a/db/data_migration/20220812212927_backfill_latest_edition_and_live_edition_on_document.rb
+++ b/db/data_migration/20220812212927_backfill_latest_edition_and_live_edition_on_document.rb
@@ -1,0 +1,12 @@
+# Populate latest_edition_id and live_edition_id for documents that don't have them defined
+all_documents = Document.where(latest_edition_id: nil, live_edition_id: nil)
+total = all_documents.count
+
+@logger.info "There are #{total} documents to populate"
+
+done = 0
+all_documents.find_in_batches do |documents|
+  documents.each(&:update_edition_references)
+  done += documents.count
+  @logger.info "Done #{done}/#{total} (#{((done / total.to_f) * 100).to_i}%)"
+end


app/models/document.rb
```diff
@@ -138,6 +138,14 @@ def humanized_document_type
     document_type.underscore.tr("_", " ")
   end
 
+  def update_edition_references!
+    latest = editions.reverse_order
+    update!(
+      latest_edition_id: latest.pick(:id),
```
davidgisbey, line 144: I think we need to exclude editions which are deleted

```
latest = editions.where.not(state:  "deleted:).reverse_order
```

test/unit/document_test.rb
```diff
@@ -311,4 +311,34 @@ class DocumentTest < ActiveSupport::TestCase
...
+
+    # 2. Publish it
+    force_publish(first_edition)
+    document.update_edition_references!
+    assert_equal first_edition.id, document.latest_edition_id
+    assert_equal first_edition.id, document.live_edition_id
+
+    # 3. Create a new draft
+    second_edition = first_edition.create_draft(create(:user))
+    document.update_edition_references!
```
davidgisbey, line 333: Do we need to call this explicitly? I feel like this should test that this method is called as part of the create/update. Wdyt?

Maybe that's fine, but if so maybe we need another test where we expect `update_document_edition_references` to be called on create and update when state changes

app/models/document.rb
ollietreend: I think this is already covered due to `Edition` having that tricksy default scope (😄😭)

https://github.com/alphagov/whitehall/blob/811ff58d063e7306fbf6d4a25f00f134c61a4c37/app/models/edition/workflow.rb#L21

However, I could add a deleted edition to the unit test for this method. That'll at least avoid a regression if ever we come to change the default scope.

What do you think?

davidgisbey: oh right yea good point. I always forget about the scoped find. All good as is then 👍

db/data_migration/20220812212927_backfill_latest_edition_and_live_edition_on_document.rb
```diff
@@ -0,0 +1,12 @@
...
+total = all_documents.count
+
+@logger.info "There are #{total} documents to populate"
+
+done = 0
+all_documents.find_in_batches do |documents|
+  documents.each(&:update_edition_references!)
+  done += documents.count
+  @logger.info "Done #{done}/#{total} (#{((done / total.to_f) * 100).to_i}%)"
+end
```
davidgisbey, line 12: I don't think it's a big deal, but this will be pretty slow right? I'm guessing it'll take a good few minutes.

Is there a downside to doing something like this?

```
execute <<-SQL
      UPDATE documents
      SET live_edition_id = (
        SELECT MAX(id)
        FROM editions
        WHERE documents.id = editions.document_id
        WHERE documents.live_edition_id IS NULL
        AND editions.state IN ('published', 'withdrawn')
      )
    SQL
```

app/models/document.rb
ollietreend: The default scope is the sort of thing that now seems like a necessary evil, because the whole app has been built around the assumption that it can never accidentally find a "deleted" Edition.

This does seem like an obvious 'new' case of that assumption being used though.

I might try throwing in a deleted Edition to this method's unit test – just to explicitly declare that assumption & avoid an accidental regression.

db/data_migration/20220812212927_backfill_latest_edition_and_live_edition_on_document.rb
davidgisbey: Yea all fair points! It takes about 45 seconds to run the raw SQL on the int dataset though so it's quite a bit quicker.

ollietreend: Oh blimey 🙈 that's WAY quicker!!

app/models/document.rb
```diff
@@ -138,6 +138,14 @@ def humanized_document_type
     document_type.underscore.tr("_", " ")
   end
 
+  def update_edition_references!
```
kevindew, line 141: nit: should only use a bang method when there is also a non-bang version: https://rubystyle.guide/#dangerous-method-bang

db/data_migration/20220812212927_backfill_latest_edition_and_live_edition_on_document.rb
kevindew: The downside of a single SQL query is that you can lock all the records during the transaction and thus not allow anyone to write to the DB until it's complete. I'm not too sure about how that plays out in MySQL but it can be an annoying for Postgres.

Though if you're talking 45s vs 1 hour I'd probably go with the fast query.

kevindew (commented):
This looks great, thanks @ollietreend.

I do feel a bit 😬 about the AR callbacks, but you've reasoned them well in the PR description and it seems reasonable to use them for denormalising. 

I think we will hit a somewhat unideal scenario with this with the weird unpublishing system. I'm not sure if it'll be a problem though, I reckon we may already have the problem. The example would be:

Context: We have a document that has been published twice (a and b). The latest and live edition are currently b.

- We go to unpublish item b.
- b becomes the latest edition
- a inexplicably becomes the live one

davidgisbey (approved):
As discussed. Current approach seems good to me 👍

db/data_migration/20220812212927_backfill_latest_edition_and_live_edition_on_document.rb
ollietreend: Not to dismiss this suggestion, but I think I'm going to leave the database migration as is since neither of us seem to hold a strong opinion either way.

However, the SQL query may prove handy on dev machines if we want to quickly backfill data without pulling down a fresh copy of the integration database.

PR #6734, opened 2022-08-15
[Part1] Remove attachment tabs from editions edit page

davidgisbey:
## Description 

This is the first of 4 pull requests. All of this work sits behind the `REMOVE_EDIT_TABS ` permission https://github.com/alphagov/whitehall/blob/main/app/models/user.rb#L32

The pull requests will do the following 

1. This PR will remove tabs from the Edition edit page for all Edition types except `DocumentCollections` and `Consultations` 
2. Remove tabs for `DocumentCollections` and add them to the Edition Summary page (otherwise there would be no way to navigate to them)
3.  Ditto for `Consultations`
4.  Remove all tabs from the Attachments index page (Document etc.)

## Trello card

https://trello.com/c/X0ZL0ouj/625-remove-attachment-tab-from-edit-edition-page-and-document-tab-from-attachments-index-page

### DIFF
--- a/test/integration/attachment_replacement_integration_test.rb
+++ b/test/integration/attachment_replacement_integration_test.rb
@@ -66,7 +66,9 @@ class AttachmentReplacementIntegrationTest < ActionDispatch::IntegrationTest
           stub_publishing_api_has_linkables([], document_type: "topic")
           visit admin_news_article_path(edition)
           click_button "Create new edition to edit"
-          click_link "Attachments 1"
+          fill_in "edition_change_note", with: "changes"
+          click_button "Save"
+          click_link "Modify attachments"
           within ".existing-attachments" do
             click_link "Edit"
           end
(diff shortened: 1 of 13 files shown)


test/integration/attachment_replacement_integration_test.rb
```diff
@@ -66,7 +66,7 @@ class AttachmentReplacementIntegrationTest < ActionDispatch::IntegrationTest
           stub_publishing_api_has_linkables([], document_type: "topic")
           visit admin_news_article_path(edition)
           click_button "Create new edition to edit"
-          click_link "Attachments 1"
+          visit admin_edition_attachments_path(edition.latest_edition)
```
ollietreend, line 69: Not something that can be solved in this ticket, but it makes me sad that we need to directly navigate to the attachments page. 😢 

Ideally this is the sort of test that'd navigate as if it were a real user (e.g. click links). But that's not possible here because there's no user friendly way to navigate from the 'edit' screen to the 'summary' and on to 'attachments'.

ollietreend (approved):
Looks good to me 👍🏻

test/integration/attachment_replacement_integration_test.rb
davidgisbey: I've actually done this for the other specs. Not sure why i was lazy here. Will update now!

PR #6735, opened 2022-08-15
Bump all Sentry gems to 5.4.1

ollietreend:
This commit bumps all three Sentry gems used by this project to version 5.4.1.

An internal API change was made in the Sentry gems which means 5.3.x gems aren't compatible with 5.4.x gems.

If all three gems aren't updated at the same time, we end up with test failures containing error messages like this:

```
exception expected, not
Class: <ArgumentError>
Message: <"wrong number of arguments (given 1, expected 2)">
```

For further details, see getsentry/sentry-ruby#1859 and getsentry/sentry-ruby#1865.


beccapearce (approved):
Great

PR #6738, opened 2022-08-16
Fix 'FakeRummagableIndex' error in development

ollietreend:
A 'constant `Whitehall::FakeRummagableIndex` is already defined' error message would frequently appear when working in a development environment.

The error would appear when changing files in the project while the Rails server is already running. On the first page reload after making code changes, this error would appear. It'd usually disappear again on a second reload.

The error message suggested that the `FakeRummagableIndex` class was being redefined. Presumably this was happening because Rails reloads source code when changes are made.

It seems that this change fixes the issue by only defining the `Whitehall::FakeRummagableIndex` class if it hasn't already been defined.

### DIFF
--- a/lib/whitehall/search_index.rb
+++ b/lib/whitehall/search_index.rb
@@ -50,37 +50,39 @@ module Whitehall
     end
   end
 
-  class FakeRummageableIndex < Rummageable::Index
-    class << self
-      attr_accessor :store
-    end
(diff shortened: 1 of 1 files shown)


brucebolt (approved):
Thanks for looking into this 🏅.  It's caused me so many problems recently!

PR #6739, opened 2022-08-16
Remove Bootstrap modal popups

ollietreend:
This PR removes the two modal popups that existed in Whitehall's admin interface.

It was a surprisingly easy change to make because standalone pages already existed for these actions. They were just being [progressively enhanced](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement) with Bootstrap JS modal dialogues.

The final commit in this PR removes the JavaScript module that was used to generate the modal popups, since it was no longer used.

### Confirm force publish

Trello: https://trello.com/c/yrvHvS2g/634-add-confirm-force-publish-page

| Before | After |
| --- | --- |
| 'Confirm force publish' before changes | 'Confirm force publish' after changes |

### Confirm unwithdrawal

Trello: https://trello.com/c/ABixlAH2/639-add-confirm-unwithdraw-page

| Before | After |
| --- | --- |
| 'Confirm unwithdrawal' before changes | 'Confirm force publish' after changes |


davidgisbey:
Looks great. Currently this will also affect corporate information pages

image

image

Personally i think that's fine and we should continue to do so going forward so we flag as little code as possible, but i'm going to run it by people at standup

davidgisbey:
We've agreed we're going to ship small changes like this in the editions and corporate information tab

davidgisbey (approved):

PR #6737, opened 2022-08-16
[Part2] Remove attachment tabs from editions edit page

davidgisbey:
## Description 

This follows on from https://github.com/alphagov/whitehall/pull/6734

It does the following:

1. Removes the tabs from the edition edit page for DocumentCollections
2. Moves the link to the `Collection documents` page to the edition summary page
3. Adds a test for the new flow

### Edition summary page 

image

#### Before 

image

#### After

image

#### Before

image

#### After 

image

## Guidance for review

The design on the section on the Edition Summary page will need some discussion with Nikin when he gets back. For now i'm just linking to the section. It's behind a flag though so it shouldn't stop us merging and iterating

## Trello card

https://trello.com/c/X0ZL0ouj/625-remove-attachment-tab-from-edit-edition-page-and-document-tab-from-attachments-index-page

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6740, opened 2022-08-16
[Part3] Remove attachment tabs from editions edit page

davidgisbey:
## Description 

This follows on from https://github.com/alphagov/whitehall/pull/6737

It does the following:

1. Removes the tabs from the edition edit page for edit Consultation page
2. Removes tabs from the public feedback page 
3. Removes tabs from the final outcome page
4. Moves the link to the `Public feedback` page to the edition summary page
5. Moves the link to the `Final outcome` page to the edition summary page

### Edition summary page 

image

#### Before 

image

#### After

image

#### Before

image

#### After 

image

#### Before

image

#### After 

image

## Trello card

https://trello.com/c/X0ZL0ouj/625-remove-attachment-tab-from-edit-edition-page-and-document-tab-from-attachments-index-page

### DIFF
--- a/app/views/admin/responses/show.html.erb
+++ b/app/views/admin/responses/show.html.erb
@@ -3,10 +3,15 @@
 
 <div class="row">
   <section class="col-md-8">
+    <% if current_user.can_remove_edit_tabs? %>
+      <span class="back">
+        <%= link_to 'Back', admin_edition_path(@edition) %>
+      </span>
+    <% end %>
+
     <h1><%= @response.friendly_name.capitalize %> for consultation</h1>
 
-    <%= consultation_editing_tabs(@edition) do %>
-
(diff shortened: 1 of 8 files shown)


app/views/admin/responses/show.html.erb
```diff
@@ -1,12 +1,18 @@
 <% page_title "#{@response.friendly_name.capitalize} for: #{@edition.title}" %>
 <% page_class "response" %>
 
+
```
beccapearce: Tiny nit pick that I don't think this line needs to be added

beccapearce: Ah I see it's removed in the next commit, feel free to ignore

davidgisbey: Ah i squashed it into the wrong commit 🤦 good catch! Thanks

PR #6742, opened 2022-08-17
Include location type when presenting World Location News

brucebolt:
This is required so we can add a label to the World Location News page telling users whether the news relates to a World Location or International Delegation.

This is dependent on https://github.com/alphagov/govuk-content-schemas/pull/1114 being merged and deployed.

[Trello card](https://trello.com/c/zxDhAkI3)


BeckaL (approved):
LGTM

BeckaL (commented):
Actually, see comment on content schemas PR. Question as to whether there's any point adding to the schema if it's always shown as "world location news". Also, this key says world location, not world location news, so still more confused, as I think world location news is the thing we're trying to put in.

But also see comment about my current state of extreme confusion 😂

BeckaL (approved):
✨

PR #6757, opened 2022-08-22
Remove expensive edition queries

ollietreend:
This PR is a follow-up to #6733. It implements step 4 of the Data Migration outlined in that PR:

> 4. Deploy code changes to use the new columns

I've removed the expensive database queries which were used to dynamically find the latest Edition of a Document, replacing them with the newly created and populated (in #6733) fields `latest_edition_id` and `live_edition_id`. I've also taken the opportunity to tidy up the various `latest_*` methods belonging to the Edition model, which were either duplicate implementations or have otherwise been made redundant.

It's probably easiest to review this PR one commit at a time. I've tried to keep each commit relatively focussed on implementing one particular change across the codebase.

## Summary of high level changes

| Method | Change made |
| --- | --- |
| `Document#latest_edition` | refactored to use new database column |
| `Document#live_edition` | refactored to use new database column |
| `Edition.latest_edition` | refactored to use new database column |
| `Edition#latest_edition` | removed in favour of `Document#latest_edition` |
| `Edition.latest_published_edition` | replaced by `Edition.live_edition` |
| `Edition#latest_published_edition` | removed |

Note: `#` signifies an instance method, where `.` represents a class method.

### `Edition#previous_edition` is a bit weird

I've noticed that the method `Edition#previous_edition` could do with some love. Its name is slightly misleading because it will only ever return the most recently live Edition of a Document (i.e. an edition that was published/withdrawn) **excluding the current Edition**. When dealing with the latest Edition, this method name makes sense. But when working with an older Edition, the method will actually return a _more recent_ Edition which seems counterintuitive.

Unfortunately `Edition#previous_edition` is not covered by Unit Tests, despite being used in several places throughout the codebase. So while its behaviour doesn't strictly match its name, it seems that various things depend on it working the way it currently does.

I'm open to suggestions about what to do about this. I guess options include:

- Do nothing (feels a bit rubbish)
- Rename the method to make its purpose more obvious (maybe even relocate it to the Document model?)
- Retrospectively add Unit Tests to make sure the current behaviour is maintained, even if it has a counterintuitive name

Or a combination of the above.

---

Trello: https://trello.com/c/iWcORsD5

### DIFF
--- a/app/helpers/public_document_routes_helper.rb
+++ b/app/helpers/public_document_routes_helper.rb
@@ -40,7 +40,7 @@ module PublicDocumentRoutesHelper
     if edition.rendering_app == Whitehall::RenderingApp::GOVERNMENT_FRONTEND
       options[:host] = URI(Plek.new.external_url_for("draft-origin")).host
     else
-      options[:preview] = edition.latest_edition.id
+      options[:preview] = edition.document.latest_edition_id
       options[:cachebust] = Time.zone.now.getutc.to_i
     end
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -555,10 +555,6 @@ EXISTS (
     end
   end
 
-  def latest_edition
-    document.editions.latest_edition.first
-  end
-
   def latest_published_edition
     document.editions.latest_published_edition.first
   end
@@ -572,7 +568,7 @@ EXISTS (
   end
 
   def is_latest_edition?
-    latest_edition == self
+    document.latest_edition == self
   end
 
   def all_nation_applicability_selected?
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -95,6 +95,8 @@ class Edition < ApplicationRecord
 
   scope :future_scheduled_editions,     -> { scheduled.where(Edition.arel_table[:scheduled_publication].gteq(Time.zone.now)) }
 
+  scope :latest_edition, -> { joins(:document).where("editions.id = documents.latest_edition_id") }
+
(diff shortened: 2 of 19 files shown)


app/models/edition.rb
```diff
@@ -247,15 +248,6 @@ def self.without_locked_documents
     .where.not("documents.locked = true")
   end
 
-  def self.latest_published_edition
```
davidgisbey, line 257: just in case scopes get misused in future i wonder if this could've just been 

```
def self.latest_published_edition
   joins(:document).where("documents.live_edition_id = editions.id AND editions.state = 'published'") }
end 
```

not big deal either way though!

app/helpers/public_document_routes_helper.rb
```diff
@@ -40,7 +40,7 @@ def preview_document_url(edition, options = {})
     if edition.rendering_app == Whitehall::RenderingApp::GOVERNMENT_FRONTEND
       options[:host] = URI(Plek.new.external_url_for("draft-origin")).host
     else
-      options[:preview] = edition.latest_edition.id
+      options[:preview] = edition.document.latest_edition.id
```
davidgisbey, line 43: Not sure if the edition is already in memory here, but if not then 

```
edition.document.latest_edition_id
```

would be quicker

ollietreend: Good spot – thanks. I've just changed that.

And this will hopefully go away altogether once frontend rendering has been removed from Whitehall ✨

davidgisbey (approved):

PR #6798, opened 2022-09-08
Add error summary component and error helper

davidgisbey:
## Description

We need to be able render an error summary component which renders above the title and link the errors to the relevant input. We also need to be able to easily render errors for fields in views. 

This PR addresses this by doing two things. Adding an error summary partial which can be used by passing in an object and an optional class name. 

These are used to render the errors and title in the component respectively.

I've also added an ErrorsHelper which can be used to render errors for fields on a form

This has two methods #errors_for_input and #errors_for.Publishing components handles errors differently for inputs and all
other components.

For inputs you pass it the object and it outputs a string

```

For all other fields it outputs an array of hashes such as:
[
  {
    text: "error message"
  },
  {
    text: "second error message"
  },
]
```

## Screenshot

image

### DIFF
--- a/app/views/shared/_error_summary.html.erb
+++ b/app/views/shared/_error_summary.html.erb
@@ -1,0 +1,12 @@
+<% if object.errors.present? %>
+  <%= render "govuk_publishing_components/components/error_summary", {
+    id: "error-summary",
+    title: "To save the #{object.class.name.demodulize.underscore.humanize.downcase} please fix the following issues:",
+    items: object.errors.errors.map do |error|
+        {
+          href:"##{object.class.to_s.underscore}_#{error.attribute.to_s}",
+          text: error.full_message,
+        }
+      end
+  } %>
+<% end %>
--- a/app/views/shared/_error_summary.html.erb
+++ b/app/views/shared/_error_summary.html.erb
@@ -1,7 +1,9 @@
+<% class_name = class_name.presence || object.class.name.demodulize.underscore.humanize.downcase %>
+
 <% if object.errors.present? %>
   <%= render "govuk_publishing_components/components/error_summary", {
     id: "error-summary",
-    title: "To save the #{object.class.name.demodulize.underscore.humanize.downcase} please fix the following issues:",
+    title: "To save the #{class_name} please fix the following issues:",
     items: object.errors.errors.map do |error|
         {
           href:"##{object.class.to_s.underscore}_#{error.attribute.to_s}",
--- a/app/views/shared/_error_summary.html.erb
+++ b/app/views/shared/_error_summary.html.erb
@@ -1,14 +1,28 @@
 <% class_name = class_name.presence || object.class.name.demodulize.underscore.humanize.downcase %>
 
 <% if object.errors.present? %>
-  <%= render "govuk_publishing_components/components/error_summary", {
-    id: "error-summary",
-    title: "To save the #{class_name} please fix the following issues:",
-    items: object.errors.errors.map do |error|
-        {
-          href:"##{object.class.to_s.underscore}_#{error.attribute.to_s}",
-          text: error.full_message,
-        }
-      end
-  } %>
+  <div class="gem-c-error-summary govuk-error-summary" data-module="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" id="error-summary" data-govuk-error-summary-module-started="true">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      To save the <%= class_name %> please fix the following issues:
+    </h2>
+
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <% object.errors.errors.map do |error| %>
+          <% analytics_action = "#{class_name}-error" %>
+          <% error_message = error.full_message %>
+
+          <%= tag.li(
+            tag.a(
+              error_message,
+              href: "##{object.class.to_s.underscore}_#{error.attribute.to_s}",
+              class: "govuk-link"
+            ),
+            data: track_analytics_data("form-error", analytics_action, error_message),
+            class: "gem-c-error-summary__list-item"
+          ) %>
+        <% end %>
+      </ul>
+    </div>
+  </div>
 <% end %>
--- a/test/unit/helpers/errors_helper_test.rb
+++ b/test/unit/helpers/errors_helper_test.rb
@@ -1,0 +1,50 @@
+require "test_helper"
+
+class ErrorsHelperTest < ActionView::TestCase
+  setup do
+    @object_with_no_errors = ErrorTestObject.new("title", Time.zone.today)
+    @object_with_errors = ErrorTestObject.new(nil, nil)
+    @object_with_errors.validate
+  end
+
+  test "#error_for_input returns nil when there are no error messages" do
+    assert_nil errors_for_input(@object_with_no_errors.errors, :title)
+  end
+
+  test "#error_for_input returns errors for the attribute passed in" do
+    assert_equal errors_for_input(@object_with_errors.errors, :title), "Title can't be blank"
+  end
+
+  test "#error_for_input formats the error message when there are multiple errors on a field" do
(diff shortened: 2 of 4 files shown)


test/unit/helpers/errors_helper_test.rb
```diff
@@ -0,0 +1,50 @@
...
+
+class ErrorsHelperTest < ActionView::TestCase
+  setup do
+    @object_with_no_errors = ErrorTestObject.new("title", Time.zone.today)
+    @object_with_errors = ErrorTestObject.new(nil, nil)
+    @object_with_errors.validate
+  end
+
+  test "#error_for_input returns nil when there are no error messages" do
+    assert_equal errors_for_input(@object_with_no_errors.errors, :title), nil
```
davidgisbey, line 11: Just realised this should use assert_nil. Will push up in a sec

app/views/shared/_error_summary.html.erb
```diff
@@ -0,0 +1,28 @@
+<% class_name = class_name.presence || object.class.name.demodulize.underscore.humanize.downcase %>
+
+<% if object.errors.present? %>
+  <div class="gem-c-error-summary govuk-error-summary" data-module="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" id="error-summary" data-govuk-error-summary-module-started="true">
```
ollietreend, line 4: It's a shame that you've had to switch out the partial for a manual HTML implementation of it. We'll fall out of sync if any changes are made in the components gem.

I'm tempted to suggest that we keep using the component from the gem, but additionally loop through the `object.errors` ourselves for the sake of GA tracking. But this alternative approach doesn't come without its own drawbacks.

For example, we'd need to create a bunch of hidden/empty HTML elements – one for each error – so that we can add:
```
data: track_analytics_data("form-error", analytics_action, error_message)
```

for the [AutoTrackEvent](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics/auto-track-event.js) JS module still kicks in and tracks the events for us.

Or alternatively we might end up needing to implement some custom JS which would track them for us.

e.g. an inline script which includes every error

```html
<script>
window.GOVUK.analytics.trackEvent(category, action, options)
</script>
```

### TL;DR

It doesn't feel great that we aren't using the error summary partial provided by the gem. But the other alternatives also feel a bit fiddly. 🤷🏻

app/views/shared/_error_summary.html.erb
```diff
@@ -0,0 +1,28 @@
...
+              class: "govuk-link"
+            ),
+            data: track_analytics_data("form-error", analytics_action, error_message),
+            class: "gem-c-error-summary__list-item"
+          ) %>
+        <% end %>
+      </ul>
+    </div>
+  </div>
+<% end %>
```
ollietreend, line 28: Re: test coverage

It's a shame we can't use [RSpec View specs](https://relishapp.com/rspec/rspec-rails/v/5-1/docs/view-specs/view-spec) here to test this partial. Here's a real-world [example from a previous project](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/d861ca773c8208168b5176065804bea3dac22541/spec/views/shared/case_type_badge.html.erb_spec.rb).

I tried looking to see if MiniTest has anything similar, but as far as I can tell the general approach is to [test views within controller tests](https://guides.rubyonrails.org/testing.html#testing-views) 😕

davidgisbey: I really strongly feel that we should use the view components gem. Here's a blog github wrote about them https://github.blog/2020-12-15-encapsulating-ruby-on-rails-views/

They're soooo fast compared to controller specs and far easier to use. They're amazing for a bunch if reasons, but in my experience they've led to far far better test coverage due to ease of use and speed.

It's so easy to unit test conditionals in views (which we have tonnes of)

it also leads to not having test files which are 500+ lines long.

They're amazing. They're the best thing i've encountered for improving a codebase (with a frontend) since becoming a dev.

app/views/shared/_error_summary.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<% if object.errors.present? %>
+  <div class="gem-c-error-summary govuk-error-summary" data-module="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" id="error-summary" data-govuk-error-summary-module-started="true">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      To save the <%= class_name %> please fix the following issues:
+    </h2>
+
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <% object.errors.errors.map do |error| %>
+          <% analytics_action = "#{object.class.name.demodulize.underscore.dasherize}-error" %>
```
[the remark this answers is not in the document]
davidgisbey: Yea good catch it def should be

app/views/shared/_error_summary.html.erb
davidgisbey: Yea the other option is me manually add it to each input for each view, but not all components in the publishing gems have the ability to it seems better to catch it here.

If the component did anything complex i'd agree, but it's something that i've tweaked myself pretty significantly to link errors correctly etc.

It really doesn't do a tonne tbh. 

The main thing is it fall out of sync with styling though. I'm not overly worried though tbh. Wdyt?

app/views/shared/_error_summary.html.erb
kevindew: @davidgisbey sorry I think I was typing as you were replying. 

Do you have ideas of how an adoption of ViewComponent would work? and/or how'd they compare to testing with ActionView::TestCase?

Seems like it'd be good to have a discussion early into this work where lots of views are changing as to whether we want to carry on as we are, adopt some degree of view testing or go all the way to view components.

davidgisbey: Yea i think that it makes sense to make any changes now as we're essentially going to eventually rebuild every view in Whitehall in the GOV.UK Design System.

They're not wildly dissimilar to using a view helper and  `ActionView::TestCase`, but imo it's a more comprehensive solution.  It's extremely cheap to test them. Each component encapsulates the required functionality which makes it really easy to use and significantly reduces cognitive complexity.  I've found it's particularly good at testing views with many conditionals

Here's a good example of why they're so great imo from one i previously built.

https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/components/candidate_interface/apply_again_call_to_action_component.html.erb

https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/components/candidate_interface/apply_again_call_to_action_component.rb

https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/spec/components/candidate_interface/apply_again_call_to_action_component_spec.rb

then it's called like this https://github.com/DFE-Digital/apply-for-teacher-training/blob/fe2f15fd513a531dbc5899505eb95c8716284258/app/components/candidate_interface/application_dashboard_component.html.erb#L11

There's so much great stuff about them. Would love to chat about it at some point.

Whichever way we go though i think we should probs decide quite quickly. Either solution will be far better than using view_tests in the controller though in terms of ease of use and performance.

kevindew: Great - do you want to have a think about putting in a chat in with Ollie and me to explore some ideas?

From my perspective, I like view components but worry about creating fresh inconsistencies and surprises, so we'd have to trade that off. Similarly I don't like it if we feel we need a cheap way to test views and don't have it.

davidgisbey: Sounds good i'll pop something in for tomorrow.

davidgisbey: We've discussed this. I'm going to spike it out and we're going to do a write up with recommendations

app/views/shared/_error_summary.html.erb
ollietreend: We've talked about this and agreed to go ahead with the approach as-is.

ollietreend (approved):

PR #6799, opened 2022-09-08
Add new translation page in bootstrap and design system

davidgisbey:
## Description

Currently, when a user wants to add a translation they are greeted with a bootstrap modal. They then select a language before seeing the edit page.

We want to get rid of this behaviour and replace the modal with a new translation endpoint.

Eventually we want to render this new view in the GOV.UK Design System. But for now it should render in Bootstrap.

This PR adds the new endpoint, adds views in Bootstrap and the GOV.UK Design System, and conditionally renders the view based on the "Preview design system permission".

### Before

image

#### Bootstrap

image

#### GOV.UK Design System

image

## Trello card 

https://trello.com/c/oiTIQR1G/637-add-new-translation-page

### DIFF
--- a/app/concerns/translation_controller_concern.rb
+++ b/app/concerns/translation_controller_concern.rb
@@ -3,7 +3,7 @@ module TranslationControllerConcern
 
   included do
     before_action :load_translatable_item
-    before_action :load_translated_models, except: [:index]
+    before_action :load_translated_models, except: %i[index new]
     helper_method :translation_locale
   end
 
--- a/app/controllers/admin/edition_translations_controller.rb
+++ b/app/controllers/admin/edition_translations_controller.rb
@@ -3,6 +3,8 @@ class Admin::EditionTranslationsController < Admin::BaseController
 
   before_action :forbid_editing_of_locked_documents
 
+  def new; end
+
   def update
     @translated_edition.change_note = "Added translation" if @translated_edition.change_note.blank?
     super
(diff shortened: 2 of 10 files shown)


ollietreend (commented):
I think you might have missed the 'clean up' step to remove the old modal code.

From what I can tell, this partial isn't needed any more:

https://github.com/alphagov/whitehall/blob/29aa06d76ddce59753d5e077d666b6336908a5e6/app/views/admin/editions/_show_metadata.html.erb#L91-L93

And we should be able to remove:
- `app/views/admin/shared/_translation_modal.html.erb`

Nice work making this both Bootstrap and Design System 👍🏻✨

ollietreend:
@davidgisbey Ah – in that case, I think we might need to:

189665344-c5147971-d701-428f-a698-66c57e51914d

### Remove

The 2nd search result in your screenshot (Editions):

https://github.com/alphagov/whitehall/blob/29aa06d76ddce59753d5e077d666b6336908a5e6/app/views/admin/editions/_show_metadata.html.erb#L91-L93

I believe this is no longer needed – since this PR replaces the modal with the two new views.

### Keep

But keep the 1st and 3rd search results in your screenshot (Contacts and Worldwide Offices), plus the partial file itself.

https://github.com/alphagov/whitehall/blob/13b467fd18ea8df751e32c3eacc0b27814988962/app/views/admin/contacts/_contact.html.erb#L42-L45

https://github.com/alphagov/whitehall/blob/13b467fd18ea8df751e32c3eacc0b27814988962/app/views/admin/worldwide_offices/_worldwide_office.html.erb#L54-L57

Is that right?

davidgisbey:
oh right cool sorry i totally missed what you meant. Have done 👍  thanks

ollietreend (approved):
Looks good 👍🏻

PR #6807, opened 2022-09-12
Don't flood Publishing API when ordering features

ollietreend:
We noticed some undesirable behaviour which resulted in excessive calls being made to the Publishing API.

When re-ordering the 'features' (featured documents) associated with an organisation, we found that duplicate requests were being made to re-publish the organisation page – 'N' requests (where 'N' is the number of features on the organisation) instead of just 1.

This was happening as a result of the ActiveRecord 'after save' callback on the Feature model. When re-ordering, every Feature in a FeatureList gets an updated sort order – so each one gets saved in sequence, resulting in numerous executions of the 'after save' republish callback.

Thankfully there aren't many places in the code that create/save/delete Features, so I've moved the re-publish call out of the 'after save' callback and have moved it to the relevant places higher up in the code.

---

Trello: https://trello.com/c/LR7xngsx/737-stop-whitehall-flooding-the-publishing-api-when-organisation-features-are-re-ordered


kevindew (approved):
Nice one, looks good to me 👍 

As an aside, I notice we have the condition: `if featurable.is_a?(Organisation)` is there something that presents these when this is not true, or are they things not pushed to Publishing API?

ollietreend:
@kevindew Good question.

So, the 'featurable' association is polymorphic:

```ruby
FeatureList.all.pluck(:featurable_type).uniq
=> ["Organisation", "WorldLocation"]
```

As for World Locations, I don't know exactly where or how they utilise Feature Lists. I believe the Platform team are looking at these at the moment, so hopefully they'll know better than me.

PR #6822, opened 2022-09-14
Remove `WorldLocationType` model

brucebolt:
We are currently using the `world_location_type_id`  column to act as a key for the entries in the WorldLocationType model, which only contains 2 entries and behaves exactly like an enum.

Therefore we can actually use an enum here, which allows us to reduce a lot of code complexity.

The key numbering is maintained in line with the existing WorldLocationType entries.

This also automatically gives us some new helper methods on the WorldLocation model: `world_location_types`, `world_location?` and `international_delegation?`.

[Trello card](https://trello.com/c/32SGcOpZ)


app/models/world_location.rb
```diff
@@ -104,7 +104,7 @@ def to_s
   end
 
   def self.all_by_type
-    ordered_by_name.group_by(&:world_location_type)
+    ordered_by_name.in_order_of(:world_location_type, %w[world_location international_delegation]).group_by(&:world_location_type)
```
BeckaL, line 107: Question out of interest more than anything - is it possible to re-use the enum here to define the ordering?

app/models/world_location_type.rb
```diff
@@ -1,24 +0,0 @@
-class WorldLocationType
```
BeckaL, line 1: 👍 there was so much unnecessary stuff in here! Really nice to get rid of this stuff

app/models/world_location.rb
brucebolt: I did wonder if we could do that, but decided against it as any future additions to the enum may not necessarily appear in the order we want them to be listed on the website.

app/models/world_location_type.rb
```diff
@@ -1,24 +0,0 @@
...
-  include ActiveRecordLikeInterface
-
-  attr_accessor :id, :name, :sort_order, :key
-
-  def slug
-    name.downcase.gsub(/[^a-z]+/, "-")
```
BeckaL, line 7: Leaving half a thought here before I have a bit more time to delve into this - was this slug used anywhere except in the find_by_slug below (and was that used anywhere)? Just want to make sure that this isn't somehow used to generate the slug of the overall document (I think not but haven't fully looked into this yet)

brucebolt: There's a `slug` database field in the `world_locations` table, so Rails automatically generates a `slug` method for these objects.

BeckaL (approved):
All looks sensible to me, and a super nice simplification of the code. ⭐ 

Originally I was thinking we'd replace the numeric ids in the database with literally the text "world_location" or "international_delegation" but this is a way nicer / less sketchy solution. 

Hopefully it'll be a helpful thing to have in place when we do further refactoring on world locations / international delegations, but even if we don't get to that this is a big improvement 👍

PR #6823, opened 2022-09-15
Add basic design system page for new and edit pages

davidgisbey:
## Description

This is (finally 🎉 ) the first PR that starts to actually port stuff over to use the GOV.UK Design System. Going forward i'm going to go with the strategy of duplicating the new and edit page and slowly porting over the views one partial (or a couple if they're small) at a time so it can be reviewed fairly easily.

As i update a partial i will update any affected tests.

This PR deals with adding a `shell` for the new and edit edition pages. It only does the following

1. Adds a get layout method in the controller which renders the design system layout for the new, edit, create and update actions
2. Renders the `new` and `edit` template in place of the `new_legacy` and `edit_legacy` views when the user has the "Preview design system" permission
3. Adds a `new` and `edit` view which continues to render the legacy partials so all the tests keep passing
4. Adds a `shell` in the GOV.UK Design System (title, backlink, error summary etc.)

### New page

image

### Edit page

image

## Trello card


app/views/admin/editions/edit.html.erb
```diff
@@ -0,0 +1,38 @@
+<% content_for :page_title, "Edit publication"%>
+<% content_for :back_link, (render "govuk_publishing_components/components/back_link", {
+    href: admin_edition_path(@edition)
+  })
+%>
+<% content_for :error_summary, render('shared/error_summary', object: @edition.becomes(Edition), class_name: @edition.class.name.demodulize.underscore.humanize.downcase) %>
```
kevindew, line 6: Corr this is a mouthful! Perhaps there's a future scope for a user_facing_name method on Editions. Not in this PR of course as consistent with legacy view.

app/views/admin/editions/edit.html.erb
```diff
@@ -0,0 +1,38 @@
+<% content_for :page_title, "Edit publication"%>
```
kevindew, line 1: Is this meant to use a variable instead of "publication" ?

kevindew (approved):
Amazing stuff @davidgisbey - really great to see this coming to life

app/views/admin/editions/edit.html.erb
davidgisbey: I think this exists actually 🤔 

Let me find it and update it 👍

PR #6825, opened 2022-09-15
Final renaming and duplication pr for edit edition page

davidgisbey:
## Description

This is the final renaming and duplication PR that is required before i'm able to begin adding the edit edition page.

It looks **far** more scary than it will actually be to review due.

This PR does a few things

1. Renames  `AdminEditionControllerScheduledPublishingTestHelpers` to  `AdminEditionControllerLegacyScheduledPublishingTestHelpers`
3. Duplicates the 3 test helpers. These are identical except the method names aren't appended with `legacy`
4. Duplicates the edition controller specs for each type and adds the `Preview design system` permission to the user and calls the non-legacy shared examples

By essentially duplicating all the specs and fixing them as and when they fail while i'm porting the edit page to the GOV.UK Design System, i will be able to ensure that we don't lose any test coverage.

When a test fails i'll either update it to work with the GOV.UK Design System, remove it if it is no longer relevant. Or commnet  mark it with a todo if it is reliant on JS that hasn't been wired in yet.

## Guidance for review

Per commit will be easiest

All that's changed in the controller specs is the "Preview design system" being passed to the current user. It might be worth sense checking that 👍 

## Trello card

### DIFF
--- a/test/functional/admin/legacy_case_studies_controller_test.rb
+++ b/test/functional/admin/legacy_case_studies_controller_test.rb
@@ -1,24 +1,25 @@
 require "test_helper"
 
-class Admin::CaseStudiesControllerTest < ActionController::TestCase
+class Admin::LegacyCaseStudiesControllerTest < ActionController::TestCase
+  tests Admin::CaseStudiesController
+
   setup do
     login_as :writer
-    @current_user.permissions << "Preview design system"
   end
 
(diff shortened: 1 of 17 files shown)


kevindew (approved):
This looks good to me. 

Thanks for explaining that these are just duplicate files. I didn't go through them in detail as I fully trust they're just copies.

PR #6826, opened 2022-09-15
Republish world location news when a feature changes

brucebolt:
When a featured document changes, we need to republish the `WorldLocationNews` content to make this change live on the website.

As these pages were previously rendered by Whitehall, it did not matter that this didn't happen.  But now we render these pages in Collections using the content item, this update is essential.

[Trello card](https://trello.com/c/9FzBvL38/282-bug-featuring-a-document-on-world-location-news-doesnt-update-on-the-live-site)

### DIFF
--- a/app/controllers/admin/features_controller.rb
+++ b/app/controllers/admin/features_controller.rb
@@ -13,7 +13,7 @@ class Admin::FeaturesController < Admin::BaseController
     end
 
     if @feature.save
-      @feature.republish_organisation_to_publishing_api
+      @feature.republish_featurable_to_publishing_api
       PublishingApiDocumentRepublishingWorker.perform_async(@feature.document_id) if @feature.document_id.present?
       redirect_to admin_feature_list_path(@feature_list), notice: "The document has been saved"
     else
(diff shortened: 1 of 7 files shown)


app/controllers/admin/features_controller.rb
```diff
@@ -13,7 +13,7 @@ def create
     end
 
     if @feature.save
-      @feature.republish_organisation_to_publishing_api
+      @feature.republish_featurable_to_publishing_api
```
JonathanHallam, line 16: So when this happens are we republishing the featured thing or the world location news page that they're featured on? I'd hope there'd be no need for the former but 'featurable' is an ambiguous name...

brucebolt: The `featurable` is the page that the item is featured on (i.e. the World Location News or Organisation page).  I chose this name as it matches the database field.

JonathanHallam: oh yeah, I've seen it before, not a complaint about your use of it, more its lack of description

JonathanHallam (approved):
👍

PR #6829, opened 2022-09-15
Add similar slug warning in the design system

davidgisbey:
## Description

If an edition has the same title as another document then we show the user a warning which lets them know that they need to update the title.

After some discussion with the design team, we've decided that presenting this as an error in the error summary  that links to the title input, is preferable to having multiple banners.

### What it looked like

image

### What it looks like now

image

## Trello card 

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -205,7 +205,7 @@ private
     return "admin" unless preview_design_system_user?
 
     case action_name
-    when "edit", "update", "new", "create"
+    when "edit", "update"
       "design_system"
     else
       "admin"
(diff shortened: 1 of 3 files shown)


app/controllers/admin/editions_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::EditionsController < Admin::BaseController
   include HistoricContentConcern
+  include Admin::EditionsHelper
```
beccapearce, line 3: Just wondering why we need this added in?

beccapearce (approved):
Looks good just a question on something I didn't understand

app/controllers/admin/editions_controller.rb
davidgisbey: show_similar_slugs_warning?(@edition) is a helper method from that file. Maybe we should just extend the helpers to the controller though 🤔

PR #6842, opened 2022-09-21
Move subtype fields and locales to design system

davidgisbey:
## Description

This builds upon https://github.com/alphagov/whitehall/pull/6840 and adds subtype and locale fields.

### Publications

image

### News stories

image

### Speeches

image

### Locales 

image

## Trello card 

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -4,7 +4,6 @@ class Admin::EditionsController < Admin::BaseController
 
   before_action :remove_blank_parameters
   before_action :clean_edition_parameters, only: %i[create update]
-  before_action :infer_blank_checkbox_params, only: %i[create update]
   before_action :clear_scheduled_publication_if_not_activated, only: %i[create update]
   before_action :find_edition, only: %i[show show_locked edit update revise diff confirm_destroy destroy update_bypass_id history]
   before_action :prevent_modification_of_unmodifiable_edition, only: %i[edit update]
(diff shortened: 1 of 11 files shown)


app/controllers/admin/editions_controller.rb
```diff
@@ -426,6 +427,12 @@ def remove_blank_parameters
     params.reject! { |_, value| value.blank? }
   end
 
+  def infer_blank_checkbox_params
+    return if edition_params.empty? || !preview_design_system_user?
+
+    edition_params[:political] = "0" if edition_params[:political].blank?
+  end
```
ollietreend, line 434: > The checkbox component does not pass a value via a hidden field when unchecked.

Could this actually be due to `remove_blank_parameters` which runs before anything else?

https://github.com/alphagov/whitehall/blob/d10c50548b3bca46df84d8434a56a26b10464d77/app/controllers/admin/editions_controller.rb#L426-L428

Perhaps a cleaner approach would be to avoid removing unchecked checkboxes in the first place, rather than removing them (`remove_blank_parameters`) and then adding them back in again (`infer_blank_checkbox_params`).

I guess something like this would probably do the trick, although there might be a nicer way to do it:

```ruby
def remove_blank_parameters
  checkbox_fields = %i[political]
  params.reject! { |key, value| value.blank? && !checkbox_fields.include?(key) }
end
```

ollietreend (commented):
This looks good – just one comment

ollietreend (approved):
Nice work 🚀 thanks for those tweaks

PR #6844, opened 2022-09-21
Fix display_type for languages which don't have singular nouns

BeckaL:
We were previously losing translations and defaulting to English for the edition display type for languages which don't tend to have singular nouns, e.g. Mandarin, since we were specifically looking for the translation for 'one', whereas for some languages only a translation for 'other' exists. This change uses I18n's internal logic for deciding which translation (/) to look for, depending on the locale, so we shouldn't be losing translations for display types now.

We know this was causing issues with world location news page translations, but the logic exists for a multitude of document types, so this fixes for all these cases.

[Trello:](https://trello.com/c/DSL6gfzK/288-fix-issue-with-mandarin-document-types-not-being-pulled-through-in-correct-language)

We will also need to republish affected documents after this in order to see this come into effect.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -577,7 +577,7 @@ EXISTS (
   end
 
   def display_type
-    I18n.t("document.type.#{display_type_key}.one")
+    I18n.t("document.type.#{display_type_key}", count: 1)
   end
 
   def display_type_key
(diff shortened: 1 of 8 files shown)


brucebolt (approved):
Looks good, thanks for extending the scope to cover all instances of this problem ⭐

PR #6846, opened 2022-09-21
Update slugs of probation offices

beccapearce:
A request has come in to update the slugs of these probation offices.

https://trello.com/c/UbISO2vY/738-4993193-slug-changes-advice-on-content-ministry-of-justice-moj

Trello card: 

### DIFF
--- a/db/data_migration/20220920150937_reslug_probation_offices.rb
+++ b/db/data_migration/20220920150937_reslug_probation_offices.rb
@@ -1,0 +1,147 @@
+slug_changes = [
+  {
+    slug: "kirklees-kirklees-probation-centre",
+    new_slug: "kirklees-dewsbury-probation-office",
+  },
+  {
+    slug: "hambleton-essex-lodge",
+    new_slug: "hambleton-northallerton-probation-office",
+  },
+  {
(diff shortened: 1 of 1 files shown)


db/data_migration/20220920150937_reslug_probation_offices.rb
```diff
@@ -0,0 +1,147 @@
...
+  {
+    slug: "haringey-haringey-probation-office",
+    new_slug: "haringey-lansdowne-road-probation-office",
+  },
+]
+
+slug_changes.each do |slug_change|
+  document = Document.find_by(slug: slug_change[:slug])
+
+  edition = document.editions.published.last
```
ollietreend, line 139: A safer way to get the live edition is to use:

```suggestion
  edition = document.live_edition
```

There's a subtle distinction because the 'currently live' edition can be either in the state 'published' or 'withdrawn'. (Since withdrawn documents are still 'live', they just have a withdrawal notice at the top of the page.)

Although in this scenario it probably doesn't make any difference – I expect they're all 'published'.

ollietreend (approved):
Looks good to me 👍🏻

PR #6853, opened 2022-09-23
Move bulk upload attachments page to design system

baisa:
[Trello card](https://trello.com/c/UCQ6nWML/613-move-bulk-upload-attachments-page-to-the-govuk-design-system)

These changes port the "Bulk upload from Zip file" pages to the Design System. 

There are two views, steps one and two below. 

Also shown is the second step submitted with the form inputs left blank. Only the titles are verified and a single generic error message is displayed at the top of the page. This is the current behaviour, which doesn't match other areas where a more strict verification is carried out.

### Step One 

Screenshot 2022-09-29 at 14 44 01

---

### Step Two

Screenshot 2022-09-29 at 14 47 13

---

### Submitted with blank fields

Screenshot 2022-09-29 at 14 58 19

### DIFF
--- a/app/controllers/admin/bulk_uploads_controller.rb
+++ b/app/controllers/admin/bulk_uploads_controller.rb
@@ -31,21 +31,14 @@ class Admin::BulkUploadsController < Admin::BaseController
     if @bulk_upload.save_attachments
       redirect_to admin_edition_attachments_path(@edition)
     else
-      render :set_titles
+      render(preview_design_system_user? ? "set_titles" : "set_titles_legacy")
     end
   end
 
 private
 
   def get_layout
-    return "admin" unless preview_design_system_user?
-
-    case action_name
-    when "new", "upload_zip"
-      "design_system"
-    else
-      "admin"
-    end
+    preview_design_system_user? ? "design_system" : "admin"
   end
 
   def find_edition
--- a/app/controllers/admin/bulk_uploads_controller.rb
+++ b/app/controllers/admin/bulk_uploads_controller.rb
@@ -3,9 +3,11 @@ class Admin::BulkUploadsController < Admin::BaseController
   before_action :limit_edition_access!
   before_action :enforce_permissions!
   before_action :prevent_modification_of_unmodifiable_edition
+  layout :get_layout
 
   def new
     @zip_file = BulkUpload::ZipFile.new
+    render(preview_design_system_user? ? "new" : "new_legacy")
   end
 
   def upload_zip
@@ -14,9 +16,9 @@ class Admin::BulkUploadsController < Admin::BaseController
     if @zip_file.valid?
       @bulk_upload = BulkUpload.from_files(@edition, @zip_file.extracted_file_paths)
       @zip_file.cleanup_extracted_files
-      render :set_titles
+      render(preview_design_system_user? ? "set_titles" : "set_titles_legacy")
     else
-      render :new
+      render(preview_design_system_user? ? "new" : "new_legacy")
     end
   end
 
@@ -34,6 +36,17 @@ class Admin::BulkUploadsController < Admin::BaseController
   end
 
 private
+
+  def get_layout
+    return "admin" unless preview_design_system_user?
+
+    case action_name
+    when "new", "upload_zip"
+      "design_system"
+    else
+      "admin"
+    end
+  end
 
   def find_edition
     @edition = Edition.find(params[:edition_id])
--- a/app/views/admin/attachments/_form_legacy.html.erb
+++ b/app/views/admin/attachments/_form_legacy.html.erb
@@ -34,7 +34,7 @@
   <% elsif attachment.is_a?(ExternalAttachment) %>
     <%= form.text_field :external_url %>
   <% else %>
-    <%= render 'attachment_data_fields', form: form %>
+    <%= render 'attachment_data_fields_legacy', form: form %>
   <% end %>
 
   <%= hidden_field_tag :type, params[:type] %>
--- a/app/views/admin/attachments/_hoc_reference_fields.html.erb
+++ b/app/views/admin/attachments/_hoc_reference_fields.html.erb
@@ -2,7 +2,7 @@
   <%= render "govuk_publishing_components/components/radio", {
     heading: "House of Commons paper number",
     heading_level: 2,
-    heading_size: "l",
+    heading_size: heading_size,
     name: "attachment[unnumbered_hoc_paper]",
     id_prefix: "hoc_paper_number_conditional",
     items: [
@@ -46,7 +46,7 @@
     name: "attachment[parliamentary_session]",
     id: "attachment_parliamentary_session",
     heading_level: 2,
-    heading_size: "l",
+    heading_size: heading_size,
     options: options,
     error_message: errors_for_input(attachment.errors, :parliamentary_session)
   } %>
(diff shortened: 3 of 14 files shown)


app/views/admin/attachments/_form_legacy.html.erb
```diff
@@ -8,7 +8,7 @@
   <% end %>
 
   <% if attachable.allows_attachment_references? %>
-    <%= render 'reference_fields_legacy', attachable: attachable, form: form, attachment: attachment %>
+    <%= render 'reference_fields_legacy', attachable: attachable, form: form, attachment: attachment, root: nil %>
```
davidgisbey, line 11: It might be easier to pass in a heading size instead of root? i.e. 

```
<%= render 'reference_fields_legacy', attachable: attachable, form: form, attachment: attachment, heading_size: "m" %>
```

app/views/admin/attachments/_hoc_reference_fields.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= render "govuk_publishing_components/components/radio", {
     heading: "House of Commons paper number",
     heading_level: 2,
-    heading_size: "l",
+    heading_size: root == "bulk" ? "m" : "l",
```
davidgisbey, line 5: then 
```
heading_size: heading_size
```

app/views/admin/bulk_uploads/set_titles.html.erb
```diff
@@ -1,34 +1,65 @@
...
-<div class="row">
-  <section class="col-md-8">
-    <h1>Set titles for your attachments</h1>
-    <%= form_for @bulk_upload, url: admin_edition_bulk_uploads_path(@edition) do |form| %>
-      <%= form.errors %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="govuk-!-margin-bottom-8">
+      <%= form_for @bulk_upload, url: admin_edition_bulk_uploads_path(@edition) do |form| %>
+        <% content_for :error_summary, render('shared/error_summary', object: @bulk_upload, parent_class: "bulk_upload", class_name: @bulk_upload.class.name.demodulize.underscore.humanize.downcase) %>
```
ollietreend, line 14: The validation is slightly odd on this form.

It looks like it doesn't actually show error messages for each invalid form field – just a generic "fill in missing fields" prompt.

Error message when adding attachment details

We're using the 'error summary' component here, meaning the error messages should be clickable and refer directly to the invalid input fields. But in this case, clicking the error message does nothing – it adds `#bulk_upload_base` to the URL, but there isn't an element present on the page with that ID.

This feels wrong, but I'm not sure what the best course of action is here. Do we spend time improving validation, or accept that it's functionally the same as what exists in Bootstrap and move on? That feels like a product decision to me.

app/views/admin/bulk_uploads/set_titles.html.erb
```diff
@@ -1,34 +1,65 @@
-<% page_title "Bulk upload for: #{@edition.title}" %>
+<% content_for :page_title, "Bulk upload for: #{@edition.title}" %>
+<% content_for :title, "Edit file attachment" %>
```
ollietreend, line 2: Is this the right title? This page is for multiple attachments, so I imagine it should at least be:

```suggestion
<% content_for :title, "Edit file attachments" %>
```

It's worth noting the Bootstrap design uses the title "Set titles for your attachments"

ollietreend (commented):
This is looking good, but I've added a few comments on bits that I think might need some attention.

app/views/admin/attachments/_form_legacy.html.erb
davidtrussler: Cheers both, thought that was looking a bit clunky. I've updated it in all the files that need it; think it looks a bit more elegant now.

app/views/admin/attachments/_hoc_reference_fields.html.erb
davidtrussler: 👍

app/controllers/admin/bulk_uploads_controller.rb
```diff
@@ -35,6 +37,17 @@ def create
...
+  def get_layout
+    return "admin" unless preview_design_system_user?
+
+    case action_name
+    when "new", "upload_zip", "create"
+      "design_system"
+    else
+      "admin"
+    end
+  end
```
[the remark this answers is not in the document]
davidtrussler: Done 👍

app/views/admin/bulk_uploads/new.html.erb
```diff
@@ -1,17 +1,25 @@
-<% page_title "Bulk upload for: #{@edition.title}" %>
+<% content_for :page_title, "Bulk upload for: #{@edition.title}" %>
+<% content_for :title, "Bulk upload" %>
+<% content_for :context, "Attachments for publication" %>
```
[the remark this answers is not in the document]
davidtrussler: Yes, good point, updated as suggested. 👍

app/views/admin/bulk_uploads/new.html.erb
```diff
@@ -1,17 +1,25 @@
...
+    href: admin_edition_attachments_path(@edition)
+  } %>
+<% end %>
 
-<div class="row">
-  <section class="col-md-8">
-    <h1>Bulk upload for: <%= @edition.title %></h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @zip_file, url: upload_zip_admin_edition_bulk_uploads_path(@edition), multipart: true do |form| %>
```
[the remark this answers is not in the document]
davidtrussler: Oh yes, so it is. Error messaging has been added.

app/views/admin/bulk_uploads/set_titles.html.erb
davidtrussler: Yes, I discussed this with Carlos at the time and we decided to go with it as it is for now (see comments on [card](https://trello.com/c/UCQ6nWML/613-move-bulk-upload-attachments-page-to-the-govuk-design-system)). As far as I am aware the multiple files aspect of this causes some problems with the error messaging as it exists. I also started writing a new ticket to cover this in future. I'll make sure that exists before moving on from this ticket.

test/functional/admin/bulk_uploads_controller_test.rb
```diff
@@ -105,11 +105,13 @@ def post_to_upload_zip(filename)
...
-  view_test "POST :create with invalid attachments re-renders the bulk upload form" do
+  view_test "POST :create with invalid attachments re-renders the bulk upload form with design system permission" do
+    @current_user.permissions << "Preview design system"
+
     post :create, params: { edition_id: @edition, bulk_upload: invalid_create_params }
 
     assert_response :success
-    assert_select ".form-errors", text: /enter missing fields/
+    assert_select ".govuk-error-summary", text: /enter missing fields/
   end
```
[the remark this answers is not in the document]
davidtrussler: Have retained Bootstrap test in addition to Design System 👍

ollietreend (commented):
This looks great @davidtrussler ✨🚀 Thanks for all the tweaks following my review!

Final thing before this is ready to merge: it'd be great to tidy up the commit history a little. I suspect we could probably get away with just a small handful of commits to cover this work – without individual commits for each of the tweaks made along the way.

ollietreend (approved):

PR #6856, opened 2022-09-23
Port new/edit html attachment pages to the Design System

davidtrussler:
The work in this PR moves the add/edit HTML attachment pages to the design system layout. See the [Trello card](https://trello.com/c/VocALuuZ/715-move-new-edit-html-attachment-page-to-the-govuk-design-system) for this change. Both new and edit pages are essentially the same and share the same components. Only the edit page is shown in the screenshots below, which show the full page and some details of expanded sections etc. 

There are a couple of areas where this layout differs from the design so we'll need to discuss how to proceed in these areas:
- the column headed "Formatting" is aligned at the top with the column headed "Title (required)" rather than aligning with the "Back" link as in the design. This is a result of the generic template we are using for all these pages so it may not be possible or desirable to change that. 
- the "List of words to avoid" link is rendered with the "Details" component which shows/hides the list rather than the link to another page that is shown in the design. 

### Full page layout
Screenshot 2022-09-23 at 14 48 45

---

### "Command paper number" and "House of Commons paper number" sections with "Numbered" option selected
Screenshot 2022-09-23 at 15 18 36

---

### "Guidance on manually numbered headings" section expanded
Screenshot 2022-09-23 at 15 21 26

---

### Sidebar: "Paste and convert to Markdown"
Screenshot 2022-09-26 at 12 33 12

---

### Sidebar: "Headings"
Screenshot 2022-09-26 at 12 36 24

---

### Sidebar: "Video links"
Screenshot 2022-09-26 at 12 39 18

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -3,6 +3,7 @@
 @import "./components/autocomplete";
 @import "./components/govspeak-editor";
 @import "./components/miller-columns";
+@import "./admin/govspeak-help";
 
 @import "./admin/modules/unpublish-display-conditions";
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -3,7 +3,7 @@
 @import "./components/autocomplete";
 @import "./components/govspeak-editor";
 @import "./components/miller-columns";
-@import "./admin/govspeak-help";
+@import "./admin/views/govspeak-help";
 
 @import "./admin/modules/unpublish-display-conditions";
 
--- a/app/helpers/admin/sidebar_helper.rb
+++ b/app/helpers/admin/sidebar_helper.rb
@@ -1,15 +1,22 @@
 module Admin::SidebarHelper
   def simple_formatting_sidebar(options = {})
-    sidebar_tabs govspeak_help: "Help" do |tabs|
-      tabs.pane id: "govspeak_help", class: "govspeak_help" do
-        tab_content = []
-        tab_content << render("admin/editions/govspeak_help", options)
-        tab_content << render("admin/editions/words_to_avoid_guidance")
-        tab_content << tag.h3("Style", class: "style-title")
-        tab_content << tag.p do
-          raw %(For style, see the #{link_to('style guide', 'https://www.gov.uk/guidance/style-guide')})
+    if current_user.has_permission? "Preview design system"
+      sidebar_content = []
+      sidebar_content << render("admin/editions/govspeak_help", options)
+      sidebar_content << render("admin/editions/style_guidance", options)
+      raw sidebar_content.join("\n")
+    else
+      sidebar_tabs govspeak_help: "Help" do |tabs|
+        tabs.pane id: "govspeak_help", class: "govspeak_help" do
+          tab_content = []
+          tab_content << render("admin/editions/govspeak_help_legacy", options)
+          tab_content << render("admin/editions/words_to_avoid_guidance")
+          tab_content << tag.h3("Style", class: "style-title")
+          tab_content << tag.p do
+            raw %(For style, see the #{link_to('style guide', 'https://www.gov.uk/guidance/style-guide')})
+          end
+          raw tab_content.join("\n")
         end
-        raw tab_content.join("\n")
       end
     end
   end
--- a/app/views/admin/attachments/_form.html.erb
+++ b/app/views/admin/attachments/_form.html.erb
@@ -23,23 +23,36 @@
 
   <% if attachment.is_a?(HtmlAttachment) %>
     <%= form.fields_for :govspeak_content do |govspeak_fields| %>
-
-      <%= govspeak_fields.check_box :manually_numbered_headings %>
-      <div class="js-manual-numbering-help<%= ' js-hidden' unless attachment.manually_numbered_headings? %>">
-        <pre>
-  ## 1. First heading
-  ### 1.1 First sub-heading
-  ### 1.2 Second sub-heading
-  ### Unnumbered sub-heading
-
-  Manually number your headings as appropriate using the above numbering scheme.</pre>
+      <div class="govuk-!-margin-bottom-8">
+        <%= render "govuk_publishing_components/components/checkboxes", {
+          name: "attachment[govspeak_content_attributes][manually_numbered_headings]",
+          heading: "Manually numbered headings",
+          heading_level: 2,
+          heading_size: "l",
+          hint_text: "",
+          items: [
+            {
+              label: "Use manually numbered headings",
+              value: "1",
+              checked: form.object.govspeak_content.manually_numbered_headings,
+              conditional: sanitize("<p>Manually number your headings using the numbering scheme below.</p><code>## 1. First heading<br />### 1.1 First sub-heading<br />### 1.2 Second sub-heading<br />### Unnumbered sub-heading</code>")
+            }
+          ]
+        } %>
       </div>
 
-      <%= content_tag :fieldset, class: ("right-to-left" if form.object.rtl_locale?) do %>
-        <%= govspeak_fields.text_area :body, rows: 20, class: "previewable", data: {
-          module: "paste-html-to-govspeak"
-        } %>
-      <% end %>
+      <%= render "govuk_publishing_components/components/textarea", {
+        label: {
+          heading_size: "l",
+          text: "Body (required)"
+        },
+        name: "attachment[govspeak_content_attributes][body]",
+        rows: 20,
+        id: "attachment_govspeak_content.body",
+        value: form.object.govspeak_content.body,
+        error_items: errors_for(attachment.errors, :"govspeak_content.body"),
+        margin_bottom: 8
+      } %>
     <% end %>
   <% elsif attachment.is_a?(ExternalAttachment) %>
     <div class="govuk-!-margin-bottom-8">
@@ -64,6 +77,10 @@
   <% if attachment.is_a?(ExternalAttachment) %>
     <%= render "govuk_publishing_components/components/button", {
       text: "Next"
+    } %>
+  <% elsif attachment.is_a?(HtmlAttachment) %>
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Save"
     } %>
   <% else %>
     <%= render "govuk_publishing_components/components/button", {
--- a/app/views/admin/attachments/edit.html.erb
+++ b/app/views/admin/attachments/edit.html.erb
@@ -11,8 +11,8 @@
   </section>
 
   <% if attachment.html? %>
-    <div class="col-md-4">
+    <aside class="govuk-grid-column-one-third">
       <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true %>
-    </div>
+    </aside>
   <% end %>
 </div>
--- a/app/views/admin/attachments/new.html.erb
+++ b/app/views/admin/attachments/new.html.erb
@@ -11,8 +11,8 @@
   </section>
 
   <% if attachment.html? %>
-    <div>
+    <aside class="govuk-grid-column-one-third">
       <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true %>
-    </div>
+    </aside>
   <% end %>
 </div>
--- a/app/views/admin/editions/_style_guidance.html.erb
+++ b/app/views/admin/editions/_style_guidance.html.erb
@@ -1,0 +1,53 @@
+<h2 class="govuk-heading-l">Style</h2>
+
+<p class="govuk-body">For styles, see the <%= link_to "style guide", "https://www.gov.uk/guidance/style-guide", class: "govuk-link" %>
+
+<p class="govuk-body">For details, see the <%= link_to "guideline on using plain English", "https://www.gov.uk/guidance/content-design/writing-for-gov-uk#plain-english", class: "govuk-link" %>
+
+<%= render "govuk_publishing_components/components/details", {
+  title: "List of words to avoid"
+} do %>
+  <ul class='gem-c-list govuk-list govuk-list--bullet'>
+    <li>agenda (unless it’s for a meeting)</li>
+    <li>advancing</li>
+    <li>collaborate (use ‘working with’)</li>
+    <li>combating</li>
+    <li>commit/pledge (we need to be more specific – we’re either doing something or we’re not)</li>
+    <li>countering</li>
+    <li>deliver (pizzas, post and services are delivered – not abstract concepts like ‘improvements’ or ‘priorities’)</li>
+    <li>deploy (unless it’s military or software)</li>
+    <li>dialogue (we speak to people)</li>
+    <li>disincentivise (and incentivise)</li>
+    <li>empower</li>
+    <li>facilitate (instead, say something specific about how you are helping)</li>
+    <li>focusing</li>
+    <li>foster (unless it’s children)</li>
+    <li>impact (as a verb)</li>
+    <li>initiate</li>
+    <li>key (unless it unlocks something. A subject/thing isn’t ‘key’ – it’s probably ‘important’)</li>
+    <li>land (as a verb. Only use if you are talking about aircraft)</li>
+    <li>leverage (unless in the financial sense)</li>
+    <li>liaise</li>
+    <li>overarching</li>
+    <li>progress (as a verb – what are you actually doing?)</li>
+    <li>promote (unless you are talking about an ad campaign or some other marketing promotion)</li>
+    <li>robust</li>
+    <li>slimming down (processes don’t diet – we are probably removing x amount of paperwork, etc)</li>
+    <li>streamline</li>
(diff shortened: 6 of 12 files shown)


app/views/admin/attachments/_form.html.erb
```diff
@@ -23,23 +23,39 @@
...
   <% if attachment.is_a?(HtmlAttachment) %>
     <%= form.fields_for :govspeak_content do |govspeak_fields| %>
+      <div class="govuk-!-margin-bottom-8">
+        <%= render "govuk_publishing_components/components/checkboxes", {
+          name: "attachment[govspeak_content_attributes][manually_numbered_headings]",
+          items: [
+            {
+              label: "Manually numbered headings",
+              value: "1",
+              checked: form.object.content_id != nil
```
ollietreend, line 33: Is this the right condition for the checkbox to be checked? It looks like this will mean the box is always checked, so long as there's a content ID.

But I was expecting to see something more like this:

```
checked: form.object.govspeak_content.manually_numbered_headings
```

i.e. reading the value of the `manually_numbered_headings` field, not `content_id`.

app/views/admin/editions/_style_guidance.html.erb
```diff
@@ -0,0 +1,61 @@
+<h2 class="govuk-heading-l">Styling</h2>
+
+<p class="govuk-body">For styles, see the <%= link_to "style guide", "https://www.gov.uk/guidance/style-guide", class: "govuk-link" %>
```
ollietreend, line 3: Is this supposed to be here? I think it effectively repeats the `<p>` below it.

Cursor_and_Editing_publication_HTML_attachment_-_GOV_UK_Whitehall_Publisher

app/views/admin/editions/_style_guidance.html.erb
```diff
@@ -0,0 +1,61 @@
...
+
+<p class="govuk-body">For details, see the <%= link_to "guideline on using plain English", "https://www.gov.uk/guidance/content-design/writing-for-gov-uk#plain-english"%>
```
ollietreend, line 5: This link is missing `class: "govuk-link"` – so it's ended up being a slightly different shade of blue 🙈 

```suggestion
<p class="govuk-body">For details, see the <%= link_to "guideline on using plain English", "https://www.gov.uk/guidance/content-design/writing-for-gov-uk#plain-english", class: "govuk-link" %>
```

app/helpers/admin/sidebar_helper.rb
```diff
@@ -1,15 +1,22 @@
...
   def simple_formatting_sidebar(options = {})
-    sidebar_tabs govspeak_help: "Help" do |tabs|
-      tabs.pane id: "govspeak_help", class: "govspeak_help" do
-        tab_content = []
-        tab_content << render("admin/editions/govspeak_help", options)
-        tab_content << render("admin/editions/words_to_avoid_guidance")
-        tab_content << tag.h3("Style", class: "style-title")
-        tab_content << tag.p do
-          raw %(For style, see the #{link_to('style guide', 'https://www.gov.uk/guidance/style-guide')})
+    if !current_user.has_permission? "Preview design system"
```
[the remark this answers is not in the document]
davidtrussler: Yes, I think it was set that way for dev purposes and got left in :face_palm: - now updated to be correct.

app/views/admin/attachments/_form.html.erb
davidtrussler: Yeah, this makes very little sense to me but I believe it's correct. The way Nikin described it to me (and the way it works currently) is that the box is checked only once the document has been saved and you return to edit it. No-one seems to know how this is useful.

app/views/admin/editions/_style_guidance.html.erb
davidtrussler: Oh yeah, class added 👍

app/assets/stylesheets/admin/_govspeak-help.scss
```diff
@@ -11,5 +11,10 @@
 
   pre {
     white-space: pre-wrap;
+    padding: govuk-spacing(2);
+    background-color: govuk-colour("light-grey");
+    word-break: normal;
+    word-wrap: normal;
+    overflow: auto;
```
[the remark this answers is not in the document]
davidtrussler: Yes, good point. Seems we have a new file for this purpose that will be rationalised at a later date. For now specific styles like this are in `app/assets/stylesheets/admin/views/_whats-new.scss`.

app/views/admin/editions/_style_guidance.html.erb
```diff
@@ -0,0 +1,61 @@
...
+  </ul>
+<% end %>
+
+<% if ENABLE_WORDS_TO_AVOID_HIGHLIGHTING %>
+  <% initialise_script "GOVUK.WordsToAvoidGuide",
+      el: ["form textarea[name='edition[summary]']", "form textarea[name='edition[body]']"],
+      wordsToAvoidList: "#js-words-to-avoid-list",
+      wordsToAvoidAlert: "#js-words-to-avoid-alert",
+      wordsToAvoidCounter: "#js-words-to-avoid-count" %>
+<% end %>
```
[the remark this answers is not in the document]
davidtrussler: Yes, that's correct. Unnecessary code has been removed.

app/views/admin/attachments/_form.html.erb
ollietreend: @davidtrussler I don't think it works as you've described – I have a feeling some wires have become crossed somewhere. It might be easier to talk this through together.

Looking through the code for the Bootstrap version of this page:
https://github.com/alphagov/whitehall/blob/049dac792091d3de01b479805ecf269cd50a4dd5/app/views/admin/attachments/_form_legacy.html.erb#L17-L26

It looks like there are a couple of things going on here:

1. The checkbox behaves like a 'normal' checkbox – i.e. it's 'checked' state is controlled by the user
    - if the user checks it & hits save, it'll stay checked when the page is reloaded
    - if they leave it unchecked & hit save, it'll remain unchecked
    - it defaults to being unchecked, and it'll never automatically check itself
2. There's some inline help which only shows when the checkbox is checked – i.e. the `<div class="js-manual-numbering-help...` is hidden if the user hasn't checked the checkbox.
    - this behaviour is implemented server-side with a conditional class:
      ```ruby
      <%= ' js-hidden' unless attachment.manually_numbered_headings? %>
      ```
    - JavaScript exists to implement this behaviour client-side, **but it's broken**.
      It targets `#attachment_manually_numbered_headings` but the correct ID to target is `#attachment_govspeak_content_attributes_manually_numbered_headings`
      https://github.com/alphagov/whitehall/blob/049dac792091d3de01b479805ecf269cd50a4dd5/app/assets/javascripts/admin/views/attachments/_form.js#L25-L29

You said:
> The way Nikin described it [...] is that the box is checked only once the document has been saved and you return to edit it.

I suspect the broken show/hide JavaScript is what led to this description – i.e. the inline help only appears after you've saved and returned to edit the HTML attachment. (Because that's when the Ruby conditional would kick in.)

But I think that's a separate issue to the first piece of behaviour – i.e. the checkbox being manually controlled by a user, rather than being automatically checked after the first save.

**TL;DR:** I still think this is the desired behaviour:

```suggestion
              checked: form.object.govspeak_content.manually_numbered_headings
```

app/assets/stylesheets/admin/_govspeak-help.scss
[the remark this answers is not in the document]
DilwoarH: Hey @davidtrussler - I think I may have not been clear when I said this is where we're storing view specific CSS.

What I meant was you should create a separate file like the `what-new.scss` file. However, looks more like a component style, therefore, I would recommend we put these in component folder in admin - `app/assets/stylesheets/admin/components/_govspeak-help.scss`.

I would also create CSS directly for the block you're trying to modify so `.app-c-govspeak_help__pre` rather than `.app-c-govspeak_help pre`

```
.app-c-govspeak_help__pre {
    padding: govuk-spacing(2);
    background-color: govuk-colour("light-grey");
    word-break: normal;
    word-wrap: normal;
    overflow: auto;
  }
}
```

FYI - I will be moving the files around so that we have the following CSS structure:

```
admin (for the new stuff)
  -> components
  -> views
  ...
admin_legacy (for all the old Bootstrap stuff)
  -> views
  -> helpers
  ...
```

I will be moving these things over in this card: https://trello.com/c/3UW25yLj/779-separate-the-asset-files-css-js-for-bootstrap-pages-vs-design-system-pages.

I hope this makes sense.

app/views/admin/attachments/_form.html.erb
```diff
@@ -23,23 +23,39 @@
...
+        } do %>
+          <p>Manually number your headings using the numbering scheme below.</p>
+          <code>## 1. First heading<br />### 1.1 First sub-heading<br />### 1.2 Second sub-heading<br />### Unnumbered sub-heading</code>
+        <% end %>
       </div>
 
-      <%= content_tag :fieldset, class: ("right-to-left" if form.object.rtl_locale?) do %>
-        <%= govspeak_fields.text_area :body, rows: 20, class: "previewable", data: {
-          module: "paste-html-to-govspeak"
+      <div class="govuk-!-margin-bottom-8">
```
DilwoarH, line 46: you can add the margin bottom flag on the text area: 

```
margin_bottom: 9,
```

app/views/admin/attachments/_form.html.erb
```diff
@@ -23,23 +23,39 @@
...
+        <%= render "govuk_publishing_components/components/textarea", {
+          label: {
+            heading_size: "l",
+            text: "Body (required)"
+          },
+          name: "attachment[govspeak_content_attributes][body]",
+          rows: 20,
```
DilwoarH, line 53:
```suggestion
          rows: 20,
          margin_bottom: 8,
```

app/views/admin/attachments/_form.html.erb
DilwoarH: https://components.publishing.service.gov.uk/component-guide/textarea/with_margin_bottom

app/views/admin/attachments/edit.html.erb
```diff
@@ -11,8 +11,8 @@
   </section>
 
   <% if attachment.html? %>
-    <div class="col-md-4">
+    <section class="govuk-grid-column-one-third">
```
DilwoarH, line 14: Maybe `aside` is a better tag (https://www.w3.org/html/wiki/Elements/aside)

```suggestion
    <aside class="govuk-grid-column-one-third">
```

app/views/admin/attachments/edit.html.erb
```diff
@@ -11,8 +11,8 @@
...
       <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true %>
-    </div>
+    </section>
```
DilwoarH, line 16:
```suggestion
    </aside>
```

app/views/admin/attachments/new.html.erb
```diff
@@ -11,8 +11,8 @@
   </section>
 
   <% if attachment.html? %>
-    <div>
+    <section class="govuk-grid-column-one-third">
       <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true %>
-    </div>
+    </section>
```
DilwoarH, line 16:
```suggestion
    </aside>
```

app/assets/stylesheets/admin/_govspeak-help.scss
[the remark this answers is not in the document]
DilwoarH: nice one

app/views/admin/attachments/_form.html.erb
davidtrussler: @ollietreend I've updated this section in [this commit](https://github.com/alphagov/whitehall/pull/6856/commits/625dd90fb39debef67aa7cda18c4b66fc9bbc1ac). There is though some weirdness going on with the checkbox component that won't show conditional items if there is no header or hint text. Whether this is a problem with the component or how it is supposed to work by design I'm not sure. 

Consequently I've implemented it as a kind of mash-up of the design, which I will also run by JP for comment as it may or may not be acceptable: for instance I think we might want different wording for "Guidance on manually numbered headings" if we go with this. 

Shown in the screenshots below compared to the supplied design. 

| Design | Unchecked | Checked |
| --- | --- | --- |
| Screenshot 2022-10-10 at 13 39 33 | Screenshot 2022-10-10 at 14 51 38 | Screenshot 2022-10-10 at 14 49 09 |

ollietreend: @davidtrussler Nice use of the conditional reveal functionality to avoid us having to write our own show/hide JS.

The need for a heading/hint is annoying – I guess it's because we're sort-of hacking the checkboxes component into serving our needs. Usually a heading would be required because the user would be choosing from a list of multiple items.

Just wondering: have you already tried setting the options `visually_hide_heading: true, no_hint_text: true`? These might do what we need [according to the docs](https://components.publishing.service.gov.uk/component-guide/checkboxes). But I'm not sure if they can both be used together?

Failing that, I wonder if we can get the best of both worlds by simply extracting the HTML we need from that component. Then we could hardcode the HTML into the view without using the partial – but still benefit from the conditional reveal functionality provided by JS.

What do you think?

app/views/admin/attachments/_form.html.erb
davidtrussler: 👍

app/views/admin/attachments/_form.html.erb
davidtrussler: 👍

app/views/admin/attachments/edit.html.erb
davidtrussler: Agreed - makes sense. 👍

app/views/admin/attachments/edit.html.erb
davidtrussler: 👍

app/views/admin/attachments/new.html.erb
```diff
@@ -11,8 +11,8 @@
   </section>
 
   <% if attachment.html? %>
-    <div>
+    <section class="govuk-grid-column-one-third">
```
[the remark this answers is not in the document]
davidtrussler: 👍

app/views/admin/attachments/new.html.erb
davidtrussler: 👍

app/views/admin/attachments/_form.html.erb
davidtrussler: Hello @ollietreend, as you are aware, the outcome of discussions with Carlos & JP (& yourself) is that we should go with this as above. 

Also, I'm going to move this ticket back to "In development" for the minute (and make this PR a draft) in order to deal with the right hand sidebar after the discussion earlier about using a different component in that.

davidtrussler:
There has been some discussion on the right hand column and the appropriate components to use. We have tried using the `details` component but decided to stick with the `accordion` component for this release, with the sidebar to be looked at later from the perspective of a better UI.

ollietreend (approved):
This looks good to me 👍🏻 

Nice work @davidtrussler.

app/assets/stylesheets/application.scss
```diff
@@ -2,6 +2,8 @@
 
 @import "./components/miller-columns";
 @import "./components/autocomplete";
+@import "./admin_legacy/govspeak-help";
```
DilwoarH, line 5: Hey - if we need anything from the legacy, we should copy it into `app/assets/stylesheets/admin/components/_govspeak-help.scss`

app/assets/stylesheets/admin/components/_govspeak-help.scss
```diff
@@ -0,0 +1,7 @@
+.govspeak_help__pre {
```
DilwoarH, line 1: As this isn't strictly a component, should we move it under `views`?

DilwoarH (changes requested):
just a few comments, nothing major - looks good otherwise

DilwoarH:
Also need to make this "Ready for review" :)

app/assets/stylesheets/application.scss
davidtrussler: No, none of the legacy stuff in this file is still needed for Design System.

DilwoarH: ah great so we don't need to import it then?

davidtrussler:
> just a few comments, nothing major - looks good otherwise

OK, I've dealt with those few things. The only thing I'm less sure about is whether we need to add anything for tracking - I'm not really sure what we're looking for when we port from Bootstrap to Design System.

app/assets/stylesheets/application.scss
```diff
@@ -2,6 +2,8 @@
...
+@import "./admin/views/govspeak-help";
```
ollietreend, line 6: @davidtrussler I think you only need one of these imports here, not both – i.e. not the `admin_legacy` one.

```suggestion
@import "./admin/views/govspeak-help";
```

davidtrussler: Yes, good point, have removed the unnecessary reference.

DilwoarH:
Looks good @davidtrussler

PR #6860, opened 2022-09-27
Add image fields partial in the design system

davidgisbey:
## Description

This adds in the image two partials, one for Case Studies and another for all other Edition types.

Currently, it works for uploading a single image, but still needs to work to allow multiple file uploads at the same time.

### Case studies

image

### Other editions 

image

## Trello card


app/views/admin/editions/_image_template.html.erb
```diff
@@ -0,0 +1,82 @@
...
+        <%= hidden_field_tag "edition[images_attributes][#{index}][_destroy]", "1" %>
+
+        <%= render "govuk_publishing_components/components/checkboxes", {
+          name: "edition[images_attributes][#{index}][_destroy]",
+          id: "edition_images_#{index}__destroy",
+          items: [
+            {
+              label: "(uncheck to remove)",
+              value: "0",
+              checked: true,
```
ollietreend, line 55: With the hardcoded `checked: true` option, I think this means the checkboxes will become re-checked if there are other validation errors on the page.

I think the user's expectation would be that the state of these checkboxes remains intact.

ollietreend (commented):
Nice work 👍🏻 this is looking good.

Just one comment from me.

app/views/admin/editions/_image_template.html.erb
davidgisbey: Good catch. Have fixed 👍 thanks for the review

ollietreend (approved):
Nice 👍🏻 looks good

PR #6861, opened 2022-09-27
Move Legacy redirect URLs pages to the GOV.UK Design System

dnkrj:

## What

Move the legacy urls page to the GOVUK design system layout.

## Why

This is part of a Whitehall transition work to move from bootstrap to GOVUK design system.

### After
whitehall-admin dev gov uk_government_admin_editions_1_document-sources_edit(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1_document-sources_edit(iPhone 12 Pro)

### Before
whitehall-admin dev gov uk_government_admin_editions_1_document-sources_edit(iPad Pro) (1)
whitehall-admin dev gov uk_government_admin_editions_1_document-sources_edit(iPhone 12 Pro) (1)

### DIFF
--- a/app/controllers/admin/document_sources_controller.rb
+++ b/app/controllers/admin/document_sources_controller.rb
@@ -2,8 +2,11 @@ class Admin::DocumentSourcesController < Admin::BaseController
   before_action :require_import_permission!
   before_action :find_edition
   before_action :forbid_editing_of_locked_documents
+  layout :get_layout
 
-  def edit; end
+  def edit
+    render :edit_legacy unless preview_design_system_user?
+  end
 
   def update
     @document_sources = params[:document_sources]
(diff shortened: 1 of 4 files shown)


davidgisbey:
Looks great 👍 Just wondering what happens if a user inputs invalid info? I might be being stupid, but i can't see any error handling at all?

dnkrj:
> Looks great 👍 Just wondering what happens if a user inputs invalid info? I might be being stupid, but i can't see any error handling at all?
> 
> I know it's pre existing behaviour but it's probs worth adding some. Wdyt?

As discussed, I had a look into the input validation and the only validation that occurs is the stripping of invalid URLs. It would be more sensible to inform the user when this occurs but seeing as this is the existing behaviour and a page that is not used very often we've decided to leave it as is.

davidgisbey (approved):
Great work. Have given it a manual test and it's working nicely.

PR #6862, opened 2022-09-27
Adds conditional reveal to "another website" checkbox

davidtrussler:
[Trello card](https://trello.com/c/OoA1ipPB/773-bug-missing-conditional-external-url-input-on-held-on-another-website-field)

This change restores a conditional input on the "Edit consultation" page which permanently hidden after porting to the Design System. This led to the user receiving an error message ("External URL can't be blank") if they checked the checkbox but they had no option to provide a value for the URL.

| Before | After |
| --- | --- |
| Screenshot 2022-09-27 at 15 45 43 | Screenshot 2022-09-27 at 15 44 36 |

### DIFF
--- a/app/views/admin/consultations/_form.html.erb
+++ b/app/views/admin/consultations/_form.html.erb
@@ -15,6 +15,7 @@
       heading: "Held on another website",
       heading_level: 3,
       heading_size: "l",
+      no_hint_text: true,
       error_items: errors_for(edition.errors, :external),
       items: [
         {
--- a/app/views/admin/consultations/_form.html.erb
+++ b/app/views/admin/consultations/_form.html.erb
@@ -16,7 +16,6 @@
       heading: "Held on another website",
       heading_level: 3,
       heading_size: "l",
-      hint_text: "",
       error_items: errors_for(edition.errors, :external),
       items: [
         {
--- a/app/views/admin/consultations/_form.html.erb
+++ b/app/views/admin/consultations/_form.html.erb
@@ -6,13 +6,12 @@
 
 <%= standard_edition_form(edition, @information, preview_design_system: true) do |form| %>
   <%= render "govuk_publishing_components/components/fieldset", {
-    legend_text: "",
+    legend_text: ""
   } do %>
     <%= form.hidden_field :external, value: "0" %>
 
     <%= render "govuk_publishing_components/components/checkboxes", {
       name: "edition[external]",
-      id: "edition_external",
       heading: "Held on another website",
       heading_level: 3,
       heading_size: "l",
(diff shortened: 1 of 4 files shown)


app/views/admin/consultations/_form.html.erb
```diff
@@ -6,36 +6,37 @@
 
 <%= standard_edition_form(edition, @information, preview_design_system: true) do |form| %>
   <%= render "govuk_publishing_components/components/fieldset", {
-    legend_text: "Held on another website",
-    heading_level: 3,
-    heading_size: "l"
+    legend_text: "",
```
DilwoarH, line 9: Is this needed?

app/views/admin/consultations/_form.html.erb
```diff
@@ -6,36 +6,37 @@
...
   } do %>
+    <%= form.hidden_field :external, value: "0" %>
+
     <%= render "govuk_publishing_components/components/checkboxes", {
       name: "edition[external]",
       id: "edition_external",
+      heading: "Held on another website",
+      heading_level: 3,
+      heading_size: "l",
+      hint_text: "",
```
DilwoarH, line 19:
```suggestion
```

app/views/admin/consultations/_form.html.erb
```diff
@@ -6,36 +6,35 @@
...
 <%= standard_edition_form(edition, @information, preview_design_system: true) do |form| %>
   <%= render "govuk_publishing_components/components/fieldset", {
-    legend_text: "Held on another website",
-    heading_level: 3,
-    heading_size: "l"
+    legend_text: "",
   } do %>
     <%= render "govuk_publishing_components/components/checkboxes", {
       name: "edition[external]",
       id: "edition_external",
```
DilwoarH, line 15: just drop this ID as we don't need it for anything so the JS won't have any effect

app/views/admin/consultations/_form.html.erb
davidtrussler: If `legend_text` is left out completely it fires an error in the component even though the `<legend>` is populated by the `heading` variable from the checkbox component. So it looks like we need it though it looks a bit odd. 

```
ActionView::Template::Error (undefined local variable or method `legend_text' for #<ActionView::Base:0x000000004e66c8>
Did you mean?  legend_classes):
    35:         %>
    36:       <% end %>
    37:     <% else %>
    38:       <%= tag.legend legend_text, class: legend_classes %>
    39:     <% end %>
    40:     <% if error_message %>
    41:       <%= render "govuk_publishing_components/components/error_message", {
```

app/views/admin/consultations/_form.html.erb
davidtrussler: 👍 Yes, we can do without that.

app/views/admin/consultations/_form.html.erb
davidtrussler: 👍 Good point - removed.

davidgisbey (approved):
lgtm

PR #6865, opened 2022-09-28
Move legacy worldwide organisation fields to the design system

davidgisbey:
## Description

This PR does two things

1. Adds the worldwide organisation input using the accessible autocomplete multi-select functionality  
2. Fixes a bug where the first_published_at field wasn't linking errors correctly when there was a blank inpt

I've had to comment out the tests around the implementation of tracking as the module needs to be rewritten to work with the accessible autocomplete.

The card for that can be found here https://trello.com/c/eKGeFCQu/975-add-tracking-in-for-associations-on-the-edit-page

### Before

image

### After 

image

image

### Error messages

Unfortunately the error message is a little clunkier due to the way the error summary component works but this will get ironed out at content review for 1b

#### Before

image

image

#### After

image

image

### DIFF
--- a/test/functional/admin/legacy_publications_controller_test.rb
+++ b/test/functional/admin/legacy_publications_controller_test.rb
@@ -83,7 +83,7 @@ class Admin::LegacyPublicationsControllerTest < ActionController::TestCase
 
   test "should validate previously_published field on create" do
     post :create, params: { edition: controller_attributes_for(:publication).except(:previously_published) }
-    assert_equal "You must specify whether the document has been published before", assigns(:edition).errors.full_messages.last
+    assert_equal "Previously published You must specify whether the document has been published before", assigns(:edition).errors.full_messages.last
   end
 
   test "should validate first_published_at field on create if previously_published is true" do
--- a/test/support/admin_edition_controller_test_helpers.rb
+++ b/test/support/admin_edition_controller_test_helpers.rb
@@ -1488,6 +1488,8 @@ module AdminEditionControllerTestHelpers
         get :new
 
         assert_select "form#new_edition" do
+          assert_select "label[for=edition_worldwide_organisation_ids]", text: "Worldwide organisations"
+
           assert_select "#edition_worldwide_organisation_ids" do |elements|
             assert_equal 1, elements.length
             assert_data_attributes_for_worldwide_organisations(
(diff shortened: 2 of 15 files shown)


test/functional/admin/legacy_publications_controller_test.rb
```diff
@@ -83,7 +83,7 @@ class Admin::LegacyPublicationsControllerTest < ActionController::TestCase
 
   test "should validate previously_published field on create" do
     post :create, params: { edition: controller_attributes_for(:publication).except(:previously_published) }
-    assert_equal "You must specify whether the document has been published before", assigns(:edition).errors.full_messages.last
+    assert_equal "Previously published You must specify whether the document has been published before", assigns(:edition).errors.full_messages.last
```
ollietreend, line 86: This is a bit rubbish 😢 although I don't know of an obvious way around it while maintaining the hyperlink in the error summary.

test/support/admin_edition_controller_test_helpers.rb
```diff
@@ -1609,10 +1613,10 @@ def assert_data_attributes_for_ministers(element:, track_label:)
...
 
   def assert_data_attributes_for_worldwide_organisations(element:, track_label:)
-    assert_equal "Worldwide organisations…", element["data-placeholder"]
-    assert_equal "track-select-click", element["data-module"]
-    assert_equal "worldwideOrganisationSelection", element["data-track-category"]
-    assert_equal track_label, element["data-track-label"]
+    # TODO: Add tracking back in. This is covered in this Trello card https://trello.com/c/eKGeFCQu/975-add-tracking-in-for-associations-on-the-edit-page
+    # assert_equal "track-select-click", element["data-module"]
+    # assert_equal "worldwideOrganisationSelection", element["data-track-category"]
+    # assert_equal track_label, element["data-track-label"]
```
ollietreend, line 1619: It doesn't feel great to be merging this in. But I'm not sure what to suggest in its place.

Does this mean the Accessible Autocomplete component doesn't support the tracking that this page needs? Should we consider this a blocker?

ollietreend:
@davidgisbey This PR looks good – but I've held off approving until you're back so we can decide what to do about my comments.

test/support/admin_edition_controller_test_helpers.rb
davidgisbey: Yea for all the associations the autocompletes aren't tracked. Dilwoar mentioned there's a module that'll need to be rewritten. There's a card tracking places it needs to be added here  https://trello.com/c/eKGeFCQu/975-add-tracking-in-for-associations-on-the-edit-page

davidgisbey:
As discussed we're going to merge with the TODOs. There is a card to reintroduce tracking in the backlog which i will pick up once all the PRs are merged in.

ollietreend (approved):

davidgisbey:
Thanks @ollietreend

PR #6867, opened 2022-09-28
Port remaining associations and attachments sections

davidgisbey:
## Description

This adds in all the association fields to the edit page. It largely uses the accessible autocomplete. However, there's an issue with it in that we've lost tracking. The JS module responsible will need to be tweaked. I've carded up affected fields in the backlog and commented out the tests with a TODO.

This also does the work to port across the HTML and inline attachments partials

## Screenshots

image

image

image

## Trello card

### DIFF
--- a/app/views/admin/editions/_appointment_fields.html.erb
+++ b/app/views/admin/editions/_appointment_fields.html.erb
@@ -1,4 +1,4 @@
-<% cache_if edition.role_appointment_ids.empty?, taggable_ministerial_role_appointments_cache_digest do %>
+<% cache_if edition.role_appointment_ids.empty?, "#{taggable_ministerial_role_appointments_cache_digest}-design-system" do %>
   <%= render "components/autocomplete", {
     id: "edition_role_appointment_ids",
     name: "edition[role_appointment_ids][]",
--- a/app/views/admin/editions/_appointment_fields.html.erb
+++ b/app/views/admin/editions/_appointment_fields.html.erb
@@ -1,0 +1,20 @@
+<% cache_if edition.role_appointment_ids.empty?, taggable_ministerial_role_appointments_cache_digest do %>
+  <%= render "components/autocomplete", {
+    id: "edition_role_appointment_ids",
+    name: "edition[role_appointment_ids][]",
+    label: {
+      text: "Ministers",
+      bold: true,
+    },
+    select: {
+      options: [""] + taggable_ministerial_role_appointments_container,
+      multiple: true,
+      selected: edition.role_appointment_ids,
+    },
+    data: {
+        module: 'track-select-click',
+        track_category: 'ministerSelection',
+        track_label: request.path
+    },
+  } %>
+<% end %>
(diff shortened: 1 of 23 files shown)


app/views/admin/editions/_appointment_fields.html.erb
```diff
@@ -0,0 +1,20 @@
+<% cache_if edition.role_appointment_ids.empty?, taggable_ministerial_role_appointments_cache_digest do %>
```
ollietreend, line 1: We may need to tweak how this caching works. Right now it looks like `appointment_fields` and `legacy_appointment_fields` will both cache to the same key.

So I think there's a good chance the rendered list will end up randomly switching between new and old design systems due to the cache key clash.

```suggestion
<% cache_if edition.role_appointment_ids.empty?, "#{taggable_ministerial_role_appointments_cache_digest}-design-system" do %>
```

Something like this might be the solution – just appending something to the end of the cache key so they don't clash.

app/views/admin/editions/_html_version_fields.html.erb
```diff
@@ -0,0 +1,12 @@
+<h3 class=govuk-heading-l>HTML attachment</h3>
```
ollietreend, line 1: This HTML attribute is missing double quotes around the value:

```suggestion
<h3 class="govuk-heading-l">HTML attachment</h3>
```

While it's [not technically necessary](https://stackoverflow.com/questions/6495310/do-you-quote-html5-attributes) in this case, it's a good habit to use them everywhere.

app/views/admin/editions/_statistical_data_set_fields.html.erb
```diff
@@ -0,0 +1,20 @@
+<% cache_if edition.statistical_data_set_document_ids.empty?, taggable_statistical_data_sets_cache_digest do %>
```
ollietreend, line 1: Ditto for this cache key.

```suggestion
<% cache_if edition.statistical_data_set_document_ids.empty?, "#{taggable_statistical_data_sets_cache_digest}-design-system" do %>
```

app/views/admin/editions/_topical_event_fields.html.erb
```diff
@@ -0,0 +1,20 @@
+<% cache_if edition.topical_event_ids.empty?, taggable_statistical_data_sets_cache_digest do %>
```
ollietreend, line 1: ☝🏻

app/views/admin/editions/_world_location_fields.html.erb
```diff
@@ -0,0 +1,22 @@
+<% required = false unless defined?(required) %>
+
+<% cache_if edition.world_location_ids.empty?, taggable_statistical_data_sets_cache_digest do %>
```
ollietreend, line 3: ☝🏻

ollietreend (commented):
Nice work plumbing this all in 🔧 👍🏻 

Just a few small comments.

app/views/admin/editions/_statistical_data_set_fields.html.erb
davidgisbey: Yea good shout. I wasn't about conflicts while people are using the flag

ollietreend (approved):

PR #6872, opened 2022-09-28
Add nations fields in the GOV.UK Design System

davidgisbey:
## Description

This adds in the nation_fields partial in the GOV.UK Design System.

The Bootstrap implementation was pretty weird in that it had a checkbox for all nations that passes a value to the controller via the `all_nation_applicability` attribute, then all 4 nations are passed as part of the nested `nation_inapplicabilities_attributes`

(see the highlighted code)

image

I've had to do a decent amount of controller munging to get this work with the publishing component checkbox component and keep it accessible.

One issue i encountered was we only need to build the nation params for publications, consultations and detailed guides. This needs to be done before the build_edition before action, which builds the edition based on the params. 

However, as we only want to build the nation params for certain types and there isn't a persisted object yet, i've used the controller_name to infer which edition object will be built.

There may well be a better way of doing it, but it's late and sleepy! Please feel free to change this if anyone can think of anything better!

## Screenshot

image

## Trello card

### DIFF
--- a/app/views/admin/editions/_nation_fields.html.erb
+++ b/app/views/admin/editions/_nation_fields.html.erb
@@ -1,0 +1,75 @@
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][0][id]", edition.nation_inapplicabilities.find_by(nation_id: 1)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][1][id]", edition.nation_inapplicabilities.find_by(nation_id: 2)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][2][id]", edition.nation_inapplicabilities.find_by(nation_id: 3)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][3][id]", edition.nation_inapplicabilities.find_by(nation_id: 4)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][0][nation_id]", 1 %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][1][nation_id]", 2 %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][2][nation_id]", 3 %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][3][nation_id]", 4 %>
+
+<%= render "govuk_publishing_components/components/checkboxes", {
+  name: "edition[all_nation_applicability][]",
+  id: "edition_nation_inapplicabilities",
+  heading: "Excluded nations",
+  error: errors_for_input(edition.errors, :nation_inapplicabilities),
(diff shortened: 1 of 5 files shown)


app/views/admin/editions/_nation_fields.html.erb
```diff
@@ -0,0 +1,75 @@
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][0][id]", edition.nation_inapplicabilities.find_by(nation_id: 1)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][1][id]", edition.nation_inapplicabilities.find_by(nation_id: 1)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][2][id]", edition.nation_inapplicabilities.find_by(nation_id: 1)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][3][id]", edition.nation_inapplicabilities.find_by(nation_id: 1)&.id %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][0][nation_id]", 1 %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][1][nation_id]", 2 %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][2][nation_id]", 3 %>
+<%= hidden_field_tag "edition[nation_inapplicabilities_attributes][3][nation_id]", 4 %>
```
ollietreend, line 8: This (and the controller munging) feels a bit like we're bending over backwards to get the [checkboxes component][] to fit our needs.

I find these `hidden_field_tags` a bit confusing – I'm not entirely sure what they're doing or why. But I suspect there could be a bug hiding in here: the first four lines all perform a `find_by(nation_id: 1)` – hardcoded to `nation_id: 1`, whereas I think they might need to be for IDs 1, 2, 3 and 4 as the following four lines are?

I wonder if a nicer approach might be to break away from using the publishing component in this case and go with custom HTML markup. That way we'd (hopefully) be able to do away with the data munging in the controller.

What do you think?

[checkboxes component]: https://components.publishing.service.gov.uk/component-guide/checkboxes

davidgisbey: Either way this should probs live in a view component. At a first glance. Does this handle passing a value when something is unchecked?

davidgisbey: Maybe we can chat this through after standup

ollietreend: > Does this handle passing a value when something is unchecked?

No 😞 plus it doesn't set the `nation_id` field for each item, either. So... actually those hidden inputs would still be needed most likely.

Let's chat after standup.

davidgisbey: After discussion we're going to stick as is for now and i'm going to whack it in a component after

ollietreend (approved):

PR #6877, opened 2022-10-03
Add support for upcoming changes section

DilwoarH:
## What
We need to add an 'Upcoming changes' section to the 'What's new' page in Whitehall Publisher.

This section would be between the introduction section and the 'recent changes' section.

The section and entries follow the same format as the 'recent changes' section.

The section shouldn't show in the contents list or on the page if there's no entries/content in the YAML file.

## Why
We're using the page to communicate with publishers upcoming changes to Whitehall Publisher.

## Additional Information
https://github.com/alphagov/whitehall/blob/main/config/locales/en/admin/whats_new.yml

https://github.com/alphagov/whitehall/tree/main/app/views/admin/whats_new

whitehall-admin dev gov uk_government_admin_whats-new(iPad Pro) (10)

https://trello.com/c/Cf4p7Ied/785-add-upcoming-changes-section-to-whats-new-page

### DIFF
--- a/config/locales/en/admin/whats_new.yml
+++ b/config/locales/en/admin/whats_new.yml
@@ -17,7 +17,7 @@ en:
         updates:
           - heading: Topic taxonomy tags page moves to the Design System
             area: Creating and updating documents
-            type: improved
+            type: improvement
             date: 5 October 2022
             body_govspeak: |
               The topic taxonomy tags page will move to the GOV.UK Design System.
@@ -28,7 +28,7 @@ en:
         updates:
           - heading: Reusing previous withdrawal dates and public explanations
             area: Creating and updating documents
-            type: improved
+            type: improvement
             date: 26 September 2022
             body_govspeak: |
               When re-withdrawing content, you can choose to reuse a previous withdrawal date and public explanation in certain circumstances.
--- a/config/locales/en/admin/whats_new.yml
+++ b/config/locales/en/admin/whats_new.yml
@@ -15,15 +15,12 @@ en:
       upcoming_changes:
         heading: Upcoming changes
         updates:
-          - heading: Topic taxonomy tagging and unpublishing and withdrawing pages move to the Design System
+          - heading: Topic taxonomy tags page moves to the Design System
             area: Creating and updating documents
             type: improved
             date: 5 October 2022
             body_govspeak: |
-              The following pages will move to the GOV.UK Design System:
-              
-              * topic taxonomy tagging
-              * unpublishing, withdrawing and unwithdrawing
+              The topic taxonomy tags page will move to the GOV.UK Design System.
 
               Topics that are tagged to a document will be listed under the ‘Selected topics’ header on the ‘Topic taxonomy tags’ page. You can also remove tags in this section.
       recent_changes:
--- a/config/locales/en/admin/whats_new.yml
+++ b/config/locales/en/admin/whats_new.yml
@@ -18,7 +18,7 @@ en:
           - heading: Topic taxonomy tags page moves to the Design System
             area: Creating and updating documents
             type: improvement
-            date: 5 October 2022
+            date: 6 October 2022
             body_govspeak: |
               The topic taxonomy tags page will move to the GOV.UK Design System.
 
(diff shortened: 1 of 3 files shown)


config/locales/en/admin/whats_new.yml
```diff
@@ -12,6 +12,17 @@ en:
...
+        heading: Upcoming changes
+        updates:
+          - heading: Reusing previous withdrawal dates and public explanations
+            area: Creating and updating documents
+            type: improved
+            date: 26 September 2022
+            body_govspeak: |
+              When re-withdrawing content, you can choose to reuse a previous withdrawal date and public explanation in certain circumstances.
+
+              [Read the guidance about when you can reuse previous withdrawal dates and explanations](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/unpublishing-and-archiving).
```
DilwoarH, line 25: @johnpauldickie - added some placeholder content, could you let me know what to put here?

johnpauldickie (approved):

config/locales/en/admin/whats_new.yml
```diff
@@ -4,14 +4,25 @@ en:
...
           This will make Whitehall Publisher more accessible, more user friendly and give you a consistent user experience throughout your publishing journey on GOV.UK.
 
           [Read more about the changes on the content community Basecamp](https://3.basecamp.com/4322319/buckets/15005645/messages/5287245970).
+      upcoming_changes:
+        heading: Upcoming changes
+        updates:
+          - heading: Topic taxonomy tags page moves to the Design System
+            area: Creating and updating documents
+            type: improved
+            date: 5 October 2022
```
kevindew, line 21: I'm wondering if having a date here is a bit confusing without context.

I interpreted the date to be the date the feature is (was) due rather than the date we posted the update.

config/locales/en/admin/whats_new.yml
```diff
@@ -4,14 +4,25 @@ en:
...
           Whitehall Publisher will be moving from the current styling to use the styling and components of the [GOV.UK Design System](https://design-system.service.gov.uk/).
           This will make Whitehall Publisher more accessible, more user friendly and give you a consistent user experience throughout your publishing journey on GOV.UK.
 
           [Read more about the changes on the content community Basecamp](https://3.basecamp.com/4322319/buckets/15005645/messages/5287245970).
+      upcoming_changes:
+        heading: Upcoming changes
+        updates:
+          - heading: Topic taxonomy tags page moves to the Design System
+            area: Creating and updating documents
+            type: improved
```
kevindew, line 20: I may be over thinking it so this is a bit nitpicky, but I wonder if past tense tags aren't appropriate for upcoming changes. I'd wonder if "improvement" would be better. 

May be good to make all the tags present tense:

Improved => Improvement
Fixed => Fix
Changed => Change

kevindew (commented):
Code looks good but I have a couple of comments on the content metadata 🙈 sorry

config/locales/en/admin/whats_new.yml
johnpauldickie: The date is the change is due to be released. I don't think there's a risk of confusion here based on the date being in the future. We've based this approach on other similar pages on Content Publisher and out there in industry.

config/locales/en/admin/whats_new.yml
kevindew: Great stuff. I unresolved this as @DilwoarH will need to change the code slightly to accomodate them.

kevindew: Oh gosh, sorry I see you fixed the code! Great job @johnpauldickie

config/locales/en/admin/whats_new.yml
kevindew: I'm confused now - I think because this date is today's date - am I understanding right from your message that we're shipping "Topic taxonomy tags" today? and putting an upcoming change that it's coming today? 😕

johnpauldickie: Sorry @kevindew - mistake in the date which I didn't spot. It is being released tomorrow but was originally meant to be today. Now updated in a new commit.

kevindew (approved):
Thanks - changes look good.

I do wonder if we should be more ambiguous with the time things are coming. Going for a particular date is a little delicate. I'd wonder if we were fine going with just a month value "October 2022". Obviously doesn't feel appropriate for something that is coming in one day but in general could be a good and less confusing approach.

PR #6880, opened 2022-10-03
Reverse show notice flag to be false by default

DilwoarH:
The logic OR assignment was treating the `false` flag we passed in as always to be [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) (which is correct).

This changes it to make sure that it only shows if you pass in a value of `true`.

https://trello.com/c/rHlLiraK/788-fix-double-notices-showing

## Before

Screenshot 2022-10-03 at 10 17 49 am

## After

Screenshot 2022-10-03 at 10 12 07 am
Screenshot 2022-10-03 at 10 12 42 am

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -1,7 +1,7 @@
 <% environment_style = GovukAdminTemplate.environment_style %>
 <% environment_label = GovukAdminTemplate.environment_label %>
 <% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
-<% show_notices ||= true %>
+<% show_legacy_notices ||= false %>
 
 <header class="
   masthead
@@ -79,7 +79,7 @@
       </li>
     </ul>
   </div>
-  <% if show_notices %>
+  <% if show_legacy_notices %>
     <section class="notices">
       <%= render partial: "shared/notices" %>
     </section>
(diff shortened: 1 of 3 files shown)


app/views/shared/_header.html.erb
```diff
@@ -1,7 +1,7 @@
 <% environment_style = GovukAdminTemplate.environment_style %>
 <% environment_label = GovukAdminTemplate.environment_label %>
 <% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
-<% show_notices ||= true %>
+<% show_legacy_notices ||= false %>
```
dnkrj, line 4: It seems like defining defaults in this way is just inviting these kind of bugs. Do you know if ruby/rails offer any better alternatives?

dnkrj (approved):
Will definitely fix the bug, just feel like there must be a less risky way of setting defaults.
Feel free to merge and we can revisit with a more general look at things.

app/views/shared/_header.html.erb
DilwoarH: This is currently our convention on setting defaults, see other examples: https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_cards.html.erb#L2

I do get what you're saying, however, this partial is only used by two templates, admin layout and design system layout.

DilwoarH: I am going to merge, but feel free to suggest any alternative solutions.

PR #6888, opened 2022-10-05
Tidy routes file (and other routes related things)

brucebolt:
The public-facing routes part of the `config/routes.rb` file had become somewhat disorganised with no logical ordering and inconsistent syntax for different routes.

This breaks that part of the file into the following sections:
- routes rendered by Whitehall to the public under the `/world` scope
- routes rendered by Whitehall to the public under the `/government` scope
- routes that exist solely for the purpose of non-English finders
- redirects rendered by Whitehall (ideally these would be moved out so they are handled at the router level)
- routes no longer rendered by Whitehall, but retained to maintain the route helpers (we will look into resolving these in a later PR)

Additionally each section has been sorted into alphabetical order.  Furthermore some other related code and tests have been updated so we don't need to maintain so many redundant routes.

[Trello card](https://trello.com/c/EfhUs8ak)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,30 +37,29 @@ Whitehall::Application.routes.draw do
     resources :worldwide_organisations, path: "worldwide-organisations", only: [:show], defaults: { format: :json }
   end
 
-  # World locations and Worldwide organisations
-  get "/world/organisations/:organisation_id/office" => redirect("/world/organisations/%{organisation_id}", prefix: "")
-  get "/world/organisations/:organisation_id/about" => redirect("/world/organisations/%{organisation_id}", prefix: "")
-  get "/world/organisations/:id(.:locale)", as: "worldwide_organisation", to: "worldwide_organisations#show", constraints: { locale: valid_locales_regex }
-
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -65,6 +65,21 @@ def external_redirect(path_prefix, target)
   scope Whitehall.router_prefix, shallow_path: Whitehall.router_prefix do
     root to: redirect("/", prefix: ""), via: :get, as: :main_root
 
+    # Redirects rendered by Whitehall
```
BeckaL, line 68: I'm a bit less sure about this grouping - do we anticipate doing the redirects outside of whitehall at some point? I guess I'm thinking that maybe it's still useful to see all the e.g. organisation stuff grouped together, even if it's a redirect, unless we know that we'll fix it at some point, in which case this list serves as a to-do list.

But this not at all a strong opinion, just discussion point.

brucebolt: I'd like to remove these redirects from Whitehall completely, as router can handle the concept of redirects.  Maybe I'll do that as part of this card, but on a later PR.

PR #6892, opened 2022-10-05
Implement Analytics tracking on the 'Topic taxonomy tags' page

DilwoarH:
## What

Implement tracking on the Design System version of the 'Topic taxonomy tags' page.

## Why

So that we can continue tracking how publishers use this page.

https://trello.com/c/EyKlvBnz/793-implement-analytics-tracking-on-the-topic-taxonomy-tags-page

### DIFF
--- a/app/assets/javascripts/modules/track-selected-taxons.js
+++ b/app/assets/javascripts/modules/track-selected-taxons.js
@@ -22,7 +22,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         taxonPath.push(taxons[i].innerText.trim())
       }
 
-      GOVUK.analytics.trackEvent('taxonSelection', taxonPath.join(' > '), { label: window.location.pathname })
+      GOVUK.analytics.trackEvent('taxonSelection', taxonPath.join(' > '), { label: this.getCurrentPath() })
     }
   }
 
@@ -43,6 +43,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         )
       })
     }
+  }
+
+  TrackSelectedTaxons.prototype.getCurrentPath = function () {
+    return window.location.pathname
   }
 
   Modules.TrackSelectedTaxons = TrackSelectedTaxons
--- a/app/assets/javascripts/modules/track-selected-taxons.js
+++ b/app/assets/javascripts/modules/track-selected-taxons.js
@@ -1,0 +1,29 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function TrackSelectedTaxons (module) {
+    this.form = module
+  }
+
+  TrackSelectedTaxons.prototype.init = function () {
+    this.form.addEventListener('submit', this.onSubmitListener.bind(this))
+  }
+
+  TrackSelectedTaxons.prototype.onSubmitListener = function (e) {
+    var selectedItems = this.form.querySelectorAll('.miller-columns-selected__list .miller-columns-selected__list-item')
+
+    for (var index = 0; index < selectedItems.length; index++) {
+      var taxonPath = []
+      var taxons = selectedItems[index].querySelectorAll('.govuk-breadcrumbs__list-item')
+
+      for (var i = 0; i < taxons.length; i++) {
+        taxonPath.push(taxons[i].innerText.trim())
+      }
+
+      GOVUK.analytics.trackEvent('taxonSelection', taxonPath.join(' > '), { label: window.location.pathname })
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/modules/track-selected-taxons.js
```diff
@@ -0,0 +1,28 @@
...
+  TrackSelectedTaxons.prototype.init = function () {
+    this.form.addEventListener('submit', this.onSubmitListener.bind(this))
+  }
+
+  TrackSelectedTaxons.prototype.onSubmitListener = function (e) {
+    var requestPath = this.requestPath
+    var selectedItems = this.form.querySelectorAll('.miller-columns-selected__list .miller-columns-selected__list-item')
+
+    Array.from(selectedItems, function(item) {
+      var taxonPath = Array.from(item.querySelectorAll('.govuk-breadcrumbs__list-item'), function(crumb) {
```
ollietreend: `Array.from` [isn't supported by IE11](https://caniuse.com/mdn-javascript_builtins_array_from) – is this going to be a problem? I'm not sure if this method gets polyfilled, or if we're yet at the stage where we can drop IE11 support?

app/views/admin/edition_tags/edit.html.erb
```diff
@@ -30,7 +30,10 @@
       margin_bottom: 2,
     } %>
 
-    <%= form_for @tag_form, url: admin_edition_tags_path(@edition), method: :put do |form| %>
+    <%= form_for @tag_form, url: admin_edition_tags_path(@edition), method: :put, data: {
+      module: "track-selected-taxons",
+      "request-path": request.path,
```
ollietreend, line 35: Do we _need_ to pass this through from the server side? 🤔

Unless I'm mistaken, I think `request.path` in Rails will give the same result as `window.location.pathname` in JavaScript.

I know you're just replicating existing behaviour, so no need to change if you want to keep it strictly like-for-like.

app/assets/javascripts/modules/track-selected-taxons.js
```diff
@@ -33,12 +33,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         var target = e.currentTarget
 
         if (target.checked) {
-          GOVUK.analytics.trackEvent('pageElementInteraction', 'checkboxClickedOn', { label: target.labels[0].innerText.trim() })
+          GOVUK.analytics.trackEvent('pageElementInteraction', 'checkboxClickedOn', { label: this.form.querySelector('label[for="' + target.id + '"]').innerText.trim() })
         } else {
-          GOVUK.analytics.trackEvent('pageElementInteraction', 'checkboxClickedOff', { label: target.labels[0].innerText.trim() })
+          GOVUK.analytics.trackEvent('pageElementInteraction', 'checkboxClickedOff', { label: this.form.querySelector('label[for="' + target.id + '"]').innerText.trim() })
```
ollietreend, line 38: Opinion: it might be nicer (more DRY) to define 

```js
var label = this.form.querySelector('label[for="' + target.id + '"]').innerText.trim()
```

above, where you already have `var target`. To avoid repeating it twice in each half of this `if` statement.

spec/javascripts/modules/track-selected-taxon.spec.js
```diff
@@ -0,0 +1,145 @@
...
+    )
+
+    expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith(
+      'taxonSelection',
+      'Parenting, childcare and children\'s services > Childcare and early years',
+      { label: '/government/admin/editions/1/tags/edit' }
+    )
+  })
+
+  it('should send two tracking event when form is submitted with withdrawal date selected', function () {
```
ollietreend, line 123: You've got the wrong test description here (but I can tell where you copied & pasted this test from 😆)

app/views/admin/edition_tags/edit.html.erb
DilwoarH: good point - we could actually, I think I went too much into isolation principles and wasn't trying to use anything other than the module scope.

app/assets/javascripts/modules/track-selected-taxons.js
DilwoarH: Updated

app/views/admin/edition_tags/edit.html.erb
DilwoarH: updated

app/assets/javascripts/modules/track-selected-taxons.js
DilwoarH: updated

spec/javascripts/modules/track-selected-taxon.spec.js
DilwoarH: updated

PR #6898, opened 2022-10-06
Use Redis.new rather than Redis.current

edwardkerry:
Redis.current will be deprecated in Redis 5.0.

As Redis.current ultimately calls Redis.new we can migrate this method now to pave the way for future upgrades.

From Redis changelog from [5.0][1]:

‘Removed Redis.current. You shouldn't assume there is a single global Redis connection, use a connection pool instead, and libaries using Redis should accept a Redis instance (or connection pool) as a config. E.g. MyLibrary.redis = Redis.new(…).’

[Redis.current][2] just calls Redis.new so should just be able to swap methods.

[trello](https://trello.com/c/oGFWhrCO/1551-migrate-away-from-deprecated-redis-method)\\

Test Updates: 
- Because taxonomy helper tests need a shared Redis instance (for stubbing), we also Stub Redis.new to return the stubbed client.
- Because we're always getting a new Redis instance, remove the block in the feature tests that previously forced a new Redis instance.

### DIFF
--- a/features/support/admin_taxonomy_helper.rb
+++ b/features/support/admin_taxonomy_helper.rb
@@ -24,11 +24,3 @@ module AdminTaxonomyHelper
   end
 end
 World(AdminTaxonomyHelper)
-
-Around do |_, block|
-  redis = Redis.new
-  Redis.new = nil
-  block.call
-ensure
-  Redis.new = redis
-end
--- a/features/support/admin_taxonomy_helper.rb
+++ b/features/support/admin_taxonomy_helper.rb
@@ -26,9 +26,9 @@ end
 World(AdminTaxonomyHelper)
 
 Around do |_, block|
-  redis = Redis.current
-  Redis.current = nil
+  redis = Redis.new
+  Redis.new = nil
   block.call
 ensure
-  Redis.current = redis
+  Redis.new = redis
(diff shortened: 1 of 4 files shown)


KludgeKML:
The problem affecting these tests is that the redis cache adapter is used in classes that don't really allow a Redis instance (or a cache adapter instance) to be passed easily to them from tests, so sharing a redis instance stubbed in the test with the user of the adapter is quite hard. We could rewrite some of the code to accept a redis client passed into it, but it feels like it would be passing it around so much that we'd end up with production code that was only there to make the tests pass. The alternative seems to be rewriting a bunch of tests, but that bunch might be quite a lot.

features/support/admin_taxonomy_helper.rb
```diff
@@ -24,11 +24,3 @@ def check_links_patched_in_publishing_api
...
 end
 World(AdminTaxonomyHelper)
-
-Around do |_, block|
-  redis = Redis.current
-  Redis.current = nil
-  block.call
-ensure
-  Redis.current = redis
-end
```
kevindew, line 34: Nice to see this go 👍

kevindew (approved):
Looks good to me

PR #6900, opened 2022-10-06
Remove email signup controller and route

brucebolt:
Now that all the English language finders have been moved out of Whitehall, there is no need to maintain the `/government/email-signup` path.

The functionality previously provided by this path was migrated to Email Alert Frontend, with only 2 places in Whitehall linking to this path (which result in a 404 anyway).  This PR fixes those 2 broken links, which are found on the non-English World Location News pages [\[example\]](https://www.gov.uk/world/france/news.fr) and JSON versions of the non-English finders [\[example\]](https://www.gov.uk/government/publications.de.json) (both still rendered by Whitehall).

Therefore removing all the code that was used to render these pages as they are no longer required.

[Trello card](https://trello.com/c/EfhUs8ak/33-tidy-whitehalls-routes-file)

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6902, opened 2022-10-06
Create route helpers for Get Involved and Topical Events

brucebolt:
The controllers for these routes were removed some time ago, however the routes (and the non-existent controller names) remained in the routes file solely so we could utilise the route helpers automatically created by Rails.

This removes those routes and creates our own custom route helper methods.

I did investigate using [`direct` routes](https://guides.rubyonrails.org/routing.html#direct-routes) as an alternative.  This would be fine to use if we only wanted fairly static URLs, e.g. `/government/get-involved` or `/government/get-involves/take-part/#{slug}`.  It appears the returned string gets passed into `url_for`, which doesn't apply any of the options (e.g. appending a cachebust string, locale or format) if you pass in a string (it does if you use a model or controller name).  Therefore we'll need to create our own helpers in order to add in these options which are quite extensively baked into Whitehall.

This PR includes a commit that fixes a test (forming atom feed URLs) that will be removed when https://github.com/alphagov/whitehall/pull/6900 is merged, as Whitehall no longer needs to create these URLs for formats that it no longer renders.

[Trello card](https://trello.com/c/EfhUs8ak/33-tidy-whitehalls-routes-file)

### DIFF
--- a/app/helpers/public_document_routes_helper.rb
+++ b/app/helpers/public_document_routes_helper.rb
@@ -112,6 +112,38 @@ module PublicDocumentRoutesHelper
     Plek.new.website_root + take_part_page_path(object, options)
   end
 
+  def topical_event_path(object, options = {})
+    slug = case object
+           when String
+             object
+           when TopicalEvent
+             object.slug
+           else
+             raise ArgumentError, "Must provide a slug or TopicalEvent"
+           end
+
+    append_url_options("/government/topical-events/#{slug}", options)
+  end
+
+  def topical_event_url(object, options = {})
+    Plek.new.website_root + topical_event_path(object, options)
+  end
+
+  def topical_event_about_pages_path(object, options = {})
+    slug = case object
+           when String
+             object
+           when TopicalEvent
+             object.slug
+           when TopicalEventAboutPage
+             object.topical_event.slug
+           else
+             raise ArgumentError, "Must provide a slug, TopicalEvent or TopicalEventAboutPage"
+           end
+
+    append_url_options("/government/topical-events/#{slug}/about", options)
+  end
+
 private
 
   def build_url_for_corporate_information_page(edition, options)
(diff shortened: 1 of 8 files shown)


app/helpers/public_document_routes_helper.rb
```diff
@@ -112,6 +112,42 @@ def take_part_page_url(object, options = {})
...
+             object.topical_event.slug
+           else
+             raise ArgumentError, "Must provide a slug, TopicalEvent or TopicalEventAboutPage"
+           end
+
+    append_url_options("/government/topical-events/#{slug}/about", options)
+  end
+
+  def topical_event_about_pages_url(object, options = {})
+    Plek.new.website_root + topical_event_about_pages_path(object, options)
```
BeckaL, line 148: Do we need this method? As far as I can tell it's only used in tests. Or do we want to make sure all content types we introduce helpers for have both 'url' and 'path' helpers for consistency with how it is when we have the rails helpers?

BeckaL (approved):
Looks good to me - question above about whether we still want the 'url' method as well as the 'path' method, but no strong opinions either way. All looks safe to me, but I think worth deploying this to integration and having a poke round the pages that will be affected by this just to be double sure?

app/helpers/public_document_routes_helper.rb
brucebolt: I'll try removing this method and seeing if any of the existing tests fail.  I think I only added it for consistency, but will definitely remove it if it's not being used anywhere.

brucebolt:
> I think worth deploying this to integration and having a poke round the pages that will be affected by this just to be double sure?

Yes, I'll definitely do this before merging.  I did test all the possible UIs locally, but a final check on integration would be good too.

app/helpers/public_document_routes_helper.rb
brucebolt: All the tests passed, so I've removed the `topical_event_about_pages_url` method and its associated tests.  Also rebased against main as this branch was quite a bit out of date.

brucebolt:
I've deployed this to integration, clicked around the relevant pages in Whitehall's admin UI and checked for errors being logged.  All looks ok, so I'm going to merge this into main.

PR #6903, opened 2022-10-06
Fix invalid topic taxonomy HTML

kevindew:
Square braces are not valid characters in an id and produces HTML like so:

```
<div class=\"govuk-checkboxes__item\"><input class=\"govuk-checkboxes__input\" type=\"checkbox\" name=\"taxonomy_tag_form[taxons][]\" value=\"44171085-15e5-4524-89ca-b409d3675f93\" id=\"taxonomy_tag_form[taxons]-0\" tabindex=\"-1\"><label class=\"govuk-label govuk-checkboxes__label\" for=\"taxonomy_tag_form[taxons]-0\">\n        Test taxon\n      </label></div>
```

I wondered if this was behind our end-to-end tests failures, but was wrong. However since I fixed it some invalid HTML it felt like I should open it as PR.

### DIFF
--- a/app/views/admin/edition_tags/edit.html.erb
+++ b/app/views/admin/edition_tags/edit.html.erb
@@ -37,7 +37,7 @@
 
       <% if @topic_taxonomy.ordered_taxons_transformed(@tag_form.selected_taxons).count() %>
         <%= render partial: "/components/miller-columns", locals: {
-          id: "taxonomy_tag_form[taxons]",
+          id: "taxonomy_tag_form-taxons",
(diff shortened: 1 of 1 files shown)


beccapearce (approved):

ollietreend (commented):
I don't think this change is needed because it's valid in HTML5.

HTML 4.01 was more strict about which characters were valid in `id` attributes – HTML5 is much more relaxed.

More info: https://mathiasbynens.be/notes/html5-id-class

kevindew:
Today I learnt! Thanks Ollie

PR #6907, opened 2022-10-07
Fix historical route generation

JonathanHallam:
https://github.com/alphagov/whitehall/pull/6888 made some changes to how routes work for historical appointments. These were slightly more complicated than initially thought and the implementation broke some route generation. This PR reverts these changes to make sure that routes are generated properly again.


BeckaL (approved):
LGTM, had a check against previous routes file prior to refactoring and they all look equivalent but now scoped under /history

PR #6909, opened 2022-10-10
[WIP] Separate the asset files (CSS & JS) for Bootstrap pages vs. Design System pages

DilwoarH:
## What

We need to separate out the assets used on the Bootstrap vs. Design System page layouts.

### New file naming conventions

No changes needed to frontend assets. For admin assets, use these conventions:

**Bootstrap assets**

- JavaScript
  - Main JS file: `app/assets/javascripts/admin_legacy.js`
  - Root JS directory: `app/assets/javascripts/admin_legacy/`
- Stylesheets
  - Main CSS file: `app/assets/stylesheets/admin_legacy.scss`
  - Root SCSS directory: `app/assets/stylesheets/admin_legacy/`

**Design System assets**

- JavaScript
  - Main JS file: `app/assets/javascripts/application.js`
  - Root JS directory: `app/assets/javascripts/admin/`
  - Subdirectories for components, views, etc.
- Stylesheets
  - Main CSS file: `app/assets/stylesheets/application.scss`
  - Root SCSS directory: `app/assets/stylesheets/admin/`
  - Subdirectories for components, views, etc.

### Split out files

Files should be split so that:

- Design System layout uses `application.js`
- Bootstrap layout uses `admin.js`
- `application.js` no longer imports `admin.js` (i.e. we should remove the `//= require admin` directive)

### Nav bar

We'll still need some CSS & JavaScript to support the nav bar, which is shared between both layouts.

Ideally this behaviour could be re-implemented without requiring jQuery – but if it ends up too complicated, then it'd be fine for us to just import the existing behaviour & jQuery.

### Other functionality

If there's any other functionality that the Design System implementation depends on from Bootstrap, we should port it across. (If jQuery is required, we should try to reimplement is with vanilla JS.)

I don't expect there will be much functionality that falls into this category – but it's worth keeping in mind. We should be especially careful to ensure Analytics tracking behaviour remains functional, since this isn't well covered by tests.

## Why

Right now, the Design System layout includes all of the JS used in the Bootstrap layout. We run the risk of the old JS interfering with the new JS, introducing unexpected behaviour or even implementing behaviour that we didn't know existed.

We've already seen examples where the Bootstrap JS is doing unwanted things on Design System pages – e.g. https://github.com/alphagov/whitehall/commit/0ec847dc9d1974ce0774f0043dc09a84a385d5e8

https://trello.com/c/3UW25yLj/779-separate-the-asset-files-css-js-for-bootstrap-pages-vs-design-system-pages


app/assets/javascripts/admin_legacy.js
```diff
@@ -0,0 +1,14 @@
+//= require govuk_publishing_components/lib
+//= require jquery-ui
+//= require extensions_legacy/jquery.distinct
+//= require vendor_legacy/jquery/chosen.jquery.js
```
kevindew: It seems unconventional to change the name of a vendor directory. Can we leave it? I imagine it's also not true that these are legacy for the frontend rendering part of the site.

app/assets/javascripts/admin_legacy.js
```diff
@@ -0,0 +1,14 @@
...
+//= require vendor_legacy/jquery/jquery.player.min.js
+//= require vendor_legacy/jquery/jquery.highlighttextarea.js
+//= require vendor_legacy/jquery/jquery-debounce.js
+//= require vendor_legacy/google-diff-match-patch.js
+//= require ./application_legacy/convert-html-pub-charts
+//= require govuk_legacy
+//= require_tree ./common_legacy
```
kevindew, line 11: Similar comment about this only being legacy in admin context, but by it's nature of being common it's things for frontend and backend.

I think we should leave common as is. If we wanted to distinguish we could copy files from common into admin_legacy

app/assets/javascripts/admin_legacy.js
```diff
@@ -0,0 +1,14 @@
...
+//= require ./application_legacy/hide-extra-rows.js
```
kevindew, line 12: application_legacy seems a bit of a strange folder name. Do we know what the purpose of the application folder is? it seems like they should either be in common, if they're shared with frontend and backend, or in admin_legacy if they're just needed for admin

app/assets/javascripts/all_legacy.js
```diff
@@ -1,18 +1,18 @@
 // This is a manifest file used by the runner for JavaScript unit tests.
```
kevindew, line 1: This all file seems to comprise both frontend and admin so it doesn't seem like something that should be legacy - it seems the intention would be to use it for all JS. 

Do we instead need different files used for JS tests? maybe we don't need an `all` file anymore?

app/assets/javascripts/frontend.js
```diff
@@ -12,10 +12,10 @@
...
+//= require vendor_legacy/jquery/jquery.player.min.js
+//= require vendor_legacy/sorttable
+//= require vendor_legacy/object-create-polyfill
 //
-//= require govuk
-//= require_tree ./common
-//= require_tree ./application
+//= require govuk_legacy
+//= require_tree ./common_legacy
+//= require_tree ./application_legacy
```
kevindew, line 21: It doesn't seem like this file should have renames as the legacy changes should only be affecting the admin part of Whitehall

spec/support/jasmine-browser.json
```diff
@@ -2,7 +2,7 @@
   "srcDir": "public/assets/whitehall",
   "srcFiles": [
     "application-*.js",
-    "all-*.js"
+    "all_legacy-*.js"
```
kevindew, line 5: I'm expecting we may want this to resemble: 

```suggestion
"admin_legacy-*.js"
"frontend-*.js"
```

since all won't mean what it used to.

app/assets/javascripts/application.js
```diff
@@ -5,13 +5,13 @@
 //= require govuk-admin-template
 //= require admin_legacy
 
-//= require components/miller-columns
-//= require components/autocomplete
+//= require admin/components/miller-columns
+//= require admin/components/autocomplete
```
kevindew, line 5: I imagine we're not distinguishing components as being admin or not - so I'd probably suggest leaving them under components

app/assets/javascripts/application.js
```diff
@@ -5,13 +5,13 @@
...
-//= require modules/analytics
-//= require modules/paste-html-to-govspeak
-//= require modules/unpublish-display-conditions
-//= require modules/unpublish-tracking
-//= require modules/track-selected-taxons
+//= require admin/modules/analytics
+//= require admin/modules/paste-html-to-govspeak
+//= require admin/modules/unpublish-display-conditions
+//= require admin/modules/unpublish-tracking
+//= require admin/modules/track-selected-taxons
```
kevindew: This looks good, I assume these only have a role in the admin side of the app?

app/assets/javascripts/application.js
```diff
@@ -3,15 +3,15 @@
 //= require govuk_publishing_components/analytics
 
 //= require govuk-admin-template
-//= require admin_legacy
```
kevindew: This commit could use a why to explain why this is removed.

app/assets/javascripts/admin/modules/navbar-toggle.js
```diff
@@ -0,0 +1,27 @@
+window.GOVUK = window.GOVUK || {}
```
kevindew, line 1: I'm a bit unclear why this extra JS is added the commit doesn't explain much.

Is it definitely needed in this PR (i.e. do things break without it) as it's already quite an intense amount of changes.

kevindew (commented):
Nice job! Looks like it's on a good journey to things being cleaner.

I've added some comments where I feel the `legacy` suffix has ventured into the frontend and shared aspect of the code where it isn't applicable (we don't have a non-legacy frontend in here so we don't need to distinguish). Comments I've made apply to both JS and stylesheets but I only added them to JS.

I've also found myself a bit confused about the extra JS added later in this PR, it's a bit unclear to me whether this is needed for the core task of this PR. Some better commit documentation would help.

Should we add a short bit documentation to this app about this new asset organisation once all is resolved?

app/assets/javascripts/admin_legacy.js
DilwoarH: yes I did move it back, looks like I missed this one

DilwoarH: I did move it back in this [commit](https://github.com/alphagov/whitehall/pull/6909/commits/39036b97c8cb22fdbdf13ac8fa8c21119c0e49d1)

app/assets/javascripts/admin_legacy.js
DilwoarH: I think maybe we should have two copies, one in `admin_legacy` and one in `frontend`. Leaving it in the root folder might indicate that these things are actually common

app/assets/javascripts/admin/modules/navbar-toggle.js
DilwoarH: So this is something that I wanted to split up but was requested that we should remove all legacy admin JS and rewrite the navigation toggle in vanilla JS

app/assets/javascripts/all_legacy.js
DilwoarH: Only thing I see that is required in the all and not the other files is:

```
//= require govuk-admin-template
```

DilwoarH: I have reverted this change as you are correct, it isn't legacy.

app/assets/javascripts/application.js
DilwoarH: Components will never be used within the `frontend` side of things as we have no plans to move any of the `frontend` code to the GOVUK design system. The plan is once we remove all of the frontend code, we will be moving the whole `admin` directory up one level.

kevindew: As I understand the frontend is already moved to the design system and makes use of components.

Even if that wasn't the case I just suspect that we're going to get caught in a muddle with the component guide expecting a configuration in some places. So I feel like trying to make them distinct with an admin prefix is likely to cause some problems or inconsistencies.

DilwoarH:
Things to do:
- Split up the PR so Navigation JS is seperate
- Restructure `spec` folder
- Move `components` to top directory

DilwoarH:
superseded by #6921

PR #6911, opened 2022-10-11
Add Change note and publishing controls in the GOV.UK Design System

davidgisbey:
## Description

This PR adds partials for the change note section and the publishing controls (save and cancel button/links) in the GOV.UK Design System.

I've got for a like for like approach with what is currently there after yesterdays discussion.

## Screenshots

image

## Trello card


app/helpers/admin/editions_helper.rb
```diff
@@ -222,6 +222,19 @@ def edition_information(information)
...
   def standard_edition_publishing_controls(form, edition)
+    tag.div(class: "publishing-controls") do
+      if edition.change_note_required?
+        concat render("change_notes", form:, edition:)
+      end
+      if current_user.can_redirect_to_summary_page?
+        concat render("save_or_cancel", form:, edition:)
+      else
+        concat form.save_or_continue_or_cancel
+      end
```
ollietreend, line 233: Do we need this conditional here? Isn't the 'redirect to summary page' feature flag going away [as part of this ticket](https://trello.com/c/6lZphXit/806-revert-changes-to-save-redirect)?

Also: I think `form.save_or_continue_or_cancel` will render Bootstrap components, which is probably not what we want since this is used on Design System pages.

https://github.com/alphagov/whitehall/blob/ad86a5ea8d7538f787c71cb586adb1b771ea08d9/lib/whitehall/form_builder.rb#L71-L74

As far as I'm aware, no Design System page should make use of the Whitehall Form Builder, right?

ollietreend (commented):
Looks good 👍🏻 just one thing I wasn't sure of.

app/helpers/admin/editions_helper.rb
davidgisbey: I think this should actually use the can_preview_design_system method. I've also updated it so use the `Save and continue` button

ollietreend (approved):

PR #6921, opened 2022-10-13
Separate the asset files (CSS & JS) for Bootstrap pages vs. Design System pages

DilwoarH:
## What

We need to separate out the assets used on the Bootstrap vs. Design System page layouts.

### New file naming conventions

No changes needed to frontend assets. For admin assets, use these conventions:

**Bootstrap assets**

- JavaScript
  - Main JS file: `app/assets/javascripts/admin_legacy.js`
  - Root JS directory: `app/assets/javascripts/admin_legacy/`
- Stylesheets
  - Main CSS file: `app/assets/stylesheets/admin_legacy.scss`
  - Root SCSS directory: `app/assets/stylesheets/admin_legacy/`

**Design System assets**

- JavaScript
  - Main JS file: `app/assets/javascripts/application.js`
  - Root JS directory: `app/assets/javascripts/admin/`
  - Subdirectories for components, views, etc.
- Stylesheets
  - Main CSS file: `app/assets/stylesheets/application.scss`
  - Root SCSS directory: `app/assets/stylesheets/admin/`
  - Subdirectories for components, views, etc.

### Split out files

Files should be split so that:

- Design System layout uses `application.js`
- Bootstrap layout uses `admin.js`
- `application.js` no longer imports `admin.js` (i.e. we should remove the `//= require admin` directive)

### Nav bar

We'll still need some CSS & JavaScript to support the nav bar, which is shared between both layouts.

Ideally this behaviour could be re-implemented without requiring jQuery – but if it ends up too complicated, then it'd be fine for us to just import the existing behaviour & jQuery.

### Other functionality

If there's any other functionality that the Design System implementation depends on from Bootstrap, we should port it across. (If jQuery is required, we should try to reimplement is with vanilla JS.)

I don't expect there will be much functionality that falls into this category – but it's worth keeping in mind. We should be especially careful to ensure Analytics tracking behaviour remains functional, since this isn't well covered by tests.

## Why

Right now, the Design System layout includes all of the JS used in the Bootstrap layout. We run the risk of the old JS interfering with the new JS, introducing unexpected behaviour or even implementing behaviour that we didn't know existed.

We've already seen examples where the Bootstrap JS is doing unwanted things on Design System pages – e.g. https://github.com/alphagov/whitehall/commit/0ec847dc9d1974ce0774f0043dc09a84a385d5e8

https://trello.com/c/3UW25yLj/779-separate-the-asset-files-css-js-for-bootstrap-pages-vs-design-system-pages


ollietreend (approved):
Nice work untangling this all @DilwoarH ✨ 🙌🏻 

This looks good to me.

PR #6922, opened 2022-10-14
Add Kwasi Kwarteng to Past Chancellors List

Aurorum:
Adds Kwasi Kwarteng to this list following his departure from government: https://www.gov.uk/government/history/past-chancellors


chao-xian:
Hi @Aurorum thanks for this change! We've checked this is ok to release and I'm about to approve and merge it. Once that's done it'll go to the live site shortly after that.

chao-xian (approved):

PR #6941, opened 2022-10-21
Replace Back buttons with Simple Cancel buttons in GOV.UK Design System pages

dnkrj:

## What
[Replace back buttons on the edition tags, attachment confirm destroy and attachment reorder pages with cancel buttons.
Remove the back button from new and edit attachment pages.](https://trello.com/c/FJaufMPB/813-replace-back-buttons-with-cancel-button-in-govuk-design-system-pages)

## Why
In our like for like approach we're using the same buttons as before for the new design system pages.
The Cancel button behaviour on these pages is simple always going back to the same page.

### After
whitehall-admin dev gov uk_government_admin_editions_1370963_tags_edit(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1370963_tags_edit(iPhone 12 Pro)
whitehall-admin dev gov uk_government_admin_editions_1374127_attachments_national-underground-asset-register-project-update-october-2022_confirm_destroy(iPad Pro)](https://user-images.githubusercontent.com/9594455/197165381-49c4d797-a4a1-4d8f-a9db-36fd0ee84e16.png)
usercontent.com/9594455/197165321-94af795c-3560-42b1-8ce0-58f653c947c4.png)
whitehall-admin dev gov uk_government_admin_editions_1374127_attachments_reorder(iPad Pro)

### Before
whitehall-admin dev gov uk_government_admin_editions_1370963_tags_edit(iPad Pro) (1)
whitehall-admin dev gov uk_government_admin_editions_1370963_tags_edit(iPhone 12 Pro) (2)
whitehall-admin dev gov uk_government_admin_editions_1374127_attachments_national-underground-asset-register-project-update-october-2022_confirm_destroy(iPad Pro) (2)
whitehall-admin dev gov uk_government_admin_editions_1374127_attachments_new(iPad Pro) (2)
whitehall-admin dev gov uk_government_admin_editions_1374127_attachments_reorder(iPad Pro) (1)


nnagewad:
Looks good

app/views/admin/needs/edit.html.erb
```diff
@@ -1,9 +1,6 @@
 <% content_for :page_title, @edition.title %>
 <% content_for :title, "Associated user needs" %>
 <% content_for :context, @edition.title %>
-<% content_for :back_link do %>
-  <%= render "govuk_publishing_components/components/back_link", href: admin_edition_path(@edition) %>
```
DilwoarH, line 5: Do we need to add a cancel button here?

app/views/admin/attachments/edit.html.erb
```diff
@@ -1,9 +1,6 @@
 <% content_for :page_title, "Editing #{attachment.attachable_model_name} #{attachment.readable_type} attachment" %>
 <% content_for :title, "Edit #{attachment.readable_type} attachment" %>
 <% content_for :context, "Attachments for #{attachment.attachable_model_name}" %>
-<% content_for :back_link do %>
-  <%= render "govuk_publishing_components/components/back_link", href: attachable_attachments_path(attachable) %>
-<% end %>
```
DilwoarH, line 6: The [forms](https://github.dev/alphagov/whitehall/blob/ad6e0737bd5191efae51a96ed2fa9ddf615a54d8/app/views/admin/attachments/_form.html.erb) partial doesn't seem to have a cancel button, do we need to update that aswell?

DilwoarH (changes requested):
Looks like some of these there are a few pages where are no `cancel` or `back` link. This means the user won't have a way to go back or get out of a flow.

app/views/admin/attachments/edit.html.erb
dnkrj: Good catch, it was removed here:
https://github.com/alphagov/whitehall/commit/ff67c0b95d575b0787934d0e8882707597e26d9e#diff-747aa398dc59115a77970d9f5e8e860d990f10cd6b79c1482f36444b8d31bfddL69

Have added it back in.

app/views/admin/needs/edit.html.erb
dnkrj: The bootstrap page doesn't have a cancel button, JP agreed that with like for like we shouldn't introduce one if there wasn't one there before.

DilwoarH:
> Looks like some of these there are a few pages where are no `cancel` or `back` link. This means the user won't have a way to go back or get out of a flow.

Hey - I know we talked about this elsewhere (not sure where) but you mentioned that you wanted to do fully `like for like` but I think we should have some sort of navigation flow for individual pages. Happy to take this off GitHub to discuss.

Here is an example where I think we need it:

image

davidgisbey:
Agreed. If we should add in a cancel link (since that seems to be the pattern we're going with) if there isn't one.

dnkrj:
I've reverted the changes to the user needs page.
[I've added a card to handle this properly.](https://trello.com/c/PZ4Q11UG/835-add-cancel-button-to-user-needs-page)

PR #6866, opened 2022-09-28
Add confirm delete page for attachments

dnkrj:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

[Introduce confirm delete page for attachments using the new design system.](https://trello.com/c/gCDiXR7n/636-add-confirm-delete-page-for-attachments)
The PR does not use the confirm delete page or introduce a bootstrap version of the page - the old attachment index page can continue to use a modal.

## Why

The new design system does not use modals. Currently confirmation that an attachment should be deleted is performed using a modal. This PR introduces a new page to be linked to in the new list of attachments.

## Screenshots

whitehall-admin dev gov uk_government_admin_editions_1_attachments_2_confirm_destroy(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1_attachments_2_confirm_destroy(iPhone 12 Pro)

PR #6943, opened 2022-10-24
Move attachments index page to gov.uk design system

baisa:
[Trello card](https://trello.com/c/rgcVKahA/686-move-attachments-index-page-to-the-govuk-design-system)

Screenshot 2022-10-31 at 12 36 50

### DIFF
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -6,7 +6,9 @@ class Admin::AttachmentsController < Admin::BaseController
 
   rescue_from Mysql2::Error, with: :handle_duplicate_key_errors_caused_by_double_create_requests
 
-  def index; end
+  def index
+    render(preview_design_system_user? ? "index" : "index_legacy")
+  end
 
   def reorder; end
 
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -7,7 +7,7 @@ class Admin::AttachmentsController < Admin::BaseController
   rescue_from Mysql2::Error, with: :handle_duplicate_key_errors_caused_by_double_create_requests
 
   def index
-    render "index_legacy" unless preview_design_system_user?
+    render_design_system("index", "index_legacy", next_release: false)
   end
 
   def reorder; end
@@ -92,6 +92,7 @@ private
 
   def get_layout
     design_system_actions = %w[edit update new create confirm_destroy reorder]
+    design_system_actions << "index" if preview_design_system?(next_release: false)
     if preview_design_system?(next_release: true) && design_system_actions.include?(action_name)
       "design_system"
     else
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -7,7 +7,7 @@ class Admin::AttachmentsController < Admin::BaseController
   rescue_from Mysql2::Error, with: :handle_duplicate_key_errors_caused_by_double_create_requests
 
   def index
-    render(preview_design_system_user? ? "index" : "index_legacy")
+    render "index_legacy" unless preview_design_system_user?
   end
 
   def reorder; end
--- a/app/views/admin/attachments/_attachments.html.erb
+++ b/app/views/admin/attachments/_attachments.html.erb
@@ -1,45 +1,38 @@
-<% initialise_script "GOVUK.ReorderableAttachmentsList", el: "ol.existing-attachments" if attachable.can_order_attachments? && attachable.attachments.many? %>
+<%= render "govuk_publishing_components/components/heading", {
+  text: "Attachments",
+  font_size: "l",
+  margin_bottom: 3,
+} %>
 
 <% if attachable.can_order_attachments? && attachable.attachments.many? %>
-  <p class="alert alert-info">
-    To change the order of your attachments, drag them up or down and then click ‘Save attachment order’ at
-    the bottom of the page.
-  </p>
+<p class="govuk-body"><%= link_to 'Reorder attachments', reorder_admin_edition_attachments_path(@edition), class: "govuk-link govuk-link--no-visited-state" %></p>
 <% end %>
 
-<%= form_tag [:order, :admin, typecast_for_attachable_routing(attachable), Attachment], method: :put do %>
-  <ol class="existing-attachments list-unstyled">
-    <% attachable.attachments.includes(:attachment_data).each do |attachment| %>
-      <%= content_tag_for(:li, attachment.becomes(Attachment), class: "well") do %>
-        <strong><%= attachment.title %></strong>
-        <span class="file">
-          <%= attachment.is_a?(ExternalAttachment) ? 'Link' : 'File' %>: <%= link_to_attachment(attachment, preview: true, full_url: true) %>
-        </span>
-        <%= attachment_uploading_status(attachment) %>
-        <%= attachment_metadata_tag(attachment) %>
-
-        <% if attachable.can_order_attachments? %>
-          <%= label_tag "ordering-#{attachment.id}", "Order:" %>
-          <%= text_field_tag "ordering[#{attachment.id}]", attachment.ordering, id: "ordering-#{attachment.id}", class: "ordering" %>
-        <% end %>
-
-        <ul class="actions list-unstyled">
-          <li>
-            <%= link_to 'Edit', [:edit, :admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)] %>
-          </li>
-          <li>
-            <%= link_to 'Delete',
-                  [:admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)],
-                  method: 'delete',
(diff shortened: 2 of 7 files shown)


features/step_definitions/attachment_steps.rb
```diff
@@ -63,12 +63,11 @@
 end
 
 Then(/^the attachments should be in the following order:$/) do |attachment_list|
```
davidgisbey, line 65: Once the above is reverted you can test the output of both by using Ollies new helper `using_design_system?` added here https://github.com/alphagov/whitehall/pull/6953

```
if using_design_system?
  attachment_names = all("table td:first").map(&:text)

  attachment_list.hashes.each_with_index do |attachment_info, index|
    attachment = Attachment.find_by(title: attachment_info[:title])
    expect(attachment.title).to eq(attachment_names[index])
  end
else
   attachment_ids = all(".existing-attachments > li").map { |element| element[:id] }
  
   attachment_list.hashes.each_with_index do |attachment_info, index|
     attachment = Attachment.find_by(title: attachment_info[:title])
     expect("attachment_#{attachment.id}").to eq(attachment_ids[index])
  end
end  
```

app/views/admin/attachments/_attachments.html.erb
```diff
@@ -1,45 +1,37 @@
 <% initialise_script "GOVUK.ReorderableAttachmentsList", el: "ol.existing-attachments" if attachable.can_order_attachments? && attachable.attachments.many? %>
```
dnkrj, line 1: Is this still needed for the design system page?

app/views/admin/attachments/index.html.erb
```diff
@@ -1,46 +1,47 @@
...
+    </ul>
 
-      <div class="alert alert-warning">
-        <p>
-          You must upload attachments in an <a href="https://www.gov.uk/guidance/content-design/planning-content#open-formats">open standards format</a>.<br/>
-        </p>
-        <p>
-          For example, if an attachment is text-based and designed to be edited it should be uploaded to GOV.UK as an .odt (OpenDocument text) file instead of a closed format like .docx.
-        </p>
+    <details class="govuk-details" data-module="govuk-details">
```
davidgisbey, line 33: I think we can use the details component here so any changes to the design system will be applied when the publishing components gem is updated

https://components.publishing.service.gov.uk/component-guide/details


```
  <%= render "govuk_publishing_components/components/details", {
      title: " An important note regarding uploading attachments",
    } do %>
      <p>You must upload attachments in an open standards format. </p>
      <p>For example, if an attachment is text-based and designed to be edited it should be uploaded to GOV.UK as an .odt (OpenDocument text) file instead of a closed format like .docx.</p>
      <p>Read more about <%= link_to "open standards formats", "https://www.gov.uk/guidance/content-design/planning-content#open-formats", class: "govuk-link govuk-link--no-visited-state" %>.</p>
    <% end %>
```

davidgisbey:
Amazing work. I've left a couple of comments, but I gave it a manual test and it's working really nicely 👍

app/views/admin/attachments/_attachments.html.erb
```diff
@@ -1,45 +1,37 @@
...
-    <%= submit_tag 'Save attachment order', class: 'btn btn-primary' %>
-  <% end %>
-
-<% end %>
+<%= render "govuk_publishing_components/components/table", {
+  rows: attachable.attachments.includes(:attachment_data).map do |attachment|
+    [
+      {
+        text: attachment.title
+      },
```
davidgisbey, line 18: If we're going like for like i wonder if this should link to the file directly 

```
 text: link_to(link_to 'attachment.title', attachment.url(preview: true, full_url: true), class: "govuk-link govuk-link--no-visited-state"),
        format: "numeric"
```

and then the next column could be removed? Probs one for JP

app/views/admin/attachments/_attachments.html.erb
```diff
@@ -1,45 +1,37 @@
...
+      {
+        text: link_to('Edit', [:edit, :admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], class: "govuk-link govuk-link--no-visited-state"),
+        format: "numeric"
+      },
+      {
+        text: link_to('Delete',
+          [:admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)],
+          class: "govuk-link govuk-link--no-visited-state",
+          method: 'delete',
+          data: { confirm: 'Are you sure you want to delete this attachment?' }),
```
dnkrj, line 32: We don't use the system modal in the design system.
This should link to the new `confirm_destroy` page.
https://github.com/alphagov/whitehall/pull/6866

app/views/admin/attachments/_attachments.html.erb
```diff
@@ -1,45 +1,37 @@
...
+      {
+        text: attachment.title
+      },
+      {
+        text: link_to(link_to 'Show', attachment.url(preview: true, full_url: true), class: "govuk-link govuk-link--no-visited-state"),
+        format: "numeric"
+      },
+      {
+        text: link_to('Edit', [:edit, :admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], class: "govuk-link govuk-link--no-visited-state"),
+        format: "numeric"
```
davidgisbey, line 25: I think we should remove ` format: "numeric"` for these as they're not numbers. It slightly changes the styling as it reduces padding, but i don't think it's an issue 

## before

image


## after

image

dnkrj (changes requested):
Just a few more changes to add David's, otherwise looks good 👍

app/views/admin/attachments/_attachments.html.erb
davidgisbey, line 32: Not something we need to worry about in this PR but there was a card for adding a confirm delete page for attachments somewhere 🤔 i'll have a look for it

app/views/admin/attachments/_attachments.html.erb
```diff
@@ -1,45 +1,37 @@
...
+  font_size: "l",
+  margin_bottom: 3,
+} %>
+
 <% if attachable.can_order_attachments? && attachable.attachments.many? %>
-  <p class="alert alert-info">
-    To change the order of your attachments, drag them up or down and then click ‘Save attachment order’ at
-    the bottom of the page.
-  </p>
+  <%= link_to 'Reorder attachments', reorder_admin_edition_attachments_path(@edition), class: "govuk-link govuk-link--no-visited-state" %>
```
dnkrj, line 10: This should use a href with the design system classes rather than the bootstrap template.

davidgisbey: link_to is actually a rails helper https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to rather than a Bootstrap template so i think we're all good here 

```
link_to 'Reorder attachments', reorder_admin_edition_attachments_path(@edition), class: "govuk-link govuk-link--no-visited-state"
```

becomes

```
 <a class="govuk-link govuk-link--no-visited-state" href="/government/admin/editions/1288858/attachments/reorder">Reorder attachments</a>
```

dnkrj: Oh, thanks David

app/views/admin/attachments/_attachments.html.erb
baisa: Well spotted @dnkrj

app/views/admin/attachments/_attachments.html.erb
baisa: We have agreed with JP and Nikin for the look like in the screenshot above 👍

features/step_definitions/attachment_steps.rb
baisa: Done, thank you @davidgisbey 👍

app/controllers/admin/attachments_controller.rb
```diff
@@ -6,7 +6,9 @@ class Admin::AttachmentsController < Admin::BaseController
 
   rescue_from Mysql2::Error, with: :handle_duplicate_key_errors_caused_by_double_create_requests
 
-  def index; end
+  def index
+    render_design_system("index", "index_legacy", next_release: true)
```
dnkrj, line 10: I believe that this the index page is release 1.4 so should not be included in the next release yet

app/controllers/admin/attachments_controller.rb
```diff
@@ -89,7 +91,7 @@ def attachable_attachments_path(attachable)
 private
 
   def get_layout
-    design_system_actions = %w[edit update new create confirm_destroy reorder]
+    design_system_actions = %w[index edit update new create confirm_destroy reorder]
```
dnkrj, line 94: We'll also need to handle that the index page is not in the same release as the other pages here.

dnkrj (commented):
Looks good other than the release 👍

davidgisbey:
Amazing work. Looks great to me! Like Daniel said there's a couple of changes around release, but it's good to once they're actioned 👍

dnkrj (approved):

dnkrj:
Handling of the `get_layout` method looks great 🎉

davidgisbey (approved):
lgtm

PR #6947, opened 2022-10-24
Move error summary into a view component

davidgisbey:


app/components/admin/error_summary_component.rb
```diff
@@ -0,0 +1,56 @@
...
+      error_item
+    end
+  end
+
+  def analytics_action
+    @analytics_action ||= "#{humanized_class_name}-error"
+  end
+
+  def parent_class
+    @parent_class ||= @parent_class.presence || object.class.to_s.underscore
```
ollietreend, line 54: Can this be simplified, or is it subtly different?

```suggestion
    @parent_class ||= object.class.to_s.underscore
```

Presumably `@parent_class` will return if it already has a value, so no need to call `#presence` on it?

ollietreend (approved):
Nice work @davidgisbey 🎉 this looks good to me

PR #6948, opened 2022-10-24
Add Datetime View Component

davidgisbey:


test/components/admin/datetime_component_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    edition = build_stubbed(:edition, first_published_at: nil)
+    render_inline(
+      Admin::DatetimeComponent.new(
+        object: edition,
+        field_name: "first_published_at",
+      ),
+    )
+
+    assert_equal page.all("h3")[0].text, "Date"
+    assert_equal page.all("label")[0].text, "Day"
```
kevindew, line 51: Would suggest you don't need these since this isn't code affected by the inputs to the component or is it the scenario described above.

These don't seem things that are worth testing anyway since they essentially are just confirming that you typed the same header and label in your HTML and aren't controlled by logic.

test/components/admin/datetime_component_test.rb
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    assert_equal page.all("select")[0].value, "1"
+    assert_equal page.all("select")[1].value, "1"
+    assert_equal page.all("select")[2].value, "2020"
+    assert_equal page.all("select")[3].value, "12"
+    assert_equal page.all("select")[4].value, "30"
+  end
+
+  test "uses the start year and and year to populate the year fields options if passed in" do
```
kevindew, line 118:
```suggestion
  test "uses the start year and and end year to populate the year fields options if passed in" do
```

test/components/admin/datetime_component_test.rb
```diff
@@ -0,0 +1,137 @@
...
+        end_year: 2022,
+      ),
+    )
+
+    [*1999..2022].each do |year|
+      assert_selector "option[value='#{year}']", count: 1
+    end
+
+    assert_selector "option[value='2023']", count: 0
+    assert_selector "option[value='1998']", count: 0
```
kevindew, line 135:
```suggestion
    assert_selector "option[value='1999'], option[value='2022']", count: 2
    assert_selector "option[value='1998'], option[value='2023']", count: 0
```

I don't think we need to do the extent of assertions we're doing here as we're just testing the Rails helper, which we already rather trust.

app/components/admin/datetime_component.html.erb
```diff
@@ -0,0 +1,64 @@
+<div class="gem-c-date-input govuk-date-input govuk-datetime <%= 'govuk-form-group--error' if errors_for_input(object.errors, field_name).present? %>">
+  <h3 class="govuk-heading-s govuk-!-margin-bottom-2 govuk-!-margin-top-4">Date</h3>
+
+  <div class="gem-c-hint govuk-hint govuk-!-margin-bottom-2">
+    For example, 01 August 2022
+  </div>
+
+  <% if errors_for_input(object.errors, field_name).present? %>
```
kevindew, line 8: should this errors_for_input check be something in the component code itself. It seems to do processing in the helper so seems like something that could be helpfully memo'd in the component.

app/components/admin/datetime_component.rb
```diff
@@ -0,0 +1,62 @@
...
+    default_date&.day
+  end
+
+  def default_hour
+    default_date&.hour
+  end
+
+  def default_minute
+    default_date&.min
+  end
```
kevindew, line 57: This all looks rather verbose.

Could all these methods be replaced with calling the date precision in the view?

I'd imagine you could cut all this down to
```
def date_value
  object.public_send(field_name) || default_date
end
```

and then to just need `&.<method>` in the ERB?

kevindew (commented):
Thanks so much for putting this together ⭐ 

Really interesting to see example implementation. I've reviewed it a bit like it was going in for real but don't worry about dealing with the more nitpicky implementation comments until we work all the bits out.

davidgisbey:
this is now being added as a publishing component

PR #6950, opened 2022-10-24
Remove code for finders

brucebolt:
The English finders are already been rendered by Finder Frontend.  The code for finders was retained solely so non-English finders could be rendered (as their content is not present in Search API, so couldn't be rendered from Finder Frontend).

FCDO have agreed these finders can be switched off, so this PR removes that code and other related code that is no longer used.

The routes will need to be unpublished using the rake task in https://github.com/alphagov/whitehall/pull/6952 before this code is merged.

Not to be merged until all World Location News and International Delegation pages are rendered by Collections, as covered by https://github.com/alphagov/whitehall/pull/6961.

[Trello card](https://trello.com/c/SZ4uUhag)


features/step_definitions/fatalities_steps.rb
```diff
@@ -1,3 +1,10 @@
+Given(/^I can navigate to the list of announcements$/) do
```
JonathanHallam: Did you say yesterday that you were going to remove this?

test/integration/navigation_selection_test.rb
```diff
@@ -6,20 +6,16 @@ class NavigationSelectionTest < ActiveSupport::TestCase
...
   TESTED_CONTROLLERS = %w[
-    announcements
     consultations
     corporate_information_pages
     email_signup_information
     histories
     home
     latest
     ministerial_roles
-    news_articles
```
JonathanHallam, line 17: It'd be good to have why these are being removed in the commit message (I'm don't actually know why these are being removed)

app/helpers/email_signup_helper.rb
```diff
@@ -5,9 +5,10 @@ def email_signup_path(atom_feed_url)
     if mhra_exists_for_atom_feed?(atom_feed_url)
       mhra_email_signup_path_from_atom_feed(atom_feed_url)
     elsif feed_uri.path.match(%r{^/world/(.*)\.atom$})
-      new_email_signups_path(email_signup: { feed: atom_feed_url })
```
JonathanHallam, line 8: In the PR description you say that this shouldn't be merged until non english World Location News pages are rendered by Collections. The Commit message says this needs to be done because we're still rendering non English WL News pages in Whitehall. 

Given that this is dependent on our next migration, can this method just be removed if it's only for non english world location news pages

test/support/document_controller_test_helpers.rb
```diff
@@ -308,93 +308,7 @@ def should_show_inapplicable_nations(document_type)
       end
     end
 
-    def should_paginate(edition_type, options = {})
```
JonathanHallam, line 311: Should all of this removal be in this commit or is the pagination one more suitable?

lib/whitehall/announcement_filter_option.rb
```diff
@@ -1,12 +0,0 @@
-module Whitehall
```
JonathanHallam, line 1: Is there any database cleanup we can do as a result of this?

features/step_definitions/fatalities_steps.rb
brucebolt: Looks like I did, but the removal was in the wrong commit (f78659554ef34e2dd3fc1ae8f1d7c8bfa50fab95 instead of this one).  I'll update that now.

test/integration/navigation_selection_test.rb
brucebolt: The reason for removing that is the same document types were removed from the tested method in `application_helper.rb`: https://github.com/alphagov/whitehall/pull/6950/commits/46b4ac87b96faf54577dfd8aea12aee588aa23e9#diff-5f3fc5e3977d242572aa1d08551f5eb557de0ccaff30370838ee9df5386ea0da.

As far as I could see, these methods were not being used anywhere (except in tests) and the method referred to a route helper which was deleted in the same commit.

I'll update the commit message to include this information.

brucebolt: Looking at this more closely, I think this code can be removed completely as it is unused.  I've raised https://github.com/alphagov/whitehall/pull/6971 to do that and will rebase this branch once merged.

test/integration/navigation_selection_test.rb
```diff
@@ -15,10 +14,6 @@ class NavigationSelectionTest < ActiveSupport::TestCase
     ministerial_roles
     organisations
     past_foreign_secretaries
-    publications
```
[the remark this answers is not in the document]
brucebolt: As above, this code will all be removed in https://github.com/alphagov/whitehall/pull/6971.

app/helpers/email_signup_helper.rb
brucebolt: I've now realised the entire `EmailSignupHelper` can be removed (which I've done in [this commit(https://github.com/alphagov/whitehall/pull/6950/commits/8dbea2de034ce4e99db83f7746deafdb1f2319cc)).

That then led me to discover some partials that weren't being used, which have been removed in [this commit](https://github.com/alphagov/whitehall/pull/6950/commits/1d785d8870ccf5ace4e47cf017e22ec0f5ab4930).

Another 144 lines of code deleted 🎉!

test/support/document_controller_test_helpers.rb
brucebolt: I think this should stay here, as it's a helper for view tests where the view has been deleted.  There's no further use of this helper elsewhere, so I deleted it at the same time as the view tests.

lib/whitehall/announcement_filter_option.rb
brucebolt: I don't think so.  None of this was committed to the database, it seemed to just be held in memory whilst the application ran.

JonathanHallam (approved):
this is excellent!!

kevindew:
Amazing job 🎉 but I'm a little sad the YOLO commit has gone.

PR #6951, opened 2022-10-24

beccapearce:
When running tests in parallel the scheduled publishing worker test has a tendency to randomly fail. This seems to be because old editions are not properly discarded or are running at the same time which causes the test to fail.

To fix this we have have added a method to pass in a different worker name to the test sidekiq, this means that the editions should no longer overlap since they are in different queues.

Co-authored-by: BeckaL <becka.lelew@digital.cabinet-office.gov.uk >

### DIFF
--- a/test/support/sidekiq_test_helpers.rb
+++ b/test/support/sidekiq_test_helpers.rb
@@ -2,10 +2,10 @@ require "govuk_sidekiq/testing"
 require "sidekiq/api"
 
 module SidekiqTestHelpers
-  def with_real_sidekiq
+  def with_real_sidekiq(worker_name = "whitehall-test")
     Sidekiq::Testing.disable! do
       Sidekiq.configure_client do |config|
-        config.redis = { namespace: "whitehall-test" }
+        config.redis = { namespace: worker_name }
       end
 
       Sidekiq::ScheduledSet.new.clear
--- a/test/unit/workers/scheduled_publishing_worker_test.rb
+++ b/test/unit/workers/scheduled_publishing_worker_test.rb
@@ -3,6 +3,21 @@ require "test_helper"
 class ScheduledPublishingWorkerTest < ActiveSupport::TestCase
   include SidekiqTestHelpers
 
+  class << self
+    attr_writer :worker_number
+
+    def worker_number
+      @worker_number || 0
+    end
+  end
+
+  parallelize_setup do |worker|
+    # The parallelize_setup block runs in the context of the class, not the
+    # instance. So we have to use a class variable rather than an instance
+    # variable to store the worker number.
+    self.worker_number = worker
+  end
+
   setup do
     @publishing_robot = create(:scheduled_publishing_robot)
   end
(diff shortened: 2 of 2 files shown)


kevindew:
Amazing that you've been looking into this ⭐. I'm a little worried that this solution relies on redis namespace as that feature is going to be disabled in Sidekiq 7 (the most recent release): https://github.com/mperham/sidekiq/blob/main/docs/7.0-Upgrade.md#redis-namespace

Would you be able to summarise the source of the flaky test? I wonder if we can fix them by making the tests more specific rather than changing the architecture?

beccapearce:
Darn! Well as far as I can see the tests fail because there is another test running at the same time (due to the parallelisation) which is adding edition to the queue so our numbers of the queue length are out. 

I think there are a few other ways to tackle this: We could just check for the specific ids of the editions that we are adding and removing from the queues, there's also another option about rewriting the test, I'm a bit dubious of this `with_real_sidekiq` method, I would say that we shouldn't be using real sidekiq in unit tests. I can see why it was done but i's be tempted to rewrite it especially given that we are losing that namespace method.

I'll have another think and try some approaches out to see if I can get one that passes.

kevindew:
Sure that makes sense, so we're asserting on global state that another test might change.

Which tests is it that are prone to failing?

beccapearce:
There's a couple just to make it more annoying: 
https://github.com/alphagov/whitehall/blob/20a0a69ecef9cfa7303ea18a5a536d842e9b4bc7/test/unit/workers/scheduled_publishing_worker_test.rb#L60-L78

https://github.com/alphagov/whitehall/blob/20a0a69ecef9cfa7303ea18a5a536d842e9b4bc7/test/unit/workers/scheduled_publishing_worker_test.rb#L97-L105

and https://github.com/alphagov/whitehall/blob/20a0a69ecef9cfa7303ea18a5a536d842e9b4bc7/test/unit/workers/scheduled_publishing_worker_test.rb#L107-L114 seem to be the main culprits.

kevindew:
Ok I've had a look. The class Sidekiq::ScheduledSet always talks to real Redis, even if Sidekiq is in fake mode. We do have access to Sidekiq::Queues however in fake mode.

I imagine the best option is to apply mocking to the tests to cut the need out for the deep integration. Mocking is a bit contrived but it seems better than needing a deep integration with Redis in a high level unit test.

A quick example could be:

```
  test ".dequeue removes a job for a scheduled edition" do
    edition = create(:scheduled_edition)
    edition_job = Sidekiq::SortedEntry.new({}, 1, "class" => "ScheduledPublishingWorker", "args" => [edition.id])
    control_job = Sidekiq::SortedEntry.new({}, 1, "class" => "ScheduledPublishingWorker", "args" => [edition.id + 1])
    Sidekiq::ScheduledSet.stubs(:new).returns([edition_job, control_job])

    edition_job.expects(:delete)

    ScheduledPublishingWorker.dequeue(edition)
  end
```

Which then won't talk to real Redis anymore and still tests the `dequeue` method.

kevindew:
Another thought I’ve had is: do we even need to remove items from the queue? It seems such a minor operation checking whether something needs publishing that it’s probably just fine to let the item happen and have no action.

In some ways I’m wondering if this extra complexity we have in trying to clear a queue is actually revealing that this worker might not do sufficient checks that an item is allowed to be published. 

I’d wonder if we just made sure the edition was scheduled and time was equal or after the publish time then we could remove the ScheduledSet stuff

PR #6952, opened 2022-10-24
Add temporary rake task to unpublish non-English finders

brucebolt:
This rake task will need to be run in production (then deleted) and serves the purpose of marking the non-English finders (and their equivalent atom and JSON feeds) as 'gone' (410 response code).

These finders don't actually have content items at the moment (the content item redirects to `/government` so we automatically fall back to Whitehall), so we first need to publish a content item for each translation before marking it as 'gone'.

FCDO have agreed these finders can be switched off, so this PR stops us from serving the finders at those routes by marking them as "gone" (which has the effect of changing the rendering application to `government-frontend`).

Syntax for running the task:
```
rake finders:temp_unpublish_non_english_finders
```

Once run in production, the rake task can be removed.

| Screenshot of a French language finder that has been marked as 'gone' |
| -- |
| Screenshot showing page marked as gone.  The text (in French) says La page que vous recherchez n'est plus disponible. Les informations présentées sur cette page ont été supprimées car elles ont été publiées par erreur. |

This task is to be run before https://github.com/alphagov/whitehall/pull/6950 is merged.

### DIFF
--- a/lib/tasks/publish_finders.rake
+++ b/lib/tasks/publish_finders.rake
@@ -8,4 +8,61 @@ namespace :finders do
       PublishFinder.call(content_item)
     end
   end
+
+  desc "Temporary task to unpublish non-English finders - this can be removed after running in production"
+  task temp_unpublish_non_english_finders: :environment do
+    announcements_finder_content_id = "88936763-df8a-441f-8b96-9ea0dc0758a1"
+    publications_finder_content_id = "b13317e9-3753-47b2-95da-c173071e621d"
+
+    locales = Locale.non_english.map(&:code)
+
+    locales.each do |locale|
+      puts "Publishing content item for /government/announcements.#{locale}"
+      publish_finder_for_locale(announcements_finder_content_id, "/government/announcements.#{locale}", locale)
+
+      puts "Unpublishing /government/announcements.#{locale}"
+      Whitehall::PublishingApi.publish_gone_async(announcements_finder_content_id, "/search/news-and-communications", nil, locale.to_s)
+
+      puts "Publishing content item for /government/publications.#{locale}"
+      publish_finder_for_locale(publications_finder_content_id, "/government/publications.#{locale}", locale)
+
+      puts "Unpublishing /government/publications.#{locale}"
+      Whitehall::PublishingApi.publish_gone_async(publications_finder_content_id, "/search/all", nil, locale.to_s)
+    end
+  end
+end
+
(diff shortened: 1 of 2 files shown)


test/unit/tasks/publish_finders_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  test "it publishes then marks as gone each non-English finder" do
+    Locale.stubs(:non_english).returns([
+      Locale.new(:cy),
+      Locale.new(:fr),
+    ])
+
+    Services.publishing_api.expects(:put_content).with("88936763-df8a-441f-8b96-9ea0dc0758a1", has_entries(
+                                                                                                 base_path: "/government/announcements.cy",
```
BeckaL, line 16: I find these tests a bit noisy but I'm not super fussed given we're going to remove them fairly quickly (and we're actually good at removing code now✨)

lib/tasks/publish_finders.rake
```diff
@@ -8,4 +8,58 @@ namespace :finders do
...
     end
   end
+
+  desc "Temporary task to unpublish non-English finders - this can be removed after running in production"
+  task temp_unpublish_non_english_finders: :environment do
+    locales = Locale.non_english.map(&:code)
+
+    locales.each do |locale|
+      puts "Publishing content item for /government/announcements.#{locale}"
+      publish_finder_for_locale("88936763-df8a-441f-8b96-9ea0dc0758a1", "/government/announcements.#{locale}", locale)
```
BeckaL, line 18: I'd be tempted to pull out these content ids and name them as variables so they're a bit less magic. 
But again, given temp, not super fussed

BeckaL (approved):
Logic looks good to me. Couple of comments about readability, but up to you whether you want to address them or not given temp (must be painful adding lines back in for you now 😂)

test/unit/tasks/publish_finders_test.rb
brucebolt: Yeah, I did think about DRY-ing them up, but as the code will only get run once then deleted, it didn't seem to be worth the effort.

lib/tasks/publish_finders.rake
brucebolt: That's a good idea, I'll do that now to make sure we're referring to the same content ID continuously.

brucebolt:
> must be painful adding lines back in for you now 😂

At least it's only temporary code 😁

PR #6956, opened 2022-10-25
Remove unused images

brucebolt:
This removes images that used to be rendered by Whitehall, but aren't anymore.

For each deleted image, I've checked to make sure we are not using it anywhere in the Whitehall codebase.

There are more images that aren't being rendered, but are still mentioned in CSS files (albeit CSS that is never rendered), but that feels outside the scope of this PR.


BeckaL (approved):
Nice one, agreed that these all look unused, cleanup full steam ahead! ⭐

PR #6957, opened 2022-10-25
Remove unused code

brucebolt:
Whilst removing code for the finders in https://github.com/alphagov/whitehall/pull/6950/commits, I came across other pieces of code that are unused.

These are therefore safe to remove now, before the finders are switched off.


BeckaL (approved):
Yup, all looks good to me, double checked all the unused stuff and I concur that I cannot find any usages, was a bit worried about dynamic invocations of the filter paths but think [this](https://github.com/alphagov/whitehall/blob/18ee16237c064e8d979605a88c5f2c667f8e6607/app/views/world_location_news/_document_list_from_rummager.html.erb#L24)  is the only example of a dynamically invoked filter path, and it's only used for announcements (and will be removed with world location news rendering anyway).

So think this is good to go 👍

PR #6958, opened 2022-10-25
Retains tracking data on submit button from bootstrap

davidtrussler:
Trello card: [External attachment - tracking needed](https://trello.com/c/X4Wj4I0U/822-external-attachment-tracking-needed)

The intention of this change is to restore the tracking that was lost in the transition from bootstrap to design system. To replicate that the button is wrapped in a div that contains the various `data` attributes that seem to be required.

### DIFF
--- a/app/views/admin/attachments/_form.html.erb
+++ b/app/views/admin/attachments/_form.html.erb
@@ -73,17 +73,26 @@
   <% end %>
 
   <%= hidden_field_tag :type, params[:type] %>
-  
+
   <div class="govuk-button-group">
     <% if attachment.is_a?(ExternalAttachment) %>
       <%= render "govuk_publishing_components/components/button", {
-        text: "Next"
+        text: "Next",
+        data_attributes: {
+          module: "track-button-click",
+          "track-category": "form-button",
+          "track-action": "#{attachment.readable_type}-attachment-button"
+        }
       } %>
     <% else %>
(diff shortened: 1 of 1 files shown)


app/views/admin/attachments/_form.html.erb
```diff
@@ -61,13 +61,15 @@
...
-  <% if attachment.is_a?(ExternalAttachment) %>
-    <%= render "govuk_publishing_components/components/button", {
-      text: "Next"
-    } %>
-  <% else %>
-    <%= render "govuk_publishing_components/components/button", {
-      text: "Save"
-    } %>
-  <% end %>
+  <div data-module="track-button-click" data-track-category="form-button" data-track-action=<%= "#{attachment.readable_type}-attachment-button" %>>
```
davidgisbey, line 64: The button component has an option to pass tracking in if that's better? https://components.publishing.service.gov.uk/component-guide/button/with_data_attributes

```
  text: "Track this!",
  margin_bottom: true,
  data_attributes: {
    module: "track-button-click",
    "data-track-category": "form-button",
    data-track-action: "#{attachment.readable_type}-attachment-button"
  }
} %>
```

davidtrussler: Yes, that would be a much better use of the component. Cheers @davidgisbey. Have updated that. 👍

davidgisbey (approved):
Lgtm!

PR #6937, opened 2022-10-20
Use Collections to render non-English World Location News

brucebolt:
Despite moving the rendering of English language World Location News pages from Whitehall to Collections, we kept non-English pages rendered by Whitehall.

This was to allow FCDO time to approve our plans regarding switching off non-English finders, as the Collections rendered page does not contain a link to the finders, nor does it list recent documents from the finders.

Now it has been confirmed these finders will be switched off, we can render non-English World Location News pages in Collections.

After deploying this change, the following rake task will need to be run to republish the existing World Location News pages:
```
publishing_api:bulk_republish:document_type[WorldLocationNews]
```

[Trello card](https://trello.com/c/5AWg31CN)

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6961, opened 2022-10-25
Remove code for rendering World Location News

brucebolt:
This removes the code for rendering World Location News, which has been migrated to the Collections app.

Not to be merged until https://github.com/alphagov/whitehall/pull/6937 has been deployed to production and all affected documents published.  Additionally, we need to sort out International Delegations at the same time.


app/views/world_locations/show.html.erb
```diff
@@ -1,142 +0,0 @@
-<% page_title @world_location.world_location_news.title, "UK and the world" %>
```
JonathanHallam, line 1: Just want to make sure that we're not going to effect things like https://www.gov.uk/world/uk-mission-to-the-united-nations/ which aren't news pages. Should make sure we test these at release or decide to remove them in favour of the news pages currently rendered by collections (Bruce in conversation on Thurs)

app/helpers/world_location_helper.rb
```diff
@@ -12,25 +12,4 @@ def group_and_sort(locations)
       # finally we remove the sorted name from the structure
       map { |(letter, names_and_locations)| [letter, names_and_locations.map { |_transliterated_name, location| location }] }
   end
-
-  def world_location_survey_url(location)
-    case location.slug
-    when "usa"
-      "https://www.surveymonkey.com/s/873FC35"
```
JonathanHallam, line 19: all of these 404 anyway 😬

app/helpers/public_document_routes_helper.rb
```diff
@@ -144,6 +144,44 @@ def topical_event_about_pages_path(object, options = {})
     append_url_options("/government/topical-events/#{slug}/about", options)
   end
 
+  def world_location_path(object, options = {})
+    slug = case object
```
JonathanHallam, line 148: Is it worth pulling this out into a method so that it can be used across these two methods?

app/views/world_locations/show.html.erb
brucebolt: This PR is dependent on us solving the issues with International Delegation pages.  So we'll only be able to merge this PR once they are migrated to use Collections

app/helpers/world_location_helper.rb
brucebolt: 😬

app/helpers/public_document_routes_helper.rb
brucebolt: Good idea.  I've moved this into a separate method method.

JonathanHallam (approved):
awesome, thanks

PR #6962, opened 2022-10-25
Introduce Cancel button ViewComponent for GOV.UK Design System pages

dnkrj:


dnkrj:
After further discussion it was decided that this was not a good candidate for a ViewComponent and the logic will instead by placed in the [`edition_routes_helper.rb`](https://github.com/alphagov/whitehall/blob/main/app/helpers/admin/edition_routes_helper.rb)

PR #6893, opened 2022-10-06
Whitehall Transition: Release 1.2

ollietreend:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #6968, opened 2022-10-26
Adapt Design System feature flag so users can preview the next release

ollietreend:
## Background

As part of our work to transition Whitehall to the GOV.UK Design System, we have a need for Design System changes to be previewed behind a 'feature flag' prior to releasing them to end users.

We need Design System feature flags for a couple of purposes:

1. **Preview all unreleased Design System changes**
   This is useful for developers and those testing Design System changes that are still being worked on.

2. **Preview the next release**
   Useful when reviewing an upcoming release as it'll appear to users, without other changes that are still being worked on.

We already have an established pattern for previewing unreleased changes (i.e. point 1 above). This Pull Request attempts to rationalise that pattern and extend it to support the second need: specifically previewing only those changes lined up for the next release.

## Overview of changes

- Rename `preview_design_system_user?` to `preview_design_system?`
- Add keyword parameter `next_release` so we can differentiate between things that are lined up for the next release vs. things that are still being worked on.
- Add helper method `render_design_system` to DRY up conditional code that decides which view to render

It may be helpful to review each commit independently, as that'll allow you to see what I've introduced and why, followed by using the methods in existing code.

## Other benefits

This will make it easier to deploy a release for general use before performing clean-up to remove "preview" feature flags.

For example, we could change:

```diff
def preview_design_system?(next_release: false)
-  current_user.can_preview_design_system? || (next_release && current_user.can_preview_next_release?)
+  current_user.can_preview_design_system? || next_release
end
```

To immediately release changes to users. And then in a follow-up PR, we can remove the use of feature flags altogether – e.g. as was done for Release 1.2 in #6893.

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -56,6 +56,14 @@ class Admin::BaseController < ApplicationController
     current_user.can_preview_design_system? || (next_release && current_user.can_preview_next_release?)
   end
 
+  def render_design_system(design_system_view, legacy_view, next_release: false)
+    if preview_design_system?(next_release:)
+      render design_system_view
+    else
+      render legacy_view
+    end
+  end
+
 private
 
   def forbidden!
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -52,8 +52,8 @@ class Admin::BaseController < ApplicationController
     end
   end
 
-  def preview_design_system_user?
-    current_user.can_preview_design_system?
+  def preview_design_system?(next_release: false)
+    current_user.can_preview_design_system? || (next_release && current_user.can_preview_next_release?)
   end
 
 private
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -18,7 +18,7 @@ class Admin::AttachmentsController < Admin::BaseController
   end
 
   def new
-    render(preview_design_system_user? ? "new" : "new_legacy")
+    render_design_system("new", "new_legacy", next_release: true)
   end
 
   def create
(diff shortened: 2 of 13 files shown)


app/controllers/admin/base_controller.rb
```diff
@@ -52,8 +52,16 @@ def forbid_editing_of_locked_documents
     end
   end
 
-  def preview_design_system_user?
-    current_user.can_preview_design_system?
+  def preview_design_system?(next_release: false)
```
dnkrj, line 55: Just a thought but I wonder if this should be renamed `show_design_system` or similar now that it will also be present for a launch?

dnkrj (approved):

baisa (approved):

davidgisbey (commented):
I really like this. I think it might be worth adding either a feature or controller test though.

davidgisbey (approved):

app/controllers/admin/base_controller.rb
ollietreend: Design System changes which have 'been released' for general use, will drop use of the `preview_design_system?` method entirely.

For example, see the changes I made to `edition_tags_controller.rb` when launching it for all users in Release 1.2: https://github.com/alphagov/whitehall/pull/6893/files#diff-a3de518365bb0b4073122ea2e28fc75bf1d2df1bb0f870c561e049bc52b02b4a

The example I gave in this PR's description:

```diff
def preview_design_system?(next_release: false)
-  current_user.can_preview_design_system? || (next_release && current_user.can_preview_next_release?)
+  current_user.can_preview_design_system? || next_release
end
```

Would only ever be a temporary/hacky way of quickly releasing the feature to users. But if used, this would need to be followed up by a PR akin the the one I've just linked as part of a "post-release tidy up" process.

So I think it's still appropriate to call this `preview_design_system` because its primary purpose is for previewing unreleased Design System changes (with that one temporary/hacky exception above).

PR #6970, opened 2022-10-26
Fix newly added cancel links

dnkrj:

The cancel links added in https://github.com/alphagov/whitehall/pull/6941 do not work.
This PR uses the link_to function to generate the anchor tags.

### DIFF
--- a/app/views/admin/attachments/confirm_destroy.html.erb
+++ b/app/views/admin/attachments/confirm_destroy.html.erb
@@ -14,7 +14,7 @@
         destructive: true,
       } %>
 
-      <a class="govuk-link" href="attachable_attachments_path(attachable)">Cancel</a>
+      <%= link_to("Cancel", attachable_attachments_path(attachable), class: "govuk-link") %>
     </div>
     <% end %>
   </section>
--- a/app/views/admin/edition_tags/edit.html.erb
+++ b/app/views/admin/edition_tags/edit.html.erb
@@ -72,7 +72,7 @@
           }
         } %>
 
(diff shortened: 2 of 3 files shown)


ollietreend (approved):

app/views/admin/edition_tags/edit.html.erb
```diff
@@ -72,6 +72,7 @@
           }
         } %>
 
+        <%= link_to("Cancel", admin_edition_path(@edition), class: "govuk-link") %>
         <a class="govuk-link" href="<%= admin_edition_path(@edition) %>">Cancel</a>
```
davidgisbey, line 76: I think this needs to come out then right?

app/views/admin/attachments/confirm_destroy.html.erb
```diff
@@ -2,6 +2,11 @@
 <% content_for :page_title, "Delete attachment" %>
 <% content_for :title, "Delete attachment" %>
 <% content_for :title_margin_bottom, 6 %>
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+    href: attachable_attachments_path(attachable)
+  } %>
+<% end %>
```
davidgisbey, line 9: Do we def want a backlink and cancel link on the same page?

dnkrj: Oh, good catch, sorry, have removed it again

app/views/admin/edition_tags/edit.html.erb
dnkrj: Thanks!

davidgisbey (approved):

PR #6982, opened 2022-10-31
Release 1.3 attachment page fixes (part1)

davidgisbey:
## Description 

This PR actions some of the changes requested for the attachments pages outlined in this document  https://docs.google.com/document/d/1QBQl5s-576_HsfUfKCWDAvwETvqukrNZLY6a-KLJtGw/edit#

## Scope 

It only focusses on the attachments page changes. Withdrawal and govspeak sidebar changes are not in scope.

It also does not include reverting the command paper number and house of commons number 

image

Or fixing errors so they link to the div not the input itself. They will be actioned in follow up PRs

## Checkboxes not working

There were a few bugs around checkboxes not working correctly. This fixes the following bugs

1. Ensures the `manually_numbered_headings` and `attachment[accessible]` checkboxes are set to false is left unchecked by adding hidden fields
2. Checks the accessible checkboxes is the attachment is accessible

### External attachment new/edit page

Updates `Next` to `Save` 

#### Before

image

#### After

image

### New bulk attachment page

- Removes backlink
- Adds cancel link
- Updates `Save` to `Upload zip`

#### Before

image

#### After

image

### Edit title for bulk attachments

Adds a cancel button

#### Before

image

#### After 

image

### Before

image

### After 

image

## Add feature specs for bulk upload in design system

There weren't any and there probs should be.

## Trello card 

https://trello.com/c/KjdWREUw/843-fixes-for-attachments-prior-to-13-release

### DIFF
--- a/app/views/admin/bulk_uploads/new.html.erb
+++ b/app/views/admin/bulk_uploads/new.html.erb
@@ -21,9 +21,13 @@
         error_items: errors_for(@zip_file.errors, :zip_file)
       } %>
 
-      <%= render "govuk_publishing_components/components/button", {
-        text: "Next",
-      } %>
+      <div class="govuk-button-group">
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Upload zip",
+        } %>
+
+        <%= link_to("Cancel", admin_edition_attachments_path(@edition), class: "govuk-link") %>
+      </div>
     <% end %>
   </div>
 </div>
--- a/app/views/admin/bulk_uploads/new.html.erb
+++ b/app/views/admin/bulk_uploads/new.html.erb
@@ -1,11 +1,6 @@
 <% content_for :page_title, "Bulk upload for: #{@edition.title}" %>
 <% content_for :title, "Bulk upload" %>
 <% content_for :context, "Attachments for #{@edition.format_name}" %>
-<% content_for :back_link do %>
-  <%= render "govuk_publishing_components/components/back_link", {
-    href: admin_edition_attachments_path(@edition)
-  } %>
-<% end %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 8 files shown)


app/views/admin/bulk_uploads/set_titles.html.erb
```diff
@@ -37,12 +38,15 @@
               <%= render 'admin/attachments/reference_fields', attachable: @edition, form: attachment_fields, attachment: attachment, heading_size: "m" %>
             <% end %>
 
+            <% form.hidden_field :accessible, value: "0" %>
```
davidtrussler, line 41: Should this be `<%=` for it to show in the markup? 

Also I think this needs to be named uniquely for each attachment, whereas here we have multiple hidden fields of the same name on the page. In fact should it have the same name as the checkbox with value 1 so that the submitted value toggles between the two?

app/views/admin/bulk_uploads/new.html.erb
```diff
@@ -1,11 +1,6 @@
 <% content_for :page_title, "Bulk upload for: #{@edition.title}" %>
 <% content_for :title, "Bulk upload" %>
 <% content_for :context, "Attachments for #{@edition.format_name}" %>
-<% content_for :back_link do %>
```
dnkrj, line 4: Could you also remove the back link from the set_titles template?

app/views/admin/bulk_uploads/set_titles.html.erb
davidgisbey: Good shout. I've just realised this isn't actually needed though as it's only used to create the objects not update.

app/views/admin/bulk_uploads/new.html.erb
davidgisbey: Good shout. Have done

dnkrj (approved):

davidtrussler (approved):
LGTM 👍

PR #6989, opened 2022-11-01
Updates tracking module to `gem-track-click` for design system

davidtrussler:
This PR updates the code for tracking, which was not giving the expected results as it was originally implemented. The reason is that the Design System uses tracking from the component which requires the `GemTrackClick` script and replaces the previous tracking functionality.


davidgisbey (approved):
lgtm

PR #6992, opened 2022-11-01
[843] Ensure errors link to a fields container not the input for attachments and withdrawal pages

davidgisbey:
## Description

At the moment, errors in the error summary component link directly to the inout. This means the label and hint text are not visible on the screen and a user loses context.

Instead of linking directly to the input, we're going to wrap the input in a div with the current input id. We will then append the id of the input with '_field'.

This adds a helper which takes an id and a block. It creates a div with the id passed in and renders the block within it.

This is scoped to the 1.3. release. I'll pick this up with stuff on the edit edition page afterwards.

### Before 

image

### After

image

## Trello card 

### DIFF
--- a/app/helpers/admin/field_container_helper.rb
+++ b/app/helpers/admin/field_container_helper.rb
@@ -1,0 +1,6 @@
+module Admin::FieldContainerHelper
+  def field_container(id:, &block)
+    content = capture(&block)
+    content_tag(:div, content, id:)
+  end
+end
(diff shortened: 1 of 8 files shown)


dnkrj (changes requested):
I'm not sure that this is the correct approach. [The design system is quite specific about specifically linking to the field.](https://design-system.service.gov.uk/components/error-summary/). [Functionality was added to scroll to the label or legend instead of the field](https://github.com/alphagov/govuk-frontend/pull/1056) but I'm not sure why this isn't working for us. Happy to investigate more tomorrow.

davidgisbey:
This has been resolved in a separate pr

PR #7001, opened 2022-11-02
Run feature tests against the Design System

ollietreend:
This PR adapts the `rake cucumber` task to run the feature tests twice:

1. With the 'Preview next release' flag enabled
2. Without the flag enabled (i.e. a normal run, as before)

This will ensure we're keeping our feature tests up to date as we work on new Design System styled pages.

For now it only applies the 'Preview next release' flag. We're currently focussing on preparing the next release (Release 1.3) of the Whitehall Design System transition, so this is the quickest/simplest thing to do. Later we should open this up to use the broader 'Preview design system' flag, which will include all changes even if they're not lined up for the next release.


davidgisbey (approved):
excellent work. That all makes sense to me! Def worth the extra build time imo

PR #7003, opened 2022-11-02
Update govspeak guidance to be behind next release flag

davidgisbey:
## Description

This is in the next release. Currently, it's behind the preview design system flag 

### DIFF
--- a/app/helpers/admin/sidebar_helper.rb
+++ b/app/helpers/admin/sidebar_helper.rb
@@ -1,22 +1,22 @@
 module Admin::SidebarHelper
   def simple_formatting_sidebar(options = {})
-    if current_user.has_permission? "Preview design system"
-      sidebar_content = []
-      sidebar_content << render("admin/editions/govspeak_help", options)
-      sidebar_content << render("admin/editions/style_guidance", options)
-      raw sidebar_content.join("\n")
-    else
-      sidebar_tabs govspeak_help: "Help" do |tabs|
-        tabs.pane id: "govspeak_help", class: "govspeak_help" do
-          tab_content = []
-          tab_content << render("admin/editions/legacy_govspeak_help", options)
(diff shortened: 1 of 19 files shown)


app/views/admin/attachments/edit.html.erb
```diff
@@ -9,7 +9,7 @@
 
   <% if attachment.html? %>
     <aside class="govuk-grid-column-one-third">
-      <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true %>
+      <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true, preview_design_system: preview_design_system?(next_release: true) %>
```
ollietreend, line 12: Is `preview_design_system?(next_release: true)` actually needed here?

This entire file should be scoped to Design System stuff since it's `edit.html.erb` rather than `edit_legacy.html.erb`, so I think it should be okay to simply hardcode it?

```suggestion
      <%= simple_formatting_sidebar hide_inline_attachments_help: true, show_footnote_help: true, show_stat_headline_help: true, preview_design_system: true %>
```

app/controllers/admin/base_controller.rb
```diff
@@ -55,6 +55,7 @@ def forbid_editing_of_locked_documents
   def preview_design_system?(next_release: false)
     current_user.can_preview_design_system? || (next_release && current_user.can_preview_next_release?)
   end
+  helper_method :preview_design_system?
```
ollietreend, line 58: ...in which case, this shouldn't be needed.

app/views/admin/attachments/edit.html.erb
davidgisbey: yea good point!

ollietreend (approved):
Nice 👍🏻

PR #7006, opened 2022-11-02
Remove redundant feature tests

ollietreend:
Following on from PR #7001, we can now remove these feature tests since they're now redundant. This functionality will be covered by the existing feature tests that these duplicated, now that the Cucumber tests run against the Design System too.

### DIFF
--- a/features/edition-attachments.feature
+++ b/features/edition-attachments.feature
@@ -13,33 +13,12 @@ Feature: Managing attachments on editions
     And I add an external attachment with the title "Beard Length Illustrations" and the URL "http://www.beardlengths.gov.uk"
     Then the publication "Standard Beard Lengths" should have 3 attachments
     When I set the order of attachments to:
-      |         title                | order |
-      | Beard Length Graphs 2012     |   0   |
-      | Beard Length Statistics 2014 |   1   |
-      | Beard Length Illustrations   |   2   |
+      | title                        | order |
+      | Beard Length Graphs 2012     | 0     |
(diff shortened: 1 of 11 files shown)


features/edition-attachments.feature
```diff
@@ -13,33 +13,12 @@ Feature: Managing attachments on editions
     And I add an external attachment with the title "Beard Length Illustrations" and the URL "http://www.beardlengths.gov.uk"
     Then the publication "Standard Beard Lengths" should have 3 attachments
     When I set the order of attachments to:
```
davidgisbey, line 15: How is this passing for the Design System once when it's on a different page?

ollietreend: I think it's passing right now because the Attachments Index page is still Bootstrap under the 'Preview next release' flag. So this feature spec is actually testing the Bootstrap functionality in both cases.

This will be something we'll need to tackle once the Attachments Index page gets tagged as `next_release: true`. I'm thinking we may need to adopt tags for features that _only_ run in Bootstrap or Design System modes rather than both – with this being the first example, due to it being a change of workflow.

davidgisbey: oh right ofc 🤦 yea all good then

davidgisbey (approved):

features/edition-attachments.feature
ollietreend: > oh right ofc 🤦

It took me some time to figure this out, too. So no shame.

PR #7013, opened 2022-11-03
Add tracking to specialist topic tagging page

dnkrj:

## What
Adds tracking functionality to the autocomplete using the `onConfirm` callback.
[Uses this functionality to add missing tracking to the specialist topic tagging page.](https://trello.com/c/plXuFwJo/850-add-tracking-to-specialist-topic-tagging-page)

## Why
The current select2 component uses the [`TrackSelectClick` module](https://github.com/alphagov/whitehall/blob/main/app/assets/javascripts/admin_legacy/modules/track_select_click.js) to for GA tracking, this functionality does not work for the accessible autocomplete.
The accessible autocomplete [suggests instead using the `onConfirm` callback](https://github.com/kevindew/accessible-autocomplete#analytics-and-tracking) to perform tracking.

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -23,7 +23,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     new window.accessibleAutocomplete.enhanceSelectElement({ // eslint-disable-line no-new, new-cap
       selectElement: $select,
       minLength: 3,
-      showNoOptionsFound: true
+      showNoOptionsFound: true,
+      onConfirm: function (value) {
+        var category = $select.getAttribute('data-track-category')
+        var label = $select.getAttribute('data-track-label')
+        var action = value
+        if (category && label) {
+          window.GOVUK.analytics.trackEvent(category, action, { label: label })
+        }
+      }
     })
   }
 
--- a/app/views/admin/base/_specialist_sector_fields.html.erb
+++ b/app/views/admin/base/_specialist_sector_fields.html.erb
@@ -13,6 +13,8 @@
     options: [""] + specialist_sector_options_for_select,
     selected: @edition.primary_specialist_sector_tag,
   },
+  track_category: "taxonSelectionPrimarySpecialist",
+  track_label: request.path
 } %>
 
 <%= render "components/autocomplete", {
@@ -27,4 +29,6 @@
     multiple: true,
     selected: @edition.secondary_specialist_sector_tags,
   },
+  track_category: "taxonSelectionAdditionalSpecialist",
+  track_label: request.path
 } %>
--- a/app/views/admin/edition_legacy_associations/edit.html.erb
+++ b/app/views/admin/edition_legacy_associations/edit.html.erb
@@ -13,6 +13,11 @@
       <div class="govuk-button-group govuk-!-margin-bottom-6">
         <%= render "govuk_publishing_components/components/button", {
           text: "Update specialist topics",
+          data_attributes: {
+            module: "gem-track-click",
+            "track-category": "form-button",
+            "track-action": "publication-button"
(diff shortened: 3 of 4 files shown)


app/assets/javascripts/components/autocomplete.js
```diff
@@ -23,7 +23,17 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+      showNoOptionsFound: true,
+      onConfirm: function (value) {
+        var category = $select.getAttribute('data-track-category')
+        var label = $select.getAttribute('data-track-label')
+        if (!category || !label) {
+          return
+        }
+
+        var action = value
+        window.GOVUK.analytics.trackEvent(category, action, { label })
```
DilwoarH, line 35: Could we potentially add another attribute called `data-track-select="true"`

```suggestion
        var trackSelect = $select.getAttribute('data-track-select') === "true"
        
        if (trackSelect) {
          var category = $select.getAttribute('data-track-category')
          var label = $select.getAttribute('data-track-label')

          window.GOVUK.analytics.trackEvent(category, value, { label: label })
        }
```

DilwoarH (changes requested):

app/assets/javascripts/components/autocomplete.js
```diff
@@ -23,7 +23,17 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-      showNoOptionsFound: true
+      showNoOptionsFound: true,
+      onConfirm: function (value) {
+        var category = $select.getAttribute('data-track-category')
+        var label = $select.getAttribute('data-track-label')
+        if (!category || !label) {
+          return
+        }
+
+        var action = value
```
ollietreend, line 34: What is the expected value of `action`?

I was hoping to see a unit test which might have given a clue about what values it'd track in different scenarios. But it looks like there isn't a test for this file at all 🙁 and obviously creating one would out-of-scope for this PR.

Looking at the [Accessible Autocomplete docs](https://alphagov.github.io/accessible-autocomplete/#onconfirm-default---), it seems that the `onConfirm` callback will receive an `Object` parameter. But I was expecting `action` to be a `String` for passing in to `GOVUK.analytics.trackEvent`. So based on that, I suspect this tracking might not work as expected 🤔

ollietreend (commented):
This approach looks good – just one potential issue in the implementation.

app/assets/javascripts/components/autocomplete.js
dnkrj: Yeah, lack of test coverage is really an issue for all of these components.

Action in this case action is a String.
The autocomplete returns [whatever the value of that option was when it was passed into it](https://github.com/kevindew/accessible-autocomplete/blob/master/src/autocomplete.js#L302) when it was initialised.
In this case we're using the enhanceSelectElement function which only [passes the option values to the autocomplete](https://github.com/kevindew/accessible-autocomplete/blob/master/src/wrapper.js#L23) which are the Strings.

app/assets/javascripts/components/autocomplete.js
dnkrj: What do you think about this vs. an explicit parameters on the autocomplete component for the tracking category and label?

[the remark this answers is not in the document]
dnkrj: It's the [shorthand for property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#property_definitions), which I suppose we shouldn't use if we're still targeting IE at all.

Have updated it.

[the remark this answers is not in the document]
DilwoarH: nice - didn't know they brought it into JS

DilwoarH (approved):

PR #7016, opened 2022-11-04
Remove Content Publisher migration functionality

kevindew:

This removes code that was added in 2019 in preparation for what was then an imminent migration of content from Whitehall to Content Publisher. A variety of things happened following the introduction of this code and ultimately it was never used in anger and remains unlikely to be used again. 

Removing this allows simplifying the codebase, should we change plans we can always revert this.

There is going to be a follow up PR for this that removes a database field that was added for this change.


ollietreend (approved):
@kevindew this PR is glorious 😍

It's always good to see more red than green in a PR 🙌🏻 

Thanks for opening this.

PR #7017, opened 2022-11-04
Remove locked column from documents

kevindew:
This is a draft which builds from: https://github.com/alphagov/whitehall/pull/7016. Once that is merged and deployed safely (i.e. we're confident we don't have to rollback) we can merge this in to complete the cleanup.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -46,8 +46,6 @@ class Document < ApplicationRecord
 
   after_create :ensure_document_has_a_slug
 
-  self.ignored_columns = %w[locked]
-
   attr_accessor :sluggable_string
 
   def self.live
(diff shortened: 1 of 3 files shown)


ollietreend (approved):
This looks good to me 👍🏻 

Just to add – this database migration isn't as scary or irreversible as you might think:

```
mysql> SELECT COUNT(*) FROM documents WHERE locked = true;
+----------+
| COUNT(*) |
+----------+
|        0 |
+----------+
```

kevindew:
> This looks good to me 👍🏻
> 
> Just to add – this database migration isn't as scary or irreversible as you might think:
> 
> ```
> mysql> SELECT COUNT(*) FROM documents WHERE locked = true;
> +----------+
> | COUNT(*) |
> +----------+
> |        0 |
> +----------+
> ```

Super thanks, yup I was expecting it to be 0 (would have been a shock if it was used) but didn't want to couple it with the code changes so rolling back would be simple should an unforeseen problem occur. Rolling back code and DB at the same time is a bit 🤮

PR #7021, opened 2022-11-04
Add search to 'Topic taxonomy tags' (Autocomplete + Miller Columns component)

DilwoarH:
## What

Add search/filtering functionality to the 'Topic taxonomy tags' page. This will involve integrating the Autocomplete and Miller Columns components so they work together.

This functionality already exists in Content Publisher, so we should be able to port it across relatively easily*.

*Kevin has created a proof of concept branch here, but its lacking tests: [https://github.com/alphagov/whitehall/compare/search-on-taxonomy](https://github.com/alphagov/whitehall/compare/search-on-taxonomy "‌")

## Why

User feedback tells us this is needed – [https://gds.slack.com/archives/C03R56148PR/p1666710324163929](https://gds.slack.com/archives/C03R56148PR/p1666710324163929 "‌")

[https://govuk.zendesk.com/agent/#/tickets/5102286](https://govuk.zendesk.com/agent/#/tickets/5102286 "‌")

https://trello.com/c/evNn3X8N/856-add-search-to-topic-taxonomy-tags-autocomplete-miller-columns-component

## Before

Screenshot 2022-11-04 at 2 55 56 pm

## After

Screenshot 2022-11-04 at 2 55 33 pm

### DIFF
--- a/app/assets/javascripts/components/miller-columns.js
+++ b/app/assets/javascripts/components/miller-columns.js
@@ -1,1 +1,88 @@
 //= require miller-columns-element
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function MillerColumns (module) {
+    this.module = module
+    this.searchable = module.getAttribute('data-searchable') === 'true'
+  }
+
+  MillerColumns.prototype.init = function () {
+    if (this.searchable) this.initSearch()
+  }
+
+  MillerColumns.prototype.initSearch = function () {
+    var element = this.module.querySelector('#js-app-c-miller-columns__search')
+    var input = this.module.querySelector('#js-app-c-miller-columns__search-input')
+    var millerColumns = this.module.querySelector('miller-columns')
+
+    if (!window.accessibleAutocomplete) {
+      element.parentNode.removeChild(element)
+      return
+    }
+
+    var topics = millerColumns.taxonomy.flattenedTopics
+    var topicSuggestions = []
+
+    topics.forEach(function (topic) {
+      topicSuggestions.push({
+        topic: topic,
+        highlightedTopicName: topic.topicName.replace(/<\/?mark>/gm, ''), // strip existing <mark> tags
+        breadcrumbs: topic.topicNames
+      })
+    })
+
+    if (!topicSuggestions.length) {
+      element.parentNode.removeChild(element)
+      return
+    }
+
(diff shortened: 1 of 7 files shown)


kevindew:
Nice one working on this, I just wanted to add a note to make sure you also get in the no-js fixes I added in my branch: https://github.com/alphagov/whitehall/commit/b8a42ce4f0388acb40f9306e5a4fe794ec347a13

DilwoarH:
> Nice one working on this, I just wanted to add a note to make sure you also get in the no-js fixes I added in my branch: [b8a42ce](https://github.com/alphagov/whitehall/commit/b8a42ce4f0388acb40f9306e5a4fe794ec347a13)

Thanks Kevin - done

app/assets/javascripts/components/miller-columns.js
```diff
@@ -1 +1,88 @@
...
+
+  MillerColumns.prototype.initSearch = function () {
+    var element = this.module.querySelector('#js-app-c-miller-columns__search')
+    var input = this.module.querySelector('#js-app-c-miller-columns__search-input')
+    var millerColumns = this.module.querySelector('miller-columns')
+
+    if (!window.accessibleAutocomplete) {
+      element.parentNode.removeChild(element)
+      return
+    }
```
ollietreend, line 23: Is this here to avoid JS errors if we've forgotten to include Autocomplete's JS?

If we think this is likely to be an issue, it might be nice to include a helpful error message. Otherwise it might end up being more confusing as it'll silently swallow the problem and look like it's doing 'nothing'.

```suggestion
    if (!window.accessibleAutocomplete) {
      element.parentNode.removeChild(element)
      console.error('Accessible Autocomplete is required if you want this Miller Columns instance to be searchable')
      return
    }
```

I'm also not sure what the benefit is of deleting the `#js-app-c-miller-columns__search` element here? It doesn't look like that `element` gets created by JS, so why do we need to remove it?

ollietreend (approved):
Nice work @DilwoarH  🎉 

I've included a suggestion which is completely just my opinion, so feel free to ignore it if you don't think it's helpful.

Still, I'm happy to approve this as-is ✅

app/assets/javascripts/components/miller-columns.js
DilwoarH: I don't think we need the error message here as we have autocomplete component as part of the Whitehall application.

The reason why we need to remove the element is that there is an input field which needs to be removed. Also, this won't cause a JS error as it won't run if not available.

PR #7024, opened 2022-11-07
Fix specialist topic tagging in design system pages

dnkrj:

## What
Copy the code to select the relevant option from the [accessible autocomplete `onConfirm` function](https://github.com/kevindew/accessible-autocomplete/blob/master/src/wrapper.js#L42) to our custom `onConfirm` function introduced for tracking.

## Why
These lines are required for the select to function property and were previously overwritten by the custom `onConfirm` function to tracking. This is a [known issue with the accessible autocomplete](https://github.com/alphagov/accessible-autocomplete/issues/322).
We cannot use the [`track-select-change`](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics/track-select-change.js) module as the accessible autocomplete does not trigger `changed` events when marking options as selected.
The loss in functionality that resulted from the tracking implementation was not detected in tests as the test [interacts directly with the select](https://github.com/alphagov/whitehall/blob/main/features/step_definitions/specialist_sector_steps.rb#L10).

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -24,13 +24,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       selectElement: $select,
       minLength: 3,
       showNoOptionsFound: true,
-      onConfirm: function (value) {
+      onConfirm: function (query) {
         var category = $select.getAttribute('data-track-category')
         var label = $select.getAttribute('data-track-label')
-        var action = value
+        var action = query
         if (category && label) {
           window.GOVUK.analytics.trackEvent(category, action, { label: label })
         }
+
+        const options = $select.options
+        let matchingOption
+        if (query) {
+          matchingOption = [].filter.call(options, option => (option.textContent || option.innerText) === query)[0]
+        } else {
+          matchingOption = [].filter.call(options, option => option.value === '')[0]
+        }
+        if (matchingOption) { matchingOption.selected = true }
       }
     })
   }


app/assets/javascripts/components/autocomplete.js
```diff
@@ -19,18 +19,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   Autocomplete.prototype.initAutoCompleteSelect = function ($select) {
+    var selectElement = $select
```
DilwoarH, line 22: Is there a reason why this is needed? Could we not just change the argument name?

dnkrj: Updated.
The was no real reason, I just wanted the code to be as close to a copy as possible but it's probably not that important with more changes to the autocomplete likely to come.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -24,13 +24,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
         }
+
+        const options = $select.options
+        let matchingOption
+        if (query) {
+          matchingOption = [].filter.call(options, option => (option.textContent || option.innerText) === query)[0]
+        } else {
+          matchingOption = [].filter.call(options, option => option.value === '')[0]
+        }
+        if (matchingOption) { matchingOption.selected = true }
```
DilwoarH, line 42: Is this ported from a specific piece of code? If so, could you reference it in your commit.

DilwoarH (approved):

PR #7025, opened 2022-11-07
Run feature tests with the 'Preview design system' flag

ollietreend:
This PR expands upon the changes made in #7001 which configured the Cucumber feature tests to run with the flag 'Preview next release' so that upcoming Design System changes would be covered by tests.

In this PR, I've configured Cucumber to run with the flag 'Preview design system', which supersedes the 'Preview next release' flag. This will run feature tests against all unreleased Design System changes, including those which haven't yet been tagged for inclusion in the next release.

I've tried to fix some of the more obvious test failures – for example, where button/label text differs – but came across a number of tests which need attention from someone more familiar with the current state of play (most likely @davidgisbey). So for those tests, I've tagged them as `@design-system-wip`.

### New tags: `@design-system-wip`, `@bootstrap-only`, `@design-system-only`

These tags will control which context feature specs will run in. By default, all feature specs must pass both with and without the 'Preview design system' flag enabled.

If a feature is still being worked on, it's possible to tag a test with `@design-system-wip` to skip it from the Design System run. Tests with this tag should be fixed prior to those features being released to the general userbase.

`@bootstrap-only` and `@design-system-only` can be used to denote functionality that only exists in one or the other layout. For example, the 'highlight words to avoid' feature is not being ported across to the Design System, so it is tagged as a `@bootstrap-only` feature.

### New helper method: `within_conditional_reveal`

Use this to scope Capybara finders to a 'conditional reveal' container belonging to a radio or checkbox.

For more details about using this, see the commit message f380f83ceb3e15ad57603b12e331fd9f600f3f69.

---

### DIFF
--- a/config/cucumber.yml
+++ b/config/cucumber.yml
@@ -3,8 +3,11 @@ rerun = File.file?('rerun.txt') ? IO.read('rerun.txt') : ""
 rerun = rerun.strip.gsub /\s/, ' '
 rerun_opts = rerun.empty? ? "--format #{ENV['CUCUMBER_FORMAT'] || 'progress'} features" : "--format #{ENV['CUCUMBER_FORMAT'] || 'pretty'} #{rerun}"
 std_opts = "--format #{ENV['CUCUMBER_FORMAT'] || 'progress'} --strict --tags 'not @wip'"
+
+default_tags = "--tags 'not @design-system-only'"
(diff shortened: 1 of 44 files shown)


davidgisbey (approved):
lgtm i'll look through the design system wip tests. The non-js ones should be able to be fixed.

PR #7027, opened 2022-11-07
Remove imported edition state

baisa:
_Imported_ edition state and associated functionality (speed_tagging, convert_to_draft) is redundant and can be removed.

[Trello card](https://trello.com/c/XzGw1GmN/829-remove-imported-edition-state-from-whitehall)


app/views/admin/editions/_edit_attachments.html.erb
```diff
@@ -16,7 +16,7 @@
           <%= f.text_field :title %>
         <% end %>
       </td>
-      <td class="attachment-filename"><%= link_to_attachment(attachment, preview: !@edition.published?, truncate: @edition.imported?) %></td>
+      <td class="attachment-filename"><%= link_to_attachment(attachment, preview: !@edition.published?, truncate: false) %></td>
```
ollietreend, line 19: It looks like this was the only place that the `truncate:` option was being set to `true`.

So it'd be great to remove it from the link_to_attachment helper altogether ✨🧹

https://github.com/alphagov/whitehall/blob/78e4f19186d2e7ba1ec0cdea7a657a996280646f/app/helpers/application_helper.rb#L47-L54

app/views/admin/editions/_legacy_first_published_at.html.erb
```diff
@@ -3,28 +3,22 @@
   <p class="required">This document <span>*</span></p>
   <div class="radio">
     <%= form.label :previously_published_false, class: 'radio' do %>
-      <%= form.radio_button(:previously_published, false, disabled: edition.imported?) %>
+      <%= form.radio_button(:previously_published, false, disabled: false) %>
```
ollietreend, line 6: I think we should be able to remove the `disabled:` option entirely now.

```suggestion
      <%= form.radio_button(:previously_published, false) %>
```

app/views/admin/editions/_legacy_first_published_at.html.erb
```diff
@@ -3,28 +3,22 @@
...
       has never been published before. It is new.
     <% end %>
   </div>
   <div class="radio">
     <%= form.label :previously_published_true, class: 'radio' do %>
-      <%= form.radio_button(:previously_published, true, disabled: edition.imported?) %>
+      <%= form.radio_button(:previously_published, true, disabled: false) %>
```
ollietreend, line 12:
```suggestion
      <%= form.radio_button(:previously_published, true) %>
```

app/views/admin/editions/_show_attachments.html.erb
```diff
@@ -11,7 +11,7 @@
     <td>
       <%= attachment.title %>
     </td>
-    <td class="attachment-filename"><%= link_to_attachment(attachment, preview: !@edition.published?, truncate: @edition.imported?, full_url: true) %></td>
+    <td class="attachment-filename"><%= link_to_attachment(attachment, preview: !@edition.published?, truncate: false, full_url: true) %></td>
```
ollietreend, line 14: Ditto here – I think `truncate:` can go entirely.

```suggestion
    <td class="attachment-filename"><%= link_to_attachment(attachment, preview: !@edition.published?, full_url: true) %></td>
```

app/views/admin/editions/_show_metadata.html.erb
```diff
@@ -128,11 +128,7 @@
       <% end %>
     </h2>
     <% if @edition.attachments.any? %>
-      <% if @edition.imported? %>
-        <%= render partial: "edit_attachments", locals: { edition: @edition } %>
```
ollietreend, line 132: I think this is the only place where `partial: "edit_attachments"` is rendered – so we should be able to delete the file [app/views/admin/editions/_edit_attachments.html.erb](https://github.com/alphagov/whitehall/blob/78e4f19186d2e7ba1ec0cdea7a657a996280646f/app/views/admin/editions/_edit_attachments.html.erb) 🗑️ 

Which means we should also be able to delete [app/assets/javascripts/admin_legacy/rename_attachments.js](https://github.com/alphagov/whitehall/blob/78e4f19186d2e7ba1ec0cdea7a657a996280646f/app/assets/javascripts/admin_legacy/rename_attachments.js) because it was only used by that partial. 💥

app/models/edition.rb
```diff
@@ -594,30 +594,17 @@ def valid_as_draft?
...
 
   def can_have_some_invalid_data?
-    imported? || deleted? || superseded?
+    deleted? || superseded?
   end
 
-  attr_accessor :trying_to_convert_to_draft, :has_previously_published_error
+  attr_accessor :has_previously_published_error
 
   def errors_as_draft
```
ollietreend, line 606: I think we should be able to delete the `valid_as_draft?` and `errors_as_draft` methods entirely. They were only being used by imported/speed tagging things.

...which should mean we can also get rid of this smelly patch file: [lib/patches/make_initialize_dup_public.rb](https://github.com/alphagov/whitehall/blob/a7d4f4c4e222ac482a1bce7f0a32f6e68a3c6212/lib/patches/make_initialize_dup_public.rb)

app/views/admin/editions/_speed_tagging.html.erb
```diff
@@ -1,66 +0,0 @@
...
-
-      <ul>
-        <% @edition.errors_as_draft.full_messages.each do |message| %>
-          <li><%= message %></li>
-        <% end %>
-      </ul>
-    </div>
-  <% end %>
-
-  <% unless @edition.document.document_sources.empty? %>
```
ollietreend, line 14: I have a _sneaky suspicion_ that we'll also be able to get rid of Document Sources. But I think this might need some more digging, so perhaps one to tackle later. (I'd like to check with Kevin first!)

config/brakeman.ignore
```diff
@@ -80,50 +80,6 @@
       ],
       "note": ""
     },
-    {
-      "warning_type": "Cross-Site Scripting",
```
ollietreend, line 84: 🎉 🎉 😍

test/unit/consultation_test.rb
```diff
@@ -8,7 +8,7 @@ class ConsultationTest < ActiveSupport::TestCase
   should_not_allow_inline_attachments
   should_protect_against_xss_and_content_attacks_on :consultation, :title, :body, :summary, :change_note
 
-  %i[imported deleted].each do |state|
+  %i[deleted].each do |state|
```
ollietreend, line 11: At first glance, I feel like the 'each' is no longer useful here because the Array only contains one item 🤔 

But I can also see that it's consistent with the approach taken for other states, e.g.
https://github.com/alphagov/whitehall/blob/8392b2f904cd0a5cab875dfc5826377c4b4bb0b8/test/unit/consultation_test.rb#L45

So am inclined to say it's fine to leave it as is, for consistency.

test/unit/edition/has_document_collection_test.rb
```diff
@@ -32,13 +32,6 @@ class Edition::HasDocumentCollectionsTest < ActiveSupport::TestCase
     assert_not edition.reload.part_of_published_collection?
   end
 
-  test "allows assignment of document collection on a saved edition" do
-    edition = create(:imported_publication)
-    document_collection = create(:document_collection, :with_group)
-    edition.document_collection_group_ids = [document_collection.groups.first.id]
-    assert_equal [document_collection], edition.document.document_collections
-  end
-
```
ollietreend, line 41: Would this test still be relevant/useful if we used an edition in a different state? 🤔 

I'm wondering if we're losing coverage of something useful by removing this test entirely.

What about if we just change `:imported_publication` to `:draft_edition` or some other state?

test/unit/services/edition_publisher_test.rb
```diff
@@ -111,15 +111,6 @@ class EditionPublisherTest < ActiveSupport::TestCase
     assert publisher.perform!
   end
 
-  test "#perform! does not notify if publishing is unsuccessful" do
-    edition  = build(:imported_edition)
```
ollietreend, line 115: Same here. Would this test still be useful to keep?

Could we change `:imported_edition` to e.g. `:draft_edition`?

ollietreend (commented):
@baisa This is looking excellent! 😍 😍 

I can see how thorough you've been, too! ✨🧹

And this is the most exciting part:
line diff

Although I think we might be able to remove even more lines 💪🏻 see the comments I've added.

app/models/edition.rb
baisa: Done! Thanks @ollietreend 👍

app/views/admin/editions/_speed_tagging.html.erb
baisa: Definitely worth investigating!

test/unit/edition/has_document_collection_test.rb
baisa: Makes sense @ollietreend - done! 👍

ollietreend (approved):
Excellent work @baisa!
🙌🏻 💫 🧹 ✨ 💥

This is one more piece of tech debt off the list. Thanks for doing it! ⭐

PR #7029, opened 2022-11-07
Track when users upload an image

ollietreend:
This adds tracking to the legacy Bootstrap admin interface so that we can work out when a user is uploading an image.

This has been requested by our Performance Analysts to help with understanding user behaviour and success metrics.

Specifically, this should allow us to work out how many image uploads result in errors, vs. those which are successful. Without this tracking, it's not possible to tell which form submissions included an image upload vs. those which didn't add/change images.


dnkrj (approved):
Code looks good, Although I'm not familiar with tag naming conventions

PR #7038, opened 2022-11-08
Introduce Cancel button with routes helper

dnkrj:

## What
Introduce a new routes helper to provide a cancel path to templates.
Use that helper on the edition edit and translation pages.
Remove back buttons that have now been replaced.

## Why
In our like for like approach we're using the same buttons as before for the new design system pages.
We are however no longer using the form builder and consequently rehoming the routes helper logic.

## Screenshots

whitehall-admin dev gov uk_government_admin_publications_1374130_edit(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1374130_translations_ar_edit(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1374130_attachments(iPad Pro) (1)


ollietreend (approved):
@dnkrj This looks good 👍🏻 but I think it'd be nice to add some tests.

Very happy to talk it through if you want to.

test/unit/helpers/admin/cancel_routes_helper_test.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class Admin::CancelRoutesHelperTest < ActionView::TestCase
+  test "admin_cancel_path take a new edition instance and generates the correct path" do
+    e = build(:edition)
+    assert_equal "/government/admin/editions", admin_cancel_path(e)
+  end
+
+  test "admin_cancel_path take a new corporate_information_page instance and generates the correct path" do
+    c = build(:corporate_information_page)
+    c.organisation = create(:organisation)
```
davidgisbey, line 11: Looking great. This is a bit nitty, but i generally go with `build_stubbed` rather than `create` wherever possible in unit tests as it's quicker

There's a little article on it here https://kolosek.com/factory-girl-build-vs-build_stub/

`create` - instantiates the object and associations
`build` - builds the object and instantiates associations 
`build_stubbed` - creates and assigns attributes to an object to make it behave like an instantiated object, but does not hit the database

davidgisbey (approved):
lgtm once rubocop is happy

PR #7047, opened 2022-11-10
Ensure display language has selected value set when present

davidgisbey:
## Description

This was missing a selected value which means that when a user goes to edit the page, it will default back to all languages.

## Before 

image

## After 

image


davidtrussler (approved):
LGTM! 👍

PR #7049, opened 2022-11-10
Move confirm force publish page to the GOV.UK Design System

dnkrj:

## What

[Move confirm force publish page to the GOV.UK Design System.
](https://trello.com/c/AYjLQh3C/890-force-publish-journey)

## Why

This is part of a Whitehall transition work to move from bootstrap to GOVUK design system.

### After
whitehall-admin dev gov uk_government_admin_editions_1374130_confirm_force_publish_lock_version=12(iPad Pro)

### Before
whitehall-admin dev gov uk_government_admin_editions_1374130_confirm_force_publish_lock_version=12(iPad Pro) (1)


davidtrussler (approved):
LGTM! 👍

app/controllers/admin/edition_workflow_controller.rb
```diff
@@ -78,6 +78,8 @@ def confirm_force_publish
     unless @edition.valid?(:publish)
       redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ")
```
davidgisbey, line 79: It looks like a couple of tests are failing

I think for this one we just need to add a test which gives the user the next release permission and check it renders confirm_publish and update this test to confirm_published_legacy

image

And for this one 
image

What's happening is it's checking for a redirect, but fails as there's a render action after the redirect

You can fix it by doing 

```
unless @edition.valid?(:publish)
   redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ")
   return
end

render_design_system(:confirm_force_publish, :confirm_force_publish_legacy, next_release: true)
```
or there's a handy bit of rails syntax 

```
redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ") and return unless @edition.valid?(:publish)

```

dnkrj: All done, thank you!

PR #7053, opened 2022-11-11
896 add RTL support for attachments

davidtrussler:
[Trello card](https://trello.com/c/RT5RbtmT/896-add-rtl-support-for-attachments)

This work fixes a bug introduced in the port to the Design System on the Attachments pages.

Previously, if the user changed the value of the Display Language to a language that is written right to left, there was a change to the class in the container elements around the title and body inputs and a corresponding style change to allow for a right to left language. This functionality was initially lost and is restored here. 

Additionally the means by which text direction on the relevant elements is implemented by the browser is updated to use the `dir` attribute in the HTML in place of using CSS. This is more semantic and allows the text to render correctly in cases where CSS is not present (such as search engines).

The changes introduced by this PR: 
- add a new module to handle the language change
- adds the `dir` attributes required to display the text direction correctly

Additionally some class names have been updated throughout the form for consistency.

### DIFF
--- a/app/assets/stylesheets/admin/views/_attachments.scss
+++ b/app/assets/stylesheets/admin/views/_attachments.scss
@@ -1,0 +1,6 @@
+.right-to-left {
+  input,
+  textarea {
+    direction: rtl;
+  }
+}
--- a/app/views/admin/attachments/_form.html.erb
+++ b/app/views/admin/attachments/_form.html.erb
@@ -1,9 +1,7 @@
-<% initialise_script "GOVUK.AdminAttachmentsForm", selector: '.js-attachment-form', right_to_left_locales: Locale.right_to_left.collect(&:to_param) %>
-
-<%= form_for attachment, url: [:admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], as: :attachment, html: { :class => "js-attachment-form" }, multipart: true do |form| %>
+<%= form_for attachment, url: [:admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], as: :attachment, html: { :class => "attachment-form", :"data-module" => "LocaleSwitcher", :"data-rtl-locales" => Locale.right_to_left.collect(&:to_param) }, multipart: true do |form| %>
   <% content_for :error_summary, render('shared/error_summary', object: attachment, parent_class: "attachment", class_name: @attachment.class.name.demodulize.underscore.humanize.downcase) %>
 
-  <div class="govuk-!-margin-bottom-8">
+  <div class="govuk-!-margin-bottom-8 attachment-form__title <%= "right-to-left" if form.object.rtl_locale? %>">
     <%= render "govuk_publishing_components/components/input", {
       label: {
         text: "Title (required)"
@@ -25,7 +23,7 @@
     <%= form.fields_for :govspeak_content do |govspeak_fields| %>
       <%= hidden_field_tag 'attachment[govspeak_content_attributes][manually_numbered_headings]', "0" %>
 
-      <div class="govuk-!-margin-bottom-8">
+      <div class="govuk-!-margin-bottom-8 attachment-form__manually-numbered-headings">
         <%= render "govuk_publishing_components/components/checkboxes", {
           name: "attachment[govspeak_content_attributes][manually_numbered_headings]",
           heading: "Manually numbered headings",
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/admin/views/_attachments.scss
```diff
@@ -0,0 +1,9 @@
+.js-title,
```
davidgisbey, line 1: So i think a few things are required to get this to work. I think we can just apply this to the .rtl class 

```
.right-to-left {
  input,
  textarea {
    direction: rtl;
  }
}

```

app/views/admin/attachments/_form.html.erb
```diff
@@ -1,9 +1,7 @@
-<% initialise_script "GOVUK.AdminAttachmentsForm", selector: '.js-attachment-form', right_to_left_locales: Locale.right_to_left.collect(&:to_param) %>
-
-<%= form_for attachment, url: [:admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], as: :attachment, html: { :class => "js-attachment-form" }, multipart: true do |form| %>
+<%= form_for attachment, url: [:admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], as: :attachment, html: { :class => "attachment-form", :"data-module" => "LocaleSwitcher", :"data-rtl-locales" => Locale.right_to_left.collect(&:to_param) }, multipart: true do |form| %>
   <% content_for :error_summary, render('shared/error_summary', object: attachment, parent_class: "attachment", class_name: @attachment.class.name.demodulize.underscore.humanize.downcase) %>
 
-  <div class="govuk-!-margin-bottom-8">
+  <div class="govuk-!-margin-bottom-8 attachment-form__title">
```
davidgisbey, line 4: and we need to add this so it applies the class 

```

  <div class="govuk-!-margin-bottom-8 attachment-form__title <%= "right-to-left" if form.object.rtl_locale? %>">

```

app/views/admin/attachments/_form.html.erb
```diff
@@ -43,20 +41,22 @@
...
-          heading_size: "l",
-          text: "Body (required)"
-        },
-        name: "attachment[govspeak_content_attributes][body]",
-        rows: 20,
-        id: "attachment_govspeak_content_body",
-        value: form.object.govspeak_content.body,
-        error_items: errors_for(attachment.errors, :"govspeak_content.body"),
-      } %>
+      <div class="govuk-!-margin-bottom-8 attachment-form__body">
```
davidgisbey, line 44: And this 

```
<div class="govuk-!-margin-bottom-8 attachment-form__body <%= "right-to-left" if form.object.rtl_locale? %>"">

```

davidgisbey:
Hey, nice work. I think a few things need to be updated so it applies the class on page load. But other than that it's good to go!

app/assets/stylesheets/admin/views/_attachments.scss
davidtrussler: Good point! Have updated this.

app/views/admin/attachments/_form.html.erb
davidtrussler: Oh yes, that was in the bootstrap version, now added. 👍

davidgisbey (approved):
lgtm!

app/views/admin/attachments/_form.html.erb
```diff
@@ -43,20 +41,22 @@
...
-          heading_size: "l",
-          text: "Body (required)"
-        },
-        name: "attachment[govspeak_content_attributes][body]",
-        rows: 20,
-        id: "attachment_govspeak_content_body",
-        value: form.object.govspeak_content.body,
-        error_items: errors_for(attachment.errors, :"govspeak_content.body"),
-      } %>
+      <div class="govuk-!-margin-bottom-8 attachment-form__body" dir="<%= form.object.rtl_locale? ? "rtl" : "ltr" %>">
```
ollietreend, line 44: I have a slight concern that we're setting `dir="rtl"` too 'high up' in the DOM.

This `<div class="attachment-form__body">` doesn't _only_ contain RTL user input (i.e. the `<textarea>` and rendered preview) – it also contains UI elements such as the `<label for="attachment_govspeak_content_body">` and the `<button>Preview</button>`.

I don't think those UI elements should switch to right-to-left just because the user input does. It seems semantically 'wrong' to say that the UI is switching direction.

For example, this is how the UI changes when I toggle that `dir` attribute.

| `dir="ltr"` | `dir="rtl"` |
| --- | --- |
| Left to right | Right to left |

**Disclaimer:** I've done this in my browser's dev tools in the integration environment, so I'm not actually running your branch. But I think this is how it'll display nonetheless.

ollietreend (commented):
I think the `dir="rtl"` approach is much better than the original CSS approach. Thanks for doing this 👍🏻 

I just have one concern, though.

app/views/admin/attachments/_form.html.erb
davidtrussler: @ollietreend Yes, fair point, this is a hangover from making changes "like-for-like". Given the update in the means by which the text direction is handled it's probably a sensible enhancement.

PR #7054, opened 2022-11-11
Make attachments index page and secondary_navigation helper methods model agnostic

davidgisbey:
## Description 

The attachments index page and the `secondary_navigation_tabs_items` helper methods have both been built with the assumption that an edition is passed in.

This means that they're both breaking when the index page is rendered for other types of models, such as Policy Groups.

## Changes

This PR does a few things:

1. Moves the attachments index page into the next release and fixes the feature specs
2. Updates the attachments index page to be model agnostic & ensures the correct endpoints are present for Policy Groups and Responses to render confirm delete and reorder pages
3. Moves all helper methods that construct nav tabs into the `TabbedNavHelper`. It then adds functionality for the Policy groups tabs. This will be easily extendable for various other objects, such as Responses and People when we get to porting those sections/pages.

image

## Trello card 

https://trello.com/b/0n3p84mY/govuk-publishing-service-publisher-experience-team-doing


app/helpers/admin/tabbed_nav_helper.rb
```diff
@@ -1,4 +1,77 @@
 module Admin::TabbedNavHelper
+  include Admin::EditionsHelper
+
+  def secondary_navigation_tabs_items(model, current_path)
```
davidgisbey, line 4: I did think about doing something similar to this then alias_methoding the descendants but it seemed a bit overkill 

https://stackoverflow.com/questions/2393697/look-up-all-descendants-of-a-class-in-ruby

app/helpers/admin/tabbed_nav_helper.rb
```diff
@@ -1,4 +1,77 @@
...
+    if model.is_a?(Edition)
```
davidgisbey, line 5: This is required so it doesn't call the editions type in the method below 

i.e.  
```
send("#{model.class.model_name.param_key}_nav_items", model, current_path)
 ```

calls 

```
consultation_nav_items(model, current_path)
```
for a consultation.

To avoid this I did think about doing something similar to this https://stackoverflow.com/questions/2393697/look-up-all-descendants-of-a-class-in-ruby

Then adding alias methods for the subclasses, but it seemed a bit overkill tbh.

baisa (approved):
Nice refactoring @davidgisbey 👏 LGTM 👍

PR #7057, opened 2022-11-14
Add fact check requests journey to the design system

davidgisbey:


nnagewad:
Hey bud,

Just wondering if we've explored a few possibilities:

- Move the Notification banner from the 1/3 container to the 2/3 container.
- Instead of using the Notification banner have we considered using the Inset component?
- The copy beneath the "Extra instructions" could be styled much like our paragraph content

Happy to discuss.

davidgisbey:
@nnagewad thanks for the feedback. Have made the changes 👍

app/views/admin/editorial_remarks/new.html.erb
```diff
@@ -1,21 +1,34 @@
...
   </div>
 
-  <div class="col-md-4">
-    <span class="back">
-      back to <%= link_to 'document', admin_edition_path(@edition) %>
-    </span>
+  <div class="govuk-grid-column-one-third">
+    <p class="govuk-body">
+      back to <%= link_to 'document', admin_edition_path(@edition), class: "govuk-link" %>
+    </p>
```
ollietreend, line 15: Would it be more consistent with other Design System'd pages if this was **removed** and replaced by a 'Cancel' link alongside the 'Submit remark' button?

i.e. something like this:

Submit or cancel

app/views/admin/editorial_remarks/new.html.erb
```diff
@@ -1,21 +1,34 @@
...
+          bold: true,
+        },
+        name: "editorial_remark[body]",
+        id: "editorial_remark_body",
+        error_items: errors_for(@editorial_remark.errors, :body),
+        rows: 20,
+      } %>
+
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Submit remark"
```
ollietreend, line 30: This button should probably become 'Add internal note' – although you may choose to do that as part of the [Move Notes/History/Fact-checking tabs](https://trello.com/c/xiy8xs27/888-move-notes-history-fact-checking-tabs-to-the-design-system) card.

app/views/admin/editorial_remarks/new.html.erb
```diff
@@ -1,21 +1,34 @@
-<% page_title "New Editorial Remark" %>
+<% content_for :page_title, "New editorial remark" %>
```
ollietreend, line 1: Ditto for the wording used here.

Also, we should probably set the page 'context' to the edition title, for consistency with other Design System'd pages – for example, the 'Topic taxonomy tags' page

https://github.com/alphagov/whitehall/blob/2b90ee26775ea4484312ce9d6c9201a8a509d9c3/app/views/admin/edition_tags/edit.html.erb#L1-L3

Again – this might be something you'd prefer to cover in the [Move Notes/History/Fact-checking tabs](https://trello.com/c/xiy8xs27/888-move-notes-history-fact-checking-tabs-to-the-design-system) card rather than this PR.

PR #7058, opened 2022-11-15

beccapearce:
When running tests in parallell the scheduled publishing worker test has a tendency to randomly fail. This seems to be because old editions are not properly discarded or are running at the same time which causes the test to fail.

To fix this we have stubbed out the sidkiq jobs so they won't interfere with eachother.


beccapearce:
Replaced by the approach in https://github.com/alphagov/whitehall/pull/7071

PR #7065, opened 2022-11-16
Tidy whitehall routes

callumknights:

### DIFF
--- a/app/helpers/public_document_routes_helper.rb
+++ b/app/helpers/public_document_routes_helper.rb
@@ -25,6 +25,7 @@ module PublicDocumentRoutesHelper
       build_url_for_corporate_information_page(edition, options)
     else
       polymorphic_url(edition.path_name, options.reverse_merge(id: edition.document))
+      # "/government/collections/#{edition.document.slug}"
     end
   end
 
@@ -142,6 +143,23 @@ module PublicDocumentRoutesHelper
            end
 
     append_url_options("/government/topical-events/#{slug}/about", options)
+  end
+
+  def document_collection_path(object, options = {})
+    slug = case object
+           when String
+             object
+           when Hash
+             object[:id].slug
+           else
+             raise ArgumentError, "Must provide a slug or DocumentCollection"
+           end
+
+    append_url_options("/government/collections/#{slug}", options)
+  end
+
+  def document_collection_url(object, options = {})
+    document_collection_path(object, options)
   end
 
 private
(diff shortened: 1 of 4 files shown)


kevindew:
Continuing from our chat on Slack.

I wonder if we want to look collectively at dealing with the Edition routes. I'm expecting we don't actually need the general helper method (`consultation_path()` etc) but actually find they all go through the same `document_url` esque helpers. It'd be good to get thoughts from @brucebolt when he's back from leave.

Some ideas I explored a little were:

Could the edition just tell us it's paths, this would mean we could remove all the `Whitehall.url_maker...` etc bits and pieces. I imagine to do this we'd need an Edition to have a public_url method and a preview_url method. I prototyped it a bit in: 

https://github.com/alphagov/whitehall/blob/cb0dd70d88ca606fd9c10c5ad08a0d97c81b83ad/app/models/document_collection.rb#L52-L84

Though I was unsure of best way to handle inheritance with Edition to avoid duplicating a lot.

The less radical idea I had was to rewrite the public_document_url and similar helpers, which would require less change throughout the codebase but would need to have a bunch of conditions to determine which Edition (and an annoying edge case for worldwide org corporate information pages):

https://github.com/alphagov/whitehall/blob/77abbc29fc584bda8a7b6202ebe23ca86c2a7649/app/helpers/public_document_routes_helper.rb#L4-L49

brucebolt:
I do feel that removing `Whitehall.url_maker` would be the best solution, as this whole area of the application is rather confusing to understand and seems it could be simplified considerably.  Maybe we should spike into removing this?

kevindew:
@brucebolt yup I like the sound of that. Do you like the idea of editions knowing their own routes as per: https://github.com/alphagov/whitehall/blob/cb0dd70d88ca606fd9c10c5ad08a0d97c81b83ad/app/models/document_collection.rb#L52-L84 or are you thinking start from something different?

brucebolt:
Editions (and other non-editionable document types) knowing their own routes would be my preferred solution.  If we can share most of the code for editionable documents, then that'd be great.

I did start going down this route with Topical Events and World Locations, but ended up tangled in a `url_maker` mess.

If we can remove `url_maker` completely, we can then go down the route of the routes being defined in the model and remove a lot of code that seems rather complicated for what it's actually doing.

kevindew:
> Editions (and other non-editionable document types) knowing their own routes would be my preferred solution. If we can share most of the code for editionable documents, then that'd be great.
> 
> I did start going down this route with Topical Events and World Locations, but ended up tangled in a `url_maker` mess.
> 
> If we can remove `url_maker` completely, we can then go down the route of the routes being defined in the model and remove a lot of code that seems rather complicated for what it's actually doing.

Aces that makes sense to me.

What seems somewhat tricky is finding the right balance of having code that doesn't repeat itself loads and is also not caught up in a complex inheritance situation :-/

PR #7067, opened 2022-11-16
Increase max width of admin GOV.UK Design System pages

davidgisbey:
## Description

This increases the max width of the pages in the admin interface of Whitehall in the GOV.UK Design System to 1140px so it is line with the Bootstrap implementation. This will allow us to develop feature parity in a like for like way.

In particular we need this to be able to port the history, notes and fact checking tabs.

## Trello card

https://trello.com/c/XQ6gelOt/894-do-we-need-to-increase-the-maximum-page-width

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,3 +1,5 @@
+$govuk-page-width: 1140px;
+
 @import "govuk_publishing_components/all_components";
 
 @import "./components/autocomplete";


app/assets/stylesheets/application.scss
```diff
@@ -27,6 +27,12 @@
   }
 }
 
+.govuk-width-container {
+  max-width: 1140px;
+  padding-left: 15px;
+  padding-right: 15px;
+}
```
DilwoarH, line 34:
```suggestion
.govuk-width-container--large {
  max-width: 1140px;
  padding: 0 govuk-spacing(3);
}
```

I would add it as an additional new BEM modifier rather than overriding the original styling. 

Also do we need the padding? should it not just keep the padding that exists?

The implementation will be as follows:

```html
<div class="govuk-width-container govuk-width-container--large"></div>
```

DilwoarH: cc @kevindew - do you have anything to add to this?

kevindew:
Just some passing thoughts on this.

It seems generally unusual to override design system styles (bit like CSS equivalent of monkey patching) so not sure if we could have some surprises.

But design system does have width set-up as a setting that we could be overriding with a variable: https://github.com/alphagov/govuk-frontend/blob/4b5ed0ada3b64722ae2074dfa0db5d4d3a45cb79/src/govuk/settings/_measurements.scss#L14 which would be a more friendly way to change it (I think we might lose some calculated changes in components: https://github.com/alphagov/govuk-frontend/search?q=%24govuk-page-width)

I'm not sure if this affects media queries at all, will large screen still start at 960?

app/assets/stylesheets/application.scss
DilwoarH: Also, 1140px is a very odd number (how did we get to this), does 1200px not work?

DilwoarH:
> Just some passing thoughts on this.
> 
> It seems generally unusual to override design system styles (bit like CSS equivalent of monkey patching) so not sure if we could have some surprises.
> 
> But design system does have width set-up as a setting that we could be overriding with a variable: https://github.com/alphagov/govuk-frontend/blob/4b5ed0ada3b64722ae2074dfa0db5d4d3a45cb79/src/govuk/settings/_measurements.scss#L14 which would be a more friendly way to change it (I think we might lose some calculated changes in components: https://github.com/alphagov/govuk-frontend/search?q=%24govuk-page-width)
> 
> I'm not sure if this affects media queries at all, will large screen still start at 960?

I can confirm adding the following to line one of `application.scss` works:


```
$govuk-page-width: 1200px;
```

davidgisbey:
Do we have a preference for one over the other?

DilwoarH:
> Do we have a preference for one over the other?

The variable will be the best implementation as it will be passed into the design system library and used for anywhere where the `$govuk-page-width` is calculated.

app/assets/stylesheets/application.scss
davidgisbey: I wasn't in the meeting but i think 1140px was decided on as it's what Boostrap has. And the idea was it would be less jarring when users jump from page to page between the design systems.

app/assets/stylesheets/application.scss
```diff
@@ -27,6 +27,8 @@
   }
 }
 
+$govuk-page-width: 1140px;
```
DilwoarH, line 30: I believe it needs to be at the top @davidgisbey.

Also the standard pattern is to put all variables at the top of the file as well.

DilwoarH (changes requested):

app/assets/stylesheets/application.scss
davidgisbey: Okay great thanks. Have made the change

ollietreend (approved):

PR #7071, opened 2022-11-17

beccapearce:
When running tests in parallell the scheduled publishing worker test has a tendency to randomly fail. This seems to be because old editions are not properly discarded or are running at the same time which causes the test to fail.

To fox this we have set up the tests to run in a different queue for each parallel process that runs. This should prevent the other queues from interfering with the current running test.

We have also move away from using the redis namespace config as it is being removed in the next major version.

### DIFF
--- a/test/support/sidekiq_test_helpers.rb
+++ b/test/support/sidekiq_test_helpers.rb
@@ -5,7 +5,10 @@ module SidekiqTestHelpers
   def with_real_sidekiq
     Sidekiq::Testing.disable! do
       Sidekiq.configure_client do |config|
-        config.redis = { namespace: "whitehall-test" }
+        # The Rails built-in test parallelization doesn't make it easy to find the worker number.
+        # So here we're using the number suffixed to the Postgres database name, because that is unique for each worker.
+        parallel_test_runner_number = ActiveRecord::Base.connection.current_database.split("-").last.to_i
+        config.redis = { db: parallel_test_runner_number }
       end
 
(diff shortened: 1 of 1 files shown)


test/support/sidekiq_test_helpers.rb
```diff
@@ -5,7 +5,8 @@ module SidekiqTestHelpers
   def with_real_sidekiq
     Sidekiq::Testing.disable! do
       Sidekiq.configure_client do |config|
-        config.redis = { namespace: "whitehall-test" }
+        parallel_test_runner_number = ActiveRecord::Base.connection.current_database.split("-").last.to_i
+        config.redis = { db: parallel_test_runner_number }
```
ollietreend, line 9: It might be worth adding a comment to explain the slightly odd way we're getting the worker number.

```suggestion
        # The Rails built-in test parallelization doesn't make it easy to find the worker number.
        # So here we're using the number suffixed to the Postgres database name, because that is unique for each worker.
        parallel_test_runner_number = ActiveRecord::Base.connection.current_database.split("-").last.to_i
        config.redis = { db: parallel_test_runner_number }
```

ollietreend (commented):
Really nice work digging into this @beccapearce – and I can't wait for this test to stop flaking!

ollietreend (approved):

brucebolt:
This is amazing, don't know how many times this has failed on me, great work @beccapearce 🎖️!

PR #7073, opened 2022-11-17
Delete document source

baisa:
They have served their purpose when publications from other websites were being recreated on the GOV.UK. The documents that have Document Sources populated are all 7 years or older.

[Trello card](https://trello.com/c/c7cIZe5m/901-delete-document-source-related-codebase-in-whitehall)


db/migrate/20221117134008_drop_document_sources.rb
```diff
@@ -0,0 +1,5 @@
+class DropDocumentSources < ActiveRecord::Migration[7.0]
+  def change
+    drop_table :document_sources
```
ollietreend, line 3: This should stop the Rubocop complaints by providing a way for this migration to rollback.

```suggestion
    drop_table :document_sources do |t|
      t.integer "document_id"
      t.string "url", null: false
      t.integer "import_id"
      t.integer "row_number"
      t.string "locale", default: "en"
      t.index ["document_id"], name: "index_document_sources_on_document_id"
      t.index ["url"], name: "index_document_sources_on_url", unique: true
    end
```

app/controllers/admin/document_sources_controller.rb
```diff
@@ -1,29 +0,0 @@
-class Admin::DocumentSourcesController < Admin::BaseController
-  before_action :require_import_permission!
```
ollietreend, line 2: It looks like we can now remove the method `require_import_permission!` from Base Controller:

https://github.com/alphagov/whitehall/blob/fc1d1208c2494257ad33e5f1563c30b2bf9ea5b4/app/controllers/admin/base_controller.rb#L21-L23

As well as the "import" permission in the User model:

https://github.com/alphagov/whitehall/blob/ba7e7191b0b5230eb59e3cddb22ad79386b76cb6/app/models/user.rb#L23

https://github.com/alphagov/whitehall/blob/ba7e7191b0b5230eb59e3cddb22ad79386b76cb6/app/models/user.rb#L83-L85

These seem to have only been used for the Document Sources functionality.

ollietreend (approved):
Great work @baisa! ✨🧹🚀

PR #7079, opened 2022-11-22
Use a single route for International Delegation pages

brucebolt:
We currently have two almost identical pages for International Delegations: `/world/slug` and `/world/slug/news`.

We are deprecating the news version of this page, so need to update the presenter to present the correct URL and redirect `/world/slug/news` to `/world/slug`.

Not to be merged until https://github.com/alphagov/collections/pull/3032 and https://github.com/alphagov/collections/pull/3049 are deployed.

Once this has been merged, the following will need to be done:
- all International Delegation pages republished, using the following rake task: `publishing_api:bulk_republish:document_type[WorldLocationNews]`
- the rake task to redirect all the news pages run: `temp_redirect_international_delegation`
- verify that International Delegation news pages (e.g. https://www.gov.uk/world/uk-mission-to-the-united-nations/news) redirect to the non-news page (e.g. https://www.gov.uk/world/uk-mission-to-the-united-nations) and the destination is now rendered by Collections
- merge https://github.com/alphagov/whitehall/pull/7082 to remove the rake task

[Trello card](https://trello.com/c/2HducX7L)


app/presenters/publishing_api/world_location_news_presenter.rb
```diff
@@ -57,7 +63,11 @@ def featured_links
     end
 
     def path_for_news_page
-      Whitehall.url_maker.world_location_news_index_path(world_location)
+      if world_location.world_location?
+        Whitehall.url_maker.world_location_news_index_path(world_location)
+      elsif world_location.international_delegation?
+        Whitehall.url_maker.world_location_path(world_location, locale: I18n.locale)
```
BeckaL, line 69: Why don't world location news need this locale in their path too?

brucebolt: It looks like I've added in the locale too early (it was supposed to be in [this PR](https://github.com/alphagov/whitehall/pull/6961/commits/ac3fc6487a540539ad3916e5a74f9bc1f366ea25#diff-e30a477531e7255e6c1245b17ca6e74324183262f63e8bf6a960102203fef22cR60) which will come after this).

At the moment, we're using the Rails automagic route helper method, so Rails deals with the locale itself.  After we remove the route in a later PR, we'll need to use our own helper which requires us to specify the locale.

I'll remove this from here as we don't need it yet.

BeckaL (approved):
Looks good to me once locale removed to add later as per your comment

brucebolt:
@BeckaL I've pushed an extra commit to this PR, as I realised we were publishing both the WorldLocation and WorldLocationNews to `/world/slug` at the same time when an International Delegation was updated.  As non-International Delegation World Locations weren't being published anyway, we can now remove all the publishing code from these models.

Would you be able to review the latest commit?

PR #7080, opened 2022-11-22
Introduce Document::PaginatedTimeline

ollietreend:
This PR introduces the new class Document::PaginatedTimeline as a means to merge the [Document::PaginatedHistory] and [Document::PaginatedRemarks] classes. It builds upon the spike PR #6868 but adds test coverage and implements some of the remaining 'to do's from that spike.

The Document ‘Timeline’ is a representation of the History (Versions) and Notes (Editorial Remarks) of a Document, combined into one chronological list.

It will allow us to combine the 'history' and 'notes' tabs shown on the edition show & edit pages into one single tab.

How does the data get combined?
===============================

I've used a [union query] to merge the queries for `Document#edition_versions` and `Document#editorial_remarks` into one result set, sorted by creation date. This provides a single chronological timeline of ‘history’ and ‘note’ entries.

Union queries aren't natively supported by Active Record. While it's possible to [extend Active Record to support union queries][activerecord-union], that starts to make things more complicated than they really need to be. We have a very specific use case for union queries, so it seems simpler to perform the query manually rather than bending Active Record to support it.

Both queries need to return the same columns so they can be merged. So we return three fields that exist in both tables: id, created_at, and model_name (so we know which table the record came from). We include created_at so the list can be sorted chronologically.

How to render timeline entries?
===============================

Call the method `Document::PaginatedTimeline#entries` to retrieve the timeline entries for rendering to the page. These entries will either be `AuditTrailEntry` (History) objects or `EditorialRemark` (Note) objects.

Change to database.yml
======================

I've enabled ‘prepared queries’ in the MySQL database adapter config. This allows us to safely escape parameters in SQL queries (avoiding SQL injection attacks). They’re currently disabled in Whitehall (which is the Rails default), but will be [enabled by default in Rails 7.2] so this should be a non-controversial change.

[Document::PaginatedHistory]: https://github.com/alphagov/whitehall/blob/f0f0adfbc349293f9dc37a995b33e791021edf62/app/models/document/paginated_history.rb
[Document::PaginatedRemarks]: https://github.com/alphagov/whitehall/blob/f0f0adfbc349293f9dc37a995b33e791021edf62/app/models/document/paginated_remarks.rb
[union query]: https://dev.mysql.com/doc/refman/8.0/en/union.html
[activerecord-union]: https://joshfrankel.me/blog/a-journey-into-writing-union-queries-with-active-record/
[enabled by default in Rails 7.2]: https://github.com/rails/rails/blob/c41e2f1bca65b51bccda7dd465d525071c66264f/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L172

---

Trello: https://trello.com/c/xiy8xs27/888-move-notes-history-fact-checking-tabs-to-the-design-system

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -25,7 +25,7 @@ class Attachment < ApplicationRecord
           joins(:attachment_data).where("attachment_data.carrierwave_file = ?", basename)
         }
 
-  scope :files, -> { where(type: FileAttachment) }
+  scope :files, -> { where(type: "FileAttachment") }
 
   scope :for_current_locale, -> { where(locale: [nil, I18n.locale]) }
 
--- a/app/models/document/paginated_history.rb
+++ b/app/models/document/paginated_history.rb
@@ -41,6 +41,12 @@ class Document::PaginatedHistory
       @preloaded_previous_version = previous_version
     end
 
+    def ==(other)
+      self.class == other.class &&
+        version == other.version &&
+        action == other.action
+    end
+
     def actor
       version.user
     end
--- a/app/models/document/paginated_timeline.rb
+++ b/app/models/document/paginated_timeline.rb
@@ -1,0 +1,127 @@
+class Document::PaginatedTimeline
+  PER_PAGE = 30
+
+  def initialize(document:, page:)
+    @document = document
+    @page = page.to_i
+  end
+
+  def entries
+    @entries ||= begin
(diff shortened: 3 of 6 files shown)


davidgisbey (approved):
I've given this a good look through and it's working nicely. I've built upon it when addings tabs,  but that'll come in a separate PR

PR #7094, opened 2022-11-28
Add GitHub Action workflow for verifying pact tests

kevindew:
Trello: https://trello.com/c/QrxjshEm/270-move-ruby-gems-from-jenkins-to-github-actions

This moves the pact verification to a GitHub Action and will no longer be performed by a Jenkins build. This workflow will also be called by GDS API Adapters as part of its CI build to assert that newly generated pacts are valid against this app. To avoid running this same task twice `pact:verify` is no longer part of the Jenkins test actions and instead the other rake default steps are included.

For more details see https://github.com/alphagov/gds-api-adapters/pull/1175

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@ node {
           echo "Running a subset of the tests to check the content schema changes"
           govuk.runRakeTask("test:publishing_schemas --trace")
         } else {
-          sh("bundle exec rake")
+          sh("bundle exec rake lint test cucumber jasmine")
         }
       }
     }
--- a/.github/workflows/pact-verify.yml
+++ b/.github/workflows/pact-verify.yml
@@ -1,0 +1,49 @@
+# Pact verify workflow
+#
+# This workflow asserts that Pact contract tests are valid against this
+# codebase. It is trigged when changes are made to this project and it
+# is explicitly called by GDS API Adapters when changes are made there.
+on:
+  pull_request:
+  push:
+  workflow_call:
+    inputs:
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -15,7 +15,7 @@ node {
           echo "Running a subset of the tests to check the content schema changes"
           govuk.runRakeTask("test:publishing_schemas --trace")
         } else {
-          sh("bundle exec rake")
+          sh("bundle exec rake lint test cucumber jasmine")
```
ChrisBAshton, line 18: Hmmm I understand why you haven't just removed `pact:verify` from the default task (devs should be able to run the full test suite locally), but I think it would be cleaner to define a new Rake task that encapsulates the above, e.g.

```ruby
task ci: %i[lint test cucumber jasmine] # Pact tests run separately on GitHub Actions
```

And then changing this to:

```suggestion
          sh("bundle exec rake ci")
```

But not a blocker.

ChrisBAshton (approved):

Jenkinsfile
kevindew: Sure, this could work but may need some refinement.

I don't like the idea of it being `ci` as pact:verify will be also a part of CI. Could call it `jenkins_ci`, but that couples it to Jenkins. I suppose could be `ci_build` hmmm. I also wonder if it's actually better to be direct in the Jenkinsfile and whether adding another task ends up being misdirection.

ChrisBAshton: Fair point 👍 happy to leave as-is.

PR #7104, opened 2022-11-30
Add aria attributes to button and preview in Govspeak Preview component

jyoung-gds:
[Trello ticket link](https://trello.com/c/oGsq2F3m/828-add-accessibility-improvements-to-preview-govspeak-component)

## WHAT
- Adds `aria-controls` and `aria-describedby` attributes to the button render
- Adds `aria-live` and `id` attributes to the preview div
- Adds `id` attribute to label (needs [this PR](https://github.com/alphagov/govuk_publishing_components/pull/3093) to be merged and released before this change can be deployed)
- Updates `govuk_publishing_components` to 34.1.0 (required for the addition of ID attribute options on label)

## WHY
To improve accessibility of the Govspeak Preview component

⚠️ Not to be merged until [this change](https://github.com/alphagov/govuk_publishing_components/pull/3093) has been released in publishing components ⚠️

### DIFF
--- a/app/views/components/_govspeak-editor.html.erb
+++ b/app/views/components/_govspeak-editor.html.erb
@@ -1,5 +1,7 @@
 <%
   id ||= "#{name}-#{SecureRandom.hex(4)}"
+  preview_id = "#{id}-preview"
+  label_id = "#{id}-label"
   label[:bold] ||= false
   hint ||= nil
   hint_id ||= "#{id}-hint"
@@ -35,6 +37,7 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-three-quarters ">
       <%= render "govuk_publishing_components/components/label", {
+        id: label_id,
         html_for: id,
         hint_id: hint_id,
         hint_text: hint,
@@ -47,7 +50,9 @@
         margin_bottom: hint ? 5 : 2,
         classes: "js-app-c-govspeak-editor__preview-button",
         data_attributes: preview_button_data_attributes,
-        type: "button"
+        aria_controls: preview_id,
+        aria_describedby: label_id,
+        type: "button",
       } %>
     </div>
   </div>
@@ -63,7 +68,7 @@
       describedby: hint_id,
     } %>
   </div>
-  <div class="app-c-govspeak-editor__preview">
+  <div class="app-c-govspeak-editor__preview" aria-live="polite" id="<%= preview_id %>">
     <p class="govuk-body">Generating preview, please wait.</p>
   </div>
 <% end %>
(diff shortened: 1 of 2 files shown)


app/views/components/_govspeak-editor.html.erb
```diff
@@ -63,7 +65,7 @@
       describedby: hint_id,
     } %>
   </div>
-  <div class="app-c-govspeak-editor__preview">
+  <div class="app-c-govspeak-editor__preview" aria-live="polite" id="<%= id %>">
```
dnkrj, line 68: I think that this `id` needs to be different to the textarea `id`

jyoung-gds: @dnkrj updated

app/views/components/_govspeak-editor.html.erb
```diff
@@ -1,5 +1,7 @@
 <%
+  binding.pry
```
kevindew, line 2: Looks like this was left behind by mistake.

app/views/components/_govspeak-editor.html.erb
```diff
@@ -47,7 +49,9 @@
         margin_bottom: hint ? 5 : 2,
         classes: "js-app-c-govspeak-editor__preview-button",
         data_attributes: preview_button_data_attributes,
-        type: "button"
+        aria_controls: "app-c-govspeak-editor__preview",
```
kevindew, line 52: I think this should be a reference to an id, so should be the preview id.

app/views/components/_govspeak-editor.html.erb
```diff
@@ -47,7 +49,9 @@
...
+        aria_describedby: id,
```
kevindew, line 53: I'm not 100% sure we want this, I understand that this should be an id of an item that describes the purpose of this button (like a span with hint text), but I wouldn't have thought the textarea does that?

app/views/components/_govspeak-editor.html.erb
jyoung-gds: oops yep good spot 🤦🏻‍♂️

app/views/components/_govspeak-editor.html.erb
jyoung-gds: It might be good to get @dnkrj's  input on this. I asked him about it in a Slack message and he said we could use the aria-describedby to give context to the button as it is read after the aria-label by screenreaders

kevindew: Sounds good - I'm certainly no expert on aria

PR #7122, opened 2022-12-06
Add UI that allows GDS Editor to update a documents slug

davidgisbey:
## Description

One of the common support requests that we get is to update the slug for a document. Currently we run a rake task to do this https://docs.publishing.service.gov.uk/manual/howto-change-slug-and-create-redirect.html

This PR adds the ability for a GDS Editor to update a documents slug via the UI.

This will likely (although TBD) will be accessible via either the edition summary or edit edition page.  You only have access to this page if you have the GDS Editor permission.

I've added in validating that the slug prior to updating it and pushing it downstream to the Publishing API and Search Index so the user doesn't get sent a 422 from the Publishing API and have an invalid slug persisted on the document.

### Invalid slug

image

### Slug not present

image

### Slug is not unique

image

### Successful 

image

### DIFF
--- a/lib/data_hygiene/document_reslugger.rb
+++ b/lib/data_hygiene/document_reslugger.rb
@@ -1,0 +1,72 @@
+module DataHygiene
+  # Reslugs a document to new_slug.
+  #
+  # When run, the following happens:
+  #
+  #  - changes the documents slug
+  #  - reindexes the document with it's new slug
+  #  - republishes the document to Publishing API (automatically handles the redirect)
+  #
+  class DocumentReslugger
+    attr_reader :document, :published_edition, :current_user, :new_slug
+
+    def initialize(document, published_edition, current_user, new_slug)
+      @document = document
+      @published_edition = published_edition
+      @current_user = current_user
+      @new_slug = new_slug
+    end
+
+    def run!
+      add_errors_if_invalid
+      return false if document.errors.present?
+
+      remove_from_search_index
+      save_document
+      republish_document
+      add_to_search_index
+      create_editorial_remark
+      true
+    end
+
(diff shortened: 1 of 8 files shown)


lib/whitehall/authority/rules/document_rules.rb
```diff
@@ -8,7 +8,12 @@ def initialize(actor, subject)
     end
 
     def can?(action)
-      actor.gds_editor? || actor.departmental_editor? || action == :create
+      case action
+      when :perform_administrative_tasks
+        actor.gds_editor? ? true : false
```
dnkrj, line 13: Do we need the ternary operator here?

dnkrj (approved):

lib/whitehall/authority/rules/document_rules.rb
davidgisbey: Thanks for the review. Yea good point! Will sort it 👍

lib/data_hygiene/document_reslugger.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+  private
+
+    def add_errors_if_invalid
+      document.errors.add(:slug, "is blank") and return if new_slug.blank?
+
+      document.errors.add(:slug, "must be unique") and return if new_slug_is_a_duplicate
+
+      document.errors.add(:slug, "is invalid") unless new_slug.starts_with?("/") && new_slug !~ %r{//} && new_slug !~ %r{./\z}
```
brucebolt, line 39: I've just tried this task and I think we want to ensure the slug doesn't start with `/`, else we add that into the slug too.

PR #7124, opened 2022-12-07
Use govuk-ruby-images for the container image base and builder.

sengi:
This greatly simplifies the Dockerfile, speeds up the build and produces a more robust and up-to-date image.

Tested: running on the integration cluster without errors.


marcpomfret (approved):
LGTM

PR #7125, opened 2022-12-07
Bump Rails Translation Manager and use new plural file

1pretz1:
We're now using the extra plural rules defined by RTM as a prod
dependency [1]. This will replace the list defined in govuk_app_config
and so the reference has also been removed. RTM's version loads
automatically [2] and so we don't have to explicitly include the file in
the app.

[1]: https://github.com/alphagov/rails_translation_manager/blob/main/config/locales/plurals.rb
[2]: https://github.com/alphagov/rails_translation_manager/blob/f18fd2e178719dd0a103981160f898771957fa43/lib/rails_translation_manager.rb#L24

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,6 +45,7 @@ gem "ptools"
 gem "rack"
 gem "rack_strip_client_ip"
 gem "rails-i18n"
+gem "rails_translation_manager"
 gem "rake"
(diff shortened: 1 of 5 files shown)


KludgeKML (approved):
LGTM

PR #7127, opened 2022-12-07
Add classes to the forbidden template so that it renders in both bootstrap and the design system

dnkrj:
## What
Add `govuk-heading-l` and `govuk-back-link` to the appropriate elements in the forbidden template.

## Why
The forbidden template is used across whitehall when a user attempts to access or edit something that they do not have access to, currently when this occurs on a page rendered using the design system layout the content is unstyled.

As adding the logic to check the layout currently being used across all of the pages that could render the forbidden message is complex and the template itself is very simple, this PR adds classes that will be ignored by the bootstrap layout but improve the appearance of the page when using the design system.

## Screenshots
|   | Before | After |
| - | ------- | ----- |
| Design System | whitehall-admin dev gov uk_government_admin_editorial_remarks_662568_confirm_destroy(iPad Pro) (2) | whitehall-admin dev gov uk_government_admin_editorial_remarks_662568_confirm_destroy(iPad Pro) (1) |
| Bootstrap | whitehall-admin dev gov uk_government_admin_editorial_remarks_662568_confirm_destroy(iPad Pro) (3) | whitehall-admin dev gov uk_government_admin_editorial_remarks_662568_confirm_destroy(iPad Pro) (4) |

### DIFF
--- a/app/views/admin/editions/forbidden.html.erb
+++ b/app/views/admin/editions/forbidden.html.erb
@@ -1,7 +1,7 @@
 <% page_title "Access denied" %>
 <div class="no-content no-content-bordered">
-  <h1>Sorry, you don’t have access to this document</h1>
-  <p class="back">
+  <h1 class="govuk-heading-l">Sorry, you don’t have access to this document</h1>
+  <p class="back govuk-back-link">
     <%= link_to 'Back to document list', admin_editions_path %>
   </p>
 </div>


davidgisbey:
Seems like the simplest way round the issue to me. It's such a simple page like you say that i don't see it being a big issue.

ollietreend (approved):
This looks good to me – a nice, simple, pragmatic solution! 🎉 

Thanks for doing this @dnkrj

DilwoarH:
This might cause design issues as this isn't an intended use of the back link

app/views/admin/editions/forbidden.html.erb
```diff
@@ -1,7 +1,7 @@
 <% page_title "Access denied" %>
 <div class="no-content no-content-bordered">
-  <h1>Sorry, you don’t have access to this document</h1>
-  <p class="back">
+  <h1 class="govuk-heading-l">Sorry, you don’t have access to this document</h1>
+  <p class="back govuk-back-link">
     <%= link_to 'Back to document list', admin_editions_path %>
   </p>
```
DilwoarH, line 6:
```
<p class="back govuk-body">
    <%= link_to 'Back to document list', admin_editions_path, class: "govuk-link" %>
  </p>
```

DilwoarH: This might be a better way of doing it

dnkrj: https://github.com/alphagov/whitehall/pull/7149

PR #7083, opened 2022-11-23
Edition summary page to GOVUK Design system

DilwoarH:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

## What

[Move the Edition Summary page to the GOV.UK Design System](https://trello.com/c/voJfdq8h/685-move-edition-summary-page-to-the-govuk-design-system)

## Why

This is part of the transition work to move Whitehall to the GOVUK Design system.

## After

whitehall-admin dev gov uk_government_admin_detailed-guides_1331735(iPad Pro) (1)

## Before

whitehall-admin dev gov uk_government_admin_detailed-guides_1331735(iPad Pro)

PR #7131, opened 2022-12-08
Add user interface for editing change notes

brucebolt:
Editing or deleting public change notes is a common request sent to Technical 2nd Line.

This PR adds a user interface that will allow anyone with GDS Admin permissions (e.g. Content 2nd Line) to edit or delete major change notes, without needing developer input.

This is a major improvement on the previous process, which required developers to run rake tasks.  One of these tasks used a fuzzy match, which did not give developers confidence the correct change note was being deleted.

This also removes the rake tasks, as they will no longer be required.

Ultimately, this will be linked to from a potential 'admin tools' section of Whitehall.

### Listing the existing change notes on a document
Screenshot showing the timestamp and change note for each major change made to a document.  Alongside each are edit and delete button.

### Editing a change note
Screenshot showing the interface for editing a change note.  There is the existing change note, a text input field for the updated note, a submit button and a cancel link.

### Flash message after editing a change note
Screenshot showing listing of changes notes with a flash message that reads "Change note updated".

### Confirm screen for deleting a change note
Screenshot showing the confirmation page for deleting a change note.  There is the existing change note, a delete button and a cancel link.

### Flash message after deleting a change note
Screenshot showing listing of changes notes with a flash message that reads "Change note deleted".

### Editorial remarks after making changes to change notes
Screenshot showing editorial remarks for a document.  One remark reads "Deleted change note: This is a major edit which has been edited" alongside the time and name of the person who deleted the change note, the other reads "Updated change note from This is a major edit to This is a major edit which has been edited" with the same information alongside.

[Trello card](https://trello.com/c/bMwFRQ3e)


app/views/admin/edition_change_notes/edit.html.erb
```diff
@@ -20,6 +20,16 @@
         <%= render "govuk_publishing_components/components/button", {
           text: "Update"
         } %>
+      </div>
+    <% end %>
+
+    <%= form_tag [:admin, :edition_change_note], id: @edition_to_change, method: :delete do %>
```
davidgisbey, line 26: Just a nit, but i'm wondering if should live as an action here instead?
https://github.com/alphagov/whitehall/pull/7131/commits/ca725a3b36f9dd5c0346045585f352f223f8ecf5#diff-b44ff3403d3f2cddd475847e71d41b5181f2762af4ecfe920c2d460a6c6133d7R27-R31

Wdyt?

davidgisbey: Also, it's not a big deal, but should this have a confirm delete page?

davidgisbey (approved):
lgtm. I've left a couple of optional/nitty comments, but no blockers

app/views/admin/edition_change_notes/edit.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    <%= form_for @edition_to_change, method: :patch, url: admin_edition_change_note_path(@edition_to_change) do |form| %>
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: "New change note:"
+        },
+        name: "new_change_note",
+        value: @edition_to_change.change_note,
+      } %>
+
+      <div class="govuk-button-group govuk-!-margin-bottom-6">
```
dnkrj, line 19: Should probably use "Save" and "Cancel" as per other pages.
https://github.com/alphagov/whitehall/blob/main/app/views/admin/attachments/_form.html.erb#L78-L90

app/views/admin/edition_change_notes/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        [
+          {
+            text: I18n.l(change_note.major_change_published_at, format: :long_ordinal),
+          },
+          {
+            text: change_note.change_note,
+          },
+          {
+            text: render("govuk_publishing_components/components/button", {
+              text: "Edit",
```
dnkrj, line 29: For other pages we've used separate links to an edit page and a confirm delete page as David mentioned.
https://github.com/alphagov/whitehall/pull/7083/files#diff-dea9c5138c955d5e779f2c0776692f5063ef6535ebb4afdefeb7ddacdbf99b21R284

dnkrj (commented):
Looks great, couple of minor nits to align this with the rest of the design system pages.
The new summary page is probably the most refined example:
https://github.com/alphagov/whitehall/pull/7083

app/controllers/admin/edition_change_notes_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+    # This is the edition the change note was originally set on, not the latest edition
+    @edition_to_change = Edition.find(params[:id])
+  end
+
+  def update
+    edition_to_change = Edition.find(params[:id])
+    old_change_note = edition_to_change.change_note
+
+    # rubocop:disable Rails/SkipsModelValidations
+    edition_to_change.update_attribute(:change_note, params[:new_change_note])
```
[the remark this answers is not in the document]
brucebolt: > Just a question. We have no validations here. Should we check for presence of input, or are we happy to trust the user since it's a GDS Admin?

I left the validations out as this is just for internal use.  We'd need to review the validations if we were to expand the usage.

> One other question. On second line i've had to change a minor change to a major change before and add a change note. Is that out of scope for this card?

I started writing code to do this, then realised it got more complicated as you had to also request a date/time from the user.  I didn't think I'd get that completed this week, so put it out of scope.  I definitely think this is something we should add in any future iterations.

app/views/admin/edition_change_notes/edit.html.erb
brucebolt: Yes, I think that's a good idea.  I'll add that in.

app/views/admin/edition_change_notes/edit.html.erb
brucebolt: Thanks, I'll update the label and add a 'Cancel' link.

davidgisbey:
lgtm! 🚢

PR #7132, opened 2022-12-08
Dockerfile cleanup.

sengi:
- Use `$APP_HOME` from the base image instead of hardcoding `/app` throughout.
- Remove the `SHELL` instruction since there are no longer any pipelines.
- Set the default uid to the non-root user from the base image.

Tested: works on the integration cluster.


nsabri1 (approved):
LGTM

PR #7135, opened 2022-12-12
Do not require a `/` at the start of slugs

brucebolt:
We are currently requiring a `/` at the start of slugs when documents are reslugged through the user interface.

Whitehall doesn't store this at the start of it's slugs in the database, so we need to update the validation to ensure this is not included.


lib/data_hygiene/document_reslugger.rb
```diff
@@ -36,7 +36,7 @@ def add_errors_if_invalid
 
       document.errors.add(:slug, "must be unique") and return if new_slug_is_a_duplicate
 
-      document.errors.add(:slug, "is invalid") unless new_slug.starts_with?("/") && new_slug !~ %r{//} && new_slug !~ %r{./\z}
+      document.errors.add(:slug, "is invalid") if new_slug.starts_with?("/") || new_slug =~ %r{//} || new_slug =~ %r{./\z}
```
BeckaL, line 39: I'm not great at regex in ruby - what is this condition in words? Could we add unit tests for each of the conditions, that way at least someone like me can read the tests and understand what's invalid :-)

davidgisbey (approved):
thanks for sorting that lgtm

PR #7148, opened 2022-12-14
Add history tab in the GOV.UK Design System

davidgisbey:
## Description

This PR follows on from https://github.com/alphagov/whitehall/pull/7144 which added tabs to the edit edition, edition summary and edit translation page.

It populated the govspeak help tab, but left the history and fact checking tabs blank.

This PR populates the history tab.

_Note: It does not add pagination via JS, which will be added in a follow up PR. Pagination does work, but it will reload the page_

## Guidance for review

By commit will be easiest.

### Edit edition page 

image

### Edition Summary page

image

### Edit translation page

image

### DIFF
--- a/app/components/admin/editions/document_history_tab_component.html.erb
+++ b/app/components/admin/editions/document_history_tab_component.html.erb
@@ -10,6 +10,8 @@
   text: sanitize("History and notes have been merged. #{tag.a('Read more about the change', href: admin_whats_new_path)}")
 } %>
 
+<%= paginate(@document_history, theme: 'history') %>
+
 <% if entries_on_newer_editions.present? %>
   <div class="app-component-editions__newer-edition-entries">
     <h3 class="govuk-heading-m">On newer editions</h3>
--- a/app/components/admin/editions/document_history_tab_component.html.erb
+++ b/app/components/admin/editions/document_history_tab_component.html.erb
@@ -1,0 +1,47 @@
+<h2 class="govuk-heading-l">History</h2>
+
+<% if editing %>
+  <p class="govuk-body">To add an internal note, save your changes.</p>
+<% end %>
+
+<p class="govuk-body"><%= link_to("Add internal note", new_admin_edition_editorial_remark_path(edition), class: "govuk-body govuk-link") %></p>
+
+<%= render "govuk_publishing_components/components/inset_text", {
+  text: sanitize("History and notes have been merged. #{tag.a('Read more about the change', href: admin_whats_new_path)}")
+} %>
+
+<% if entries_on_newer_editions.present? %>
+  <div class="app-component-editions__newer-edition-entries">
+    <h3 class="govuk-heading-m">On newer editions</h3>
+
+    <ul class="gem-c-list govuk-list">
+      <% entries_on_newer_editions.each do |entry| %>
+        <%= render_entry(entry) %>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
+
+<% if entries_on_current_edition.present? %>
+  <div class="app-component-editions__current-edition-entries">
+    <h3 class="govuk-heading-m">On this edition</h3>
+
+    <ul class="gem-c-list govuk-list">
+      <% entries_on_current_edition.each do |entry| %>
+        <%= render_entry(entry) %>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
+
+<% if entries_on_previous_editions.present? %>
+  <div class="app-component-editions__previous-edition-entries">
+    <h3 class="govuk-heading-m">On previous editions</h3>
+
+    <ul class="gem-c-list govuk-list">
+      <% entries_on_previous_editions.each do |entry| %>
+        <%= render_entry(entry) %>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
--- a/app/controllers/admin/edition_translations_controller.rb
+++ b/app/controllers/admin/edition_translations_controller.rb
@@ -54,8 +54,12 @@ private
   end
 
   def load_translated_models
-    @document_remarks = Document::PaginatedRemarks.new(@edition.document, params[:remarks_page])
-    @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
+    if preview_design_system?(next_release: false)
+      @document_history = Document::PaginatedTimeline.new(document: @edition.document, page: params[:page] || 1)
+    else
+      @document_remarks = Document::PaginatedRemarks.new(@edition.document, params[:remarks_page])
+      @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
+    end
     @translated_edition = LocalisedModel.new(@edition, translation_locale.code)
   end
 
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -208,8 +208,12 @@ private
   end
 
   def fetch_version_and_remark_trails
-    @document_remarks = Document::PaginatedRemarks.new(@edition.document, params[:remarks_page])
-    @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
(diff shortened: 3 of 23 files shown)


test/unit/models/document/paginated_timeline_test.rb
```diff
@@ -72,8 +72,7 @@ class PaginatedTimelineTest < ActiveSupport::TestCase
                           submitted
                           updated
                           rejected
-                          submitted
-                          created]
```
ollietreend, line 76: One of the implicit responsibilities of this test is to prove it correctly differentiates between `"created"` and `"editioned"` actions.

The difference between the two is whether or not the edition is the first on a document – i.e. `@is_first_edition`:

https://github.com/alphagov/whitehall/blob/ea2b5baec0a081d59b4d0b6b5ab463311edd9f49/app/models/document/paginated_history.rb#L54-L57

That gets populated by `PaginatedTimeline#get_versions_hash`:

https://github.com/alphagov/whitehall/blob/ea2b5baec0a081d59b4d0b6b5ab463311edd9f49/app/models/document/paginated_timeline.rb#L145-L147

But by reducing `Document::PaginatedTimeline::PER_PAGE` we lose the initial `"created"` action from this test, so the test becomes less useful.

Instead, could we mock the `PER_PAGE` value in this test so we still get the full list of actions?

We should be able to use the `mock_pagination` helper – something like:

```ruby
mock_pagination(per_page: 30) do
  timeline = Document::PaginatedTimeline.new(document: @document, page: 1)
  entries = timeline.entries.select { |e| e.instance_of?(Document::PaginatedHistory::AuditTrailEntry) }
  expected_actions = %w[updated
                        editioned
                        published
                        submitted
                        updated
                        rejected
                        submitted
                        created]
end
```

test/components/admin/editions/editorial_remark_component_test.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require "test_helper"
+
+class Admin::Editions::EditorialRemarkComponentTest < ViewComponent::TestCase
+  test "it constructs output based on the editorial remark when an author is present" do
+    editorial_remark = build_stubbed(:editorial_remark, created_at: Time.zone.local(2020, 1, 1, 11, 11))
+    author = editorial_remark.author
+    render_inline(Admin::Editions::EditorialRemarkComponent.new(editorial_remark:))
+
+    assert_equal page.text, "\n  #{editorial_remark.body}\n  #{author.name}  1 January 2020 11:11am\n\n"
```
ollietreend, line 11: **Opinion:** you'd be able to make this a bit less pernickety by stripping whitespace:

```suggestion
    assert_equal page.text.strip, "#{editorial_remark.body}\n  #{author.name}  1 January 2020 11:11am"
```

Although it doesn't solve the internal whitespace within the string. So may not be all that useful 🤷🏻

app/controllers/admin/editions_controller.rb
```diff
@@ -207,9 +207,13 @@ def get_layout
     end
   end
 
-  def fetch_version_and_remark_trails
-    @document_remarks = Document::PaginatedRemarks.new(@edition.document, params[:remarks_page])
-    @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
+  def fetch_version_and_remark_trails(preview_design_system: false)
+    if preview_design_system
```
ollietreend, line 211: Does `preview_design_system` need to be defined as a parameter on this method?

Couldn't the method call `preview_design_system?(next_release: false)` directly itself?

```suggestion
  def fetch_version_and_remark_trails
    if preview_design_system?(next_release: false)
```

app/controllers/admin/edition_translations_controller.rb
```diff
@@ -54,8 +54,12 @@ def translated_item_name
   end
 
   def load_translated_models
-    @document_remarks = Document::PaginatedRemarks.new(@edition.document, params[:remarks_page])
-    @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
+    if preview_design_system?(next_release: false)
+      @document_history = Document::PaginatedTimeline.new(document: @edition.document, page: params[:page] || 1)
```
ollietreend, line 58: **Minor naming opinion:**

Would it make more sense to call this `@document_timeline`?

app/components/admin/editions/document_history_tab_component.html.erb
```diff
@@ -0,0 +1,47 @@
+<h2 class="govuk-heading-l">History</h2>
+
+<% if editing %>
+  <p class="govuk-body">To add an internal note, save your changes.</p>
+<% end %>
+
+<p class="govuk-body"><%= link_to("Add internal note", new_admin_edition_editorial_remark_path(edition), class: "govuk-body govuk-link") %></p>
```
ollietreend, line 7: Is the "Add internal note" button supposed to be hidden if editing?

app/controllers/admin/edition_translations_controller.rb
ollietreend: **Counter argument:**

It gets rendered inside the "History" tab, by the "Document History Tab" component.

So I guess `@document_history` fits in with that naming convention!

app/views/admin/edition_translations/edit.html.erb
```diff
@@ -78,7 +78,12 @@
         {
           id: "history_tab",
           label: "History",
-          content: render(Admin::Editions::DocumentHistoryTabComponent.new(edition: @edition, document_history: @document_history, editing: true))
+          content: render(Admin::Editions::DocumentHistoryTabComponent.new(
+            edition: @edition,
+            document_history: @document_history,
+            editing: true,
+            pagination: paginate(@document_history, theme: 'history', param_name: 'history_page'),
```
ollietreend, line 85: Is there a reason why you chose to pass the "rendered" pagination into the component?

This looks like the sort of thing we'd be able to do from within the component itself by using the `paginate` helper directly.

app/controllers/admin/editions_controller.rb
```diff
@@ -207,9 +207,13 @@ def get_layout
...
+      @document_history = Document::PaginatedTimeline.new(document: @edition.document, page: params[:page] || 1)
```
ollietreend, line 212: Is this param name correct?

```
page: params[:page]
```

In the views, we pass `param_name: 'history_page'` to the helper method:

https://github.com/alphagov/whitehall/blob/ea2b5baec0a081d59b4d0b6b5ab463311edd9f49/app/views/admin/editions/show/_sidebar.html.erb#L21

Although I might be misinterpreting what that helper param does.

ollietreend (commented):
Really nice use of View Components 🎉 it's lovely to see them in action.

I've added a few comments, but there's nothing huge – mostly just small things and/or opinions. Otherwise this is looking good.

test/unit/models/document/paginated_timeline_test.rb
davidgisbey: Cool thanks will sort

app/components/admin/editions/document_history_tab_component.html.erb
davidgisbey: It should show. The copies just a prompt to save as it takes you off the page. Maybe the copy should be.

```
Before adding an internal note, save your changes.
```

test/components/admin/editions/editorial_remark_component_test.rb
davidgisbey: Thanks. I still think it's better. I'll update it now

app/views/admin/edition_translations/edit.html.erb
davidgisbey: As discussed calling it within the view component causes a bunch of issues. We're going pass it in and feature test it.

ollietreend (approved):

DilwoarH:
@davidgisbey - Hey, moved the stylesheets and updated the namespace in this PR #7170

PR #7154, opened 2022-12-14
Add the basics for models knowing their own routes

brucebolt:
We have started to remove the routes for content no longer rendered by Whitehall ([example](https://github.com/alphagov/whitehall/pull/6902)).  However in the comments on [this PR](https://github.com/alphagov/whitehall/pull/7065), we decided this could result in us creating _a lot_ of new methods which are equally complicated.

We decided to do a spike into whether we can do the following:
- having methods on each model for the public and preview URLs
- remove `UrlMaker`, `PublicDocumentRoutesHelper`, `LocalisedUrlPathHelper` and `Admin::EditionRoutesHelper`

This spike has resulted in this PR, which allows a number of different document types to know their own route.  This will form the basis of gradually migrating across all models to know their own routes, and the removal of `UrlMaker`.

This will ultimately make Whitehall's routes file easier to work with by only including the routes it actually serves.

[Trello card](https://trello.com/c/8R2p8CGd)

### DIFF
--- a/app/helpers/public_document_routes_helper.rb
+++ b/app/helpers/public_document_routes_helper.rb
@@ -109,23 +109,4 @@ private
       self
     end
   end
-
-  def append_url_options(path, options = {})
-    if options[:format] && options[:locale]
-      path = "#{path}.#{options[:locale]}.#{options[:format]}"
-    elsif options[:locale] && options[:locale] != I18n.default_locale
-      path = "#{path}.#{options[:locale]}"
-    elsif options[:format]
-      path = "#{path}.#{options[:format]}"
-    end
-
-    if options[:cachebust]
-      query_params = {
-        cachebust: options[:cachebust],
-      }
-      path = "#{path}?#{query_params.to_query}"
-    end
-
-    path
-  end
 end
--- a/app/helpers/public_document_routes_helper.rb
+++ b/app/helpers/public_document_routes_helper.rb
@@ -2,7 +2,8 @@ module PublicDocumentRoutesHelper
   include ActionDispatch::Routing::PolymorphicRoutes
 
   def document_path(edition, options = {})
-    document_url(edition, options.merge(only_path: true))
+    options = locale_options(edition, options)
+    edition.public_path(options)
   end
 
   def public_document_path(edition, options = {})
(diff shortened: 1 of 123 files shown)


app/models/application_record.rb
```diff
@@ -1,3 +1,22 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  def append_url_options(path, options = {})
+    if options[:format] && options[:locale]
+      path = "#{path}.#{options[:locale]}.#{options[:format]}"
+    elsif options[:locale] && options[:locale] != I18n.default_locale
```
JonathanHallam, line 7: am I being dim or are `options[:locale]` and `options[:locale]` not the same? what's going on with this line?

app/presenters/publishing_api/payload_builder/polymorphic_path.rb
```diff
@@ -21,7 +21,11 @@ def call
     private
 
       def base_path
-        @base_path ||= Whitehall.url_maker.polymorphic_path(item)
+        @base_path ||= if item.respond_to?(:public_path)
+                         item.public_path
```
JonathanHallam, line 25: Am assuming that this means polymorphic paths can't be translated?

test/unit/application_record_test.rb
```diff
@@ -0,0 +1,23 @@
+require "test_helper"
+
+class ApplicationRecordTest < ActiveSupport::TestCase
+  test "append_url_options adds locale" do
+    assert_equal "/government/foo.cy", Edition.new.append_url_options("/government/foo", locale: "cy")
```
JonathanHallam, line 5: at what point will we be able to remove this file? Just thinking out loud given we're adding functionality to append locales

app/models/corporate_information_page.rb
```diff
@@ -174,6 +174,16 @@ def alternative_format_provider_required?
     attachments.any? { |a| a.is_a?(FileAttachment) }
   end
 
+  def base_path
+    if worldwide_organisation.present?
```
JonathanHallam, line 178: well this is grim isn't it. Really hoping we can just axe this need.

app/models/publication.rb
```diff
@@ -157,6 +157,12 @@ def assign_statistics_announcement
     end
   end
 
+  def base_path
+    return "/government/statistics/#{slug}" if statistics? || national_statistic?
```
JonathanHallam, line 161: why is this here rather than on the statistics model? I'm sure there's a horrible whitehall reason that statistics and statistics_announcements aren't the same thing but just interested :)

app/models/document_collection.rb
```diff
@@ -25,7 +25,7 @@ def search_index
   end
 
   def search_link
-    Whitehall.url_maker.public_document_path(self)
+    base_path
```
JonathanHallam, line 28: nice!

app/helpers/public_document_routes_helper.rb
```diff
@@ -82,23 +82,6 @@ def get_involved_url(options = {})
     Plek.website_root + get_involved_path(options)
   end
 
-  def take_part_page_path(object, options = {})
```
JonathanHallam, line 98: Really nice, I know this won't show up on the code removal metrics but this kind of simplification is great

JonathanHallam (commented):
Have asked a few questions in line but this is looking super tidy. Thank you!!

app/models/application_record.rb
brucebolt: The first part (before the `&&`) checks `:locale` exists as a key within the `options` hash, whilst the second part compares it's value.

The default value would be `nil`, so I can probably remove the first part.

app/models/edition.rb
```diff
@@ -688,6 +688,25 @@ def attribute_names
     @attributes.keys
   end
 
+  def base_path
+    url_slug = slug || id.to_param
+    "/government/generic-editions/#{url_slug}"
+  end
+
+  def public_path(options = {})
```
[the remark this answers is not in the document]
brucebolt: Have updated the commit message.

app/models/publication.rb
brucebolt: There's no model for statistics or national statistics (statistics announcements and statistical data sets are different things).

test/unit/application_record_test.rb
brucebolt: TBH, I think this will probably stay here.  It's a method that's inherited by nearly all models, so the `ApplicationRecord` seemed the best place to include this.

PR #7155, opened 2022-12-14
Update CI GHA workflow to use reusable components

theseanything:
This updates the CI workflow to use the individual reusable workflows from govuk-infrastructure, instead of using inputs to skip steps. This allows workflows to be less noisey with steps that aren't relevant for a particular app. Also allows more flexibility in configuration, as extra jobs can easily be added directly without modifying a shared workflow.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,13 +8,38 @@ on:
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


theseanything:
closed in favour of https://github.com/alphagov/whitehall/pull/7164

PR #7161, opened 2022-12-16
Fix a JS error by always rendering the foreign language only field

dnkrj:

## What
Always render the foreign language only field.
Toggles CSS classes to hide and show the field.

## Why
Fixes a JS error that would occur when the page was rendered without one.

```js
Uncaught TypeError: Cannot read properties of
null (reading 'querySelector')
    at t.setupworldNewsStoryVisibilityToggle
    at t. init
    at Object.start
    at HTMLDocument.<anonymous>
```

### DIFF
--- a/app/views/admin/editions/_locale_fields.html.erb
+++ b/app/views/admin/editions/_locale_fields.html.erb
@@ -1,39 +1,37 @@
-<% if edition.locale_can_be_changed? %>
-  <div class="edition-form--locale-fields">
-    <%= render "govuk_publishing_components/components/checkboxes", {
-      name: "edition[create_foreign_language_only]",
-      id: "edition_create_foreign_language_only",
-      heading: "Foreign language only #{edition.model_name.human.downcase}",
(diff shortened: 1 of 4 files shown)


davidgisbey (approved):
lgtm

app/views/admin/editions/_locale_fields.html.erb
```diff
@@ -1,39 +1,37 @@
...
-                text: language,
-                value: value,
-                selected: edition.primary_locale == value
-              }
-            end
-          })
-        }
-      ]
-    } %>
+<div class="app-view-edit-edition__locale-field <%= "app-view-edit-edition__locale-field--hidden" unless edition.locale_can_be_changed? && edition.world_news_story? %>">
```
davidgisbey, line 1: sorry i missed this before. This field should render for Consultations and NewsArticles so this could be updated to 

```
edition.locale_can_be_changed? && (edition.is_a?(Consultation) || edition.world_news_story?)
```

PR #7162, opened 2022-12-16
Move sidebar actions into a ViewComponent

dnkrj:

## What
Move sidebar actions into a ViewComponent as part of the process to [move the edition summary page to ViewComponents](https://trello.com/c/5vwim1WH/985-refactor-edition-summary-page-using-view-components-where-appropriate).
Add tests for the ViewComponent.

## Why
There was a logic in the sidebar actions template. By moving it into a ViewComponent that code can be better organised and tested.


test/components/admin/editions/show/sidebar_actions_component_test.rb
```diff
@@ -0,0 +1,197 @@
+# frozen_string_literal: true
```
DilwoarH, line 1: Is this intended?

DilwoarH:
Looks great, one small comment above.

test/components/admin/editions/show/sidebar_actions_component_test.rb
dnkrj: It's a ruby "magic comment" that's present in all of the ViewComponent files to improve memory performance by only defining a single instance of each unique string.

app/components/admin/editions/show/sidebar_actions_component.rb
```diff
@@ -0,0 +1,177 @@
...
+          class: "govuk-link",
+        )
+      end
+    end
+  end
+
+  def add_data_action
+    if @edition.publicly_visible?
+      actions << link_to("View on website (opens in new tab)", @url_maker.public_document_url(@edition), class: "govuk-link", target: "_blank", rel: "noopener")
+      actions << link_to("View data about page", helpers.content_data_page_data_url(@edition), class: "govuk-link", data: {
```
davidgisbey, line 169: This is super super nitty so feel free to ignore, but i wonder where this should be two methods as it actually adds two actions. 
```
def add_view_action; end
def add_data_action; end
```

Or pluralise it 

```
def add_data_actions; end
```

It's a tiny thing though. I only noticed as i was playing with generating the objects returned by methods within the array rather than being piped in and had to pop them in an array and flatten it.

davidgisbey (approved):
I've left a couple of small comments, but nothing blocking. 

In terms of style, there might be a few things that i'd do differently like assigning some of the instance variables in the initialisation to private methods and some instance variables to private attr readers, but really it's just a matter of preference.

You'll definitely see it done both ways throughout the codebase.

I found this really easy to follow and the testing was super clear ✅

app/components/admin/editions/show/sidebar_actions_component.rb
dnkrj: Thanks, I've added another method

app/components/admin/editions/show/sidebar_actions_component.rb
```diff
@@ -0,0 +1,177 @@
...
+    @scheduler = Whitehall.edition_services.scheduler(edition)
+    @force_scheduler = Whitehall.edition_services.force_scheduler(edition)
+    @publisher = Whitehall.edition_services.publisher(edition)
+    @force_publisher = Whitehall.edition_services.force_publisher(edition)
+
+    @edition = edition
+  end
+
+  def render?
+    actions.any?
```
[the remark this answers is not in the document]
dnkrj: It happens for superseded documents, there was a test for them but was asserting a `li` count of 0, I've updated the test to instead assert that the container is not rendered

PR #7164, opened 2022-12-16
Update CI GHA workflow to use reusable components

theseanything:
This updates the CI workflow to use the individual reusable workflows from govuk-infrastructure, instead of using inputs to skip steps. This allows workflows to be less noisey with steps that aren't relevant for a particular app. Also allows more flexiblity in configuration, as extra jobs can easily be added directly without modifying a shared workflow.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,13 +8,40 @@ on:
   pull_request:
 
 jobs:
-  test:
-    name: Test Rails
-    uses: alphagov/govuk-infrastructure/.github/workflows/test-rails.yaml@main
+  cache-dependencies:
+    name: Cache dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-cache-dependencies.yaml@main
+
+  security-analysis-ruby:
+    name: Security Analysis Ruby
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-security-analysis-ruby.yaml@main
+
+  lint-scss:
+    name: Lint SCSS
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-lint-scss.yaml@main
+
+  lint-javascript:
+    name: Lint JavaScript
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-lint-javascript.yaml@main
+
+  lint-ruby:
+    name: Lint Ruby
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-lint-ruby.yaml@main
+
+  test-javascript:
+    name: Test JavaScript
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-test-javascript.yaml@main
+
+  test-ruby:
+    name: Test Ruby
+    needs: cache-dependencies
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-test-ruby.yaml@main
     with:
-      requiresJavaScript: true
-      requiresRedis: true
-      requiresMySQL: true
-      mysqlUsername: whitehall
-      mysqlPassword: whitehall
+      enableMySQL: true
+      enableRedis: true
       extraSystemDependencies: 'ghostscript'


kevindew:
Hey @theseanything this stuff looks cool. What's the scope of this work? Is this intended to be a full future CI configuration or is this experimenting?

I wondered at what point would be good to feedback on things? I have a few thoughts but didn't know what stage you were at?

theseanything:
> Hey @theseanything this stuff looks cool. What's the scope of this work? Is this intended to be a full future CI configuration or is this experimenting?
> 
> I wondered at what point would be good to feedback on things? I have a few thoughts but didn't know what stage you were at?

Hey @kevindew - currently on working implementing CI in GHA for the publishing apps.  Any feedback you have would be great, so feel free to leave comments!

kevindew:
> > Hey @theseanything this stuff looks cool. What's the scope of this work? Is this intended to be a full future CI configuration or is this experimenting?
> > I wondered at what point would be good to feedback on things? I have a few thoughts but didn't know what stage you were at?
> 
> Hey @kevindew - currently on working implementing CI in GHA for the publishing apps. Any feedback you have would be great, so feel free to leave comments!

Great stuff, ok let's see feedback wise.

So it'd be good to involve people in the teams a bit more than previous, I think people were quite blindsided by the initial introduction of the Whitehall GHA CI, so seeking more input or informing of changes and why could be useful for the code owners.

At a fundamental level, I feel that perhaps the drive to have heavy re-use is potentially adds more complexity than it saves. GHA files are generally low on the amount of config needed to achieve an action and the re-use patterns are rather rigid with low flexibility. I do somewhat wonder if life would be simpler starting from a place without re-use (as per: https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#gov-uk-rails-application-with-postgres-redis-yarn-and-gov-uk-content-schemas-dependencies) and then see where the pain points appear to embrace re-use. Not that I don't see any advantages to re-use but do worry it'll cause my headaches than it solves.

---

On the embracing re-use front, if we do go down that path, I feel like it could be better to have one generic action that can run an app task rather than lots of specific ones. For example, the test ruby one, made a bit more generic could be used for a few scenarios with a bit of tweaking:

```
  app-js-test:
    name: Javascript Tests
    needs: cache-dependencies
    uses: alphagov/govuk-infrastructure/.github/workflows/ci-govuk-ruby-app-task.yaml@main
    with:
      task: bundle exec jasmine:ci

  app-test:
    name: Minitest
    needs: cache-dependencies
    uses: alphagov/govuk-infrastructure/.github/workflows/ci-govuk-ruby-app-task.yaml@main
    with:
      enableMySQL: true
      enableRedis: true
      setup: bundle exec rails db:reset
      task: bundle exec test

  app-cucumber:
    name: Cucumber
    needs: cache-dependencies
    uses: alphagov/govuk-infrastructure/.github/workflows/ci-govuk-ruby-app-task.yaml@main
    with:
      enableMySQL: true
      enableRedis: true
      setup: bundle exec rails db:reset
      task: bundle exec cucumber
```

and could imagine you could use that to also cover most of the other tasks with that one task (it'd also mean we could have the [pact-verify](https://github.com/alphagov/whitehall/blob/main/.github/workflows/pact-verify.yml) action use it too.

--- 

It'd be good to have a think about how things with fit with the [govuk-saas-config](https://github.com/alphagov/govuk-saas-config/blob/a28ef11f9cda891363dad3334eaf625982b4c81a/github/lib/configure_repo.rb#L106) that sets up the required jobs checks. At the moment all of the replatforming GHA test repos files have been exepmpt as they use `ci.yaml` rather than [the convention of `ci.yml`](https://github.com/alphagov/govuk-saas-config/blob/a28ef11f9cda891363dad3334eaf625982b4c81a/github/lib/configure_repo.rb#L135) which I assume was unintentional. I'd guess we'd want a way to specify which jobs are required if we go down the multiple task routes. I'd guess we could do the same thing we do in gems of having a test action that GitHub requires which [requires other jobs to pass](https://github.com/alphagov/govuk_schemas/blob/13931cd6aef331a32bb7a97774049192c9efdc24/.github/workflows/ci.yml#L28-L35) though it's a bit clunky.

---

Another area that I think would be good looked at, especially in the case of Whitehall is utilising a [different test runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners) so we can try keep the performance the same or better than Jenkins with more CPUs. I imagine we'd have some set-up faff getting a runner set-up (I don't have the access to turn one on myself) but I imagine that's where we could really save some time and would be good that a design can utilise them.

---

Then on the more more minor side a few things come to mind:

- We're missing cucumber from this job
- Great to see the env var to set db versions. I'm sure it won't be long until we have some divergence, particularly when we need to update a db
- The filenames of the reusable actions seem rather generic for rather specific things that seems like could be hinderance. For example: ci-test-javascript is very much just something for a Ruby app and wouldn't be usable by a Javascript project of ours, ci-test-ruby seems rather GOV.UK Rails specific (is that a search-api concession) - I think they could be clearer in scope
- It'd be good to test production asset precompiling, as that catches people out: https://github.com/alphagov/whitehall/pull/7150 - we lost it from Jenkins CI as part of an optimisation and seems to have caught people out a bunch since: https://github.com/alphagov/govuk-jenkinslib/commit/a6d8f52b5951c874c07a5e52750d4303f2972014

theseanything:
Thanks Kevin!


> So it'd be good to involve people in the teams a bit more than previous, I think people were quite blindsided by the initial introduction of the Whitehall GHA CI, so seeking more input or informing of changes and why could be useful for the code owners.


Yup. Will do better a job at letting teams know!  

> At a fundamental level, I feel that perhaps the drive to have heavy re-use is potentially adds more complexity than it saves …. 

 Whilst I agree there is a little added complexity in re-use, from the experience we’ve had building out the initial CI workflow for frontend apps and implementing the deployment pipeline, the toil involved updating 45~ individuals repos is significantly more time consuming. The re-use is more valuable when rolling out the initial implementation, where tweaks and updates to shared logic are more common (e.g. backing service configuration, dependency versions, job configuration etc…). Tho this shouldn’t prevent us from switching to individual configuration in future. 


 Thanks for pointing out those dev docs - will add a reminder to make sure we keep them up to date and make sure we’re consistent in what we actually do vs what’s written down. 


> On the embracing re-use front, if we do go down that path, I feel like it could be better to have one generic action that can run an app task rather than lots of specific ones …

 Good suggestion about the generic app task job. Still investigating how varied our app jobs are - currently considering GitHub’s composite actions to capture some of the shared logic like configuring backing services. Could remove the need for re-usable workflows altogether and bring back that type flexibility in configuration.  - https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

 
> It'd be good to have a think about how things with fit with the [govuk-saas-config](https://github.com/alphagov/govuk-saas-config/blob/a28ef11f9cda891363dad3334eaf625982b4c81a/github/lib/configure_repo.rb#L106) that sets up the required jobs checks …

 Thanks for heads up on govuk-saas-config, didn’t know it configured the required checks. For right now, we probably don’t want make these CI jobs (in GHA) required checks - as they’re only used for the EKS platform, and we don’t want to get in the way of the existing development process during implementation. But we’ve got a note make sure they’re eventually required. 


> Another area that I think would be good looked at, especially in the case of Whitehall is utilising a [different test runner](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners) so we can try keep the performance the same or better than Jenkins with more CPUs … 


 Yup - we’ve got a goal to make CI tests in GHA at least as fast as our existing infra and conscious that Whitehall tests are slower right now. Larger test runners is one of the possible solutions, something will come back too after the initial implementation. 

> We're missing cucumber from this job

 Yup - on the todo list. 

> Great to see the env var to set db versions …

 Yup - on the iteration todo list 

> * The filenames of the reusable actions seem rather generic for rather specific things that seems like could be hinderance …

Yup agree - will have a look at better naming

> * It'd be good to test production asset precompiling …


We currently precompile the assets during image build - so at the moment, we have some safety from trying deploy a release where it doesn’t work. But yup on the iteration todo list.

DilwoarH:
This is excellent work, this will help us identify exactly where things are failing.

kevindew:
Thanks Sean.

>  Whilst I agree there is a little added complexity in re-use, from the experience we’ve had building out the initial CI workflow for frontend apps and implementing the deployment pipeline, the toil involved updating 45~ individuals repos is significantly more time consuming. The re-use is more valuable when rolling out the initial implementation, where tweaks and updates to shared logic are more common (e.g. backing service configuration, dependency versions, job configuration etc…). Tho this shouldn’t prevent us from switching to individual configuration in future. 

Sure, it has certainly looked painful when you've had to update all the deploy workflows. On the other hand, with the gems we've had some of the individual workflows for years without much toil - so I imagine it might not be as bad as you fear. I do think this approach you've taken for re-use seems better than the previous iteration as it offers some escape routes of rolling own actions for more bespoke app needs.

The other thing I wonder about is what the plan for ownership of these are? It doesn't seem ideal to me that Replatforming are implementing these (aside from a consistency benefit) as they're not the team that'll have to deal with any challenges/longer term maintenance. Is the plan that replatforming will hand over ownership to teams or would it be considered the teams from the point of merging?
  
>  Good suggestion about the generic app task job. Still investigating how varied our app jobs are - currently considering GitHub’s composite actions to capture some of the shared logic like configuring backing services. Could remove the need for re-usable workflows altogether and bring back that type flexibility in configuration. - https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

Yup, composite actions sound pretty neat. 👍 

It seems such a pain how you can't seem to share services easily as they seem the most verbose part of actions overall.

> 
> > It'd be good to have a think about how things with fit with the [govuk-saas-config](https://github.com/alphagov/govuk-saas-config/blob/a28ef11f9cda891363dad3334eaf625982b4c81a/github/lib/configure_repo.rb#L106) that sets up the required jobs checks …
> 
>  Thanks for heads up on govuk-saas-config, didn’t know it configured the required checks. For right now, we probably don’t want make these CI jobs (in GHA) required checks - as they’re only used for the EKS platform, and we don’t want to get in the way of the existing development process during implementation. But we’ve got a note make sure they’re eventually required. 

Yeah, I assumed you didn't want them required but probably hadn't realised it was only down to a file extension spelling. It's annoying that the world hasn't agreed on yaml or yml. 

> > * It'd be good to test production asset precompiling …
> 
> We currently precompile the assets during image build - so at the moment, we have some safety from trying deploy a release where it doesn’t work. But yup on the iteration todo list.

Yeah that was how @dnkrj found out the failure of the build I referenced - which is kinda the same boat with Jenkins where it fails on deploy. It'd be good to catch them pre-merge, it's a shame we lost that from our Jenkins CI - I guess we used to have it lingering until only recently when we used to build docker images in Jenkins CI for e2e tests.

theseanything:
> The other thing I wonder about is what the plan for ownership of these are? ...

Ownership will probably eventually fall back to teams? (unclear what GOV.UK's long term intention is for who supports infra/dev tooling/processes)  - but at the very least don't think teams should need to feel fully responsible before we migrate to EKS. We absolutely do not want to be building something, then immediately dump the maintenance burden onto teams and give zero support.

For right now, Replatforming needs a CI/CD pipeline as part of building out the infrastructure. We definitely don't want to be obstructing teams' current development processes and existing work, and we have no plans for changing the existing Jenkins CI/CD process. Realistically having Replatforming set up an initial working CI pipeline for all the applications will be far quicker than putting the onus on each team to do (they have their own priorities and might not yet have the knowledge to do so). Having said that, we don't want to exclude teams from this process - should be part of communicating the changes and allowing opportunity to give input. Replatforming do plan on providing support and training for teams as we migrate towards EKS, including the CI/CD processes. 

> It seems such a pain how you can't seem to share services easily as they seem the most verbose part of actions overall.

Yeah absolutely - tho they're essentially a neat way of spinning up a container. Which composite action might be a good replacement to allow reusability. 

> Yeah, I assumed you didn't want them required but probably hadn't realised it was only down to a file extension spelling. It's annoying that the world hasn't agreed on yaml or yml.

Ahah yah, defo a fluke that we weren't caught out by sass config!

kevindew:
Great thanks @theseanything all sounds positive.

With regards to ownership: 

> Ownership will probably eventually fall back to teams? (unclear what GOV.UK's long term intention is for who supports infra/dev tooling/processes)

The direction we're aiming towards is ownership boundaries are defined by repositories (with hopefully minimal cases of multiple teams working in same repository). So how'd that pan out with things like this is that Publishing Experience own the actions in this repo, which may utilise workflows part of by govuk-infrastructure (and therefore owned by Platform Reliabality) - they offer something of a public API to services govuk-infrastructure look after. So teams would fulfil a similar model to an open source project maintainer role.

theseanything:
Closed in favour of https://github.com/alphagov/whitehall/pull/7197

PR #7168, opened 2022-12-19
Add async pagination for the History tab

davidgisbey:
## Description

We've recently added the history tab to the edition, summary, edition edit and edit translation pages.

image

This tab already has pagination built into it, but using it reloads the page.

This adds in dynamic pagination via Ajax requests

## Guidance for review

I've left a couple of comments on things i'm unclear on.

## Trello card 

https://trello.com/c/xiy8xs27/888-move-notes-history-fact-checking-tabs-to-the-design-system

### DIFF
--- a/app/assets/javascripts/admin/modules/document-history-paginator.js
+++ b/app/assets/javascripts/admin/modules/document-history-paginator.js
@@ -1,0 +1,49 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function DocumentHistoryPaginator (module) {
+    this.module = module
+  }
+
+  DocumentHistoryPaginator.prototype.init = function () {
+    var fetch = window.fetch
+    if (!fetch) { return }
+
+    this.setupEventListeners()
+  }
+
+  DocumentHistoryPaginator.prototype.setupEventListeners = function () {
+    var module = this.module
+    var newerLink = module.querySelector('.app-view-document-history-tab__newer-pagination-link')
+    var olderLink = module.querySelector('.app-view-document-history-tab__older-pagination-link')
+
+    if (newerLink) {
+      this.addLinkEventListener(newerLink)
+    }
+
+    if (olderLink) {
+      this.addLinkEventListener(olderLink)
+    }
+  }
+
+  DocumentHistoryPaginator.prototype.addLinkEventListener = function (link) {
+    var module = this.module
+
+    link.addEventListener('click', function (e) {
+      e.preventDefault()
+
+      window.fetch(new URL(document.location.origin + link.dataset.remotePagination))
+        .then(function (response) { return response.text() })
+        .catch(function () { window.location.replace(new URL(link.href)) })
+        .then(function (html) {
+          module.innerHTML = html
+
+          var documentHistoryModule = new GOVUK.Modules.DocumentHistoryPaginator(module)
+          documentHistoryModule.init()
+        })
+    })
+  }
+
+  Modules.DocumentHistoryPaginator = DocumentHistoryPaginator
+})(window.GOVUK.Modules)
--- a/features/admin-history-pagination.feature
+++ b/features/admin-history-pagination.feature
@@ -1,0 +1,30 @@
+Feature: Viewing a documents history
+
+  Background:
+    Given I am a writer
+
+  @design-system-only
+  Scenario: Viewing multiple pages of history
+    Given a draft news article "Stubble to be Outlawed" exists
+    When "Stubble to be Outlawed" has two pages of history
+    Then I can see the first ten items in the History tab
+
+    When I click the "Older" link
+    Then I can see the second page of history
+
+    When I click the "Newer" link
+    Then I can see the first ten items in the History tab
+
+  @javascript @design-system-only
+  Scenario: Viewing multiple pages of history with JavaScript
+    Given a draft news article "Stubble to be Outlawed" exists
+    When "Stubble to be Outlawed" has two pages of history
+    Then I can see the first ten items in the History tab
+
+    When I click the "Older" link
+    Then I can see the second page of history
+    And the History tab is still showing
+
+    When I click the "Newer" link
+    Then I can see the first ten items in the History tab
+    And the History tab is still showing
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,6 +12,7 @@
 //= require admin/analytics
 
 //= require admin/modules/app-analytics
+//= require admin/modules/document-history-paginator
 //= require admin/modules/navbar-toggle
 //= require admin/modules/paste-html-to-govspeak
 
--- a/app/assets/stylesheets/admin/views/_document-history-tab.scss
+++ b/app/assets/stylesheets/admin/views/_document-history-tab.scss
@@ -1,8 +1,8 @@
-.app-view-document-history-tab__pagination-link {
-  margin-left: govuk-spacing(3);
-}
-
-.app-view-document-history-tab__pagination-link:last-child {
+.app-view-document-history-tab__older-pagination-link {
   float: right;
   margin-right: govuk-spacing(3);
+}
+
+.app-view-document-history-tab__newer-pagination-link {
+  margin-left: govuk-spacing(3);
 }
--- a/app/components/admin/editions/document_history_tab_component.html.erb
+++ b/app/components/admin/editions/document_history_tab_component.html.erb
@@ -10,7 +10,7 @@
   text: sanitize("History and notes have been merged. #{tag.a('Read more about the change', href: admin_whats_new_path)}")
 } %>
 
-<%= paginate(@document_history, theme: 'history') %>
+<%= paginate(@document_history, theme: 'history', editing:) %>
 
 <% if entries_on_newer_editions.present? %>
   <div class="app-view-editions__newer-edition-entries">
--- a/app/controllers/admin/edition_audit_trail_controller.rb
+++ b/app/controllers/admin/edition_audit_trail_controller.rb
@@ -3,6 +3,12 @@ class Admin::EditionAuditTrailController < Admin::EditionsController
 
   def index
     @edition = Edition.find(params[:id])
-    @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
+    if preview_design_system?(next_release: false)
+      @document_history = Document::PaginatedTimeline.new(document: @edition.document, page: params[:page] || 1)
+
+      render(html: Admin::Editions::DocumentHistoryTabComponent.new(edition: @edition, document_history: @document_history, editing: params[:editing]).render_in(view_context))
+    else
+      @document_history = Document::PaginatedHistory.new(@edition.document, params[:page])
+    end
   end
 end
--- a/app/helpers/admin/audit_trail_helper.rb
+++ b/app/helpers/admin/audit_trail_helper.rb
@@ -18,10 +18,10 @@ module Admin::AuditTrailHelper
     html << absolute_time(entry.created_at, class: "created_at")
   end
 
-  def paginated_audit_trail_url(page)
+  def paginated_audit_trail_url(page, editing = nil)
     url_for(
       params.to_unsafe_hash
-            .merge(controller: "admin/edition_audit_trail", action: "index", page: (page <= 1 ? nil : page))
+            .merge(controller: "admin/edition_audit_trail", action: "index", page: (page <= 1 ? nil : page), editing:)
             .symbolize_keys,
     )
   end
--- a/app/views/admin/edition_translations/edit.html.erb
+++ b/app/views/admin/edition_translations/edit.html.erb
@@ -78,11 +78,14 @@
         {
           id: "history_tab",
           label: "History",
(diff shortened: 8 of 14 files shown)


app/assets/javascripts/admin/modules/document-history.js
```diff
@@ -0,0 +1,45 @@
...
+    var documentHistoryTab = document.querySelector('#history_tab')
+
+    document.addEventListener('click', function (e) {
+      var newerLink = document.querySelector('.app-component-document-history-tab__newer-pagination-link')
+      var olderLink = document.querySelector('.app-component-document-history-tab__older-pagination-link')
+
+      if (newerLink && newerLink.contains(event.target)) {
+        e.preventDefault()
+        // eslint-disable-next-line
+        Rails.ajax({
```
davidgisbey, line 23: The linter here complains that

```
standardx: Use JavaScript Standard Style (tweaked by standardx) (https://github.com/standard/standardx)
  /govuk/whitehall/app/assets/javascripts/admin/modules/document-history.js:22:9: 'Rails' is not defined.
  /govuk/whitehall/app/assets/javascripts/admin/modules/document-history.js:31:9: 'Rails' is not defined.
```
My understanding from googling is that when you use sprockets all that is required here is to add 

`//= require rails-ujs` to the `application.js` file.

any thoughts?

features/admin-history-pagination.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Viewing a documents history
```
davidgisbey, line 1: I've tested the functionality via feature specs since it's a pretty e2e bit of functionality. Do people think it's enough coverage?

dnkrj: Seems sensible to me

dnkrj (commented):
Generally looks good to me, in a more modern web app we'd probably have a library that would mean that `innerHTML` isn't called directly but this seems to be the preferred method with Ruby. Will leave it to Dilwoar to approve just in case he has more knowledge there.

app/assets/javascripts/admin/modules/document-history.js
```diff
@@ -0,0 +1,45 @@
...
+  function DocumentHistory (module) {
+    this.module = module
+  }
+
+  DocumentHistory.prototype.init = function () {
+    this.setupEventListeners()
+  }
+
+  DocumentHistory.prototype.setupEventListeners = function () {
+    var documentHistoryTab = document.querySelector('#history_tab')
```
DilwoarH, line 14:
```suggestion
    var documentHistoryTab = this.module.querySelector('#history_tab')
```

app/assets/javascripts/admin/modules/document-history.js
```diff
@@ -0,0 +1,45 @@
...
+
+    document.addEventListener('click', function (e) {
```
DilwoarH, line 16: This is dangerous as you're listening to click events on the whole document. I think what you need is click listeners on the individual links.

app/assets/javascripts/admin/modules/document-history.js
DilwoarH, line 23: I would avoid using rails ajax and stick to using native `fetch` or  something like [this](https://github.com/alphagov/whitehall/blob/main/app/assets/javascripts/components/govspeak-editor.js#L21)

DilwoarH (dismissed):

app/assets/javascripts/admin/modules/document-history.js
davidgisbey: The problem i found is if i scope it to the links is you end up in a bit of an infinite loop. The older and newer links get re-rendered when the component is passed back via html. You have to readd them, but you run into the same issue next time you click one of the links.

There's probs a simple way round this, but i'm not overly sure what the right thing to do is. 

Maybe just add it to the history tab? Or i guess just grab the html after this 

image

and render that?

ollietreend: I think you should be fine to add the click handler to a parent DOM element, without needing to add it to the entire document.

For example, I'm assuming the element at `this.module` is a parent of `#history_tab`. So while we replace the contents of `#history_tab`, the parent `this.module` element persists so will keep its event listeners.

```suggestion
    this.module.addEventListener('click', function (e) {
```

app/assets/javascripts/admin/modules/document-history.js
davidgisbey: So unfortunately the module actually only returns the link to the tab not the tab itself which are both children of the govuk-tabs div.

I can do this to avoid using the document if it's better?   

` var documentHistoryTab = this.module.closest('.govuk-tabs').querySelector('#history_tab')` 

wdyt?

app/assets/javascripts/admin/modules/document-history.js
davidgisbey: Thanks! Have sorted

app/assets/javascripts/admin/modules/document-history.js
```diff
@@ -0,0 +1,38 @@
...
+    documentHistoryTab.addEventListener('click', function (e) {
+      var newerLink = documentHistoryTab.querySelector('.app-view-document-history-tab__newer-pagination-link')
+      var olderLink = documentHistoryTab.querySelector('.app-view-document-history-tab__older-pagination-link')
+
+      if (newerLink && newerLink.contains(event.target)) {
+        e.preventDefault()
+
+        window.fetch(new URL(document.location.origin + newerLink.dataset.remotePagination))
+        .then(function (response) { return response.text()})
+        .then((html) => { documentHistoryTab.innerHTML = html })
```
kevindew, line 25: Looks like you've got ES6 JS in this, which won't work in IE and I expect will fail uglify compliation.

app/assets/javascripts/admin/modules/document-history.js
```diff
@@ -0,0 +1,38 @@
...
+    var documentHistoryTab = this.module.closest('.govuk-tabs').querySelector('#history_tab')
+
+    documentHistoryTab.addEventListener('click', function (e) {
+      var newerLink = documentHistoryTab.querySelector('.app-view-document-history-tab__newer-pagination-link')
+      var olderLink = documentHistoryTab.querySelector('.app-view-document-history-tab__older-pagination-link')
+
+      if (newerLink && newerLink.contains(event.target)) {
+        e.preventDefault()
+
+        window.fetch(new URL(document.location.origin + newerLink.dataset.remotePagination))
```
kevindew, line 23: Do we have a polyfill for fetch? 

Do we know what happens when this promise fails?

app/assets/javascripts/admin/modules/document-history.js
kevindew: Sorry for the drive by comment, saw it pop up in my email.

I imagine that if the elements you are interacting with are outside the module you probably want to move the module to contain it. But I guess you don't need that as I think you're just listening for clicks on elements within the tab? so you could probably event listeners on the newer/older links

davidgisbey: with the current implementation the whole tab gets replaced including the links. So adding them to the parent elements seems like the easiest solution.

We could go down the route of only adding in the section below where the links are rendered by munging the response, but this seemed simpler. Wdyt?

kevindew: Ah sure, that makes sense. I think in that case you do have a bit of an isolation issue as the replacement also replaces the element that generated the module - as that'd no longer be in the DOM after a page change. This isn't a cardinal sin but does feel like a bit of a smell as there isn't encapsulation

I think you'd do best to create an element inside the history which is a wrapper for this i.e the tab contents contain:

```
<div data-module="edition-history-paginator">
  <%= history_with_pagination %>
</div>
```

Then that div can be the module and can have it's contents replaced while still persisting with each pagination and only interacting with elements inside of it.

app/assets/javascripts/admin/modules/document-history.js
ollietreend: @kevindew with regards to IE11 support, [the GDS service manual says](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2022):

> You will still need to support Internet Explorer 11 if your analytics data shows at least 2% of your users arriving at the service’s start page are using it.

According to Analytics data for Whitehall, IE11 usage [dropped off a cliff](https://analytics.google.com/analytics/web/#/report/visitors-browser/a26179049w83104518p86186851/_u.date00=20220101&_u.date01=20221231&explorer-table.plotKeys=%5B%5D&_r.drilldown=analytics.browser:Internet%20Explorer&_.gSectionId=summary&explorer-graphOptions.selected=analytics.nthMonth) throughout the course of last year:

Number of IE11 users

_(admittedly it was a small cliff to begin with)_

As of December 2022, **0.26%** of Whitehall users were on IE – so well below the 2% threshold required of public facing services.

It's hard to know when to draw a line, but I'm wondering whether it's now safe to assume that the user's browser supports `window.fetch` without requiring a polyfill? It's been [supported by](https://caniuse.com/fetch) evergreen browsers for the past 8 years now.

kevindew: @ollietreend I think it's fine for us to drop IE11 support in a considered way but I don't think we should just do it casually without some involvement from wider GOV.UK frontenders (which I believe @dnkrj is exploring at the moment). 

The complicating factors seem to be:

a) deciding what not supporting IE means, there's a decision as to whether it's a "we don't care if it's broken" or if it's a "we purposefully degrade it to fall back to plain HTML" so it can still kinda work with progressive enhancement (there's more on this in [design system issue 2607](https://github.com/alphagov/govuk-frontend/discussions/2607) which talks about functional/compliant browser definitions)
b) we're still part of a GOV.UK that does support IE11 - naturally the direction of travel is to end support, so it's ok to lead that way, but shouldn't just diverge without coordination and communication.

kevindew: Thinking pragmatically, if we’re happy this is a progressive enhancement, then a short term approach could be to use object detection as a way to use fetch without opening up errors for IE

I.e. we could have the start of the module exit early if window.fetch is falsey. It’d then just fall back to no-js behaviour.

davidgisbey: As discussed when the promise fails it falls back to the non-js behaviour and reloads the page with the next or previous set of history.

app/assets/javascripts/admin/modules/document-history-paginator.js
```diff
@@ -0,0 +1,55 @@
...
+
+    this.setupEventListeners()
+  }
+
+  DocumentHistoryPaginator.prototype.setupEventListeners = function () {
+    var module = this.module
+    var newerLink = module.querySelector('.app-view-document-history-tab__newer-pagination-link')
+    var olderLink = module.querySelector('.app-view-document-history-tab__older-pagination-link')
+
+    if (newerLink) {
```
dnkrj, line 21: Since this is really the same code twice you could define a setupLink function and the call it with `newerLink` and `olderLink`

dnkrj (approved):
Looks great! Just one suggestion about you initialisation

app/assets/javascripts/admin/modules/document-history-paginator.js
davidgisbey: Thanks for the review. Great shout. I'll sort it before merge

app/assets/javascripts/admin/modules/document-history-paginator.js
```diff
@@ -0,0 +1,49 @@
...
+
+  DocumentHistoryPaginator.prototype.addLinkEventListener = function (link) {
+    var module = this.module
+
+    link.addEventListener('click', function (e) {
+      e.preventDefault()
+
+      window.fetch(new URL(document.location.origin + link.dataset.remotePagination))
+        .then(function (response) { return response.text() })
+        .catch(function () { window.location.replace(new URL(link.href)) })
```
ollietreend, line 38: I'm not sure if `window.location.replace()` is the method you want to be using here.

The [W3Schools definition](https://www.w3schools.com/jsref/met_loc_replace.asp) explains:

> `replace()` removes the current URL from the document history.
>
> With `replace()` it is not possible to use "back" to navigate back to the original document.

I think the expected behaviour is that you'd still be able to use the "back" button to go back a page – i.e. the same as if JavaScript was disabled. In this case, you probably want to set `window.location.href` instead:

```js
window.location.href = new URL(link.href)
```

And since you're not manipulating the URL itself, just passing it through, you'd be fine to just pass it through as a String. (I actually didn't know about [URL objects](https://javascript.info/url) before seeing this PR! 😅)

```suggestion
        .catch(function () { window.location.href = link.href })
```

app/assets/javascripts/admin/modules/document-history-paginator.js
```diff
@@ -0,0 +1,49 @@
...
+        .then(function (html) {
+          module.innerHTML = html
+
+          var documentHistoryModule = new GOVUK.Modules.DocumentHistoryPaginator(module)
+          documentHistoryModule.init()
```
kevindew, line 43: Assuming you're going to be making an amend anyway as per @ollietreend's comment above.

We don't need to create a new module here as this one still exists (the module element is still in the dom), so rather than doing this we should re-initialise the event listeners by calling `this.setupEventListeners()`

Issue #6954, opened 2022-10-24
Add View Components


davidgisbey:
## Description of issue

Whitehall has large complex views. The established pattern is to break down the pages into many partials. Each of these partials have different functions and vary in levels of complexity. We then test these views through a combination of feature specs and view controller specs.

These specs are slow when it comes to performance. They're also often somewhat laborious to write as you are unable to isolate the section of the view that you are working on and subsequently interested in testing. As many views are large and complex, you often have to handle interactions with other applications. For example you may well have to stub out some calls to Publishing API despite only being interested in the fact that when you add some fields to the page, they render and have the correct values.

Often there is quite a lot of logic that sits within partials in the form of conditionals, or is pulled in from the controller or helpers.

An example of this can be seen in the way we construct images for editions 

https://github.com/alphagov/whitehall/blob/ecb22a6a76ccbfbd83de00b65bd8e6cccdf8c39e/app/views/admin/editions/_image_template.html.erb

Within this template we conditionally render 5 different inputs or pieces of content based on the object's state. 

<img width="516" alt="image" src="https://user-images.githubusercontent.com/42515961/197198186-22148ea8-e62c-4a2b-b640-380d7ea7e39b.png">

<img width="820" alt="image" src="https://user-images.githubusercontent.com/42515961/197198292-92f15ae4-0a26-47d2-9c39-8cdc3f8d1ff8.png">

<img width="603" alt="image" src="https://user-images.githubusercontent.com/42515961/197198217-ceb1f100-5916-465c-8942-0dcf10a52963.png">

<img width="708" alt="image" src="https://user-images.githubusercontent.com/42515961/197198324-f131cbcb-6e33-4f59-b823-096e899ee6cf.png">

<img width="590" alt="image" src="https://user-images.githubusercontent.com/42515961/197198358-8a78e462-7cb1-4ee0-ba08-0c1571999a5b.png">

We also often see the use of variables within views And many locals being passed into partials.

<img width="677" alt="image" src="https://user-images.githubusercontent.com/42515961/197200615-754e2d0e-08e4-4002-a3db-f1b4047a4ee5.png">

In order to properly test all the combinations of conditionals and locals passed in, we're required to write an exhaustive amount of controller view specs, which often are essentially functioning as unit tests, but with none of the performance.

In reality, a lot of logic in the views is not properly tested.

[...]

PR #7170, opened 2022-12-20
Add CSS docs and fix inconsistent file structures

DilwoarH:
## What

Add CSS docs for our new standards and fix inconsistent file structures.

## Why

We only have documentation for older legacy CSS but nothing for the new standards. This PR tries to address this and also fix any inconsistency that were merged before the standards were documented.


davidgisbey (approved):
Thanks for that. I'll follow the guide going forward 👍

docs/css.md
```diff
@@ -1,115 +1,125 @@
 # Whitehall CSS Styleguide
 
-To be read in conjunction with the [GDS Way CSS styleguide](https://gds-way.cloudapps.digital/manuals/programming-languages/css.html)
+To be read in conjunction with the following:
 
-## File structure and scope
+- [GDS Way CSS styleguide](https://gds-way.cloudapps.digital/manuals/programming-languages/css.html)
+- [GOVUK Component Conventions](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_conventions.md)
```
kevindew, line 6:
```suggestion
- [GOV.UK Component Conventions](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_conventions.md)
```

app/components/admin/editions/editorial_remark_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<li class="app-component-editions-editorial-remark__list-item govuk-!-margin-bottom-4">
+<li class="app-c-editions-editorial-remark__list-item govuk-!-margin-bottom-4">
```
kevindew, line 1: This isn't a GOV.UK component, this is a view component. Should it still have an `app-c` prefix?

If I remember right we were going to style these based on views so I'd have thought it should be using `app-view` ? whereas only things in `app/views/components` would use `app-c` since they're for GOV.UK components.

docs/css.md
```diff
@@ -1,115 +1,125 @@
...
+
+## Namespacing
+
+### Components
```
kevindew, line 10: We should have something here to explain the difference between handling components in GOV.UK components and view components, as this seems an area to easily get confused with.

app/components/admin/editions/editorial_remark_component.html.erb
kevindew: More info in here: https://github.com/alphagov/whitehall/issues/6954

docs/legacy_css.md
```diff
@@ -0,0 +1,115 @@
+# Whitehall Legacy CSS Styleguide
```
kevindew, line 1: It'd be useful to explain what legacy CSS is and when you might consider this guide. 

Are all the bits of this still relevant too? I wonder if there are bits that should be deleted or if even need this file at all.

kevindew (commented):
Nice one Dilwoar, just added a few comments. Tricky area seems difference between GOV.UK components and view components.

app/components/admin/editions/editorial_remark_component.html.erb
davidgisbey: We had a long discussion about this the other day in the dev catchup. There was some discussion about maybe using something like "app-vc-*", or `app-c`, or whether they'd live in the views folder.

Since view components can live in multiple views, i think it's better for them to have their own folder rather than living in `stylesheets/admin/view_components/_component-name.scss`

DilwoarH: We could move it to its own namespace, which may be better for differentiating between the 2 different types of components.

docs/css.md
DilwoarH: Thanks - great point, I'll add something in

docs/legacy_css.md
DilwoarH: I think a lot of them might not be relevant, will look through and see what is not needed a bit more carefully. The other thought I had was to just delete this file as we're no longer planning to support it.

app/components/admin/editions/editorial_remark_component.html.erb
kevindew: As I recall the agreement was that view components were always to be coupled to views and if it was shared it should use a regular publishing component, thus we shouldn't have the shared view component concern (and the purpose collision that has with publishing components). 

It looks like we only have one now and I imagine that should probably be just a helper since it's just a wrapper around a publishing component. So maybe we'll still safe to return to that intention?

I don't mind too much if they have a different namespace to app-view (though I think that should be fine if we stick with the view coupling intention), but do think we should avoid app-c since it'd imply it's a publishing component.

docs/legacy_css.md
kevindew: Sure, I think it's likely fine to delete it only seems useful if we expect people to read it to know how to do something.

app/components/admin/editions/editorial_remark_component.html.erb
DilwoarH: Should we have them in all under `/views` as we're only going to be using view components for views?

Where we have shared partials, we can store them under `/views/shared`.

The less we use the word `component`, the less likely there will be confusion.

In terms of the namespace, it might be good to keep the work `component` out of it. My suggestion would be to just put everything under `app-view-*` (as all views eventually will become view components

ollietreend: > @davidgisbey: Since view components can live in multiple views

> @kevindew: As I recall the agreement was that view components were always to be coupled to views and if it was shared it should use a regular publishing component

I think there's a subtle distinction here that might be helpful to acknowledge.

Our use of View Components is coupled to the View layer, but I don't think they have to be tied directly to one single 'view' file (i.e. one page). I think they could also be used in place of partials – i.e. reusable chunks of the View layer that might appear in more than one place – and still fit within our definition of acceptable use of View Components.

So another way to think about this CSS class namespacing problem is: what namespace should we be using for shared view partials? Because the same answer should probably apply to our View Components.

kevindew: Yup good shout. I think if you were to do a partial e.g. `app/views/admin/editions/_change-notes.html.rb` you'd probably use a prefix that referred to that partial e.g. `app-view-editions-change-notes` ?

docs/css.md
DilwoarH: Done

app/components/admin/editions/editorial_remark_component.html.erb
DilwoarH: I have updated the documentation

docs/legacy_css.md
DilwoarH: I have added a notice for users to make sure they understand it's legacy and will be deleted once bootstrap can be removed.

kevindew (approved):

PR #7171, opened 2022-12-20
Add tracking for release 1.4

jyoung-gds:

[Trello ticket link](https://trello.com/c/oi4jFQW5/1012-add-app-wide-page-view-user-organisation-tracking)

## What
- Extends the `analyics.js` file to allow custom dimensions to be added using meta tags.
- Adds accompanying spec file to test the extension
- Uses meta tags to add page view tracking for user organisation to custom dimension 8 on the design system layout page, providing tracking throughout the app.

## Why
This behaviour was present in the bootstrap admin layout and needed to be replicated with the move to design system.

### Before
image

### After
image

### DIFF
--- a/app/assets/javascripts/admin/modules/analytics.js
+++ b/app/assets/javascripts/admin/modules/analytics.js
@@ -1,6 +1,18 @@
 (function (GOVUK) {
   'use strict'
 
+  GOVUK.setCustomDimensionsFromMetaTags = function () {
+    var metas = document.querySelectorAll("meta[name^='custom-dimension']")
+
+    for (var i = 0; i < metas.length; i++) {
+      var meta = metas[i]
+      var dimensionId = parseInt(meta.getAttribute('name').split('custom-dimension:')[1])
+      var content = meta.getAttribute('content')
+
+      GOVUK.analytics.setDimension(dimensionId, content)
+    }
+  }
+
   GOVUK.Analytics.load()
   GOVUK.analyticsVars = { primaryLinkedDomains: [document.domain] }
   GOVUK.analytics = new GOVUK.Analytics({
@@ -8,6 +20,7 @@
     cookieDomain: document.domain
   })
 
+  GOVUK.setCustomDimensionsFromMetaTags()
   GOVUK.analytics.trackPageview()
   GOVUK.analyticsPlugins.externalLinkTracker()
 })(window.GOVUK)
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -1,5 +1,11 @@
 <% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
 
+<% if user_signed_in? %>
+  <% content_for :head do %>
+    <meta name="custom-dimension:8" content="<%= current_user.organisation_slug.presence || '(not set)' %>">
+  <% end %>
+<% end %>
+
 <%= render 'govuk_publishing_components/components/layout_for_admin',
   product_name: "Whitehall Publisher",
   environment: environment,
--- a/spec/javascripts/admin/modules/analytics.spec.js
+++ b/spec/javascripts/admin/modules/analytics.spec.js
@@ -1,0 +1,69 @@
+describe('GOVUK.analytics', function () {
+  'use strict'
+  var GOVUK = window.GOVUK
+
+  function addGoogleAnalyticsSpy () {
+    if (typeof window.ga === 'undefined') {
+      window.ga = function () {}
+    }
+    spyOn(window, 'ga')
+  }
+
+  describe('setCustomDimensionsFromMetaTags', function () {
+    var metaTag
+
+    beforeEach(function () {
+      metaTag = document.createElement('meta')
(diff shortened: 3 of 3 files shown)


app/assets/javascripts/admin/modules/analytics.js
```diff
@@ -1,13 +1,26 @@
...
+  }
+
   GOVUK.Analytics.load()
   GOVUK.analyticsVars = { primaryLinkedDomains: [document.domain] }
   GOVUK.analytics = new GOVUK.Analytics({
     universalId: 'UA-26179049-6', // GOVUK Apps GA ID
     cookieDomain: document.domain
   })
 
+  GOVUK.setCustomDimensionsFromMetaTags()
```
DilwoarH, line 23: Is there a reason why this is attached to the GOVUK namespace?

jyoung-gds: I was having trouble accessing the function in the test file, attaching it to GOVUK allowed me to call it when testing. I wasn't sure if there's a better/cleaner way to do this.

app/views/layouts/design_system.html.erb
```diff
@@ -1,5 +1,11 @@
 <% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
 
+<% if user_signed_in? && current_user.has_email? %>
```
ollietreend, line 3: What's the significance of using `current_user.has_email?` here?

Is there a case where a user will be signed in _without_ an email address? And if so, why don't we track their organisation slug (if they have one)?

app/assets/javascripts/admin/modules/analytics.js
ollietreend: This seems like a tricky one. This file (`admin/modules/analytics.js`) isn't a typical [GOV.UK JS module](https://docs.publishing.service.gov.uk/repos/govuk_publishing_components/javascript-modules.html) because it automatically runs on page load, rather than exporting a Class/Prototype/Function that will later be instantiated on a DOM element.

That makes it pretty difficult to test since it will only be executed once, on page load, when the Jasmine test runner first loads up. Whereas the unit tests need to be able to instantiate Analytics multiple times over, to assert that the correct Custom Dimensions are set.

@DilwoarH is there a better / established pattern that can be used here instead of putting it in the `GOVUK` namespace?

app/views/layouts/design_system.html.erb
```diff
@@ -1,5 +1,11 @@
...
+  <% content_for :head do %>
+    <meta name="custom-dimension:8" content="<%= current_user.organisation_slug.present? ? current_user.organisation_slug : '(not set)' %>">
```
ollietreend, line 5: I think it should be possible to shorten this by using [Object#presence](https://apidock.com/rails/Object/presence):

```suggestion
    <meta name="custom-dimension:8" content="<%= current_user.organisation_slug.presence || '(not set)' %>">
```

app/views/layouts/design_system.html.erb
ollietreend: Also, it'd be nice to adapt/reuse/duplicate [this integration test](https://github.com/alphagov/whitehall/blob/c0ad13dd04e6c2f3b53cd4373379a0b0f3dc307f/test/integration/admin_analytics_test.rb) which covers the Bootstrap version of this behaviour.

jyoung-gds: From what I can tell there may be a case where a signed in user has no email address, but I don't see a reason not to track their organisation slug in this case. This logic was used in the old admin layout file, they don't give any indication why it's in there in the PR and I've checked the integration tests and they don't test for a case where the user has no email, so I'm assuming it's safe to remove. Thanks for flagging @ollietreend

app/assets/javascripts/admin/modules/analytics.js
ollietreend: Our Jasmine test suite is only really suitable for unit testing JavaScript code. The ideal 'unit' of JavaScript would be an isolated function that can be called & torn down repeatedly. And since these unit tests all run within a single page in the browser, they shouldn't really rely on global state belonging to the overall page (e.g. `<meta>` tags present on the Jasmine test page).

> @DilwoarH: Is there a way to add the meta tag before the jasmine JS runs?

Therefore, this doesn't feel like a great option. It's kind of bending the unit test suite to do something it wasn't designed to do, which is why it feels a bit weird – it's the code telling us it's a bad fit for unit testing.

It feels like we have a couple of options:

1. **Unit test:** If we want to unit test this, we need the behaviour to be accessible from our Jasmine tests. This is the route @jyoung-gds has taken by 'exporting' that `setCustomDimensionsFromMetaTags` function to the `window.GOVUK` global object, so our unit tests can call it.
2. **Integration test:** If we want this to effectively be an 'anonymous' or private function which only runs once on page load, then it'd probably be more appropriate to cover its behaviour in a higher-level JS-enabled integration test. That way we'll be able to reload the page to easily test its behaviour under specific conditions.

Maybe we should be exploring Option 2 here. What do you think?

DilwoarH (approved):
Approving this for now so that it doesn't block the release. This will be refactored as part of the JS review and refactor

PR #7172, opened 2022-12-20
Republish organisation when promotional feature is saved or destroyed

davidgisbey:
## Description

Currently, you have to return to the organisation page and click save to push changes to promotional features and promotional feature links downstream to the PublishingApi

This updates the `PromotionalFeatureItemsController` and `PromotionalFeaturesController` `create`, `update` and `destroy` actions to republish the organisation on a successful action (save or delete).

## Trello card

https://trello.com/c/bmFyYErA/1008-automatically-republish-organisation-when-promotional-features-are-changed

### DIFF
--- a/app/views/admin/promotional_feature_items/new.html.erb
+++ b/app/views/admin/promotional_feature_items/new.html.erb
@@ -1,4 +1,5 @@
 <% page_title "Feature item for #{@promotional_feature.title}" %>
 <h1>New feature item for <%= @promotional_feature.title %></h1>
+<p class="warning">Warning: newly created feature items will appear instantly on the live site.</p>
 
 <%= render "form", promotional_feature_item: @promotional_feature_item, promotional_feature: @promotional_feature, organisation: @organisation %>
--- a/app/controllers/admin/promotional_feature_items_controller.rb
+++ b/app/controllers/admin/promotional_feature_items_controller.rb
@@ -11,6 +11,7 @@ class Admin::PromotionalFeatureItemsController < Admin::BaseController
   def create
     @promotional_feature_item = @promotional_feature.promotional_feature_items.build(promotional_feature_item_params)
     if @promotional_feature_item.save
+      Whitehall::PublishingApi.republish_async(@promotional_feature_item.organisation)
       redirect_to_feature "Feature item added."
     else
       render :new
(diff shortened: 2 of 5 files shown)


app/models/promotional_feature_item.rb
```diff
@@ -10,4 +10,13 @@ class PromotionalFeatureItem < ApplicationRecord
   accepts_nested_attributes_for :links, allow_destroy: true, reject_if: ->(attributes) { attributes["url"].blank? }
 
   mount_uploader :image, ImageUploader
+
+  after_save :republish_organisation_to_publishing_api
+  after_destroy :republish_organisation_to_publishing_api
```
ollietreend, line 15: There's a risk with this approach that any bulk actions performed on `PromotionalFeatureItem`s will trigger a barrage of republish calls to the Publishing API. This could prove problematic once we add the custom sort functionality which will end up updating every `PromotionalFeatureItem` on a `PromotionalFeature` to set its sort order.

This has hurt us in the past – for example, I remember a situation where [document attachments behaved similarly](https://docs.google.com/document/d/11zmKzkbUxlJtd8Din3Y_x0u46mPhL1lX0saQiVnuzfo/edit#) and caused a GOV.UK incident. And a [similar incident occurred](https://docs.google.com/document/d/1exAZhfDg1ZmlCQ0hmE3flhFh9EpXO3qDr-EG4a6Dz1Y/edit#) on orgs with large numbers of featured documents.

I expect this is less likely to result in incidents due to only two pages using promotional features, and neither using very many. So they won't result in huge spikes of republish jobs. However, it'd be ideal if we could find somewhere 'further up' in the code so this republishing only gets triggered once per user-performed action.

Off the top of my head, I don't have any suggestions for how/where to implement that. (Sorry!) But I'm hoping we can find something.

davidgisbey: Okay cool. I think we can just add it into the promotional features controller 👍

kevindew: Yeah I think moving things to controllers (or services) is preferable to using active record callbacks. Ideally we'd get rid of all the active record callbacks that trigger writes to the Publishing API.

davidgisbey: okay cool they're in a few other places. I can look at them over the holiday break

app/views/admin/promotional_feature_items/new.html.erb
```diff
@@ -1,4 +1,5 @@
 <% page_title "Feature item for #{@promotional_feature.title}" %>
 <h1>New feature item for <%= @promotional_feature.title %></h1>
+<p class="warning">Warning: newly created feature items will appear instantly on the live site.</p>
```
ollietreend, line 3: Nice touch ✨

ollietreend (approved):
This looks great – nice work @davidgisbey! ✨🚀

PR #7178, opened 2022-12-21
Make Analytics code testable

DilwoarH:
## What

Moves analytics code into a module to make it easier to unit test.

## Why

Previously the code was not testable and this meant that things could be potentially failing without us knowing.

https://trello.com/c/tUuuK8IG/842-review-js-modules


app/views/layouts/design_system.html.erb
```diff
@@ -19,7 +19,7 @@
     </div>
   </div>
 
-  <div class="govuk-width-container">
+  <div class="govuk-width-container" data-module="app-analytics">
```
ollietreend, line 22: Is there any reason why you've chosen to apply the `data-module` to this HTML element in particular?

I realise there's no way to directly apply it to the `<html>` or `<body>` elements since they're provided by the govuk_publishing_components gem.

I'd be tempted to suggest creating an empty standalone `<div>` element to hold it. That way we can avoid any future confusion where developers might assume it's coupled to this container div specifically – or worse yet, implement functionality that depends on `this.module` being this specific element.

It'd also mean we can include a comment to explain the reason for this div to the casual passer by.

```html
<!-- This element exists to initialise the JS module that configures custom Analytics behaviour -->
<div data-module="app-analytics"></div>
```

DilwoarH: Good point, I will update

ollietreend (approved):
Nice – looks good to me @DilwoarH 👍🏻

PR #7179, opened 2022-12-21
Add reordering of promotional features

davidgisbey:
## Description

We are adding the ability to reorder promotional features. 

This PR does the following:

1. Backfills the `PromotionalFeature#ordering` column.
2. Sets the ordering value before save if blank (should mainly just assign before creation)
3. Orders the `Organisation#promotional_features` or `order` 
4. Adds the reorder endpoint

### Promotional feature index page

image

### Reorder page 

image

### DIFF
--- a/test/functional/admin/promotional_features_controller_test.rb
+++ b/test/functional/admin/promotional_features_controller_test.rb
@@ -84,4 +84,49 @@ class Admin::PromotionalFeaturesControllerTest < ActionController::TestCase
     assert_not PromotionalFeature.exists?(promotional_feature.id)
     assert_equal "Promotional feature deleted.", flash[:notice]
   end
+
+  test "GET :reorder loads the promotional feature belonging to the organisation" do
+    promotional_feature1 = create(:promotional_feature, organisation: @organisation)
+    promotional_feature2 = create(:promotional_feature, organisation: @organisation)
+
+    get :reorder, params: { organisation_id: @organisation }
+
+    assert_response :success
+    assert_template :reorder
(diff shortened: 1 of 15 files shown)


test/functional/admin/promotional_features_controller_test.rb
```diff
@@ -84,4 +84,43 @@ class Admin::PromotionalFeaturesControllerTest < ActionController::TestCase
...
+    assert_redirected_to admin_organisation_promotional_features_path(@organisation)
+  end
+
+  test "PATCH on :update_order reorders the promotional_features and republishes the organisation" do
+    Organisation.stubs(:allowed_promotional).returns(organisations_array = mock)
+    organisations_array.expects(:find).with(@organisation.slug).returns(@organisation)
+    @organisation.expects(:reorder_promotional_features).with do |value|
+      value["2"] == "1" && value["1"] == "2"
+    end
+    Whitehall::PublishingApi.expects(:republish_async).with(@organisation)
```
ollietreend, line 113: Just a suggestion (not even sure if it'll work)

Can we specify here that the Publishing API is **only called once**?

I'm thinking that'd be a nice way to lock-in the expectation that we don't end up republishing the organisation more than once. For example, it's protect against a future regression where somebody refactors the Publishing API call into an `after_save` callback on the ActiveRecord model.

However – it looks like you've stubbed out `Organisation#reoder_promotional_features` here, so actually this whole suggestion might be a moot point because that method isn't covered by this test.

ollietreend (approved):
This looks excellent ✨🚀 

Really nice, clean & tidy work @davidgisbey! Thanks for doing this.

PR #7185, opened 2022-12-23
Updates payload to support images in Preview

DilwoarH:
## What

Fixes payload so images load correctly when users are previewing.

## Why

A bug was reported where image attachments would not show when previewing in the new design system layout.

Dec-23-2022 11-42-56 am

https://trello.com/c/XHW0LKvX/1023-images-are-not-working-in-preview-mode

### DIFF
--- a/app/assets/javascripts/components/govspeak-editor.js
+++ b/app/assets/javascripts/components/govspeak-editor.js
@@ -14,15 +14,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     return document.querySelector('meta[name="csrf-token"]').content
   }
 
-  GovspeakEditor.prototype.getRenderedGovspeak = function (content, callback) {
-    var body = new FormData()
-    body.append('body', content)
+  GovspeakEditor.prototype.getRenderedGovspeak = function (body, callback) {
+    var data = this.generateFormData(body)
 
     var request = new XMLHttpRequest()
     request.open('POST', '/government/admin/preview', false)
     request.setRequestHeader('X-CSRF-Token', this.getCsrfToken())
     request.onreadystatechange = callback
-    request.send(body)
+    request.send(data)
+  }
+
+  GovspeakEditor.prototype.generateFormData = function (body) {
+    var data = new FormData()
+    data.append('body', body)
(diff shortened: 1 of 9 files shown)


nnagewad:
Nice

kevindew:
I'm a little confused about the scope of this so wanted to enquire.

It looks like this fixes the issue for editing an edition, which as I understand is not a feature we've released yet (as that's still in bootstrap). However I was under the impression that the more immediate problem was that images that were embedded in a HTML attachment were not previewable, which - if I understand right - this won't resolve?

I think we're also missing the ability to preview attachments as per https://github.com/alphagov/whitehall/blob/99f2bda53e3dda05d0e70cb023772027eb151803/app/assets/javascripts/admin_legacy/preview.js#L71 ?

DilwoarH:
> I'm a little confused about the scope of this so wanted to enquire.
> 
> It looks like this fixes the issue for editing an edition, which as I understand is not a feature we've released yet (as that's still in bootstrap). However I was under the impression that the more immediate problem was that images that were embedded in a HTML attachment were not previewable, which - if I understand right - this won't resolve?
> 
> I think we're also missing the ability to preview attachments as per
> 
> https://github.com/alphagov/whitehall/blob/99f2bda53e3dda05d0e70cb023772027eb151803/app/assets/javascripts/admin_legacy/preview.js#L71
> 
> ?

I think there are two separate issues. This one fixes the issue with the preview in the edit draft view. The issue with the images is a separate issue - which is still being investigated.

DilwoarH:
Decision made outside of PR:

1. Update the component to take in 3 new properties `image_ids`, `attachment_ids`, and `alternative_format_provider_id`.
2. Stores these as data attributes within the component
3. Update the JS to pull from data attributes
4. Create an additional module to listen to dynamic elements like `alternative_format_provider_id`

cc @kevindew

DilwoarH:
@kevindew ready for re-review

davidtrussler (approved):
The code changes look good to me. I haven't been able to verify it works as expected locally.

DilwoarH:
> The code changes look good to me. I haven't been able to verify it works as expected locally.

I've tested it on integration multiple times

PR #7196, opened 2022-12-30
Update govspeak syntax for images

baisa:
[Trello card](https://trello.com/c/9oMxpTMN/991-update-govspeak-markdown-syntax)

New syntax in Bootstrap design:
Screenshot 2022-12-30 at 11 35 09
Screenshot 2022-12-30 at 11 35 51

New syntax in the new Design System:
Screenshot 2022-12-30 at 11 39 12
Screenshot 2022-12-30 at 11 39 21

### DIFF
--- a/app/controllers/admin/preview_controller.rb
+++ b/app/controllers/admin/preview_controller.rb
@@ -1,10 +1,12 @@
 class Admin::PreviewController < Admin::BaseController
   before_action :find_attachments
   before_action :limit_attachment_access!
+  include GovspeakHelper
 
   def preview
     if Govspeak::HtmlValidator.new(params[:body]).valid?
-      @images = Image.find(params.fetch(:image_ids, []))
+      images = Image.find(params.fetch(:image_ids, []))
+      @images = images_setup(images)
       @alternative_format_contact_email = alternative_format_contact_email
       render layout: false
     else
--- a/app/views/admin/editions/_image_template.html.erb
+++ b/app/views/admin/editions/_image_template.html.erb
@@ -67,7 +67,7 @@
             bold: true,
           },
           name: "readonly",
-          value: "!!#{index + 1}",
+          value: "[Image:#{image.image_data.carrierwave_image}]",
           readonly: true,
         } %>
       <% end %>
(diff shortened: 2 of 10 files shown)


app/views/admin/editions/_image_template.html.erb
```diff
@@ -67,7 +67,7 @@
             bold: true,
           },
           name: "readonly",
-          value: "!!#{index + 1}",
```
DilwoarH, line 70: What will happen to old documents that use the `!!n` syntax?

baisa: @DilwoarH the old syntax will still be valid and working - please let me know if you noticed otherwise

dnkrj (commented):
Looks good, we'll need to update the govspeak help too
https://github.com/alphagov/whitehall/blob/main/app/views/admin/editions/_govspeak_help.html.erb#L92-L95

baisa:
> Looks good, we'll need to update the govspeak help too https://github.com/alphagov/whitehall/blob/main/app/views/admin/editions/_govspeak_help.html.erb#L92-L95

Good point @dnkrj Have added it now

app/controllers/admin/preview_controller.rb
```diff
@@ -35,4 +35,20 @@ def limit_attachment_access!
...
+        image_data_id: image.image_data_id,
+        edition_id: image.edition_id,
+        alt_text: image.alt_text,
+        url: image.url,
+        caption: image.caption,
+        created_at: image.created_at,
+        updated_at: image.updated_at,
+      }
+    end
+  end
```
DilwoarH, line 53: Do you think we should share the mapping function in the helper? So we don't have duplicate code?

DilwoarH:
Looks great, one small comment - it'll be good to get @johnpauldickie's thoughts on the implementation / visuals

DilwoarH:
I just thought of a question - what happens when we have two images with the same filename?

kevindew:
> I just thought of a question - what happens when we have two images with the same filename?

I can confirm that will be a problem. The carrierwave_image field only stores a (normalised) version of the original filename and will have conflicts. 

I imagine we probably need to create a new field on the image table that represents the key for embedding.

ollietreend:
Superseded by #7589

PR #7198, opened 2023-01-03
Reintroduce and fix the basics for models knowing their own routes

brucebolt:
This reintroduces the code for starting to migrate models to know their own routes, which was removed in https://github.com/alphagov/whitehall/pull/7183 as it caused an incident.

A test has been added to cover the scenario in the incident and the bug fixed.  Additionally this revealed a somewhat related issue with statistical announcements, which is also fixed here.

[Trello card](https://trello.com/c/86Gc4TqZ)


BeckaL (approved):
Looks good to me. Nice to have caught the base path /public path thing too. Let's deploy it on integration first to do a sanity check (assuming you haven't already)?

PR #7201, opened 2023-01-03
Add "position: sticky" to the miller columns

dnkrj:

In longer lists the children of the item of a miller column element are sometimes not visible when you click on it. Using `position: sticky` ensures that they are displayed on screen when possible.

## Before
https://user-images.githubusercontent.com/9594455/210401714-a8fb1fac-4ca6-47f4-a03c-e8eab3247347.mov

## After
https://user-images.githubusercontent.com/9594455/210401729-a4d0af56-b8ac-44bf-a691-dbded31348ff.mov

https://trello.com/c/YJ8n9Hc9/1097-add-position-sticky-to-miller-columns

### DIFF
--- a/app/assets/stylesheets/components/_miller-columns.scss
+++ b/app/assets/stylesheets/components/_miller-columns.scss
@@ -3,3 +3,8 @@
 .app-c-miller-columns__search {
   margin-bottom: govuk-spacing(6);
 }
+
+.miller-columns__column {
+  position: sticky;
+  top: 0;
+}


kevindew:
This looks neat 👍 

It seems like something that we may want on the component itself rather than as a specific style override: https://github.com/alphagov/miller-columns-element

DilwoarH:
Should we make the search input sticky too?

dnkrj:
> 

Making the search bar sticky is a bit more difficult as it would block the first line of the miller columns when the page scrolls. I think the real win comes in doing the column.

> This looks neat 👍
> 
> It seems like something that we may want on the component itself rather than as a specific style override: https://github.com/alphagov/miller-columns-element

Currently miller columns support IE 11, I think that we can update it in line with GOV.UK Frontend support matrix and reduce IE support to functional but will do that in a follow up.

davidgisbey (approved):
this is super cool 👍

DilwoarH:
Do you think we need to push this upstream on the package itself?

PR #7216, opened 2023-01-09
Allow outstanding Models to define their own routes

callumknights:
Allow outstanding Models to define their own routes

This change allows certain models to move away from the use of UrlMaker, which is necessary to migrate
rendering of these Models outside of Whitehall, without leaving unnecessary routes behind just to take 
advantage of Rails routing helpers. 

Both Worldwide Organisations and Operational Fields are still rendered by Whitehall, so I've left the route intact. 
In the case of those two Models this PR will simply make migrating the rendering later on, easier. 

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -2,9 +2,11 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   def append_url_options(path, options = {})
-    if options[:format] && options[:locale]
+    locale = normalise_locale(options[:locale])
+
+    if options[:format] && options[:locale] && locale != I18n.default_locale
       path = "#{path}.#{options[:locale]}.#{options[:format]}"
-    elsif options[:locale] && options[:locale] != I18n.default_locale
+    elsif options[:locale] && locale != I18n.default_locale
       path = "#{path}.#{options[:locale]}"
     elsif options[:format]
       path = "#{path}.#{options[:format]}"
@@ -20,5 +22,18 @@ class ApplicationRecord < ActiveRecord::Base
     path = "#{path}##{options[:anchor]}" if options[:anchor]
 
     path
+  end
+
+private
+
+  def normalise_locale(locale)
+    case locale
+    when Struct
+      locale.code
+    when String
+      locale.to_sym
+    when Symbol
+      locale
+    end
   end
 end
--- a/app/models/policy_group.rb
+++ b/app/models/policy_group.rb
@@ -60,6 +60,18 @@ class PolicyGroup < ApplicationRecord
     true
   end
 
+  def base_path
+    "/government/groups/#{slug}"
+  end
+
+  def public_path(options = {})
+    append_url_options(base_path, options)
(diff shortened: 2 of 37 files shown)


app/models/policy_group.rb
```diff
@@ -73,6 +85,6 @@ def summary_or_name
              description: :summary
 
   def search_link
-    Whitehall.url_maker.policy_group_path(slug)
+    public_path
```
brucebolt, line 88: As the `search_link` method is only used once, is it possible to inline `:public_path` on line 83 rather than defining this method?

app/models/worldwide_organisation.rb
```diff
@@ -84,7 +84,7 @@ def sponsoring_organisation
              format: "worldwide_organisation"
 
   def search_link
-    Whitehall.url_maker.worldwide_organisation_path(slug)
+    public_path
```
brucebolt, line 87: Same here with the `search_link` method (I think it applies in other models too).

app/models/application_record.rb
```diff
@@ -2,9 +2,11 @@ class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
 
   def append_url_options(path, options = {})
-    if options[:format] && options[:locale]
+    locale = normalise_locale(options[:locale])
+
+    if options[:format] && options[:locale] && locale != I18n.default_locale
```
brucebolt, line 7: Nice find!  I'd missed the possibility of `:en` being passed in as a parameter, so this should make the code more robust.

app/models/application_record.rb
```diff
@@ -21,4 +23,17 @@ def append_url_options(path, options = {})
 
     path
   end
+
+private
+
+  def normalise_locale(locale)
+    case locale
+    when Struct
+      locale.to_a[0]
```
brucebolt, line 32: I think there's a `code` method on the `Struct` that can be used instead of making an assumption the code is always the first element.

test/functional/ministerial_roles_controller_test.rb
```diff
@@ -180,33 +180,6 @@ class MinisterialRolesControllerTest < ActionController::TestCase
     assert_equal whips, assigns(:whips_by_organisation)
   end
 
-  view_test "shows the cabinet minister's name and role" do
```
brucebolt, line 183: These tests are probably still needed, since Whitehall still renders these pages.

brucebolt (commented):
This is looking great, thanks for taking on this work @callumknights 🎖️.

I've left a few minor comments, but otherwise everything looks ready to approve.

test/functional/ministerial_roles_controller_test.rb
callumknights: Added back in :)

app/models/application_record.rb
callumknights: Sorted 👍

brucebolt (approved):
Thanks for making the changes, everything looks good to me 👍

PR #7246, opened 2023-01-13
Add JS for Speeches to the EditionForm module

davidgisbey:
## Description

When a user creates a speech, one of the possible subtypes is `Authored article`. Unlike the other speeches, this is written not spoken. 

Due to this, when an `Authored article` is selected the following needs to happen:

1. The location field needs to be hidden and the value reset
2. Any reference to `Speaker` in labels needs to be updated to `Writer`

Then when the subtype moves from `Authored edition` to another subtype the reverse needs to the occur: 

1. the location field needs to be displayed
2. Any reference to `Writer` needs to be updated to `Speaker`

This PR adds the speech form to displays the correct fields/values on page load and adds a JS module which does the above dynamically.

### With authored article subtype

image

image

### With any other subtype

image

image

## Trello card

https://trello.com/c/KOGs7sJq/1072-add-javascript-for-authoredarticles-to-hide-show-fields-subtype-of-speech

### DIFF
--- a/app/assets/javascripts/admin/views/edition-form.js
+++ b/app/assets/javascripts/admin/views/edition-form.js
@@ -74,20 +74,20 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var hasProfileRadioLabel = form.querySelector('#edition_role_appointment label[for="edition_role_appointment_speaker_on_govuk"]')
     var noProfileRadioLabel = form.querySelector('#edition_role_appointment label[for="edition_role_appointment_speaker_not_on_govuk"]')
     var deliveredOnLabel = form.querySelector('#edition_delivered_on .govuk-fieldset__legend')
-    var locationDiv = form.querySelector('.js-app-view-edition-form__speech-location')
+    var locationDiv = form.querySelector('.js-app-view-edit-edition__speech-location-field')
     var locationInput = locationDiv.querySelector('input[name="edition[location]"]')
     var authoredArticleId = '6'
 
     select.addEventListener('change', function (event) {
       if (event.currentTarget.value === authoredArticleId) {
-        locationDiv.classList.add('govuk-visually-hidden')
+        locationDiv.classList.add('app-view-edit-edition__speech-location--hidden')
         deliveredByLabel.textContent = 'Writer (required)'
         hasProfileRadioLabel.textContent = 'Writer has a profile on GOV.UK'
         noProfileRadioLabel.textContent = 'Writer does not have a profile on GOV.UK'
         deliveredOnLabel.textContent = 'Written on (required)'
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/admin/views/edition-form.js
```diff
@@ -62,5 +63,38 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    var hasProfileRadioLabel = form.querySelector('#edition_role_appointment label[for="edition_role_appointment_speaker_on_govuk"]')
+    var noProfileRadioLabel = form.querySelector('#edition_role_appointment label[for="edition_role_appointment_speaker_not_on_govuk"]')
+    var deliveredOnLabel = form.querySelector('#edition_delivered_on .govuk-fieldset__legend')
+    var locationDiv = form.querySelector('.js-app-view-edition-form__speech-location')
+    var locationInput = locationDiv.querySelector('input[name="edition[location]"]')
+    var authoredArticleId = '6'
+
+    select.addEventListener('change', function (event) {
+      if (event.currentTarget.value === authoredArticleId) {
+        locationDiv.classList.add('govuk-visually-hidden')
```
dnkrj, line 83: `'govuk-visually-hidden'` will only remove this for non-screenreader users, I think we want it to be hidden for everyone, so we should probably use the `hidden` property.

```suggestion
        locationDiv.hidden = true
```

app/assets/javascripts/admin/views/edition-form.js
```diff
@@ -62,5 +63,38 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        deliveredByLabel.textContent = 'Writer (required)'
+        hasProfileRadioLabel.textContent = 'Writer has a profile on GOV.UK'
+        noProfileRadioLabel.textContent = 'Writer does not have a profile on GOV.UK'
+        deliveredOnLabel.textContent = 'Written on (required)'
+        locationInput.value = ''
+      } else {
+        locationDiv.classList.remove('govuk-visually-hidden')
```
dnkrj, line 90:
```suggestion
        locationDiv.hidden = false
```

dnkrj (commented):
Looks good, just think we need to hide the input for everyone?

app/assets/javascripts/admin/views/edition-form.js
davidgisbey: When looking at this i noticed for the locale switcher added a class with `display: none`

image

So i've just piggy backed off of that so it'll apply correctly on page load. Does that sound reasonable?

dnkrj (approved):

davidgisbey:
thanks for the review @dnkrj

PR #7251, opened 2023-01-16
Remove constraint and comment on Mailer version in Gemfile

davidgisbey:
Previously, there was an issue with the mailer gem in release 2.8.0 mikel/mail#1489.

This has been resolved so we can remove the comment and use the latest stable release.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,6 @@ gem "jbuilder"
 gem "jquery-ui-rails"
 gem "kaminari"
 gem "link_header"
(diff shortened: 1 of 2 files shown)


kevindew:
Thanks!

I've just realised that mail is also not a direct dependency, so we can actually remove it from the Gemfile directly. It was only added in: https://github.com/alphagov/whitehall/commit/33ce47b95b8c4cb3f9cca812ba69475b36947f48

davidgisbey:
Thanks. I'll have a look through the other apps and updated the comment/remove it where possible.

kevindew (approved):
Thanks, sorry about not realising mail should be removed. Hope you hadn't got too far.

PR #7256, opened 2023-01-17
Create data_migration to change Queen slugs

beccapearce:
adds a data migration to change slugs mentioning the Queen to mention the King.

As it is a data migration it will need to be manually run in the relevant environment.

Trello card: https://trello.com/c/7KBZxFl6


db/data_migration/20230104104917_reslug_her_majesty_to_his_majesty.rb
```diff
@@ -0,0 +1,105 @@
...
+    statistics_announcement = StatisticsAnnouncement.find_by(slug: old_slug)
+
+    unless statistics_announcement
+      puts "!! no statistics announcement found for: #{old_slug}"
+      next
+    end
+
+    Whitehall::SearchIndex.delete(statistics_announcement)
+    statistics_announcement.update!(slug: new_slug)
+    Whitehall::SearchIndex.add(statistics_announcement)
```
ollietreend, line 59: Double check: does this need to be sent to the Publishing API?

**I think it probably doesn't** – because StatisticsAnnouncement includes PublishesToPublishingApi, which will publish to Publishing API after committing to the database.

https://github.com/alphagov/whitehall/blob/c465d1d5a806d31333a7b2f01461f0a38ee59461/lib/publishes_to_publishing_api.rb#L6

But worth double checking in case our Friday brains are failing us.

db/data_migration/20230104104917_reslug_her_majesty_to_his_majesty.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  else
+    document = Document.find_by(slug: old_slug)
+
+    unless document
+      puts "!! no document found for: #{old_slug}"
+      next
+    end
+
+    edition = document.latest_edition
```
ollietreend, line 92: Should this be `live_edition` or `latest_edition`? 🤔

I'm not sure if this _and_ the HTML attachment route should both use the same?

db/data_migration/20230104104917_reslug_her_majesty_to_his_majesty.rb
```diff
@@ -0,0 +1,105 @@
...
+    Whitehall::PublishingApi.republish_async(html_attachment)
+
+    # remove the most recent edition from the search index
+    Whitehall::SearchIndex.delete(edition)
+
+    # send edition to publishing api
+    PublishingApiDocumentRepublishingWorker.new.perform(document.id)
+
+    # add edition to search index
+    Whitehall::SearchIndex.add(edition)
```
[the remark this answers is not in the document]
ollietreend: Also: could we use the `DataHygiene::DocumentReslugger` to handle this – once we've located the HTML attachment's document and updated the slug?

(We think probably not, because this is to update the slug of individual HTML attachments rather than the parent publication.)

If not: it would be good to follow-up by creating a `HtmlAttachmentReslugger` for future re-use.

db/data_migration/20230104104917_reslug_her_majesty_to_his_majesty.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    unless edition
+      puts "!! edition not found for: #{old_slug}"
+      next
+    end
+
+    Whitehall::SearchIndex.delete(edition)
+    document.update!(slug: new_slug)
+    PublishingApiDocumentRepublishingWorker.new.perform(document.id)
+    Whitehall::SearchIndex.add(edition)
```
ollietreend, line 102: Could this be replaced by `DataHygiene::DocumentReslugger`?

(We think maybe yes – but this was with our Friday brains)

ollietreend (approved):

PR #7257, opened 2023-01-17
See if reverting bc5d438 fixes Jasmine failures.

sengi:
This reverts commit bc5d438005be468fe39310221351bae2a2d9acf3.


sengi:
Nope, still [fails in the same way](https://github.com/alphagov/whitehall/actions/runs/3941842937/jobs/6744763949).

PR #7262, opened 2023-01-18
Delete role appointment for speech

baisa:
So the duplicated role can be removed as requested in https://govuk.zendesk.com/agent/tickets/5182511

As a result, role appointment can be deleted and consequently, so can duplicated role.

### DIFF
--- a/db/migrate/20230118130656_remove_role_appointment_for_speech.rb
+++ b/db/migrate/20230118130656_remove_role_appointment_for_speech.rb
@@ -1,0 +1,14 @@
+# Zendesk ticket https://govuk.zendesk.com/agent/tickets/5182511
+# Need to unlink the speech from the role appointment.
+# This speech in published state is already linked to person's new role.
+# Because the speech is superseded and would raise the validation
+# "cannot be modified when edition is in the superseded state"
+# it must be saved without running the validations.
+
+class RemoveRoleAppointmentForSpeech < ActiveRecord::Migration[7.0]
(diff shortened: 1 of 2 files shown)


robinjam (approved):

db/migrate/20230118130656_remove_role_appointment_for_speech.rb
```diff
@@ -0,0 +1,14 @@
...
+# "cannot be modified when edition is in the superseded state"
+# it must be saved without running the validations.
+
+class RemoveRoleAppointmentForSpeech < ActiveRecord::Migration[7.0]
+  def change
+    speech = Speech.find(1_216_085)
+    speech.role_appointment_id = nil
+    speech.save!(validate: false)
+  end
+end
```
DilwoarH, line 14: This causes issues locally where the data may not be present.

My recommendation would be to change to this:
```
class RemoveRoleAppointmentForSpeech < ActiveRecord::Migration[7.0]
  def change
    speech = Speech.find(1_216_085)
    if speech
      speech.role_appointment_id = nil
      speech.save!(validate: false)
    end
  end
end
```

Ideally, this could have been done better by the cli rather than a migration.

kevindew:
Good shout @DilwoarH, we should indeed change this as it will catch out devs in a few scenarios.

As an alternative Whitehall in future Whitehall does have a data migration pattern which is used for changes to data content rather than database structure.

PR #7267, opened 2023-01-19
Fix for remove-role-appointment-for-speech migration

baisa:
This is to update changes made in https://github.com/alphagov/whitehall/pull/7262


kevindew:
Change looks good, have you tested it locally with an empty db to see that it's safe?

baisa:
> Change looks good, have you tested it locally with an empty db to see that it's safe?

I have tested it locally with non-existing speech.

kevindew (approved):
Great stuff

PR #7268, opened 2023-01-19
Add Datetime fields Publishing Component

davidgisbey:
## Description

At the moment, we've got a fairly complex `datetime-fields` partial that we use to render the DateTime fields for editions.

In it's current implementation, it's coupled to editions which isn't ideal as we've got a need for DateTime inputs for a variety of objects.

Due to this, it makes sense to create a Datetime fields  Publishing Component. This should be model agnostic so it can be reused throughout Whitehall.

## Trello card 

https://trello.com/c/TLp8zlHT/899-add-publishing-component-for-datetime-fields

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -3,6 +3,7 @@ $govuk-page-width: 1140px;
 @import "govuk_publishing_components/all_components";
 
 @import "./components/autocomplete";
+@import "./components/datetime-fields";
 @import "./components/govspeak-editor";
 @import "./components/inset-prompt";
 @import "./components/miller-columns";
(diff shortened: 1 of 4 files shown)


davidtrussler:
@davidgisbey I wonder if rendering the component with a hint should be a separate example. It looks a little confusing when the default state already displays a hint without that being specified.

davidgisbey:
Yea that's a good shout. I'll add it in now. Thanks for the review.

davidgisbey:
@davidtrussler i've added the hint as an optional param

davidtrussler (approved):
Looks good! 
I gave some thought as to whether we ought to be styling this with `flexbox` - I suspect the layout might look a bit different in IE11 - but I think it will not be a problem in this case. 
👍

davidgisbey:
Thanks for the review @davidtrussler

PR #7273, opened 2023-01-20
Reintroduce polling ajax for broken link checker

dnkrj:

## What
[Reintroduce polling AJAX for broken link checker.](https://trello.com/c/doeqw0lS/990-reintroduce-polling-ajax-for-broken-link-checker)
Adds javascript to handle the button click and automatically refreshing the link checking report.
Introduces json template for the link check report.

## Why
This functionality existed in the bootstrap version of the page so this brings the page to like for like.
The bootstrap version used a .js template to send the data to the client, json is preferred as it does not require client side execution.

### With Javascript
https://user-images.githubusercontent.com/9594455/213690598-21914d6a-3235-4e31-bc72-5a28c93bb815.mov

### Without Javascript
https://user-images.githubusercontent.com/9594455/213690613-0ba6a764-bb74-4705-a073-0002a6a84d73.mov


app/assets/javascripts/admin/helpers/smart-poller.js
```diff
@@ -0,0 +1,19 @@
+(function (GOVUK) {
+  'use strict'
+
+  GOVUK.smartPoller = function (wait, polledFunction) {
```
DilwoarH, line 4: Is this used anywhere else?

We've generally moved away from attaching things to the global GOVUK scope. So it might be worth considering moving this into the module itself

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,60 @@
...
+  function BrokenLinksReport (module) {
+    this.module = module
+  }
+
+  BrokenLinksReport.prototype.init = function () {
+    if (!fetch) return
+    var button = this.module.querySelector('button')
+    var authenticityTokenField = this.module.querySelector('[name="authenticity_token"]')
+    if (button && authenticityTokenField) this.setupEventListener(button, authenticityTokenField.value)
+    var refreshLink = this.module.querySelector('.js-broken-links-refresh')
```
DilwoarH, line 14: Could we change this class name to `js-app-view-broken-links__refresh-link` or something similar?

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,60 @@
...
+    if (refreshLink) {
+      this.setupPolling(refreshLink.href)
+      refreshLink.remove()
+    }
+  }
+
+  BrokenLinksReport.prototype.setupEventListener = function (button, authenticityToken) {
+    var action = this.module.querySelector('form').action
```
DilwoarH, line 22: Should we target a class rather than a `form` tag? for example `js-app-view-broken-links__refresh-form`

My reasoning behind this would primarily be that we may introduce another form or change the form element to something else. Having an explicit JS reference class will help developers be wary that a JS listener is attached this.

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,60 @@
...
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ authenticity_token: authenticityToken })
+      }).then(function (response) { return response.json() })
+        .catch(function () { location.reload() })
+        .then(function (json) {
+          this.replaceContentsAndCopyClasses(json.html)
+          this.init()
+        }.bind(this))
+      event.preventDefault()
```
DilwoarH, line 34: Generally, if `event.preventDefault()` is used, it's usually added to the first line of the event function.

```js
button.addEventListener('click', function (event) {
      event.preventDefault()
      ....
})

```

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,60 @@
...
+      fetch(href + '.json').then(function (response) { return response.json() })
+        .then(function (json) {
+          if (json.inProgress) {
+            retry()
+          } else {
+            this.replaceContentsAndCopyClasses(json.html)
+            this.init()
+          }
+        }.bind(this))
+        .catch(function () { location.reload() })
```
DilwoarH, line 49: I don't think a reload is a good thing to do here because the user may be in the middle of editing something and would suddenly find the page abruptly refreshing.

app/assets/javascripts/admin/views/broken-links-report.js
dnkrj: Added 👍

app/assets/javascripts/admin/views/broken-links-report.js
dnkrj: The logic behind refreshing is that that's the default behaviour without javascript.
The button should only but present on the summary page so no changes should be lost.
What do you think?
Happy to keep or remove it, it feel quite unlikely that the promise will fail.

app/assets/javascripts/admin/helpers/smart-poller.js
dnkrj: Had a look and bootstrap only used it for link checking so have rolled it in 👍

app/assets/javascripts/admin/views/broken-links-report.js
DilwoarH: I think we notify the user that an error occurred rather than refreshing. It could be that we display a message saying an error occurred and then give them a button to refresh the page, but it should be a user decision.

I hope that makes sense.

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  BrokenLinksReport.prototype.init = function () {
+    if (!fetch) return
+    var button = this.module.querySelector('button')
+    var authenticityTokenField = this.module.querySelector('[name="authenticity_token"]')
+    if (button && authenticityTokenField) this.setupEventListener(button, authenticityTokenField.value)
+    var refreshLink = this.module.querySelector('.js-broken-links-refresh')
+    if (refreshLink) {
+      this.setupPolling(refreshLink.href)
```
DilwoarH, line 16: This passing in the href to start polling
```
      this.setupPolling(refreshLink.href)
```

but the function has two expected variables:

```
startPolling = function (wait, polledFunction)
```

where is this set?

app/assets/javascripts/admin/views/broken-links-report.js
dnkrj: As discussed I've updated the code to re-show the refresh button in the case that we fail 10 times and updated this to simply call `retry()`

app/assets/javascripts/admin/views/broken-links-report.js
dnkrj: `setupPolling` and `startPolling` were two different functions.
I've restructured to `setupPolling` and `poll` to simplify things.

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  BrokenLinksReport.prototype.setupPolling = function (refreshLink) {
+    refreshLink.hidden = true
+
+    var retries = 10
+    var retry = function () {
+      retries -= 1
+      if (retries === 0) {
+        refreshLink.hidden = false
```
DilwoarH, line 44: if the refresh link shows, will the context change to let the user know that they need to try again?

app/assets/javascripts/admin/views/broken-links-report.js
```diff
@@ -0,0 +1,76 @@
...
+          this.init()
+        }
+      }.bind(this))
+      .catch(function () { retry() })
+  }
+
+  BrokenLinksReport.prototype.replaceContentsAndCopyClasses = function (html) {
+    this.module.innerHTML = html
+    this.module.classList = this.module.firstChild.classList
+    this.module.firstChild.outerHTML = this.module.firstChild.innerHTML
```
DilwoarH, line 72: Is there a reason why we need to do this? Can we not just have a message placeholder div which we place?

DilwoarH:
Should we have some styling on the "In progress" view?

DilwoarH:
Other than a few questions, looks good to me 👍

app/assets/javascripts/admin/views/broken-links-report.js
dnkrj: This way the rendering is the same for in progress and the final report.

We need it in the template anyway since that will be displayed if the page is loaded while a report is being generated or javascript is not being used.

dnkrj:
Styling changes will be covered in https://trello.com/c/ALQzQdNT/1114-fixes-for-15-review-doc

app/assets/javascripts/admin/views/broken-links-report.js
DilwoarH: Is there a card for this?

app/assets/javascripts/admin/views/broken-links-report.js
dnkrj: What kind of context change would you suggest?
The important thing for the user is seeing the report which involves navigating to this part of the page either way.
It's only viewing the report that requires either waiting or clicking the "Check progress" link.

davidgisbey (approved):
Gave it a good test and it lgtm

PR #7282, opened 2023-01-24
Add tracking to the Edit Edition page

davidgisbey:
## Description

This works through this document https://docs.google.com/document/d/1E-FWgS_mc-UfRjtTGoPX7Gbher_6KKL5t3jjFlJ2Kmo/edit#heading=h.ezkkr7egiwe8 and adds missing tracking on the Edit Edition page.

This does the following:

1. adds subtype tracking as a custom dimension

image

2. Add tracking for adding an internal note

image

3. Add tracking to the save & save and continue buttons 

image

image

4. Ensure flash message warning is passed to GA even when flash isn't rendered

We only ever show 1 banner. This means that a flash warning isn't rendered on the page when a error summary banner is showing.

We still need the event passed so this ensures the error summary component is wrapped in a div which passes the flash warning message to GA

image

5. Add tracking for lead and supporting orgs

image

image

### Event for adding an image 

This will now be handled in the new images workflow

### Tracking of associations 

We are going to put selecting associations into blocked for now. The accessible autocomplete is a bit of a nightmare to track due to it's niche behaviour.

We can hack around it using keyup and click event listeners on the module itself, but the behaviour is complex and niche so any changes to the accessible autocomplete would likely break it. We're going to cycle back to this when the new PA starts

## Trello card

https://trello.com/c/ya9GBOuG/42-tracking-to-be-added-to-edit-edition-page

### DIFF
--- a/app/components/admin/editions/document_history_tab_component.html.erb
+++ b/app/components/admin/editions/document_history_tab_component.html.erb
@@ -4,7 +4,17 @@
   <p class="govuk-body">To add an internal note, save your changes.</p>
 <% end %>
 
-<p class="govuk-body"><%= link_to("Add internal note", new_admin_edition_editorial_remark_path(edition), class: "govuk-body govuk-link govuk-link--no-visited-state") %></p>
+<p class="govuk-body">
+  <%= link_to("Add internal note",
+    new_admin_edition_editorial_remark_path(edition),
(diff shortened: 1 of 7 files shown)


davidtrussler (approved):
Looking good! 👍

PR #7284, opened 2023-01-25
Adds required modules to allow miller-columns to render correctly

davidtrussler:
[Trello](https://trello.com/c/Run0Mkiq/1094-add-a-polyfill-for-custom-elements)

We have found that the newly-added miller columns on Whitehall are currently very broken in IE11 (nothing displays at all). This is due to lack of support for custom elements (this component uses a `miller-columns` element). The solution that has been used on Content Publisher is to use the [@webcomponents/custom-elements](https://www.npmjs.com/package/@webcomponents/custom-elements) polyfill. 

These changes implement that in Whitehall as well. 

During the process it became clear that even with this there were still some problems due to the use of the `Reflect` JS Object. In Content Publisher this is remedied with the [core-js-bundle](https://www.npmjs.com/package/core-js-bundle) module. That has also been added here as the quickest tried and tested method.

|Before|After|
|-|-|
|before|after|

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -2,6 +2,12 @@
 //= require govuk_publishing_components/all_components
 //= require govuk_publishing_components/analytics
 
+// support ES6 (promises, functions, etc. - see docs)
+//= require core-js-bundle/index.js
+
+// support ES6 custom elements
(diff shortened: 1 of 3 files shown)


DilwoarH (approved):

DilwoarH:
Nice work

PR #7288, opened 2023-01-26
Fix bug with history pagination on edit translation page

davidgisbey:
## Description 

At the moment, this is broken in the Bootstrap and GOV.UK Design System implementation.

It looks like it's not ever worked in Bootstrap. This is due to the controller attempting to find the edition by params[:id].

For translations this is the locale, so it blows up. This handles this by using params[:edition_id] if present.


davidtrussler (approved):
Good spot ... LGTM! 👍

PR #7298, opened 2023-01-30
Add images tab

jyoung-gds:
Trello ticket [link 1](https://trello.com/c/0M9vyh5i/1106-create-separate-images-page), [link 2](https://trello.com/c/AvguABf6/1056-new-image-tab), [link 3](https://trello.com/c/koVf63dk/1118-create-separate-images-page)

## What
This PR creates a separate resource for images, adding a controller and an index endpoint. It then uses this endpoint when creating a tab for adding/removing/editing images on the edit edition page, which is hidden behind a feature flag. 

**NOTE:** None of the UI is functional at the moment, this is just a backbone for the upcoming update to images workflow.

## Why
This is required for the upcoming work on updating the images workflow.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -270,6 +270,7 @@ Whitehall::Application.routes.draw do
             post :upload_zip, on: :collection
             get :set_titles, on: :member
           end
+          resources :images, controller: "edition_images", only: %i[index]
         end
 
         get "/editions/:id" => "editions#show"
--- a/test/functional/admin/edition_images_controller_test.rb
+++ b/test/functional/admin/edition_images_controller_test.rb
@@ -1,0 +1,4 @@
+require "test_helper"
+
+class Admin::EditionImagesControllerTest < ActionDispatch::IntegrationTest
+end
--- a/test/functional/admin/edition_images_controller_test.rb
+++ b/test/functional/admin/edition_images_controller_test.rb
@@ -1,4 +1,21 @@
 require "test_helper"
 
 class Admin::EditionImagesControllerTest < ActionDispatch::IntegrationTest
+  test "redirects users without the 'Preview images update' permission" do
+    edition = create(:draft_publication)
+    login_as create(:gds_editor)
+    get admin_edition_images_path(edition.id)
+    assert_equal 302, status
+    follow_redirect!
+    assert_equal "/government/admin/publications/#{edition.id}/edit", path
+  end
+
+  test "forbids unauthorised users from viewing the images index endpoint" do
+    edition = create(:draft_publication)
+    user = create(:world_editor)
+    user.permissions << "Preview images update"
+    login_as user
+    get admin_edition_images_path(edition.id)
+    assert_equal 403, status
+  end
 end
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -1,4 +1,28 @@
 class Admin::EditionImagesController < Admin::BaseController
-  def index
+  before_action :find_edition
+  before_action :redirect_unless_user_can_preview_images_update
+  before_action :enforce_permissions!
+  layout "design_system"
+
+  def index; end
+
+private
+
+  def find_edition
+    edition = Edition.find(params[:edition_id])
+    @edition = LocalisedModel.new(edition, edition.primary_locale)
+  end
+
+  def redirect_unless_user_can_preview_images_update
+    redirect_to edit_admin_edition_path(@edition) unless current_user.can_preview_images_update?
+  end
+
+  def enforce_permissions!
+    case action_name
+    when "index"
+      enforce_permission!(:see, @edition)
+    else
+      raise Whitehall::Authority::Errors::InvalidAction, action_name
+    end
   end
 end
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -1,0 +1,4 @@
+class Admin::EditionImagesController < Admin::BaseController
+  def index
+  end
+end
--- a/app/helpers/admin/tabbed_nav_helper.rb
+++ b/app/helpers/admin/tabbed_nav_helper.rb
@@ -1,19 +1,20 @@
 module Admin::TabbedNavHelper
   include Admin::EditionsHelper
 
-  def secondary_navigation_tabs_items(model, current_path)
+  def secondary_navigation_tabs_items(user, model, current_path)
     if model.is_a?(Edition)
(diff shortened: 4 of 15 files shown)


config/routes.rb
```diff
@@ -270,6 +270,7 @@ def external_redirect(path_prefix, target)
             post :upload_zip, on: :collection
             get :set_titles, on: :member
           end
+          resources :images
```
davidgisbey, line 273: generally when adding endpoints it's good to just add the ones you need as you need them. It might be that you end up needing new/create/edit/update etc., but for now i'd just add 

```
resources :images, only: %i[index]
```

app/controllers/admin/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def build_blank_image
+    if @edition.allows_image_attachments? && @edition.images.none?(&:new_record?)
+      image = @edition.images.build
+      image.build_image_data
+    end
+  end
+
+  def enforce_permissions!
```
davidgisbey, line 23: i think we need to enforce a similar permission to the edition summary page. This is from the editions controller

```
  def enforce_permissions!
    case action_name
    when "index", "topics", "history"
      enforce_permission!(:see, edition_class || Edition)
    when "show"
      enforce_permission!(:see, @edition)
    when "new"
      enforce_permission!(:create, edition_class || Edition)
    when "create"
      enforce_permission!(:create, @edition)
    when "edit", "update", "revise", "diff", "update_bypass_id"
      enforce_permission!(:update, @edition)
    when "destroy", "confirm_destroy"
      enforce_permission!(:delete, @edition)
    when "export", "confirm_export"
      enforce_permission!(:export, edition_class || Edition)
    else
      raise Whitehall::Authority::Errors::InvalidAction, action_name
    end
  end
```

So i think we just need 

```
enforce_permission!(:see,  Edition)
```

then for the new/edit/create/update actions later it would be 

```
:create, Edition
```

app/controllers/admin/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    unless current_user.can_preview_images_update?
+      redirect_to edit_admin_edition_path(@edition)
```
davidgisbey, line 25: I think it would be good to move this into a separate before action which lives at the top of the before_actions.

Something like

```
before_action :find_edition
before_action :redirect_unless_user_can_preview_image_update
before_action :enforce_permissions!
before_action :build_blank_image
layout "design_system"


def redirect_unless_user_can_preview_image_update
  redirect_to edit_admin_edition_path(@edition) unless current_user.can_preview_images_update?
end
```

app/controllers/admin/images_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Admin::ImagesController < Admin::BaseController
+  before_action :find_edition
```
davidgisbey, line 2: Is this controller always going to be coupled to images that belong to an edition? It seems likely that's the case for now as we use mounts for images on other objects (promotional features etc.)

I'm assuming that will be the case here as the designs are so customised for the editions flow (has it's own tab etc.)

Just double checking as `find_edition` couples this controller to editions.

If that is the case though maybe we should call this `Admin::EditionImagesController` so it's a bit more explicit that it's only going to be used for images as lots of other objects have them too

davidgisbey:
Great work Johnny. I've left a couple of comments but no big blockers 🚀

app/views/admin/images/_uploaded_images.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Lead image",
+    font_size: "m",
+    padding: true
+  } %>
+
+  <%= render "govuk_publishing_components/components/details", {
+        title: "Using a lead image",
+      } do %>
+    <p>Using a lead image is optional. To use a lead image either select the
```
dnkrj, line 15: Worth using class="govuk-body" for the paragraphs here and below.

jyoung-gds:
@davidgisbey @dnkrj I've pushed some changes up to address both of your comments so this is ready for a re-review. I took out the `build_blank_image` method in the controller because it wasn't needed, it was something I mistakenly left in whilst trying to figure out how images currently work (it may be needed at some point but not right now).

davidgisbey:
Looks great Johnny. Very happy to approve if you'd like to get this merged in and then iterate on it.

test/functional/admin/edition_images_controller_test.rb
```diff
@@ -0,0 +1,4 @@
+require "test_helper"
+
+class Admin::EditionImagesControllerTest < ActionDispatch::IntegrationTest
+end
```
ollietreend, line 4: It would be nice to cover the `redirect_unless_user_can_preview_images_update` and `enforce_permissions!` behaviour here.

There is some prior art for testing permissions in other controllers, for example the attachments controller test:

https://github.com/alphagov/whitehall/blob/4e087da57a78a4fb29cb0ae5b57db91e142e38e8/test/functional/admin/attachments_controller_test.rb#L442-L446

But I don't think this test will copy-and-paste nicely. 😕  The attachments controller test is a 'functional style' test that inherits from `ActionController::TestCase`. This type of test [is discouraged nowadays](https://api.rubyonrails.org/classes/ActionController/TestCase.html).

https://github.com/alphagov/whitehall/blob/4e087da57a78a4fb29cb0ae5b57db91e142e38e8/test/functional/admin/attachments_controller_test.rb#L3

Whereas the images controller test inherits from `ActionDispatch::IntegrationTest` because it's an 'integration style' test. This is the recommended approach and default Rails behaviour when creating new controller tests, which is how you've ended up with it here.

https://github.com/alphagov/whitehall/blob/f506f75b0d95d132558980da7dfa9c6f08239971/test/functional/admin/edition_images_controller_test.rb#L3

These different types of tests operate in slightly different contexts, so the way you write tests differs too.

I think something like this should work:

```ruby
edition = create(:draft_case_study)
login_as create(:world_editor)
get admin_edition_images_path(edition.id)
assert_equal 403, status
```

ollietreend (commented):
Nice work @jyoung-gds 👍🏻 

I've just added a couple of things – one is small, and the other is (hopefully!) small-ish.

jyoung-gds:
@davidgisbey yes the plan is to iterate on this and implement things like showing images currently on an edition etc. once this is merged. I think this will speed things up because it will allow more asynchronous working earlier on. I've written up some trello cards for the remaining work

@ollietreend thanks for review! I'll make those changes today

ollietreend (approved):
Great work @jyoung-gds ✨🚀

PR #7307, opened 2023-02-01
Prepare 1.5 release 

davidgisbey:
**Merging this will trigger the 1.5 release. Do not merge until comms has gone out**

## Description

This PR does two things:

1. Temporarily forces the 'next release' of Design System to be shown to everybody
2. Fixes all the tests that fail as a result of it

A follow up PR will come after the release that will clean up all the old code.

## Trello card

https://trello.com/c/2fLcyBrE/1-prepare-15-release-force-permissions-and-fix-tests

### DIFF
--- a/app/views/components/_miller-columns-list.html.erb
+++ b/app/views/components/_miller-columns-list.html.erb
@@ -3,7 +3,7 @@
     <li>
       <div class="govuk-checkboxes__item">
         <input class="govuk-checkboxes__input" type="checkbox" name="<%= name %>" value="<%= item[:value] %>" id="<%= id %>-<%= index %>"
-          <% if item[:checked].present? && item[:checked] %> checked <% end %>
+          <% if check_item?(item) %> checked <% end %>
         >
         <label class="govuk-label govuk-checkboxes__label" for="<%= id %>-<%= index %>">
           <%= item[:label] %>
--- a/test/integration/shareable_preview_generate_new_link_test.rb
+++ b/test/integration/shareable_preview_generate_new_link_test.rb
@@ -23,17 +23,21 @@ class ShareablePreviewGenerateNewLinkIntegrationTest < ActionDispatch::Integrati
       test "it revokes the previous links and generates a new one" do
         get admin_case_study_path(edition)
 
+        find(".govuk-details__summary", text: "Share document preview").click
+
         current_preview_url = all("input").first.value
         current_query_string = Rack::Utils.parse_query URI(current_preview_url).query
         current_token = current_query_string["token"]
 
-        click_link "Generate new link"
+        click_button "Generate new link"
+
+        find(".govuk-details__summary", text: "Share document preview").click
 
         new_preview_url = all("input").first.value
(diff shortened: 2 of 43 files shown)


test/integration/shareable_preview_generate_new_link_test.rb
```diff
@@ -22,18 +22,20 @@ class ShareablePreviewGenerateNewLinkIntegrationTest < ActionDispatch::Integrati
 
       test "it revokes the previous links and generates a new one" do
         get admin_case_study_path(edition)
+        find(".govuk-details__summary").click
```
ollietreend, line 25: I'm wondering if this would work and be more readable – (and more future proof if we add a second 'details' component to the page down the line)

```suggestion
        find("summary", text: "Share document preview").click
```

(and ditto for where it's used on line 33)

app/views/components/_miller-columns-list.html.erb
```diff
@@ -3,7 +3,14 @@
...
         <input class="govuk-checkboxes__input" type="checkbox" name="<%= name %>" value="<%= item[:value] %>" id="<%= id %>-<%= index %>"
-          <% if item[:checked].present? && item[:checked] %> checked <% end %>
+          <% if item[:checked].present? &&
+                  item[:checked] ||
+                    item[:items].any? do |child|
+                      child[:checked] ||
+                      child[:items].any? do |grandchild|
+                        grandchild[:checked]
+                      end
+                    end %> checked <% end %>
```
ollietreend, line 13: It'd be nice if this logic existed in a helper method, rather than directly in the view – although I'm not sure there's an obvious place for it short of creating a new Miller Columns helper.

ollietreend: Also: is this fixed regression covered by a test? If not, does it need to be? (Maybe not – just asking!)

ollietreend (commented):
This is looking great @davidgisbey ✨ 

Just a couple of small comments/questions.

test/integration/shareable_preview_generate_new_link_test.rb
davidgisbey: yea good shout. Will update it

app/views/components/_miller-columns-list.html.erb
davidgisbey: Okay cool i'll add a miller columns helper since it's the place that makes most sense to me.

It's actually covered in a controller spec https://github.com/alphagov/whitehall/blob/main/test/functional/admin/edition_world_tags_controller_test.rb#L50-L57

```
  view_test "should check a child taxon and its parents when only a child taxon is selected" do
    stub_publishing_api_links_with_taxons(@edition.content_id, [world_grandchild_taxon_content_id])

    get :edit, params: { edition_id: @edition }

    assert_select "input[value='#{world_child_taxon_content_id}'][checked='checked']"
    assert_select "input[value='#{world_grandchild_taxon_content_id}'][checked='checked']"
  end

```

PR #7315, opened 2023-02-03
Remove 1.5 legacy code

davidgisbey:
## Description

Now released 1.5 and are need to go through and remove all the legacy code for the Bootstrap implementation. This goes through and pulls out legacy views, removes css styling & any related JS.

It also does a bit of tidying up of the fact check request/editorial remark code so it's BEM compliant.

## Trello

https://trello.com/c/G0S7YKYF/5-post-15-release-cleanup


app/controllers/admin/editions_controller.rb
```diff
@@ -205,8 +202,8 @@ def get_layout
     end
   end
 
-  def fetch_version_and_remark_trails(next_release: true)
-    if preview_design_system?(next_release:)
+  def fetch_version_and_remark_trails(govuk_design_system: false)
+    if govuk_design_system
```
ollietreend, line 206: **No need to change this**, but it's just occurred to me that we probably don't need any parameters on this method since we could just call `get_layout` to work that out (making that the single source of truth).

i.e.

```suggestion
  def fetch_version_and_remark_trails
    if get_layout == "design_system"
```

app/views/admin/editions/_standard_elements.html.erb
```diff
@@ -2,14 +2,16 @@
...
+<div dir="<%= "rtl" if edition.rtl? %>">
+  <h2 class="app-view-editions-edition__document__title govuk-heading-l"><%= edition.title %></h2>
 
-<% unless edition.summary.blank? %>
-  <div class="summary">
-    <p class="govuk-body"><em><%= edition.summary %></em></p>
-  </div>
-<% end %>
+  <% unless edition.summary.blank? %>
+    <div class="app-view-editions-edition__document__summary">
```
ollietreend, line 9: Nice work BEMing these class names! ⭐

app/controllers/admin/editions_controller.rb
davidgisbey: great shout. Will update that now

PR #7316, opened 2023-02-03
Updates `input` and `textarea` elements on edit translations page for `dir` attribute

davidtrussler:
[Trello](https://trello.com/c/peLddSGk/69-add-rtl-to-translations)

This PR adds a value for the `dir` attribute in the title, summary and body fields of the Edit Translation page. This ensures that the language displays correctly where it is a right-to-left displaying language.

It roughly follows the way the definition for `rtl_locale?` is done on the attachments page but could do with some checking from a ruby perspective. Otherwise it adds the values using the newly-updated input and textarea components.

|LTR language|RTL language|
|-|-|
|Screenshot 2023-02-03 at 12 42 03|Screenshot 2023-02-03 at 12 47 11|

### DIFF
--- a/app/views/admin/edition_translations/edit.html.erb
+++ b/app/views/admin/edition_translations/edit.html.erb
@@ -18,6 +18,8 @@
           heading_size: "l",
           value: @translated_edition.title,
           error_items: errors_for(form.object.errors, :title),
+          right_to_left: form.object.translation_rtl?,
+          right_to_left_help: false
         } %>
 
         <h3 class="app-view-edition-translation__english-content govuk-heading-m govuk-!-margin-bottom-2">English title content:</h3>
@@ -34,6 +36,8 @@
         value: @translated_edition.summary,
         rows: 2,
         error_items: errors_for(form.object.errors, :summary),
+        right_to_left: form.object.translation_rtl?,
+        right_to_left_help: false
       } %>
 
       <h3 class="app-view-edition-translation__english-content govuk-heading-m govuk-!-margin-bottom-2">English summary content:</h3>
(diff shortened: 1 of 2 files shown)


app/views/admin/edition_translations/edit.html.erb
```diff
@@ -18,6 +18,8 @@
           heading_size: "l",
           value: @translated_edition.title,
           error_items: errors_for(form.object.errors, :title),
+          right_to_left: form.object.rtl_locale?(translation_locale),
```
davidgisbey, line 21: I think we can use this method https://github.com/alphagov/whitehall/blob/main/app/models/edition/translatable.rb#L51-L53

So  it would just be `form.object.rtl?`

davidtrussler: Actually, it seems that method returns a value based on the `primary_local`, which is always `en`. I can add a new method like below which does what is required. Does that seem like a better option? 

```
def translation_rtl?
  Locale.find_by_code(translation_locale.code).rtl?
end
```

davidgisbey: Oh you're absolutely correct. I think you were bang on before then, but i think it should probably live in the translatable concern next to the `rtl?` method. Wdyt?

davidtrussler: Or alternatively
```
def translation_rtl?
  Locale.new(translation_locale).rtl?
end
```

davidtrussler: @davidgisbey I've updated that to add the new method in translatable and use that in the view. I'll squash the commits if that looks OK.

davidgisbey (approved):
great work. LGTM!

PR #7317, opened 2023-02-03
Remove rendering of Past Prime Ministers pages

brucebolt:
Once rendering of the Past Prime Ministers index page and the individual Past Prime Minister pages have been migrated to Collections, the rendering can be removed from Whitehall.

[Trello card](https://trello.com/c/daPwZeIO)


BeckaL (approved):
This looks good to me. I do notice that we've actually progressed more on the individual PM pages although maybe the index page will catch up. Possibly worth this PR just waiting for all of that to be done and removing everything in one go.

BeckaL (approved):
This all looks good to me 🥳 . Looks to me like everything for past pms is cleaned up here. Having a general look through whitehall for historic stuff, I found [one additional file](https://github.com/alphagov/whitehall/blob/main/app/assets/stylesheets/frontend/views/_history-people.scss) that looks like it can also be removed (I think it was just used for foreign secretaries so could have gone before)

brucebolt:
> Having a general look through whitehall for historic stuff, I found [one additional file](https://github.com/alphagov/whitehall/blob/main/app/assets/stylesheets/frontend/views/_history-people.scss) that looks like it can also be removed (I think it was just used for foreign secretaries so could have gone before)

Good find.  I'll add that as an extra commit here as it's somewhat related.

BeckaL (approved):

PR #7319, opened 2023-02-03
Use edition.format_name in email notifications

ollietreend:
There were a few places where we just downcased the edition type instead of using its #format_name method.

This resulted in multi-word edition types from rendering incorrectly in emails.

For example, 'detailed guide' was being rendered as 'detailedguide'.

---


davidgisbey (approved):
lgtm

PR #7325, opened 2023-02-06
Remove ability to run broken links report from Edit Edition page

davidgisbey:
## Description

In Bootstrap you can only run the broken links report from the Edition Summary page.

Currently, in the GOV.UK Design System we always allow this.

This adds in a bit of functionality to hide the run report buttons if a
`render_broken_links_button: false` is passed into the partial.

## Trello card

https://trello.com/c/KLzIk1bW/70-broken-link-checker-on-edit-page

### DIFF
--- a/app/views/admin/editions/_govspeak_help.html.erb
+++ b/app/views/admin/editions/_govspeak_help.html.erb
@@ -6,7 +6,7 @@
   link_check_report ||= nil
 %>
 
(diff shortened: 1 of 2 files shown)


beccapearce (approved):

dnkrj:
This won't cover the case when the user loads the edit page while the report is still in progress. The JS will replace the content of the report from the server with the button in it.
The simplest way to do this would probably be to add a CSS rule to hide the button, otherwise the JS would need to be updated.

PR #7330, opened 2023-02-08
Add filter to the history tab

davidgisbey:
## Description

Based on feedback of merging history and notes we've decided to provide users with the option to filter to see the following:

1. history and notes
2. history
3. notes

This PR updates the `PaginatedTimeline` model to be able to take an optional filter param. If "internal notes" is passed in it'll create a timeline that only includes `EditorialRemarks`. And if "history" is passed in it will only return `Document::PaginatedHistory::AuditTrail` objects.

It then adds a select with the available options and fires off an event listener on change to refresh the tabs content with the users selection.

It handles pagination correctly, so if you click next page you continue to only see the filtered content.

## Screenshots

image

image

## Trello card

https://trello.com/b/MKAU3hcD/govuk-publishing-service-whitehall-transition


test/components/admin/editions/document_history_tab_component_test.rb
```diff
@@ -16,6 +16,9 @@ class Admin::Editions::DocumentHistoryTabComponentTest < ViewComponent::TestCase
                    </nav>".html_safe
 
     Admin::Editions::DocumentHistoryTabComponent.any_instance.stubs(:paginate).returns(pagination)
+
+    params = ActionController::Parameters.new(controller: "admin/editions", action: "show", id: @document.latest_edition.id)
+    Admin::Editions::DocumentHistoryTabComponent.any_instance.stubs(:params).returns(params)
```
davidgisbey, line 21: having to stub this feels like a bit of a smell to me. Should probs be passed in as an arg. Although i think it could probs be picked up in a follow up PR. Wdyt?

ollietreend: @davidgisbey Yeah, I agree this feels like a bit of a smell.

What about if you used `@document_history.filter` as the 'source of truth' for the currently selected filter option? That feels like it could be a nicer way to handle it, because then the view doesn't need to 'know' how to retrieve that value from the controller request params directly.

davidgisbey: So this is actually blowing up due to the helper method that constructs the remote pagination url using the params, rather than params being used to determine state anywhere inside the view component.

ollietreend: I managed to resolve this by using the built-in route helper method `document_history_admin_edition_path` rather than the custom method `paginated_document_history_url`.

This removes our dependency on having access to the controller `params`, since we already have everything we need to generate a link with the correct URL params:

- `page` comes from `current_page`
- `editing` is a local variable
- `only` is available from `@document_history.only`

ollietreend (approved):

PR #7333, opened 2023-02-09
Populate content item for past prime ministers index

callumknights:
In preparation for moving the rendering for the [past prime ministers index page](https://www.gov.uk/government/history/past-prime-ministers) out of whitehall, we need to publish a content item for this page.

This PR adds a presenter for this page, behaviour to ensure that the page is republished when a relevant update to a model is made, and a rake task to allow us to republish the page.

Depends on [related addition to content schemas](https://github.com/alphagov/publishing-api/pull/2285) - github actions won't pass until this is merged in, but I've pointed the jenkins run at the publishing api branch, so this should be green.

[Trello](https://trello.com/c/0NcK4ojz/420-populate-content-item-for-past-prime-ministers-index)

### DIFF
--- a/app/presenters/publishing_api/historical_accounts_index_presenter.rb
+++ b/app/presenters/publishing_api/historical_accounts_index_presenter.rb
@@ -19,7 +19,9 @@ module PublishingApi
 
       content.merge!(
         base_path:,
-        details: {},
+        details: {
+          appointments_without_historical_accounts:,
+        },
         document_type: "historic_appointments",
         public_updated_at: Time.zone.now,
         rendering_app: Whitehall::RenderingApp::WHITEHALL_FRONTEND,
@@ -27,6 +29,22 @@ module PublishingApi
       )
 
       content.merge!(PayloadBuilder::Routes.for(base_path))
+    end
+
+    def appointments_without_historical_accounts
+      role = Role.friendly.find("prime-minister")
+      people_to_present = (role.role_appointments.historic.map(&:person) - HistoricalAccount.all.map(&:person)).uniq
+      people_to_present.map do |person|
+        { title: person.name,
+          dates_in_office:
+            person.role_appointments.historic.map do |appointment|
+              {
+                start_year: appointment.started_at.year,
+                end_year: appointment.ended_at.year,
+              }
+            end,
+          image_url: person.image_url }
+      end
     end
 
     def base_path
--- a/app/presenters/publishing_api/historical_accounts_index_presenter.rb
+++ b/app/presenters/publishing_api/historical_accounts_index_presenter.rb
@@ -1,0 +1,42 @@
+module PublishingApi
+  class HistoricalAccountsIndexPresenter
+    attr_accessor :update_type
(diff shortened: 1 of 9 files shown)


app/presenters/publishing_api/historical_accounts_index_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+        schema_name: "historic_appointments",
+      )
+
+      content.merge!(PayloadBuilder::Routes.for(base_path))
+    end
+
+    def find_appointments_without_accounts(role)
+      past_pms = role.role_appointments
+      ids_of_pms_without_accounts = past_pms.map(&:person_id) - HistoricalAccount.all.map(&:person_id)
+      Person.where(id: ids_of_pms_without_accounts)
```
brucebolt, line 37: Rails can do the association stuff for you, without needing to convert IDs to people:
```suggestion
      role.role_appointments.map(&:person) - HistoricalAccount.all.map(&:person)
```

If you change this, it might be worth just inlining it on lines 41-42 instead of being a method.

app/presenters/publishing_api/historical_accounts_index_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    def appointments_without_historical_accounts
+      role = Role.friendly.find("prime-minister")
+      people_to_present = find_appointments_without_accounts(role)
+      people_to_present.map do |person|
+        previous_role_appointments = RoleAppointment.where(person_id: person.id)
```
brucebolt, line 44:
```suggestion
        previous_role_appointments = person.role_appointments
```

app/presenters/publishing_api/historical_accounts_index_presenter.rb
```diff
@@ -0,0 +1,69 @@
...
+                                                    .where.not(RoleAppointment::CURRENT_CONDITION)
```
brucebolt, line 45: I wonder if it's worth creating a `historic` scope of the `RoleAppointment` model (since we already have one called `current`, doing the inverse of this)?

Something like:
```
 scope :historic, -> { where.not(CURRENT_CONDITION) }
```

Then this could be updated to, then in-lined on line 49:
```
previous_role_appointments = person.role_appointments.historic
```

brucebolt (commented):
Looks good to me.  There's just a few small Rails-y things that'll make the code simpler to read that I've commented on in one commit.

app/presenters/publishing_api/historical_accounts_index_presenter.rb
BeckaL: Good suggestion. Also made me realise that I wasn't actually just presenting historic role appointments here, meaning that this errored if someone had both historic and current appointments. Amended the test to add this case in too

brucebolt (approved):
Thanks for making those changes, the code is much easier to read now.

All looks good except a typo in one of the commit messages (`appoointments`, instead of `appointments`).

PR #7336, opened 2023-02-10
Adds support for RTL languages on Foreign language only consultations

davidtrussler:
[Trello](https://trello.com/c/aCuYoaQX/72-add-rtl-support-for-foreign-language-only-consultations)

These changes add support for right-to-left languages to the Foreign language only consultation page, for example [this page](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/consultations/661078/edit). When a consultation document is being edited or created the user has the option to select “Foreign language only consultation” and choose a language for the document. Where the language selected is a RTL language the appropriate attribute should be updated in the relevant fields.

Screenshot 2023-02-10 at 10 31 28

The work here
- Adds the correct value to the `dir` attribute when the page is loaded
- Updates the existing `LocaleSwitcher` module so that it can be used more widely than the page on which it is currently used. 
- Adds that module to this page so that the changes are made dynamically
- Updates the attachment page to use the more generic module.

### DIFF
--- a/app/assets/javascripts/admin/views/locale-switcher.js
+++ b/app/assets/javascripts/admin/views/locale-switcher.js
@@ -1,10 +1,20 @@
+/* This module is used where only certain elements within a form are required to update the value of their `dir` attribute in response to a change in a `select` element.
+** By default it will affect this change on `input` and `textarea` elements but can be extended to other elements as below.
+** Usage:
+** - Add to the form with `data-module="LocaleSwitcher"`
+** - Add collection of RTL languages on the form element with `data-rtl-locales=<array of LTR languages>`
+** - Wrap the language selector with the class `js-locale-switcher-selector`
+** - Wrap fields to display RTL languages with `js-locale-switcher-field`
+** - Add to any custom elements the class `js-locale-switcher-custom`
+*/
+
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
   function LocaleSwitcher (module) {
     this.module = module
-    this.rightToLeftLocales = module.dataset.rtlLocales.split(' ')
+    this.rightToLeftLocales = JSON.parse(module.dataset.rtlLocales)
   }
 
   LocaleSwitcher.prototype.init = function () {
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/admin/views/locale-switcher.js
```diff
@@ -14,24 +24,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
 
-    select.addEventListener('change', function () {
-      if (rightToLeftLocales.indexOf(this.value) > -1) {
-        title.setAttribute('dir', 'rtl')
-        body.setAttribute('dir', 'rtl')
-        preview.setAttribute('dir', 'rtl')
+    select.addEventListener('change', function (event) {
+      var value = event.target.value
+
+      if (rightToLeftLocales.indexOf(value) > -1) {
```
ollietreend, line 37: ### Opinion (no action needed - feel free to ignore)

A nicer way to write this might be:

```suggestion
      if (rightToLeftLocales.includes(value)) {
```

However, `Array.prototype.includes` is not supported by IE11. I'm not sure if we've 'officially' dropped JS in IE11 yet (might be worth checking with @dnkrj). So I also understand if you'd rather leave this as is.

app/views/components/_govspeak-editor.html.erb
```diff
@@ -70,11 +69,12 @@
...
       describedby: hint_id,
-      right_to_left: right_to_left
+      right_to_left: right_to_left,
+      right_to_left_help: false
     } %>
   </div>
 
-  <%= tag.div class: "app-c-govspeak-editor__preview", 'aria-live': "polite", id: preview_id, dir: dir do %>
+  <%= tag.div class: "app-c-govspeak-editor__preview js-locale-switcher-custom", 'aria-live': "polite", id: preview_id, dir: dir do %>
     <p class="govuk-body">Generating preview, please wait.</p>
```
ollietreend, line 78: Should this `<p>` have an explicit `dir="ltr"` for the sake of completeness?

```suggestion
    <p class="govuk-body" dir="ltr">Generating preview, please wait.</p>
```

So that this part of the UI still displays correctly, even if the overall document is in an RTL language.

ollietreend (commented):
Nice work @davidtrussler – it's great to have this as a reusable module 🎉 

I've left a couple of comments, but nothing huge. I'm also happy to approve this PR as is if you don't think they're worth addressing.

app/assets/javascripts/admin/views/locale-switcher.js
davidtrussler: Yes, I agree `includes` is more elegant but I think I'd keep the IE11 support here for the minute. From what I can see we're not using `includes` in this context anywhere else yet in Whitehall.

app/views/components/_govspeak-editor.html.erb
davidtrussler: Yes, I've added that - good spot.

PR #7338, opened 2023-02-10
Render uploaded images with details in their tab

dnkrj:

# What
[Renders the images attached to an edition in the new images tab and adds a new CSS file to style the layout.
](https://trello.com/c/cb3oyumy/1135-add-display-uploaded-images-functionality-to-images-tab)
# Screenshots

whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPad Pro) (1)

whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPhone 12 Pro) (1)

### DIFF
--- a/app/views/admin/edition_images/_uploaded_images.html.erb
+++ b/app/views/admin/edition_images/_uploaded_images.html.erb
@@ -38,3 +38,25 @@
     margin_top: 0
   } %>
 </div>
+
+<% if edition.images %>
+  <ul class="govuk-list">
+    <% edition.images.each_with_index do |image, index| %>
+      <li class="govuk-grid-row">
+        <div class="govuk-grid-column-one-third">
+          <% if image.url.present? %>
+            <img src="<%= image.url %>" alt="Image <%= index + 1 %>" class="app-view-edition-images__image">
+          <% end %>
+        </div>
+        <ul class="app-view-edition-images__details govuk-grid-column-two-thirds govuk-list govuk-list--bullet">
+          <li><strong>Caption: </strong><%= image.caption.blank? ? "None" : image.caption %></li>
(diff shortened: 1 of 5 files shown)


app/views/admin/edition_images/_uploaded_images.html.erb
```diff
@@ -38,3 +38,25 @@
...
+          <% end %>
+        </div>
+        <ul class="app-view-edition-images__details govuk-grid-column-two-thirds govuk-list govuk-list--bullet">
+          <li><strong>Caption: </strong><%= image.caption.blank? ? "None" : image.caption %></li>
+          <% if image.alt_text %><li><strong>Alt text: </strong><%= image.alt_text.blank? ? "None" : image.alt_text %></li><% end %>
+          <li><strong>Markdown code: </strong><span>[Image:&nbsp;<%=image.image_data.carrierwave_image %>]</span></li>
+        </ul>
+      </li>
+      <% if image != edition.images.last %>
+        <li aria-hidden="true"><hr class="app-view-edition-images__section-break govuk-section-break govuk-section-break--m govuk-section-break--visible"></li>
```
davidtrussler, line 58: I just wondered if this was the best way to add a separator between rows (as opposed to just setting a bottom border) as it adds an otherwise unnecessary element. I can see it works better with the grid though so it's not a huge deal I guess.

dnkrj: Yeah, it's unfortunate. We need a separate element so that the edges line up with the images.
I tried putting the `hr` in the previous `li` element but because of the columns it has the same issues as using `border-bottom`.

davidtrussler: Sure, OK, let's live with it then.

davidtrussler (approved):
LGTM 👍

PR #7340, opened 2023-02-10
Add page to allow images to be deleted

dnkrj:

# What
[Adds routes for destroy and confirm destroy and defines methods on the controller and a template to handle image deletion.](https://trello.com/c/b1CsjEWj/1057-deleting-an-image)

# Screenshots
whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPad Pro)

whitehall-admin dev gov uk_government_admin_editions_1370963_images_445052_confirm_destroy(iPad Pro)

whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPad Pro) (2)

whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPhone 12 Pro)

### DIFF
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -6,7 +6,24 @@ class Admin::EditionImagesController < Admin::BaseController
 
   def index; end
 
+  def confirm_destroy; end
+
+  def destroy
+    filename = image.image_data.carrierwave_image
+    image.destroy!
+    redirect_to admin_edition_images_path(@edition), notice: "#{filename} has been deleted"
+  end
+
 private
+
+  def image
+    @image ||= find_image
+  end
+  helper_method :image
+
+  def find_image
+    @edition.images.find(params[:id]) if params[:id]
+  end
 
   def find_edition
     edition = Edition.find(params[:edition_id])
@@ -21,6 +38,8 @@ private
     case action_name
     when "index"
       enforce_permission!(:see, @edition)
+    when "destroy", "confirm_destroy"
+      enforce_permission!(:update, @edition)
     else
       raise Whitehall::Authority::Errors::InvalidAction, action_name
     end
--- a/app/views/admin/edition_images/_uploaded_images.html.erb
+++ b/app/views/admin/edition_images/_uploaded_images.html.erb
@@ -53,6 +53,9 @@
           <% if image.alt_text %><li><strong>Alt text: </strong><%= image.alt_text.blank? ? "None" : image.alt_text %></li><% end %>
           <li><strong>Markdown code: </strong><span>[Image:&nbsp;<%=image.image_data.carrierwave_image %>]</span></li>
         </ul>
+        <div class="app-view-edition-images__actions govuk-grid-column-full">
+          <%= link_to("Delete image", confirm_destroy_admin_edition_image_path(@edition, image), class: "govuk-link gem-link--destructive") %>
+        </div>
       </li>
       <% if image != edition.images.last %>
         <li aria-hidden="true"><hr class="app-view-edition-images__section-break govuk-section-break govuk-section-break--m govuk-section-break--visible"></li>
--- a/features/edition-images.feature
+++ b/features/edition-images.feature
@@ -18,5 +18,16 @@ Feature: Images tab on edit edition
     Given I am a writer
     And I have the "Preview images update" permission
     And a draft document with images exists
-    When I visit the images tab of the the document with images
-    Then I should see a list of the images
+    When I visit the images tab of the document with images
+    Then I should see a list with 2 images
+
+  Scenario: Images can be deleted from the images tab
+    Given I am a writer
+    And I have the "Preview images update" permission
+    And a draft document with images exists
+    When I visit the images tab of the document with images
+    Then I should see a list with 2 images
+    When I click to delete an image
+    And I confirm the deletion
+    Then I should see a success banner
+    And I should see a list with 1 image
--- a/app/assets/stylesheets/admin/views/_edition-images.scss
+++ b/app/assets/stylesheets/admin/views/_edition-images.scss
@@ -10,3 +10,7 @@
 .app-view-edition-images__details {
   padding-left: govuk-spacing(6);
 }
+
+.app-view-edition-images__actions {
+  margin-top: govuk-spacing(2);
+}
--- a/app/views/admin/edition_images/confirm_destroy.html.erb
+++ b/app/views/admin/edition_images/confirm_destroy.html.erb
@@ -1,0 +1,21 @@
+<% content_for :context, @edition.title %>
+<% content_for :page_title, "Delete image" %>
+<% content_for :title, "Delete image" %>
+<% content_for :title_margin_bottom, 6 %>
+
+<div class="govuk-grid-row">
+  <section class="govuk-grid-column-two-thirds">
+    <%= form_tag [:admin, :edition, :image], id: image, method: :delete do %>
+      <p class="govuk-body govuk-!-margin-bottom-7">Are you sure you want to delete <%= image.image_data.carrierwave_image %>?</p>
+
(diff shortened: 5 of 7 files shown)


app/controllers/admin/edition_images_controller.rb
```diff
@@ -6,7 +6,22 @@ class Admin::EditionImagesController < Admin::BaseController
 
   def index; end
 
+  def confirm_destroy; end
+
+  def destroy
+    image.destroy!
+    redirect_to admin_edition_images_path(@edition), notice: "#{image.image_data.carrierwave_image} has been deleted"
+  end
```
ollietreend, line 14: My instinct tells me that this is the sort of functionality we should be covering with tests. It's always easier to add tests at the beginning rather than retro-fit them later.

However, I'm struggling to find good examples of other tests in Whitehall which cover the deletion journey. I was hoping there'd be some decent feature specs for attachments, since they follow a similar 'CRUD' workflow. But annoyingly the [edition-attachments.feature](https://github.com/alphagov/whitehall/blob/7ccf59cae75651bd9f295fd305a446c0c3b62f58/features/edition-attachments.feature) spec stops short of testing the deletion journey.

Despite not having a solid example to follow, I still think it'd be good to have the 'D' part of this 'CRUD' journey covered by a feature test. What do you think? 🤔

app/views/admin/edition_images/_uploaded_images.html.erb
```diff
@@ -38,3 +38,28 @@
     margin_top: 0
   } %>
 </div>
+
+<% if edition.images %>
+  <ul class="govuk-list">
+    <% edition.images.each_with_index do |image, index| %>
+      <li class="app-view-edition-images__row govuk-grid-row">
```
ollietreend, line 45: I notice you've given this `<li>` the class name `app-view-edition-images__row`, but it doesn't look like you've used it anywhere (e.g. there are no CSS or JS selectors for it).

Does this class name serve a purpose?

app/views/admin/edition_images/_uploaded_images.html.erb
```diff
@@ -38,3 +38,28 @@
...
+          <% if image.alt_text %><li><strong>Alt text: </strong><%= image.alt_text.blank? ? "None" : image.alt_text %></li><% end %>
+          <li><strong>Markdown code: </strong><span>[Image:&nbsp;<%=image.image_data.carrierwave_image %>]</span></li>
+        </ul>
+        <div class="app-view-edition-images__actions govuk-grid-column-full">
+          <%= link_to("Delete image", confirm_destroy_admin_edition_image_path(@edition, image), class: "govuk-link gem-link--destructive") %>
+        </div>
+      </li>
+      <% if image != edition.images.last %>
+        <li aria-hidden="true"><hr class="app-view-edition-images__section-break govuk-section-break govuk-section-break--m govuk-section-break--visible"></li>
+      <% end %>
```
ollietreend, line 62: It seems like this `<hr>` exists purely for styling, rather than having a semantic meaning. I'm assuming this based on the use of `aria-hidden` attribute to hide it from accessibility tools.

In which case, is there a reason why you didn't consider implementing this visual separator in CSS?

For example, something like this:

```css
.app-view-edition-images__row {
  border-bottom: 1px solid /* some colour */;
}

.app-view-edition-images__row:last-child {
  border-bottom: 0;
}
```

I have a feeling you might have already tried this, given the existence of the `app-view-edition-images__row` class name on the main `<li>` elements. It'd certainly be nice to avoid the HTML bloat, if possible. Did you already try it, and was there a reason why you landed on the additional `<li>` and `<hr>` instead?

If the reasoning is maintain consistency with existing Design System styles, it might be good enough to simply [re-use the same colour variable](https://github.com/alphagov/govuk-frontend/blob/503e076e5d9242fd9c5a7c843dcc470a6965d24a/src/govuk/core/_section-break.scss#L49-L51) that a GOV.UK Section Break uses – e.g.

```scss
border-bottom: 1px solid $govuk-border-colour;
```

This is the approach that was taken on the What's New page:

https://github.com/alphagov/whitehall/blob/cabec75c1951d2a5dea0d6976139d7a63f207604/app/assets/stylesheets/admin/views/_whats-new.scss#L5-L10

dnkrj: Yeah, this `hr` is unfortunately a little awkward. David T brought it up as well in his review as well.
https://github.com/alphagov/whitehall/pull/7338

You're right in thinking that I tried `border-bottom` with `.app-view-edition-images__row` first. The issue is that because of the list item padding there needs to be an element within the `li` to get the border to be the correct width. This felt slightly better to me than appending some text divs at the end of the previous `li` and unfortunately using `:after` would require `calc` in the css which I don't think we quite want to get into yet.

ollietreend: Very sadly, this makes sense 🙈 

Thanks for explaining it @dnkrj.

app/controllers/admin/edition_images_controller.rb
dnkrj: Good point, have added a test now 🌞

features/edition-images.feature
```diff
@@ -19,4 +19,15 @@ Feature: Images tab on edit edition
     And I have the "Preview images update" permission
     And a draft document with images exists
     When I visit the images tab of the the document with images
-    Then I should see a list of the images
\ No newline at end of file
+    And I should see a list with 2 images
```
ollietreend, line 22: I think this probably makes more sense to remain as a 'then' statement.

```suggestion
    Then I should see a list with 2 images
```

features/edition-images.feature
```diff
@@ -19,4 +19,15 @@ Feature: Images tab on edit edition
...
+  Scenario: Images can be deleted from the images tab
+    Given I am a writer
+    And I have the "Preview images update" permission
+    And a draft document with images exists
+    When I visit the images tab of the the document with images
+    And I should see a list with 2 images
+    And I click to delete an image
+    And I confirm the deletion
+    Then I should see a success banner
+    And I should see a list with 1 image
```
ollietreend, line 33: I think it should be possible to use multiple when/then statements in sequence to describe this journey.

```suggestion
    When I visit the images tab of the the document with images
    Then I should see a list with 2 images
    When I click to delete an image
    And I confirm the deletion
    Then I should see a success banner
    And I should see a list with 1 image
```

I've seen this pattern used before when describing multi-step journeys. For example, [admin-history-pagination.feature](https://github.com/alphagov/whitehall/blob/e9996106d76a2e31da4f7d0fbc0a24de1dc250da/features/admin-history-pagination.feature) follows a similar structure:

https://github.com/alphagov/whitehall/blob/e9996106d76a2e31da4f7d0fbc0a24de1dc250da/features/admin-history-pagination.feature#L11-L18

ollietreend (commented):
Really nice @dnkrj – thanks for adding the feature spec! 🎉 🤖 

I've added a couple of suggestions for different [step keywords](https://cucumber.io/docs/gherkin/reference/#keywords) to make the feature spec read more clearly. However, technically this all looks good 👍🏻

features/edition-images.feature
```diff
@@ -19,4 +19,15 @@ Feature: Images tab on edit edition
...
     When I visit the images tab of the the document with images
-    Then I should see a list of the images
\ No newline at end of file
+    Then I should see a list with 2 images
+
+  Scenario: Images can be deleted from the images tab
+    Given I am a writer
+    And I have the "Preview images update" permission
+    And a draft document with images exists
+    When I visit the images tab of the the document with images
```
ollietreend, line 28: Actually, sorry, one final thing. I've spotted a typo in this step name.

I know you didn't introduce it, but would you mind fixing it before merging this?

```suggestion
    When I visit the images tab of the document with images
```

dnkrj:
Thanks Ollie! Have updated

ollietreend (approved):

PR #7345, opened 2023-02-15
Add page to allow image details to be edited

dnkrj:

# What
Add a page to edit image details with fields for caption and alt text and image information.

# Trello cards
[Existing images](https://trello.com/c/QeifKsuT/1060-edit-details)
[New images](https://trello.com/c/tBzITHVi/1059-image-details-part-2-of-uploading-an-image)

# Screenshots
## Edit page
whitehall-admin dev gov uk_government_admin_editions_1370963_images_445034_edit(iPad Pro) (2)

## Edit page with alt text
whitehall-admin dev gov uk_government_admin_editions_1370963_images_445034_edit(iPad Pro) (1)

## Index page
whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPad Pro) (1)

## Success banner
whitehall-admin dev gov uk_government_admin_editions_1370963_images(iPad Pro) (2)

## Mobile
whitehall-admin dev gov uk_government_admin_editions_1370963_images_445034_edit(iPhone 12 Pro) (2)

### DIFF
--- a/app/assets/stylesheets/admin/views/_edition-images.scss
+++ b/app/assets/stylesheets/admin/views/_edition-images.scss
@@ -1,4 +1,5 @@
 .app-view-edition-images__section-break {
+  margin-top: 0;
   margin-bottom: govuk-spacing(5);
 }
 
@@ -13,4 +14,5 @@
 
 .app-view-edition-images__actions {
   margin-top: govuk-spacing(2);
+  margin-bottom: 0;
 }
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -14,6 +14,14 @@ class Admin::EditionImagesController < Admin::BaseController
     redirect_to admin_edition_images_path(@edition), notice: "#{filename} has been deleted"
   end
 
+  def update
+    if image.update(params.require(:image).permit(:caption, :alt_text))
+      redirect_to admin_edition_images_path(@edition), notice: "#{image.image_data.carrierwave_image} details updated"
+    else
+      render :edit
+    end
+  end
+
 private
 
   def image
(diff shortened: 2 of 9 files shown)


ollietreend (commented):
This looks good to me @dnkrj.

I have a slight concern that we're getting a bit heavy on feature tests. In cases where there could be various permutations of a journey – for example, the conditional show/hide of alt text depending on if it's already present – it's often nicer to cover them a bit further down in [the testing pyramid](http://www.agilenutshell.com/episodes/41-testing-pyramid).

In this case, I think it might be preferable to have one feature spec to cover the "edit image details" journey – essentially a basic journey/happy path for the feature. And then cover the show/hide alt text behaviour in the [edition_images_controller_test.rb](https://github.com/alphagov/whitehall/blob/a82e7ed6d531105dbd849354449168a95c8bab6b/test/functional/admin/edition_images_controller_test.rb) file. From there, it should be possible to assert the existence or absence of the alt text input using selectors (similar to how it's done in [whats_new_test.rb](https://github.com/alphagov/whitehall/blob/dcbe575af372012312b886272ae84029de1c1520/test/integration/whats_new_test.rb)).

That, in turn, would allow us to keep from overcrowding the feature specs and associated steps files. I also believe (in theory) that the integration tests should run slightly faster than the Cucumber tests.

This is all just based in opinion, rather than an objective judgement, so I'm very happy to talk about it more. Also feel free to go ahead with the current feature tests instead.

dnkrj:
I've moved the `alt-text` testing from the feature into the controller test 🌞. My original thinking that the controller test should test logic in the controller rather than the template but don't really have any strong feelings.

ollietreend (approved):
Great work 🎉 thanks @dnkrj!

PR #7347, opened 2023-02-16
Stop pre-computing HTML attachment Govspeak

ollietreend:
This PR effectively undoes the functionality introduced (a long time ago) in PR #1753.

When Whitehall rendered and served the GOV.UK frontend to users, it made sense for HTML Attachments to have their Govspeak content pre-computed as HTML and cached in the database. This meant Govspeak only need to be computed once for those pages.

However, things have moved on, and Whitehall no longer renders HTML Attachment pages direct to end users. Instead, they're rendered once and pushed into the Publishing API. Downstream of the Publishing API, the Content Store now represents the 'cache' of GOV.UK page content.

Therefore, the pre-computed HTML cache is no longer useful to us here in Whitehall.

---

Trello: https://trello.com/c/NTg9rcsG/1101-stop-rendering-html-attachment-content-asynchronously


test/unit/presenters/publishing_api/html_attachment_presenter_test.rb
```diff
@@ -74,6 +68,7 @@ def present(record)
     assert_equal "cy", present(html_attachment).content[:locale]
   end
 
+  # Is this really useful?
```
beccapearce: I think we discussed this briefly, that we'd expect it to be covered by some kind of validation rule but we should check.

ollietreend: 💯 

I did have a look into this, and I think we're safe to remove it thanks to validation rules.

See the final commit on this PR: 798eae08cee75f4249c5847359a293b59d4b6290

beccapearce: Aha! Yep just got to it!

beccapearce (approved):
I think this all looks good and removes so much complexity and redundant code! 

I think we should test on integration first just because it's removing a lot of data but looks good to go!

Might be interesting to see how much space we drop from the database too?

ollietreend:
Thanks @beccapearce!

Yep – I'll definitely take it for a spin on integration first.

I've been thinking about how to measure the change in database size. I think it might be tricky to tell immediately, because I don't think MySQL will free up unused disk space immediately. However, we should be able to see the relative change in database size by waiting for the overnight integration database dumps to start showing up in S3.

PR #7351, opened 2023-02-17
Only present past dates of relevant role in historical accounts index presenter

BeckaL:
On the past prime ministers page, we display their dates in role. 

For past prime ministers without historic appointments, in the content item we were presenting their dates in role for any historic role they held (for example, dates when they were chancellors). This fixes that. 

Note that since we're not yet using this content item, this isn't affecting the data displaying on the website, which is still accurate.

[Trello](https://trello.com/c/0NcK4ojz/420-populate-content-item-for-past-prime-ministers-index)


brucebolt (approved):
Good spot!

PR #7361, opened 2023-02-21
Disabled button state on submission

DilwoarH:
What

Create a design solution to notify to users that a form has been submitted.

Why

We found that we allow users to submit multiple times if they continue to click on the submit button. This isn't a problem noticed on a fast connection, but on slower connections, the user doesn't receive a visual indication that something is happening.

Built a quick prototype of how we could potentially do it but needs design thinking: https://codepen.io/DilwoarH/pen/jOxyMRy

Design system solution: https://design-system.service.gov.uk/components/button/#stop-users-from-accidentally-sending-information-more-than-once

The design system solution doesn't work as it only stops it for within the first second.

Background

Original problem: https://trello.com/c/DNw6WuuA/726-using-repeated-events-as-a-metric-for-success

https://trello.com/c/UTUQWWYr/21-disabled-button-state-on-submission


davidgisbey:
Looks good. Are there any other places this needs to be applied that we've already ported?

davidgisbey (approved):

DilwoarH:
> Looks good. Are there any other places this needs to be applied that we've already ported?

We'll have to do it as and when we notice it. Let's get this in and add other prs for any other bits

PR #7364, opened 2023-02-22
Convert Sitewide Settings Index page to Design system

DilwoarH:
## What

Moved the `index` page from bootstrap to GOVUK Design System.

## Why

We want to remove bootstrap from the Whitehall application and improve user experience.

## Getting settings locally
```ruby
minister_reshuffle = SitewideSetting.new(
      key: :minister_reshuffle_mode,
      on: true,
      govspeak: "Test minister [reshuffle](http://example.com) message",
    )
minister_reshuffle.save!
```

## Before
whitehall-admin dev gov uk_government_admin_sitewide_settings(iPad Pro) (1)

## After
whitehall-admin dev gov uk_government_admin_sitewide_settings(iPad Pro)

https://trello.com/c/dNEXewSW/16-sitewide-settings-listing-page


test/functional/admin/legacy_sitewide_settings_controller_test.rb
```diff
@@ -0,0 +1,32 @@
+require "test_helper"
+
+class Admin::SitewideSettingsControllerTest < ActionController::TestCase
```
davidgisbey, line 3: At the moment it's complaining as there are two test files with the same class name.

If you do this it'll work 

```
class Admin::LegacySitewideSettingsControllerTest < ActionController::TestCase
  tests SitewideSettingsController

```

davidgisbey: I think we also need to add the preview design system permission to the user in the non-legacy test 

We can just that helper method so it'll be 

```
login_as_preview_design_system_user  :gds_editor
```

DilwoarH: ahh thanks

DilwoarH:
@davidgisbey - all tests now passing

davidgisbey (approved):

PR #7367, opened 2023-02-22
Add organisation index page in the GOV.UK Design System

davidgisbey:
## Description

This PR moves the organisation index page to the GOV.UK Design System.

In order to do this it does a few things:

1. Duplicates the `OrganisationControllerTest` file and adds the `Preview design system` permission to the users in the non-legacy file. 
2. Ensures the controller renders the correct layout and view based on the whether the user has the  `Preview design system` permission
3.  Adds the view in the GOV.UK Design System
4. Updates any tests that break

##  Filtering 

I've chosen to add filtering to the page as it is already part of the Table Publishing Component and it took 10 seconds. I know it's not strictly like for like but it significantly improves the user experience at zero cost to us.

I can pull it out if the consensus is we shouldn't though.

### Before

image

### After 

image

image

## Trello card

https://github.com/alphagov/whitehall/pull/7367


nnagewad:
This looks neat @davidgisbey. With regards to the filter, does the list get smaller as a publisher starts typing into that field?

davidgisbey:
yea it fuzzy searches all columns @nnagewad like this 

image

image


image

image

davidgisbey:
So it's actually pretty straight forward to configure the columns width. 

```
.app-view-organisation-index__table .govuk-table__row .govuk-table__header:nth-child(1) {
  width: 10%;
}

.app-view-organisation-index__table .govuk-table__row .govuk-table__header:nth-child(2) {
  width: 30%;
}

.app-view-organisation-index__table .govuk-table__row .govuk-table__header:nth-child(3) {
  width: 30%;
}

.app-view-organisation-index__table .govuk-table__row .govuk-table__header:nth-child(4) {
  width: 15%;
}

.app-view-organisation-index__table .govuk-table__row .govuk-table__header:nth-child(5) {
  width: 15%;
}

```

Those were the column width values in Bootstrap which makes it look like this 

image


If i change the last two values to be 10% and 20% it doesn't have to wrap the links in the last column

image

There's also the option to use this styling within the pub component if we'd rather use that 

image

nnagewad:
I've also created a new Github Issue regarding the inconsistent table width: https://github.com/alphagov/govuk_publishing_components/issues/3271

davidgisbey:
I've ensured the table is fixed and set the column widths so it's as like for like as possible here https://github.com/alphagov/whitehall/pull/7367/commits/18ded3c4687e3712e6a65a9164ea00882d0243ac

app/views/admin/organisations/index.html.erb
```diff
@@ -1,33 +1,102 @@
...
+            {
+              text: organisation.govuk_status,
+            },
+            {
+              text: link_to('[gov.uk]', organisation.public_path, class: "govuk-link") +
+                      (link_to('[current site]', organisation.url, class: "govuk-link") if organisation.govuk_status != 'live'),
+            }
+          ]
+        end
+    } %>
```
DilwoarH, line 98: should this logic be moved up to the top?

DilwoarH:
Looks good, just one minor comment

app/views/admin/organisations/index.html.erb
davidgisbey: tbh i think it should go into a helper method and be tested. There's quite a lot going on in it. I'll sort that this morning.

davidgisbey:
Thanks for the review @DilwoarH. I've added a helper method as discussed

DilwoarH (approved):
LGTM

PR #7368, opened 2023-02-22
Add upload image to the images tab

jyoung-gds:
[Trello ticket link](https://trello.com/c/C7Aby79t/1137-add-image-upload-functionality-to-images-tab)

## What
- Adds the ability to upload images via the images tab. 
- Adds error handling of image uploads to the index page.

**NOTE:** The error handling mirrors existing error handling for images, but this needs refreshing. The error messages do not represent those in the Figma designs, and when uploading a file which is the wrong file type the error message does not accurately describe the issue to the user (it gives a "file cannot be blank" error). I will write up a separate ticket for this work as it was getting quite involved and blocking the ticket from progressing.

## Why
This is part of the images update epic

### DIFF
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -22,6 +22,21 @@ class Admin::EditionImagesController < Admin::BaseController
     end
   end
 
+  def create
+    @new_image = @edition.images.build
+    @new_image.build_image_data(image_params["image_data"])
+
+    if @new_image.save
+      redirect_to edit_admin_edition_image_path(@edition, @new_image.id)
+    else
+      # Removes @new_image from the edition, otherwise the index page will attempt to render it and error
+      @edition.images.delete(@new_image)
+      render :index
+    end
+  end
+
+  def edit; end
+
 private
 
   def image
@@ -46,10 +61,14 @@ private
     case action_name
     when "index"
       enforce_permission!(:see, @edition)
-    when "edit", "update", "destroy", "confirm_destroy"
+    when "edit", "update", "destroy", "confirm_destroy", "create"
       enforce_permission!(:update, @edition)
     else
       raise Whitehall::Authority::Errors::InvalidAction, action_name
     end
+  end
+
+  def image_params
+    params.fetch(:image, {}).permit(image_data: [:file])
   end
 end
--- a/app/views/admin/edition_images/index.html.erb
+++ b/app/views/admin/edition_images/index.html.erb
@@ -1,6 +1,7 @@
 <% content_for :context, "#{@edition.title}" %>
 <% content_for :page_title, "Images for #{@edition.format_name}" %>
 <% content_for :title, "Images for #{@edition.format_name}" %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @new_image, parent_class: "edition_images")) if @new_image.present? %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
--- a/features/edition-images.feature
+++ b/features/edition-images.feature
@@ -1,28 +1,27 @@
 @design-system-only
 Feature: Images tab on edit edition
 
-  Scenario: Accessing the images tab with correct permissions
+  Background:
     Given I am a writer
+
+  Scenario: Images tab is hidden by default
+    And I start drafting a new publication "Standard Beard Lengths"
+    When I am on the edit page for publication "Standard Beard Lengths"
+    Then the page should not have an images tab
+
+  Scenario: Accessing the images tab with correct permissions
     And I have the "Preview images update" permission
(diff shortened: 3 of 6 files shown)


app/views/admin/edition_images/preview.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+    href: admin_edition_images_path(@edition)
+  } %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body"> The image you selected is exactly the right size. The whole image will be used. </p>
+    <img src="<%= @image.url %>" height="640" alt="Image preview" class="app-view-edition-images__image govuk-!-margin-bottom-4"><br>
```
ollietreend: For the sake of consistency, I think we should either include both `width` and `height` attributes, or neither.

It seems safe enough to hardcode these values, given they're also hardcoded in the 'Image information' sidebar on this page.

features/edition-images.feature
```diff
@@ -52,3 +49,19 @@ Feature: Images tab on edit edition
...
     And I update the image details and save
     Then the publication "Standard Beard Lengths" should have 1 image attachment
+
+  Scenario: Small image uploaded
+    And I have the "Preview images update" permission
+    And I start drafting a new publication "Standard Beard Lengths"
+    When I am on the edit page for publication "Standard Beard Lengths"
+    And I navigate to the images tab
+    And I upload a 64x96 image
+    Then I should get the error message "Images image data file must be 960px wide and 640px tall, but is 64px wide and 96px tall"
```
ollietreend, line 58: This is not the prettiest of error messages – but you've already said you're going to follow up with some more changes to improve error messaging 👍🏻

app/controllers/admin/edition_images_controller.rb
```diff
@@ -26,9 +26,10 @@ def create
     new_image = @edition.images.build
     new_image.build_image_data(image_params["image_data"])
 
-    if @edition.save
+    if @edition.save && new_image.valid?
       redirect_to admin_edition_image_preview_path(@edition, new_image.id)
     else
+      @edition.reload
```
ollietreend, line 32: I'm not entirely sure if/why this `@edition.reload` call is needed? 🤔 

I would have expected that to actually reset the `@edition.errors` state and disconnect the `new_image` object from that edition.

Does it work without this line?

app/controllers/admin/edition_images_controller.rb
```diff
@@ -22,6 +22,20 @@ def update
     end
   end
 
+  def create
+    new_image = @edition.images.build
+    new_image.build_image_data(image_params["image_data"])
+
+    if @edition.save && new_image.valid?
```
ollietreend, line 29: In fact, I'm wondering if this would actually be better as:

```suggestion
    @new_image = @edition.images.build
    @new_image.build_image_data(image_params["image_data"])

    if @new_image.save
```

Since `@new_image` is already associated with the `@edition` – because it was created with `@edition.images.build`.

And then in the view, you'd be able to use:

```ruby
errors_for(@new_image.errors, :"image_data.file")
```

app/controllers/admin/edition_images_controller.rb
jyoung-gds: As per our Slack conversation: This line was to remove the `new_image` object from `@edition`. Without this, the index page errors as it attempts to render the `new_image` object in the "Uploaded Images" section. 

We agreed to change this to `@edition.images.delete(@new_image)`, to avoid reloading the entire edition which could be costly for large editions with lots of images and to include a comment to explain why this line is needed.

app/views/admin/edition_images/index.html.erb
```diff
@@ -1,6 +1,7 @@
 <% content_for :context, "#{@edition.title}" %>
 <% content_for :page_title, "Images for #{@edition.format_name}" %>
 <% content_for :title, "Images for #{@edition.format_name}" %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @edition, parent_class: "edition")) %>
```
ollietreend, line 4: Does this actually work?

I'm surprised you haven't needed to set `object: @new_image` rather than `object: @edition` 🤔

```suggestion
<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @new_image, parent_class: "edition")) if @new_image.present? %>
```

ollietreend (commented):
Great stuff @jyoung-gds! 🎉 

I've just commented one more thing (sorry) which I missed in my last review.

app/views/admin/edition_images/index.html.erb
jyoung-gds: Good spot, this was a hang up from changing the `edition_images_controller`. I've just updated it

ollietreend (approved):

PR #7369, opened 2023-02-22
Ports edit consultation page to design system

davidtrussler:
[Trello card](https://trello.com/c/6x9humqo/10-move-consultation-response-edit-page-to-the-govuk-design-system)

This makes the relevant additions to the 2 pages (same view) for editing a consultation document: 
- Public feedback
- Final outcome

The changes include
- Updating the controller
- Adding a test
- Adding a view

||Bootstrap|Design System|
|-|-|-|
|**Public feedback**|Screenshot 2023-02-22 at 16 02 21|Screenshot 2023-02-22 at 15 55 27|
|**Final outcome**|Screenshot 2023-02-22 at 15 59 05|Screenshot 2023-02-22 at 15 56 45|

### DIFF
--- a/app/controllers/admin/responses_controller.rb
+++ b/app/controllers/admin/responses_controller.rb
@@ -4,6 +4,7 @@ class Admin::ResponsesController < Admin::BaseController
   before_action :enforce_edition_permissions!
   before_action :prevent_modification_of_unmodifiable_edition
   before_action :find_response, only: %i[edit update]
+  layout :get_layout
 
   def show
     @response = response_class.find_by(edition_id: @edition) || response_class.new(published_on: Time.zone.today)
@@ -20,18 +21,29 @@ class Admin::ResponsesController < Admin::BaseController
   end
 
   def edit
-    render :edit_legacy
+    render_design_system("edit", "edit_legacy", next_release: false)
   end
 
   def update
     if @response.update(response_params)
       redirect_to [:admin, @edition, @response.singular_routing_symbol], notice: "#{@response.friendly_name.capitalize} updated"
     else
-      render :edit_legacy
+      render_design_system("edit", "edit_legacy", next_release: false)
     end
   end
 
 private
+
+  def get_layout
+    design_system_actions = []
+    design_system_actions += %w[edit update] if preview_design_system?(next_release: false)
+
+    if design_system_actions.include?(action_name)
+      "design_system"
+    else
+      "admin"
+    end
+  end
 
   def find_consultation
     @edition = Consultation.find(params[:consultation_id])
--- a/app/views/admin/responses/_form.html.erb
+++ b/app/views/admin/responses/_form.html.erb
@@ -1,0 +1,48 @@
+<%= form_for [:admin, consultation, consultation_response], url: [:admin, consultation, consultation_response.singular_routing_symbol] do |form| %>
+  <%= render "govuk_publishing_components/components/fieldset", {
+    legend_text: "Published on",
+    heading_size: "l"
+  } do %>
+    <%= render "components/datetime_fields", {
+      prefix: "consultation_#{consultation_response.singular_routing_symbol}",
+      field_name: "published_on",
+      date_only: true,
+      error_items: errors_for(consultation_response.errors, :published_on),
+      year: {
+        value: consultation_response.published_on&.year,
+        start_year: 1997,
+        end_year: Date.today.year + 5
+      },
+      month: {
+        value: consultation_response.published_on&.month
+      },
+      day: {
+        id: "consultation_#{consultation_response.singular_routing_symbol}_published_on",
+        value: consultation_response.published_on&.day
+      }
+    } %>
+  <% end %>
+
+  <%= render "components/govspeak-editor", {
+    label: {
+      heading_size: "l",
+      text: "Detail/Summary"
+    },
+    name: "consultation_#{consultation_response.singular_routing_symbol}[summary]",
+    rows: 20,
+    id: "consultation_#{consultation_response.singular_routing_symbol}_summary",
+    value: consultation_response.summary,
+    error_items: errors_for(form.object.errors, :summary),
+    data_attributes: {
+      alternative_format_provider_id: @edition && @edition.alternative_format_provider_id ? @edition.alternative_format_provider_id : current_user.organisation.try(:id)
+    }
+  } %>
+
+  <div class="govuk-button-group">
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Save"
+    } %>
+
+    <%= link_to("Cancel", "#{admin_edition_path(consultation)}/public_feedback", class: "govuk-link govuk-link--no-visited-state") %>
+  </div>
+<% end %>
--- a/test/functional/admin/legacy_responses_controller_test.rb
+++ b/test/functional/admin/legacy_responses_controller_test.rb
@@ -1,10 +1,12 @@
 require "test_helper"
 
-class Admin::ResponsesControllerTest < ActionController::TestCase
-  should_be_an_admin_controller
+class Admin::LegacyResponsesControllerTest < ActionController::TestCase
+  tests Admin::ResponsesController
+
+  legacy_should_be_an_admin_controller
 
   setup do
-    login_as_preview_design_system_user(:writer)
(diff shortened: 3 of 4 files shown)


app/controllers/admin/responses_controller.rb
```diff
@@ -20,19 +21,29 @@ def create
   end
 
   def edit
-    render :edit_legacy
+    render_design_system("edit", "edit_legacy", next_release: true)
```
davidgisbey, line 24: i think this needs to be `next_release: false` since it won't be going out in the next release

app/controllers/admin/responses_controller.rb
```diff
@@ -20,19 +21,29 @@ def create
...
   end
 
   def update
     if @response.update(response_params)
       redirect_to [:admin, @edition, @response.singular_routing_symbol], notice: "#{@response.friendly_name.capitalize} updated"
     else
-      render :edit_legacy
+      render_design_system("edit", "edit_legacy", next_release: true)
```
davidgisbey, line 31: ditto

app/views/admin/responses/_form.html.erb
```diff
@@ -0,0 +1,49 @@
+<%= form_for [:admin, consultation, consultation_response], url: [:admin, consultation, consultation_response.singular_routing_symbol] do |form| %>
+  <%= render "govuk_publishing_components/components/date_input", {
```
davidgisbey, line 2: I think these need to be selects. I think we can use this helper method 

https://apidock.com/rails/ActionView/Helpers/DateHelper/date_select

And steal a lot of the styling from the datetime fields component 

https://github.com/alphagov/whitehall/blob/main/app/assets/stylesheets/components/_datetime-fields.scss
https://github.com/alphagov/whitehall/blob/main/app/views/components/_datetime_fields.html.erb

test/functional/admin/legacy_responses_controller_test.rb
```diff
@@ -0,0 +1,116 @@
+require "test_helper"
+
+class Admin::LegacyResponsesControllerTest < ActionController::TestCase
+  tests Admin::ResponsesController
+
+  legacy_should_be_an_admin_controller
+
+  setup do
+    login_as :writer
```
davidgisbey, line 9: Thanks for duplicating the controller! I think we just need to update the non-legacy controller to use the `login_as_preview_design_system_user` so it runs that set of tests with a user who has the permission

davidgisbey (commented):
looking good. Mostly minor comment other than the select one. Sorry i know that's going to be a bit of a pain. I wonder if down the line we might want to make that a publishing component.

davidtrussler:
> looking good. Mostly minor comment other than the select one. Sorry i know that's going to be a bit of a pain. I wonder if down the line we might want to make that a publishing component.

Yeah I wondered about the select thing but just went with the nearest available existing component. No problem, I'll look at that. Cheers!

davidgisbey:
Yea i totally get why you picked that component. Ideally we would use it, but currently, we have't got anything in place to handle all the validations for inputs. If a user inputs an invalid date it'll just blow up and we won't present any helpful error messages back to the user. It might well be something we add down the line though!

app/controllers/admin/responses_controller.rb
davidtrussler: 👍

app/controllers/admin/responses_controller.rb
davidtrussler: 👍

app/controllers/admin/responses_controller.rb
```diff
@@ -20,19 +21,29 @@ def create
...
     end
   end
 
 private
 
+  def get_layout
+    design_system_actions = %w[edit update]
+
+    if design_system_actions.include?(action_name)
```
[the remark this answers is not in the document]
davidtrussler: 👍 Oops, yes, makes sense.

app/views/admin/responses/_form.html.erb
davidtrussler: Replaced date inputs with updated datetime_fields component

test/functional/admin/legacy_responses_controller_test.rb
davidtrussler: @davidgisbey I've updated the test here to use the appropriate login and removed the legacy one previously added. I'm not entirely confident that's what's required here though. See the [last commit](https://github.com/alphagov/whitehall/pull/7369/commits/7c7f8f73ab33c21cf57fc85109ca8a1f237b14a0). Could you check that?

test/functional/admin/legacy_responses_controller_test.rb
```diff
@@ -1,116 +0,0 @@
-require "test_helper"
```
davidgisbey: So we should leave this file as is so we continue to test the Bootstrap implementation until we release the page in  the design system. At that point we can remove this file.

davidtrussler: Oh yeah, that makes sense!! I've restored that file.

app/views/admin/responses/_form.html.erb
```diff
@@ -0,0 +1,47 @@
+<%= form_for [:admin, consultation, consultation_response], url: [:admin, consultation, consultation_response.singular_routing_symbol] do |form| %>
+  <%= render "govuk_publishing_components/components/fieldset", {
+    legend_text: "Published on",
+    heading_size: "s"
+  } do %>
+    <%= render "components/datetime_fields", {
+      prefix: "consultation_#{consultation_response.singular_routing_symbol}",
+      field_name: "published_on",
+      # date_heading: "Published on",
+      date_only: true,
```
davidgisbey, line 10: we need to link errors here so we need to add 

```
error_items: errors_for(@consultation_response.errors, :published_at
```

It should also have a large heading

```
    heading_size: "l",
```

app/views/admin/responses/_form.html.erb
```diff
@@ -0,0 +1,47 @@
+<%= form_for [:admin, consultation, consultation_response], url: [:admin, consultation, consultation_response.singular_routing_symbol] do |form| %>
+  <%= render "govuk_publishing_components/components/fieldset", {
+    legend_text: "Published on",
+    heading_size: "s"
```
davidgisbey, line 4: I think this should be "l" s it matches the govspeak editor input size

app/views/admin/responses/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+  } do %>
+    <%= render "components/datetime_fields", {
+      prefix: "consultation_#{consultation_response.singular_routing_symbol}",
+      field_name: "published_on",
+      # date_heading: "Published on",
+      date_only: true,
+      year: {
+        value: consultation_response.published_on.year,
```
davidgisbey, line 12:
```
`value: consultation_response.published_on&.year,`
```

We need to use a safe navigator here or it blows up when the page is submitted with blank values

app/views/admin/responses/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+        start_year: 1997,
+        end_year: Date.today.year + 5
+      },
+      month: {
+        value: consultation_response.published_on.month
+      },
+      day: {
+        value: consultation_response.published_on.day
```
davidgisbey, line 20: if we assign this the id of `"consultation_#{consultation_response.singular_routing_symbol}_published_on"` the errors will link and focus on the day input

app/views/admin/responses/_form.html.erb
davidtrussler: Yes, wasn't sure about that from a like-for-like point of view but probably makes more sense 👍

app/views/admin/responses/_form.html.erb
davidtrussler: 👍

app/views/admin/responses/_form.html.erb
davidtrussler: 👍

davidtrussler:
Thanks for the comments on the error stuff @davidgisbey - I've addressed those now.

davidgisbey (approved):
great work. tested it locally and it's behaving nicely 👍

PR #7370, opened 2023-02-22
Make the historical account presenter work

JonathanHallam:
Things weren't being added to the content item because links need content id's not urls

### DIFF
--- a/app/presenters/publishing_api/historical_account_presenter.rb
+++ b/app/presenters/publishing_api/historical_account_presenter.rb
@@ -54,16 +54,14 @@ module PublishingApi
         .distinct(&:person)
         .order(:started_at)
         .map(&:person)
+        .select { |person| person.historical_accounts.present? }
 
       person_to_present = historical_account.person
 
       return [] unless all_appointees.include?(person_to_present)
 
       neighbouring_role_holders(all_appointees, person_to_present).map do |person|
-        {
-          "title" => person.name,
-          "base_path" => person.historical_accounts.present? ? person.historical_accounts.for_role(role.id).first.public_path : HistoricalAccountsIndexPresenter.base_path,
-        }
+        person.historical_accounts.for_role(role.id).first.content_id
       end
     end
 
--- a/test/unit/presenters/publishing_api/historical_account_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/historical_account_presenter_test.rb
@@ -63,12 +63,7 @@ class PublishingApi::HistoricalAccountPresenterTest < ActiveSupport::TestCase
       person: [
         historical_account.person.content_id,
       ],
-      ordered_related_items: [
-        {
-          "title" => "Some Other Person",
-          "base_path" => "/government/history/past-prime-ministers/some-other-person",
-        },
-      ],
+      ordered_related_items: [person2.historical_accounts.first.content_id],
     }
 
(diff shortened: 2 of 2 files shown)


test/unit/presenters/publishing_api/historical_account_presenter_test.rb
```diff
@@ -102,7 +97,7 @@ class PublishingApi::HistoricalAccountPresenterTest < ActiveSupport::TestCase
 
     historical_accounts_in_descending_order.each_with_index do |historical_account, index|
       links = PublishingApi::HistoricalAccountPresenter.new(historical_account).links
-      actual_names_of_related_people = links[:ordered_related_items].map { |person_hash| person_hash["title"] }
+      actual_names_of_related_people = links[:ordered_related_items].map { |c_id| HistoricalAccount.where(content_id: c_id).first.person.forename }
```
BeckaL, line 100: It's a bit unecessary to do this via people's names now, since we can just assert on the expected content ids.

E.g:

```
      actual_content_ids_of_related_people = links[:ordered_related_items]
      expected_content_ids_of_related_people = expected_indices_of_surrounding_prime_ministers[index].map do |i|
        historical_accounts_in_descending_order[i].content_id
      end
      assert_equal expected_content_ids_of_related_people, actual_content_ids_of_related_people
```

app/presenters/publishing_api/historical_account_presenter.rb
```diff
@@ -60,10 +60,7 @@ def related_pms
       return [] unless all_appointees.include?(person_to_present)
 
       neighbouring_role_holders(all_appointees, person_to_present).map do |person|
-        {
-          "title" => person.name,
-          "base_path" => person.historical_accounts.present? ? person.historical_accounts.for_role(role.id).first.public_path : HistoricalAccountsIndexPresenter.base_path,
-        }
+        person.historical_accounts.present? ? person.historical_accounts.for_role(role.id).first.content_id : HistoricalAccountsIndexPresenter.content_id
```
BeckaL, line 63: I think there's going to be an issue here - where the person has no historical account (e.g. recent PMs), link expansion isn't going to work here, since the content id will not be that of a person. Effectively, we'll not get their name. So I wonder if, for those without historical accounts, we should add them in a separate 'people without historical accounts' hash (we'd also need to make sure their roles were expanded so we could get dates in office when sorting - or just add an assumption into the frontend that anybody without a historical account is the most recent?). 

Sorry for the spanner, this bit of functionality seems ever more of a pain especially since I imagine it's barely used (we could also not fall victim to sunk cost fallacy and consider scrapping it and saying it's not worth the dev time it'd take to do this and faff around with it on the frontend too 🙈)

Maybe we can discuss as devs, there are other alternatives to solving but I increasingly wonder if it's worth it.

brucebolt, line 63: Just thinking about this a bit more, `HistoricalAccountsIndexPresenter.content_id` will mean we display "Past Prime Ministers" as the link text for those without a historical account, instead of their name.  This could potentially appear multiple times in the list of related links.

Maybe in these cases, we should just exclude them and show less than 5 links?  It might mean `neighbouring_role_holders` could be simplified too?

BeckaL (approved):
Looks good to me, nice simplification too I think :-)

PR #7372, opened 2023-02-23
add body to worldwide organisation presenter

kashifatcha:
## What
Add the body to the content item for worldwide organisations. 

## Why
This will allow us to render worldwide organisations using an app outside of Whitehall.
This is part of ongoing work to shutdown Whitehall's frontend. 

Relies on - https://github.com/alphagov/publishing-api/pull/2294

Trello - https://trello.com/c/LrXHcFUM/382-add-body-to-worldwide-organisation-presenter

### DIFF
--- a/app/presenters/publishing_api/worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/worldwide_organisation_presenter.rb
@@ -19,6 +19,7 @@ module PublishingApi
       content.merge!(
         description:,
         details: {
+          body:,
           logo: {
             crest: "single-identity",
             formatted_title: item.logo_formatted_name,
@@ -46,6 +47,10 @@ module PublishingApi
 
     def description
       item.summary
+    end
+
+    def body
+      Whitehall::GovspeakRenderer.new.govspeak_to_html(item.body) || ""
     end
 
     def ordered_contacts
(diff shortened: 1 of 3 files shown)


app/presenters/publishing_api/worldwide_organisation_presenter.rb
```diff
@@ -48,6 +48,15 @@ def description
...
     end
 
+    def body
+      [
+        {
+          content_type: "text/govspeak",
+          content: item.body || "",
+        },
+      ]
+    end
```
brucebolt, line 58: In some presenters, we just use the parsed govspeak as HTML directly as the body content (e.g. [`DetailedGuidePresenter`](https://github.com/alphagov/whitehall/blob/860cb32a8d71f28cf01f80c96bc83198d15d2190/app/presenters/publishing_api/detailed_guide_presenter.rb#L79-L81)), but in others we use a hash like this (e.g. [`RolePresenter`](https://github.com/alphagov/whitehall/blob/860cb32a8d71f28cf01f80c96bc83198d15d2190/app/presenters/publishing_api/role_presenter.rb#L81-L88)).  What's the reason for choosing this approach over sending the parsed govspeak downstream?

JonathanHallam: Sorted!

brucebolt (approved):
Looks good to me.  There's two line breaks that have been removed that probably shouldn't have been, but it's not a blocking issue.

PR #7375, opened 2023-02-24
Adds option to date-time component to render date only

davidtrussler:
[Trello](https://trello.com/c/YNK1v6ta/27-create-a-new-publishing-component-to-use-select-elements-for-a-date-inputs)

- Updates view
- Updates documentation
- Updates spacing for consistency

These changes add an option to the Datetime fields component to allow the user to choose to render the date fields only. 

This option defaults to false so that it is compatible with existing uses of the component.

When set to true it removes the time fields and the “Date” heading, which becomes superfluous in this use case. 

It also applies a small change to the vertical spacing between the date and time fields for consistency when only one of these is displayed. This applies in particular when error messages are included.

Screenshot 2023-02-24 at 12 44 19

### DIFF
--- a/app/views/components/_datetime_fields.html.erb
+++ b/app/views/components/_datetime_fields.html.erb
@@ -1,6 +1,8 @@
 <%
  prefix = prefix
  field_name = field_name
+ date_only ||= false
+ date_heading ||= nil
 
  error_items = error_items ||= nil
 
@@ -46,12 +48,14 @@
 %>
 
 <%= tag.div class: root_classes, data: data_attributes do %>
-  <%= render "govuk_publishing_components/components/heading", {
-    text: "Date",
-    heading_level: heading_level || 3,
-    font_size: heading_size || "m",
-    padding: true,
-  } %>
+  <% unless date_only && !date_heading %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: date_heading || "Date",
+      heading_level: heading_level || 3,
+      font_size: heading_size || "m",
+      padding: true,
+    } %>
+  <% end %>
 
   <% if date_hint %>
     <%= render "govuk_publishing_components/components/hint", {
@@ -66,7 +70,7 @@
     } %>
   <% end %>
 
-  <div class="govuk-!-margin-bottom-3 app-c-datetime-fields__date-time-wrapper">
+  <div class="app-c-datetime-fields__date-time-wrapper">
     <div class="app-c-datetime-fields__date-time">
       <%= render "govuk_publishing_components/components/label", {
         text: "Day",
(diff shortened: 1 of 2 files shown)


app/views/components/_datetime_fields.html.erb
```diff
@@ -46,12 +47,14 @@
 %>
 
 <%= tag.div class: root_classes, data: data_attributes do %>
-  <%= render "govuk_publishing_components/components/heading", {
-    text: "Date",
-    heading_level: heading_level || 3,
-    font_size: heading_size || "m",
-    padding: true,
-  } %>
+  <% unless date_only %>
```
davidgisbey, line 50: Great work! Just one comment. It feels like you should still be able to provide a heading here like you can in the date_input component https://components.publishing.service.gov.uk/component-guide/date_input



Maybe we could do something like 

```
  <% unless date_only || !date_heading_label %>
      text: date_heading || "Date",
      heading_level: heading_level || 3,
      font_size: heading_size || "m",
      padding: true,
    } %>
  <% end %>

```

That way you can pass it in as an option if you choose to. Wdyt?

app/views/components/_datetime_fields.html.erb
```diff
@@ -1,6 +1,7 @@
 <%
  prefix = prefix
  field_name = field_name
+ date_only = date_only || false
```
davidgisbey, line 4: I think we should memoise this since it's used multiple times

date_only ||= date_only || false

app/views/components/_datetime_fields.html.erb
```diff
@@ -130,61 +133,65 @@
...
-  <% end %>
-
-  <div class="app-c-datetime-fields__date-time-wrapper">
-    <div class="app-c-datetime-fields__date-time">
-      <%= render "govuk_publishing_components/components/label", {
-        text: "Hour",
-        html_for: hour_select_id,
-        id: hour_label_id,
+  <% unless date_only %>
+    <div class="govuk-!-margin-top-3 ">
```
DilwoarH, line 137:
```suggestion
    <div class="govuk-!-margin-top-3">
```

extra space left

app/views/components/_datetime_fields.html.erb
```diff
@@ -1,6 +1,7 @@
 <%
  prefix = prefix
  field_name = field_name
```
DilwoarH, line 3: I think this can be removed lol

app/views/components/_datetime_fields.html.erb
davidtrussler: Yes, that makes sense. Though it needs to be AND rather than OR for the `unless` clause.

app/views/components/_datetime_fields.html.erb
davidtrussler: 👍

app/views/components/_datetime_fields.html.erb
davidtrussler: 👍

app/views/components/_datetime_fields.html.erb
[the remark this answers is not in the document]
davidtrussler: Updated this as per comment above.

app/views/components/_datetime_fields.html.erb
```diff
@@ -1,6 +1,8 @@
...
+ date_only ||= date_only || false
+ date_heading = date_heading || nil
```
DilwoarH, line 5: Isn't this the same as:

```suggestion
 date_only ||= false
 date_heading ||= nil
```

app/views/components/_datetime_fields.html.erb
davidtrussler: I think `field_name` needs to be there otherwise `field_name: "#{field_name}(3i)"` causes an error.

app/views/components/_datetime_fields.html.erb
davidtrussler: Hmm, I'm not too sure. I understood this to be a means of gaining efficiency though on reflection I'm not sure that stacks up. I guess those should both be consistent at least.

app/views/components/_datetime_fields.html.erb
davidgisbey: apologies you're correct!

nnagewad:
Great job @davidtrussler 👍🏽

app/views/components/_datetime_fields.html.erb
davidgisbey: sorry it should be date_only ||= false

app/views/components/_datetime_fields.html.erb
davidtrussler: This is updated as in the comment above for consistency.

app/views/components/_datetime_fields.html.erb
davidtrussler: I've left these as they are.

davidgisbey (approved):
Just gave it a good manual test and it's working nicely. Thanks for sorting that!

PR #7377, opened 2023-02-24
Add padding to pages for users without javascript

dnkrj:

https://trello.com/c/Li0EWfF2/1162-make-the-navbar-work-for-non-js-users

# What
Add padding/margin to the top of the main page container using a `:before` pseudo element. 

# Why
This ensures that the page content is visible to users who do not use JavaScript (and IE11 users) as the menus are always expanded.
Using a combination of fixed height and percentage based margin top aids with ensuring that the padding is appropriate across many screen sizes.
Using a pseudo element ensures that the styling does not clash with the `.govuk-width-container` styling.

# Video showcasing different screen widths

https://user-images.githubusercontent.com/9594455/221230042-a5af1653-4bc1-41e5-afdf-18a0c7bfbde5.mov

### DIFF
--- a/app/assets/stylesheets/admin/_layout.scss
+++ b/app/assets/stylesheets/admin/_layout.scss
@@ -1,0 +1,23 @@
+// These styles ensure that appropriate padding is added to the top of the page
+// to display the uncollapsed menu for users without javascript.
+$collapse-tabs-width: 710px;
+
+.legacy-whitehall .navbar-wrapper:after {
+  content: "";
+  display: block;
+  height: 400px;
+  margin-top: -20%;
+
+  @media (max-width: $collapse-tabs-width) {
+    height: 320px;
+  }
+
+  @media (min-width: $govuk-page-width) {
+    height: 150px;
+    margin-top: 0;
+  }
+}
+
+.js-enabled .legacy-whitehall .navbar-wrapper:after {
+  display: none;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -9,6 +9,7 @@ $govuk-page-width: 1140px;
 @import "./components/miller-columns";
 @import "./components/secondary-navigation";
 
+@import "./admin/layout";
 @import "./admin/views/audit-trail";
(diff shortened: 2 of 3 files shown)


DilwoarH:
Great work - one small change I would recommend is to attach the spacing to the navbar rather than the width container. This way, even if the width container isn't used - it'll still have the spacing applied. 

`_layout.scss`:

(I changed a few of the spacing values, so please have a look)

```scss
.legacy-whitehall .navbar-wrapper:after {
  content: "";
  display: block;
  height: 500px;
  margin-top: -20%;

  @include govuk-media-query($until: desktop) {
    height: 320px
  }
}

.js-enabled .legacy-whitehall .navbar-wrapper:after {
  display: none;
}
```

`views/layouts/design-system.html.erb`:

`.... # line 20 add navbar-wrapper class`

```html


 <div class="legacy-whitehall">
   <div class="environment-<%= environment %> navbar-wrapper">
      <%= render partial: "shared/header" %>
    </div>
  </div>
...
```

dnkrj:
The issue with adding the spacing to the navbar is that the percentage negative margin changes with full page width not the content width which means that content is once again hidden on wider displays.

This could be solved with another media query but I'm hesitant to add it as it's not needed, what do you think?

Screenshot 2023-02-27 at 17 17 20

DilwoarH:
> The issue with adding the spacing to the navbar is that the percentage negative margin changes with full page width not the content width which means that content is once again hidden on wider displays.
> 
> This could be solved with another media query but I'm hesitant to add it as it's not needed, what do you think?
> 
> Screenshot 2023-02-27 at 17 17 20

We can if you want but it seems to work fine on a 16-inch mac pro. Unless we know people have bigger screens (which I don't think is the case)

Screenshot 2023-02-27 at 5 49 45 pm

dnkrj:
I've change the code to attach the spacing to the navbar as you suggested, but added an extra media query and wasn't able to change to the media query you suggested as the tabs collapse at a different point

DilwoarH (approved):
Great work! 🥇🥇🥇

PR #7378, opened 2023-02-27
Default multi selects to show all values, clean up and add a configuration parameter to the template

dnkrj:

# What
[Remove unused arguments for the autocomplete template.](https://trello.com/c/uZEwCGpe/1153-clean-up-multi-select-and-a-showallvalues)
Add a configuration options argument that is forwarded to the autocomplete initialisation.
Default the `showAllValues` configuration option to `true` for multiselects.

# Why
The autocomplete was copied from content publisher and there has been much confusion about which arguments are supported and needed by the different versions. By forwarding the configuration options directly to the JavaScript we reach some level of consistency.

# Screenshot
Screenshot 2023-02-26 at 22 45 42


app/views/components/_autocomplete.html.erb
```diff
@@ -1,27 +1,18 @@
...
 
   root_classes = %w(app-c-autocomplete govuk-form-group)
-  root_classes << "app-c-autocomplete--#{width}" if width
-  root_classes << "app-c-autocomplete--search" if search
-  root_classes << "app-js-only" if jsonly
   root_classes << "govuk-form-group--error" if error_items.any?
 %>
 
-<%= tag.div class: root_classes, data: data_attributes.merge(module: "autocomplete", "autocomplete-type": type) do %>
+<%= tag.div class: root_classes, data: data_attributes.merge(module: "autocomplete") do %>
```
davidgisbey: One thing that i noticed about this is it overrides any existing modules? Is this the correct behaviour?

davidgisbey:
Functionality is working perfectly 👍 We've got a few of these scattered through the codebase (user needs specialist sector & the edit edition page). Should we fix that as part of this PR or in a follow up PR. Happy either way!

davidgisbey (approved):
lgtm!

app/views/components/_autocomplete.html.erb
```diff
@@ -1,27 +1,24 @@
...
   select ||= {}
-  input ||= {}
-  search ||= false
-  jsonly ||= false
+  autocomplete_configuration_options ||= {}
+
+  data_attributes ||= {}
+  data_attributes[:module] ||= ""
+  data_attributes[:module] << " autocomplete"
+  data_attributes[:module] = data_attributes[:module].strip!
```
ollietreend, line 14: _Tiny nitpick_ about the difference between [`String#strip` and `String#strip!`](https://ruby-doc.org/core-3.1.1/String.html#method-i-strip)...

The `strip!` method mutates the existing String object, so you shouldn't need to assign it back to itself.

Whereas `strip` returns a new String object, but doesn't mutate the existing one. And therefore you need to assign it if you want to keep it.

So this could be rewritten as:

```suggestion
  data_attributes[:module].strip!
```

or alternatively:

```suggestion
  data_attributes[:module] = data_attributes[:module].strip
```

We don't need to do both (i.e. re-assign and also mutate the String).

ollietreend (approved):
This looks good to me.

I've added a _teeny tiny_ nitpick. However, what you've got already works perfectly fine as is.

PR #7382, opened 2023-02-27
Fix the this attachment is accesible checkbox

beccapearce:
There was a missing `=` meaning that the this attachment is accessible checkbox could not be unchecked.


davidgisbey (approved):
lgtm!

PR #7394, opened 2023-03-02
Add pagination component

DilwoarH:
This ports the pagination component from the GOVUK Design System - https://design-system.service.gov.uk/components/pagination/

We need this component for listing pages where pagination is required.

https://trello.com/c/SF2D6O7S/47-create-pagination-component

screenshot-whitehall-admin dev gov uk-2023 03 02-17_14_51

### DIFF
--- a/app/views/components/_pagination.html.erb
+++ b/app/views/components/_pagination.html.erb
@@ -1,0 +1,60 @@
+<%
+  id ||= "pagination-#{SecureRandom.hex(4)}"
+  aria_label ||= "results"
+  items ||= []
+  previous_href ||= false
+  next_href ||= false
+%>
+
+<nav id="<% id %>" class="app-c-pagination govuk-pagination" role="navigation" aria-label="<%= aria_label %>">
+  <% if previous_href %>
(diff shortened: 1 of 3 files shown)


nnagewad:
very nice!!!!

davidgisbey (approved):
lgtm

davidgisbey:
Sorry just had a thought. What do we do for a page like this where the pagination can go to a signifcant amount of pages?

image

image

DilwoarH:
> This is looking amazing! Great work.
> 
> Just had a thought. What do we do for a page like this where the pagination can go to a significant amount of pages?
> 
> image
> 
> image
> 
> Just wondering if we need something like the 'Larger number of pages' section here https://design-system.service.gov.uk/components/pagination/

Ah yes thanks - I'll add that support

DilwoarH:
I've added the ellipses support but there isn't a pattern for first and last links

davidgisbey (approved):
Excellent work! Looks like this needs a quick rebase onto main. The fatality notice schema has been updated  on the publishing API so the pact tests are failing

PR #7399, opened 2023-03-03
8 consultation response show to design system

davidtrussler:

### DIFF
--- a/app/views/admin/responses/show.html.erb
+++ b/app/views/admin/responses/show.html.erb
@@ -1,42 +1,53 @@
-<% page_title "#{@response.friendly_name.capitalize} for: #{@edition.title}" %>
-<% page_class "response" %>
+<% content_for :page_title, "#{@response.friendly_name.capitalize} for consultation" %>
+<% content_for :title, "#{@response.friendly_name.capitalize} for consultation" %>
+<% content_for :context, "#{@edition.title}" %>
 
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "components/secondary_navigation", {
+      aria_label: "Document navigation tabs",
+      items: secondary_navigation_tabs_items(current_user, @edition, request.path)
+    } %>
 
-<div class="row">
-  <section class="col-md-8">
-    <h1><%= @response.friendly_name.capitalize %> for consultation</h1>
-    <%= consultation_editing_tabs(@edition) do %>
-      <% if @response.persisted? %>
-        <div class="summary">
-          <h3><%= @response.friendly_name.capitalize %> summary</h3>
-          <%= govspeak_to_html @response.summary %>
-          <p>Published on <%= @response.published_on.to_fs(:long_ordinal) %></p>
-        </div>
+    <% if @response.persisted? %>
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "#{@response.friendly_name.capitalize} summary",
+        font_size: "l",
+        margin_bottom: 3
+      } %>
 
-        <ul class="actions">
-          <li><%= link_to "Edit #{@response.friendly_name}", [:edit, :admin, @edition, @response.singular_routing_symbol] %></li>
-        </ul>
+      <%= govspeak_to_html @response.summary %>
 
-        <h3>Attachments (optional)</h3>
-        <ul class="actions">
-          <li><%= link_to 'Upload new file attachment', new_admin_response_attachment_path(@response) %></li>
-          <li><%= link_to 'Add new HTML attachment', new_admin_response_attachment_path(@response, type: "html") %></li>
-        </ul>
-        <%= render('attachments_legacy', attachable: @response) if @response.attachments.any? %>
+      <p class="govuk-body">Published on <%= @response.published_on.to_fs(:long_ordinal) %></p>
 
-      <% else %>
-        <p class="alert alert-info">
-          <%= consulation_response_help_text(@response) %>
-        </p>
+      <p class="govuk-body">
+        <%= link_to "Edit #{@response.friendly_name}", [:edit, :admin, @edition, @response.singular_routing_symbol], class: "govuk-link" %>
+      </p>
 
-        <%= render partial: 'legacy_form', locals: { consultation: @edition, consultation_response: @response } %>
-      <% end %>
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Attachments (optional)",
+        font_size: "l",
+        margin_bottom: 3
+      } %>
+
+      <%= render "govuk_publishing_components/components/list", {
+        items: [
+          (link_to 'Upload new file attachment', new_admin_response_attachment_path(@response), class: "govuk-link"), 
+          (link_to 'Add new HTML attachment', new_admin_response_attachment_path(@response, type: "html"), class: "govuk-link")
+        ]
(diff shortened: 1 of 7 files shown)


app/views/admin/responses/_attachments.html.erb
```diff
@@ -0,0 +1,38 @@
+<% if attachable.can_order_attachments? && attachable.attachments.many? %>
+  <p class="govuk-body"><%= link_to 'Reorder attachments', reorder_admin_response_attachments_path(@response), class: "govuk-link govuk-link--no-visited-state" %></p>
```
davidtrussler, line 2: I'm struggling to get the required path here to show the attachments that are added to a response.

features/step_definitions/consultation_steps.rb
```diff
@@ -101,19 +101,19 @@
...
-
-Then(/^the responses attachments should be in the following order:$/) do |attachment_list|
-  attachment_ids = all(".existing-attachments > li").map { |element| element[:id] }
-
-  attachment_list.hashes.each_with_index do |attachment_info, index|
-    attachment = Attachment.find_by(title: attachment_info[:title])
-    expect("attachment_#{attachment.id}").to eq(attachment_ids[index])
-  end
-end
+# When(/^I set the order of the responses attachments to:$/) do |attachment_order|
```
davidtrussler, line 104: I'm not sure how to deal with these tests in the design system scenario where there is no ordering of attachments present.

davidtrussler:
Hi @davidgisbey - I'm struggling in a couple of places on the Consultation response page (and added comments in the appropriate places) and wondered if you could point me in the right direction: 
- When we add the link to go to the page to reorder attachments I can't see how to form the correct URL
- Given the re-ordering no longer happens on this page there are failing tests for this operation. I can't see how to exclude those for the design system scenario whilst retaining them in the existing bootstrap version. 

Screenshot 2023-03-03 at 16 30 29

app/views/admin/responses/_attachments.html.erb
davidgisbey: Ah i see whats going on. We're using polymorphism to determine which class the attachment belongs to. An attachment can belong to an `Edition` or a `Response`, both of which have different data models.

The same view is used in both contexts so it needs to be model agnostic if possible & if not it needs to account for the different data models. 

On this line https://github.com/alphagov/whitehall/blob/main/app/views/admin/attachments/reorder.html.erb#L1
it's trying to set the context to the attachables title. When that's an edition it's fine, but when it's an object that doesn't have  a title like a response it blows up.

We can change this line of code 

```
<% content_for :context, attachable.title %>
```

to 

```
<% content_for :context, attachable.try(:title) %>
```

so it retuns nil if aan object doesn't implement the title method.

Let me know if that makes no sense!

features/step_definitions/consultation_steps.rb
davidgisbey: Cool. We've got prior art for this here 

https://github.com/alphagov/whitehall/blob/main/features/step_definitions/attachment_steps.rb#L52-L81

It might be that you can literally just use the attachment methods as i think they do the exact same thing to the last couple of blocks in the feature spec would be 

```
  When I set the order of attachments to:
      | title                    | order |
      | Beard Length Graphs 2012 | 0     |
      | Outcome attachment title | 1     |
    Then the attachments should be in the following order:
      | title                    |
      | Beard Length Graphs 2012 |
      | Outcome attachment title |
```

app/views/admin/responses/_attachments.html.erb
davidtrussler: Ah great, yes that makes sense.

app/views/admin/responses/_attachments.html.erb
```diff
@@ -0,0 +1,38 @@
+<% if attachable.can_order_attachments? && attachable.attachments.many? %>
```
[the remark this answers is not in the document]
davidtrussler: Yes, that would be better. I think the only problem is that we don't need the heading in the `attachments/attachments` partial in this instance. I can't see if there's a way to optionally exclude this, otherwise I guess we would need to change the other view that calls this to include the heading at that point.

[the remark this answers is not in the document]
davidgisbey: Oh right i see. Yea i think we can pull the heading into the views above like you said. The other option is to pass something into the locales but i think that's probs a bit overkill

davidgisbey, line 1: Now we've cleaned up the partial so it works with both we can render it and pull the dupe code out

```
    <%= render('/attachments/attachments', attachable: attachable) if attachable.attachments.any? %>

```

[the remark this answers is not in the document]
davidtrussler: Yes we can. 👍

app/views/admin/responses/show_legacy.html.erb
```diff
@@ -0,0 +1,42 @@
...
+        </div>
+
+        <ul class="actions">
+          <li><%= link_to "Edit #{@response.friendly_name}", [:edit, :admin, @edition, @response.singular_routing_symbol] %></li>
+        </ul>
+
+        <h3>Attachments (optional)</h3>
+        <ul class="actions">
+          <li><%= link_to 'Upload new file attachment', new_admin_response_attachment_path(@response) %></li>
+          <li><%= link_to 'Add new HTML attachment', new_admin_response_attachment_path(@response, type: "html") %></li>
```
davidgisbey, line 22: both those links need `class: "govuk-link`

app/views/admin/responses/show.html.erb
```diff
@@ -1,42 +1,54 @@
...
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Attachments (optional)",
+        font_size: "l",
+        margin_bottom: 3
+      } %>
+
+      <%= render "govuk_publishing_components/components/list", {
+        items: [
+          (link_to 'Upload new file attachment', new_admin_response_attachment_path(@response)), 
+          (link_to 'Add new HTML attachment', new_admin_response_attachment_path(@response, type: "html"))
```
davidgisbey, line 36: both these links need `class: "govuk-link"

app/views/admin/responses/show.html.erb
```diff
@@ -1,42 +1,54 @@
...
+        ]
+      } %>
+
+      <%# render('attachments', attachable: @response) if @response.attachments.any? %>
```
davidgisbey, line 40: and this can come out now 👍

davidgisbey:
i've left a couple of minor comments, but it's looking like it's just about there. I've given it a manual test and the functionality is behaving nicely. Once your happy with the commit history etc. as discussed give me a yell and i'll give it a quick glance then ✅

app/views/admin/responses/show.html.erb
davidtrussler: 👍

app/views/admin/responses/show.html.erb
davidtrussler: 👍

cbernardojr-gds:
@davidtrussler Few changes:

- Public feedback tab: Move the Edit Public feedback link under the Public feedback summary title
- Attachments: Make the Delete red (I believe adding the class "app-link--destructive")

Thanks!

PR #7403, opened 2023-03-06
Implement presenter for 'How Government Works' page

brucebolt:
This content into the presenter for the 'How Government Works' page, which will allow us to render it outside of Whitehall.

The page gets republished when:
- the relevant rake task is run
- a new prime minister is appointed
- a ministerial role appointment is created or updated
- an organisation is created or updated
- reshuffle mode is switched on or off

Depends on: https://github.com/alphagov/publishing-api/pull/2303

[Trello card](https://trello.com/c/ckwRKhIR)

### DIFF
--- a/app/presenters/publishing_api/how_government_works_presenter.rb
+++ b/app/presenters/publishing_api/how_government_works_presenter.rb
@@ -20,10 +20,10 @@ module PublishingApi
       content.merge!(
         base_path:,
         description: "About the UK system of government. Understand who runs government, and how government is run.",
-        document_type: "special_route",
+        document_type: "how_government_works",
         public_updated_at: Time.zone.now,
         rendering_app: Whitehall::RenderingApp::WHITEHALL_FRONTEND,
-        schema_name: "special_route",
+        schema_name: "how_government_works",
       )
 
       content.merge!(PayloadBuilder::Routes.for(base_path))
--- a/app/presenters/publishing_api/how_government_works_presenter.rb
+++ b/app/presenters/publishing_api/how_government_works_presenter.rb
@@ -24,10 +24,7 @@ module PublishingApi
         public_updated_at: Time.zone.now,
         rendering_app: Whitehall::RenderingApp::WHITEHALL_FRONTEND,
         schema_name: "how_government_works",
-        details: {
-          department_counts:,
-          ministerial_role_counts:,
-        },
+        details:,
       )
 
       content.merge!(PayloadBuilder::Routes.for(base_path))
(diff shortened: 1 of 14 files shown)


test/unit/presenters/publishing_api/how_government_works_presenter_test.rb
```diff
@@ -0,0 +1,110 @@
...
+        current_prime_minister: [
+          @current_pm.content_id,
+        ],
+      }
+
+      presenter = PublishingApi::HowGovernmentWorksPresenter.new
+
+      assert_equal expected_hash, presenter.content
+      assert_equal expected_links, presenter.links
+      assert_valid_against_publisher_schema(presenter.content, "how_government_works")
```
BeckaL, line 68: Could we also add a line to validate that the links are valid against the links schema à la [this](https://github.com/alphagov/whitehall/blob/15ed64ee5616e097eba47b325c12333ec4f826db/test/unit/presenters/publishing_api/operational_field_presenter_test.rb#L61) (sidenote, I think someone should go through whitehall and add these tests for all presenters at some point)

app/presenters/publishing_api/how_government_works_presenter.rb
```diff
@@ -0,0 +1,94 @@
...
+          reshuffle_in_progress: reshuffle_in_progress?,
+        }
+      end
+    end
+
+    def department_counts
+      {
+        ministerial_departments: Organisation.listable.ministerial_departments.count,
+        non_ministerial_departments: Organisation.listable.non_ministerial_departments.count,
+        agencies_and_public_bodies: Organisation.listable.select { |o| o.type.agency_or_public_body? }.count,
```
BeckaL, line 63: I note that we were hard-coding this before, so this seems like a good improvement. How are we going to display this on the frontend? Round down to nearest hundred and say x+? (I.e. at the moment we have 444, so does this go to saying "400+ agencies" rather than the current "300+"?

test/unit/sitewide_setting_test.rb
```diff
@@ -1,7 +1,17 @@
 require "test_helper"
 
 class SitewideSettingTest < ActiveSupport::TestCase
-  # test "the truth" do
```
BeckaL, line 4: 😬 glad we're making use of this now!

BeckaL (commented):
This looks good to me, just a request to add in a test line and a question about how we're going to use some info on the frontend. I added [thoughts](https://github.com/alphagov/publishing-api/pull/2303#pullrequestreview-1326599616) into your schemas PR about whether there was any other information we should include here.

app/presenters/publishing_api/how_government_works_presenter.rb
brucebolt: My plan was to have logic in the frontend so if it's less than 300, show the actual number.  If it's more, we'll show 300+.  But I thought having the actual number in the content item will make it easier to change this in the future.

BeckaL: Yeah I vote for rounding personally, but we can see when we get to that bit 👍

test/unit/presenters/publishing_api/how_government_works_presenter_test.rb
brucebolt: Nice idea.  This led me to notice I'd put the links in the wrong place in the schema 🤦‍♂️.  Agree we should add these for all presenters.

brucebolt:
I've just added an extra line in https://github.com/alphagov/whitehall/pull/7403/commits/451ce37692735c3a22f5052a59f659838a7695c0 as I forgot to make a call to `patch_links`, so the current PM never appeared in the content item.

PR #7406, opened 2023-03-06
Add 'Select with Search' component

ollietreend:
This PR adds a 'Select with Search' component. This is an app-specific "govuk_publishing_component" which can be used as a drop-in replacement for the ["Select" component](https://components.publishing.service.gov.uk/component-guide/select), but with some extra enhancements.

The generated `<select>` tag is enhanced with [Choices.js](https://github.com/Choices-js/Choices) to make it searchable.

## Features

- Users can filter / search the list of options (results are 'fuzzy matched')
- Supports option groups
- Progressively enhanced - without JavaScript, it's a regular `<select>`
- Styled to look like a GOV.UK Design System component

## What it looks like

Since this is a `govuk_publishing_component`, you can preview it in the Component Guide in development at:
http://whitehall-admin.dev.gov.uk/component-guide/select-with-search

| Screenshot | Description |
| --- | --- |
|  | A select with grouped options |
|  | A select where the options have been filtered with the search term "edin" <br><br> Note: results are 'fuzzy matched' |

## Usage

Use it in the same way you'd use the select component. It accepts all the same parameters.

### A basic example

```erb
<%= render "components/select-with-search", {
  id: "dropdown1",
  label: "My Dropdown",
  options: [
    {
      text: "Option one",
      value: "option1"
    },
    {
      text: "Option two",
      value: "option2",
      selected: true
    },
    {
      text: "Option three",
      value: "option3"
    }
  ]
} %>
```

### With grouped options

```erb
<%= render "components/select-with-search", {
  id: "dropdown-grouped",
  label: "Select a city",
  grouped_options: [
    [
      "England",
      [
        { text: "Bath", value: "bath" },
        { text: "Bristol", value: "bristol" },
        { text: "London", value: "london" },
        { text: "Manchester", value: "manchester" }
      ]
    ],
    [
      "Northern Ireland",
      [
        { text: "Bangor", value: "bangor" },
        { text: "Belfast", value: "belfast" }
      ]
    ],
    [
      "Scotland",
      [
        { text: "Dundee", value: "dundee" },
        { text: "Edinburgh", value: "edinburgh" },
        { text: "Glasgow", value: "glasgow" }
      ]
    ],
    [
      "Wales",
      [
        { text: "Cardiff", value: "cardiff" },
        { text: "Swansea", value: "swansea" }
      ]
    ]
  ]
} %>
```

### With tracking

```erb
<%= render "components/select-with-search", {
  id: "dropdown-with-tracking",
  label: "Document type",
  track_category: "DocumentTypeSelection",
  track_label: current_path, # optional
  options: [
    {
      text: "News article",
      value: "news-article"
    },
    {
      text: "Press release",
      value: "press-release"
    },
    {
      text: "Government response",
      value: "government-response"
    }
  ]
} %>
```

---

Trello: https://trello.com/c/G8Uagrjg/1154-create-an-app-component-for-single-select-with-search

### DIFF
--- a/app/assets/stylesheets/components/_select-with-search.scss
+++ b/app/assets/stylesheets/components/_select-with-search.scss
@@ -1,0 +1,415 @@
+.app-c-select-with-search {
+  //
+  // A theme for Choices.js styled to look like a GOV.UK Design System component
+  //
+  $background: #ffffff;
+  $focus-color: $govuk-focus-colour;
+  $focus-outline: $govuk-focus-width solid $govuk-focus-colour;
+
+  .choices * {
+    // Something inside .choices needs this – I'm not sure what yet
+    box-sizing: border-box;
+  }
+
+  .choices {
+    @include govuk-font($size: 19);
+    position: relative;
+    margin-bottom: 24px;
+  }
+
+  .choices:focus {
+    outline: none;
+  }
+
+  .choices:last-child {
+    margin-bottom: 0;
+  }
+
+  // Disabled state
+  .choices.is-disabled {
+    .choices__inner,
+    .choices__input {
+      background-color: #eaeaea;
+      cursor: not-allowed;
+      -webkit-user-select: none;
+      user-select: none;
+    }
+
+    .choices__item {
+      cursor: not-allowed;
+    }
+  }
+
+  // This is an intentionally high specificity selector to avoid using !important when targetting
+  // the <select hidden> element that would otherwise end up picking up `display: block` from the
+  // `.choices[data-type*="select-one"] .choices__input` rule below.
+  .choices .choices__inner select[hidden] {
+    display: none;
+  }
+
+  // Type: Single Select
+  .choices[data-type*="select-one"] {
+    cursor: pointer;
+
+    .choices__input {
+      display: block;
+      width: 100%;
+      padding: 10px;
+      border-bottom: 1px solid #dddddd;
+      background-color: $background;
+      margin: 0;
+    }
+
+    .choices__button {
+      background-image: url("data:image/svg+xml,%3Csvg width='21' height='21' viewBox='0 0 21 21' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='%23000' fill-rule='evenodd'%3E%3Cpath d='M2.592.044l18.364 18.364-2.548 2.548L.044 2.592z'/%3E%3Cpath d='M0 18.364L18.364 0l2.548 2.548L2.548 20.912z'/%3E%3C/g%3E%3C/svg%3E");
+      padding: 0;
+      background-size: 8px;
+      position: absolute;
+      top: 50%;
+      right: 0;
+      margin-top: -10px;
+      margin-right: 25px;
+      height: 20px;
+      width: 20px;
+      border-radius: 10em;
+      opacity: .25;
+
+      &:hover,
+      &:focus {
+        opacity: 1;
+      }
+
+      &:focus {
+        outline: $focus-outline;
+      }
+    }
+
+    .choices__item[data-value=""] .choices__button {
+      display: none;
+    }
+
+    // Dropdown arrow ▼
+    &:after {
+      content: "";
+      height: 0;
+      width: 0;
+      border-style: solid;
+      border-color: #333333 transparent transparent;
+      border-width: 5px;
+      position: absolute;
+      right: 11.5px;
+      top: 50%;
+      margin-top: -2.5px;
+      pointer-events: none;
+    }
+
+    &.is-open:after {
+      border-color: transparent transparent #333333;
+      margin-top: -7.5px;
+    }
+
+    &[dir="rtl"]:after {
+      left: 11.5px;
+      right: auto;
+    }
+
+    &[dir="rtl"] .choices__button {
+      right: auto;
+      left: 0;
+      margin-left: 25px;
+      margin-right: 0;
+    }
+  }
+
+  // Type: Select Multiple / Text
+  .choices[data-type*="select-multiple"],
+  .choices[data-type*="text"] {
+    .choices__inner {
+      cursor: text;
+    }
+
+    // Delete button X
+    .choices__button {
+      position: relative;
+      top: -2px;
+      display: inline-block;
+      margin-top: 0;
+      margin-right: -4px;
+      margin-bottom: 0;
+      margin-left: 8px;
+      padding-left: 16px;
+      border-left: 1px solid #929191;
+      background-image: url("data:image/svg+xml,%3Csvg width='21' height='21' viewBox='0 0 21 21' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='%230b0c0c' fill-rule='evenodd'%3E%3Cpath d='M2.592.044l18.364 18.364-2.548 2.548L.044 2.592z'/%3E%3Cpath d='M0 18.364L18.364 0l2.548 2.548L2.548 20.912z'/%3E%3C/g%3E%3C/svg%3E");
+      background-size: 12px;
+      width: 12px;
+      height: 100%;
+      line-height: 1;
+      opacity: .75;
+      border-radius: 0;
+
+      &:hover,
+      &:focus {
+        opacity: 1;
+      }
+    }
+  }
+
+  .choices__inner {
+    display: inline-block;
+    vertical-align: top;
+    width: 100%;
+    background-color: $background;
+    padding: 5px;
+    border: 2px solid #0b0c0c;
+    font-size: 19px;
+    min-height: 44px;
+    overflow: hidden;
+  }
+
+  &.govuk-form-group--error .choices:not(.is-active, .is-focused, .is-open) .choices__inner {
+    border-color: $govuk-error-colour;
+  }
+
+  .choices.is-focused .choices__inner,
+  .choices.is-open .choices__inner {
+    outline: $focus-outline;
+    outline-offset: 0;
+    box-shadow: inset 0 0 0 2px;
+  }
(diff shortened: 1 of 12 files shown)


app/views/components/_select-with-search.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  is_page_heading ||= false
+
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  heading_size = false unless shared_helper.valid_heading_size?(heading_size)
+  select_helper = GovukPublishingComponents::Presenters::SelectWithSearchHelper.new(local_assigns)
+
+  aria_describedby = { describedby: select_helper.describedby }
+%>
+
+<%= content_tag :div, class: select_helper.css_classes, data: { module: "select-with-search" } do %>
```
davidgisbey, line 14: I think we need an option to pass in data modules on init. For example we track selection and deselection of supporting and lead orgs on the edit page with the `track-select-click` module

davidgisbey:
I think we need to update the styling a little 

image

davidgisbey:
Hey, sorry i'm struggling to find time to give this a proper tech review, but i've left a couple of comments.

lib/govuk_publishing_components/presenters/select_with_search_helper.rb
```diff
@@ -0,0 +1,66 @@
...
+               :hint_id,
+               :hint,
+               :label_classes,
+               :select_classes,
+               to: :@select_helper
+
+      def initialize(local_assigns)
+        @select_helper = SelectHelper.new(local_assigns.except(:options, :grouped_options))
+        @options = local_assigns[:options]
+        @grouped_options = local_assigns[:grouped_options]
```
dnkrj, line 24: I could go either way on this, but did you consider a single options parameter that calls the appropriate method based on structure?

dnkrj (commented):
Generally looks good to me 🌞

lib/govuk_publishing_components/presenters/select_with_search_helper.rb
ollietreend: Yeah – I did try that out first, as it was my initial preference too.

We essentially just massage the data a little before passing it through to the Rails helpers [options_for_select] or [grouped_options_for_select].

I was originally expecting to be able to just sniff the type of options object passed in – e.g. an Array for 'flat' options, or a Hash for grouped options – but I soon realised it wasn't as simple as that. It turns out [grouped_options_for_select] accepts both Array and Hash inputs.

So without going too far into the detail and sniffing out different types of Array structure, I ended up just re-using the same approach that Rails itself uses – namely, different methods/parameters to differentiate between 'flat' vs grouped lists of options. That felt like the easiest and most conventional way to solve the problem.

> I could go either way on this

I feel the same. I don't have a strong feeling either way, so am happy to change it if you think the other approach would be better.

[options_for_select]: https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/options_for_select
[grouped_options_for_select]: https://apidock.com/rails/v5.2.3/ActionView/Helpers/FormOptionsHelper/grouped_options_for_select

dnkrj: Sounds good, that makes sense, just a shame that rails strayed from HTML conventions on this occasion.

app/assets/stylesheets/components/_select-with-search.scss
```diff
@@ -0,0 +1,416 @@
+.app-c-select-with-search {
+  //
+  // A theme for Choices.js styled to look like a GOV.UK Design System component
+  //
+  $background: #ffffff;
+  $focus-color: $govuk-focus-colour;
+  $focus-outline: $govuk-focus-width solid $govuk-focus-colour;
```
DilwoarH, line 7: I believe our standard is to keep all variables at the top.

app/assets/stylesheets/components/_select-with-search.scss
```diff
@@ -0,0 +1,416 @@
...
+      display: block;
+      width: 100%;
+      padding: 10px;
+      border-bottom: 1px solid #dddddd;
+      background-color: $background;
+      margin: 0;
+    }
+
+    .choices__button {
+      background-image: url("data:image/svg+xml,%3Csvg width='21' height='21' viewBox='0 0 21 21' xmlns='http://www.w3.org/2000/svg'%3E%3Cg fill='%23000' fill-rule='evenodd'%3E%3Cpath d='M2.592.044l18.364 18.364-2.548 2.548L.044 2.592z'/%3E%3Cpath d='M0 18.364L18.364 0l2.548 2.548L2.548 20.912z'/%3E%3C/g%3E%3C/svg%3E");
```
DilwoarH, line 65: could we move this to a file or not possible?

app/views/components/docs/select-with-search.yml
```diff
@@ -0,0 +1,161 @@
+name: Select with search
+description: A dropdown select with search
+body: |
+  Use this component to create a JavaScript-enhanced dropdown select.
+
+  It's powered by [Choices.js][], which is similar to [Select2][] but without the jQuery dependency.
+  And it's styled to look like [Accessible Autocomplete][], or any other [GOV.UK Design System][] component.
```
DilwoarH, line 7: Missing links in doc

app/views/components/docs/select-with-search.yml
```diff
@@ -0,0 +1,161 @@
...
+      label: Choose your preferred thing
+      hint: You might need some more information here
+      hint_id: optional-hint-id
+      options:
+      - text: Something
+        value: option1
+      - text: Something else
+        value: option2
+  with_tracking:
+    # description: Tracking can be enabled by passing a minimum of data_track_category and data_track_action. Other tracking attributes are optional. Tracking is provided by the track-select-change module.
```
DilwoarH, line 98: shall we remove this

config/routes.rb
```diff
@@ -404,5 +404,5 @@ def call(*args)
     mount DisableSlimmer.new(Sidekiq::Web), at: "/sidekiq"
   end
 
-  mount GovukPublishingComponents::Engine, at: "/component-guide" if Rails.env.development?
+  mount GovukPublishingComponents::Engine, at: "/component-guide" unless Rails.env.production?
```
DilwoarH, line 407: thanks you for this !

DilwoarH:
This is looking great! A few minor comments but looks good to merge

app/views/components/_select-with-search.html.erb
ollietreend: I've added support for tracking in 6feb67404f083c7148accc268620db626b33099c (2nd commit on this PR)

The `track-select-click` module can't be used directly because the `<select>` element is not preserved once Choices.js has kicked in. So I've gone with an implementation similar to the one used on the Autocomplete component (see #7013).

app/assets/stylesheets/components/_select-with-search.scss
ollietreend: I didn't want to pollute the global namespace, which is why I kept them inside `.app-c-select-with-search { }`.

What do you recommend?

app/assets/stylesheets/components/_select-with-search.scss
ollietreend: To be honest, I don't think `.choices__button` ever gets created. It's used when Choices.js is in 'multiple select' mode – but the Whitehall component only ever runs it in 'single select' mode.

However I don't really want to go through and strip support for 'multiple select' from the theme. I'd rather leave this intact as is, where the SVG lives directly where it's used, and the theme supports all the potential use cases of the JS library (in case our needs ever change in the future).

app/views/components/docs/select-with-search.yml
ollietreend: Nice catch @DilwoarH - thanks.

ollietreend:
@davidgisbey the z-index issue is now resolved

Select with Search z-index fixed

app/assets/stylesheets/components/_select-with-search.scss
DilwoarH: Makes sense

app/assets/stylesheets/components/_select-with-search.scss
```diff
@@ -0,0 +1,415 @@
...
+    padding: 10px;
+    font-size: 19px;
+    border-bottom: 1px solid #b1b4b6;
+
+    &:last-child {
+      border-bottom: 0;
+    }
+
+    &:nth-of-type(even) {
+      background-color: #fafafa;
```
DilwoarH, line 287:
```suggestion
      background-color: govuk-colour("light-grey");
```

I think we should update this colour with something with a bit more contrast.


## Before
Screenshot 2023-03-27 at 4 38 10 pm

## After

Screenshot 2023-03-27 at 4 40 56 pm

DilwoarH (approved):
LGTM - one minor suggestion, not required if you think otherwise

app/assets/stylesheets/components/_select-with-search.scss
ollietreend: Thanks @DilwoarH 🎉 

I've tried as much as possible to re-use the existing styles established by [Accessible Autocomplete](https://alphagov.github.io/accessible-autocomplete/).

That includes the alternating colours of options, which Accessible Autocomplete defines as:

```css
.autocomplete__option--odd {
  background-color: #FAFAFA;
}
```

([view source](https://github.com/alphagov/accessible-autocomplete/blob/a79d8f7c07e6e1af296ee0f913447a248faa550d/src/autocomplete.css#L104-L106))

So that's why I've gone with `#FAFAFA` in this component, too.

PR #7411, opened 2023-03-07
Replacing uglifier gem with terser for es6-friendly JS compression

kornysietsma:
Note that name mangling is disabled for now, for extra safety - need to discuss whether mangling is safe.  (mangling was on for uglifier)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,6 @@ gem "sidekiq-scheduler"
 gem "slimmer"
 gem "sprockets-rails"
 gem "transitions", require: ["transitions", "active_record/transitions"]
-gem "uglifier"
 gem "validates_email_format_of"
 gem "view_component"
 gem "whenever", require: false
@@ -102,3 +101,5 @@ group :cucumber, :test do
   gem "govuk_test"
   gem "launchy"
 end
+
+gem "terser", "~> 1.1"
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -102,3 +101,5 @@ group :cucumber, :test do
   gem "govuk_test"
   gem "launchy"
 end
+
+gem "terser", "~> 1.1"
```
ollietreend, line 105: It might be nice to move this `gem "terser"` line up so it sits alongside the other 'ungrouped' gems.

I don't think this is strictly enforced (there doesn't seem to be a linter rule for it) but it seems like the existing convention is for ungrouped gems to sit together, and be listed alphabetically.

So for this gem I think it'd sit on line 59, underneath `sprockets-rails` and above `transitions`.

ollietreend (commented):
This looks great. As we discussed, I think it should be fine to go ahead with name mangling enabled, since it was enabled on the previous Uglifier gem too.

Gemfile
kornysietsma: makes sense - I'll also remove the `~-> 1.1` specification for consistency.

ollietreend (approved):
Looks good to me! 🚀

PR #7413, opened 2023-03-08
Add image cropping to the new images flow

dnkrj:

# What
[Check image dimensions before upload and redirect oversized images to a crop page appending the image as a dataUrl.
Use cropper.js to allow the user to crop the image.
On form submission attach the cropped image to the page and continue in the regular upload flow.](https://trello.com/c/X7unauwt/1061-cropping-image)

# Why
Whitehall does not have an asset store, consequently only images ready to be displayed on a page can be uploaded. By cropping the image before upload users a given the ability to crop images.

# Screenshot
whitehall-admin dev gov uk_government_admin_editions_1374126_images(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1374126_images(iPhone 12 Pro)


app/assets/javascripts/components/image-cropper.js
```diff
@@ -0,0 +1,78 @@
+//= require cropperjs/dist/cropper.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function ImageCropper ($imageCropper) {
+    this.$imageCropper = $imageCropper
+    this.$image = this.$imageCropper.querySelector('.app-c-image-cropper__image')
+  }
```
ollietreend, line 9: It might be nice to define the 960x640 dimensions as constants/variables so they're a bit more obvious to the casual reader. Right now they're hardcoded as [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

```suggestion
  function ImageCropper ($imageCropper) {
    this.$imageCropper = $imageCropper
    this.$image = this.$imageCropper.querySelector('.app-c-image-cropper__image')
    this.$targetWidth = 960
    this.$targetHeight = 640
  }
```

app/assets/javascripts/components/image-cropper.js
```diff
@@ -0,0 +1,78 @@
...
+      return
+    }
+
+    var width = this.$image.clientWidth
+    var naturalWidth = this.$image.naturalWidth
+    var scaledRatio = width / naturalWidth
+
+    // Adjust the crop box limits to the scaled image
+    var minCropBoxWidth = Math.round(960 * scaledRatio)
+    var minCropBoxHeight = Math.round(640 * scaledRatio)
```
ollietreend, line 36:
```suggestion
    var minCropBoxWidth = Math.round(this.$targetWidth * scaledRatio)
    var minCropBoxHeight = Math.round(this.$targetHeight * scaledRatio)
```

app/assets/javascripts/components/image-cropper.js
```diff
@@ -0,0 +1,78 @@
...
+      })
+    }
+  }
+
+  ImageCropper.prototype.setupFormListener = function () {
+    var input = this.$imageCropper.querySelector('.js-cropped-image-input')
+    input.form.addEventListener('submit', function (event) {
+      event.preventDefault()
+      this.cropper.getCroppedCanvas({
+        width: 960
```
ollietreend, line 60:
```suggestion
        width: this.$targetWidth
```

app/assets/javascripts/components/image-cropper.js
```diff
@@ -0,0 +1,78 @@
...
+          this.$imageCropper.dataset.filename,
+          {
+            type: 'image/jpeg',
+            lastModified: new Date().getTime()
+          })
+        var container = new DataTransfer()
+        container.items.add(file)
+        input.files = container.files
+        input.form.submit()
+      }.bind(this), 'image/jpeg', 0.8)
```
ollietreend, line 73: What about images that were uploaded as PNGs or GIFs?

It would be nice to pass the original mime type in to [HTMLCanvasElement.toBlob()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob) so that we don't inadvertently end up turning crisp PNGs into blurry JPEGs.

ollietreend: Also, minor side note but hopefully not a blocker: it looks like Safari [doesn't support the 'quality' parameter](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob#browser_compatibility). So presumably (?) JPEGs are just exported at 100% quality 🤷🏻

Post-processing [happens server-side anyway](https://github.com/alphagov/whitehall/blob/6274099c39bab0d84761a2b549c703f56814eab3/app/uploaders/image_uploader.rb#L14-L16) so this shouldn't result in uncompressed JPEGs being published to GOV.UK in practice.

In fact, might even be worth using 100% quality all the time anyway, to avoid double-compressing the images. 🤷🏻 But I don't have a strong opinion on that.

app/views/admin/edition_images/crop.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :context, @edition.title %>
+<% content_for :page_title, "Preview image" %>
+<% content_for :title, "Preview image" %>
```
ollietreend, line 3: Should the page title be "**Crop image**"?

```suggestion
<% content_for :page_title, "Crop image" %>
<% content_for :title, "Crop image" %>
```

ollietreend (commented):
Using Base64 encoded Data URLs is a nice way to work around the image storage problem to allow traditional client/server navigation between the 'upload' and 'crop' pages.

I do wonder whether there's an upper limit in terms of file size. Especially since Base64 adds about 37% to the size of a file (so a 10mb JPEG will become ~13.7mb of page weight). But maybe that's not going to be a problem in practice.

app/assets/javascripts/admin/views/redirect-upload-to-crop.js
```diff
@@ -0,0 +1,58 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function RedirectUploadToCrop (module) {
```
ollietreend, line 5: I wonder if we could avoid the need for this `RedirectUploadToCrop` module altogether.

Just thinking out loud, but since we're already committing to a client => server => client request cycle, what if we leave Rails to handle the image dimension validation? (Which it's already doing anyway.)

Instead of relying on JavaScript to recognise the need for cropping, what if we detected it in Rails and rendered the cropping page when the image is oversized? We would already have the image data at that point, so could Base64 encode it and still send it back to the client as a Data URL.

ollietreend (commented):
One more thought...

app/assets/javascripts/admin/views/redirect-upload-to-crop.js
```diff
@@ -0,0 +1,58 @@
...
+
+      event.preventDefault()
+
+      var reader = new FileReader()
+      reader.readAsDataURL(file)
+      reader.onload = function () {
+        var img = document.createElement('img')
+        img.src = reader.result
+        img.onload = function () {
+          if (img.width > 960 && img.height > 640) {
```
jyoung-gds, line 29: I think the logic here needs to be inclusive of the size limits to include files which are the right width/height but the wrong height/width
```suggestion
          if ((img.width > 960 && img.height >= 640) || (img.width >= 960 && img.height > 640)) {
```

app/assets/javascripts/components/image-cropper.js
jyoung-gds, line 36: Is there a reason you used Math.round() for these values? I think this will cause problems with rounding down, because the minimum crop size will be smaller than 960x640 pixels meaning we couldn't support the image.

app/views/admin/edition_images/crop.html.erb
dnkrj: Double checked this with Nikin and he agreed

app/assets/javascripts/admin/views/redirect-upload-to-crop.js
dnkrj: Good point!

app/assets/javascripts/components/image-cropper.js
dnkrj: I copied these over from content publisher, but you're right, have changed to `ceil` and specified the width and height for the canvas

app/assets/javascripts/components/image-cropper.js
dnkrj: Have added code to maintain the type and removed the compression parameter 👍

app/assets/javascripts/admin/views/redirect-upload-to-crop.js
dnkrj: Have updated the code to validate and base64 encode the image on the server. It means using `MiniMagick` to get the image dimensions but overall I think that it's an improvement. Thanks!

test/unit/image_data_test.rb
```diff
@@ -30,6 +28,18 @@ class ImageDataTest < ActiveSupport::TestCase
     assert image_data.valid?
   end
 
+  test "requires images larger than 960x640 to be cropped" do
+    image_data = build_example("960x960_jpeg.jpg")
+    assert_not image_data.valid?
+    assert_equal ["must be 960px wide and 640px tall, but is 960px wide and 960px tall"], image_data.errors[:file]
```
jyoung-gds, line 34: I think updating the `:needs_cropping` error message I mentioned will help make this test more specific, making it a more specific cropping error message rather than a regular wrong size error. Is there also a way of checking that we get an error with type `:needs_cropping` in this test? Using similar syntax from the `edition_images_controller` maybe something like `assert image_data.errors.of_kind?(:file, :needs_cropping)`?

test/unit/image_data_test.rb
```diff
@@ -30,6 +28,18 @@ class ImageDataTest < ActiveSupport::TestCase
...
+  end
+
+  test "images cannot be cropped if one side is too small" do
+    image_data = build_example("300x1000_png.png")
+    assert_not image_data.valid?
+    assert_equal ["must be 960px wide and 640px tall, but is 300px wide and 1000px tall"], image_data.errors[:file]
+  end
```
jyoung-gds, line 41: Similarly here, I think re-wording our error message for images which are too small will improve the specificity of this test. It's hard to tell from reading the two tests what sets one apart from the other. Again using the `.kind_of?` method but using `assert_not` would make it clearer that in this scenario the image can't be cropped.

jyoung-gds (commented):
Just left a few comments on the recently changed stuff which I think needs changing. Looking really good, only small changes!

app/validators/image_validator.rb
```diff
@@ -44,8 +44,10 @@ def validate_size(record, image)
     target_width = @size[0]
     target_height = @size[1]
 
-    unless actual_width == target_width && actual_height == target_height
+    if actual_width < target_width || actual_height < target_height
```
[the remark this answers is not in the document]
dnkrj: I changed the number of the function here to make clear that it's checking each dimension individually, let me know if you have any ideas as to how it could be made even clearer

dnkrj (commented):
Thanks Johnny

ollietreend (approved):
This looks excellent. I can't wait for it to be in the hands of our users!

Nice work ✨🚀

PR #7417, opened 2023-03-09
Move New and Edit People page to the GOV.UK Design System

davidgisbey:
## Description 

This ports the new/edit people page to the GOV.UK Design System. It appends `legacy` to the current new/edit views and partials. Then adds the new views in the GOV.UK Design System and conditionally renders the Boostrap or GOV.UK Design System based on whether the user has the "Preview design system" permission.

## Trello card 

https://trello.com/c/gyhTZQLT/42-new-edit-person-page

#### Before

image

#### After 

image

#### Before

image

#### After 

image

**Updated image sizing**

image


app/views/admin/people/_form.html.erb
```diff
@@ -3,38 +3,118 @@
...
+    error_items: errors_for(person.errors, :letters)
+  } %>
+
+  <%= render "govuk_publishing_components/components/fieldset", {
+    legend_text: "Image",
+    heading_size: "l"
+  } do %>
     <% if person.image_url %>
-      <%= image_tag person.image_url %>
+      <%= image_tag person.image.url(:s465) %>
```
DilwoarH, line 70: What does `:s465` do?

DilwoarH: I'm guessing related to file size.

Maybe what we need to make this work for all screens is to add a wrapper around it to stop the image from overflowing.

DilwoarH:
Small comment but other than that - LGTM 👍

cbernardojr-gds:
@davidgisbey 
Few changes:

- Warning message in black
- Remove the warning at the bottom
- Optional: Can we have the Rt Hon as a question using radio? If not, maybe we could have in the checkbox something like [x] Add the prefix Ht Hon

newperson

app/views/admin/people/_form.html.erb
davidgisbey: Is the best way to do this just to do something like

```
responsive-image { 
  max-width: 100%; 
  height: auto;
}
```

in the stylesheets and add the wrapper?

Where should the styling live. It won't be view specific. Should we have a top level file? How do we make it BEM compliant?

DilwoarH: Maybe something like this - I would put it in a separate view CSS file
```scss
.app-view-edit-person__image { 
  max-width: 100%; 
  height: auto;
}
```

davidgisbey: Isn't this something we will want to reuse on things like the show page (and anywhere else in Whitehall we use images?)

app/views/admin/people/edit.html.erb
```diff
@@ -1,12 +1,16 @@
...
-    <h1>Edit person</h1>
-    <p class="warning">Warning: changes to people appear instantly on the live site.</p>
-    <%= render partial: "form", locals: {person: @person, show_instantly_live_warning: true} %>
-  </section>
-  <div class="col-md-4">
-    <%= legacy_simple_formatting_sidebar %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/warning_text", {
+      text: "Warning: changes to people appear instantly on the live site.",
```
CL-GDS, line 8: Carlos advised, remove 'Warning:' from message

app/views/admin/people/new.html.erb
```diff
@@ -1,12 +1,16 @@
...
-    <h1>New person</h1>
-    <p class="warning">Do not create ministerial people without consulting GDS.</p>
-    <%= render partial: "form", locals: {person: @person, show_consult_gds_warning: true} %>
-  </section>
-  <div class="col-md-4">
-    <%= legacy_simple_formatting_sidebar %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/warning_text", {
+      text: "Do not create ministerial people without consulting GDS.",
```
CL-GDS, line 8: Expand acronym for clarity

CL-GDS (changes requested):
Small changes other then that LGTM

app/views/admin/people/new.html.erb
ollietreend: I'm not sure a pattern has been established within Whitehall's UI for expanding acronyms yet. For example, other areas in the 'Machinery of Government' have similar text which mention 'GDS' without expansion. Likewise the Dashboard page includes a link to the "GDS support form (Zendesk)".

I'm fairly confident that anybody using Whitehall will already be aware of what the acronym 'GDS' means – they will already have interacted with GDS through publisher training courses and the like.

So for now, I think it's safe to leave this as is. At least until the team has a Content Designer who's able to iterate the UI copy beyond like-for-like.

app/views/admin/people/edit.html.erb
CL-GDS: Sorry yes this was discussed in a call earlier. He advised remove 'Warning:' from the string's as a warning is explanitory

ollietreend (commented):
This is look good – one small thing from me.

app/controllers/admin/people_controller.rb
```diff
@@ -70,10 +70,7 @@ def update_order_role_appointments
 private
 
   def get_layout
-    design_system_actions = %w[reorder_role_appointments update_order_role_appointments]
```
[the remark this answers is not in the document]
davidgisbey: oh yea 🤦 good catch!

app/views/admin/people/_form.html.erb
ollietreend: I tend to agree with @davidgisbey – this does seem like a useful utility class that'd end up being used in more than one place. But I can't think of any examples off the top of my head – so maybe something to do later if/when the need arises. (I suspect other MoG pages might end up needing it.)

app/views/admin/people/_form.html.erb
```diff
@@ -0,0 +1,111 @@
...
+    id: "person_privy_counsellor",
+    name: "person[privy_counsellor]",
+    items: [
+      {
+        label: "Add the prefix Rt Hon",
+        value: "1",
+        checked: person.privy_counsellor
+      }
+    ]
+  } %>
```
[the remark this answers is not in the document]
davidgisbey: yup it's super annoying 😢

app/views/admin/people/edit.html.erb
davidgisbey: Have actioned. Thanks!

ollietreend (approved):
Looks good 👍🏻

PR #7421, opened 2023-03-10
38 move edit people translation page to the design system

davidgisbey:


cbernardojr-gds:
All good, @davidgisbey 👍

app/assets/stylesheets/admin/views/_translations.scss
```diff
@@ -0,0 +1,4 @@
+.app-view-edition-translation__english-content,
+.app-view-person-translation__english-content {
+  color: $govuk-secondary-text-colour;
+}
```
ollietreend, line 4: I think you've done the right thing here by consolidating these rules into one place.

You might want to go a step further and just define them as one class.

```suggestion
.app-view-translation__english-content {
  color: $govuk-secondary-text-colour;
}
```

..and update the HTML views accordingly.

ollietreend (approved):
Nice work @davidgisbey ✨🚀 

This is an exemplar PR for like-for-like transition work!

I've left one comment/suggestion, but have approved regardless since I don't consider it a deal breaker.

app/assets/stylesheets/admin/views/_translations.scss
davidgisbey: yea sounds good. Got a view on how is should be structured to be BEM compliant? I'm never really sure where something like this should live? It's not really an override & it is shared across many views

davidgisbey: cool yea that sounds good to me. It's still super intuitive what it does and where to look for it that way. I'll action that before merge.

Thanks for the speedy response!

PR #7446, opened 2023-03-15
Update design of the corporate information index page

davidgisbey:
## Description

This updates the design of the corporate information pages index page based on the feedback in this PR.

https://github.com/alphagov/whitehall/pull/7376#issuecomment-1470106151

image

And the consultation response show page based on this feedback 

https://github.com/alphagov/whitehall/pull/7399#issuecomment-1470125505

image

## Trello card 

https://trello.com/c/62Jqs595/8-move-consultation-response-show-page-to-the-govuk-design-system

https://trello.com/c/p6vopzC9/22-org-corporate-information-index-page

https://trello.com/c/qL6X06aL/44-worldwide-org-corporate-information-page


cbernardojr-gds:
@davidgisbey Looks great!

app/helpers/admin/corporate_information_pages_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def index_table_state_row(edition)
+    state = if edition.force_published?
+              "Force published"
+            elsif edition.unpublishing.present?
+              "Unpublished"
+            else
+              admin_edition_state_text(edition)
+            end
```
ollietreend, line 17: I think withdrawn editions will inadvertently get caught out by this.

Withdrawals are modelled as 'unpublishing' records – so they will have `edition.unpublishing.present?` and therefore be mislabelled as "Unpublished".

But, this does make me think. Should we actually make this change 'upstream' – i.e. update the `admin_edition_state_text` helper method, rather than wrapping more functionality around it?

That helper method is only used in a couple of places, so it shouldn't be too disruptive to change it 🤞🏻

app/helpers/admin/corporate_information_pages_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+             tag.span(state, class: "govuk-tag govuk-tag--s")
+           when "Published"
+             tag.span(state, class: "govuk-tag govuk-tag--s govuk-tag--green")
+           when "Scheduled"
+             tag.span(state, class: "govuk-tag govuk-tag--s govuk-tag--turquoise")
+           when "Rejected"
+             tag.span(state, class: "govuk-tag govuk-tag--s govuk-tag--red")
+           when "Withdrawn", "Unpublished"
+             tag.span(state, class: "govuk-tag govuk-tag--s govuk-tag--grey")
+           end
```
ollietreend, line 34: At some point soon, I expect these tags will be needed in more places than just the Corporate Information Pages index. I'd hope for the sake of consistency that we end up re-using it on the Documents index page, and other sections that need to list Editions.

So if not now, I think we'll _very soon_ need to consider moving this into the Editions helper rather than a specific CorporateInformationPages helper.

test/unit/helpers/admin/corporate_information_pages_helper_test.rb
```diff
@@ -0,0 +1,106 @@
...
+    },
+    {
+      state: :unpublished,
+      expected_tag_classes: "govuk-tag govuk-tag--s govuk-tag--grey",
+      label_text: "Unpublished",
+    },
+  ].each do |hash|
+    test "#index_table_state_row returns the correct tag for the #{hash[:state]} state" do
+      state = hash[:state]
+      edition = state == :unpublished ? create(:edition, state) : build(:edition, state)
```
ollietreend, line 58: This conditional _could be_ the reason why the test hasn't picked up on the "withdrawn" vs "unpublished" issue.

Does the bug present itself if this is changed to:

```suggestion
      edition = create(:edition, state)
```

?

ollietreend (commented):
A couple of comments from me!

I'm pretty confident there's a bug in here, but the test implementation could be hiding it.

app/helpers/admin/corporate_information_pages_helper.rb
DilwoarH, line 34: I think this can be simplified a lot by doing the following:

```suggestion
    tag_colour_class = case state
           when "Force published" then "govuk-tag--yellow"
           when "Draft" then "govuk-tag--blue"
           when "Published" then "govuk-tag--green"
           when "Scheduled" then "govuk-tag--turquoise"
           when "Rejected" then "govuk-tag--red"
           when "Withdrawn", "Unpublished" then "govuk-tag--grey"
           end
           
    tag.span(state, class: ["govuk-tag govuk-tag--s govuk-tag", tag_colour_class].join(" "))
```

or even

```
    def get_tag_class(state)
        tag_colour_class = case state
               when "Force published" then "govuk-tag--yellow"
               when "Draft" then "govuk-tag--blue"
               when "Published" then "govuk-tag--green"
               when "Scheduled" then "govuk-tag--turquoise"
               when "Rejected" then "govuk-tag--red"
               when "Withdrawn", "Unpublished" then "govuk-tag--grey"
               end
       ["govuk-tag govuk-tag--s govuk-tag", tag_colour_class].join(" ")
    end
    
    tag.span(state, class: get_tag_class(state))
```

app/views/admin/corporate_information_pages/index.html.erb
```diff
@@ -35,27 +32,30 @@
...
            },
            {
-             text: linked_author(edition.last_author, class: "govuk-link")
+             text: tag.p("#{time_ago_in_words edition.updated_at} ago", class: "govuk-!-margin-0") +
+                    "by " +
+                    linked_author(edition.last_author, class: 'govuk-link')
            },
            {
-             text: "#{time_ago_in_words edition.updated_at} ago",
+             text: index_table_state_row(edition),
```
DilwoarH, line 53: is there a reason why this is repeated?

app/views/admin/corporate_information_pages/index.html.erb
```diff
@@ -35,27 +32,30 @@
...
            },
            {
-             text: admin_edition_state_text(edition),
-           }
+             text: sanitize("<a class='govuk-link' href='#{admin_edition_path(edition)}'>Edit <span class='govuk-visually-hidden'> #{edition.title}</span></a>"),
```
DilwoarH, line 56: can we not do this using the `link_to` helper?

DilwoarH: something like this maybe (untested)

```ruby
link_to sanitize("Edit <span class='govuk-visually-hidden'> #{edition.title}</span>", admin_edition_path(edition), class: "govuk-link"
```

DilwoarH:
Left some comments, overall - looks great

app/helpers/admin/corporate_information_pages_helper.rb
davidgisbey: It feels like this is a little bit of a different use case since the other method purposely doesn't use Force published or Unpublished as states as they're not in the state machine.

Maybe they should be? It feels like at the very least unpublishing should be imo.

app/helpers/admin/corporate_information_pages_helper.rb
davidgisbey: Yea good shout. I'll move it to the editions helper.

app/views/admin/corporate_information_pages/index.html.erb
davidgisbey: good shout have done 👍

test/components/admin/editions/tags_component_test.rb
```diff
@@ -0,0 +1,98 @@
...
+    },
+    {
+      state: :unpublished,
+      expected_tag_classes: "govuk-tag govuk-tag--s govuk-tag--grey",
+      label_text: "Unpublished",
+    },
+  ].each do |hash|
+    test "returns the correct tag for the #{hash[:state]} state" do
+      state = hash[:state]
+      edition = state == :unpublished ? create(:edition, state) : build(:edition, state)
```
ollietreend, line 50: Would it be simpler just to `create` them all here?

```suggestion
      edition = create(:edition, hash[:state])
```

I imagine the `unpublished` state needs the Edition to exist in the database (i.e. with `create`) so the separate Unpublishing record can be created and attached to it. But that seems like a detail which probably isn't helpful context for a developer to need when working in this area of code.

ollietreend (commented):
@davidgisbey nice work on the View Component. This feels like a much better place for the code to live.

test/components/admin/editions/tags_component_test.rb
davidgisbey: yea that's exactly right, i just did this to speed the tests up but happy to use create.

app/components/admin/editions/tags_component.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def tags
+    sanitize(
+      [
+        state_tag,
+        limited_access_tag,
+        broken_links_report_tag,
+      ]
+      .compact
+      .join(""),
```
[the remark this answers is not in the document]
davidgisbey: have done thanks 👍

ollietreend (approved):
Nice one @davidgisbey! 👍🏻

PR #7457, opened 2023-03-20
Fix admin host check to work with any domain suffix.

sengi:
Needed for Replatforming switchover to avoid complicating the configuration. Doesn't represent any actual widening of access.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,7 +6,7 @@ class AdminRequest
   end
 
   def self.valid_admin_host?(host)
-    [Whitehall.admin_host, Whitehall.internal_admin_host].include? host
+    host.starts_with?("whitehall-admin")
   end
 end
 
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -1,13 +1,13 @@
+def valid_admin_host?(host)
```
kevindew, line 1: It's not ideal having this defined here as this will become globally available to the whole application. I imagine it'd be betteras a static method of the AdminRequest class so that you can call it with AdminRequest.valid_admin_host? You'd then need to change the matches method to use `self.class.valid_admin_host?()`

theseanything (approved):

config/routes.rb
theseanything: > Or you could inline it below with:
> 
> ```
> constraints: ->(request) { request.host.starts_with? "whitehall-admin" }
> ```

Checks also happen within the admin namespace routes that use the class method, so just saves some repetition.

kevindew (approved):

PR #7461, opened 2023-03-21
Remove UrlMaker for various `html_attachments`

callumknights:
This follows a similar vein in the below PR of making models aware of their own routes so we can stop relying on Rails helper methods to generate URL's. Not quite so simple in this case as this model covered 5 individual routes that were no longer rendered by Whitehall, so rather than completely re-do how URL's for this model are provided elsewhere, I've slightly rejigged the normal application of `public_path` and `public_url`.

This also involved an addition to the `append_url_options` method to allow previews URL's to be generated.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -12,12 +12,11 @@ class ApplicationRecord < ActiveRecord::Base
       path = "#{path}.#{options[:format]}"
     end
 
-    if options[:cachebust]
-      query_params = {
-        cachebust: options[:cachebust],
-      }
-      path = "#{path}?#{query_params.to_query}"
-    end
+    query_params = {}
+    query_params[:cachebust] = options[:cachebust] if options[:cachebust]
+    query_params[:preview] = options[:preview] if options[:preview]
+
+    path = "#{path}?#{query_params.to_query}" unless query_params.empty?
 
     path = "#{path}##{options[:anchor]}" if options[:anchor]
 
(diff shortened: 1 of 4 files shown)


app/models/application_record.rb
```diff
@@ -19,6 +19,8 @@ def append_url_options(path, options = {})
       path = "#{path}?#{query_params.to_query}"
     end
 
+    path = "#{path}?preview=#{options[:preview]}" if options[:preview]
```
brucebolt, line 22: This won't work if both `cachebust` and `preview` are present together as you'll end up with `path?cachebust=1234?preview=1234` instead of `path?cachebust=1234&preview=1234`.  The value for `preview` will need to be added into `query_params` further up.

Perhaps:

```ruby
    query_params = {}
    query_params[:cachebust] = options[:cachebust] if options[:cachebust]
    query_params[:preview] = options[:preview] if options[:preview]

    path = "#{path}?#{query_params.to_query}"
```

A test scenario where both `cachebust` and `preview` are both used at the same time would've caught this.

brucebolt (commented):
This is looking good @callumknights.

I've suggested a little refactor in the model that should make the `url` method less complex.

Also I think you'll need to add a test for the change to the `append_url_options` method, as that would've caught a problem that exists when both `cachebust` and `preview` are present.

app/models/application_record.rb
callumknights: Also sorted, cheers for picking this up, didn't think about both being applied at the same time 👍

callumknights: Got there in the end 👍

brucebolt (approved):
Thanks for making the changes and adding a test.  All looks good now.

PR #7467, opened 2023-03-21
71 add another component

davidtrussler:
[Trello](https://trello.com/c/dCR4WbZO/71-create-add-another-js-component-and-apply-it-to-casualties-fields)

This creates a new component to duplicate fields in a form, based on a couple of existing features, as shown below.

This is designed to be a generic component though it will initially be used in this specific location ("New fatality notice" page). This PR also adds the component to that location.

The main functions it implements are
- Adds an "Add" button to the section when the component loads. The actual text of this button is configurable but defaults to "Add another"
- On clicking the "Add" button 
  - the existing fields are copied
  - the values in the copied fields are cleared
  - the values for the `id`, `for` and `name` attributes are incremented in the newly created fields
  - the new fields are appended to the section
  - a "Delete" button is added to all the fields
  - the focus is set to the first field of the newly added set
- On clicking any of the "Delete" buttons
  - the appropriate set of fields is removed
  - the values for the `id`, `for` and `name` attributes are updated in all the remaining fields to be sequential
  - if only one field is displayed as a result of this action the "Delete" button no longer displays
  - the focus is set to the first field of the section

|[Existing bootstrap implementation](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/fatalities/new)|[Similar MOJ component](https://design-patterns.service.justice.gov.uk/components/add-another/)|This component|
|-|-|-|
|Screenshot 2023-03-21 at 16 52 07|Screenshot 2023-03-21 at 16 53 43|Screenshot 2023-03-23 at 14 22 28|


cbernardojr-gds:
@davidtrussler Few changes:
- Add button on top and green
- Delete as a link instead of the Remove as a button
- Grey hr line separating the elements

casualities

davidtrussler:
@cbernardojr-gds Thanks for that. I have a couple of queries on those changes: 
- Should we use a link in this instance ("Delete") since it is not a link to take the user elsewhere? I would say it is, semantically speaking, a button since it causes a change in the UI on some user action. I can't see if there is an equivalent elsewhere but maybe there is. Certainly we use a button (albeit destructive) for deleting something I think. 
- The problem with moving the "Add" button to the top is that it makes it awkward for users who use the keyboard to navigate through the form. They have to tab backwards each time to add another section. When it is at the bottom I think it is more intuitive for the user to tab from the field that is created once to the "Add" button.

For the moment I've updated it as below. 

Screenshot 2023-03-22 at 18 09 46

davidtrussler:
> 

@DilwoarH Yes, we changed that in the end - and after a conversation with the accessibility community. So it now has the "Add" button at the bottom and it's reverted to the grey style. Should be reflected in the screenshot at the top under "This component".

CL-GDS (approved):
LGTM. 
Investigating whether we need header tags (`_form` line `37`) or if we use a component. This can be addressed later as an actionable task for Myself, once I find out and have implemented it into the guide.

davidgisbey (approved):

PR #7471, opened 2023-03-22
Added Edition model diagram

kornysietsma:
New diagram added to show some details of the Edition and related model objects.

Also rebuilt the edition_state_transitions diagram as the old one had no source and was obsolete, added some clarifying comments in the edition_workflow page and moved all diagrams into a single directory


docs/edition_workflow.md
ollietreend, line 1: Ah - I'd thought you'd created this entire file when you first demoed the branch. I didn't realise `edition_workflow.md` already existed!

There's some out of date info in the **Unpublishing / Withdrawal** section. Since Whitehall no longer renders frontend pages, the `DocumentsController` is no longer relevant in this doc:

> If a document has an edition in one of the `Edition::PUBLICLY_VISIBLE_STATES` (`published` or `withdrawn`), Whitehall's `DocumentsController` will render it. If an `Unpublishing` exists, the withdrawal notice will be rendered near the top of the page as well.
>
> If a document does not have an edition in one of these states, it will use the `Unpublishing` to determine the response. If the `Unpublishing` is a redirect, ie a `Consolidated` `Unpublishing`, it will redirect to the alternative url. Otherwise it will render the `unpublished` view.

Feel free to tidy this up if you want to! But otherwise no worries - we'll clear it up another time.

ollietreend (approved):
These changes look great @kornysietsma. Thanks for doing this!

It's great to have PlantUML source for these diagrams, rather than them being lost to the mists of time.

docs/edition_workflow.md
kornysietsma: I'm not quite confident with what this section _should_ say, as I haven't really looked in this area.  Happy to make a change if you offer an alternative wording!

beccapearce (approved):

PR #7472, opened 2023-03-22
Add presenter for World index page

brucebolt:
This will present the content item for `gov.uk/world`, which is an index page for all world locations.

The page will be updated when world locations are created, updated or destroyed.

[Trello card](https://trello.com/c/7kBEEWUL)


lib/publish_world_index_page.rb
```diff
@@ -0,0 +1,9 @@
+class PublishWorldIndexPage
```
BeckaL, line 1: I do wonder here whether we might be able to combine some of these individual classes to publish single pages into a single thing that takes an argument of the presenter? Feels like we're adding a fair few of these now.

app/models/world_location.rb
```diff
@@ -32,6 +32,8 @@ class WorldLocation < ApplicationRecord
 
   scope :ordered_by_name, -> { with_translations(I18n.default_locale).order("world_location_translations.name") }
 
+  after_commit :republish_world_index_page_to_publishing_api
```
BeckaL, line 35: Now that we have this, do we also need to update the seeds file to skip this (e.g. as per [this](https://github.com/alphagov/whitehall/blob/c3a35269ce989b01c1c8e94da13394b5e1b03591/db/seeds.rb#L90)) - this was noticed by new devs setting up whitehall e.g. [here](https://gds.slack.com/archives/C02L13S214K/p1679061305102379?thread_ts=1679057170.131029&cid=C02L13S214K) from some of our recent work, and I think this might cause the same issue as we have world locations in the seeds?

BeckaL (commented):
Nice, generally looks good to me! Just a couple of comments / suggestions here.

lib/publish_world_index_page.rb
brucebolt: Nice idea.  I'll add this as a separate commit.

lib/present_page_to_publishing_api.rb
```diff
@@ -1,6 +1,6 @@
-class PublishWorldIndexPage
-  def publish
-    payload = PublishingApi::WorldIndexPresenter.new
+class PresentPageToPublishingApi
```
BeckaL, line 1: nice!

BeckaL (commented):
Nice refactor, looks good to me!

PR #7473, opened 2023-03-22
Prefix memcache keys and make the prefix configurable.

sengi:
Make Whitehall prefix its memcache keys so that it can share a memcache with other apps without name clashes. Most of the other GOV.UK apps already do this.

Unlike the other apps, Whitehall needs the prefix to be configurable. This is because the same Whitehall binary is run in two different configurations as `whitehall-admin` and `whitehall-frontend`. This way we can be sure that keys from those two deployments aren't going to collide (not that it would necessarily be a practical problem if they did, but it would be unexpected and unnecessarily complex behaviour so allowing the user to keep the keys separate is clearly the least-surprise choice here).

This will allow us to enable the Rails cache for `whitehall-admin`, which we used to have (albeit per EC2 instance) on the old hosting setup.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -68,8 +68,8 @@ Whitehall::Application.configure do
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
 
-  # Use a different cache store in production.
-  config.cache_store = :mem_cache_store
+  # Prefix memcache keys to avoid name clashes when sharing a cache.
+  config.cache_store = :mem_cache_store, nil, { namespace: ENV.fetch("MEMCACHE_KEY_PREFIX", "whitehall"), compress: true }
 
   # Use a real queuing backend for Active Job (and separate queues per environment)
   # config.active_job.queue_adapter     = :resque


sengi:
Friendly ping?

config/environments/production.rb
```diff
@@ -69,7 +69,7 @@
   config.log_tags = [:request_id]
 
   # Use a different cache store in production.
-  config.cache_store = :mem_cache_store
+  config.cache_store = :mem_cache_store, nil, { namespace: ENV.fetch("MEMCACHE_KEY_PREFIX", "whitehall"), compress: true } unless ENV["HEROKU_APP_NAME"]
```
kevindew, line 72: Would you be to explain why we need namespacing as I'm not too clear? 

Also we won't need a heroku conditional for this app.

sengi: Ah that's fab, I'll get rid of the Heroku bit, thanks!

The prefix is needed so that we don't have to worry about name clashes with other apps using the same memcache.

(We could have separate memcaches per app, but we don't want that because what we'd theoretically gain in resource isolation doesn't outweigh the admin overhead and complexity and waste that it'd entail, plus we know the app is well-behaved in the face of failing memcache requests anyway so even the resource isolation thing is moot.)

sengi: Done.

kevindew: Ah ok - yeah separate memcaches per app seem more consistent with our direction following https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md. I know we've just got a looming challenge too to separate out the namespaced redis instances since support ends for that with Sidekiq 7. Do agree that memcache is more friendly on failures. Is there a place that documents the shared memcache decision just so it's clear why it goes a different route to our other shared data direction of travel? Not really a concern for this PR.

Any chance you could update the commit message to state that the purpose of the namespace is so that it can be using a shared instance? that's the part I needed explaining.

sengi: This is an improvement that I've been working on in my spare time — I'm not really able to handle the bureaucracy, sorry.

sengi: Apologies for being a bit grumpy there earlier. I've written an extensive description that explains the rationale, the wider context into which this change fits and the direction of travel.

That's interesting that Sidekiq 7 is dropping support for the redis-namespace wrapper. (Looking at the implementation of redis-namespace I can see why, but there's clearly a need for it to be a first class feature of Sidekiq. Redis's numbered "databases" don't solve the same problem at all because there's no sensible way to map the numerical IDs automatically.) Sounds like an ideal opportunity to switch to in-cluster Redises per app: less toil, more developer-friendly, less cost, and we'd be able to test with the same software that we use in prod. The argument for Elasticache for Redis was that running our own was more hassle so it was worth the expense and the config complexity, but the reverse is true now.

kevindew (approved):
Thanks for the explanation on the commit message 👍 Code change looks good.

This was a bit of a surprise to us in senior tech due to the expectation of following: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md#proposal - @aldavidson was keen to have a chat to you to learn more about your position here

PR #7474, opened 2023-03-22
Make StatisticsAnnouncements aware of it's own base path

callumknights:
Done as part of the work to remove the use of UrlMaker from routes no longer rendered by whitehall.

### DIFF
--- a/app/views/admin/statistics_announcements/show.html.erb
+++ b/app/views/admin/statistics_announcements/show.html.erb
@@ -11,7 +11,7 @@
         <%= @statistics_announcement.title %>
       </span>
     </h1>
-    <%= link_to "View announcement on GOV.UK", statistics_announcement_url(@statistics_announcement, public_and_cachebusted_url_options) %>
+    <%= view_on_website_link_for @statistics_announcement %>
 
     <% if @statistics_announcement.cancelled? %>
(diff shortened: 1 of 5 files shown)


app/views/admin/statistics_announcements/show.html.erb
```diff
@@ -11,7 +11,7 @@
         <%= @statistics_announcement.title %>
       </span>
     </h1>
-    <%= link_to "View announcement on GOV.UK", statistics_announcement_url(@statistics_announcement, public_and_cachebusted_url_options) %>
+    <%= link_to "View announcement on GOV.UK", @statistics_announcement.public_url(public_and_cachebusted_url_options) %>
```
BeckaL, line 14: I wonder if here we should be using the `view_on_website_link_for` helper rather than specifying this link like this. It will then use the public_url method under the hood, and will make things it more consistent - this was the only case I could find outside of that helper file of `public_and_cachebusted_url_options` being called directly outside of that helper, so might be a useful thing for when we eventually rationalise that helper file.

BeckaL (commented):
This looks great! I can't find any other examples of the helper method being used. I've just left a question here about whether we might be able to update the view to use a helper, that is tangentially related to the changes here.

BeckaL (approved):
👍

PR #7475, opened 2023-03-22
Transition Roles index page to GDS Design System 

chriswhitehouse-tw:
## Description

This ports the roles index page to the GOV.UK Design System. Then adds the new view in the GOV.UK Design System and conditionally renders the Bootstrap or GOV.UK Design System based on whether the user has the "Preview design system" permission.

### On landing:
Screenshot 2023-03-24 at 11 18 23
Screenshot 2023-03-24 at 11 18 39
Screenshot 2023-03-24 at 11 18 52

### Confirmation of Delete view
Screenshot 2023-03-24 at 11 19 28

### Following deletion of role
Screenshot 2023-03-24 at 11 19 54

## Trello

https://trello.com/c/MApp6pCM

### DIFF
--- a/app/views/admin/roles/index.html.erb
+++ b/app/views/admin/roles/index.html.erb
@@ -1,17 +1,19 @@
 <% content_for :page_title, "Roles" %>
 <% content_for :title, "Roles" %>
-
-<%= render "govuk_publishing_components/components/warning_text", {
-  text: "Do not create ministerial roles without consulting GDS.",
-  highlight_text: true
-} %>
+<% content_for :title_margin_bottom, 6 %>
 
 <%= render "govuk_publishing_components/components/button", {
-  title: "Create a role",
-  text: "Create role",
-  secondary_solid: true,
-  href: new_admin_role_path
+  title: "Create new role",
+  text: "Create new role",
+  href: new_admin_role_path,
+  margin_bottom: 6
 } %>
+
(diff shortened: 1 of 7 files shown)


app/views/admin/roles/index.html.erb
```diff
@@ -1,42 +1,72 @@
...
+      },
+      {
+        text:
+            if role.destroyable?
+                form_tag admin_role_path(role), method: :delete  do
+                  render "govuk_publishing_components/components/button", {
+                    text: "delete",
+                    destructive: true
+                  }
+                end
```
DilwoarH, line 65: We need a intermediary page that shows the delete confirmation

@davidgisbey @dnkrj - do you have an example PR?

cbernardojr-gds:
@chriswhitehouse-tw Few comments:
- "Create role" button should be green, and the text "Create new role". Also, add some space between the button and the table. The space between warning message, button, title and table should be the same.
- Warning message should be black. Can the warning be under the button instead? (I would argue that we don't need the warning there, and that information should be a page after you click on Create new role, but maybe let's just move to not to complicate it for now)
- We don't need the "Cannot delete" text since it doesn't explain why you can't, so not having the button/link will have the same meaning
- Delete button should be a red link.
- Can we have a column for Edit as a blue link instead of the Name linked?
-

cbernardojr-gds:
@chriswhitehouse-tw All good 👍

PR #7477, opened 2023-03-22
Reduce noise in the build output

floehopper:
There's quite a lot of noise (i.e. expected output) in the build output. Reducing this noise will make it easier to see any *unexpected* output which might be important to address. I've tried to categorise the noise that we could reduce:

* [x] Various calls to `Kernel#puts` within code under test.
* [ ] The Sidekiq worker logger seems to be configured to use stdout, e.g. `test/unit/workers/check_organisation_links_worker_test.rb`
* [ ] Asset precompilation looks like it's generating a lot of output, i.e. `RAILS_ENV=test bundle exec rails assets:clobber assets:precompile`
* [ ] The Pact specs seems to be run in a very verbose mode, i.e. `SPEC_OPTS='' /root/.rbenv/versions/3.1.2/bin/ruby -S pact verify --pact-helper /govuk/whitehall/spec/service_consumers/pact_helper.rb`
* [ ] Where does this message come from? "Locale files are in sync, nice job!"
* [ ] There are multiple occurrences of this Ruby parser warning (possibly generated by Rubocop?): `warning: parser/current is loading parser/ruby31, which recognizes 3.1.3-compliant syntax, but you are running 3.1.2. Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.`
* [ ] Do we need to generate coverage reports locally? Maybe disable unless `CI` env var or similar is defined.

/cc @chrisroos @chrislo

### DIFF
--- a/test/unit/reports/organisation_attachments_report_test.rb
+++ b/test/unit/reports/organisation_attachments_report_test.rb
@@ -8,7 +8,9 @@ class Reports::OrganisationAttachmentsReportTest < ActiveSupport::TestCase
     Timecop.freeze do
       path = Rails.root.join("tmp/#{organisation.slug}-attachments_#{Time.zone.now.strftime('%d-%m-%Y_%H-%M')}.csv")
 
-      Reports::OrganisationAttachmentsReport.new(organisation.slug).report
+      capture_io do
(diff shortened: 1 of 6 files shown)


floehopper:
I'm going to close this draft PR - I plan to open a new one soon.

PR #7481, opened 2023-03-23
Add GA4 tracking to HTML attachment links

AshGDS:
Hi @andysellick / @JamesCGDS - would you be able to approve this if all is OK?

## What

- Adds GA4 link tracking to HTML attachment links

## Why

- It's part of our interactions for our GA4 tracking.

Trello card: https://trello.com/c/nJ05htIn/491-add-tracking-attachment-link-clicks

## Testing

It's a bit difficult to Test as getting a local Whitehall page up isn't too easy at the moment. I managed to render the attachment HTML component on the Whitehall Admin dashboard instead (thanks @andysellick for the idea). Then I checked that the data attributes were correctly attached to the HTML from there. If you want to see the HTML rendered, what I did was:

- Run `make whitehall` then `govuk-docker up whitehall-app`
- In `whitehall`, copy everything from `app/views/documents/_attachment.html.erb` and paste it into `app/views/admin/dashboard/index.html.erb`
- Add `attachment = ExternalAttachment.new()` above the attachment code in `index.html.erb`
- Go to `app/models/external_attachment.rb` and find the method `def html?`. Change it so that it returns `true`. 
- If you go to http://whitehall-admin.dev.gov.uk/government/admin/ you'll see the HTML attachment with GA4 attributes on it. The CSS will be broken but that isn't needed here.
- There's no GA4 JavaScript in the Whitehall Admin dashboard so actually clicking the link won't do anything as there's no `dataLayer` or GA4 modules. But what you can do is copy the `ga4-link` JSON, and paste it onto a link on one of our GOVUK homepage links. Then, if you clicked the link you'll see that the attachment link JSON I wrote works fine with our link tracker. 


app/views/documents/_attachment.html.erb
```diff
@@ -13,8 +13,18 @@
...
   preview = controller.nil? ? nil : params[:preview]
+
+  attachment_data = {}
+  if attachment.html?
+    attachment_data[:module] = "ga4-link-tracker"
+    attachment_data[:ga4_track_links_only] = ""
+    attachment_data[:ga4_link] = {
+      "event_name": "navigation",
+      "type": "html attachment"
+    }.to_json
```
andysellick, line 24: Readability suggestion:

```Ruby
if attachment.html?
  attachment_data = {
    module: "ga4-link-tracker",
    ga4_track_links_only: "",
    ga4_link: {
      event_name: "navigation",
      type: "html attachment",
    },
  }.to_json
end
```

app/views/documents/_attachment.html.erb
```diff
@@ -13,8 +13,18 @@
   # to bust cache and view a specific edition, isn't applicable when publishing
   # to publishing-api.
   preview = controller.nil? ? nil : params[:preview]
+
+  attachment_data = {}
+  if attachment.html?
```
andysellick, line 18: We were going to make tracking be optional for anything inside the page content although this might be a special case. Did you look into making it optional and how complex was it?

AshGDS: @andysellick I couldn't think of a way to toggle it - it might require a Ruby developer to implement, as the only way I could think of doing that was if there was a checkbox on the admin UI for HTML attachments which allowed you to turn off tracking. But then I guess there would also need to be some kind of comms between content and performance analysts to make sure tracking is only turned off for certain reasons?

app/views/documents/_attachment.html.erb
AshGDS: @andysellick Thanks, made the change :+1:

AshGDS:
Thanks @andysellick I've made those changes. Jenkins is failing but that's due to GitHub's RSA key change.

app/views/documents/_attachment.html.erb
andysellick: That's a good point. I wonder if this is the first time we've encountered this situation? Might be worth expanding our ruleset for how we implement this. Great explanation, let's keep it like this and have a wider discussion soon.

andysellick (approved):

PR #7482, opened 2023-03-24
Get Involved page to GOVUK Design system

DilwoarH:
## What

This moves the get involved page to GOVUK Design system

## Why

This work is part of the transition to GOVUK Design project for Whitehall.

https://trello.com/c/z2EmXoY5/71-get-involved-page

### Before

whitehall-admin dev gov uk_government_admin_get-involved(iPad Pro) (1)

### After

whitehall-admin dev gov uk_government_admin_get-involved(iPad Pro)

### DIFF
--- a/app/views/admin/get_involved/index.html.erb
+++ b/app/views/admin/get_involved/index.html.erb
@@ -1,9 +1,26 @@
-<% page_title 'Get involved' %>
+<% content_for :page_title, "Get involved" %>
+<% content_for :title, "Get involved" %>
+<% content_for :title_margin_bottom, 4 %>
 
-<div class="get-involved-header">
-  <h1>Get involved</h1>
-  <%= link_to "View on website", get_involved_url %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">
+      <%= link_to "View on website", get_involved_url, class: "govuk-link", target: "_blank" %>
+    </p>
+
+    <%= render "components/secondary_navigation", {
+      aria_label: "Get involved tabs",
+      items: [
+        {
+          label: "Get involved",
+          href: admin_get_involved_path,
+          current: true,
+        },
+        {
+          label: "Take part pages",
+          href: admin_take_part_pages_path,
+        }
+      ]
+    } %>
+  </div>
 </div>
-<section class="get-involved">
-  <%= get_involved_tab_navigation %>
-</section>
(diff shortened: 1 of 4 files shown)


app/views/admin/get_involved/index.html.erb
```diff
@@ -1,9 +1,26 @@
...
+<% content_for :title, "Get involved" %>
+<% content_for :title_margin_bottom, 2 %>
 
-<div class="get-involved-header">
-  <h1>Get involved</h1>
-  <%= link_to "View on website", get_involved_url %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">
+      <%= link_to "View on website", get_involved_url, class: "govuk-link" %>
```
CL-GDS, line 8: Please adjust View on Website link to match latest in the Guide to match Carlos's approved style. 
Latest commit here: https://github.com/alphagov/whitehall/pull/7487

app/views/admin/get_involved/index.html.erb
```diff
@@ -1,9 +1,26 @@
...
+    </p>
+
+    <%= render "components/secondary_navigation", {
+      aria_label: "Get involved tabs",
+      items: [
+        {
+          label: "Get involved",
+          href: admin_get_involved_path,
+          current: true,
```
CL-GDS, line 17: @davidgisbey   Do these need paths to be set like this: 

current: path == admin_get_involved_path($INSTANCE) 

If it is I will update into the guide.

CL-GDS (changes requested):
View on Website link needs adjusting to latest standard. 
Comment on Pathing issued for secondary review. Possible changes needed on this after comment
Other then that LGTM.

davidgisbey:
Looks good to me. I think we probs should add a bit of margin like Conor says so it matches the other pages, but all good other than that.

Tbh i don't see the point of this at all, but since we're like for like let's worry about that later

davidgisbey (approved):

app/views/admin/get_involved/index.html.erb
DilwoarH: So:

```suggestion
      <%= link_to "View on website", get_involved_url, class: "govuk-link", target: "_blank" %>
```

Is that correct?

app/views/admin/get_involved/index.html.erb
DilwoarH: So this section is a bit awkward because the `take part` pages live in another folder. Let's leave this as it is now, but the two sections probably need to be combined at some point.

app/views/admin/get_involved/index.html.erb
DilwoarH: I've updated the spacing and the link target:

Screenshot 2023-03-28 at 2 58 50 pm

PR #7491, opened 2023-03-29
Add unique filename validation to edition images controller

jyoung-gds:
[Trello ticket link](https://trello.com/c/MTqCZng8/1196-add-validation-so-images-have-unique-filenames)

## What
- Refactors code to make use of delegations and methods introduced in [PR #7520 ](https://github.com/alphagov/whitehall/pull/7520)
- Adds a custom validation to the `image_data` model which checks whether an image has a unique filename when compared with other images on the same edition. This requires an edition and image, on which the image data will be validated against, to be set via an attribute accessor and for a file to be present.
- Makes use of the `image_data` validation in `edition_images#create` by passing in the uploaded image and edition to ensure images with duplicated filenames cannot be uploaded to an edition.
### Additional information
Passing the edition and uploaded image in as an attribute accessor was required as the `image_data` model does not have access to either in non-persisted state.
## Why
The images update proposed a new way of generating markdown for images using the image filename on upload. This means that we need to avoid having duplicate filenames on an edition, otherwise the markdown for both will be the same. 

## Screenshots
whitehall-admin dev gov uk_government_admin_editions_1408685_images(iPad Pro) (1)

### DIFF
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -27,7 +27,7 @@ class Admin::EditionImagesController < Admin::BaseController
     @new_image.build_image_data(image_params["image_data"])
 
     if @new_image.save
-      redirect_to edit_admin_edition_image_path(@edition, @new_image.id), notice: "#{@new_image.image_data.carrierwave_image} successfully uploaded"
+      redirect_to edit_admin_edition_image_path(@edition, @new_image.id), notice: "#{@new_image.filename} successfully uploaded"
     elsif new_image_needs_cropping?
       @data_url = image_data_url
       render :crop
--- a/app/controllers/admin/edition_images_controller.rb
+++ b/app/controllers/admin/edition_images_controller.rb
@@ -26,6 +26,8 @@ class Admin::EditionImagesController < Admin::BaseController
     @new_image = @edition.images.build
     @new_image.build_image_data(image_params["image_data"])
 
+    @new_image.image_data.validate_on_image = @new_image
+
     if @new_image.save
       redirect_to edit_admin_edition_image_path(@edition, @new_image.id), notice: "#{@new_image.filename} successfully uploaded"
     elsif new_image_needs_cropping?
--- a/app/models/image_data.rb
+++ b/app/models/image_data.rb
@@ -1,6 +1,8 @@
 require "mini_magick"
 
 class ImageData < ApplicationRecord
+  attr_accessor :validate_on_image
+
   VALID_WIDTH = 960
   VALID_HEIGHT = 640
 
@@ -10,6 +12,7 @@ class ImageData < ApplicationRecord
 
   validates :file, presence: true
   validates_with ImageValidator, size: [VALID_WIDTH, VALID_HEIGHT]
+  validate :filename_is_unique
 
   delegate :width, :height, to: :dimensions
   delegate :content_type, to: :file
@@ -46,5 +49,16 @@ private
                       image = MiniMagick::Image.open file.path
                       Dimensions.new(image[:width], image[:height])
                     end
+  end
+
+  def filename_is_unique
+    return if validate_on_image.blank? || file.blank?
+
+    image = validate_on_image
+    edition = validate_on_image.edition
+
+    if edition.images.excluding(image).joins(:image_data).exists?(["image_data.carrierwave_image = ?", filename])
+      errors.add(:file, message: "name is not unique. All your file names must be different. Do not use special characters to create another version of the same file name.")
+    end
   end
 end
--- a/app/views/admin/edition_images/_image_upload.html.erb
+++ b/app/views/admin/edition_images/_image_upload.html.erb
@@ -10,7 +10,7 @@
   <%= render "govuk_publishing_components/components/file_upload", {
     name: "image[image_data][file]",
     id: "edition_images_image_data_file",
-    hint: "Images can be JPEG, PNG, SVG or GIF files.",
+    hint: raw('Images can be JPEG, PNG, SVG or GIF files. If you are uploading more than one image, <a class="govuk-link" href="https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos">read the image guidance</a> on using unique file names.'),
     error_items: @new_image.present? ? errors_for(@new_image.errors, :"image_data.file") : nil,
   } %>
 
--- a/test/unit/image_data_test.rb
+++ b/test/unit/image_data_test.rb
@@ -36,6 +36,25 @@ class ImageDataTest < ActiveSupport::TestCase
     assert image_data.errors.of_kind?(:file, :too_large)
   end
 
+  test "rejects images with duplicate filename on edition" do
+    image_data = build_example("960x640_jpeg.jpg")
+    edition = create(:news_article, images: [build(:image, image_data:)])
+    image = build(:image, image_data:, edition:)
+
+    image_data.validate_on_image = image
+
+    assert_not image_data.valid?
+    assert_equal ["name is not unique. All your file names must be different. Do not use special characters to create another version of the same file name."], image_data.errors[:file]
+  end
+
+  test "does not validate unique filename if validate_on_image is not assigned" do
+    image_data = build_example("960x640_jpeg.jpg")
+    edition = create(:news_article, images: [build(:image, image_data:)])
+    build(:image, image_data:, edition:)
+
+    assert image_data.valid?
+  end
+
   test "#bitmap? is false for SVG files" do
     assert_not build_example("test-svg.svg").bitmap?
   end
--- a/features/edition-images.feature
+++ b/features/edition-images.feature
@@ -76,3 +76,15 @@ Feature: Images tab on edit edition
     And I navigate to the images tab
     And I click upload without attaching a file
     Then I should get the error message "Image data file can't be blank"
+
+  Scenario: Uploading a file with duplicated filename
+    And I have the "Preview images update" permission
+    And I start drafting a new publication "Standard Beard Lengths"
+    When I am on the edit page for publication "Standard Beard Lengths"
+    And I navigate to the images tab
+    And I upload a 960x640 image
+    And I update the image details and save
+    And I upload a 960x640 image
+    Then I should get the error message "Image data file name is not unique. All your file names must be different. Do not use special characters to create another version of the same file name."
+
+
--- a/test/functional/admin/edition_images_controller_test.rb
+++ b/test/functional/admin/edition_images_controller_test.rb
@@ -72,6 +72,18 @@ class Admin::EditionImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select ".govuk-error-summary li", "Image data file is too small. Select an image that is 960 pixels wide and 640 pixels tall"
   end
 
+  test "#create shows a validation error if image has a duplicated filename" do
+    login_authorised_user
+    edition = create(:news_article)
+    file = upload_fixture("images/960x640_gif.gif")
+    create(:image, edition:, image_data: create(:image_data, file:))
+
+    post admin_edition_images_path(edition), params: { image: { image_data: { file: } } }
+
+    assert_template "admin/edition_images/index"
+    assert_select ".govuk-error-summary li", "Image data file name is not unique. All your file names must be different. Do not use special characters to create another version of the same file name."
+  end
+
   def login_authorised_user
     user = create(:gds_editor)
     user.permissions << "Preview images update"


app/models/image.rb
```diff
@@ -15,6 +15,11 @@ def url(*args)
     image_data.file_url(*args)
   end
 
+  def filename_is_unique?
+    filename = image_data.file&.identifier
+    edition.images.none? { |i| i.image_data.carrierwave_image == filename }
+  end
```
ollietreend, line 21: I think there might be some unexpected behaviour in here.

If we take an existing image from the database, `#filename_is_unique?` _will always be_ `false`. Even if it's the only image on that edition.

That's because `edition.images` will include the image we're working with, since it's already saved to the database. So it'll always match against itself.

For example:

```ruby
edition = Edition.last

# assuming an Edition with 1 image
edition.images.count # => 1

edition.images.first.filename_is_unique? # => false

# and that's because edition.images contains edition.images.first
edition.images.include?(edition.images.first) # => true

# and so there will never be 'none'
edition.images.none? { |i| i.image_data.carrierwave_image == filename } # => always false
```

This isn't a problem in our particular use case (i.e. in the edition images controller) because we call the `#filename_is_unique?` method _before we've persisted_ the new image to the database. So it won't be included in `edition.images`.

However I'd suggest we consider making this method more true to its name, and make it ignore itself from the list of images. For example, to extend the existing approach:

```suggestion
  def filename_is_unique?
    filename = image_data.file&.identifier
    edition.images.excluding(self).none? { |i| i.image_data.carrierwave_image == filename }
  end
```

(of course, we'd want to make sure we've covered this scenario in a unit test)

app/controllers/admin/edition_images_controller.rb
```diff
@@ -26,7 +26,11 @@ def create
     @new_image = @edition.images.build
     @new_image.build_image_data(image_params["image_data"])
 
-    if @new_image.save
+    unless @new_image.filename_is_unique?
+      @new_image.image_data.errors.add(:file, message: "called #{@new_image.image_data.file.file&.filename} has already been uploaded. Choose a different image or rename the file.")
+    end
```
ollietreend, line 31: This might be nicer if implemented in the model, rather than directly in the controller. It feels more like a 'data layer' concern than something the controller should be aware of.

It should be possible (I think?) to move this into a [custom validation method](https://guides.rubyonrails.org/active_record_validations.html#custom-methods) on the model.

ollietreend (commented):
Nice work @jyoung-gds – this is looking good.

I've added a few comments – primarily on the approach you've taken in your unit tests, as well as a potential implementation bug.

I'm happy to talk any of this through if you want to.

app/controllers/admin/edition_images_controller.rb
```diff
@@ -26,10 +26,13 @@ def create
     @new_image = @edition.images.build
     @new_image.build_image_data(image_params["image_data"])
 
+    @new_image.image_data&.validate_on_edition = @edition
+    @new_image.image_data&.validate_on_image = @new_image
```
kevindew, line 30: It seems unusual to have the safe operator on an assignment. I'd imagine the code would be more readable putting it inside a conditional block

i.e. 

```
if @new_image.image_data
   @new_image.image_data.validate_on_edition = @edition
   @new_image.image_data.validate_on_image = @new_image
end
```

app/controllers/admin/edition_images_controller.rb
```diff
@@ -26,10 +26,13 @@ def create
...
+
     if @new_image.save
       redirect_to edit_admin_edition_image_path(@edition, @new_image.id), notice: "#{@new_image.filename} successfully uploaded"
     else
-      # Removes @new_image from the edition, otherwise the index page will attempt to render it and error
+      ## Remove @new_image from @edition.images array, otherwise the view will render it in the 'Uploaded images' list
```
kevindew, line 35: You've ended up with a double `#` here

app/models/image_data.rb
```diff
@@ -47,4 +50,15 @@ def dimensions
...
                     end
   end
+
+  def filename_is_unique
+    return if validate_on_edition.blank? || validate_on_image.blank? || file.blank?
+
+    edition = validate_on_edition
+    image = validate_on_image
+
+    if edition.images.excluding(image).any? { |i| i.filename == filename }
```
kevindew, line 60: Is this association already loaded in memory? Just wondering if it's potentially a big query we're running to load them all in which could be a query on the DB table with filename (that said I'm not sure where we get the filename from)

app/models/image_data.rb
```diff
@@ -47,4 +50,15 @@ def dimensions
...
+      errors.add(:file, message: "called #{filename} has already been uploaded. Choose a different image or rename the file.")
```
kevindew, line 61: Is this filename normalised (i.e. has it had nasty characters removed and such)? Just wondering if this may come out with a different filename potentially to what the user uploaded.

On the otherhand, if it's not normalised I wonder if that's a problem as it could breaking embedding.

test/unit/image_data_test.rb
```diff
@@ -30,6 +30,41 @@ class ImageDataTest < ActiveSupport::TestCase
...
+    image_data = build_example("960x640_jpeg.jpg")
+    images = [create(:image, image_data:)]
+    edition = create(:news_article, images:)
+    create(:image, image_data:)
+
+    image_data.validate_on_edition = edition
+
+    assert image_data.valid?
+  end
+
```
kevindew, line 67: Tests look good 👍

app/views/admin/edition_images/index.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :context, "#{@edition.title}" %>
 <% content_for :page_title, "Images for #{@edition.format_name}" %>
 <% content_for :title, "Images for #{@edition.format_name}" %>
-<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @new_image, parent_class: "edition_images")) if @new_image.present? %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @new_image.image_data, parent_class: "edition_images")) if @new_image.present? %>
```
kevindew, line 4: Looks like we'd lose any validation issues on the @new_image object with this change, I can see that still has validation rules.

app/views/admin/edition_images/_image_upload.html.erb
```diff
@@ -9,9 +9,9 @@
 ) do %>
   <%= render "govuk_publishing_components/components/file_upload", {
     name: "image[image_data][file]",
-    id: "edition_images_image_data_file",
+    id: "edition_images_file",
```
kevindew, line 12: It's unclear to me why this has changed

app/views/admin/edition_images/_image_upload.html.erb
```diff
@@ -9,9 +9,9 @@
...
     hint: "Images can be JPEG, PNG, SVG or GIF files.",
-    error_items: @new_image.present? ? errors_for(@new_image.errors, :"image_data.file") : nil,
+    error_items: @new_image.present? ? errors_for(@new_image.image_data.errors, :"file") : nil,
```
kevindew, line 14: I'm struggling to understand why this changed, is it to change the format of the error messages?

kevindew (commented):
This is looking good Jonny, I did find myself a bit confused with the last commit where there seems to be a bunch of other aspects than the validation made as changes - which made things a bit trickier to review.

app/views/admin/edition_images/_image_upload.html.erb
jyoung-gds: Ah yea this is a hang-up from the previous solution, I forgot to change this back after Ollie's review. Apologies!

app/controllers/admin/edition_images_controller.rb
```diff
@@ -26,10 +26,15 @@ def create
...
+      @new_image.image_data.validate_on_edition = @edition
+      @new_image.image_data.validate_on_image = @new_image
+    end
+
     if @new_image.save
-      redirect_to edit_admin_edition_image_path(@edition, @new_image.id), notice: "#{@new_image.image_data.carrierwave_image} successfully uploaded"
+      redirect_to edit_admin_edition_image_path(@edition, @new_image.id), notice: "#{@new_image.filename} successfully uploaded"
     else
-      # Removes @new_image from the edition, otherwise the index page will attempt to render it and error
+      # Remove @new_image from @edition.images array, otherwise the view will render it in the 'Uploaded images' list
```
ollietreend, line 37: I think this change should go away if you rebase on `main` now

app/controllers/admin/edition_images_controller.rb
```diff
@@ -26,10 +26,15 @@ def create
     @new_image = @edition.images.build
     @new_image.build_image_data(image_params["image_data"])
 
+    if @new_image.image_data
+      @new_image.image_data.validate_on_edition = @edition
+      @new_image.image_data.validate_on_image = @new_image
+    end
```
ollietreend, line 32: Are there any circumstances where this `if` statement _won't_ be true? 🤔 

Considering we `build_image_data` in the line above, I'm thinking this will _always_ be present. But maybe I'm missing an edge case?

app/views/admin/edition_images/_image_upload.html.erb
```diff
@@ -10,7 +10,7 @@
   <%= render "govuk_publishing_components/components/file_upload", {
     name: "image[image_data][file]",
     id: "edition_images_image_data_file",
-    hint: "Images can be JPEG, PNG, SVG or GIF files.",
+    hint: raw("Images can be JPEG, PNG, SVG or GIF files. If you are uploading more than one image, <a class='govuk-link' href='https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos'>read the image guidance</a> on using unique file names."),
```
ollietreend, line 13: This is a nit-pick, but it's usually preferable to use double-quotes for HTML attributes.

```suggestion
    hint: raw('Images can be JPEG, PNG, SVG or GIF files. If you are uploading more than one image, <a class="govuk-link" href="https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos">read the image guidance</a> on using unique file names.'),
```

app/models/image_data.rb
```diff
@@ -47,4 +50,15 @@ def dimensions
                       Dimensions.new(image[:width], image[:height])
                     end
   end
+
+  def filename_is_unique
+    return if validate_on_edition.blank? || validate_on_image.blank? || file.blank?
+
+    edition = validate_on_edition
+    image = validate_on_image
```
ollietreend, line 58: You shouldn't actually need the Edition object because the Image will already be associated with it.

```suggestion
    return if validate_on_image.blank? || file.blank?

    edition = validate_on_image.edition
    image = validate_on_image
```

app/models/image_data.rb
```diff
@@ -47,4 +50,15 @@ def dimensions
...
+
+    if edition.images.excluding(image).joins(:image_data).exists?(["image_data.carrierwave_image = ?", filename])
```
ollietreend, line 60: Nice work implementing this as a database query 🤓 👍🏻

app/models/image_data.rb
```diff
@@ -1,6 +1,8 @@
 require "mini_magick"
 
 class ImageData < ApplicationRecord
+  attr_accessor :validate_on_edition, :validate_on_image
```
ollietreend, line 4:
```suggestion
  attr_accessor :validate_on_image
```

ollietreend (commented):
This is looking good @jyoung-gds - nice work!

Just a couple of small comments from me.

app/models/image_data.rb
ollietreend: We don't do anything specific to clean filenames – but CarrierWave does some default sanitization, so we implicitly use that.

https://github.com/carrierwaveuploader/carrierwave/blob/1531a67366f0e25e3d298133a72c81b6c9c0dc83/lib/carrierwave/sanitized_file.rb#L297-L305

test/unit/image_data_test.rb
```diff
@@ -30,6 +30,41 @@ class ImageDataTest < ActiveSupport::TestCase
...
   end
 
+  test "rejects images with duplicate filename on edition" do
+    image_data = build_example("960x640_jpeg.jpg")
+    images = [create(:image, image_data:)]
+    edition = create(:news_article, images:)
+    image = create(:image, image_data:)
+
+    image_data.validate_on_edition = edition
+    image_data.validate_on_image = image
```
ollietreend, line 40: Make sure the new `image` is associated with the existing `edition`. (Assuming you remove `validate_on_edition`)

```suggestion
    image_data = build_example("960x640_jpeg.jpg")
    edition = create(:news_article, images: [build(:image, image_data:)])
    image = build(:image, edition:, image_data:)

    image_data.validate_on_image = image
```

app/controllers/admin/edition_images_controller.rb
jyoung-gds: I think you're right, I was thinking we'd need this if we had an empty file upload for this scenario but the image data still exists in this case.

app/models/image_data.rb
jyoung-gds: With the CarrierWave sanitization, there are instances where image_data will flag a duplicate filename error on filenames which seem unique to the user - for example `"my>file.jpg"` and `"my$file.jpg"` will both sanitize to `"my_file.jpg"`. To avoid confusing the user in this scenario, @rosaryou suggested an update to the upload guidance and error message.

ollietreend (approved):
This looks good to me @jyoung-gds – nice work! ✨🚀

PR #7499, opened 2023-03-30
Send the description to publishing api as html

JonathanHallam:
This will make it so that our rendering app doesn't have to parse MD

Trello - https://trello.com/c/alzDKosv/472-add-rendering-of-field-of-operation-pages-to-government-frontend

### DIFF
--- a/test/unit/presenters/publishing_api/operational_field_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/operational_field_presenter_test.rb
@@ -5,7 +5,7 @@ class PublishingApi::OperationalFieldPresenterTest < ActiveSupport::TestCase
     @operational_field = create(
       :operational_field,
       name: "Operational Field name",
-      description: "Operational Field description",
+      description: "Operational Field description \n\n##Some title",
     )
     @fatality_notices_for_operational_field = (0..4).map { |_i| create(:published_fatality_notice, operational_field: @operational_field) }
     create(:published_fatality_notice, operational_field: create(:operational_field))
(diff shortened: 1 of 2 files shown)


test/unit/presenters/publishing_api/operational_field_presenter_test.rb
```diff
@@ -5,7 +5,7 @@ class PublishingApi::OperationalFieldPresenterTest < ActiveSupport::TestCase
     @operational_field = create(
       :operational_field,
       name: "Operational Field name",
-      description: "Operational Field description",
+      description: "Operational Field description \n\n##Some title",
```
BeckaL, line 8: It seems a bit odd here to add some extra stuff and then not actually use it anywhere because we stub the govspeak renderer. Wondering if instead we follow the pattern of e.g. the [featured document presenter test](https://github.com/alphagov/whitehall/blob/230ea28ad8e4ffa542b863f68d5e29eae8bdceed/test/unit/presenters/publishing_api/featured_documents_presenter_test.rb#L36) where it just uses the govspeak renderer in the expected content

BeckaL (commented):
Looks generally good to me, just a suggestion here about tests

test/unit/presenters/publishing_api/operational_field_presenter_test.rb
JonathanHallam: yeah, this shouldn't have been here, didn't proof my code properly. This came from when we were seeing what it looked like when run through the `govspeak_to_html` thing. 

Excellent spot on the featured document thing, we were cribbing from the[ consultation presenter test](https://github.com/alphagov/whitehall/blob/main/test/unit/presenters/publishing_api/consultation_presenter_test.rb#L409-L421), but this is much better, have sorted, thanks :)

BeckaL (approved):
LGTM!

PR #7501, opened 2023-03-30
Release 1.6 fixes for DateTime fields

davidgisbey:
## Description

Various bug fixes related to DateTime fields for the new/edit edition pages.

### Show (required) on the Date heading to indicate time is optional

image

The more i thought about it. I think we should retain the required on the fieldset for fields that are required.

Otherwise there's no way to distinguish between optional fields which require a date, but not time and optional fields that if the user chooses to provide require a date, but not time. Very open to differing opinions on this though!

image

### Remove first published on from consultations

This snuck in when i originally ported the page and is incorrect.

#### Before

image

#### After

image

#### Before

image

image

#### After 

image

#### Before

image

image

#### After

image

## Guidance for review

Per commit would probs easiest since it's broken down into one issue per commit.

## Trello card 

https://trello.com/c/D65y76ql/86-update-datetime-component-to-specify-date-is-required

### DIFF
--- a/app/views/admin/editions/_first_published_at.html.erb
+++ b/app/views/admin/editions/_first_published_at.html.erb
@@ -62,6 +62,8 @@
       legend_text: "First published",
       heading_size: "l",
       hint: "For example, 31 3 2000",
+      id: "edition_first_published_at",
+      error_message: errors_for_input(edition.errors, :first_published_at),
     } do %>
       <%= first_published_at_fields %>
     <% end %>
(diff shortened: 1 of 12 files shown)


app/views/admin/editions/_first_published_at.html.erb
```diff
@@ -39,7 +39,7 @@
       <%= render "govuk_publishing_components/components/radio", {
         name: "edition[previously_published]",
         id: "edition_previously_published",
-        error_items: errors_for(edition.errors, :previously_published),
+        error_items: errors_for(edition.errors, :previously_published) || errors_for(edition.errors, :first_published_at),
```
ollietreend, line 42: Oooh - nice catch!

PR #7505, opened 2023-03-31
Update schema for /world/embassies

BeckaL:
Previously, this was using a schema_name of 'placeholder'. Placeholder content items are not sent to router, meaning that router did not know about this content item, and when the prefix route for /world was removed, this page 404'd. Updating this to use the barebones schema for embassies to avoid this problem when we republish the /world content item. Later work to migrate this page out of being rendered by whitehall will add additional details to this content item and schema.

This also updates the branch of publishing api that is checked out for schemas to be 'main' instead of 'deployed_to_production', since the concept of a 'deployed to production' branch is no longer used in the new world.

Depends on [schemas PR](https://github.com/alphagov/publishing-api/pull/2334)

[Trello](https://trello.com/c/7kBEEWUL/517-create-content-item-for-help-and-services-around-the-world)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@
 library("govuk")
 
 REPOSITORY = 'whitehall'
-DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+DEFAULT_SCHEMA_BRANCH = 'main'
 
 node {
   govuk.setEnvar("TEST_DATABASE_URL", "mysql2://root:root@127.0.0.1:33068/whitehall_test")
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -3,7 +3,7 @@
 library("govuk")
 
 REPOSITORY = 'whitehall'
-DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+DEFAULT_SCHEMA_BRANCH = 'main'
```
JonathanHallam, line 6: Happy with all the changes here but it could be that this is better off in it's own PR, don't have massively strong feelings either way, and i guess actually the new system doesn't yet use release tags in the same way so just being in its own commit as is would be fine. Sorry for stream of consciousness, have approved

PR #7506, opened 2023-03-31
Use Search With Select Component on the New/Edit Edition pages

davidgisbey:
## Description

Now the `search-with-select` component has been merged we can use it in the new/edit edition pages.

I've left all the native selects as native selects with one exception. 

image

image

There's ~100 languages in the dropdown so it seems like a free win. I can remove it if people disagree.

I've tweaked the component so that opt groups so there's an option to pass in a blank option

image

In order to get the feature specs passing i've had to write some additional functionality into the select helper. It was a bit of a mare. More info in that commit message.

## Trello card

https://trello.com/c/fxqZiF3E/88-use-new-select-with-search-component-on-edit-edition-page

### DIFF
--- a/lib/govuk_publishing_components/presenters/select_with_search_helper.rb
+++ b/lib/govuk_publishing_components/presenters/select_with_search_helper.rb
@@ -22,6 +22,7 @@ module GovukPublishingComponents
         @select_helper = SelectHelper.new(local_assigns.except(:options, :grouped_options))
         @options = local_assigns[:options]
         @grouped_options = local_assigns[:grouped_options]
+        @include_blank = local_assigns[:include_blank]
         @local_assigns = local_assigns
       end
 
@@ -33,15 +34,17 @@ module GovukPublishingComponents
 
       def options_html
         if @grouped_options.present?
-          grouped_options_for_select(
-            transform_grouped_options(@grouped_options),
-            selected_option,
-          )
+          blank_option_if_include_blank +
+            grouped_options_for_select(
+              transform_grouped_options(@grouped_options),
+              selected_option,
+            )
         elsif @options.present?
-          options_for_select(
-            transform_options(@options),
-            selected_option,
-          )
+          blank_option_if_include_blank +
+            options_for_select(
+              transform_options(@options),
+              selected_option,
+            )
         end
       end
 
@@ -69,6 +72,12 @@ module GovukPublishingComponents
         grouped_options.map do |(group, options)|
           [group, transform_options(options)]
         end
+      end
+
+      def blank_option_if_include_blank
+        return "".html_safe if @include_blank.blank?
+
+        options_for_select([["", ""]])
       end
     end
   end
--- a/app/assets/javascripts/components/select-with-search.js
+++ b/app/assets/javascripts/components/select-with-search.js
@@ -23,6 +23,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       shouldSort: false // show options and groups in the order they were given
     })
 
+    this.module.choices = this.choices
+
     if (this.enableTracking) {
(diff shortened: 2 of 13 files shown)


lib/govuk_publishing_components/presenters/select_with_search_helper.rb
```diff
@@ -22,6 +22,7 @@ def initialize(local_assigns)
         @select_helper = SelectHelper.new(local_assigns.except(:options, :grouped_options))
         @options = local_assigns[:options]
         @grouped_options = local_assigns[:grouped_options]
+        @include_blank = local_assigns[:include_blank]
```
ollietreend, line 25: ## Question: is `include_blank` only applicable to grouped options?

`include_blank` only applies to grouped options right now.

When a blank/placeholder is needed on 'simple' options, it's manually added in the view. For example here, where we prepend a `[nil]` option:

https://github.com/alphagov/whitehall/blob/379ba86e8aaa381f17a0e5578a3231306444d3fc/app/views/admin/corporate_information_pages/_standard_fields.html.erb#L12

Do we want to take the opportunity to make this behaviour consistent? i.e. remove instances of manually adding `[nil]` in the view, and have the `SelectWithSearchHelper` add the blank/placeholder option instead?

ollietreend (commented):
This looks excellent.

Really nice work getting Capybara to play nicely with Choices.js. I know it might feel a bit clunky, but you've done a good job of adding comments on the especially 'weird' bits. I can't think of a significantly 'cleaner' way to implement it – but hopefully we won't need to interact directly with Choices.js now that this helper handles the low-level details.

I've also added a couple of nitpick comments for grammar/wording in some comments.

lib/govuk_publishing_components/presenters/select_with_search_helper.rb
davidgisbey: Yea i think you're right i think it's a good opportunity to do it. The only thing is that the native govuk-select doesn't have the option to so we will still have to use it for that

davidtrussler:
It feels like we've lost some screen reader support where the "search-with-select" component has been introduced. Specifically if the user tabs from the "Foreign language only consultation" checkbox to the languages dropdown there is no announcement as to what field they are now in, instead the unhelpful placeholder text "Select one" is read to the user. The current implementation announces that the field is the "Document language" one. Not sure at a glance what the solution would be for this would be though.

lib/govuk_publishing_components/presenters/select_with_search_helper.rb
```diff
@@ -32,7 +33,13 @@ def css_classes
...
 
       def options_html
-        if @grouped_options.present?
+        if @grouped_options.present? && @include_blank.present?
+          options_for_select([["", ""]]) +
+            grouped_options_for_select(
+              transform_grouped_options(@grouped_options),
+              selected_option,
+            )
+        elsif @grouped_options.present?
```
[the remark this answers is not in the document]
davidgisbey: There's some really weird behaviour when you do this. If you add "" to the output it renders the component with no options. No idea why 🤔 This works though

```
        if @grouped_options.present? && @include_blank.present?
          blank_option + optgroups_output
        elsif @grouped_options.present?
          optgroups_output
        elsif @options.present? && @include_blank.present?
          blank_option + options_output
        elsif @options.present?
          options_output
        end

     def blank_option
        options_for_select([["", ""]])
      end

      def optgroups_output
        grouped_options_for_select(
          transform_grouped_options(@grouped_options),
          selected_option,
        )
      end

    def options_output
        options_for_select(
          transform_options(@options),
          selected_option,
        )
      end
```

It's so bizarre!

davidgisbey:
Thanks for all the feedback @ollietreend i've pushed up the changes

ollietreend (approved):

PR #7507, opened 2023-03-31
Move edit role appointments page to GDS Design System 

chriswhitehouse-tw:
## Description

This ports the role appointments edit page to the GOV.UK Design System. Then adds the new view in the GOV.UK Design System and conditionally renders the Bootstrap or GOV.UK Design System based on whether the user has the "Preview design system" permission.

#### On landing (can delete, no speeches)
Screenshot 2023-03-31 at 15 39 33

#### On landing (can't delete, with speeches)
Screenshot 2023-03-31 at 15 37 47

#### Error (no started at date, no ended at date)
Screenshot 2023-03-31 at 15 38 19

#### Error (no started at date, but with end date)
Screenshot 2023-03-31 at 15 38 37

#### On landing (can delete, no speeches)
Screenshot 2023-04-05 at 17 00 39

#### On landing (can't delete, with speeches)
Screenshot 2023-04-06 at 11 28 46

#### On landing (single option on drop down for edit)
Screenshot 2023-04-06 at 11 31 02

#### Error (no started at date, no ended at date)
Screenshot 2023-04-05 at 17 02 03

#### Error (no started at date, but with end date)
Screenshot 2023-04-05 at 17 02 24

## Trello

https://trello.com/c/R5A32Kau

Add content_for page_title, title, and error_summary

### DIFF
--- a/app/views/admin/role_appointments/edit.html.erb
+++ b/app/views/admin/role_appointments/edit.html.erb
@@ -1,42 +1,64 @@
-<% page_title "Edit appointment to #{@role_appointment.role.name}"  %>
+<% content_for :page_title, "Edit appointment to #{@role_appointment.role.name}"  %>
+<% content_for :title, "Edit appointment to #{@role_appointment.role.name}" %>
+<% content_for :title_margin_bottom, 6 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @role_appointment)) %>
 
-<div class="row">
-  <div class="col-md-8">
-    <h1>Edit appointment to position of <%= @role_appointment.role.name %> </h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for [:admin, @role_appointment] do |form| %>
+      <%= render 'admin/role_appointments/form', form: form %>
 
-    <%= form_for [:admin, @role_appointment], html: {class: 'well'} do |form| %>
-      <%= render partial: 'admin/role_appointments/form', locals: { form: form } %>
-
-      <%= form.save_or_cancel cancel: edit_admin_role_path(@role_appointment.role) %>
+      <div class="govuk-button-group">
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Save",
+          data_attributes: {
+            module: "gem-track-click",
+            "track-category": "form-button",
+            "track-action": "#{form.object.class.name.demodulize.underscore.dasherize}-button",
+            "track-label": "Save"
+          }
+        } %>
+        <% if @role_appointment.destroyable? %>
+          <%= link_to("Delete", confirm_destroy_admin_role_appointment_path(@role_appointment), class: "govuk-link govuk-link--no-visited-state gem-link--destructive") %>
+        <% end %>
+        <%= link_to("Cancel", edit_admin_role_path(@role_appointment.role), class: "govuk-link") %>
+      </div>
     <% end %>
 
-    <% if @role_appointment.destroyable? %>
-      <%= button_to "Delete", admin_role_appointment_path(@role_appointment), method: :delete, class: "btn btn-danger" %>
-    <% else %>
-      <p>Cannot delete</p>
+    <% unless @role_appointment.destroyable? %>
+      <%= render "govuk_publishing_components/components/inset_text", {
+        text: "Note: This appointment cannot be deleted"
+      } %>
     <% end %>
   </div>
 
-  <div class="col-md-4">
+  <div class="govuk-grid-column-one-third">
     <section class="speeches">
-      <h2>Speeches during this appointment</h2>
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Speeches during this appointment",
+        margin_bottom: 4
+      } %>
 
       <% if @role_appointment.speeches.any? %>
-        <ul class="speeches">
-        <% @role_appointment.speeches.each do |speech| %>
-          <li><%= link_to speech.title, [:admin, speech] %></li>
-        <% end %>
-        </ul>
+        <%= render "govuk_publishing_components/components/list", {
+          visible_counters: true,
+          items: @role_appointment.speeches.map do |speech|
+              link_to(speech.title, [:admin, speech], class: "govuk-link")
+            end
+        } %>
       <% else %>
-        <p>None</p>
+        <p class="govuk-body">None</p>
       <% end %>
     </section>
 
-    <h2>Other appointments to this role</h2>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Other appointments to this role",
+      margin_bottom: 4
+    } %>
     <% if @role_appointment.other_appointments_for_same_role.any? %>
-      <%= render partial: "admin/role_appointments/legacy_list", locals: { role_appointments: @role_appointment.other_appointments_for_same_role }%>
+      <%= render "admin/role_appointments/list", role_appointments: @role_appointment.other_appointments_for_same_role %>
     <% else %>
-      <p>None</p>
+      <p class="govuk-body">None</p>
     <% end %>
   </div>
 </div>
--- a/app/views/admin/role_appointments/edit.html.erb
+++ b/app/views/admin/role_appointments/edit.html.erb
@@ -8,6 +8,12 @@
     <%= form_for [:admin, @role_appointment] do |form| %>
       <%= render 'admin/role_appointments/form', form: form %>
 
+      <% unless @role_appointment.destroyable? %>
+        <%= render "govuk_publishing_components/components/inset_text", {
+          text: "This appointment cannot be deleted"
+        } %>
+      <% end %>
+
       <div class="govuk-button-group">
         <%= render "govuk_publishing_components/components/button", {
           text: "Save",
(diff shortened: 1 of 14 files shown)


chriswhitehouse-tw:
Couple of areas I could do with specific guidance:

* Is the single drop down option ok for the select? I was unsure how to disable select.
* Rendering of the 'none' for speeches, and 'Cannot delete' doesn't look great. These are both p tags. Is there an alternative  styling I should be using?
* There is a server error if a user tries to submit a blank started at date, but populated ended at date. Presumably to do with model validations, but not sure why the error summary is not handling this?

app/views/admin/role_appointments/edit.html.erb
```diff
@@ -1,40 +1,74 @@
-<% page_title "Edit appointment to #{@role_appointment.role.name}"  %>
+<% content_for :page_title, "Edit appointment to #{@role_appointment.role.name}"  %>
+<% content_for :title, "Edit appointment to #{@role_appointment.role.name}" %>
+<% content_for :title_margin_bottom, 6 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @role_appointment, parent_class: "role_appointment")) %>
```
davidgisbey, line 4:
```
<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @role_appointment)) %>
```

This won't break anything, but we only need to pass a `parent_class` in when the model is inheriting from a parent class and the fields use the parent class.

For example, we use one on the edit edition page as the names of the fields are `edition[title]` etc. but the class of the objects if you call`@edition.class` is `Consultation, Case Study` etc. 

Without passing in the parent_class to the error summary it would look for fields names `consultation[title]` etc.

Let me know if that doesn't make sense

app/views/admin/role_appointments/edit.html.erb
```diff
@@ -1,40 +1,74 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for [:admin, @role_appointment] do |form| %>
+      <%= render 'admin/role_appointments/form', form: form %>
 
-    <%= form_for [:admin, @role_appointment], html: {class: 'well'} do |form| %>
-      <%= render partial: 'admin/role_appointments/form', locals: { form: form } %>
-
-      <%= form.save_or_cancel cancel: edit_admin_role_path(@role_appointment.role) %>
+      <div class="govuk-button-group">
+        <%= render "govuk_publishing_components/components/button", {
```
davidgisbey, line 12: I think we can make this a bit simpler

```
  <div class="govuk-button-group">
          text: "Save",
          data_attributes: {
            module: "gem-track-click",
            "track-category": "form-button",
            "track-action": "role-appointment-button",
            "track-label": "Save"
          }
        } %>
```

If you look a the [component](https://components.publishing.service.gov.uk/component-guide/button), all you need to provide is text and it defaults to having a type of submit etc.

Providing a name/value is only required when the controller is figuring out where to send a user submits a form when there are 2 save buttons. So it can do something like

```
if object.save 
  if params[:save] == "save and continue"
    redirect to x
  else
    redirect to x
  end
else 
  render same page with errors
end
```

app/views/admin/role_appointments/edit.html.erb
```diff
@@ -1,40 +1,74 @@
...
+            "track-label": "Save"
+          }
+        } %>
+        <%= link_to("Cancel", edit_admin_role_path(@role_appointment.role), class: "govuk-link") %>
+      </div>
     <% end %>
 
     <% if @role_appointment.destroyable? %>
-      <%= button_to "Delete", admin_role_appointment_path(@role_appointment), method: :delete, class: "btn btn-danger" %>
+      <%= form_for [:admin, @role_appointment], method: "delete" do |form| %>
```
davidgisbey, line 29: Sorry we should put this in the docs but we should add confirm delete pages for destroying objects. 

To follow the established pattern i 'd think this should have a red delete link inline between the save button and cancel link

app/views/admin/role_appointments/edit.html.erb
```diff
@@ -1,40 +1,74 @@
...
       <% end %>
     </section>
 
-    <h2>Other appointments to this role</h2>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Other appointments to this role"
+    } %>
     <% if @role_appointment.other_appointments_for_same_role.any? %>
-      <%= render partial: "admin/role_appointments/legacy_list", locals: { role_appointments: @role_appointment.other_appointments_for_same_role }%>
+      <%= render "admin/role_appointments/list", { role_appointments: @role_appointment.other_appointments_for_same_role }%>
```
davidgisbey, line 71: <%= render "admin/role_appointments/list",  role_appointments: @role_appointment.other_appointments_for_same_role %>

app/views/admin/role_appointments/edit.html.erb
```diff
@@ -1,40 +1,74 @@
...
-        <% end %>
-        </ul>
+        <%= render "govuk_publishing_components/components/list", {
+          visible_counters: true,
+          items: @role_appointment.speeches.map do |speech|
+              link_to(speech.title, [:admin, speech], class: "govuk-link")
+            end
+        } %>
       <% else %>
         <p>None</p>
```
davidgisbey, line 63: <p class="govuk-body>None</p>

test/functional/admin/role_appointments_controller_test.rb
```diff
@@ -2,7 +2,7 @@
 
 class Admin::RoleAppointmentsControllerTest < ActionController::TestCase
   setup do
-    login_as :writer
+    login_as_preview_design_system_user(:gds_editor)
```
davidgisbey, line 5:
```
  login_as_preview_design_system_user(:writer)
```

Ideally, we should keep the same level of permissions on the user so it's testing the exact same stuff.

davidgisbey:
Addressing your questions

```
Is the single drop down option ok for the select? I was unsure how to disable select.
```

If this is a common occurrence, since the component doesn't allow no blank options (i think we should build this in down the line) i think we could do something like:

```
selected:  selected: form.object.person_id == person_id || disambiguated_people_names.count == 1
```
The better option is building a `include_blank` option into the component long term though.

```
Rendering of the 'none' for speeches, and 'Cannot delete' doesn't look great. These are both p tags. Is there an alternative styling I should be using?
```

Carlos settled on this pattern on the people page

image

I think we should stick with that unless there's a good reason not to.

```
There is a server error if a user tries to submit a blank started at date, but populated ended at date. Presumably to do with model validations, but not sure why the error summary is not handling this?
```

I can se that this is caused by this validator being run and blowing up on nil
https://github.com/alphagov/whitehall/blob/main/app/models/role_appointment.rb#L36-L54

As this is doing a bunch of work to see if there are overlaps it will always blow up if the start date isn't present. There's a bunch of stuff we can do within the validator to check for presence in the conditionals.

However,  it's already going to add an error if start date isn't presence due to this line

```
validates :role_id, :person_id, :started_at, presence: true
```
so the easiest thing to do is probably only run the vallidator if start date is present

```
  validates_with Validator, if: -> {  start_date.present? }
```

app/views/admin/role_appointments/edit.html.erb
chriswhitehouse-tw: Cool, yes that makes perfect sense. I can see how the ErrorSummaryComponent handles this now.

app/views/admin/role_appointments/edit.html.erb
chriswhitehouse-tw: Cool, yes, that makes perfect sense as well. That was me being lazy and copy pasting over from the role edit page, without thinking hard about whether each of those parameters are actually required.

app/views/admin/role_appointments/edit.html.erb
chriswhitehouse-tw: cool, yup, good spot, thanks.

app/views/admin/role_appointments/edit.html.erb
chriswhitehouse-tw: cool, thank you. That looks better.

chriswhitehouse-tw:
cool, thanks for the comments @davidgisbey.

I've implemented text_inset for notes, and i've added the Validator conditional and associated test.

I've not implemented the the suggested code for the drop down, as I don't think it aligns with how the disambiguated_people_names is actually being whittled down to 1. As currently implemented, disambiguated_people_names will always return a full list, it is the select method on line 13 that is responsible to filtering this down to one person if edit page is being rendered, or a list of all people, if the new page is being rendered.

app/views/admin/role_appointments/edit.html.erb
chriswhitehouse-tw: Thanks, yup, have updated to reflect this.

app/views/admin/role_appointments/edit.html.erb
davidgisbey: No worries happens to me all  the time

app/views/admin/role_appointments/edit.html.erb
davidgisbey: Thanks! Sorry we should add this to the card when required.

davidgisbey:
Thanks for doing that Chris. I'll have a look through this first thing tomorrow am 👍

davidgisbey (approved):
lgtm!

nnagewad:
This looks good. Just wondering about the inset text. Does it appear after the date has been set? Should that be the case? For the interim could the inset text be located above the buttons? Happy to discuss with @chriswhitehouse-tw + @davidgisbey

chriswhitehouse-tw:
Thanks for the review @nnagewad. I've updated the code and the screenshots. ✅

PR #7524, opened 2023-04-05
Republish topical event page on publish of about page

BeckaL:
Previously content editors were updating about pages and were confused as to why the corresponding topical event page wasn't also being updated.

Normally this would be done by dependency resolution, but since the topical event can be created without an about page, we'd need to define this logic on create of an about page too. Alternatively we could add a reverse link in the about page, but this is the simpler option for now.

[Trello](https://trello.com/c/r99oD6Hz/557-republish-topical-events-when-their-about-pages-change)

### DIFF
--- a/app/models/topical_event_about_page.rb
+++ b/app/models/topical_event_about_page.rb
@@ -12,6 +12,8 @@ class TopicalEventAboutPage < ApplicationRecord
   validates_with SafeHtmlValidator
   validates_with NoFootnotesInGovspeakValidator, attribute: :body
 
+  after_commit :republish_topical_event_to_publishing_api
+
   searchable title: :name,
              link: :search_link,
              content: :indexable_content,
@@ -35,5 +37,9 @@ class TopicalEventAboutPage < ApplicationRecord
 
   def public_url(options = {})
     Plek.website_root + public_path(options)
+  end
+
+  def republish_topical_event_to_publishing_api
+    Whitehall::PublishingApi.republish_async(topical_event)
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/topical_event_about_page.rb
```diff
@@ -12,6 +12,9 @@ class TopicalEventAboutPage < ApplicationRecord
   validates_with SafeHtmlValidator
   validates_with NoFootnotesInGovspeakValidator, attribute: :body
 
+  after_save :republish_topical_event_to_publishing_api
+  after_destroy :republish_topical_event_to_publishing_api
```
JonathanHallam, line 16: Could use `after_commit` here to cover both cases. Would also mean we're calling `after_update` though and I'm not 100% sure what that does...

https://apidock.com/rails/ActiveRecord/Transactions/ClassMethods/after_commit

floehopper: Thanks! 👍

I think `after_save` is the equivalent of `after_create` and `after_update`, so `after_commit` does seem appropriate here.

Also using `after_commit` will mean we're guaranteeing that the change is complete in the database before calling `Whitehall::PublishingApi.republish_async` which means we're making fewer assumptions about what that code does.

floehopper: Addressed in https://github.com/alphagov/whitehall/pull/7524/commits/de3e3bb23d473fd034d9197964202ac32e9717ea

floehopper:
@JonathanHallam 

> Could you just fix everything up into one commit please :)

Done! ✅

JonathanHallam (approved):
nice one, thanks :)

PR #7526, opened 2023-04-05
Remove '/latest' route

callumknights:
Unused seemingly, removing resulted in zero test failures.

Done as part of the work to remove routes just used for helper methods out of whitehall.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -88,7 +88,6 @@ Whitehall::Application.routes.draw do
     # End of public facing routes still rendered by Whitehall
 
     # Routes no longer rendered by Whitehall, but retained to maintain the route helpers
-    get "/latest", as: "latest", to: rack_404
     get "/organisations", as: "organisations", to: rack_404
     get "/statistical-data-sets/:id", as: "statistical_data_set", to: rack_404
     get "/statistics(.:locale)", as: "statistics", to: "statistics#index", constraints: { locale: valid_locales_regex }


config/routes.rb
```diff
@@ -88,7 +88,7 @@ def external_redirect(path_prefix, target)
     # End of public facing routes still rendered by Whitehall
 
     # Routes no longer rendered by Whitehall, but retained to maintain the route helpers
```
floehopper, line 90: I see that this comment suggests that none of these paths are rendered by Whitehall, but is there any way we can verify that it is definitely the case, perhaps by looking at what's in the GOV.UK router?

I also see that this comment was introduced in f69c61bc379701483e3d6ae05874f123bed6007e which says:

> These routes exist solely so Rails creates route helpers. They are not served by Whitehall, so we should list them separately from everything else.

So perhaps that's good enough! Also the fact that they were previously responding with a 404 Not Found means we're unlikely to cause a significant problem by making this change.

floehopper (commented):
It looks as if the last use of `latest_path` was removed in 6826d2442a83b6812ddbeb5f6db48763b03c60a4. It doesn't look as if `latest_url` has ever been used. It might be worth mentioning all this in the commit note / PR description.

Otherwise this looks good to me! 👍

config/routes.rb
floehopper: Re-reading my comment I've convinced myself that even if in the unlikely event something was relying on whitehall to render this path removing this route isn't going to make things any worse. So I'll mark this as resolved.

config/routes.rb
```diff
@@ -88,7 +88,7 @@ def external_redirect(path_prefix, target)
...
-    get "/latest", as: "latest", to: rack_404
+    # get "/latest", as: "latest", to: rack_404
```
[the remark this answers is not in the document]
floehopper: This has been resolved in 817a448dba4c58998959e63b7f57bb82d79594aa.

floehopper (approved):
Looks good to me 👍

PR #7522, opened 2023-04-05
Add Policy Groups Index & Confirm Destroy Pages in the GOV.UK Design System

davidgisbey:
## Description

This PR is the first of two which transitions the Policy Groups section to the GOV.UK Design System. This PR adds the index & confirm destroy pages.

It does the following:

1. Duplicates the policy groups controller tests and assigns the non-legacy specs user the "Preview design system permission.
2. Prepends legacy to the index view for the bootstrap implementation and adds the logic to the controller to render the correct implementation based on the users permissions
3. Adds the view in the GOV.UK Design System
4. Adds a confirm delete page for policy groups (follows the established pattern)

#### Before

image

#### After 

image

image

### Confirm delete

image

image

## Trello card

https://trello.com/c/5tyvWY9S/94-edit-group

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #7528, opened 2023-04-05
Add Policy Groups New and Edit Pages in the GOV.UK Design System

davidgisbey:
## Description

Folllow on pr from  https://github.com/alphagov/whitehall/pull/7522 which added the policy groups index & confirm delete page

This PR adds the new/create policy group page in the GOV.UK Design System

#### Before

image

#### After

image

#### Before

image

#### After

image

## Trello cards

https://trello.com/c/BrKNhVe9/93-new-group
https://trello.com/c/5tyvWY9S/94-edit-group


app/views/admin/policy_groups/new.html.erb
```diff
@@ -1,13 +1,16 @@
-<% page_title "New group" %>
+<% content_for :page_title, "New group" %>
+<% content_for :title, "New group" %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @policy_group)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/warning_text", {
+      text: "Do not create groups without consulting GDS. Changes to groups appear instantly on the live site."
```
davidgisbey, line 8: I combined the two warning messages here which might be wrong @nnagewad. Adding two warnings felt a bit weird though. Let me know if you'd like it changed

nnagewad: Just saw that message. As per our slack conversation, there is a lot to explore. Do we need to incorporate the second sentence at all?

For now it's good as is, but we can explore whether it's even necessary...

app/helpers/admin/tabbed_nav_helper.rb
```diff
@@ -77,7 +77,7 @@ def policy_group_nav_items(group, current_path)
       {
         label: "Group",
         href: edit_admin_policy_group_path(group),
-        current: current_path == edit_admin_policy_group_path(group),
+        current: current_path == edit_admin_policy_group_path(group) || current_path == admin_policy_group_path(group),
```
chriswhitehouse-tw, line 80: Not sure I fully understand this. Would be good to discuss how the tabbed_nav works.

app/views/admin/policy_groups/_form.html.erb
```diff
@@ -0,0 +1,76 @@
+<%= form_for [:admin, policy_group] do |policy_group_form| %>
+  <%= render "govuk_publishing_components/components/input", {
+    label: {
+      text: "Name (required)"
+    },
+    name: "policy_group[name]",
+    id: "policy_group_name",
+    value: policy_group.name,
+    heading_size: "l",
```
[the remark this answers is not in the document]
davidgisbey: Yea i've seen that too. All of the pages in the documents section use large headings. I think either is probs fine for now, but we should make them consistent before we release stuff. One for the doc Nik made i think

app/helpers/admin/tabbed_nav_helper.rb
davidgisbey: The tabbed nav takes an array of hashes with a  label, href and current (functions a bit like selected or checked for radios/selects/checkboxes).

When current == true it gives the  tab the selected styling (blue underline)

image

This is just a helper method that returns those array of hashes. I think there's probably a subtle bug at play here which goes further than just this tab.

When an edit page is passed in as the current path, then an unsuccessful save occurs, the update action in the controller re-renders the edit view. Due to this the current path is the update actions path, not the edit path. i.e.

```
/government/admin/policy-groups/1330557
```

instead of
```
/government/admin/policy-groups/1330557/edit
```

Due to this the tab was re-rendering with no tabs showing the selected styling

CL-GDS (approved):
LGTM, I like the thread of desciptions. I will get this copied into the transition guide.

app/helpers/admin/tabbed_nav_helper.rb
chriswhitehouse-tw: Thanks @davidgisbey, that makes sense to me. It was actually the 'send' method in tabbed_nav_helper that was throwing me. But understand now.

PR #7533, opened 2023-04-10
Handle lead images in the new images tab

dnkrj:

# What
Move uploaded images to View Component.
[Pull out lead image from uploaded images.
Add button to hide or show the lead image for Case Studies.](https://trello.com/c/4Eogck71/1136-add-like-for-like-lead-image-functionality-to-images-tab)

# Why
This brings the new images tab functionality up to the existing page.
We have decided not to alter the lead image behaviour in the initial image workflow redesign.

# Screenshots
whitehall-admin dev gov uk_government_admin_editions_1374131_images(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1374131_images(iPhone 12 Pro)

### DIFF
--- a/app/components/admin/edition_images/uploaded_images_component.html.erb
+++ b/app/components/admin/edition_images/uploaded_images_component.html.erb
@@ -39,9 +39,9 @@
   } %>
 </div>
 
-<% if edition.images %>
+<% if @edition.images %>
   <ul class="govuk-list">
-    <% edition.images.each_with_index do |image, index| %>
+    <% @edition.images.each_with_index do |image, index| %>
       <li class="govuk-grid-row">
         <div class="govuk-grid-column-one-third">
           <% if image.url.present? %>
@@ -58,7 +58,7 @@
           <%= link_to("Delete image", confirm_destroy_admin_edition_image_path(@edition, image), class: "govuk-link gem-link--destructive") %>
         </div>
       </li>
-      <% if image != edition.images.last %>
+      <% if image != @edition.images.last %>
         <li aria-hidden="true"><hr class="app-view-edition-images__section-break govuk-section-break govuk-section-break--visible"></li>
       <% end %>
     <% end %>
--- a/app/components/admin/edition_images/uploaded_images_component.html.erb
+++ b/app/components/admin/edition_images/uploaded_images_component.html.erb
@@ -2,7 +2,8 @@
   text: "Uploaded images",
   font_size: "l",
 } %>
-<div>
+
+<% if can_have_lead_image? %>
   <%= render "govuk_publishing_components/components/heading", {
     text: "Lead image",
     font_size: "m",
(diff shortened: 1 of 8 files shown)


app/models/edition/first_image_pulled_out.rb
```diff
@@ -13,4 +13,8 @@ module Edition::FirstImagePulledOut
   def image_disallowed_in_body_text?(index)
     index == 1
   end
+
+  def first_image_pulled_out?
+    true
```
minhngocd, line 18: I think maybe using `@edition.include?(Edition::FirstImagePulledOut)` in the component class might be more extensible than having to define this method on all document types that will need to go through this in the future?

app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -4,4 +4,37 @@ class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
...
+      markdown: "[#{image.image_data.carrierwave_image}]",
+      links: links_for_image(image),
+    }
+  end
+
+  def links_for_image(image)
+    links = []
+    links << link_to("Edit details", edit_admin_edition_image_path(@edition, image), class: "govuk-link")
+    links << link_to("Delete image", confirm_destroy_admin_edition_image_path(@edition, image), class: "govuk-link gem-link--destructive")
+    links
```
minhngocd, line 38: Just thinking that the links are very view specific, I'm not sure it's more readable getting extracted to the ruby class rather than staying in the view.

app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -4,4 +4,37 @@ class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
...
+  def document_images
+    lead_image_count = @edition.first_image_pulled_out? ? 1 : 0
+    if @edition.images.length > lead_image_count
+      @edition.images.drop(lead_image_count).map.with_index(1) { |image, index| image_to_object image, index }
+    end
+  end
+
+private
+
+  def image_to_object(image, index)
```
minhngocd, line 23: image_to_hash for more ruby-esque convention

app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -4,4 +4,37 @@ class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
...
+
+  def lead_image
+    if @edition.first_image_pulled_out? && !@edition.images.empty?
+      image_to_object @edition.images.first, 0
+    end
+  end
+
+  def document_images
+    lead_image_count = @edition.first_image_pulled_out? ? 1 : 0
+    if @edition.images.length > lead_image_count
```
minhngocd, line 16: can this just be if @edition.includes(Edition::FirstImagePulledOut) && @edition.images.any?

app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -4,4 +4,37 @@ class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
   def initialize(edition:)
     @edition = edition
```
minhngocd, line 5: Also thinking we can simplify some logic duplication and improve readability with something like this

```
  def initialize(edition:)
    @edition = edition
    if has_lead_image?
      @lead_image = @edition.images.shift
    end
    @document_images = @edition.images
  end

  def has_lead_image?
    @edition.first_image_pulled_out? && @edition.images.any?
  end

  def lead_image
    image_to_object @lead_image, 0
  end
  
```

app/components/admin/edition_images/uploaded_images_component.html.erb
```diff
@@ -20,19 +20,39 @@
...
-      <img src="<%= lead_image[:url] %>" alt="<%= lead_image[:preview_alt_text] %>" class="app-view-edition-images__image">
-    </div>
-    <ul class="app-view-edition-images__details govuk-grid-column-two-thirds govuk-list govuk-list--bullet">
-      <li><strong>Caption: </strong><%= lead_image[:caption] %></li>
-      <li><strong>Alt text: </strong><%= lead_image[:alt_text] %></li>
-    </ul>
-    <div class="app-view-edition-images__actions govuk-grid-column-full govuk-button-group">
-      <% if @edition.type == "CaseStudy" %>
-      <% lead_image[:links].each do |link| %><%= link %><% end %>
+  <% if lead_image %>
```
minhngocd, line 23: Is there a way to separate the logics between the blocks to show the lead image and the form? feels like it's a little bit difficult to read, but I get that div nestings could have an impact to how it's rendered...

e.g.
<govuk grid div>
[ if lead image exists >> show lead image block ]
Show the form
[ if case study then show case study specific text & buttons else normal buttons ] < this  logic can also be pulled into rb file?
[ if lead image exists >> show links ]

app/components/admin/edition_images/uploaded_images_component.rb
dnkrj: I moved it to the class to avoid duplicating it in the view (between lead and document document images)
I don't think that it's an issue as this is ViewComponent class and hence directly tied to the view

app/models/edition.rb
```diff
@@ -411,6 +411,10 @@ def image_disallowed_in_body_text?(_index)
     false
   end
 
+  def first_image_pulled_out?
+    false
+  end
+
```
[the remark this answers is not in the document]
dnkrj: Thanks!

minhngocd (approved):

app/components/admin/edition_images/uploaded_images_component.html.erb
```diff
@@ -0,0 +1,90 @@
...
+
+<%= render "govuk_publishing_components/components/heading", {
+  text: "Images available to use in document",
+  font_size: "m",
+  padding: true
+} %>
+
+<% if document_images %>
+  <%= render "govuk_publishing_components/components/inset_text", {
+    text: "Use the 'Insert image' button in the document tab to add images to body copy.",
```
ollietreend, line 64: Sorry I didn't pick up on this sooner, but: isn't this undoing the change made in #7549?

We need to [avoid references to the 'Insert image' button](https://trello.com/c/vOZPG9Pt/1216-remove-references-to-insert-image-modal-because-its-not-part-of-release-16) because it isn't in this upcoming release.

dnkrj: This PR went in first - the other was based on this so I think right now we're good!
https://github.com/alphagov/whitehall/blob/main/app/components/admin/edition_images/uploaded_images_component.html.erb#LL64C45-L64C45

PR #7534, opened 2023-04-10
Display old markdown for editions with duplicate image filenames

dnkrj:

# What
[For editions with duplicate image filenames display the old markdown format e.g. !!1](https://trello.com/c/PcwkD3X1/1208-display-old-markdown-for-editions-with-duplicate-image-filenames)

# Why
Whilst we can potentially prevent new images from being uploaded with duplicated filenames, we are not updating images that have already been uploaded.


minhngocd (approved):
only request is to maybe put the test and the implementation in 1 commit?

app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -26,11 +26,16 @@ def image_to_object(image, index)
...
-      markdown: "[#{image.image_data.carrierwave_image}]",
+      markdown: unique_names? ? "[#{image.filename}]" : "!!#{index}",
       links: links_for_image(image),
     }
   end
 
+  def unique_names?
+    names = @edition.images.map(&:filename)
+    names.uniq.length == names.length
+  end
```
ollietreend, line 37: Let's [memoize](https://michaeljherold.com/articles/memoization-in-ruby/) this value to avoid an exponential loop. Otherwise we'll end up running `@edition.images.map` once for every image on the edition.

Since this is a boolean value, we can't use the usual `@unique_names ||=` approach to memoizing as that won't work for `false` values.

Instead, how about something like this:

```ruby
  def unique_names?
    return @unique_names if defined? @unique_names

    names = @edition.images.map(&:filename)
    @unique_names = (names.uniq.length == names.length)
  end
```

Or alternatively, a less 'clever' solution is just to run the method once in the class constructor, and then use `@unique_names` everywhere else:

```ruby
  def initialize(edition:)
    @edition = edition
    @unique_names = unique_names?
  end
```

ollietreend (commented):
This looks good, but I've just spotted one potential performance issue.

It'd be great to resolve it before this gets merged in.

app/components/admin/edition_images/uploaded_images_component.rb
dnkrj: Good point, thanks Ollie, have updated the code

PR #7541, opened 2023-04-12
73 take part page to design system

davidtrussler:
[Trello](https://trello.com/c/elfoyfNE/73-add-new-take-part-page)

The changes in this PR port the "Take part" landing page to the design system. Screenshots below.

Some points to note: 
- Although the field for the image description (alt text) is currently marked as required this does not return an error when the user saves the page without that so it appears to not be required. 
- The current implementation includes a warning to the user when they exceed 160 characters in the summary field that is not included in the Design System port. This is consistent with other pages that include a character count though the actual limit is inconsistent across Whitehall. The error that is generated when the text is too long is actually when that text exceeds 255 characters so the hint text here reflects that. 
- Additionally the title field is subject tot he same limit so a hint has been added to that. 

| Bootstrap | Design System |
| - | - |
| Screenshot 2023-04-12 at 11 51 28 | Screenshot 2023-04-12 at 11 52 46 |

### DIFF
--- a/app/views/admin/take_part_pages/_form.html.erb
+++ b/app/views/admin/take_part_pages/_form.html.erb
@@ -1,24 +1,92 @@
-<div class="row">
-  <div class="col-md-8">
-    <p class="warning">
-      Warning: changes to take part pages appears instantly on the live site.
-    </p>
+<div class="govuk-grid-row">
+  <section class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/warning_text", {
+      text: "Changes to take part pages appears instantly on the live site."
+    } %>
 
-    <%= form_for take_part_page, url: [:admin, take_part_page] do |form| %>
-      <%= form.errors %>
-      <%= form.text_field :title %>
-      <%= form.text_area :summary, rows: 2, class: 'js-summary-length-counting', data: { 'count-message-threshold' => 160, 'count-message-selector' => '.summary-length-info' } %>
-      <div class="summary-length-info">Summary text should be 160 characters or fewer. <span class="count"></span></div>
-      <%= form.text_area :body, rows: 20, class: "previewable", required: true, data: {
-        module: "paste-html-to-govspeak"
+    <%= form_for take_part_page, url: [:admin, take_part_page], multipart: true do |form| %>
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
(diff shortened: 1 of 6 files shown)


nnagewad:
This looks great. I am wondering for consistency purposes there should be a character limit notice for title as well.

nnagewad:
As per our conversation. Use hint text with 255 character limit for the time being.

Overall, great job @davidtrussler

CL-GDS (approved):
Other then changing `edit` line #4  
   From: text: "Warning: changes to take part pages appears instantly on the live site."
   to:       text: "Changes to take part pages appears instantly on the live site."
   
   LGTM

davidtrussler:
> 

Ah yes, good spot - the component appends the word "warning" for screen readers so it's not needed in the text 👍

app/views/admin/take_part_pages/_form.html.erb
```diff
@@ -1,24 +1,92 @@
...
+      <%= render "govuk_publishing_components/components/textarea", {
+        label: {
+          text: "Summary (required)",
+          heading_level: 2,
+          heading_size: "l",
+        },
+        name: "take_part_page[summary]",
+        id: "take_part_page_summary",
+        rows: 2,
+        hint: "Summary text should be 255 characters or fewer.",
```
davidgisbey, line 30: i can see above it says `      <div class="summary-length-info">Summary text should be 160 characters or fewer. <span class="count"></span></div>
`

Should this be 160?

davidtrussler: > i can see above it says ` <div class="summary-length-info">Summary text should be 160 characters or fewer. <span class="count"></span></div>`
> 
> Should this be 160?

There's a few inconsistencies around this across Whitehall. I discussed this with @nnagewad yesterday: The validation when the form is submitted reports an error when the user inputs > 255 characters on both the summary and the title so we've added a hint for both to that effect (see comments above). It's something we should look at though - we haven't implemented a client-side validation here at all (as exists on bootstrap) for now because of these inconsistencies. Also 255 characters seems somewhat overlong for a title.

davidgisbey: Oh right apologies i missed that. LGTM then!

davidgisbey (approved):

PR #7543, opened 2023-04-12
DRY-up publish classes

brucebolt:
There are a number of classes that present non-editionable and non-PublishedToPublishingApi content to Publishing API.

These all have exactly the same behaviour, so updating to have only one which accepts the presenter name as an argument.

This change was first introduced in https://github.com/alphagov/whitehall/pull/7472, but reverted as part of https://github.com/alphagov/whitehall/pull/7498 (to resolve an incident).  As this is unrelated to the incident, this code is safe to remain in place.


floehopper (approved):
LGTM 👍

PR #7547, opened 2023-04-12
Update presenter for ministers index page

brucebolt:
The content item for the ministers index page currently contains a lot of information (e.g. all historic ministers) but lacks other information needed to render the page outside of Whitehall (e.g. the types of whip).  The amount of information is so vast that browser-based JSON formatters fail to format the page.

Therefore removing the inverse link from the role content item (to stop all this content appearing in the index page) and updating the presenter to include only the required information.

This is dependent on https://github.com/alphagov/publishing-api/pull/2341 and https://github.com/alphagov/publishing-api/pull/2344.

Once merged, the following rake task will need to be run to publish the updated version of this page:
```
rake publishing_api:republish:republish_ministers_index
```

[Trello card](https://trello.com/c/SFVenvAi)

### DIFF
--- a/app/presenters/publishing_api/ministers_index_presenter.rb
+++ b/app/presenters/publishing_api/ministers_index_presenter.rb
@@ -35,6 +35,11 @@ module PublishingApi
         ordered_cabinet_ministers: ordered_cabinet_ministers_content_ids,
         ordered_also_attends_cabinet: ordered_also_attends_cabinet_content_ids,
         ordered_ministerial_departments: ordered_ministerial_departments_content_ids,
+        ordered_house_of_commons_whips: ordered_whips_content_ids(Whitehall::WhipOrganisation::WhipsHouseOfCommons),
+        ordered_junior_lords_of_the_treasury_whips: ordered_whips_content_ids(Whitehall::WhipOrganisation::JuniorLordsoftheTreasury),
+        ordered_assistant_whips: ordered_whips_content_ids(Whitehall::WhipOrganisation::AssistantWhips),
+        ordered_house_lords_whips: ordered_whips_content_ids(Whitehall::WhipOrganisation::WhipsHouseofLords),
+        ordered_baronessess_and_ladies_in_waiting_whips: ordered_whips_content_ids(Whitehall::WhipOrganisation::BaronessAndLordsInWaiting),
       }
     end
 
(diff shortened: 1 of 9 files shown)


app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -49,9 +49,7 @@ def details
     end
 
     def base_path
-      return "/government/ministers" if I18n.locale == :en
-
-      "/government/ministers.#{I18n.locale}"
+      "/government/ministers"
```
chrislo, line 61: I think it would be worth explaining why we only present and render this page in English in the commit note if we know - is that a policy, or in the styleguide or something?

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -51,5 +54,40 @@ def details
     def base_path
       "/government/ministers"
     end
+
+    def reshuffle_in_progress
+      SitewideSetting.find_by(key: :minister_reshuffle_mode)
```
chrislo, line 65: could you do `SitewideSetting.find_by(key: :minister_reshuffle_mode) || false` or similar here to avoid needing to `&` try in each of the call sites of this method?

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -89,5 +90,15 @@ def ordered_also_attends_cabinet
 
       sorted_ministers.map(&:content_id)
     end
+
+    def ordered_ministerial_departments
```
chrislo, line 99: it looks like this method returns content_ids so perhaps `ordered_ministerial_department_content_ids` would be a better name.

app/presenters/publishing_api/ministers_index_presenter.rb
```diff
@@ -100,5 +105,16 @@ def ordered_ministerial_departments
         .uniq
         .map(&:content_id)
     end
+
+    def ordered_whips(whip_type)
```
chrislo, line 109: This method returns content ids, so perhaps `ordered_whip_content_ids` or something would be a better name?

chrislo (approved):
Looks good to me. A couple of minor suggestions, but I wouldn't let those stop you from merging this.

app/presenters/publishing_api/ministers_index_presenter.rb
brucebolt: Will add this information to the commit message (most roles don't have translations).

app/presenters/publishing_api/ministers_index_presenter.rb
brucebolt: Nice idea.  I'll change this.

app/presenters/publishing_api/ministers_index_presenter.rb
brucebolt: Good idea.  I'll change the method names.

PR #7550, opened 2023-04-13
Remove ministers index page from links in roles

brucebolt:
At the moment, we are including a link to the ministers index page in the links for every ministerial role (including past roles).

This results in a reverse link in the ministers index page content item that has become too large to work with.  The content item is not currently being used by any frontend applications.

In later work, this will be replaced with a link to the current and relevant ministers only in the index page's content item.

We are having to make this change in two steps (step one: empty the array, step two: remove the `ministerial` link completely), as patching the links won't remove the link if the key is not included. A later PR will remove the key completely from the presenter.

After deployment, all roles will need to be republished using the following rake task:

```
publishing_api:bulk_republish:document_type[Role]
```

[Trello card](https://trello.com/c/SFVenvAi)

### DIFF
--- a/app/presenters/publishing_api/role_presenter.rb
+++ b/app/presenters/publishing_api/role_presenter.rb
@@ -30,7 +30,7 @@ module PublishingApi
     def links
       {
         ordered_parent_organisations: item.organisations.pluck(:content_id).compact,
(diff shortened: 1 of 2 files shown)


BeckaL (approved):
Nice, looks good to me, will be nice to get the ministers content item down to manageable size :-D

BeckaL:
Everything republished now on integration and production

PR #7552, opened 2023-04-13
72 take part page

davidtrussler:
[Trello](https://trello.com/c/isQwsOg9/72-take-part-page)

Ports the "Take part" landing page, update order page and delete page to the design system. Screenshots below. 

| Page | Design System |
|-|-|
| Landing page | Screenshot 2023-04-20 at 14 38 28 |
| Update order page | Screenshot 2023-04-19 at 12 38 43 |
| Delete page | Screenshot 2023-04-19 at 12 31 31 |

### DIFF
--- a/app/controllers/admin/take_part_pages_controller.rb
+++ b/app/controllers/admin/take_part_pages_controller.rb
@@ -8,6 +8,7 @@ class Admin::TakePartPagesController < Admin::BaseController
 
   def index
     @take_part_pages = TakePartPage.in_order
+    render_design_system(:index, :legacy_index, next_release: false)
   end
 
   def new
@@ -37,10 +38,18 @@ class Admin::TakePartPagesController < Admin::BaseController
     end
   end
 
+  def confirm_destroy
+    @take_part_page = TakePartPage.friendly.find(params[:id])
+  end
+
   def destroy
     @take_part_page = TakePartPage.friendly.find(params[:id])
     @take_part_page.destroy!
     redirect_to [:admin, TakePartPage], notice: %(Take part page "#{@take_part_page.title}" deleted!)
+  end
+
+  def update_order
+    @take_part_pages = TakePartPage.in_order
   end
 
   def reorder
@@ -53,7 +62,7 @@ private
 
   def get_layout
     design_system_actions = []
-    design_system_actions += %w[new create] if preview_design_system?(next_release: false)
+    design_system_actions += %w[new create index confirm_destroy update_order] if preview_design_system?(next_release: false)
 
     if design_system_actions.include?(action_name)
       "design_system"
--- a/app/views/admin/take_part_pages/index.html.erb
+++ b/app/views/admin/take_part_pages/index.html.erb
@@ -1,38 +1,63 @@
-<% page_title 'Get involved' %>
+<% content_for :page_title, "Get involved" %>
+<% content_for :title, "Get involved" %>
+<% content_for :title_margin_bottom, 4 %>
 
-<div class="get-involved-header">
-  <h1>Get involved</h1>
-  <%= link_to "View on website", get_involved_url %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <p class="govuk-body">
+      <%= link_to "View on website", get_involved_url, class: "govuk-link", target: "_blank" %>
+    </p>
+  </div>
 </div>
 
-<section class="get-involved">
-  <%= get_involved_tab_navigation %>
-
-  <h2>Take part pages</h2>
-  <p class="warning">
-    Warning: changes to take part page ordering appears instantly on the live site.
-  </p>
-
-  <% if @take_part_pages.any? %>
-    <%= form_tag reorder_admin_take_part_pages_path, method: :post do %>
-      <fieldset class="sortable">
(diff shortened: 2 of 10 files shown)


nnagewad:
Had a great chat with @davidtrussler.

Some minor changes:

- Update width to full as reorder component only works for full-width container
    - Will need to revisit the component itself and make necessary updates
- Make the "Update order" button primary
- Make "Add new take part page" a link and have it right next to the primary button

nnagewad:
as per conversation on slack, designs have been updated.

davidtrussler:
> 

@nnagewad I've updated the styles on the Edit/Delete links in the table to match the example you gave. That will be more consistent and looks better I think. I've updated the screenshot for the landing page above ☝️

nnagewad:
Looks great @davidtrussler

app/views/admin/take_part_pages/index.html.erb
```diff
@@ -1,38 +1,59 @@
-<% page_title 'Get involved' %>
+<% content_for :page_title, "Get involved" %>
+<% content_for :title, "Get involved" %>
+<% content_for :title_margin_bottom, 4 %>
 
-<div class="get-involved-header">
-  <h1>Get involved</h1>
-  <%= link_to "View on website", get_involved_url %>
+<div class="govuk-grid-row">
```
davidgisbey, line 5: Does this need to live inside a row as it's full width anyway?

app/views/admin/take_part_pages/index.html.erb
```diff
@@ -1,38 +1,59 @@
...
+  <div class="govuk-grid-column-full">
+    <p class="govuk-body">
+      <%= link_to "View on website", get_involved_url, class: "govuk-link", target: "_blank" %>
```
davidgisbey, line 8: Does this need to live inside a row as it's full width anyway?

app/views/admin/take_part_pages/index.html.erb
```diff
@@ -1,38 +1,59 @@
...
 </div>
 
-<section class="get-involved">
-  <%= get_involved_tab_navigation %>
-
-  <h2>Take part pages</h2>
-  <p class="warning">
-    Warning: changes to take part page ordering appears instantly on the live site.
-  </p>
+<%= render "components/secondary_navigation", {
```
davidgisbey, line 13: On other pages that use 2/3 width this is in the 2/3 column. I think we need to do that here

app/controllers/admin/take_part_pages_controller.rb
```diff
@@ -53,7 +62,7 @@ def reorder
 
   def get_layout
     design_system_actions = []
-    design_system_actions += %w[new create] if preview_design_system?(next_release: false)
+    design_system_actions += %w[new create index edit confirm_destroy update_order] if preview_design_system?(next_release: false)
```
davidgisbey, line 65: I think edit might have snuck in here by accident  

Should it be
```
%w[new create index confirm_destroy update_order]

```

app/views/admin/take_part_pages/index.html.erb
```diff
@@ -1,38 +1,59 @@
...
+      <div class="app-view-take-part__page-list">
+        <%= render "govuk_publishing_components/components/table", {
+          rows: @take_part_pages.map do |take_part_page|
+            [
+              {
+                text: take_part_page.title
+              },
+              {
+                text: (
+                  link_to("Edit", [:edit, :admin, take_part_page]) +
```
davidgisbey, line 47: this is missing the `govuk-link` class

app/views/admin/take_part_pages/index.html.erb
```diff
@@ -1,38 +1,64 @@
...
-  <% else %>
-    <p class="no-content no-content-bordered">No take part pages</p>
-  <% end %>
+    <% if @take_part_pages.any? %>
+      <%= form_tag reorder_admin_take_part_pages_path, method: :post do %>
+        <%= render "govuk_publishing_components/components/reorderable_list", {
+          items: @take_part_pages.map do |take_part_page|
+            {
+              id: take_part_page.id,
+              title: link_to(take_part_page.title, [:edit, :admin, take_part_page])
```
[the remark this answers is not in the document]
davidtrussler: Yes, good point. And the delete link needs it as well in fact (for focus styles at least).

app/views/admin/take_part_pages/index.html.erb
davidtrussler: I guess strictly speaking it doesn't, but it seems to be the established pattern to use `govuk-grid-column-full` inside `govuk-grid-row` where we have full width elements, so I'd be inclined to leave it like this here as well.

app/views/admin/take_part_pages/index.html.erb
davidtrussler: Yes, I think that's correct. Actually it isn't that way in the design for the page but it is more consistent, not least with the "Get involved" tab.

app/views/admin/take_part_pages/index.html.erb
davidtrussler: Yes, good point. And the delete link needs it as well in fact (for focus styles at least).

davidgisbey (approved):
lgtm!

PR #7553, opened 2023-04-13
Update caption/credit/alt text guidance

dnkrj:

# What

[Rename the caption field to cover both caption and credit.](https://trello.com/c/Z0OYnxvo/1209-update-copy-around-caption-credit-field)
Update the alt text copy.
https://docs.google.com/document/d/1c8Cd8TBXm0wSsUdrleWWbGZXy6FddlyKwC5ScQonEo0/edit#

# Why
We are not adding a credit field to Whitehall

# Screenshots
whitehall-admin dev gov uk_government_admin_editions_1374131_images_445118_edit(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1374131_images_445118_edit(iPhone 12 Pro)

### DIFF
--- a/app/views/admin/edition_images/edit.html.erb
+++ b/app/views/admin/edition_images/edit.html.erb
@@ -11,21 +11,21 @@
 
       <%= render "govuk_publishing_components/components/textarea", {
         label: {
-          text: "Caption",
+          text: "Caption and credit",
           heading_size: "l",
         },
         name: "image[caption]",
         id: "image_caption",
         value: image.caption,
-        rows: 5,
-        hint: tag.p("Name a person in a photo or label a graph, infographic or diagram. The caption appears next to the image.", class: "govuk-!-margin-bottom-0 govuk-!-margin-top-0"),
+        rows: 2,
+        hint: raw("Name a person in a photo or add image credit. <a class='govuk-link' href='https://www.gov.uk/guidance/content-design/images'>Read the image copyright standards</a>. The text appears next to the image."),
         margin_bottom: 4,
       } %>
 
       <% if image.alt_text.blank? %>
         <h2 class="govuk-heading-l">Alt text</h2>
 
-        <p class="govuk-body">If you need to give an alt text description, write it in the body copy below the image so it’s available to everyone.</p>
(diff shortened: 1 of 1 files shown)


app/views/admin/edition_images/edit.html.erb
```diff
@@ -11,21 +11,21 @@
...
+          text: "Caption and credit",
           heading_size: "l",
         },
         name: "image[caption]",
         id: "image_caption",
         value: image.caption,
-        rows: 5,
-        hint: tag.p("Name a person in a photo or label a graph, infographic or diagram. The caption appears next to the image.", class: "govuk-!-margin-bottom-0 govuk-!-margin-top-0"),
+        rows: 2,
+        hint: raw("Name a person in a photo or add image credit. <a class='govuk-link' href='//www.gov.uk/guidance/content-design/images'>Read the image copyright standards</a>. The text appears next to the image."),
```
jyoung-gds, line 21: Stylistically this differs from the rest of Whitehall where all links are prefixed with `https` - a minor thing and something I'll leave at your discretion whether or not to change because the behaviour is the same.

app/views/admin/edition_images/edit.html.erb
```diff
@@ -43,7 +43,7 @@
         } %>
       <% end %>
 
-      <p class="govuk-body"><a href="https://www.gov.uk/guidance/content-design/images" class="govuk-link" target="_blank">How to use alt text (opens in new tab)</a></p>
+      <p class="govuk-body"><a href="//www.gov.uk/guidance/content-design/images" class="govuk-link" target="_blank">Using informative and decorative images on GOV.UK (opens in new tab)</a></p>
```
jyoung-gds, line 46: Same thing here with the `https` prefix

jyoung-gds (approved):
I've left a couple of comments pointing out some minor styling differences with the rest of Whitehall, it's up to you whether to update them or not because the behaviour is the same with or without. Approved so you don't need to request a re-review.

app/views/admin/edition_images/edit.html.erb
dnkrj: Ah, good point about the consistency, have added the protocol 👍

PR #7558, opened 2023-04-14
Fix bug with rejected_by returning the wrong user

davidgisbey:
## Description

At the moment, rejected_by returns the last user to update a rejected edition, rather than the user who first rejected the user.

It grabs all the audits where the state is rejected and orders them desc, then selects the first one.

This means that the audit with the highest id is returned. If we switch the ordering the asc it returns
the user who first rejected the edition.

## Zendesk ticket

https://govuk.zendesk.com/agent/tickets/5298020


dnkrj (approved):
Nice one

PR #7574, opened 2023-04-20
Add audit history for non-editionable models

jkempster34:
https://trello.com/c/7snah8vf

This is a proof of concept to add an audit history to non-editionable models. 

We need to remove the rendering of Worldwide organisation pages away from Whitehall. Currently there’s a load of complexity in those pages because they get the majority of their content from an attached Corporate information page. These CI pages are editionable and hold history, FCDO don’t want to lose this entirely so we’ve agreed to add some basic history functions to them.

This uses the existing `Edition::AuditTrail` model in order to persist the audit history, despite it's namespace it is quite generic and is not Edition specific. Advantages of this are (a) that we can re-use the existing Version model; (b) we can make use of the existing before filter (setting the user) and callbacks (storing the event & author); and (c) we’re not introducing another audit trail implementation. A couple of downsides are (a) the name is confusing (because it’s in the Edition namespace), but we could change that; and (b) we’d need to implement WorldwideOrganisation#state, but we could hard-code that to something like “non-editionable”

However, the view components are new, as the existing view components rely on Edition specific functionality. This PR does include pagination, but for the moment, it's not JavaScript pagination (which editionable audit history is); if we wanted to do this, we may have to create a separate (simpler) implementation. As this is a proof of concept, we haven't spent much time on styling.

We would also possibly want to backfill the audit history of the existing Worldwide Organisations.

Ultimately, we want to know whether this approach of reusing the `Edition::AuditTrail`, while creating new view components makes sense, or if there may be a more sensible path to take?

image

### DIFF
--- a/app/controllers/admin/worldwide_organisations_controller.rb
+++ b/app/controllers/admin/worldwide_organisations_controller.rb
@@ -1,4 +1,6 @@
 class Admin::WorldwideOrganisationsController < Admin::BaseController
+  VERSIONS_PER_PAGE = 2
+
   respond_to :html
 
   before_action :find_worldwide_organisation, except: %i[index new create]
@@ -28,7 +30,12 @@ class Admin::WorldwideOrganisationsController < Admin::BaseController
     respond_with :admin, @worldwide_organisation
   end
 
-  def show; end
+  def show
+    @versions = Kaminari
+                  .paginate_array(@worldwide_organisation.versions_desc)
+                  .page(params[:page])
+                  .per(VERSIONS_PER_PAGE)
+  end
 
   def access_info
     @access_and_opening_times = @worldwide_organisation.access_and_opening_times ||
--- a/app/models/worldwide_organisation.rb
+++ b/app/models/worldwide_organisation.rb
@@ -39,6 +39,10 @@ class WorldwideOrganisation < ApplicationRecord
 
   include PublishesToPublishingApi
 
+  attr_accessor :state
+
+  include Edition::AuditTrail
+
   extend FriendlyId
   friendly_id
 
--- a/app/components/admin/model_history_component.html.erb
+++ b/app/components/admin/model_history_component.html.erb
@@ -1,0 +1,7 @@
+<div>
+  <h4><%= action %></h4>
+
+  <p>
+    <%= time %> by <%= actor %>
+  </p>
+</div>
--- a/app/components/admin/model_history_component.rb
+++ b/app/components/admin/model_history_component.rb
@@ -1,0 +1,29 @@
+class Admin::ModelHistoryComponent < ViewComponent::Base
+  include ApplicationHelper
+
+  attr_reader :version
+
+  def initialize(version:)
+    @version = version
+  end
+
+private
+
+  def action
+    case version.event
+    when "update" then "Document updated"
+    when "create" then "Document created"
+    else "Unknown action"
+    end
+  end
+
+  def actor
+    user = version.user
+
+    user ? linked_author(user, class: "govuk-link") : "User (removed)"
+  end
+
+  def time
+    absolute_time(version.created_at)
+  end
+end
--- a/app/components/admin/model_history_list_component.html.erb
+++ b/app/components/admin/model_history_list_component.html.erb
@@ -1,0 +1,12 @@
+<div class="">
+  <%= link_to_previous_page(versions, 'Newer') %>
+  <%= link_to_next_page(versions, 'Older') %>
+</div>
+
+<ul class="list-unstyled">
+  <% versions.each do |version| %>
+    <li>
+      <%= render Admin::ModelHistoryComponent.new(version: version) %>
+    </li>
+  <% end %>
+</ul>
--- a/app/components/admin/model_history_list_component.rb
+++ b/app/components/admin/model_history_list_component.rb
@@ -1,0 +1,7 @@
+class Admin::ModelHistoryListComponent < ViewComponent::Base
+  attr_reader :versions
+
+  def initialize(versions:)
+    @versions = versions
+  end
+end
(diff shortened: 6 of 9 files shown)


app/controllers/admin/worldwide_organisations_controller.rb
```diff
@@ -28,7 +30,12 @@ def update
     respond_with :admin, @worldwide_organisation
   end
 
-  def show; end
+  def show
+    @versions = Kaminari
+                  .paginate_array(@worldwide_organisation.versions_desc)
+                  .page(params[:page])
+                  .per(VERSIONS_PER_PAGE)
```
ollietreend, line 37: We should avoid using `Kaminari.paginate_array` if at all possible.

That method is designed to handle ['dumb' Array objects](https://github.com/kaminari/kaminari#paginating-a-generic-array-object) which hold an entire result set and need to be sliced up in-memory. It has caused significant performance issues in the past on Documents with large edit histories. That's why we [stopped using it in the Editions controller](https://github.com/alphagov/whitehall/pull/6706/commits/5bd53cb37cc4812949d4b55553dcb8cbbe00e832) when we needed to [improve page load performance](https://trello.com/c/SKFiAakd/298-handle-large-change-histories-more-efficiently-in-whitehall).

In practice, `paginate_array` will load _every_ Version from the database into memory, and then immediately discard those which aren't needed for the current page.

Instead, we should be able to rely on Kaminari's ability to [paginate ActiveRecord results](https://github.com/kaminari/kaminari#query-basics) at the database layer – i.e. by using SQL's LIMIT and OFFSET clauses.

**This is untested**, but I expect the following will work:

```suggestion
    @versions = @worldwide_organisation.versions_desc
                  .page(params[:page])
                  .per(VERSIONS_PER_PAGE)
```

app/models/worldwide_organisation.rb
```diff
@@ -39,6 +39,10 @@ class WorldwideOrganisation < ApplicationRecord
 
   include PublishesToPublishingApi
 
+  attr_accessor :state
+
+  include Edition::AuditTrail
```
ollietreend, line 44: To build upon @floehopper's [comment](https://github.com/alphagov/whitehall/pull/7574#issuecomment-1516578688) and @davidgisbey's [comment in Slack](https://gds.slack.com/archives/C02L13S214K/p1682011364360699?thread_ts=1682005365.826279&cid=C02L13S214K):

Hardcoding `state: "non-editionable"` seems like a bit of a hack – necessary to work around the fact that `Edition::AuditTrail` expects the object to have a `state`, but if we're honest it's probably a bit of a smell.

Instead, I'd suggest we avoid storing a `state` for non-editioned models altogether.

In fact, it looks like the `state` field is already optional. The [database schema allows NULL values](https://github.com/alphagov/whitehall/blob/a691d049a8f54a649b35a9a1bc4fa634c09d5895/db/schema.rb#L1082), and the [Version model](https://github.com/alphagov/whitehall/blob/ad86a5ea8d7538f787c71cb586adb1b771ea08d9/app/models/version.rb) doesn't specify validation rules on that field.

### Make `Edition::AuditTrail` more generic

We're starting to use `Edition::AuditTrail` for something slightly broader than what it was originally designed for. To date it's only been used on Edition-type models – so it has always been safe to assume they had a `state`.

Now that we're using it on non-Edition-type models, I'd suggest we consider making it more generic.

1. Rename `Edition::AuditTrail` to `AuditTrail`. I think we should be able to store this at `app/models/concerns/audit_trail.rb`.
2. Stop assuming that `state` is defined on the model – fall back to `nil` if it isn't present

```diff
  def record_create
    user = Edition::AuditTrail.whodunnit
+   state = self.try(:state)
    versions.create!(event: "create", user:, state:)
    alert!(user)
  end
```

### Beware other Edition-specific logic

I also want to point out the `Edition::AuditTrail#alert!` method which assumes it's dealing with an Edition-type model:

https://github.com/alphagov/whitehall/blob/73e1f38e0c64e505a6f1fbc5f54bb940447a5f05/app/models/edition/audit_trail.rb#L55-L59

We should consider this when making `AuditTrail` more generic.

ollietreend (commented):
This is looking good – I think you're on the right track by re-using existing behaviour.

Just pointing out a couple of things which I think might need addressing.

app/models/worldwide_organisation.rb
jkempster34: I wonder if those alerts could be shifted elsewhere, maybe into another concern `Edition::Alerts`

app/controllers/admin/worldwide_organisations_controller.rb
jkempster34: Thanks for this. This was a TODO as I couldn't remember how to do it, but what you have wrote there works great! ⭐

app/models/worldwide_organisation.rb
ollietreend: Hmm 🤔 having taken a closer look, I think this might actually be dead code.

The `CO_NSS_WATCHKEEPER_EMAIL_ADDRESS` environment variable is not set in any of our live environments.

I'll double check with the team, but suspect the `alert!` functionality could be left alone in this PR, and a separate clean-up card created to remove it.

ollietreend: That seems a fair point @jkempster34. Feel free to leave that alert functionality as is for now - we may well open a card to refactor it into a more appropriate place later.

floehopper: Even though ‘edition_published_by_monitored_user’ has the word edition in the method name, the only parameter that is passed is the ‘user’ and I think it’s possible/likely that whoever needs to be alerted about changes made by the special user might well want to be alerted about *any* changes made by that special user. I think the method is badly named.

jkempster34:
Closing this. Opened a more complete PR here https://github.com/alphagov/whitehall/pull/7583

PR #7576, opened 2023-04-24
Ensure speech params are cleaned in editions create action

davidgisbey:
## Description

At the moment, the clear_role_appointment_param_on_override is called from the SpeechesController. As the SpeechesController inherits from the EditionsController it runs the before_action after the before_action's in the EditionsController.

This means by the time the params for the delivered by are cleaned the edition has already been built.

By moving this to the clean_edition_parameters before_action in the EditionsController the params get cleaned prior to assignment.

## Vid of behaviour before

https://user-images.githubusercontent.com/42515961/234551038-37ccb4ac-6ced-43d1-9f70-50483dc94fb8.mov

## Vid of behaviour after

https://user-images.githubusercontent.com/42515961/234550635-a0e3e057-8925-436c-b93f-8693957c47ef.mov

## Trello card

https://trello.com/c/K8rPB3wP/85-speaker-field-clear-unused-conditional-reveal-fields-on-save

### DIFF
--- a/app/controllers/admin/speeches_controller.rb
+++ b/app/controllers/admin/speeches_controller.rb
@@ -1,21 +1,22 @@
 class Admin::SpeechesController < Admin::EditionsController
-  before_action :clear_role_appointment_param_on_override, only: %i[update create]
-
 private
 
   def edition_class
     Speech
   end
 
-  def clear_role_appointment_param_on_override
+  def clean_edition_parameters
+    super
     design_system = preview_design_system?(next_release: false)
 
-    if design_system && params[:speaker_radios] == "yes"
+    if design_system && edition_params[:speaker_radios] == "yes"
       edition_params[:person_override] = nil
-    elsif design_system && params[:speaker_radios] == "no"
+    elsif design_system && edition_params[:speaker_radios] == "no"
       edition_params[:role_appointment_id] = nil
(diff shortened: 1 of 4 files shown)


app/controllers/admin/speeches_controller.rb
```diff
@@ -1,6 +1,4 @@
 class Admin::SpeechesController < Admin::EditionsController
-  before_action :clear_role_appointment_param_on_override, only: %i[update create]
```
ollietreend, line 2: If you want to avoid putting speech-specific stuff in the main Editions controller, you could always extend the `clean_edition_parameters` method in Speeches controller instead.

```ruby
class Admin::SpeechesController < Admin::EditionsController
  def clean_edition_parameters
    super
    clear_role_appointment_param_on_override
  end

private

  # ...
end
```

I'm undecided whether it's better for this to live in Editions controller or in Speeches controller. I don't have a strong opinion either way – so feel free to ignore & leave it as is, if you think that's better.

ollietreend (approved):
Nice work @davidgisbey 👍🏻 

This looks good to me. I've left one potential alternative approach for your consideration - but I don't have a strong opinion either way, so am approving as is.

app/controllers/admin/speeches_controller.rb
davidgisbey: yea i think this is better 👍 Have updated it

PR #7577, opened 2023-04-24
Tidy code used to generate /world/embassies

chrisroos:
Soon we'd like to move the rendering of /world/embassies to the collection app. The code in Whitehall for rendering this page was quite complicated and tangled up the business logic with the presentation. This PR attempts to simplify the code, improve the test coverage and make the separation between business logic and presentation clearer so that it is easier move the rendering part elsewhere. 

We spent quite a lot of time trying to simplify or remove the "special cases" that were previously in the EmbassyPresenter (and are now in Embassy). We know that when we introduce a "publishing" step we'll be adding some workflow changes here - namely that any changes to these special cases (that previously would have been immediately reflected in the rendered page) will in the future necessitate publishing a document. We'd hoped to avoid those by either removing the special cases, modelling them explicitly in data, or relying on existing behaviour to cover them. Unfortunately we haven't been able to do that yet, all of the special cases are required to maintain the existing behaviour. We have started talking to colleagues in FCDO to see if we can change content to remove the need for some of them. We've decided for now to not puruse this any further and work out what to do as and when we introduce the publishing step. Hopefully the changes we have made so far will make it easier to revisit these in the future. 

In this PR we've improved the embassies/index template so that it more clearly communicates the 3 different scenarios it caters for: 

- Pointing British citizens to where they can receive help in a specific world location
- Pointing British citizens to where they can receive help if it is available in a different world location
- Falling back to a generic "contact local authorities" message if neither case holds. 

This should make it more straight-forward to replicate these three cases in a rendering app. 

The majority of these commits are refactoring changes (i.e. they don't affect the behaviour) but there are also 3 commits were we change behaviour to fix bugs that we discovered while building an understanding of the current behavior and refactoring. It will hopefully be easier to review the commits in sequence rather than looking at the combined diff as we use earlier refactorings to make subsequent ones easier. 


app/presenters/embassy_presenter.rb
```diff
@@ -4,27 +4,17 @@ class EmbassyPresenter < SimpleDelegator
...
-  end
-
-  def remote_office_name
-    if special_case?
-      SPECIAL_CASES[name][:building]
-    elsif has_remote_service?
-      organisation.name
-    end
-  end
+  RemoteOffice = Struct.new(:name, :location, :path)
```
floehopper: Minor: It might be nice to use the `keyword_init` option so that constructor call-sites have to use keyword arguments which might make things clearer, i.e.

```ruby
RemoteOffice = Struct.new(:name, :location, :path, keyword_init: true)

ro = RemoteOffice.new(name: 'name', location: 'location', path: 'path)
```

test/unit/models/embassy_test.rb
```diff
@@ -56,4 +56,32 @@ class EmbassyTest < ActiveSupport::TestCase
     assert_equal "Legoland", location.remote_services_country.name
     assert_equal "British Embassy Legoland", location.offices.first.title
   end
+
+  test '#can_assist_british_nationals? returns true if the world location is a special case' do
+    location_name = 'Central African Republic'
```
floehopper: Minor: Perhaps using `Embassy::SPECIAL_CASES.keys.first` would make things clearer here and avoid the need to make an assertion against the test setup, i.e. `assert Embassy::SPECIAL_CASES.has_key?(location_name)` (below).

test/unit/models/embassy_test.rb
```diff
@@ -42,7 +42,6 @@ class EmbassyTest < ActiveSupport::TestCase
     location = Embassy.new(toytown)
 
     assert_equal "Legoland", location.remote_services_country.name
-    assert_equal "British Embassy Legoland", location.offices.first.title
```
floehopper, line 74: Minor: Can you explain in the commit note *why* it's OK to remove this assertion?

test/unit/models/embassy_test.rb
```diff
@@ -1,49 +1,12 @@
...
-  end
-
   test "it delegates to the world location" do
-    embassy = Embassy.new(@location)
+    location = create(:world_location, :with_worldwide_organisations, name: "Narnia")
+    embassy = Embassy.new(location)
     assert_equal "Narnia", embassy.name
   end
 
-  test "remote_services_office and remote_services_country" do
```
floehopper, line 51: Minor: Can you explain in the commit note *why* it's OK to remove these tests?

test/unit/models/embassy_test.rb
```diff
@@ -71,4 +34,59 @@ class EmbassyTest < ActiveSupport::TestCase
 
     refute embassy.can_assist_british_nationals?
   end
+
+  test '#remote_office returns special case data where world location is a special case' do
+    location_name = 'Central African Republic'
```
floehopper: As before:

> Minor: Perhaps using `Embassy::SPECIAL_CASES.keys.first` would make things clearer here and avoid the need to make an assertion against the test setup, i.e. `assert Embassy::SPECIAL_CASES.has_key?(location_name)` (below).

test/unit/models/embassy_test.rb
```diff
@@ -1,49 +1,12 @@
 require "test_helper"
 
 class EmbassyTest < ActiveSupport::TestCase
-  setup do
```
floehopper, line 4: Minor: Can you explain *why* you've moved the `setup` into the relevant tests?

app/models/embassy.rb
```diff
@@ -5,7 +5,7 @@ def initialize(world_location)
 
   delegate :name, to: :@world_location
 
-  def consular_services_organisations
+  def organisations_with_embassy_offices
```
floehopper, line 8: Minor: Can you explain *why* you've renamed the method.

test/unit/models/embassy_test.rb
```diff
@@ -1,92 +1,155 @@
...
-    embassy = Embassy.new(location)
-
-    assert_nil embassy.remote_office
+  context "when there is an organisation with two embassy offices, one in an unknown location and one in another world location" do
+    before do
+      organisation = create(:worldwide_organisation,
+                            world_locations: [world_location],
+                            name: "org-name",
+                            slug: "org-slug")
+      @other_location = create(:world_location)
```
floehopper, line 133: Minor: It might be clearer to use a `let` here given that you've used `let` for other instance vars in the test.

test/unit/models/embassy_test.rb
```diff
@@ -152,4 +152,34 @@ class EmbassyTest < ActiveSupport::TestCase
...
+
+  context "when there is an organisation with two offices in two remote countries and one isn't an embassy" do
+    before do
+      organisation = create(:worldwide_organisation,
+                            world_locations: [world_location],
+                            name: "org-name",
+                            slug: "org-slug")
+      @second_location = create(:world_location)
+      @third_location = create(:world_location)
+      second_location_contact = create(:contact, street_address: "street-address", country: @second_location)
```
floehopper, line 164: Minor: Do you need to specify the `street_address` here? It doesn't seem to be used in the test and it looks as if the factory will set one by default.

app/views/embassies/index.html.erb
```diff
@@ -23,21 +23,21 @@
...
                 <ul class="govuk-list govuk-list--spaced govuk-!-margin-top-0">
-                  <% if !embassy.has_consular_service_in_location? || embassy.has_remote_service? %>
-                  <li>
-                    <p class="govuk-body">
-                      <%= embassy.text %>
-                    </p>
-                    <%= embassy.embassy_path %>
-                  </li>
+                  <% if embassy.can_assist_in_location? %>
+                    <%= render partial:"organisation", collection: embassy.organisations_with_embassy_offices %>
```
floehopper, line 29: Very minor: I know you didn't add this, but it would be nice to take the opportunity to add a space between `partial:` and "organisation" here!

floehopper (approved):
Overall this feels like a good set of improvements. 👍

While I think the logic is still a bit hard to understand in places, I think this is now more to do with the actual logic (and the reasons behind it) rather than the code being hard to follow (which was previously the case). In particular the new unit tests for the `Embassy` class make things a lot clearer.

One thing I found a bit hard to follow in the PR was the "Extract Embassy#offices_in_location?" commit:
  * I wasn't sure what the "before" and "after" text in the commit note was referring to - do some of those lines appear as headings in the actual page? If so, it would be good to make that clearer using some kind of formatting or clarifying text in the commit note.
  * I think it would help if you could clarify in the commit note whether the you're confident the behaviour change is a bug fix and/or whether you've checked this with FCDO.
  * I wonder if the logic changes in this commit could be split into two or more commits to make it easier to follow, e.g. start using guard conditions in a first commit and only then extract the new methods in another commit?
 
Is it worth adding unit tests for `Embassy#organisations_with_embassy_offices` for completeness?

I've also added some minor points inline, but I don't think any of those should prevent you merging the PR.

app/presenters/embassy_presenter.rb
chrislo: Resolved this by editing the history. Force pushing with the changes.

test/unit/models/embassy_test.rb
chrislo: Addressed in 93b235c1d0

test/unit/models/embassy_test.rb
chrislo: Addressed in 93b235c1d0

test/unit/models/embassy_test.rb
chrislo: Addressed in 7e532085c4 which I'll fixup before merging.

test/unit/models/embassy_test.rb
chrislo: There's a contact_with_country factory that we can use, I've made that change in cfba5a4ed7

app/views/embassies/index.html.erb
chrislo: Fixed in 7692345ab2

chrislo:
@floehopper I've made some changes in fixup commits to address your feedback on specific areas of the code, thank you! To address your suggestions to improve the commit messages I'm going to have to edit the commits and force-push again, I'll fixup the changes when I do that too.

app/models/embassy.rb
chrislo: I've now added additional explanation in the commit message and force-pushed.

test/unit/models/embassy_test.rb
chrislo: I've now added additional explanation in the commit message and force-pushed.

test/unit/models/embassy_test.rb
chrislo: I've now added additional explanation in the commit message and force-pushed.

test/unit/models/embassy_test.rb
chrislo: I've now added additional explanation in the commit message and force-pushed.

chrislo:
Thanks for the review @floehopper! 

I've added some additional detail to the commit note for the "Extract Embassy#offices_in_location?" commit, to make the "before" and "after" examples hopefully clearer and to explain why we're confident to make this small change in behaviour. 

I've added tests for Embassy#organisations_with_embassy_offices in 76269ecb61 - thanks for catching that. 

@chrisroos and I also struggled with the naming of Embassy, since it doesn't quite capture what this particular page cares about - the assistance provided by an office in a particular location (whether it's an embassy or not). I think for now I'm not going to make any further name changes or clarifications because we've already made quite a lot. I think the test might be when we come to migrate this page to a rendering app and write code to populate a content item, if we keep getting confused by or tripped up on the name we can have another pass at clarifying things.

ollietreend (approved):
Nice work tidying this up! This refactor looks good to me - and nice work spotting & fixing the bugs too. 🚀

PR #7583, opened 2023-04-26
Audit worldwide organisations

jkempster34:
> Part 2 of a 2 part deployment. [Part 1 here](https://github.com/alphagov/whitehall/pull/7590)

### Move `AuditTrail` out of the `Edition` namespace
We want to be able to create audit trails for non-editionable models.

There is an existing audit trail implementation that is only used for
editionable models. However, as it is quite generic we want to repurpose it so
that it can be used for both non-editonable and editionable models.

This moves the `AuditTrail` concern out of the `Edition` namespace.

The `alert!` method of the `AuditTrail` is possibly non-generic, this will be
refactored or removed  at a later point. 

### Include the `AuditTrail` concern in `WorldwideOrganisation`
This adds the `AuditTrail` concern to the `WorldwideOrganisation` model. From
this point, any create or update events will be recorded as `Versions`. 

`AuditTrail` currently requires a `state`. Instead, we fall back to nil if
`state` is not set.

### Show an audit trail for worldwide organisations
Now that we create an audit trail for worldwide organisations, we want to
display the history on the admin page.

ViewComponents for editionable models already exist. But, because these contain
edition specific logic, rather than modifying them, we want to create separate
ViewComponents for non-editionable models .

This adds a paginated audit trail alongside the details of the worldwide
organisations. Any initial audits are shown as "No history before this time" by
"User (unknown)"

### Remove any mentions of Papertrail

We used to use the PaperTrail gem to keep audits until it was removed in favour
of a custom solution (https://github.com/alphagov/whitehall/pull/195).

Remove any remaining mentions of PaperTrail as they are misleading. The
`@previous_papertrail_whodunnit` variable is unused and safe to remove
entirely.

image


db/data_migration/20230424100641_add_initial_worldwide_organisation_audits.rb
```diff
@@ -0,0 +1,19 @@
+EVENT_TYPE = "initial".freeze
```
jkempster34, line 1: Couple of thoughts about this

1. Is this the most sensible thing to do? We could use the `updated_at` to set the `created_at` of the `Version` instead.
2. Am I correct that there are only 435 worldwide organisations, so we don't need to worry about batching the requests.
3. How do we ensure that we run this migration before any users make further changes to worldwide organsiations (that will now be audited). Could we move this to the normal migrations folder so it runs automatically, or is there another approach that we take on this app?

floehopper (commented):
This all looks good to me 👍 

However, stepping back from this PR and given what I've discovered in [this Trello comment](https://trello.com/c/hwo2GEzk/578-create-the-new-db-tables-for-worldwide-organisations-in-whitehall#comment-644169ff9b28dd820a544047), I'm wondering that if the whole approach of moving the about-us `CorporateInformationPage` fields into the `WorldwideOrganisation` model is being called into question, then perhaps the changes in this PR may no longer be required. /cc @JonathanHallam

Otherwise I've made one minor suggestion inline.

Given my comment above and that I was involved in the initial work on this, it might be sensible for someone else to review the PR, so I'm not going to approve it for now even though it's fine!

jkempster34:
> This all looks good to me 👍
> 
> However, stepping back from this PR and given what I've discovered in [this Trello comment](https://trello.com/c/hwo2GEzk/578-create-the-new-db-tables-for-worldwide-organisations-in-whitehall#comment-644169ff9b28dd820a544047), I'm wondering that if the whole approach of moving the about-us `CorporateInformationPage` fields into the `WorldwideOrganisation` model is being called into question, then perhaps the changes in this PR may no longer be required. /cc @JonathanHallam
> 
> Otherwise I've made one minor suggestion inline.
> 
> Given my comment above and that I was involved in the initial work on this, it might be sensible for someone else to review the PR, so I'm not going to approve it for now even though it's fine!

Ah yes, I see. Perhaps this can be converted back to a draft while we gather more information.

floehopper:
@jkempster34:

> However, stepping back from this PR and given what I've discovered in [this Trello comment](https://trello.com/c/hwo2GEzk/578-create-the-new-db-tables-for-worldwide-organisations-in-whitehall#comment-644169ff9b28dd820a544047), I'm wondering that if the whole approach of moving the about-us `CorporateInformationPage` fields into the `WorldwideOrganisation` model is being called into question, then perhaps the changes in this PR may no longer be required.

I chatted with @JonathanHallam earlier and he's confident that this is the right direction, so my concern above is no longer a blocker for getting this merged.

> Given my comment above and that I was involved in the initial work on this, it might be sensible for someone else to review the PR...

I think this ☝️ is still true though - worth getting another set of eyes on it - and getting answers to all your questions.

jkempster34:
@ollietreend Hi Ollie, would you mind taking another look at this work please?

test/components/admin/audit_trail_component_test.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Admin::AuditTrailComponentTest < ViewComponent::TestCase
+  setup do
+    @non_editionable_item = create(:worldwide_organisation)
+  end
+
+  test "renders audit trail entries" do
+    create(:version, event: "create", item: @non_editionable_item)
+    create(:version, event: "update", item: @non_editionable_item)
+    versions = Kaminari.paginate_array(@non_editionable_item.versions_desc).page(1)
```
ollietreend, line 11: As with the implementation in the controller, you should be able to avoid using `paginate_array` here.

While it's not the end of the world to use it here (it's never going to cause any performance issues..) it sets a bad example for how to implement this when using Kaminari.

```suggestion
    versions = @non_editionable_item.versions_desc.page(1)
```

db/data_migration/20230424100641_add_initial_worldwide_organisation_audits.rb
ollietreend: > How do we ensure that we run this migration before any users make further changes to worldwide organsiations (that will now be audited). Could we move this to the normal migrations folder so it runs automatically, or is there another approach that we take on this app?

Good question. Unfortunately even if this was moved to the normal migrations folder, there's still the risk that users will make changes before it's run – the hosting platform [makes no guarantee](https://govuk-k8s-user-docs.publishing.service.gov.uk/manage-app/access-ci-cd/#deploying-a-release-of-a-gov-uk-app) that database migrations will complete before new code hits the database.

I think in practice, this is very unlikely to be a problem. However it's probably best not to tempt fate.

I think the only foolproof way to achieve this is to split it into a two-stage deployment:

1. Deploy & run the "create initial audits" data migration first
2. Then follow up with the other changes, including where `include AuditTrail` is introduced to the `WorldwideOrganisation` model

ollietreend (commented):
This looks good to me 👍🏻 just one minor comment, and a follow-up to your migration question

jkempster34:
@ollietreend Great thanks, 

If performing a two stage deployment, we would be missing an audit if a user makes a change before part 2 is deployed. I'm not sure if that's a huge problem TBH, and perhaps better than misordered audits if a user makes a change before the migration has run in a one-stage deployment.

In terms of design, how do you feel about "No history before this time" ..."user (unknown)". James suggested another approach [here](https://github.com/alphagov/whitehall/pull/7583#discussion_r1177925906) that's worth considering.

Once we have the above figured out, I'm happy to open another 2 PRs for each stage (if thats the agreed direction)!

db/data_migration/20230424100641_add_initial_worldwide_organisation_audits.rb
ollietreend: @jkempster34

> In terms of design, how do you feel about "No history before this time" ..."user (unknown)". James suggested another approach https://github.com/alphagov/whitehall/pull/7583#discussion_r1177925906 that's worth considering.

I think the current approach you've taken seems reasonable – i.e. a "No history before this time" event with today's date (or the date this gets deployed/migrated). I think that's the most truthful representation of the situation. It'll be the date we started tracking edit history.

Whereas it feels a bit misleading to backdate 'created' and 'updated' events based on the timestamps in the records – since it'd give an incomplete picture of the edit history.

ollietreend (approved):
Looks good to me 👍🏻 and bonus points for removing old PaperTrail references. Thanks! ✨🚀

PR #7589, opened 2023-04-27
Update Govspeak/Markdown syntax for images

dnkrj:

# What
[Support `[Image:example.jpg]` syntax for images in GovSpeak.](https://trello.com/c/9oMxpTMN/991-update-govspeak-markdown-syntax-for-images)
Continue to support the `!!1` syntax.
This PR is based on https://github.com/alphagov/whitehall/pull/7196.

# Why
The new images UI provides users with the syntax as it means that image references do not shift if one is deleted.

### DIFF
--- a/app/views/admin/editions/_legacy_govspeak_help.html.erb
+++ b/app/views/admin/editions/_legacy_govspeak_help.html.erb
@@ -108,9 +108,9 @@
   <h3><a data-toggle="collapse" href="#govspeak-images" aria-expanded="false">Images</a></h3>
   <div class="collapse" id="govspeak-images">
     <p>First upload your image(s), then:</p>
-    <pre>!!<em>n</em></pre>
+    <pre>[Image:&nbsp;filename]</pre>
     <p>eg for the first image:</p>
-    <pre>!!1</pre>
+    <pre>[Image:&nbsp;example.jpg]</pre>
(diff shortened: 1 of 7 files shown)


app/views/admin/editions/_legacy_govspeak_help.html.erb
ollietreend, line 1: It looks like this PR is missing an update to the non-legacy partial.

We need to update both partials:

- [_legacy_govspeak_help.html.erb](https://github.com/alphagov/whitehall/blob/73e1f38e0c64e505a6f1fbc5f54bb940447a5f05/app/views/admin/editions/_legacy_govspeak_help.html.erb) (this file)
- [_govspeak_help.html.erb](https://github.com/alphagov/whitehall/blob/e7d7f0ba23156bd09abec229a432289d5ecebcfa/app/views/admin/editions/_govspeak_help.html.erb) (for Design System pages)

It might be okay to update _only_ the new Govspeak Help partial. However I'm not 100% sure if the legacy partia still going to be used somewhere obscure, or if Release 1.6 stops its use entirely.

dnkrj: Good point, I've added both, I think that originally I just copied the wrong one over.

PR #7590, opened 2023-04-27
Create initial audits for worldwide organisations

jkempster34:
> Part 1 of a 2 part deployment. [Part 2 here](https://github.com/alphagov/whitehall/pull/7583)

### [Create initial audits for worldwide organisations](https://github.com/alphagov/whitehall/pull/7590/commits/f5ee5f3b1452c8abd27a6f07b112b0d02df1abd2)

We will soon be auditing worldwide organisations; this will create a `Version`
record when a worldwide organisation is created or updated. However, until
auditing is added, we have no way of knowing what changes were made or by
whom.

This migration creates a `Version` with an "initial" event type for all worldwide
organisations. We can then handle this in the UI in an appropriate way by
displaying "No history before this time".

We want to deploy the worldwide organisation auditing work in two steps -
first, by running this migration, then, by merging separate work that adds audits.
There may be audits that are not recorded in the period between the two steps.
We have decided that this is only a minor issue, and we will attempt to deploy
the follow up work shortly after this migration is run in order to reduce the
likelihood of this happening.

image

### DIFF
--- a/db/data_migration/20230424100641_add_initial_worldwide_organisation_audits.rb
+++ b/db/data_migration/20230424100641_add_initial_worldwide_organisation_audits.rb
@@ -1,0 +1,19 @@
+EVENT_TYPE = "initial".freeze
+
+def log_message
+  %(
+    Number of WorldwideOrganisations: #{WorldwideOrganisation.count},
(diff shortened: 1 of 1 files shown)


ollietreend (approved):
Looks good 👍🏻

PR #7591, opened 2023-04-27
Add summary and body to worldwide organisation translations

floehopper:
Trello: https://trello.com/c/hwo2GEzk

This is an initial draft PR to illustrate what we might do if we want to move the content currently stored on the "about" `CorporateInformationPage` for a `WorldwideOrganisation` onto the `WorldwideOrganisation` itself.

This PR doesn't yet include code changes which make use of these new fields since the Trello card doesn't seem to want me to do that. However, I feel extremely nervous about making these database changes without first seeing how we'll make use of them and, in particular, how we'll change the editor UI.

This PR currently just includes a regular database migration (to add the new columns) and a data migration (to copy data to those new columns); eventually we'd probably want to delete all the "about" `CorporateInformationPages`. There are a few potential gotchas in the data migration:

1. Although the Trello card talks about moving the fields from the "about" `CorporateInformationPages` to the `WorldwideOrganisation`, in fact I think we need to move them from the translations associated with the former to the translations for the same locale for the latter.
2. Some `WorldwideOrganisations` have a draft "about" `CorporateInformationPage`. Only data in the latest published edition of the "about" `CorporateInformationPage` will be copied. If we want to retain any other data, we'll need to make separate provision for that.
3. Some `WorldwideOrganisations` do not currently have a translation for all the locales translated on the "about" `CorporateInformationPage`. The data migration will create them, but this means the translations for the `name` & `services` fields will be blank for that locale.

I haven't yet dealt with attachments, but the plan would be to include the `Attachable` concern into `WorldwideOrganisations` and update the `attachable_id` & `attachable_type` on the `Attachments` associated with the latest published edition of the "about" `CorporateInformationPage`. However, again I'm pretty nervous about making this change in the absence of making the changes to the editor UI.

We _might_ also want to capture some of the edition-related document history from the "about" `CorporateInformationPage` in the new non-editionable document history being created in [this Trello card](https://trello.com/c/7snah8vf).

If this still feels like the right direction, then I think I'd prefer to convert the Trello card into something more complete (even if it's only just a spike), i.e. to actually change the editor UI to use the new fields.

### DIFF
--- a/db/data_migration/20230427150648_copy_summary_and_body_from_worldwide_organisation_about_page_to_worldwide_organisation.rb
+++ b/db/data_migration/20230427150648_copy_summary_and_body_from_worldwide_organisation_about_page_to_worldwide_organisation.rb
@@ -1,0 +1,38 @@
+WorldwideOrganisation.all.each do |worldwide_organisation|
+  puts "Processing #{worldwide_organisation.slug}"
+
+  if worldwide_organisation.about_us.nil?
+    puts "No published about page".indent(2)
+    next
+  end
+
+  if worldwide_organisation.about_us.translations.none?
+    puts "No translations for published about page".indent(2)
+    next
+  end
+
+  worldwide_organisation.about_us.translations.each do |source_translation|
+    puts "Processing source translation for #{source_translation.locale}".indent(2)
+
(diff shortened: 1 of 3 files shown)


floehopper:
I'm going to close this now, because I think it's served its purpose. I plan to open a new more fleshed out PR soon.

PR #7596, opened 2023-05-02
Publish embassies index

chrisroos:
Trello link: https://trello.com/c/MMUcRbIT

This PR enables Whitehall to publish an `embassy_index` document to the Publishing API and keep that document up-to-date with changes made to its underlying data. We need this document to be published so that we can switch over the rendering of the `/world/embassies` page from Whitehall to Collections. 

---

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -20,6 +20,8 @@ class Contact < ApplicationRecord
   after_create :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
 
+  after_commit :republish_embassies_index_page_to_publishing_api
+
   include TranslatableModel
   translates :title,
              :comments,
@@ -74,5 +76,9 @@ class Contact < ApplicationRecord
 
   def missing_translations
     super & contactable.non_english_translated_locales
+  end
+
+  def republish_embassies_index_page_to_publishing_api
+    PresentPageToPublishingApi.new.publish(PublishingApi::EmbassiesIndexPresenter)
   end
 end
--- a/app/models/embassy.rb
+++ b/app/models/embassy.rb
@@ -21,7 +21,7 @@ class Embassy
     elsif can_assist_in_other_location?
       remote_office = offices_in_remote_location.first
       RemoteOffice.new(name: remote_office.worldwide_organisation.name,
-                       location: remote_office.country,
+                       location: remote_office.country.name,
                        path: remote_office.worldwide_organisation.public_path)
     end
   end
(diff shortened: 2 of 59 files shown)


lib/whitehall/publishing_app.rb
```diff
@@ -0,0 +1,5 @@
+module Whitehall
+  module PublishingApp
+    WHITEHALL = "whitehall".freeze
```
BeckaL, line 3: Lovely little improvement, thank you ⭐

app/models/embassy.rb
```diff
@@ -21,7 +21,7 @@ def remote_office
     elsif can_assist_in_other_location?
       remote_office = offices_in_remote_location.first
       RemoteOffice.new(name: remote_office.worldwide_organisation.name,
-                       location: remote_office.country,
+                       location: remote_office.country.name,
```
BeckaL, line 24: Just a check since you mention in the commit message about the template doing this transformation - does this affect the existing rendering (i.e. are we going to be calling `.name` in a template now on a string)?

app/models/contact.rb
```diff
@@ -20,6 +20,8 @@ class Contact < ApplicationRecord
   after_create :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
 
+  after_commit :republish_embassies_index_page_to_publishing_api
```
BeckaL, line 23: Just leaving a comment here unrelated to the line but in response to the decisions you've outlined in this commit message that I agree with all the decisions you've made there, thanks for outlining them! 😄

BeckaL (commented):
Looks great! Just one question here to double check that we're not affecting any existing rendering by changing something on the model which I imagine you've already thought about so just wanted to check, but happy to approve after that.

app/models/embassy.rb
chrislo: Previously we were relying on the `to_s` method [defined by WorldLocation](https://github.com/alphagov/whitehall/blob/publish-embassies-index/app/models/world_location.rb#L61). So we've decided to call `name` explicitly here to populate the json. This doesn't affect the existing rendering.

app/models/contact.rb
chrislo: Thank you!

chrislo:
@BeckaL - we've double checked the rendering question you asked, and we don't think there'll be an issue - let us know if you have any other thoughts.

BeckaL (approved):
Great, LGTM :shipit:

chrislo:
Thanks @BeckaL! Rebasing and force-pushing before merging.

PR #7605, opened 2023-05-03
Spike use prefix route for organisation content items

floehopper:
Trello: https://trello.com/c/sajxUZnZ

We're hoping that changing the main route for an organisation content item from "exact" to "prefix" will allow us to handle the various redirects for URL paths under the organisation page using Rails routing in the Collections app without breaking other content items that are published with exact routes under the organisation page.

I've sketched out the corresponding changes to the Collections app in https://github.com/alphagov/collections/pull/3244.

### DIFF
--- a/test/unit/presenters/publishing_api/organisation_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/organisation_presenter_test.rb
@@ -50,7 +50,7 @@ class PublishingApi::OrganisationPresenterTest < ActionView::TestCase
       publishing_app: "whitehall",
       rendering_app: "collections",
       routes: [
-        { path: public_path, type: "exact" },
+        { path: public_path, type: "prefix" },
         { path: public_atom_path, type: "exact" },
       ],
       redirects: [],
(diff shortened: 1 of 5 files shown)


test/unit/presenters/publishing_api/organisation_presenter_test.rb
```diff
@@ -294,7 +294,7 @@ def organisation.summary
     presented_item = present(organisation)
 
     assert_equal("collections", presented_item.content[:rendering_app])
-    assert_equal([{ path: "/courts-tribunals/court-at-mid-wicket", type: "exact" }], presented_item.content[:routes])
+    assert_equal([{ path: "/courts-tribunals/court-at-mid-wicket", type: "prefix" }], presented_item.content[:routes])
```
BeckaL, line 297: Not an issue for testing this out, but I wonder for later whether we want to leave the courts and tribunals orgs alone since they're not 'ordinary' org pages, and I don't think e.g. have atom feeds, but not going to cause issues for now

floehopper: Good spot, although I don't think it's particularly relevant that they don't have an atom feed. I suppose the question is whether the redirected paths like `/government/organisations/:organisation_id/chiefs-of-staff` *ever* applied to those organisations. I can see that they now live under `/courts-tribunals`, but I wonder if that was always the case...?

floehopper:
I'm closing this draft PR - I'm planning to open a new PR to actually implement the change.

test/unit/presenters/publishing_api/organisation_presenter_test.rb
floehopper: Addressed in https://github.com/alphagov/whitehall/pull/7621/commits/d4f7bc6299d0fc9eca610b98089e76d01628826a#diff-e861a54893d500d742a133f8c52baa2e9c2a9954262368606f8e5d44ca5924aaR66.

PR #7618, opened 2023-05-09
[DO NOT MERGE] Make data-ga4-attachment-link appear on all attachment links

AshGDS:
Do not merge as we are deploying/changing the attachment tracking in a stepped process, so that we don't affect the existing `file_download` tracking .

Previously we were tracking HTML attachments as 'attachment' and all other attachment links as file_download using JS. This change makes it so that every attachment link will get the type 'attachment' set on it in our ga4-link-tracker.js in publishing components.

https://trello.com/c/MHFXw6H2/552-change-filedownloads-on-attachment-pages-to-have-attachment-instead-of-generic-download-type

### DIFF
--- a/app/views/documents/_attachment.html.erb
+++ b/app/views/documents/_attachment.html.erb
@@ -14,12 +14,9 @@
   # to publishing-api.
   preview = controller.nil? ? nil : params[:preview]
 
-  attachment_data = {}
-  if attachment.html?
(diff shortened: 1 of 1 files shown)


ollietreend:
I'm closing this pull request as Whitehall no longer renders attachments itself following #7932.

PR #7621, opened 2023-05-09
Handle legacy organisation-related redirects

floehopper:
Trello: https://trello.com/c/sajxUZnZ

Currently a number of legacy organisation-related redirects are implemented in whitehall as Rails routes:

https://github.com/alphagov/whitehall/blob/f871f66c9e4f21c23132fa9cf9505e7a449dcacc/config/routes.rb#L66-L73

Changing the main route for an organisation content item from "exact" to "prefix" will allow us to handle the paths for those legacy organisation-related redirects using Rails routing in the collections app without breaking other content items that are published with exact routes under the organisation page, e.g. corporate information pages. The latter will not be affected, because they have more specific routes which take precedence in the router.

This will allow us to remove the legacy redirects from whitehall's routes moving us another step closer to being able to switch off whitehall-frontend.

Although courts & tribunals are organisations and a republished using the `PublishingApi::OrganisationPresenter`, they're published under a different path, e.g. `/courts-tribunals/court-of-protection`. This means that the legacy redirects don't make sense for them and so I've added logic in `PublishingApi::OrganisationPresenter#use_prefix_route?` to retain the "exact" route type in this case.

Note that we did consider adding a set of more-specific routes to the organisation content item (instead of a single prefix route), but unfortunately some awkwardness with locales meant that it would've made the whitehall code a lot more complicated and it didn't seem worth it just to implement these old redirects.

Important: this PR should not be deployed until [the corresponding PR in the collections app](https://github.com/alphagov/collections/pull/3247) is deployed. And once this PR is deployed we will need to republish all of the organisations before finally we'll be able to actually remove the redirects from the whitehall Rails routes.

### DIFF
--- a/app/presenters/publishing_api/organisation_presenter.rb
+++ b/app/presenters/publishing_api/organisation_presenter.rb
@@ -32,7 +32,7 @@ module PublishingApi
         schema_name:,
       )
       content.merge!(
-        PayloadBuilder::PolymorphicPath.for(item, additional_routes:),
+        PayloadBuilder::PolymorphicPath.for(item, prefix: use_prefix_route?, additional_routes:),
       )
       content.merge!(PayloadBuilder::AnalyticsIdentifier.for(item))
     end
@@ -60,6 +60,10 @@ module PublishingApi
 
     def schema_name
       "organisation"
+    end
+
+    def use_prefix_route?
+      !court_or_tribunal?
     end
 
     def additional_routes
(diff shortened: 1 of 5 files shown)


BeckaL (approved):
All looks sensible to me, thanks for excellent explanations in commit messages and adding in that extra test

BeckaL:
A minor thing I just thought - you can use the 'do not merge' label on this PR as an extra marker here that there is a dependent PR to be merged in first. Doesn't do anything to block the merge, but can be a useful signifier for other people in case

floehopper:
https://github.com/alphagov/collections/pull/3247 has been merged and deployed, so I've removed the do-not-merge labels.

PR #7623, opened 2023-05-09
Update copy around lead images

dnkrj:

# What
Implement new copy for lead images and add logic for case studies.
Move the message into the component class.

# Why
The original copy was written for functionality that has not been implemented.
Having the message in the component class fixes a bug where the paragraph tags were being rendered twice in the template.

# Screenshots

## Before

whitehall-admin dev gov uk_government_admin_editions_1374131_images(iPad Pro) (1)

### Case Studies
whitehall-admin dev gov uk_government_admin_editions_1374131_images(iPad Pro)

### Other Editions
whitehall-admin dev gov uk_government_admin_editions_1374126_images(iPad Pro) (1)

### DIFF
--- a/app/components/admin/edition_images/uploaded_images_component.rb
+++ b/app/components/admin/edition_images/uploaded_images_component.rb
@@ -39,6 +39,14 @@ class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
     "#{new_image_display_option == 'no_image' ? 'Remove lead' : 'Use default'} image"
   end
 
+  def lead_image_guidance
+    if @edition.type == "CaseStudy"
+      tag.p("Using a lead image is optional and can be shown or hidden. The first image you upload is used as the lead image.", class: "govuk-body") + tag.p("The lead image appears at the top of the document. The same image cannot be used in the body text.", class: "govuk-body")
+    else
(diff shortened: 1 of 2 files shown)


app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -39,6 +39,14 @@ def update_image_display_option_button_text
     "#{new_image_display_option == 'no_image' ? 'Remove lead' : 'Use default'} image"
   end
 
+  def lead_image_guidance
+    if @edition.type == "CaseStudy"
+      tag.p("Using a lead image is optional and can be shown or hidden. The first image you upload as the lead image.", class: "govuk-body") + tag.p("The lead image appears at the top of the document. It cannot be used in the document body.", class: "govuk-body")
```
ollietreend, line 44: Should this be "is" instead of "as"?

```suggestion
      tag.p("Using a lead image is optional and can be shown or hidden. The first image you upload is the lead image.", class: "govuk-body") + tag.p("The lead image appears at the top of the document. It cannot be used in the document body.", class: "govuk-body")
```

dnkrj: Good catch!

ollietreend (approved):

PR #7625, opened 2023-05-10
[DO NOT MERGE] Ensure GOV.UK Design System has full test coverage prior to 1.6 release

davidgisbey:
## Description

This ensures that the Boostrap and GOV.UK Design System implementation are fully tested. I've done this by doing the following: 

1. forced the `preview_design_system?` to always return true to force all tests to run in the design system
2. noted down out all tests that aren't in a legacy test file that failed
3. Either duplicated the test file and prepended one with legacy. Then gave the non-legacy files user the `Preview design system` permission or duplicated the test and gave the new tests user the permission
4. removed the forced permission

### DIFF
--- a/test/functional/admin/detailed_guides_controller_test/national_applicability_test.rb
+++ b/test/functional/admin/detailed_guides_controller_test/national_applicability_test.rb
@@ -5,8 +5,7 @@ class Admin::DetailedGuidesControllerTest < ActionController::TestCase
     tests Admin::DetailedGuidesController
 
     setup do
-      login_as create(:writer, organisation: create(:organisation))
-      current_user.permissions << "Preview design system"
+      login_as_preview_design_system_user(:writer)
 
       stub_request(
         :get,
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -358,10 +358,9 @@ private
   def build_national_exclusion_params
     design_system_controllers = %w[consultations detailed_guides publications]
     return unless design_system_controllers.include?(controller_name) && preview_design_system?(next_release: true)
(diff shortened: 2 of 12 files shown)


test/functional/admin/detailed_guides_controller_test/national_applicability_test.rb
```diff
@@ -6,6 +6,8 @@ class NationalApplicabilityTest < ActionController::TestCase
 
     setup do
       login_as create(:writer, organisation: create(:organisation))
+      current_user.permissions << "Preview design system"
```
ollietreend, line 9: Would this work instead?

```suggestion
      login_as_preview_design_system_user :writer
```

ollietreend, line 9: Would this work instead?

```suggestion
      login_as_preview_design_system_user :writer
```

ollietreend (approved):
This looks great. Really nice work spotting all the uncovered changes. And bonus points for getting through the gnarly National Inapplicabilities fixes! 💪🏻 🚀

test/functional/admin/detailed_guides_controller_test/national_applicability_test.rb
ollietreend: Ah, you might be right. I'd assumed `create(:writer)` would give the user an org automatically. But looking at [the factory](https://github.com/alphagov/whitehall/blob/ab98fb14d01fddd38abf79a430fb68dbcf365de8/test/factories/users.rb), I think I might be wrong on that!

No stress if it causes tests to fail – just leave it as is!

davidgisbey: it went 🟢  so all good. Thanks for that

PR #7634, opened 2023-05-11
Simplify "I can see the attachment title" cucumber step

floehopper:
The "I can see the attachment title" attachment step is now only used in the context of the standard admin attachments index page rendered using the design system via `app/views/admin/attachments/index.html.erb` & `app/views/admin/attachments/_attachments.html.erb`; not in the context of the consultation response attachments index page rendered using legacy templates via `app/views/admin/responses/show_legacy.html.erb` & `app/views/admin/responses/_attachments_legacy.html.erb`.

This means there's no longer any point in having this step support both the design system and legacy templates. This in turn means the "Previewing HTML attachment" scenario no longer needs to be tagged as `design-system-only`.

Note that the final stage of moving the standard admin attachments index page to the design system appears to have been implemented in 7d4dd7a77ef9b90ea0688cc3d633dc27623d7d8b as part of #7219.


BeckaL (approved):
👍

features/edition-attachments.feature
```diff
@@ -24,7 +24,6 @@ Feature: Managing attachments on editions
       | Beard Length Statistics 2014 |
       | Beard Length Illustrations   |
 
-  @design-system-only
```
davidgisbey, line 27: Might be worth double checking with @ollietreend but i think this should probs be left in.

We run the suite twice, once with a user that has the "Preview design system" permission and one without. This permission controls whether a user sees the Bootstrap or GOV.UK Design System implementation while the page is being transitioned. 

During this period we need to test both implementations. However, once the page has been fully ported , both users will always see the new implementation so it effectively becomes a duplicate test if we run it twice.

By adding `@design-system-only` we only run the test in the run that tests views are using the GOV.UK Design System.

floehopper: @davidgisbey Thanks for looking at this. I think I understand all of what you've said, but I'm still confident this can be removed given that this scenario (which concerns a publication) only touches pages which have *already* been switched over to *only* use the design system, i.e. the design system version of the relevant page(s) is rendered whether or not the design-system toggle is set. Or am I missing something? /cc @ollietreend

ollietreend: Well spotted @floehopper - and thanks for suggesting this simplification.

You're right - edition attachments now run entirely under the Design System layout, so there's no reason why we need to keep the `preview_design_system?` conditional in this case.

Would you mind going ahead and also removing the `@design-system-only` tag from the other scenario in this feature file? (`Scenario: Adding and reordering attachments`) I believe that one should also be a no-op at this point. And then we're at least being consistent within this particular feature file.

More generally speaking, I acknowledge that there will likely be other areas within our codebase and test suites which are in a sort of 'half way' point with regards to the transition from Bootstrap to Design System. We have a team actively working on transitioning the remaining areas of Whitehall's admin UI, so I think it's acceptable for these bits to be lying around right now. As the Design System transition work concludes, we'll be making a conscious effort to rid the codebase of all references to `preview_design_system?`, `using_design_system?` and friends.

floehopper: @ollietreend 

> Would you mind going ahead and also removing the @design-system-only tag from the other scenario in this feature file?

Yes, no problem.

Re being at a halfway point, I completely understand - no worries. My motivation for wanting to change this particular bit is that I'm about to make `WorldwideOrganisation` support attachments (as part of [this card](https://trello.com/c/hwo2GEzk) and I want to reuse some of these Cucumber steps. Making this change first should make that change easier.

floehopper: Addressed in 551bb8d325ccfaa307b77b939eb0394f4f9cee6d

floehopper:
@ollietreend Can you double-check my 2nd commit when you have time? Thanks.

ollietreend (approved):
Excellent – thanks for this @floehopper! 👍🏻

PR #7638, opened 2023-05-11
133 single image upload component

davidtrussler:
[Trello Card](https://trello.com/c/Bm92hDFN/133-in-page-single-image-upload-reusable-pattern-dev)

This PR 
- creates a new component for reuse in areas where a single image needs to be uploaded within the page, i.e. there is no dedicated image tab. The component deals with 4 scenarios: 
  - the default: the user is creating a new page
  - the user saves the newly created page without uploading an image triggering an error
  - the user edits an existing page which already has an uploaded image associated with it. In this case the uploaded image is displayed and the user has the option to replace that with a different one. 
  - the user edits an existing page which already has an uploaded image associated with it and tries to save the page which has errors. In this case a text reference to the uploaded image is displayed rather than the image itself. 
- applies this component to the "Take Part" pages 

|Scenario|Component|
|-|-|
|Default| Screenshot 2023-05-11 at 11 27 43 |
|With errors| Screenshot 2023-05-11 at 11 28 02 |
|Edit page when an image is already uploaded| Screenshot 2023-05-11 at 11 28 16 |
|Edit page when there are errors on the page| Screenshot 2023-06-06 at 14 30 33 |

|Scenario|Component implemented on "Take Part"|
|-|-|
|Default| Screenshot_1 |
|With errors| Screenshot_2 |
|Edit page when an image is already uploaded| Screenshot_3 |
Edit page when there are errors on the page| Screenshot 2023-06-06 at 14 39 22 |

### DIFF
--- a/app/views/components/_single-image-upload.html.erb
+++ b/app/views/components/_single-image-upload.html.erb
@@ -1,0 +1,72 @@
+<%
+  error_items ||= []
+  page_errors ||= false
+  data_attributes ||= {}
+  image_src ||= nil
+  image_alt ||= nil
+  image_cache ||= nil
+
+  root_classes = %w(app-c-single-image-upload govuk-form-group)
+%>
+
+<%= tag.div class: root_classes, data: data_attributes do %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Image (required)",
+    font_size: "l",
+    margin_bottom: 3
+  } %>
+
+  <% if image_src %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Uploaded image",
+      heading_level: 3,
+      font_size: "m",
+      margin_bottom: 3
+    } %>
+
+    <% if page_errors %>
+      <%= hidden_field_tag "#{name}[image_cache]", image_cache %>
+      <p class="govuk-body"><strong>File name: </strong><%= filename %></p>
+      <p class="govuk-body"><strong>Alt text: </strong><%= image_alt == "" ? "None" : image_alt %></p>
+    <% else %>
+      <div class="govuk-grid-row app-c-single-image-upload__uploaded-image">
+        <div class="govuk-grid-column-one-quarter">
+          <img src="<%= image_src %>" alt="" class="app-view-edition-images__image">
+        </div>
+
+        <div class="govuk-grid-column-three-quarters">
+          <p class="govuk-body govuk-!-margin-0"><strong>Alt text: </strong><%= image_alt == "" ? "None" : image_alt %></p>
+        </div>
+      </div>
+    <% end %>
+
+    <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+    <% image_label = "Replace image" %>
+    <% label_size = "m" %>
+  <% else %>
+    <% image_label = "Upload image" %>
+    <% label_size = nil %>
+  <% end %>
+
+  <%= render "govuk_publishing_components/components/file_upload", {
+    label: {
+      text: image_label,
+      heading_size: label_size
+    },
+    name: "#{name}[image]",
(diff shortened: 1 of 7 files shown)


app/views/components/_single-image-upload.html.erb
```diff
@@ -0,0 +1,71 @@
...
+      value: image_src,
+      error_items: error_items
+    } %>
+  <% else %>
+    <%= render "govuk_publishing_components/components/file_upload", {
+      label: {
+        text: "Upload image"
+      },
+
+      name: "#{name}[image]",
```
davidgisbey, line 55: generally we've gone down the pattern of passing in the full name and id of a field. I think it's probs best to follow that to avoid confusion. Wdyt?

So this would be 

```
name: name,
id: id,

```
and then we would pass in 

```
  <%= render "components/single-image-upload", {
        id: "take_part_page_image",
        name: "take_part_page[image]",
        error_items: errors_for(form.object.errors, :image),
        image_src: form.object.image.url,
        image_alt: form.object.image_alt_text
      } %>
```

app/views/components/_single-image-upload.html.erb
```diff
@@ -0,0 +1,71 @@
...
+    } %>
+  <% end %>
+
+  <%= render "govuk_publishing_components/components/input", {
+    label: {
+      text: "Image description (alt text)"
+    },
+    name: "#{name}[image_alt_text]",
+    id: "#{id}_image_alt_text",
+    value: image_alt == "" ? "None" : image_alt # form.object.image_alt_text
```
davidgisbey, line 69: would this be nil since if one is passed in not ""?

If that is the case i think this might be slightly cleaner.

```
 value: image_alt.presence || "None"
```

`.presence` returns either the object if one is present or nil if not

davidgisbey:
If i update the image, then force a failed save (i made the title blank)

image

the current image disappears

image

If i then fill in the title field and click save, the original image is retained.

image

I'm not 100% sure on what the correct thing to do here is. We could:

1. show the currently uploaded image by passing in the carrierwave image
2. show the new image (although i'm not actually sure we can if it hasn't been uploaded to asset store. Might be one to check with Ollie if it can't wait until Monday)
3. let the user know the image they've uploaded will be uploaded on a successful save

app/views/components/_single-image-upload.html.erb
davidtrussler: Yeah that's true, though we do need to display "None" next to the image when there is no alt text set, so I've move that to that point. 
```
<p class="govuk-body govuk-!-margin-0"><strong>Alt text: </strong><%= image_alt == "" ? "None" : image_alt %></p>
```

davidtrussler:
> I'm not 100% sure on what the correct thing to do here is. We could:
> 
> 1. show the currently uploaded image by passing in the carrierwave image
> 2. show the new image (although i'm not actually sure we can if it hasn't been uploaded to asset store. Might be one to check with Ollie if it can't wait until Monday)
> 3. let the user know the image they've uploaded will be uploaded on a successful save

Hmm, yes I think this will need some wider discussion. I don’t really understand the way uploads are handled well enough to know how to solve the problems we are seeing here, though I guess they are not new but just exposed by the changes we are making to how we present the uploading to the user.

davidgisbey:
Hey @davidtrussler apologies for the delay reviewing this.

I manually tested this and found that on an unsuccessful save it wasn't keeping the new image cached. I wasn't exactly sure how to fix this so i gave it a go and found that this resolves the issue.

https://github.com/alphagov/whitehall/commit/a52dfd686ddbf3353e718fb80443ded24b1d1db8

It's not the best though!

Also the carrierwave image is the currently uploaded image, so it was continuing to show the previous filename.

davidgisbey (approved):
lgtm!

PR #7644, opened 2023-05-11
116-promotional-features-index-page-to-gds

farahTW:
**Description**
This PR transitions the Promotional Features Index page to the GOV.UK Design System.

It does the following:

Duplicates the promotional features controller tests and assigns the non-legacy specs user the "Preview design system permission".
Prepends legacy to the existing index html for the bootstrap implementation and adds the logic to the controller to render the correct implementation based on the users permissions
Adds the new index html  implemented with  GOV.UK Design System
Modified a feature test for coverage of promotional feature index page.

**Screen shots**
**Before:**
Old_index_page

**After :**
Index_gds_pf

**Screen shot with no promotional features:**
Index_with_no_pf

**Trello:**
https://trello.com/c/P2TvBUC6/116-promotional-features-index-page

### DIFF
--- a/app/views/admin/promotional_features/index.html.erb
+++ b/app/views/admin/promotional_features/index.html.erb
@@ -6,7 +6,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <p class="govuk-body">
-      <%= view_on_website_link_for @organisation, class: "govuk-link" %>
+      <%= view_on_website_link_for @organisation, class: "govuk-link", target: "blank" %>
     </p>
 
     <%= render "components/secondary_navigation", {
@@ -47,9 +47,9 @@
         end
       } %>
     <% else %>
-        <div class="govuk-body app-view-features-search-results__no_documents">
-          No promotional features
-        </div>
+      <%= render "govuk_publishing_components/components/inset_text", {
+        text: "No promotional features."
+      } %>
     <% end %>
   </div>
 </div>
--- a/app/views/admin/promotional_features/index.html.erb
+++ b/app/views/admin/promotional_features/index.html.erb
@@ -9,7 +9,7 @@
 
 <%= render "components/secondary_navigation", {
   aria_label: "Promotional features navigation tabs",
-  items: secondary_navigation_tabs_items(current_user, @organisation, request.path)
+  items: secondary_navigation_tabs_items( @organisation, request.path)
 } %>
 
 <%= render "govuk_publishing_components/components/warning_text", {
--- a/app/views/admin/promotional_features/index.html.erb
+++ b/app/views/admin/promotional_features/index.html.erb
@@ -1,39 +1,54 @@
-<% page_title @organisation.name %>
+<% content_for :context, "Organisation" %>
+<% content_for :title, @organisation.name %>
+<% content_for :title_margin_bottom, 4 %>
 
-<div class="organisation-header">
-  <h1><%= @organisation.name %></h1>
-  <%= view_on_website_link_for @organisation %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+<p class="govuk-body">
+  <%= view_on_website_link_for @organisation, class: "govuk-link" %></p>
+
+<%= render "components/secondary_navigation", {
+  aria_label: "Promotional features navigation tabs",
+  items: secondary_navigation_tabs_items(current_user, @organisation, request.path)
+} %>
+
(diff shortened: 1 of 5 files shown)


features/step_definitions/promotional_features_steps.rb
```diff
@@ -158,10 +169,19 @@
 end
 
 Then(/^the promotional features should be in the following order:$/) do |promotional_feature_list|
-  promotion_feature_ids = all(".promotional_feature").map { |element| element[:id] }
+  if using_design_system?
```
davidgisbey, line 172: This page only ever uses the [GOV.UK Design System](https://github.com/alphagov/whitehall/pull/7644/files#diff-1de1f6ca4df7f1675999edf0f4c9af90bbcfcd1c001d84a5d134c68859ff3a20L67) so i think we just need one bit of logic that works here.

app/views/admin/promotional_features/index.html.erb
```diff
@@ -1,39 +1,54 @@
...
-</section>
+<% content_for :page_title, @organisation.name %>
+<% content_for :context, "Organisation" %>
+<% content_for :title, @organisation.name %>
+<% content_for :title_margin_bottom, 4 %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+<p class="govuk-body">
+  <%= view_on_website_link_for @organisation, class: "govuk-link" %></p>
```
davidgisbey, line 9: super nitty but i think it's a bit more readable if we move the closing tag to the same indentation as the opening tag
```
<p class="govuk-body">
  <%= view_on_website_link_for @organisation, class: "govuk-link" %>
</p>
```

app/views/admin/promotional_features/index.html.erb
```diff
@@ -1,39 +1,54 @@
...
+          format: "numeric",
+        }
+      ]
+    end
+  } %>
+<% else %>
+  <div class="govuk-body app-view-features-search-results__no_documents">
+    No promotional features
+  </div>
+<% end %>
```
davidgisbey, line 52: https://github.com/alphagov/whitehall/pull/7644/files#diff-0c62d94e9441510a913c4adf568b1756985e7487be4beb1bd945a4388e3b38f6R8-R52

Please could we outdent this by 4 so it's clear it's within the 2/3 width div.

app/views/admin/promotional_features/index.html.erb
```diff
@@ -1,39 +1,54 @@
...
+
+<%= render "govuk_publishing_components/components/button", {
+  text: "New promotional feature",
+  href: new_admin_organisation_promotional_feature_path,
+  margin_bottom: 6,
+} %>
+
+<% if @promotional_features.many? %>
+  <p class="govuk-body">
+    <%= link_to "Reorder promotional features", reorder_admin_organisation_promotional_features_path(@organisation), class: "govuk-link", target: "_blank" %>
```
davidgisbey, line 28: I don't think this needs the `target: "_blank" `. Generally we don't open links on a new tab unless there's a good reason to as it's hard for screen reader users to follow what's going on.

When we do we would indicate that a new tab is being opened by doing something like

`"Reorder promotional features (opens in new tab)"`

For this though.i think it's totally fine it does it in the same tab

app/views/admin/promotional_features/index.html.erb
```diff
@@ -1,39 +1,54 @@
...
+      ]
+    end
+  } %>
+<% else %>
+  <div class="govuk-body app-view-features-search-results__no_documents">
+    No promotional features
+  </div>
+<% end %>
+  </div>
+</div>
```
davidgisbey, line 54: We leave a new line at the end of files. There's normally an option in you ide to set this up.

Explanation here if you're bored and want to read why https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

davidgisbey:
I know it was a step up from the previous pages you've worked on, but this is a fantastic 1st attempt at this page! 

I've left a few comments but they're all fairly minor. I've tested it locally and it's working excellently. Great job!

features/step_definitions/promotional_features_steps.rb
farahTW: If i put logic for preview design system then its failing in default mode

farahTW:
Thanks David !

CL-GDS (approved):
LGTM, changes requested have been implemented

nnagewad:
Looks good from a design perspective 👍🏽

PR #7646, opened 2023-05-11
Reduce noise in test output

floehopper:
There's quite a lot of noise (i.e. expected or unintended output) in the build output. Reducing this noise will make it easier to see any *unexpected* output which might be important to address.

This PR doesn't deal with *all* the noise, but hopefully it's a significant improvement. We can always do more and (ideally) fail fast if any more gets added.

I don't think there's anything particularly controversial in this PR, but I'd be particularly interested in any thoughts about the Rummager-related fix I've done in the "Avoid RestClient warning from RummageableTest" commit.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,7 +17,12 @@ rescue LoadError
   # Pact isn't available in all environments
 end
 
+desc "Run the 'pact verify' command with the RSpec 'progress' formatter"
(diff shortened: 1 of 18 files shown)


ollietreend (approved):
Excellent work tidying this all up! ⭐ 

This is a big improvement for DX!

PR #7652, opened 2023-05-15
Remove further uses of `UrlMaker`

brucebolt:
There were some uses of `UrlMaker` that we had not previously identified.

This removes all the direct uses of `UrlMaker`.  There are also indirect uses via `Whitehall.url_maker`, which will be written up as another card.

[Trello card](https://trello.com/c/swZNpdoQ)

### DIFF
--- a/app/components/admin/editions/show/sidebar_actions_component.rb
+++ b/app/components/admin/editions/show/sidebar_actions_component.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class Admin::Editions::Show::SidebarActionsComponent < ViewComponent::Base
+  include Admin::EditionRoutesHelper
+
   def initialize(edition:, current_user:)
-    @url_maker = Whitehall::UrlMaker.new(host: Plek.find("whitehall"))
     @enforcer = Whitehall::Authority::Enforcer.new(current_user, edition)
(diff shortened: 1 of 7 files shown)


floehopper (approved):
Overall this all looks good to me! 👍 

One minor question: The Trello card mentions that we're trying to get rid of `Admin::EditionRoutesHelper` as well as `UrlMaker`, but the first commit introduces a new use of `Admin::EditionRoutesHelper`. I assume that getting rid of `Admin::EditionRoutesHelper` is going to be tackled as a separate step, but it might be nice to explain in the commit note why it's OK to introduce it here...? However, I'm happy to leave that up to your discretion.

Marking as approved! ✅

brucebolt:
> One minor question: The Trello card mentions that we're trying to get rid of `Admin::EditionRoutesHelper` as well as `UrlMaker`, but the first commit introduces a new use of `Admin::EditionRoutesHelper`. I assume that getting rid of `Admin::EditionRoutesHelper` is going to be tackled as a separate step, but it might be nice to explain in the commit note why it's OK to introduce it here...? However, I'm happy to leave that up to your discretion.

When doing the `UrlMaker` work, I think we realised we wouldn't be removing `Admin::EditionRoutesHelper` as that's not related to the public rendering paths.  But we never updated the Trello card.  I'll remove reference to it now.

PR #7654, opened 2023-05-15
Ensure first_published_at param is cleaned on update

davidgisbey:
## Description

At the moment, there's an issue with the first_published_at value not being set to nil when a user chooses 'This document has never been published before.' for the document status radios.

When you update the radio from previously published to not previously published the first_published_at params are still submitted on save. As these params are not cleaned before save they continue to be persisted on the edition.

## Trello card

https://trello.com/c/7O1ypSvX/145-document-status-conditional-field-cannot-be-changed

## Video of before & after

https://github.com/alphagov/whitehall/assets/42515961/50f56d0b-4680-47d3-85f2-c1ca332913aa

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -465,6 +465,14 @@ private
     edition_params.delete(:create_foreign_language_only)
     edition_params[:external_url] = nil if edition_params[:external] == "0"
     edition_params[:change_note] = nil if edition_params[:minor_change] == "true"
+
+    if edition_params[:previously_published] == "false"
+      edition_params["first_published_at(1i)"] = ""
+      edition_params["first_published_at(2i)"] = ""
+      edition_params["first_published_at(3i)"] = ""
+      edition_params["first_published_at(4i)"] = ""
+      edition_params["first_published_at(5i)"] = ""
+    end
   end
 
   def clear_scheduled_publication_if_not_activated
--- a/test/functional/admin/legacy_publications_controller_test.rb
+++ b/test/functional/admin/legacy_publications_controller_test.rb
@@ -71,6 +71,7 @@ class Admin::LegacyPublicationsControllerTest < ActionController::TestCase
            edition: controller_attributes_for(
              :publication,
              first_published_at: Time.zone.parse("2001-10-21 00:00:00"),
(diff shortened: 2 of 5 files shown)


app/controllers/admin/editions_controller.rb
```diff
@@ -467,6 +468,16 @@ def clean_edition_parameters
     edition_params[:change_note] = nil if edition_params[:minor_change] == "true"
```
ollietreend, line 468: Any reason why you've introduced a new method rather than extending the `clean_edition_parameters` method?

It seems like it'd be a good fit given the fact that it already handles `external_url` and `change_note`.

```suggestion
    edition_params[:change_note] = nil if edition_params[:minor_change] == "true"
    if edition_params[:previously_published] == "false"
      edition_params["first_published_at(1i)"] = ""
      edition_params["first_published_at(2i)"] = ""
      edition_params["first_published_at(3i)"] = ""
      edition_params["first_published_at(4i)"] = ""
      edition_params["first_published_at(5i)"] = ""
    end
```

test/functional/admin/generic_editions_controller_test.rb
```diff
@@ -30,4 +30,22 @@ class Admin::GenericEditionsControllerTest < ActionController::TestCase
 
     assert_select ".govuk-error-summary a", text: "Title is already used on GOV.UK. Please create a unique title", href: "#edition_title"
   end
+
+  test "PATCH :update updates first_published_at to nil when previously_published is false" do
```
ollietreend, line 34: Is the reverse of this already covered somewhere?

i.e. that the `first_published_at` field gets set correctly when the radio value is `"true"`?

davidgisbey: yea the 4 tests below cover that behaviour. They started failing without it as this defaults to false in the tests.

app/controllers/admin/editions_controller.rb
davidgisbey: yea good shout. i think there's a few param cleaning methods so i sort of just followed the pattern, but they're more complex so it sort of makes sense they're discreet.

ollietreend (approved):
Nice! Thanks @davidgisbey ✨🚀

PR #7655, opened 2023-05-15
126 promotional- feature- item -confirm -delete

farahTW:
**Description**
This PR creates the new Promotional Features Items Confirm-Destroy page to the GOV.UK Design System.

**It does the following:**

- Duplicates the promotional features items controller tests and assigns the non-legacy specs user the "Preview design system permission".
- Prepends legacy to the existing controller and adds the logic to the non legacy controller to render the correct implementation based on the users permissions.
- Adds the confirm destroy html implemented with GOV.UK Design System.
- Add confirm destroy end point test to the controller test file.

**Screen shots:**
**Before:**
Screenshot 2023-05-15 at 12 12 12

**After:**
Screenshot 2023-05-15 at 12 12 36

**Trello:**
https://trello.com/c/ltljsK1j/126-promotional-feature-item-confirm-delete-page


app/controllers/admin/legacy_promotional_feature_items_controller.rb
```diff
@@ -0,0 +1,92 @@
+class Admin::PromotionalFeatureItemsController < Admin::BaseController
```
davidgisbey, line 1: Hey, so we don't need to duplicate the controller itself. Just the tests like you've done here https://github.com/alphagov/whitehall/pull/7655/commits/b535da2272ecfbb933340f5090b869907f6936c4#diff-d54c5fbac923edc9765a4dc8052b520ed67628c4d98deb1957827f03514e9d14R1-R135

The controller itself handles all the routing of whether the user should see the [ GOV.UK Design System layout ](https://github.com/alphagov/whitehall/pull/7655/commits/b535da2272ecfbb933340f5090b869907f6936c4#diff-73f8697b8bb7cee92fcaf82cdeeae53252c0178481893a628c48e1d866ab3b6fR49-R56) and views.

Let's have a chat about the "Preview design system" permission when we're pairing later as it'll be easier to explain the context then rather than me typing out an essay. Apologies i should've gone though it earlier!

app/views/admin/promotional_feature_items/_promotional_feature_item.html.erb
```diff
@@ -22,9 +22,5 @@
   </p>
 
   <%= link_to "Edit", edit_admin_organisation_promotional_feature_item_path(@organisation, @promotional_feature, promotional_feature_item), class: "btn btn-default"%>
-  <%= button_to "Delete",
-        admin_organisation_promotional_feature_item_path(@organisation, @promotional_feature, promotional_feature_item),
-        method: "delete",
-        class: "btn btn-danger",
-        data: { confirm: "Are you sure you want to delete this feature item?" } %>
+  <%= link_to "Delete", confirm_destroy_admin_organisation_promotional_feature_item_path(@organisation,@promotional_feature, promotional_feature_item), class: 'btn btn-danger btn-sm' %>
```
davidgisbey, line 25: I thin for now we should leave this as a button, and update it to a link in the GOV.UK Design System implementation

```
<%= button_to "Delete",
        confirm_destroy_admin_organisation_promotional_feature_item_path(@organisation,@promotional_feature, promotional_feature_item),
        method: "get",
        class: "btn btn-danger",
```

davidgisbey, line 25: I think we should continue to use implementation for Bootstrap. i.e. not hit the confirm delete and just use the code deleted here

image

We should use link to the confirm delete page for our new implementation though (so on the index page you just added)

app/views/admin/promotional_feature_items/confirm_destroy.html.erb
```diff
@@ -0,0 +1,24 @@
+<% content_for :context, @organisation.name %>
+<% content_for :page_title, "Delete #{@promotional_feature_item.title}" %>
+<% content_for :title, "Delete #{@promotional_feature_item.title}" %>
+<% content_for :title_margin_bottom, 6 %>
+
+<% page_title "Feature item for #{@promotional_feature_item.title}" %>
```
davidgisbey, line 6: We should only ever be hitting this page when we use the GOV.UK Design System so we don't need to worry about setting the page title for the bootstrap implementation here

app/controllers/admin/promotional_feature_items_controller.rb
```diff
@@ -44,9 +44,16 @@ def destroy
     Whitehall::PublishingApi.republish_async(@organisation)
     redirect_to_feature "Feature item deleted."
   end
-
+  def confirm_destroy; end
 private
-
+  def get_layout
+    design_system_actions = %w[confirm_destroy]
+    if preview_design_system?(next_release: false) && design_system_actions.include?(action_name)
```
davidgisbey, line 51: we don't have a bootstrap implementation for the confirm delete page so i think we can just have 

```
 if design_system_actions.include?(action_name)
  "design_system"
 else
   "admin"
 end
  end
```

test/functional/admin/promotional_feature_items_controller_test.rb
```diff
@@ -131,4 +131,14 @@ class Admin::PromotionalFeatureItemsControllerTest < ActionController::TestCase
     assert_not PromotionalFeatureItem.exists?(promotional_feature_item.id)
     assert_equal "Feature item deleted.", flash[:notice]
   end
+
+
+  test "GET :confirm_destroy calls correctly" do
```
davidgisbey, line 136: This commit looks good but it needs a rubocop please

farahTW: Ok I ran the rubocop..will check n run again no worries

app/controllers/admin/legacy_promotional_feature_items_controller.rb
davidgisbey: So this controller can be completely removed. We already have this controller here https://github.com/alphagov/whitehall/pull/7655/files#diff-73f8697b8bb7cee92fcaf82cdeeae53252c0178481893a628c48e1d866ab3b6fR1-R64

We only duplicate the controller specs not the controller itself

davidgisbey:
Looks good other than the one comment i've left about removing the duplicate controller. 

Please could we remove that and squash some of the commits together.

I think having 1 commit or 2 where one duplicates the controller tests and the other adds the endpoint is fine.

davidgisbey (approved):

PR #7658, opened 2023-05-15
WIP: Remove special cases

chrisroos:


db/data_migration/20230515143400_remove_central_african_republic_embassy_special_case.rb
```diff
@@ -0,0 +1,14 @@
+location = WorldLocation.find_by!(slug: "central-african-republic")
+organisation = WorldwideOrganisation.find_by!(slug: "british-embassy-kinshasa")
+organisation.world_locations += [location] unless organisation.world_locations.include?(location)
+
+location.reload
+
+embassy = Embassy.new(location)
+unless embassy.organisations_with_embassy_offices == [organisation]
```
BeckaL, line 8: Is this all run in a transaction so raising will rollback? I wasn't sure it was failing fast to do this check at this point, but imagine it's because it's wrapped in a transaction

db/data_migration/20230515143400_remove_central_african_republic_embassy_special_case.rb
```diff
@@ -0,0 +1,14 @@
+location = WorldLocation.find_by!(slug: "central-african-republic")
+organisation = WorldwideOrganisation.find_by!(slug: "british-embassy-kinshasa")
+organisation.world_locations += [location] unless organisation.world_locations.include?(location)
+
+location.reload
+
+embassy = Embassy.new(location)
```
BeckaL, line 7: I think I'm missing something here, what does this line do (i.e. obviously it creates an embassy, but Embassies aren't stored in the database anywhere so I was reading this as having no effect)

BeckaL (commented):
Great work on following this through and reaching all these agreements! Really nice to tidy this stuff up. Generally looks good to me, thanks for all the detail in the commit messages, just two questions below

db/data_migration/20230515143400_remove_central_african_republic_embassy_special_case.rb
chrisroos: We use the [`Embassy#organisations_with_embassy_offices`](https://github.com/alphagov/whitehall/blob/d5e8993a2059c0deddb98438f9e85988ec151938/app/models/embassy.rb#L8) method to detect whether the embassy organisations for that location match what we expect them to be after we've changed the data. Does that make sense?

db/data_migration/20230515143400_remove_central_african_republic_embassy_special_case.rb
chrisroos: Although there appears to be a precedent for [using `raise` in these data migrations](https://github.com/search?q=repo%3Aalphagov%2Fwhitehall+raise+path%3A%2F%5Edb%5C%2F%2F+path%3A%2F%5Edb%5C%2Fdata_migration%5C%2F%2F&type=code), you've made me realise that this probably isn't what I want to happen. I think it'll be good enough to write any unexpected data to stdout for us to manually investigate and fix. I'll make this change now.

db/data_migration/20230515143400_remove_central_african_republic_embassy_special_case.rb
BeckaL: Oh 🤦‍♀️ 🤦‍♀️ 🤦‍♀️  sorry yes was thinking about that failing fast stuff separately from this

db/data_migration/20230515143400_remove_central_african_republic_embassy_special_case.rb
chrisroos: I've addressed this in https://github.com/alphagov/whitehall/pull/7658/commits/68706888d0165ca0de809c1916c3cb1c2571539e.

chrisroos:
Thanks @BeckaL! I've addressed your feedback. Can you take another look when you have a mo, please?

chrisroos:
Thanks, @BeckaL! :+1:

PR #7660, opened 2023-05-15
Develop updated connected documents functionality

chriswhitehouse-tw:
## Description

This PR implements a new flow for the connection of documents to statistics announcements, incorporating the GOV.UK design system.

It has established two new routes:

* /government/admin/statistics_announcements/{statistics_announcement_slug}/publication
* /government/admin/statistics_announcements/{statistics_announcement_slug}/publication/{publication_id}/connect

Note, this flow is not yet fully implemented end to end, and will be completed as part of 'move the statistics announcement summary to design system' story.

In addition, it appears manual testing of the filtered results suggests some slightly odd behaviour, with un permitted documents being filtered out of the page results. As such some paginated sets have less than 15 results.

The PR does the following:

1. Add statistics announcements publication index page
2. Add basic document table, conditional rendering on search param being provided
3. Add view links to results table
4. Add connect links and functionality to results table
5. Add error handling to connect links functionality
6. Add pagination to search results

#### Before search string entered
whitehall-admin dev gov uk_government_admin_statistics_announcements_title--13_publication (1)

#### With search results
whitehall-admin dev gov uk_government_admin_statistics_announcements_title--13_publication_search=transport

## Trello
https://trello.com/c/HNwwVfU0

### DIFF
--- a/app/controllers/admin/statistics_announcement_publications_controller.rb
+++ b/app/controllers/admin/statistics_announcement_publications_controller.rb
@@ -8,8 +8,21 @@ class Admin::StatisticsAnnouncementPublicationsController < Admin::BaseControlle
     end
   end
 
-  private
+  def connect
+    @statistics_announcement.assign_attributes(publication_params)
+
+    @statistics_announcement.save
+
+    redirect_to [:admin, @statistics_announcement], notice: "Announcement updated successfully"
+  end
+
+private
+
   def find_statistics_announcement
     @statistics_announcement = StatisticsAnnouncement.friendly.find(params[:statistics_announcement_id])
+  end
+
+  def publication_params
+    { publication_id: params[:publication_id] }
   end
 end
--- a/app/controllers/admin/statistics_announcement_publications_controller.rb
+++ b/app/controllers/admin/statistics_announcement_publications_controller.rb
@@ -26,11 +26,11 @@ private
   end
 
   def get_editions
-    Edition.statistical_publications.with_title_containing(params[:search])
+    Edition.with_title_containing(params[:search])
   end
 
   def params_filters
-    params.slice(:type, :state, :organisation, :author, :page, :title, :world_location, :from_date, :to_date, :only_broken_links)
+    params.slice(:page)
           .permit!
           .to_h
   end
@@ -40,15 +40,16 @@ private
   end
 
   def edition_filter_options
-    filter_options = params_filters_with_default_state
+    params_filters_with_default_state
                        .symbolize_keys
                        .merge(
                          include_unpublishing: true,
                          include_link_check_reports: true,
                          include_last_author: true,
+                         type: "publication",
+                         subtypes: @statistics_announcement.publication_type,
+                         per_page: Admin::EditionFilter::GOVUK_DESIGN_SYSTEM_PER_PAGE,
                        )
-
-    filter_options.merge(per_page: Admin::EditionFilter::GOVUK_DESIGN_SYSTEM_PER_PAGE)
   end
 
   def find_statistics_announcement
--- a/app/controllers/admin/statistics_announcement_publications_controller.rb
+++ b/app/controllers/admin/statistics_announcement_publications_controller.rb
@@ -3,20 +3,27 @@ class Admin::StatisticsAnnouncementPublicationsController < Admin::BaseControlle
   layout "design_system"
 
   def index
-    if params[:search].present?
-      @editions = Edition.statistical_publications.with_title_containing(params[:search])
-    end
+    get_editions
   end
 
   def connect
     @statistics_announcement.assign_attributes(publication_params)
 
-    @statistics_announcement.save
-
-    redirect_to [:admin, @statistics_announcement], notice: "Announcement updated successfully"
+    if @statistics_announcement.save
+      redirect_to [:admin, @statistics_announcement], notice: "Announcement updated successfully"
+    else
+      get_editions
+      render :index
+    end
   end
 
 private
+
+  def get_editions
+    if params[:search].present?
+      @editions = Edition.statistical_publications.with_title_containing(params[:search])
+    end
+  end
 
   def find_statistics_announcement
     @statistics_announcement = StatisticsAnnouncement.friendly.find(params[:statistics_announcement_id])
--- a/app/controllers/admin/statistics_announcement_publications_controller.rb
+++ b/app/controllers/admin/statistics_announcement_publications_controller.rb
@@ -22,7 +22,11 @@ class Admin::StatisticsAnnouncementPublicationsController < Admin::BaseControlle
 private
 
   def filter
-    @filter ||= Admin::EditionFilter.new(Edition, current_user, edition_filter_options)
+    @filter ||= Admin::EditionFilter.new(edition_scope, current_user, edition_filter_options)
+  end
+
+  def edition_scope
+    Edition.with_translations(I18n.locale)
   end
 
   def params_filters
@@ -39,9 +43,6 @@ private
     params_filters_with_default_state
                        .symbolize_keys
                        .merge(
-                         include_unpublishing: true,
-                         include_link_check_reports: true,
-                         include_last_author: true,
                          type: "publication",
                          subtypes: @statistics_announcement.publication_type,
                          per_page: Admin::EditionFilter::GOVUK_DESIGN_SYSTEM_PER_PAGE,
--- a/app/controllers/admin/statistics_announcement_publications_controller.rb
+++ b/app/controllers/admin/statistics_announcement_publications_controller.rb
@@ -3,7 +3,9 @@ class Admin::StatisticsAnnouncementPublicationsController < Admin::BaseControlle
   layout "design_system"
 
   def index
-    get_editions
+    if params[:search].present?
+      filter
+    end
   end
 
   def connect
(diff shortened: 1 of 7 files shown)


app/controllers/admin/statistics_announcement_publications_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+private
+
+  def filter
+    @filter ||= Admin::EditionFilter.new(get_editions, current_user, edition_filter_options)
+  end
+
+  def get_editions
+    Edition.statistical_publications.with_title_containing(params[:search])
```
davidgisbey, line 29: hey chris, great work. I noticed one thing here that i think is causing a bit of an issue. Often when i try and link the document i'm hitting this error


image

Looking at a statistics announcement i can see that it can either be linked a publication with the subtype of official on national statistics.

image

Looking at the JS implementation i can see that the publication_type_id is passed into the JS module

image

This is used to only return publications with a title of that subtype when the ajax call is made

image

image

I think we need to update this method to only pass in editions with the statistics announcements `publication_type_id`

As you've set this up in a before_action we can just do
```
    Edition.statistical_publications.where(publication_type_id: @statistics_announcement.publication_type_id).with_title_containing(params[:search])

```
When i pass this extra where clause it reduces the amount of returned editions from 250 to 20. All of which can be linked 

image


image

app/controllers/admin/statistics_announcement_publications_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  before_action :find_statistics_announcement
+  layout "design_system"
+
+  def index
+    if params[:search].present?
+      filter
+    end
+  end
+
+  def connect
```
davidgisbey, line 11: Do you think we could just call this action `update`? that way we can just use an update resource rather than a custom route? 

```
resources :statistics_announcement_publications, as: "publication", path: "publication", only: %i[index update]
```

app/controllers/admin/statistics_announcement_publications_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def get_editions
+    Edition.statistical_publications.with_title_containing(params[:search])
+  end
+
+  def params_filters
+    params.slice(:type, :state, :organisation, :author, :page, :title, :world_location, :from_date, :to_date, :only_broken_links)
+          .permit!
+          .to_h
```
davidgisbey, line 35: So i think we can simplify things here a bit. If we update the name of the search field to title, we can pass that into the edition filter and it will do the work to filter on title for us

```
def params_filters
    params.slice(:title)
          .permit!
          .to_h
end
````
then we can pass the publication type and subtype in as filters like this
```
def edition_filter_options
    filter_options = params_filters_with_default_state
                       .symbolize_keys
                       .merge(
                         include_unpublishing: true,
                         include_link_check_reports: true,
                         include_last_author: true,
                         type: "publication",
                         subtypes: @statistics_announcement.publication_type_id,
                       )

    filter_options.merge(per_page: Admin::EditionFilter::GOVUK_DESIGN_SYSTEM_PER_PAGE)
  end

```

File: app/controllers/admin/statistics_announcement_publications_controller.rb [hunk 4]
```diff
@@ -0,0 +1,61 @@
+class Admin::StatisticsAnnouncementPublicationsController < Admin::BaseController
+  before_action :find_statistics_announcement
+  layout "design_system"
+
+  def index
+    if params[:search].present?
```
davidgisbey, line 6: If we do this then this would need be updated to `params[:title]` https://github.com/alphagov/whitehall/pull/7660/files#r1195196818

app/views/admin/statistics_announcement_publications/index.html.erb
```diff
@@ -0,0 +1,33 @@
+<% content_for :page_title, "Connect document" %>
+<% content_for :context, @statistics_announcement.title %>
+<% content_for :title, "Connect document" %>
+<% content_for :title_margin_bottom, 6 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @statistics_announcement)) %>
+
+<%= form_with url: admin_statistics_announcement_publication_index_path(@statistics_announcement), method: :get do |_form| %>
+  <%= render "govuk_publishing_components/components/search", {
+    name: "search",
```
davidgisbey, line 9: If we do this then this would need be updated to `title` https://github.com/alphagov/whitehall/pull/7660/files#r1195196818

davidgisbey:
I've left some comments with some potential changes, but this is excellent work Chris!

test/functional/admin/statistics_announcement_publications_controller_test.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    get :index, params: { statistics_announcement_id: @announcement, search: @search }
+  end
+
+  view_test "GET :index with search value renders paginated results" do
+    editions = []
+    16.times { editions << @statistics_publication }
+
+    stub_filter = stub_edition_filter({ editions:, options: { per_page: 15 } })
+    Admin::EditionFilter.stubs(:new).returns(stub_filter)
```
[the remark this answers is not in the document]
chriswhitehouse-tw: Hi @ollietreend, this is kinda where I started with the test, but found that the test user didn't have the necessary permissions to see the test publications, so they were being filtered out by the permitted_only method in the EditionFilter.

I picked up the stubbing from the Editions controller tests, which is probably better practice anyhow. Although, there doesn't seem to be a standalone test suite for the EditionFilter.

test/functional/admin/statistics_announcement_publications_controller_test.rb
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  test "GET :connect will add a document to a statistics announcement" do
+    get :connect, params: { statistics_announcement_id: @announcement, publication_id: @statistics_publication }
+
+    assert_equal @statistics_publication, @announcement.reload.publication
+    assert_redirected_to admin_statistics_announcement_path(@announcement)
+  end
+
+  view_test "GET: connect will handle an error in adding not a statistics publication to statistics announcement" do
```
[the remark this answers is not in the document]
chriswhitehouse-tw: ✅

app/controllers/admin/statistics_announcement_publications_controller.rb
chriswhitehouse-tw: Cool, I spent a bit of time playing around with this to get the button to style like a link, but wasn't happening quickly. So have decided to stick with current implementation.

app/controllers/admin/statistics_announcement_publications_controller.rb
chriswhitehouse-tw: ✅

app/views/admin/statistics_announcement_publications/index.html.erb
chriswhitehouse-tw: ✅

app/controllers/admin/statistics_announcement_publications_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    params_filters.reverse_merge("state" => "active")
+  end
+
+  def edition_filter_options
+    params_filters_with_default_state
+                       .symbolize_keys
+                       .merge(
+                         include_unpublishing: true,
+                         include_link_check_reports: true,
+                         include_last_author: true,
```
davidgisbey, line 44: sorry i should've asked this before, but are we using these 3 params

```
include_unpublishing: true,
include_link_check_reports: true,
include_last_author: true,
```

chriswhitehouse-tw: Yes, double checked this agains the documents search controller, and it doesn't look like they are necessary, so I have removed them.

app/views/admin/statistics_announcement_publications/index.html.erb
```diff
@@ -0,0 +1,33 @@
...
+<% content_for :context, @statistics_announcement.title %>
+<% content_for :title, "Connect document" %>
+<% content_for :title_margin_bottom, 6 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @statistics_announcement)) %>
+
+<%= form_with url: admin_statistics_announcement_publication_index_path(@statistics_announcement), method: :get do |_form| %>
+  <%= render "govuk_publishing_components/components/search", {
+    name: "title",
+    value: params[:title]
+  } %>
```
davidgisbey, line 11: Sorry Chris i just noticed one thing 

In Niks design the search has a title 

image

We can add one to it like this 
```
<%= render "govuk_publishing_components/components/search", {
 name: "title",
  value: params[:title],
  label_text: "Search by title or slug"
  label_size: "xl",
} %>
```

test/functional/admin/statistics_announcement_publications_controller_test.rb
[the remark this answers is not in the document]
ollietreend: Oh, weird. I would have thought `login_as_preview_design_system_user(:gds_editor)` would have permission to see those documents 🤷🏻 

I tend to prefer using factories (i.e. 'real' objects) rather than stubbing wherever possible, just because it gives a greater depth of test coverage by exercising more code. Plus it means the stub doesn't need updating if ever the public interface of the 'real' object changes.

But I don't have a strong opinion here. If `create_list(:published_statistics, 15)` isn't working, then feel free to stick with the stubs.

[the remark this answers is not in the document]
chriswhitehouse-tw: Cool, thanks Ollie, I had another play with this, and still can't get the publications to render within the test. So will leave as is.

app/views/admin/statistics_announcement_publications/index.html.erb
chriswhitehouse-tw: thanks David, good spot. I've updated to reflect this now.

davidgisbey (approved):
thanks for that. Great work 👍

nnagewad:
Looks good from a design perspective 👍🏽

PR #7661, opened 2023-05-15
Remove embassies index

chrislo:
Trello: https://trello.com/c/MMUcRbIT/534-move-rendering-of-the-embassies-page-from-whitehall-to-a-frontend-app

The embassies index page (/world/embassies) is now [being rendered](https://www.gov.uk/world/embassies) by the collections app so this PR removes the corresponding rendering code from Whitehall.


brucebolt (approved):
Looks good to me.

I'm a little undecided about how I feel with removing the translations.  One of Platform Health's work streams (about 2 years ago) added translations for all locale keys, but as discussed on Slack, they aren't actually being used for anything so can be removed.

chrislo:
Thanks @brucebolt. I understand the concern, but I think the cost of having code around that's not reachable is not insignificant. I'm confident that we can recover these translations from the original source or the git history if we do need them at some point. And as we have no mechanism to ensure that English strings are re-translated if they ever change, it feels like we'd need to review these translations anyway even if we were ever to use them.

PR #7663, opened 2023-05-15
Move Cabinet minister endpoints to the GOV.UK Design System

davidgisbey:
## Description 

We're moving over the Cabinet minister ordering page to the GOV.UK Design System. This page has quite a lot going on within it. In the previous implementation there were 4 reorderable lists for ministers, also attends cabinet, whips and ministerial organisations.

This uses the new design which breaks these 4 lists into 4 tabs. Each of these has a reorder link which goes off to a separate page to do the reordering.

I've decided to use 4 separate endpoints rather than using a query string to determine which list to reorder as it feels cleaner.

I'd very much recommend reviewing this per commit as it will a lot easier.

#### Before

image
image
image

#### After

image

image

image

image

image

image

image

image

## Trello card

https://trello.com/c/eJJv8zRq/95-reordering-page


chriswhitehouse-tw (commented):
Hi David, 

This all looks really good to me. The only things I've spotted through manual testing are:

* The cancel links do not redirect to the specific refer tab
* Similarly, if you navigate to the edit page and then click cancel you are redirected to the roles index page, and not back to the reordering page. Not sure if this is an issue?

davidgisbey:
Thanks for the feedback. Have pushed up a fix for the first bullet in the last commit. As discussed i think the second bullet will be handled separately.

chriswhitehouse-tw (approved):
Code looks great, and works well. 👍

nnagewad:
This looks great! Good shout on the guidance for the reorder

PR #7665, opened 2023-05-16
Remove all remaining uses of `UrlMaker`

brucebolt:
We have updated all models to know their own routes (through the `public_path` and `public_url` methods), with an aim to removing the "magic" that was previously used to determine the public path of a document.

This removes all remaining use of `UrlMaker` from the application.

`UrlMaker` (and it's associated code) will be removed in a later PR.

[Trello card](https://trello.com/c/Fs7mTsmi)

### DIFF
--- a/test/functional/corporate_information_pages_controller_test.rb
+++ b/test/functional/corporate_information_pages_controller_test.rb
@@ -1,6 +1,8 @@
 require "test_helper"
 
 class CorporateInformationPagesControllerTest < ActionController::TestCase
+  include Rails.application.routes.url_helpers
+
   should_be_a_public_facing_controller
 
   view_test "show renders the summary as plain text" do
@@ -99,7 +101,7 @@ class CorporateInformationPagesControllerTest < ActionController::TestCase
   test "finds unpublishing for a corporate information page" do
     organisation = create(:organisation)
     cip = create(:corporate_information_page, :unpublished, organisation:)
-    alternative_url = Whitehall.url_maker.root_url
+    alternative_url = Plek.website_root
     cip.unpublishing.update!(redirect: true, slug: cip.slug, alternative_url:)
 
     get :show, params: { organisation_id: cip.organisation, id: cip.slug }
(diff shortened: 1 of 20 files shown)


test/unit/services/link_reporter_csv_service_test.rb
```diff
@@ -1,6 +1,9 @@
 require "test_helper"
 
 class LinkReporterCsvServiceTest < ActiveSupport::TestCase
+  include Rails.application.routes.url_helpers
```
chrislo, line 4: I wonder if this commit should be squashed into the previous one - keeping the implementation and test changes together?

test/functional/corporate_information_pages_controller_test.rb
```diff
@@ -99,7 +101,7 @@ class CorporateInformationPagesControllerTest < ActionController::TestCase
   test "finds unpublishing for a corporate information page" do
     organisation = create(:organisation)
     cip = create(:corporate_information_page, :unpublished, organisation:)
-    alternative_url = Whitehall.url_maker.root_url
+    alternative_url = "https://www.gov.uk"
```
chrislo, line 104: Is there a helper method available that returns this string, to avoid duplicating it here?

test/functional/corporate_information_pages_controller_test.rb
```diff
@@ -111,8 +113,7 @@ class CorporateInformationPagesControllerTest < ActionController::TestCase
     organisation = create(:organisation)
     organisation2 = create(:organisation, slug: "another_organisation")
     cip = create(:corporate_information_page, :unpublished, organisation:)
-    alternative_url = Whitehall.url_maker.root_url
-    cip.unpublishing.update!(redirect: true, slug: cip.slug, alternative_url:)
+    cip.unpublishing.update!(redirect: true, slug: cip.slug, alternative_url: "https://www.gov.uk")
```
chrislo, line 116: Is there a helper method available that returns this string, to avoid duplicating it here?

chrislo (approved):
Looks good to me @brucebolt! I've made a few small suggestions/queries but they are minor and shouldn't stop you from merging.

test/unit/presenters/api/worldwide_organisation_presenter_test.rb
```diff
@@ -72,7 +72,7 @@ class Api::WorldwideOrganisationPresenterTest < PresenterTestCase
   end
 
   test "json includes public world organisations url as web_url" do
-    assert_equal Whitehall.url_maker.worldwide_organisation_url(@world_org), @presenter.as_json[:web_url]
+    assert_equal @world_org.public_url, @presenter.as_json[:web_url]
```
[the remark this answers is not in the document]
brucebolt: Same as the other one, I did originally have this in one commit, but split it out into two.

test/functional/corporate_information_pages_controller_test.rb
brucebolt: Yes, will update to use Plek's `website_root` method.

PR #7666, opened 2023-05-16
Remove `UrlMaker` and associated code

brucebolt:
Now that models know their own public route (through the `public_path` and `public_url` methods), we no longer need to use magic to guess what their route could be.

Therefore removing all of the code that was previously used to do this.

[Trello card](https://trello.com/c/Fs7mTsmi/602-remove-uses-of-whitehallurlmaker)


app/models/application_record.rb
```diff
@@ -12,9 +12,16 @@ def append_url_options(path, options = {})
...
+    permitted_query_params = %i[
+      cachebust
+      preview
+      token
+      utm_campaign
+      utm_medium
+      utm_source
+    ]
+
+    query_params = options.slice(*permitted_query_params).compact
```
BeckaL, line 24: oh very nice! It slightly bugged me the way we were doing this before but wasn't sure what the better ruby option was

BeckaL (approved):
Looks great to me! Wondered if you could add a basic line to each of the first few commits to just explain the context of them so they can be understood individually if e.g. looking at a blame locally, but minor so approving anyway.

brucebolt:
> Wondered if you could add a basic line to each of the first few commits to just explain the context of them so they can be understood individually if e.g. looking at a blame locally, but minor so approving anyway.

Good idea 👍.  Have added a more detailed explanation to each commit.

PR #7669, opened 2023-05-16
Added new summary card component and documentation.

farahTW:
GOV UK design system recently added the Summary Card Component. We decided to use the same during the transition.
This adds the relevant component with documentation.
This will be used for promotional feature item index page transition.

**Screen shots:**
image
image
image
image
image
image
image
image

**Trello:**
https://trello.com/c/G7epREYd/156-add-summary-card-component

### DIFF
--- a/app/views/components/_summary_card_component.html.erb
+++ b/app/views/components/_summary_card_component.html.erb
@@ -1,0 +1,41 @@
+<%
+  id ||= nil
+  data_attributes ||= {}
+  summary_card_actions ||= []
+  rows ||=[]
+%>
+<%= tag.div class: "app-c-summary-card-component", id: id, data: data_attributes do %>
+  <div class="govuk-summary-card">
+    <div class="govuk-summary-card__title-wrapper">
+      <h2 class="govuk-summary-card__title"><%= title %></h2>
+      <ul class="govuk-summary-card__actions">
+        <% summary_card_actions.each do |action| %>
+          <li class="govuk-summary-card__action">
+            <%= link_to sanitize(action[:label] + tag.span(" #{title}", class: "govuk-visually-hidden")), action[:href], class: "govuk-link govuk-link--no-visited-state #{"gem-link--destructive govuk-!-font-weight-bold" if action[:destructive]}".strip %>
+          </li>
+        <% end %>
+      </ul>
+    </div>
+    <div class="govuk-summary-card__content">
+      <dl class="govuk-summary-list">
+        <% rows.each do |row| %>
+          <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">
+              <%= row[:key] %>
+            </dt>
+            <dd class="govuk-summary-list__value">
+              <%= row[:value] %>
+            </dd>
+            <% if row[:actions].present? %>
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_summary-card-component.scss
```diff
@@ -0,0 +1,7 @@
+.app-c-summary-card-component {
+  .govuk-summary-list__row {
+    .govuk-summary-list__value {
+      width: 30%;
```
davidtrussler, line 4: I don't think there's any need to specify this width: the `key` elements are already given an explicit width so this effectively sets both the `key` and `value` elements to 50%. 

In that case there would be no need to add specific styles at all in addition to what are already provided.

app/views/components/_summary_card_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  rows ||=[]
+%>
+<%= tag.div class: "app-c-summary-card-component", id: id, data: data_attributes do %>
+  <div class="govuk-summary-card">
+    <div class="govuk-summary-card__title-wrapper">
+      <h2 class="govuk-summary-card__title"><%= title %></h2>
+      <ul class="govuk-summary-card__actions">
+        <% summary_card_actions.each do |action| %>
+          <li class="govuk-summary-card__action">
+            <%= link_to action[:label], action[:href], class: "govuk-link govuk-link--no-visited-state #{"gem-link--destructive govuk-!-font-weight-bold" if action[:destructive]}".strip %>
```
davidtrussler, line 14: For accessibility where we have links we should provide assistive technology with more information so that there isn't a collection of "edit" (or whatever) links with no context. Instead a screen reader would read "edit key one", "edit key two" etc. 

We can either add this with an aria-label or provide some hidden text like `<a>edit<span class="govuk-visually-hidden"> title</span></a>`

app/assets/stylesheets/components/_summary-card-component.scss
davidgisbey: i swear it was moving the edit/delete links down for the rows without this but now it seems to be working fine 😂 have removed it 👍

app/views/components/_summary_card_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+            <dt class="govuk-summary-list__key">
+              <%= row[:key] %>
+            </dt>
+            <dd class="govuk-summary-list__value">
+              <%= row[:value] %>
+            </dd>
+            <% if row[:actions].present? %>
+              <dd class="govuk-summary-list__actions">
+                <% row[:actions].each do |action| %>
+                  <%= link_to action[:label], action[:href], class: "govuk-link govuk-link--no-visited-state govuk-!-margin-left-2 #{"gem-link--destructive" if action[:destructive]}".strip %>
```
davidtrussler, line 32: Oops sorry, we'd want to put the hidden text in here as well - so we have `<a class="govuk-link govuk-link--no-visited-state govuk-!-margin-left-2" href="#1">View<span class="govuk-visually-hidden"> key one</span></a>` etc.

davidtrussler (approved):
👍 LGTM!

PR #7674, opened 2023-05-17
Move worldwide organisation about page content onto worldwide organisation

floehopper:
This PR moves the details for a worldwide organisation page (e.g. https://www.gov.uk/world/organisations/british-embassy-paris) from the CorporateInformationPage model into the worldwide organisation. 

This fixes an oddity in Whitehall, where the contents of the Worldwide Organisation page are not directly editable from the Worldwide Organisation itself (even though the contents are viewable under the details tab on the admin page), but instead these have to be edited on a separate 'About Us' corporate information page which is itself not published anywhere on GOV.UK. This is a confusing experience, and has led to tech debt where we still publish a draft of the Worldwide Org's about page to the draft content store, and this can conflict with the base path for the Worldwide Organisation itself.

It's worth noting that _Organisations_ (as opposed to Worldwide Organisations) can have About Us pages that do appear on GOV.UK (with a separate URL path) and shouldn't be affected by any of these changes.

One implication of this change is that, since Corporate Information pages are editioned, but Worldwide Organisations are not, the editionable functionality of the body and summary text for the Worldwide Organisation is lost. This has been discussed and agreed with FCDO, and we have [introduced a simple audit history](https://github.com/alphagov/whitehall/pull/7583) for Worldwide Organisations to make up for some of this lost functionality.

This PR contains:
* Some initial tidying commits that make the switch easier
* A migration to copy data from Worldwide Organisation About Us pages (and translations) to the Worldwide Organisation, and updates to the UI to make this content editable under the Worldwide Organisation. This migration took ~11 secs to run on a developer machine against a recent database dump.
* A migration to update Worldwide Organisation About Us pages' attachments to instead be attached to the Worldwide Organisation, and adds the component to edit attachments to the Worldwide Organisation. This migration took ~3 secs to run on a developer machine against a recent database dump.
* Logic to suppress the About Us pages from being viewable from the Worldwide Organisation's corporate information pages, and some explanatory text to explain that this content should now be edited under the main details tab
* A pending migration to delete the Worldwide Organisation about us pages

UI Screenshots:

| Feature | Previous to this PR | After this PR |
| ------ | ----- | ------- |
| Editing the Worldwide Organisation | Screenshot 2023-05-18 at 15 55 12 |whitehall-admin dev gov uk_government_admin_worldwide_organisations_british-consulate-general-milan_edit (1)|
| Editing the Worldwide Organisation Translation |Screenshot 2023-05-18 at 15 55 43 |whitehall-admin dev gov uk_government_admin_worldwide_organisations_british-consulate-general-milan_translations_it_edit|
| Viewing all corporate information pages for a Worldwide Org |redacted_screenshot | whitehall-admin dev gov uk_government_admin_worldwide_organisations_british-consulate-general-milan_corporate_information_pages|

Screenshots of adding attachments for the body: 
whitehall-admin dev gov uk_government_admin_worldwide_organisations_british-consulate-general-milan
whitehall-admin dev gov uk_government_admin_worldwide_organisations_british-consulate-general-milan_attachments

### DIFF
--- a/db/migrate/20230515141021_add_summary_and_body_to_worldwide_organisation_translations.rb
+++ b/db/migrate/20230515141021_add_summary_and_body_to_worldwide_organisation_translations.rb
@@ -5,6 +5,7 @@ class AddSummaryAndBodyToWorldwideOrganisationTranslations < ActiveRecord::Migra
   class CorporateInformationPage < Edition
     ABOUT_US_TYPE_ID = 20
     scope :published, -> { where(state: "published") }
+    scope :draft, -> { where(state: "draft") }
     translates :summary, :body
   end
 
@@ -18,6 +19,12 @@ class AddSummaryAndBodyToWorldwideOrganisationTranslations < ActiveRecord::Migra
         corporate_information_page_type_id: CorporateInformationPage::ABOUT_US_TYPE_ID,
       )
     end
+
+    def draft_about_us
+      @draft_about_us ||= corporate_information_pages.draft.find_by(
+        corporate_information_page_type_id: CorporateInformationPage::ABOUT_US_TYPE_ID,
+      )
+    end
   end
 
   def up
@@ -26,17 +33,43 @@ class AddSummaryAndBodyToWorldwideOrganisationTranslations < ActiveRecord::Migra
       t.column :body, :text, size: :medium
     end
 
+    number_of_worldwide_organisations = 0
+    number_of_worldwide_organisations_without_about_page = 0
+    number_of_worldwide_organisations_with_draft_content = 0
+    number_of_worldwide_organisation_translations_created = 0
+
     WorldwideOrganisation.all.each do |worldwide_organisation|
-      next if worldwide_organisation.about_us.nil?
+      write "Copying content from WorldOrganisation: #{worldwide_organisation.slug}"
+      number_of_worldwide_organisations += 1
+
+      if worldwide_organisation.about_us.nil?
+        write "  WorldOrganisation #{worldwide_organisation.slug} has no about page"
+        number_of_worldwide_organisations_without_about_page += 1
+        next
+      end
+
+      if worldwide_organisation.draft_about_us.present?
+        write "  Draft content for WorldOrganisation #{worldwide_organisation.slug} exists but will not be copied"
+        number_of_worldwide_organisations_with_draft_content += 1
+      end
 
       worldwide_organisation.about_us.translations.each do |source_translation|
+        write "  Processing #{source_translation.locale} translation from about page"
         target_translation = worldwide_organisation.translations.find_by(locale: source_translation.locale)
         if target_translation.nil?
+          write "    Creating #{source_translation.locale} translation for WorldwideOrganisation; name attribute will not be set so rendered page will fallback to english name"
+          number_of_worldwide_organisation_translations_created += 1
           target_translation = worldwide_organisation.translations.create!(locale: source_translation.locale)
         end
+        write "    Copying content from about page to WorldwideOrganisation"
         target_translation.update!(summary: source_translation.summary, body: source_translation.body)
       end
     end
+
+    write "Number of WorldwideOrganisations: #{number_of_worldwide_organisations}"
+    write "Number of WorldwideOrganisations without an about page: #{number_of_worldwide_organisations_without_about_page}"
+    write "Number of WorldwideOrganisations with draft content: #{number_of_worldwide_organisations_with_draft_content}"
+    write "Number of WorldwideOrganisation translations created: #{number_of_worldwide_organisation_translations_created}"
   end
 
   def down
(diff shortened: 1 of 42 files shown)


db/migrate/20230515141021_add_summary_and_body_to_worldwide_organisation_translations.rb
```diff
@@ -0,0 +1,48 @@
...
+      t.column :body, :text
+    end
+
+    WorldwideOrganisation.all.each do |worldwide_organisation|
+      next if worldwide_organisation.about_us.nil?
+
+      worldwide_organisation.about_us.translations.each do |source_translation|
+        target_translation = worldwide_organisation.translations.find_by(locale: source_translation.locale)
+        if target_translation.nil?
+          target_translation = worldwide_organisation.translations.create!(locale: source_translation.locale)
```
BeckaL, line 35: I guess this has the impact of publishing translations for worldwide organisations that did not previously have them. I had a look at a few translations in integration that fell into this camp, and they seem to have been created this year in March (i.e. they're not so old that nobody cares about them, but old enough that I'm surprised they haven't been published or complained about). I'm wondering if we should add these to the list of things we want to check with FDCO that they're OK with being published before we merge this.

floehopper: We've added this to our list of things to ask FCDO about. Marking as resolved.

chrisroos:
I've just rebased on master to resolve a conflict and force pushed.

db/pending_migration_cleanups/20230516160033_delete_worldwide_organisation_about_pages.rb
```diff
@@ -0,0 +1,27 @@
...
+    has_many :corporate_information_pages, through: :edition_worldwide_organisations, source: :edition, class_name: "::CorporateInformationPage"
+
+    def about_us_page_editions
+      corporate_information_pages.where(corporate_information_page_type_id: CorporateInformationPage::ABOUT_US_TYPE_ID)
+    end
+  end
+
+  def up
+    WorldwideOrganisation.all.each do |worldwide_organisation|
+      worldwide_organisation.about_us_page_editions.destroy_all
```
ollietreend, line 20: Will this delete attachments on old editions, too?

I've only glanced, but it looks like it might be necessary to call `delete_all_attachments`:

https://github.com/alphagov/whitehall/blob/ab98fb14d01fddd38abf79a430fb68dbcf365de8/app/models/attachable.rb#L178-L180

because the `has_many` association doesn't have `dependent: :destroy`

ollietreend (commented):
Just one small question from me.

But overall this looks very thorough 👍🏻 and I'd be happy to approve it if you need it.

db/pending_migration_cleanups/20230516160033_delete_worldwide_organisation_about_pages.rb
floehopper: Hmm. Good question. Since we've moved (rather than copied) the attachments from each of the about page published edition to the relevant WO, I don't think we want to delete those attachments. However, we *do* want to delete the attachments associated with about page non-published editions which I had overlooked. Perhaps we need two steps, one where we delete the about page published editions (but not their attachments) and one where we delete the about page non-published editions *and* their attachments...?

@chrisroos Are you happy to look into this? Let me know if you want to pair on it.

ollietreend: > two steps, one where we delete the about page published editions (but not their attachments) and one where we delete the about page non-published editions and their attachments

Won't the 'current' published Edition technically have _zero_ attachments? Since you've already moved them (rather than copied). So calling `delete_all_attachments` should technically do nothing, since they're no longer associated to that Edition? 🤔 Or maybe I'm missing something.

floehopper: Ah, yes - you're right - thanks!

chrisroos: Hi @ollietreend and @floehopper - I believe I've addressed this in https://github.com/alphagov/whitehall/pull/7674/commits/c216f8f1945c923052bd7185c5e05b26121b520f. I'm fairly confident it's working as expected but would really appreciate a second or third pair of eyes on it given the destructive nature of the migration.

It takes about 15 mins to run on my machine. Is that an acceptable amount of time for a standard Rails migration to run? Or should we make this a data migration when we're happy to run it?

chrisroos:
I've converted this back to a draft while I make some changes:

- Prevent users from creating new "About us" Corporate Information Pages for Worldwide Organisations.
- Investigate whether the migration to move attachments from Corporate Information Pages to Worldwide Organisations is working as expected - in my development db, the `attachable_type` ends up as `MoveAttachmentsFromWorldwideOrganisationAboutPagesToWorldwideOrganisations::WorldwideOrganisation` instead of `WorldwideOrganisation`.

chrisroos:
We've decided not to go with this approach for now so I'm closing this PR. The short version is that we're uncomfortable removing functionality (notably 2i of the About us Corporate Information Pages) that FCDO rely on. There's more detail in https://trello.com/c/EusAiUcv/579-move-the-body-data-to-worldwide-organisations#comment-649300642198d8a985d8e49a.

As I understand it, part of the motivation for this PR was to avoid exceptions that we currently see when trying to publish About us Corporate Information Pages for worldwide organisations - because their base path conflicted with that of the worldwide organisation itself. We're going to tackle that specific problem in #7789.

PR #7675, opened 2023-05-18
Remove old migrations

chrislo:
It seems unlikely that we'd need to re-run migrations older than 2023, so this PR removes the database and data migrations older than that date.

The advantage of doing this is it makes things like grepping the code and fixing linting errors when rules change easier.

There is precedence for doing this in, for example 5b873a6bd6.


ollietreend (approved):
Thanks for opening this @chrislo.

I've just had a chat with @kevindew about it, and he pointed out that there isn't an established convention within GOV.UK for deleting old migrations.

However, given the fact that the precedent has been set within this particular codebase (as you pointed out - 5b873a6), I see no strong reason to keep these old migrations lying around. As you've said, they're unlikely to be useful at this point.

So, this looks good to me 👍🏻

chrislo:
Great, thanks @ollietreend and @kevindew. I feel like the costs outweigh the benefits (given that it's fairly easy to pull deleted things out of the git history), so happy to merge.

PR #7676, opened 2023-05-18
Update behaviour of select-with-search component

davidgisbey:
## Description

Based of some feedback we got on our documents search page we're making some changes to the select-with-search components behaviour.

This does the following:

1. Removes the "Press to select" text from options when highlighted
2. Updates fuzzy search to match sequential characters at any part of the string
3. Removes the padding from the right of options where "Press to select" was located. It's now 5px instead of 100px
4. Apply 'word-break: break-word;' so that results are more readable
5. Increase search result limit to 100

All of these changes bring the component more in line with the previous behaviour that Select2 exhibits.

The updates to fuzzy search are using this documentation https://fusejs.io/api/options.html & the update 1 & 6 can be found here https://github.com/Choices-js/Choices#itemselecttext & https://github.com/Choices-js/Choices#searchresultlimit-4

### removal of "press to select", fuzzy search update, reduced padding & word break

|Bootstrap|Current implementation|New implementation|
|-----------|-----------|------------|
|image|image|image|

### Increase in result limit

The Bootstrap and new implementation are scrollable

|Bootstrap|Current implementation|New implementation|
|-----------|-----------|------------|
|image|image|image|

## Trello cards

https://trello.com/c/6BE62bVa/164-bug-search-doesnt-display-results-for-shot-versions-of-departments
https://trello.com/c/f8k0oLhf/165-words-cut-off-on-pick-lists

### DIFF
--- a/app/assets/javascripts/components/select-with-search.js
+++ b/app/assets/javascripts/components/select-with-search.js
@@ -20,7 +20,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.choices = new window.Choices(this.select, {
       allowHTML: true,
       searchPlaceholderValue: 'Search in list',
-      shouldSort: false // show options and groups in the order they were given
+      shouldSort: false, // show options and groups in the order they were given
+      itemSelectText: '',
+      searchResultLimit: 100,
+      // https://fusejs.io/api/options.html
+      fuseOptions: {
+        ignoreLocation: true, // matches any part of the string
+        threshold: 0 // only matches when characters are sequential
+      }
     })
 
     this.module.choices = this.choices
--- a/app/assets/stylesheets/components/_select-with-search.scss
+++ b/app/assets/stylesheets/components/_select-with-search.scss
@@ -212,7 +212,6 @@
     box-shadow: 0 2px 0 #929191;
     line-height: 1;
     color: #0b0c0c;
-    word-break: break-all;
     box-sizing: border-box;
 
     &[data-deletable] {
@@ -246,7 +245,6 @@
     border-top-width: 0;
     top: 100%;
     overflow: hidden;
-    word-break: break-all;
     will-change: visibility;
 
     &.is-active {
@@ -295,7 +293,7 @@
   @media (min-width: 640px) {
     .choices__list--dropdown .choices__item--selectable,
     .choices__list[aria-expanded] .choices__item--selectable {
-      padding-right: 100px;
+      padding-right: 5px;
     }
 
     .choices__list--dropdown .choices__item--selectable:after,


app/assets/javascripts/components/select-with-search.js
```diff
@@ -20,7 +20,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.choices = new window.Choices(this.select, {
       allowHTML: true,
       searchPlaceholderValue: 'Search in list',
-      shouldSort: false // show options and groups in the order they were given
+      shouldSort: false, // show options and groups in the order they were given
+      itemSelectText: '',
+      searchResultLimit: 100,
```
davidgisbey, line 25: This is a completely arbitrary number. It could just as easily be 20 or 50. Happy to go with whatever the consensus is.

app/assets/stylesheets/components/_select-with-search.scss
```diff
@@ -295,7 +295,8 @@
   @media (min-width: 640px) {
     .choices__list--dropdown .choices__item--selectable,
     .choices__list[aria-expanded] .choices__item--selectable {
-      padding-right: 100px;
+      padding-right: 5px;
+      word-break: break-word;
```
ollietreend, line 299: The CSS rule `word-break: break-word;` [is deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), so we shouldn't really be using it.

However, I think probably what we're trying to achieve here is to restore the browser's default line wrap behaviour. So instead of setting it here, I think we should try removing the following rules:

https://github.com/alphagov/whitehall/blob/2bb0c25121226c2aea98d5087a5288182ca10163/app/assets/stylesheets/components/_select-with-search.scss#L215

https://github.com/alphagov/whitehall/blob/2bb0c25121226c2aea98d5087a5288182ca10163/app/assets/stylesheets/components/_select-with-search.scss#L249

Which should mean we don't need this CSS rule at all.

app/assets/javascripts/components/select-with-search.js
ollietreend: 100 seems a perfectly reasonable choice to me! 👍🏻 

Ideally we wouldn't go for _much more_ than 100, though, because we know Choices.js can get a little janky when rendering 100's of results at once. For example, I expect it'll feel a bit jankier if rendering 500 items.

app/assets/stylesheets/components/_select-with-search.scss
davidgisbey: Good spot! That fixed it. Thanks for the review 👍

ollietreend (approved):
Looks good to me! ✨🚀

Thanks for doing this.

PR #7679, opened 2023-05-18
Make migration to add content ID to Worldwide Offices more robust

jkempster34:
This migration seems to have failed part way through when running on the live
environments.

This means that it has not been added to the schema migrations table as a
successful migration, and is trying to run again. Because the first part of the
migration (adding the column) was successful, we now see an error
`Mysql2::Error: Duplicate column name 'content_id'` on subsequent attempts.

This makes the migration conditional on the presence of the column, and moves
the part that generates the UUIDs into a separate data migration so that we can
run it in a more controlled manner.

---


chrisroos (approved):
Looks good to me, @jkempster34 :+1:

PR #7684, opened 2023-05-19
Present `access_and_opening_times` as `nil` when they are blank

jkempster34:
### Present `access_and_opening_times` as `nil` when they are blank
Because we convert access and opening times to Govspeak, they get wrapped in a
tag when empty (e.g., `<div class=\"govspeak\">\n</div>`).

It will be easier for us to deal with these in the rendering app if they are
`nil` instead.

> I know this is unrelated sorry👇 , but I just wanted to tidy up as I go

### Remove unused view partial
This was used by the world locations show page (removed in #6961), but is now
not referenced anywhere , so it can be safely removed.

---


app/views/worldwide_organisations/_worldwide_organisation.html.erb
```diff
@@ -1,29 +0,0 @@
-<%= content_tag_for(:div, worldwide_organisation) do %>
-  <% worldwide_organisation.sponsoring_organisations.each do |organisation| %>
```
BeckaL, line 2: Thanks for tidying up! Would it be possible to put reference to the commit that removed the last ref to this partial in the commit message? I was just having a trawl through to see what had happened here but wondered if you'd already done this digging

PR #7687, opened 2023-05-19
155 topical events details

davidtrussler:
[Trello](https://trello.com/c/eSW5NOkF/155-details-page)

This PR ports the Topical Events Details page to the Design System. 

|Bootstrap|Design System|
|-|-|
| Screenshot 2023-05-22 at 10 39 51 | Screenshot 2023-05-22 at 15 24 02 |

### DIFF
--- a/app/views/admin/topical_events/show.html.erb
+++ b/app/views/admin/topical_events/show.html.erb
@@ -1,51 +1,43 @@
-<% page_title @topical_event.name %>
+<% content_for :page_title, @topical_event.name %>
+<% content_for :title, @topical_event.name %>
+<% content_for :title_margin_bottom, 4 %>
 
-<% content_for :current_tab do %>
-  <table class="<%= model_name %> table table-bordered add-top-margin">
-    <tr>
-      <th>Summary</th>
-      <td class="summary"><%= @topical_event.summary %></td>
-    </tr>
+<p class="govuk-body"><%= view_on_website_link_for @topical_event, class: "govuk-link" %></p>
 
-    <tr>
-      <th>Description</th>
-      <td class="description"><%= @topical_event.description %></td>
-    </tr>
+<div class="govuk-!-margin-bottom-8">
+  <%= render "components/secondary_navigation", {
+    aria_label: "Topical Events tabs",
+    items: [
+      {
+        label: "Details",
+        href: [:admin, @topical_event],
+        current: true,
+      },
+      {
+        label: "About page",
+        href: [:admin, @topical_event, :topical_event_about_pages]
+      },
+      {
+        label: "Features",
+        href: [:admin, @topical_event, :topical_event_featurings]
+      }
+    ]
+  } %>
+</div>
 
-    <tr>
-      <th>Details</th>
-      <td class="details"><%= topical_event_contents_breakdown(@topical_event) %></td>
-    </tr>
-
-    <% if @topical_event.is_a?(TopicalEvent) %>
-      <tr>
-        <th>Duration</th>
-        <td class="duration">
-          <% if @topical_event.start_date && @topical_event.end_date %>
-            <%= @topical_event.start_date %> to <%= @topical_event.end_date %>
-          <% end %>
-        </td>
-      </tr>
-    <% end %>
-
-    <% if @topical_event.respond_to?(:featured_links) %>
-      <tr>
-        <th>Featured links</th>
-        <td class="featured-links">
-          <% if @topical_event.featured_links.any? %>
-            <%= render 'shared/featured_links', links: @topical_event.featured_links %>
-          <% else %>
-            <em>None</em>
-          <% end %>
-        </td>
-      </tr>
-    <% end %>
-  </table>
-
-
-  <div class="form-actions">
-    <%= link_to "Edit", [:edit, :admin, @topical_event], class: "btn btn-lg btn-primary"%>
-  </div>
-<% end %>
-
-<%= render partial: 'tab_wrapper', locals: { model: @topical_event } %>
+<%= render "govuk_publishing_components/components/summary_list", {
+  title: "Details",
+  items: [
+    { field: "Summary", value: @topical_event.summary },
+    { field: "Description", value: @topical_event.description },
+    { field: "Details", value: topical_event_contents_breakdown(@topical_event) },
+    { field: "Duration",
+      value: ("#{@topical_event.start_date} to #{@topical_event.end_date}" if @topical_event.start_date && @topical_event.end_date)
+    }
+  ],
+  edit: {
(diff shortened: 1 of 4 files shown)


nnagewad:
Looks good from the design front 👍🏽

app/views/admin/topical_events/show.html.erb
```diff
@@ -1,51 +1,53 @@
...
+  href: [:edit, :admin, @topical_event],
+  secondary_solid: true
+} %>
+
+<%= render "govuk_publishing_components/components/summary_list", {
+  items: [
+    { field: "Summary", value: @topical_event.summary },
+    { field: "Description", value: @topical_event.description },
+    { field: "Details", value: topical_event_contents_breakdown(@topical_event) },
+    (if @topical_event.is_a?(TopicalEvent)
```
davidgisbey, line 39: I know this was pre-existing code, but do we know if this is ever not a TopicalEvent? It feels like it should be in this controller?

app/views/admin/topical_events/show.html.erb
```diff
@@ -1,51 +1,53 @@
...
-<%= render partial: 'tab_wrapper', locals: { model: @topical_event } %>
+<% content_for :page_title, @topical_event.name %>
+<% content_for :title, @topical_event.name %>
+<% content_for :title_margin_bottom, 4 %>
+
+<p class="govuk-body"><%= view_on_website_link_for @topical_event, class: "govuk-link" %></p>
+
+<div class="govuk-!-margin-bottom-8">
+  <%= render "components/secondary_navigation", {
+    aria_label: "Get involved tabs",
```
davidgisbey, line 9: I think this should be `Topical events tabs`

app/views/admin/topical_events/show.html.erb
```diff
@@ -1,51 +1,53 @@
...
+        value:
+          if @topical_event.start_date && @topical_event.end_date
+            "#{@topical_event.start_date} to #{@topical_event.end_date}"
+          end
+      }
+    end),
+    *(
+      { field: "Featured links",
+        value: @topical_event.featured_links.any? ? (render 'shared/featured_links', links: @topical_event.featured_links) : "None"
+      } if @topical_event.respond_to?(:featured_links)
```
davidgisbey, line 50: I've just tracked this down https://github.com/alphagov/whitehall/commit/1accacf45c48d33c513b4a3bca1e6c81c8442c40

It looks like a lot of this is a hangover from a bunch of deleted code.

This will always return false as topical events don't have featured links.

app/views/admin/topical_events/show.html.erb
```diff
@@ -1,51 +1,53 @@
...
-    <%= link_to "Edit", [:edit, :admin, @topical_event], class: "btn btn-lg btn-primary"%>
-  </div>
-<% end %>
-
-<%= render partial: 'tab_wrapper', locals: { model: @topical_event } %>
+<% content_for :page_title, @topical_event.name %>
+<% content_for :title, @topical_event.name %>
+<% content_for :title_margin_bottom, 4 %>
+
+<p class="govuk-body"><%= view_on_website_link_for @topical_event, class: "govuk-link" %></p>
```
davidgisbey, line 5: should this all be in 2/3 width?  probs one for Nik.

image

app/views/admin/topical_events/show.html.erb
```diff
@@ -1,51 +1,53 @@
...
+  } %>
+</div>
+
+<%= render "govuk_publishing_components/components/button", {
+  text: "Edit details",
+  href: [:edit, :admin, @topical_event],
+  secondary_solid: true
+} %>
+
+<%= render "govuk_publishing_components/components/summary_list", {
```
davidgisbey, line 34: It's a real shame that you can't pass in custom label text for the summary list component or we could do this which seems a lot cleaner

image

app/views/admin/topical_events/show.html.erb
```diff
@@ -1,51 +1,53 @@
...
+  items: [
+    { field: "Summary", value: @topical_event.summary },
+    { field: "Description", value: @topical_event.description },
+    { field: "Details", value: topical_event_contents_breakdown(@topical_event) },
+    (if @topical_event.is_a?(TopicalEvent)
+      { field: "Duration",
```
davidgisbey, line 40: maybe?

``` 
{ field: "Duration", value: "#{@topical_event.start_date} to #{@topical_event.end_date}" if  @topical_event.start_date && @topical_event.end_date }
```

davidgisbey: So the whole thing would be 
```
<%= render "govuk_publishing_components/components/summary_list", {
  title: "Details",
  edit: {
    href: [:edit, :admin, @topical_event],
    text: "Edit"
  },
  items: [
    {
      field: "Summary",
      value: @topical_event.summary
    },
    {
      field: "Description",
      value: @topical_event.description
    },
    {
      field: "Details",
      value: topical_event_contents_breakdown(@topical_event)
    },
    {
      field: "Duration",
      value: ("#{@topical_event.start_date} to #{@topical_event.end_date}" if @topical_event.start_date && @topical_event.end_date),
    },
  ]
} %>
```

app/views/admin/topical_events/show.html.erb
davidtrussler: Have removed the conditional clause here.

app/views/admin/topical_events/show.html.erb
davidtrussler: Removed featured links section.

app/views/admin/topical_events/show.html.erb
davidtrussler: Yes, that's better, though I've put the if clause to the whole item - I suspect we would not wan the "Duration" field to show at all if there are no dates. Not entirely sure about that though TBH.

davidtrussler: Actually the tests don't like that - seem to expect a date field so restored that and just applied the if clause to the field's value.

davidgisbey: cool we can always use a splat to remove it conditionally like this is we want to do that down the line

```
*([{
      field: "Duration",
      value: "#{@topical_event.start_date} to #{@topical_event.end_date}",
}]  if @topical_event.start_date && @topical_event.end_date),
```

app/views/admin/topical_events/show.html.erb
davidtrussler: Have looked over the width issue with Nik and decided to stay with full width here. The narrower width starts to look a bit awkward at narrow, non-mobile, widths.

davidgisbey (approved):
lgtm!

PR #7691, opened 2023-05-22
Add Offset Links Tab to the Topical Events Featurings Index Page in the GOV.UK Design System

davidgisbey:
## Description

While adding the features page for World locations news i hadn't realised that there is also a features page for topical events. 
Unfortunately there's fairly subtle differences in data modelling so i'm going to have to clean up and reuse a bunch of the partials used for World news locations and Orgs feature pages.

Unlike World location news and orgs, Topical events use TopicalEventFeaturings to feature documents. To keep the PRs reviewable i'm going to open 1 PR for each of the tabs.

This PR tackles the easiest of the tabs, the Non-gov links tab (offsite links).

The currently featured and document tab will be added in the next two prs.

## Screenshots

image

image

## Trello card

https://trello.com/c/EgRz8GpE/151-featured-page


app/helpers/admin/topical_event_featurings_helper.rb
```diff
@@ -0,0 +1,9 @@
+module Admin::TopicalEventFeaturingsHelper
```
chriswhitehouse-tw, line 1: Took me a while to piece together what this helper function was doing. Just wondering if there would be a way to do this more at the model layer rather than the view layer?

chriswhitehouse-tw (commented):
Great, looks really good @davidgisbey. A few comments, which are probably nice to have rather than essential.

Only other observation, which is probably more for tech debt, is that when creating of editing an offsite link, the validation does not handle adding whitespace before URL gracefully. (i.e put a space before the url, causes  Addressable::URI::InvalidURIError)

app/helpers/admin/topical_event_featurings_helper.rb
davidgisbey: great shout. Will sort this

davidgisbey:
Regarding your comment about whitespace, i agree it's bad! There's various gems that strip whitespace from a users input, or we could add our own method that does it down the line. We should def sort this though. We can adda card to the ideas/tech debt column

app/views/admin/offsite_links/confirm_destroy.html.erb
```diff
@@ -13,8 +13,11 @@
           text: "Delete",
           destructive: true,
         } %>
-
-        <%= link_to("Cancel", polymorphic_url([:features, :admin, @parent]) + "#non_govuk_government_links_tab", class: "govuk-link govuk-link--no-visited-state") %>
+        <% if @parent.is_a?(TopicalEvent) %>
```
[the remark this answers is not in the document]
davidgisbey: Have added this in 👍

chriswhitehouse-tw (approved):
Looks good to me David.

I guess in an ideal world we would have the cancel links from edit and new link pages navigate back to the anchor as well, but understand this might be a bit of a faff to implement. So looks good to me.

davidgisbey:
ah cool i hadn't realised the cancel link wasn't working there. I'll sort that now 👍

PR #7714, opened 2023-05-24
Add currently featured tab to the Topical Event Featurings index page

davidgisbey:
## Description 

This follows on from https://github.com/alphagov/whitehall/pull/7691 which added the Non-GOV.UK government links tab.

This adds the currently featured tab and the reorder & confirm delete featurings pages.

I've had to do a fairly chunky refactor here to get the CurrentlyFeaturedTabComponent to work for  features and topical event featurings. 

I've tried to keep the commits sensible so it's probs worth reviewing per commit 

### index page currently featured tab

image

### Confirm delete 

image

### Reorder 

image

## Trello card

https://trello.com/c/EgRz8GpE/151-topical-event-featuring-index-page
 

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #7718, opened 2023-05-25
Add Documents tab to the Featurings Index page 

davidgisbey:
## Description

This follows on from:

https://github.com/alphagov/whitehall/pull/7691 - added the offsite links tab 
https://github.com/alphagov/whitehall/pull/7714 - added the recently added tab

This is the final PR for the featurings index page and adds the documents tab. This has the exact same design as the documents tab for the features page, so i've done a few things:

1. Refactor the featurable_editions partial so that it works with featurings and features.
2. moved shared partials into the shared folder
3. updated the controller to handle 15 editions per page 
4. added some feature specs

## Screenhots

image

## Trello card 

https://trello.com/c/EgRz8GpE/151-featured-page


app/views/admin/shared/_search_results.html.erb
```diff
@@ -0,0 +1,45 @@
+<p class="govuk-heading-s govuk-!-margin-bottom-3"><%= pluralize(number_with_delimiter(paginator.total_count), "document") %></p>
```
chriswhitehouse-tw, line 1: Just wondering if it is strictly necessary to move this into a shared directory as its only the featurable_editions component that calls this partial?

chriswhitehouse-tw (approved):
Great, thanks @davidgisbey. This good.

Apart form comment about the necessity of the partial being shared, it looks good to me.

app/views/admin/shared/_search_results.html.erb
davidgisbey: Yea i'm not sure exactly where this would live as it's called from multiple views by multiple controllers. Perhaps  `featurable_editions/search_results` 🤔

PR #7722, opened 2023-05-25
Update people show page to use summary list component

davidgisbey:
## Description

This uses the summary list component rather than edit and delete buttons and the biography.

It adds rows for the rt hon prefix, title, forename, surname, image and biography and only renders the row if a value is present.

It also adds in the persons title as a lead paragraph between the title and link to view them on GOV.UK

### Before

image

When deleteable

image

### After

image

image

I've updated the caption to "People"

image

### DIFF
--- a/app/assets/stylesheets/admin/_overrides.scss
+++ b/app/assets/stylesheets/admin/_overrides.scss
@@ -9,3 +9,13 @@
 code {
   color: govuk-colour("red");
 }
+
+.gem-c-summary-list--with-image {
+  .govuk-summary-list__key {
+    vertical-align: top;
+  }
+
+  img {
+    display: block;
+  }
+}
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -115,6 +115,10 @@ class Person < ApplicationRecord
     end
   end
 
+  def current_role_appointments_title
+    current_role_appointments.collect(&:role_name).to_sentence
+  end
+
 private
 
   def name_as_words(*elements)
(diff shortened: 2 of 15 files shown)


nnagewad:
When you're back, need to add the screenshot of the changes

nnagewad:
LGTM on the design front 👍🏽

nnagewad:
> Carlos asked us to use the name of the tab for each context block. If we want to change that i think we should do it all for all sections so it's consistent.
> 
> Let me have a look at how many places we're doing that in.

Right on.

Sorry to be a bother. Was re-reading how captions work, and they kinda help set the stage that the current page is a child to something bigger: https://design-system.service.gov.uk/styles/typography/#headings-with-captions

davidgisbey:
All good i thought stuff had been released with it the other way but i think we're all good. We can sort the other stuff that's been done the other way as part of that card you've written up 👍

davidtrussler:
This looks good to me. The only small concern I have is whether an image belongs semantically in a data list and what value we would expect that to impart to assistive technology, like maybe to say that an image is present - the actual content of the image is less important since it is self-evidently an image of the person. This may be getting a little bit pedantic though and at this stage may be something to come back to later.

app/assets/stylesheets/admin/_overrides.scss
```diff
@@ -9,3 +9,9 @@
 code {
   color: govuk-colour("red");
 }
+
+.gem-c-summary-list--with-image {
+  .govuk-summary-list__key {
+    vertical-align: top;
+  }
```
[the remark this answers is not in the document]
davidgisbey: have pushed up the changes 👍

app/assets/stylesheets/admin/_overrides.scss
```diff
@@ -9,3 +9,13 @@
...
+
+  image {
```
davidtrussler, line 18: Sorry - my fault - should be `img` 🤦

davidgisbey: sorry i'm on autopilot!

PR #7724, opened 2023-05-26
Remove the access and opening times table

jkempster34:
https://trello.com/c/lA84xqw5

We added a “pending“ migration to remove the `access_and_opening_times` table as the data has been migrated to fields on the Worldwide Organisation and Worldwide offices (#7611)

Now that we are confident that the change is working as expected, we can perform the pending migration.

---


brucebolt (commented):
I think the migration's name (i.e. the timestamp) will need updating, as it is older than the latest migration, meaning Rails won't run the migration automatically (you should expect to see the `version` update in `db/schema.rb`).

jkempster34:
> > I think the migration's name (i.e. the timestamp) will need updating, as it is older than the latest migration, meaning Rails won't run the migration automatically (you should expect to see the `version` update in `db/schema.rb`).
> 
> Rails maintains a table of migrations that have been performed in the table associated with `ActiveRecord::SchemaMigration.all`.
> 
> We can test this if you would be happy to pull down this branch, Rails should raise a `PendingMigrationError` (I think)
> 
> I also believe it's best practice to select the later timestamp when resolving conflicts in the schema (even if it's not the migration you added)

Link explaining it here https://stackoverflow.com/a/36608615

PR #7732, opened 2023-05-30
Create call for evidence

beccapearce:
https://trello.com/c/Bx2zLsHM/1312-create-the-basic-call-for-evidence-document-type-in-whitehall

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -269,6 +269,21 @@ private
             ],
           },
         ],
+        call_for_evidence_participation_attributes: [
+          :id,
+          :link_url,
+          :email,
+          :postal_address,
+          {
+            call_for_evidence_response_form_attributes: [
+              :id,
+              :title,
+              :_destroy,
+              :attachment_action,
+              { call_for_evidence_response_form_data_attributes: %i[id file file_cache] },
+            ],
+          },
+        ],
         nation_inapplicabilities_attributes: %i[id nation_id alternative_url excluded],
         fatality_notice_casualties_attributes: %i[id personal_details _destroy],
       },
--- a/db/migrate/20230617100117_add_call_for_evidence_response_schemas.rb
+++ b/db/migrate/20230617100117_add_call_for_evidence_response_schemas.rb
@@ -1,0 +1,34 @@
+class AddCallForEvidenceResponseSchemas < ActiveRecord::Migration[7.0]
+  def up
+    create_table "call_for_evidence_response_form_data", id: :integer, charset: "utf8mb3", force: :cascade do |t|
+      t.string "carrierwave_file"
+      t.datetime "created_at", precision: nil
+      t.datetime "updated_at", precision: nil
+    end
+
+    create_table "call_for_evidence_response_forms", id: :integer, charset: "utf8mb3", force: :cascade do |t|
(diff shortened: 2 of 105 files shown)


app/models/call_for_evidence_response.rb
```diff
@@ -0,0 +1,53 @@
+class CallForEvidenceResponse < ApplicationRecord
```
ryanb-gds, line 1: I think this model should not be part of this PR - to be added in the outcome PR

ryanb-gds (changes requested):
Looking pretty good to me. Suggest removing the `CallForEvidenceResponse` model from the PR and adding in the "Outcome" PR. Also might be nice if the rename of `Response` to `ConsultationResponse` was its own PR, would reduced the difficulty of reviewing this PR a little.

And obvs still needs the feature flag to be applied!

app/controllers/admin/editions_controller.rb
```diff
@@ -269,6 +269,21 @@ def permitted_edition_attributes
...
+          {
+            call_for_evidence_response_form_attributes: [
+              :id,
+              :title,
+              :_destroy,
+              :attachment_action,
+              { call_for_evidence_response_form_data_attributes: %i[id file file_cache] },
+            ],
+          },
+        ],
```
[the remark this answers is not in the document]
minhngocd: I've added a trello card to track this in the backlog

db/migrate/20230617100117_add_call_for_evidence_response_schemas.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    create_table "call_for_evidence_participations", id: :integer, charset: "utf8mb3", force: :cascade do |t|
+      t.integer "edition_id", foreign_key: true
+      t.string "link_url"
+      t.datetime "created_at", precision: nil
+      t.datetime "updated_at", precision: nil
+      t.string "email"
+      t.integer "call_for_evidence_response_form_id"
+      t.text "postal_address"
+      t.index %w[call_for_evidence_response_form_id], name: "index_cons_participations_on_cons_response_form_id"
```
ryanb-gds, line 24: I don't think this index name is right? "cons" should maybe be "cfes"? What's a bit odd is I think if there's an index with the same name on the consultation participation table the migration should fail, but tests suggest that isn't happening 🤔

db/schema.rb
```diff
@@ -59,6 +59,31 @@
...
 
+  create_table "call_for_evidence_participations", id: :integer, charset: "utf8mb3", force: :cascade do |t|
+    t.integer "edition_id"
+    t.string "link_url"
+    t.datetime "created_at", precision: nil
+    t.datetime "updated_at", precision: nil
+    t.string "email"
+    t.integer "call_for_evidence_response_form_id"
+    t.text "postal_address"
+    t.index ["call_for_evidence_response_form_id"], name: "index_cons_participations_on_cons_response_form_id"
```
ryanb-gds, line 70: Same problem as above - "cons" in index name

ryanb-gds (commented):
Essentially ready to merge but I think it's worth checking that potential duplicate index name?

db/schema.rb
minhngocd: fixed

ryanb-gds (approved):

PR #7736, opened 2023-05-30
Move 1.8 pages into the next release

davidgisbey:
## Description

I did the following to prepare this release

1. Moved pages going out in 1.8 into next release
2. Add return true if next_release override to the preview_design_system? method in the base controller
3. run tests and fix/duplicate non-legacy tests to both implementations have full coverage
4. extend override to include return true if next_release && !Rails.env.test? so override only applies to prod

When i overrode the permission and ran the tests, only legacy tests for the Bootstrap implementation failed so we can be confident we've got full coverage.

## Trello card 

https://trello.com/c/YIwuhpye/223-prepare-release-18

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -42,6 +42,9 @@ class Admin::BaseController < ApplicationController
   end
 
   def preview_design_system?(next_release: false)
+    # Temporarily force this flag to 'on' for users in production, regardless of their permissions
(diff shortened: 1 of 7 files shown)


chriswhitehouse-tw (approved):
Great, looks good to me @davidgisbey.

I double checked that you have updated the right controllers.

And the change to force the legacy tests to run seems appropriate.

PR #7746, opened 2023-06-01
Use custom class instead of format numeric for tables with actions

davidgisbey:
## Description 

At the moment, we're using the 'format: "numeric"' option on the Table
Publishing Component to float actions to the right. We shouldn't really
be using an unrelated class for styling.

This commit adds styling for '.govuk-table--with-actions' and applies
it to tables with actions in place of using format numeric.

## Trello card

https://trello.com/c/K7PsPxlV/221-add-a-new-class-to-stylesheets-for-tables-with-actions-being-floated-right-and-apply-instead-of-using-numeric-formatting

### DIFF
--- a/app/assets/stylesheets/admin/_overrides.scss
+++ b/app/assets/stylesheets/admin/_overrides.scss
@@ -2,6 +2,12 @@
   table-layout: fixed;
 }
 
+.govuk-table--with-actions {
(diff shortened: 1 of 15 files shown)


farahTW:
Reviewed looks  good to me.

PR #7759, opened 2023-06-02
Fix ability to edit translation documents

minhngocd:
[Trello ticket link](https://trello.com/c/EfvXj4b5/1378-bug-fix-editing-featured-links-for-translations)

Previously, locale was not added to the publishing API publish command, meaning the default english document is loaded, and attempting to publish an already published document returns an error 409 unable to publish already published document. With the fix, publish command should be able to load the correct document with the previous draft created in the PUT command.

### DIFF
--- a/test/unit/lib/present_page_to_publishing_api_test.rb
+++ b/test/unit/lib/present_page_to_publishing_api_test.rb
@@ -1,0 +1,17 @@
+require "test_helper"
+
+class PresentPageToPublishingApiTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  describe "#publish" do
+    it "should update content, patch links and publish new document" do
+      I18n.with_locale(:cy) do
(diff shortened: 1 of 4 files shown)


test/unit/lib/present_page_to_publishing_api_test.rb
```diff
@@ -0,0 +1,17 @@
+require "test_helper"
+
+class PresentPageToPublishingApiTest < ActiveSupport::TestCase
```
beccapearce, line 3: We were wondering why this is in a separate file to the test/unit/present_page_to_publishing_api_test.rb tests. As they relate to the same class.

beccapearce:
Just one thing that @jyoung-gds and I were curious about is whether the new test needs to be in a separate file. Not an issue we were just wondering. Happy to approve otherwise

beccapearce (approved):
LGTM, would love to chat about the tests later too :)

PR #7768, opened 2023-06-05
Update take part pages styling

farahTW:
This PR updates take-part-pages as per suggested styling and design specifications.
It does the following:

- It removes hint text from Title and Summary input and added character count.
- It modifies space between last the field and button as 50 px.

**Note:** Image component is still same as PF for new image component requested is still not completed.
This will be done once new image component is ready to be incorporated.

**Screen shots:**
image

image

**Trello:**
https://trello.com/c/SAfAUOhr/230-update-take-part-page-styling-and-components

### DIFF
--- a/app/views/admin/take_part_pages/_form.html.erb
+++ b/app/views/admin/take_part_pages/_form.html.erb
@@ -5,31 +5,36 @@
     } %>
 
     <%= form_for take_part_page, url: [:admin, take_part_page], multipart: true do |form| %>
-      <%= render "govuk_publishing_components/components/input", {
-        label: {
-          text: "Title (required)"
+      <%= render "govuk_publishing_components/components/character_count", {
+        textarea: {
+          label: {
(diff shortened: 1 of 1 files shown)


app/views/admin/take_part_pages/_form.html.erb
```diff
@@ -62,17 +67,17 @@
...
+          label: {
+            text: "Image description (alt text)",
+          },
+          heading_size: "l",
+          name: "take_part_page[image_alt_text]",
+          id: "take_part_page_image_alt_text",
+          value: form.object.image_alt_text
+        } %>
+      </div>
       <div class="govuk-button-group">
```
davidgisbey, line 81: If it's easier i think we can just add 

```
<div class="govuk-button-group govuk-!-margin-top-8">
```

davidgisbey (approved):
lgtm!

PR #7773, opened 2023-06-06
UI improvements to role listing page

farahTW:
This PR does the UI improvements for Role listing page.
It has the following changes.

- First column items are bold.
- Removed header for: Edit and Delete.
- Removed word “translation” from Manage translation link.
- Removed “role type” column.

**Screen shots:**
image

**Trello:**
https://trello.com/c/1unFs900/234-ui-improvements-to-roles-listing-page

### DIFF
--- a/app/views/admin/roles/index.html.erb
+++ b/app/views/admin/roles/index.html.erb
@@ -2,19 +2,19 @@
 <% content_for :title, "Roles" %>
 <% content_for :title_margin_bottom, 6 %>
 
-<%= render "govuk_publishing_components/components/button", {
-  title: "Create new role",
-  text: "Create new role",
-  href: new_admin_role_path,
-  margin_bottom: 6
-} %>
-
 <div class="govuk-!-margin-bottom-6">
   <%= render "govuk_publishing_components/components/warning_text", {
     text: "Do not create ministerial roles without consulting GDS."
   } %>
 </div>
 
+<%= render "govuk_publishing_components/components/button", {
+  title: "Create new role",
+  text: "Create new role",
+  href: new_admin_role_path,
+  margin_bottom: 6
+} %>
+<div class="app-view-roles__page-list">
 <%= render "govuk_publishing_components/components/table", {
   head: [
     {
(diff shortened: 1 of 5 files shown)


app/views/admin/roles/index.html.erb
```diff
@@ -23,53 +23,52 @@
...
-    },
     {
       text: "Currently appointed"
     },
     {
       text: "Translations"
     },
     {
-      text: "Edit"
+      text: tag.span(class: "govuk-visually-hidden")
```
davidgisbey, line 33: hey, so we should still have text here so screen readers pick it up.

One thing i noticed with this is that it's two separate columns. Whereas in all other places we've got them as one "Actions" column.

I think we should update this to 

```
      text: tag.span("Actions", class: "govuk-visually-hidden")
```

and then merge the edit and delete columns. Wdyt?

app/views/admin/roles/index.html.erb
```diff
@@ -23,53 +23,52 @@
...
+            ""
+          else
+            role.current_person_name
+          end
       },
       {
         text:
           content_tag_for(:div, role) do
-            link_to(sanitize("Manage translations#{tag.span(role.name, class: "govuk-visually-hidden")}"), admin_role_translations_path(role), title: "Manage translations of #{role.name}", class: "govuk-link")
+            link_to(sanitize("Manage#{tag.span(role.name, class: "govuk-visually-hidden")}"), admin_role_translations_path(role), title: "Manage translations of #{role.name}", class: "govuk-link")
```
davidgisbey, line 58: sorry i think we should leave a space between manage and `role.name` otherwise it will read it as one word 

```
"Manage #{tag.span(role.name, class: "govuk-visually-hidden")}"
```

app/views/admin/roles/index.html.erb
```diff
@@ -23,53 +23,52 @@
...
       },
       {
         text:
-            if role.destroyable?
-                link_to(sanitize("Delete#{tag.span(role.name, class: "govuk-visually-hidden")}"), confirm_destroy_admin_role_path(role), class: "govuk-link govuk-link--no-visited-state gem-link--destructive")
-            else
-              ""
-            end
+          if role.destroyable?
+            link_to(sanitize("Delete#{tag.span(role.name, class: "govuk-visually-hidden")}"), confirm_destroy_admin_role_path(role), class: "govuk-link govuk-link--no-visited-state gem-link--destructive")
```
davidgisbey, line 67: same here regarding spaces,

app/views/admin/roles/index.html.erb
farahTW: yeah ..make sense

davidgisbey (approved):
one of the tests is failing but other than that it lgtm

PR #7777, opened 2023-06-06
Test file structure

minhngocd:
The test directory structure did not mirror application code, which was difficult to differentiate which part of the code the tests belong to. Refactoring the directory structure also highlighted 6 test files that have been duplicated since they were created in different folders within test. These tests have been merged. Still pending review of individual tests to check for duplicate test cases.

The duplicate test files that have been merged into 1 test file and require reviewing later are: 
- `attachment_test`
-  `corporate_information_page_test`
- `featured_link_test`
- `sitewide_setting_test`
- `social_media_account_test`
- `unpublishing_test`


kevindew:
Wow this looks amazing! Thanks so much, the test structure has bugged me for ages.

With Rails it seems to be a convention that tests for files in the app directory aren't nested in an app directory in the test directory (some examples in https://guides.rubyonrails.org/testing.html). So for a more idiomatic I'd suggest you do: `mv test/unit/app/* test/unit` so the app directories are at the same level as lib.

kevindew:
Thanks @minhngocd - no problem about the delay. I'll reply inline to each point. Again many thanks for working on this 🙇 


> 1. The default convention set out in Rails is everything being under `/test` or `/spec` so us introducing `/test/unit/` is already adding a layer of structure that I think isn't so different from `/test/unit/app`, which gives us the benefit of easier organisation and navigation to address following problems.

Sure, yup I understand test/unit is unconventional too. I thought it would be starting to make things more conventional. As moving things into `test/unit/*` is then not too big a step to then move them into `test/` to get even more conventional. Although it's not necessarily part of Rails doctrine it's relatively common to have lib tests standalone in the test directory. while app tests are in root (it's definitely an approach seen in a few GOV.UK apps: [example](https://github.com/alphagov/content-publisher/tree/main/spec))

> 2. I believe having app tests in a separate directory / package highlights the fact that we have a bunch of tests that are not necessarily unit tests in the unit folder (sql test, inflection test, rails config tests, locale tests) as well as separating unit tests for the lib folder

Yeah I'd agree they're not unit tests. They certainly seem to be types of integration tests.

> 3. which begs another question - should tests for lib be in unit testing? Are we over-configuring dependent "external" libs? How much should we be testing libs and should they just be covered by integration testing?

The code in lib shouldn't be external libs (perhaps you were thinking of vendor) as it's basically the home for any code that doesn't naturally fit in the regular app directories - which is often more thorough aspects of business logic as well as code that could be extracted to a library but doesn't necessarily have wider use.

> 4. need to also sort out definition between 'Integration' and 'unit' as there doesn't seem to be much difference between the test files in these 2 folders. e.g. tests in integration are testing 1 class under test, unit tests are creating things in the database. Do we have a clear definition as to which test belongs to which category?

Yeah, it'd be useful to have a definition. I bet we had one in the past (and I've found: https://github.com/alphagov/whitehall/blob/099f53e35cdd0ea63a2349be3766c98e65521ce3/docs/testing.md which doesn't reference integration) and I imagine convenience has led to it slipping. 

Having a folder called "integration" is itself a bit of a pain as they overlap as a concept with other types of integration tests like request tests (which I think we call functional tests in Whitehall) and system tests (which we do as cucumber tests). I'd wonder if we wanted to make "integration" explicitly for testing things like configuration or working directly with lower level databases in a ways that can't be tested at unit level, whereas if it's part of a feature it should be tested at cucumber or controller level. We could document approaches, this is a way we've documented how to test before: https://github.com/alphagov/content-publisher/blob/main/docs/testing-strategy.md

> I personally feel that separating out app unit test code this is a necessary stepping stone if we do want to get to a point where we have clean test structure and clean test definition separation, sort out the problems above, and only then if we do want to, move to default directory structure if it makes navigation easier. Right now merging all this together feels like obfuscating the existing complexity we have in test definitions.

Sure, maybe get a couple of other opinions about folder organisation to see what feels more natural for devs. My opinion is based around an anticipation that we want to reduce the snowflake aspect of it and gradually get it more idiomatic for Rails - but it could be that we collectively prefer something more custom? Either way - the work to move things in this PR is a huge improvement.

kevindew (approved):
Great stuff, are there any courtesies we need to give around merging this? Any team stakeholders who need to be in the loop (perhaps @ollietreend) or any PRs that'll need painful rebases?

PR #7778, opened 2023-06-07
Ensure correct layout renders for pages being ported to GOV.UK Design System

davidgisbey:
## Description

We've had an issue a few times where we've accidentally rendered atemplate built in the GOV.UK Design System with the Bootstrap layout. This is due to pages accidentally being moved into the next release.

This PR couples view rendering with the get_layout method in each controller that conditionally renders the view. If `get_layout` returns `design_system` that is passed downstream to the `preview_design_system?` method.

It then adds a shared test to all admin controllers that still need to be ported to the GOV.UK Design System.

This test checks whether the `get_layout` method is defined on the controller. If not then the test returns as we can be confident it's rendering the `admin` layout

If defined, it creates a user with the "Preview next release" permission, then iterates through each non-inherited controller
action (new create edit update etc.) and checks that get_layout returns "admin".

Once pages are moved into the next release the test will fail and it can be removed.

## Where this doesn't work

If we try to break releases into smaller sections where only some of a controllers actions are going out in a release this wouldn't work as it iterates through all the controller actions. I've manually added exceptions for actions we know are always in the design system (reorder confirm destroy etc.), but if for example a new page was meant to be in the design system in 1. controller but the edit page wasn't, it wouldn't work.

We could adapt this to take an arg which filters out some actions etc. down the line.

This shouldn't be an issue though as all our future releases involve releasing full sections so all the controller actions should port to using the design system at the same time.

## Screenshot of failure

When i moved the roles index action into the design system layout i get this error

image

## Trello card 

https://trello.com/c/0DVQgb9L/241-add-test-which-ensures-that-views-have-use-the-correct-layout

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -49,8 +49,8 @@ class Admin::BaseController < ApplicationController
   end
   helper_method :preview_design_system?
 
-  def render_design_system(design_system_view, legacy_view, next_release: false)
-    if preview_design_system?(next_release:)
+  def render_design_system(design_system_view, legacy_view)
+    if get_layout == "design_system"
       render design_system_view
     else
       render legacy_view
--- a/app/controllers/admin/cabinet_ministers_controller.rb
+++ b/app/controllers/admin/cabinet_ministers_controller.rb
@@ -8,7 +8,7 @@ class Admin::CabinetMinistersController < Admin::BaseController
     @whip_roles = MinisterialRole.includes(:translations).whip.order(:whip_ordering)
     @organisations = Organisation.ministerial_departments.excluding_govuk_status_closed.order(:ministerial_ordering)
 
-    render_design_system(:show, :legacy_show, next_release: false)
+    render_design_system(:show, :legacy_show)
   end
 
   def reorder_cabinet_minister_roles
--- a/test/support/controller_test_helpers.rb
+++ b/test/support/controller_test_helpers.rb
@@ -2,6 +2,8 @@ module ControllerTestHelpers
   extend ActiveSupport::Concern
 
   module ClassMethods
+    LAYOUT_ALWAYS_DESIGN_SYSTEM = %w[reorder update_order confirm_destroy].freeze
+
     def should_be_an_admin_controller
       test "should be an admin controller" do
         assert @controller.is_a?(Admin::BaseController), "the controller should be an admin controller"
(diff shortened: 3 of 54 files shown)


app/controllers/admin/base_controller.rb
```diff
@@ -49,8 +49,8 @@ def preview_design_system?(next_release: false)
   end
   helper_method :preview_design_system?
 
-  def render_design_system(design_system_view, legacy_view, next_release: false)
-    if preview_design_system?(next_release:)
+  def render_design_system(design_system_view, legacy_view, layout: "admin")
+    if preview_design_system?(next_release: layout == "design_system")
```
ollietreend, line 53: I think we might be actually able to simplify this to just:

```suggestion
  def render_design_system(design_system_view, legacy_view)
    if self.respond_to?(:get_layout) && get_layout == "design_system"
```

Rather than passing around the `next_release` or `layout` values at all – because I think that's already worked out by the `get_layout` method of each controller, right?

For example, the specific actions and user permissions are already tested here:

https://github.com/alphagov/whitehall/blob/b20bc5d3c1927e34695ffc681c64987de1ed414a/app/controllers/admin/corporate_information_pages_controller.rb#L30-L39

app/controllers/admin/cabinet_ministers_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
     @whip_roles = MinisterialRole.includes(:translations).whip.order(:whip_ordering)
     @organisations = Organisation.ministerial_departments.excluding_govuk_status_closed.order(:ministerial_ordering)
 
-    render_design_system(:show, :legacy_show, next_release: false)
+    render_design_system(:show, :legacy_show, layout: get_layout)
```
ollietreend, line 11: Which would mean you could ditch that third parameter altogether:

```suggestion
    render_design_system(:show, :legacy_show)
```

test/support/controller_test_helpers.rb
```diff
@@ -20,18 +16,22 @@ def should_be_a_public_facing_controller
...
+
+        user = login_as(:writer)
+        user.permissions << "Preview next release"
+
+        # Get all non-inherited controller actions and test they use the admin layout.
+        @controller.class.instance_methods(false).map(&:to_s).reject { |action| LAYOUT_ALWAYS_DESIGN_SYSTEM.include?(action) }.each do |action|
+          @controller.action_name = action
+          assert_equal "admin", @controller.send(:get_layout), "#{@controller.class}##{action} is not rendering the admin layout"
         end
       end
```
ollietreend, line 31: Very clever test 🤓 looks like it should be handy to catch any mistakes while we're still in this transitionary period.

app/controllers/admin/base_controller.rb
davidgisbey: yea great shout. Have implemented it 👍

app/controllers/admin/base_controller.rb
```diff
@@ -49,8 +49,8 @@ def preview_design_system?(next_release: false)
   end
   helper_method :preview_design_system?
 
-  def render_design_system(design_system_view, legacy_view, next_release: false)
-    if preview_design_system?(next_release:)
+  def render_design_system(design_system_view, legacy_view)
+    if self.respond_to?(:get_layout) && get_layout == "design_system"
```
ollietreend, line 53: Thinking about it, you can probably do without the `respond_to?(:get_layout)` call altogether.

After all, this will only run when you call `render_design_system` in a controller. So it should never be called on a controller without that method – and it's probably a good thing if the tests explode in that scenario.

```suggestion
    if get_layout == "design_system"
```

app/controllers/admin/base_controller.rb
davidgisbey: Just realised this actually needs to be as it returns false 

```
self.class.private_method_defined?(:get_layout) && get_layout == "design_system"
```

app/controllers/admin/base_controller.rb
davidgisbey: yup good shout. Makes sense to me

PR #7788, opened 2023-06-09
Display a feedback banner on design system pages

dnkrj:

[When the what's new banner is not being shown, instead use the phase banner to link to the publishing feedback email.](https://trello.com/c/m2SbCZ9U/1328-update-banner-for-changed-content)

# Screenshot

whitehall-admin dev gov uk_government_admin(iPad Pro)


app/views/layouts/design_system.html.erb
```diff
@@ -23,7 +23,7 @@
   </div>
 
   <div class="govuk-width-container">
-    <% if t('admin.whats_new.show_banner') %>
+    <% if t('admin.whats_new.show_banner') || t('admin.feedback.show_banner') %>
```
ollietreend, line 26: Remove the conditional here

```suggestion
```

app/views/shared/_whats_new_banner.html.erb
```diff
@@ -1,21 +1,34 @@
...
-          "read more about the changes",
-          admin_whats_new_path,
-          {
-            class: "govuk-link",
-            data: {
-              module: tracking_module,
-              track_category: 'phaseBanner',
-              track_action: 'whats-new-banner-click',
-              track_label: 'Whats new'
+<%if t('admin.whats_new.show_banner') %>
```
ollietreend, line 3: ...because it's already used here

app/views/shared/_whats_new_banner.html.erb
```diff
@@ -1,21 +1,34 @@
...
+              }
             }
-          }
-        )
-      }.", attributes: %w(class href data-module data-track-category data-track-action data-track-label))
-  } %>
+          )
+        }.", attributes: %w(class href data-module data-track-category data-track-action data-track-label))
+    } %>
+<% else %>
```
ollietreend, line 23: ...and then explicitly apply the 'else if' conditional here

```suggestion
<% elsif t('admin.feedback.show_banner') %>
```

test/integration/whats_new_test.rb
```diff
@@ -24,8 +24,8 @@ class WhatsNewTest < ActionDispatch::IntegrationTest
     login_as create(:gds_editor)
     get admin_whats_new_path
 
-    if I18n.t("admin.whats_new.show_banner")
-      assert_select ".gem-c-phase-banner"
+    if I18n.t("admin.whats_new.show_banner") || I18n.t("admin.feedback.show_banner")
+      assert_select ".gem-c-phase-banner", count: 1
     else
       assert_select ".gem-c-phase-banner", count: 0
     end
```
ollietreend, line 31: This test isn't all that great, because it doesn't stub/override the values of `I18n.t("admin.whats_new.show_banner")` or `I18n.t("admin.feedback.show_banner")`.

But again, this is a pre-existing smell – not something you've introduced. It's also not exactly mission critical functionality. So (wearing my pragmatic hat) I think this is probably okay as is.

ollietreend (commented):
I'm not a fan of using locale files as if they were configuration files. But I know you're just extending the existing approach that's in use for the "What's new" banner.

I also see what you mean about the purpose of the banner changing slightly (i.e. it's no longer just a "What's new" banner). Without boiling the ocean, I wonder if a sensible approach could be to rename the partial to `_admin_banner.html.erb` to reflect its evolved purpose.

I'd also be tempted to suggest we stop using the `if` conditional in the main layout file, and instead just implement them just in the partial file itself.

ollietreend (approved):

PR #7833, opened 2023-06-20
Fix issue with preview links showing before change note added

davidgisbey:
## Description 

There's an issue that a few publishers have encountered where they try and share a preview link with colleagues before an edition has been saved after a new edition has been created.

This occurs in the following situation:

1. a published document exists
2. the publisher creates a new draft document
3. when they arrive on the edit page they don't add a change note or mark the version as minor.
4. they leave the edit page via exiting the browser or by clicking the cancel link.
5. they have access to preview links and the sharable preview link functionality
6. if they share a preview link it doesn't work
7. if they click the `Preview on website (opens in new tab)` link they see the previous draft 

When a draft is successfully saved, it is pushed downstream to PublishingAPI and then to Draft Content Store where it updates existing content item.

This means that until this save occurs the Draft Content Store will return the latest draft that was pushed to it. In this case that would be the draft of the now published edition.

When the user clicks the 'Preview on website (opens in new tab)' link they'll hit draft-origin, which will make the call to Draft Content Store for the latest draft for that content_id. As stated this will load the latest draft it knows about which is now out of date.

This means you can do things like add attachments and update the taxonomy of the latest draft on Whitehall, and click the preview link and none of these changes will be reflected downstream as the edition is invalid due to not having a change note or being marked as a minor version.

Additionally, the sharable preview link will continue to show, but won't work for the following reason.

- the new draft edition is created with a new auth_bypass_id
- the sharable preview link is rendered with a token which is constructed using the new draft editions auth_bypass_id
- the user takes the link
- the old auth_bypass_id lives on the old draft in the Draft Content Store
- it attempts to decrypt the token constructed with the new auth_bypass_id using the value of the old auth_bypass_id

This PR updates edition show page to only render the links to draft preview and sharable links if the change note section has been completed when it is required on an edition (not the first draft edition etc.)

### Before

image

### After 

image

### DIFF
--- a/app/views/admin/editions/show/_main.html.erb
+++ b/app/views/admin/editions/show/_main.html.erb
@@ -5,21 +5,15 @@
     </div>
 
     <div>
-      <%=
-        status_text = [@edition.state.capitalize]
-        # TODO: remove unpublishing information once we have a separate state for unpublished editions
-        status_text << "(unpublished #{time_ago_in_words(@edition.unpublishing.created_at)} ago)" if @edition.unpublishing.present?
-
-        render "govuk_publishing_components/components/summary_list", {
-          borderless: true,
-          items: [
-            { field: "Type of document", value: edition_type(@edition) },
-            { field: "Status", value: status_text.join(' ') },
-            { field: "Change type", value: @edition.minor_change? ? 'Minor' : 'Major' },
-            *([{ field: "Organisations", value: joined_list(@edition.organisations.map { |o| o.name }) }] if @edition.respond_to?(:organisations)),
-          ]
-        }
-      %>
+      <%= render "govuk_publishing_components/components/summary_list", {
+        borderless: true,
+        items: [
+          { field: "Type of document", value: edition_type(@edition) },
+          { field: "Status", value: status_text(@edition) },
+          { field: "Change type", value: @edition.minor_change? ? 'Minor' : 'Major' },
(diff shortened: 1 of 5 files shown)


app/views/admin/editions/show/_main.html.erb
```diff
@@ -91,6 +91,17 @@
           <% end %>
         <% end %>
       <% end %>
+    <% elsif !@edition.publicly_visible? && !@edition.change_note_section_completed? %>
+      <%= render "govuk_publishing_components/components/heading", {
```
davidgisbey, line 95: i know this is a duplication, but i'm going to add a follow up PR which moves the preview logic into a view component as there's tonnes of untested logic here right now.

app/views/admin/editions/show/_main.html.erb
```diff
@@ -97,6 +91,17 @@
...
+    <% elsif !@edition.publicly_visible? && !@edition.versioning_completed? %>
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Preview",
+        heading_level: 2,
+        font_size: "l",
+        margin_bottom: 3,
+      } %>
+
+      <%= render "govuk_publishing_components/components/inset_text", {
+        text: "To preview this document or share a document preview, add a change note or update the change type to minor."
```
davidgisbey, line 103: This is just placeholder content and needs content review please

beccapearce (approved):
LGTM 🚀

PR #7834, opened 2023-06-21
Link draft attachments to their parent document

brucebolt:
Asset Manager has a concept of linking assets to their parent document. We use this URL in the CSV Previews to retrieve the metadata about the parent document.

At the moment, Whitehall only sends the parent document URL for live assets.

This extends that to send the parent document URL for draft assets too. Doing this will allow us to offer previews of draft CSV files to users.

[Trello card](https://trello.com/c/y2vagYMl)

### DIFF
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -139,6 +139,15 @@ class AttachmentData < ApplicationRecord
     draft_attachable.is_a?(Edition) ? draft_attachable : nil
   end
 
+  def draft_attachment
+    attachments.find { |attachment| attachment.attachable_type == "Edition" && attachment.attachable&.draft? }
+  end
+
+  def draft_edition
+    draft_attachable = draft_attachment&.attachable
+    draft_attachable.is_a?(Edition) ? draft_attachable : nil
+  end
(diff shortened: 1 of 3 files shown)


app/models/attachment_data.rb
```diff
@@ -139,6 +139,15 @@ def draft_edition_for(user)
     draft_attachable.is_a?(Edition) ? draft_attachable : nil
   end
 
+  def draft_attachment
+    attachments.find { |attachment| attachment.attachable_type == "Edition" && attachment.attachable&.draft? }
```
JonathanHallam, line 143: is the `&.` thing doing a check to make sure the attachable exists to stop it from falling over half way thorugh?

JonathanHallam (approved):
looks good, tiny non blocking question about how code works :)

app/models/attachment_data.rb
brucebolt: Yeah, it is theoretically possible to have an orphaned attachment which isn't attached to anything (and presumably has been an issue in the past, as the other similar methods all include the safe navigation operator.

PR #7843, opened 2023-06-22
Architecture Decision Record for new asset model

kornysietsma:
Adding an Architecture Decision Record for changes to how Assets are modeled - follow-on from previous RFC and discussion.


tunylund:
This page is best to use for reviewing the ADR
https://github.com/alphagov/whitehall/blob/adr-0002-new-asset-model/docs/adr/0002-new-asset-model.md

kevindew:
This sounds great to me, thanks @kornysietsma - very few things few that I have to comment on.

The question I did have was about what the relationship between a resource and an Asset, namely how a resource knows about its assets? I have a feeling I'm not phrasing this question well so might be best illustrated with some examples:

for the situation of ImageData we have one ImageData model and multiple Asset models for each image size variant, what would be the way that these know of each other? Some routes I could see are: foreign_keys on ImageData (i.e. fields for s300_asset_id, s640_asset_id ...) and a very rigid way of storing variants; an intermediary model e.g. ImageDataAsset that stores image_data_id, asset_id, variant; or a polymorphic relationship from Asset, where it tracks the owner Id, owner class and variant?

ollietreend:
Just leaving a comment here to let you know I've seen this PR and think it looks good – but have the same question as Kevin, re: how the `AttachmentData`/`ImageData` to `Asset` relationship works.

It would be good to understand how you're thinking about implementing that. I suspect a polymorphic relationship is most in keeping with Whitehall's current approach (for example, how `Attachment` records are associated to their 'attachable' model objects). But that's not a strongly held opinion if you think there's a better way to do it.

kevindew (approved):
Looks good to me 👍

PR #7844, opened 2023-06-22
Update the base path of the Worldwide Corporate Information about pages

jkempster34:
https://trello.com/c/cxmT8SUk

The worldwide corporate information about pages differ from the organisation
corporate information about pages in that they are not rendered on a separate
page.

Currently, we attempt to prevent worldwide corporate information about pages
from being published at all. When they do get published, they raise an error as
the base_path conflicts with the existing content item for the worldwide
organisation itself.

Instead, we have decided to intentionally publish all worldwide corporate
information about pages as `placeholder` items with the base_path
`about/about`. This means we can avoid the conflicting base_path errors that
occur when they are published.

This:
- updates the base_path for corporate information pages as described above.
- Removes a guard statement that was added some time ago in an attempt to
prevent worldwide corporate information about pages from being published.

---

### DIFF
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -175,13 +175,11 @@ class CorporateInformationPage < Edition
   end
 
   def base_path
-    if worldwide_organisation.present?
-      url = edition_worldwide_organisation.worldwide_organisation.base_path.to_s + "/about/#{slug}"
-      url.gsub("/about/about", "")
-    elsif organisation.present?
-      url = edition_organisation.organisation.base_path.to_s + "/about/#{slug}"
-      url.gsub("/about/about", "/about")
-    end
+    return if owning_organisation.blank?
+
+    url = owning_organisation.base_path + "/about/#{slug}"
+    url.gsub!("/about/about", "/about") if organisation.present?
+    url
   end
 
 private
--- a/app/services/edition_service.rb
+++ b/app/services/edition_service.rb
@@ -42,11 +42,6 @@ class EditionService
 
 private
 
-  def is_whitehall_corp_info_page?
-    edition.type == "CorporateInformationPage" &&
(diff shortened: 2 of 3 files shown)


app/models/corporate_information_page.rb
```diff
@@ -175,13 +175,15 @@ def alternative_format_provider_required?
...
-      url = edition_worldwide_organisation.worldwide_organisation.base_path.to_s + "/about/#{slug}"
-      url.gsub("/about/about", "")
-    elsif organisation.present?
-      url = edition_organisation.organisation.base_path.to_s + "/about/#{slug}"
-      url.gsub("/about/about", "/about")
-    end
+    return if owning_organisation.blank?
+
+    url = owning_organisation.base_path + "/about/#{slug}"
+    # The corporate information pages associated with worldwide organisations are not rendered
```
JonathanHallam, line 181: We tend to try and avoid comments if possible. I'd suggest adding a line or two to the commit message so that we can remove these.

app/models/corporate_information_page.rb
```diff
@@ -175,13 +175,15 @@ def alternative_format_provider_required?
...
+    # on their own page. Instead, we publish them as `placeholder` content items with the path
+    # /about/about`. On the other hand, the corporate information pages associated with organisations
+    # are rendered on their own page, which we want to have the path `/about`.
+    url.gsub!("/about/about", "/about") if organisation.present?
```
JonathanHallam, line 185: I think adding an explanation of this line to the commit message would suffice to remove comments. Basically just saying that the presence of an organisation rather than a worldwide_organisation is enough for us to work out what's going on

app/models/corporate_information_page.rb
jkempster34: I thought this might be warranted because it's quite confusing. Happy to remove it if you disagree!

JonathanHallam (approved):
Thanks!

PR #7847, opened 2023-06-23
Rename consultation responses

minhngocd:
Response model is only associated with Consultation, therefore should be renamed to something more specific such as ConsultationResponse. This is also to prepare us for introducing CallForEvidenceResponse. We want to avoid having polymorphic relationships where we create ConsultationResponse and CallForEvidenceResponse to inherit Response because this will create 2 layers of polymorphism, including the 2 classes that currently inherit from (Consultation)Response - ConsultationOutome and ConsultationPublicFeedback. To keep things simple and easy to navigate and maintain in the future, we have opted to rename Response
ConsultationResponseFormUploader was renamed since it was an uploader for documents. There was nothing in the uploader that signified anything specific to consultation, and also it was for uploading documents given the file extensions. So rather than calling it a form uploader, we can make it a generic document uploader to reuse for both Consultation and Call For Evidence and anything else that might require a document in a similar way.

https://trello.com/c/Bx2zLsHM/1312-create-the-basic-call-for-evidence-document-type-in-whitehall


app/uploaders/response_document_uploader.rb
```diff
@@ -1,4 +1,4 @@
-class ConsultationResponseFormUploader < WhitehallUploader
+class ResponseDocumentUploader < WhitehallUploader
```
ryanb-gds, line 1: Is this right that we've removed the "Consultation" element from this? I suppose the uploader can easily be shared?

ryanb-gds (approved):
One question but assuming the answer is yes then good to go I think.

app/uploaders/response_document_uploader.rb
minhngocd: Yes we have! In contrast to the other renaming, this one actually has nothing tied to Consultation in the code. So we thought we could rename it to something generic between the two and reuse for any type of Response Document (judging by the file extension allowlist)

PR #7848, opened 2023-06-23
Cleanup nested documents collections

davidgisbey:
## Description

There's currently a workaround to add document collections to a DocumentCollectionGroup that a large group of users have used to circumvent this validation.

Instead of directly attaching the document, users attach other collections as ‘external links’.

This is documented in this card
https://trello.com/c/RKMn2oIs/55-nested-document-collections-are-not-allowed-but-they-exist

We've decided that we should just remove this validation as it doesn't seem to be causing any massive issues.

## Trello card

https://trello.com/c/RKMn2oIs/55-nested-document-collections-are-not-allowed-but-they-exist


ollietreend (approved):
Nice

PR #7860, opened 2023-06-27
Remove the use of show legacy endpoint for file attachments

lauraghiorghisor-tw:

This first slice of changes is replacing the `GET` calls to the legacy `/whitehall_assets/:legacy_url_path` with calls to `/assets/:id` - which are the `/show` endpoints in Asset Manager. The legacy call returns the asset's document ID which is then used for all subsequent calls to delete or update. We want whitehall to use only the non-legacy `/show` endpoint, in line with the rest of the publishing applications, thus also removing this redundant translation step and the usage of `legacy_url_path`.  

In order to accomplish this, the asset manager document ID returned by the `/create` endpoint is stored in an `Asset` model. The new `Asset` model has a `belongs_to` relationship to the `AttachmentData` and currently only serves the file attachments flow. It also includes a version, e.g. `file` or `thumbnail`.

All current changes are behind `User::Permissions::USE_NON_LEGACY_ENDPOINTS`. In addition to the user permission, the code checks if there are any `Assets` on the given `AttachmentData` to decide whether or not it should use the old or new flow. 

Any relevant unit/integration tests have been added. Where applicable, the tests are split in a "feature flag on/feature flag off" manner to enable easy removal of flag. The expected behaviour has also been manually tested in the integration environment. When the `USE_NON_LEGACY_ENDPOINTS` permission is enabled the code uses the non-legacy `/show` endpoint rather than the legacy one. 

[RFC here](https://docs.google.com/document/d/1YLsnqhu0BNNCnHJa62yHr_BbyYJS8pIWTRFZKcIezwI/edit#heading=h.3m6qxrmo17ls)


app/services/asset_manager/attachment_updater/access_limited_updates.rb
```diff
@@ -5,15 +5,28 @@ def self.call(attachment_data)
...
         )
+
+        if attachment_data.pdf?
+          enum.yield AssetManager::AttachmentUpdater::Update.new(
+            nil, attachment_data, attachment_data.file.thumbnail, access_limited:
+          )
+        end
+      end
+    else
+      # Both the file and its thumbnail are now represented by an Asset
```
lauraghiorghisor-tw, line 21: In all these files `xxx_updates` we have kept the two possible flows coexisting and the new one should only run if assets are present.

test/unit/services/asset_manager/attachment_updater/access_limited_updates_test.rb
```diff
@@ -8,89 +8,189 @@ class AssetManager::AttachmentUpdater::AccessLimitedUpdatesTest < ActiveSupport:
     let(:attachment_data) { attachment.attachment_data }
     let(:update_worker) { mock("asset-manager-update-asset-worker") }
 
-    around do |test|
-      AssetManager.stub_const(:AssetUpdater, update_worker) do
-        test.call
+    describe "Attachment Data has no assets" do
```
lauraghiorghisor-tw, line 11: For this test and other similar to it, we have split the tests in Feature flag on and off respectively (in our case: assets present or not). We have then replicated all tests to match the new implementation and on occasion added more tests or slightly adjusted the logic. 
The Asset model simplifies the tests a little bit in that there is no need to always tests whether the attachment is a pdf since the new `variant` field helps deduce whether we're dealing with original or thumbnail (or in the future specific image versions).

PR #7861, opened 2023-06-28
New image uploader

minhngocd:
This new image uploader is used for featured images (including Lead Images, Person Images, Featured Images) and hence do not allow SVG file uploads.

Most non-inline images currently use ImageUploader, but only ever send a specific size over to publishing API, resulting in an error or some other unexpected behaviour when SVGs are uploaded.
The logic required to add SVG blocking using in conjunction with the other ImageUploader and also managing error messages seemed overly complex as compared to adding another uploader for these types of Featured Images which encompasses the resizing as well as the extension blocking and its validation messages.

SVGs should be used for graphs and charts and not as person images, featured images, and lead images, so we have introduced a new uploader to support this type of images.

https://trello.com/c/ZjWXXW93/1438-add-wrong-file-type-errors-to-topical-events-logo-uploads

### DIFF
--- a/app/uploaders/featured_image_uploader.rb
+++ b/app/uploaders/featured_image_uploader.rb
@@ -1,0 +1,23 @@
+class FeaturedImageUploader < WhitehallUploader
+  include CarrierWave::MiniMagick
+
+  def extension_allowlist
+    %w[jpg jpeg gif png]
+  end
+
+  configure do |config|
+    config.remove_previously_stored_files_after_update = false
+    config.validate_integrity = true
+  end
+
+  version(:s960) { process resize_to_fill: [960, 640] }
+  version(:s712, from_version: :s960) { process resize_to_fill: [712, 480] }
+  version(:s630, from_version: :s960) { process resize_to_fill: [630, 420] }
+  version(:s465, from_version: :s960) { process resize_to_fill: [465, 310] }
+  version(:s300, from_version: :s960) { process resize_to_fill: [300, 195] }
+  version(:s216, from_version: :s960) { process resize_to_fill: [216, 140] }
+
+  def image_cache
+    file.file.gsub("/govuk/whitehall/carrierwave-tmp/", "") if send("cache_id").present?
+  end
+end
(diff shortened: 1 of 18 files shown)


ryanb-gds (approved):
Looks good to me

app/uploaders/featured_image_uploader.rb
```diff
@@ -0,0 +1,23 @@
+class FeaturedImageUploader < WhitehallUploader
```
ollietreend, line 1: Generally speaking, I feel like the words "feature" and "featured" are a bit over-used in Whitehall. And it's making me wonder if there could be a slightly more intention revealing name for this uploader class.

For somebody new coming into this codebase:

- The `FeaturedImageUploader` is used for `Feature#image` – and with a name this similar, it sort of sounds like it was made specifically for this use case. _Is it supposed to be coupled to this model?_ 🤔
- But it's also used on `PromotionalFeatureItem#image`. _So maybe it's supposed to be used for images on "featured" things?_ 🤔
- But then it's also used on `Person#image`. _Does that mean it's been re-purposed for something it wasn't originally intended?_ 🤔

I suspect you've probably already thought through different naming options before landing on this. So I'm totally happy for you to stick with this name. This is not a strongly held opinion 🙂 but the only suggestion I might throw into the ring is to maybe call it `BitmapImageUploader` – which would hopefully reveal the intention of this being used for uploading images that cannot be vectors.

ollietreend (commented):
Lovely work @minhngocd and @jyoung-gds 😎

Just a small opinion about naming.

app/uploaders/featured_image_uploader.rb
minhngocd: We struggled a lot with this as well and was also teetering between BitmapImageUploader and FeaturedImageUploader. We thought about if people were implementing new document types with images, how do they know when to use ImageUploader and BitmapImageUploader - so we thought that having FeaturedImageUploader would be clearer about when it should be used (and maybe renaming ImageUploader to InlineImageUploader or something equivalent). 

But agree it's not the best name - we have a thread trying to ask design folks what they would call these types of images as opposed to the inline ones and "Featured" was one of the things that was suggested.

minhngocd: I'm either way writing up a tech debt card in the backlog to maybe rethink image scaling for these types of images (backend vs frontend) - maybe i'll also include in there how to categorise and name image uploaders?

ollietreend: Thanks for explaining @minhngocd – that makes sense! ✨ ✅

ollietreend (approved):
Looks good to me 👍🏻

PR #7862, opened 2023-06-28
Remove unused attachable helper methods

ryanb-gds:
Removing some of the attachable helper methods because they appear to be unused


beccapearce (approved):
Looks great! Nice catch! 🌟

PR #7863, opened 2023-06-28
Add calls for evidence to the document filter

beccapearce:
Adding calls for evidence to the list of document types that can be filtered on the document index page

Before:
Screenshot 2023-06-28 at 11 54 45

After: 
Screenshot 2023-06-28 at 12 02 04


ryanb-gds (approved):
Short and sweet!

PR #7864, opened 2023-06-28
Fix bug with filter for topical event featurings

davidgisbey:
## Description

At the moment, this method returns all featurable editions for a topical event featurings.

This isn't actually the use case needed. What this needs to do is take a subset of editions and filter out any that belong to a topical event featuring.

This allows it to work with the paginator which only renders 15 per page.

This isn't the perfect solution as it it filters 1 out the paginator will only render 14 results, but a big refactor of the filter is beyond the scope of this piece of work. It should be something we take on post-transition though

### Before 

image

### After

image

## Trello card

https://trello.com/c/e5MmKlGd/329-filter-not-working

### DIFF
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -207,7 +207,7 @@ class TopicalEvent < ApplicationRecord
     end
   end
 
-  def featurable_editions
+  def featurable_editions(editions)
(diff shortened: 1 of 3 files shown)


chriswhitehouse-tw (approved):
LGTM

PR #7869, opened 2023-06-28
292 organisation financial reports

davidtrussler:
[Trello](https://docs.google.com/document/d/1XsSgkcWi2gYz36R7pkJT_Fzdi0DJOCqzC6a4n9X6HUc/edit)

Ports the Organisation / Financial Reports landing page to the Design System and creates the Delete view (screenshots below). 
- Updates controller for design_system
- Updates controller test for design_system
- Creates UI for design_system on index page
- Adds confirm_destroy view

|Page description|View|
|-|-|
|Landing page with no reports|Screenshot 2023-06-28 at 14 22 02|
|Landing page with reports|Screenshot 2023-06-28 at 14 21 15|
|Confirm delete page|Screenshot 2023-06-28 at 14 31 01|

### DIFF
--- a/app/controllers/admin/financial_reports_controller.rb
+++ b/app/controllers/admin/financial_reports_controller.rb
@@ -1,6 +1,11 @@
 class Admin::FinancialReportsController < Admin::BaseController
   before_action :load_organisation
   before_action :load_financial_report, only: %i[edit update destroy]
+  layout :get_layout
+
+  def index
+    render_design_system(:index, :legacy_index)
+  end
 
   def edit; end
 
@@ -30,7 +35,22 @@ class Admin::FinancialReportsController < Admin::BaseController
     redirect_to admin_organisation_financial_reports_path(@organisation), notice: "Deleted Successfully"
   end
 
+  def confirm_destroy
+    @financial_report = @organisation.financial_reports.find(params[:id])
+  end
+
 private
+
+  def get_layout
+    design_system_actions = []
+    design_system_actions += %w[index confirm_destroy] if preview_design_system?(next_release: false)
+
+    if design_system_actions.include?(action_name)
+      "design_system"
+    else
+      "admin"
+    end
+  end
 
   def load_financial_report
     @financial_report = @organisation.financial_reports.find(params[:id])
--- a/app/views/admin/financial_reports/confirm_destroy.html.erb
+++ b/app/views/admin/financial_reports/confirm_destroy.html.erb
@@ -1,0 +1,21 @@
+<% content_for :context, @organisation.name %>
+<% content_for :page_title, "Delete Financial report" %>
+<% content_for :title, "Delete Financial report" %>
+<% content_for :title_margin_bottom, 4 %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body govuk-!-margin-bottom-8">Are you sure you want to delete this report?</p>
+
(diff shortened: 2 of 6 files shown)


app/controllers/admin/financial_reports_controller.rb
```diff
@@ -1,6 +1,9 @@
 class Admin::FinancialReportsController < Admin::BaseController
   before_action :load_organisation
   before_action :load_financial_report, only: %i[edit update destroy]
+  layout :get_layout
+
+  def index; end
```
davidgisbey, line 6: I think we're missing the bootstrap implementation here. So this needs to be 

```
def index
  render_design_system(:index, :legacy_index)
end

```

And we need to retain the boostrap implementation in the `legact_index` view. Otherwise the user will always see the design system page.

app/views/admin/financial_reports/confirm_destroy.html.erb
```diff
@@ -0,0 +1,21 @@
+<% content_for :context, "Financial report" %>
```
davidgisbey, line 1: I suspect this should be

```
<% content_for :context, @organisation.name %>
```

might be worth double checking with Nik, but that's generally the pattern we've followed.

app/views/admin/financial_reports/index.html.erb
```diff
@@ -1,39 +1,74 @@
...
+              text: "Spending"
+            },
+            {
+              text: tag.span("Actions", class: "govuk-visually-hidden")
+            }
+          ],
+          rows: @organisation.financial_reports.map do |report|
+            [
+              {
+                text: tag.span(report.year, class: "app-view-financial-reports__listing__year")
```
davidgisbey, line 50: can we just do

```
 text: tag.span(report.year, class: "govuk-!-font-weight-bold")
```

app/views/admin/financial_reports/index.html.erb
```diff
@@ -1,39 +1,74 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Financial Reports",
+      margin_bottom: 6
+    } %>
+
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Create new report",
+      href: {action: 'new'},
```
davidgisbey, line 26: i can see this works but i wonder if we can use

```
new_admin_organisation_financial_report_path(@organisation)
```

We already use a few patterns for creating href's so we can use one of those

davidgisbey: Also Ruth has asked us to track all primary buttons so can we add 

```
data_attributes: {
        module: "gem-track-click",
        "track-category": "form-button",
        "track-action": "create-financial-report-button",
        "track-label": "Create new report"
      }
```

app/views/admin/financial_reports/confirm_destroy.html.erb
davidtrussler: > I suspect this should be
> 
> ```
> <% content_for :context, @organisation.name %>
> ```
> 
> might be worth double checking with Nik, but that's generally the pattern we've followed.

Yes actually that would make more sense, and be more consistent. 👍

app/views/admin/financial_reports/index.html.erb
davidtrussler: > 

Yeah, that would be simpler wouldn't it? 😆

davidgisbey (approved):

PR #7873, opened 2023-06-28
Ensure cancel link works on unsuccessful save for new featurings page

davidgisbey:
## Description

 Ensure cancel link works on unsuccessful save for new featurings page

When the page saves unsuccessfully the cancel link doesn't construct the link correctly. When you click it no HTTP request
is made.

This only occurs when an anchor is present. As the PATCH update resource has the same url the GET index endpoint instead of reloading the page, it simply enacts the anchors behaviour when you click on it.

As far as i can tell there's no way to force a reload without using JSor adding a query string. This isn't an ideal situation but will force a page reload when clicked.

## Trello card

https://trello.com/c/xYVJjW0f/326-cancel-button-doesnt-work-if-there-is-an-error

### DIFF
--- a/app/views/admin/topical_event_featurings/new.html.erb
+++ b/app/views/admin/topical_event_featurings/new.html.erb
@@ -56,7 +56,8 @@
         <%= link_to "Cancel",
           polymorphic_path(
             [:admin, @topical_event, :topical_event_featurings],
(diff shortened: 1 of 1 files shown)


chriswhitehouse-tw (approved):
LGTM

PR #7879, opened 2023-06-29
Fix call for evidence path

beccapearce:
We had a mismatch in our code of whether the admin paths for call for evidence should have hyphens or underscores.

This PR sets the path for calls for evidence to be '/calls-for-evidence' to ensure it is standardised.


ryanb-gds (approved):
LGTM. Kinda weird that Rails doesn't hyphenate paths by default

PR #7880, opened 2023-06-29
Revert "Switch the rendering of worldwide CIPs to government-frontend"

jkempster34:
We haven't yet handled RTL direction translations. I'm fairly confident that there are no translations of RTL languages currently for worldwide CIPs, but reverting until we have fixed.

Reverts alphagov/whitehall#7867


JonathanHallam (approved):
thanks

PR #7882, opened 2023-06-30
Save asset manager id on file creation

lauraghiorghisor-tw:

In order to accomplish this, the asset manager document ID returned by the `/create` endpoint is stored in an `Asset` model. The new `Asset` model has a `belongs_to` relationship to `AttachmentData` and currently only serves the file attachments flow. It also includes a `version`, e.g. `file` or `thumbnail`.

All current changes are behind `User::Permissions::USE_NON_LEGACY_ENDPOINTS`. Where applicable, the tests are split in a "feature flag on/feature flag off" manner to enable easy removal of flag. The expected behaviour has also been manually tested in the integration environment. 

[RFC](https://docs.google.com/document/d/1YLsnqhu0BNNCnHJa62yHr_BbyYJS8pIWTRFZKcIezwI/edit#heading=h.3m6qxrmo17ls)
[ADR](https://github.com/alphagov/whitehall/blob/adr-0002-new-asset-model/docs/adr/0002-new-asset-model.md)

### DIFF
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -3,7 +3,7 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
 
   sidekiq_options queue: "asset_manager"
 
-  def perform(file_path, legacy_url_path, draft = false, attachable_model_class = nil, attachable_model_id = nil, auth_bypass_ids = [])
+  def perform(file_path, legacy_url_path, model_id = nil, asset_variant = nil, draft = false, attachable_model_class = nil, attachable_model_id = nil, auth_bypass_ids = [])
     return unless File.exist?(file_path)
 
     file = File.open(file_path)
@@ -18,7 +18,9 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
       end
     end
 
-    asset_manager.create_whitehall_asset(asset_options)
+    response = asset_manager.create_whitehall_asset(asset_options)
+
+    create_asset_manager_asset(model_id, asset_variant, response)
 
     if attachable_model
       # The AttachmentData we want to set the timestamp on may not
@@ -32,5 +34,24 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
     file.close
     FileUtils.rm(file)
     FileUtils.rmdir(File.dirname(file))
+  end
+
+private
+
+  def create_asset_manager_asset(model_id, asset_variant, response)
+    return unless model_id && asset_variant
+
+    response_id = get_asset_id(response)
+    save_asset_id_to_assets(model_id, asset_variant, response_id)
+  end
+
+  def get_asset_id(response)
+    attributes = response.to_hash
+    url = attributes["id"]
+    url[/\/assets\/(.*)/, 1]
+  end
+
+  def save_asset_id_to_assets(model_id, variant, asset_manager_id)
+    Asset.create!(asset_manager_id:, attachment_data_id: model_id, variant:)
   end
 end
--- a/test/integration/asset_manager_test.rb
+++ b/test/integration/asset_manager_test.rb
@@ -2,6 +2,8 @@ require "test_helper"
 
 class AssetManagerIntegrationTest
   class CreatingAFileAttachment < ActiveSupport::TestCase
+    extend Minitest::Spec::DSL
+
     setup do
       @filename = "960x640_jpeg.jpg"
       @attachment = FactoryBot.build(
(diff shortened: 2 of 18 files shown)


app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -18,7 +18,9 @@ def perform(file_path, legacy_url_path, draft = false, attachable_model_class =
       end
     end
 
-    asset_manager.create_whitehall_asset(asset_options)
+    response = asset_manager.create_whitehall_asset(asset_options)
+
+    create_asset_manager_asset(model_id, asset_variant, response)
```
ollietreend, line 23: I'm slightly nervous about this.

The code comment just a few lines below this says:

> The AttachmentData we want to set the timestamp on may not exist yet

which is why the job `AssetManagerAttachmentSetUploadedToWorker` runs asynchronously ½ second later – i.e. to make sure the `AttachmentData` model has been persisted to the database.

And if it hasn't been persisted yet, presumably it won't have an ID yet. So I'm sort of expecting `model_id` to be `nil`. In which case, I don't think we should be able to create an `Asset` at this point in time/code?

I saw your earlier commit which implemented this in the `AssetManagerAttachmentSetUploadedToWorker` – which makes more sense to me.

Am I missing something? 🤔

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -33,4 +35,24 @@ def perform(file_path, legacy_url_path, draft = false, attachable_model_class =
...
+
+  def get_asset_id(response)
+    attributes = response.to_hash
+    url = attributes["id"]
+    url[/\/assets\/(.*)/, 1]
+  end
+
+  def save_asset_id_to_assets(model_id, variant, asset_manager_id)
+    asset = Asset.new(asset_manager_id:, attachment_data_id: model_id, variant:)
+    asset.save!
```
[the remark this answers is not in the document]
ollietreend: Also, if we had the `AttachmentData` object, this would be more idiomatic if implemented using the [auto-generated ActiveRecord association method](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#module-ActiveRecord::Associations::ClassMethods-label-Auto-generated+methods):

```ruby
attachment_data.create_asset!(asset_manager_id:, variant:)
```

this way it'll automatically be associated with that Attachment Data object.

app/workers/asset_manager_create_whitehall_asset_worker.rb
ollietreend: One more thought: I wonder if this situation isn't occurring in test because Sidekiq jobs don't execute asynchronously. For example, the integration tests make use of `Sidekiq::Testing.inline!` to cause worker jobs to be executed synchronously.

Could it be possible that the `AttachmentData` record will always exist in tests, but not in the real world? 🤔

test/integration/asset_manager_test.rb
```diff
@@ -10,33 +12,73 @@ class CreatingAFileAttachment < ActiveSupport::TestCase
...
-      Services.asset_manager.expects(:create_whitehall_asset).with(file_and_legacy_url_path_matching(/#{@filename}/))
+    describe "use_non_legacy_endpoints is false" do
+      test "sends the attachment to Asset Manager" do
+        Services.asset_manager.expects(:create_whitehall_asset).with(file_and_legacy_url_path_matching(/#{@filename}/))
 
-      Sidekiq::Testing.inline! do
-        @attachment.save!
+        Sidekiq::Testing.inline! do
+          @attachment.save!
+        end
```
ollietreend, line 21: I wonder if this is giving us a false sense of security that the `AttachmentData` will already have been created, despite the code comment saying otherwise.

test/unit/whitehall/asset_manager_storage_test.rb
```diff
@@ -112,6 +113,63 @@ def assets_protected?
...
+
+      context "use_non_legacy_endpoints permission is true" do
+        setup do
+          @model.use_non_legacy_endpoints = true
+        end
+
+        test "calls worker with model_id and default origin asset_variant" do
+          variant = Asset.variants[:original]
+
+          AssetManagerCreateWhitehallAssetWorker.expects(:perform_async).with(anything, anything, @model.id, variant, anything, anything, anything, anything)
```
ollietreend, line 132: The AttachmentData has only been built, but not saved to the database yet.

https://github.com/alphagov/whitehall/blob/291a38416f62a1bf44b7b6f328eb1ac7b17188f4/test/unit/whitehall/asset_manager_storage_test.rb#L120

Therefore I believe `@model.id` will be `nil`.

Which means that you're actually expecting a `nil` value for `model_id` here:

https://github.com/alphagov/whitehall/blob/291a38416f62a1bf44b7b6f328eb1ac7b17188f4/test/unit/whitehall/asset_manager_storage_test.rb#L132

I don't think this is what you intended this test to be doing.

ollietreend (commented):
Thanks for opening this PR @lauraghiorghisor-tw and @syed-ali-tw.

At a high level, this PR makes sense and I think it looks good. However there are a couple of themes in my comments which I think we'll need to address before this ready to go.

1. **Use of a feature flag** – and if one is actually needed for this piece of work
2. **Sync vs. async sequencing** – and when it's safe to do what

I'd be very happy to talk through this review with you if it's helpful. Just let me know!

test/unit/whitehall/asset_manager_storage_test.rb
lauraghiorghisor-tw: Thanks for catching that!

app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -33,4 +35,24 @@ def perform(file_path, legacy_url_path, draft = false, attachable_model_class =
...
+
+    response_id = get_asset_id(response)
+    save_asset_id_to_assets(model_id, asset_variant, response_id)
+  end
+
+  def get_asset_id(response)
+    attributes = response.to_hash
+    url = attributes["id"]
+    url[/\/assets\/(.*)/, 1]
+  end
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Same code (that we know of) exists in the `AssetManager::ServiceHelper` for extracting ID out of the response coming from a legacy call to GET asset - in order to then make a PUT request with said ID (quite a mouthful). But that code will go away as per the next PR (well, once the feature flags are removed). So, yes, this is probably alright in here 👍🏻

app/workers/asset_manager_create_whitehall_asset_worker.rb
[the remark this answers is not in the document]
lauraghiorghisor-tw: That would have been lovely indeed, if only for not passing a billion arguments to this worker (quite the eyesore). But one cannot simply pass an object (insert LOTR meme here).

app/workers/asset_manager_attachment_set_uploaded_to_worker.rb
```diff
@@ -35,16 +35,27 @@ def perform(model_class, model_id, legacy_url_path)
       if attachment_data.path == legacy_url_path
         found = true
         attachment_data.uploaded_to_asset_manager!
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Definitely something to consider as part of the other epic around state syncing. 

Also on this one, I am very blurry as to the usefulness of the updates being called inside that method (given they also get called from the controller and other places) because the updates don't really include anything referring to this timestamp. Update calls sprinkled everywhere feels like a catch-all to cover all the points where the async nature of it all might crack.

@kornysietsma Tagging you so that this lives in your brain too - as mastermind of syncing.

ollietreend (approved):
Looks good to me – thanks @lauraghiorghisor-tw & @syed-ali-tw! 🚀

PR #7883, opened 2023-06-30
Remove usage of legacy url path from file attachment updates

lauraghiorghisor-tw:

This first slice of changes is replacing the `GET` calls to the legacy `/whitehall_assets/:legacy_url_path` with calls to `/assets/:id ` (non-legacy endpoints used by other publishing apps). In the as-is, the legacy call returns the asset's document ID which is then used for all subsequent calls to delete, update etc. We want `whitehall` to use only the non-legacy `/show` endpoint thus also removing this redundant translation step and the usage of `legacy_url_path` (eventually). Most of the required changes are around the various attachment updaters that support this flow.

Preliminary work around saving asset manager ID in a new Asset model has been done [in this PR](https://github.com/alphagov/whitehall/pull/7882).

Whilst the changes for saving the ID are behind a user permission, the changes in this PR check for the existence of `assets` on the associated `AttachmentData` to select whether the old or new flow should be used. The flows will co-exist until the migration work is laid out and user permission can be removed. 

Where applicable, the tests are split in a "feature flag on/feature flag off" manner to enable easy removal of flag. The expected behaviour has also been manually tested in the integration environment.

[RFC](https://docs.google.com/document/d/1YLsnqhu0BNNCnHJa62yHr_BbyYJS8pIWTRFZKcIezwI/edit#heading=h.3m6qxrmo17ls)
[ADR](https://github.com/alphagov/whitehall/blob/adr-0002-new-asset-model/docs/adr/0002-new-asset-model.md)

### DIFF
--- a/app/services/asset_manager/attachment_updater/access_limited_updates.rb
+++ b/app/services/asset_manager/attachment_updater/access_limited_updates.rb
@@ -5,15 +5,28 @@ class AssetManager::AttachmentUpdater::AccessLimitedUpdates
       access_limited = AssetManagerAccessLimitation.for(attachment_data.access_limited_object)
     end
 
-    Enumerator.new do |enum|
-      enum.yield AssetManager::AttachmentUpdater::Update.new(
-        attachment_data, attachment_data.file, access_limited:
-      )
-
-      if attachment_data.pdf?
+    if attachment_data.assets.empty?
+      Enumerator.new do |enum|
         enum.yield AssetManager::AttachmentUpdater::Update.new(
-          attachment_data, attachment_data.file.thumbnail, access_limited:
+          nil, attachment_data, attachment_data.file, access_limited:
         )
+
+        if attachment_data.pdf?
+          enum.yield AssetManager::AttachmentUpdater::Update.new(
+            nil, attachment_data, attachment_data.file.thumbnail, access_limited:
+          )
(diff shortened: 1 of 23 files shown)


app/services/asset_manager/attachment_updater/access_limited_updates.rb
```diff
@@ -5,15 +5,28 @@ def self.call(attachment_data)
...
+        if attachment_data.pdf?
+          enum.yield AssetManager::AttachmentUpdater::Update.new(
+            nil, attachment_data, attachment_data.file.thumbnail, access_limited:
+          )
+        end
+      end
+    else
+      # Both the file and its thumbnail are now represented by an Asset
+      # and rely solely on the asset_manager_id rather than the legacy_url_path for updates.
+      # Therefore, it is not needed to check whether or not the Attachment is a pdf.
```
ollietreend, line 23: Nice! 🤓 This will simplify things.

ollietreend (approved):
Nice work! 🤩

It's great to see you eating this elephant [one bite at a time](https://www.psychologytoday.com/us/blog/mindfully-present-fully-alive/201804/the-only-way-to-eat-an-elephant). I'm looking forward to seeing the back of `legacy_url_path`! 🙌🏻

PR #7887, opened 2023-07-03
273 edit people

chriswhitehouse-tw:
## Description

This PR describes the transitioning from the previous Bootstrap method of implementation to the new Government Design System components. 

It does the following:

1. Prepend people page with legacy
2. Add people page into GOV.UK design system
3. Add reorder people endpoint to organisations
4. Refactor people endpoints into organisation people controller

### Before

When you first visit this page, this is what you will see:

whitehall-admin integration publishing service gov uk_government_admin_organisations_government-digital-service_people

There is another view that can be accessed, as we can see below this screenshot showcases a Government Department and Agency that has more than one type of roles:

Screenshot 2023-07-12 at 16 03 48

Below is a view of a Government Department and Agency that has no roles associated with it:

whitehall-admin integration publishing service gov uk_government_admin_organisations_financial-reporting-council_people

### After

My organisation with only one role type:
whitehall-admin dev gov uk_government_admin_organisations_government-digital-service_people

My organisation with multiple role types
whitehall-admin dev gov uk_government_admin_organisations_cabinet-office_people

Not my organisation - no reorder link:
whitehall-admin dev gov uk_government_admin_organisations_office-for-disability-issues_people

No roles associated with organisation:
whitehall-admin dev gov uk_government_admin_organisations_academy-for-justice-commissioning_people

Reorder page:
whitehall-admin dev gov uk_government_admin_organisations_cabinet-office_people_reorder_title=Ministers type=ministerial

Reorder success banner:
whitehall-admin dev gov uk_government_admin_organisations_cabinet-office_people (3)

## Trello
https://trello.com/c/rvouyXnr


app/controllers/admin/organisations_controller.rb
```diff
@@ -96,7 +96,7 @@ def destroy
 
   def get_layout
     design_system_actions = %w[confirm_destroy]
-    design_system_actions += %w[index show features] if preview_design_system?(next_release: false)
+    design_system_actions += %w[index show features people update] if preview_design_system?(next_release: false)
```
davidgisbey: i suspect `update` probably shouldn't be in here as it's linked to the edit view (which also isn't in here)

app/views/admin/organisations/_people_order.html.erb
```diff
@@ -0,0 +1,37 @@
+<%= render "govuk_publishing_components/components/summary_list", {
+  title:,
+  edit: {
+    href: "#",
+    link_text: sanitize("Reorder <span class=\"govuk-visually-hidden\">#{title}</span>"),
```
davidgisbey: i think

```
sanitize("Reorder tag.span(title, class"govuk-visually-hidden")),
```

is slighty cleaner. Wdyt?

app/views/admin/people/_form.html.erb
```diff
@@ -106,6 +106,6 @@
        }
     } %>
 
-    <%= link_to "Cancel", admin_people_path, class: "govuk-link" %>
+    <%= link_to "Cancel", :back, class: "govuk-link govuk-link--no-visited-state" %>
```
davidgisbey, line 109: the reason i'd generally advise that we don't use the referrer to set the backlink is that it doesn't behave nicely when the page reloads after an unsuccessful save 

https://github.com/alphagov/whitehall/assets/42515961/cd6a0073-1e2f-402b-a1d2-54bcbabb5cac

I can completely see why we would want to implement something that can set the cancel link back to the people page since we link there from the summary card, but i think unfortunately we will have to do it in a more hands on way which we should handle in another card.

Previously we've done stuff like pass a query string like `return_to=path_to_return_to` then handle routing in the controller via before actions 

Something like

```
before_action : set_cancel_link

def set_cancel_link
 @cancel_link =  if params[:return_to] == "people"
                               admin_organisation_people_path(@organisation)
                            else
                              admin_people_path
                            end 
```

def worth doing but also a lot of work do i wouldn't worry about it for now

app/views/admin/organisations/_people_order.html.erb
```diff
@@ -0,0 +1,37 @@
...
+                           label: "Edit",
+                           href: edit_admin_person_path(organisation_role.role.current_person)
+                         },]
+             }]
+            else
+              [{
+                 key: "Occupant",
+                 value: "Vacant",
+               }]
+            end),
```
davidgisbey: This is using a splat operator which removes the hash from the array. i.e. *["a] == "a". 

It can also work inversely i.e. *"a" == ["a"]

I think you can probably simplify this a bit by just removing the array from the conditional and  removing the splat.

```
  rows: [
        if organisation_role.role.current_person
          {
            key: "Occupant",
            value: organisation_role.role.current_person_name,
            actions: [
              {
                label: "Edit",
                href: edit_admin_person_path(organisation_role.role.current_person)
              },
            ]
          }
        else
          {
            key: "Occupant",
            value: "Vacant",
          }
        end,
      ]
```

app/views/admin/organisations/people.html.erb
```diff
@@ -0,0 +1,46 @@
...
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% if @ministerial_organisation_roles.any? ||
+          @special_representative_organisation_roles.any? ||
+          @management_organisation_roles.any? ||
+          @traffic_commissioner_organisation_roles.any? ||
+          @chief_professional_officer_roles.any? ||
+          @military_organisation_roles.any?
```
davidgisbey: totally optional stylistic change 
```
  [
        @ministerial_organisation_roles,
        @special_representative_organisation_roles,
        @management_organisation_roles,
        @traffic_commissioner_organisation_roles,
        @chief_professional_officer_roles,
        @military_organisation_roles,
      ].select(&:any?)
```

app/views/admin/organisations/people.html.erb
```diff
@@ -0,0 +1,46 @@
...
+    %>
+      <%= render "govuk_publishing_components/components/tabs", {
+        tabs: [
+          *(people_tab(@ministerial_organisation_roles, "Ministers") if @ministerial_organisation_roles.any?),
+          *(people_tab(@special_representative_organisation_roles, "Special representatives") if @special_representative_organisation_roles.any?),
+          *(people_tab(@management_organisation_roles, "Management") if @management_organisation_roles.any?),
+          *(people_tab(@traffic_commissioner_organisation_roles, "Traffic commissioners") if @traffic_commissioner_organisation_roles.any?),
+          *(people_tab(@chief_professional_officer_roles, "Chief professional officers") if @chief_professional_officer_roles.any?),
+          *(people_tab(@military_organisation_roles, "Military") if @military_organisation_roles.any?),
```
davidgisbey: this is awesome 👍 absolutely love it

app/views/admin/organisations/_people_order.html.erb
```diff
@@ -1,10 +1,11 @@
 <%= render "govuk_publishing_components/components/summary_list", {
   title:,
-  edit: {
-    href: "#",
+  edit: { **(if @render_reorder
```
davidgisbey: I think we also need to add something about more than 1 role being present
```
if @render_reorder && organisation_roles.many?
```

image

davidgisbey: also this might be slightly easier to parse, totally optional as it's def working!

```
  edit: if @render_reorder && organisation_roles.many?
          {
            href: reorder_admin_organisation_people_path(@organisation, type:, title:),
            link_text: sanitize("Reorder <span class=\"govuk-visually-hidden\">#{title}</span>"),
            link_text_no_enhance: true
          }
        else
          {}
        end 
```

app/controllers/admin/organisations_controller.rb
```diff
@@ -94,9 +102,14 @@ def destroy
 
 private
 
+  def organisation_roles(type)
+    @organisation.organisation_roles.joins(:role)
+                 .merge(Role.send(type)).order(:ordering)
```
davidgisbey, line 83: we shouldn't really be using `send` as it doesn't respect an objects interface i.e. it will work with private methods.
https://stackoverflow.com/questions/30401970/what-is-the-difference-between-rubys-send-and-public-send-methods

we can use `public_send` instead

```
.merge(Role.public_send(type)).order(:ordering)
```

app/controllers/admin/organisation_people_controller.rb
```diff
@@ -0,0 +1,55 @@
+class Admin::OrganisationPeopleController < Admin::BaseController
```
davidgisbey, line 1: this is an excellent refactor guys

davidgisbey:
Really excellent work guys. I've left a few comments but nothing major. Mainly stylistic.

Really solid test coverage and implementation and love the refactor at the end 👍

app/controllers/admin/organisations_controller.rb
```diff
@@ -28,30 +28,6 @@ def show
     render_design_system(:show, :legacy_show)
   end
 
-  def people
```
davidgisbey, line 31: do we need to cleanup the get_layout method for these actions?

app/controllers/admin/organisations_controller.rb
davidgisbey: Just double checked and this is the case 

image

chriswhitehouse-tw: Nice, good catch ✅

app/views/admin/people/_form.html.erb
chriswhitehouse-tw: Cool, thanks @davidgisbey. I did wonder if the back link was a bit to simple to be true. We've implemented the conditional cancel link on params.

app/views/admin/organisations/_people_order.html.erb
chriswhitehouse-tw: Yeah, for sure. We've made this change. ✅

app/views/admin/organisations/_people_order.html.erb
chriswhitehouse-tw: Cool, thanks @davidgisbey. Yeah, got a bit mixed up with the necessity of the splat operator there.

app/views/admin/organisations/people.html.erb
chriswhitehouse-tw: Nice, yup. That's better syntax.

app/views/admin/organisations/people.html.erb
chriswhitehouse-tw: Ta!

app/controllers/admin/organisations_controller.rb
chriswhitehouse-tw: Cool, thank you. I hadn't appreciated that. Good to know!

app/views/admin/organisations/_people_order.html.erb
chriswhitehouse-tw: Cool, thanks 👍

davidgisbey (approved):
Amazing work again both!

app/views/admin/organisations/_people_order.html.erb
```diff
@@ -1,10 +1,14 @@
...
-  edit: {
-    href: "#",
-    link_text: sanitize("Reorder #{tag.span(title, class: "govuk-visually-hidden")}"),
-    link_text_no_enhance: true
-  }
+  edit: if @render_reorder && organisation_roles.many?
+          {
+            href: reorder_people_admin_organisation_path(@organisation, type:, title:),
+            link_text: sanitize("Reorder #{tag.span(title, class: "govuk-visually-hidden")}"),
+            link_text_no_enhance: true
```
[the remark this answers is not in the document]
chriswhitehouse-tw: Yes, good thought, had a play around with this, and it appears to be unnecessary.

PR #7901, opened 2023-07-04
293 add edit financial reports

davidtrussler:
[Trello](https://trello.com/c/0mZOqlro/293-add-edit-financial-reports)

This adds the New and Edit Financial reports pages (screenshots below). 
- Updates controller for design_system
- Creates new form_fields partial
- Updates UI for design_system for New & Edit pages
- Adds tracking
- Updates tests for design_system

|New page|Edit page|
|-|-|
|**Default**|**Default**|
|Screenshot 2023-07-05 at 11 02 51|Screenshot 2023-07-05 at 11 03 02|
|**With errors**|**With errors**|
|Screenshot 2023-07-07 at 12 52 05|Screenshot 2023-07-07 at 12 53 22|

### DIFF
--- a/app/models/financial_report.rb
+++ b/app/models/financial_report.rb
@@ -1,8 +1,10 @@
 class FinancialReport < ApplicationRecord
   belongs_to :organisation
-  validates_associated :organisation
 
-  validates :year, presence: true, uniqueness: { scope: :organisation_id }, numericality: { only_integer: true }
+  validates :organisation_id, presence: true
+  validates :year, presence: true
+  validates :year, uniqueness: { scope: :organisation_id }
+  validates :year, numericality: { only_integer: true, allow_blank: true }
   # We allow nil because data suggests some organisations are missing some data, 0 would be inaccurate in these cases
   validates :spending, numericality: { only_integer: true }, allow_nil: true
   validates :funding, numericality: { only_integer: true }, allow_nil: true
--- a/app/views/admin/financial_reports/edit.html.erb
+++ b/app/views/admin/financial_reports/edit.html.erb
@@ -1,20 +1,7 @@
-<% page_title @organisation.name + " financial reports" %>
-<div class="organisation-header">
-  <h1><%= @organisation.name %></h1>
-  <%= view_on_website_link_for @organisation %>
-</div>
+<% content_for :page_title, "Edit financial report for #{@organisation.name}" %>
+<% content_for :title, "Edit financial report" %>
+<% content_for :context, @organisation.name %>
+<% content_for :title_margin_bottom, 8 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @financial_report)) %>
 
-<%= tab_navigation_for(@organisation) do %>
-  <h2 class="add-bottom-margin">Updating the <%= @financial_report.year %> Financial Report from <%= @organisation.name %></h2>
-  <div class="row">
-    <div class="col-md-8">
-      <%= form_for @financial_report, url: {action: "update"}, html: {class: 'well'} do |f| %>
-        <%= f.errors %>
-        <%= f.text_field :year %>
-        <%= f.text_field :funding %>
-        <%= f.text_field :spending %>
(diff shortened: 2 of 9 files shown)


nnagewad:
Looking good. Just wondering if the year has no input, shall we just say that the year is blank. 🤔

davidtrussler:
> 

@nnagewad Hmm, I suspect that is harder to do than it seems. I think all error messages follow the same format where the error is a missing value, such as "<name of field> can't be blank". Not 100% sure about that though but I strongly suspect that's the case. We can enquire further if we want to make that change but if it's not a big deal I suggest we stick with it for now at least.

davidtrussler:
> > 
> 
> @nnagewad Hmm, I suspect that is harder to do than it seems. I think all error messages follow the same format where the error is a missing value, such as "<name of field> can't be blank". Not 100% sure about that though but I strongly suspect that's the case. We can enquire further if we want to make that change but if it's not a big deal I suggest we stick with it for now at least.

Oh actually I've found that is a fairly simple change so happy to do that.

davidtrussler:
> > > 
> > 
> > 
> > @nnagewad Hmm, I suspect that is harder to do than it seems. I think all error messages follow the same format where the error is a missing value, such as " can't be blank". Not 100% sure about that though but I strongly suspect that's the case. We can enquire further if we want to make that change but if it's not a big deal I suggest we stick with it for now at least.
> 
> Oh actually I've found that is a fairly simple change so happy to do that.

In fact I was misunderstanding the original request. I think to update the error messages displayed for the year to only display that it needs to be numeric if the filed is not blank is a worthwhile improvement but out of scope for this ticket. I will make a note of it for a future piece of work in the Trello card.

app/models/financial_report.rb
```diff
@@ -1,7 +1,7 @@
 class FinancialReport < ApplicationRecord
   belongs_to :organisation
-  validates_associated :organisation
 
+  validates :organisation_id, presence: true
   validates :year, presence: true, uniqueness: { scope: :organisation_id }, numericality: { only_integer: true }
```
davidgisbey, line 5: If we want to get rid of that extra validation we can use

```
  validates :year, presence: true, uniqueness: { scope: :organisation_id }, numericality: { only_integer: true, allow_blank: true }
```

app/views/admin/financial_reports/_form_fields.html.erb
```diff
@@ -0,0 +1,53 @@
...
+  error_items: errors_for(form.object.errors, :year)
+} %>
+
+<%= render "govuk_publishing_components/components/input", {
+  label: {
+    text: "Funding"
+  },
+  name: "financial_report[funding]",
+  id: "financial_report_funding",
+  value: form.object.funding,
```
davidgisbey, line 20: As discussed this will fix the weird behaviour with their input turning to 0 when the validation calls `.to_i` on the string
```
value: params.dig("financial_report", "year") || form.object.year,
```

app/views/admin/financial_reports/edit.html.erb
```diff
@@ -1,20 +1,13 @@
...
-        <%= f.errors %>
-        <%= f.text_field :year %>
-        <%= f.text_field :funding %>
-        <%= f.text_field :spending %>
-        <%= f.save_or_cancel cancel: admin_organisation_financial_reports_path(@organisation) %>
-      <% end %>
-    </div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @financial_report, url: {action: "update"} do |form| %>
```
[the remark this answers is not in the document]
davidtrussler: Ah great, thanks, I thought something like that should be doable.

app/views/admin/financial_reports/_form_fields.html.erb
davidtrussler: Great, that's a lot better. Cheers! 👍

davidtrussler:
> 

Actually, that has been done now - I'll delete the Trello card note.

davidgisbey (approved):
lgtm!

PR #7914, opened 2023-07-07
Fix rebuild taxonomy cache command

ryanb-gds:
The mlanett-redis-lock gem API has never, as far as I can see, accepted a duration for the life option. Without this fix, the command resulted in a `Unsupported command argument type: ActiveSupport::Duration` error

I briefly attempted writing a test for this but it seemed very high effort for uncertain value. I'm not sure how to mock the publishing API adapter for this case or if it would even be possible, and the classes that handle the rebuild have their own tests, so I've decided against it.


ryanb-gds:
Unsure about this as command appears to run in production

PR #7915, opened 2023-07-07
Call for Evidence email reminders

minhngocd:
After a Call For Evidence document is closed, at 8 weeks, we would like to send an email reminder to remind the user to publish the response. Since it is optional for Call For Evidence to have a response, we don't need to send anything after the initial reminder.

There is an existing daily job that runs and sends publisher notification emails for consultation reminders. We can reuse this to also send CFE Reminders. Cron job is set to run daily and config can be found for each environment in Helm Charts. For more info about emails, read the email_delivery.md docs

Guidance link is <Placeholder> for now while we are still working on it.

https://trello.com/c/VVJx9AY0/1313-implement-1-reminder-email-at-8-weeks-for-call-for-evidence

### DIFF
--- a/app/models/call_for_evidence_reminder.rb
+++ b/app/models/call_for_evidence_reminder.rb
@@ -1,0 +1,24 @@
+class CallForEvidenceReminder
+  WEEKS_FROM_CLOSE_TO_REMINDER = 8
+
+  class << self
+    def send_reminder
+      CallForEvidence.awaiting_response.closed_at_or_within_24_hours_of(WEEKS_FROM_CLOSE_TO_REMINDER.weeks.ago).each do |call_for_evidence|
+        log(call_for_evidence)
+        MultiNotifications.call_for_evidence_reminder(call_for_evidence).map(&:deliver_now)
+      end
+    end
+
+  private
+
+    def log(call_for_evidence)
+      Rails.logger.info("Sending reminder for call for evidence ##{call_for_evidence.id} '#{call_for_evidence.title}' to #{obfuscated_email_addresses(call_for_evidence)}")
+    end
+
+    def obfuscated_email_addresses(call_for_evidence)
+      call_for_evidence.authors.uniq.map { |author|
+        author.email.gsub(/^(.{2}).*(.{2})$/, '\1*****\2')
+      }.to_sentence
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


app/models/call_for_evidence_reminder.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    end
+
+  private
+
+    def log(call_for_evidence)
+      Rails.logger.info("Sending reminder for call for evidence ##{call_for_evidence.id} '#{call_for_evidence.title}' to #{obfuscated_email_addresses(call_for_evidence)}")
+    end
+
+    def obfuscated_email_addresses(call_for_evidence)
```
ryanb-gds, line 18: This doesn't really feel to me like something the reminder model should be responsible for, but given the Consultation reminder also is I think a separate PR would be appropriate if we want to invest in separating that responsibility

ryanb-gds (approved):
LGTM, notwithstanding the suggestion about the obfuscation responsibility

app/views/mail_notifications/call_for_evidence_reminder.text.erb
```diff
@@ -0,0 +1,6 @@
+This a reminder that the call for evidence "<%= @title %>" closed 8 weeks ago and you may want to publish responses. You can do this in the Outcome section.
+
+It is optional to publish responses for a call for evidence.
+
+Read how to write good call for evidences pages here:
+{placeholder for link}
```
ollietreend, line 6: A few (small!) things:

1. I think the first sentence is missing the word `"is"`.
2. Could we remove the entire reference to Call for Evidence guidance, since we don't have the link yet? So we don't accidentally launch with a `{placeholder}`. It seems like the email will still be just as useful without that placeholder reference for now.
3. It might be nice to add a link to the document, either on GOV.UK or in Whitehall. Other emails ([example](https://github.com/alphagov/whitehall/blob/0e0ffc8ec6b4667bacecb17bfe9bba55e3e2581f/app/views/mail_notifications/edition_rejected.text.erb), [example](https://github.com/alphagov/whitehall/blob/0e0ffc8ec6b4667bacecb17bfe9bba55e3e2581f/app/views/mail_notifications/edition_published.text.erb)) include a link to the document – although I realise the consultation ones don't ([example](https://github.com/alphagov/whitehall/blob/0e0ffc8ec6b4667bacecb17bfe9bba55e3e2581f/app/views/mail_notifications/consultation_deadline_passed.text.erb)), so it's probably fine to leave it as is.

```suggestion
This is a reminder that the call for evidence "<%= @title %>" closed 8 weeks ago and you may want to publish responses. You can do this in the Outcome section.

It is optional to publish responses for a call for evidence.
```

ollietreend (commented):
I've just added a tiny nitpick comment! But otherwise this looks great ✨

ollietreend (approved):
Looks good! 👍🏻 thanks @minhngocd

PR #7933, opened 2023-07-11
Remove usage of legacy url path from file attachment deletion

lauraghiorghisor-tw:
Following the [PR on file attachment updates](https://github.com/alphagov/whitehall/pull/7883), this piece of work handles the deletion of file attachments. 

There are a few different flows around asset deletion, all calling `AssetDeleter` one way or another. The scope for these changes is file attachments. Images and logos will be dealt with in future cards. For file attachments, the updates are triggered by the `MetadataWorker` -> `AttachmentDeleter` -> `AssetDeleter`.

The approach here is similar to the one in the updates PR, checking for the existence assets (the new `Asset` model) on the current `AttachmentData` to decide whether to use `asset_manager_id` (if assets are present) or `legacy_url_path`. 
The intention is to have the two flows coexist until the old one can be safely removed. The creation of new assets remains behind the `Use non legacy endpoints` user permission which is not yet enabled in production.

[Trello card](https://trello.com/c/AE0rfUJZ/96-story-use-assetmanagerid-for-deleting-a-file-attachment)

### DIFF
--- a/app/controllers/admin/promotional_feature_items_controller.rb
+++ b/app/controllers/admin/promotional_feature_items_controller.rb
@@ -32,7 +32,7 @@ class Admin::PromotionalFeatureItemsController < Admin::BaseController
       Whitehall::PublishingApi.republish_async(@organisation)
       if legacy_image_url.present?
         current_legacy_image_url = @promotional_feature_item.image.file&.instance_variable_get("@legacy_url_path")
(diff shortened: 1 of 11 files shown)


ollietreend (approved):
Looks good to me! 👍🏻

tunylund (approved):

PR #7941, opened 2023-07-13
Add additional world location information to /world content item

brucebolt:
We are replacing the Whitehall World Location API with the content item for `/world`.

Therefore adding the additional information needed that is currently supplied by the API.

Depends on https://github.com/alphagov/publishing-api/pull/2447.

After deployment, the following rake task will need to be run in production:
```
publishing_api:republish:republish_world_index
```

[Trello card](https://trello.com/c/g4ypLS8V)

### DIFF
--- a/test/unit/presenters/publishing_api/world_index_presenter_test.rb
+++ b/test/unit/presenters/publishing_api/world_index_presenter_test.rb
@@ -2,9 +2,30 @@ require "test_helper"
 
 class PublishingApi::WorldIndexPresenterTest < ActiveSupport::TestCase
   setup do
-    @world_location_1 = create(:world_location, name: "Location 1", slug: "location-1", active: true)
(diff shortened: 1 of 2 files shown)


test/unit/presenters/publishing_api/world_index_presenter_test.rb
```diff
@@ -2,9 +2,30 @@
...
+      analytics_identifier: "B2",
+      iso2: "CD",
+    )
+    @international_delegation = create(
+      :international_delegation,
+      name: "Delegation",
+      slug: "delegation",
+      active: true,
+      analytics_identifier: "C3",
+      iso2: nil,
```
brucebolt, line 27: I've left this value present, but `nil` rather than just not including the key, as this reflects the existing behaviour of the Whitehall API.

jkempster34 (approved):

PR #7946, opened 2023-07-13
Fix file attachments caching behaviour

lauraghiorghisor-tw:
When encountering a validation error on a file attachment, the current UI offers the option to also re-upload another file (or same file). When this happens, both the new file and the previous cached file get sent to the `AttachmentsController`, as part of the `attachment_data_attributes`.
Only the last file uploaded is associated with an `AttachmentData` and stored in Whitehall. Nonetheless, the Carrierwave flow gets triggered for the cached file as well, resulting in duplicated entries of the same `legacy_url_path` in Asset Manager.
Additionally, an `AttachmentDataNotFoundTransient` error is thrown by the `SetUploadedToWorker` which tries to find an `AttachmentData` with the legacy path of the cached file.

The changes in this commit remove this behaviour by dropping the `file_cache` param at the controller level, if a `file` has already been sent through.

### DIFF
--- a/app/controllers/admin/attachments_controller.rb
+++ b/app/controllers/admin/attachments_controller.rb
@@ -108,7 +108,7 @@ private
   end
 
   def attachment_params
-    params.fetch(:attachment, {}).permit(
+    attachment_params = params.fetch(:attachment, {}).permit(
       :title,
       :locale,
       :isbn,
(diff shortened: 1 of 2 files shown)


tunylund (approved):
nice one 👍

PR #7955, opened 2023-07-17
Update TagsComponent to include `Force scheduled`

davidgisbey:
## Description 

This was missed when the TagComponent was created which has led to the scheduled filter on the editions index/search page to look like it's returning force_published documents, when in reality they're force scheduled.

## Before 

image

## After

image

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -701,6 +701,10 @@ EXISTS (
     website_root + public_path(options)
   end
 
+  def force_scheduled?
+    force_published? && state == "scheduled"
+  end
+
 private
 
   def date_for_government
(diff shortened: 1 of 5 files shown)


app/models/edition.rb
```diff
@@ -701,6 +701,10 @@ def public_url(options = {})
     website_root + public_path(options)
   end
 
+  def force_scheduled?
```
ChrisBAshton, line 704: The commit references `force_published?` but should be `force_scheduled?`

ChrisBAshton (approved):
Nitpick below. Thanks for fixing the bug!

app/models/edition.rb
davidgisbey: Good spot! Thanks. Have sorted

PR #7961, opened 2023-07-18
Add Worldwide Organisation Details And History Pages  in the GOV.UK Design System

davidgisbey:
## Description 

This adds the Worldwide Org Details and History Pages in the GOV.UK Design System. It:

- prepends legacy to the boostrap implementation for the details page
- adds the GOV.UK Design System implementation
- handles routing in the get_layout method and show action
- adds a new histroy endpoint

I've chosen to duplicate the existing audit trail components rather than updating them to take a `design_system` argument and conditionally rendering the classes within them based on which design system it should use. I know there's a lot of duplication but it's far easier to remove if we do this and requires a tonne less testing with users being setup in various states.

I've  largely used the same design as we use for edition history for the history we present here too. With the exception being that we use our own generic pagination rather than older and newer links.

## Screenshots

image

image

## Trello card

https://trello.com/c/yo35eBiD/315-details-page
https://trello.com/c/kUAABVEh/347-dedicated-about-tab-for-world-organisation


nnagewad:
Looks good. We can remove the history tab :)

davidgisbey:
as discussed we're going to 

- move the history into the about tab
- use a large heading for details
- not render the body as govspeak and cap it at 160-200 chars

nnagewad:
DOPE!

farahTW:
LG

farahTW (approved):
LGTM

PR #7962, opened 2023-07-18
Refactor attachment updater

lauraghiorghisor-tw:
Now that the Asset model has been introduced to manage how whitehall talks to asset manager, it is also easier to group and fire updates.

Whereas before additional checks were required to ascertain which variants an attachment can potentially have, the Asset model now provides that out of the box. This enables cleaner logic around asset manipulation.

[Trello card](https://trello.com/c/IKTNJyCF/97-story-use-asset-manager-id-when-updating-attachment-states-from-a-rake-task)

### DIFF
--- a/app/services/asset_manager/attachment_updater.rb
+++ b/app/services/asset_manager/attachment_updater.rb
@@ -82,8 +82,6 @@ class AssetManager::AttachmentUpdater
       draft_edition.public_url(draft: true)
     elsif visible_edition.present?
       visible_edition.public_url
-    else
-      []
     end
   end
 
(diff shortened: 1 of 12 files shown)


app/services/asset_manager/attachment_updater.rb
```diff
@@ -9,15 +9,25 @@ def self.call(
   )
     return if attachment_data.deleted?
 
-    updates = []
+    if attachment_data.assets.empty?
```
kornysietsma, line 12: Please add a comment to clarify that this is a feature flag and the code will eventually be removed!

tunylund (approved):

PR #7978, opened 2023-07-20
Add Organisation About Page 

davidgisbey:
## Description

We're moving the about us information to it's own tab for organisations. This adds the tab to the helper method responsible for rendering the organisation tabs and then adds a new endpoint which handles rendering the about page

image

## Trello card

https://trello.com/c/HW4NnOjF/348-update-organisations-details-page-and-add-an-about-page

### DIFF
--- a/app/views/admin/organisations_about/show.html.erb
+++ b/app/views/admin/organisations_about/show.html.erb
@@ -1,0 +1,49 @@
+<% content_for :page_title, @organisation.name %>
+<% content_for :title, @organisation.name %>
+<% content_for :context, organisation_context_block(current_user, @organisation)  %>
+<% content_for :title_margin_bottom, 4 %>
+
+<p class="govuk-body">
+  <%= view_on_website_link_for @organisation, class: "govuk-link", target: "blank" %>
+</p>
+
+<%= render "components/secondary_navigation", {
+  aria_label: "Organisation navigation tabs",
+  items: secondary_navigation_tabs_items(@organisation, request.path)
+} %>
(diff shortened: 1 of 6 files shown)


davidtrussler:
@davidgisbey @nnagewad Just an observation when looking at this PR: if I go to the "About" tab of an organisation that doesn't have such a thing I get the message "No about us page.", which is fine but maybe it's not clear to a user in this scenario how they would create an "About Us" page. Should we think about adding a link to the page where this is done?

nnagewad:
@davidtrussler maybe that can be a separate ticket 👍🏽

app/views/admin/organisations_about/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+<% content_for :context, organisation_context_block(current_user, @organisation)  %>
+<% content_for :title_margin_bottom, 4 %>
+
+<p class="govuk-body">
+  <%= view_on_website_link_for @organisation, class: "govuk-link", target: "blank" %>
+</p>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "components/secondary_navigation", {
```
davidtrussler, line 12: The `secondary_navigation` component needs to be within a full-width container - in this case the tabs overflow the container.

davidgisbey: great catch have sorted thanks

davidtrussler (approved):
👍

PR #7985, opened 2023-07-20
Include the World Location(s) for a Worldwide Organisation when presenting to search

brucebolt:
This will allow us to use Search API to find the Worldwide Organisations that are associated with a specific World Location.

By doing this, we can then change the data source in `gds-api-adapters` meaning the Whitehall API can be deprecated.

[Trello card](https://trello.com/c/jJ0SH0N3)

### DIFF
--- a/app/models/worldwide_organisation.rb
+++ b/app/models/worldwide_organisation.rb
@@ -145,4 +145,8 @@ class WorldwideOrganisation < ApplicationRecord
 
     offices.each { |office| Whitehall::PublishingApi.republish_async(office) }
   end
+
+  def search_index
+    super.merge("world_locations" => world_locations.map(&:slug))
+  end
 end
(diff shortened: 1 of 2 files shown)


app/models/worldwide_organisation.rb
```diff
@@ -145,4 +145,8 @@ def republish_worldwide_offices
 
     offices.each { |office| Whitehall::PublishingApi.republish_async(office) }
   end
+
+  def search_index
+    super.merge("world_locations" => world_locations.map(&:slug))
```
CristinaRO, line 150: What happens when a world location has its slug changed? Does everything search-related get updated automatically?

brucebolt: Unfortunately at some point in the past, slug was chosen as the way of linking to world locations in Search API.  This means some manual work is required if a location's slug changes.  We have that [documented](https://docs.publishing.service.gov.uk/manual/rename-a-country.html#6-remove-duplicate-search-results) as a step in the "Rename a country" playbook.

JonathanHallam (approved):
👍

PR #7986, opened 2023-07-20
Add Worldwide Organisation About Page

davidgisbey:
## Description

We're moving the about us information to it's own tab for worldwide organisations. This adds the tab to the helper method responsible for rendering the worldwide organisation tabs and then adds a new endpoint.

image

## Trello card

https://trello.com/c/HW4NnOjF/348-update-organisations-details-page-and-add-an-about-page

### DIFF
--- a/app/controllers/admin/worldwide_organisations_about_controller.rb
+++ b/app/controllers/admin/worldwide_organisations_about_controller.rb
@@ -1,0 +1,7 @@
+class Admin::WorldwideOrganisationsAboutController < Admin::BaseController
+  layout "design_system"
+
+  def show
+    @worldwide_organisation = WorldwideOrganisation.friendly.find(params[:id])
+  end
+end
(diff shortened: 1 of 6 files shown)


nnagewad:
DOPE!

davidtrussler:
@nnagewad It's a bit picky perhaps but the text in the "Body" section here has a slightly annoying top margin that pushes the text below the work "Body" in a way that doesn't happen in the "Summary" section. I think it would need a general solution to the component itself rather than being specific to this piece of work. @davidgisbey

davidgisbey:
ah right i see what's happening. the `simple_format` method creates a para for it which adds margin to the top by default. I think we can override it by passing in `simple_format(truncate(@organisation.body, length: 500), class: "govuk-!-margin-top-0")`

Great spot will push up the fix.

davidtrussler (approved):
LGTM! 👍

PR #8002, opened 2023-07-24
Sort World Locations in world index content item

brucebolt:
The pages consuming this data are expecting the world locations to be in alphabetical order, excluding special characters.
    
Therefore updating the presenter to sort the locations in the content item.

Also adds missing tests to the `WorldLocationHelper`.  I decided not to spend time DRY-ing up the `group_and_sort` method since this will be deleted when public rendering of the Whitehall API is switched off.

After deployment, the world index page will need to be republished.

[Trello card](https://trello.com/c/eKjf7gPC)

### DIFF
--- a/test/unit/app/helpers/world_location_helper_test.rb
+++ b/test/unit/app/helpers/world_location_helper_test.rb
@@ -58,4 +58,45 @@ class WorldLocationHelperTest < ActionView::TestCase
       assert_equal expected_order, group_and_sort(unsorted)
     end
   end
+
+  context "#sort" do
+    test "sorts correctly where special characters are used anywhere other than the first character" do
+      world_location_1 = create(:world_location, name: "Cow")
+      world_location_2 = create(:world_location, name: "Côt")
+      unsorted = [world_location_1, world_location_2]
+
+      expected_order = [
+        world_location_2,
+        world_location_1,
+      ]
+
+      assert_equal expected_order, sort(unsorted)
+    end
+
+    test "sorts correctly where special characters are used as the first character" do
+      world_location_1 = create(:world_location, name: "England")
+      world_location_2 = create(:world_location, name: "Éire")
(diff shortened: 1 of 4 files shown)


test/unit/app/helpers/world_location_helper_test.rb
```diff
@@ -0,0 +1,102 @@
...
+        [
+          world_location_2,
+          world_location_1,
+        ],
+      ]]
+
+      assert_equal expected_order, group_and_sort(unsorted)
+    end
+
+    test "ignores The prefix when sorting" do
```
jkempster34, line 40: Can we make this name a bit clearer, I only understand it because I know what it means. Maybe just some backticks: "ignores `The` prefix when sorting"

test/unit/app/helpers/world_location_helper_test.rb
```diff
@@ -0,0 +1,102 @@
...
+      world_location_2 = create(:world_location, name: "The Gambia")
+      unsorted = [world_location_1, world_location_2]
+
+      expected_order = [[
+        "G",
+        [
+          world_location_2,
+        ],
+      ],
+                        [
```
jkempster34, line 51: This indentation is a bit weird. Has Rubocop forced this, or can it be made to be in line with the array above?

jkempster34 (approved):
Two very minor comments, looks great otherwise 👍

test/unit/app/helpers/world_location_helper_test.rb
brucebolt: Unfortunately this was done by rubocop.  Lining it up meant we had lint failures.

test/unit/app/helpers/world_location_helper_test.rb
brucebolt: Good point.  I've updated this.

PR #8005, opened 2023-07-25
Add New and Edit Organisation to GOV.UK Design System

davidgisbey:
## Description

This adds the New and Edit Org pages in the GOV.UK Design System.

It's a bit of a giant page so apologies it's going to be quite chunky to review.

The JS on the page didn't actually work properly, but the correct behaviour is.

- custom logo file input appears if user selects custom logo in the logo crest field
- `non_departmental_public_body` fields show when a user selects an org type where `non_departmental_public_body` is `true` in the org model
- We've used radios in a few places that used checkboxes before

### Without custom logo and not a `non_departmental_public_body`

image

image

### With custom logo and a `non_departmental_public_body`

image

image

image

## Trello card

https://trello.com/c/1Q5wzkjC/377-new-edit-organisation-page-dev


davidgisbey:
Nik, there's currently 60px for between the last field and button in the reusable partial we use.

image

 Is this correct? I can see why there's larger spacing for a slightly smaller page like this, but it looks slightly off in the screenshots above

image

It also uses `Add another featured link` on other pages. I'm guessing we want to have the same text for all of them. Let me know what you'd like to use

We can have different spacing on the different pages, but it'll require a little more work. Maybe we want the spacing and text should probably be the same on all the pages though as they're essentially the same fields? Happy to do whatever though

nnagewad:
Hey @davidgisbey, So the design intends to have a horizontal rule between the last item and the add button. To be honest, I would like to revisit this component as a whole because I do find Bootstrap version is more elegant and clear that there is a dedicated block.

Happy to discuss more IRL.

davidgisbey:
Okay cool shall we add that as a success criteria for the card to fix this up then and just crack on as is for now then? Maybe we can pick it up next sprint.

nnagewad:
> Okay cool shall we add that as a success criteria for the card to fix this up then and just crack on as is for now then? Maybe we can pick it up next sprint.

Totally. I can whip something up. Run it by you and then share with Denis/Murilo/Mike

nnagewad:
Created a ticket in the backlog for add/part: https://trello.com/c/GO9WScmQ

RodneyJohnsonGDS:
Hi @davidgisbey I have reviewed the PR and found really good test coverage, particularly including tests around the JS module in the Organisation Form. Relevant tests are all passing including the legacy tests too. Nice, clean and efficient work done here - happy to approve!

PR #8008, opened 2023-07-25
Remove unused asset manager rake tasks

lauraghiorghisor-tw:
These rake tasks were used for a one-off migration, to populate `auth_bypass_ids` on assets when the field was first introduced. No longer needed.

[Trello card](https://trello.com/c/IKTNJyCF/97-story-use-asset-manager-id-when-updating-attachment-states-from-a-rake-task)


lauraghiorghisor-tw:
@davidgisbey Hey David, I see you were the last one to touch this area. We changed the worker as it was using the `AssetUpdater` we had business with. Tests were fine - passing hash with symbols, expecting hash with symbols - but the rake task broke when manually testing (see the `ArgumentError` in commit message). 

Thought it would be good to have you look at the changes. You may also have some insight into wether we need to maintain these rake tasks since it seems they haven't been run in some time (they would have errored the same prior to our changes).  We are moving towards using non-legacy endpoints for `whitehall` as well, it seems all the `auth_bypass_ids` values are being set just fine in `asset-manager`. Perhaps there is not a business need here anymore?

The strict args are set in the `Sidekiq` initializer. I am also wondering if there is anything we can do to make it break the tests so this situation doesn't happen again.

kornysietsma (approved):
Looks fine to me - agree it might be good to check with David

davidgisbey:
Great catch. Code looks good to me. I suspect these can probably be removed i think they were used once to populate the auth_bypass_id when it was added. Might be worth double checking with a member of the Experience team though

lauraghiorghisor-tw:
> Great catch. Code looks good to me. I suspect these can probably be removed i think they were used once to populate the auth_bypass_id when it was initially added. Might be worth double checking with a member of the Experience team though

@ollietreend Dragging you in to have a look at these rake tasks, whether we need them. 
The worker itself (which is so beautifully reusable) is something I'd keep around and maybe use it further down the line if we bring both File and Image Attachments flows to some level of consistency.

ollietreend:
Hi @lauraghiorghisor-tw 

Both `AssetManagerUpdateWhitehallAssetWorker` and the `asset_manager:*` Rake tasks were implemented in #6570 to support the introduction of `auth_bypass_ids` on assets. This was required as part of an "epic" to introduce the "shareable previews" feature to Whitehall.

As I think you & David have already concluded, these Rake tasks would have been used for a **one-off migration** when the "shareable previews" feature was initially launched. I don't believe it's been used (or needed) since then, so we probably should have cleared it up at the time.

They would have been run over a year ago now, long before "strict args" were required in the app. So as you've already said, the error you're seeing would not have happened at the time.

- **If these Rake tasks are useful to you:** you'll need to adapt them to use strict args (as you've already done in this PR) 😄
- **If you don't need them:** please do feel free to delete them rather than fix them 🗑️

Either way you'll be improving the codebase! ✨🧹

As for the tests not picking up on the Sidekiq strict args requirement, that does seem a bit annoying. Have you tried adding `Sidekiq.strict_args!` to the [test/support/sidekiq_test_helpers.rb](https://github.com/alphagov/whitehall/blob/2b90ee26775ea4484312ce9d6c9201a8a509d9c3/test/support/sidekiq_test_helpers.rb) file? It looks like that is included in both the MiniTest and Cucumber test suites, and hopefully will configure Sidekiq so it behaves the same as in the 'real world'.

lauraghiorghisor-tw:
Thanks for confirming we can get rid of them @ollietreend 🙏🏻

Will have a play around with the `strict_args` 🛝

PR #8010, opened 2023-07-25
Allow AssetManagerCreateWhitehallAssetWorker to accept temporary asset related params

syed-ali-tw:
These changes are fix for production issue:
https://govuk.sentry.io/issues/4336826185/?project=202259&query=&referrer=issue-stream&statsPeriod=14d&stream_index=3 
this issue was introduced when the logic for saving assets was moved to AssetManagerCreateAssetWorker and the number of argument that was passed to the perform method inside AssetManagerCreateWhitehallAssetWorker was reduced from 8 to 6.
The changes introduced are temporary to be able to execute one job that is currently failing and retrying.

build that caused the issue: https://github.com/alphagov/whitehall/actions/runs/5647571117/job/15297962359

### DIFF
--- a/app/workers/asset_manager_create_whitehall_asset_worker.rb
+++ b/app/workers/asset_manager_create_whitehall_asset_worker.rb
@@ -3,7 +3,29 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
 
   sidekiq_options queue: "asset_manager"
 
-  def perform(file_path, legacy_url_path, draft = false, attachable_model_class = nil, attachable_model_id = nil, auth_bypass_ids = [])
+  WITH_TEMPORARY_ASSET_RELATED_PARAMS = 8
+  EXPECTED_PARAMS = 6
+
+  def perform(*args)
+    # We eventually only run the code that is inside "perform_with_assets_params"
+    # these changes are introduced to fix a prod issue that was caused by change in
+    # arguments( decreased from 8 to 6 ) that were accepted by perform method
+    # these cause one of the job that was about to execute just before application deployment
+    # to fail. the changes are temporary.
+    if args.length == EXPECTED_PARAMS
+      perform_without_assets_params(*args)
+    elsif args.length == WITH_TEMPORARY_ASSET_RELATED_PARAMS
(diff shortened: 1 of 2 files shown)


app/workers/asset_manager_create_whitehall_asset_worker.rb
```diff
@@ -3,7 +3,26 @@ class AssetManagerCreateWhitehallAssetWorker < WorkerBase
...
-  def perform(file_path, legacy_url_path, draft = false, attachable_model_class = nil, attachable_model_id = nil, auth_bypass_ids = [])
+  WITH_TEMPORARY_ASSET_RELATED_PARAMS = 8
+  def perform(*args)
+    # We eventually only run the code thats inside "perform_with_assets_params"
+    # these changes are introduced to fix a prod issue that was caused by change in
+    # arguments( decreased from 8 to 6 ) that were accepted by perform method
+    # these cause one of the job that was about to execute just before application deployment
+    # to fail. the changes are temporary.
+
+    if args.length < WITH_TEMPORARY_ASSET_RELATED_PARAMS
```
kornysietsma, line 14: Minor thing - I'd prefer to have explicit equality checks, something like
```
if args.length == EXPECTED_PARAMS
...
else if args.length == WITH_TEMPORARY_ASSET_RELATED_PARMS
...
else raise "invalid parameter count in sidekiq job"
```
I guess that might be tricky with optional parameters though.  But I wonder - why do we have optional parameters? Are they ever not passed???

anyway, as a temporary change that will last only days, this isn't vital.

kornysietsma (approved):
Looks fine to me, though see comment for one optional improvement.

app/workers/asset_manager_create_whitehall_asset_worker.rb
syed-ali-tw: incorporated these changes. having explicit equality checks made more sense.

kornysietsma (approved):
Looks good. We still need to revisit what "good" sidekiq job parameters look like, but in later stories - this is fine to get the prod fix in.

PR #8015, opened 2023-07-27
Update minitest -> v15.19.0 & mocha -> v2.1.0

floehopper:
In minitest v15.19.0, the very old `MiniTest` module name has been dropped, so we need to use the `Minitest` module name everywhere. This also requires us to upgrade mocha to v2.1.0. See https://github.com/freerange/mocha/issues/614 for more details.


kevindew (approved):
Looks great - thanks for doing this so fast.

PR #8020, opened 2023-07-27
Make the Audit Trail thread-safe

ollietreend:
This PR fixes a thread-safety bug which has resulted in user actions being misattributed in document history timelines.

The class attribute `AuditTrail.whodunnit` was used to hold the current user whose request is being served by the app. This class attribute value is shared with all Threads on the Process which the app runs in.

This never used to be an issue because the web server used to serve requests via multiple single-threaded processes. So there was no risk of the variable being shared between multiple requests. However this recently changed so Whitehall is now served in a multi-threaded environment. This has resulted in multiple Threads having shared access to the same `AuditTrail.whodunnit` variable.

When multiple HTTP requests were served in parallel by the same Process (in separate Threads), the `AuditTrail.whodunnit` variable would sometimes be written by one Thread and then overwritten by the second Thread before Audit Trail records were written. The result was that the wrong user would sometimes be attributed to a document change.

In this commit, I've introduced a new `Current` model which has a thread-safe `user` attribute. This has been achieved with the help of the Rails `ActiveSupport::CurrentAttributes` super class, which ensures that attribute values are thread-isolated and reset automatically before and after each request.

So now the current user will be available at `Current.user` rather than `AuditTrail.whodunnit` throughout the app.

## Race condition test

A test has been written which demonstrates the thread-safety bug.

```
bundle exec rails test test/integration/race_condition_test.rb
```

This test originally failed, and now it passes.

I'm undecided whether or not this test is valuable to keep long-term. It's pretty heavy-weight because it includes `sleep` statements to reproduce the desired environmental conditions within two threads. So it'll add a good couple of seconds to the overall test suite run time.

---


beccapearce:
This looks good to me and thank you for the commit messages, makes it much easier to follow.

Regarding the test, I would kind of like it kept in for a short about of time just to protect against the issue occurring again. But I'd also be ok with it being removed as a separate commit so there is a record of it in the PR.

beccapearce (approved):
Happy to approve, keeping in mind my comment about the test :)

ollietreend:
I've merged this keeping the integration test in place.

The previous implementation of "whodunnit" hadn't really changed in about 10 years – so I expect this implementation will be equally unlikely to change, meaning there's probably a low risk of regression. So at some point in the future, if anybody cares and remembers about it, it _should be_ safe to remove that test.

PR #8026, opened 2023-07-28
Sidekiq Web UI integrated authenticated behind signon

kevindew:
This is a proof of concept to illustrate an approach we could use to integrate the Sidekiq Web user interface into GOV.UK apps, now that Sidekiq Monitoring [1] is no longer operational. It was put together after a chat with Jon Hallam about what a future without Sidekiq Monitoring might be. I'd love to receive some feedback on if people think we should go down this route.

This approach offers a number of advantages to the Sidekiq Monitoring set-up of before. This one has authentication, using Signon, where users require a "Sidekiq Admin" permission to access and thus won't require the VPN for access. It will allow post requests and thus users are able to interact with the admin interface to delete dead jobs etc. It also uses the Sidekiq code of Whitehall and won't have the issue that Sidekiq Monitoring code doesn't necessarily match the version of Sidekiq used in an app.

---

Example of successful login:
Jul-28-2023 18-13-37

Example of unsuccessful login:
Jul-28-2023 18-13-57

---

I don't expect this think to be a prototype that other apps should copy. As Whitehall has frontend rendering there's a couple of extra esoteric bits of code (Slimmer and Admin host) that mean this has some app specific code. Instead I'd imagine it'd make sense if govuk_sidekiq was able to provide a class that apps could use to mount a GDS SSO ready middleware to avoid each app repeating logic. However prior to merging I expect we should move the middleware to a separate file in lib and add some tests.

There's a couple of other problems that a wider roll out of this approach would have to resolve:

1) apps that are only on the VPN without a public hostname - we'd need
   to establish a way to route a request on a public hostname to them to
   expose the Sidekiq endpoint
2) API only apps - the GDS SSO gem doesn't support session
   authentication for Rails apps that are `api_only` [2] as it didn't
   need to and that provided better errors. We may need to work out a
   way that can work with this.

Finally there's a question still on approach for testing, I'm not sure if apps should have a test for this in them (though if we have a middleware class in govuk_sidekiq that should be unit tested). I do wonder if it'd make sense to have Smokey check access to Sidekiq.

[1]: https://github.com/alphagov/sidekiq-monitoring
[2]: https://github.com/search?q=repo%3Aalphagov%2Fgds-sso%20api_only&type=code

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,5 @@
+require "sidekiq/web"
+
 class AdminRequest
   def self.matches?(request)
     # Allow access to all routes in development, and restrict to the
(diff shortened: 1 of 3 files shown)


ollietreend (approved):
This looks great - thanks for doing this @kevindew 🚀

PR #8037, opened 2023-08-01
Publish worldwide corporate information about pages as redirects

jkempster34:
https://trello.com/c/En4l42d7

Worldwide corporate information about pages aren't rendered, instead they are
surfaced in the body of the owning worldwide organisation.

There have been various attempts to devise a strategy for these pages.
Currently, they are published with the `schema_name` of
`placeholder_corporate_information_page` at the path
`world/organisation/:organisation_id/about/about`.

Unfortunately, anywhere in Whitehall that links to the base path of these pages
now 404s. Instead, we should redirect to the worldwide organisation itself.

This:
- Updates the base path of worldwide corporate
  information pages to
  `world/organisation/:organisation_id/about` to
  match the non-worldwide
  organisation corporate information pages.
- Adds a generic presenter for a redirect.
- Publishes worldwide corporate information pages as
  redirects.

This means any worldwide corporate information about pages will now live at
`world/organisation/:organisation_id/about`. When we republish these content
items, Pulishing API will create additional redirect from each
`world/organisation/:organisation_id/about/about` to
`world/organisation/:organisation_id/about` as well.

---

### DIFF
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -110,6 +110,12 @@ class CorporateInformationPage < Edition
     organisations
   end
 
+  def api_presenter_redirect_to
+    raise "only worldwide about pages should redirect" unless about_page? && worldwide_organisation.present?
+
+    worldwide_organisation.public_path(locale: I18n.locale)
+  end
+
   def self.for_slug(slug)
     if (type = CorporateInformationPageType.find(slug))
       find_by(corporate_information_page_type_id: type.id)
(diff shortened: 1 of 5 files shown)


app/models/corporate_information_page.rb
```diff
@@ -175,8 +181,7 @@ def base_path
     return if owning_organisation.blank?
 
     url = owning_organisation.base_path + "/about/#{slug}"
-    url.gsub!("/about/about", "/about") if organisation.present?
-    url
+    url.gsub("/about/about", "/about")
```
brucebolt, line 184: I wonder if this entire method could now be cleaned up, to remove the confusing gsub?

E.g.

```
    return if owning_organisation.blank?

    url = if slug == "about"
                owning_organisation.base_path + "/about"
              else
                owning_organisation.base_path + "/about/#{slug}"
              end
```

app/presenters/publishing_api/redirect_presenter.rb
```diff
@@ -0,0 +1,39 @@
...
+    def initialize(item, update_type: nil)
+      self.item = item
+      self.update_type = update_type || "major"
+    end
+
+    delegate :content_id, to: :item
+
+    def content
+      {
+        title: nil,
```
brucebolt, line 14: Do we need to specify a nil title, as it is forbidden in [the content schema](https://github.com/alphagov/publishing-api/blob/b0cf0ebd9c25b1bb918fb9491c93a3118fb2599e/content_schemas/formats/redirect.jsonnet#L7)?

app/models/corporate_information_page.rb
jkempster34: Done 👍

brucebolt (approved):

PR #8046, opened 2023-08-02
Setup Google Tag Manager

dnkrj:

# What
Add code for Google Tag Manager

https://trello.com/c/EDrVRCsv/1515-set-up-ga4-property-and-gtm-container

# Why
Needed as part of the process of moving to GA4

### DIFF
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -9,6 +9,13 @@
 <% end %>
 
 <% content_for :head do %>
+  <% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+    <%= render "govuk_publishing_components/components/google_tag_manager_script", {
+      gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
+      gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
+      gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"]
+    } %>
+  <% end %>
   <%= stylesheet_link_tag "admin_legacy.css", integrity: false %>
   <%= csrf_meta_tags %>
 <% end %>
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -1,5 +1,15 @@
 <% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
 
+<% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+  <% content_for :head do %>
+    <%= render "govuk_publishing_components/components/google_tag_manager_script", {
+      gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
+      gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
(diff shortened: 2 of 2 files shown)


app/views/layouts/admin.html.erb
```diff
@@ -9,10 +9,20 @@
 <% end %>
 
 <% content_for :head do %>
+  <%= render "govuk_publishing_components/components/google_tag_manager_script", {
+    gtm_id: "GTM-P93SHJ4Z"
```
kevindew, line 13: It'd be good if we could set this with an environment variable - then we'd not have the analytics running in people's local environments and could, if we chose to, have different analytics for non-prod environments like staging and integration to not pollute prod data (would save a whole bunch of filtering).

Would also mean we don't need to duplicate the key as well in separate files.

kevindew: Here's an example from content publisher code: https://github.com/alphagov/content-publisher/blob/6fb72ab4038ae3150a777364ed51dfeadeaca47f/app/views/layouts/application.html.erb#L16-L22

One of the interesting things from Content Publisher, that you may also have to face and potentially resolve is also that some datalayer stuff is configured before the component (which seems a bit in-elegant): https://github.com/alphagov/content-publisher/blob/6fb72ab4038ae3150a777364ed51dfeadeaca47f/app/views/layouts/application.html.erb#L7-L15 - which if we have to do similar we might want to extend the component.

app/views/layouts/admin.html.erb
```diff
@@ -9,10 +9,20 @@
...
   <%= stylesheet_link_tag "admin_legacy.css", integrity: false %>
   <%= csrf_meta_tags %>
 <% end %>
 
+<% content_for :navbar do %>
+  <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-P93SHJ4Z"
+    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+  <!-- End Google Tag Manager (noscript) -->
+<% end %>
```
kevindew, line 24: If this is something we should track for Whitehall it's probably something we should do for other apps and have a component. I was going to suggest it should be part of the tag manager component, but I see about the difference in head or not. I do wonder though - we probably could just have the whole component at the start of the body element I guess and also include noscript.

Also as discussed face-to-face, I did wonder whether this noscript approach was something that was potentially an option as a way to help @patrickpatrickpatrick's [RFC](https://github.com/alphagov/govuk-rfcs/pull/160) where it suggests using ES5 for IE to still get analytics but it'd be so much better if they just got a lo-fi analytics approach like this. Though do wonder about cookie consent?

app/views/layouts/admin.html.erb
dnkrj: Good point! Have updated to match content publisher and added a PR to set the environmental variable
https://github.com/alphagov/govuk-helm-charts/pull/1243

app/views/layouts/admin.html.erb
dnkrj: Yes, have removed this for now as it's not needed for our initial testing but will consider the approach in future.

ollietreend (approved):
Looks good to me 👍🏻

PR #8047, opened 2023-08-02
Stop duplicate review reminders being sent

ollietreend:
We've observed a bug in the Review Reminder feature which resulted in duplicate reminder emails being repeatedly sent to users.

### What was going wrong

The `ReviewReminderNotifierWorker` sends an email and then updates the `ReviewReminder` record to have a `reminder_sent_at` timestamp. But there's a validation rule that ensures the `review_at` date cannot be in the past, which meant the `update!` call failed with an exception and resulted in the Sidekiq job being put into the retry queue.

In turn, this meant we were sending the same reminder to users multiple times while the Sidekiq job retried (and continually failed).

### How this PR fixes it

This PR updates the worker test to expose the bug. I've then done the minimum work required to get the test passing.

Ideally I'd like to follow up with another PR to refactor some of this code – including changing the validation rules so we don't need to bypass them in this scenario.

Trello: https://trello.com/c/c5D5qVcL/1552-stop-duplicate-review-reminders-being-sent

---

### DIFF
--- a/app/workers/review_reminder_notifier_worker.rb
+++ b/app/workers/review_reminder_notifier_worker.rb
@@ -12,7 +12,7 @@ class ReviewReminderNotifierWorker < WorkerBase
         recipient_address: email_address,
       ).deliver_now
 
-      review_reminder.update!(reminder_sent_at: Time.zone.now)
+      review_reminder.update_columns(reminder_sent_at: Time.zone.now)
     end
   end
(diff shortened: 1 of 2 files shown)


beccapearce (approved):
Looks good to me! 

Let's make sure we make a card for the follow up tidying that needs to be done :)

PR #8058, opened 2023-08-04
Fix broken draft previews for corporate information pages

callumknights:
Work to move rendering out of whitehall unintentionally broke draft previews for corporate
information pages. This PR fixes the issue by allowing the publication of draft 'about_us'
bodies to publishing-api.

WorldwideOrganisations not being editionable doesn't allow them to have a draft workflow that uses normal publishing methods, so the `put_content` solution is necessary to prevent publishing-api defaulting to live content. Making WorldwideOrganisations editionable would be the ideal solution in the future, but is a lot of work and far out of the scope of this bugfix. 

Trello: https://trello.com/c/dlZFPXcK/749-fix-the-broken-previews-for-corporate-information-pages

### DIFF
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -33,11 +33,20 @@ class CorporateInformationPage < Edition
   scope :accessible_documents_policy, -> { where(corporate_information_page_type_id: CorporateInformationPageType::AccessibleDocumentsPolicy.id) }
 
   def republish_owning_organisation_to_publishing_api
-    Whitehall::PublishingApi.republish_async(owning_organisation) if owning_organisation.present?
+    if state == "draft" && owning_organisation.is_a?(WorldwideOrganisation)
+      presenter = PublishingApi::WorldwideOrganisationPresenter.new(owning_organisation, state: "draft")
+      Services.publishing_api.put_content(presenter.content_id, presenter.content)
+    elsif owning_organisation.present?
+      Whitehall::PublishingApi.republish_async(owning_organisation)
+    end
   end
 
   def republish_about_page_to_publishing_api
-    about_us = owning_organisation&.about_us
+    about_us = if state == "draft"
+                 owning_organisation&.draft_about_us
+               else
+                 owning_organisation&.about_us
+               end
     return unless about_us
 
     PublishingApiDocumentRepublishingWorker.perform_async_in_queue(
--- a/app/models/has_corporate_information_pages.rb
+++ b/app/models/has_corporate_information_pages.rb
@@ -12,6 +12,10 @@ module HasCorporateInformationPages
     about_us.summary if about_us.present?
   end
 
+  def draft_summary
+    draft_about_us.summary if draft_about_us.present?
+  end
+
   def body
     about_us.body if about_us.present?
   end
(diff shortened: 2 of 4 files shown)


app/models/has_corporate_information_pages.rb
```diff
@@ -31,10 +31,10 @@ def published_corporate_information_pages
   end
 
   def about_us
-    @about_us ||= corporate_information_pages.published.for_slug("about")
+    corporate_information_pages.published.for_slug("about")
   end
 
   def draft_about_us
-    @draft_about_us ||= corporate_information_pages.draft.for_slug("about")
+    corporate_information_pages.draft.for_slug("about")
```
CristinaRO, line 38: Do the `summary` and the `body` (which are defined based upon the `about_us`) need to be draft-aware as well?

callumknights: I don't believe `body` does, in fact I'm not 100% sure it's used at all. But you are correct about the `summary` that's still broken, more work to do 🙃

callumknights: This is fixed now Cristina :) I've not touched the potentially unused `body` method, as I think refactoring that is out of scope of this card.

app/models/corporate_information_page.rb
```diff
@@ -33,11 +33,20 @@ def process_associations_before_save(new_edition)
   scope :accessible_documents_policy, -> { where(corporate_information_page_type_id: CorporateInformationPageType::AccessibleDocumentsPolicy.id) }
 
   def republish_owning_organisation_to_publishing_api
-    Whitehall::PublishingApi.republish_async(owning_organisation) if owning_organisation.present?
+    if state == "draft" && owning_organisation.is_a?(WorldwideOrganisation)
+      presenter = PublishingApi::WorldwideOrganisationPresenter.new(owning_organisation, state: "draft")
+      Services.publishing_api.put_content(presenter.content_id, presenter.content)
```
CristinaRO, line 38: It makes me confused that the path for dealing with the special case requires knowledge of the internals of the publishing API process, and manipulating content directly through those internals.

Whereas the "default" case has a neatly packaged method (`republish_async`).

Is there no way to use the more "typical" endpoints of the publishing mechanism? It could be that the answer is "no", or "not without significant refactoring", and that's fair, but I'm wary of coding ourselves into deeper and deeper exception paths and making any bugs harder to spot and diagnose.

app/models/has_corporate_information_pages.rb
CristinaRO: Thanks!

I don't suppose there's a way to make the entity that `has_corporate_information_pages` aware of `state`, such as `about_us` and `draft_about_us` become one method:
```
def about_us
  corporate_information_pages.send(state.to_sym).for_slug("about")
end
```
or similar?

app/models/corporate_information_page.rb
callumknights: I agree 100%, and I don't like the fact that I'm depending on the institutional knowledge Bruce has of these publishing internals to know the solution to this problem. Unfortunately WorldwideOrganisations being non-editionable, but also having a `draft` state is an exception to the general rule of how whitehall content is published. 

Ultimately the solution would be to make WorldwideOrganisations editionable, and then draft editions can be published normally, however that's a significant piece of work and definitely outside the scope of this bugfix.

CristinaRO: Would you be willing to include this context in the commit message, to communicate the background to future devs?

callumknights: Sure, I'll amend it now :)

CristinaRO (approved):
Thank you!

PR #8064, opened 2023-08-07
Fix review reminder timezone issue

ollietreend:
**Note:** Changes in this PR are best reviewed commit by commit.

## Context

Document ‘review at’ dates were being stored as DateTime values, which intrinsically come with a timezone in Rails. However in the UI they’re only able to choose dates, not times.

The database values were being stored in UTC meaning the values entered by users were being wound back 1 hour from BST (current timezone) to UTC. That means a user supplied date of “07/08/2023” was represented as the DateTime “07/08/2023 00:00:00 BST” which got stored in the database as “06/08/2023 23:00:00 UTC”. Notice the date actually changed because of this automatic timezone conversion.

## What this PR changes

This PR makes a few changes to resolve the timezone issue and improve code readability:

1. **Converts `review_at` to Date rather than a DateTime.** This helps avoid timezone confusion because dates do not have timezones. Users can only choose a review _date_ and cannot specify a _time_ – so it makes sense for us to store values in the same format users provide them in.
2. **Introduces the scope `ReviewReminder.reminder_due`** for finding documents that are due for a review reminder email. This means we can simplify the 'send review reminder emails' Rake task by removing the business logic required to find reminders that are due.
3. **Introduces methods `ReviewReminder#reminder_due?` and `ReviewReminder#reminder_sent!`** to simplify the logic required when working out if a particular reminder is due, and then marking it as sent.

## Why

- So users get reminders on the day they expected.
- Consolidate business logic for review reminders into one place (i.e. the `ReviewReminder` model).

Follows on from #8047
Trello: https://trello.com/c/HOwOTb4u/1558-fix-a-timezone-bug-in-review-reminders

---

### DIFF
--- a/app/models/review_reminder.rb
+++ b/app/models/review_reminder.rb
@@ -2,6 +2,12 @@ class ReviewReminder < ApplicationRecord
   belongs_to :document
   belongs_to :creator, class_name: "User"
 
+  scope :reminder_due, lambda {
+    where(review_at: ..Time.zone.today, reminder_sent_at: nil)
+      .joins(document: :latest_edition)
+      .where.not(document: { editions: { first_published_at: nil } })
+  }
+
   validates :document, :creator, :review_at, :email_address, presence: true
   validates :email_address, format: { with: URI::MailTo::EMAIL_REGEXP, if: -> { email_address.present? } }
   validate :review_date_cannot_be_in_the_past, if: -> { review_at.present? }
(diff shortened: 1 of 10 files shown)


beccapearce (approved):
Looks good! Thanks for the clear commit messages! ⭐

beccapearce (approved):

PR #8067, opened 2023-08-07
Remove unused Whitehall Frontend translations

brucebolt:
As Whitehall no longer renders any content to the public, we can remove any residual keys that are no longer being used.

This is the result of running `i18n-tasks unused`, then manually checking each key.

[Trello card](https://trello.com/c/lEgxXEDP)


jkempster34 (approved):
All looks good to me, thanks for splitting it up for easy review. ✅ 

I noticed GitHub has truncated one of your commit messages which you might want to fix as it isn't the prettiest. I don't know why though, I thought that would only happen over 72 characters 🤷

brucebolt:
> I noticed GitHub has truncated one of your commit messages which you might want to fix as it isn't the prettiest. I don't know why though, I thought that would only happen over 72 characters 🤷

Good spot.  I didn't notice that.  Have updated and shortened the commit message.

PR #8070, opened 2023-08-08
Remove Slimmer

brucebolt:
Slimmer is middleware that inserts the standard GOV.UK header and footer (retrieved from static) into each page.

As Whitehall no longer renders any content to the public, slimmer can be removed as it is not being used.

[Trello card](https://trello.com/c/Nsf89d2v)


floehopper:
@brucebolt @jkempster34 

While I was investigating something else, I noticed you've removed the `/auth/failure` route in this PR, so I think you can also remove the corresponding action:

https://github.com/alphagov/whitehall/blob/5141d52f714315f76e6e4bcbe18179fbc3a42ee1/app/controllers/admin/base_controller.rb#L8-L10

And the corresponding template:

https://github.com/alphagov/whitehall/blob/5141d52f714315f76e6e4bcbe18179fbc3a42ee1/app/views/authentications/failure.html.erb#L1-L13

Also the `except: %i[auth_failure]` option on this line:

https://github.com/alphagov/whitehall/blob/5141d52f714315f76e6e4bcbe18179fbc3a42ee1/app/controllers/admin/base_controller.rb#L6

You can see that at least some of this was added in [this commit](https://github.com/alphagov/whitehall/commit/f4bbca65ccc9c31d4af964bbb5af0d6bf56637e1) in #5096.

brucebolt:
Thanks and good spot @floehopper!  I've raised https://github.com/alphagov/whitehall/pull/8571 to remove the code that I previously missed.

PR #8072, opened 2023-08-08
Warn users about an attachment's markdown changing

dnkrj:

# What
If a user updates a file attachment to a new file with a different filename display a warning that any markdown referencing the file must be updated.
Updates the info text at file upload to reference markdown when replacing a file.

# Why
Markdown is now based on filename `[Attachment: example.pdf]` and hence changes if the file is replaced with one with a different filename. Content publisher displays a message in the notice but it is not conditional on the filename actually having changed.

# Screenshots
whitehall-admin dev gov uk_government_admin_editions_1374132_attachments_6940872_edit(iPad Pro)
whitehall-admin dev gov uk_government_admin_editions_1374132_attachments(iPad Pro)


baisa (approved):
LGTM @dnkrj

PR #8082, opened 2023-08-08
Add Release workflow to version deployments

theseanything:
This adds a release workflow that creates a new GitHub Release (and git tag) with an incremental version number (v1, v2, etc..). This version tag is then used to tag images and referenced in the rest of the deployment system. This makes it easier to compare references (vs commit SHAs). The version tag is calculated as the last version incremented by the number of merge commits until the reference commit.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,18 +19,16 @@ on:
         - staging
         - production
         default: 'integration'
-  workflow_run:
-    workflows: [CI]
-    types: [completed]
-    branches: [main]
+  release:
+    types: [released]
 
 jobs:
   build-and-publish-image:
-    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success'
+    if: github.event_name == 'workflow_dispatch' || startsWith(github.ref_name, 'v') 
     name: Build and publish image
     uses: alphagov/govuk-infrastructure/.github/workflows/build-and-push-image.yml@main
     with:
-      gitRef: ${{ inputs.gitRef || github.ref }}
+      gitRef: ${{ inputs.gitRef || github.ref_name }}
     secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -19,18 +19,16 @@ on:
...
-    workflows: [CI]
-    types: [completed]
-    branches: [main]
+  release:
+    types: [released]
 
 jobs:
   build-and-publish-image:
-    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success'
+    if: github.event_name == 'workflow_dispatch' || startsWith(github.ref_name, 'v')
```
richardTowers, line 27: This is quite a loose pattern - would this also match branches with names like `view-components`? Could that cause any confusion?

sengi (approved):
LGTM (modulo addressing @richardTowers's question)

.github/workflows/deploy.yml
theseanything: This workflow is triggered by a release event or manually (workflow_dispatch event). If it's a "workflow_dispatch" event,  the first part of the "if" statement will always be true `github.event_name == 'workflow_dispatch'. So the second part of the "if" statement will only be reached on a release event. For release events `github.ref_name` will always by a reference to a git tag (never a branch name).

It's possible that someone manually creates a release with v prefix, but realistically manually release are rare on GOV.UK for our deployable apps.

theseanything: For reference, the docs are here for the `github` context object:
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Whilst the docs describe `github.ref_name` as
> The short ref name of the branch or tag that triggered the workflow run. 

For release triggers, it's always [a tag that's triggered the run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release).

PR #8083, opened 2023-08-08
Remove "Preview call for evidence" feature flag

ollietreend:
We're ready to release the new Call for Evidence content type for general usage now. So we need to remove the feature flag that was used while the feature was still being previewed by developers and select users.

This effectively reverts the commit
c3a456987abf75a2d0a1bacc7859f2f4f2edecf4 which introduced this flag.

I've also removed use of the "Preveiw design system" feature flag in two of the functional tests because they weren't actually needed. The tests still pass fine without this flag.

Trello: https://trello.com/c/3ervkwyc/1567-release-call-for-evidence

---


baisa (approved):
Looks great @ollietreend 👌

PR #8086, opened 2023-08-08
remove documents controller

JonathanHallam:
Trello - https://trello.com/c/h8JG6EVI/769-remove-the-documentcontroller


brucebolt (approved):
Could you just add a note to the commit message to say that this isn't being used anywhere, which is why it is being deleted.  Just in case someone returns to this in the future and asks why the code was deleted.

PR #8091, opened 2023-08-08
Change the namespace of multiple helpers

JonathanHallam:
These things are only used by the admin interface so they've been moved to that namespace

Trello - https://trello.com/c/mcIdxTTJ/762-update-namespace-for-some-whitehall-code

### DIFF
--- a/app/helpers/admin/javascript_helper.rb
+++ b/app/helpers/admin/javascript_helper.rb
@@ -1,4 +1,4 @@
-module JavascriptHelper
+module Admin::JavascriptHelper
   def initialise_script(constructor_or_singleton_name, params = {})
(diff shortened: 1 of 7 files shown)


brucebolt (commented):
Could you add a line to each of the commit messages explaining why the code is being moved (i.e. it is now only used in the `admin` side of the site after the public rendering has been removed), so it is easier to find the reason for this change without needing to know which PR it came from.

JonathanHallam:
> Could you add a line to each of the commit messages explaining why the code is being moved (i.e. it is now only used in the `admin` side of the site after the public rendering has been removed), so it is easier to find the reason for this change without needing to know which PR it came from.

Done!

brucebolt (approved):

PR #8104, opened 2023-08-09
Removed calls for evidence text from consultations form

mjwhitehouse:
## What

This PR removes the guidance to use calls for evidence as part of a consultation content type, as call for evidence is now live.

### Text before

Screenshot 2023-08-09 at 17 55 13

We now do not advise to use a consultation content type to produce a call for evidence, so this is a text update 

## Why 

To ensure guidance to publishers is clear

## Trello card 

https://trello.com/c/YCr9rinL/1303-update-current-guidance-on-consultations-in-whitehall

### DIFF
--- a/app/views/admin/consultations/_form.html.erb
+++ b/app/views/admin/consultations/_form.html.erb
@@ -1,7 +1,6 @@
 <div class="format-advice">
-  <p class="govuk-body"><strong>Use this format for:</strong> Consultations (officially “documents requiring collective agreement across government”, including calls for evidence).</p>
-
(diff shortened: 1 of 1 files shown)


app/views/admin/consultations/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="format-advice">
-  <p class="govuk-body"><strong>Use this format for:</strong> Consultations (officially “documents requiring collective agreement across government”, including calls for evidence).</p>
+  <p class="govuk-body"><strong>Use this format for:</strong> Consultations (officially “documents requiring collective agreement across government.”)</p>
```
baisa, line 2: Do we want to have a dot inside or outside of the bracket?

PR #8110, opened 2023-08-10
Fix an error in the "What's new" page YAML

ollietreend:
There was a missing list item in the YAML which meant the two most recent "What's new" entries were showing as one.

The heading was showing "New call for evidence content type" but the rest of the content was being overwritten by the "New images" item.

This fixes the error introduced by PR #8101 / commit 753201a1d738c57855bd65bd03b0287d379c988f.

| Before | After |
| --- | --- |
| whats_new_before | whats_new_after |


ryanb-gds (approved):
Seems like it should do the trick 👍

PR #8119, opened 2023-08-14
Added documentation of the current state of the audit trail behaviour

ryanb-gds:
I've spent a fair bit of time recently investigating some support tickets where I needed to be able to identify when changes had been applied to an edition. I wasn't initially able to find any documentation explaining at a high level how the audit trail worked from a developer's point of view, so I have added a basic markdown document to the Whitehall docs folder.

Questions to reviewers:
- Is this the correct location for this documentation?
- Is this documentation valuable?
- Is this document accurate? I'm pretty new to Whitehall still.
- What else would be useful to add?

### DIFF
--- a/docs/auditing.md
+++ b/docs/auditing.md
@@ -1,0 +1,9 @@
+# Whitehall Content Auditing
+
+Some content types in Whitehall are versioned so that we are able to explain how a content item reached its current state.
+
+Versioning behaviour is implemented in the [audit trail module](../app/models/audit_trail.rb). At time of writing, this module is included in the `Edition` and `WorldwideOrganisation` models.
+
+Every time an auditable model is created or updated, the audit trail module adds a new record to the versions table. The version data includes the type and ID of the auditable model, the user ID of the user who triggered the event, if available (for updates performed via Rake tasks or queue workers this may not be possible), the new state of the model, and a timestamp. Note that user IDs are tracked in the `whodunnit` column, which is an idea borrowed from the [Paper Trail rails gem](https://github.com/paper-trail-gem/paper_trail).
+
+The `whodunnit` value is populated by fetching the user from the [Current model](../app/models/current.rb). The `Current` model extends the Active Support `CurrentAttributes` [class](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html), which provides convenient access to global request data in a thread-safe manner.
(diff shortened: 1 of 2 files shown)


docs/auditing.md
```diff
@@ -0,0 +1,9 @@
+# Whitehall Content Auditing
```
beccapearce, line 1: Just thinking about findability in the dev docs (all the documentation in the docs folder is imported into the dev docs), is it worth putting Audit trail in the title of this page?

docs/auditing.md
```diff
@@ -0,0 +1,9 @@
...
+
+Some content types in Whitehall are versioned so that we are able to explain how a content item reached its current state.
```
beccapearce, line 3: Do we know which content types are versioned? Maybe it's not necessary to say but I don't know if we actually have it documented anywhere

ryanb-gds: We can work it out I think - I think in my head I thought being slightly vague in some areas was future-proofing in case which types are auditable changes in future, but actually that's not really helping the reader!

beccapearce:
Thank you for adding this! Dead useful to have it.

The only other thing I might like to have added, if you agree, is a code snippet of have to see the versions on an edition. I'm assuming it's something like: `edition.versions.last` and an example of what is returned. (Feel free to disagree on this, I do like a code snippet a little much)

beccapearce:
To make sure I cover your questions
> Is this the correct location for this documentation?
Yes I believe so, it'll be imported into the dev docs which is useful for 2ndline.
> Is this documentation valuable?
For me yes, I always have to remember how auditing works in whitehall
> Is this document accurate? I'm pretty new to Whitehall still.
All seems correct to me
> What else would be useful to add?
added in the comments

Thank you for adding it! ⭐

docs/auditing.md
beccapearce: Hmmm, yeah you're totally right about the future proofing 🤔 

It would almost be easier to say which models don't have auditing as we're unlikely to add it, but that might be a very confusing way to put it

docs/auditing.md
ryanb-gds: Good shout

docs/auditing.md
ryanb-gds: Added

beccapearce (approved):
Looks great! 🎉

PR #8121, opened 2023-08-14
Delete search index records when an edition becomes unindexable

ryanb-gds:
This commit solves an issue that resulted in 404s from some search results pages. At present, only English language content should appear in search results. Currently, if a document is created with a primary locale of English, and then the locale is switched to another language, the document will still appear in the search index. This is mostly commonly seen with consultation documents. This commit ensures that the document is deleted from the search index if it is no longer indexable, but the previous edition was.

This solution was preferred over preventing users from changing the locale of a document once it has been created, due to the disruption this would cause to their workflows. We believe it is common for users to create a draft consultation in English and then switch to Welsh where necessary.

### DIFF
--- a/app/services/service_listeners/search_indexer.rb
+++ b/app/services/service_listeners/search_indexer.rb
@@ -8,6 +8,10 @@ module ServiceListeners
 
         Whitehall::SearchIndex.add(edition)
         reindex_collection_documents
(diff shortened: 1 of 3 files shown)


ollietreend (approved):
Nice work @ryanb-gds 🎉 

Thank you for digging in to the gnarly details of this one. I think this looks good!

PR #8123, opened 2023-08-15
Ignore empty specialist topic group when creating document collection

catalinailie:
There are a few empty specialist topic groups with empty names. These make the creation of document collection fail for those topic pages with `Validation failed: Heading can't be blank`.


app/services/draft_document_collection_builder.rb
```diff
@@ -44,7 +44,7 @@ def initialise_document_collection
       mapped_specialist_topic_content_id: specialist_topic[:content_id],
     ) do |document_collection|
       document_collection.title = "Specialist topic import: #{specialist_topic[:title]}"
-      document_collection.summary = specialist_topic[:description]
+      document_collection.summary = specialist_topic[:description].empty? ? "[Placeholder text for department to fill in]" : specialist_topic[:description]
```
hannako, line 47: I would just call this "placeholder text"

app/services/draft_document_collection_builder.rb
```diff
@@ -56,6 +56,8 @@ def initialise_document_collection
   def add_groups_to_document_collection(document_collection)
     groups =
       specialist_topic_groups.map do |group|
+        next if group[:name].empty?
```
hannako, line 59: This works as longs as it isn't possible for a group to have an array of content id's with no name... do we know that's the case?

catalinailie: Hmm that's a good point. All the groups with empty names I saw so far had no content ids. I tried to test it on integration but I can't create a group with empty name so I assume this is an old issue that was fixed.

catalinailie: Maybe I can add a placeholder group name for this one as we do with the summary and then the CD can delete the group if it's empty

hannako: Maybe we should instead do `next if empty_group?(group)` that checks if both heading and content_id's are empty. Just in case. It will blow up again if nameless groups _do_ exist, and we can look into it then.

catalinailie: Yes, this is much better. Thanks

app/services/draft_document_collection_builder.rb
```diff
@@ -112,6 +112,10 @@ def empty_group?(group)
     group[:name].empty? && group[:content_ids].empty?
   end
 
+  def document_withdrawn?(document)
+    document.editions.last.state == "withdrawn"
```
hannako, line 116: Tiny point, I think there's a latest_edition method (line 35)

hannako (approved):
I am approving the code changes implemented by Catalina. But as I've rebased to include some test commits authored by me, I better get an additional pair of eyes. cc @chao-xian 🙏

Tetrino (approved):

PR #8126, opened 2023-08-15
Redirect to document summary page after saving an edition

ryanb-gds:
### What

We think as part of editor quick wins that we should remove the taxonomy from being the next page on save and continue, so that publishers don’t add it each time

Save and continue should go to the document summary

### Why

We believe that removing topic tags from the workflow will save publishers time when editing and help publishers stay on task.

We believe this will not have negative consequences when they first create a document and they'll be able to easily add topic tags from summary or a secondary nav instead

**We know this is true when:**

- We see overall less time spent on topic tags page (because instead of forcing publishes to land on it each time, they'll only navigate to topic tags when they actually need to)
- Positive feedback from publishers

**Trello**: https://trello.com/c/DcdTBWJv

### DIFF
--- a/app/views/admin/statistics_announcements/show/_main.html.erb
+++ b/app/views/admin/statistics_announcements/show/_main.html.erb
@@ -135,7 +135,7 @@
     <% end %>
     <% unless @edition_taxons.any? %>
       <%= render "govuk_publishing_components/components/warning_text", {
-        text: "Please add a tag before publishing"
+        text: "Add topic tags before you can publish this document."
       } %>
     <% end %>
   </section>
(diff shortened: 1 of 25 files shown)


app/views/admin/statistics_announcements/show/_main.html.erb
```diff
@@ -135,7 +135,7 @@
     <% end %>
     <% unless @edition_taxons.any? %>
       <%= render "govuk_publishing_components/components/warning_text", {
-        text: "Please add a tag before publishing"
+        text: "Add topic tags before you can publish this document."
```
ryanb-gds, line 138: I'm not sure this reads perfectly. Should it say "You must add topic tags before..."?

ryanb-gds (commented):
Looks good, just one query about the wording of the tag warning

baisa (approved):
LGTM!

PR #8135, opened 2023-08-17
Improvements to search indexing documentation

ryanb-gds:
Added sections explaining how the Searchable module and the edition Search Indexer work. Changed the document to being generally about search indexing rather than just how to set up a local search index. Pointed out edition service initialiser and searchable config options. As always, keen to get feedback and also would appreciate a thorough check for accuracy!

I've taken the liberty of making the existing local search setup guide into a general search document, because having them in separate documents seemed a little unhelpful to me, but let me know if I should split them out again.


docs/search_indexing.md
```diff
@@ -1,14 +1,61 @@
-# Search setup guide
+# Search Indexing
 
-## Setup search locally
+Whitehall interacts directly with the Search API, unlike other publishing apps which leave publishing API to manage indexing.
```
ollietreend, line 3:
```suggestion
Whitehall interacts directly with the Search API, unlike other publishing apps which leave Publishing API to manage indexing.
```

ollietreend (approved):
My preference would be for us to capitalise app names as "Whitehall" and "Publishing API" consistently throughout this document. Although I appreciate this document already used lowercase "whitehall" before you made these edits, so you haven't added a 'new' problem per se.

Anyway – case sensitivity aside (🙈) this is a great addition to our docs. Thank you for taking the time to write this up @ryanb-gds! ⭐

PR #8146, opened 2023-08-21
Update current_release_date to use FK association

davidgisbey:
## Description 

This updates the statistics_announcements model to use the new backfilled FK rather than mnaually sorting the statistics_announcement_dates based on the most recently created.

This will both be more performant and allow us to more easily scope and order DB queries based on the current_release_date of the statistics_announcement.

Using this in place of the current_release_date association in the factory requires us to update some tests to use create rather than build due to various callbacks occurring that require the current_release_date_id
to be persisted and the association reloaded.

## Trello card

https://trello.com/c/8578hcqj/335-statistic-announcements-ordering-borked

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -28,7 +28,7 @@ class Publication < Publicationesque
   after_save :touch_statistics_announcement
   def touch_statistics_announcement
     if published? && !statistics_announcement.nil?
-      statistics_announcement.touch
+      statistics_announcement.reload.touch
     end
   end
 
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -52,7 +52,7 @@ class StatisticsAnnouncement < ApplicationRecord
               message: "must be a statistical type",
             }
 
-  accepts_nested_attributes_for :current_release_date
+  accepts_nested_attributes_for :statistics_announcement_dates
 
   scope :with_title_containing,
         lambda { |*keywords|
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -25,7 +25,10 @@ class StatisticsAnnouncement < ApplicationRecord
   belongs_to :publication
 
   has_one  :current_release_date,
-           -> { order("created_at DESC") },
+           lambda {
+             joins(:statistics_announcement)
+                .where("statistics_announcements.current_release_date_id = statistics_announcement_dates.id")
+           },
            class_name: "StatisticsAnnouncementDate",
            inverse_of: :statistics_announcement
   has_many :statistics_announcement_dates,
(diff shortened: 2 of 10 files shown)


app/models/statistics_announcement.rb
```diff
@@ -25,7 +25,10 @@ def can_publish_to_publishing_api?
   belongs_to :publication
 
   has_one  :current_release_date,
-           -> { order("created_at DESC") },
+           lambda {
+             joins(:statistics_announcement)
```
ollietreend, line 29: I think this should be:

```suggestion
             joins(:statistics_announcement_dates)
```

right?

test/factories/statistics_announcements.rb
```diff
@@ -12,22 +12,22 @@
     organisations { FactoryBot.build_list :organisation, 1 }
 
     association :creator, factory: :writer
-    association :current_release_date, factory: :statistics_announcement_date
+    statistics_announcement_dates { [create(:statistics_announcement_date)] }
```
ollietreend, line 15: I reckon this could probably be `build` rather than `create`? 🤔 

```suggestion
    statistics_announcement_dates { build_list(:statistics_announcement_date, 1) }
```

ollietreend (commented):
A couple of questions, but otherwise looks good.

It looks like this change was more fiddly than I'd expected!

app/models/statistics_announcement.rb
davidgisbey: So this is the pattern you established in documents here 

https://github.com/alphagov/whitehall/blob/main/app/models/document.rb#L27-L30

When you update it to your suggestion you get

image

davidgisbey: If fact removing inverse_of while less performant would remove the need to reload before a touching the statistics announcement again 🤔

ollietreend: Oh right, I see, my bad.

I was confused because it looked like we were trying to do `StatisticsAnnouncement.joins(:statistics_announcement)` and I was wondering why it was working. But I'd missed the `class_name` bit.

ollietreend: Re: using `inverse_of` and needing to `reload`. I wasn't 100% clear why reload was needed to touch? 🤔 

For what it's worth, the Document model joins `latest_edition` using the `inverse_of` option too.

https://github.com/alphagov/whitehall/blob/2694ef04a54098144998dd28c75e48e4f1a12c15/app/models/document.rb#L27-L30

davidgisbey: Yea i shamelessly copied your implementation 😂 basically the inverse_of means that it uses the preloaded object that has since been updated and is now stale. So when it updates the publication from draft to published it isn't able to see that change in the assoicated document (stale one)

test/factories/statistics_announcements.rb
davidgisbey: good shout. Have sorted

app/models/publication.rb
```diff
@@ -28,7 +28,7 @@ class Publication < Publicationesque
   after_save :touch_statistics_announcement
   def touch_statistics_announcement
     if published? && !statistics_announcement.nil?
-      statistics_announcement.touch
+      statistics_announcement.reload.touch
```
ollietreend, line 31: Is this deffo still needed? I seem to remember you saying it was to do with the way the associations were set up – just double checking it's still needed, because it doesn't seem 100% obvious why we should need to reload it.

ollietreend (approved):
Nice work! 🚀 Looks good to me.

app/models/publication.rb
davidgisbey: it is i'm afraid. The inverse relationship causes an eager load of the publication for the statistics announcement which doesn't capture the state transition that occurs from draft to published. 

I know it sucks 😢

PR #8150, opened 2023-08-22
Add a preview link to the edit page

dnkrj:

# What
- Add a link to the preview a document directly below the body field when a document is saved, publicly visible and has an appropriate change note.
- Add a link to preview an HTML Attachment in the edit page.

https://trello.com/c/sNmuIgvs/1562-add-document-preview-link-to-edit-screens

# Why
- To reduce the clicks for editors to preview back and forth.

# Screenshot
whitehall-admin dev gov uk_government_admin_editions_1374136_attachments_sed-fringilla-mauris-sit-amet_edit(iPad Pro)

### DIFF
--- a/app/views/admin/attachments/_form.html.erb
+++ b/app/views/admin/attachments/_form.html.erb
@@ -59,6 +59,27 @@
           }
         } %>
       </div>
+
+      <% if !attachment.new_record? %>
+        <%= render "govuk_publishing_components/components/inset_text", {
+          } do %>
+          <p class="govuk-body">
+            <%= link_to("Preview on website (opens in new tab)",
+            attachment.url(preview: true, full_url: true),
+            class: "govuk-link",
+            target: "_blank",
+            data: {
+              module: "gem-track-click",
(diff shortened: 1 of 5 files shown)


minhngocd:
Looks good, thanks Dan!

2 questions from me: 
- what are your thoughts on creating a new partial and just passing in the URL as param, it means the class and structure etc. is all in the same place
- what do we do about tests? is there a feature test or view test for this?

dnkrj:
> * what are your thoughts on creating a new partial and just passing in the URL as param, it means the class and structure etc. is all in the same place

I'm not sure that it's worth it at this point - with only two instances and a fairly short block of code - but definitely think that it makes sense if this becomes a pattern that we use across the application. At the same time don't have a strong opinion on it.

> * what do we do about tests? is there a feature test or view test for this?

We don't have much in the way of testing for bits of the views like this - you're suggesting adding a test to ensure it renders when we expect it to and doesn't render when a document hasn't been saved?

minhngocd (approved):

PR #8157, opened 2023-08-22
Update historical accounts associations to reflect usage and fix bugs

davidgisbey:
## Description

There's a bunch of details on the [trello card](https://trello.com/c/XBvhkwbl/390-historical-accounts-is-not-working-as-expected) but essentially after some digging and discussion with other teams who have touched this area of the codebase, historical accounts should only be for past prime ministers, and they should only be able to have one, not many. 

Other historical accounts have now been hardcoded on Collections.

Due to this we've decided to do the following:

1. Cleanup the data in the DB so that the Prime Minister role is the only role that can have historical accounts &  delete all other existing historical accounts (they don't work anyway). This was handled in this PR https://github.com/alphagov/whitehall/pull/8163
2. Update the people model to only have one historical account
3. Update the historical account model to only have one role
4. Remove the "Role" select from the new/edit historical account page and pass the Prime Minister role to the historical account in the controller
5. Hide the Create historical account button if one is already present on the person.

## Trello card

https://trello.com/c/XBvhkwbl/390-historical-accounts-is-not-working-as-expected

### DIFF
--- a/app/controllers/admin/historical_accounts_controller.rb
+++ b/app/controllers/admin/historical_accounts_controller.rb
@@ -4,15 +4,15 @@ class Admin::HistoricalAccountsController < Admin::BaseController
   layout "design_system"
 
   def index
-    @historical_accounts = @person.historical_accounts.includes(roles: :translations)
+    @historical_account = @person.historical_account
   end
 
   def new
-    @historical_account = @person.historical_accounts.build
+    @historical_account = @person.build_historical_account
   end
 
   def create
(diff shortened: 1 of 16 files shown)


davidgisbey:
the 1st and last commit lgtm! If you're happy with the 2-4 feel free to approve and we can run the migration/merge post lunch

farahTW:
The second ,third and last commit looks good to me as well !

PR #8158, opened 2023-08-22
Render About us as edition when presenting worldwide organisation

CristinaRO:
The code for rendering an edition does all the necessary processing of
the attachments. In order to use the edition rendering, we need to pass
the About us object to the rendering method, instead of its body.

I've added a test because this functionality is quite complex, relying
as it does on special treatment of a worldwide corporate info page.

[Trello card](https://trello.com/c/DFzKJtrm/764-inline-attachments-dont-show-on-about-us-worldwide-organisation-corporate-information-pages)

### DIFF
--- a/app/presenters/publishing_api/worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/worldwide_organisation_presenter.rb
@@ -59,7 +59,7 @@ module PublishingApi
     end
 
     def body
-      Whitehall::GovspeakRenderer.new.govspeak_to_html(item.body) || ""
+      Whitehall::GovspeakRenderer.new.govspeak_edition_to_html(item.about_us) || ""
     end
 
     def main_office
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/worldwide_organisation_presenter.rb
```diff
@@ -59,7 +59,7 @@ def description
     end
 
     def body
-      Whitehall::GovspeakRenderer.new.govspeak_to_html(item.body) || ""
+      Whitehall::GovspeakRenderer.new.govspeak_edition_to_html(item.about_us) || ""
```
brucebolt, line 62: Good idea removing the ambiguity of the `body` vs `about_us` here 👍.

brucebolt (approved):
Nice change.  All looks good to me.

app/presenters/publishing_api/worldwide_organisation_presenter.rb
CristinaRO: Credit to @jkempster34 :)

PR #8159, opened 2023-08-22
Remove custom stylelint rules

dnkrj:

# What
- Remove custom stylelint rules

# Why
- Following #8023 we no longer lint our legacy css for which these exceptions were introduced


ryanb-gds (approved):
Had a good poke  around locally at anything I thought might render govspeak or attachments and couldn't find anything amiss, so looks like we were good to delete the helpers

PR #8163, opened 2023-08-23
Add data migration to clean up historical accounts

davidgisbey:
## Description

Historical accounts should now only be able to be created for prime minsters and should be have one rather than many. The work to do this will be added in a separate PR.

This does two things:
1. Destroys any historical accounts that exist for foreign secretaries
2. Ensures that the only role which supports historical accounts is the prime minister.

## Trello card

https://trello.com/c/XBvhkwbl/390-historical-accounts-is-not-working-as-expected


farahTW (approved):
LGTM

PR #8168, opened 2023-08-24
Prevent image cropper from being initialised twice

dnkrj:

# What
- Early exit `initCropper` when the cropper has already been initialised.

# Why
- There is an edge case where both an image is both 'complete' and triggers a 'load' event.
- This was causing the image cropper to be initialised twice resulting in a bug for the user a cropped image could not be saved.

https://trello.com/c/Bn1TJJZv/1669-investigate-bug-where-image-cropper-doesnt-always-work


ollietreend (approved):
Great work squashing this pesky little bug! 🏅 🙌🏻

It seems wild to me that there isn't more predictability about how & when the `load` event should fire vs. the `complete` property become true. But 🤷🏻 this seems like a reliable way around it.

PR #8171, opened 2023-08-24
Use govuk date component for schedule publication date field

ryanb-gds:
### What

This PR switches the schedule publication field to use a new date time field component that leverages the GOVUK shared date field component. It maintains the existing request shape and data types. Once all uses of the existing date time component have been replaced with our new component, we will delete the old component and remove the awkward `with_govuk_date_component` filename suffix from the new one.

The switch to the GOVUK component requires us to add date validation which was previously considered unnecessary because the select element limited the options a user could input.

### Why

We are making this change to make Whitehall more consistent with other GOVUK interfaces and because we believe the GOVUK component offers a better user experience. However, it is also more easily possible to input invalid dates to the GOVUK component, so this PR also adds date validation based on the behaviour in content publisher to ensure that users have a good experience if they make a mistake.

### DIFF
--- a/app/views/admin/editions/_scheduled_publication_fields.html.erb
+++ b/app/views/admin/editions/_scheduled_publication_fields.html.erb
@@ -13,24 +13,33 @@
         label: "Schedule for publication",
         value: 1,
         checked: edition.scheduled_publication.present?,
-        conditional: render("components/datetime_fields", {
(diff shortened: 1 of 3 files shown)


ryanb-gds:
Closing this PR as unable to find a technically satisfactory approach to achieving this that maintained a good user experience

PR #8172, opened 2023-08-24
Remove contacts guidance in formatting help

mjwhitehouse:
## What

Currently the in-app guidance on Whitehall suggests that users can search for a contact by typing ‘[Contact: ’ and then type a keyword to find your contact (if it exists).

Since the move to the GOV.UK design system, this hasn't worked, and the functionality no longer exists. We think that it's difficult to make this accessible, so this PR removes the guidance to suggest that as an option for users. 

Also updated what appeared to be a broken link in the guidance as it didn't go anywhere, and have linked to the organisations guidance.

## Why

We received a Zendesk ticket asking about this feature, although it's only one piece of feedback from one user. We believe the feature is under-utilised, and actually most publishers copy and paste from the organisations tab, which has also been made clearer in the move to the design system.

## What has been removed?

Screenshot 2023-08-24 at 21 06 01

## Links

https://trello.com/c/tsyOqfLj
https://govuk.zendesk.com/agent/#/tickets/5351274

### DIFF
--- a/app/views/admin/editions/_govspeak_help.html.erb
+++ b/app/views/admin/editions/_govspeak_help.html.erb
@@ -203,8 +203,8 @@
     <p class='govuk-body'>You can also embed contact information:</p>
     <pre class='govspeak-help__pre'>[Contact:<em>n</em>]</pre>
     <p class='govuk-body'>(n is the ID of the contact you want to embed.)</p>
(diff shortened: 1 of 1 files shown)


davidgisbey (approved):
lgtm! 🚀

PR #8174, opened 2023-08-25
Move flash to view component

beccapearce:
As part of the work to remove the topic taxonomy from the edition workflow (trello card below). We wanted to render a flash[:notice] with a link. However we found that while all flash[:alert] messages were marked as html_safe no flash[:notice] message could be marked as safe. While there were no flash[:alert] messages which showed user input, this was not documented and meant that an alert with user input may have been added by an unknowing dev. 

This work was intended to allow both flash[:notice] and flash[:alert] messages to be able to be marked as html_safe when safe and default to not html_safe in order to keep our application secure and prevent any malicious content from being injected. 

We added this to a view component instead of leaving it in the view to ensure that we could test the work properly.

This work is a precursor to the work done in https://github.com/alphagov/whitehall/pull/8126

Trello card: https://trello.com/c/2jmWV2Ur/1638-update-buttons-in-topic-taxonomy-tags-page

### DIFF
--- a/app/components/admin/flash_alert_component.rb
+++ b/app/components/admin/flash_alert_component.rb
@@ -1,0 +1,9 @@
+class Admin::FlashAlertComponent < Admin::FlashMessageComponent
+  def component_name
+    "error_alert"
+  end
+
+  def data_track_action
+    "alert-danger"
+  end
+end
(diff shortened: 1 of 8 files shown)


ryanb-gds (approved):
This would be a good opportunity to use the classic [replace conditional with polymorphism](https://refactoring.guru/replace-conditional-with-polymorphism) which would benefit us if we ever wanted to add extra behaviour, but given I'd imagine that's somewhat unlikely more than happy to approve as-is. If you felt [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) that would be fair enough

ollietreend (approved):
Sorry, one final thing. I wonder if the test description

PR #8186, opened 2023-08-29
Work in Progress - moving Alternative Format Provider Email for Attac…

minhngocd:
…hments

- just moved the field from the view to the attachments page.
- commented out tests that require attachment email provider to be available. These tests will need to move to the attachments page.
- Added a null check for yielding blocks in editions helper since Corporate Information Page only requires the block to render the email provider. We might as well not pass in the block in that scenario.

Other things to consider:
- When moving the field, we will now have to default the email for edition creation. This needs to be done
- What happens when the lead org doesn't have an email set, what do we default it to?
- The text for Image and Attachments need to stay in the document forms - but as a separate piece of info label
- Creating new page for saving the actual email

https://trello.com/c/b6ftq7IE/1563-move-accessible-attachments-email-to-attachments-section


ryanb-gds:
I'm going to close this because we've completely changed the navigation since @minhngocd started working on it. I feel it would be easier for me to start from a clean slate

PR #8193, opened 2023-08-31
Add Group Documents (memberships) Index and Confirm Destroy pages

davidgisbey:
## Screenshots

image

image

image

https://trello.com/c/szAfNI2T/526-create-a-group-landing-page-for-adding-documents

### DIFF
--- a/app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
+++ b/app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
@@ -1,0 +1,9 @@
+.app-view-document-collection-group-memberships-index__table {
+  .govuk-table__cell {
+    word-break: break-word;
+
+    &:last-child {
+      width: 7em;
+    }
+  }
+}
--- a/app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
+++ b/app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
@@ -1,9 +1,12 @@
-.app-view-document-collection-group-memberships-index__table {
-  .govuk-table__cell {
-    word-break: break-word;
+.app-view-document-collection-group-memberships-index__heading {
+  display: flex;
+  justify-content: space-between;
+}
 
-    &:last-child {
-      width: 7em;
-    }
+.app-view-document-collection-group-memberships-index__table .govuk-table__cell{
+  word-break: break-word;
+
+  &:last-child {
+    white-space: nowrap;
   }
 }
--- a/app/views/admin/document_collection_group_memberships/index.html.erb
+++ b/app/views/admin/document_collection_group_memberships/index.html.erb
@@ -45,7 +45,7 @@
               },
               {
                 text: link_to(sanitize("View #{tag.span(title, class: "govuk-visually-hidden")}"), membership.document.latest_edition.public_url, class: "govuk-link") +
-                  link_to(sanitize("Delete #{tag.span(title, class: "govuk-visually-hidden")}"), confirm_destroy_admin_document_collection_group_document_collection_group_membership_path(@collection, @group, membership), class: "govuk-link gem-link--destructive govuk-!-margin-left-3")
+                  link_to(sanitize("Remove #{tag.span(title, class: "govuk-visually-hidden")}"), confirm_destroy_admin_document_collection_group_document_collection_group_membership_path(@collection, @group, membership), class: "govuk-link gem-link--destructive govuk-!-margin-left-3")
               }
             ]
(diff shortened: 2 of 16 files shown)


app/views/admin/document_collection_group_memberships/index.html.erb
```diff
@@ -0,0 +1,62 @@
...
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-three-quarters">
+        <%= render "govuk_publishing_components/components/heading", {
+          text: "Documents",
+          margin_bottom: 6,
+        } %>
+      </div>
+
+      <div class="govuk-grid-column-one-quarter">
```
davidtrussler, line 30: I think there should be a more efficient way to deal with the layout requirements here: as it is the "Add document" link breaks across lines strangely and floats below and to the right in smaller viewports. The roughly equivalent layout in the summary card component uses flexbox so this might be the best way to deal with it (I assume we are OK with support for this) otherwise it would need to float and respond to breakpoints which could get a bit awkward.

app/views/admin/document_collection_group_memberships/index.html.erb
```diff
@@ -0,0 +1,62 @@
...
+      <div class="govuk-table--with-actions app-view-document-collection-group-memberships-index__table">
+        <%= render "govuk_publishing_components/components/table", {
+          rows: @group.memberships.map do |membership|
+            title =  membership.document.latest_edition.title
+            [
+              {
+                text: title,
+              },
+              {
+                text: link_to(sanitize("View #{tag.span(title, class: "govuk-visually-hidden")}"), membership.document.latest_edition.public_url, class: "govuk-link") +
```
davidtrussler, line 47: Minor thing but I think elsewhere the space between "View"/"Remove" links and the visually hidden text has been part of the hidden text. Otherwise the link underline spans the text and the space making it look slightly odd.

app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
```diff
@@ -0,0 +1,9 @@
+.app-view-document-collection-group-memberships-index__table {
+  .govuk-table__cell {
+    word-break: break-word;
+
+    &:last-child {
+      width: 7em;
+    }
```
davidtrussler, line 7: I think I would try to avoid specifying explicit widths for textual elements because of variations in how devices might render the text, particularly if it's magnified or some-such. Personally I don't have a problem with the links wrapping onto new lines and I think it happens elsewhere but if it is to be avoided I think applying something like `white-space: nowrap;` to the cell would achieve this.

minhngocd: the white-space nowrap is a much nicer touch - thank you! I will update

app/views/admin/document_collection_group_memberships/index.html.erb
minhngocd: Hi @davidtrussler - I don't see where else it has the space being part of the hidden text, do you have an example? 
I searched for `#{tag.span(` and the results all have the space outside the hidden span, with the one exception of World Locations.

davidtrussler: @minhngocd Oh yes, you're right. In fact I see now that the `govuk-visually-hidden` CSS class actually adds a space before that text so really we should delete the space here and it will be taken care of that way. But since we haven't done that anywhere else (apart from the one you mention) maybe it doesn't really matter. Does look a bit wrong to me though.

minhngocd: @davidtrussler oh dear 🙈 i will add a ticket to the tech debt column

davidtrussler:
> 

Hi @minhngocd - That works, but I think the flexbox solution could be a bit more elegant if you simplify the markup to put the `<h2>` and the `<p>` elements together into the same containing `<div>` then style that as `display: flex; justify-content: space-between;`.

app/views/admin/document_collection_group_memberships/index.html.erb
davidtrussler: Yes, good idea, we should make it all consistent but in the meantime it's not exactly a show-stopper! Might be worth a bit more thought as well because I have this idea (that might well be nonsense) that screen-readers get thrown when a mid-sentence word starts in uppercase, as it does here.

minhngocd:
@davidtrussler ohh amazing - thanks !! I've pushed the changes. 🙏

davidtrussler (approved):
Great, LGTM! 👍

PR #8195, opened 2023-08-31
Add option to filter documents by overdue review reminders

davidgisbey:
## Description

This adds in an additional filter to the documents search page which allows users to be able to filter by review reminders that are overdue.

Unforunately i've had to manually adds the joins to the documents and review reminder tables due to some custom sql written in the :with_title_containing scope

```
in_default_locale
  .includes(:document)
  .where("edition_translations.title LIKE :like_clause OR documents.slug = :slug", like_clause:, slug: keywords)
  .references(:document)
```

This expects the documents table to be called documents, but when you use active record to inner join a has one relationship it aliases the table to singular.

So in this case, when you inner join the document it aliases the documents table to 'document'. This causes  the above where clause to blow up as it can't find the documents table. Interestingly the same behaviour doesn't seem to occur for left outer join.

## Screenshots

image

## Trello card

https://trello.com/c/IX8jDbW9/379-add-review-overdue-to-the-states-search-list-on-the-document-search-page

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -72,37 +72,38 @@ class Edition < ApplicationRecord
           escaped_like_expression = keywords.gsub(/([%_])/, "%" => '\\%', "_" => '\\_')
           like_clause = "%#{escaped_like_expression}%"
 
-          in_default_locale
-            .includes(:document)
-            .where("edition_translations.title LIKE :like_clause OR documents.slug = :slug", like_clause:, slug: keywords)
-            .references(:document)
+          scope = in_default_locale.includes(:document)
+          scope
+            .where("edition_translations.title LIKE :like_clause", like_clause:)
+            .or(scope.where(document: { slug: keywords }))
         }
 
-  scope :in_pre_publication_state,      -> { where(state: Edition::PRE_PUBLICATION_STATES) }
-  scope :force_published,               -> { where(state: "published", force_published: true) }
-  scope :not_published,                 -> { where(state: %w[draft submitted rejected]) }
-  scope :without_not_published,         -> { where.not(state: %w[draft submitted rejected]) }
+  scope :in_pre_publication_state, -> { where(state: Edition::PRE_PUBLICATION_STATES) }
+  scope :force_published, -> { where(state: "published", force_published: true) }
+  scope :not_published, -> { where(state: %w[draft submitted rejected]) }
+  scope :without_not_published, -> { where.not(state: %w[draft submitted rejected]) }
 
-  scope :announcements,                 -> { where(type: Announcement.concrete_descendants.collect(&:name)) }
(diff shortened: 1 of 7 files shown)


app/models/edition.rb
```diff
@@ -245,6 +245,12 @@ def self.only_broken_links
     )
   end
 
+  def self.review_overdue
+    joins("INNER JOIN documents ON documents.id = editions.document_id INNER JOIN review_reminders ON review_reminders.document_id = documents.id")
+    .where(document: { review_reminders: { review_at: ..Time.zone.today } })
+    .where.not(first_published_at: nil)
+  end
+
```
minhngocd, line 253: Have just tested this and it seems to produce the correct SQL 
```
  scope :review_overdue, -> {
    joins(document: :review_reminder)
      .where(review_at: ..Time.zone.today)
      .where.not(first_published_at: nil)
  }
```

Produces
```
"SELECT * FROM `editions` 
INNER JOIN `documents` ON `documents`.`id` = `editions`.`document_id` 
INNER JOIN `review_reminders` ON `review_reminders`.`document_id` = `documents`.`id` 
WHERE `editions`.`state` != 'deleted' 
AND `editions`.`review_at` <= '2023-09-04' 
AND `editions`.`first_published_at` IS NOT NULL"
```

minhngocd: I think it might be a bit more readable? what do you think?
I also feel like it should be a scope rather than a self method to accurately portray what the method is doing

davidgisbey: hmm so interestingly if i use this scope 

```
  scope :review_overdue, -> {
    joins(document: :review_reminder)
      .where(document: { review_reminders: { review_at: ..Time.zone.today } })
      .where.not(first_published_at: nil)
  }
```

my output is 

```
 FROM `editions` INNER JOIN `documents` `document` ON `document`.`id` = `editions`.`document_id` INNER JOIN `review_reminders` ON `review_reminders`.`document_id` = `document`.`id` WHERE `editions`.`state` != 'deleted' AND `review_reminders`.`review_at` <= '2023-09-04' AND `editions`.`first_published_at` IS NOT NULL"
```
It's interesting that mine is aliasing the table to `document`.

Due to that, it's causing this error

https://github.com/alphagov/whitehall/assets/42515961/a7041702-5564-49bb-a46b-a7c87408aba6

I'd be interested to see if that bug is replicable for you if you get a bit of time please 🙏 

steps to recreate would be

1. create a review reminder where the review date has passed (probs easiest from console)
2. search for it by title/slug while the review overdue filter is checked

Def agreed that using a scope is way nicer 👍 I'll sort that now

davidgisbey:
Amazing that's working perfectly now 👍 thanks so much for sorting that!

minhngocd (approved):

PR #8199, opened 2023-09-01
Fix test so that it tests the right method

mtaylorgds:
There are a set of tests for `errors_for_input` and another set for `errors_for` in `errors_helper_test.rb`, but one of the tests for `errors_for` actually calls `errors_for_input`. It looks like this was copy/pasted and the test name was updated, but the test body was not.


ryanb-gds (approved):
Thanks Mark, nice catch!

PR #8201, opened 2023-09-01
Replace errant 'style' attribute with 'class'

mtaylorgds:
A paragraph tag is using a 'style' attribute when it should be using a 'class' one (and is in fact setting the style to a class, not a style).


minhngocd (approved):
The inconsistent double-quote and single-quote usage in this file is causing my eyes to twitch but it's not the point of this PR nor introduced, so LGTM

PR #8207, opened 2023-09-04
Add translated textarea and input components

davidgisbey:
## Description

At the moment, we have quite a few translation pages that render textareas and inputs then the english translation below them.

We've decided that we're going to migrate the h2 and english translation for both to a details component.

While doing this I think it makes sense to add reusable components which handle spacing etc. so they're all consistent. This PR adds 2 components which handle this.

If details text, but no heading it will default to `English translation` and have `for {label_name_with_(required)_stripped}` in a visually hidden span.

## Guidance for review

I've potentially made some fault assumptions about the correct behaviour of this component.

1.  If there is no english translation for the field, i've hidden the details component. It might be that there be something that states there's no translation for that field
2. i've added a default title for the details component which simply states "English translation" with a visually hidden span which adds the label to the end ("English translation for Title" etc.)

It might be that these pages should have more specific titles.

I probs need to speak to Nik about this when he's back off leave before i'll know how to proceed.

### Translated textarea component

image

image

### Translated input component

image

image

## Trello card

https://trello.com/c/8vp9mX1f/365-ui-revisit-translation-page-update-original-language-inside-the-details-component
https://trello.com/c/S0kuPlXk/470-create-new-endpoint-for-new-translation-page

### DIFF
--- a/app/views/components/docs/translated_input.yml
+++ b/app/views/components/docs/translated_input.yml
@@ -1,0 +1,43 @@
+name: Translated input
+description: Renders an input field and the english translation within a details component if one is present
+body: |
+  This combines two other publishing components:
+
+  [input]: https://components.publishing.service.gov.uk/component-guide/input
+  [details]:  https://components.publishing.service.gov.uk/component-guide/details
+
+  It should be used on pages where translations of objects can be created or edited.
+accessibility_criteria: |
+  - accept focus
+  - be focusable with a keyboard
+  - be usable with a keyboard
+  - indicate when it has focus
+examples:
+  default:
+    data:
+      input:
+        label:
+          text: Input (required)
+        heading_size: l
+        name: input
+  with_details_text:
+    data:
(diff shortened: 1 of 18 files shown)


app/views/components/docs/translated_input.yml
```diff
@@ -0,0 +1,43 @@
+name: Translated input
+description: Renders an input and the english translation within a details component if one is present
+body: |
+  This combines two other publishing components:
+
+  [input]: https://components.publishing.service.gov.uk/component-guide/input
+  [details]:  https://components.publishing.service.gov.uk/component-guide/details
+
+  It should on pages where translations of objects can be created or edited.
```
davidtrussler, line 9: This should probably read something like "It should be used on pages ..."

davidtrussler:
Hi @minhngocd 
I just left a *very* minor comment on this, otherwise looks fine.

PR #8209, opened 2023-09-05
Apply translated components to translation pages

davidgisbey:
## Description

This follows on from https://github.com/alphagov/whitehall/pull/8207 which added two new components

- translated_input
- translated_textarea

Essentially they combine a text input or textarea and a details component. This PR adds these components to the various translation pages.

I've also cleaned up a few things like:

- govspeak-component being used on the edit page but not the edit translation page
- inconsistent formatting
- textareas being used on the edit translation page when a text input is used on the edit page

### Contact translations

image

### Organisation translations

image

### Person translations

image

### Role translations

image

### World news story translations

image

### Worldwide org translations

image

### Worldwide office translations

image

### Edition translations

image

## Trello cards

https://trello.com/c/T6gEl6CR/486-create-translated-textarea-and-input-components
https://trello.com/c/8vp9mX1f/365-ui-revisit-translation-page-update-original-language-inside-the-details-component

### DIFF
--- a/app/assets/stylesheets/admin/_overrides.scss
+++ b/app/assets/stylesheets/admin/_overrides.scss
@@ -25,3 +25,10 @@ code {
     display: block;
   }
 }
+
+.app-c-translated-textarea,
+.app-c-translated-input {
+  & > *:first-child > .govuk-form-group {
+    margin-bottom: govuk-spacing(2) !important; // stylelint-disable-line declaration-no-important
+  }
+}
--- a/app/assets/stylesheets/admin/views/_translation.scss
+++ b/app/assets/stylesheets/admin/views/_translation.scss
@@ -1,3 +1,0 @@
-.app-view-translation__english-content {
-  color: $govuk-secondary-text-colour;
-}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -42,7 +42,6 @@ $govuk-page-width: 1140px;
 @import "./admin/views/summary";
 @import "./admin/views/take-part";
 @import "./admin/views/topical-events-index";
-@import "./admin/views/translation";
 @import "./admin/views/unpublish-withdrawal";
 @import "./admin/views/whats-new";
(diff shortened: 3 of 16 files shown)


app/assets/stylesheets/admin/_overrides.scss
```diff
@@ -25,3 +25,11 @@ code {
...
   }
 }
+
+.app-c-translated-textarea, .app-c-translated-input {
+  & > *:first-child {
+    .govuk-form-group {
+      margin-bottom: 10px !important; // stylelint-disable-line declaration-no-important
+    }
+  }
+}
```
davidtrussler, line 35: A few small things on this now that I've seen this in context (sorry I didn't think of this earlier):
- I've realised there is a built-in way to override the design-system margins, which is to use the [Spacing helpers](https://design-system.service.gov.uk/styles/spacing/#applying-spacing-in-your-own-css), which means we don't need the `!important` declaration (which feels better :smile:). 
- It's best not to use explicit pixel dimensions but rather use the [Responsive spacing scale](https://design-system.service.gov.uk/styles/spacing/#responsive-spacing) so that the spacing can adapt for different viewports. 
- we shouldn't need the `& > *:first-child` rule to target the `govuk-form-group` class (provided that it's the only use of that class within the parent element which I think is the case here).

So I think this can be written as
```
.app-c-translated-textarea, .app-c-translated-input {
  .govuk-form-group {
    @include govuk-responsive-margin(2, "bottom");
  }
}
```

davidtrussler:
Hi @minhngocd 
I've just added a comment on the CSS - the rest of it looks fine.

app/assets/stylesheets/admin/_overrides.scss
minhngocd: Thanks @davidtrussler ! Unfortunately I can't seem to remove the need for !important... maybe would be good to pair on this tomorrow, but I've incorporated the other changes here.

davidtrussler (approved):
LGTM! 👍

PR #8211, opened 2023-09-05
Add new endpoint for worldwide_organisations_translations

davidgisbey:
## Description 

Currently, we use the edit and update endpoint for creating new translations for objects. When you select the language you want it posts to the create endpoint, which then redirects to the edit page.

This isn't particularly idiomatic for Rails and stands out from the rest of the codebase which tends to use Rails resources traditionally (new & create for non-persisted objects and edit & update for updating persisted ones).

Interestingly, the current behaviour for a localised model for which a translation is not yet persisted is to display the primary language translations attrs.

Once a translation is persisted to the database, the attrs on the translation are then used on the edit page.

To get round this behaviour you can build a translation for the locale on the object itself, so in this case a Worldwide Organisation, then use this object in the context of the new action.

It's not an ideal solution, but it's the simplest way i can think of to ensure the correct behaviour for the new action and when the new view is re-rendered on an unsuccessful save

## Screenshots

image

image

## Trello card

https://trello.com/c/S0kuPlXk/470-create-new-endpoint-for-new-translation-page

### DIFF
--- a/app/assets/stylesheets/admin/views/_translation.scss
+++ b/app/assets/stylesheets/admin/views/_translation.scss
@@ -1,3 +1,0 @@
-.app-view-translation__english-content {
-  color: $govuk-secondary-text-colour;
-}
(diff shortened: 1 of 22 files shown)


davidgisbey:
closing this it's stale

PR #8215, opened 2023-09-06
Use design system date component for scheduled publication input

ryanb-gds:
### What

Replace the scheduled publication date selection component with the date input component from the [GOV.UK publishing components](https://components.publishing.service.gov.uk/component-guide/date_input) 

### Why

We feel this is a better user experience because users are able to input a date without scrolling through a large number of options. This is also bringing us in line with other publishing applications.

One thing to note is that until now there was no validation of the scheduled publication input because it was assumed to be impossible to enter an invalid date using the select inputs. Rails also automatically attempted to use the input to instantiate a Date object. This is no longer safe to do, so we have had to add a check to ensure that the input can be safely parsed to a date and to add an error to the edition if it cannot.

We've also fixed a bug where the scheduled publication active checkbox was losing its state on a validation error.

### Next Steps

We're going to gradually replace the other document date fields

### Screenshots
Screenshot 2023-09-06 at 12 45 26

Screenshot 2023-09-06 at 12 53 26

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -54,6 +54,7 @@ class Edition < ApplicationRecord
   validates_each :first_published_at do |record, attr, value|
     record.errors.add(attr, "can't be set to a future date") if value && Time.zone.now < value
   end
+  validate :scheduled_publication, :valid_date
 
   UNMODIFIABLE_STATES = %w[scheduled published superseded deleted].freeze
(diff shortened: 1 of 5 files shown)


baisa (approved):

mjwhitehouse:
@ryanb-gds - Is the error message in x's? I'm not sure this matches the pattern https://design-system.service.gov.uk/components/error-message/

PR #8216, opened 2023-09-06
Update buttons on edit topics page

davidgisbey:
**This is coupled to https://trello.com/c/DcdTBWJv/1561-remove-topic-tags-from-the-editor-workflow. Do no merge until both are ready to be shipped**

## Description

This PR updates the workflow for adding topics and specialist topics to an edition. It does the following:

- updates the copy for the "Update tags" button to "Save",
- removes the "Update and review specialist topic tags" button
- removes the routing previously required for multiple flows from the edition tags controller
- removes additional routing from the legacy associations controller (specialist tags)
- updates the feature and controller specs to use/pass with the new flow

## Guidance for review

I've not added a flash for when the user clicks the `Cancel` link. Generally flashes are reserved for successful/unsuccessful changes in the state of objects.

I suspect it's not worth adding this for some additional reasons:

1. there are `Cancel` links throughout Whitehall, none of which trigger flashes so it would be inconsistent with all other usages
2. it adds additional complexity as the document summary page would need to be arrived at with a query stringwould be checked for before rendering the page. If we consistently do this it'll add a good amount of extra work and controller bloat.

Very happy to discuss.

## Screenshots

image

image

## Trello card

https://trello.com/c/2jmWV2Ur/1638-update-buttons-in-topic-taxonomy-tags-page

### DIFF
--- a/app/controllers/admin/edition_legacy_associations_controller.rb
+++ b/app/controllers/admin/edition_legacy_associations_controller.rb
@@ -4,31 +4,17 @@ class Admin::EditionLegacyAssociationsController < Admin::BaseController
   before_action :limit_edition_access!
   layout "design_system"
 
-  def edit
-    @path = get_path
-  end
+  def edit; end
 
   def update
     @edition.assign_attributes(edition_params)
     if updater.can_perform? && @edition.save_as(current_user)
       updater.perform!
     end
-    redirect_to get_path, saved_confirmation_notice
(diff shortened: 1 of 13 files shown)


beccapearce (approved):
This looks good, We're planning on releasing the first part of this tomorrow so we can merge it in then. We might want to let the designers double check it on integration tomorrow as well

PR #8219, opened 2023-09-07
Ensure scheduled publication is set to a future date

ryanb-gds:
Now that we're allowing users free rein to enter their dates using text inputs, we need to start validating that they make sense in the context of the application.

Related Trello: https://trello.com/c/IFGDegsk


ryanb-gds:
Apparently according to https://govuk.zendesk.com/agent/filters/25324916 this may break users' workflows, so we may need to have a wider discussion before merging this.

beccapearce (approved):

PR #8222, opened 2023-09-08
Modify scheduled publication date validation error message

ryanb-gds:
The updated message is closer to the guidance provided at https://design-system.service.gov.uk/components/date-input/, but the guidance doesn't cover cases where the user enters, for example, a string instead of a number for one of the date parts. It also assumes that invidividual parts of the date can be validated separately. For this reason I have had to improvise a little, but I believe the new message is closer to the spirit of the guidance and does not feature the unapproved XX style formatting advice. Given the format is already provided in the input hint there is no need to specify it again in the error message.

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -736,7 +736,7 @@ EXISTS (
 
   def valid_date
     if @date_field_validity.present? && @date_field_validity[:scheduled_publication] == false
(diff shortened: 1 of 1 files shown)


beccapearce (approved):
I this is definitely an improvement so I'll approve. I'm wondering though if we want to ask a content designer to have a think on it as well?

PR #8226, opened 2023-09-08
Ensure ministerial reordering publishes changes to Publishing API

davidgisbey:
## Description

We've had tickets come through stating that the ministerial ordering page is not working correctly.

At the moment, when you reorder ministerial roles and organisations, it calls update_all which skips validations and callbacks.

This means that the after_save which pushes changes to Publishing API is not called, so the changes are not passed downstream.

We've decided to retain the behaviour where validations are skipped as it feels like this functionality shouldn't be blocked via an invalid role or organisation.

We may want to come back and discuss if this is appropriate after the fix is rolled out.

However, save!(validate: false) does trigger the callbacks so the roles organisations will continue to be republished.

### DIFF
--- a/app/controllers/admin/cabinet_ministers_controller.rb
+++ b/app/controllers/admin/cabinet_ministers_controller.rb
@@ -60,7 +60,9 @@ private
     return unless params.include?(key)
 
     params[key]["ordering"].keys.each do |id|
-      Role.where(id:).update_all("#{column}": params[key]["ordering"][id.to_s])
+      role = Role.find(id)
+      role.assign_attributes({ "#{column}": params[key]["ordering"][id.to_s] })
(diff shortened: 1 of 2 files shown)


beccapearce (approved):
Looks good!

beccapearce:
I think this page now takes a while to load after clicking update order. Maybe a good thing to look into going forward is how to speed this up or do it asynchronously.

We should test the timings on integration to make sure it doesn't time out.

beccapearce (changes requested):
We should move this into workers to stop page time outs.

davidgisbey:
Minister reshuffles don't get pushed downstream without the sitewide setting for it being switched on by design so i think we can close this PR.

PR #8228, opened 2023-09-08
Ensure org is republished after reordering people

davidgisbey:
## Description 

At the moment, update_column is used to ensure that the uniq validation on ordering is not triggered while the updates take place.

However, this also skips any callbacks which handle republishing the org.

I think it makes sense to retain this behaviour and manually republish the org, so the org isn't republished for each role that is reordered.


beccapearce (approved):
This looks good to me!

PR #8234, opened 2023-09-11
Use govuk date component for review dates

ryanb-gds:
### What

We are gradually replacing all usages of the select-based date inputs with the GOV.UK Publishing date input component, which uses text inputs. This PR applies that change to the review date field on review reminders.

### Why

The Publishing date component is backed by user research which suggests text inputs are more efficient for users when they are entering arbitrary dates.

### Development Notes

As part of this PR I have included a refactoring to abstract the date validation logic into its own concern. The concern can be applied as follows:

```ruby
class MyModel < ApplicationRecord
  include DateValidation
  date_attributes :my_date_attribute, :my_other_date_attribute
end
```
Now, if a user provides an incorrectly formatted or invalid date for any of the specified attributes, an error message will be added to the model.

### DIFF
--- a/app/models/date_validation.rb
+++ b/app/models/date_validation.rb
@@ -1,0 +1,48 @@
+module DateValidation
+  extend ActiveSupport::Concern
+
+  attr_reader :invalid_date_attributes
+
+  included do
+    validates_with DateValidator
+
+    # @param attribute Symbol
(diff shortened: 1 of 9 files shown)


ryanb-gds:
Closing for now as I've found a bug in how it interacts with review date validation

PR #8236, opened 2023-09-11
Use govuk date component for review dates

ryanb-gds:
### What

We are gradually replacing all usages of the select-based date inputs with the GOV.UK Publishing date input component, which uses text inputs. This PR applies that change to the review date field on review reminders.

### Why

The Publishing date component is backed by user research which suggests text inputs are more efficient for users when they are entering arbitrary dates.

### Development Notes

As part of this PR I have included a refactoring to abstract the date validation logic into its own concern. The concern can be applied as follows:

```ruby
class MyModel < ApplicationRecord
  include DateValidation
  date_attributes :my_date_attribute, :my_other_date_attribute
end
```
Now, if a user provides an incorrectly formatted or invalid date for any of the specified attributes, an error message will be added to the model.

I also had to do some work to ensure that the presence error does not also appear in cases where the date attribute was set to nil because it was invalid

### DIFF
--- a/app/models/date_validation.rb
+++ b/app/models/date_validation.rb
@@ -1,0 +1,48 @@
+module DateValidation
+  extend ActiveSupport::Concern
+
+  attr_reader :invalid_date_attributes
+
+  included do
+    validates_with DateValidator
+
+    # @param attribute Symbol
+    # @param date Hash|Date|DateTime|String|nil
+    # @return Hash|Date|DateTime|String|nil
+    def pre_validate_date_attribute(attribute, date)
+      if date.is_a?(Hash)
+        begin
+          raise ArgumentError if date.values.any?(&:nil?)
+          raise ArgumentError if date[1].to_i.zero? || date[2].to_i.zero? || date[3].to_i.zero?
+
+          Date.new(date[1], date[2], date[3])
+        rescue ArgumentError
+          @invalid_date_attributes = [] if @invalid_date_attributes.nil?
+          @invalid_date_attributes << attribute
+          date = nil
+        end
+      end
+      date
+    end
+  end
+
+  class_methods do
+    def date_attributes(*attributes)
+      attributes.each do |attribute|
+        define_method("#{attribute}=") do |value|
+          super(pre_validate_date_attribute(attribute, value))
+        end
+      end
+    end
+  end
+
+  class DateValidator < ActiveModel::Validator
+    def validate(record)
+      return if record.invalid_date_attributes.nil?
+
+      record.invalid_date_attributes.each do |date_attribute|
+        record.errors.add date_attribute, "must be a valid date in the correct format"
+      end
+    end
+  end
+end
--- a/app/models/date_validation.rb
+++ b/app/models/date_validation.rb
@@ -27,11 +27,8 @@ module DateValidation
   def pre_validate_date_attribute(attribute, date)
     if date.is_a?(Hash)
       begin
-        raise ArgumentError if date.values.any?(&:nil?)
-        raise ArgumentError if date[1].to_i.zero? || date[2].to_i.zero? || date[3].to_i.zero?
-
         Date.new(date[1], date[2], date[3])
-      rescue ArgumentError
+      rescue ArgumentError, TypeError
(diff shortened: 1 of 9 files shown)


app/models/date_validation.rb
```diff
@@ -0,0 +1,62 @@
...
+        end
+      end
+    end
+  end
+
+  def pre_validate_date_attribute(attribute, date)
+    if date.is_a?(Hash)
+      begin
+        raise ArgumentError if date.values.any?(&:nil?)
+        raise ArgumentError if date[1].to_i.zero? || date[2].to_i.zero? || date[3].to_i.zero?
```
ollietreend, line 31: This might be more succinct:

```suggestion
        raise ArgumentError if date.values.any? { |v| v.to_i.zero? }
```

But just so I understand: what scenario is this protecting against?

These values should already have been cast to integers at this point, which is why `.to_i` hasn't been called on line 33 (i.e. the `Date.new` line). Non-numeric values will have become `0`.

The `Date.new` constructor already rejects `0` input values:

```ruby
Date.new(2023, 1, 0)
# => Date::Error: invalid date
```

So I'm not sure if this guard clause protects against a scenario that wasn't already covered? 🤔

test/unit/app/models/date_validation_test.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  class StubModel < StubBaseRecord
+    include DateValidation
+    validates :some_date, presence: true
+    date_attributes :some_date
+  end
+
+  test "should be valid when date attribute is a valid date" do
+    model = StubModel.new(some_date: { 1 => 2023, 2 => 9, 3 => 10, 4 => 0, 5 => 0 })
```
ollietreend, line 17: What do keys `4` and `5` represent in this hash – hours and minutes?

I'm not sure the `DateValidation` concern will handle DateTime values correctly right now – it only expects 3 parameters to be passed in, and it calls `Date.new(date[1], date[2], date[3])`.

Does this concern need to be made compatible with DateTime fields too?

ollietreend (commented):
This is looking good – just a couple of comments/questions from me.

app/models/date_validation.rb
ryanb-gds: I used a loop originally, then I had to do this: https://github.com/alphagov/whitehall/pull/8221

ryanb-gds: I've found a better way to do this by catching TypeErrors which are triggered when a non-integer value is passed to the date constructor. Update to follow

test/unit/app/models/date_validation_test.rb
ryanb-gds: Because we never pass the result of the date instantiation on to the model we don't actually need to worry about it, the original time values will still be used, and other validation rules will get the original data. As you mentioned on Slack, if we were to use text inputs for the time fields we would need to validate those too.

test/unit/app/models/date_validation_test.rb
```diff
@@ -0,0 +1,41 @@
...
+    model = StubModel.new(some_date: { 1 => 2023, 2 => "January", 3 => 20 })
+    assert_not model.valid?
+  end
+
+  test "should not validate presence if date was invalid" do
+    model = StubModel.new(some_date: { 1 => 2023, 2 => "January", 3 => 20 })
+    model.valid?
+    assert model.errors.where(:some_date, :blank).empty?
+  end
+end
```
ollietreend, line 41: We just talked about adding a test like:

```ruby
model = StubModel.new(some_date: { 1 => 2023, 2 => "January", 3 => 20 })
assert_not model.valid?
model.some_date = { 1 => 2023, 2 => 1, 3 => 20 }
assert model.valid?
```

Which I don't think will pass right now, given the fact that the 'valid' path of the setter doesn't clear/reset the validation state.

ryanb-gds: I've added a test that covers this

ollietreend (approved):
Nice! 🚀 This looks great. Thanks @ryanb-gds!

PR #8242, opened 2023-09-12
Move preview edition functionality into a View Component

davidgisbey:
## Description

We've had quite a few bugs historically around previewing documents. This moves out the logic for previewing documents into a view component and unit tests. I have changed a couple of things:

-  I've updated the logic about when the preview link should show for the default locale.

It's my understanding that this should always show, regardless of edition type. There was some weird behaviour going on where it was checking for a translation and if it didn't find one, it wouldn't render the link. In reality all translatable editions (all edition types) have an english translation by default. However, they don't when created by the factories. This was causing the link not to show unless a translation was specifically created.

Either way i'm pretty sure we always want to show this link.

- An extra space was present before `(opens in new tab)` when there was only 1 translation. This has been removed
- I've added   `(opens in new tab)` to the end of the link for non-English translations when multiple translations are present

Having this show for one translation but not others, when the behaviour is the same doesn't really make sense.

## Before

image

## After

image


app/components/admin/editions/show/preview_component.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Admin::Editions::Show::PreviewComponent < ViewComponent::Base
+  include Admin::UrlOptionsHelper
+
+  attr_reader :edition
```
ryanb-gds, line 6: Does this actually need an `attr_reader`? Is that just a style thing so we don't have to use `@` everywhere inside the class?

ryanb-gds (approved):
Great to have this additional test coverage ⭐

app/components/admin/editions/show/preview_component.rb
davidgisbey: Yea i should probably actually define it as private though so it's not available on the public interface. Will sort that now 👍 

Thanks for the review

PR #8243, opened 2023-09-12
530 reordering groups

davidtrussler:
[Trello](https://trello.com/c/tvSssCqa/530-allow-for-the-reordering-of-groups)

The changes in this PR
- Add the "Reorder group" link to the Group landing page
- Creates the new reorder page

Screenshots below

|Page|Screenshot|
|-|-|
|Landing page|Screenshot 2023-09-12 at 15 09 53|
|Reorder page|Screenshot 2023-09-12 at 15 10 08|
|Landing page after reorder|Screenshot 2023-09-12 at 15 10 25|

### DIFF
--- a/app/controllers/admin/document_collection_groups_controller.rb
+++ b/app/controllers/admin/document_collection_groups_controller.rb
@@ -1,6 +1,7 @@
 class Admin::DocumentCollectionGroupsController < Admin::BaseController
   before_action :load_document_collection
   before_action :load_document_collection_group, only: %i[confirm_destroy destroy edit update show]
+  before_action :check_new_design_system_permissions, only: %i[show order reorder]
   layout :get_layout
 
   def index
@@ -14,9 +15,7 @@ class Admin::DocumentCollectionGroupsController < Admin::BaseController
     render_design_system(:index, :legacy_index)
   end
 
-  def show
-    forbidden! unless new_design_system?
-  end
+  def show; end
 
   def new
     @group = @collection.groups.build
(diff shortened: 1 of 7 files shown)


nnagewad:
Looks great from the screenshots. @davidtrussler let me know if you want to do a 1-2-1 demo session sometime today or tomorrow

app/controllers/admin/document_collection_groups_controller.rb
```diff
@@ -47,6 +47,17 @@ def update
     render_design_system(:edit, :legacy_edit)
   end
 
+  def reorder; end
```
minhngocd, line 50: For both reorder, and order, do we need to also add some forbidden checks unless users have ability to preview design system? i feel like with this, people can just access the order page through URL regardless of permissions

app/views/admin/document_collection_groups/reorder.html.erb
```diff
@@ -0,0 +1,29 @@
...
+          {
+            id: group.id,
+            title: group.heading
+          }
+        end
+      } %>
+
+      <div class="govuk-button-group govuk-!-margin-bottom-6">
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Save",
```
minhngocd, line 23: Do we need to add Google Analytics to these buttons?

minhngocd:
Hi @davidtrussler - thank you! it looks and works great! 
A couple of questions from me for the bits above ☝️

app/controllers/admin/document_collection_groups_controller.rb
davidtrussler: Yes, that makes sense. I've added that for both of those (think it's working correctly)

app/views/admin/document_collection_groups/reorder.html.erb
davidtrussler: Yes, good point. I've added that, I think the values are what would be expected.

minhngocd (approved):

PR #8244, opened 2023-09-12
Provide a relative date validator

ryanb-gds:
At present, each model duplicates date validation logic, or relies on setting constraints on select inputs which won't be viable when me move towards the design system date component. It therefore seems sensible to provide a single validator which can be responsible for validating dates relative to other dates.

I will gradually apply the relative date validator to other fields whilst working on replacing the date components

I had to disable validation on the review reminder factory in circumstances where the `review_at` date was being set to the current day (i.e. the review is due), because generally users can only set future dates. I tweaked the factory traits a bit so that they can be composed together for flexibility rather than duplicating the logic.

### DIFF
--- a/test/factories/review_reminder.rb
+++ b/test/factories/review_reminder.rb
@@ -11,23 +11,20 @@ FactoryBot.define do
 
     trait(:reminder_due) do
       review_at { Time.zone.today }
-      reminder_sent_at { nil }
+      # Attempting to set review_at to today usually causes a validation error, because review date should be set to a future date
+      to_create { |instance| instance.save(validate: false) }
     end
 
     trait(:reminder_sent) do
-      review_at { Time.zone.today }
       reminder_sent_at { Time.zone.now }
     end
 
-    trait(:due_but_never_published) do
-      review_at { Time.zone.today }
-      reminder_sent_at { nil }
+    trait(:document_never_published) do
       document { create(:document, editions: [build(:draft_edition, first_published_at: nil)]) }
     end
 
     trait(:not_due_yet) do
       review_at { 1.month.from_now }
-      reminder_sent_at { nil }
     end
   end
 end
--- a/test/unit/app/validators/relative_date_validator_test.rb
+++ b/test/unit/app/validators/relative_date_validator_test.rb
@@ -1,0 +1,70 @@
+require "test_helper"
+
+class RelativeDateValidatorTest < ActiveSupport::TestCase
+  class StubModel
+    include ActiveModel::API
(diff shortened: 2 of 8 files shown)


test/factories/review_reminder.rb
```diff
@@ -11,23 +11,20 @@
...
     trait(:reminder_sent) do
-      review_at { Time.zone.today }
       reminder_sent_at { Time.zone.now }
     end
 
     trait(:due_but_never_published) do
-      review_at { Time.zone.today }
-      reminder_sent_at { nil }
       document { create(:document, editions: [build(:draft_edition, first_published_at: nil)]) }
     end
```
ollietreend, line 24: It might be worth renaming this to reflect it no longer affects the 'due' date.

There's an argument to say this should be a trait on the `documents` factory rather than here, but I think it's only really relevant to this set of unit tests.

```suggestion
    trait(:document_never_published) do
      document { create(:document, editions: [build(:draft_edition, first_published_at: nil)]) }
    end
```

test/unit/app/validators/relative_date_validator_test.rb
```diff
@@ -0,0 +1,66 @@
...
+    assert_equal before_message, model.errors[:some_date].first
+  end
+
+  test "after error message can be overriden" do
+    after_message = "should be in the future"
+    model = StubModel.new(some_date: Time.zone.local(2022, 1, 1, 1, 1))
+    validator = RelativeDateValidator.new(attributes: :some_date, after: -> { Time.zone.local(2022, 1, 1, 1, 2) }, after_message:)
+    validator.validate(model)
+    assert_equal after_message, model.errors[:some_date].first
+  end
```
ollietreend, line 65: It might be nice to have tests to cover the _default_ message too. The other tests only `assert_not_empty model.errors` rather than the message itself.

ollietreend (commented):
Nice refactor 👍🏻 I've just dropped in a couple of thoughts, but they're not deal-breakers.

test/factories/review_reminder.rb
ryanb-gds: Ah yes, good spot

test/unit/app/validators/relative_date_validator_test.rb
ryanb-gds: Good point, there's a bug in there because I'm not calling the lambda before attempting to cast it to a string 😱

ollietreend (approved):
Nice 👍🏻 looks good!

PR #8245, opened 2023-09-13
Push content with non legacy endpoints to pub api

lauraghiorghisor-tw:
These changes implement sending non-legacy urls in the format `/media/:id/:filename` to the publishing api, for attachments created with the `use_non_legacy_endpoints` flag. This replaces previous behaviour where attachments have a legacy url such as `"/government/uploads/system/uploads/attachment_data/file/:id/:filename".

### URL generation
URLs are composed in `asset_manager_storage`, based on an ID  and a `filename` returned from `asset-manager` upon asset creation. We opted to preemptively save the filename in the `whitehall` database in order to cater for any future changes around what versions `carrierwave` can generate. 

### Previous implementation
In the previous implementation, the `legacy_url_path` was always known by `whitehall` and "dictated" to `asset-manager`. Thus, rendering urls in `whitehall` or building the payload for `publishing-api` were never an issue. Nonetheless, this still meant that on occasion assets could be unavailable; in this case, placeholders would be seen on page where a `404` was encountered (e.g. if asset was not yet available in `asset-manager`). 

### Current solution
Now that the asset model is in place, `whitehall` has a better grasp as to what assets have been successfully uploaded to `asset-manager`. But since there is now a dependency on `asset-manager` to return the asset data, we face the issue of assets not being available either when rendering in `whitehall` or sending data to the `publishing-api`. We are in this case unable to generate urls in the way the legacy path was done before. We have addressed this issue by: 
- replacing the `<a>` tag with an unclickable `<span>` in the views in `whitehall`
- filtering the attachments whose assets are not yet available from the attachments payload for `publishing-api` 
- filtering the attachments whose assets are not yet available from the `govspeak` renderer payload (which fixes both the `whitehall` preview and the `publishing-api` rendering)
- the `UPLOADING` label now shows if any of the assets are still uploading/missing (not just the original file).

### Blocking publishing
There was blocking behaviour already in place at the time of publishing: if any original assets were pending upload, then a publisher would not have been able to publish an edition and they would have seen an error message on screen. We have extended this behaviour so that publishing is blocked if any assets are missing (for file attachments, this would be pdf with a missing thumbnail). 

With the blocking and filtering behaviours in place, in order to ensure that the data will be sent to the `publishing-api` when assets do become available, we are triggering the edition `draft_updater` from the `CreateAssetWorker`.

### Feature flag changes
In order to support the release of these changes, we now save the flag `use_non_legacy_endpoints` as a database field on `AttachmentData` rather than using the in-memory field (derived from user permissions) that we originally opted for.  We are planning to release the changes and then run the migration (which will populate assets for all `AttachmentData`). The flag will also signify that the migration has not yet been run and that the legacy flow should be used in the interim. The migration will set the flag as true in the database for the `AttachmentData`s being migrated.

After introducing the new database flag, we have additionally made changes to the delete and update workers so that the behaviour is toggled based on the flag and not the presence of assets as it was before. 

[Trello card](https://trello.com/c/CrfXnC5c/155-story-whitehall-should-push-edition-content-with-non-legacy-asset-urls-to-publish-api-and-content-api-attachmentdata)

### DIFF
--- a/app/helpers/attachments_helper.rb
+++ b/app/helpers/attachments_helper.rb
@@ -60,7 +60,9 @@ module AttachmentsHelper
   end
 
   def block_attachments(attachments = [], alternative_format_contact_email = nil)
-    attachments.map do |attachment|
+    attachments
+      .select { |attachment| !attachment.file? || attachment.attachment_data.all_asset_variants_uploaded? }
+      .map do |attachment|
       render(
         partial: "govuk_publishing_components/components/attachment",
         locals: {
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -63,7 +63,9 @@ module GovspeakHelper
   end
 
   def prepare_attachments(attachments, alternative_format_contact_email)
-    attachments.map do |attachment|
+    attachments
+      .select { |attachment| !attachment.file? || attachment.attachment_data.all_asset_variants_uploaded? }
+      .map do |attachment|
       attachment_component_params(attachment, alternative_format_contact_email:)
     end
   end
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -16,7 +16,7 @@ class AttachmentData < ApplicationRecord
   validates :file, presence: true
   validate :file_is_not_empty
 
-  attr_accessor :to_replace_id, :attachable, :use_non_legacy_endpoints
+  attr_accessor :to_replace_id, :attachable
 
   belongs_to :replaced_by, class_name: "AttachmentData"
   validate :cant_be_replaced_by_self
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -63,7 +63,7 @@ class AttachmentData < ApplicationRecord
 
   def all_asset_variants_uploaded?
     if use_non_legacy_endpoints
-      return pdf? ? assets.size == 2 : assets.size == 1
+      return assets.size == (pdf? ? 2 : 1)
     end
 
     uploaded_to_asset_manager?
@@ -185,7 +185,7 @@ class AttachmentData < ApplicationRecord
     attachable && attachable.respond_to?(:auth_bypass_id) ? [attachable.auth_bypass_id] : []
   end
 
-  private
+private
 
   def filtered_attachments(include_deleted_attachables: false)
     if include_deleted_attachables
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -27,15 +27,15 @@ class AttachmentData < ApplicationRecord
   attribute :present_at_unpublish, :boolean, default: false
 
   def filename
-    url && File.basename(url)
+    file.present? && file.file.filename
   end
 
   def filename_without_extension
-    url && filename.sub(/.[^.]*$/, "")
+    filename && filename.sub(/.[^.]*$/, "")
   end
 
   def file_extension
-    File.extname(url).delete(".") if url.present?
+    File.extname(filename).delete(".") if filename.present?
   end
 
   def pdf?
@@ -47,7 +47,9 @@ class AttachmentData < ApplicationRecord
   end
 
   def csv?
-    file_extension.casecmp("csv").zero?
+    return file_extension.casecmp("csv").zero? if file_extension
+
+    false
   end
 
   # Is in OpenDocument format? (see https://en.wikipedia.org/wiki/OpenDocument)
@@ -57,6 +59,14 @@ class AttachmentData < ApplicationRecord
 
   def indexable?
     AttachmentUploader::INDEXABLE_TYPES.include?(file_extension)
+  end
+
+  def all_asset_variants_uploaded?
+    if use_non_legacy_endpoints
+      return pdf? ? assets.size == 2 : assets.size == 1
+    end
+
+    uploaded_to_asset_manager?
   end
 
   def update_file_attributes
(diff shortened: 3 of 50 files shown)


app/helpers/attachments_helper.rb
```diff
@@ -60,7 +60,9 @@ def attachment_component_params(attachment, alternative_format_contact_email: ni
   end
 
   def block_attachments(attachments = [], alternative_format_contact_email = nil)
-    attachments.map do |attachment|
+    attachments
```
tunylund, line 63: could make the calling code use `renderer.block_attachments(item.attachments_ready_for_publishing...)`

app/helpers/govspeak_helper.rb
```diff
@@ -63,7 +63,9 @@ def prepare_images(images)
   end
 
   def prepare_attachments(attachments, alternative_format_contact_email)
-    attachments.map do |attachment|
+    attachments
+      .select { |attachment| !attachment.file? || attachment.attachment_data.all_asset_variants_uploaded? }
```
tunylund, line 67: could make the calling code use `prepare_attachments(item.attachments_ready_for_publishing...)`

app/models/attachment_data.rb
```diff
@@ -27,15 +27,15 @@ class AttachmentData < ApplicationRecord
   attribute :present_at_unpublish, :boolean, default: false
 
   def filename
-    url && File.basename(url)
+    file.present? && file.file.filename
   end
 
   def filename_without_extension
```
tunylund, line 33: it looks like this function could be deleted as it is only used in 2 tests

app/presenters/publishing_api/payload_builder/attachments.rb
```diff
@@ -19,7 +19,9 @@ def call
       def attachments
         items.flat_map do |item|
           if item
-            item.attachments.map(&:publishing_api_details)
+            item.attachments
+                .select { |attachment| !attachment.file? || attachment.attachment_data.all_asset_variants_uploaded? }
```
tunylund, line 23: could use `attachments_ready_for_publishing`

lib/whitehall/asset_manager_storage.rb
```diff
@@ -38,11 +38,14 @@ def store!(file)
 
   def retrieve!(identifier)
     asset_path = uploader.store_path(identifier)
-    File.new(asset_path)
+    File.new(asset_path, uploader.model, uploader.version_name)
   end
 
   class File
-    def initialize(asset_path)
+    def initialize(asset_path, model = nil, version_name = nil)
```
tunylund, line 45: Since the initializer has changed, shouldn't the `store!` method also return a `File.new(uploader.store_path, uploader.model, uploader.version_name)`?

lib/whitehall/asset_manager_storage.rb
```diff
@@ -51,17 +54,27 @@ def delete
     end
 
     def url
-      URI.join(Plek.asset_root, Addressable::URI.encode(@legacy_url_path)).to_s
-    end
-
-    def filename
-      ::File.basename(@legacy_url_path)
+      if @model && @model.respond_to?("use_non_legacy_endpoints") && @model.use_non_legacy_endpoints
+        return nil unless @model.assets.any?
```
tunylund, line 58: this return seems unnecessary line 62 has a nilcheck already

lib/whitehall/asset_manager_storage.rb
```diff
@@ -51,17 +54,27 @@ def delete
...
+
+        asset_variant = @version ? Asset.variants[@version] : Asset.variants[:original]
+        asset = @model.assets.where(variant: asset_variant).first
```
tunylund, line 61: Instead of a database lookup, this just loop through assets and pick the correct variant. That way the test doesn't need to do `model.save!`

test/unit/lib/whitehall/asset_manager_storage_test.rb
```diff
@@ -265,8 +232,48 @@ class Whitehall::AssetManagerStorage::FileTest < ActiveSupport::TestCase
...
+  test "constructs the url of the file using the assets root and legacy url path" do
+    expected_asset_url = URI.join("http://assets-host", @asset_url_path).to_s
+
+    assert_equal expected_asset_url, @file.url
+  end
+
+  test "returns file url using asset_manager_id when the model has the original asset" do
+    model = build(:attachment_data_with_asset)
+    model.use_non_legacy_endpoints = true
+    model.save!
```
tunylund, line 247: I reckon this save! is unnecessary. The code in the `url` method does a database lookup when it doesn't have to.

tunylund (changes requested):
Excellent commit messages and pull request message and super thorough testing ❤️

app/helpers/attachments_helper.rb
tunylund: As discussed, this is better for now, as we the alternative would force us to write excessive tests or refactor the abstractions in Presenters to better represent the requirements - not the right time for that.

app/helpers/govspeak_helper.rb
tunylund: As discussed, this is better for now, as we the alternative would force us to write excessive tests or refactor the abstractions in Presenters to better represent the requirements - not the right time for that.

test/unit/lib/whitehall/asset_manager_storage_test.rb
tunylund: As discused, carrierwave magickery makes it so that `build(:attachment_data_with_asset)` causes the `file` instance to be an instance of `Carrierwave::File`. Whereas if we do save and reload, the file instance ends up being `AssetManagerStorage::File`. This is sad and makes save and reload necessary for these tests.

tunylund (approved):
as discussed, this looks good 👍

PR #8250, opened 2023-09-14
Migrate edition published before date to design system

ryanb-gds:
### What

Replace the Whitehall date component with the GOVUK design system date input for the first published at field.

Also includes a tweak to the relative date validator to make the date constraints inclusive, and a bug fix to persist the state of the document status radio toggle when validation fails.

Happily these changes allow us to delete one of the custom validators which we no longer need :tada: 

### Why

User research indicates that the design system date component provides a superior user experience.

### Screenshots

Before:

Screenshot 2023-09-14 at 11 15 53

After:
Screenshot 2023-09-14 at 11 16 27

After with validation error:

Screenshot 2023-09-14 at 11 25 01

### Links


app/validators/relative_date_validator.rb
```diff
@@ -2,11 +2,11 @@ class RelativeDateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.nil?
 
-    if options[:before].present? && options[:before].call <= value
-      record.errors.add(attribute, :date_before, message: options[:before_message] || "must be before #{options[:before].call}")
+    if options[:before].present? && options[:before].call < value
```
ollietreend, line 5: @ryanb-gds and I chatted on Slack about potentially refactoring this Relative Date Validator class to make use of [Ranges](https://ruby-doc.org/3.2.2/Range.html).

Just noting this here for visibility – but I don't expect that to be changed in this PR 🙂

ollietreend (approved):
Looks good to me 👍🏻

PR #8251, opened 2023-09-14
Remove placeholder logic

brucebolt:
We were previously sending some content items to Publishing API with a schema type prefixed `placeholder_` to indicate that content item was still rendered by Whitehall.

This is no longer the case, so we can remove that logic.

The `GenericEditionPresenter` is used solely in tests, reverting this to use a generic `placeholder` schema as the logic will be removed from Publishing API in https://github.com/alphagov/publishing-api/pull/2504.

This will need to be updated again when the `placeholder` schema is completely removed.

[Trello card](https://trello.com/c/xAZ7nU8h)

### DIFF
--- a/app/presenters/publishing_api/generic_edition_presenter.rb
+++ b/app/presenters/publishing_api/generic_edition_presenter.rb
@@ -19,7 +19,7 @@ module PublishingApi
         document_type:,
         public_updated_at: item.public_timestamp || item.updated_at,
         rendering_app: Whitehall::RenderingApp::GOVERNMENT_FRONTEND,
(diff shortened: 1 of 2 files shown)


jkempster34 (approved):
You mention that this will need updating further once the placeholder content schema is removed, do you mean beyond this? https://github.com/alphagov/whitehall/pull/8252

brucebolt:
> You mention that this will need updating further once the placeholder content schema is removed, do you mean beyond this? #8252

Nope.  Other than removing `GenericEditionPresenter` (which feels out of our team's scope), I think your PR covers the change that's needed here.

PR #8253, opened 2023-09-14
Switch order of parameters in 'if' statement

ollietreend:
I've switched the order of these parameters in the 'if' statement.

Whilst only a subtle change, this should allow clicks of the "Save" button (i.e. `params[:save].present?`) to avoid calling out to the Publishing API (i.e. `@edition.has_been_tagged?`).

Despite being a relatively small optimisation, it should slightly speed up these requests by removing an external API call.

Clicks of the "Save and continue" button will still trigger that Publishing API call, which is required to work out which flash message to show.

---

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -316,7 +316,7 @@ private
   end
 
   def saved_confirmation_notice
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
Nice catch :+1:

PR #8258, opened 2023-09-15
CSV previews to work with modern urls

syed-ali-tw:
**Fix Asset Preview to work with modern url**

This PR is trying to fix 2 problems related to csv preview

1. To move away from legacy url path, we needed csv previews to work with `/media/:id/:filename` endpoint in asset manager. So we are publishing csv preview url to be `/media/:id/:filename` instead of legacy_url.

2. Currently gov-speak body preview for csv does not work correctly when we try to view online, the reason being it is not configured correctly to use frontend domain. So we have made the changes to allow it to use frontend domain

We added check for non legacy endpoints so that we can return the non legacy preview url correctly.

[trello](https://trello.com/c/LV4TnjlP/176-story-csv-previews-should-work-with-modern-urls)

### DIFF
--- a/app/helpers/attachments_helper.rb
+++ b/app/helpers/attachments_helper.rb
@@ -14,7 +14,14 @@ module AttachmentsHelper
   end
 
   def preview_path_for_attachment(attachment)
-    "/government/uploads/system/uploads/attachment_data/file/#{attachment.attachment_data.id}/#{attachment.filename}/preview"
+    if attachment.attachment_data.use_non_legacy_endpoints
+      if attachment.attachment_data.all_asset_variants_uploaded?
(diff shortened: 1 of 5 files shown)


tunylund (approved):

richardTowers:
A point for next week, but we should try to avoid using absolute URLs, and we should definitely avoid `Plek.external_url_for("assets-origin")`.

Absolute URLs lead to confusing environment sync issues - if something's published in production, and then the nightly sync moves it down into content-store in staging and integration, the link will still point at production, which is not what we want.

assets-origin is supposed to be a behind-the-scenes domain. User facing queries should always hit the CDN on assets.publishing.service.gov.uk (which then forwards requests to assets-origin). Users shouldn't hit assets-origin directly.

Let's have a chat about the need for this next week and see if we can work out a better approach.

PR #8261, opened 2023-09-18
Replace relative date validator with built-in Rails validation

ryanb-gds:
Turns out we can pass a Proc to the inclusion validator that will be evaluated at runtime, making it possible to validate against a range. Therefore, our relative date validator is no longer required, we can use Rails built-in validation instead. Hurrah!

The validation message for the `first_published_at` field on the edition model had to change to reflect the fact that we can only provide a single message, but there are plans to revisit error messages in general anyway.


app/models/edition.rb
```diff
@@ -55,8 +55,8 @@ class Edition < ApplicationRecord
   validates :summary, presence: true, if: :summary_required?, length: { maximum: 65_535 }
   validates :previously_published, inclusion: { in: [true, false], message: "You must specify whether the document has been published before" }
   validates :first_published_at, presence: true, if: -> { previously_published || published_major_version }
-  validates :first_published_at, relative_date: { after: -> { Date.parse("1900-01-01") }, before: -> { Time.zone.now }, before_message: "can't be set to a future date" }
-  validates :scheduled_publication, relative_date: { after: -> { Time.zone.now }, after_message: "must be in the future" }, if: :draft?
+  validates :first_published_at, inclusion: { in: proc { Date.parse("1900-01-01")..Time.zone.now }, message: "must be between 1/1/1900 and the present" }, if: :draft?, allow_blank: true
```
ollietreend, line 58: Just wondering if this wording might be more readable? I guess the likely scenario is that someone's given a future date in rather than a historic date – and the 'not before 1900' constraint seems unlikely to be an issue, but worth pointing out.

Just an opinion – no need to use it 🙂

Only a suggestion - feel free to ignore.

```suggestion
  validates :first_published_at, inclusion: { in: proc { Date.parse("1900-01-01")..Time.zone.now }, message: "must be in the past (but no earlier than 1900)" }, if: :draft?, allow_blank: true
```

ollietreend (approved):
Looks good to me!

I've added one small suggestion about wording of one of the validation messages. But not a blocker.

app/models/edition.rb
ryanb-gds: I think, as we discussed briefly last week, what I'll do is once I've finished replacing all of these I'll get Rosa to go through and do a sweep of the error messages and we'll sort the lot out at once, if that's ok

PR #8265, opened 2023-09-18
Fix ConsultationResponseForm & CallForEvidenceResponseForm bug

davidgisbey:
## Description

There's a bug at the moment, where you can persist a consultation/call for evidence response form without uploading a file. You can do this by adding a title to the `Downloadable response form title` field, but not adding a file.

This PR addresses this bug and does some minor refactors. It:

- adds a presence validation to the consultation & call of evidence response form to check that response form data is present
- ensures that the new validation error is displayed above the file input when triggered
- refactors the forms to use `fields_for` instead using a bunch of `hidden_field_tag`'s 
- handles building dependencies in the controller, rather than constructing them within the view

### Call for evidence

image

image

### Consultations

image

image

 ## Trello card 

https://trello.com/c/VY1YoFIa/1544-whitehall-crashes-when-consultation-response-form-file-is-missing

### DIFF
--- a/app/controllers/admin/calls_for_evidence_controller.rb
+++ b/app/controllers/admin/calls_for_evidence_controller.rb
@@ -39,4 +39,11 @@ private
   def document_can_be_previously_published
     false
   end
+
+  def build_edition_dependencies
+    super
+    participation = @edition.call_for_evidence_participation || @edition.build_call_for_evidence_participation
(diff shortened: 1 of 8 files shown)


ryanb-gds (approved):
Looks good to me, although to be honest I'm still somewhat unclear about the modelling of the response/participation/feedback area. Feels like we could do with a nice little entity relationship diagram explaining that. Nice refactor to use the `fields_for` method, that's a big improvement

PR #8266, opened 2023-09-18
Consultation date component migration

ryanb-gds:
### What

Migrate to GOVUK publishing date input component for consultation and consultation response forms. 

Also includes moving away from custom validation of open/close relative dates and a workaround for some peculiar Rails behaviour around handling non-numeric string inputs for the year fields.

### Why

User research indicates that the GOVUK publishing component date input provides a superior user experience to the Rails date helper component

### Screenshots

Before:

Consultation Opening Date
Screenshot 2023-09-18 at 16 46 18

Consultation Closing Date
Screenshot 2023-09-18 at 16 46 27

Consultation Response Published On
Screenshot 2023-09-18 at 16 46 43

After (with validation errors present):

Consultation Opening Date
Screenshot 2023-09-18 at 16 38 50

Consultation Closing Date
Screenshot 2023-09-18 at 16 38 55

Consultation Response Published On
Screenshot 2023-09-18 at 16 49 11

### DIFF
--- a/app/models/date_validation.rb
+++ b/app/models/date_validation.rb
@@ -26,6 +26,10 @@ module DateValidation
     @invalid_date_attributes = Set.new if @invalid_date_attributes.nil?
     if date.is_a?(Hash)
       begin
+        # Rails will cast the year part of the date to 0 if the year input parameter is a non-numeric string
+        # This only seems to happen to the year part, other parts remain as strings
+        raise TypeError if (date[1]).zero?
+
         Date.new(date[1], date[2], date[3])
         @invalid_date_attributes.delete(attribute)
       rescue ArgumentError, TypeError
--- a/app/models/consultation.rb
+++ b/app/models/consultation.rb
@@ -9,9 +9,9 @@ class Consultation < Publicationesque
 
   validates :opening_at, presence: true, unless: ->(consultation) { consultation.can_have_some_invalid_data? }
   validates :closing_at, presence: true, unless: ->(consultation) { consultation.can_have_some_invalid_data? }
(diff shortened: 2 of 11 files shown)


baisa (approved):
LGTM!

app/models/date_validation.rb
```diff
@@ -26,6 +26,10 @@ def pre_validate_date_attribute(attribute, date)
     @invalid_date_attributes = Set.new if @invalid_date_attributes.nil?
     if date.is_a?(Hash)
       begin
+        # Rails will cast the year part of the date to 0 if the year input parameter is a non-numeric string
+        # This only seems to happen to the year part, other parts remain as strings
```
ollietreend, line 30: Just to explain why I think this is needed. I don't think it's specifically doing anything special for 'year'.

My understanding is that Rails calls `.to_i` on every value in the hash, so they're all cast to integers. None of them will remain as strings.

As you said, non-numeric strings will become `0`. In other words:

```ruby
"5".to_i == 5 # => true
"0".to_i == 0 # => true
"abc".to_i == 0 # => true
```

However, the `Date.new` constructor only accepts `0` for the year parameter:

- `0` is a valid year (as is any other integer, positive **or** negative)
- `0` **is not** a valid month (must be >= 1)
- `0` **is not** a valid day (must be >= 1)

I think this is why you only saw this causing problems for the year. Because non-numeric strings for month and day would have triggered a `TypeError` on the Date constructor already – whereas `0` is a valid value for the year.

ryanb-gds: Oh, that makes sense. I was trying to read through the attribute assignment code in Rails but it was a bit tricky to follow. Thanks for clarifying

PR #8268, opened 2023-09-18
Enable use non legacy endpoints for file attachments

lauraghiorghisor-tw:
This is the first release of the epic related to migrating `whitehall` to use non-legacy endpoints in `asset-manager`.

We are enabling the `use_non_legacy_endpoints` for all attachments that belong to an `AttachmentData`. The scope of this change includes all document types that include the `Attachable` concern. 

Whilst `csv` files pertain to this scope, they are a special case due to the `/preview` functionality so they will be released separately.

[Trello card](https://trello.com/c/lnpH3zBG/100-task-roll-out-attachmentdata-to-all-users-except-for-csvs)

### DIFF
--- a/app/controllers/admin/bulk_uploads_controller.rb
+++ b/app/controllers/admin/bulk_uploads_controller.rb
@@ -26,7 +26,7 @@ class Admin::BulkUploadsController < Admin::BaseController
     @bulk_upload.attachments_attributes = create_params[:attachments_attributes]
     @bulk_upload.attachments.each do |attachment|
       attachment.attachment_data.attachable = @edition
-      attachment.attachment_data.use_non_legacy_endpoints = use_non_legacy_endpoints?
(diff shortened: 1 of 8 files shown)


app/controllers/admin/bulk_uploads_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
     @bulk_upload.attachments_attributes = create_params[:attachments_attributes]
     @bulk_upload.attachments.each do |attachment|
       attachment.attachment_data.attachable = @edition
-      attachment.attachment_data.use_non_legacy_endpoints = use_non_legacy_endpoints?
+      attachment.attachment_data.use_non_legacy_endpoints = use_non_legacy_endpoints? unless attachment.attachment_data.csv?
```
tunylund, line 29: it would be a bit cleaner if the decision about csvs would be in the `use_non_legacy_endpoints?` method. it can take attachment_data as an argument.

PR #8273, opened 2023-09-19
Spike out using a form object to update an editions change note

davidgisbey:
## Description

This is a first attempt at implementing [form objects](https://dev.to/drbragg/rails-design-patterns-form-object-4d47). These will help us simplify building out editing over multiple pages.

I chose to use change notes for the spike as it's fairly small and self contained. It's also easily linked to from the document summary page.

One thing that i ran into that was slightly inconvenient in that there's already an `edition_change_notes_controller` which handles updating change notes for published editions. This was built in as it was a common 2ndline task.

I've got around this by naming this controller the `DraftEditionChangeNotesController` but tbh i think it probably makes more sense to rename the existing controller to something like `gds_admin_change_note_controller` or something similar and using `EditionChangeNotesController` for this. Or even better, start using subfolders so it could live in

`admin/editions/change_note_controller.rb` 

If we use subfolders i think it would probably be part of a larger refactor though.

For now though since this is mainly proof of concept, i've gone with the path of least resistance.

Likewise edit view is essentially copy paste of the code used in the edit page. I've not refactored it into a partial though, but it should be if we decide to go ahead and use this as some point.

## Screenshots

image

image

image

## Trello card

https://trello.com/c/DKQeargF/1716-spike-out-using-form-objects-to-perform-partial-validation-of-edition-input

### DIFF
--- a/app/controllers/admin/draft_edition_change_notes_controller.rb
+++ b/app/controllers/admin/draft_edition_change_notes_controller.rb
@@ -1,0 +1,36 @@
+class Admin::DraftEditionChangeNotesController < Admin::BaseController
+  before_action :find_edition, :enforce_permissions!, :redirect_to_show_page_unless_editable
+  layout "design_system"
+
+  def edit
+    @change_note_form = ChangeNoteForm.build_from_edition(@edition)
+  end
+
+  def update
+    @change_note_form = ChangeNoteForm.new(change_note_params)
+
+    if @change_note_form.save!(@edition)
+      redirect_to admin_edition_path(@edition), notice: "Change note updated successfully"
+    else
+      render :edit
+    end
+  end
+
+private
+
+  def find_edition
+    @edition = Edition.find(params[:edition_id])
+  end
+
+  def enforce_permissions!
+    enforce_permission!(:update, @edition)
+  end
+
+  def redirect_to_show_page_unless_editable
+    redirect_to admin_edition_path(@edition) unless @edition.editable?
+  end
+
+  def change_note_params
+    params.require(:change_note_form).permit(:minor_change, :change_note)
+  end
+end
--- a/app/forms/change_note_form.rb
+++ b/app/forms/change_note_form.rb
@@ -1,7 +1,9 @@
 class ChangeNoteForm
   include ActiveModel::Model
+  include ActiveModel::Attributes
 
-  attr_accessor :minor_change, :change_note
+  attribute :minor_change, :boolean, default: false
+  attribute :change_note
 
   validates :minor_change, inclusion: { in: [true, false] }
   validates :change_note, presence: true, if: :major_version?
@@ -14,8 +16,6 @@ class ChangeNoteForm
   end
 
   def save!(edition)
-    self.minor_change = ActiveModel::Type::Boolean.new.cast(minor_change)
-
     return false unless valid?
 
     if major_version?
--- a/app/forms/change_note_form.rb
+++ b/app/forms/change_note_form.rb
@@ -1,0 +1,39 @@
+class ChangeNoteForm
+  include ActiveModel::Model
+
+  attr_accessor :minor_change, :change_note
+
+  validates :minor_change, inclusion: { in: [true, false] }
+  validates :change_note, presence: true, if: :major_version?
+
+  def self.build_from_edition(edition)
+    new(
+      minor_change: edition.minor_change,
+      change_note: edition.change_note,
+    )
+  end
+
+  def save!(edition)
+    self.minor_change = ActiveModel::Type::Boolean.new.cast(minor_change)
+
+    return false unless valid?
+
+    if major_version?
+      edition.update!(
+        change_note:,
+        minor_change:,
+      )
+    else
+      edition.update!(
+        change_note: nil,
+        minor_change:,
+      )
+    end
+  end
+
+private
+
+  def major_version?
+    minor_change == false
+  end
+end
--- a/config/locales/en/admin/change_note_form.yml
+++ b/config/locales/en/admin/change_note_form.yml
@@ -1,0 +1,11 @@
+en:
+  activemodel:
+      errors:
+        models:
+          change_note_form:
+            attributes:
+              minor_change:
+                inclusion: Select whether this is a major or minor update
+              change_note:
+                blank: "Enter a change note"
+
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -172,6 +172,9 @@ Whitehall::Application.routes.draw do
         get :edit_slug, on: :member, controller: :edition_slug
         patch :update_slug, on: :member, controller: :edition_slug
 
+        get :edit_change_note, to: "draft_edition_change_notes#edit"
+        patch :update_change_note, to: "draft_edition_change_notes#update"
+
         collection do
           post :export
           get :confirm_export
--- a/test/forms/change_note_form_test.rb
+++ b/test/forms/change_note_form_test.rb
@@ -1,0 +1,53 @@
+require "test_helper"
+
+class ChangeNoteFormTest < ActiveSupport::TestCase
+  include ActionDispatch::TestProcess
+
+  test "is invalid when minor_change is not present" do
+    change_note_form = ChangeNoteForm.new(change_note: "This is present", minor_change: nil)
+    assert_not change_note_form.valid?
+  end
+
+  test "is invalid when it is a major change and change_note is not present" do
+    change_note_form = ChangeNoteForm.new(change_note: nil, minor_change: false)
+    assert_not change_note_form.valid?
+  end
+
+  test "is valid when it is a minor change and a change_note is not present" do
+    change_note_form = ChangeNoteForm.new(change_note: nil, minor_change: true)
+    assert change_note_form.valid?
+  end
+
+  test "is valid when it is a major change and change_note is present" do
+    change_note_form = ChangeNoteForm.new(change_note: "This is present", minor_change: false)
+    assert change_note_form.valid?
+  end
+
+  test "#save returns false and adds errors to the form object when invalid" do
+    change_note_form = ChangeNoteForm.new
+    edition = build(:edition)
+
+    assert_equal false, change_note_form.save!(edition)
+    assert_equal ["Minor change Select whether this is a major or minor update"], change_note_form.errors.full_messages
+  end
+
+  test "#save updates the edition successfully when it is a major change and a change note is present" do
+    edition = create(:edition)
+    change_note_form = ChangeNoteForm.new(change_note: "This is present", minor_change: "false")
+
+    change_note_form.save!(edition)
+
+    assert_equal "This is present", edition.reload.change_note
+    assert_equal false, edition.minor_change
+  end
+
+  test "#save updates the edition successfully and sets the change note to nil when it is a minor change" do
+    edition = create(:edition, change_note: "Lots of changes.")
+    change_note_form = ChangeNoteForm.new(change_note: "I'm still present", minor_change: "true")
+
+    change_note_form.save!(edition)
+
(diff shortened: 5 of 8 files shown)


app/forms/change_note_form.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def self.build_from_edition(edition)
+    new(
+      minor_change: edition.minor_change,
+      change_note: edition.change_note,
+    )
+  end
+
+  def save!(edition)
+    self.minor_change = ActiveModel::Type::Boolean.new.cast(minor_change)
```
davidgisbey, line 17: this feels a bit clunky. It's only really needed if the save is unsuccessful because Rails is smart enough to do cast on save.

```
object.attribute
=> true

object.update(attribute: "false")

object.attribute
=> false
```
It needs be cast again before the page is re-rendered though, as the radios treat any string as true. Without the minor radio will always be selected.

It could potentially live somewhere else 


```
if invalid?
  self.minor_change = ActiveModel::Type::Boolean.new.cast(minor_change)
  return false
elsif major_version?
else 
end 
```

etc.
Or it could be a public method called in the controller. It might be fine as is though. Whichever way is slightly annoying but i don't think it's a massive deal.

Keen to hear peoples thoughts.

app/views/admin/draft_edition_change_notes/edit_change_note.html.erb
```diff
@@ -0,0 +1,50 @@
+<% content_for :page_title, "#{@edition.title}: Change note" %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @change_note_form)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @change_note_form, url: update_change_note_admin_edition_path, method: :patch do |form| %>
+      <%= render "govuk_publishing_components/components/radio", {
```
davidgisbey, line 7: this is almost a direct rip from the `_change_notes` partial. I've not bothered refactoring it into a partial as the design might change etc. but it may well be worth doing if we proceed down this route

config/routes.rb
```diff
@@ -172,6 +172,9 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
         get :edit_slug, on: :member, controller: :edition_slug
         patch :update_slug, on: :member, controller: :edition_slug
 
+        get :edit_change_note, on: :member, controller: :draft_edition_change_notes
+        patch :update_change_note, on: :member, controller: :draft_edition_change_notes
```
davidgisbey, line 176: There's a pretty well established of using resources where possible. I've used custom routes here for 2 reasons:

1. There's already a change notes controller that handles editing published change notes. Only GDS Admins have access to this. But currently, it's using the `editions/:id/change_notes` endpoints.

If we proceed with this i'd suggest we rename that controller to something that conveys it's use case.

2. the other reason custom endpoints may make sense to use though, is when you create rails resources by default they assume an id exists for the object, which in the case of form objects it rarely does.

We could potentially use shallow nesting https://guides.rubyonrails.org/routing.html#shallow-nesting instead.

I'm don't feel particularly strong about deviating from using resources, but keen to hear other peoples opinions. There could also def be something fairly simple that i'm missing that would allow us to build the resources without it adding an additional id param.

config/locales/en/admin/change_note_form.yml
```diff
@@ -0,0 +1,11 @@
+en:
+  activemodel:
+      errors:
+        models:
+          change_note_form:
+            attributes:
+              minor_change:
+                inclusion: Select whether this is a major or minor update
+              change_note:
+                blank: "Enter a change note"
```
davidgisbey, line 10: This is a new pattern of adding error messages, rather than adding them to the models themselves. I know during discussions around improving errors down the line, we've discussed extracting them from the models and moving them to locales, so i figured this is probs a good place to start.

These definitely could just be added to the validations in the form object. This would def help to alleviate a bit of model bloat though,

app/forms/change_note_form.rb
ryanb-gds: Would including`ActiveRecord::AttributeAssignment` on the form model help with this? I used it [here](https://github.com/alphagov/whitehall/blob/44cefdf037a58d32f68f21577da5a725599fefbb/test/unit/app/models/date_validation_test.rb#L11C4-L11C46) to get Rails to do its magic casting in a test

config/locales/en/admin/change_note_form.yml
ryanb-gds: I'm in favour. Would also make our tests more maintainable as it would mean we could just reference the locale string instead of having to update our tests every time we tweak an error message

config/routes.rb
ryanb-gds: Hmm, I think I agree with your choice. Using shallow nesting would be going against its intended use of nesting index and create endpoints for a resource. Do [singular resources](https://guides.rubyonrails.org/routing.html#singular-resources) work in this case? I'm guessing having create and delete routes for change notes doesn't really make much sense?

test/forms/change_note_form_test.rb
```diff
@@ -0,0 +1,53 @@
...
+  test "#save returns false and adds errors to the form object when invalid" do
+    change_note_form = ChangeNoteForm.new
+    edition = build(:edition)
+
+    assert_equal false, change_note_form.save!(edition)
+    assert_equal ["Minor change Select whether this is a major or minor update"], change_note_form.errors.full_messages
+  end
+
+  test "#save updates the edition successfully when it is a major change and a change note is present" do
+    edition = create(:edition)
```
ryanb-gds, line 35: Potentially controversial, but what do you think of potentially stubbing the edition doing something like the [date validation test](https://github.com/alphagov/whitehall/blob/main/test/unit/app/models/date_validation_test.rb). I did a little bit of profiling of the Whitehall test suite using [StackProf](https://github.com/tmm1/stackprof) the other day and found out about 2/3 of the test suite run time is spent waiting on the database. So anything we can do to reduce the number of database calls in our tests could make a big impact.

ryanb-gds: Correction, I used https://github.com/test-prof/test-prof, I think with the StackProf integration. Might have been the rubyprof one though. Anyway, it's a pretty cool tool

config/routes.rb
davidgisbey: Yea good shout i think removing the `on: :member` makes sense here so that it expects :edition_id rather than just :id.

Yea i think the majority of the resources aren't required for this since it's just an edit page. 

I've ended up going with

```
 get :edit_change_note, to: 'draft_edition_change_notes#edit'
 patch :update_change_note, to: 'draft_edition_change_notes#update'
```

So the action names can be generic. it required quite a lot of noisy changes (file name changes etc.) so i just squashed it into the second commit. Let me know what you think

config/locales/en/admin/change_note_form.yml
davidgisbey: Great shout. I should def be using that in the test. Will sort that now

test/forms/change_note_form_test.rb
davidgisbey: I think build_stubbed won't work when it needs to interact with the database though right? Which it will for these tests

app/forms/change_note_form.rb
davidgisbey: I could def be wrong but i suspect it already has access to `assign_attributes` as it inherits from active model.

I just tested using `assign_attributes` locally and it doesn't seem to cast booleans. I might well be being dense though! Waiting for the coffee to kick in 😂 maybe we can discuss after standup?

test/forms/change_note_form_test.rb
ryanb-gds: Would it not be sufficient to check that `update!` was called with the right parameters? I thought `build_stubbed` did that but apparently not - but we could use our own stub to do it.

On the profiling side, I didn't check the specific directory but we easily could if you want to do a bit more poking around. Might be worth seeing if we can pair on that at some point

app/forms/change_note_form.rb
jkempster34: Apologies for being nosey again, I was intrigued by the PR title

You could maybe use the `attribute` method like we've done here https://github.com/alphagov/transition/blob/ac1217bd251d42b8d3618aa93f825fb9fc8fe9e8/app/forms/site_form.rb#L15

ollietreend: I think you probably want to be using [ActiveModel::Attributes](https://rubydoc.info/github/rails/rails/main/ActiveModel/Attributes) for this rather than [AttributeAssignment](https://rubydoc.info/github/rails/rails/main/ActiveModel/AttributeAssignment).

> Active Model Attributes are aware of data types, can have default values, and can handle casting and serialization.

So you could have something like:

```ruby
class ChangeNoteForm
  include ActiveModel::Model
  include ActiveModel::Attributes

  attribute :minor_change, :boolean, default: false
end
```

and that should handle casting the value automatically.

app/controllers/admin/draft_edition_change_notes_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    @edition = Edition.find(params[:edition_id])
+  end
+
+  def enforce_permissions!
+    enforce_permission!(:update, @edition)
+  end
+
+  def redirect_to_show_page_unless_editable
+    redirect_to admin_edition_path(@edition) unless @edition.editable?
+  end
```
ollietreend, line 31: Presumably we could do away with the other 'GDS Admin only' change note controller by tweaking this, right?

(Of course that's not in scope for this spike, but just thinking about how we could avoid maintaining two very similar controllers in future.)

app/forms/change_note_form.rb
```diff
@@ -0,0 +1,39 @@
+class ChangeNoteForm
```
ollietreend, line 1: If we were doing this for real, I assume we'd want to namespace these form objects a bit?

e.g. `app/forms/edition/change_note_form.rb` as `Edition::ChangeNoteForm`?

test/forms/change_note_form_test.rb
davidgisbey: Yea agreed. I must admit i'm always a bit way of heavy mocking/creating stubbed classes. But stubbing the update call seems like it's an excellent pattern to follow to me for unit tests

app/controllers/admin/draft_edition_change_notes_controller.rb
davidgisbey: Yea i think so. It would require a decent refactor though as the existing controller is built by taking 2 id's, the current edition's and a historical one (for the edition where the change note will be updated). 

It's not immediately obvious to me why the current editions id needs to be passed. I suspect i can probs be removed.

But yes, we could enforce different permissions for editions in different states & only allow GDS admins to see the index endpoint etc.

app/forms/change_note_form.rb
davidgisbey: Yea agreed. I'm not sure if we need `Admin::` thrown in there as well? Probs overkill since the rendering is going

app/forms/change_note_form.rb
davidgisbey: Thanks! Have sorted it

test/forms/change_note_form_test.rb
davidgisbey: Have updated the tests to use build_stubbed 👍

app/forms/change_note_form.rb
davidgisbey: Btw i also think we should do this for other things.

We have about 6 controllers with `Edition` prepended to them. Feels like they should be name spaced too.

ryanb-gds:
Closing this as I think it's unlikely to merge soon. Might well be useful for future reference though!

PR #8274, opened 2023-09-19
Use GOVUK publishing date input for statistical announcements

ryanb-gds:
User research indicates that teh GOVUK Publishing date input provides a superior user experience to the Rails date selection component, so we are replacing it across Whitehall.

I also removed the extracted release dates partial, as the abstraction did not survive switching components very well at all, there are too many variations on the fields name/id/parameter access required. I chose not to fix the issue where the precision input loses state if the form validation fails, as that is presently an issue in production and the fix wasn't obvious.

### Screenshots

Before

Create New Announcement
Screenshot 2023-09-20 at 08 44 15

Change Announcement Date
Screenshot 2023-09-20 at 08 44 58

After (with errors)

Create New Announcement
Screenshot 2023-09-20 at 08 48 12

Change Announcement Date
Screenshot 2023-09-20 at 08 55 21


davidgisbey (approved):
lgtm!

PR #8277, opened 2023-09-20
Give unpublished documents a new removed state

baisa:
This is to make the Whitehall UI more intuitive and the state of draft more meaningful to users. It should also reduce the number of errors that occur with syncing content with the Publishing API.

Ryan's notes (this is a joint effort):

Now that we have a removed state, the re-publishing flow is for the user to simply create a new edition of the document, superseding the removed edition. This does mean that "unwithdraw" now works differently from "republishing" - I've added a [ticket to see if we want to consider doing "unwithdraw" the same way](https://trello.com/c/qiLxHj4h).

I had to do a bit of work to make sure the removed edition could be superseded effectively. While I was there, I saw that the unwithdraw was doing something slightly illogical, which was transitioning the withdrawn edition straight to superseded, and then the immediate force publish was doing a no-op superseding. I thought it would make much more sense and be more consistent with the workflow model if unwithdraw set the state back to published and then the old edition is immediately superseded when the new edition is force published.

I've also deleted some code that allowed you to unpublish draft editions or withdraw already withdrawn editions, which made no sense to me at all. According to  f82ff054caa2bc8d46d05c5efefd37df5d6a1722 that was there to facilitate updating the withdrawal explanation on withdrawn/unpublished documents, but that doesn't seem to be required any longer.

I'm also interested in how this impacts the republishing task, assets and attachments. There seems to be quite a bit of logic such as [this](https://github.com/alphagov/whitehall/blob/edb6f0c95fa124cdcfe3e598ab1b7d5d54a775c3/app/services/edition_unpublisher.rb#L37) and [this](https://github.com/alphagov/whitehall/blob/edb6f0c95fa124cdcfe3e598ab1b7d5d54a775c3/app/workers/publishing_api_document_republishing_worker.rb#L40) that I suspect can now be removed, but I'm not 100% sure and I don't think it's causing any problems having it hanging around right now. Would appreciate feedback on that.

[Trello card](https://trello.com/c/4ttikeTH/1529-add-a-removed-state-or-withdrawn-state)


app/services/edition_unwithdrawer.rb
```diff
@@ -23,12 +23,12 @@ def failure_reasons
 private
 
   def prepare_edition
-    # The withdrawn edition needs to be in the published state so as a draft can be created
-    edition.update!(state: :published)
     edition.reload
```
ryanb-gds, line 26: Might be safe to get rid of this but I'm not 100% sure

test/unit/app/services/service_listeners/edition_dependencies_test.rb
```diff
@@ -52,49 +52,6 @@ class ServiceListeners::EditionDependenciesTest < ActiveSupport::TestCase
...
-          dependable_speech.title = "New speech title"
-          dependable_speech.submit!
-
-          expect_publishing(dependable_speech)
-          expect_republishing(dependent_article)
-          assert Whitehall.edition_services.send(service_name, dependable_speech).perform!
-        end
-      end
-    end
-  end
```
ryanb-gds, line 85: I'm going to investigate whether this is genuinely safe to remove. We think speeches have to depend on a news article. I think it should be safe because there shouldn't be any harm in having a removed edition preserve its dependencies, but needs confirming

ryanb-gds:
As agreed with @ollietreend and @beccapearce we're going to change to calling the new state "unpublished" instead of "removed". Adding that change to this PR would be difficult to rebase into a nice sequence of commits, so I'm going to close this PR and open a new one

PR #8278, opened 2023-09-20
Migrate topical event and role appointment date input components

ryanb-gds:
### What

Speeding things up a bit now as it seems we've worked out most of the tricky parts. This updates the topical events and role appointment date inputs to use the GOV.UK Publishing date input.

### Why

There's a lot of user research behind https://components.publishing.service.gov.uk/component-guide/date_input. That wasn't the case with the existing Whitehall date selection components.

### Screenshots (with errors)

Topical Events:
Screenshot 2023-09-20 at 14 26 05

Role Appointments:
Screenshot 2023-09-20 at 14 25 41

### Links

### DIFF
--- a/app/models/topical_event.rb
+++ b/app/models/topical_event.rb
@@ -1,7 +1,10 @@
 class TopicalEvent < ApplicationRecord
+  include DateValidation
   include PublishesToPublishingApi
   include Searchable
   include SimpleWorkflow
+
+  date_attributes(:start_date, :end_date)
 
   searchable title: :name,
              link: :search_link,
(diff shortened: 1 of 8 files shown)


app/models/topical_event.rb
```diff
@@ -1,8 +1,11 @@
 class TopicalEvent < ApplicationRecord
+  include DateValidation
   include PublishesToPublishingApi
   include Searchable
   include SimpleWorkflow
 
+  date_attributes(:start_date, :end_date)
```
davidgisbey, line 7: Same question as the other PR. Is end date nullable? If not should we validate start date is before end date?

davidgisbey (approved):
looks good. I just have that general question about whether we should be validating start date is before end date if both are present

app/models/topical_event.rb
ryanb-gds: Covered by another validate call on line 87, `validate :start_and_end_dates`. Again there's quite a bit of logic in there so I decided to leave it be

PR #8279, opened 2023-09-20
Migrate remaining date inputs to govuk publishing component

ryanb-gds:
### What

Replace the last remaining select-based date inputs on Whitehall with the GOV.UK Publishing date input.

### Why

The GOV.UK Publishing date input provides a better user experience, according to user research

### Screenshots (with errors)

Organisation Close Date
Screenshot 2023-09-20 at 15 05 27

Government Start Date
Screenshot 2023-09-20 at 15 51 04

Government End Date
Screenshot 2023-09-20 at 15 53 38

Offsite Links Date
Screenshot 2023-09-20 at 15 53 38

### Links

https://trello.com/c/IFGDegsk

### DIFF
--- a/app/models/government.rb
+++ b/app/models/government.rb
@@ -1,5 +1,8 @@
 class Government < ApplicationRecord
+  include DateValidation
   include PublishesToPublishingApi
+
+  date_attributes(:start_date, :end_date)
 
   validates :name, presence: true, uniqueness: { case_sensitive: false }
   validates :slug, presence: true, uniqueness: { case_sensitive: false }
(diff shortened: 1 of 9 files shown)


app/models/government.rb
```diff
@@ -1,6 +1,9 @@
 class Government < ApplicationRecord
+  include DateValidation
   include PublishesToPublishingApi
 
+  date_attributes(:start_date, :end_date)
```
davidgisbey, line 5: should we validate start date is before end date? Or i guess maybe not as end date can be nil?

ryanb-gds: It's already covered on line 12, there's `validate :not_overlapping?, if: -> { start_date.present? }` which checks quite a few things so I decided not to mess about with it.

PR #8280, opened 2023-09-20
Created rake task for AttachmentData migration to Assets

syed-ali-tw:
In order to be able to populate Asset with Attachment details we want to request asset_manager_id from asset-manager and save them along with other detail like attachment_data_id. This rake task intends to do this.

[trello](https://trello.com/c/jB2kP508/104-task-migrate-database-for-attachmentdata-to-use-new-asset-id-instead)

### DIFF
--- a/app/workers/create_asset_relationship_worker.rb
+++ b/app/workers/create_asset_relationship_worker.rb
@@ -1,0 +1,76 @@
+class CreateAssetRelationshipWorker < WorkerBase
+  def perform(start_id, end_id)
+    logger.info("CreateAssetRelationshipWorker start!")
+    assetable_type = AttachmentData
+    assetables = assetable_type.where(id: start_id..end_id)
+    logger.info "Number of #{assetable_type} found: #{assetables.count}"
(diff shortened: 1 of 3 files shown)


app/workers/create_asset_relationship_worker.rb
```diff
@@ -0,0 +1,69 @@
...
+    assetables = assetable_type.where(id: counter_start..counter_end)
+    logger.info "Number of #{assetable_type.to_s} found: #{assetables.count.to_s}"
+    logger.info "Creating Asset for #{counter_end} records"
+
+    count = 0
+    assetables.each do |assetable|
+      break if counter_start > counter_end
+      assetable.use_non_legacy_endpoints = true;
+      assetable.save
+      begin
```
tunylund, line 14: saving the original and the thumbnail in the same begin-rescue is problematic. If the orginal raises an error, the code will never attempt to save the thumbnail

tunylund (changes requested):

tunylund (approved):

PR #8283, opened 2023-09-21
Remove current user test

ryanb-gds:
This test was added just to confirm that the Current model in Rails could be used in a thread safe way. We have been using Current without seeing any regressions so this test can probably be safely removed, as it's really just testing Rails rather than any of our code.


ryanb-gds:
Closing this as @ollietreend pointed out this test is quite useful in confirming that we've achieved thread safety effectively, in case we were to use a mechanism other than the `CurrentAttributes` module at any point in future (as other parts of Whitehall do).

PR #8210, opened 2023-09-05
Format `.js` and `.scss` files using Prettier

dnkrj:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

# What
- Introduce and configure prettier
- Update files to reflect new formatting rules

# Why
- Align with govuk-frontend

https://trello.com/c/p8IqI4zH/1653-tech-spike-more-opinionated-es6-rules

PR #8284, opened 2023-09-21
Add ERB lint as a dependency

davidgisbey:


lib/tasks/lint.rake
```diff
@@ -1,5 +1,10 @@
 desc "Run all linters"
 task lint: :environment do
   sh "bundle exec rubocop"
+  if Rails.env.development?
+    sh "bundle exec erblint --lint-all --autocorrect"
```
davidgisbey, line 5: I've copied the behaviour of Collections Publisher as i don't have particularly strong feelings on whether this should autocorrect locally. Happy to go with the conensus.

dnkrj (approved):
🎉

ollietreend (approved):
I know we've talked about implementing ERB linting for a while now, and the number of changes that it'd inevitably require always felt prohibitive.

I think we should bite the bullet and merge this in 🚀 there's no time like the present, as otherwise it'll only get out of sync and need rebasing.

This seems an especially good time to introduce ERB linting since we've recently tightened up our SCSS and JS linting too (i.e. #8178 and #8210).

This looks great @davidgisbey! 🎉

davidgisbey:
Cool thanks both. Have merged 👍

PR #8292, opened 2023-09-22
Enable AttachmentData migration for CSV files.

tunylund:
The migration was previously blocked by csv preview not being ready yet.


tunylund:
closed as this is done in another pr

PR #7803, opened 2023-06-13
Pass locale into sidebar actions link

callumknights:
Fixes an issue where non-english editions weren't linking to the public site correctly from the admin interface.

Trello: https://trello.com/c/dmI3Fr77/691-5359625-5357779-translated-pages-broken-links

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #8299, opened 2023-09-25
Use correct locale for non-English edition links

ollietreend:
The method `Edition#public_url` returns the publicly visible URL for a document on the GOV.UK website.

However it was not returning the correct URL for editions with a non-English primary locale – for example, for foreign-language only news articles or consultations.

This bug was partially fixed in #7803 but only for the "View on website" link shown in the sidebar for published editions.

We've since received a [Zendesk ticket][1] telling us that the "Preview on website" link, as well as the links in publisher notification emails, were still missing the locale from the end of the URL. This meant users got "Page not found" errors when following those links.

This commit fixes the root cause of the problem, meaning I can safely remove the fix applied in #7803 as it's no longer required.

---

For reference, the "Preview on website" link does not explicitly set a locale when getting the public URL:

https://github.com/alphagov/whitehall/blob/6e86aaea8b86cedc6e6689deebcec87a12269bf7/app/components/admin/editions/show/preview_component.html.erb#L11

Nor do the email notifications:

https://github.com/alphagov/whitehall/blob/cf42f9e3202fbb19b90c3bfe9e893ab5e28904cd/app/services/author_notifier_service.rb#L40-L42

Therefore the changes made to `Edition#public_url` should fix these broken links (and any others throughout the app).

[1]: https://govuk.zendesk.com/agent/tickets/5491481

---

Trello: https://trello.com/c/xzQvVgtw/1767-fix-broken-links-for-non-english-editions

### DIFF
--- a/app/components/admin/editions/show/sidebar_actions_component.rb
+++ b/app/components/admin/editions/show/sidebar_actions_component.rb
@@ -269,7 +269,7 @@ private
   def add_view_action
     if @edition.publicly_visible?
       actions << link_to("View on website (opens in new tab)",
-                         @edition.public_url(locale: @edition.primary_locale),
+                         @edition.public_url,
                          class: "govuk-link",
                          target: "_blank",
                          rel: "noopener",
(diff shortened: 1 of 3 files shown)


ryanb-gds (approved):
Looks like a nasty bug well fixed 👍

PR #8301, opened 2023-09-26
Keep list of user permissions in one place

ollietreend:
Whitehall keeps a list of user permissions that it accepts from Signon users in the module User::Permissions as constants.

The only exception to that was the "Sidekiq Admin" permission, which was defined in SidekiqGdsSsoMiddleware.

To simplify things I've decided to centralise the list of 'known' user permissions. This will make it easier to keep track of the permissions, and gives us an easy way to answer the question: "Which permissions does Whitehall use?".

Now that the list is kept in one place, the answer is simple. If it's not listed in User::Permissions, then Whitehall doesn't know or care about it.


beccapearce (commented):
Looks good to me!

PR #8304, opened 2023-09-27
Put unpublished editions into an unpublished state

ryanb-gds:
As opposed to the current behaviour of reverting them to the draft state

This is intended to reduce confusion for users, who have reported that the behaviour does not match their expectations. The current behaviour also means that we have thousands of draft editions that we do not expect will ever be published again. This is inconsistent with other publishing apps.

In addition, the reversion to draft state makes it difficult to synchronise the content with the publishing API. Simply putting the edition into a dedicated "unpublished" state and requiring users to supersede the edition as usual makes the synchronisation logic much simpler, as we no longer need to check if the draft edition was once in the published state or send a new draft to the publishing API when we unpublish an edition.

This PR includes a data migration that updates the state of draft editions with an unpublishing record to the new "unpublished" state, and instructs Publishing API to discard the associated draft content item.

Trello: https://trello.com/c/4ttikeTH


test/unit/app/models/edition/publishing_test.rb
```diff
@@ -125,34 +131,14 @@ class Edition::PublishingTest < ActiveSupport::TestCase
...
     edition = build(:published_edition)
 
     assert_nil edition.unpublished_edition
   end
 
-  test "#unpublished_edition returns unpublished edition if unpublished" do
-    edition = create(:draft_edition, :with_document)
-    edition.build_unpublishing(edition:)
+  test "#unpublished_edition returns itself if unpublished" do
+    edition = create(:unpublished_edition)
```
ryanb-gds, line 141: This is weird and this whole method needs to be refactored away, but I will add a separate PR for that

test/unit/lib/data_hygiene/publishing_api_document_republisher_test.rb
```diff
@@ -30,17 +30,4 @@ class DataHygiene::PublishingApiDocumentRepublisherTest < ActiveSupport::TestCas
...
-    edition = create(:unpublished_edition, title: "Unpublished edition", document:)
-    unpublishing = edition.unpublishing
-
-    PublishingApiUnpublishingWorker.expects(:new).returns(worker = mock)
-    worker.expects(:perform).with(unpublishing.id, true)
-
-    Sidekiq::Testing.inline! do
-      DataHygiene::PublishingApiDocumentRepublisher.new(edition.class, NullLogger.instance).perform
-    end
-  end
```
ryanb-gds, line 45: This is essentially a duplicate of the test in `test/unit/app/workers/publishing_api_document_republishing_worker_test.rb`

ollietreend (commented):
@ryanb-gds I'm sorry that this is an incomplete review 😓 but rather than keeping you hanging, here's what stood out to me so far. I expect it's probably all the 'big' stuff anyway.

app/models/edition/publishing.rb
```diff
@@ -42,7 +42,7 @@ def change_note_required?
     elsif draft? && new_record?
       false
     else
-      other_editions.published.exists?
+      other_editions.where(state: %i[published unpublished]).exists?
```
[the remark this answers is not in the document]
ryanb-gds: Good shout, I'll give it a try

app/workers/publishing_api_document_republishing_worker.rb
```diff
@@ -38,9 +38,7 @@ def perform(document_id, bulk_publishing = false)
       document.lock!
 
       if the_document_has_been_unpublished?
-        send_draft_and_unpublish
-      elsif the_document_has_been_withdrawn?
-        send_published_and_withdraw
+        send_published_and_unpublish
```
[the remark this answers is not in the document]
ryanb-gds: Yeah I wasn't 100% sure on this. I think it does need to be updated, but I might have made the wrong change, thinking about it. I was working based on the fact that there was previously code to handle unpublished documents,  which rerendered and sent the draft that was created during unpublishing and resent the "unpublishing" data to Publishing API. There isn't a draft to send any longer, so this doesn't make sense and I think should be changed.

I based the new behaviour on what happens with withdrawn documents. For withdrawn documents, it resends the document, which I believe republishes it to Publishing API, and then resends the unpublishing to ensure it's immediately unpublished again. However, because unpublished documents are not rendered, unlike withdrawn documents, that isn't really necessary. I think for an unpublished document the worker should really perform a no-op. Would you agree?

db/data_migration/20230927101500_migrate_unpublished_draft_editions_to_unpublished_state.rb
```diff
@@ -0,0 +1,15 @@
...
+# We are therefore migrating all draft editions with an associated `Unpublishing` record to the unpublished state, and
+# asking that the Publishing API discard its draft of the content item.
+
+unpublished_editions = Edition.joins(:unpublishing).draft
+
+unpublished_editions.find_each(batch_size: 100) do |edition|
+  edition.state = "unpublished"
+  edition.save!
+  PublishingApiDiscardDraftWorker.perform_async(edition.content_id, edition.primary_locale)
+end
```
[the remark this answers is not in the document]
ryanb-gds: Makes sense to me. I wrote very much the simplest possible migration so it's good to get some guidance on what's a useful level of optimization for future migrations

test/factories/editions.rb
```diff
@@ -157,24 +157,28 @@
     end
 
     trait(:unpublished) do
+      state { :unpublished }
       after(:create) do |edition|
         edition.unpublishing = build(:unpublishing, edition:)
       end
```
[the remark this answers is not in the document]
ryanb-gds: Nope, just thoughtless copy pasta

app/services/edition_unpublisher.rb
```diff
@@ -29,11 +27,6 @@ def fire_transition!
     mark_attachments_as_present_at_unpublish
   end
 
-  def prepare_edition
-    edition.force_published = false
-    edition.reset_version_numbers
-  end
```
ollietreend, line 35: My [brief search of the code](https://github.com/search?q=repo%3Aalphagov%2Fwhitehall+%28published_major_version+OR+published_minor_version%29&type=code) suggests the version numbers on editions are used to decide whether the "First published at" field is shown to users. And also maybe when building the "change notes" history that get sent to the Publishing API. But I don't think the values themselves are sent downstream to the Publishing API.

In any case, I'm wondering what the desired behaviour should be here? 🤔

It's worth noting that the `reset_version_numbers` method doesn't actually set them back to `0`. Instead it [rolls them back](https://github.com/alphagov/whitehall/blob/fe3e2d61dec9f4c881d9d16c15258496bb2d3284/app/models/edition/publishing.rb#L87-L90) to whatever versions the previous edition had.

ollietreend: This whole thing is a bit mind boggling, but my instinct is to say we leave it in rather than remove it.

But 🤯 🤷🏻

app/workers/publishing_api_document_republishing_worker.rb
```diff
@@ -78,11 +76,7 @@ def the_document_has_an_edition_to_check?
   end
 
   def the_document_has_been_unpublished?
-    pre_publication_edition && pre_publication_edition.unpublishing
-  end
-
-  def the_document_has_been_withdrawn?
-    live_edition && live_edition.unpublishing
+    live_edition && (live_edition.unpublished? || live_edition.withdrawn?)
```
ollietreend, line 79: We talked about:

I don't think it's safe to consolidate the `the_document_has_been_unpublished?` and the `the_document_has_been_withdrawn?` methods because `live_edition` has to be publicly visible – which means it can only ever be `published` or `withdrawn`.

Unpublished editions are not publicly visible, so are not 'live' per se.

test/factories/editions.rb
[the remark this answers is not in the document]
ryanb-gds: As mentioned, this proposed change broke some tests. I don't exactly know why, but essentially when the after create hook is used the test/unit/app/services/asset_manager/attachment_updater_test.rb:216 fails because the edition doesn't seem to have a thumbnail any longer. It also affects other tests around attachments and assets.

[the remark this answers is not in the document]
ryanb-gds: This change exposed some incorrect tests which I have resolved.

app/services/edition_unpublisher.rb
ryanb-gds: I'm not sure I agree with you on this one, though it's not a strong disagreement. The unpublished version remains a genuine version that did exist and was publicly available, to my mind. If we reset the version number, we'll end up with the previous edition and the unpublished edition having the same version number, which doesn't seem right and might lead to confusing change notes?

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -178,7 +178,7 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
 
     describe "redirect url updates" do
       let(:unpublished_edition) { FactoryBot.create(:unpublished_edition) }
-      let(:redirect_url) { unpublished_edition.public_url }
+      let(:redirect_url) { unpublished_edition.unpublishing.document_url }
```
ryanb-gds, line 181: This fix was required when we tweaked the edition factories not to use an after create hook to add the unpublishing. The test was actually incorrect looking at the underlying behaviour, but because the factory was forcibly setting up the unpublishing the values of the edition's public url and the unpublishing document_url were coincidentally the same.

app/services/edition_unpublisher.rb
ryanb-gds: Reverted in latest push

app/workers/publishing_api_document_republishing_worker.rb
ryanb-gds: I've reverted back to the old implementation until we figure out a way to make this work. It's not currently working for unpublished editions

app/workers/publishing_api_document_republishing_worker.rb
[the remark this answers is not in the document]
ryanb-gds: Reverted these changes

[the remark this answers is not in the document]
ryanb-gds: We ended up sending the unpublishing request again so that the unpublishing explanation can be updated if necessary, and the sending the draft if there is one.

app/workers/publishing_api_document_republishing_worker.rb
ryanb-gds: We had to check for an unpublished edition separately in the end. The behaviour for withdrawn and unpublished documents remains separate

PR #8313, opened 2023-09-29
Only render attachments in previews when we allow inline attachments

beccapearce:
Some of our documents that do not allow inline attachments have been wrongly showing inline attachments in the body preview on the edit edition page. This has led to publishers being confused at where their embedded attachment goes when they publish the document.

This was caused because we were passing the attachment ids to the component that renders the preview based on whether they `allow_attachments?` However publications are a document_type that `allow_attachments?` but doesn't `allows_inline_attachments?`. And so incorrectly shows inline attachments in the preview.

This fixes this issue by only passing the attachment ids through to the component that renders the preview if the edition `allows_inline_attachments?`

trello card: https://trello.com/c/24R4gThb

Before: 
Screenshot 2023-09-29 at 17 59 31

After:
Screenshot 2023-09-29 at 18 00 27

### DIFF
--- a/app/views/admin/editions/_standard_fields.html.erb
+++ b/app/views/admin/editions/_standard_fields.html.erb
@@ -63,7 +63,7 @@
       right_to_left: form.object.translation_rtl?,
       data_attributes: {
         image_ids: edition.images.map { |img| img[:id] }.to_json,
-        attachment_ids: edition.allows_attachments? ? edition.attachments.map(&:id) : [],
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
Seems like it will solve the issue. As discussed, would be nice to have test coverage, but in this instance the value to effort trade off is not in our favour. Hopefully we will get an opportunity to improve view test coverage during the work to move to the document hub and spoke interface model

PR #8316, opened 2023-10-02
Push content with non legacy endpoints to publishing api - image data

lauraghiorghisor-tw:
These changes ensure that we only send healthy media urls for images to publishing api and that whitehall similarly only renders such urls. 

Most of the functionality for migrating legacy urls to non-legacy/media urls was already in place due to the changes in the carrierwave config file that we implemented for file attachments. In addition, these changes implement:
- Images whose asset variants have not all finished uploading (all 7 variants) are not rendered in whitehall preview/govspeak
- Images whose asset variants have not all finished uploading (all 7 variants) cannot be viewed in draft/live stack
- Same applies to lead images
- If assets are still uploading, we can see a "Processing" label for each image
- Back fix for files which are now displaying the same "Processing" label rather than "Uploading"

[Trello card here](https://trello.com/c/pTVgisgn/169-story-whitehall-should-push-edition-content-with-non-legacy-asset-urls-to-publish-api-imagedata)

### DIFF
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -48,7 +48,9 @@ module GovspeakHelper
   end
 
   def prepare_images(images)
-    images.map do |image|
+    images
+      .select { |image| image.image_data&.all_asset_variants_uploaded? }
+      .map do |image|
       {
         id: image.image_data.carrierwave_image,
         image_data_id: image.image_data_id,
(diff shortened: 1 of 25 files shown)


app/models/edition/images.rb
```diff
@@ -19,6 +19,10 @@ def process_associations_after_save(edition)
     accepts_nested_attributes_for :images, reject_if: :no_substantive_attributes?, allow_destroy: true
 
     add_trait Trait
+
+    def images_ready_for_publishing
+      images.select { |image| image.image_data.all_asset_variants_uploaded? }
```
tunylund, line 24: maybe add a coalescing operator to be on the safe side

app/helpers/govspeak_helper.rb
```diff
@@ -25,7 +25,7 @@ def govspeak_with_attachments_to_html(body, attachments = [], alternative_format
   end
 
   def bare_govspeak_edition_to_html(edition)
-    images = prepare_images(edition.try(:images) || [])
+    images = prepare_images(edition.try(:images_ready_for_publishing) || [])
```
tunylund, line 28: cool but the govspeak admin helper needs this too

tunylund (approved):

PR #8323, opened 2023-10-05
Add a confirm publish page to edition workflow

beccapearce:
In order to prevent cases of accidental publishing we have added an interstitial page to the publish journey.

Trello card: https://trello.com/c/US0cgYnR

### DIFF
--- a/app/controllers/admin/edition_workflow_controller.rb
+++ b/app/controllers/admin/edition_workflow_controller.rb
@@ -33,7 +33,7 @@ class Admin::EditionWorkflowController < Admin::BaseController
       enforce_permission!(:update, @edition)
     when "reject"
       enforce_permission!(:reject, @edition)
-    when "publish", "schedule"
+    when "publish", "schedule", "confirm_publish"
       enforce_permission!(:publish, @edition)
     when "force_publish", "confirm_force_publish", "force_schedule", "confirm_force_schedule"
       enforce_permission!(:force_publish, @edition)
@@ -63,6 +63,12 @@ class Admin::EditionWorkflowController < Admin::BaseController
     end
     redirect_to new_admin_edition_editorial_remark_path(@edition),
                 notice: "Document rejected; please explain why in an internal note"
+  end
+
+  def confirm_publish
+    unless @edition.valid?(:publish)
+      redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ")
+    end
   end
 
   def publish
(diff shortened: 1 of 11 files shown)


app/controllers/admin/edition_workflow_controller.rb
```diff
@@ -167,7 +171,7 @@ def approve_retrospectively
 private
 
   def get_layout
-    design_system_actions = %w[confirm_approve_retrospectively confirm_force_schedule confirm_unpublish confirm_unschedule confirm_unwithdraw unpublish confirm_force_publish]
+    design_system_actions = %w[confirm_approve_retrospectively confirm_force_schedule confirm_publish confirm_unpublish confirm_unschedule confirm_unwithdraw unpublish confirm_force_publish]
```
ryanb-gds, line 174: At this point do we still need this? They're all using the design system now, right?

test/support/css_selectors.rb
```diff
@@ -74,7 +74,7 @@ def world_locations_selector
   end
 
   def publish_form_selector(document)
```
ryanb-gds, line 76: I feel it might be sensible to update the name of this method, given its purpose has changed?

app/controllers/admin/edition_workflow_controller.rb
```diff
@@ -65,6 +65,10 @@ def reject
                 notice: "Document rejected; please explain why in an internal note"
   end
 
+  def confirm_publish
+    redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ") and return unless @edition.valid?(:publish)
```
ryanb-gds, line 69: Would this be a bit more readable if it wasn't a one-liner? I _think_ I know what it does... but I'd like to be more confident

ryanb-gds (approved):
Looking good, a few minor suggestions you may or may not wish to adopt 🙌

app/controllers/admin/edition_workflow_controller.rb
```diff
@@ -293,6 +297,8 @@ def action_name_as_human_interaction(action_name)
       "unpublish this edition"
     when "confirm_force_publish"
       "force publish this edition"
+    when "confirm_publish"
+      "Publish this edition"
```
ollietreend, line 301: Should this be lowercase, as per the other action names?

```suggestion
      "publish this edition"
```

app/controllers/admin/edition_workflow_controller.rb
beccapearce: Ooh I think so, I'll have a go at removing it in a separate PR just to keep it separate and hopefully clearer

app/controllers/admin/edition_workflow_controller.rb
beccapearce: Yes I agree, which do you think is more clear?
```
    if @edition.valid?(:publish)
      nil
    else
       redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ")
     end
```
or 
```
    unless @edition.valid?(:publish)
      redirect_to admin_edition_path(@edition), alert: @edition.errors[:base].join(". ")
    end
```

or something else?

ryanb-gds: I like the unless

PR #8345, opened 2023-10-10
Fix several errors relating to date validation

ryanb-gds:
Includes fixes for the following issues:
- When the year was nil, a NoMethodError was being thrown because we were attempting to call "zero?" on nil. We're now rescuing from the NoMethodError and marking the date invalid
- Entering a negative month value would trigger a MultiparameterAssigmentError, because Ruby Date constructor allows months to be set using a negative offset but Rails date assignment does not

### DIFF
--- a/app/models/date_validation.rb
+++ b/app/models/date_validation.rb
@@ -28,11 +28,14 @@ module DateValidation
       begin
         # Rails will cast the year part of the date to 0 if the year input parameter is a non-numeric string
         # This only seems to happen to the year part, other parts remain as strings
-        raise TypeError if (date[1]).zero?
+        raise TypeError if date[1].zero?
+
+        # Rails does not accept negative month values, but the Date constructor does
+        raise TypeError if (date[2]).negative?
 
         Date.new(date[1], date[2], date[3])
         @invalid_date_attributes.delete(attribute)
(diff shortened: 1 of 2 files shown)


app/models/date_validation.rb
```diff
@@ -28,11 +28,14 @@ def pre_validate_date_attribute(attribute, date)
       begin
         # Rails will cast the year part of the date to 0 if the year input parameter is a non-numeric string
         # This only seems to happen to the year part, other parts remain as strings
-        raise TypeError if (date[1]).zero?
+        raise TypeError if date[1].zero?
+
+        # Rails does not accept negative month values, but the Date constructor does
+        raise TypeError if (date[2]).negative?
```
davidgisbey, line 34: does the date constructor handle negative values for date[1] and date[3] then?

ryanb-gds: Well, a negative year is legitimate (BC years), and yes it does prevent negative days: https://docs.ruby-lang.org/en/master/Date.html#method-c-new

davidgisbey: cool thanks. I suspect the person putting in negative values was purposefully doing to try and break it 😂 

LGTM!

davidgisbey (approved):

PR #8318, opened 2023-10-03
531 reordering of groups

RodneyJohnsonGDS:
[Trello](https://trello.com/c/kEcGRQ1X/531-allow-for-the-reordering-of-documents-within-a-group)

The changes in this PR
- Add the "Reorder document" link to the Group landing page
- Creates the new reorder documents page
- Introduces the reordering of documents functionality
- Display success flash message 

Screenshots below:

|Page|Screenshot|
|-|-|
|Landing page:|Screenshot 2023-09-12 at 15 09 53|
|Reorder page|Screenshot 2023-09-12 at 15 10 08|
|Landing page after reorder|Screenshot 2023-09-12 at 15 10 25|

PR #8347, opened 2023-10-10
changed the add documents and reorder links from p tags to list tags …

RodneyJohnsonGDS:
…as these are more semantically and stylistally correct

This commit is a continuation from a previously closed and merged branch, can be founder here:
https://github.com/alphagov/whitehall/pull/8318#discussion_r1352028169

This commit user story can be found in this Trello link: https://trello.com/c/kEcGRQ1X/531-allow-for-the-reordering-of-documents-within-a-group

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@
 /node_modules
 /yarn-error.log
 .cache/
+.DS_Store
--- a/app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
+++ b/app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
@@ -1,6 +1,25 @@
 .app-view-document-collection-group-memberships-index__heading {
-  display: flex;
-  justify-content: space-between;
+  .gem-c-heading {
+    display: inline-block;
+    @include govuk-responsive-margin(2, "right");
+  }
+
+  .govuk-list {
+    float: right;
+
+    li {
+      display: inline-block;
+      @include govuk-responsive-padding(2, "left");
+      @include govuk-responsive-margin(2, "left");
+      border-left: 1px solid $govuk-border-colour;
+
+      &:first-child {
+        margin-left: 0;
+        padding-left: 0;
+        border-left: none;
+      }
+    }
+  }
 }
 
 .app-view-document-collection-group-memberships-index__table .govuk-table__cell {
@@ -8,16 +27,5 @@
 
   &:last-child {
     white-space: nowrap;
-  }
-}
-
-.app-view-document-collection-group-memberships-index__heading-links {
-  display: flex;
-  align-items: center;
-
-  & > p:first-child {
-    @include govuk-responsive-padding(2, "right");
-    @include govuk-responsive-margin(2, "right");
-    border-right: 1px solid $govuk-border-colour;
   }
 }
--- a/app/views/admin/document_collection_group_memberships/index.html.erb
+++ b/app/views/admin/document_collection_group_memberships/index.html.erb
@@ -26,17 +26,15 @@
             text: "Documents",
             margin_bottom: 6,
           } %>
-
-          <div class="app-view-document-collection-group-memberships-index__heading-links">
-            <p class="govuk-body govuk-!-text-align-right">
+          <ul class="govuk-list">
+            <li>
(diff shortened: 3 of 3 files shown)


davidtrussler:
@RodneyJohnsonGDS I think that markup is more appropriate now and consistent but I think the use of `display: flex` here is a little problematic as it is aligning the links centrally to the header and distorting their horizontal alignment, albeit only marginally, so I would prefer to style it without that. 

Also, we should probably be a bit more generic in the way the margins and borders for the links are applied so that it could potentially work for single or multiple links. 

I'll suggest an alternative in the CSS file.

app/assets/stylesheets/admin/views/_document-collection-group-memberships-index.scss
```diff
@@ -15,7 +15,7 @@
   display: flex;
   align-items: center;
 
-  & > p:first-child {
+  & > li:first-child {
     @include govuk-responsive-padding(2, "right");
     @include govuk-responsive-margin(2, "right");
     border-right: 1px solid $govuk-border-colour;
```
davidtrussler, line 21: I think it would be better to lose the whole `app-view-document-collection-group-memberships-index__heading-links` class and add a more generic style nested in the `app-view-document-collection-group-memberships-index__heading` element like below. 

That way we can just allow the `<li>` elements to flow more naturally and add the separating border in a way that doesn't rely on a specific number of links to be present. It also centres the vertical rule better and keeps the horizontal alignment with the header. 

I think I'd also lose the flexbox layout in the containing element as well because that makes it display oddly on smaller viewports. By floating the `<ul>` element we can let it flow underneath the heading in this case, though we do need to change the heading from a block level element to allow this, and give it a bit of a margin to avoid the elements clashing. 

Once we're nested inside the very specific `app-view-document-collection-group-memberships-index__heading` element I think we can afford to be far more generic in the way we reference elements so I don't think we need to add specific classes for those. 

See what you think. I know it's quite a big change but it makes it closer to the sort of styles we are using in related components. (Also I think flexbox can be overused. 😄)

```
.app-view-document-collection-group-memberships-index__heading {
  .gem-c-heading {
    display: inline-block;
    @include govuk-responsive-margin(2, "right");
  }

  .govuk-list {
    float: right; 

    li {
      display: inline-block;
      @include govuk-responsive-padding(2, "left");
      @include govuk-responsive-margin(2, "left");
      border-left: 1px solid $govuk-border-colour;

      &:first-child {
        margin-left: 0;
        padding-left: 0;
        border-left: none;
      }
    }
  }
}
```

app/views/admin/document_collection_group_memberships/index.html.erb
```diff
@@ -26,17 +26,15 @@
             text: "Documents",
             margin_bottom: 6,
           } %>
-
-          <div class="app-view-document-collection-group-memberships-index__heading-links">
-            <p class="govuk-body govuk-!-text-align-right">
+          <ul class="govuk-list app-view-document-collection-group-memberships-index__heading-links">
```
davidtrussler, line 29: If we use the approach above we can lose the `app-view-document-collection-group-memberships-index__heading-links` class here.

app/views/admin/document_collection_group_memberships/index.html.erb
```diff
@@ -47,9 +45,11 @@
             margin_bottom: 6,
           } %>
 
-          <p class="govuk-body govuk-!-text-align-right">
-            <%= link_to "Add document", admin_document_collection_group_search_options_path(@collection, @group), class: "govuk-link" %>
-          </p>
+          <ul class="govuk-list govuk-body govuk-!-text-align-right">
```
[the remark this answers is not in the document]
RodneyJohnsonGDS: Looks good, done!

.gitignore
```diff
@@ -29,3 +29,4 @@
 /node_modules
 /yarn-error.log
 .cache/
+.DS_Store/
```
davidtrussler, line 32: This would need to be more generic to catch all instances of this file as it occurs in many directories and subdirectories (e.g. the one in the `app` directory has survived). I'm not sure we should need to add this though if we make sure we don't commit those files.

davidtrussler (approved):
LGTM 👍

PR #8348, opened 2023-10-10
Improve CI run time

jyoung-gds:
# What
[Trello ticket link](https://trello.com/c/IS5QUoaK/1757-split-and-parallelise-minitest-job-in-ci)
This PR improves the time it takes our CI tests to run by:
- Parallelising the integration tests so that one process runs the `cucumber:ok` task and one process runs the `cucumber:preview_design_system` task
- Parallelising the MiniTest workflow using a matrix strategy, randomly distributing tests across 4 nodes. 

**Benchmark:** Approximately 17m39s (average of the 10 most recent merge to main CI runs)

With the changes proposed, this PR reduces CI run time to approximately 7 minutes.

# Why
This is to help reduce developer toil and time to deploy changes. The current run time is excessive and slows down development.

# Additional information
## Order of implementation
To avoid having a period where there are no Minitest or integration checks in our CI/CD pipeline, these changes must be implemented in the following order:
1. Merge the new parallelised tests to main
2. Create new branch rules so that parallelised tests must pass before merging
3. Remove non-parallelised branch rules
4. Remove non-parallelised workflows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,10 @@ jobs:
   test-ruby:
     name: Test Ruby
     uses: ./.github/workflows/minitest.yml
+  
+  test-ruby-parallel:
+    name: Test Ruby
+    uses: ./.github/workflows/minitest-parallel.yml
 
   integration-tests-parallel:
     name: Integration tests
--- a/.github/workflows/integration-tests-parallel.yml
+++ b/.github/workflows/integration-tests-parallel.yml
@@ -1,57 +1,15 @@
-name: CI
+name: Run integration tests
 
-on:
-  workflow_dispatch: {}
-  push:
-    branches:
-      - main
-    paths-ignore:
-      - "Jenkinsfile"
-      - ".git**"
-  pull_request:
+on: workflow_call
 
 jobs:
-  security-analysis:
-    name: Security Analysis
-    uses: alphagov/govuk-infrastructure/.github/workflows/brakeman.yml@main
-
-  lint-scss:
-    name: Lint SCSS
-    uses: ./.github/workflows/lintscss.yml
-
-  lint-javascript:
-    name: Lint JavaScript
-    uses: ./.github/workflows/lintjs.yml
-
-  lint-prettier:
-    name: Prettier
-    uses: ./.github/workflows/lintprettier.yml
-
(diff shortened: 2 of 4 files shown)


jyoung-gds:
I've got some food for thought on this:

With the current implementation, MiniTest tests are manually assigned to nodes. This potentially puts us at risk of not running the full suite: if we've missed some tests, or if future tests get added to locations which lie outside of the assignment. This could be solved by randomly assigning tests across nodes, potentially resulting in a small increase in run time. The random assignment of tests would make it easier to scale up - as a node can be added without worrying about re-allocating tests across nodes - but increases complexity.

.github/workflows/minitest.yml
```diff
@@ -17,6 +17,14 @@ jobs:
   run-minitest:
     name: Run Minitest
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ci_node_index: [
```
ollietreend, line 23: Might be nice to rename this variable

```suggestion
        test_paths: [
```

.github/workflows/integration-tests-design-system.yml
```diff
@@ -0,0 +1,52 @@
+name: Run integration tests on design system
+
+on: workflow_call
+
+jobs:
+  run-integration-tests-design-system:
+    name: Run integration tests on design system
+    runs-on: ubuntu-latest
+    steps:
```
ollietreend, line 9: Would it be possible to 'DRY this up' by using a matrix strategy, as you've already done with MiniTest?

That way we'd only need one `.github/workflows/integration-tests.yml` file, right?

jyoung-gds: Definitely. I went for this version because I thought it could be easier to spot when tidying up after design system is all ported over, but maybe it doesn't make much difference.

davidgisbey:
this is great 🚀

ryanb-gds (approved):
Looks reasonable to me. I wasn't expecting us to be quite so fine-grained with the unit test directories, but if this works well then I don't have any objections

davidgisbey:
Seems like a good approach to me. Makes me feel a lot more comfortable about us not missing tests 👍

davidgisbey (approved):
lgtm!

PR #8350, opened 2023-10-10
Remove references to slug on document filter

beccapearce:
The word slug is not well used by publishers and it is usually the same as the page title. It will be less confusing to users if we just use one.

trello card: https://trello.com/c/Uttw5Q63
Before:
image

After: 
image


beccapearce:
I think it makes sense to update it for both to keep it consistent. but maybe I should double check that that is correct.

And yes The behaviour should stay the same and match on both, people still search by slug so i think we should keep that

PR #8356, opened 2023-10-11
Spike out adding custom error messages via locales

davidgisbey:
### The problem

A common ask from design is that we give more descriptive error messages. At the moment the implementation uses default rails messages. We use `#full_message` https://api.rubyonrails.org/v6.1.4/classes/ActiveModel/Error.html#method-i-full_message in the[ error summary component](https://github.com/alphagov/whitehall/blob/8b7b867216ec45a3c15b6d00db5507187882a41c/app/components/admin/error_summary_component.rb#L30)

Essentially this combines the attribute and standard error message to construct the error message

For example a blank title would produce

```
Title can't be blank
```
as `error.attribute == :title` and `error.message == "can't be blank"

Eventually we want to end up with a full set of error messages which live in locales rather than the model itself. Getting to that point is actually slightly tricky as we have many models, all of which need custom error messages. Replacing all of these in one go in one PR would be a pretty insurmountable task.

It feels like we need a way to slowly add these in. Most likely 1 section or even model at a time.

Due to this we need to ensure the original implementation continues to work when custom error messages aren't present in the locales for that specific error.

### Proposed solution

This attempts to solve this issue by extending the error summary component. For each error we check whether an error message is present in the locales. If one is, then we use that error message in place of the error found on the object. If not, then we continue to use `#full_message`.

There's actually some more hidden complexity here as often there are nested errors which require some additional work to figure out where the locale should exist.

Originally, if one was found in the locales, i just used `error.message` but i noticed that errors added in the model itself take precedent. This can be fixed by removing the message from the model itself, but for shared errors (for example the `invalid_date` one added recently by Ryan it's used in many places.

I do think that probably the right thing to do is update all those messages in 1 go and remove the error message from the `DateValidation` concern. 

### Error text above inputs

I've not worried about extending this implementation to the [ErrorsHelper](https://github.com/alphagov/whitehall/blob/8b7b867216ec45a3c15b6d00db5507187882a41c/app/helpers/errors_helper.rb#L1) which handles rendering the error message above the relevant field. If people are happy with this approach it could quite easily be extended to this helper though

### DIFF
--- a/app/components/admin/error_summary_component.rb
+++ b/app/components/admin/error_summary_component.rb
@@ -27,7 +27,7 @@ private
   def error_items
     errors.map do |error|
       error_item = {
-        text: error.full_message,
+        text: error_message_text(error),
         data_attributes: track_analytics_data("form-error", analytics_action, error.full_message),
       }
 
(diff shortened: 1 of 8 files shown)


app/components/admin/error_summary_component.rb
```diff
@@ -51,4 +51,34 @@ def errors
                   object.errors
                 end
   end
+
+  def error_message_text(error)
+    if custom_error_present_in_locales?(error)
+      custom_error_message(error)
```
davidgisbey, line 57: i actually think this should just be `error.message` tbh

I ran into an issue with this being overwritten by [this](https://github.com/alphagov/whitehall/blob/8b7b867216ec45a3c15b6d00db5507187882a41c/app/models/date_validation.rb#L59) custom error message.

I think the right thing to do though is add custom messages for each affected attribute in the locales files.

davidgisbey:
THere's a wayyy more elegant solution for this so we'll go with that.

https://edgeguides.rubyonrails.org/configuring.html#configuring-active-model

PR #8361, opened 2023-10-12
Add custom error messaging for "invalid_date" for datetime fields

davidgisbey:
## Description

A common ask from design is that we give more descriptive error messages. At the moment the implementation uses default rails messages. We use `#full_message` https://api.rubyonrails.org/v6.1.4/classes/ActiveModel/Error.html#method-i-full_message in the[ error summary component](https://github.com/alphagov/whitehall/blob/8b7b867216ec45a3c15b6d00db5507187882a41c/app/components/admin/error_summary_component.rb#L30)

Essentially this combines the attribute and standard error message to construct the error message

For example a blank title would produce

```
Title can't be blank
```
as `error.attribute == :title` and `error.message == "can't be blank"

Eventually we want to end up with a full set of error messages which live in locales rather than the model itself. Getting to that point is actually slightly tricky as we have many models, all of which need custom error messages. Replacing all of these in one go in one PR would be a pretty insurmountable task.

Fortunately, Rails provides us with the functionality override the format of the `full_message` for individual models/attrs.

https://edgeguides.rubyonrails.org/configuring.html#configuring-active-model

This adds this setting to the config and adds a custom error message for datetime fields when the user inputs a date in an invalid format.

This PR does the minimal work possible to move this validation https://github.com/alphagov/whitehall/blob/ce464394bb23298757e621bd8def5f8bb2094bc2/app/models/date_validation.rb#L59
into a locale.

However, since we need to update the text for from `First published at` to `First published date` we've had to manually update all the error messages for that attr.

### Call for evidence & consultation response edit pages

image

image

## Guidance for review

One thing we need to be careful of is making the error messages too generic. For example, this page can have multiple date fields. 

If we use the same error message `Date is not in the correct format` we could have two links with the exact same text which would be an accessibility concern.

## Trello card

https://trello.com/c/ZuDpF2Cq/1646-design-investigate-formatting-help

### DIFF
--- a/app/models/date_validation.rb
+++ b/app/models/date_validation.rb
@@ -56,7 +56,7 @@ module DateValidation
   class DateValidator < ActiveModel::Validator
     def validate(record)
       record.invalid_date_attributes&.each do |date_attribute|
-        record.errors.add date_attribute, :invalid_date, message: "must be a valid date in the correct format"
+        record.errors.add date_attribute, :invalid_date
       end
     end
   end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -55,7 +55,7 @@ class Edition < ApplicationRecord
   validates :summary, presence: true, if: :summary_required?, length: { maximum: 65_535 }
   validates :previously_published, inclusion: { in: [true, false], message: "You must specify whether the document has been published before" }
   validates :first_published_at, presence: true, if: -> { previously_published || published_major_version }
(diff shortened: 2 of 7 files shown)


ryanb-gds:
What's the status of this @davidgisbey? Worth my having a review of it?

davidgisbey:
So this was ready for review. I took the approach Ollie wanted to take for adding custom error messages. We could merge it or hold off as it doesn't seem like we want to add the ability for this in Whitehall right now

ryanb-gds (approved):
I really like this and I feel this should be the default approach for validation messages. I'd say we should merge it and share with the team at dev catch up to say this is how we want to do error messages going forwards

PR #8363, opened 2023-10-12
Use "test_helper" rather than "fast_test_helper"

ollietreend:
The GovspeakLinkValidator test was failing when run independently of the rest of the MiniTest test suite:

```
rails test test/unit/lib/data_hygiene/govspeak_link_validator_test.rb
/Users/ollie.treend/govuk/whitehall/test/unit/lib/data_hygiene/govspeak_link_validator_test.rb:3:in `<main>': uninitialized constant Edition (NameError)

class Edition::GovspeakLinkValidatorTest < ActiveSupport::TestCase
      ^^^^^^^
```

This was happening because the test was including the "fast test helper", which only sets up a minimal set of dependencies, compared to the regular "test helper" which includes the entire application.

It was causing the newly introduced parallel tests to fail, for example here:
https://github.com/alphagov/whitehall/actions/runs/6495804055/job/17642125464

---

### DIFF
--- a/test/unit/lib/data_hygiene/govspeak_link_validator_test.rb
+++ b/test/unit/lib/data_hygiene/govspeak_link_validator_test.rb
@@ -1,4 +1,4 @@
-require "fast_test_helper"
+require "test_helper"
 
 class Edition::GovspeakLinkValidatorTest < ActiveSupport::TestCase
   test "should be valid if the input is nil" do


jyoung-gds (approved):
Thanks for debugging this, I was just looking at it. Could there be more cases?

ollietreend:
@jyoung-gds it looks like only a small handful of other tests use the `fast_test_helper` – and they all seem to pass fine when run independently.

I suspect we could probably get rid of the `fast_test_helper` altogether, because I'm not convinced it brings much value, but that's a refactor for another day I reckon!

PR #8368, opened 2023-10-16
Update slug field on the edit edition page to page address

davidgisbey:
## Description

On the edit edition page there is a read only input which shows a documents slug. This is confusing for users for 2 reasons:

1. they think they can edit it
2. they don't really know what a slug is

This updates the design to simply show the url of the document on GOV.UK.

### Before

image

### After

image

The screenshot here shows dev.gov.uk because it's on my local env. It would show

```
http://www.integration.gov.uk/government/publications/futures-toolkit-for-policy-makers-and-analysts
```

on int and 

```
http://www.gov.uk/government/publications/futures-toolkit-for-policy-makers-and-analysts
```

on prod

## Trello card 

https://trello.com/b/0n3p84mY/publishing-experience-team

### DIFF
--- a/features/step_definitions/edition_update_slug_steps.rb
+++ b/features/step_definitions/edition_update_slug_steps.rb
@@ -8,10 +8,10 @@ And(/^I update the slug to "([^"]*)"$/) do |new_slug|
   click_button "Update"
 end
 
-Then(/^I can see the slug has been updated to "([^"]*)"$/) do |new_slug|
+Then(/^I can see the edition's public URL contains "([^"]*)"$/) do |new_slug|
   visit admin_edition_path(@edition)
   click_button "Create new edition"
-  expect(find("#edition_slug").value).to eq new_slug
+  expect(page.find(".app-view-edit-edition__page-address .govuk-hint")).to have_content new_slug
 end
 
 Then(/^I am told I do not have access to the document$/) do
--- a/app/views/admin/editions/_standard_fields.html.erb
+++ b/app/views/admin/editions/_standard_fields.html.erb
@@ -9,13 +9,6 @@
     margin_bottom: 6,
   } %>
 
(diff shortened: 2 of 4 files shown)


features/step_definitions/edition_update_slug_steps.rb
```diff
@@ -8,10 +8,10 @@
   click_button "Update"
 end
 
-Then(/^I can see the slug has been updated to "([^"]*)"$/) do |new_slug|
+Then(/^I can see the editions public url contains "([^"]*)"$/) do |new_slug|
```
ollietreend, line 11:
```suggestion
Then(/^I can see the edition's public URL contains "([^"]*)"$/) do |new_slug|
```

ryanb-gds (approved):
Seems like a good idea to me. I take it there are no concerns (thinking accessibility, valid HTML usage) around using the hint component without an associated form input?

features/step_definitions/edition_update_slug_steps.rb
```diff
@@ -8,10 +8,10 @@
...
   visit admin_edition_path(@edition)
   click_button "Create new edition"
-  expect(find("#edition_slug").value).to eq new_slug
+  expect(page).to have_content new_slug
```
ollietreend, line 14: This matcher feels slightly misleading, because it would in theory match against anything on the page now – rather than directly in the `#edition_slug` element. I don't have an immediate solution. Maybe this is good enough tbh.

Another option could be to match against the entire page URL? e.g. something like:

```
I can see the edition's public URL contains "/government/edition-type/new-slug"
```

Which feels less likely to clash with anything else on the page.

No change needed necessarily – just something that I noticed.

ollietreend (commented):
Just a couple of things. One is a nitpick, the other is just me thinking out loud.

PR #8378, opened 2023-10-18
Bump Ruby to 3.2.2

davidgisbey:
## Description

We aim to maintain our apps [as close to the latest version as is practical](https://docs.publishing.service.gov.uk/manual/keeping-software-current.html). Starting with Whitehall is probably the best approach because it is likely to be the most challenging to update, and will therefore give us a good indication of how much effort will be required to update our other applications.


davidgisbey:
Being done elsewhere

PR #8381, opened 2023-10-18
Transition main navigation bar to new design system.

farahTW:
This PR transits main navigation bar to design system.

**It does the following:**
- Duplicated existing header partial view to legacy.
- Created new header partial with design system logic for main navigation.
- Added new method in base controller to handle toggle between legacy and new  header views.
- Updated admin and design_system views with new header partial view render.
- Added view tests to BaseController to verify new header renders for design system permissions.

**Screen shots:**
**Development env:**
image

**Integration env:**
image

**Production env:**
image

**Trello:**
https://trello.com/c/UFcEeugT/560-transition-main-navigation-bar-to-new-design-system

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -53,6 +53,11 @@ class Admin::BaseController < ApplicationController
     end
   end
 
+  def show_new_header?
+    current_user.can_preview_design_system?
+  end
+  helper_method :show_new_header?
+
 private
 
   def new_design_system?
--- a/app/views/layouts/admin.html.erb
+++ b/app/views/layouts/admin.html.erb
@@ -21,7 +21,7 @@
 <% end %>
 
 <% content_for :navbar do %>
-  <%= render "shared/header", admin_template: true %>
+  <%= render "shared/legacy_header", admin_template: true %>
   <% if t('admin.whats_new.show_banner') %>
     <div class="govuk-design-system-styling">
       <div class="<%= yield(:full_width) === "true" ? "app-width-container--full-width" : "container" %>">
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -26,11 +26,15 @@
 
   <%= render "govuk_publishing_components/components/skip_link" %>
 
-  <div class="legacy-whitehall">
-    <div class="environment-<%= environment %> navbar-wrapper">
-      <%= render partial: "shared/header" %>
+  <% if show_new_header? %>
+    <%= render partial: "shared/header" %>
+  <% else %>
+    <div class="legacy-whitehall">
+      <div class="environment-<%= environment %> navbar-wrapper">
+        <%= render partial: "shared/legacy_header" %>
+      </div>
     </div>
-  </div>
+  <% end %>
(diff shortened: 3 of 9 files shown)


app/views/layouts/admin.html.erb
minhngocd, line 1: I don't think we need to change this file at all, right? 
When we drew up the quadrant of users, the group that uses admin layout should never really see the new header. 
it should always render legacy.

app/controllers/admin/base_controller.rb
minhngocd, line 1: Are we also missing some tests to check if design preview users see the new header, and the other users see legacy?

app/views/shared/_header.html.erb
```diff
@@ -9,47 +9,43 @@
   navbar-default
   navbar-inverse
   navbar-static-top
-  <% if !t('admin.whats_new.show_banner') && admin_template %>add-bottom-margin<% end %>
-  <% if environment_style %>environment-indicator<% end %>" role="banner">
```
davidtrussler, line 13: I think we should remove the `<header>` element at this stage so that we do not nest one of these elements within another, which could affect the way it renders.

I don't think it matters that this will also remove the navigation from header because we are also going to have to adopt a slightly different structure to the existing one when we use the `layout_header` component, so that the navigation is not structurally part of the header but is a separate `<nav>` element (but that is part of the next piece of work).

app/views/layouts/design_system.html.erb
```diff
@@ -28,7 +28,11 @@
...
   <div class="legacy-whitehall">
     <div class="environment-<%= environment %> navbar-wrapper">
-      <%= render partial: "shared/header" %>
+      <% if show_new_header? %>
+        <%= render partial: "shared/header" %>
+      <% else %>
+        <%= render partial: "shared/legacy_header" %>
+      <% end %>
     </div>
   </div>
```
davidtrussler, line 37: I think this will need to be restructured a bit like below so that when using the new header we don't nest it in the classes we don't need (`legacy-whitehall`, `navbar-wrapper` etc.)

```
<% if show_new_header? %>
    <%= render partial: "shared/header" %>
<% else %>
  <div class="legacy-whitehall">
    <div class="environment-<%= environment %> navbar-wrapper">
        <%= render partial: "shared/legacy_header" %>
    </div>
  </div>
<% end %>
```

app/views/shared/_header.html.erb
farahTW: @davidtrussler 
As discussed I have removed the header tag and update the design view. Now the view will look like below and it will get fixed once sub-navigation will get implemented.

image

davidtrussler (approved):
The changes made in regard to the comments I left look good to me. 👍

app/controllers/admin/base_controller.rb
farahTW: done

minhngocd (approved):

farahTW:
Hi @nnagewad !
Please find below screen shots for highlighted tab in main navigation header

nnagewad:
Nice! The changes look great!

mtaylorgds:
Latest changes LGTM.

PR #8383, opened 2023-10-19
562 add more page

davidtrussler:
Trello: [562 | Link More drop down in the sub navigation to new page with "more" links](https://trello.com/c/UCCxgSnI/562-link-more-drop-down-in-the-sub-navigation-to-new-page-with-more-links)

This PR creates a new view which users will be directed to when they click on the "More" link in the navigation. Screenshot below. 

Changes: 
- Updates routes file for the new page
- Adds a new controller for the new page
- Adds a new view for the new page
- Adds a test for the new page
- Adds temporary "forbidden" toggle for users without Design System permissions
- Reorders links and changes "Departments & Agencies" to "Organisations"

Screenshot 2023-10-25 at 16 27 26

### DIFF
--- a/app/controllers/admin/more_controller.rb
+++ b/app/controllers/admin/more_controller.rb
@@ -1,0 +1,5 @@
+class Admin::MoreController < Admin::BaseController
+  layout "design_system"
+
+  def index; end
+end
--- a/app/controllers/admin/more_controller.rb
+++ b/app/controllers/admin/more_controller.rb
@@ -1,5 +1,22 @@
 class Admin::MoreController < Admin::BaseController
-  layout "design_system"
+  before_action :check_new_design_system_permissions, only: %i[index]
+  layout :get_layout
 
   def index; end
+
+private
+
+  def check_new_design_system_permissions
+    forbidden! unless new_design_system?
+  end
+
+  def get_layout
+    design_system_actions = %w[index] if preview_design_system?(next_release: false)
+
+    if design_system_actions&.include?(action_name)
+      "design_system"
+    else
+      "admin"
+    end
+  end
 end
--- a/app/helpers/admin/url_helper.rb
+++ b/app/helpers/admin/url_helper.rb
@@ -28,7 +28,7 @@ module Admin::UrlHelper
   end
 
   def admin_organisations_link
-    admin_link "Departments & agencies", admin_organisations_path
+    admin_link "Organisations", admin_organisations_path
   end
 
   def admin_roles_header_menu_link
--- a/app/helpers/admin/url_helper.rb
+++ b/app/helpers/admin/url_helper.rb
@@ -74,7 +74,11 @@ module Admin::UrlHelper
   end
 
   def admin_header_menu_link(name, path)
-    tag.li(link_to(name, path, role: "menuitem"), class: "masthead-menu-item")
+    if preview_design_system?
+      link_to(name, path, class: "govuk-link")
+    else
+      tag.li(link_to(name, path, role: "menuitem"), class: "masthead-menu-item")
+    end
   end
 
   def admin_header_link(name, path, path_matcher = nil, options = {})
--- a/app/helpers/admin/url_helper.rb
+++ b/app/helpers/admin/url_helper.rb
@@ -19,28 +19,56 @@ module Admin::UrlHelper
     admin_header_menu_link "Topical events", admin_topical_events_path
   end
 
+  def admin_topical_events_link
+    admin_link "Topical events", admin_topical_events_path
+  end
+
   def admin_organisations_header_menu_link
     admin_header_menu_link "Departments & agencies", admin_organisations_path
+  end
+
+  def admin_organisations_link
+    admin_link "Departments & agencies", admin_organisations_path
   end
 
   def admin_roles_header_menu_link
     admin_header_menu_link "Roles", admin_roles_path
   end
 
+  def admin_roles_link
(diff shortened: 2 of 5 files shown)


minhngocd:
Hi @davidtrussler All looks good to me, just a question about the tests. Should we add a test somewhere that checks if the user is a preview design user they should see the 1 link without the list, and if they are without the permission they should?

davidtrussler:
> Hi @davidtrussler All looks good to me, just a question about the tests. Should we add a test somewhere that checks if the user is a preview design user they should see the 1 link without the list, and if they are without the permission they should?

Thinking about the test, and if I understand what you are saying correctly, wouldn't that be a test that would belong in the [Transition sub navigation bar to the new design system](https://trello.com/c/LsOoy6Dv/563-transition-sub-navigation-bar-to-the-new-design-system) card? This page only exists in the design system scenario, but the "More" link appears in both but depends on the user's permission as to what it actually is.

minhngocd:
> > Hi @davidtrussler All looks good to me, just a question about the tests. Should we add a test somewhere that checks if the user is a preview design user they should see the 1 link without the list, and if they are without the permission they should?
> 
> Thinking about the test, and if I understand what you are saying correctly, wouldn't that be a test that would belong in the [Transition sub navigation bar to the new design system](https://trello.com/c/LsOoy6Dv/563-transition-sub-navigation-bar-to-the-new-design-system) card? This page only exists in the design system scenario, but the "More" link appears in both but depends on the user's permission as to what it actually is.

Heyya @davidtrussler . 
I think we still need a test for what the content is based on user permission. I can see there is logic for "if user is design system then show 1 link, otherwise show a list of tags" but I don't see a test for this logic.

davidtrussler:
> > > Hi @davidtrussler All looks good to me, just a question about the tests. Should we add a test somewhere that checks if the user is a preview design user they should see the 1 link without the list, and if they are without the permission they should?
> > 
> > 
> > Thinking about the test, and if I understand what you are saying correctly, wouldn't that be a test that would belong in the [Transition sub navigation bar to the new design system](https://trello.com/c/LsOoy6Dv/563-transition-sub-navigation-bar-to-the-new-design-system) card? This page only exists in the design system scenario, but the "More" link appears in both but depends on the user's permission as to what it actually is.
> 
> Heyya @davidtrussler . I think we still need a test for what the content is based on user permission. I can see there is logic for "if user is design system then show 1 link, otherwise show a list of tags" but I don't see a test for this logic.

@minhngocd Ah right, I understand. That's in the helper, right? That doesn't quite do that, it just formats the stuff that's returned in a way that is more useable in the (list) component we are using when we're in the design system. I tried to cover that with the existing test (checking for a `govuk-link` element) though there might be a better way. In other words the "more" view will only render in the design system rather than the same view rendering differently depending on permissions.

minhngocd:
> > > > Hi @davidtrussler All looks good to me, just a question about the tests. Should we add a test somewhere that checks if the user is a preview design user they should see the 1 link without the list, and if they are without the permission they should?
> > > 
> > > 
> > > Thinking about the test, and if I understand what you are saying correctly, wouldn't that be a test that would belong in the [Transition sub navigation bar to the new design system](https://trello.com/c/LsOoy6Dv/563-transition-sub-navigation-bar-to-the-new-design-system) card? This page only exists in the design system scenario, but the "More" link appears in both but depends on the user's permission as to what it actually is.
> > 
> > 
> > Heyya @davidtrussler . I think we still need a test for what the content is based on user permission. I can see there is logic for "if user is design system then show 1 link, otherwise show a list of tags" but I don't see a test for this logic.
> 
> @minhngocd Ah right, I understand. That's in the helper, right? That doesn't quite do that, it just formats the stuff that's returned in a way that is more useable in the (list) component we are using when we're in the design system. I tried to cover that with the existing test (checking for a `govuk-link` element) though there might be a better way. In other words the "more" view will only render in the design system rather than the same view rendering differently depending on permissions.

@davidtrussler  Ahhh Now that i've dug into what this actually does I get it now. Sorry about that.
I was under the impression we were also going to remove the dropdown for the More on the header itself. I think maybe we can try and shift this branch so that it is based off the Main header transition branch and directly on the new header partial (as opposed to the legacy header), we can remove the dropdown of More. What do you think?

davidtrussler:
> > @minhngocd Ah right, I understand. That's in the helper, right? That doesn't quite do that, it just formats the stuff that's returned in a way that is more useable in the (list) component we are using when we're in the design system. I tried to cover that with the existing test (checking for a `govuk-link` element) though there might be a better way. In other words the "more" view will only render in the design system rather than the same view rendering differently depending on permissions.
> 
> @davidtrussler Ahhh Now that i've dug into what this actually does I get it now. Sorry about that. I was under the impression we were also going to remove the dropdown for the More on the header itself. I think maybe we can try and shift this branch so that it is based off the Main header transition branch and directly on the new header partial (as opposed to the legacy header), we can remove the dropdown of More. What do you think?

@minhngocd I spoke with Nik about this and we've moved the bit about removing the dropdown (on both "More" and "New document") to the Acceptance Criteria for the subnav card, and kept the two new pages associated with those just to be adding views. So as it stands that is part of that work rather than this, which I think makes more sense as it's less reliant on work in progress elsewhere.

davidtrussler:
I've made a couple of updates to this PR to make some changes discussed: 
- Reordered the links and changed "Departments & Agencies" to "Organisations" @nnagewad 
- Added helper methods to display the links in the dropdown correctly @minhngocd

PR #8385, opened 2023-10-19
Update edition workflow document to reflect existence of unpublished state

ryanb-gds:
Missed updating this part of the documentation when I was working on the new "unpublished" state in #8304


jyoung-gds (approved):
👍

PR #8389, opened 2023-10-20
Prepare release for User and Error pages.

farahTW:
This PR make changes to the User  pages to prepare for release
It does the following

It removes the get_layout method from user controller.
It replaces the layout method in user controller with "design_system".
It removes bootstrap implementation legacy files.
It removes bootstrap implementation from feature tests.

**Note:**  No changes required for Error pages for this release as these are implemented in design system layout.

**Trello:**
https://trello.com/c/tIUQ9pDb/578-prepare-release-for-user-and-error-pages


app/controllers/admin/users_controller.rb
```diff
@@ -1,22 +1,17 @@
...
 
-  def show
-    render_design_system(:show, :legacy_show)
-  end
+  def show; end
 
   def edit
     unless @user.editable_by?(current_user)
       head :forbidden
-      return
```
minhngocd, line 17: I believe you might still need a return here otherwise it will render the edit page? 
Can try it out?

minhngocd (commented):
Are the Error pages also part of this release? or does it require something different?

features/step_definitions/user_steps.rb
```diff
@@ -35,14 +30,9 @@
 
 Then(/^I should see that I am logged in as a "([^"]*)"$/) do |role|
   visit admin_user_path(@user)
-  click_link "#user_settings"
```
davidtrussler, line 38: I haven't tried running this but I would have expected this line to still be necessary since it's outside the `if` clause.

farahTW: Hi David as part of new implementation we do have user setting link anymore.We will check it again though.

app/controllers/admin/users_controller.rb
farahTW: Hi Minno,
Ideally Edit link should not appear if user is not having GDS editor permission.
I will check again

minhngocd: Hey Farah - the link would not appear, but if you visit the URL, i think the page will still render unless we have this return line

features/step_definitions/user_steps.rb
mtaylorgds: @davidtrussler The line above is already visiting the user settings page, I believe? So clicking the link takes them to the page they're already on.

If the desire is to test the link then maybe we should be visiting another page and following the link from there, otherwise I don't think we need the link…thoughts?

davidtrussler: @farahTW / @mtaylorgds Yes, that's fine. Probably the link shouldn't have been there still but anyway LGTM.

davidtrussler (approved):
LGTM, but there is still an outstanding discussion to be resolved before merging.

minhngocd:
return renders resolved, LGTM

farahTW:
Merging the PR after all discussions resolved with @minhngocd  and @davidtrussler  as a part of  code review.

PR #8390, opened 2023-10-20
Stop marking attachments as present at unpublished

ryanb-gds:
Now that we have an unpublished state for editions, we no longer need to keep track of whether attachments were present at the edition's unpublishing, because unpublished editions now have to be superseded. Previously, unpublished editions were put into the draft state and it was therefore necessary to track which attachments were present when it was unpublished so that we could distinguish them from attachments which had been added to the draft since unpublishing.

I have added a migration to remove the database column from the attachment data table as well. I was undecided about whether to include this in the PR. On the one hand it's an irreversible change, so it would probably make sense to merge this PR and run for a few days without marking attachments as present to check that there are no ill effects. On the other, not including the migration presents the risk that we might forget to come back to tidy up and we'll have a dead database column hanging around. Let me know if the preferred option would be option 1 and I'll remove the migration commit from the PR.

Trello: https://trello.com/c/X4KpEBA7

### DIFF
--- a/db/migrate/20231020151031_drop_present_at_unpublishing_column_from_attachments.rb
+++ b/db/migrate/20231020151031_drop_present_at_unpublishing_column_from_attachments.rb
@@ -1,0 +1,5 @@
+class DropPresentAtUnpublishingColumnFromAttachments < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :attachment_data, :present_at_unpublish, :boolean, if_exists: true
+  end
+end
--- a/app/models/attachment_data.rb
+++ b/app/models/attachment_data.rb
@@ -24,8 +24,6 @@ class AttachmentData < ApplicationRecord
 
   OPENDOCUMENT_EXTENSIONS = %w[ODT ODP ODS].freeze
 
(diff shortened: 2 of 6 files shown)


beccapearce (commented):
This is looking good to me! Have we tested on integration? just cautious since it's a remove column migration

davidgisbey:
Looks good to me too 👍

db/migrate/20231020151031_drop_present_at_unpublishing_column_from_attachments.rb
```diff
@@ -0,0 +1,5 @@
+class DropPresentAtUnpublishingColumnFromAttachments < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :attachment_data, :present_at_unpublish, :boolean
```
davidgisbey, line 3: If you deploy this out for testing you might run into an issue on merge where the build fails due to the column already having been dropped (unless you wait for the db to sync overnight).

If you want to merge today it might be worth adding if_exists

```
remove_column :attachment_data, :present_at_unpublish, :boolean, if_exists: true
```

beccapearce (approved):

PR #8391, opened 2023-10-20
Fix:Hide Edit link when user has no GDS Editor permission.

farahTW:
This PR adds logic to hide the "Edit" link at 'Individual User' page, when logged in user has no 'GDS Editor' permission.

It does the following:
Added helper method to 'show' view to hide "Edit" link based on the user permission.

**Screen shots:**
**User with GDS Editor permission:**
image

**User without GDS Editor permission:**
image

**Trello:**
https://trello.com/c/5CLy1wbs/582-fix-individual-user-page-to-hide-edit-link-when-user-has-no-gds-editor-permission

### DIFF
--- a/test/functional/admin/users_controller_test.rb
+++ b/test/functional/admin/users_controller_test.rb
@@ -70,4 +70,18 @@ class Admin::UsersControllerTest < ActionController::TestCase
     put :update, params: { id: @user.id, user: { world_location_ids: [world_location.id] } }
     assert_response :forbidden
   end
+
+  view_test "show: displays Edit link if user has gds editor permission to edit the record." do
+    login_as_preview_design_system_user(:gds_editor)
+
+    get :show, params: { id: @user.id }
+
+    assert_select ".govuk-summary-list__actions", text: /Edit/
+  end
+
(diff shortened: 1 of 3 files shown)


minhngocd:
Can we add a test to catch this, please?

minhngocd (changes requested):
Can we add a test to catch this scenario, please?

farahTW:
> Can we add a test to catch this scenario, please?

Hi Minno
Test exists for this scenario

test/functional/admin/users_controller_test.rb
```diff
@@ -70,4 +70,14 @@ class Admin::UsersControllerTest < ActionController::TestCase
     put :update, params: { id: @user.id, user: { world_location_ids: [world_location.id] } }
     assert_response :forbidden
   end
+
+  view_test "show: hides Edit link if user has no gds editor permission to edit the record or display the link if user has gds editor permission." do
```
mtaylorgds, line 74: I think this might be clearer as two separate tests—one with permission, and one without?

minhngocd (approved):
👍

PR #8397, opened 2023-10-23
Add rake task to report on a regex in editionable content

brucebolt:
Sometimes we want to identify which published content matches a regular expression (e.g. if we make a change to govspeak and need to republish affected content).

Therefore adding a rake task that will report on the content that includes a given regex in the currently published edition.

This is being broken down into batches of 1000, as our infrastructure does not support large queries being made on a Rails console.

[Trello card](https://trello.com/c/jD9VLG1A)

### DIFF
--- a/lib/tasks/reporting.rake
+++ b/lib/tasks/reporting.rake
@@ -12,4 +12,17 @@ namespace :reporting do
   task published_attachments_report: :environment do
     Reports::PublishedAttachmentsReport.new.report
   end
+
+  desc "Prints a list of content IDs that documents whose live edition contains a given regular expression"
+  task :matching_docs, [:regex] => :environment do |_, args|
+    regex = Regexp.new(/#{args[:regex]}/)
+
+    Document.where.not(live_edition_id: nil).find_in_batches(batch_size: 1000) do |batch|
+      batch.each do |document|
+        next unless document.editions.published.any?
+
+        puts document.content_id if regex.match?(document.editions.published.last.body)
+      end
+    end
+  end
 end
--- a/test/unit/lib/tasks/reporting_test.rb
+++ b/test/unit/lib/tasks/reporting_test.rb
@@ -1,0 +1,26 @@
+require "test_helper"
+require "rake"
+
+class ReportingRake < ActiveSupport::TestCase
+  setup do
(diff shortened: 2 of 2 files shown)


test/unit/lib/tasks/reporting_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    @document_2 = create(:draft_edition, body: "Some text 2")
+    @document_3 = create(:published_edition, body: "Some other text 1")
+  end
+
+  teardown do
+    Rake::Task["reporting:matching_docs"].reenable
+  end
+
+  test "it prints the content IDs of the matching documents from published editions" do
+    assert_output(/#{@document_1.document.content_id}/) { Rake.application.invoke_task "reporting:matching_docs[Some text]" }
```
jkempster34, line 16: I wondered if these could not use regex like `assert_output(@document_1.document.content_id) { Rake.application.invoke_task "reporting:matching_docs[Some text]" }`, but I don't know if there is a way to do `contains`

lib/tasks/reporting.rake
```diff
@@ -12,4 +12,17 @@ namespace :reporting do
   task published_attachments_report: :environment do
     Reports::PublishedAttachmentsReport.new.report
   end
+
+  desc "Prints a list of content IDs that documents whose live edition contains a given regular expression"
+  task :matching_docs, [:regex] => :environment do |_, args|
+    regex = Regexp.new(/#{args[:regex]}/)
+
+    Document.where.not(live_edition_id: nil).find_in_batches(batch_size: 1000) do |batch|
```
jkempster34, line 20: I saw on the Trello card that you were investigating other table,  are there none?

jkempster34 (approved):

lib/tasks/reporting.rake
brucebolt: I've constrained this PR to just the editionable documents.  I was going to raise in today's stand-up what the scope of the spike is, as I've already hit the 2 days on the ticket.

test/unit/lib/tasks/reporting_test.rb
brucebolt: Yeah, I couldn't find an equivalent method that didn't require a regex.

PR #8402, opened 2023-10-23
561 new document sub navigation page

RodneyJohnsonGDS:
Trello: [561 | Link New Document drop down in the sub navigation to new page with radio selection](https://trello.com/c/BTRbQkCR/561-link-new-document-drop-down-in-the-sub-navigation-to-new-page-with-radio-selection)

This PR creates a new view which users will be directed to when they click on the "New document" link in the navigation bar.

Changes:

- When a user makes a selection on the type of new document they would like to create they are redirected to the correct new document type.
- When there are no radio buttons selected - they are shown an error and they remain on the New document page - this is to ensure we are not influencing their selection.
- Tests coverage includes testing the users permission granting only certain users the ability to view this page, functionality re when a user selects a radio button / does not select a button the test ensures they are shown the correct new document page or an error respectively.
- This PR also includes changes to the routes file and the necessary controller implementations

Screenshot below:

This screenshot is what the user sees when they first land on the landing page:

whitehall-admin dev gov uk_government_admin_new-document (3)

---

This screenshot is what the user sees when they select Next without selecting a radio button:

whitehall-admin dev gov uk_government_admin_new-document (4)

### DIFF
--- a/app/controllers/admin/new_document_controller.rb
+++ b/app/controllers/admin/new_document_controller.rb
@@ -1,0 +1,45 @@
+class Admin::NewDocumentController < Admin::BaseController
+  before_action :check_new_design_system_permissions, only: %i[index]
+
+  layout :get_layout
+
+  def index; end
+
+  def new_document_options_redirect
+    redirect_options = {
+      "call-for-evidence": new_admin_call_for_evidence_path,
+      "case-study": new_admin_case_study_path,
+      "consultation": new_admin_consultation_path,
+      "detailed-guide": new_admin_detailed_guide_path,
+      "document-collection": new_admin_document_collection_path,
+      "fatality-notice": new_admin_fatality_notice_path,
+      "news-article": new_admin_news_article_path,
+      "publication": new_admin_publication_path,
+      "speech": new_admin_speech_path,
+      "statistical-data-set": new_admin_statistical_data_set_path,
+    }
+
+    if params[:new_document_options].present?
+      selected_option = params.require(:new_document_options).to_sym
+      redirect_to redirect_options[selected_option]
+    else
+      redirect_to admin_new_document_path, alert: "Please select a new document option"
+    end
+  end
+
+private
+
+  def check_new_design_system_permissions
+    forbidden! unless new_design_system?
+  end
+
+  def get_layout
+    design_system_actions = %w[index new_document_options_redirect] if preview_design_system?(next_release: false)
+
+    if design_system_actions&.include?(action_name)
+      "design_system"
+    else
+      "admin"
+    end
+  end
+end
--- a/app/views/admin/new_document/index.html.erb
+++ b/app/views/admin/new_document/index.html.erb
@@ -1,0 +1,99 @@
+<% content_for :page_title, "New document" %>
+<% content_for :title, "New document" %>
+<% content_for :title_margin_bottom, 6 %>
+
+<%
+  new_document_radio_items = [
+    {
+      value: "call-for-evidence",
+      text: "Call for evidence",
+      bold: true,
+      hint_text: "Use this to request people's views when it is not a consultation.",
+    },
+    {
+      value: "case-study",
+      text: "Case study",
+      bold: true,
+      hint_text: "Use this to share real examples that help users understand a process or an important" +
+        " aspect of government policy covered on GOV.UK.",
+    },
+    {
+      value: "consultation",
+      text: "Consultation",
+      bold: true,
+      hint_text: "Use this for documents requiring a collective agreement across government, and requests for people's view on a question with an outcome.",
+    },
+    {
+      value: "detailed-guide",
+      text: "Detailed guide",
+      bold: true,
+      hint_text: "Use this to tell users the steps they need to take to complete a clearly defined task. They are usually aimed at specialist or professional audiences.",
+    },
+    {
+      value: "document-collection",
+      text: "Document collection",
+      bold: true,
+      hint_text: "Use this to group related documents on a single page for a specific audience or around a specific theme.",
+    },
+    {
+      value: "fatality-notice",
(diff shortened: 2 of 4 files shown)


nnagewad:
This looks great 👍🏽

app/views/admin/new_document/index.html.erb
```diff
@@ -0,0 +1,99 @@
...
+      bold: true,
+      hint_text: "Use this for data that you publish monthly or more often without analysis.",
+    },
+  ]
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with url: admin_new_document_options_path do %>
+      <%= render "govuk_publishing_components/components/radio", {
```
davidtrussler, line 74: Where we have a bunch of radio controls we should also have a `<legend>` element inside the `<fieldset>` which is created as part of the `Form radio button ` component. This could be flagged as an accessibility failure/warning. It can be set via the `heading` parameter in the component and should ideally use the page title as its value, removing the page title itself.

farahTW (commented):
Hey @RodneyJohnsonGDS 
After refactoring the test controller code wise it looks ok to me.
Just wondering if can squash all commits in to one that will look  more neater as all files are related to one component.

app/controllers/admin/new_document_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      "publication": new_admin_publication_path,
+      "speech": new_admin_speech_path,
+      "statistical-data-set": new_admin_statistical_data_set_path,
+    }
+
+    if params[:new_document_options].present?
+      selected_option = params.require(:new_document_options).to_sym
+      redirect_to redirect_options[selected_option]
+    else
+      redirect_to admin_new_document_path, alert: "Please select a new document option"
```
davidtrussler, line 26: Rather than using the alert here we should be using the error component, which is more consistent with what we have done elsewhere and more closely follows the design.

davidtrussler:
@RodneyJohnsonGDS I've left a couple of comments on here about refactoring the page title into a `<legend>` element and setting up error messages.

davidtrussler:
> Hey @RodneyJohnsonGDS After refactoring the test controller code wise it looks ok to me. Just wondering if can squash all commits in to one that will look more neater as all files are related to one component.

I'd agree with squashing the commits. If we think about it as how it would be read in the context of the overall commit history I think a single commit that says this introduces a new page for selecting a new document would be easier to read and understand.

test/functional/admin/new_document_controller_test.rb
```diff
@@ -0,0 +1,54 @@
+require "test_helper"
```
mtaylorgds, line 1: It would be nice if the tests were formatted consistently in terms of blank lines—some have no blank lines, others separate the arrange/act/assert parts. Personally, I find it easier to skim through tests when they are separated into Arrange/Act/Assert sections by [separating each section by blank lines](https://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern/#77fd4d083b8541eab04a75783bd46719).

test/functional/admin/new_document_controller_test.rb
```diff
@@ -0,0 +1,54 @@
...
+    assert_select ".govuk-inset-text", text: "Check the content types guidance if you need more help in choosing a content type."
+  end
+
+  test "access to the New document index page is forbidden for users without design system permissions" do
+    login_as :writer
+    get :index
+    assert_response :forbidden
+  end
+
+  test "POST #new_document_options_redirect redirects to #call-for-evidence if search option passed is call-for-evidence" do
```
mtaylorgds, line 27: Should we be testing all the links to make sure they all redirect as expected? Maybe use parameterized tests to avoid having to duplicate the same test multiple for each link.

RodneyJohnsonGDS: Sounds good - I've now setup a parametized test that ensures each radio buttons redirects the user to the correct path, thanks @mtaylorgds

test/functional/admin/new_document_controller_test.rb
RodneyJohnsonGDS: Done!

app/controllers/admin/new_document_controller.rb
RodneyJohnsonGDS: Thanks for the feedback David! As discussed we will pick up how we handle errors across the codebase as a separate user story. For the time being, we'll keep implementation as is.

app/views/admin/new_document/index.html.erb
RodneyJohnsonGDS: Accessibility is important - well spotted! Done!

test/functional/admin/new_document_controller_test.rb
mtaylorgds: This still doesn't seem to be consistent to me? Happy to jump on a quick call if my comment isn't clear.

app/views/admin/new_document/index.html.erb
davidtrussler: Hi @RodneyJohnsonGDS - the `<legend>` change is perfect, ta! Only thing is that we still need the `<% content_for :page_title, "New document" %>` line that has been removed so that the page title appears in the browser tab.

test/functional/admin/new_document_controller_test.rb
mtaylorgds: These look good now 👍

app/views/admin/new_document/index.html.erb
RodneyJohnsonGDS: Done!

davidtrussler (approved):
LGTM! 👍

PR #8407, opened 2023-10-24
Update reporting rake task to include all govspeak columns

brucebolt:
This adds all the fields which could contain govspeak into the `reporting:matching_docs` rake task, allowing us to see all document types that would need to be republished if part of the govspeak gem were to be updated.

[Trello card](https://trello.com/c/jD9VLG1A)

### DIFF
--- a/lib/tasks/reporting.rake
+++ b/lib/tasks/reporting.rake
@@ -13,16 +13,40 @@ namespace :reporting do
     Reports::PublishedAttachmentsReport.new.report
   end
 
-  desc "Prints a list of content IDs that documents whose live edition contains a given regular expression"
+  desc "Prints a list of content IDs for documents whose govspeak content contains a given regular expression"
   task :matching_docs, [:regex] => :environment do |_, args|
     regex = Regexp.new(/#{args[:regex]}/)
 
-    Document.where.not(live_edition_id: nil).find_in_batches(batch_size: 1000) do |batch|
-      batch.each do |document|
-        next unless document.editions.published.any?
+    Document.where.not(live_edition_id: nil).find_each do |object|
+      next unless object.editions.published.any?
 
-        puts document.content_id if regex.match?(document.editions.published.last.body)
-      end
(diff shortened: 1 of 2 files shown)


lib/tasks/reporting.rake
```diff
@@ -13,16 +13,50 @@ namespace :reporting do
...
+      end
+    end
+
+    WorldLocationNews.find_in_batches(batch_size: 1000) do |batch|
+      batch.each do |object|
+        document_includes_regex(regex, object.content_id, object.class.name, object.mission_statement)
+      end
+    end
+
+    WorldwideOffice.find_in_batches(batch_size: 1000) do |batch|
```
jkempster34, line 52: I'm not sure if it's as easy for Documents and Editions, but could these be a db query instead?
```ruby
WorldwideOffice.where("access_and_opening_times REGEXP ?", '#{regex}').find_each { |d| puts d.content_id }
```

Would that be significantly faster, or not really?

jkempster34 (approved):
LGTM, left a comment about whether querying at the db level would return the same result and be quicker

lib/tasks/reporting.rake
```diff
@@ -13,16 +13,50 @@ namespace :reporting do
...
-        next unless document.editions.published.any?
+      batch.each do |object|
+        next unless object.editions.published.any?
 
-        puts document.content_id if regex.match?(document.editions.published.last.body)
+        document_includes_regex(regex, object.content_id, object.class.name, object.editions.published.last.body)
+      end
+    end
+
+    HtmlAttachment.find_in_batches(batch_size: 1000) do |batch|
```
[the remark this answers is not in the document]
brucebolt: Nice!  I didn't realise it also used `find_in_batches`.  Have updated that.

lib/tasks/reporting.rake
brucebolt: I didn't see any change in speed (there's so few records anyway) for Person, PolicyGroup, WorldLocationNews and WorldwideOffice.  It may speed up for Document and HtmlAttachment, but the query would likely become complex as I'd need to include some joins.

PR #8408, opened 2023-10-24
Create rake task for Image Data Relationship

yuetylauiris:
We want to migrate the DefaultNewsOrganisationImageData to FeaturedImageData so that we do the migration of assets before switching over to use FeaturedImageData for non legacy flow.

[Trello Card](https://trello.com/c/hxnlRSx1/205-story-defaultnewsorganisationimagedatafeaturedimagedata-to-use-assetid)


db/schema.rb
```diff
@@ -1092,6 +1094,8 @@
     t.date "end_date"
     t.string "content_id"
     t.text "summary"
+    t.bigint "default_news_organisation_image_data_id"
```
tunylund, line 1097: this doesn't belong here

PR #8409, opened 2023-10-24
Refactor AttachmentUpdaterTest to reduce use of stubs and improve readability

ryanb-gds:
This test was making very heavy use of let and also doing some unnecessary mocking and stubbing when we could use real factory-built objects instead, so I thought it would be worth refactoring a bit. Previously I found the test very difficult to modify when I had to add a case for checking that unpublished documents had their attachments updated as expected. Hopefully this should help if we need to make any further changes.

### DIFF
--- a/test/unit/app/services/asset_manager/attachment_updater_test.rb
+++ b/test/unit/app/services/asset_manager/attachment_updater_test.rb
@@ -4,326 +4,248 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
   extend Minitest::Spec::DSL
 
   describe AssetManager::AttachmentUpdater do
-    let(:updater) { AssetManager::AttachmentUpdater }
-    let(:update_service) { mock("asset-manager-update-asset-worker") }
-    let(:attachment_data) { attachment.attachment_data }
-    let(:simple_pdf) { File.open(fixture_path.join("simple.pdf")) }
-    let(:attachment) { FactoryBot.create(:file_attachment, file: simple_pdf) }
-    let(:original) { Asset.variants[:original] }
-    let(:thumbnail) { Asset.variants[:thumbnail] }
-    let(:original_asset) { Asset.new(asset_manager_id: "original_asset_manager_id", variant: original, filename: "simple.pdf") }
-    let(:thumbnail_asset) { Asset.new(asset_manager_id: "thumbnail_asset_manager_id", variant: thumbnail, filename: "thumbnail_simple.pdf.png") }
-
-    around do |test|
-      AssetManager.stub_const(:AssetUpdater, update_service) do
-        test.call
-      end
-    end
-
-    before do
-      attachment_data.assets = [original_asset, thumbnail_asset]
-    end
-
-    context "when the attachment has been deleted" do
-      before do
-        attachment.delete
-      end
-
-      it "does not update the asset" do
-        AssetManager::AssetUpdater.expects(:call).never
-
-        updater.call(attachment_data, redirect_url: true, draft_status: true)
-      end
-    end
-
-    describe "access limited updates" do
-      context "when attachment's attachable is access limited" do
-        before do
-          access_limited_object = stub("access-limited-object")
-          @organisation = create(:organisation)
-          AssetManagerAccessLimitation.stubs(:for).with(access_limited_object).returns([@organisation.content_id])
-
-          attachment_data.stubs(:access_limited?).returns(true)
-          attachment_data.stubs(:access_limited_object).returns(access_limited_object)
-        end
-
-        it "updates the access limited state of all assets" do
-          update_service.expects(:call)
-                          .with(original_asset.asset_manager_id, attachment_data, nil, { "access_limited_organisation_ids" => [@organisation.content_id] })
-          update_service.expects(:call)
-                        .with(thumbnail_asset.asset_manager_id, attachment_data, nil, { "access_limited_organisation_ids" => [@organisation.content_id] })
-
-          updater.call(attachment_data, access_limited: true)
-        end
-      end
-
-      context "when attachment's attachable is not access limited" do
-        before do
-          attachment_data.stubs(:access_limited?).returns(false)
-        end
-
-        it "updates all assets to have an empty access_limited array" do
-          update_service.expects(:call)
-                          .with(original_asset.asset_manager_id, attachment_data, nil, { "access_limited_organisation_ids" => [] })
-          update_service.expects(:call)
-                        .with(thumbnail_asset.asset_manager_id, attachment_data, nil, { "access_limited_organisation_ids" => [] })
-
-          updater.call(attachment_data, access_limited: true)
-        end
-      end
-    end
-
-    describe "draft status updates" do
-      let(:draft) { true }
-      let(:unpublished) { false }
-      let(:replaced) { false }
-
-      before do
-        attachment_data.stubs(:draft?).returns(draft)
-        attachment_data.stubs(:unpublished?).returns(unpublished)
-        attachment_data.stubs(:replaced?).returns(replaced)
-      end
+    context "when the attachment's attachable is a draft" do
+      let(:edition) { create(:draft_news_article) }
+      let(:attachment) { create(:file_attachment, attachable: edition) }
 
       it "marks corresponding assets as draft" do
-        update_service.expects(:call)
-                        .with(original_asset.asset_manager_id, attachment_data, nil, { "draft" => true })
-        update_service.expects(:call)
-                      .with(thumbnail_asset.asset_manager_id, attachment_data, nil, { "draft" => true })
+        expected_attribute_hash = {
+          "draft" => true,
+        }
 
-        updater.call(attachment_data, draft_status: true)
+        attachment.attachment_data.assets.each do |asset|
+          AssetManager::AssetUpdater.expects(:call).with(asset.asset_manager_id, attachment.attachment_data, nil, expected_attribute_hash)
+        end
+
+        AssetManager::AttachmentUpdater.call(attachment.attachment_data, draft_status: true)
       end
 
-      context "and attachment is not draft" do
-        let(:draft) { false }
+      it "sets parent_document_url for all assets" do
+        expected_attribute_hash = { "parent_document_url" => edition.public_url(draft: true) }
 
-        it "marks corresponding assets as not draft" do
-          update_service.expects(:call)
-                          .with(original_asset.asset_manager_id, attachment_data, nil, { "draft" => false })
-          update_service.expects(:call)
-                        .with(thumbnail_asset.asset_manager_id, attachment_data, nil, { "draft" => false })
-
-          updater.call(attachment_data, draft_status: true)
+        attachment.attachment_data.assets.each do |asset|
+          AssetManager::AssetUpdater.expects(:call).with(asset.asset_manager_id, attachment.attachment_data, nil, expected_attribute_hash)
         end
+
+        AssetManager::AttachmentUpdater.call(attachment.attachment_data, link_header: true)
       end
 
-      context "and attachment is replaced" do
-        let(:replaced) { true }
+      it "updates all assets to have an empty access_limited array when the draft is not access limited" do
+        expected_attribute_hash = {
+          "access_limited_organisation_ids" => [],
+        }
 
-        it "marks corresponding assets as not draft even though attachment is draft" do
-          update_service.expects(:call)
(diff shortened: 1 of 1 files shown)


test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
-        end
+        AssetManager::AttachmentUpdater.call(attachment.attachment_data, draft_status: true)
+      end
 
-        it "updates all assets to have an empty access_limited array" do
-          update_service.expects(:call)
-                          .with(original_asset.asset_manager_id, attachment_data, nil, { "access_limited_organisation_ids" => [] })
-          update_service.expects(:call)
-                        .with(thumbnail_asset.asset_manager_id, attachment_data, nil, { "access_limited_organisation_ids" => [] })
+      it "sets parent_document_url for attachment using draft hostname" do
```
jyoung-gds, line 41: As mentioned in our call, it would be nice for this naming to match with the test on line 124 because they're testing the same thing but just draft vs non-draft states

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
+          AssetManager::AssetUpdater.expects(:call).with(asset.asset_manager_id, attachment.attachment_data, nil, expected_attribute_hash)
+        end
 
-        updater.call(attachment_data, draft_status: true)
+        AssetManager::AttachmentUpdater.call(attachment.attachment_data, access_limited: true)
       end
 
-      context "and attachment is not draft" do
-        let(:draft) { false }
+      it "marks corresponding assets as not draft even though the attachable is a draft when the attachment has been replaced" do
```
jyoung-gds, line 63: I think this can change to `it "marks corresponding assets as not draft when the attachment has been replaced"` to make it more readable. It's already in the context of `"when the attachment's attachable is a draft"` which covers `"even though the attachable is a draft"` part of the test name.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
-        let(:parent_document_url) { draft_edition.public_url(draft: true) }
+        AssetManager::AttachmentUpdater.call(attachment.attachment_data, draft_status: true)
+      end
 
-        it "sets parent_document_url for attachment using draft hostname" do
-          update_service.expects(:call)
-                          .with(original_asset.asset_manager_id, attachment_data, nil, { "parent_document_url" => parent_document_url })
-          update_service.expects(:call)
-                        .with(thumbnail_asset.asset_manager_id, attachment_data, nil, { "parent_document_url" => parent_document_url })
+      it "sets parent_document_url for all assets" do
```
jyoung-gds, line 124: See my comment on line 41 about these test names matching to convey similar behaviour

jyoung-gds (changes requested):
As discussed in the call we just had, all tests look good and coverage is replicated. I've just recommended some small changes to test names to make them more readable.

jyoung-gds (approved):
👍🚀

tunylund:
I really like the changes. I can't get myself to reading the meaning of all these changes in relation to the meaning of tests and making sure that no test is testing the wrong thing. I'm very happy to have a call about these.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
-    before do
-      attachment_data.assets = [original_asset, thumbnail_asset]
-    end
-
-    context "when the attachment has been deleted" do
-      before do
-        attachment.delete
-      end
+    context "when the attachment does not belong to an attachable" do
+      it "does not update the status of any assets" do
```
lauraghiorghisor-tw, line 8: I am not sure what this is meant to test. I believe all classes that include the `Attachable` concern have attachments and `AttachmentData`. Implicitly, they are the attachable. I doubt there is an attachment that does not have an attachable. On top of that, most attachables are editions. Now, there is a funny outlier, which is `PolicyGroup`.  For some reason the factories for attachments do set an attachable by default but it is `PolicyGroup` of all things. We actually decided to leave it be like that so that people are required to specify the attachable when it is relevant in the test. See [factory here](https://github.com/alphagov/whitehall/blob/c015d3cf5ad7e4a5adf3e1629f45f8d6a5671f10/test/factories/attachments.rb#L5). If it makes more sense to change the default to some edition, go for it. 
In this test, it passes because the link header update does not run as it is only set for editions, I think.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
       end
+    end
 
-      context "and attachment is replaced" do
-        let(:replaced) { true }
+    context "when the attachment's attachable is a draft and is access limited" do
+      it "updates the access limited state of all assets" do
+        edition = create(:draft_news_article, :access_limited)
+        attachment = create(:file_attachment, attachable: edition)
+        assets = [
```
lauraghiorghisor-tw, line 82: This whole thing can be removed and you can loop straight through attachment.attachment_data.assets. Factory should have the assets.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
-        let(:sample_rtf) { File.open(fixture_path.join("sample.rtf")) }
-        let(:sample_docx) { File.open(fixture_path.join("sample.docx")) }
-        let(:attachment_data) { AttachmentData.create!(file: sample_rtf, replaced_by: replacement) }
-        let(:replacement) { AttachmentData.create!(file: sample_docx) }
-
-        it "raises a AssetNotFound error" do
-          attachment_data.assets = [thumbnail_asset]
-          attachment_data.assets.create!(asset_manager_id: "asset_manager_id", variant: original, filename: "sample.rtf")
-          replacement.assets.create!(asset_manager_id: "replacement_asset_manager_id", variant: original, filename: "sample.docx")
+    context "when the attachment's attachable is unpublished" do
```
lauraghiorghisor-tw, line 216: Most of this context now wraps the replacement logic tests but I don't think there is a strict correlation there.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
+
+        AssetManager::AttachmentUpdater.call(attachment.attachment_data, access_limited: true)
       end
     end
 
-    describe "link header updates" do
-      let(:edition) { FactoryBot.create(:published_edition) }
-      let(:parent_document_url) { edition.public_url }
-      let(:attachment) { FactoryBot.create(:file_attachment, file: simple_pdf, attachable: edition) }
+    context "when attachment's attachable is published" do
```
lauraghiorghisor-tw, line 100: Most of this context now wraps the replacement logic tests but I don't think there is a strict correlation there.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
 
-    describe "access limited updates" do
-      context "when attachment's attachable is access limited" do
-        before do
-          access_limited_object = stub("access-limited-object")
-          @organisation = create(:organisation)
-          AssetManagerAccessLimitation.stubs(:for).with(access_limited_object).returns([@organisation.content_id])
+    context "when the attachment's attachable is a draft" do
+      let(:edition) { create(:draft_news_article) }
+      let(:attachment) { create(:file_attachment, attachable: edition) }
```
lauraghiorghisor-tw, line 19: The entire context runs fine with the default factory assets, no need to generate ✅
`let(:assets) { attachment.attachment_data.assets }`

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
-
-        it "raises a AssetNotFound error" do
-          attachment_data.assets = [thumbnail_asset]
-          attachment_data.assets.create!(asset_manager_id: "asset_manager_id", variant: original, filename: "sample.rtf")
-          replacement.assets.create!(asset_manager_id: "replacement_asset_manager_id", variant: original, filename: "sample.docx")
+    context "when the attachment's attachable is unpublished" do
+      it "updates redirect URL for all assets" do
+        edition = create(:unpublished_edition)
+        attachment = create(:file_attachment, attachable: edition)
+        assets = [
```
lauraghiorghisor-tw, line 220: Same as above.

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -4,281 +4,234 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
...
     end
 
-    describe "link header updates" do
-      let(:edition) { FactoryBot.create(:published_edition) }
-      let(:parent_document_url) { edition.public_url }
-      let(:attachment) { FactoryBot.create(:file_attachment, file: simple_pdf, attachable: edition) }
+    context "when attachment's attachable is published" do
+      let(:edition) { create(:published_news_article) }
+      let(:attachment) { create(:file_attachment, attachable: edition) }
+      let(:assets) do
```
lauraghiorghisor-tw, line 103: Same as above

lauraghiorghisor-tw:
Nice one. Definitely more readable.  I think it really does depend where your interest lies though, for us it helped to see it broken down in the type of updates. I'm sure you've already dug up the history, but it used to be [like this](https://github.com/alphagov/whitehall/tree/bebba154bd8d1e481193dfdb18f625b91ecc0569/app/services/asset_manager/attachment_updater) with the associated test files. The separation was nice in some ways. Nonetheless, I think it flows nicely in each context as you have it now. It matches better the style you see in their integration counterparts `test/integration/attachment_link_header_test.rb` and the like.

👍🏻

test/unit/app/services/asset_manager/attachment_updater_test.rb
ryanb-gds: Yeah I ended up removing this test in the other PR because you're right, there is no way for an attachment not to have an attachable. I did initially try changing it to "not an attachable" as you suggest, but it turns out according to [this test](https://github.com/alphagov/whitehall/blob/c015d3cf5ad7e4a5adf3e1629f45f8d6a5671f10/test/integration/attachment_draft_status_integration_test.rb#L100) the `AttachmentUpdater` should work for a `PolicyGroup` too. So I concluded that the test for not having an attachable did not make any sense.

test/unit/app/services/asset_manager/attachment_updater_test.rb
ryanb-gds: Thanks, I will remove these

test/unit/app/services/asset_manager/attachment_updater_test.rb
ryanb-gds: In fact, thinking about it, I would expect replacement only to apply for drafts? Presumably once the attachable has been published the user has to create a new edition before they can replace the attachment data?

test/unit/app/services/asset_manager/attachment_updater_test.rb
ryanb-gds: I have removed this test from this PR

test/unit/app/services/asset_manager/attachment_updater_test.rb
lauraghiorghisor-tw: That makes more sense yes. I think most updates are in draft, except for the ones triggered by the edition listeners, which are then caused by some user action such as publishing etc.

ryanb-gds: Having done a bit of further digging, I've found that the replacement logic is only invoked during bulk uploads. That user interface is only available when an edition is in draft state so I think testing with a draft attachable is appropriate.

PR #8412, opened 2023-10-25
Refactor attachment updater to separate methods

ryanb-gds:
Whilst [fixing this bug](https://github.com/alphagov/whitehall/pull/8403) I had to extend the Whitehall AttachmentUpdater class to handle unpublished editions, but it was difficult to do so owing to the fact that the class provides a single method to handle updating the attachment metadata, creating a redirect, and replacing the attachment data with a new version.

Separating out that method into three separate method calls will decouple the three operations so that each can be changed independently if needed in future, and will generally aid developers in understanding this area of the code. I've also moved a lot of the logic into the `AttachmentData` model which seems like a better home for it, as the attachment data should be responsible for managing its own state and exposing that to consumers. This approach avoids an [anaemic domain model](https://martinfowler.com/bliki/AnemicDomainModel.html)

I would still like to move some of the responsibility from the `AttachmentData` model on to the `Attachment` itself, as the `AttachmentData` model does far too much reaching across to the `Attachable` via the `Attachment` for my liking, resulting in excessive coupling.

The `AssetUpdater` API is also inefficient because it accepts a hash of attributes that it could derive itself from the `AttachmentData` parameter it accepts, so I might open a quick PR for that too. However, the `AssetUpdater` has quite a few usages across the codebase so I'd need to be careful about backward compatibility

Trello: https://trello.com/c/HKzC61Ub

### DIFF
--- a/test/unit/app/services/asset_manager/attachment_updater_test.rb
+++ b/test/unit/app/services/asset_manager/attachment_updater_test.rb
@@ -41,20 +41,6 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
 
         AssetManager::AttachmentUpdater.call(attachment.attachment_data, access_limited: true)
       end
-
-      it "marks corresponding assets as not draft when the attachment has been replaced" do
-        attachment.attachment_data.replace_with!(AttachmentData.create!(file: File.open(fixture_path.join("sample.docx"))))
-
-        expected_attribute_hash = {
-          "draft" => false,
-        }
-
-        attachment.attachment_data.assets.each do |asset|
-          AssetManager::AssetUpdater.expects(:call).with(asset.asset_manager_id, attachment.attachment_data, nil, expected_attribute_hash)
-        end
-
-        AssetManager::AttachmentUpdater.call(attachment.attachment_data, draft_status: true)
-      end
     end
 
     context "when the attachment's attachable is a draft and is access limited" do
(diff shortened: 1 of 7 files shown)


ryanb-gds:
Blocked pending merge of #8409

test/unit/app/services/asset_manager/attachment_updater_test.rb
```diff
@@ -208,7 +173,7 @@ class AssetManager::AttachmentUpdaterTest < ActiveSupport::TestCase
 
           AssetManager::AssetUpdater.expects(:call).never
 
-          AssetManager::AttachmentUpdater.call(attachment.attachment_data, redirect_url: true, draft_status: true)
+          AssetManager::AttachmentUpdater.call(attachment.attachment_data)
```
jyoung-gds, line 176: Should this also be calling `AssetManager::AttachmentUpdater.redirect`? The old test behaviour seems to suggest it but I'm not 100% sure

ryanb-gds: So that test was making a call that the application never does. There was nowhere in the app that passed those two flags at the same time, so it didn't seem worth having a test case that covered both

tunylund:
These changes make totale sense to me ❤️

lauraghiorghisor-tw:
I love what you have done here, it is taking the previous PR in an awesome direction. These updates were over-dried for no reason. And breaking them down in this fashion is making the updater test so pretty 🤩 Moving the methods to `AttachmentData` model works great 👍🏻 Parts of the replacement logic was already there, it makes all the sense.

tunylund (approved):
❤️

PR #8415, opened 2023-10-25
Prepare release for document collections.

farahTW:
This PR prepares document collections feature with design system for release.
**It does the following:**

- Remove get_layout method from the controller.
- Remove legacy controller test file.
- Update the feature test to render only design system layout.
- Update the unit test to render only design system layout.

**Following controllers are part of this release.**

- document_collection_groups_controller.rb
- document_collection_group_memberships_controller.rb
- document_collection_group_document_search_controller.rb

**Trello:**
https://trello.com/c/mx0YJ8HL/552-code-clean-up-for-document-collections


minhngocd (approved):
LGTM - some of the render refactorings can be tackled as part of code cleanup - will add them to those stories

PR #8424, opened 2023-10-26
Topical event cleanup

tunylund:
[Trello](https://trello.com/c/l7AHikFh/233-story-implement-non-legacy-urls-for-topical-event-logo)


db/schema.rb
```diff
@@ -77,7 +77,7 @@
     t.string "email"
     t.integer "call_for_evidence_response_form_id"
     t.text "postal_address"
-    t.index ["call_for_evidence_response_form_id"], name: "index_cons_participations_on_cons_response_form_id"
+    t.index ["call_for_evidence_response_form_id"], name: "index_cfes_participations_on_cfes_response_form_id"
```
tunylund, line 80: this index name doesn't belong in this pull request but the new value is more appropriate than the previous one. i'm thinking we should keep it

lauraghiorghisor-tw (approved):
👍🏻

PR #8416, opened 2023-10-25
Add edition_lead_image table and lead image association

davidgisbey:
## Background 

This is an alternate implementation to the one laid out in this PR - https://github.com/alphagov/whitehall/pull/8392 which added a foreign key to the editions table for a lead_image

## Description

Currently, a lead image is assigned for News Articles and Case Studies by adding an image to the edition. There's no way to change select a lead image. This means that if the publisher wants to assign a new lead image, they're forced to remove and 
reupload all the images.

We want to build in functionality that allows a user to select a lead image.

This draft PR does the following:

1. Adds a join table for images and editions for lead images (EditionLeadImage) using a has _one association.This has a unique constraint on edition_id to make sure there's only ever 1 lead_image
2. handles creating the edition_lead_image association when the `#create_draft` method is called (creates a new draft from a published edition)
3. handles creating the lead image association when the first image is created
3. handles creating the lead image association when a lead_ mage is deleted
4. handles deleting the `edition_lead_image` when the user toggles lead image visibility off/default image (default org image) for case studies
7. adds a migration to backfill lead_image_id
8. updated places that used `images.first` to use the new lead image association

Once i get some initial feedback, i'll break the above out into 3 PRs

1. will add the association to the db
2. will handle adding the callbacks which populate `lead_image_id` & add the backfill data migration

At this point i'll run the backfill on int, staging and prod

3. will update the code to use the new association in place of `edition.images.first`

## Out of scope

- Adding the ability for users to choose a lead image via the UI (we're still waiting or some finalised designs on this)
- Updating the validation which checks the lead_image is not being used in the markdown of the document. (i'll handle this in another PR)

## Trello card

https://trello.com/c/5vCiIxeQ/1064-lead-image-handling-backend-work

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #8429, opened 2023-10-30
Add edition_lead_image joins table and model associations

davidgisbey:
## Description

We're going to add in functionality for users to select a lead image. An approach has been agreed in this PR https://github.com/alphagov/whitehall/pull/8416.

The approach will be to create an  lead_image table, which contains and image_id and edition_id. This PR adds a migration for this table, adds the EditionLeadImage model and adds the relevant associations in the models edition and image models.

This should be a has_one as an edition should only ever be able to have one lead image. I've added a unique constraint to on edition_id to enforce this.

## Trello card

### DIFF
--- a/app/models/edition_lead_image.rb
+++ b/app/models/edition_lead_image.rb
@@ -1,0 +1,4 @@
+class EditionLeadImage < ApplicationRecord
+  belongs_to :edition
+  belongs_to :image
+end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -1,6 +1,7 @@
 class Image < ApplicationRecord
   belongs_to :image_data
   belongs_to :edition
+  has_one :edition_lead_image, dependent: :destroy
 
   validates :alt_text, presence: true, allow_blank: true, length: { maximum: 255 }, unless: :skip_main_validation?
   validates :image_data, presence: { message: "must be present" }
--- a/db/migrate/20231030075543_create_edition_lead_images.rb
+++ b/db/migrate/20231030075543_create_edition_lead_images.rb
@@ -1,0 +1,11 @@
+class CreateEditionLeadImages < ActiveRecord::Migration[7.0]
+  def change
+    create_table :edition_lead_images do |t|
+      t.integer "edition_id", foreign_key: true
+      t.integer "image_id", foreign_key: true
+      t.index %w[edition_id], name: "index_lead_image_on_edition_id", unique: true
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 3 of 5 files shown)


app/models/edition_lead_image.rb
```diff
@@ -0,0 +1,4 @@
+class EditionLeadImage < ApplicationRecord
+  belongs_to :edition
+  belongs_to :image
```
ryanb-gds, line 3: I would argue that the EditionLeadImage has one Image. But I wouldn't argue it very forcefully

app/models/image.rb
```diff
@@ -1,6 +1,7 @@
 class Image < ApplicationRecord
   belongs_to :image_data
   belongs_to :edition
+  has_one :edition_lead_image, dependent: :destroy
```
ryanb-gds, line 4: If you did agree EditionLeadImage has one Image, then the Image would belong to the EditionLeadImage

db/migrate/20231030075543_create_edition_lead_images.rb
```diff
@@ -0,0 +1,12 @@
+class CreateEditionLeadImages < ActiveRecord::Migration[7.0]
+  def change
+    create_table :edition_lead_images do |t|
+      t.integer "edition_id", foreign_key: true
+      t.integer "image_id", foreign_key: true
+      t.index %w[edition_id], name: "index_lead_image_on_edition_id", unique: true
+      t.index %w[image_id], name: "index_lead_image_on_image_id"
```
ryanb-gds, line 7: Do we definitely need this index? Is there a scenario where we'll need to look up a lead image from the image? Deletion, maybe, if we can't load the lead image from the edition side 🤔

ryanb-gds (approved):
Happy for it to merge as-is but I have added a couple of comments just to throw a few ideas at you

db/migrate/20231030075543_create_edition_lead_images.rb
davidgisbey: Yea agreed, creating a joins table does them by default, but you're right it feels unlikely it'll be needed. I'll 🪓 it

app/models/edition_lead_image.rb
davidgisbey: yea agreed, that makes more sense to me 👍

davidgisbey: We can do something like `  has_one :edition_lead_image, ->(i) { EditionLeadImage.where(image_id: i.id) }, dependent: :destroy` i guess?

ryanb-gds: Hmm maybe leave it as it is - I'd say it's not worth the hassle

PR #8433, opened 2023-10-30
Use lead image association throughout codebase

davidgisbey:
## Description

This moves over any code that uses `edition.images.first`  to determine the lead image to the new lead image association. The places this is found are:

- the case study presenter
- the uploaded images component
- the document summary page
- the lead_image model

It also updates the validation which stops publishers using the lead image within markdown. This was out of date and only covered the old syntax (`!!1`). I've updated this in 2 ways:

1. it now tests for the old syntax by using the index of the lead_image + 1. So if the lead image is the 3rd image it'll look for `!!3`
2. also matches against the new syntax (`[Image: {filename}]`

## Trello card

### DIFF
--- a/app/components/admin/edition_images/uploaded_images_component.rb
+++ b/app/components/admin/edition_images/uploaded_images_component.rb
@@ -3,12 +3,6 @@
 class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
   def initialize(edition:)
     @edition = edition
-    @lead_image = has_lead_image? ? @edition.images.first : nil
-    @document_images = @edition.images.drop(has_lead_image? ? 1 : 0)
-  end
-
-  def has_lead_image?
-    can_have_lead_image? && @edition.images.any?
   end
 
   def can_have_lead_image?
@@ -16,23 +10,30 @@ class Admin::EditionImages::UploadedImagesComponent < ViewComponent::Base
   end
 
   def lead_image
-    image_to_hash @lead_image, 0 if has_lead_image?
+    @lead_image ||= can_have_lead_image? ? @edition.lead_image : nil
+  end
+
+  def lead_image_hash
+    image_to_hash(lead_image, 0) if lead_image
   end
 
   def document_images
-    @document_images.map.with_index(1) { |image, index| image_to_hash image, index }
+    @document_images ||= (@edition.images - [lead_image].compact)
+                          .map.with_index(1) { |image, index| image_to_hash(image, index) }
   end
 
   def new_image_display_option
-    if @edition.image_display_option == "no_image"
-      return has_lead_image? ? "custom_image" : "organisation_image"
+    if @edition.image_display_option == "no_image" && @edition.images.present?
+      "custom_image"
+    elsif @edition.image_display_option == "no_image"
+      "organisation_image"
+    else
+      "no_image"
     end
-
-    "no_image"
   end
 
   def update_image_display_option_button_text
-    if has_lead_image?
+    if @edition.images.present?
       return "#{new_image_display_option == 'no_image' ? 'Hide' : 'Show'} lead image"
     end
 
@@ -56,7 +57,7 @@ private
   def image_to_hash(image, index)
     {
       url: image.url,
-      preview_alt_text: index.zero? ? "Lead image" : "Image #{index}",
+      preview_alt_text: lead_image == image ? "Lead image" : "Image #{index}",
       caption: image.caption.presence || "None",
       alt_text: image.alt_text.presence || "None",
       markdown: unique_names? ? "[Image: #{image.filename}]" : "!!#{index}",
--- a/app/models/edition/custom_lead_image.rb
+++ b/app/models/edition/custom_lead_image.rb
@@ -2,16 +2,7 @@ module Edition::CustomLeadImage
   extend ActiveSupport::Concern
 
   included do
-    validates :body,
-              format: {
-                without: /^!!1([^\d]|$)/,
-                message: "cannot have a reference to the first image in the text",
-                multiline: true,
-              }
-  end
-
-  def image_disallowed_in_body_text?(index)
-    index == 1
+    validate :body_does_not_contain_lead_image_markdown
   end
 
   def update_lead_image
@@ -36,5 +27,23 @@ private
     return if edition_lead_image.blank?
 
     edition_lead_image.destroy!
+  end
+
+  def body_does_not_contain_lead_image_markdown
+    return if lead_image.blank?
+
+    if body_contains_lead_image_filename_markdown? || body_contains_lead_image_index_markdown?
+      errors.add(:body, "cannot have a reference to the lead image in the text")
+    end
+  end
+
+  def body_contains_lead_image_filename_markdown?
+    body.match?(/\[Image:\s*#{Regexp.escape(lead_image.filename)}\s*\]/)
+  end
+
+  def body_contains_lead_image_index_markdown?
+    lead_image_index = images.find_index(lead_image)
+
+    body.match?(/^!!#{lead_image_index + 1}([^\d]|$)/)
   end
 end
--- a/test/unit/app/models/edition/custom_lead_image_test.rb
+++ b/test/unit/app/models/edition/custom_lead_image_test.rb
@@ -4,58 +4,43 @@ class Edition::CustomLeadImageTest < ActiveSupport::TestCase
   include ActionDispatch::TestProcess
 
   def edition_with_custom_lead_image(options = {})
-    build(:draft_news_article, options)
-  end
+    image1 = build(:image)
+    image2 = build(:image)
 
-  test "reports that the first image is not available for adding inline" do
-    assert edition_with_custom_lead_image.image_disallowed_in_body_text?(1)
-  end
-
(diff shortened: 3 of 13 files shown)


test/unit/app/models/edition/custom_lead_image_test.rb
```diff
@@ -4,28 +4,40 @@ class Edition::CustomLeadImageTest < ActiveSupport::TestCase
...
-    build(:draft_news_article, options)
-  end
+    edition = build(:draft_news_article, options)
+    image1 = build(:image)
+    image2 = build(:image)
 
-  test "reports that the first image is not available for adding inline" do
-    assert edition_with_custom_lead_image.image_disallowed_in_body_text?(1)
-  end
+    edition.stubs(:images).returns([image1, image2])
```
ryanb-gds, line 11: What's the advantage of stubbing over setting up the association by doing `edition.images = [image1, image2]`? Or even creating the images first and passing them to the factory?

test/unit/app/models/edition/lead_image_test.rb
```diff
@@ -2,41 +2,41 @@
 
 class Edition::LeadImageTest < ActiveSupport::TestCase
   test "should use placeholder image if none had been uploaded" do
-    model = stub("Target", images: [], lead_organisations: [], organisations: []).extend(Edition::LeadImage)
+    model = stub("Target", lead_image: nil, lead_organisations: [], organisations: []).extend(Edition::LeadImage)
```
ryanb-gds, line 5: I know you didn't write the original test, but do you know why this test uses stub everywhere instead of just building a model? I definitely think we should try to avoid that in favour of using factories with build stubbed if at all possible. Might be worth a separate PR for a refactor.

test/unit/app/models/speech_test.rb
```diff
@@ -145,8 +145,9 @@ class SpeechTest < ActiveSupport::TestCase
   end
 
   test "search_index includes default image_url if it has one" do
-    image = create(:image)
-    speech = create(:published_speech, images: [image])
+    image = build(:image)
+    speech = build(:published_speech, document: build(:document))
```
ryanb-gds, line 149: Does the below work? Again trying to avoid magic stubbing
```suggestion
    speech = build(:published_speech, document: build(:document), lead_image: image)
```

ryanb-gds (commented):
A few comments, but generally looking good. Possibly a bit heavy use of stubbing in the test, but that might just be a taste thing

app/models/edition/custom_lead_image.rb
```diff
@@ -4,14 +4,11 @@ module Edition::CustomLeadImage
   included do
     validates :body,
               format: {
-                without: /^!!1([^\d]|$)/,
-                message: "cannot have a reference to the first image in the text",
+                without: ->(edition) { /^!!#{edition.lead_image_index + 1}([^\d]|$)|\[Image: #{edition.lead_image.filename}\]/ },
```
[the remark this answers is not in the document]
ollietreend: Or alternatively, I wonder if it could be refactored into a custom validator method or class, rather than being defined as a 'format' rule?

That might make more 'room' to expand it into something more readable.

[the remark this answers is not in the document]
ollietreend: I should also point out that the `[Image:]` syntax allows any amount of whitespace (including none) before the filename.

For example, these are all equivalent:
- `[Image:example.jpg]`
- `[Image: example.jpg]`
- `[Image:    example.jpg]`

So it'd be good to have test coverage for this (and then the regex will need updating accordingly).

For what it's worth, [the regex used by Govspeak](https://github.com/alphagov/govspeak/blob/ae8136a9bba22ed55479974544fba27c2f98bd0a/lib/govspeak.rb#L356C24-L356C48) is:

```ruby
/^\[Image:\s*(.*?)\s*\]/
```

Where `\s` whitespace is allowed `*` times (zero or more).

Also, we should really make use of `Regex.escape()` for the image filename. Otherwise the `.` in the file extension will technically match _any character_, not just a literal `.` fullstop/period.

So something like:

```ruby
/\[Image: #{Regex.escape(edition.lead_image.filename)}\]/
```

test/unit/app/models/edition/lead_image_test.rb
ollietreend: Maybe I'm too stub-averse, but I sort of feel the same about the stubbing done here, too:

https://github.com/alphagov/whitehall/blob/0005c564cbef8e74760c6c948eaf5a0449cdffbd/test/unit/app/models/edition/custom_lead_image_test.rb#L36-L37

It feels a bit brittle. What if we chain something else on to that Active Record call? e.g. `where()`

But presumably this is the 'cost' of using `build_stubbed` rather than `create` in these sorts of scenarios – e.g. when querying associations via the Active Record interface, that would usually expect data to have been persisted.

davidgisbey: yea unfortunately build stubbed isn't smart enough to handle active chaining active record queries. Generally if it's just one call like this i stub it out and any more i just use create. I just know we're trying to reduce our db hits.

Happy to use a create instead if thats the consensus.

test/unit/app/models/speech_test.rb
davidgisbey: thanks have sorted

test/unit/app/models/edition/custom_lead_image_test.rb
davidgisbey: thanks have sorted

app/components/admin/edition_images/uploaded_images_component.rb
```diff
@@ -56,7 +59,7 @@ def all_image_asset_variants_uploaded?(image)
   def image_to_hash(image, index)
     {
       url: image.url,
-      preview_alt_text: index.zero? ? "Lead image" : "Image #{index}",
+      preview_alt_text: lead_image == image ? "Lead image" : "Image #{index}",
```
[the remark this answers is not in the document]
davidgisbey: Yea agreed. I sort of half refactored this with the intention to come back after. Let me do a bit of work on that now and i'll push it up a fresh commit.

app/models/edition/custom_lead_image.rb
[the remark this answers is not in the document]
davidgisbey: i've I've pulled the validation into it's own method now in this commit https://github.com/alphagov/whitehall/pull/8433/commits/71207de43df034120ee0f35ad9d657caad7269b lmn what you think

test/unit/app/models/edition/lead_image_test.rb
davidgisbey: Regarding your first question, no i'm not actually sure. Happy to go back and refactor it to just use a case study or another model, but maybe in a follow up PR?

ryanb-gds: It definitely feels like we're fighting ActiveRecord a bit trying to stub database reads. I do see ActiveRecord as sacrificing testability for convenience to a large degree, we're so coupled to the database it's very hard to test our logic without it. I'd probably take the hit on the create for Ollie's example in favour of a simpler test, personally

davidgisbey: okay have updated

app/models/edition/custom_lead_image.rb
```diff
@@ -37,4 +34,20 @@ def remove_lead_image
...
+      errors.add(:body, "cannot have a reference to the lead image in the text")
+    end
+  end
+
+  def body_contains_lead_image_filename_markdown?
+    body.match?(/\[Image:\s*#{Regexp.escape(lead_image.filename)}\s*\]/)
+  end
+
+  def body_contains_lead_image_index_markdown?
+    body.match?(/^!!#{lead_image_index + 1}([^\d]|$)/)
```
ollietreend, line 51: Could we try and see if this passes?

```suggestion
    body.match?(/!!#{lead_image_index + 1}[^\d]/)
```

app/models/edition/custom_lead_image.rb
```diff
@@ -26,6 +17,12 @@ def update_lead_image
     edition_lead_image.save!
   end
 
+  def lead_image_index
+    return if lead_image.blank?
+
+    images.find_index(lead_image)
+  end
```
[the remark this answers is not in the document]
davidgisbey: have sorted this

app/models/edition/custom_lead_image.rb
davidgisbey: this doesn't work unfortunately. After some discussion we've decided to go as is

app/components/admin/edition_images/uploaded_images_component.rb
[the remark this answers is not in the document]
davidgisbey: this was done in this PR https://github.com/alphagov/whitehall/pull/8440

app/models/edition/custom_lead_image.rb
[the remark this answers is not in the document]
davidgisbey: I think we're okay on this front now but let me know if that's not the case

ollietreend (approved):
Thanks for doing so much hard work on this ⭐

PR #8434, opened 2023-10-31
Fix bug with update_lead_image method

davidgisbey:
## Description

The 'and return if' isn't returning when the `image_display_option` is `no_image` or `organisation_image`.

This means that when an image is created and the lead_image isn't set the association is created.

It still wouldn't be pushed downstream to Publishing API as the presenter handles passing the image based on the `image_display_option`, but we shouldn't have lead images sitting in the db when there isn't a lead image.

This updates the method to explicitly call return within the conditional which resolves the issue.

## Trello card

### DIFF
--- a/app/models/edition/custom_lead_image.rb
+++ b/app/models/edition/custom_lead_image.rb
@@ -15,7 +15,10 @@ module Edition::CustomLeadImage
   end
 
   def update_lead_image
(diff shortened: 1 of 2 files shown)


beccapearce (approved):
nice one!

PR #8436, opened 2023-10-31
Publish organisation after logo has been uploaded

tunylund:
In order for the organisation logo to be visible in the organisation page the organisation needs to be published after the logo file has been uploaded. Only then can OrganisationPresenter get the logo url and send it to Publish Api

### DIFF
--- a/app/workers/asset_manager_create_asset_worker.rb
+++ b/app/workers/asset_manager_create_asset_worker.rb
@@ -35,12 +35,13 @@ private
   end
 
   def enqueue_downstream_service_updates(assetable_id, assetable_type, attachable_model_class, attachable_model_id)
+    if assetable_type == Organisation.to_s
+      publish_document(Organisation.find(assetable_id), Organisation.find(assetable_id))
+    end
+
     if assetable_type == FeaturedImageData.to_s
       assetable = FeaturedImageData.find(assetable_id)
-      featured_imageable = assetable.featured_imageable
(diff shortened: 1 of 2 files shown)


syed-ali-tw (approved):

app/workers/asset_manager_create_asset_worker.rb
```diff
@@ -53,6 +54,12 @@ def enqueue_downstream_service_updates(assetable_id, assetable_type, attachable_
     end
   end
 
+  def publish_document(document, assetable)
+    if assetable.all_asset_variants_uploaded? && document.class.ancestors.include?(PublishesToPublishingApi) && document.can_publish_to_publishing_api?
+      PublishingApiWorker.perform_async(document.class.to_s, document.id)
+    end
+  end
```
ollietreend, line 61: I'm guessing this will lead to two publishes in quick succession, in practice, right?

In other words, from the user's point of view:

1. User uploads new logo and clicks "Save"
2. Page gets immediately republished – with a broken logo
3. After upload to Asset Manager completes, the page gets republished again – with a working logo

Where the gap between Step 2 and Step 3 is hopefully pretty quick – i.e. seconds rather than minutes.

Is that right?

tunylund: the #2 is slightly wrong.
When whitehall publishes to publish-api, it won't publish the image if the image is not ready. Therefore there won't be a broken logo. There just won't be a logo

tunylund: The presenter for the image is filtering out unfinished images here
https://github.com/alphagov/whitehall/blob/main/app/presenters/publishing_api/organisation_presenter.rb#L157

PR #8443, opened 2023-11-02
Allow WWOrganisation to use FeaturedImageData

yuetylauiris:
This Change makes WorldwideOrganisation use Asset manager ids instead of legacy url path for WorldwideOrganisation default news images. It also shifts the responsiblity for Image configuration into an intermediate class FeaturedImageData which is used for all similar use cases in whitehall. This should make things more consistent across these classes.

[Trello Card](https://trello.com/c/kZOQQSoh/213-story-whitehall-should-push-edition-content-with-non-legacy-asset-urls-to-publish-api-defaultorganisationimagedata-featuredimage)

### DIFF
--- a/app/views/admin/worldwide_organisations/_form.html.erb
+++ b/app/views/admin/worldwide_organisations/_form.html.erb
@@ -62,20 +62,22 @@
     rows: 4,
   } %>
 
-  <%= render "components/single-image-upload", {
-    title: "Default news image",
-    name: "worldwide_organisation[default_news_image_attributes]",
-    id: "worldwide_organisation_default_news_image",
-    image_id: "worldwide_organisation_default_news_image_file",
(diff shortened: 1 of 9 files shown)


app/views/admin/worldwide_organisations/_form.html.erb
```diff
@@ -62,6 +62,7 @@
     rows: 4,
   } %>
 
+  <%= hidden_field_tag "worldwide_organisation[default_news_image_attributes][id]", worldwide_organisation.default_news_image.id %>
```
tunylund, line 65: a little bit more railsy would be to `<%= form.fields_for :default_news_image do |_image_fields| %>`

lauraghiorghisor-tw (approved):
👍🏻

PR #8445, opened 2023-11-02
Add ability for user to select lead image for case studies and news articles

davidgisbey:
## Description

This adds the ability to select a lead image for news articles and case studies. It does the following:

- adds a feature spec for updating the lead image
- adds a route for updating the lead image
- adds an edition_lead_images_controller with an update action
- updates the image_component to render a form/button for each non-lead
image which hits the new endpoint
- updates the copy in the lead image guidance
- updates the UI so the user can select between no lead image, the orgs default lead image or a custom lead image for case studies
- actually renders the orgs default image if one is present and the user is using it
- ensures that the draft is passed down to publishing API when the lead image is updated so it works on the preview

#### With default image and no uploaded images

image

#### With no image & no uploaded images

image

#### WIth no images and images uploaded 

image

#### With default image and images uploaded

image

#### With custom image

image

#### With no image uploaded it shows default org image 

image

#### With image uploaded

image

#### With multiple images uploaded 

image

## Video of implementation 

https://github.com/alphagov/whitehall/assets/42515961/e3296d63-fa8b-40fd-930b-5ab378fb3d8f

## Trello card

https://trello.com/c/QB4hHGwB/2164-lead-image-handling-selecting-a-new-lead-image

## Designs 

https://docs.google.com/document/d/14Xz67PJYrh3dKgtDaJXDM3MiWR3G8Rm4YThpggvQSG0/edit#heading=h.614wri5x7vxr

### DIFF
--- a/app/components/admin/edition_images/image_component.html.erb
+++ b/app/components/admin/edition_images/image_component.html.erb
@@ -16,9 +16,18 @@
     } %>
   </div>
 
-  <div class="app-view-edition-resource__actions govuk-grid-column-full govuk-button-group">
-    <% links.each do |link| %>
-      <%= link %>
+  <div class="app-view-edition-resource__actions govuk-grid-column-full govuk-button-group govuk-!-margin-top-4">
+    <%= link_to("Edit details", edit_admin_edition_image_path(edition, image), class: "govuk-link govuk-!-margin-top-2 app-view-edition-resource__actions__link") %>
+    <%= link_to("Delete image", confirm_destroy_admin_edition_image_path(edition, image), class: "govuk-link gem-link--destructive govuk-!-margin-top-2 app-view-edition-resource__actions__link") %>
+
+    <% if edition.can_have_custom_lead_image? %>
+      <%= form_with url: admin_edition_lead_image_path(edition, image), method: :patch do |form| %>
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Select as lead image",
+          secondary_solid: true,
+          margin_bottom: 4,
+        } %>
+      <% end %>
     <% end %>
   </div>
 </li>
--- a/app/components/admin/edition_images/image_component.rb
+++ b/app/components/admin/edition_images/image_component.rb
@@ -38,11 +38,4 @@ private
   def find_image_index
     edition.images.find_index(image)
   end
-
-  def links
-    links = []
-    links << link_to("Edit details", edit_admin_edition_image_path(@edition, image), class: "govuk-link")
-    links << link_to("Delete image", confirm_destroy_admin_edition_image_path(@edition, image), class: "govuk-link gem-link--destructive")
-    links
-  end
 end
--- a/app/components/admin/edition_images/lead_image_component.rb
+++ b/app/components/admin/edition_images/lead_image_component.rb
@@ -40,13 +40,7 @@ private
   end
 
   def new_image_display_option
-    @new_image_display_option ||= if image_display_option_is_no_image? && edition_has_images?
-                                    "custom_image"
-                                  elsif image_display_option_is_no_image?
-                                    "organisation_image"
-                                  else
-                                    "no_image"
-                                  end
+    @new_image_display_option ||= image_display_option_is_no_image? ? "organisation_image" : "no_image"
   end
 
   def image_display_option_is_no_image?
@@ -54,21 +48,7 @@ private
   end
 
   def update_image_display_option_button_text
-    return image_display_option_button_text_when_image_has_been_uploaded if edition_has_images?
-
-    image_display_option_button_text_when_no_images_uploaded
-  end
-
-  def image_display_option_button_text_when_image_has_been_uploaded
-    return "Hide lead image" if new_image_display_option_is_no_image?
-
-    "Show lead image"
-  end
-
-  def image_display_option_button_text_when_no_images_uploaded
-    return "Remove lead image" if new_image_display_option_is_no_image?
-
-    "Use default image"
+    new_image_display_option_is_no_image? ? "Remove lead image" : "Use default image"
   end
 
   def new_image_display_option_is_no_image?
--- a/app/components/admin/edition_images/lead_image_component.rb
+++ b/app/components/admin/edition_images/lead_image_component.rb
@@ -27,6 +27,10 @@ private
     edition.type == "CaseStudy"
   end
 
+  def news_article?
+    edition.type == "NewsArticle"
+  end
+
   def lead_image
     @lead_image ||= edition.lead_image
   end
@@ -37,6 +41,14 @@ private
 
   def alt_text
     lead_image.alt_text.presence || "None"
+  end
+
+  def show_default_lead_image?
+    if case_study?
+      edition.emphasised_organisation_default_image_available? && [nil, "organisation_image"].include?(edition.image_display_option)
+    elsif news_article?
+      edition.has_lead_image?
+    end
   end
 
   def new_image_display_option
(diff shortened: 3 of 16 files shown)


app/components/admin/edition_images/lead_image_component.rb
```diff
@@ -37,36 +39,28 @@ def alt_text
...
   end
 
-  def update_image_display_option_button_text
-    return image_display_option_button_text_when_image_has_been_uploaded if edition_has_images?
-
-    image_display_option_button_text_when_no_images_uploaded
+  def show_default_lead_image?
+    case_study? &&
+      default_organisation_image.present? &&
+      [nil, "organisation_image"].include?(edition.image_display_option)
```
ryanb-gds, line 51: I've seen code like this checking that the image display option is "x" or included in "[x,y]" in a few places. Feels like we might be missing an abstraction? Not sure exactly what it should be, maybe a value object for the image display option 🤔 probably not worth expending too much effort on if there's not an obvious refactor though

app/models/edition_lead_image.rb
```diff
@@ -1,4 +1,6 @@
 class EditionLeadImage < ApplicationRecord
   belongs_to :edition
   belongs_to :image
+
+  accepts_nested_attributes_for :edition
```
ryanb-gds, line 5: This is not a very useful comment, but I'd just like to say that this thing gives me the heebies. It's just so incredibly magical

features/edition-images.feature
```diff
@@ -29,7 +29,15 @@ Feature: Images tab on edit edition
...
-    Then I should see a button to show the lead image
+    Then I should see a button to select a lead image
+
+  Scenario: User selects a new lead image
+    And a draft case study with images with the alt text "First image uploaded" and "Second image uploaded" exists
+    When I visit the images tab of the document with images
+    And I make the image with alt text "First image uploaded" the lead image
+    Then I can see that the image with alt text "First image uploaded" is the lead image
+    And I make the image with alt text "Second image uploaded" the lead image
+    Then I can see that the image with alt text "Second image uploaded" is the lead image
```
ryanb-gds, line 40: Is there any value in covering the organisation image fallback as well? If indeed it isn't already covered

test/functional/admin/edition_lead_images_controller_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    .once
+
+    get :update, params: { edition_id: edition.id, id: image.id }
+
+    assert_equal image, edition.reload.lead_image
+    assert_redirected_to admin_edition_images_path(edition)
+    assert_equal "Lead image updated to minister-of-funk.960x640.jpg", flash[:notice]
+  end
+
+  test "PATCH :update does not update the lead image when the edition is invalid " do
```
ryanb-gds, line 23: Is this a useful test case? There's no way to make an edition invalid via the lead images controller is there?

ryanb-gds (commented):
A few questions for your consideration

app/components/admin/edition_images/image_component.rb
```diff
@@ -45,4 +45,8 @@ def links
     links << link_to("Delete image", confirm_destroy_admin_edition_image_path(@edition, image), class: "govuk-link gem-link--destructive")
     links
   end
+
+  def can_have_lead_image?
+    edition.can_have_custom_lead_image?
+  end
```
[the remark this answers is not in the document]
davidgisbey: yea this is a hangover form when it was called multiple times, but it should have been `can_have_custom_lead_image?`. I've removed it and just used `edition.can_have_custom_lead_image?` in the view since it's only called once.

app/components/admin/edition_images/image_component.html.erb
```diff
@@ -16,11 +16,27 @@
...
 
-  <div class="app-view-edition-resource__actions govuk-grid-column-full govuk-button-group">
-    <% links.each do |link| %>
-      <%= link %>
+  <% if can_have_lead_image? %>
+    <%= form_with url: admin_edition_lead_image_path(edition, image), method: :patch do |form| %>
+      <div class="app-view-edition-resource__actions govuk-grid-column-full govuk-button-group govuk-!-margin-top-4">
+        <% links.each do |link| %>
+          <%= link %>
+        <% end %>
```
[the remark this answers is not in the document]
davidgisbey: Yea i think it was just for styling purposes 


with: 

image


without:

image

I'll move it out the form and manually align the content

app/models/edition_lead_image.rb
davidgisbey: yup it's the worst. We can pull a lot of this out once we move to form objects, but until we have a way of pulling the validations together and not relying on the updater pub/sub service i think we're stuck with it 😞

test/functional/admin/edition_lead_images_controller_test.rb
davidgisbey: I think so just because the change note validation error triggers a lot, so in that situation we want to make sure the action redirects correctly and populates the flash with the error.

ryanb-gds: Ah so basically it's possible to arrive on this page with the edition in an invalid state because it is missing a change note?

davidgisbey: Yea unfortunately i think that's a reasonably common occurrence. The other option here could be to test the user tires to update the lead image to an image with the markdown in the body it handles that successfully. Both of those situations should trigger an unsuccessful save. I'll add another test for that.

features/edition-images.feature
davidgisbey: cool have a test for that [here](https://github.com/alphagov/whitehall/pull/8445/commits/06da2ff2029df4ef74ded21e50adf0f47a1a1e4e#diff-2919dd3689ccd52b537a04bcc33a7a7936eb2502ecaaf13578b8e748dc42dfa5L29-R35)

app/components/admin/edition_images/lead_image_component.rb
davidgisbey: Yea this is an interesting one. There's def some weird behaviour here that can probably be extracted into methods so the behaviour is more explicit. It's basically an enum that can be

`no_image`. `organisation_image`, `custom_image` or `nil`.

and combinations of them do specific things like:

`organisation_image || nil`  means the default image is passed down by the presenter to Publishing API and is rendered on the frontend.

But there's other pairings that make sense in other places. E.g. when you update the lead_image association after you toggle from `custom_image` to  `%w[no_image organisation_image]` you should delete the `edition_lead_image`.

It sort of feels like with the current implementation it should be set to `organisation_image` on create since that's the behaviour, but as it's a shared attr in the db across all editions that's not really an option.

Some other ideas would be:

1. lead image could be polymorphic and be an edition image, default_news image or speakers image (seems unlikely due to the variety of implementations of images)
2. lead_image_preferences could be it's own model and case studies could `have_one`

Tbh though it really feels like we really should give images a proper look and challenge some assumptions around implementation. Having a bunch of different subtypes editions behaving slightly differently and often having functionality being overwritten by the rendering application causes so many headaches when you work on it for what seems like pretty limited value.

It might well be useful to be able to toggle default image visibility for specific types of edition. But some having lead images, others not, some having the ability to toggle between a default image and others not feels a bit weird. Especially as we've had people request lead images for other content types.

Probs beyond the scope of this bit of work though.

ryanb-gds: Fully agree about trying to come to a logical solution that works across content types. One to take to Darren, I think. Thanks for your response, some really good thoughts there.

I did sort of wonder also about whether trying to select an image to be a lead image, we just had a completely separate field for lead image, and whether that would make a lot of our problems go away. Probably a bit late for that though and I suspect that's going to cascade changes all the way through publishing API and schemas and all the rest of it.

ryanb-gds (approved):
Good to go! Nice work, that was quite a tricky job 👍

PR #8446, opened 2023-11-02
Install FlipFlop gem

ryanb-gds:
We are installing the [FlipFlop](https://github.com/voormedia/flipflop) gem to enable us to use feature flags with a lower barrier to entry that the current process of using user permissions. The FlipFlop gem can store feature flag information inside the Whitehall database, meaning that we no longer have to create permissions inside of signon to deploy code without releasing it to users.

We have disabled the cookie strategy because our main use case at the moment is toggling features on and off in different environments, but we may consider using it in future if our needs change.

We have also only enabled the Flipflop user interface in local development environments. Developers will use the rake task provided by the library to toggle features in cloud environments.

This PR also switches the FlipFlop strategy to use an in-memory store for feature flag data during controller tests. I don't think we should be testing for feature flags lower than the controller level so this seems reasonable, but let me know if I have missed a possible case where we might need to do that.

Trello: https://trello.com/c/8f9xSARb


config/routes.rb
```diff
@@ -397,5 +397,7 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
 
   mount SidekiqGdsSsoMiddleware, at: "/sidekiq"
 
+  mount Flipflop::Engine => "/flipflop"
```
ollietreend, line 400: I was sort of expecting to see the development environment constraint defined here, as is done for the `GovukPublishingComponents::Engine` in the line below. If it doesn't functionally change anything, it might be nicer to keep this logic in the same place rather than defining it in `application.rb`.

e.g. this

```suggestion
  mount Flipflop::Engine => "/flipflop" if Rails.env.development?
```

and then remove the `config.flipflop.dashboard_access_filter` before filter from `application.rb`.

beccapearce (approved):
This looks good to me!

ollietreend (commented):
This is looking good. One minor comment from me.

Thanks for adding a doc for this too ⭐

config/routes.rb
ryanb-gds: Good shout

ryanb-gds: Actually this isn't going to work because that config option is required by the FlipFlop engine. I could return true from it and move the environment check into the routes file but that doesn't seem a massive improvement.

ollietreend: That seems fair. My comment definitely wasn't a _strong_ opinion, so all good.

PR #8448, opened 2023-11-02
Refactor body_does_not_contain_lead_image validator

ollietreend:
This commit refactors the `body_does_not_contain_lead_image` custom validation method to reduce its coupling to Govspeak.

Right now it uses regular expressions to check if the lead image has been embedded into the document body. This relies upon the validator knowing the specific Govspeak syntax that would be used to embed the image (of which there are two different variants).

This refactor takes a more 'black box' approach by rendering the Govspeak as HTML, and then inspecting the output to see if it contains the lead image. This way the validator doesn't need to parse the Govspeak syntax directly – making it more robust against future changes to the syntax.

I imagine this new approach is slightly less performant. It now renders and inspects the resultant HTML, rather than simply matching a regex against the raw Govspeak. However I expect this will have negligible overall impact on performance, so would be a net benefit given the improvements to code readability and cleanliness.

---

### DIFF
--- a/app/models/edition/custom_lead_image.rb
+++ b/app/models/edition/custom_lead_image.rb
@@ -2,7 +2,7 @@ module Edition::CustomLeadImage
   extend ActiveSupport::Concern
 
   included do
-    validate :body_does_not_contain_lead_image_markdown
+    validate :body_does_not_contain_lead_image
   end
 
   def update_lead_image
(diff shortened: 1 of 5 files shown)


app/models/edition/custom_lead_image.rb
```diff
@@ -33,21 +33,14 @@ def remove_lead_image
     edition_lead_image.destroy!
   end
 
-  def body_does_not_contain_lead_image_markdown
-    return if lead_image.blank?
+  def body_does_not_contain_lead_image
+    return if lead_image.blank? || images.none?
```
davidgisbey, line 37: nice!

test/unit/app/models/edition/custom_lead_image_test.rb
```diff
@@ -3,11 +3,18 @@
 class Edition::CustomLeadImageTest < ActiveSupport::TestCase
   include ActionDispatch::TestProcess
 
+  def stubbed_image(name)
+    build(:image).tap do |image|
+      image.stubs(:filename).returns(name)
+      image.stubs(:url).returns("http://example.com/#{name}")
```
davidgisbey, line 9: Do we need to stub this out? Can we not just create an image with the filename we want?

davidgisbey (approved):
looks good to me. Seems like a better approach to test the output itself to me 👍

test/unit/app/models/edition/custom_lead_image_test.rb
ollietreend: I tried avoiding a stub in the first place, but it looks like the [Image](https://github.com/alphagov/whitehall/blob/ff071c7f2dfe5b9e949216d6a9fee08bfdc894f4/test/factories/images.rb) and [ImageData](https://github.com/alphagov/whitehall/blob/ff071c7f2dfe5b9e949216d6a9fee08bfdc894f4/test/factories/image_data.rb) factories have been changed with the recent work done by the Assets team.

From what I can tell, the factory can only create an SVG image or a JPEG image. And the filename isn't configurable – it seems like the JPEG is hardcoded as `minister-of-funk.960x640.jpg` and the SVG is `test-svg.svg`.

Really I want to create two 'normal' images (e.g. not SVGs, because we know they sometimes get treated differently) that have different filenames.

So I figured this stub was an easy way to avoid changing how that factory works. (Perhaps that's a bit lazy of me!?)

davidgisbey: Yea i've seen quite a lot of stuff that's not used the factories which probably isn't ideal (if you search for `build(:image_data)` you'll see what i mean. This seems fine to me though.

PR #8460, opened 2023-11-06
Create a migration that creates Assets for each Topical Event Featuring Image data item

tunylund:
in preparation of making asset id references from the database instead of carrierwave configuration


lauraghiorghisor-tw (approved):
👍🏻

PR #8467, opened 2023-11-07
Use summary card for translations section on document summary page

davidgisbey:
## Description

This ports the translations section of the document summary page to use the summary card component. It's behind the `document_hub` feature flag.

I've chosen not to duplicate these view_tests https://github.com/alphagov/whitehall/blob/e958a8cb68e5178f9a3d4a0447381a8299ce26b5/test/functional/admin/generic_editions_controller_tests/translation_test.rb#L3 

as the view component now tests this functionality. I think this file can be removed once we remove the feature flag. Interested to see what approach people think we should use in this situation though.

Currently, if we merge this the tests will continue to pass until we remove the `document_hub` feature flag, at which point they'll fail. Is that okay? Should we indicate/record somewhere that this file should be removed at that point? If this is a common occurrence, then many tests will fail at the same when the flag is removed. Is that a problem? 

### With no translations and edition is pre-publication

image

### With translation and edition is pre-publication

image

### With no translations and edition has been published

image

### With translations and edition has been published

image

## Trello card

https://trello.com/c/2NUrOcjY/1835-translations-summary-card-on-summary-page-like-for-like

### DIFF
--- a/app/views/admin/editions/show/_main.html.erb
+++ b/app/views/admin/editions/show/_main.html.erb
@@ -249,7 +249,10 @@
       <% end %>
     </section>
   <% end %>
-  <% if @edition.translatable? %>
+
+  <% if Flipflop.document_hub? %>
+    <%= render Admin::Editions::Show::TranslationsSummaryCardComponent.new(edition: @edition) %>
+  <% elsif @edition.translatable? %>
     <section class="app-view-summary__section">
       <%= render "govuk_publishing_components/components/heading", {
         text: "Translations",
--- a/app/components/admin/editions/show/translations_summary_card_component.html.erb
+++ b/app/components/admin/editions/show/translations_summary_card_component.html.erb
@@ -1,4 +1,5 @@
 <%= render "components/summary_card_component", {
   title: "Translations",
   summary_card_actions: summary_card_actions,
+  rows: rows,
 } %>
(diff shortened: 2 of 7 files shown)


davidgisbey:
re-commenting for visibility 

I've chosen not to duplicate these view_tests https://github.com/alphagov/whitehall/blob/e958a8cb68e5178f9a3d4a0447381a8299ce26b5/test/functional/admin/generic_editions_controller_tests/translation_test.rb#L3 

as the view component now tests this functionality. I think this file can be removed once we remove the feature flag. Interested to see what approach people think we should use in this situation though.

Currently, if we merge this the tests will continue to pass until we remove the `document_hub` feature flag, at which point they'll fail. Is that okay? Should we indicate/record somewhere that this file should be removed at that point? If this is a common occurrence, then many tests will fail at the same when the flag is removed. Is that a problem?

app/views/admin/editions/show/_main.html.erb
```diff
@@ -249,7 +249,10 @@
       <% end %>
     </section>
   <% end %>
-  <% if @edition.translatable? %>
+
+  <% if Flipflop.document_hub? %>
+    <%= render Admin::Editions::Show::TranslationsSummaryCardComponent.new(edition: @edition) %>
+  <% elsif @edition.translatable? %>
```
ryanb-gds, line 255: Personally I'd do the `translatable?` check outside of the feature flag check and then not need to do it again inside the view component, but I can see why you've gone for this option with a view towards removing the feature flag

ryanb-gds (approved):
Looks good. In response to your question about the controller test failing once the feature flag is removed, I think that's ideal because it will remind us to delete the test once it is no longer needed, if I understand you correctly

PR #8480, opened 2023-11-09
Add Summary Card List Whitehall Publishing Component

davidgisbey:
## Description 

This adds a new component which functions almost identically to the summary card component with one notable exception. It uses an ordered list instead of a descriptive list.

This is needed for the document summary page for images, tags and specialist tags.

The summary card component requires key value pairs for summary list items. For this we simply need a value rendered in an ordered list.

## Screenshot

image

## Trello card

https://trello.com/c/n6xpHLB5/1834-images-summary-card-on-summary-page-like-for-like

### DIFF
--- a/app/views/components/_summary_card_list.html.erb
+++ b/app/views/components/_summary_card_list.html.erb
@@ -1,0 +1,46 @@
+<%
+  id ||= nil
+  data_attributes ||= {}
+  summary_card_actions ||= []
+  rows ||=[]
+%>
+
+<%= tag.div class: "app-c-summary-card-list", id: id, data: data_attributes do %>
+  <div class="govuk-summary-card">
+    <div class="govuk-summary-card__title-wrapper">
+      <h2 class="govuk-summary-card__title"><%= title %></h2>
+      <ul class="govuk-summary-card__actions">
+        <% summary_card_actions.each do |action| %>
+          <li class="govuk-summary-card__action">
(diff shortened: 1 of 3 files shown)


app/views/components/_summary_card_list.html.erb
```diff
@@ -0,0 +1,46 @@
...
+          <li class="govuk-summary-card__action">
+            <%= link_to sanitize(action[:label] + tag.span(" #{title}", class: "govuk-visually-hidden")), action[:href], class: "govuk-link govuk-link--no-visited-state #{"gem-link--destructive govuk-!-font-weight-bold" if action[:destructive]}".strip %>
+          </li>
+        <% end %>
+      </ul>
+    </div>
+
+    <% if rows.present? %>
+      <div class="govuk-summary-card__content">
+        <ol class="gem-c-list govuk-summary-list">
```
ryanb-gds, line 23: Is ordered list appropriate for tags? Sorry to potentially complicate matters a bit!

ryanb-gds (approved):
Looks good. I'm sorta tempted to make components out of the title wrapper and the row actions to reduce the duplication between this and the original summary card, but we can consider that in a separate PR. I appreciate that may not be consistent with publishing component styles anyway

app/views/components/_summary_card_list.html.erb
dnkrj: I've taken a deeper look into this and I think that we might actually be best using `ul` for all of these.

WCAG differentiates between `ul` and `ol` not based on whether the data is ordered but on whether the content "[has the visual appearance of a numbered list](https://www.w3.org/TR/WCAG20-TECHS/H48.html)".

Given that we don't ever display numbers for these I think that a `ul` would actually be appropriate for everything. Happy to put together a small follow up PR, sorry for saying the opposite earlier!

PR #8481, opened 2023-11-09
Upgrade carrierwave from 2.x to 3.x

tunylund:
Inorder to complete the upgrade, we need to:

i> Use TransactionAwareClient for AssetManagerCreateAssetWorker. Because
   Carrierwave has changed the way it hooks with the related ActiveRecord models.
   store-method used to be called on an `after_commit` callback, but it is now invoked
   on the `after_save` callback. Important difference here is that any Database query
   might not have finished yet when the worker runs. Therefore the Worker might not
   have all the necessary information available. In this case we saw that Assetable
   model was not yet available from the database even though saving was successful for
   the ActiveRecord model.

ii> Add missing expectation to failing tests. We're unsure why the tests were passing before.

iii> Carrierwave has a new de-duplication feature for filenames which makes it hard to
     guess resulting filename. This happended with the test
     'call_for_evidence_presenter_test.rb -test "ways to respond"'

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,10 +4,11 @@ gem "rails", "7.0.8"
 
 gem "activemodel-serializers-xml"
 gem "addressable"
+gem "after_commit_everywhere"
 gem "babosa"
 gem "bootsnap", require: false
 gem "bootstrap-kaminari-views"
-gem "carrierwave", "< 3" # pin at v2 to avoid breaking changes
+gem "carrierwave"
 gem "carrierwave-i18n"
 gem "chronic"
 gem "dalli"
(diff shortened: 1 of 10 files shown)


lauraghiorghisor-tw (approved):
Difficult to understand why some of those tests broke. I would have expected the fix to make it possible to revert the changes. 
Otherwise 👍🏻

tunylund:
Yeah I tried reverting them but the tests failed for decent reasons. `auth_bypass_id` related test failed because of timing issue with database transactions. `reload` in the other test was a bit of a mystery but probably the same reason.

PR #8489, opened 2023-11-10
Fix selenium based tests failing in CI for an unknown reason

tunylund:
For an unknown reason selenium driver based image-tests began to fail in the CI workflow when executed in Github. Running the tests from a separate feature-file seems to fix the issue


sengi (approved):
Very strange. I guess if it works, it works. 🤷 (and it's a harmless workaround)

PR #8498, opened 2023-11-13
Fix test file naming as it makes CI fail the test run

tunylund:
When tests are run in parallel, no 2 tests can have the same name.


lauraghiorghisor-tw (approved):
👍🏻

PR #8503, opened 2023-11-14
Cleanup regarding Orphan Document fix for Document Collections

minhngocd:
The fix was pushed out to production so a lot of the refactorings and cleanups had to be pushed to another PR. This is to address some comments that was made on the previous PR, and general refactorings.
- Created new partial for warning banner.
- Moved logic for generating document collection group membership title and links into the helper
- The helper is now responsible for deciding on title and links, smart enough to make the decision based on the membership being whitehall or non-whitehall, with edition or no edition etc.

https://github.com/alphagov/whitehall/pull/8484
https://trello.com/c/QHOvEjag/565-bug-unpublishing-deleting-a-document-after-its-been-added-to-a-document-collection-group-creates-an-error

### DIFF
--- a/app/helpers/admin/document_collection_group_memberships_helper.rb
+++ b/app/helpers/admin/document_collection_group_memberships_helper.rb
@@ -1,9 +1,26 @@
 module Admin::DocumentCollectionGroupMembershipsHelper
+  UNAVAILABLE_DOCUMENT_TITLE = "Unavailable Document".freeze
+
   def document_collection_group_member_title(membership)
     return membership.non_whitehall_link.title if membership.non_whitehall_link
+    return unavailable_document_title_tag unless membership.document&.latest_edition
 
     membership.document.latest_edition.title
   end
+
+  def document_collection_group_member_links(collection, group, membership)
+    links = [remove_link(collection, group, membership)]
+    links.prepend(view_link(membership)) unless document_collection_group_member_unavailable?(membership)
+    sanitize(links.join)
+  end
+
+  def unavailable_document_count(memberships)
+    memberships.count do |membership|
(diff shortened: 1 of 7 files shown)


app/helpers/admin/document_collection_group_memberships_helper.rb
```diff
@@ -1,14 +1,17 @@
 module Admin::DocumentCollectionGroupMembershipsHelper
+  UNAVAILABLE_DOCUMENT_TITLE = "Unavailable Document".freeze
+
   def document_collection_group_member_title(membership)
     return membership.non_whitehall_link.title if membership.non_whitehall_link
+    return UNAVAILABLE_DOCUMENT_TITLE unless membership.document&.latest_edition
```
ollietreend, line 6: Does the "Unavailable Document" label need to retain its **bold** styling? Or has that requirement been dropped?

```suggestion
    return tag.span(UNAVAILABLE_DOCUMENT_TITLE, class: "govuk-!-font-weight-bold") unless membership.document&.latest_edition
```

ollietreend (approved):
Nice refactor 😍 this certainly feels tidier.

Thank you for following up with these changes! 🙇🏻 

I've just added one comment (not a blocker) which affects the design.

davidtrussler (commented):
Apart from the comment from @ollietreend about the bold styling of "Unavailable document", this looks good to me as well. As far as I am aware there was no requirement to change the styling here (but maybe there was).

app/helpers/admin/document_collection_group_memberships_helper.rb
minhngocd: There wasn't!! I completely missed - thank you! will change before merge

PR #8504, opened 2023-11-14
Update New Document Page to show document type option based on user permission & organisation.

farahTW:
This PR updates the New Document page for header navigation to show document type options based on user permission and organisation.
**It does the following:**

- Updates New Document Page controller and View to show and hide 'Fatality Notice' option based on User permission and Organisation.
- Adds tests to NewDocumentControllerTest to cover the scenarios.
- Adds tests to MoreControllerTest to cover the scenario for user permission and organisation.

**Screen shots:**
**Case 1:** User permission is GDS editor and Organisation is GDS: Fatality Notice is shown

image

**Case 2:** User Organisation in Ministry Of Defence irrespective of permission
image

**Case 3: When organisation is Cabinet Minister and no GDS Editor permission**

image

**Trello:**
https://trello.com/c/eUrMNKbx/629-update-links-based-on-your-organization

### DIFF
--- a/app/controllers/admin/new_document_controller.rb
+++ b/app/controllers/admin/new_document_controller.rb
@@ -6,22 +6,9 @@ class Admin::NewDocumentController < Admin::BaseController
   def index; end
 
   def new_document_options_redirect
-    redirect_options = {
-      "call-for-evidence": new_admin_call_for_evidence_path,
-      "case-study": new_admin_case_study_path,
-      "consultation": new_admin_consultation_path,
-      "detailed-guide": new_admin_detailed_guide_path,
-      "document-collection": new_admin_document_collection_path,
-      "fatality-notice": new_admin_fatality_notice_path,
-      "news-article": new_admin_news_article_path,
-      "publication": new_admin_publication_path,
-      "speech": new_admin_speech_path,
-      "statistical-data-set": new_admin_statistical_data_set_path,
-    }
-
     if params[:new_document_options].present?
-      selected_option = params.require(:new_document_options).to_sym
-      redirect_to redirect_options[selected_option]
+      new_document_type = params.require(:new_document_options).to_sym
+      redirect_to redirect_path(new_document_type)
     else
       redirect_to admin_new_document_path, alert: "Please select a new document option"
     end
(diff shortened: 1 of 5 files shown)


nnagewad:
Looking good so far. But will Fatality Notice, Field of Operation, and Sitewide Settings be hidden depending on your organization and permission?

farahTW:
> Looking good so far. But will Fatality Notice, Field of Operation, and Sitewide Settings be hidden depending on your organization and permission?

Yes It will show and hide based on organisation and user permission.
For MorePage it is already existing logic same as in legacy code.
For NewDocument Page we have implemented same as legacy logic.

farahTW:
**1.More Page screen shot when user's organisation is Ministry Of Defence:**
image

2.More Page screen shot when user is GDS Editor and Organisation is GDS:

davidtrussler:
> > Looking good so far. But will Fatality Notice, Field of Operation, and Sitewide Settings be hidden depending on your organization and permission?
> 
> Yes It will show and hide based on organisation and user permission. For MorePage it is already existing logic same as in legacy code. For NewDocument Page we have implemented same as legacy logic.

Maybe I'm missing something but 
- I can't see where the logic is in the "New Document" page to exclude items based on organisation. 
- I don't think there is any logic at all along these lines in the "More" page

Also I'm not sure the intended logic is correct: shouldn't we be excluding some content on the basis of below?
- on the "New Document" page if the user is from an organisation other than GDS/MOD
- on the "More" page if the user is from an organisation other than GDS

farahTW:
> > > Looking good so far. But will Fatality Notice, Field of Operation, and Sitewide Settings be hidden depending on your organization and permission?
> > 
> > 
> > Yes It will show and hide based on organisation and user permission. For MorePage it is already existing logic same as in legacy code. For NewDocument Page we have implemented same as legacy logic.
> 
> Maybe I'm missing something but
> 
> * I can't see where the logic is in the "New Document" page to exclude items based on organisation.
> * I don't think there is any logic at all along these lines in the "More" page
> 
> Also I'm not sure the intended logic is correct: shouldn't we be excluding some content on the basis of below?
> 
> * on the "New Document" page if the user is from an organisation other than GDS/MOD
> * on the "More" page if the user is from an organisation other than GDS

@davidtrussler  it is the same logic as was in legacy implementation, we have added new document helper that handles the criteria

farahTW:
> > Looks good! just one thing to check with @nnagewad - the existing logic also removes the "Get involved" link (as well as "Fields of operation" and "Sitewide settings") on the "More" page for organisations other than GDS/MOD. Is this expected behaviour?
> 
> Good shout David. Looks like something I've missed as well.
> 
> @farahTW it looks like Get Involved is only present if your permission is set to GDS Editor.

Yes @nnagewad

davidtrussler:
> > Looks good! just one thing to check with @nnagewad - the existing logic also removes the "Get involved" link (as well as "Fields of operation" and "Sitewide settings") on the "More" page for organisations other than GDS/MOD. Is this expected behaviour?
> 
> Good shout David. Looks like something I've missed as well.
> 
> @farahTW it looks like Get Involved is only present if your permission is set to GDS Editor.

That looks like what we have - so will approve 👍

davidtrussler (approved):
Looks great, nice work! 👍

PR #8509, opened 2023-11-15
Ensure that SVGs cannot be selected as a lead image for editions that can have custom lead images

davidgisbey:
## Description

At the moment, an SVG is set to the lead image for news articles and case studies in the following contexts.

case studies:

- user selects a SVG as the lead image via the images index page

News articles 

- users first image they upload is a SVG
- user selects SVG as the lead image

I've added a data migration to clean up the 5 editions which have SVGs as lead editions. I'll run this once this PR is merged

### SVGs are set as lead image on creation for news articles 

image

### SVGs can be selected as lead image from Images index page

image

### SVGs blow up when they are selected as lead image

image

### No ability so select SVG as lead image

image

### Not set as lead image when current lead image is deleted

image

### Not set as lead image when first image uploaded is SVG

image

## Trello card

https://trello.com/c/To2f26k4/2166-fix-svg-lead-image-behaviour

### DIFF
--- a/app/models/edition/custom_lead_image.rb
+++ b/app/models/edition/custom_lead_image.rb
@@ -11,9 +11,9 @@ module Edition::CustomLeadImage
       return
     end
 
-    return if lead_image.present? || images.blank?
+    return if lead_image.present? || non_svg_images.blank?
 
-    edition_lead_image = build_edition_lead_image(image: oldest_image)
+    edition_lead_image = build_edition_lead_image(image: oldest_non_svg_image)
     edition_lead_image.save!
   end
 
@@ -23,8 +23,12 @@ module Edition::CustomLeadImage
 
 private
 
-  def oldest_image
-    images.order(:created_at, :id).first
+  def non_svg_images
+    @non_svg_images ||= images.includes(:image_data).reject(&:svg?)
+  end
+
+  def oldest_non_svg_image
+    non_svg_images.min_by { |image| [image.created_at, image.id] }
   end
 
   def remove_lead_image
--- a/app/models/edition/custom_lead_image.rb
+++ b/app/models/edition/custom_lead_image.rb
@@ -11,10 +11,14 @@ module Edition::CustomLeadImage
       return
     end
 
-    return if lead_image.present? || non_svg_images.blank?
+    return if lead_image.present? || images.blank?
 
-    edition_lead_image = build_edition_lead_image(image: oldest_non_svg_image)
-    edition_lead_image.save!
+    image = oldest_non_svg_image
+
+    if image
+      edition_lead_image = build_edition_lead_image(image:)
+      edition_lead_image.save!
+    end
   end
 
   def non_lead_images
@@ -23,12 +27,8 @@ module Edition::CustomLeadImage
 
 private
 
-  def non_svg_images
-    @non_svg_images ||= images.includes(:image_data).reject(&:svg?)
-  end
-
   def oldest_non_svg_image
-    non_svg_images.min_by { |image| [image.created_at, image.id] }
+    images.includes(:image_data).detect { |image| !image.svg? }
   end
 
   def remove_lead_image
--- a/db/data_migration/20231115101909_remove_svg_lead_images.rb
+++ b/db/data_migration/20231115101909_remove_svg_lead_images.rb
@@ -1,0 +1,7 @@
+edition_lead_images = EditionLeadImage.joins(image: :image_data).where("carrierwave_image LIKE '%.svg'")
+
+edition_lead_images.each do |edition_lead_image|
+  edition = edition_lead_image
+  edition_lead_image.destroy!
+  edition.update_lead_image
+end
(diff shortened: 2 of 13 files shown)


app/models/edition/custom_lead_image.rb
```diff
@@ -1,6 +1,8 @@
 module Edition::CustomLeadImage
   extend ActiveSupport::Concern
 
+  SVG_CONTENT_TYPE = "image/svg+xml".freeze
```
ryanb-gds, line 4: Feels to me like this constant should live on the Image class, given it doesn't really apply to lead images?

app/models/edition/custom_lead_image.rb
```diff
@@ -23,8 +25,12 @@ def non_lead_images
 
 private
 
-  def oldest_image
-    images.order(:created_at, :id).first
+  def non_svg_images
+    images.reject { |image| image.content_type == SVG_CONTENT_TYPE }
```
ryanb-gds, line 29: Would it make any sense to have this be a scope on the Image model and then use the scope in this class?

app/models/edition/custom_lead_image.rb
```diff
@@ -23,8 +25,12 @@ def non_lead_images
...
+  end
+
+  def oldest_non_svg_image
+    non_svg_images.min_by { |image| [image.created_at, image.id] }
```
ryanb-gds, line 33: Similarly, it might be interesting to consider moving some of these types of relationship query methods into [active record extensions](https://guides.rubyonrails.org/association_basics.html#association-extensions). That's probably out of the scope of this PR though as I don't believe I've seen AR extensions used at all in Whitehall

db/data_migration/20231115101909_remove_svg_lead_images.rb
```diff
@@ -0,0 +1,7 @@
+edition_lead_images = EditionLeadImage.joins(image: :image_data).where("carrierwave_image LIKE '%.svg'")
+
+edition_lead_images.each do |edition_lead_image|
+  edition = edition_lead_image
+  edition_lead_image.destroy!
+  edition.update_lead_image
+end
```
ryanb-gds, line 7: Well done on tracking these down! 👏

test/unit/app/models/edition/custom_lead_image_test.rb
```diff
@@ -35,14 +35,15 @@ def body_text_valid(body)
     assert body_text_valid("foo\[Image: non_lead_image.jpg]\nbar")
   end
 
-  test "#update_lead_image updates the lead_image association to the oldest image" do
-    image1 = build(:image)
+  test "#update_lead_image updates the lead_image association to the oldest non-svg image" do
+    svg_image_data = build(:image_data, file: File.open(Rails.root.join("test/fixtures/images/test-svg.svg")))
+    image1 = build(:image, image_data: svg_image_data)
```
ryanb-gds, line 40: I know the original test didn't do it, but do you think it might be worth setting the created timestamp on these images just to be really explicit about how this should work? I don't like relying on the implicit order too much.

ryanb-gds (approved):
A few thoughts, but nothing that ought to stop this merging if you're happy with it. Nice job :+1:

app/models/edition/custom_lead_image.rb
davidgisbey: Yea i did think about popping it there, but it's currently only used in this context so i wasn't sure if it was good to keep it closer to the source type thing. Happy to move it over though.

ryanb-gds: The other approach if you think it is makes sense to keep it on the lead image might be to change it to `INVALID_CONTENT_TYPES` or maybe allow list rather than ban list. But then you've got to deal with a set, which is mildly less ergonomic and potentially YAGNI. If we're not sure though then might as well leave it I would say, no point changing things for questionable benefit

app/models/edition/custom_lead_image.rb
davidgisbey: Yea what makes this tricky is that the content type lives of the file object generated via carrierwave. We could add a scope like 

```
scope :svg, -> { joins(:image_data).where("carrierwave_image LIKE '%.svg'") }
```

but it feels a bit gross. Wdyt?

ryanb-gds: Ah ok. Are we in any danger of N+1 queries here, in that case, or is the `image_data` table always eager loaded?

davidgisbey: I might just be a bit clueless on carrierwave 😂 but i think it's more that we don't have the content type persisted to the database. Carrierwave grabs the asset from asset manager and provides us access to the content_type as part of the API.

ryanb-gds: I'm having a proper dig into this, bear with me...

app/models/edition/custom_lead_image.rb
davidgisbey: I've ended up moving it to the images model and adding a `sgv?` method

PR #8513, opened 2023-11-15
This commit removes the previous errors forbidden page found in the a…

RodneyJohnsonGDS:
…dmin/editions directory and replaces it with the correct fobidden page view found in admin/erros/forbidden

The previous 403 forbidden error page looks like this:

The new 403 forbidden error page looks like this:

whitehall-admin dev gov uk_government_admin_fatalities_new (1)


nnagewad:
Very nice!

minhngocd:
Ideally i feel like we can also get rid of that `forbidden!` method, and replace with an error 403, which should automatically render the correct template. But this refactoring can be done at a later stage too.

minhngocd (approved):

PR #8517, opened 2023-11-16
Remove unused files for pagination

farahTW:
This PR analyses the usage of files inside kaminari folder for pagination and removes unused ones.

Following files are removed:

- app/views/kaminari/_next_page.html.erb:2
- app/views/kaminari/_prev_page.html.erb:2
- app/views/kaminari/_paginator.html.erb

**Trello:**
https://trello.com/c/fzZMMDAG/623-investigate-whether-pagination-overrides-need-to-exist


minhngocd:
Quick question @farahTW on how you went about finding out whether it was used or not? These are overrides of the Kaminari gem views, so it might not be explicitly called within our views, but they may still be used. 
And what led to the decision of removing these, but not the ones under history, or govuk_paginator?

minhngocd:
I had a look again, I see where you might be coming from. 

seems like we're mostly using the `govuk_paginator` theme and the `history` theme. The only usage of the default bootstrap 'theme' is in legacy_search_results. Might be a wider question of if we can remove the bootstrap kaminari view dependency. But for this one, maybe we merge the removal of legacy views and this should be fine.

mtaylorgds (approved):

PR #8518, opened 2023-11-16
cleanup take_part_page

syed-ali-tw:
Clean up TakePartPage and remove DefaultNewsImage

i> remove carrierwave_image column from take_part_page which is a side-effect from : https://github.com/alphagov/whitehall/pull/8491 
ii> delete default_news_organisation_image_data.rb which is no longer required as its been replace by the use of FeaturedImageData 
iii>bring back the required text for image

[trello](https://trello.com/c/YEURlcwg/258-clean-up-takepartpage)


lauraghiorghisor-tw (approved):
👍🏻

PR #8522, opened 2023-11-17
625 admin legacy styles

davidtrussler:
[Trello](https://trello.com/c/DbE1Zaii/625-deal-with-styles-from-admin-legacy-that-are-used-in-views)

This PR: 
- Removes the CSS files below
- Removes references to legacy styles in Design System views
- Removes views that are no longer in use
- Removes JS files no longer in use

It might be easier to review individual commits on this - there are many files being removed when various dependencies are taken into account that make more sense when viewed that way. 

CSS files removed: 
- app/assets/stylesheets/admin_legacy/_document.scss
- app/assets/stylesheets/admin_legacy/_forms.scss
- app/assets/stylesheets/admin_legacy/_sidebar.scss
- app/assets/stylesheets/admin_legacy/helpers/_broken-links-report.scss
- app/assets/stylesheets/admin_legacy/views/_edition-tags-edit.scss
- app/assets/stylesheets/admin_legacy/helpers/_document-finder.scss


RodneyJohnsonGDS:
Some really good code cleanup here @davidtrussler. Perhaps we can also look at some of the scss files that are unused and still referenced in the _base.scss file. 

- I notice the styles within admin_legacy/helpers/_statistics-announcements.scss are unused and can be removed. 
- Likewise, in the admin_legacy/helpers/_player-container.scss file there are a number of styles that are unused 

May be worthwhile giving a once over to pick up as part of the cleanup; styling that not longer in use.

app/controllers/admin/link_check_reports_controller.rb
mtaylorgds, line 1: Is it safe to delete this?

There's still a route for it and I get a 404 when trying to do a link check report locally with this controller deleted…

mtaylorgds (changes requested):

app/controllers/admin/link_check_reports_controller.rb
davidtrussler: @mtaylorgds I have restored the deleted files here as discussed.

mtaylorgds (approved):

PR #8530, opened 2023-11-20
Remove graphviz_transitions dependency

richardTowers:
This was added in 2c7c7df4ed, but I suspect nobody knows it's there or how to use it.

The answer is:

```
rake transitions:draw CLASS=Edition
```

Which used to produce a nice PNG image like this:

edition_transitions

Unfortunately, this has been broken since rails 5 because of a breaking change to the `.parameterize()` method - https://github.com/itkin/graphviz_transitions/pull/1

Less dependencies is better dependencies, so let's bin it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,7 +73,6 @@ end
 group :development do
   gem "better_errors"
   gem "binding_of_caller"
-  gem "graphviz_transitions"
   gem "mechanize"
 end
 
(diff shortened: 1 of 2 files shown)


richardTowers:
This diagram is also already in the codebase in plant UML form - https://github.com/alphagov/whitehall/blob/main/docs/diagrams/edition_state_transitions.png

beccapearce (approved):
Looks good!

PR #8534, opened 2023-11-21
Clean up document searches legacy code.

farahTW:
This PR removes files related to document searches legacy code.
It removes the files which are no longer in use after transition to design system.

**Trello:**
https://trello.com/c/wvra2ses


mtaylorgds (approved):
LGTM!

PR #8535, opened 2023-11-21
Use republishing method from Whitehall::PublishingApi instead of calling publishworker directly

tunylund:
Republish method sets an `update_type: 'republish'` flag on the published content that has different side-effects than the default flag `'major'`. For example on regular publish, a `'major'` value is set as the `update_type` and that causes emails to be sent of the update. Republish uses `'republish'` flag which means that a change note is not required and there are no extra emails sent.

[Trello](https://trello.com/c/6t2gbpLB/241-story-promotional-feature-image-to-use-asset-ids)

### DIFF
--- a/test/integration/take_part_page_test.rb
+++ b/test/integration/take_part_page_test.rb
@@ -9,27 +9,25 @@ class TakePartPageTest < ActiveSupport::TestCase
     Services.asset_manager.stubs(:create_asset).returns("id" => "http://asset-manager/assets/asset_manager_id_s300", "name" => "s300_minister-of-funk.960x640.jpg")
   end
 
-  test "TakePartPage is published to the Publishing API on save" do
-    republish_count_from_create_asset_worker = 7
+  test "TakePartPage is published to Publishing API on save and republished when images finish uploading" do
     publish_count_from_after_commit = 1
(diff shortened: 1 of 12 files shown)


test/integration/take_part_page_test.rb
```diff
@@ -9,27 +9,25 @@ class TakePartPageTest < ActiveSupport::TestCase
...
-    republish_count_from_create_asset_worker = 7
+  test "TakePartPage is published to Publishing API on save and republished when images finish uploading" do
     publish_count_from_after_commit = 1
-    expected_number_of_times_published = republish_count_from_create_asset_worker + publish_count_from_after_commit
+    republish_count_from_create_asset_worker = 7
 
     Sidekiq::Testing.inline! do
-      presenter = PublishingApiPresenters.presenter_for(@take_part_page)
+      publish_presenter = PublishingApiPresenters.presenter_for(@take_part_page)
+      republish_presenter = PublishingApiPresenters.presenter_for(@take_part_page, { update_type: "republish" })
```
lauraghiorghisor-tw, line 18: Uuuu, nice! 😎

lauraghiorghisor-tw (approved):
👍🏻

syed-ali-tw (approved):

PR #8536, opened 2023-11-22
Add worker that Republishes large pages to Publishing API

davidgisbey:
## Description

At the moment, we republish large pages like the ministers index page and how government works pages synchronously. These are expensive and when chained together when a collection of models are reordered can lead to page timeouts.

This commit adds a worker which can be used to present these pages downstream async and uses the new worker in place of the sync calls.

## Performance increase

tested on organisation edit page for GDS https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/organisations/government-digital-service

3 saves 

1. 14.55 seconds
2. 10.02 seconds
3. 17.74 seconds

And with this applied

1. 0.9 seconds
2. 0.89 seconds
3. 0.89 seconds

## Trello card

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -83,7 +83,7 @@ class Contact < ApplicationRecord
   end
 
   def republish_embassies_index_page_to_publishing_api
-    PresentPageToPublishingApi.new.publish(PublishingApi::EmbassiesIndexPresenter)
+    PresentPageToPublishingApiWorker.perform_async("PublishingApi::EmbassiesIndexPresenter")
   end
 
   def publishing_api_presenter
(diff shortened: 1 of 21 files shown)


ryanb-gds (approved):
🏎️ 🏎️ 🏎️ 

This does worry me slightly because it seems like such an obvious improvement that I'm amazed nobody has made it before.

I also noted [this commit](https://github.com/alphagov/whitehall/commit/22f76d0039e72c75275e3817f4929d67dbee5328) which says "potential race condition". Well, we've just made race conditions much more likely 😬 

However... I think we are going to have to subject this to the scream test 🚢

davidgisbey:
okay i'll keep an eye out for any tickets etc. for a while, but yea i take your point, it is strange this hasn't been done before 🤔

PR #8537, opened 2023-11-22
Feature to use asset ids

yuetylauiris:
Implement `FeaturedImageData` for `Feature` to make it more consistent with other similar classes that have a carrierwave `mount_uploader` on them, as well as enabling `Feature` to use the assets flow rather than legacy url path.

This change also publishes `medium_resolution_url` and `high_resolution_url` images of any Feature. These are eventually used by the Collections application to render different size images of the Feature.

[Trello Card](https://trello.com/c/JYIbvFb0/248-feature-to-use-assetids)


app/models/edition/lead_image.rb
```diff
@@ -34,8 +34,6 @@ def lead_image_caption
   end
 
   def lead_image_has_all_assets?
```
tunylund, line 36: this function seems redundant now. If we delete it, or delegate it, we can then delete the related tests as well

app/presenters/publishing_api/speech_presenter.rb
```diff
@@ -139,8 +139,6 @@ def has_image?
     end
 
     def image_has_all_assets?
```
tunylund, line 141: this looks like a redundant function now

app/models/edition/lead_image.rb
lauraghiorghisor-tw: But the method is used in case study presenter and news article presenter, it has to be defined within the concern. There are no filtering tests at the presenter level for lead image assets missing. It would have been too repetitive to sprinkle those tests everywhere that includes the lead image concern. So these little tests cover that a bit, albeit shallowly.

PR #8540, opened 2023-11-22
Refactor cabinet ministers controller

davidgisbey:
## Description

This breaks out the CabinetMinistersController#update action into 4 separate endpoints which handle updating each of the 4 reorderable lists on the [cabinet ministers show page](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/cabinet_ministers#cabinet_minister).

Each of these have a separate page to reorder them, but currently share the same endpoint which handles updating them. Due to this there's a bunch of logic that figures out which reorder page has hit the endpoint and uses the referrer to constuct the anchor for the redirect. 

It's pretty hacky and quite tricky to parse. Separate endpoints which handle updating that specific subset of ministerial roles with are slightly more verbose seems preferable.

For now i've left a good amount of duplication in each update order actions. The next thing i'm working on is going to be a refactor for all the reorder endpoints. If i'm unable to refactor the logic from these controller actions elsewhere i'll consider a direct refactor of the endpoints.

## Trello card

### DIFF
--- a/app/controllers/admin/cabinet_ministers_controller.rb
+++ b/app/controllers/admin/cabinet_ministers_controller.rb
@@ -12,6 +12,14 @@ class Admin::CabinetMinistersController < Admin::BaseController
     @roles = MinisterialRole.includes(:translations).where(cabinet_member: true).order(:seniority)
   end
 
+  def order_cabinet_minister_roles
+    params["ordering"].each do |id, ordering|
+      Role.find(id).update_column(:seniority, ordering)
+    end
+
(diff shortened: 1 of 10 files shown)


minhngocd (approved):
Thanks for this David! it reads much better!
A very small note on the test assertions for consistency some have been updated to to assert `(anchor: ... ` where as some are still assert redirected to `#anchor` - but it's not a blocking need.
Also amazing for spotting we've had the wrong header for the organisations reordering 😱

davidgisbey:
Great spot i'll update those for consistency now. Thanks so much for the review!

PR #8542, opened 2023-11-23
Use asset-id for for call_for_evidence_response_form_data

syed-ali-tw:
In order for call_for_evidence_response_form to move away from using legacy_url_path It needs to use asset-id. This PR creates the relationship to allow call_for_evidence_response_form to use_non_legacy_endpoints and create assets in whitehall. It also introduces the functionality to publish call_for_evidence only when all the assets related to the call_for_evidence_response_form is uploaded i;e block publishing if assets are not uploaded.

[Trello](https://trello.com/c/tgZd1abb/237-story-call-for-evidence-response-form-to-use-asset-ids)

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -463,7 +463,7 @@ private
   end
 
   def clear_response_form_file_cache
-    response_form_params = edition_params.dig(:consultation_participation_attributes, :consultation_response_form_attributes, :consultation_response_form_data_attributes)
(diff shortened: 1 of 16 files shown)


app/controllers/admin/calls_for_evidence_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::CallsForEvidenceController < Admin::EditionsController
   before_action :cope_with_call_for_evidence_response_form_data_action_params, only: [:update]
+  before_action :clear_file_cache, only: [:update]
```
tunylund, line 3: needs to be create as well

tunylund (approved):

PR #8547, opened 2023-11-23
Add documentation on contact expansion in govspeak

richardTowers:
I've been learning about how Whitehall manages dependencies between Editions and Contacts, and makes sure Editions are kept up to date when Contacts are updated.

Writing it down for the benefit of others, as well as future me who will no doubt forget all of this basically immediately.


ryanb-gds (approved):
Thanks for this @richardTowers , sorry it took so long for a review!

PR #8549, opened 2023-11-23
Fix caching for consultation response form data.

syed-ali-tw:
Scenario:
CREATE a consultation with a consultation response form(file-1) 
EDIT the consultation and replace the consultation response form (file-2) 
REMOVE title from the form to cause a validation error(or cause any validation error) 
REPLACE the consultation response form (file-3) and fix the validation error 
SAVE the consultation.
Given above steps currently it saves and renders file-2 whereas the last uploaded is file-3. this is due to caching file not being replaced by the actual final file. This fix is to allow to save file over cache-file if present. This fix is across Attachments and Images too.

[Trello](https://trello.com/c/dD8j8HqT/239-story-consultation-response-form-to-use-asset-ids)

### DIFF
--- a/test/functional/admin/consultations_controller_test.rb
+++ b/test/functional/admin/consultations_controller_test.rb
@@ -339,6 +339,64 @@ class Admin::ConsultationsControllerTest < ActionController::TestCase
     assert_equal "greenpaper.pdf", consultation.consultation_participation.consultation_response_form.consultation_response_form_data.carrierwave_file
   end
 
+  test "PUT :update discards file_cache when a file is provided" do
+    two_pages_pdf = upload_fixture("two-pages.pdf")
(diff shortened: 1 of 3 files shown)


test/functional/admin/consultations_controller_test.rb
```diff
@@ -339,6 +339,64 @@ class Admin::ConsultationsControllerTest < ActionController::TestCase
...
+                          id: response_form.consultation_response_form_data.id,
+                          file: greenpaper_pdf,
+                          file_cache: response_form_data.file_cache,
+                        },
+                      },
+                    },
+                  } }
+  end
+
+  test "CREATE :discards file_cache when a file is provided" do
```
tunylund, line 373: the naming of these tests seems to be `test "<http method> <action> ..."

tunylund (approved):

PR #8557, opened 2023-11-27
Refactor has many associations to use extension with shared reorder! method

davidgisbey:
## Description

We've got loads of reordering going on using params[:ordering] with heavy amounts of duplication.

This takes a first step at cleaning some of this up by adding an association extension which handles reordering the `ordering` column for the has_many associations.

It doesn't clean up the cabinet ministers controller duplication, which will come in a follow up PR 

There's quite a lot going on in this PR, i'm going to leave a few comments throughout on areas that i think could use some discussion.

## Trello card

### DIFF
--- a/app/controllers/admin/people_controller.rb
+++ b/app/controllers/admin/people_controller.rb
@@ -55,12 +55,9 @@ class Admin::PeopleController < Admin::BaseController
 
   def update_order_role_appointments
     current_role_appointment_orderings = @person.current_role_appointments.map(&:ordering)
+    new_ordering = map_ordering_to_current_role_appointment_ordering_values(current_role_appointment_orderings)
 
-    params[:ordering].each do |appointment_row|
-      id, ordering = appointment_row
-      role_appointment = @person.role_appointments.find(id)
-      role_appointment.update!(ordering: current_role_appointment_orderings[ordering.to_i - 1])
-    end
+    @person.role_appointments.reorder!(new_ordering)
 
     flash[:notice] = "Role appointments reordered successfully"
     redirect_to admin_person_path(@person)
@@ -104,5 +101,15 @@ private
     if person_params.dig(:image_attributes, :file_cache).present? && person_params.dig(:image_attributes, :file).present?
       person_params[:image_attributes].delete(:file_cache)
     end
+  end
+
+  def map_ordering_to_current_role_appointment_ordering_values(current_role_appointment_orderings)
+    update_order_role_appointments_params.transform_values do |value|
+      current_role_appointment_orderings[value.to_i - 1]
+    end
+  end
+
+  def update_order_role_appointments_params
+    params.require(:role_appointments)["ordering"]
   end
 end
--- a/app/models/user_orderable_extension.rb
+++ b/app/models/user_orderable_extension.rb
@@ -7,9 +7,9 @@ module UserOrderableExtension
     end
   end
 
-  def reorder!(new_order)
+  def reorder!(new_order, column = :ordering)
     new_order.each do |id, ordering|
-      find(id).update!(ordering:)
+      find(id).update!("#{column}": ordering)
     end
   end
 end
--- a/app/models/user_orderable_extension.rb
+++ b/app/models/user_orderable_extension.rb
@@ -1,0 +1,15 @@
+module UserOrderableExtension
+  def reorder_without_callbacks!(new_order, column = :ordering)
+    transaction do
+      new_order.each do |id, ordering|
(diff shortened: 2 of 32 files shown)


app/controllers/admin/people_controller.rb
```diff
@@ -50,12 +50,9 @@ def reorder_role_appointments
 
   def update_order_role_appointments
     current_role_appointment_orderings = @person.current_role_appointments.map(&:ordering)
+    new_ordering = map_ordering_to_current_role_appointment_ordering_values(current_role_appointment_orderings)
 
-    params[:ordering].each do |appointment_row|
-      id, ordering = appointment_row
-      role_appointment = @person.role_appointments.find(id)
-      role_appointment.update!(ordering: current_role_appointment_orderings[ordering.to_i - 1])
```
davidgisbey, line 57: Removing this `update!` causes model callbacks not to trigger which is why the controller test is failing. Before we worry about trying to fix the controller i thought it would be worth discussing a few options:

1. adding functionality to the reorder! method which toggles `update_column` to `update!`
2. manually call republishing all the dependencies. Maybe use a service object or something similar which wraps up calling the 3-4 methods the callbacks do
3. not using the extension in this use case

Let me know what you think!

davidgisbey: Yes that's exactly the case. I guess we could pass an additional argument like `callbacks: true` which could toggle between the two.

I'm just aware it makes the method slightly more complex. I'll put something together for that now 👍 in a new commit.

ryanb-gds: I'm generally [against boolean flag args](https://martinfowler.com/bliki/FlagArgument.html) in favour of separate methods, but not strongly in this case. I was hoping we could just apply callbacks for everything now without exception

app/controllers/admin/people_controller.rb
```diff
@@ -55,12 +55,9 @@ def reorder_role_appointments
 
   def update_order_role_appointments
     current_role_appointment_orderings = @person.current_role_appointments.map(&:ordering)
+    new_ordering = map_ordering_to_current_role_appointment_ordering_values(current_role_appointment_orderings)
```
ryanb-gds, line 58: Question: why is this reordering different to all of the other ones in that the values seem to be off-by-one? Might be worth a chat if you have time

ryanb-gds (approved):
Looks good. I wouldn't mind understanding a bit more about the role appointments ordering because that's not very easy to follow at the moment. But I think we're in good shape to merge this, could always look at the appointments in a separate PR

app/models/user_orderable_extension.rb
```diff
@@ -0,0 +1,15 @@
+module UserOrderableExtension
+  def reorder_without_callbacks!(new_order, column = :ordering)
+    transaction do
+      new_order.each do |id, ordering|
+        find(id).update_column(column, ordering)
+      end
+    end
+  end
+
+  def reorder!(new_order)
```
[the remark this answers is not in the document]
davidgisbey: yea i think you're right. Will sort that before merge 👍

app/controllers/admin/people_controller.rb
davidgisbey: Yea this is a bit of a snowflake at the moment, and i suspect a follow up PR can probably iron out some of the issues. Basically an ordering has always been passed down explicitly rather than implicitly (based on the order of the collection).

It looks like role appointments are retrieved via link expansion rather than being nested within the person object. Due to this i guess there needs be an explicit way to order their previous roles as you they're ordered on this page 

https://www.gov.uk/government/people/boris-johnson

However, there's an additional need to order a persons current role appointments as the order of a persons roles is used to construct their title.

For example on Rishi Sunaks page https://www.gov.uk/government/people/rishi-sunak

You could shuffle `Minister for the Civil Service` ahead of `First Lord of the Treasury` and the context on that page would move from

`Prime Minister, First Lord of the Treasury, Minister for the Union, and Minister for the Civil Service`

to 

`Prime Minister, Minister for the Civil Service, Minister for the Union, and First Lord of the Treasury`

Historically there has been a need to do this for ministers. I remember it happening a few times since i've joined GDS.

There's probably a way we could order these in Publishing API to ensure the links are ordered in created_at order and then mark specific ones as current to use for reordering the title but that's probably beyond the scope of this. Wdyt?

davidgisbey: Whatever we decide i think it should be tackled outside of this PR so i'll sort the above comment and merge but happy to pick this up via slack on carry on chatting here 👍

PR #8562, opened 2023-11-28
Spike out rendering readable errors when GovspeakLinkValidator blocks force publishing & force scheduling  

davidgisbey:
## Description

We've had a Zendesk ticket come through where a user was unable to force schedule a document for publication due to a link violating our guidelines. Overall the error message is pretty unhelpful 

image

The link which triggered the error in this instance is 

```
[Find out about call charges](/call-charges).
```

When the broken link checker runs it says this is a valid link (which it should) as it hits gov.uk/call-charges which returns a 200.

However, when we run our own validations against links in the body, it checks that external links (links to documents created outside of Whitehall, in this case Publisher) use the full URL.

So the user it told there are no broken links and are given no extra details on which link has violated linking guidelines and why.

We have details on which links have caused errors and potential fixes on the validator. This PR spikes out formatting and rendering these in a way that would be helpful to the user

## Implementation

I think the most complete implementation of this would probably be to add a view component which handles rendering the sidebar notices (all the information on why you can't publish a document). This would allow us to use the same pattern in presenting errors as we do in the broken links report.

The problem with that is we would be reliant on parsing the failure reasons from the updater and inferring when the failure is due to a violation of the linking guidance by doing something like 

```
failure_reason == "This edition contains links which violate linking guidelines"
```
which doesn't feel particularly great.

There's some other potential solutions. One of which i've added as the 3rd commit in this PR (i also don't really like this implementation)

I've opened this as a draft just to get some initial feedback. Eager to hear peoples thoughts on this.

### Before

image

image

### After 

image

image

### DIFF
--- a/lib/data_hygiene/govspeak_link_validator.rb
+++ b/lib/data_hygiene/govspeak_link_validator.rb
@@ -12,14 +12,14 @@ module DataHygiene
 
         fix = if link.first == "/"
                 unless self.class.is_internal_admin_link?(link)
-                  "Please use either absolute paths for documents created in publisher, e.g. /government/admin/publications/3373, or full URLs for other GOV.UK links"
+                  "Please use either absolute paths for documents created in Whitehall, e.g. /government/admin/publications/3373, or full URLs for other GOV.UK links"
                 end
               elsif self.class.is_internal_admin_link?("/#{link}")
                 "This is an invalid admin link.  Did you mean /#{link} instead of #{link}?"
               elsif !%r{^(?:https?://|mailto:|#)}.match?(link)
                 "Non-document or external links should start with http://, https://, mailto:, or # (for linking to sections on the same page, eg #actions on a policy)"
               elsif link.match?(/whitehall-admin/)
-                "This links to the whitehall-admin domain. Please use paths, eg /government/admin/publications/3373, for documents created in publisher (see guidance on creating links) or full URLs for other GOV.UK links."
+                "This links to the whitehall-admin domain. Please use paths, eg /government/admin/publications/3373, for documents created in Whitehall (see guidance on creating links) or full URLs for other GOV.UK links."
               end
 
         { link:, start: match.begin(0), end: match.end(0), fix: } if fix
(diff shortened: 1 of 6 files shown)


lib/data_hygiene/govspeak_link_validator.rb
```diff
@@ -12,20 +12,39 @@ def errors
...
               elsif link.match?(/whitehall-admin/)
-                "This links to the whitehall-admin domain. Please use paths, eg /government/admin/publications/3373, for documents created in publisher (see guidance on creating links) or full URLs for other GOV.UK links."
+                "This links to the whitehall-admin domain. Please use paths, eg /government/admin/publications/3373, for documents created in Whitehall (see guidance on creating links) or full URLs for other GOV.UK links."
               end
 
         { link:, start: match.begin(0), end: match.end(0), fix: } if fix
       end
     end
 
+    def errors_to_html
```
davidgisbey, line 29: Honestly, i don't really like this approach, but it seemed like the quickest way to get a draft PR open which highlights and resolves the issue.

My preferred approach would be to move a lot of this logic into a view component which handles rendering the sidebar notices (all the inset text components)

The issue with this is that the view component would have to infer when the failure reason was due to link violations and then do a bunch of work to munge that error reason into html.

Which might be fine but seems a little brittle. I guess the edition service object coukd have a constant like 

```
LINK_VIOLATION_POLICY_ERROR_MESSAGE = "This edition contains links which violate linking guidelines".freeze
```

Then the view component could use that constant to determine what action to take.

All the solutions feel a bit sub-optimal though tbh.

davidgisbey: The problem with moving failure reasons into enums would be that the error message is often constructed dynamically based on the errors on the edition.

https://github.com/alphagov/whitehall/blob/82f2d884acb78bed08bae1568c5e1e8ca17663be/app/services/edition_scheduler.rb#L12-L23

Moving them into their own objects would work though.

A refactor like that's probably beyond the scope of what i can achieve on 2ndline though.

Would you like to to add a view component in this PR, or hand it over to Publishing to put into the backlog? 

It's a real pain for users when they encounter it, but as far as i'm aware it's not a particularly frequent problem.

davidgisbey: I've carded up something and moved it to the top of ideas https://trello.com/c/pNiIMunP/2224-users-are-given-no-help-identifying-inks-violate-our-policy-in-the-body-for-editions

I'll close this PR for now but it's linked to the Trello ticket so it can be referenced

PR #8565, opened 2023-11-28
Make the previous edition method do what its name suggests it should

ryanb-gds:
Before this commit, the Edition#previous_edition method was actually fetching the last post-publication edition for the document other than itself. That is not the behaviour consumers will expect based on the function name.

This commit changes the behaviour to fetch the previous post publication edition relative to this edition. Given that all tests pass when we provide a created_at date for the previous edition older than the current edition, this seems safe. Hopefully this will reduce confusion and potential misuse of the function in the future.

Trello: https://trello.com/c/AAeP7jad


ryanb-gds:
Closing this as agreed with @minhngocd, as we're agreed that this probably isn't much of an improvement, and actually in many cases consumers are probably expecting something more like "get me the published edition unless it's this then get me the previous one", but that's not a very nice method name. We're going to get our heads together to explore potentially moving the behaviour on to the document model, if that makes sense for consumers

PR #8575, opened 2023-11-30
Remove "tabs" from secondary nav aria labels

ryanb-gds:
A recent accessibility review of Whitehall raised a WCAG 2.1 level A failure for secondary navigation within Whitehall. Specifying an aria label using the word "tabs" implied that the secondary navigation links should behave as tabs, instead of as links, which could be confusing for non-sighted users.

Trello: https://trello.com/c/fdYHMbqp


davidgisbey (approved):
lgtm!

PR #8576, opened 2023-11-30
Further accessibility improvements to secondary navigation usage

ryanb-gds:
A recent accessibility audit of Whitehall returned a level A WCAG failure because navigation menus were labelled as "tabs" on some Whitehall pages. Elements labelled "tabs" should have tab-like behaviour, instead of menu link behaviour. We are therefore removing all mentions of "tabs" from secondary navigation aria labels.

Trello: https://trello.com/c/fdYHMbqp

### DIFF
--- a/app/views/admin/topical_event_organisations/index.html.erb
+++ b/app/views/admin/topical_event_organisations/index.html.erb
@@ -12,7 +12,7 @@
 
 <div class="govuk-!-margin-bottom-8">
   <%= render "components/secondary_navigation", {
-    aria_label: "Document tabs",
+    aria_label: "Document navigation",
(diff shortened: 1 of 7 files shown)


app/views/admin/topical_event_organisations/index.html.erb
```diff
@@ -12,7 +12,7 @@
 
 <div class="govuk-!-margin-bottom-8">
   <%= render "components/secondary_navigation", {
-    aria_label: "Document tabs",
+    aria_label: "Document navigation",
```
davidgisbey, line 15: This is a pre-existing issue but it looks like this was copy pasted and should be `"Topical Events navigation"`

PR #8577, opened 2023-11-30
Copy yarn 4 config into Docker image

ryanb-gds:
To use Yarn 4 the image must have the configuration file and the release file present. When they are not present the yarn install command fails during the Docker build because the --frozen-lockfile option prevents the Yarn 1 install from rewriting the lockfile to version 1 syntax (which we don't want anyway).

Apparently `yarn workspaces focus` is the new `yarn install -production` 🤷 : https://yarnpkg.com/migration/guide#renamed-commands


ryanb-gds:
Closing this because we can't merge while the accessible autocomplete multiselect package won't build with Node 18

PR #8586, opened 2023-12-04
Remove all Admin Legacy preview functionality

minhngocd:
Previews from Admin Legacy are no longer used and need to be removed.
There is functionality to enhanceYouTubeVideoLinks, but manually testing this, it doesn't seem to be replacing anything, could be due to how we are rendering the preview panel.
Either way, these files are not included in application.js so isn't included in the asset bundle to go to production anyways, hence currently doesn't seem to have any use.

https://trello.com/c/r8eG9T6o/2240-remove-bootstrap


ryanb-gds (approved):
Weird about that Youtube embedding thing 🤷

minhngocd:
Have tested the Youtube Link embedding on integration - this change does not seem to affect the rendering of Youtube Videos on the frontend, but maybe raises a question of whether or not we should be doing the same embedded video rendering upon preview. 
Regardless, this change does not affect anything in that aspect.

PR #8589, opened 2023-12-04
Provide unique aria labels for attachment edit and removal links

ryanb-gds:
In order to make it possible for screen readers to differentiate between links when users are navigating outside of the context (e.g. when moving through a list of all links on the page), we should reference the title of each attachment in the accessible link label.

Trello: https://trello.com/c/kiv7vMXj


davidgisbey (approved):
lgtm!

PR #8594, opened 2023-12-05
Refactor homepage list controller

minhngocd:
# What
- Removed unused variables
- Moved the extension of the Module to a more visible place on the controllers
- Add and Remove methods don't seem to be called from anywhere within Whitehall.
- Searched for the usage of the routes and did not see anything. Should be safe to remove.

# Why
This is a particularly confusing module which was selected for refactoring during our mobbing sessions. The proposed changes make it easier to read and understand.


davidgisbey (approved):
approved via mob

PR #8596, opened 2023-12-06
Add concern which handles reordering records within a class

davidgisbey:
## Description

This sort of follows on from https://github.com/alphagov/whitehall/pull/8557 which adds an association extension for reordering a models associations using the various ordering columns.

This adds very similar functionality but within a concern which handles reordering records within a class .

The final couple of commits update TakePartPages' to use the new concern. This may or may not be better. Very open to discussion. I've laid out my logic for doing so in the commit message 

## Trello card


app/models/user_orderable_class.rb
```diff
@@ -0,0 +1,13 @@
+module UserOrderableClass
```
ryanb-gds, line 1: I'd probably call this one `UserOrderable` and the extension `UserOrderableExtension` or `UserOrderableRelation` I think. Calling a module a class is potentially confusing!

I'd actually consider nesting the extension inside this module, but not too sure what the impact of that would be.

ryanb-gds (changes requested):
I think I'd like for us to decouple the reorderable module tests from our application models, if at all possible.

ryanb-gds:
> I think I'd like for us to decouple the reorderable module tests from our application models, if at all possible.

Also, and I should have led with this, great work! Love that we're pulling database logic down into the model layer where it belongs and wrapping it in a single API

app/models/user_orderable_class.rb
davidgisbey: cool have done. I'll give nesting the extension a go

davidgisbey:
I think this ones ready for a re-review when you get a chance please @ryanb-gds

ryanb-gds (approved):
Looks great, super job 🚀

davidgisbey:
Thanks so much for all the feedback & the review 👍

PR #8600, opened 2023-12-06
Ensure focus remains on Doc History show select

davidgisbey:
## Description

When you switch between seeing document history, internal notes or both on the select, it loses focus of the element. This is poor behaviour for screen readers as they jump back to the top of the page.

This is untested atm. I've added a card to the backlog to add some [here](https://trello.com/c/Tj7FQ0bP/2248-add-tests-for-doc-history-paginator-js-module) 

## Before

image

image

## After 

image

image

### DIFF
--- a/app/assets/javascripts/admin/modules/document-history-paginator.js
+++ b/app/assets/javascripts/admin/modules/document-history-paginator.js
@@ -75,6 +75,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
         })
         .then(function (html) {
           module.innerHTML = html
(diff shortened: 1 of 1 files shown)


dnkrj (approved):
Ideally we wouldn't rerender the select as it is announced again by the screenreader when that happens, but avoiding a refactor, this approach makes sense.

It's probably also worth considering if the "older" and "newer" buttons should emulate this behaviour (or something similar to it), but again that's a separate discussion.

davidgisbey:
Okay great i'll update the trello card to update the implementation so it only inserts everything below the select when pagination occurs.

PR #8601, opened 2023-12-07
Rename component files

andysellick:

## What
Rename component files to follow gem component conventions, in order to work with component auditing tools. See individual commits for details.

## Why
Part of https://github.com/alphagov/govuk_publishing_components/issues/3369

## Visual changes
None, hopefully, although this does include a change to a classname for the summary card component, which I'm unable to test as I can't run whitehall locally. This also includes a renaming of some component YML files, which might have an impact on the URLs for components within the component guide, which I'm also unable to check.


davidgisbey (approved):
lgtm. It's updated the component urls but it's consistent now which is great. Thanks so much for sorting this!

PR #8604, opened 2023-12-07
Refactor asset updater

yuetylauiris:
Stop raising AssetAlreadyDeleted error since nobody is caring about them :( They just cause noise that hides other errors we care about.

[Trello](https://trello.com/c/NpDeRmj6/272-task-stop-raising-assetalreadydeleted)

### DIFF
--- a/app/services/asset_manager/asset_updater.rb
+++ b/app/services/asset_manager/asset_updater.rb
@@ -1,15 +1,9 @@
 class AssetManager::AssetUpdater
   include AssetManager::ServiceHelper
 
-  class AssetAlreadyDeleted < StandardError
-    def initialize(identifier, asset_data_id)
-      super("Asset '#{identifier}' for Attachment Data #{asset_data_id} expected not to be deleted in Asset Manager")
-    end
-  end
-
   class AssetAttributesEmpty < StandardError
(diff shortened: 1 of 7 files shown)


app/workers/asset_manager_update_whitehall_asset_worker.rb
syed-ali-tw, line 1: we might be able to delete this.. I could not find a usage of this.

app/services/asset_manager/asset_updater.rb
```diff
@@ -1,36 +1,30 @@
...
   end
 
   def self.call(*args)
     new.call(*args)
   end
 
-  def call(asset_manager_id, asset_data, new_attributes = {})
-    update_with_asset_manager_id(asset_manager_id, asset_data, new_attributes)
+  def call(asset_manager_id, new_attributes)
+    update_with_asset_manager_id(asset_manager_id, new_attributes)
```
syed-ali-tw, line 15: Think there is no need for the other method here.

lauraghiorghisor-tw: How about removing the `update_with_asset_manager_id` itself?

PR #8605, opened 2023-12-07
Add more descriptive aria labels to page links

jyoung-gds:
[Trello ticket link](https://trello.com/c/5wCEpjGA/2212-make-pagination-links-more-descriptive-for-screenreader-users)

# What
This updates the aria labels of page links so that they are more accessible and descriptive when using a screen reader.

# Why
This was flagged as high priority in the most recent accessibility report.

# Additional information
The accessibility report called for more context on page links when using a screen reader and gave the example of adding the aria label "Page 1 of search results". Myself and @ryanb-gds investigated adding the context of "search results" to the aria label but could not find a clean way to do it. As the `govuk_paginator` theme is used in multiple places, a suitable context would need to be passed in for each use case. However, the paginate method from Kaminari is restrictive in the options you can provide it. For the time being, we propose proceeding without the additional context. The changes proposed in this PR bring the code in line with the Design System which has aria labels like "Page 1".

### DIFF
--- a/app/helpers/admin/pagination_helper.rb
+++ b/app/helpers/admin/pagination_helper.rb
@@ -99,6 +99,7 @@ module Admin::PaginationHelper
         href: build_path_for(1),
         label: "1",
         current: current_page == 1,
+        aria_label: "Page 1",
       }
     end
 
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
lgtm if we're happy it's in line with the design system component which it sounds like it is

PR #8620, opened 2023-12-12
Ensure edition is republished when assoication is updated

davidgisbey:
## Description

We want to be able to republish documents when an editions associations are updated. At the moment, we often don't do this which is unintuitive to the user. To get round this they can go to the edit edition page and click save to send the changes downstream, but it's no ideal behaviour.

We've had quite a few 2ndline tickets come in around this issue across Whitehall.

This updates the PublishingApiDocumentRepublishingWorker to not publish a draft edition downstream if it is invalid, then calls it in the controllers where associations are created/updated/deleted etc.

The edition largely is only invalid when it's carried over from a post publication state to draft as we skip validations while duping the edition and it's associations.

## Trello card

https://trello.com/c/HqQvJOq3/2259-ensure-draft-content-store-is-updated-when-a-user-saves-a-page

### DIFF
--- a/app/controllers/admin/call_for_evidence_responses_controller.rb
+++ b/app/controllers/admin/call_for_evidence_responses_controller.rb
@@ -13,6 +13,7 @@ class Admin::CallForEvidenceResponsesController < Admin::BaseController
     @response = response_class.new(response_params)
     @response.call_for_evidence = @edition
     if @response.save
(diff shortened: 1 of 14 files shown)


ryanb-gds (approved):
Beautiful 💅 this is a great improvement

PR #8628, opened 2023-12-13
Gracefully stop processing attachments when the related document has been destroyed

tunylund:
An issue in production showed that sometimes user might have uploaded a file and for whatever reason, the create asset worker executes much later. So much so that the user might have destroyed the actual original edition.

If this happens, the worker should stop processing the attachment instead of failing and forever retrying.


lauraghiorghisor-tw (approved):
✅

PR #8636, opened 2023-12-14
Fix name of test so that it runs in CI

minhngocd:
https://trello.com/c/r8eG9T6o/2240-remove-bootstrap


minhngocd:
Annoyingly, would like to remove the default email `default from: "<winston@alphagov.co.uk>"` , but the tests fail on the error `SMTP From address may not be blank: nil` 

Will test this branch on integration to make sure email notifications still happen

davidgisbey (approved):
looks good. Give me a yell if you a hand testing it on int or getting access to the shared mailbox stuff goes to

minhngocd:
Tested this in Integration with Laura - the email sending still works and looks the same as previous

PR #8637, opened 2023-12-14
Rails 7.1 upgrade

ryanb-gds:
This was a little bit tricky. Whitehall, it turns out, is very, very exposed to low level changes in Rails. However, I think we're just about there. A few notable aspects of the upgrade are documented below, and you may find it easier to review this commit by commit. Frustratingly, most of the changes that caused issues for us were not noted in the official upgrade guide.

### Deprecation of Rails secrets in favour of credentials

`Rails.application.secrets` has been deprecated in favour of `Rails.application.credentials`. However, because we use env vars to manage our production secrets, we'd much rather it hadn't been. @KludgeKML suggested a handy workaround to merge the existing secrets into the credentials object.

### Refactor of cancel routes helper necessitated by mysterious Active Model equality evaluation change

Something (wasn't able to figure out what) in Active Model has changed which means that `Edition === object` evaluates to false for instances of classes that extend the `Edition` class. Previously it evaluated to true. This caused the case statement in `cancel_routes_helper.rb` to behave differently following the upgrade. I've worked around this by overriding the corporate information page form template and moving away from the case statement in `cancel_routes_helper`. I don't really like the helper anyway, I feel we would be much better off deciding what link to use within each edition type template, where we have all the context we need, rather than having a helper checking the type of the edition.

### Serialization method signature change

The signature of the active record `serialize` method has changed. The old one is still available, but deprecated. The new one requires us to specify a "coder" for the serialized attribute. By default Rails used to assume the `YAML` coder, so that is what I have used in all cases.

### View test implementation changes

The old `view_test` implementation no longer functioned with Rails 7.1. I've copied the latest from RSpec again and tweaked it so that it behaves the same as the previous implementation.

### Active Record config overrides

Some of the new Rails default configurations made significant changes to how active record callbacks behaved. Unfortunately, many aspects of Whitehall depend on the callback behaviour, so I have overridden the new default values for the following configuration keys:

- config.active_record.run_commit_callbacks_on_first_saved_instances_in_transaction
- config.active_record.before_committed_on_all_records
- config.active_record.run_after_transaction_callbacks_in_order_defined

In some cases the affected behaviour was provided by Carrierwave, so it may be that future versions of Carrierwave will enable us to adopt the new defaults.

### Links

Trello: https://trello.com/c/ZAp0nyoz


test/support/view_rendering.rb
```diff
@@ -41,11 +46,68 @@ def find_all(*args)
...
           format: template.format,
+=======
+          format: template_format(template),
+>>>>>>> ff631047f0 (Rubocop fixes)
           locals: [],
         )
       end
     end
+<<<<<<< HEAD
+=======
```
minhngocd: left over merge conflict?

test/support/view_rendering.rb
```diff
@@ -41,11 +46,68 @@ def find_all(*args)
           template.identifier,
           EmptyTemplateHandler,
           virtual_path: template.virtual_path,
+<<<<<<< HEAD
           format: template.format,
+=======
```
minhngocd: merge conflicts

test/support/view_rendering.rb
```diff
@@ -3,8 +3,13 @@
 module ViewRendering
   extend ActiveSupport::Concern
 
+  # @!attribute [r]
+  # Returns the controller object instance under test.
```
minhngocd, line 8: left over comments?

app/helpers/admin/cancel_routes_helper.rb
```diff
@@ -1,24 +1,10 @@
...
-      case object
-      when CorporateInformationPage
-        polymorphic_path([:admin, object.owning_organisation, CorporateInformationPage])
-      when Edition
-        admin_editions_path
-      else
-        polymorphic_path([:admin, object.class])
-      end
+    if object.is_a? Edition
+      object.new_record? ? admin_editions_path : admin_edition_path(object)
```
minhngocd, line 5: hmmm i'm just wondering if we can't move "cancel_path" to the model... (might be overloading the model a bit, maybe can have a discussion). so Edition always has a "cancel_path", and corporate information page can override that, which means we don't have the change the form view either, just for the cancel link

ryanb-gds: I would actually go the other way and just call the correct method from the template, creating a new template if necessary for each usage. I didn't do that because I considered it outside of the scope of the PR, but happy to add a follow up.

minhngocd (approved):
Looks good to me!
As discussed, refactoring of the cancel link helper can be done later as part of a separate PR not related to this.

PR #8640, opened 2023-12-15
Present basic EditionableWorldwideOrgs

callumknights:
Present basic content item to Publishing-api for new EditionableWorldwideOrgs model

Trello: https://trello.com/c/dXOk7nz4/958-present-wworgs-to-the-publishing-api

### DIFF
--- a/app/models/editionable_worldwide_organisation.rb
+++ b/app/models/editionable_worldwide_organisation.rb
@@ -4,4 +4,12 @@ class EditionableWorldwideOrganisation < Edition
   def display_type_key
     "editionable_worldwide_organisation"
   end
+
+  def publishing_api_presenter
+    PublishingApi::EditionableWorldwideOrganisationPresenter
+  end
+
+  def base_path
+    "/editionable-world/organisations/#{slug}"
+  end
 end
--- a/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
@@ -1,0 +1,40 @@
+module PublishingApi
+  class EditionableWorldwideOrganisationPresenter
+    include Rails.application.routes.url_helpers
+    include ActionView::Helpers::UrlHelper
+
+    attr_accessor :item, :update_type, :state
+
+    def initialize(item, update_type: nil, state: "published")
+      self.item = item
+      self.update_type = update_type || "major"
+      self.state = state
+    end
+
+    delegate :content_id, to: :item
+
+    def content
+      content = BaseItemPresenter.new(
+        item,
+        update_type:,
+      ).base_attributes
+
+      content.merge!(
+        details: {
+          logo: {
+            crest: "single-identity",
+          },
+        },
+        document_type: item.class.name.underscore,
+        public_updated_at: item.updated_at,
+        rendering_app: Whitehall::RenderingApp::GOVERNMENT_FRONTEND,
+        schema_name: "worldwide_organisation",
+      )
+      content.merge!(PayloadBuilder::PolymorphicPath.for(item))
+    end
+
+    def links
+      {}
+    end
+  end
+end
--- a/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
@@ -34,7 +34,9 @@ module PublishingApi
     end
 
     def links
-      {}
+      {
+        sponsoring_organisations: item.organisations.map(&:content_id),
+      }
     end
   end
 end
--- a/test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
+++ b/test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
@@ -1,0 +1,42 @@
+require "test_helper"
+
+class PublishingApi::EditionableWorldwideOrganisationPresenterTest < ActiveSupport::TestCase
+  def present(...)
+    PublishingApi::EditionableWorldwideOrganisationPresenter.new(...)
+  end
+
+  test "presents a Worldwide Organisation ready for adding to the publishing API" do
+    worldwide_org = create(:editionable_worldwide_organisation)
+
+    public_path = worldwide_org.public_path
+
+    expected_hash = {
+      base_path: public_path,
+      title: worldwide_org.title,
+      schema_name: "worldwide_organisation",
+      document_type: "editionable_worldwide_organisation",
+      locale: "en",
+      publishing_app: Whitehall::PublishingApp::WHITEHALL,
+      rendering_app: Whitehall::RenderingApp::GOVERNMENT_FRONTEND,
+      public_updated_at: worldwide_org.updated_at,
+      routes: [{ path: public_path, type: "exact" }],
+      redirects: [],
+      details: {
+        logo: {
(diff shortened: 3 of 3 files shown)


test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
```diff
@@ -0,0 +1,38 @@
...
+      },
+      update_type: "major",
+    }
+
+    presented_item = present(worldwide_org)
+
+    assert_equal expected_hash, presented_item.content
+    assert_equal "major", presented_item.update_type
+    assert_equal worldwide_org.content_id, presented_item.content_id
+  end
```
brucebolt, line 37: Would it be possible to add an assertion that the presented item is valid against the schema, e.g. https://github.com/alphagov/whitehall/blob/b814a4ebd2600285ddd226ca5a135e156638199a/test/unit/app/presenters/publishing_api/news_article_presenter_test.rb#L196-L197

test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
```diff
@@ -0,0 +1,38 @@
...
+  test "presents a Worldwide Organisation ready for adding to the publishing API" do
+    worldwide_org = create(:editionable_worldwide_organisation)
+
+    public_path = worldwide_org.public_path
+
+    expected_hash = {
+      base_path: public_path,
+      title: worldwide_org.title,
+      schema_name: "worldwide_organisation",
+      document_type: "editionable_worldwide_organisation",
```
brucebolt, line 17: Has this `document_type` been defined in Publishing API schemas?

app/models/editionable_worldwide_organisation.rb
```diff
@@ -4,4 +4,20 @@ class EditionableWorldwideOrganisation < Edition
   def display_type_key
     "editionable_worldwide_organisation"
   end
+
+  def publishing_api_presenter
+    PublishingApi::EditionableWorldwideOrganisationPresenter
+  end
+
+  def base_path
+    "/editionable-world/organisations/#{slug}"
```
brucebolt, line 13: I assume we'll have to change this before these are made live to publishers?

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+            crest: "single-identity",
+          },
+        },
+        document_type: item.class.name.underscore,
+        public_updated_at: item.updated_at,
+        rendering_app: Whitehall::RenderingApp::GOVERNMENT_FRONTEND,
+        schema_name: "worldwide_organisation",
+      )
+      content.merge!(PayloadBuilder::PolymorphicPath.for(item))
+    end
```
brucebolt, line 34: Do we need to add in the analytics identifier (like [this](https://github.com/alphagov/whitehall/blob/b814a4ebd2600285ddd226ca5a135e156638199a/app/presenters/publishing_api/worldwide_organisation_presenter.rb#L42C11-L42C11) in the old presenter)?

If we can't add it yet, can you make sure there's a card written so we remember to add it in.

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def links
+      {}
```
brucebolt, line 37: We already have the organisations linked to the model [here](https://github.com/alphagov/whitehall/blob/b814a4ebd2600285ddd226ca5a135e156638199a/app/models/editionable_worldwide_organisation.rb#L2).  Maybe the organisations can already be included in the links?

brucebolt (commented):
This is looking great.  I've added a few comments, mostly just questions.

app/models/editionable_worldwide_organisation.rb
JonathanHallam: yeah, we'll also have to republish everything i think, but it'll give us a path in integration to check stuff out. We'll probably need to make some changes to government-frontend to render stuff at this path

brucebolt: I _think_ government-frontend uses the schema to decide how to render a page, rather than the route, so hopefully no changes needed there 🤞.

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
callumknights: Analytics identifier was defined in the old model, so I think it falls under the scope of the work you are doing, or at least subsequent work.

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
callumknights: Only `sponsoring_organisations` are included in the links for the original presenter, given that association doesn't exist yet that probably needs to wait.

test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
callumknights: https://github.com/alphagov/publishing-api/pull/2578

test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
callumknights: Added with a TODO note as the assertion currently fails, and wont pass till more work is done on the model

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
brucebolt: Cool.  Could you write a Trello card to add the analytics identifier, as it won't be covered under the associations card I'm currently working on.

test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
brucebolt: Could you write Trello cards for the things that will need to be added to this presenter in order to make this test pass (I think it's just `formatted_title` for now).

Also, it should be safe to add a test for the links now:

```ruby
assert_valid_against_links_schema({ links: presented_item.links }, "worldwide_organisation")
```

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
brucebolt: The wording of "Sponsoring Organisations" in the UI has changed to just be "Organisations" in the editionable workflow, which we already have as an association on the model.  So you should be able to include these as the `sponsoring_organisations`, e.g.

```ruby
    def links
      {
        sponsoring_organisations: item.organisations.map(&:content_id)
      }
    end
```

Plus a corresponding test.

app/models/editionable_worldwide_organisation.rb
```diff
@@ -4,4 +4,20 @@ class EditionableWorldwideOrganisation < Edition
...
+  end
+
+  def public_path(options = {})
```
brucebolt, line 16: These methods shouldn't be needed, since the `Edition` model has them already.

JonathanHallam (approved):
This looks good to me, happy to be able to get WWOrgs rolling a bit quicker

PR #8644, opened 2023-12-18
Fix code injection vulnerability

davidgisbey:
## Description

In the organisation roles controller #send is used directly on the Role class.

This has caused a [security alert](https://github.com/alphagov/whitehall/security/code-scanning/9)

This provides an attack vector for scripting. Particularly because it's using `#send` in place of `#public_send` so it has access to the private interface too.

There's some hidden complexity here because the type param is actually one of 6 scopes called on the Role model.

This is a pretty rough and ready fix and there are probably better solutions. I'd love to give this whole bit of the codebase a refactor as there's a bunch of weird stuff going on, but for now i think this is probably the simplest way to resolve the vulnerability.

Ideally, i think the type query param would move to the path and we could think about using constraints to match on type
https://guides.rubyonrails.org/routing.html#specifying-constraints

We could also potentially add a separate reorder endpoint for each type like we did for the [CabinetMinistersController](https://github.com/alphagov/whitehall/blob/bd0113e7737e1eef73050d808b052e0cf1212f08/app/controllers/admin/cabinet_ministers_controller.rb#L4). But i think that's probably a bit beyond the scope of this investigation & fix.

## Link to security issues for this PR 

You can see that the alert has been resolved by this PR here 

https://github.com/alphagov/whitehall/security/code-scanning?query=is%3Aopen+pr%3A8644

### DIFF
--- a/app/controllers/admin/organisation_people_controller.rb
+++ b/app/controllers/admin/organisation_people_controller.rb
@@ -4,17 +4,23 @@ class Admin::OrganisationPeopleController < Admin::BaseController
 
   def index
     @render_reorder = can?(:edit, @organisation)
-    @ministerial_organisation_roles = organisation_roles(:ministerial)
-    @management_organisation_roles = organisation_roles(:management)
-    @traffic_commissioner_organisation_roles = organisation_roles(:traffic_commissioner)
-    @military_organisation_roles = organisation_roles(:military)
-    @special_representative_organisation_roles = organisation_roles(:special_representative)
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
Ah, another great Whitehall example of where we should compose multiple API endpoints out of lower level primitives, rather than use inheritance to make a single API endpoint work for several different use cases. This will do very nicely for now though 👏

davidgisbey:
Thanks for the review!

PR #8645, opened 2023-12-18
Remove access limited from edition rake task

lauraghiorghisor-tw:
There have been cases of users accidentally removing their own access by not setting the correct organisations, thus making it impossible for themselves to publish. Adding rake task to temporarily remove access limited from editions so that the users can fix the organisation.

[Trello card](https://trello.com/c/DKQPwuNN/2273-add-a-rake-task-to-disable-access-limiting-for-an-edition)


beccapearce (approved):
Rake task looks good! I won't comment on the documentation as I had a hand in it!

ryanb-gds:
Thanks for this @lauraghiorghisor-tw , great to have this feature documented 🤓

PR #8647, opened 2023-12-19
Add seed data for Editionable Worldwide Organisations

brucebolt:
This should make it easier for us to test usability in the development environment by giving us some existing seed data.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -196,6 +196,9 @@ en:
       edition_with_appointment:
         one: Edition with appointment
         other: Editions with appointment
+      editionable_worldwide_organisation:
+        one: Worldwide organisation
+        other: Worldwide organisations
       equality_and_diversity:
         one: Equality and diversity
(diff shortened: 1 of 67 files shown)


config/locales/en.yml
```diff
@@ -196,6 +196,9 @@ en:
       edition_with_appointment:
         one: Edition with appointment
         other: Editions with appointment
+      editionable_worldwide_organisation:
```
JonathanHallam, line 199: Can you run the rake task so these are in all the locales?

brucebolt: Done.  Involved quite a bit of hacking as the rake task doesn't correctly populate the plural rules 😬.

JonathanHallam: I'm assuming that's why some are `one, other` and others have a load more numbers?

JonathanHallam (approved):
👍 thanks

PR #8653, opened 2023-12-19
Factor out opening and closing date behaviour into a concern

ryanb-gds:
Both consultations and calls for evidence have opening and closing dates. This commit factors out the duplication into a reusable concern which might potentially be used for other content types requiring this behaviour in the future.


beccapearce (approved):
Nice one!!!!

PR #8655, opened 2023-12-19
Add role association to Editionable Worldwide Organisations

brucebolt:
This allows Roles to be assigned to Editionable Worldwide Organisations during the edit workflow.

This is a change from the current behaviour, as the worldwide organisations are currently assigned on the role's edit page (rather than the organisation's edit page)

By moving this to be the other way round (i.e. the edit is made on the worldwide organisation itself), the option is more obvious to the user and allows the change to be made in advance (e.g. as part of scheduled publishing).

Screenshot:
Screenshot showing the roles field as a multi-select.  A role ("Prime Minister") has already been selected.

[Trello card](https://trello.com/b/G9rDOoC9)

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -234,6 +234,7 @@ private
         lead_organisation_ids: [],
         supporting_organisation_ids: [],
         organisation_ids: [],
+        role_ids: [],
         world_location_ids: [],
         worldwide_organisation_ids: [],
         related_policy_ids: [],
--- a/app/helpers/admin/taggable_content_helper.rb
+++ b/app/helpers/admin/taggable_content_helper.rb
@@ -93,6 +93,14 @@ module Admin::TaggableContentHelper
     end
   end
 
+  # Returns an Array that represents the taggable roles. Each element of the
+  # array consists of two values: the role name and its ID
+  def taggable_roles_container
+    Rails.cache.fetch(taggable_roles_cache_digest, expires_in: 1.day) do
+      Role.order(:name).map { |w| [w.name, w.id] }
+    end
+  end
+
   # Returns an Array that represents the taggable alternative format providers.
   # Each element of the array consists of two values: the label (organisation
   # and the email address if avaiable) and the ID of the organisation.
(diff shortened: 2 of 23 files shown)


app/helpers/admin/taggable_content_helper.rb
```diff
@@ -93,6 +93,14 @@ def taggable_world_locations_container
     end
   end
 
+  # Returns an Array that represents the taggable roles. Each element of the
```
JonathanHallam, line 96: Absolutely not a complaint, but this file uses a lot of comments and we tend to not use them, any idea why?

app/controllers/admin/editions_controller.rb
```diff
@@ -234,6 +234,7 @@ def permitted_edition_attributes
         lead_organisation_ids: [],
```
JonathanHallam, line 234: Again, not complaining, but this implies that no editionable content types can have associated roles. Definitely another chunk of Whitehall weirdness

app/helpers/admin/taggable_content_helper.rb
brucebolt: I've no idea as I just copied this from another example.  I'm tempted to remove all the comments completely, as I don't think the code is complicated enough to need an explanation.

JonathanHallam (approved):
Cracking! Couple of questions but happy with code

app/controllers/admin/editions_controller.rb
brucebolt: I think this just implies that the `roles_ids` should be an array, rather than a string or anything else.

JonathanHallam: i more mean this commit in general, like you've had to add the role tagging functionality from scratch to editions because it wasn't there already. Basically I'm just thinking about code that can be removed once (if ever) all types are editionable

brucebolt: Ah, yes, I see.

We could associate with role appointments (e.g. on speeches, where we want to say the name of the actual person who gave a speech and what their role was at the time), but that wouldn't work in this case as we want the organisation associated with the role (otherwise it'll need to be updated manually each time someone new is appointed).

I'm hoping this code may be needed for other things we could make editionable (e.g. normal organisations).

app/helpers/admin/taggable_content_helper.rb
brucebolt: I'll do that as a separate PR.

JonathanHallam: I guess if you do that it'll be one to ask publishing experience about just to make sure they're happy and there's not a bigger reason that they're here

PR #8656, opened 2023-12-19
Ensure document retains slug when edition title has special chars

davidgisbey:
## Description

We've been doing some investigative work into a bunch of sentry errors. I've gone down a bit of a rabbit hole and came across some fairly strange behaviour.

There are a subset of documents in our database that don't have a slug. This is despite their being an after create for document which sets the slug to the documents id if one is not set based on the title of the first edition for whatever reason.

This is being overridden by the #update_slug_if_possible method which is triggered by the #update_document_slug before_validation method. This updates the slug to the downcased and underscored edition title.

The title is passed to the #update_slug_if_possible as an arg. If it contains special character or characters from other languages with different alphabets friendly_id  updates the documents slug to nil.

We want to retain the default behaviour of the slug being defaulted to the id of the document. This commit ensures that's the case.

## Trello card

https://trello.com/c/ocdba6KI/2254-investigate-gdsapi-httpunprocceessibleentity-errors-on-sentry

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -86,6 +86,11 @@ class Document < ApplicationRecord
     candidate_slug = normalize_friendly_id(new_title)
     unless candidate_slug == slug
       update!(sluggable_string: new_title)
+      # when special characters or scipts are used from the non-latin alphabets
+      # friendly_id sets the documents slug to nil. This ensures that it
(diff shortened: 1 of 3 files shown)


beccapearce (approved):
Looks good!

PR #8665, opened 2023-12-21
Removed unnecessary cancel routes helper

ryanb-gds:
Trying to have a single cancel routes helper for every use case is adding unnecessary complexity. Simply using the appropriate route helper for each template reduces indirection.


minhngocd (approved):
looks good to me ! thanks for doing this Ryan!

PR #8667, opened 2023-12-21
Revert Rails 7.1 upgrade

ryanb-gds:
We're rolling back because the way we are importing our secrets into the Rails credentials object doesn't seem to be working. We started seeing that the GOV.UK Notify API client was not configured because the notify API key was not set.

Related Sentry error: https://govuk.sentry.io/issues/4744278800/events/63c8413b0aeb4eef8797fb02bac1d9d8/


davidgisbey (approved):
lgtm

PR #8668, opened 2023-12-21
Retry Rails 7.1 upgrade

ryanb-gds:
We had to revert this because of an issue relating to loading environment variables into the Rails credentials object


ryanb-gds:
Going to merge this after Xmas now as I don't think 2nd line would be that happy with us if we merged such a risky change on the last day before we break up for Xmas!

minhngocd (approved):

PR #8669, opened 2023-12-21
Fix Document Collections to only search for published editions

minhngocd:
When we introduced pagination, we inadvertently removed the scope of published editions. In addition to putting the scope back to how it was, we also removed the filtering since we didn't need to do any other filtering other than search via title. This should simplify both the code and performance. Hopefully the scoping will also be helpful for search performance in production as well.

https://trello.com/c/fEbWtdv8/2280-document-collections-add-by-title-scoping-to-only-published-editions

### DIFF
--- a/app/controllers/admin/document_collection_group_document_search_controller.rb
+++ b/app/controllers/admin/document_collection_group_document_search_controller.rb
@@ -18,30 +18,17 @@ class Admin::DocumentCollectionGroupDocumentSearchController < Admin::BaseContro
 
   def add_by_title
     flash.now[:alert] = "Please enter a search query" if params[:title] && params[:title].empty?
-    @editions = filter.editions if params[:title].present?
+    if params[:title].present?
+      results = Edition.published.with_title_containing(params[:title].strip)
+      @editions = results
+                    .page(params[:page])
+                    .per(Admin::EditionFilter::GOVUK_DESIGN_SYSTEM_PER_PAGE)
+    end
   end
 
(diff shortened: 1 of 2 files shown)


lauraghiorghisor-tw (approved):
lgtm :)

app/controllers/admin/document_collection_group_document_search_controller.rb
```diff
@@ -18,31 +18,18 @@ def search
 
   def add_by_title
     flash.now[:alert] = "Please enter a search query" if params[:title] && params[:title].empty?
-    @editions = filter.editions if params[:title].present?
+    if params[:title].present?
+      results = Edition.published.with_title_containing(params[:title].strip)
+      @editions = results
```
ryanb-gds, line 23: Just out of interest, did you give this a go without calling page in the controller? I'm still curious about whether we actually need to make the pagination call ourselves or whether we can pass it to that Kaminari pagination helper and it'll automatically do it

ryanb-gds (approved):
I like this

app/controllers/admin/document_collection_group_document_search_controller.rb
minhngocd: i have not! can definitely give it a go

minhngocd: tested - does not work - we need to do pages and params according to the kaminari gem i believe

PR #8684, opened 2023-12-29
Editionable worldwide organisation logo formatted name

jkempster34:
https://trello.com/c/12wQrTlH

### Add title method to worldwide organisation
The new editionable worldwide organisation model uses the `title` method
instead of `name`.

This adds `title` to the legacy worldwide organisation model so that views and
helpers etc. can be compatible with both model types.

### Use the `worldwide_organisation` document type
We don't need a new document type here, we can use the existing one.

### Add logo formatted name to editions
This is required for editionable worldwide organisations.

### Present logo formatted title for worldwide organisations
These have been broken for some time. They remained broken when the rendering
of these pages was moved to Government Frontend.

Presenting these correctly will allow us to fix them on the frontend and
restore their proper functionality.

### Present logo formatted name for editionable worldwide organisations
The presenter isn't currently valid against the schema because it's missing the
logo formatted name.

### Add logo formatted name textarea for editionable worldwide organisations
This allows users to add a logo formatted name for the new editionable
worldwide organisations.

image

---

### DIFF
--- a/app/models/worldwide_organisation.rb
+++ b/app/models/worldwide_organisation.rb
@@ -32,6 +32,7 @@ class WorldwideOrganisation < ApplicationRecord
   translates :name
 
   alias_method :original_main_office, :main_office
+  alias_method :title, :name
 
   validates_with SafeHtmlValidator
   validates :name, presence: true
--- a/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
@@ -25,7 +25,7 @@ module PublishingApi
             crest: "single-identity",
           },
         },
-        document_type: item.class.name.underscore,
+        document_type: "worldwide_organisation",
         public_updated_at: item.updated_at,
         rendering_app: Whitehall::RenderingApp::GOVERNMENT_FRONTEND,
         schema_name: "worldwide_organisation",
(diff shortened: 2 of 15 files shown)


app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
```diff
@@ -23,9 +25,10 @@ def content
         details: {
           logo: {
             crest: "single-identity",
+            formatted_title: worldwide_organisation_logo_name(item),
           },
         },
-        document_type: item.class.name.underscore,
+        document_type: "worldwide_organisation",
```
jkempster34, line 31: Is this okay? I was in the middle of changing Publishing API, and I wondering if we actually needed these to be different? We can distinguish the old ones from new through the base path instead?

jkempster34: We could revert this if this is okay https://github.com/alphagov/publishing-api/pull/2578

If not, happy to add in the `editionable_worldwide_organisation` to the schema types.

app/models/worldwide_organisation.rb
```diff
@@ -72,6 +72,10 @@ def sponsoring_organisation
              content: :summary,
              format: "worldwide_organisation"
 
+  def title
+    name
+  end
```
brucebolt, line 77: Could this be done with `alias_method`?  E.g. `alias_method :title :name`.

app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
brucebolt: It's probably fine to keep both editionable and non-editionable as having the same document type, since the same code will be rendering both versions.

brucebolt (approved):
This all looks good to me 👍.  Just added a couple of non-blocking comments.

app/models/worldwide_organisation.rb
jkempster34: I'm going to leave this as is because the `display_name` method below follows the same format

PR #8689, opened 2024-01-02
Add an index to the asset manager ID column on the assets table

ryanb-gds:
This is a re-opening of #8672. I initially closed that PR because I thought it might be possible to make the index unique, and to change the asset deletion code to load a single record. However, I have discovered that in some cases multiple asset records share an Asset Manager ID. There was also a question about whether adding an index would deliver the expected performance improvements, but I think it will if I have read the MySQL docs correctly.

Here is the explanation for this PR:

The lack of an index on the asset manager ID column was causing lock timeout errors occasionally during concurrent delete operations.

The asset deletion worker runs a query to delete all assets matching a particular ID in asset manager. For delete queries, InnoDB will issue a lock for each record that is scanned, regardless of whether or not that record will subsequently be deleted. Consequently, adding an index to reduce the number of records scanned will reduce the risk of locks caused by concurrent deletes.

Trello: https://trello.com/c/13dDUcoj

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_12_28_163205) do
+ActiveRecord::Schema[7.1].define(version: 2024_01_02_114732) do
   create_table "assets", charset: "utf8mb3", force: :cascade do |t|
     t.string "asset_manager_id", null: false
     t.string "variant", null: false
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -77,7 +78,7 @@
     t.string "email"
     t.integer "call_for_evidence_response_form_id"
     t.text "postal_address"
-    t.index ["call_for_evidence_response_form_id"], name: "index_cfes_participations_on_cfes_response_form_id"
+    t.index ["call_for_evidence_response_form_id"], name: "index_cons_participations_on_cons_response_form_id"
```
davidgisbey, line 81: Seems unrelated but is probs fine?

davidgisbey (approved):

db/schema.rb
minhngocd: hmmm does seem unrelated but why has it been renamed from cfe to cons (surely cfe's is the correct one) 🤔

ryanb-gds: Ah it's not fine, it's something weird going on with my local DB. Going to fix this

PR #8691, opened 2024-01-02
Editionable worldwide organisation offices

jkempster34:
https://trello.com/c/y8RHsxZ1

https://github.com/alphagov/whitehall/assets/47089130/32f54941-84e8-40bb-8e13-589660dec024

### Move main office actions into their own controller
Moving these actions into their own controller will make the migration to
editionable worldwide organisations easier, it also follows the best practice
of not creating controller actions outside of the default 7
http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/.

### Copy office related methods to editionable worldwide organisation
These methods are needed to handle the association between worldwide
organisations and worldwide offices. As part of the creation of the new
editionable worldwide organisation model, we need to copy them over.

### Use `title` for the worldwide office views
Now that we have a title method for the legacy worldwide organisation model, we
can make these view compatible with both the legacy model and the new
editionable model.

### Use `title` in the worldwide office presenter
Now that we have a title method for the legacy worldwide organisation model, we
can use in in the  worldwide office presenter so that it is compatible with
both worldwide organisations and editionable worldwide organisations.

### Make the worldwide offices work for editionable worldwide organisations
This modifies the existing worldwide offices controller and model so that they
are compatible with editionable worldwide organisations as well as legacy
worldwide organisations.

Note: When the editionable worldwide organisations feature flag is enabled,
performing actions related to the worldwide offices of legacy worldwide
organisations will not function (and vice versa). This should be okay as we
don't want these features running side-by-side. The next step will be to hide
the legacy actions when the flag is enabled for an extra level of safety.

### Validate that an office can't have both and organisation and edition
This model is temporarily supporting both the legacy worldwide organisations as
well as the new editionable worldwide organisations. We don't want to be in a
position where an office is associated with both.

### Add offices to the editionable worldwide organisation API presenter
Now that we can associate editionable worldwide organisations with offices, we
can add them to the Publishing API presenter.

---


test/unit/app/models/editionable_worldwide_organisation_test.rb
```diff
@@ -70,8 +88,8 @@ class EditionableWorldwideOrganisationTest < ActiveSupport::TestCase
     worldwide_organisation.roles << create(:ambassador_role, :vacant)
     worldwide_organisation.roles << create(:deputy_head_of_mission_role, :vacant)
     worldwide_organisation.roles << create(:worldwide_office_staff_role, :vacant)
-
     assert_nil worldwide_organisation.primary_role
+
```
brucebolt, line 91: Minor nitpick: this line break probably wasn't intended to be moved.

app/models/editionable_worldwide_organisation.rb
```diff
@@ -19,6 +19,39 @@ def display_type_key
     "editionable_worldwide_organisation"
   end
 
+  alias_method :original_main_office, :main_office
+
+  # I'm trying to use a domain centric design rather than a persistence
+  # centric design, so I do not want to expose a has_many :home_page_lists
+  # and all that this implies. I really only want to expose a list of
+  # contacts (in order) that should be shown on the home page, and some
+  # simple (explicit) methods for manipulating them.
```
brucebolt, line 28: It's probably not worth copying this comment over, since this code has existed for a long time and I'm not convinced it adds any additional information.

brucebolt (approved):
Great work on this!

I think reusing the existing controller is fine here, since we'll never have a situation where we need to support offices associated to both non-editionable and editionable content.

brucebolt:
Oh, one other thing: would you be able to add some seed data?

test/unit/app/models/editionable_worldwide_organisation_test.rb
jkempster34: Apologies, this will be a result of rebasing 400 times!

PR #8693, opened 2024-01-03
Add analytics identifier for Editionable Worldwide Organisations

brucebolt:
This adds analytics identifiers to Editionable Worldwide Organisations, in the same way as they are associated to editionable Worldwide Organisations.

[Trello card](https://trello.com/c/zITYCrxl)

### DIFF
--- a/app/models/editionable_worldwide_organisation.rb
+++ b/app/models/editionable_worldwide_organisation.rb
@@ -8,6 +8,9 @@ class EditionableWorldwideOrganisation < Edition
   include Edition::Roles
   include Edition::WorldLocations
 
+  include AnalyticsIdentifierPopulator
+  self.analytics_prefix = "WO"
+
   def base_path
     "/editionable-world/organisations/#{slug}"
   end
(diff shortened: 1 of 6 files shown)


app/models/editionable_worldwide_organisation.rb
```diff
@@ -8,6 +8,9 @@ class EditionableWorldwideOrganisation < Edition
   include Edition::Roles
   include Edition::WorldLocations
 
+  include AnalyticsIdentifierPopulator
```
jkempster34, line 11: I'm assuming we are going to copy these over from the existing orgs, which will mean they are out of sync with the model ID. Shouldn't be a problem, but thought it was worth mentioning.

jkempster34 (approved):

app/models/editionable_worldwide_organisation.rb
brucebolt: Yes, we will copy them.  My (hopefully not too dangerous) assumption is that the existing Worldwide Organisation IDs are considerably lower than the next Edition ID, so there will never be any duplication of numbering.

PR #8696, opened 2024-01-03
Move concerns into concerns directory

davidgisbey:
## Description

We've got a lot of concerns in our model directories. These should really live in the concerns folder in their own subdirectories when appropriate so it's clear which are concerns and included in other models and which are base models

This moves all the concerns in the models to the concerns folder.

Top level concerns are model agnostic sit on the top level and concerns which are model specific sit within the relevant subdir 


ryanb-gds (approved):
Sorry, missed reviewing this last week

PR #8704, opened 2024-01-05
Allow for nil image

beccapearce:
We have had cases where the featured item of an organisation returns nil, this alerts us via sentry and is not an immediately actionable alert. This changes the code to allow for a nil image once it has reached this point so we are not alerted unduly

Trello card: https://trello.com/c/prZOCIqa/2255-investigate-no-method-errors-on-sentry


ryanb-gds (approved):
🎉

PR #8709, opened 2024-01-08
Add emergency banner publishing

brucebolt:
The emergency banner is currently published and unpublished using [a rake task in the static application](https://github.com/alphagov/static/blob/6d606f04a7b09ae57008b528eb7e30652d40e287/lib/tasks/emergency_banner.rake#L4-L18). 

This adds a user interface, so the banner can be deployed without needing developer intervention.

I've avoided adding any models (as this complicated things where we're not using the same data store for this as the rest of Whitehall's data) so needed to add a tiny bit of custom error handling into the controller.

This can be made compliant with [RFC-143](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md) by simply adding a second Redis adapter, when we get round to splitting the shared Redis instance by app.

## Screenshots

| Action | Screenshot |
| -- | -- |
| No emergency banner currently deployed | Screenshot 2024-01-09 at 09 47 58 |
| Adding or editing the banner | Screenshot 2024-01-09 at 09 48 12 | 
| After editing the banner | Screenshot 2024-01-09 at 09 48 45 | 
| Removing the banner | Screenshot 2024-01-09 at 09 48 51 |

### DIFF
--- a/test/functional/admin/emergency_banner_controller_test.rb
+++ b/test/functional/admin/emergency_banner_controller_test.rb
@@ -11,12 +11,18 @@ class Admin::EmergencyBannerControllerTest < ActionController::TestCase
 
   should_be_an_admin_controller
 
-  %i[show].each do |action_method|
+  %i[edit show].each do |action_method|
     test "#{action_method} action is not permitted to non-GDS admins" do
       login_as :departmental_editor
       get action_method
       assert_response :forbidden
     end
+  end
+
+  test "update action is not permitted to non-GDS admins" do
+    login_as :departmental_editor
+    patch :update
+    assert_response :forbidden
   end
 
   context "when the banner is disabled" do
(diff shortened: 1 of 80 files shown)


test/functional/admin/emergency_banner_controller_test.rb
```diff
@@ -0,0 +1,70 @@
+require "fakeredis/minitest"
+require "test_helper"
+
+class Admin::EmergencyBannerControllerTest < ActionController::TestCase
+  extend Minitest::Spec::DSL
+
+  setup do
+    login_as :gds_admin
+    @controller = Admin::EmergencyBannerController.new
```
davidgisbey, line 9: is this needed?

test/functional/admin/emergency_banner_controller_test.rb
```diff
@@ -11,7 +11,7 @@ class Admin::EmergencyBannerControllerTest < ActionController::TestCase
 
   should_be_an_admin_controller
 
-  %i[show].each do |action_method|
+  %i[edit show update].each do |action_method|
     test "#{action_method} action is not permitted to non-GDS admins" do
       login_as :departmental_editor
       get action_method
```
davidgisbey, line 17: Not the end of the world but does this need to be a PATCH for update? I'm assuming this is hitting the same endpoint as show atm.

davidgisbey (approved):
Amazing work! I've left a couple of fairly nitty comments but no big blockers

test/functional/admin/emergency_banner_controller_test.rb
brucebolt: Unfortunately, yes.  As the context blocks obscure the name of the test, minitest doesn't attempt to automatically decide which controller to use.

test/functional/admin/emergency_banner_controller_test.rb
brucebolt: Ooh, nice spot.  I'll update that.

brucebolt: And I'll do the same for the `destroy` action.

robinjam:
Amazing! 🌟 

If you feel like working on this some more, I think the validation & persistence logic could live in an [ActiveModel](https://guides.rubyonrails.org/active_model_basics.html) model (i.e. not backed by a database table). That would allow you to use the standard Rails validation logic instead of hardcoding stuff in the controller.

(It's probably fine like this though, if you want to leave it)

PR #8717, opened 2024-01-09
Full text document title search

ryanb-gds:
### What

This is an initial attempt to apply a full text index to Whitehall's document search.

### Why

Whitehall's document search is extremely slow if we are not able to use an index. This change will make it possible to search documents quickly using a full text index on the document title.

### Problems

It isn't possible to write automated tests for code that uses the fulltext index without disabling the automatic transaction wrapping behaviour. Currently I am working around that by using a feature flag to swap out the implementation when unit testing, but we wouldn't be able to remove the feature flag without rewriting a lot of the test code for tests such as test/functional/admin/document_collection_group_document_search_controller_test.rb. I don't imagine stubbing the scope would be very easy.

Trello: https://trello.com/c/Qkl62e4y

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -76,9 +76,14 @@ class Edition < ApplicationRecord
           like_clause = "%#{escaped_like_expression}%"
 
           scope = in_default_locale.includes(:document)
-          scope
-            .where("edition_translations.title LIKE :like_clause", like_clause:)
-            .or(scope.where(document: { slug: keywords }))
+
+          if Flipflop.fulltext_document_search?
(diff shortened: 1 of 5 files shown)


davidgisbey (approved):
code looks good to me. I have a question around using the feature flag long term but i'll ask in slack

ryanb-gds:
Thanks @davidgisbey. I'm actually going to close this because we don't have a clear path to production at the moment. I don't think we can introduce this without redesigning the documents search page to have the items ordered by relevance, which needs some user research. I don't want the unused code hanging around until we get a chance to do that though.

PR #8725, opened 2024-01-11
Inform the user that an invalid last updated date filter has not been applied

lauraghiorghisor-tw:
There was no validation on last updated `from_date` and `to_date` filters. As a result, the user was redirected to session stored filters or default filter (organisation and active state), without being notified that their filtering was not successful.

Adding a validation check and notice to inform them that the filter has not been applied.

[Trello card](https://trello.com/c/dPoVTx0z/2272-improve-handling-of-invalid-date-format-in-whitehall-document-last-updated-date-filter-input)


ryanb-gds (approved):
Looks good to me

PR #8727, opened 2024-01-11
Add new link to footer for the GOVUK Status page

davidtrussler:
As part of some associated work during GIFT week we would like to add a new link to the footer in Whitehall. This is to give users a link to the GOVUK Status Page so that they can see if there are any incidents or scheduled maintenance on the publishing apps that might effect them. We have discussed the change with Darren and Ruth in the Publishing Experience Team. 

|Current|Updated|
|-|-|
|Screenshot 2024-01-11 at 12 22 12|Screenshot 2024-01-11 at 12 31 01|


ryanb-gds (approved):
Sounds reasonable to me!

PR #8748, opened 2024-01-19
Inline the updating of asset redirects

ryanb-gds:
### What

This PR inlines the code for updating the redirect_url for each asset to make it easier to follow rather than having to read multiple files.

Previously the AttachmentUpdater was responsible for updating the redirect URL for each asset, but the abstraction wasn't providing much value.

### Why

I was trying to look into https://govuk.sentry.io/issues/4725413878/events/00f57c264fcf41b0ac7225047e56ef76/ and finding it difficult to follow the code. Reducing the depth of the stack by one level helps make it a bit easier to understand what it going on.


dnkrj (approved):
Nice!

PR #8749, opened 2024-01-22
Editionable worldwide organisation corporate information pages

jkempster34:
https://trello.com/c/MUhuggob

**This is part of the work to make worldwide organsiations editionable**

### Organisation document with CIPs

image

### Add title method to organisation model
The new editionable worldwide organisation model uses the title method
instead of name.

This adds title to the organisation model so that views and helpers etc. can be
compatible with both model types.

### Use title for the corporate information page views
Now that we have a title method for the organisation model, we can make these
views compatible with the worldwide organisation model, the editionable
worldwide organisation model, and the organisation model.

### Use title for the corporate information page type
This class must be compatible with (currently) worldwide organisations,
editionable worldwide organisations, and organisations.

This uses `title` instead of `name` as all of the above classes now have a
`title` method.

### Use title method for the search title
This must be compatible with (currently) worldwide organisations,
editionable worldwide organisations, and organisations.

This uses `title` instead of `name` as all of the above classes now have a
`title` method.

### Add association between document and CIPs
For now, the simplest way of associating an edition (corporate information
page) with another edition (editionable worldwide organisation) is to
associate it with its document. This means that users can add and edit
corporate information pages no matter what state the owning worldwide
organisation is in.

This adds an association between corporate information pages and document,
using a join table to avoid adding more columns to the document or edition
tables

### Add `Edition::CorporateInformationPages` module
This adds a new module for editions which are related to corporate informaiton
pages.

It it derived from the `HasCorporateInformationPages` module, but does not
contain the "about" page functionality.

### Only one of each CIP can be created for an editionable worldwide org
An editionable worldwide organisation should only have (at most) one of each
corporate information page type.

### Remove option to create an about us page for editionable worldwide orgs
Part of the move from using the legacy worldwide organisations to using the new
editionable worldwide organisations involves removing the "about us" corporate
information page.

Instead, this information will live on the editionable worldwide organisation
itself.

This adds a class method to the corporate information types so we can return
everything except the about page. Organisations (non-worldwide) however, still
need the about page.

### Make `delete_all_attachments` method more generic
We want to hook into the end of a deletion so that we can remove all of the
corporate informaiton pages associated with an editionable worldwide
organisation when it is deleted.

This gives the `delete_all_attachments` a more generic name so that we can use
that.

### Make the CIP pages work for editionable worldwide organisations
This modifies the existing corporate information controller and model so that
they are compatible with editionable worldwide organisations as well as legacy
worldwide organisations.

The owning organisation of a CIP must be an editionable worldwide organisation
and is used for the base path of the CIP, we fetch the live edition if
available, or the most recent version otherwise.

Note: When the editionable worldwide organisations feature flag is enabled,
performing actions related to the corporate information pages of legacy
worldwide organisations will not function (and vice versa). This should be okay
as we don't want these features running side-by-side. The feature flag hides
the legacy worldwide organisations for an extra level of safety.

### Include CIPs in the editionable worldwide organisation presenter
We want to the result of this API presenter to equal the result of the (now
legacy) worldwide organisation API presenter.

Now that editionable worldwide organisations have associated corporate
information pages, this copies over the related links and details.

### Redirect to the edition page when editionable worldwide organisation
We want to redirect back to the edition page of the editionable worldwide
organisation when a corporate informaiton page is deleted as we aren't using
the index page for these.

### Use minimal table for document corporate information pages
We want a list of corporate information pages related to a document, but
without the pagination and extra detail.

This moves the table itself into a partial and uses this for document corporate
information pages.

### Validate owning organisation has published edition when publishing CIP
A published corporate information page shouldn't exist if it doesn't have a
pbulished worldwide organisation.

---

### DIFF
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -149,7 +149,7 @@ class CorporateInformationPage < Edition
   end
 
   def title_prefix_organisation_name
-    [owning_organisation.name, title].join(" \u2013 ")
+    [owning_organisation.title, title].join(" \u2013 ")
   end
 
   def title(_locale = :en)
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -17,6 +17,7 @@ class CorporateInformationPage < Edition
   validate :unique_organisation_and_page_type, on: :create, if: :organisation
   validate :unique_worldwide_organisation_and_page_type, on: :create, if: :worldwide_organisation
   validate :unique_editionable_worldwide_organisation_and_page_type, on: :create, if: :editionable_worldwide_organisation
+  validate :owning_organisation_has_published_edition, on: :publish, if: :editionable_worldwide_organisation
 
   add_trait do
     def process_associations_before_save(new_edition)
@@ -273,6 +274,12 @@ private
 
     if duplicate_scope.exists?
       errors.add(:base, "Another '#{display_type_key.humanize}' page was already published for this worldwide organisation")
+    end
+  end
+
+  def owning_organisation_has_published_edition
+    if owning_organisation.document.live_edition.blank?
+      errors.add(:base, "The owning organisation must be published before the corporate information page can be published")
     end
   end
 end
--- a/app/models/corporate_information_page.rb
+++ b/app/models/corporate_information_page.rb
@@ -16,6 +16,7 @@ class CorporateInformationPage < Edition
   delegate :slug, :display_type_key, to: :corporate_information_page_type
   validate :unique_organisation_and_page_type, on: :create, if: :organisation
   validate :unique_worldwide_organisation_and_page_type, on: :create, if: :worldwide_organisation
+  validate :unique_editionable_worldwide_organisation_and_page_type, on: :create, if: :editionable_worldwide_organisation
 
   add_trait do
     def process_associations_before_save(new_edition)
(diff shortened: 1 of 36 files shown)


app/models/document_corporate_information_page.rb
```diff
@@ -0,0 +1,6 @@
+class DocumentCorporateInformationPage < ApplicationRecord
```
brucebolt, line 1: Nice idea using a link table here.

app/models/corporate_information_page.rb
```diff
@@ -244,4 +245,20 @@ def unique_worldwide_organisation_and_page_type
       errors.add(:base, "Another '#{display_type_key.humanize}' page was already published for this worldwide organisation")
     end
   end
+
+  def unique_editionable_worldwide_organisation_and_page_type
+    duplicate_scope = CorporateInformationPage
+                        .joins(:document_corporate_information_page)
+                        .where("document_corporate_information_pages.owning_document_id = ?", document_corporate_information_page.owning_document_id)
+                        .where(corporate_information_page_type_id:)
+                        .where("state not like 'superseded'")
```
brucebolt, line 269: Could the SQL be upper cased, just to make it clearer this is SQL and not some rails thing, e.g. `.where("state NOT LIKE 'superseded'")`?

app/models/corporate_information_page_type.rb
```diff
@@ -8,6 +8,12 @@ def self.find(slug)
     all.detect { |type| type.slug == slug } || raise(ActiveRecord::RecordNotFound)
   end
 
+  def self.for(organisation)
+    return all.reject { |page| page.slug == "about" } if organisation.is_a?(EditionableWorldwideOrganisation)
+
+    all
```
brucebolt, line 14: Would this be better as an if statement?  E.g.
```
if organisation.is_a?(EditionableWorldwideOrganisation)
  all.reject { |page| page.slug == "about" }
else
  all
end
```

brucebolt (approved):
This all makes sense to me.  Great work seeing this through all the complications!

I've added a few non-blocking comments.

brucebolt (commented):
Could you add an example of a CIP associated with an editionable worldwide organisation to the seeds file.

jkempster34:
We've (Publishing Platform team) decided to take a different approach and represent CIPs as multi-part content instead.

app/models/document.rb
```diff
@@ -33,6 +33,8 @@ class Document < ApplicationRecord
   has_many :document_collection_groups, through: :document_collection_group_memberships
   has_many :document_collections, through: :document_collection_groups
   has_many :features, inverse_of: :document, dependent: :destroy
+  has_many :document_corporate_information_pages, foreign_key: :owning_document_id
+  has_many :corporate_information_pages, through: :document_corporate_information_pages, class_name: "CorporateInformationPage"
```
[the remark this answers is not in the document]
jkempster34: This particular issue could be resolved (I think) by following a similar pattern to the one here https://github.com/alphagov/whitehall/blob/38b38dfcb306ce1ff0a25717cd3fe15d37d55f58/app/models/concerns/edition/editionable_worldwide_organisations.rb

PR #8771, opened 2024-01-26
Remove raising behaviour from PersonPresenter

lauraghiorghisor-tw:
The raising behaviour was previously introduced to get a stack trace to help debug broken Person urls reported in prod. 

Narrowed down the error to the filename check with an action point to reconsider that logic and make it
not rely on filename.

Adding a log to track this in the future.

[Trello card](https://trello.com/c/COWkIRaH/2312-investigate-cause-of-carrierwave-path-in-person-image)

### DIFF
--- a/app/presenters/publishing_api/person_presenter.rb
+++ b/app/presenters/publishing_api/person_presenter.rb
@@ -35,8 +35,7 @@ module PublishingApi
       details_hash = {}
 
       if item.image&.all_asset_variants_uploaded? && item.image&.url(:s465)
-        raise "person image path should not use carrierwave path" if item.image.url(:s465).include?("carrierwave-tmp")
-
+        logger.error("PersonPresenter: Person of ID##{item.id} has image with url '#{item.image&.url(:s465)}'") if item.image.url(:s465).include?("carrierwave-tmp")
         details_hash[:image] = { url: item.image.url(:s465), alt_text: item.name }
       end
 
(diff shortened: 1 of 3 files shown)


app/presenters/publishing_api/person_presenter.rb
```diff
@@ -35,8 +35,7 @@ def details
       details_hash = {}
 
       if item.image&.all_asset_variants_uploaded? && item.image&.url(:s465)
-        raise "person image path should not use carrierwave path" if item.image.url(:s465).include?("carrierwave-tmp")
-
+        logger.info("PersonPresenter: Person of ID##{item.id} has image with url '#{item.image&.url(:s465)}'")
```
ryanb-gds, line 38: I was imagining we'd log only if the invalid carrierwave temporary path is used? I'm not sure the output is that interesting otherwise. Also in that case I think we'd log at the error level.

test/unit/app/presenters/publishing_api/person_presenter_test.rb
```diff
@@ -119,15 +119,14 @@ def present(...)
...
-    person.image = image
 
-    assert_raises do
-      present(person).content
-    end
+    image.expects(:url).times(3).with(:s465).returns("/uploads/carrierwave-tmp/1704204355-973510646575090-0009-0387/s465_test-img.jpg")
+    logger.expects(:info).with("PersonPresenter: Person of ID#1 has image with url '/uploads/carrierwave-tmp/1704204355-973510646575090-0009-0387/s465_test-img.jpg'").once
+
+    person.image = image
+    present(person).content
```
ryanb-gds, line 130: Whilst I applaud your dedication to test coverage, I'm not sure in this case that testing a logging call is sufficiently valuable to justify the performance and maintenance cost of the test (even though it's very minimal)

ryanb-gds (commented):
This isn't quite what I imagined, but happy to have a chat about the approach as I'm sure you have good reasons

ryanb-gds (approved):
Thanks for this :star:

PR #8775, opened 2024-01-29
Refactor asset manager storage and creation worker

ryanb-gds:
Laura and I spent a bit of time on Friday looking at the ongoing People image upload issue. We found it very difficult to reason about because the code that runs when an asset is uploaded is very convoluted.

This PR is intended to simplify some of that code, but there is quite a long way to go. I will update this description to provide a better overview of the changes once it is ready to fully review.


ryanb-gds:
There is not a clear path forward here

PR #8780, opened 2024-01-30
Remove the publishing step when trying to republish HTML attachments …

minhngocd:
…for unpublished or withdrawn documents

The publishing of the HTML attachment could leave the document in a bad state and especially dangerous around sensitive documents where they should be unpublished or withdrawn, but are then inadvertently published during the republishing process. We should be able to reapply the redirects for unpublishing without the need to publish the article first.

https://trello.com/c/e75Mp1Ap/2301-investigate-if-we-can-avoid-temporary-publishing-of-html-attachments-during-republishing-of-unpublished-edition


app/services/service_listeners/publishing_api_html_attachments.rb
```diff
@@ -18,8 +18,20 @@ def publish
     alias_method :unwithdraw, :publish
 
     def republish
-      update_publishing_api_content
-      unpublish_if_required
+      if Edition::PRE_PUBLICATION_STATES.include?(edition.state)
+        update_draft(update_type: "republish")
+      elsif edition.unpublishing && edition.withdrawn?
```
ryanb-gds, line 23: Would just `edition.withdrawn?` be sufficient? I don't _think_ you can have a withdrawn edition without an unpublishing... 🤔

app/services/service_listeners/publishing_api_html_attachments.rb
```diff
@@ -18,8 +18,20 @@ def publish
...
+        do_publish("republish")
+        discard_drafts(deleted_html_attachments)
+        withdraw
+      elsif edition.unpublishing
```
ryanb-gds, line 27: Similarly we could use `edition.unpublished?` here, I think

app/services/service_listeners/publishing_api_html_attachments.rb
```diff
@@ -18,8 +18,20 @@ def publish
...
+        update_draft(update_type: "republish")
```
ryanb-gds, line 28: What would happen if we didn't update the draft here? Strictly speaking it doesn't feel to me that it should be necessary, but possibly the unpublish request might fail because `allow_draft` is set to `true`?

test/unit/app/services/service_listeners/publishing_api_html_attachments_test.rb
```diff
@@ -486,10 +486,16 @@ class Republish < PublishingApiHtmlAttachmentsTest
...
+          true,
         )
         call(publication)
       end
+
+      test "for a publication that has been unpublished does not publish the attachment in order to unpublish again" do
+        publication = create(:unpublished_publication_in_error_no_redirect)
+        PublishingApiWorker.any_instance.expects(:perform).never
+        call(publication)
+      end
```
ryanb-gds, line 498: I'm personally not a huge fan of tests that test something doesn't happen, or at least not beyond an initial TDD cycle. Theoretically the set of things we could test don't happen is infinite so I always feel a bit dubious about it.

ryanb-gds (commented):
A few comments for your consideration - sorry we didn't get a chance to pair on this as there is definitely enough complexity in this small PR to warrant it. Maybe let's jump on a call to discuss. Pretty sure everything is correct anyway but just want to confirm my understanding.

test/unit/app/services/service_listeners/publishing_api_html_attachments_test.rb
minhngocd: In a documentation sense that we explicitly want to call out that publishing "should not" happen, then I think it should be tested for things that we definitely do not want to happen otherwise bad things happen. To prevent in the future if people do add that functionality back in, they can see the test and thing "hmm do i really want to do this"

app/services/service_listeners/publishing_api_html_attachments.rb
minhngocd: hmm i was thinking strictly from the point of view of keeping whitehall and publishing API up to date. additionally, if there were changes to the draft, surely we would want to see it reflected in draft origin?

ryanb-gds: Oh, yes, that makes sense. Thanks for explaining

app/services/service_listeners/publishing_api_html_attachments.rb
minhngocd: yup I can apply this

app/services/service_listeners/publishing_api_html_attachments.rb
minhngocd: yup I can apply this

PR #8781, opened 2024-01-31
Fix issue where `[nil]` is presented when no contacts

brucebolt:
We were including an array containing a single value of `nil`, if a worldwide organisation had no contacts.

This is not valid against the schema. Therefore updating to publish an empty array instead.

[Trello card](https://trello.com/c/yNJwHw4K)

### DIFF
--- a/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
@@ -90,9 +90,9 @@ module PublishingApi
     end
 
     def contacts
-      return [] unless item.main_office&.contact || item.home_page_offices&.map(&:contact)&.any?
+      return [] unless item.main_office_contact || item.home_page_office_contacts&.any?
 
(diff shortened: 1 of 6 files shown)


app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
```diff
@@ -90,6 +90,8 @@ def home_page_offices
     end
 
     def contacts
+      return [] unless item.main_office&.contact || item.home_page_offices&.map(&:contact)&.any?
```
JonathanHallam, line 93: would it be good to pull `item.main_office&.contact` and `item.home_page_offices&.map(&:contact)` into variables? As is they're used in a few places, not just in this file and am thinking it might be a bit easier to read

brucebolt: Done 👍

JonathanHallam (approved):
thanks!

PR #8788, opened 2024-02-01
Fix unpublishing of Worldwide Organisation Corporate Information Pages

brucebolt:
Unpublishing an editionable piece of content requires a `document_type` method in the Publishing API presenter.

Therefore adding the method and a test for the scenario.

Also unpublishes the affected Worldwide Organisation Corporate Information Pages and adds the missing database record for this, as they were not unpublished due to this error.

[Trello card](https://trello.com/c/xSMLrrf1)

### DIFF
--- a/app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
+++ b/app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
@@ -22,7 +22,7 @@ module PublishingApi
       content.merge!(PayloadBuilder::PolymorphicPath.for(item))
 
       content.merge!(
-        document_type: item.display_type_key,
+        document_type:,
         schema_name: "worldwide_corporate_information_page",
       )
     end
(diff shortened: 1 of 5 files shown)


app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
```diff
@@ -33,5 +33,9 @@ def links
         worldwide_organisation: [item.worldwide_organisation.content_id],
       }
     end
+
+    def document_type
```
JonathanHallam, line 37: More of a musing than a real comment and I don't think it'll work because of the way that the classes are arranged but, it'd be nice if this could go at a higher level because I'd assume that there are a few presenters that need to have this exact method implementation

JonathanHallam (approved):
Nice one, thanks :)

app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
brucebolt: Yeah, it could certainly be a default if we were inheriting from another class.

I'll leave as-is for now, but we may want to think about this in the future (e.g. as things like `BaseItemPresenter` are included in every presenter, so we may want that to be a default too.

PR #8795, opened 2024-02-01
Speed up `reporting:matching_docs` rake task

brucebolt:
This was previously retrieving all published editions and other documents (in batches of 1,000) into memory then looping through each to check the regex.

As of MySQL 8, there is support for `REGEXP` where queries. Therefore we can switch to use these, which speeds up the task from taking around one hour to 10 minutes.

This also fixes an issue where some runs of the rake task abort with an error as the pod runs out of memory.

[Trello card](https://trello.com/c/K3jNNRgX)


JonathanHallam (approved):
Really nice improvement, thank you :)

PR #8797, opened 2024-02-02
Separate asset storage classes

ryanb-gds:
This is part one of a refactor I'm working on to try to make the attachment and asset code in Whitehall more maintainable. I'd suggest reviewing it commit by commit.

This PR breaks the Whitehall Asset storage class down into several separate classes. There are three main categories of assets:

- Attachments, which require the "attachable" to be accessed in the create asset worker
- Previewables (e.g.) images, where we want to pass auth bypass IDs so that they can be loaded in shared previews
- Default asset types, which are not attachable or associated with things that can be previewed (e.g. people)

This allows to make some assumptions about what data is available from the uploder model (e.g. we must have an attachment because the attachment storage is only configured for the attachment uploader), which means we can simplify the code quite substantially.

One thing that is annoying is that I had to add the "attachable" everywhere when creating test `AttachmentData` objects. In a way it's annoying, but actually it's really exposing the fact that a large number of our tests were potentially incorrect because they were not invoking the full behaviour of the asset storage class. Attachment assets were being stored without the attachable being passed to the asset creation worker, which is not valid. In a future PR I intend to make this unnecessary.

In a future PR I will apply the same approach to the asset creation worker, which should be easier now that we can call a different worker depending on the type of asset we are storing.

Trello: https://trello.com/c/N1My8EtT


lib/whitehall/asset_manager_storage.rb
```diff
@@ -1,38 +1,16 @@
...
+    def self.store!(file)
+      temporary_location = ::File.join(
+        Whitehall.asset_manager_tmp_dir,
+        SecureRandom.uuid,
+        ::File.basename(file),
+      )
+
+      FileUtils.mkdir_p(::File.dirname(temporary_location))
+      FileUtils.cp(file, temporary_location)
+      temporary_location
```
ryanb-gds, line 12: I considered adding a unit test for this, but given it interacts so extensively with the file system and it's covered by integration and end to end tests I decided not to

app/uploaders/whitehall_uploader.rb
```diff
@@ -11,4 +11,12 @@ def relative_path
   def assets_protected?
     false
   end
+
+  def asset_params
+    {
+      assetable_id: model.id,
+      asset_variant: version_name ? Asset.variants[version_name] : Asset.variants[:original],
+      assetable_type: model.class.to_s,
+    }.deep_stringify_keys
```
ryanb-gds, line 20: I toyed with the idea of creating a model for this, e.g. `AssetReference`, so that I can store the reference in the database. That would enable me to just pass the reference ID to the worker, instead of all the assetable reference data.

test/unit/app/models/asset_test.rb
```diff
@@ -2,12 +2,12 @@
 
 class AssetTest < ActiveSupport::TestCase
   setup do
-    @attachment_data = create(:attachment_data_with_no_assets)
+    @image_data = create(:image_data_with_no_assets)
```
ryanb-gds, line 5: It was easier to use image data to test the Asset class, as the Image storage does not require an attachable

ryanb-gds (commented):
Sorry this is such a big PR, if you take the attachable test fixture wrangling out it wouldn't have been too bad. Reviewing commit by commit might be slightly easier 🙏

ryanb-gds:
> I think this all makes sense and it's a sensible break down of the assets.
> 
> Let's make sure to test on integration before deploying just to be sure :)

This was my test plan on integration. Any cases that I might have missed? https://docs.google.com/spreadsheets/d/1yYTOF7zZNA--KL05agGZWOKBCAi3PaULMXiqDiH0wn8/edit?usp=sharing

beccapearce:
I can't think of any others off the top of my head. Happy to do some clicking on integration when it's up. We could even do a mob try to break it if we think we've missed some cases

PR #8812, opened 2024-02-06
Enable automatic recompiling of stylesheets

ryanb-gds:
Follows guidance from https://docs.publishing.service.gov.uk/manual/migrate-to-dart-sass-from-libsass.html

One thing to note is that the basic watch command doesn't seem to work inside govuk docker. I had to add the option to poll for changes to the development configuration.

Trello: https://trello.com/c/n9bObkAU


dnkrj (approved):
Don't really have any expertise on this, but looks good!

PR #8821, opened 2024-02-08
Republish ministerial index pages when ministerial people or roles are updated

brucebolt:
We are missing callbacks on the `Person` and `MinisterialRole` models which means the ministers index page isn't updated when either of these are changed.

This adds that callback, so the page will be updated if something on the model (e.g. a person's role or name changes).

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -161,5 +161,10 @@ private
       PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksPresenter")
       PresentPageToPublishingApiWorker.perform_async("PublishingApi::MinistersIndexPresenter")
     end
+
+    if current_or_previous_prime_minister?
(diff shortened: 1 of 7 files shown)


app/models/person.rb
```diff
@@ -155,4 +155,11 @@ def touch_role_appointments
   def truncated_biography
     biography&.split(/\n/)&.first
   end
+
+  def republish_ministerial_pages_to_publishing_api
+    if role_appointments.map(&:ministerial?).any?
```
richardTowers, line 160: You can golf this a little

```suggestion
    if role_appointments.any?(&:ministerial?)
```

richardTowers (approved):

app/models/person.rb
brucebolt: Ooh, nice!  Didn't realise this was possible.

PR #8823, opened 2024-02-08
Expose default news image on WorldwideOrganisation

mec:
The concept of a default news image is used across different content
types, however the impementation vairies.

We want to make it consistent, here we expose the default news image in
the WorldwideOrganisationPresenter which in turn will allow the
PublishingAPI to treat it as a link, this is how the same attribute is
handled for Organisation.

We first exptract the helper methods to a module and then make the
necessary changes the schema has already been updated to reflect this
change.

Part of: https://trello.com/c/k4iXUPJU


jkempster34 (approved):
✅ 

Do we need this test as well? https://github.com/alphagov/whitehall/blob/7adcb23d788f4422c43a2f882756fac601f2e471/test/unit/app/presenters/publishing_api/organisation_presenter_test.rb#L391

PR #8826, opened 2024-02-09
Separate attachment asset creation worker

ryanb-gds:
## What

Create a separate asset creation worker for creating attachment assets

## Why

Having a single asset creation worker which tried to handle all possible asset creation requirements, regardless of whether the asset belonged to an attachment, an image, or another kind of model, led to the worker being very difficult to understand.

Attachments in particular needed a lot of special case handling, so I have created a separate asset creation worker specifically for attachment assets.

I have not yet DRYed the two workers up even though they duplicate a lot of code, because I am not yet completely confident that I wouldn't need to pull some of that code back into the concrete classes. I think we may want to change the behaviour of the attachment worker further (see TODO comment about Policy Groups). Once I am confident that I fully understand what the correct behaviour of the attachment asset creation worker should be I will remove the duplication.

Trello: https://trello.com/c/N1My8EtT

### DIFF
--- a/app/uploaders/storage/attachment_storage.rb
+++ b/app/uploaders/storage/attachment_storage.rb
@@ -5,7 +5,7 @@ class Storage::AttachmentStorage < CarrierWave::Storage::Abstract
 
     logger.info("Saving to Asset Manager for model AttachmentData with ID #{uploader.model&.id || 'nil'}")
 
-    AssetManagerCreateAssetWorker.perform_async(temporary_location, uploader.asset_params, true, uploader.model.attachable.class.to_s, uploader.model.attachable.id, uploader.model.auth_bypass_ids || [])
+    AssetManagerCreateAttachmentAssetWorker.perform_async(temporary_location, uploader.asset_params, uploader.model.attachable.class.to_s, uploader.model.attachable.id, true, uploader.model.auth_bypass_ids || [])
 
     Whitehall::AssetManagerStorage::File.new(uploader.store_path(::File.basename(original_file)), uploader.model, uploader.version_name)
   end
--- a/app/workers/asset_manager_create_asset_worker.rb
+++ b/app/workers/asset_manager_create_asset_worker.rb
@@ -23,7 +23,8 @@ class AssetManagerCreateAssetWorker < WorkerBase
     authorised_organisation_uids = get_authorised_organisation_ids(attachable_model_class, attachable_model_id)
     asset_options[:access_limited_organisation_ids] = authorised_organisation_uids if authorised_organisation_uids
 
-    create_asset(asset_options, asset_variant, assetable_id, assetable_type)
+    response = create_asset(asset_options)
+    save_asset(assetable_id, assetable_type, asset_variant, response.asset_manager_id, response.filename)
 
     enqueue_downstream_service_updates(assetable_id, assetable_type, attachable_model_class, attachable_model_id)
 
@@ -33,13 +34,6 @@ class AssetManagerCreateAssetWorker < WorkerBase
   end
 
 private
-
-  def create_asset(asset_options, asset_variant, assetable_id, assetable_type)
-    response = asset_manager.create_asset(asset_options)
-    asset_manager_id = get_asset_id(response)
-    filename = get_filename(response)
-    save_asset(assetable_id, assetable_type, asset_variant, asset_manager_id, filename)
-  end
 
   def enqueue_downstream_service_updates(assetable_id, assetable_type, attachable_model_class, attachable_model_id)
     assetable = assetable_type.constantize.find(assetable_id)
@@ -68,16 +62,5 @@ private
     asset.asset_manager_id = asset_manager_id
     asset.filename = filename
     asset.save!
-  end
-
-  def get_asset_id(response)
-    attributes = response.to_hash
-    url = attributes["id"]
-    url[/\/assets\/(.*)/, 1]
-  end
-
-  def get_filename(response)
-    attributes = response.to_hash
-    attributes["name"]
   end
 end
--- a/app/workers/asset_manager_create_asset_worker.rb
+++ b/app/workers/asset_manager_create_asset_worker.rb
@@ -9,44 +9,22 @@ class AssetManagerCreateAssetWorker < WorkerBase
   def perform(temporary_location, asset_params, auth_bypass_ids = [])
     return unless File.exist?(temporary_location)
 
-    file = File.open(temporary_location)
-
     assetable_id, assetable_type, asset_variant = asset_params.values_at("assetable_id", "assetable_type", "asset_variant")
+    assetable = assetable_type.constantize.where(id: assetable_id).first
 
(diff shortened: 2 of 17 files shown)


app/workers/asset_manager_create_asset_worker.rb
```diff
@@ -6,78 +6,25 @@ class AssetManagerCreateAssetWorker < WorkerBase
   # Use TransactionAwareClient for this worker to ensure that the commit is finished before the worker is executed.
   sidekiq_options queue: "asset_manager", client_class: Sidekiq::TransactionAwareClient
 
-  def perform(temporary_location, asset_params, draft = false, attachable_model_class = nil, attachable_model_id = nil, auth_bypass_ids = [])
+  def perform(temporary_location, asset_params, auth_bypass_ids = [])
```
lauraghiorghisor-tw, line 9: I think we're going to have to make the changes to the signature in a different PR, or rather, across several PRs, otherwise there are going to be `ArgumentError`s. There will be jobs of the old signature queued up when we release the new one.  

If I remember correctly, the way we typically address this is by doing this painful staggered PRs dance, where we create the new worker, release, allow for time to empty the queues of the old worker, then delete the old worker. In case we want to still use the old worker name, then even more dancing is required. 
[This](https://keygen.sh/blog/how-to-change-the-signature-of-a-sidekiq-job/) sums it up nicely.

app/uploaders/storage/attachment_storage.rb
```diff
@@ -5,7 +5,7 @@ def store!(carrierwave_file)
 
     logger.info("Saving to Asset Manager for model AttachmentData with ID #{uploader.model&.id || 'nil'}")
 
-    AssetManagerCreateAssetWorker.perform_async(temporary_location, uploader.asset_params, true, uploader.model.attachable.class.to_s, uploader.model.attachable.id, uploader.model.auth_bypass_ids || [])
+    AssetManagerCreateAttachmentAssetWorker.perform_async(temporary_location, uploader.asset_params, uploader.model.attachable.class.to_s, uploader.model.attachable.id, true, uploader.model.auth_bypass_ids || [])
```
lauraghiorghisor-tw, line 8: 👍🏻

app/workers/asset_manager_create_attachment_asset_worker.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    file = File.open(temporary_location)
+
+    asset_options = { file:, auth_bypass_ids:, draft: }
+    if attachable.respond_to?(:access_limited?) && attachable.access_limited?
+      asset_options[:access_limited_organisation_ids] = AssetManagerAccessLimitation.for(attachable)
+    end
+
+    response = create_asset(asset_options)
```
lauraghiorghisor-tw, line 28: I have the old version too deeply ingrained in my brain to be able to assess if inlining this makes it more readable. I do not dislike it though 👍🏻

app/workers/asset_manager_create_attachment_asset_worker.rb
```diff
@@ -0,0 +1,51 @@
+class AssetManagerCreateAttachmentAssetWorker < WorkerBase
+  include AssetManager::ServiceHelper
+
+  # Carrierwave runs on an after_save hook and the transaction that inserts Assetable into the database
+  # might not be committed yet. This can cause a race condition where the worker runs before the assetable is readable.
+  # Use TransactionAwareClient for this worker to ensure that the commit is finished before the worker is executed.
+  sidekiq_options queue: "asset_manager", client_class: Sidekiq::TransactionAwareClient
+
+  def perform(temporary_location, asset_params, attachable_model_class, attachable_model_id, draft = false, auth_bypass_ids = [])
```
lauraghiorghisor-tw, line 9: I'm thinking the correct default for draft should be true for `AttachmentWorker`. Though hard to tell since we're catering for both the `PolicyGroup` and `Edition` cases. I believe you have plans to break this further.

app/workers/asset_manager_create_attachment_asset_worker.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    asset = attachment_data.assets.where(variant: asset_params["asset_variant"]).first_or_initialize
+    asset.asset_manager_id = response.asset_manager_id
+    asset.filename = response.filename
+    asset.save!
+
+    if attachable.is_a?(Edition)
+      PublishingApiDraftUpdateWorker.perform_async(attachable.class.to_s, attachable.id)
+    else
+      # At present we can only reach this branch when the attachable is a PolicyGroup. No other non-editionable models have attachments.
```
lauraghiorghisor-tw, line 42: I still believe we should nonetheless update the `publishing-api` here too. That'd be the most consistent thing to do, and addresses the case below: 

The user may save the document with attachments still pending, meaning they would be filtered out at the presenter level (`WorkingGroupPresenter` -> uses default filtering in `govspeak_helper.rb`) so the document would be sent without any attachments. In typical editions we have added validation so that the user would be informed they cannot publish as attachments are still processing. `PolicyGroup` is the outlier since it uses the `PublishesToPublishingApi` concern flow. 

I'd prefer it if we followed up with a publishing-api call to republish the group, making sure the correct urls are sent here if they haven't already.

app/workers/asset_manager_create_asset_worker.rb
ryanb-gds: Oh I love migrations

app/workers/asset_manager_create_attachment_asset_worker.rb
ryanb-gds: By "attachments still pending" do you mean "not yet uploaded to Asset Manager"?

ryanb-gds: Also I don't think that would work because we'd have to push a draft and call publish in Publishing API to handle that, and generally we shouldn't be calling publish here

ryanb-gds: Trying to  have attachments with drafts and policy groups without is a flawed model, though

lauraghiorghisor-tw: With the "PROCESSING" tag, yes.

lauraghiorghisor-tw: > By "attachments still pending" do you mean "not yet uploaded to Asset Manager"?

With the "PROCESSING" tag, yes.

lauraghiorghisor-tw: > Also I don't think that would work because we'd have to push a draft and call publish in Publishing API to handle that, and generally we shouldn't be calling publish here

I was thinking implementing the `republish_on_assets_ready` equivalent for the `PolicyGroup` too.

ryanb-gds:
I'm going to close this, because although I think it probably has some value it isn't adding as much as I would like. I think the improvements we can make in this direction are pretty limited. I think a more significant improvement would require a change to the way Asset Manager works to allow Publishing apps to create the ID, so that we don't need to wait for the response from Asset Manager to update the Publishing API. There are some tradeoffs there though, so it needs some careful thought.

PR #8830, opened 2024-02-09
Remove need for editionable worldwide orgs to be associated with a taxon

callumknights:
Editionable worldwide orgs are not associated with taxons, so requiring a user to do this is unnecessary.

Trello: https://trello.com/c/knyAKE82/1023-dont-require-editionable-worldwide-organisations-to-be-associated-with-a-taxon

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -73,7 +73,11 @@ class Admin::EditionsController < Admin::BaseController
   def show
     fetch_version_and_remark_trails
 
-    @edition_taxons = EditionTaxonsFetcher.new(@edition.content_id).fetch
+    @edition_taxons = if @edition.requires_taxon?
+                        EditionTaxonsFetcher.new(@edition.content_id).fetch
+                      else
+                        []
+                      end
 
     if @edition.can_be_tagged_to_worldwide_taxonomy?
       @edition_world_taxons = EditionTaxonsFetcher.new(@edition.content_id).fetch_world_taxons
@@ -316,7 +320,7 @@ private
   end
 
   def saved_confirmation_notice
-    if params[:save].present? || @edition.has_been_tagged?
+    if params[:save].present? || @edition.has_been_tagged? || !@edition.requires_taxon?
       notice = "Your document has been saved"
       html_safe = false
     else
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -47,7 +47,7 @@ class Edition < ApplicationRecord
 
   validates_with SafeHtmlValidator
   validates_with NoFootnotesInGovspeakValidator, attribute: :body
-  validates_with TaxonValidator, on: :publish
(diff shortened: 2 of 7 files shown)


app/validators/taxon_validator.rb
```diff
@@ -1,6 +1,6 @@
 class TaxonValidator < ActiveModel::Validator
   def validate(edition)
-    if missing_taxons?(edition)
+    if missing_taxons?(edition) && edition.requires_taxon?
```
brucebolt, line 3: Could you add a test for this, e.g. in `TaxonValidatorTest`.

brucebolt: Alternatively, you could add the condition into the Edition model, e.g. `validates_with TaxonValidator, on: :publish, if: :requires_taxon?` then test it in `EditionTest`.

brucebolt (commented):
It's probably also worth making [this line](https://github.com/alphagov/whitehall/blob/fcb18dfbbf49271c9ab56c022d0c29b093715ffe/app/controllers/admin/editions_controller.rb#L76) conditional on `requires_taxon?` being true for the edition, so we don't make an unnecessary call to the Publishing API if we're not going to show those tags to users.

app/validators/taxon_validator.rb
jkempster34: If you go the `on: :publish` route, you have to call `invalid?(:publish)` in your test. Took me a while to figure that out 😓 

https://github.com/alphagov/whitehall/blob/9e7d05ec09af294ec3a07c642e3c976ce58afaf8/test/unit/app/models/corporate_information_page_test.rb#L617

test/unit/app/models/edition_test.rb
```diff
@@ -382,6 +382,11 @@ class EditionTest < ActiveSupport::TestCase
     assert_not build(:edition, summary: nil).valid?
   end
 
+  test "should be be valid without taxons when editionable_worldwide_organisation" do
+    edition = build(:draft_editionable_worldwide_organisation)
+    assert edition.valid?
+  end
```
brucebolt, line 388: This probably ought to be in the `EditionableWorldwideOrganisationTest`, since it applies only to that model.

callumknights: Moved 👍

app/validators/taxon_validator.rb
callumknights: Done 👍

test/unit/app/models/edition_test.rb
brucebolt: I think you can remove the test here too, as it's been copied into the other place.

brucebolt (commented):
All looks good, just one really minor thing where the test you copied over can be removed from the `EditionTest`.

callumknights:
My bad, removed it but didn't `cmd + s`. Sorted

brucebolt (approved):

PR #8834, opened 2024-02-12
Add ww orgs to filter

jkempster34:
https://trello.com/c/am4j97F6

### Add missing factories for all editionable worldwide organisation states
All of the factories for editionable models have these factories which are used
by various helper tests.

### Include worldwide organisations in the filter list
When the editionable worldwide organisation feature flag is enabled, we want to
display the option to filter documents by the worldwide organisation type.

Because the model name is EditionableWorldwideOrganisation, this uses the
`format_name` method as in #8706.

|Before|After|
|-|-|
|imageimage|imageimage|

### DIFF
--- a/lib/whitehall.rb
+++ b/lib/whitehall.rb
@@ -166,11 +166,12 @@ module Whitehall
       Publication,
       Speech,
       StatisticalDataSet,
+      *(EditionableWorldwideOrganisation if Flipflop.editionable_worldwide_organisations?),
     ]
   end
 
   def self.edition_route_path_segments
-    %w[news speeches policies publications consultations priority detailed-guides case-studies statistical-data-sets fatalities collections supporting-pages calls-for-evidence]
+    %w[news speeches policies publications consultations priority detailed-guides case-studies statistical-data-sets fatalities collections supporting-pages calls-for-evidence editionable-worldwide-organisations]
   end
 
   def self.analytics_format(format)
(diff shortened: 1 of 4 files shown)


lib/whitehall.rb
```diff
@@ -166,11 +166,12 @@ def self.edition_classes
       Publication,
       Speech,
       StatisticalDataSet,
+      *(EditionableWorldwideOrganisation if Flipflop.editionable_worldwide_organisations?),
     ]
   end
 
   def self.edition_route_path_segments
-    %w[news speeches policies publications consultations priority detailed-guides case-studies statistical-data-sets fatalities collections supporting-pages calls-for-evidence]
+    %w[news speeches policies publications consultations priority detailed-guides case-studies statistical-data-sets fatalities collections supporting-pages calls-for-evidence editionable-worldwide-organisations]
```
brucebolt, line 174: Does this also need to be conditional on the feature flag?

brucebolt (approved):

lib/whitehall.rb
jkempster34: Um, this is used to check whether edition links are valid (part of the `ADMIN_EDITION_PATH` regex), so there is no user facing impact.

Do you think it should be behind the flag? I'm not sure if anything could go wrong tbh 🤔

brucebolt: Ah, yes I see where this is used now.  Doesn't look like anything could go wrong by having it always included here.

PR #8835, opened 2024-02-12
Add missing callbacks for Editionable Worldwide Organisations

brucebolt:
This adds missing callbacks for republishing associated editions (at the moment only Editionable Worldwide Organisations) when roles and role appointments are changed.

This has all been made generic, so any editionable content types added in the future will automatically be republished when their associated roles or people are updated.

We don't need a callback on `Person`, as these will be updated through link expansion because it is not possible to associate a person with a worldwide organisation, only indirectly through the role appointment.

Full explanation in the commit messages.

[Trello card](https://trello.com/c/cHHrOnAj)

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -54,7 +54,7 @@ class Role < ApplicationRecord
 
   before_destroy :prevent_destruction_unless_destroyable
   after_update :touch_role_appointments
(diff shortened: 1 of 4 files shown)


jkempster34 (approved):
Makes sense!

PR #8843, opened 2024-02-14
Add home page offices to worldwide organisation presenter tests

brucebolt:
We were not including home page offices in the worldwide organisation presenter tests, therefore meaning the code for these was untested.

This adds a factory that allows us to have home page offices on worldwide organisations and tests they are currently included in the presented content.


jkempster34 (approved):
Good catch 🚢

PR #8844, opened 2024-02-14
Implement Worldwide Organisation offices as multi-part content

brucebolt:
We are switching Worldwide Organisations to include the content of their offices in the main content item for the organisation.  This adds that into the presenters for these content types.

In a later PR, we will remove the links and update routes once the rendering code (https://github.com/alphagov/government-frontend/pull/3095) has been updated.

Depends on https://github.com/alphagov/publishing-api/pull/2638.

### DIFF
--- a/app/helpers/presenters/publishing_api/worldwide_office_helper.rb
+++ b/app/helpers/presenters/publishing_api/worldwide_office_helper.rb
@@ -1,0 +1,26 @@
+module Presenters::PublishingApi::WorldwideOfficeHelper
+  def worldwide_office_details(worldwide_office)
+    {
+      access_and_opening_times: access_and_opening_times(worldwide_office),
+      services: services(worldwide_office),
+      type: worldwide_office.worldwide_office_type.name,
+    }
+  end
+
+private
+
+  def access_and_opening_times(worldwide_office)
+    return if worldwide_office.access_and_opening_times.blank?
+
+    Whitehall::GovspeakRenderer.new.govspeak_to_html(worldwide_office.access_and_opening_times)
+  end
+
+  def services(worldwide_office)
+    worldwide_office.services.map do |service|
+      {
+        title: service.name,
+        type: service.service_type.name,
+      }
+    end
+  end
+end
--- a/app/helpers/presenters/publishing_api/worldwide_office_helper.rb
+++ b/app/helpers/presenters/publishing_api/worldwide_office_helper.rb
@@ -7,6 +7,16 @@ module Presenters::PublishingApi::WorldwideOfficeHelper
     }
   end
 
+  def worldwide_office_parts(worldwide_offices)
+    worldwide_offices.map do |worldwide_office|
+      worldwide_office_details(worldwide_office).merge(
+        contact_content_id: worldwide_office.contact.content_id,
+        slug: worldwide_office.base_path.gsub("#{worldwide_office.worldwide_organisation.base_path}/", ""),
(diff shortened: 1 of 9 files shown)


jkempster34 (approved):
👍

test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
```diff
@@ -39,10 +41,35 @@ def present(...)
       description: worldwide_org.summary,
       details: {
         body: "<div class=\"govspeak\"><p>Information about the organisation with <em>italics</em>.</p>\n</div>",
+        home_page_office_parts: [
+          {
+            access_and_opening_times: Whitehall::GovspeakRenderer.new.govspeak_to_html(worldwide_org.reload.home_page_offices.first.access_and_opening_times),
+            contact_content_id: worldwide_org.reload.home_page_offices.first.contact.content_id,
+            services: [],
+            slug: worldwide_org.reload.home_page_offices.first.slug,
```
jkempster34, line 49: Just checking is this the full path `offices/office-1` or just `office-1`?

jkempster34: I haven't looked at your Gov Frontend code yet, but in the spike I did the other week, this would have had to be `offices/office-1`

brucebolt: It's `offices/office-1`.  I wasn't really sure whether to add the `offices/` here or in the frontend app.

I went with the frontend app in the end, as it isn't really part of the `slug`.

I wonder whether full base path (e.g. `/world/organisation/foo/offices/bar`) might actually be better than slug?

jkempster34: I kind of see it relative to the content item it's part of... if that makes sense?

brucebolt: Yeah, that's true.  Let's go for `office/office-1` here, then I can removed `office/` from the Government Frontend code.

app/helpers/presenters/publishing_api/worldwide_office_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+      services: services(worldwide_office),
+      type: worldwide_office.worldwide_office_type.name,
+    }
+  end
+
+  def worldwide_office_parts(worldwide_offices)
+    worldwide_offices.map do |worldwide_office|
+      worldwide_office_details(worldwide_office).merge(
+        contact_content_id: worldwide_office.contact.content_id,
+        slug: "office/#{worldwide_office.slug}",
```
jkempster34, line 14: Is there any way we could DRY this up to avoid hardcoding it here. 

Maybe raising the path as a constant in `WorldwideOffice`, or something?

jkempster34: I don't know what the chance of someone changing the `office` path is, but if they did, then this would break

brucebolt: Good point.  I've used the `base_path` method of the `WorldwideOffice`, then gsubbed out the base path of the `WorldwideOrganisation` and a slash.

PR #8850, opened 2024-02-16
Fix Jasmine specs in govuk-docker on Apple silicon

richardTowers:
Replay of https://github.com/alphagov/signon/pull/2723. Note: I haven't tested this, because I haven't got an Apple Silicon device to use.

We've been seeing the following error running the Jasmine specs in a govuk-docker container running on Apple silicon:

    SessionNotCreatedError: session not created:
      Chrome failed to start: exited normally.

For a full diagnosis of the problem see this govuk-docker PR [1] which was a first attempt at fixing the problem.

Note that the problem was also triggering a warning like the following even on non-Apple silicon:

    The chromedriver version (120.0.6099.199) detected in PATH at
    /usr/bin/chromedriver might not be compatible with the detected chrome
    version (121.0.6167.85); currently, chromedriver 121.0.6167.85 is
    recommended for chrome 121.*, so it is advised to delete the driver in
    PATH and retry

This PR forces yarn to install v4.17.0 of the `selenium-webdriver` node package which includes a fix [2] to the `selenium-manager` CLI app included in the `selenium-webdriver` node package so that it now correctly finds the chromium binary. This fixes both the error on Apple silicon and the warning on all devices.

[1]: alphagov/govuk-docker#724
[2]: SeleniumHQ/selenium#12890

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "stylelint-config-gds": "^2.0.0"
   },
   "resolutions": {
+    "selenium-webdriver": "4.17.0",
     "string-width": "4.2.3"
   }
 }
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
Can confirm this fixed it for me. Thanks @richardTowers

PR #8853, opened 2024-02-19
Switch worldwide offices to be included in worldwide organisation content items

brucebolt:
This will allow the same content item to be used by the offices, in addition to the parent worldwide organisation.

Full details are in each commit message.

Depends on https://github.com/alphagov/publishing-api/pull/2643.


jkempster34 (approved):
LGTM. Only thing is that commit 1 says that you amend `additional_routes` to `suffixes` so that it could be used elsewhere, but then commit 2 adds `additonal_paths`

brucebolt:
@jkempster34 I've added two more commits to this PR since you reviewed (it made sense to deploy those changes at the same time).  Would you mind taking a look at them?

jkempster34 (approved):

PR #8857, opened 2024-02-21
Empty office links for worldwide organisations

brucebolt:
Offices are now being included in the main details of the content item, to allow us to use the same content item to render all pages.

We therefore don't need these pages to be included in links, so need to empty them.

In a later PR (once all worldwide organisations are republished), we will remove these links entirely.

Depends on: https://github.com/alphagov/whitehall/pull/8853, https://github.com/alphagov/content-store/pull/1238 and https://github.com/alphagov/publishing-api/pull/2644.


brucebolt:
Closing, as we are no longer proceeding with multi-part content for worldwide organisations.

PR #8858, opened 2024-02-21
Remove office links for worldwide organisations

brucebolt:
In https://github.com/alphagov/whitehall/pull/8857, we emptied the office links from Worldwide Organisations, as offices are now being included in the main details of the content item, to allow us to use the same content item to render all pages.

After these have been republished, we no longer need the office links to be present in the presenter, so can remove them.

Depends on: https://github.com/alphagov/whitehall/pull/8857


brucebolt:
Closing, as we are no longer proceeding with multi-part content for worldwide organisations.

PR #8859, opened 2024-02-21
Remove Worldwide Office presenter

brucebolt:
We no longer need to present Worldwide Offices to Publishing API, as the content of those pages is now included in the Worldwide Organisation content item.

Depends on: https://github.com/alphagov/whitehall/pull/8853


brucebolt:
The commits in here have been moved to https://github.com/alphagov/whitehall/pull/8853.

PR #8861, opened 2024-02-21
Revert "Switch worldwide offices to be included in worldwide organisation content items"

brucebolt:
Reverting this PR as there are a number of issues on translated worldwide organisations that need to be corrected.

Reverts alphagov/whitehall#8853


jkempster34 (approved):
👍

PR #8862, opened 2024-02-21
Remove specialist topics from edition summary page

beccapearce:
## What
Remove the option within the UI to add a specialist topic from the edition summary page.

## Why
Specialist topics have been retired on GOV.UK, by removing this code we prevent tech debt and having unused code to maintain.

Trello card: https://trello.com/c/9C6tLJBL


Tetrino (approved):
LGTM!

app/presenters/publishing_api/links_presenter.rb
```diff
@@ -70,13 +69,8 @@ def worldwide_organisation_ids
       (item.try(:worldwide_organisations) || []).map(&:content_id)
     end
 
-    def topic_content_ids
-      item.specialist_sectors.map(&:topic_content_id)
-    end
-
     def parent_content_ids
```
ryanb-gds, line 72: Can we lose this method? Or is parent part of the links schema?

ryanb-gds (approved):
Glad to see this go 🧹 one or two questions but pretty sure this has everything covered

app/presenters/publishing_api/links_presenter.rb
beccapearce: Since parent content_ids are used elsewhere e.g. in HTML attachments I thought it might be easier to keep the method in.

Happy to give it a go at changing it. It caused a few test failures if I remember rightly which is why I thought I'd leave the empty array in.

Also please let me know if there is a smarter way of doing this 🙏🏻

app/presenters/publishing_api/generic_edition_presenter.rb
```diff
@@ -29,7 +29,6 @@ def content
     def links
       LinksPresenter.new(item).extract([
         :organisations,
-        :topics,
```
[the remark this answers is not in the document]
beccapearce: I _believe_ that this is only specialist topics. Other topics include stats announcements topics which I thought were handled differently. Otherwise a topic could also be a taxon but I _think_ in the code we usually call a taxon a taxon and only call it a topic on the user facing bit.

Let me know if I've missed any though or if I've misunderstood

app/presenters/publishing_api/links_presenter.rb
ryanb-gds: Well, HTML attachments have [their own presenter](https://github.com/alphagov/whitehall/blob/main/app/presenters/publishing_api/html_attachment_presenter.rb) that defines its own "links" object, but maybe this presenter is also used with HTML attachments. I think you'd probably know better than I would tbh

app/presenters/publishing_api/generic_edition_presenter.rb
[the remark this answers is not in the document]
ryanb-gds: No that helps, thanks

app/presenters/publishing_api/links_presenter.rb
beccapearce: How could I have doubted you! I just checked it again and it was the document collections I thought did some magic but I think it can be removed. Will push up the commit and double check on integration

beccapearce:
Retested on integration and think it's good to go!

PR #8873, opened 2024-02-28
Only surface image when not using default news image

jkempster34:
https://trello.com/c/k4iXUPJU

Unless a user has added a specific news image for a news article, then a default image is retrieved from an associated organisation or worldwide organisation.

Currently, the default image is surfaced as an `image` in the details of a news story. However, this is problematic because the default image does not update through dependency resolution when the image is changed on the associated organisation or worldwide organisation. 

This sets the image as nil if the article has a default image. Government Frontend will render the default news image when the details image is absent (government-frontend#3096), allowing us to replicate the functionality we are removing here through link expansion instead.

---


brucebolt (approved):

jkempster34:
We considered including default news image as a link instead of relying on ActiveRecord callbacks. However:
> 1. That leaves us with inconsistency between the records in Whitehall, and the content item (unless we add additional fields to the content item)
> 2. There is already a [callback pattern](https://github.com/alphagov/whitehall/blob/cee99a9c6a7d59afc72cf6d262af5cd40c2006f2/app/models/featured_image_data.rb#L34) in place to deal with the delay in uploading assets. 

See https://github.com/alphagov/whitehall/pull/8877

PR #8877, opened 2024-02-29
Featured image for editionable worldwide orgs

jkempster34:
https://trello.com/c/k4iXUPJU

> [!NOTE]  
> We considered including default news image as a link instead of relying on ActiveRecord callbacks. However:
> 1. That leaves us with inconsistency between the records in Whitehall, and the content item (unless we add additional fields to the content item)
> 2. There is already a [callback pattern](https://github.com/alphagov/whitehall/blob/cee99a9c6a7d59afc72cf6d262af5cd40c2006f2/app/models/featured_image_data.rb#L34) in place to deal with the delay in uploading assets. 

See https://github.com/alphagov/whitehall/pull/8873 and https://github.com/alphagov/government-frontend/pull/3096

https://github.com/alphagov/whitehall/assets/47089130/5f666310-035b-422d-931c-a4ac1a66e3b5

### Add default news image to editionable worldwide organisation
Editionable worldwide organisation are to replace worldwide organisations. Add
default news image to achieve parity.

### Add default news image to editionable worldwide organisation forms
Allow users to upload a default news image when creating or updating an
editionable worldwide organisation.

### Add edition editionable worldwide organisations table
Add database table for the association between editionable worldwide
organisations and other models (such as news stories).

The naming is unfortunate here, but follows existing convention.

### Add editionable worldwide organisations module
Add module for the association between editionable worldwide organisations and
other models (such as news stories).

This follows the same pattern defined in the statistical data sets module.

### Add missing test for worldwide organisations
There isn't any coverage for the featured image of world news stories where
their worldwide organisation has a default news image.

### Surface editionable worldwide organisation lead image
This will surface the default news image of an editionable worldwide
organisation when a model (such as a world news story) does not have its own
image.

This is displayed in Whitehall and also presented to the Publishing API.

### Republish world news stories when default news image is changed
All world news stories that use a default news image must be republished when
the editionable worldwide organisation default news image is updated.

This follows the existing pattern for organisations and worldwide organisations.

### Include the feature flags helper in model tests
This is included in controller tests but not model tests.

### Associate editionable worldwide organisations with world news stories
Allows users to select an editionable worldwide organisation when creating or
updating a world news story.

As editionable worldwide organisations are set to replace worldwide
organisations, this is behind a feature flag ready for the eventual switch.

### Republish world news stories after commit
An editionable worldwide organisation may have a change of state such as being
withdrawn. When this happens, any world news stories that are using the default
news image of the organisation must be republished.

### Clone default news image when creating a new draft
The default news image must be cloned when creating a new draft of an
editionable worldwide organisaiton.

---

### DIFF
--- a/app/controllers/admin/editionable_worldwide_organisations_controller.rb
+++ b/app/controllers/admin/editionable_worldwide_organisations_controller.rb
@@ -4,4 +4,9 @@ private
   def edition_class
     EditionableWorldwideOrganisation
   end
+
+  def build_edition_dependencies
+    super
+    @edition.default_news_image || @edition.build_default_news_image
+  end
 end
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -286,6 +286,7 @@ private
             ],
           },
         ],
+        default_news_image_attributes: %i[file file_cache id],
         nation_inapplicabilities_attributes: %i[id nation_id alternative_url excluded],
         fatality_notice_casualties_attributes: %i[id personal_details _destroy],
         document_attributes: [
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -254,6 +254,7 @@ private
         related_detailed_guide_ids: [],
         role_appointment_ids: [],
         statistical_data_set_document_ids: [],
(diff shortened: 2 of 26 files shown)


brucebolt (approved):
Looks good to me 👍

PR #8883, opened 2024-03-07
Remove email override tagging UI

beccapearce:
## Context
In https://github.com/alphagov/whitehall/pull/7950 a UI was added to Whitehall, to allow a very limited number of users with special permissions to see an extra “Email notifications” tab on the Edit page for document collections. This was to aid in the removal of specialist topics since specialist topics redirected to document collections, the email signup button on those document collection pages should have the option to sign up to the topic rather than just the page (the page being the default behaviour)

## What
This PR is to remove the temporary UI which allowed users with permission to change the notification type on document collections. We however also what to be able to show an information page int he email notifications tab where the email signup button is different from the standard behaviour i.e. it has topic level signup rather than page level signups.

The desired behaviour is:
- When a user visits the edit view of a new draft that has never been published (`taxonomomy_topic_email_overide.nil?`) they do not see the Email Notifications tab
- When a user visits the edit view of a published collection that has `taxonomomy_topic_email_overide.present?` they see the Email Notifications tab 

## Why

We will be replacing this self service UI with a rake task and developer documentation, as this feature was always intended to be temporary, and was added just for the department to be able to retire their specialist topics.

Trello card: https://trello.com/c/xB8VQy4D


ryanb-gds (approved):
🧹 🧹 🧹

app/views/admin/document_collection_email_subscriptions/edit.html.erb
```diff
@@ -18,12 +18,6 @@
...
 
     <h2 class="govuk-heading-l">Email notifications</h2>
-
-    <% if taxonomy_topic_cannot_be_set?(@collection) %>
-      <%= render partial: "email_override_summary_page" %>
-    <% else %>
-      <p class="govuk-body">Choose the type of email updates users will get if they sign up for notifications.</p>
-      <%= render partial: "form" %>
-    <% end %>
+    <%= render partial: "email_override_summary_page" %>
```
hannako, line 21: Shall we simplify this template? 
- We could lose lines 4-11 as we won't ever have errors
- We could move the content from the email override summary page into this partial, given it's now so simple.

hannako (approved):
Looks great, thank you Becca! 🏅

PR #8885, opened 2024-03-07
Only include current roles for worldwide organisations

brucebolt:
We are currently including all roles that a person has held at a Worldwide Organisation, when we only need to include their current roles.

This has led to a situation where a person's former role incorrectly appears on the Worldwide Organisation page.

After deployment: all worldwide organisations will need republishing.

[Trello card](https://trello.com/c/V79CVWB3)


jkempster34 (approved):
LGTM.

PR #8908, opened 2024-03-19
Use edition links for Worldwide Offices

brucebolt:
We are working to make Worldwide Organisations and their offices editionable.

Therefore we need to be able to link the links to editions, rather than documents, to prevent draft updates becoming live before publication.

Depends on https://github.com/alphagov/publishing-api/pull/2686.  After this is merged, all Worldwide Offices will need republishing.  Then another PR can be raised to remove the non-edition links.

[Trello card](https://trello.com/c/5Nr1NieG)


JonathanHallam (approved):
👍

PR #8909, opened 2024-03-19
Enable adding mainstream guide sub-pages to document collections

ryanb-gds:
This fixes an error that users would see when trying to add a mainstream guide sub-page to a document collection. The GovukUrl class, unlike the GovukUrlValidator class, did not allow for sub-page paths when verifying that a content item existed at the path provided. We refactor the validator to check the record can access a content item rather than fetching the content item inside the validator so that we can avoid duplicating the behaviour.

Trello: https://trello.com/c/aP2ZQQ9b

### DIFF
--- a/test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
+++ b/test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
@@ -30,6 +30,23 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
     assert url.valid?
   end
 
+  test "should be valid when a mainstream guide sub-page url is used" do
+    content_id = SecureRandom.uuid
+    stub_publishing_api_has_lookups("/foo" => content_id)
+    stub_publishing_api_has_item(content_id:,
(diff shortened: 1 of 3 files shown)


test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -90,6 +107,23 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
     assert url.errors.full_messages.include?("Url must reference a GOV.UK page")
   end
 
+  test "should be valid when a non-mainstream guide sub-page url is used" do
```
minhngocd, line 110: should this be "should be invalid"?

minhngocd (approved):
Amazing work - thanks @ryanb-gds !! very confusing piece of code - It's much more DRY and simple logic now 👍

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
ryanb-gds: Whoops, yes, thank you

PR #8910, opened 2024-03-19
Do not present non-edition links for Worldwide Offices

brucebolt:
In https://github.com/alphagov/whitehall/pull/8908, we switched Worldwide Offices to use edition links, so the draft version of the document can have different links to the parent.

Once all Worldwide Offices have been republished with edition links, we can remove the code to publish the non-edition links.

Depends on: https://github.com/alphagov/whitehall/pull/8908

[Trello card](https://trello.com/c/5Nr1NieG)


JonathanHallam (approved):
👍

PR #8912, opened 2024-03-20
Add editionable worldwide organisation links to world news stories

jkempster34:
https://trello.com/c/5nBCpGZr

### Link to editionable worldwide organisations for news articles
We are currently migrating worldwide organisations to "editionable worldwide
organisations".

We must include the new editionable worldwide organisations where the old
worldwide organisations were, this inlcludes where other content items have
links to worldwide organisations.

### Fix worldwide organisations select element JavaScript
There is JavaScript on this page which shows/hides the worldwide organisation
select element depending on the news article type.

The JavaScript depends on the
`app-view-edit-edition__world-organisation-fields` class.

Wrap both editionable worldwide organisations and worldwide organisations in
this class so that the JavaScript works for either.

---

### DIFF
--- a/app/presenters/publishing_api/links_presenter.rb
+++ b/app/presenters/publishing_api/links_presenter.rb
@@ -65,6 +65,8 @@ module PublishingApi
     end
 
     def worldwide_organisation_ids
+      return (item.try(:editionable_worldwide_organisations) || []).map(&:content_id) if item.try(:editionable_worldwide_organisations)&.any?
+
       (item.try(:worldwide_organisations) || []).map(&:content_id)
     end
 
--- a/app/views/admin/news_articles/_form.html.erb
+++ b/app/views/admin/news_articles/_form.html.erb
@@ -18,15 +18,13 @@
 
       <%= render "topical_event_fields", form: form, edition: edition %>
 
-      <% if Flipflop.editionable_worldwide_organisations? %>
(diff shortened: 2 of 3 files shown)


app/presenters/publishing_api/links_presenter.rb
```diff
@@ -65,6 +65,8 @@ def world_location_ids
     end
 
     def worldwide_organisation_ids
+      return (item.try(:editionable_worldwide_organisations) || []).map(&:content_id) if item.try(:editionable_worldwide_organisations)&.any?
```
jkempster34, line 68: So, this is based on the assumption that models will only have editionable worldwide organisations after the migration.

We can alternatively use the feature flag here, but if we do so this won't be fully-functional when using the cookies strategy (maybe not a problem)

jkempster34: I'm now thinking this a is bad assumption and we should just use the feature flag

callumknights: I think this is a good assumption, assuming that migration works exactly as planned feels wobbly to me

callumknights (approved):
LGTM 👍

app/presenters/publishing_api/links_presenter.rb
jkempster34: We decided what I've done is fine here. I've highlighted it on the migration card so we can address it when we have more info.

PR #8913, opened 2024-03-20
Delete office translations when the org translation is deleted

JonathanHallam:
This makes sure that when the translation on an editionable worldwide org is deleted
that all of the translations of offices are deleted too

This stops users from being able to translate offices into languages that the original
org isn't translated into

Trello - https://trello.com/c/BzQdhTx6/1045-delete-associated-translated-offices-when-a-translation-is-deleted

### DIFF
--- a/app/controllers/concerns/translation_controller_concern.rb
+++ b/app/controllers/concerns/translation_controller_concern.rb
@@ -26,6 +26,7 @@ module TranslationControllerConcern
 
   def destroy
     translatable_item.remove_translations_for(translation_locale.code)
+    translatable_item.destroy_associated(translation_locale.code)
     redirect_to destroy_redirect_path, notice: notice_message("deleted")
   end
 
--- a/app/models/editionable_worldwide_organisation.rb
+++ b/app/models/editionable_worldwide_organisation.rb
@@ -57,6 +57,12 @@ class EditionableWorldwideOrganisation < Edition
     "worldwide organisation"
   end
 
+  def destroy_associated(locale)
+    offices.each do |office|
+      office.contact.remove_translations_for(locale)
+    end
+  end
+
   alias_method :original_main_office, :main_office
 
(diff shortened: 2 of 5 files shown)


app/models/application_record.rb
```diff
@@ -30,6 +30,8 @@ def append_url_options(path, options = {})
     path
   end
 
+  def destroy_associated(locale); end
```
jkempster34, line 33: Can we move this to `TranslatableModel`, I think it makes more sense there.

jkempster34 (commented):
LGTM apart from the small comment.

Should we add a card to warn users that they are removing translations for associations? Just a bit of text on the delete confirmation screen maybe

app/models/application_record.rb
JonathanHallam: yeah, i did originally, but it causes a tonne of errors because not everything that includes the translation concern uses that model. Have pushed up a quick wip commit but here's the run https://github.com/alphagov/whitehall/actions/runs/8392753788/job/22986130575

jkempster34: Try `lib/translatable_model.rb`, that's where `remove_translations_for` lives

app/controllers/concerns/translation_controller_concern.rb
```diff
@@ -27,6 +27,7 @@ def confirm_destroy; end
   def destroy
     translatable_item.remove_translations_for(translation_locale.code)
     redirect_to destroy_redirect_path, notice: notice_message("deleted")
+    translatable_item.destroy_associated(translation_locale.code)
```
jkempster34, line 30: Also, does this need moving above the `redirect_to`?

app/models/application_record.rb
JonathanHallam: you're a gentleman and a scholar! Done

jkempster34 (approved):
💯

PR #8923, opened 2024-03-25
Fix weak URL validation regex and make it const.

sengi:
- Match dots correctly when validating that a URL has a GOV.UK domain.
- Make the (private) validation regex a class constant instead of constructing it every time via an unnecessary private method.

Fixes https://github.com/alphagov/whitehall/security/code-scanning/11.


ryanb-gds (approved):
Thanks @sengi

PR #8940, opened 2024-04-05
Fix docs link with newline in the middle

richardTowers:
This is rendering as a link to:

https://www.gov.uk/guidance/how-to-publish-on-gov-uk/%0Aintroduction-and-access-to-whitehall-publisher#filter-documents

Note the %0A - hex code for a newline

Removing that fixes the link.


JonathanHallam (approved):
Nice spot

PR #8941, opened 2024-04-05
Load and configure Google Analytics 4

ryanb-gds:
We will run both GA4 and Universal Analytics in tandem while we migrate event tracking to GA4.

Note that we have to accept the default cookie settings on the user's behalf to enable GA4 tracking. This is considered reasonable given that all publishers are part of government.

I have removed the rendering of the Google Tag Manager script component as the GTM script is loaded when initialising GA4

Trello: https://trello.com/c/yauxjItY

### DIFF
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -1,17 +1,10 @@
 <% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
 
-<% if ENV["GOOGLE_TAG_MANAGER_ID"] %>
-  <% content_for :head do %>
-    <%= render "govuk_publishing_components/components/google_tag_manager_script", {
-      gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
-      gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
-      gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"],
-    } %>
-  <% end %>
-<% end %>
-
 <% if user_signed_in? %>
   <% content_for :head do %>
+    <meta name="govuk:components_gem_version" content="<%= GovukPublishingComponents::VERSION %>">
+    <%= javascript_include_tag "admin/domain-config" %>
+    <%= javascript_include_tag "govuk_publishing_components/load-analytics" %>
     <meta name="custom-dimension:8" content="<%= current_user.organisation_slug.presence || "(not set)" %>">
   <% end %>
 <% end %>
@@ -21,7 +14,7 @@
            environment: environment,
            browser_title: ("Error: " if yield(:error_summary).present?).to_s + sanitize((yield(:page_title).presence || yield(:title))) do %>
 
-  <!-- This element exists to initialise the JS module that configures custom Analytics behaviour -->
+  <!-- This element exists to initialise the JS module that configures custom Universal Analytics behaviour -->
   <div data-module="app-analytics"></div>
 
   <%= render "govuk_publishing_components/components/skip_link" %>
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,4 +1,5 @@
 //= link_tree ../images
 //= link_tree ../builds
 //= link application.js
+//= link admin/domain-config.js
 //= link components/visual-editor.js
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/admin/auto-accept-cookies.js
```diff
@@ -0,0 +1,5 @@
+window.GOVUK = window.GOVUK || {};
+(function (GOVUK) {
+  'use strict'
+  GOVUK.setDefaultConsentCookie()
+}(window.GOVUK))
```
andysellick, line 5: A simpler way of doing this (which I think should work, haven't tested but [seen implemented successfully elsewhere](https://github.com/alphagov/content-tagger/pull/1657/files)) would be to add the following at the end of `application.js`. The analytics code is initialised in a document.ready so it's fine to do it like this. Also I think you also need the second line.

```
window.GOVUK.approveAllCookieTypes()
window.GOVUK.cookie('cookies_preferences_set', 'true', { days: 365 })
```

app/views/layouts/design_system.html.erb
```diff
@@ -21,7 +11,25 @@
...
 
-  <!-- This element exists to initialise the JS module that configures custom Analytics behaviour -->
+  <%= javascript_tag nonce: true do -%>
+    ;(function (window) {
+      window.GOVUK = window.GOVUK || {}
+      window.GOVUK.vars = window.GOVUK.vars || {}
+      window.GOVUK.vars.extraDomains = [
+        {
+          name: 'whitehall-admin',
+          domains: [document.location.host],
```
andysellick, line 21: This is a nice shortcut of enabling the analytics code on any domain where whitehall runs but is this what you really want to do? Our default setup on GOV.UK uses different GA properties for integration, staging and production, so we can clearly separate out our analytics data.

app/views/layouts/design_system.html.erb
```diff
@@ -1,15 +1,5 @@
...
-      gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
-      gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
-      gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"],
-    } %>
-  <% end %>
-<% end %>
-
 <% if user_signed_in? %>
   <% content_for :head do %>
     <meta name="custom-dimension:8" content="<%= current_user.organisation_slug.presence || "(not set)" %>">
```
andysellick, line 5: I think you're also going to need the following here:

```
  <meta name="govuk:components_gem_version" content="<%= GovukPublishingComponents::VERSION %>" />
  <%= javascript_include_tag "govuk_publishing_components/load-analytics" %>
```

I also think the JavaScript you've added on line 14 will need to be included in the HEAD prior to this.

app/views/layouts/design_system.html.erb
ryanb-gds: I'd assumed that's what the `preview` variable does. In that case what does the preview variable do?

ryanb-gds: It seems that this is how we're currently set up looking at https://github.com/alphagov/whitehall/blob/main/app/assets/javascripts/admin/analytics.js but I will check with the performance analysis team what they would prefer

andysellick: I guess it depends on what your required GA4 setup is. Ours is that we have the three environments ([production](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/domain-config.js#L22), [staging](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/domain-config.js#L35), [integration](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/domain-config.js#L50)) and each one has a separate entry in the domain config (see links).

All three have the same value for `id`, staging and integration each have their own unique values for `auth` and `preview`, whereas production has only `id` (the other two are optional). If your setup is 'all the data from any environment goes to the same place' then that's fine, although that seems like it would be quite polluted (e.g. for development we have separate `auth` and `preview` values).

ryanb-gds: Ah, so are you saying that if I use the same `name` in all environments then the data will end up in a single GA4 property, even if the `auth` and `preview` values are different?

ryanb-gds: I've spoken to the Performance Analysts and in Publishing we have separate tag IDs for each environment - just waiting to be provided with those.

ryanb-gds: Correction, we have a single tag ID and a different preview and auth value for the integration environment. We do not have an analytics environment set up for staging. I have added the appropriate configuration

PR #8955, opened 2024-04-15
Rename `FeaturedDocumentsPresenter`

brucebolt:
The name implied this is a presenter, but it is in fact a collection of helper methods used in other presenters.

Therefore renaming to reflect this isn't a presenter.


jkempster34 (approved):
👍

PR #8956, opened 2024-04-15
Move `LinksPresenter`

brucebolt:
The name of this presenter implies it is a presenter, however it builds a hash that is included in other presenters.

Therefore moving to be present under `PayloadBuilder`, in the same way as all other similar presenters.


jkempster34 (approved):
Makes sense

PR #8957, opened 2024-04-15
Rename `WorkingGroupPresenter`

brucebolt:
Renaming the `WorkingGroupPresenter` to be more consistently named compared to the other Publishing API presenters.

### DIFF
--- a/app/presenters/publishing_api/policy_group_presenter.rb
+++ b/app/presenters/publishing_api/policy_group_presenter.rb
@@ -1,5 +1,5 @@
 module PublishingApi
-  class WorkingGroupPresenter
+  class PolicyGroupPresenter
     attr_accessor :item, :update_type
 
     def initialize(item, update_type: nil)
--- a/app/models/policy_group.rb
+++ b/app/models/policy_group.rb
@@ -81,6 +81,6 @@ class PolicyGroup < ApplicationRecord
              description: :summary
 
   def publishing_api_presenter
-    PublishingApi::WorkingGroupPresenter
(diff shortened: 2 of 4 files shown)


app/presenters/publishing_api/policy_group_presenter.rb
```diff
@@ -49,12 +49,7 @@ def details
     end
 
     def body
-      # It looks 'wrong' using the description as the body, but it isn't
-      if item.description.present?
-        Whitehall::GovspeakRenderer.new.govspeak_with_attachments_to_html(item.description, item.attachments, item.email)
-      else
-        ""
```
jkempster34, line 56: Is this empty string important in the frontend app? I.e., will changing this from "" to nil break something?

jkempster34 (approved):

app/presenters/publishing_api/policy_group_presenter.rb
jkempster34: I didn't read your commit comment 🤦. Ignore this

jkempster34 (changes requested):
Going back to my earlier comment, I think this will raise errors in Government Frontend.

See this module https://github.com/alphagov/government-frontend/blob/main/app/presenters/content_item/body.rb#L2

Empty strings are propogated through Publishing API, I've just done a test by editing one of the Policy Groups in Whitehall.

brucebolt:
Good spot!  I've dropped the second commit and just kept the first one which renames the presenter.

PR #8958, opened 2024-04-15
Add a user interface for republishing past prime ministers page

Gweaton:

This PR is the first step towards being able to run all our republishing jobs via a user interface, to allow GDS Admins to republish a document when necessary. In future, we'll be adding all other `republishing` and `bulk_republishing` rake tasks to this, to eliminate the need for developers to run the tasks manually.

**Still to do:**

Figure out a way to prevent double clicks of the "Confirm republishing" button.

### Republishing index page

image

### Confirmation page

image

### Index page with confirmation banner

image


jlewr:
I think the content here is probably fine, given this is aimed at GDS Admins. 

My only concern would be that this might be intimidating for non-developers who won't appreciate things like 'govspeak' and the 'API'.

lib/whitehall/authority/rules/document_rules.rb
```diff
@@ -8,7 +8,11 @@ def initialize(actor, subject)
     end
 
     def can?(action)
-      actor.gds_editor? || actor.departmental_editor? || action == :create
+      if action == :republish
```
jkempster34, line 11: I _think_ these rules map to model types. It might be more appropriate to place this in `MiscellaneousRules`

app/controllers/admin/republishing_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Admin::RepublishingController < Admin::BaseController
+  before_action :enforce_permissions!
+
+  def index; end
+  end
```
jkempster34, line 19: Extra `end` in this commit

app/views/admin/republishing/index.html.erb
```diff
@@ -31,13 +31,11 @@
         text: "Past prime ministers",
       },
       {
-        text: button_to("Republish",
-          admin_republishing_republish_past_prime_ministers_index_path,
-          method: :post,
+        text: link_to(sanitize("Republish #{tag.span('Past prime ministers', class: 'govuk-visually-hidden')}"),
```
jkempster34, line 34: Nice 🙌

jkempster34:
This all looks good to me!

I wonder if the commits here follow your own personal journey on this PR, i.e., adding a confirmation page when agreed yesterday. That might be unnecessary to know in future when looking through the commit history

app/controllers/admin/republishing_controller.rb
```diff
@@ -9,9 +9,26 @@ def republish_past_prime_ministers_index
...
 
   def enforce_permissions!
     enforce_permission!(:republish, Document)
   end
+
+  def republishable_documents
+    [{
+      title: "Past prime ministers",
+      slug: "past-prime-ministers",
+      path: "/government/admin/republishing/republish-past-prime-ministers-index",
```
yndajas, line 31: Do we need the path? If we added `-index` to this specific slug, could we reach the path programmatically from the slug (and ideally a path helper, possibly extracted into a helper/util method)? Would that work for other pages?

app/views/admin/republishing/confirm.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :page_title, "Are you sure you want to republish \"#{@document_title}\"?" %>
+<% content_for :title, "Are you sure you want to republish \"#{@document_title}\"?" %>
```
yndajas, line 2: For the one that's the document title not the heading, should we simplify to `"Republish \"#{@document_title}\""`? I can't remember if using different content for the document title and `h1` is bad practice

app/views/admin/republishing/confirm.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% content_for :title_margin_bottom, 6 %>
+<div class="govuk-grid-row">
+  <section class="govuk-grid-column-two-thirds">
+    <p class="govuk-body govuk-!-margin-bottom-7">
+      This will schedule the page to be republished.
+    </p>
+
+    <%= button_to("Confirm republishing",
```
yndajas, line 10: "Confirm republish request"?

app/views/admin/republishing/index.html.erb
yndajas: Might be good to add "page" at the end

yndajas: Maybe not worth it yet, but we could add a helper method for this like on RODA: https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/main/app/helpers/application_helper.rb#L16-L24 (the RODA method would need updating to remove an extra space if used in v5 of the design system)

config/routes.rb
```diff
@@ -23,6 +23,7 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
       resources :users, only: %i[index show edit update]
 
       get "republishing" => "republishing#index", as: :republishing_index
+      get "republishing/:document_slug/confirm" => "republishing#confirm", as: :confirm_republishing
```
yndajas, line 26: I'm curious if we need these `as`es here

test/functional/admin/republishing_controller_test.rb
```diff
@@ -19,6 +19,23 @@ class Admin::RepublishingControllerTest < ActionController::TestCase
     assert_response :forbidden
   end
 
+  test "GDS Admin users should be able to access GET :confirm with an existing document slug" do
```
yndajas, line 22: `an existing document slug` -> `a republishable document slug`?

test/functional/admin/republishing_controller_test.rb
```diff
@@ -19,6 +19,23 @@ class Admin::RepublishingControllerTest < ActionController::TestCase
...
+    get :confirm, params: { document_slug: "past-prime-ministers" }
+    assert_response :ok
+  end
+
+  test "GDS Admin users should see a 404 page when trying to confirm republishing a document that doesn't exist" do
+    get :confirm, params: { document_slug: "non-existent" }
+    assert_response :not_found
+  end
```
yndajas, line 30: And:
- `that doesn't exist` -> `that is not republishable via this interface`
- `document_slug: "not-republishable"`

lib/whitehall/authority/rules/document_rules.rb
yndajas: Does mapping this rule to the `Document` model feel off here?

jkempster34: I think so yeah if you look at `RULESMAP`. That to me means the actions performed on a Document.

Whereas, the `MiscellaneousRules` include other administrative task like the emergency banner, and sitewide settings

app/views/admin/republishing/index.html.erb
```diff
@@ -0,0 +1,38 @@
...
+  Sometimes it may be necessary to republish content to the Publishing API. This will refresh the content on the website.
+</p>
+
+<p class="govuk-body">
+  For example if we make an update to govspeak and a publishing application pre-renders that content prior to its submission to Publishing API, that would require us to re-render and save new HTML for content.
+</p>
+
+<p class="govuk-body">
+  The following actions will allow you to republish content that was originally published in this application.
+  Any linked editions will also be republished through dependency resolution.
```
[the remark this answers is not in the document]
jkempster34: I think editions is correct/fine https://docs.publishing.service.gov.uk/repos/publishing-api/dependency-resolution.html

app/controllers/admin/republishing_controller.rb
```diff
@@ -9,9 +9,26 @@ def republish_past_prime_ministers_index
     redirect_to(admin_republishing_index_path)
   end
 
+  def confirm
+    republishable_document = republishable_documents.find { |document| document[:slug] == params[:document_slug] }
+
+    return render "admin/errors/not_found", status: :not_found unless republishable_document
```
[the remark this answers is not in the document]
yndajas: I mistook this action for the one that actually schedules the republishing

features/step_definitions/republishing_published_documents_steps.rb
```diff
@@ -0,0 +1,20 @@
...
+  person_a = create_person("A Person")
+  @prime_minister_role = create(:ministerial_role, name: "Prime Minister", cabinet_member: true)
+  @person_a_prime_minister_appointment = create(:role_appointment, person: person_a, role: @prime_minister_role, started_at: Date.parse("2020-01-01"))
+end
+
+Given(/^the prime minister subsequently changed$/) do
+  @person_a_prime_minister_appointment.update!(ended_at: Date.parse("2024-01-01"))
+  person_b = create_person("B Person")
+  create(:role_appointment, person: person_b, role: @prime_minister_role, started_at: Date.parse("2024-01-01"))
+end
```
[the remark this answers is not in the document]
jkempster34: Agreed!

app/views/admin/republishing/confirm.html.erb
Gweaton: I think I prefer the original, unless you've got strong feelings about this one?

config/routes.rb
Gweaton: We've decided to keep these to make the routes a bit more readable.

app/views/admin/republishing/confirm.html.erb
Gweaton: We've changed the title here.

lib/whitehall/authority/rules/document_rules.rb
yndajas: We had a chat about this and felt that republishing a document is an action performed on a document (the updated at date will change, I think?). The miscellaneous rules felt a little more meta. Might be worth chatting about this one tomorrow in case you have more context on this though?

app/views/admin/republishing/index.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          text: "Document title",
+        },
+        {
+          text: "Action",
+        },
+      ],
+      rows: [
+        [
+          {
+            text: "Past prime ministers",
```
brucebolt, line 33: Maybe this could link to [the page](https://www.gov.uk/government/history/past-prime-ministers) that's actually being republished, so it's very clear to the user which page is being updated?

brucebolt (commented):
This looks great to me.  It's amazing to see this take shape after we've been speaking about how annoying these rake tasks have been for years!

I've added a small comment about adding a link to the table, but otherwise (pending a decision on @jkempster34's questions about permissions), all looks good to me.

A front end developer in the Publishing Experience team should be able to advise about the bottom margins.

lib/whitehall/authority/rules/document_rules.rb
jkempster34: Happy to be overruled on this.

For me, this would map directly to a DocumentsController. Ultimately, it doesn’t really matter though!

The updated_at won’t change with this operation on the Whitehall models, only the updated_at of the content item will change.

Gweaton: We've decided to go for your suggestion here as we're not updating the Whitehall model.

brucebolt (approved):

PR #8966, opened 2024-04-17
Build on republishing work

yndajas:
This builds on the republishing work in #8958, renaming and DRYing up various bits of functionality and content in order to prepare us to extend the republishing functionality to further documents, other content types, and bulk republishing

---

### DIFF
--- a/app/controllers/admin/republishing_controller.rb
+++ b/app/controllers/admin/republishing_controller.rb
@@ -2,7 +2,7 @@ class Admin::RepublishingController < Admin::BaseController
   before_action :enforce_permissions!
 
   def index
-    @republishable_documents = republishable_documents
+    @republishable_pages = republishable_pages
   end
 
   def republish_past_prime_ministers_index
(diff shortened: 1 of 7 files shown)


app/controllers/admin/republishing_controller.rb
```diff
@@ -11,13 +11,13 @@ def republish_past_prime_ministers_index
     redirect_to(admin_republishing_index_path)
   end
 
-  def confirm
-    republishable_document = republishable_documents.find { |document| document[:slug] == params[:document_slug] }
+  def confirm_commonly_republished_document
```
Gweaton, line 18: Hmm, is there a shorter name we can use here to distinguish between republishing a document with a slug and one without? I'm not sure if these are any more commonly-republished than the other type, and  I was under the impression neither of these were relatively commonly-republished! To help figure out the name for this type of document, what was the name you'd planned for the other type?

config/routes.rb
```diff
@@ -22,9 +22,13 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
 
       resources :users, only: %i[index show edit update]
 
-      get "republishing" => "republishing#index", as: :republishing_index
-      get "republishing/:document_slug/confirm" => "republishing#confirm_commonly_republished_document", as: :confirm_commonly_republished_document_republishing
-      post "republishing/:document_slug/republish" => "republishing#republish_commonly_republished_document", as: :republish_commonly_republished_document
+      scope :republishing do
+        root to: "republishing#index", as: :republishing_index, via: :get
+        scope :document do
```
Gweaton, line 27: This routing makes sense, but I wonder if it conflicts with our understanding of this entity as a "document" vs it being "content"? I think we need to decide what to call these different types of "thing" being published as it'll help with the other question I had re: what to call the things we're republishing.

Gweaton (commented):
Looks like some good tidyups. Let's figure out what to call the different publishable entities!

app/controllers/admin/republishing_controller.rb
yndajas: Probably just `confirm_document`! I wonder if it's worth asking the others why these ones have specific Rake tasks - that might reveal a name?

app/views/admin/republishing/index.html.erb
```diff
@@ -1,5 +1,5 @@
-<% content_for :page_title, "Republish documents" %>
```
yndajas, line 1: Update the permission here from `:republish_documents` to `:republish_content`

Gweaton (approved):
Looks good, thanks for this!

PR #8967, opened 2024-04-18
Do not require body for Editionable Worldwide Organisations

brucebolt:
Non-editionable worldwide organisations do not always have body content, so matching that behaviour for the editionable versions.


brucebolt:
Closing as this only affects one Worldwide Organisation.  We can skip the validation when doing the migration.

PR #8969, opened 2024-04-18
Remove references to mapped_specialist_topic_content_id

unoduetre:

# What

As part of work to retire specialist topics, we temporarily stored a mapping between the archived specialist topic and its document collection conversion. We should remove this once we’re done.

This PR is the first one and it prepares for the database column removal.

# Why

To keep the system clean and avoid unnecessary data.

[Trello card](https://trello.com/c/T6GgX4Vj/2503-remove-mappedspecialisttopiccontentid-from-document-collections-m-l)

### DIFF
--- a/app/models/document_collection.rb
+++ b/app/models/document_collection.rb
@@ -56,10 +56,6 @@ class DocumentCollection < Edition
     "/government/collections/#{slug}"
   end
 
-  def specialist_topic_conversion?
-    mapped_specialist_topic_content_id.present?
-  end
-
   def publishing_api_presenter
     PublishingApi::DocumentCollectionPresenter
   end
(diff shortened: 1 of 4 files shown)


beccapearce (commented):
This is looking good. Just have some naming suggestions if that's ok.
What do you think of changing the PR name to something like `Remove references to mapped_specialist_topic_content_id from document collections` or something like that. It would be great to avoid the part 1 and part 2 given they are so distinct. 

I was also wondering if the commit message could be changed a bit too, Maybe something like `Remove the mapped_specialist_topic_content_id from code`?

unoduetre:
> This is looking good. Just have some naming suggestions if that's ok. What do you think of changing the PR name to something like `Remove references to mapped_specialist_topic_content_id from document collections` or something like that. It would be great to avoid the part 1 and part 2 given they are so distinct.
> 
> I was also wondering if the commit message could be changed a bit too, Maybe something like `Remove the mapped_specialist_topic_content_id from code`?

I used the name of the ticket as the PR name, as I usually do. I wonder if it's better to use more descriptive names or to keep using ticket titles? What do you think?

With respect to the commit name, the one proposed is too long for the first line, so instead I could e.g. add it as the second line?

beccapearce:
Ah I see I think having it more descriptive than the ticket is useful since one ticket could have multiple PRs. 

And yes 2 lines would be great or maybe `Remove mapped specialist topics from code` could work?

unoduetre:
@beccapearce I've updated the PR names and commit messages.

PR #8971, opened 2024-04-18
Migrate Worldwide Organisations to Editionable Worldwide Organisations

brucebolt:
This migrates all `WorldwideOrganisation` records (and their associations) to `EditionableWorldwideOrganisation` records.

The migration is done inside a transaction, with a check of various things (e.g. that the presented content remains the same) so the entire migration can be aborted if any part fails.

I've gone against our convention of not adding code comments, as I feel they are useful here in explaining what is actually being migrated in each step.


callumknights (approved):
Think I'm as happy with this as I can reasonably be. Especially as this has run in integration with no issues 👍

Gweaton (approved):
Same as Callum - I've combed through as best I can and all seems to make sense, particularly after going through it together yesterday.

PR #8975, opened 2024-04-22
Add visual editor private beta permission

lauraghiorghisor-tw:
[Trello card](https://trello.com/c/eHnQgH5D/2480-show-visual-editor-based-on-private-beta-user-permission-and-feature-toggle)


features/step_definitions/visual_editor_steps.rb
```diff
@@ -0,0 +1,31 @@
+Given(/^The visual editor private beta feature flag is (enabled|disabled)$/) do |enabled|
+  @test_strategy ||= Flipflop::FeatureSet.current.test!
+  @test_strategy.switch!(:govspeak_visual_editor, enabled == "enabled")
+end
+
+When(/^I start creating a new publication$/) do
+  visit new_admin_publication_path
+end
+
+Then(/^I should see the visual editor instead of the govspeak editor$/) do
```
ryanb-gds, line 10: The name of this "then" clause would imply to me that we are checking that the govspeak editor is not visible as well. How easy that is in reality I don't know, checking things are not visible is very tricky sometimes

features/step_definitions/visual_editor_steps.rb
```diff
@@ -0,0 +1,31 @@
...
+  visit new_admin_edition_attachment_path(publication, type: "html")
+end
+
+Given(/^a draft publication with an HTML attachment "(.*?)" exists$/) do |title|
+  html_attachment = create(:html_attachment, title:)
+  @publication = create(:publication, attachments: [html_attachment])
+end
+
+When(/^I edit the HTML attachment "(.*?)"$/) do |title|
+  html_attachment = Attachment.find_by(title:)
```
ryanb-gds, line 29: I know we can't set the ID ourselves, but perhaps we could use the base path or something that we can be confident in the format of? Matching on strings with no specified format is just one of those things I try to avoid wherever possible, even though it's probably perfectly safe in this case

ryanb-gds (approved):
Looks good, couple of observations but both pretty minor

features/step_definitions/visual_editor_steps.rb
lauraghiorghisor-tw: Good shout, let me try something.

features/step_definitions/visual_editor_steps.rb
lauraghiorghisor-tw: We can always use a `@html_attachment` and not include the title at all. It seemed a pattern in other tests, in the sense that it reads well.

PR #8977, opened 2024-04-22
Switch Worldwide Corporate Information Pages to use edition links

brucebolt:
In https://github.com/alphagov/publishing-api/pull/2708, we allowed Worldwide Corporate Information Pages to use edition links. This change switches those pages to use edition links, allowing us to move towards having Worldwide Organisation editionable.

This will empty the non-edition links. Once all these pages are republished, another PR will remove the non-edition links method.

[Trello card](https://trello.com/c/2VqUBhuS)

### DIFF
--- a/app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
+++ b/app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
@@ -24,13 +24,17 @@ module PublishingApi
       content.merge!(
         document_type:,
         schema_name: "worldwide_corporate_information_page",
+        links: {
+          parent: [item.worldwide_organisation.content_id],
+          worldwide_organisation: [item.worldwide_organisation.content_id],
+        },
       )
     end
 
     def links
       {
-        parent: [item.worldwide_organisation.content_id],
-        worldwide_organisation: [item.worldwide_organisation.content_id],
+        parent: [],
+        worldwide_organisation: [],
       }
     end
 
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
```diff
@@ -24,13 +24,17 @@ def content
       content.merge!(
         document_type:,
         schema_name: "worldwide_corporate_information_page",
+        links: {
+          parent: [item.worldwide_organisation.content_id],
+          worldwide_organisation: [item.worldwide_organisation.content_id],
+        },
```
yndajas, line 30: I don't have much context on this, but I'm guessing this is the expected format for/location of edition links?

yndajas (approved):

app/presenters/publishing_api/worldwide_corporate_information_page_presenter.rb
brucebolt: Yeah, the edition links exist within the content item, which is sent to Publishing API as part of the `put_content` request.  An example is in the `WorldwideOrganisationPresenter`: https://github.com/alphagov/whitehall/blob/9b9656da2c5e1025d0d727e6cc03fbd5e5c79998/app/presenters/publishing_api/worldwide_organisation_presenter.rb#L30

Whereas non-edition links exist in a separate method, which is sent to Publishing API as a `patch_links` request.

PR #8983, opened 2024-04-23
Allow Editionable Worldwide Organisations to have attachments

brucebolt:
Corporate Information Pages (CIPs) attached to non-editionable Worldwide Organisations currently support attachments.

As we are embedding the "About Us" CIP into the main Editionable Worldwide Organisation, we need them to support attachments too, so these can be migrated across.

[Trello card](https://trello.com/c/V3UmyopH)

### DIFF
--- a/app/models/editionable_worldwide_organisation.rb
+++ b/app/models/editionable_worldwide_organisation.rb
@@ -7,6 +7,7 @@ class EditionableWorldwideOrganisation < Edition
   include Edition::Organisations
   include Edition::Roles
   include Edition::WorldLocations
+  include Attachable
 
   has_many :pages, class_name: "WorldwideOrganisationPage", foreign_key: :edition_id, dependent: :destroy, autosave: true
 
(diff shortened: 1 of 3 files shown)


app/models/editionable_worldwide_organisation.rb
```diff
@@ -64,6 +65,11 @@ def process_associations_before_save(new_edition)
   include AnalyticsIdentifierPopulator
   self.analytics_prefix = "WO"
 
+  delegate :alternative_format_contact_email, to: :sponsoring_organisation, allow_nil: true
+  def sponsoring_organisation
+    lead_organisations.first
+  end
```
yndajas, line 71: Are these required for supporting attachments?

brucebolt: Yes, turns out it was also required in this model.

That's because we eventually need it [here](https://github.com/alphagov/whitehall/blob/bf678254b8bdb3930735f569f065c1985b668e1c/app/helpers/govspeak_helper.rb#L31C70-L31C102), which is called from this model's [presenter](https://github.com/alphagov/whitehall/blob/bf678254b8bdb3930735f569f065c1985b668e1c/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb#L76C39-L76C63).

yndajas (approved):

PR #8986, opened 2024-04-23
Rake task to set a taxonomy topic email override

beccapearce:
When a document collection has a taxonomy topic email override field set, government frontend will show an email signup link for the taxonomy topic stored in the field. See code [here](https://github.com/alphagov/government-frontend/blob/main/app/views/shared/_document_collections_email_signup.html.erb#L1-L21)

This less than ideal complexity in the email sign up journey from some document collection pages was a necessary compromise to unblock the retirement of specialist topics. 

There will be documentation added to the developer docs to describe when this task should be run.

Trello card:
https://trello.com/c/OSr529Am/2080-workflow-to-manage-future-requests-from-hmrc-to-set-topic-level-email-notifications-on-new-not-yet-published-document-collection

### DIFF
--- a/lib/tasks/set_taxonomy_topic_email_override.rake
+++ b/lib/tasks/set_taxonomy_topic_email_override.rake
@@ -1,0 +1,25 @@
+desc "override email notifications for a document collection to a given taxonomy topic, can only be run on against a document that has never been published."
+task :set_taxonomy_topic_email_override, %i[document_collection_id taxon_content_id confirmation_string] => :environment do |_, args|
+  document_collection_id = args[:document_collection_id]
+  taxon_content_id = args[:taxon_content_id]
+  confirmation_string = args[:confirmation_string]
+
+  if document_collection_id.blank? || taxon_content_id.blank?
+    raise "Document collection ID and taxon content ID are required arguments"
+  end
+
+  document_collection = DocumentCollection.find_by(id: document_collection_id)
+  raise "Cannot find document collection with ID: #{document_collection_id}" unless document_collection
+
+  raise "Cannot set a taxonomy topic email override on previously published documents" if document_collection.document.live?
+
+  taxon_content_item = Services.publishing_api.get_content(taxon_content_id)
+  sub_message = "#{taxon_content_item.to_h['title']} for document collection #{document_collection.id}, #{document_collection.title}."
+
+  if confirmation_string == "run_for_real"
+    document_collection.update!(taxonomy_topic_email_override: taxon_content_id)
+    puts "Taxonomy topic email override set to #{sub_message}"
+  else
+    puts "This was a dry run. Taxonomy topic email override would have been set to #{sub_message}"
+  end
+end
--- a/test/unit/lib/tasks/set_taxonomy_topic_email_override_test.rb
+++ b/test/unit/lib/tasks/set_taxonomy_topic_email_override_test.rb
@@ -1,0 +1,74 @@
+require "test_helper"
+require "rake"
+
+class SetTaxonomyTopicEmailOverrideRake < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  teardown { task.reenable }
+
+  describe "#set_taxonomy_topic_email_override" do
+    let(:task) { Rake::Task["set_taxonomy_topic_email_override"] }
+
+    it "raises an error unless document collection ID and taxon content ID are both provided" do
+      e = assert_raises(StandardError) { task.invoke }
+      assert_equal e.message, "Document collection ID and taxon content ID are required arguments"
+    end
+
+    it "raises an error if document collection does not exist" do
+      e = assert_raises(StandardError) { task.invoke("nothing-at-this-id", "taxon-content-id") }
+      assert_equal e.message, "Cannot find document collection with ID: nothing-at-this-id"
+    end
+
+    it "raises an error if document collection has previously been published" do
+      document_collection = create(:document_collection, :published)
+      e = assert_raises(StandardError) { task.invoke(document_collection.id.to_s, "taxon-content-id") }
+      assert_equal e.message, "Cannot set a taxonomy topic email override on previously published documents"
+    end
+
+    it "raises an error if no content item is returned for the given taxon content ID" do
+      document_collection = create(:document_collection, :draft)
+      bad_taxon_content_id = "123abc"
+      stub_publishing_api_does_not_have_item(bad_taxon_content_id)
+      assert_raises(GdsApi::HTTPNotFound) do
+        task.invoke(document_collection.id.to_s, bad_taxon_content_id.to_s)
+      end
+    end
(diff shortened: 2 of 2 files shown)


hannako:
Thanks for pushing up the WIP commit! Some initial thoughts:

- To avoid code duplication and lots of test set up, shall we just have a single task where you pass in a dry run flag? The behaviour would be that unless you explicitly pass in `dry_run: false`, then by default the task is run in dry run mode. 
- I think the namespace should match the rake task file name, currently one is email_override and the other is set_email_override
- I know you're kind of midway between having a rake task with all the logic, and splitting it out into a module. So perhaps the test file name will change later, but in case it's helpful I think it would be more normal to have the test that's invoking the rake task in the test/unit/lib/tasks folder - that's the pattern in this repo right? 
- It might be easier as you suggested earlier, to just put all the logic in the rake task rather than have the module.
- I think you can add an `extend Minitest::Spec::DSL` line to your test file, and then you can have context and describe blocks.

lib/tasks/set_taxonomy_topic_email_override.rake
```diff
@@ -0,0 +1,25 @@
+desc "override email notifications for a document collection to a given taxonomy topic, can only be run on against a document that has never been published."
+task :set_taxonomy_topic_email_override, %i[document_collection_id taxon_content_id dry_run] => :environment do |_, args|
+  document_collection_id = args[:document_collection_id]
+  taxon_content_id = args[:taxon_content_id]
+  dry_run = args[:dry_run] || "dry_run_true"
```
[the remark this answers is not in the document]
unoduetre: Also maybe instead of `dry_run_false` something like `no_dry_run` as it sounds a bit more natural, at least to me? What do you think?

lib/tasks/set_taxonomy_topic_email_override.rake
```diff
@@ -0,0 +1,25 @@
+desc "override email notifications for a document collection to a given taxonomy topic, can only be run on against a document that has never been published."
+task :set_taxonomy_topic_email_override, %i[document_collection_id taxon_content_id dry_run] => :environment do |_, args|
```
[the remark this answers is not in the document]
hannako: Thanks Mat. Yeah this is a really good point, but the weirdness was somewhat intentional: 

If a document collection is published after this field is set, there is no way to unset it. So if it was set in error and then published, the only solution would be to un-publish the document, and create a fresh new document without the field. But there would also be clear-up required if users sign up to alerts in the intervening period. We are creating this task as there's a chance a single dept might want to customise their emails in the near future (and we'd agreed to let them do so). I'm working on documentation in the dev docs at the moment to link to some content/process guidance. TLDR: I wanted to make the chances of it being run by accident as **minimal** as possible. Hence defaulting to run in `dry run` unless one very specific non-boolean string was passed in. But I take your point, I don't want it to be confusing. Maybe `confirmation_string` is better?

How about this:

```
desc "override email notifications for a document collection to a given taxonomy topic, can only be run on against a document that has never been published."
task :set_taxonomy_topic_email_override, %i[document_collection_id taxon_content_id confirmation_string] => :environment do |_, args|
  document_collection_id = args[:document_collection_id]
  taxon_content_id = args[:taxon_content_id]
  confirmation_string = args[:confirmation_string]

  if document_collection_id.blank? || taxon_content_id.blank?
    raise "Document collection ID and taxon content ID are required arguments"
  end

  document_collection = DocumentCollection.find_by(id: document_collection_id)
  raise "Cannot find document collection with ID: #{document_collection_id}" unless document_collection

  raise "Cannot set a taxonomy topic email override on previously published documents" if document_collection.document.live_edition_id.present?

  taxon_content_item = Services.publishing_api.get_content(taxon_content_id)
  sub_message = "#{taxon_content_item.to_h['title']} for document collection #{document_collection.id}, #{document_collection.title}."

  if confirmation_string == "run_for_real"
    document_collection.update!(taxonomy_topic_email_override: taxon_content_id)
    puts "Taxonomy topic email override set to #{sub_message}"
  else
    puts "This was a dry run. Taxonomy topic email override would have been set to #{sub_message}"
  end
end
```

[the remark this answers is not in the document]
unoduetre: That looks really nice and the code is shorter.

unoduetre (approved):

lib/tasks/set_taxonomy_topic_email_override.rake
```diff
@@ -0,0 +1,25 @@
...
+  confirmation_string = args[:confirmation_string]
+
+  if document_collection_id.blank? || taxon_content_id.blank?
+    raise "Document collection ID and taxon content ID are required arguments"
+  end
+
+  document_collection = DocumentCollection.find_by(id: document_collection_id)
+  raise "Cannot find document collection with ID: #{document_collection_id}" unless document_collection
+
+  raise "Cannot set a taxonomy topic email override on previously published documents" if document_collection.document.live_edition_id.present?
```
ryanb-gds, line 14: There's a [shorthand for this](https://github.com/alphagov/whitehall/blob/39df09dd26544ef5989dd162e1fa1bc7c123e72f/app/models/document.rb#L97) if you want it

lib/tasks/set_taxonomy_topic_email_override.rake
```diff
@@ -0,0 +1,25 @@
+desc "override email notifications for a document collection to a given taxonomy topic, can only be run on against a document that has never been published."
+task :set_taxonomy_topic_email_override, %i[document_collection_id taxon_content_id confirmation_string] => :environment do |_, args|
```
ryanb-gds, line 2: I'd say the general convention for the `confirmation_string` argument would be to call it `dry_run`. This convention is already present in Whitehall. The [change note remover](https://github.com/alphagov/whitehall/blob/39df09dd26544ef5989dd162e1fa1bc7c123e72f/lib/tasks/data_hygiene.rake#L5) makes use of it

ryanb-gds (approved):
Couple of suggestions but essentially looks good

lib/tasks/set_taxonomy_topic_email_override.rake
hannako: Thanks - there was convo about this [here](https://github.com/alphagov/whitehall/pull/8986#pullrequestreview-2028137615) and in an earlier version Becca had two rake tasks (dry run and for reals).

Shall I go back to `dry_run` with a boolean? As app owner I think you get casting vote!

ryanb-gds: Haha, I hadn't read that. @unoduetre is right about it typically being true by default. Probably the nicest thing would be a confirmation step so the operator would have to type "y" or similar to proceed. Quite a few CLI libraries make that fairly easy but Rake does not, as far as I can see, so probably stick with what you have for now. I don't feel strongly about it.

PR #8990, opened 2024-04-24
Add offices to home page lists on new edition creation

brucebolt:
When a new edition is created, we need to assign the home page offices to the "home page list" for the new edition.

This must be done after save, as otherwise the record can't be associated correctly with the new edition.


JonathanHallam (approved):
Nice one, thank you :)

PR #8993, opened 2024-04-24
Add UI for republishing a document by slug

yndajas:
[Trello](https://trello.com/c/egFDeEYg)

This is the final individual content republishing task that needs a UI. This one is a little more complicated than the rest since it's actually a document's editions that are being republished, and which editions will be republished is the responsibility of a worker class. However, this does still play back to the user which content will be republished

## Screenshots

Note that there have been a few small changes since these screenshots:
- "Status" -> "State" in the editions table
- "Live" -> "Published" and "Live, withdrawn" -> "Withdrawn" for specific state descriptions (we now take the state directly from the edition being republished)

### Before

image

#### Full flow (draft edition)

image

image

image

image

#### Non-existing document

image

#### Live and draft

Screenshot 2024-04-25 at 16 20 59

#### Live

image

#### Unpublished

image

#### Withdrawn

image

---


app/workers/publishing_api_document_republishing_worker.rb
```diff
@@ -12,46 +12,80 @@
...
+        { method: :send_live_edition, edition: live_edition, status: "Live" },
+      ]
+    elsif there_is_a_newer_draft?
+      [
+        { method: :patch_links },
+        { method: :send_live_edition, edition: live_edition, status: "Live" },
+        { method: :send_draft_edition, edition: pre_publication_edition, status: "Draft" },
+      ]
+    end
+  end
```
yndajas, line 87: Not sure how I feel about this pattern/methods that return (arrays of) these kinds of custom meta-ish-data objects. The `status` key is also very much a presentational detail, which might be out of place in a worker, but I suppose this is a `#preview` method

It works, but I'm definitely open to other suggestions for how to go about fetching a list of editions for playing back to the user

I did consider having an entirely separate method returning this information, rather than abstracting into this actions methods used in both `#perform` and `#preview`, but then there's much greater risk of them getting out of sync and previews being inaccurate

test/unit/app/workers/publishing_api_document_republishing_worker_test.rb
```diff
@@ -106,4 +106,76 @@ class PublishingApiDocumentRepublishingWorkerTest < ActiveSupport::TestCase
...
+      document = create(:document, editions: [build(:superseded_edition)])
+
+      assert_equal PublishingApiDocumentRepublishingWorker.new.preview(document.id), []
+    end
+
+    context "when there are non-superseded editions" do
+      test "returns an action for the latest unpublished edition when the document has been unpublished" do
+        first_unpublished_edition = build(:unpublished_edition)
+        last_unpublished_edition = build(:unpublished_edition)
+        document = create(:document, editions: [first_unpublished_edition, last_unpublished_edition])
```
yndajas, line 121: A lot of repeated setup between `#perform` and `#preview`

app/workers/publishing_api_document_republishing_worker.rb
```diff
@@ -27,6 +27,27 @@ def perform(document_id, bulk_publishing = false)
...
+      return nil if action == :patch_links
+
+      edition = {
+        refresh_latest_unpublished_edition: latest_unpublished_edition,
+        refresh_pre_publication_edition: pre_publication_edition,
+        refresh_published_edition: published_edition,
+        refresh_withdrawn_edition: withdrawn_edition,
+      }[action]
+
+      raise Error if edition.blank? # unknown action
```
yndajas, line 45: Is there a way we can rewrite this so that we don't rely on errors being raised to tell us we're missing something? Would some kind of case statement guarantee that all of a defined set of actions are covered? Probably not in typeless Ruby land, right?

app/workers/publishing_api_document_republishing_worker.rb
```diff
@@ -27,6 +27,22 @@ def perform(document_id, bulk_publishing = false)
     end
   end
 
+  def preview_refreshed_editions(document_id)
```
ryanb-gds, line 30: I agree about this not being a great fit for the worker. Could always move this on to a `DocumentRepublisher` service or similar, which would be quite a Whitehall-ish thing to do.  I do like the "actions" concept, that's a nice way to organise the workflow.

Another option might be to move some of this onto the `Document` model. You could probably have `Document#has_republishable_editions?` and `Document#republishing_actions` methods which could be called from either the worker or the controller. Personally I'd probably go that route, I like rich Active Record models rather than passing models around to services, but either would probably be an improvement on this living in the worker.

You could probably go even further and the actions start becoming command objects in their own right or something... 🤔 might be overkill though 😂

yndajas: I like the sound of shifting this to `Document` model. Though given the actions are essentially the names of private methods in the worker, would it be odd for the model to know/be tied to those? I suppose we could consider the worker's responsibility simply to be defining an implementation for (and running) the actions as defined in the model

ryanb-gds: Yeah I haven't thought it through too much. Maybe therefore the actions do need to be first class concepts instead of just private methods. Anyway, don't let my philosophical musings get in the way of you getting a republishing UI into the world!

yndajas: And a fixed version: https://github.com/alphagov/whitehall/pull/8993/commits/697343f6af092e7cd6fa64b7c3f609ae00765b59

ryanb-gds: Looks pretty good, could we possibly do without the metaprogramming? If I'm correct that the metaprogramming is there to provide memoisation then I suspect we could do without it, thinking about it, because Active Record already [caches queries](https://guides.rubyonrails.org/caching_with_rails.html#sql-caching). So I don't think we're gaining much from it.

yndajas: It looks like that caching is [within the context of controller actions](https://guides.rubyonrails.org/caching_with_rails.html#sql-caching:~:text=it%27s%20important%20to%20note%20that%20query%20caches%20are%20created%20at%20the%20start%20of%20an%20action%20and%20destroyed%20at%20the%20end%20of%20that%20action%20and%20thus%20persist%20only%20for%20the%20duration%20of%20the%20action). I'd imagine that would work for the use case in this PR, but this worker is also used outside of controllers, and sometimes with `perform_async` or `perform_async_in_queue` - I'm not sure the caching would apply in those contexts?

ryanb-gds: I think Sidekiq also applies caching: https://github.com/sidekiq/sidekiq/wiki/Problems-and-Troubleshooting#activerecord-query-cache

You could probably have a play around using `binding.pry` or `byebug` to see what it actually does, maybe.

yndajas: Definitely looks like caching is happening based on the Rails server output, so I'm happy to remove the fancy memoisation! I'll sort out working those changes into the commit history

brucebolt (approved):
Looks good to me, just left a couple of very small suggestions.

app/controllers/admin/republishing_controller.rb
```diff
@@ -119,6 +119,17 @@ def republish_role
 
   def find_document; end
 
+  def search_document
+    document = Document.find_by(slug: params[:document_slug])
+
+    unless document
+      flash[:alert] = "Published document with slug '#{params[:document_slug]}' not found"
```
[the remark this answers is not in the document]
yndajas: Ah yes - I think that's a relic of my initial understanding of this task!

app/models/document.rb
```diff
@@ -147,6 +147,29 @@ def has_republishable_editions?
...
+        (:refresh_pre_publication_edition if pre_publication_edition&.valid?),
+      ].compact
+    elsif withdrawn_edition.present?
+      return [:refresh_withdrawn_edition]
+    end
+
+    [
+      :patch_links,
+      (:refresh_published_edition if published_edition.present?),
+      (:refresh_pre_publication_edition if pre_publication_edition&.valid?),
```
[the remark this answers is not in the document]
brucebolt: I think the word "refresh" is more descriptive of the actual process, but we've used "republish" in many places (including documentation and other apps), so I'd be inclined to stick with "republish".

PR #8996, opened 2024-04-25
Copy translations of social media accounts when creating new edition

brucebolt:
When creating a new edition of an edition with social media accounts, we need to also copy the translations.

This is not yet in use, as it is behind the Editionable Worldwide Organisation feature flag.


app/models/concerns/edition/social_media_accounts.rb
```diff
@@ -2,9 +2,15 @@ module Edition::SocialMediaAccounts
   extend ActiveSupport::Concern
 
   class Trait < Edition::Traits::Trait
-    def process_associations_before_save(edition)
+    def process_associations_after_save(edition)
```
yndajas, line 5: Is it worth mentioning the shift to after save in the commit message?

yndajas (approved):

PR #8997, opened 2024-04-25
Bump Ruby to 3.2.4

dj-maisy:
## What?
This bumps Ruby to 3.2.4.

## Related

* https://github.com/alphagov/govuk-ruby-images/pull/92


sengi:
We might be able to drop the patch version entirely and just use the `3.2` tag. (Or does whitehall still have some issue that stops us from doing that?)

sengi (commented):
Oh we should update /.ruby-version to match as well.

ryanb-gds (approved):

kevindew:
Just to flag - as per a Slack chat - that .ruby-version needs a full version for rbenv which is the tool used in development environments.

If we want to go down this road we'd need to update govuk-docker and change guidance on dev environments to use a different ruby version manager.

sengi:
Oh I think this was just a mistake — thanks for spotting!

PR #9004, opened 2024-04-29
Save visual editor content for editions

lauraghiorghisor-tw:
- Adding feature tests to cover the happy path
- Ensuring validation errors are rendered correctly on the edition form

[Trello card](https://trello.com/c/YXixNjiJ/2489-save-visual-editor-content-to-document-editions)


features/support/javascript.rb
```diff
@@ -76,12 +76,19 @@ def select_from_choices(value, options)
     select_has_value_as_text = all(".choices__list.choices__list--single", visible: false, text: value, wait: false).present?
     return if select_has_value_as_text || select.value == value
 
-    choices_option = div.find(
-      "div[role='option'][class='choices__item choices__item--choice choices__item--selectable']",
-      text: value,
-      visible: false,
-    )
-
+    choices_option = if options[:from] == "edition_publication_type_id"
```
dnkrj, line 79: I wonder if instead of adding a special case for `edition_publication_type_id` we try `role='option'` first and if that's not found change to `role='treeitem'`?
What do you think?

dnkrj (approved):
Looks good, just one comment

PR #9007, opened 2024-04-29
Make worldwide organisation body content consistent

brucebolt:
The body of an editionable worldwide organisation without any content was different to that of a non-editionable worldwide organisation (i.e. we were including an empty div as govspeak conversion was taking place).

Updating the edtionable presenter to ensure both types are consistent.

### DIFF
--- a/test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
+++ b/test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
@@ -212,4 +212,28 @@ class PublishingApi::EditionableWorldwideOrganisationPresenterTest < ActiveSuppo
 
     assert_not presented_item.content[:details].key? :default_news_image
   end
+
+  test "does not convert an absent body to govspeak" do
+    worldwide_org = create(:editionable_worldwide_organisation)
+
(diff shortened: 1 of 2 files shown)


test/unit/app/presenters/publishing_api/editionable_worldwide_organisation_presenter_test.rb
```diff
@@ -183,4 +183,28 @@ def present(...)
 
     assert_equal [lead_organisation_2.content_id, lead_organisation_1.content_id], presented_item.content.dig(:links, :sponsoring_organisations)
   end
+
+  test "does not convert an absent body to govspeak" do
+    worldwide_org = create(:editionable_worldwide_organisation)
+
+    # rubocop:disable Rails/SkipsModelValidations
```
callumknights, line 190: What's the reason for these Rubocop skips? Just curious more than anything

brucebolt: I've explained the reason in the commit message.

callumknights: Whoops, didn't see that, all good then

callumknights (approved):

PR #9008, opened 2024-04-29
Include `default_news_image` for editionable worldwide organisation

brucebolt:
This attribute is [included in the presenter](https://github.com/alphagov/whitehall/blob/607d5d19cf94d5b2681e1eb6c712d3ff4f80a81f/app/presenters/publishing_api/worldwide_organisation_presenter.rb#L299) for non-editionable worldwide organisations, but not the editionable worldwide organisation.

Therefore adding it to the presenter and copying the tests that are included for the non-editionable presenter.

### DIFF
--- a/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
@@ -4,6 +4,7 @@ module PublishingApi
     include ActionView::Helpers::UrlHelper
     include ApplicationHelper
     include OrganisationHelper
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/editionable_worldwide_organisation_presenter.rb
```diff
@@ -76,6 +65,24 @@ def body
       Whitehall::GovspeakRenderer.new.govspeak_edition_to_html(item)
     end
 
+    def details
```
callumknights, line 68: Nice little refactor

callumknights (approved):

PR #9009, opened 2024-04-29
Provide a rake task to allow developers to reinstate a deleted document

ryanb-gds:
Sometimes a document gets into a state where its live edition is deleted. We don't currently understand how this happens, but Publishing API and Content Store are not updated to reflect the fact that the edition has been deleted.

It is not possible to recover from this situation via the Whitehall user interface, but the Edition model does expose a method for creating a new draft edition from a deleted edition. This commit provides a rake task to call the method safely so that we can allow users to recover the deleted document.

Trello: https://trello.com/c/mIxPZUew

Original Zendesk ticket: https://govuk.zendesk.com/agent/tickets/5799605

### DIFF
--- a/lib/data_hygiene/deleted_document_restorer.rb
+++ b/lib/data_hygiene/deleted_document_restorer.rb
@@ -1,0 +1,32 @@
+class DataHygiene::DeletedDocumentRestorer
+  def initialize(document_id, user_email)
+    @document_id = document_id
+    @user_email = user_email
+  end
+
+  def run!
+    latest_edition = Edition.unscoped.where(document_id: @document_id).last
+
+    unless latest_edition.deleted?
+      raise RestoreDocumentError.latest_edition_not_deleted
+    end
+
+    user = User.where(email: @user_email).first
+
+    if user.nil?
+      raise RestoreDocumentError.user_not_found(@user_email)
+    end
+
+    latest_edition.create_draft(user, allow_creating_draft_from_deleted_edition: true)
+  end
+
+  class RestoreDocumentError < StandardError
+    def self.latest_edition_not_deleted
+      new("This document's latest edition is not deleted")
+    end
+
+    def self.user_not_found(user_email)
+      new("This document doesn't exist for user with email #{user_email}")
+    end
+  end
+end
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -32,4 +32,16 @@ namespace :data_hygiene do
       speech.save!(validate: false)
     end
   end
+
+  desc "Restore a deleted document by creating a new edition from its latest deleted draft"
+  task :restore_deleted_document, %i[document_id user_email] => :environment do |_, args|
+    deleted_document_restorer = DataHygiene::DeletedDocumentRestorer.new(args[:document_id], args[:user_email])
+
+    begin
+      deleted_document_restorer.run!
+      puts "Created a new draft for document with ID #{args[:document_id]}"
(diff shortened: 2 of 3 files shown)


lib/tasks/data_hygiene.rake
```diff
@@ -32,4 +32,18 @@ namespace :data_hygiene do
...
+  task :restore_deleted_document, %i[document_id user_email] => :environment do |_, args|
+    deleted_document_restorer = DataHygiene::DeletedDocumentRestorer.new(args[:document_id], args[:user_email])
+
+    begin
+      deleted_document_restorer.run!
+      puts "Created a new draft for document with ID #{args[:document_id]}"
+    rescue DataHygiene::DeletedDocumentRestorer::DocumentStateError
+      puts "Error: This document's most recent edition is not deleted."
+    rescue DataHygiene::DeletedDocumentRestorer::UserNotFound
+      puts "Error: Unable to find user with email #{args[:user_email]}"
```
ChrisBAshton, line 46: These `puts` statements duplicate what you've captured in the `deleted_document_restorer.rb`. I'd opt for an approach like:

```suggestion
  rescue DataHygiene::DeletedDocumentRestorer::DocumentStateError, DataHygiene::DeletedDocumentRestorer::UserNotFound => e
    puts "Error: #{e.message}"
```

Maybe even have the exception classes inherit from a parent so that the exception-handling code is easier to read:

```suggestion
  rescue DataHygiene::DeletedDocumentRestorer::Exception => e
    puts "Error: #{e.message}"
```

lib/data_hygiene/deleted_document_restorer.rb
```diff
@@ -0,0 +1,26 @@
+require "active_record"
```
ChrisBAshton, line 1: Surprised the linter doesn't shout about the missing newline 🤔 

```suggestion
require "active_record"

```

lib/tasks/data_hygiene.rake
```diff
@@ -32,4 +32,18 @@ namespace :data_hygiene do
       speech.save!(validate: false)
     end
   end
+
+  desc "Restore a deleted document by creating a new edition"
```
ChrisBAshton, line 36: I'd probably be a bit more explicit here, just so it's crystal clear for the developer who stumbles across this in a year's time.

```suggestion
  desc "Restore a deleted document by creating a new edition from its latest deleted draft"
```

Do you think the task should also only fail if the latest non-deleted edition is "superseded"?
Or should this task be generic enough to, e.g., retrieve a draft that someone has deleted but now wants to get back (but where the published document is otherwise in a healthy state)?

ChrisBAshton (commented):
LGTM - a few questions/suggestions below, but nothing major 👍

lib/data_hygiene/deleted_document_restorer.rb
ryanb-gds: I don't even need active record, I was toying with the idea of throwing its `RecordNotFound` error but that didn't feel right. Forgot to take it out but have done so now.

lib/tasks/data_hygiene.rake
ryanb-gds: Good shout on the description.

I was pondering the question about whether to constrain the task to documents where the latest non-deleted edition is "superseded". I couldn't really decide, but I couldn't see any obvious harm so I left it as is for simplicity's sake. If you'd feel more comfortable with ensuring it can only be used in the superseded case, I'd be happy to do so.

ChrisBAshton (approved):

PR #9017, opened 2024-05-02
Set `.ruby-version` to `3.2.4`

pezholio:
Trying to get Whitehall set up using `govuk-docker` and I get the following error:

```
ruby-build: definition not found: 3.2

The following versions contain `3.2' in the name:
....
```

My rbenv is up to date, and after chatting with some other devs, this leads me to believe that rbenv doesn’t support minor versions.


kevindew (approved):
Looks good 👍 

Yes, rbenv doesn't support fuzzy matching, nor does asdf. rvm and chruby do but have much lower userbases.

pezholio:
Ta for the approval, can someone merge for me please, as I don't have the right permissions

PR #9018, opened 2024-05-02
Implement exiting the visual editor experience for private-beta users (Editions)

lauraghiorghisor-tw:
As part of releasing a private-beta trial for a new visual editor feature, we are offering the possibility to also exit the trial for a specific document. The private-beta will only offer limited content editing options, so we must ensure that users with advanced markdown needs can fall back to using the standard govspeak markdown editor.

Users in the trial will have a special permission, which has already been implemented.

A user in the trial will always create a new document with the visual editor enabled, and that document wil remain marked with a DB flag ensuring that all other users with permissions will continue editing the document in the visual editor. A user in the trial may exit the visual editor, and that document will remain exited for all subsequent users. A user outside the trial may also irreversibly exit a document. Users with JS disabled, be it in the trial or outside the trial, should always have a seamless experience where they can use the default textarea. Users with JS disabled will
also implicitly exit the document from the trial.

The implementation is done primarily in JS, hiding or showing the visual editor and govspeak editor respectively. We have used a hidden field whose value we set on the server, and then modify on the client via JS. The value of the hidden field gets committed on save.

Changes:
- Add a migration to introduce a visual_editor field to editions
- Set the field to true for all editions created by users in the visual editor experiment
- Render it as a hidden field on the edition form defaulting to false (To handle the non-JS case)
- Toggle the field to true when the visual editor is shown (To ensure persistence of the visual editor)
- Introduce a button to exit the visual editor (Displayed only when the visual editor is shown)
- Always render the Govspeak Editor (instead of a backup textarea as before)
- JS to handle toggling of the editor and hidden field when the button is clicked
- The hidden field name is required in the visual editor template to help facilitate extending the behaviour to other document types (produces a clear error message)
- JS to make sure that the ID is set to either the contenteditable or govspeak textarea as appropriate
to ensure that errors are correctly linked
- Modify the Govspeak Editor JS to get textarea by classname instead of ID to ensure compatibility with previous change

Co-authored-by: Daniel Karaj <daniel.karaj@digital.cabinet-office.gov.uk>

[Trello card](https://trello.com/c/geeWlpCj/2492-implement-exiting-the-visual-editor-editionsl)

Screenshot 2024-05-03 at 16 06 53

### DIFF
--- a/app/assets/javascripts/components/visual-editor.js
+++ b/app/assets/javascripts/components/visual-editor.js
@@ -5,24 +5,63 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
 ;(function (Modules) {
   function VisualEditor(module) {
     this.module = module
-
-    this.content = module.querySelector('.app-c-visual-editor__content')
-    this.container = module.querySelector('.app-c-visual-editor__container')
-    this.textarea = module.querySelector(
-      '.app-c-visual-editor__textarea-wrapper textarea'
-    )
   }
 
   VisualEditor.prototype.init = function () {
-    this.textarea.classList.add('app-c-visual-editor__textarea--hidden')
-    const id = this.textarea.getAttribute('id')
-    this.textarea.removeAttribute('id')
+    this.content = this.module.querySelector('.app-c-visual-editor__content')
+    this.container = this.module.querySelector(
+      '.app-c-visual-editor__container'
+    )
+    this.textarea = this.module.querySelector(
+      '.app-c-visual-editor__govspeak-editor-wrapper textarea'
+    )
 
     new window.GovspeakVisualEditor(this.content, this.container, this.textarea) // eslint-disable-line no-new
 
-    this.container
-      .querySelector('div[contenteditable="true"]')
-      .setAttribute('id', id)
+    this.govspeakEditorwrapper = this.module.querySelector(
+      '.app-c-visual-editor__govspeak-editor-wrapper'
+    )
+    this.visualEditorWrapper = this.module.querySelector(
+      '.app-c-visual-editor__visual-editor-wrapper'
+    )
+    this.exitButton = this.module.querySelector(
+      '.js-app-c-visual-editor__exit-button'
+    )
+    this.visual_editor_flag = document.querySelector(
+      '.app-c-visual-editor__hidden-field'
+    )
+    this.contentEditable = this.container.querySelector(
+      'div[contenteditable="true"]'
+    )
+
+    this.exitButton.addEventListener('click', this.hideVisualEditor.bind(this))
+    this.showVisualEditor()
+  }
+
+  VisualEditor.prototype.showVisualEditor = function () {
+    this.govspeakEditorwrapper.classList.add(
+      'app-c-visual-editor__govspeak-editor-wrapper--hidden'
+    )
+    this.visualEditorWrapper.classList.add(
+      'app-c-visual-editor__visual-editor-wrapper--show'
+    )
+    this.visual_editor_flag.value = true
+
+    this.contentEditable.setAttribute('id', this.textarea.getAttribute('id'))
+    this.textarea.removeAttribute('id')
+  }
+
+  VisualEditor.prototype.hideVisualEditor = function () {
+    this.govspeakEditorwrapper.classList.remove(
+      'app-c-visual-editor__govspeak-editor-wrapper--hidden'
+    )
+    this.visualEditorWrapper.classList.remove(
+      'app-c-visual-editor__visual-editor-wrapper--show'
+    )
+    this.visual_editor_flag.value = false
+
+    this.textarea.setAttribute('id', this.contentEditable.getAttribute('id'))
+    this.contentEditable.removeAttribute('id')
   }
 
   Modules.VisualEditor = VisualEditor
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -239,6 +239,7 @@ private
       :read_consultation_principles,
       :all_nation_applicability,
       :speaker_radios,
+      :visual_editor,
       :logo_formatted_name,
       {
         all_nation_applicability: [],
@@ -338,6 +339,9 @@ private
     edition_locale = edition_params[:primary_locale] || I18n.default_locale
     I18n.with_locale(edition_locale) do
       @edition = LocalisedModel.new(new_edition, edition_locale)
+      if @edition.visual_editor.nil?
+        @edition.visual_editor = Flipflop.govspeak_visual_editor? && current_user.can_see_visual_editor_private_beta?
+      end
     end
   end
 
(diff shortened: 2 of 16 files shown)


app/controllers/admin/editions_controller.rb
```diff
@@ -338,6 +339,9 @@ def build_edition
     edition_locale = edition_params[:primary_locale] || I18n.default_locale
     I18n.with_locale(edition_locale) do
       @edition = LocalisedModel.new(new_edition, edition_locale)
+      if @edition.visual_editor.nil?
+        @edition.visual_editor = Flipflop.govspeak_visual_editor? && current_user.can_see_visual_editor_private_beta?
+      end
```
ryanb-gds, line 344: I'm on the fence about whether this should be done on the controller. There's an [after_initialize](https://guides.rubyonrails.org/active_record_callbacks.html#after-initialize-and-after-find) callback we could use to do this automatically on the model. It does feel like more part of the domain model logic than the application logic. On the other hand Active Record models knowing about FlipFlop and the current user seems a bit grim, but that's sort of inevitable with AR.

app/views/admin/editions/_standard_fields.html.erb
```diff
@@ -64,17 +64,24 @@
...
         },
         name: "edition[body]",
+        id: "edition_body",
         value: edition.body,
+        rows: 20,
         error_items: errors_for(edition.errors, :body),
         right_to_left: form.object.translation_rtl?,
-        id: "edition_body",
+        data_attributes: {
+          image_ids: edition.images.map { |img| img[:id] }.to_json,
```
ryanb-gds, line 80: Is this in the scope of this PR?

db/migrate/20240501173819_add_visual_editor_to_editions.rb
```diff
@@ -0,0 +1,5 @@
+class AddVisualEditorToEditions < ActiveRecord::Migration[7.1]
+  def change
+    add_column :editions, :visual_editor, :boolean
```
ryanb-gds, line 3: Part of me still wonders whether we're going to want to show when the Visual Editor was exited at some point. I suppose we can do another migration if and when that becomes necessary

features/renders_govspeak_editor_no_js.feature
```diff
@@ -6,8 +6,15 @@ Feature: Renders govspeak editor if JS is disabled
 
   Scenario: I create a new publication
     When I start creating a new publication
-    Then I should see the textarea instead of the visual editor
+    Then I should only see the textarea for the body field
+
+  Scenario: I edit a publication saved with visual editor
+    When I edit a publication saved with visual editor
+    Then I should only see the textarea for the body field
```
ryanb-gds, line 13: Should we say "govspeak editor" instead of "textarea" here (and elsewhere in these tests) for clarity, or is that two different things?

features/step_definitions/visual_editor_steps.rb
```diff
@@ -53,9 +78,29 @@
...
+  publication = create(:publication, visual_editor: nil)
+  visit edit_admin_publication_path(publication)
+end
+
+When(/^I update the publication in the govspeak editor$/) do
+  fill_in "edition_body", with: "Any old iron"
+end
+
+When(/^I update the publication in the visual editor$/) do
+  find(".ProseMirror").base.send_keys("Any old iron")
```
ryanb-gds, line 95: This is interesting, bet it took some figuring out how to do

features/visual-editor.feature
```diff
@@ -10,15 +10,49 @@ Feature: Save edition content with visual editor
...
+    And I save and go to document summary
+    Then I should see the govspeak editor on subsequent edits of the publication
+
+  @javascript
+  Scenario: I edit a publication that has been previously exited
+    When I edit a publication that has been previously exited
+    Then I should see the govspeak editor
+    When I update the publication in the govspeak editor
+    And I save and go to document summary
+    Then I should see the govspeak editor on subsequent edits of the publication
```
ryanb-gds, line 41: I feel these lines are a bit redundant, but I can understand if you want to keep them around for completeness sake. Likewise for the test below about pre-existing documents

ryanb-gds (approved):
👏  Lovely stuff. A few questions and suggestions, but nowt serious.

app/assets/javascripts/components/visual-editor.js
```diff
@@ -5,24 +5,63 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
...
+  VisualEditor.prototype.showVisualEditor = function () {
+    this.govspeakEditorwrapper.classList.add(
+      'app-c-visual-editor__govspeak-editor-wrapper--hidden'
+    )
+    this.visualEditorWrapper.classList.add(
+      'app-c-visual-editor__visual-editor-wrapper--show'
+    )
+    this.visual_editor_flag.value = true
+
+    this.contentEditable.setAttribute('id', this.textarea.getAttribute('id'))
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: The error summary link needs to point to this + we need to be able to give focus to the correct invalid element, and I believe it needs to be the body/textarea ID because the error pertains to a specific model field.

app/views/admin/editions/_standard_fields.html.erb
lauraghiorghisor-tw: The data attributes no, but we are now passing them down because we need to use them in the conditional rendering of the govspeak editor (from within the visual editor) 🥲😑☠️ 
And that we decided to do (instead of rendering the textarea as before) because otherwise when we exit the visual editor, we lose the preview functionality.

features/renders_govspeak_editor_no_js.feature
lauraghiorghisor-tw: Subtle differences left, right and centre.

If no JS -> then it is truly only the textarea, since the preview button is also not shown. This one makes some sense but yes the name of the step is a bit too technical. Let me think of an alternative.


The more concerning naming yet! is where we have gospeak editor (original one) vs govspeak editor (rendered from the visual editor template). Really nothing in it to the user but I just could not have steps named the same 😅

features/step_definitions/visual_editor_steps.rb
lauraghiorghisor-tw: Not when we have Demandadan.

features/visual-editor.feature
lauraghiorghisor-tw: I think it is good to call out that a user with permission does not have the ability to revert the flag set on the document (be it pre-existing or exited).

features/renders_govspeak_editor_no_js.feature
ryanb-gds: Hmm. Ideally cucumber tests would reflect the experience of the user, being very much a BDD framework. It is a bit of a blunt tool though so fair enough.

app/controllers/admin/editions_controller.rb
lauraghiorghisor-tw: I tried a naive 
```
  def set_visual_editor_flag
    if visual_editor.nil? && created_at.nil?
      self.visual_editor = Flipflop.govspeak_visual_editor? && current_user.can_see_visual_editor_private_beta?
    end
  end
```
but it seems we do not have access to the current user in here.

ryanb-gds: Yeah, fair enough. Seems the controller is the Railsy way

PR #9019, opened 2024-05-02
Update `.gitignore` for log files

brucebolt:
When running locally, log files like `log/development.log.0` are created.

These are currently not being ignored by the `.gitignore` file, so adding these to be ignored.


JonathanHallam (approved):
cheerrs

PR #9023, opened 2024-05-03
Republish organisation page when contact is updated

ChrisBAshton:
A bug was reported in Zendesk recently:
https://govuk.zendesk.com/agent/tickets/5818031

Someone set a contact to appear on the org homepage, but they never appeared. Running the
`publishing_api:republish:organisation_by_slug` rake task fixes the issue, so clearly the organisation is not being republished when it should be.

We already have callbacks for updating the organisation when a contact is created or destroyed, but not when a contact is updated. I've been able to consistently reproduce the bug whereby editing contact info does not get reflected on the org page. Adding this callback should fix that bug, and should prevent further issues of this nature.

The test is based on the code added in #4552, which describes a similar problem.

Tested on integration, works well: https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/organisations/online-procedure-rule-committee/contacts

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -16,6 +16,7 @@ class Contact < ApplicationRecord
 
   after_update :republish_dependent_editions
   after_update :republish_dependent_policy_groups
+  after_update :republish_organisation_to_publishing_api
 
   after_create :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
Thanks Chris. I always wonder if Publishing API could do this upstream via [dependency resolution](https://github.com/alphagov/publishing-api/blob/main/docs/dependency-resolution.md), but that doesn't seem to do what I think it should. Several people have explained to me why that is but I just can't seem to understand it.

PR #9024, opened 2024-05-03
Implement exiting the visual editor experience for private-beta users (HTML attachments)

lauraghiorghisor-tw:
As part of releasing a private-beta trial for a new visual editor feature, we are offering the possibility to also exit the trial for a specific document, including HTML attachments. The private-beta will only offer limited content editing options, so we
must ensure that users with advanced markdown needs can fall back to using the standard govspeak markdown editor.

Users in the trial will have a special permission, which has already been implemented.

A user in the trial will always create a new document with the visual editor enabled, and that document wil remain marked with a DB flag ensuring that all other users with permissions will continue editing the document in the visual editor.
A user in the trial may exit the visual editor, and that document will remain exited or all subsequent users. A user outside the trial may also irreversibly exit a document. Users with JS disabled, be it in the trial or outside the trial, should always have a seamless experience where they can use the default textarea. Users with JS disabled will also implicitly exit the document from the trial.

The implementation is done primarily in JS, hiding or showing the visual editor and govspeak editor respectively. We have used a hidden field whose value we set on the server, and then modify on the client via JS. The value of the hidden field gets committed on save.

[Trello card](https://trello.com/c/PhE8eKf6/2536-implement-exiting-the-visual-editor-html-attachments)


features/renders_govspeak_editor_no_permission.feature
```diff
@@ -19,4 +19,12 @@ Feature: Renders govspeak editor if no visual editor permission is given
   @javascript
   Scenario: I create a new HTML attachment
     When I start creating a new HTML attachment for publication "Publication with HTML attachments and visual editor"
-    Then I should see the govspeak editor
\ No newline at end of file
+    Then I should see the govspeak editor
+
+  @javascript
+  Scenario: I edit an HTML attachments saved with visual editor
```
ryanb-gds, line 25: Very minor, but incorrect pluralisation here

ryanb-gds (approved):
LGTM

PR #9030, opened 2024-05-09
Speed up republishing index

yndajas:
The republishing index page is currently very slow to load. It's running the entire `content` method on seven presenters only to use the title. This moves the simple title logic into its own method, so this can be called directly instead of calling a larger method and throwing away most of the data

I've deployed this to integration and the page now loads without noticeable delay


Gweaton (approved):
Nicely done!

PR #9046, opened 2024-05-15
Hide markdown help when visual editor is showing

lauraghiorghisor-tw:
## User story
As a Private Beta user
I do not want to see the GovSpeak Markdown guidance whilst using the VIsual Editor
So that I don’t get confused as it is not usable with the Visual Editor

## Acceptance Criteria
**Given** I am a Private Beta trial user
**When** I edit a new document
**And** I see the Visual Editor
**Then** the Markdown Help Guidance on the right side is not shown

[Trello card](https://trello.com/c/8Cqmgw0E/2490-hide-markdown-guidance-when-using-the-visual-editor)


dnkrj (approved):
It would be nice if we didn't have to reach out of the component all the way to the `govspeak-help` class but I think that with the way we write our javascript this solution is the most sensible.

PR #9048, opened 2024-05-16
Add translation validations for Editionable Worldwide Organisation parts

brucebolt:
This ensures that Worldwide Organisation Pages and Worldwide Offices can only have a translation if the parent Editionable Worldwide Organisation also has the same translation.

The condition was enforced in the controllers, but adding a validation to the models ensures this conditions is enforced everywhere.

[Trello card](https://trello.com/c/QrJ74EK3)

### DIFF
--- a/test/unit/app/models/contact_test.rb
+++ b/test/unit/app/models/contact_test.rb
@@ -292,4 +292,21 @@ class ContactTest < ActiveSupport::TestCase
       office.contact.destroy!
     end
   end
+
+  test "when associated with a WorldwideOffice should be valid when translated into a language that the worldwide office's associated organisation has" do
+    worldwide_organisation = create(:editionable_worldwide_organisation, translated_into: %i[de es fr])
+    office = create(:worldwide_office, edition: worldwide_organisation, worldwide_organisation: nil)
+    contact = create(:contact, contactable: office, translated_into: %i[fr])
+
+    assert contact.valid?
+  end
+
+  test "when associated with a WorldwideOffice should not be valid when translated into a language that the worldwide office's associated organisation does not have" do
+    worldwide_organisation = create(:editionable_worldwide_organisation, translated_into: %i[de es fr])
+    office = create(:worldwide_office, edition: worldwide_organisation, worldwide_organisation: nil)
+    contact = create(:contact, contactable: office, translated_into: %i[cy es-419])
+
+    assert_not contact.valid?
+    assert contact.errors[:base].include?("Translations 'cy, es-419' do not exist for this worldwide organisation")
+  end
 end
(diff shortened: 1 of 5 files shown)


test/unit/app/models/contact_test.rb
```diff
@@ -292,4 +292,21 @@ class ContactTest < ActiveSupport::TestCase
       office.contact.destroy!
     end
   end
+
+  test "when associated with a WorldwideOrganisationPage should be valid when translated into a language that the worldwide organisation has" do
```
yndajas, line 296: `when associated with a WorldwideOffice should be valid when translated into a language that the worldwide office's associated organisation has`?

test/unit/app/models/contact_test.rb
```diff
@@ -292,4 +292,21 @@ class ContactTest < ActiveSupport::TestCase
...
+    worldwide_organisation = create(:editionable_worldwide_organisation, translated_into: %i[de es fr])
+    office = create(:worldwide_office, edition: worldwide_organisation, worldwide_organisation: nil)
+    contact = create(:contact, contactable: office, translated_into: %i[fr])
+
+    assert contact.valid?
+  end
+
+  test "when associated with a WorldwideOrganisationPage should not be valid when translated into a language that the worldwide organisation does not have" do
```
yndajas, line 304: `when associated with a WorldwideOffice should not be valid when translated into a language that the worldwide office's associated organisation does not have`?

test/unit/app/models/editionable_worldwide_organisation_test.rb
```diff
@@ -179,7 +179,7 @@ class EditionableWorldwideOrganisationTest < ActiveSupport::TestCase
 
   test "should clone office and contact associations when new draft of published edition is created" do
     contact = create(:contact, translated_into: [:es])
-    published_worldwide_organisation = create(:editionable_worldwide_organisation, :published)
+    published_worldwide_organisation = create(:editionable_worldwide_organisation, :published, translated_into: [:es])
```
yndajas, line 182: Not sure if GDS likes to avoid red commits, but if so is it worth squashing this into the last commit?

yndajas (approved):

test/unit/app/models/contact_test.rb
brucebolt: Thanks.  I copy and pasted this from the other commit, but forgot to change the model name! 🤦‍♂️

test/unit/app/models/contact_test.rb
brucebolt: Have updated.

test/unit/app/models/editionable_worldwide_organisation_test.rb
brucebolt: Seems like a good idea.  I've squashed them into one now.

PR #9061, opened 2024-05-20
Rake task: Mark documents as political for an organisation

tw-raulgracia:
New task to mark all documents as political 

- New rake task under election namespace to mark all documents                                 
for a given organisation as politcal and then republish them                                                             
- Note: it updates both published and in pre publication state editions                                                       
for each document                                                                                                           

[Trello Task](https://trello.com/c/2MVkrWue)


lib/tasks/election.rake
```diff
@@ -59,4 +59,25 @@ namespace :election do
...
+  task :mark_documents_as_political_for, [:slug] => :environment do |_t, args|
+    org = Organisation.find_by!(slug: args[:slug])
+    puts "Marking all documents as political for #{org.name}..."
+
+    editions = org.editions
+    editions.published.update_all(political: true)
+    editions.in_pre_publication_state.update_all(political: true)
+
+    org.editions.published.map(&:document_id).each do |document_id|
+      PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
```
ryanb-gds, line 77: Thinking about it @tw-raulgracia, we don't actually need to do this here, because when the election actually happens there's a task which republishes everything anyway: https://github.com/alphagov/whitehall/blob/00ae5b36e7ab8f219eaac3a1eaa8063030151351/lib/tasks/election.rake#L16

So we can drop the call to the republishing worker

lib/tasks/election.rake
```diff
@@ -59,4 +59,25 @@ namespace :election do
       puts e.message
     end
   end
+
+  desc "
+  Mark all documents of a given organisation as political
+  Usage:
+    rake election:mark_documents_as_political_for[organisation_slug]
+  "
+  task :mark_documents_as_political_for, [:slug] => :environment do |_t, args|
```
ryanb-gds, line 68: @johnpauldickie has confirmed that we need to apply this only to documents published on GOV.UK after a certain date for each organisation, so we'll need to add a date parameter to the command

ryanb-gds (changes requested):
We have a bit more clarity on requirements now, so we need to change a couple of things

ryanb-gds (approved):

lib/tasks/election.rake
```diff
@@ -59,4 +59,27 @@ namespace :election do
...
     end
   end
+
+  desc "
+  Mark all documents of a given organisation as political
+  Usage:
+    rake election:mark_documents_as_political_for[organisation_slug, '30-12-2024']
+  "
+  task :mark_documents_as_political_for, %i[slug date] => :environment do |_t, args|
+    date = Date.parse(args[:date])
```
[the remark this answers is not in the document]
ryanb-gds: Unfortunately the first published date for the organisations doesn't match the provided cut-off date, so we'll have to go with this for now.

ryanb-gds (approved):

PR #9063, opened 2024-05-21
Begin modelling a Content Library within Whitehall

pezholio:
In the Content Modelling team, we're beginning a piece of work to build a Content Library that will enable the embedding and sharing of discrete pieces of content across the Publishing estate. We wanted to ensure the following:

- Content Library items have the same publishing process as documents (2i etc)
- Content Library items can be categorised with meaning (so they're not just blocks of text)
- Content Library items can have a structure which is customisable

With this in mind, it made sense to host this within Whitehall and have Content Items inherit from the `Edition` model. To save polluting the codebase and ensure it was clear what this new code belonged to, I've decided to namespace all the code so far in a `ContentLibrary` module (I did consider an engine, but decided that _might_ be overkill).

To ensure extendability for Content Library Items, I've added a new JSON field to Editions, which ensures we can make the schemas for Content Items as flexible as we can without having to add a ton of extra fields to the table (I know there's already a ton of fields that are used in some places, but not others).

I'm not usually a big fan of pushing code that isn't used anywhere yet, but I wanted to ensure that this code got under the noses of enough people so the wider Whitehall team is aware of what we're planning and make sure I'm not going down a route that doesn't fit with the wider technical direction.

Thoughts welcome!

### DIFF
--- a/app/controllers/admin/object_store/items_controller.rb
+++ b/app/controllers/admin/object_store/items_controller.rb
@@ -1,0 +1,27 @@
+module Admin::ObjectStore
+  class ItemsController < Admin::EditionsController
+  private
+
+    def edition_class
+      ObjectStore::Item
+    end
+
+    def permitted_edition_attributes
+      (super << ObjectStore.field_names_for_item_type(params[:item_type])).flatten
+    end
+
+    def show_or_edit_path
+      if params[:save].present?
+        edit_admin_object_store_item_path(@edition.item_type, @edition)
+      else
+        admin_object_store_item_path @edition
+      end
+    end
+
+    def new_edition
+      edition = edition_class.new(item_type: params[:item_type])
+      edition.assign_attributes(new_edition_params)
+      edition
+    end
+  end
+end
--- a/app/helpers/admin/object_store/items_helper.rb
+++ b/app/helpers/admin/object_store/items_helper.rb
@@ -1,0 +1,39 @@
+module Admin
+  module ObjectStore
+    module ItemsHelper
+      def edition_name(edition)
+        edition.item_type.titleize
+      end
+
+      def render_partial_path(edition, partial_name)
+        render partial: partial_path(edition, partial_name), object: edition, as: edition.item_type
+      end
+
+      def partial_path(edition, partial_name)
+        dirname = edition.item_type.pluralize
+        File.join("admin", "object_store", dirname, partial_name)
+      end
+
+      def object_store_item_form(edition)
+        form_for edition, url: form_url_for_object_store_item(edition), as: edition.item_type do |form|
+          yield(form)
+          concat render("govuk_publishing_components/components/button", {
+            text: "Save",
+            value: "save",
+            name: "save",
+            data_attributes: {
+              module: "gem-track-click",
+              "track-category": "form-button",
+              "track-action": "object-store-item-button",
+              "track-label": "Save",
+            },
+          })
+        end
+      end
+
+      def form_url_for_object_store_item(edition)
+        admin_object_store_items_path(item_type: edition.item_type)
+      end
+    end
+  end
+end
--- a/app/models/object_store/item.rb
+++ b/app/models/object_store/item.rb
@@ -4,6 +4,10 @@ module ObjectStore
 
     after_initialize :add_field_accessors
 
+    validates :item_type, presence: true, on: :create
+    validate :item_type_cannot_change
+    validates :item_type, inclusion: { in: ObjectStore.item_types }
+
     def summary_required?
       false
     end
@@ -28,6 +32,12 @@ module ObjectStore
       end
     rescue UnknownItemType
       # Ignored
+    end
+
+    def item_type_cannot_change
+      if persisted? && item_type_changed?
+        errors.add :item_type, "cannot be changed after creation"
+      end
     end
   end
 end
--- a/app/models/object_store/item.rb
+++ b/app/models/object_store/item.rb
@@ -1,0 +1,33 @@
+module ObjectStore
+  class Item < Edition
+    store_accessor :details, :item_type
+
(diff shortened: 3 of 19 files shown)


ryanb-gds:
Thanks for sharing @pezholio!

My first instinct is not to build this in Whitehall, because there's a Publishing architecture diagram in my head that looks like this

Screenshot 2024-05-21 at 17 19 23

However, I think that's not generally considered a direction Publishing wants to go in, and I know we want to leverage the idea of editions for reusable content, so for cost reasons if nothing else building in Whitehall probably makes sense right now.

What I might suggest is to see if we can find a way of achieving this without bringing along all the baggage that comes with an edition to reusable content. I'm guessing for example that we don't want to make reusable content searchable, which is included on the Edition model by default, and there are several other edition behaviours that probably aren't relevant for reusable content.

To start with, perhaps we could look at moving the [workflow concern](https://github.com/alphagov/whitehall/blob/main/app/models/concerns/edition/workflow.rb) out of the edition namespace and decoupling it from the Edition model so that it can be included in any model with a "state" attribute?  Being able to compose these behaviours would potentially allow Whitehall to become much more flexible.

To give an example, at the moment whenever we have to opt out of a behaviour we have to [override the base model](https://github.com/alphagov/whitehall/blob/6bdbeb304047014060d39e16ded1f2b08486803b/app/models/editionable_worldwide_organisation.rb#L167) and then check if we've overridden it in the [one shared edition template](https://github.com/alphagov/whitehall/blob/6bdbeb304047014060d39e16ded1f2b08486803b/app/views/admin/editions/_standard_fields.html.erb#L132). A classic case of favouring inheritance over composition.

If, on the other hand, we used composition (i.e. compose traits to add behaviour to models rather than inherit from a single model), we would have more templates (e.g. editions and reusable content would be separate templates), but those templates would be much simpler and easier to reason about because they would have far fewer conditions in them - a sizeable benefit in my opinion. Likewise our models would be simpler because they would not need to override attributes that are not relevant to them.

Let me know if you think any of the above is worth a discussion - more than happy to jump on a call!

pezholio:
@ryanb-gds Thanks for the thoughts - that's all really interesting stuff. I agree I think there's probably too much baggage in that Edition model and I think sharing the traits around would probably give us much for flexibility. I'll see what I can do and hit you up for a chat later 👍

richardTowers:
Thanks for this @pezholio - great to have something concrete to talk about.

I wrote out a (too) long comment, but then got distracted and navigated away without saving it, so I'll summarise the key points before I get distracted again:

1) I'm fine with this being in Whitehall (or maybe Mainstream - possible discussion about authorization to be had)
2) I take a different view to Ryan on the desired architecture - I don't really like things like content-tagger being separate apps, and all other things being equal I'd like as few apps as possible presenting bits of user interface to publishers. Hence I'm happy for this to be in Whitehall, rather than a new micoservice / micro-publishing-app (🙅🏻 ).
3) I agree with Ryan about trying to avoid the baggage of Edition. A lot of its included modules / relations / scopes / validations don't make sense here (and probably for a fair few of the other models that inherit from it). Many of them do though, so it's worth doing a bit of thinking about how to factor it. I could imagine having a deeper inheritance structure like below, but that might cause Ryan to blow a gasket - I'm sure there's a better way.
4) Not having to explicitly exclude Content Library things from Edition searches is another nice benefit of not having them live in the edition table.
5) It would be interesting to think about how these things might look in publishing API - which fields will they have on them, and will we want to use links to compose them into bigger molecules (e.g. an organisation has many contacts, a contact has many email addresses etc.), or will we just refer to the atoms by their IDs?

```mermaid
classDiagram
    ApplicationRecord <|-- LiteEdition
    LiteEdition <|-- Edition
    LiteEdition <|-- ContentLibraryItem
    ContentLibraryItem <|-- EmailAddress
    ContentLibraryItem <|-- PhoneNumber
    LiteEdition : include Workflow
    LiteEdition : include AuditTrail
    LiteEdition : ... etc.
    Edition : include NullWorldLocations
    Edition : after_update republish_topical_event_to_publishing_api
    Edition : ... etc.
```

pezholio:
Cool, lots to think about here - I like the concept of a "Lite Edition" here. I've been playing with a bunch of things, and I do think that (unfortunately) no matter what we do, we're still going to have to need to use the Editions table, as without it, the relationship to the Documents model breaks down and we'll have to create a new Document-like model each time, which is probably not ideal. 

I think it might be worth having an "Editional" behaviour that sets up the relationship between our model and the Document, and defines the "editions" table as the one we want to use, we can then import the behaviours we want/need, while: 

a) keeping the relationship between the Document and our model, but also 
b) removing some of the baggage we have around our bloated Edition model. 

We still have a lot of baggage behind the scenes as far as the table goes, but I wonder if we can start to trim that down gradually by migrating some of the additional fields to a JSON column on an edition by edition basis. I could probably kick that work off by having the column I've called `content_item_data` just `data`, so it can be used for anything Edition related.

ryanb-gds:
That's a shame about the dependency on the document -> editions relationship. I was just having a look myself and concluded that to try to share the edition concerns isn't actually possible without refactoring more or less the entirety of Whitehall, because the concerns depend on fields and methods outside of their concern all over the place, e.g. [the Publishing concern knows about attachments](https://github.com/alphagov/whitehall/blob/main/app/models/concerns/edition/publishing.rb).

Until we refactor these concerns to decouple them from each other and from assuming knowledge about the model they're included in we're pretty stuck when it comes to sharing behaviour with non-edition models. Edition models also access other editions belonging to their document, which again prohibits sharing of behaviour. We could probably refactor these things but it would be a lot of effort. Having said that, if we did choose to refactor in future, having reusable content items in the mix as well won't make things any easier!

pezholio:
Closing in favour of https://github.com/alphagov/whitehall/pull/9112

PR #9074, opened 2024-05-24
Add rake task to output worldwide organisation version history

brucebolt:
This is a temporary task and will only be run once to extract the history of all worldwide organisation corporate information pages.

Once it has been run, the code will be deleted.

[Trello card](https://trello.com/c/EQFZBnI0)


Gweaton (approved):
As this isn't writing anything to the DB, I think I can live with there not being any tests!

PR #9075, opened 2024-05-24
Add bulk republishing page for all about us pages

Gweaton:

[Trello card](https://trello.com/c/EH75S676/1166-add-a-user-interface-for-whitehalls-all-about-us-republishing-rake-task)

## Changes in this PR

Adds new bulk republishing section to the republishing index page

Adds individual page for republishing all Organisation "About Us" pages, submitting the form on which will perform the same behaviour as the `all_about_pages` Rake task [here](https://github.com/alphagov/whitehall/blob/add-bulk-republishing-page-for-all-about-us-pages/lib/tasks/publishing_api.rake#L127).

This should replace the need for developers to run that rake task.

### New section on index page

image

### New republishing page

image

### Confirmation after republishing

image

### DIFF
--- a/app/controllers/admin/bulk_republishing_controller.rb
+++ b/app/controllers/admin/bulk_republishing_controller.rb
@@ -1,0 +1,40 @@
+class Admin::BulkRepublishingController < Admin::BaseController
+  include Admin::RepublishingHelper
+
+  before_action :enforce_permissions!
+
+  def confirm_all
+    @bulk_content_type_metadata = bulk_content_type_metadata[params[:bulk_content_type].underscore.to_sym]
+    return render "admin/errors/not_found", status: :not_found unless @bulk_content_type_metadata
+
+    @republishing_event = RepublishingEvent.new
+  end
+
+  def republish_all_organisation_about_us_pages
+    bulk_content_type_key = :all_organisation_about_us_pages
+    bulk_content_type_value = RepublishingEvent.bulk_content_types.fetch(bulk_content_type_key)
+    @bulk_content_type_metadata = bulk_content_type_metadata.fetch(bulk_content_type_key)
+    action = "#{@bulk_content_type_metadata[:name].upcase_first} have been queued for republishing"
+
+    @republishing_event = build_republishing_event(action:, bulk_content_type: bulk_content_type_value)
+
+    if @republishing_event.save
+      BulkRepublisher.new.republish_all_organisation_about_us_pages
+
+      flash[:notice] = action
+      redirect_to(admin_republishing_index_path)
+    else
+      render "confirm_all"
+    end
+  end
+
+private
+
+  def enforce_permissions!
+    enforce_permission!(:administer, :republish_content)
+  end
+
+  def build_republishing_event(action:, bulk_content_type:)
+    RepublishingEvent.new(user: current_user, reason: params.fetch(:reason), action:, bulk_content_type:, bulk: true)
+  end
+end
--- a/app/helpers/admin/republishing_helper.rb
+++ b/app/helpers/admin/republishing_helper.rb
@@ -1,0 +1,14 @@
+module Admin::RepublishingHelper
+  include Rails.application.routes.url_helpers
+
+  def bulk_content_type_metadata
+    @bulk_content_type_metadata ||= {
+      all_organisation_about_us_pages: {
+        id: "all-organisation-about-us-pages",
+        name: "all organisation 'About Us' pages",
+        republishing_path: admin_bulk_republishing_all_organisation_about_us_pages_republish_path,
+        confirmation_path: admin_bulk_republishing_all_confirm_path("all-organisation-about-us-pages"),
+      }
+    }
+  end
+end
--- a/app/helpers/admin/republishing_helper.rb
+++ b/app/helpers/admin/republishing_helper.rb
@@ -8,7 +8,23 @@ module Admin::RepublishingHelper
         name: "all organisation 'About Us' pages",
         republishing_path: admin_bulk_republishing_all_organisation_about_us_pages_republish_path,
         confirmation_path: admin_bulk_republishing_all_confirm_path("all-organisation-about-us-pages"),
-      }
+      },
     }
+  end
+
+  def republishing_index_bulk_republishing_rows
+    bulk_content_type_metadata.values.map do |content_type|
+      [
+        {
+          text: content_type[:name].upcase_first,
+        },
+        {
+          text: link_to(sanitize("Republish #{tag.span(content_type[:name], class: 'govuk-visually-hidden')}"),
+                        content_type[:confirmation_path],
+                        id: content_type[:id],
+                        class: "govuk-link"),
+        },
+      ]
+    end
   end
 end
--- a/app/views/shared/_republishing_form.html.erb
+++ b/app/views/shared/_republishing_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with(url: republishing_url, method: :post, data: {
+<%= form_with(url: republishing_path, method: :post, data: {
     module: "prevent-multiple-form-submissions",
   }) do %>
     <%= render("govuk_publishing_components/components/textarea", {
--- a/app/views/shared/_republishing_form.html.erb
+++ b/app/views/shared/_republishing_form.html.erb
@@ -1,0 +1,17 @@
+<%= form_with(url: republishing_url, method: :post, data: {
+    module: "prevent-multiple-form-submissions",
+  }) do %>
+    <%= render("govuk_publishing_components/components/textarea", {
+      label: {
+        text: "What is the reason for republishing?",
+        heading_size: "m",
+      },
+      name: "reason",
+      id: "republishing_event_reason",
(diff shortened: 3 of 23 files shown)


test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      PublishingApiDocumentRepublishingWorker.expects(:perform_async_in_queue).with(
+        "bulk_republishing",
+        about_us_page_2.id,
+        true,
+      ).in_sequence(sequence)
+
+      BulkRepublisher.new.republish_all_organisation_about_us_pages
+    end
+  end
```
yndajas, line 26: I think the issue must be that some other tests are creating about us pages that are not cleaned up. Running this test in isolation is fine, but when running the full suite, another ID is expected to be passed to the worker

app/controllers/admin/bulk_republishing_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    @bulk_content_type_metadata = bulk_content_type_metadata[params[:bulk_content_type].underscore.to_sym]
+    return render "admin/errors/not_found", status: :not_found unless @bulk_content_type_metadata
+
+    @republishing_event = RepublishingEvent.new
+  end
+
+  def republish_all_organisation_about_us_pages
+    bulk_content_type = :all_organisation_about_us_pages
+    bulk_content_type_enum_index = RepublishingEvent.bulk_content_types.fetch(bulk_content_type)
+    action = "All organisation 'About Us' pages have been queued for republishing"
```
yndajas, line 16: Did we want to use string interpolation here to keep this change-proof?

yndajas, line 16: Did we want to use string interpolation here to keep this change-proof?

test/functional/admin/bulk_republishing_controller_test.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  test "GDS Admin users should be able to POST :republish_all_organisation_about_us_pages, creating a RepublishingEvent for the current user" do
+    BulkRepublisher.any_instance.expects(:republish_all_organisation_about_us_pages).once
+
+    post :republish_all_organisation_about_us_pages, params: { reason: "this needs republishing" }
+
+    newly_created_event = RepublishingEvent.last
+    assert_equal newly_created_event.user, current_user
+    assert_equal newly_created_event.reason, "this needs republishing"
+    assert_equal newly_created_event.action, "All organisation 'About Us' pages have been queued for republishing"
```
yndajas, line 35: Should we assert on the other two/three fields?

app/helpers/admin/republishing_helper.rb
```diff
@@ -4,11 +4,27 @@ module Admin::RepublishingHelper
...
       all_organisation_about_us_pages: {
-      id: "all-organisation-about-us-pages",
-      name: "all organisation 'About Us' pages",
-      republishing_path: admin_bulk_republishing_all_organisation_about_us_pages_republish_path,
-      confirmation_path: admin_bulk_republishing_all_confirm_path("all-organisation-about-us-pages"),
+        id: "all-organisation-about-us-pages",
+        name: "all organisation 'About Us' pages",
+        republishing_path: admin_bulk_republishing_all_organisation_about_us_pages_republish_path,
+        confirmation_path: admin_bulk_republishing_all_confirm_path("all-organisation-about-us-pages"),
+      },
```
yndajas: This should be in the previous commit

test/unit/app/helpers/admin/republishing_helper_test.rb
```diff
@@ -0,0 +1,16 @@
+require "test_helper"
+
+class Admin::RepublishingHelperTest < ActionView::TestCase
+  setup do
+    @first_row = bulk_content_type_index_rows.first
+  end
+
+  test "#bulk_content_type_index_rows capitalises the first letter of the bulk content type" do
+    assert_equal "All organisation 'About Us' pages", @first_row.first[:text]
```
yndajas, line 9: I think the assertion is the wrong away around here and below

app/helpers/admin/republishing_helper.rb
```diff
@@ -4,11 +4,27 @@ module Admin::RepublishingHelper
...
     }
-  }
+  end
+
+  def bulk_content_type_index_rows
```
yndajas, line 15: `bulk_content_type_republishing_index_rows` or `republishing_index_bulk_republishing_rows`?

features/bulk-republishing-content.feature
```diff
@@ -0,0 +1,13 @@
...
+  I want to be able to republish content in bulk
+  So that they reflect changes to their dependencies when this doesn't happen automatically
+
+  Background:
+    Given I am a GDS admin
+
+  Scenario: Republish all Organisation "About Us" pages
+    Given Published Organisation "About Us" pages exist
+    When I request a bulk republishing of the Organisation "About Us" pages
+    Then I can see the Organisation "About Us" pages have been scheduled for republishing
```
yndajas, line 12: "queued"

features/step_definitions/bulk_republishing_content_steps.rb
```diff
@@ -0,0 +1,23 @@
...
+  organisation_1 = create(:organisation)
+  create(
+    :about_corporate_information_page,
+    organisation: organisation_1,
+  )
+  organisation_2 = create(:organisation)
+  create(
+    :about_corporate_information_page,
+    organisation: organisation_2,
+  )
```
yndajas, line 11: This can probably be simplified as well:

```rb
2.times { create(:about_corporate_information_page) }
```

app/views/shared/_republishing_form.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    module: "prevent-multiple-form-submissions",
+  }) do %>
+    <%= render("govuk_publishing_components/components/textarea", {
+      label: {
+        text: "What is the reason for republishing?",
+        heading_size: "m",
+      },
+      name: "reason",
+      id: "republishing_event_reason",
+      error_items: errors_for(@republishing_event.errors, :reason),
```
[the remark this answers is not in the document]
Gweaton: ah sorry I must have missed that commit somehow... I suppose we could pass in the `republising_event` as a local too 👍

app/controllers/admin/bulk_republishing_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def confirm_all
+    @bulk_content_type_metadata = bulk_content_type_metadata[params[:bulk_content_type].underscore.to_sym]
+    return render "admin/errors/not_found", status: :not_found unless @bulk_content_type_metadata
+
+    @republishing_event = RepublishingEvent.new
+  end
+
+  def republish_all_organisation_about_us_pages
+    bulk_content_type = :all_organisation_about_us_pages
+    bulk_content_type_enum_index = RepublishingEvent.bulk_content_types.fetch(bulk_content_type)
```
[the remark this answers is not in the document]
Gweaton: I found `bulk_content_type` a little confusing, as I thought it was actually a `type` rather than an integer, maybe `bulk_content_type_integer` would be a diplomatic compromise?

test/unit/app/helpers/admin/republishing_helper_test.rb
Gweaton: Can never get these the right way round, I tried putting the opposite of what I thought it should be 😅

test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -0,0 +1,22 @@
...
+      queue_sequence = sequence("queue")
+
+      2.times do
+        about_us_page = create(:about_corporate_information_page)
+
+        PublishingApiDocumentRepublishingWorker
+          .expects(:perform_async_in_queue)
+          .with("bulk_republishing", about_us_page.id, true)
+          .in_sequence(queue_sequence)
+      end
```
yndajas, line 17: If there's a preference for doing setup before (pre-)assertions, we could do this, but the current approach feels easier to read

```rb
      Array.new(2).map { create(:about_corporate_information_page) }.each do |about_us_page|
        PublishingApiDocumentRepublishingWorker
          .expects(:perform_async_in_queue)
          .with("bulk_republishing", about_us_page.id, true)
          .in_sequence(queue_sequence)
      end
```

test/unit/app/services/bulk_republisher_test.rb
yndajas: So I think the actual issue was that it need to be the `document_id` not the `id`. The lack of cleanup curiously seems not to matter

app/controllers/admin/bulk_republishing_controller.rb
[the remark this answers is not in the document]
yndajas: I've gone with key and value for now, but I'm open. It's a little confusing since `bulk_content_types` on the class returns a hash of string keys and integer values while `bulk_content_type` on an instance just returns the string key

yndajas (approved):

PR #9080, opened 2024-05-28
Duplicate contact numbers and services on new edition

brucebolt:
We are currently copying the office and the associated contact when creating a new edition, however not the contact numbers or services.

This adds that duplication, so these things are not lost when a new edition is created.

[Trello card](https://trello.com/c/w7nXW8eZ)

### DIFF
--- a/app/models/editionable_worldwide_organisation.rb
+++ b/app/models/editionable_worldwide_organisation.rb
@@ -32,6 +32,10 @@ class EditionableWorldwideOrganisation < Edition
           new_office.contact.contact_numbers << contact_number.dup
         end
 
+        office.services.each do |service|
+          new_office.services << service
+        end
+
         new_office.save!
 
         if @edition.office_shown_on_home_page?(office)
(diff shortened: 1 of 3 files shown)


app/models/editionable_worldwide_organisation.rb
```diff
@@ -28,6 +28,12 @@ def process_associations_after_save(new_edition)
 
         new_office.contact = office.contact.dup
 
+        office.contact.contact_numbers.each do |contact_number|
+          new_office.contact.contact_numbers << contact_number.dup
+        end
+
+        new_office.save!
```
Gweaton, line 39: Was this actually saving before, or was the only "saving" happening on `add_office_to_home_page!` below? Struggling to understand how things were persisted before otherwise!

Gweaton (approved):
Looks good - if the answer to my question was "it wasn't saving before..." and was a bug, it might be worth updating the commit message to explain that too, but I may have misunderstood the original code.

app/models/editionable_worldwide_organisation.rb
brucebolt: I'll update the commit message, but the reason was we weren't saving some records (I think it was only those that weren't being adding to the homepage).

PR #9081, opened 2024-05-29
Add, then remove, default `bulk` value for `RepublishingEvent`s

Gweaton:

## Context

When we added the `bulk` field in https://github.com/alphagov/whitehall/pull/9075/commits/578590b2781ce85e1bbcdac1822a0a11b1c0ee22 it seemed to default to `false` when creating new `RepublishingEvent`s without us needing to set a default value.

It appears that before merging that PR, a rubocop update was merged and the PR branch not rebased with `main` after this. This caused Rubocop to fail on `main` due to this rubocop issue.

## Changes in this PR

Here we add an explicit `false` default value for `RepublishingEvent`s and ignore the Rubocop violation on the previous migration. We then remove the false value as all existing `RepublishingEvent`s will have been migrated to the new default value.

I've had to use `up` rather than `change` here as `change_column` is not reversible. According to a comment on the accepted answer on [this Stack Overflow
post](https://github.com/alphagov/whitehall/pull/9075/commits/578590b2781ce85e1bbcdac1822a0a11b1c0ee22),

```
If you need reversible migrations, put this in an up block rather
than a change block. You can leave the down block empty. It won't revert
the table to the original condition but the migration can be rolled
back.
```

### DIFF
--- a/db/migrate/20240521150403_add_bulk_and_bulk_content_type_to_republishing_events.rb
+++ b/db/migrate/20240521150403_add_bulk_and_bulk_content_type_to_republishing_events.rb
@@ -1,7 +1,7 @@
 class AddBulkAndBulkContentTypeToRepublishingEvents < ActiveRecord::Migration[7.1]
   def change
     change_table :republishing_events, bulk: true do |t|
-      t.boolean :bulk, null: false
+      t.boolean :bulk, null: false # rubocop:disable Rails/NotNullColumn
       t.integer :bulk_content_type
     end
   end
--- a/db/migrate/20240529093425_make_republishing_events_bulk_false_by_default.rb
+++ b/db/migrate/20240529093425_make_republishing_events_bulk_false_by_default.rb
@@ -1,0 +1,7 @@
+class MakeRepublishingEventsBulkFalseByDefault < ActiveRecord::Migration[7.1]
+  def up
+    change_column :republishing_events, :bulk, :boolean, default: false
+  end
+
+  def down; end
+end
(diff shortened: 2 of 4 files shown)


db/migrate/20240529093425_make_republishing_events_bulk_false_by_default.rb
```diff
@@ -0,0 +1,7 @@
+class MakeRepublishingEventsBulkFalseByDefault < ActiveRecord::Migration[7.1]
+  def up
+    change_column :republishing_events, :bulk, :boolean, default: false
+  end
```
yndajas, line 4: I thought it was previously only defaulting to `false` for existing records, but new records were correctly erroring if no value was provided? Will this mean we could end up with an accidental default in future?

Gweaton: You're right - I forgot we were passing in the `bulk: false` when Republishing. Will update the commit message. As we need a default, I think `false` is a sensible one, and safe.

Gweaton: Or are you suggesting we should ignore the error instead?

Gweaton: As discussed, we're now migrating all existing `RepublishingEvent`s `bulk` values where `nil` to `false`, then unset the default value to prevent accidentally creating non bulk `RepublishingEvent`s

db/migrate/20240521150403_add_bulk_and_bulk_content_type_to_republishing_events.rb
```diff
@@ -1,7 +1,7 @@
 class AddBulkAndBulkContentTypeToRepublishingEvents < ActiveRecord::Migration[7.1]
```
yndajas, line 1: In the commit message, it should be "prevent accidentally setting `RepublishingEvent`s as `bulk: false`"

> This is an intermediate step before setting this back to have no `null` value to prevent accidentally setting `RepublishingEvent`s as `bulk` when they're not in future changes to the code.

yndajas (approved):

PR #9087, opened 2024-05-29
Prevent long Nation Inapplicability Alternative URLs from breaking the database transaction

minhngocd:
URLs which are longer than 255 characters break the database limit, so we are introducing a character limit for URLs to prevent this from happening. Users should seek to shorten their URLs.

https://trello.com/c/CPM8XMsf

### DIFF
--- a/app/models/nation_inapplicability.rb
+++ b/app/models/nation_inapplicability.rb
@@ -10,6 +10,7 @@ class NationInapplicability < ApplicationRecord
 
   validates :nation_id, inclusion: { in: Nation.potentially_inapplicable.map(&:id) }
   validates :alternative_url, uri: true, allow_blank: true
+  validates :alternative_url, length: { maximum: 255 }
 
   attr_accessor :excluded
 
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -1,5 +1,8 @@
 en:
   activerecord:
+    attributes:
+      publication/nation_inapplicabilities:
+        alternative_url: "Alternative URL for excluded nation"
     errors:
       messages:
         invalid_date: "is not in the correct format"
@@ -11,3 +14,7 @@ en:
               invalid_date: First published date is not in the correct format
               blank: Enter a first published date
               inclusion: First published date must be between 1/1/1900 and the present
+        nation_inapplicability:
+          attributes:
+            alternative_url:
+              too_long: "is too long. Please shorten the URL to less than 255 characters."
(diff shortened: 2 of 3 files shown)


minhngocd:
Error message content might need a review

image

image

app/models/nation_inapplicability.rb
```diff
@@ -10,6 +10,7 @@ class NationInapplicability < ApplicationRecord
 
   validates :nation_id, inclusion: { in: Nation.potentially_inapplicable.map(&:id) }
   validates :alternative_url, uri: true, allow_blank: true
+  validates :alternative_url, length: { maximum: 255, too_long: "URL too long. Please shorten the URL to under 255 characters." }
```
ryanb-gds, line 13:
```suggestion
  validates :alternative_url, length: { maximum: 255, too_long: "is too long. Please shorten the URL to less than 255 characters." }
```

might be good enough for now?

ryanb-gds (commented):
I have a suggestion on the message. My suspicion is we might wait a while for content review. I don't like that the message starts with the field name which is not the same as the label either. We could avoid that by putting the message in [a locale file](https://github.com/alphagov/whitehall/blob/edition-refactoring/config/locales/en/errors.yml), which might be nice

minhngocd:
@ryanb-gds updated with your suggestions and also updated the screenshots to reflect 👍

minhngocd:
@ryanb-gds Managed to change the humanised version of the attribute so it should read better (updated screenshot). 

The only gripe now is the anchor link doesn't quite take you to the field because of the attribute id being coded into the view, and the error summary component not having any connection to this so will not know which of the 3 fields to link to. 

Could override the href so that if it is related to Nation Inapplicability alternative URL to just send it to Excluded Nations field anchor instead. but this would create an exceptional case in Error Summary Component which would be ugly. Open to suggestions or i think we could just call it functional enough here.

ryanb-gds:
> @ryanb-gds Managed to change the humanised version of the attribute so it should read better (updated screenshot).
> 
> The only gripe now is the anchor link doesn't quite take you to the field because of the attribute id being coded into the view, and the error summary component not having any connection to this so will not know which of the 3 fields to link to.
> 
> Could override the href so that if it is related to Nation Inapplicability alternative URL to just send it to Excluded Nations field anchor instead. but this would create an exceptional case in Error Summary Component which would be ugly. Open to suggestions or i think we could just call it functional enough here.

🤯 

I think it's probably functional enough? Having tested this on the current codebase the link for the alternative URL field in the error summary doesn't actually work anyway. Rails and conditional fields are not friends

ryanb-gds (approved):
Thanks for picking this up ⭐

PR #9089, opened 2024-05-30
Update / test remove MP letters task

richardTowers:
There were a few edge cases which came up when I ran this task last night:

1) MySQL LIKE statements are case insensitive, so the query found things like "Francis Implementation Team" (as it's got a lower case mp in) (not sure why this is in the letters of someone's name, but let's not visit that rabbit hole).
2) Even if it was case sensitive, `%MP%` would match people with MPhil degrees.
3) The `puts` statement being outside of the if statement caused some confusion - it would log things as updated that matched the query, but not the ruby gsub, even though they hadn't been updated.

I considered making the MySQL query stricter, but then we might actually miss things (like someone with MP in their letters, but separated with a comma, or lower case). Instead, I've just made it clear which ones are being skipped in the logs.

While writing tests for this, I also noticed an edge case where it would replace letters like "CBE MP VC" with "CBEVC" (i.e. it would remove both surrounding spaces from MP). Better to do it with split / join to avoid this I think.

### DIFF
--- a/lib/tasks/election.rake
+++ b/lib/tasks/election.rake
@@ -3,12 +3,14 @@ namespace :election do
   task remove_mp_letters: :environment do
     puts "Removing MP from MP's letters:"
     Person.where('letters LIKE "%MP%"').find_each do |person|
-      puts "updating #{person.name}"
-      new_letters = person.letters.gsub(/(^|\s)MP(\s|$)/, "")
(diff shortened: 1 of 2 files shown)


MahmudH (approved):
Thanks, LGTM 👍🏽

PR #9100, opened 2024-05-31
Allow Better Errors to work with Docker

pezholio:
I noticed that, in some circumstances, error traces were showing in plaintext when using Whitehall in Docker. After some research I found [this issue][1], the solution for which allows the host machine, Docker and Better Errors to play together nicely!

[1]: https://github.com/BetterErrors/better_errors/issues/270

### DIFF
--- a/config/initializers/better_errors.rb
+++ b/config/initializers/better_errors.rb
@@ -1,0 +1,3 @@
+if Rails.env.development?
+  BetterErrors::Middleware.allow_ip! "0.0.0.0/0"
+end


yndajas (approved):

kevindew:
It's cool that you've fixed it, however I would ask do we really want better_errors? It's been broken a long time so it's a dependency that can probably be shed for simplification of the project - and I really don't think it's a part of anyone's workflow for the app.

pezholio:
@kevindew - I totally get that, but I have found it useful when doing spikes / exploration. It might be worth flagging it for future discussion, but it'd be good if we could keep it around for now 👍

PR #9084, opened 2024-05-29
Delete blank contacts

callumknights:
Change to allow blank phone numbers to be submitted for Worldwide Office contact details.

Trello: https://trello.com/c/yaokYwP2/1162-bug-it-is-not-possible-to-delete-the-final-phone-number-from-a-worldwide-office

Credit to @brucebolt for this!

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #9111, opened 2024-06-04
Correct issue with new phone number deletion logic

callumknights:
NilClass errors where appearing for certain WWorgs, this should account for that.

### DIFF
--- a/app/controllers/admin/worldwide_offices_controller.rb
+++ b/app/controllers/admin/worldwide_offices_controller.rb
@@ -21,10 +21,8 @@ class Admin::WorldwideOfficesController < Admin::BaseController
 
     update_params[:service_ids] ||= []
 
-    if (contact_numbers_attributes = update_params.dig(:contact_attributes, :contact_numbers_attributes))
-      contact_numbers_attributes.each do |_id, contact_number_attribute|
-        contact_number_attribute[:_destroy] = true unless contact_number_attribute[:label].present? || contact_number_attribute[:number].present?
-      end
+    update_params.dig(:contact_attributes, :contact_numbers_attributes)&.each do |_id, contact_number_attribute|
+      contact_number_attribute[:_destroy] = true unless contact_number_attribute[:label].present? || contact_number_attribute[:number].present?
     end
 
     if @worldwide_office.update(update_params)
--- a/app/controllers/admin/worldwide_offices_controller.rb
+++ b/app/controllers/admin/worldwide_offices_controller.rb
@@ -23,7 +23,7 @@ class Admin::WorldwideOfficesController < Admin::BaseController
 
     if (contact_numbers_attributes = update_params.dig(:contact_attributes, :contact_numbers_attributes))
       contact_numbers_attributes.each do |_id, contact_number_attribute|
-        contact_number_attribute[:_destroy] = true if contact_number_attribute[:label].empty? && contact_number_attribute[:number].empty?
+        contact_number_attribute[:_destroy] = true unless contact_number_attribute[:label].present? || contact_number_attribute[:number].present?
       end
     end
 
--- a/test/functional/admin/worldwide_offices_controller_test.rb
+++ b/test/functional/admin/worldwide_offices_controller_test.rb
@@ -292,8 +292,6 @@ class Admin::WorldwideOfficesControllerTest < ActionController::TestCase
               contact_numbers_attributes: {
                 "0" => {
                   id: contact_number.id,
-                  label: contact_number.label,
-                  number: contact_number.number,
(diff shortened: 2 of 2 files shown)


app/controllers/admin/worldwide_offices_controller.rb
```diff
@@ -23,7 +23,7 @@ def update
 
     if (contact_numbers_attributes = update_params.dig(:contact_attributes, :contact_numbers_attributes))
       contact_numbers_attributes.each do |_id, contact_number_attribute|
-        contact_number_attribute[:_destroy] = true if contact_number_attribute[:label].empty? && contact_number_attribute[:number].empty?
+        contact_number_attribute[:_destroy] = true unless contact_number_attribute[:label].present? && contact_number_attribute[:number].present?
```
yndajas, line 26: Should the `&&` be `||` here now? Or is the contact number only valid if **both** fields are present?

app/controllers/admin/worldwide_offices_controller.rb
```diff
@@ -23,7 +23,7 @@ def update
 
     if (contact_numbers_attributes = update_params.dig(:contact_attributes, :contact_numbers_attributes))
       contact_numbers_attributes.each do |_id, contact_number_attribute|
```
yndajas, line 25: Could also do this in a separate commit to simplify this code, but it's not important!

https://github.com/alphagov/whitehall/pull/9084/files#r1624701069

app/controllers/admin/worldwide_offices_controller.rb
brucebolt, line 26: Will this mark for destruction if any one of either the number and label are not present?

Two tests with one for each of `label` and `number` being blank would clarify this.

brucebolt (commented):
Could you update the `PUT :update deletes contact numbers that are marked as to be destroyed` test to reflect that we've realised `label` and `number` aren't present when a contact has been deleted in the UI.  That will let us catch if this issue is reintroduced in the future.

app/controllers/admin/worldwide_offices_controller.rb
callumknights: Only if both are present that's correct

callumknights: Updated, can confirm it will mark a contact number for destruction if it's missing either field.

test/functional/admin/worldwide_offices_controller_test.rb
```diff
@@ -292,8 +292,8 @@ class Admin::WorldwideOfficesControllerTest < ActionController::TestCase
               contact_numbers_attributes: {
                 "0" => {
                   id: contact_number.id,
-                  label: contact_number.label,
-                  number: contact_number.number,
+                  label: "",
+                  number: "",
```
brucebolt, line 296: It appears these values aren't being included at all in the payload, hence the `undefined method `empty?' for nil:NilClass` error.  So both these lines can just be deleted.

test/functional/admin/worldwide_offices_controller_test.rb
```diff
@@ -306,6 +306,58 @@ class Admin::WorldwideOfficesControllerTest < ActionController::TestCase
...
+              },
+            },
+          },
+          id: office,
+          worldwide_organisation_id: worldwide_organisation,
+        }
+
+    assert_not ContactNumber.exists?(contact_number.id)
+  end
+
```
brucebolt, line 360: Thanks for adding both these tests, they look great.

test/functional/admin/worldwide_offices_controller_test.rb
callumknights: Sorted :)

test/functional/admin/worldwide_offices_controller_test.rb
```diff
@@ -306,6 +304,58 @@ class Admin::WorldwideOfficesControllerTest < ActionController::TestCase
...
+                  number: "",
+                },
+              },
+            },
+          },
+          id: office,
+          worldwide_organisation_id: worldwide_organisation,
+        }
+
+    assert_not ContactNumber.exists?(contact_number.id)
```
brucebolt, line 356: We shouldn't be deleting the record if a user leaves one field blank (as they will lose the data they've entered into the other field).  Instead there should be a validation error displayed (so change the `&&` to `||` in the controller, or switch back to using `if` but with `blank?` instead of `empty?`, as `nil` responds to that method).

callumknights: This should all be solved now

brucebolt (approved):

PR #9112, opened 2024-06-04
Initial spike for object store

pezholio:
This follows on from https://github.com/alphagov/whitehall/pull/9063 to add the initial scaffolding for an object store in Whitehall, which will allow us to manage reusable blocks of content.

An item for an object store is an editionable object which inherits from the root `Edition` model. We did look at defining a smaller subset of code that would make a model editionable, but this turned out to be too much of a rabbit hole.

Items can have a shape which is defined in config, for now we're having one content type, an email address, but going forward, we will want many different types of object. Ideally, we'd also like these content types to be defined by non-developers.

I've added a very basic create/edit flow for items too (which will need iteration, both for design and functionality), but this gives us a good idea of how this will all fit together, as well as highlighting the changes that need to be made upstream for us to send content to the Publishing API.

### DIFF
--- a/app/models/object_store/item.rb
+++ b/app/models/object_store/item.rb
@@ -1,0 +1,33 @@
+module ObjectStore
+  class Item < Edition
+    store_accessor :details, :item_type
+
+    after_initialize :add_field_accessors
+
+    def summary_required?
+      false
+    end
+
+    def body_required?
+      false
+    end
+
+    def previously_published
+      false
+    end
+
+  private
+
+    def add_field_accessors
+      config = ObjectStore.item_type_by_name(item_type)
+      config.fields.each do |field|
+        singleton_class.class_eval { store_accessor :details, field.name }
+        if field.required?
+          singleton_class.class_eval { validates field.name, presence: true }
+        end
+      end
+    rescue UnknownItemType
+      # Ignored
+    end
+  end
+end
--- a/app/models/object_store/item.rb
+++ b/app/models/object_store/item.rb
@@ -4,6 +4,10 @@ module ObjectStore
 
     after_initialize :add_field_accessors
 
+    validates :item_type, presence: true, on: :create
+    validate :item_type_cannot_change
+    validates :item_type, inclusion: { in: ObjectStore.item_types }
+
     def summary_required?
       false
     end
@@ -28,6 +32,12 @@ module ObjectStore
       end
     rescue UnknownItemType
       # Ignored
+    end
+
+    def item_type_cannot_change
+      if persisted? && item_type_changed?
+        errors.add :item_type, "cannot be changed after creation"
+      end
     end
   end
 end
--- a/app/models/object_store/item.rb
+++ b/app/models/object_store/item.rb
@@ -1,5 +1,7 @@
 module ObjectStore
-  class Item < Edition
+  class Item < ApplicationRecord
+    include Edition::LiteEdition
+
     store_accessor :details, :item_type
 
     after_initialize :add_field_accessors
@@ -8,16 +10,8 @@ module ObjectStore
     validate :item_type_cannot_change
     validates :item_type, inclusion: { in: ObjectStore.item_types }
 
-    def summary_required?
-      false
-    end
-
-    def body_required?
-      false
-    end
-
-    def previously_published
-      false
+    def title_required?
+      true
     end
 
   private
--- a/config/object_store/fields.yml
+++ b/config/object_store/fields.yml
@@ -1,0 +1,6 @@
+email_address:
+  properties:
+    email_address:
+      type: string
+  required:
+    - email_address
--- a/lib/object_store.rb
+++ b/lib/object_store.rb
@@ -1,0 +1,55 @@
+module ObjectStore
+  class UnknownItemType < StandardError
+    def initialize(item_type)
+      super("Unknown item type: #{item_type}")
+    end
+  end
+
+  class ItemType
+    attr_accessor :name, :fields
+
+    def field_names
+      fields.map(&:name)
+    end
+  end
+
+  class Field
+    attr_accessor :name, :type, :required
+
+    def required?
+      required == true
+    end
+  end
+
+  def self.configuration
+    @configuration ||= OpenStruct.new
+  end
+
+  def self.configure
+    yield(configuration)
+  end
+
+  def self.items
+    @items ||= configuration["fields"].map do |name, config|
+      properties = config["properties"] || []
+      item_type = ItemType.new
+      item_type.name = name
+      item_type.fields = properties.map do |field_name, _v|
+        field = Field.new
+        field.name = field_name
+        field.type = properties.dig(field_name, "type")
+        field.required = config["required"]&.include?(field_name)
+        field
+      end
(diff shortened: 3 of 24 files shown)


lib/object_store.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def self.configure
+    yield(configuration)
+  end
+
+  def self.config_for_item_type(item_type)
+    configuration.fields[item_type] || raise(UnknownItemType, item_type)
+  end
+
+  def self.fields_for_item_type(item_type)
```
tahb, line 20: It feels like there could be two small objects here: `item_type` and `field` as a separate concept from "field_definition" (the config representation).

We already want to send messages to them such as `item_type.required?` and `field.required?`, and I imagine this behaviour is going to extend somewhat?

lib/object_store.rb
```diff
@@ -0,0 +1,39 @@
...
+      super("Unknown item type: #{item_type}")
+    end
+  end
+
+  def self.configuration
+    @configuration ||= OpenStruct.new
+  end
+
+  def self.configure
+    yield(configuration)
```
tahb, line 13: I wonder if we should validate the structure of is configuration so we can make certain assumptions about it, ensuring a certain structure and making sure that we don't accidentally break it.

Something like [Psych](https://docs.ruby-lang.org/en/master/Psych.html#module-Psych-label-Reading+from+a+file) to load the yml and check the structure? We should probably figure out a way to put that into the CI tests too.

test/unit/lib/object_store_test.rb
```diff
@@ -0,0 +1,67 @@
+require "test_helper"
+
+class ObjectStoreTest < ActiveSupport::TestCase
+  setup do
```
tahb, line 4: Given we are making guards against missing fields I think we ought to test the sad path here too, covering happens when this config file is missing keys or contains unexpected keys or values.

app/models/object_store/item.rb
```diff
@@ -0,0 +1,33 @@
+module ObjectStore
+  class Item < Edition
```
tahb, line 2: I'm not sure yet what we need from Edition but given we said earlier we don't care about a lot of stuff it adds, do we need inheritance here or should we use composition?

```ruby
module LiteEdition
  def self.included(base)
    base.class_eval do
      store_accessor :details, :field1, :field2
      …

      validates :field1, presence: true
      …
    end
  end
end

class Item < ApplicationRecord
  include LiteEdition
end
```

app/models/object_store/item.rb
```diff
@@ -0,0 +1,33 @@
...
+      false
+    end
+
+    def previously_published
+      false
+    end
+
+  private
+
+    def add_field_accessors
```
tahb, line 25: If we were to reach for a field object we could decouple ourselves from ObjectStore a little with something like:

```ruby
def fields
  @fields ||= ObjectStore.fields_for_item_type(item_type)
end

def add_field_accessors
  @fields.each do |field|
    singleton_class.class_eval { store_accessor :details, field.name }
    if field.required?
      singleton_class.class_eval { validates fields.name, presence: true }
    end
  end
rescue UnknownItemType
  # Ignored
end
```

test/integration/object_store/create_email_test.rb
```diff
@@ -0,0 +1,32 @@
+require "test_helper"
+require "capybara/rails"
+
+class CreateEmailTest < ActionDispatch::IntegrationTest
```
tahb, line 4: I read this and mistook it for creating/sending an email. Maybe worth renaming to `create_email_content_item_test.rb` or `create_email_address_test.rb`.

tahb (commented):
Thanks for breaking down this change into all the steps 👍 

As someone who's just joined without any background of the plan I've got quite a few thoughts and given this is a foundational piece they feel like good conversations to have now, rather than later, so forgive the number of them please!

Naming: I notice we use "item" and "content" in different places. I think "item" may be too generic and so "content_item" and "content_item_type" might provide more clarity.

config/object_store/fields.yml
```diff
@@ -0,0 +1,6 @@
+email_address:
+  properties:
+    email_address:
+      type: string
```
ryanb-gds, line 4: Is it going to be possible to validate that this input is, in fact, a valid email address etc.? Or do, say, relative validation against another field value?

app/models/object_store/item.rb
ryanb-gds: I have been considering what a more composition based approach might look like in https://github.com/alphagov/whitehall/blob/edition-refactoring/docs/adr/0003-introduce-reusable-content-items.md, which might be of interest. A `LiteEdition` module might be an interesting step in that direction, to see how much behaviour we can lift off of the edition model.

ryanb-gds (commented):
I think we want to be really, really confident that we want to start storing data using JSON before we go down this path. I'd echo @tahb concerns about breaking with the convention of using database columns lightly. My suspicion is that as we start adding more options to the configuration, such as validation, multiple value fields etc., which are features people always want in CMS, we'll think it might have been easier to define the types in code after all and leverage Active Record.

If we do want to stick to JSON and a config file, then should we just be using the schema from Publishing API instead? I don't much like the idea of having to keep a config file in Whitehall in sync with a config file in Publishing API.

config/object_store/fields.yml
pezholio: Yeah, it's definitely worth considering, but I wanted to get the pipework in first

pezholio:
@ryanb-gds Thanks for the feedback I did give the idea of a "Lite Edition" a go in the previous PR, but I was concerned that it was going to eat up too much time and risk impacting the wider team too much, as there's a ton of moving parts in the concerns that make assumptions about the shape of the inheriting model.

Part of my thinking re: JSON columns was to give flexibility to create new content types on the fly. I did consider using the Publishing API schemas to define the types, but I was actually thinking about making the schemas on the Publishing API less strict, so an "Object Store Item" schema could, in effect accept any shape / size of object. 

Would be good to have a bit more of a detailed chat about all of this, there's definitely some interesting stuff in that ADR we could adapt too.

lib/object_store.rb
pezholio: That's a good call that, still getting my head back into a Ruby-ish way of thinking!

tahb:
@pezholio Great to see LiteEdition emerging 🎉 

It's difficult for me to see and feel confident that we've drawn the lines of responsibility right. 

Are we the first to have the need for a `LiteEdition`? I'm thinking that if any part of the codebase has needed something similar in the past it might be an idea to compare their needs and review what we've got.

If we're setting the path and the concept is agreed on then I wonder if a next step could be to move this set of changes into a Whitehall fork so that we can take it forward at pace. I think it would be good to test this theory out and be able to tweak it without having to overly worry about all the considerations of merging it into `main`. Once we've proved LiteEdition works and we've got a reasonable boundary we could make a clearer proposition? I'm assuming we can do a meaningful test with a local set up that has all parts of the stack we need?

pezholio:
Closing in favour of https://github.com/alphagov/whitehall/pull/9204

PR #9113, opened 2024-06-04
Add republishable content types method

yndajas:
We'll soon be adding the UI for bulk republishing content by type. The plan is to provide a select element with the republishable content types as options. This adds a couple of methods to support that

---

### DIFF
--- a/test/unit/app/helpers/admin/republishing_helper_test.rb
+++ b/test/unit/app/helpers/admin/republishing_helper_test.rb
@@ -1,6 +1,33 @@
 require "test_helper"
 
 class Admin::RepublishingHelperTest < ActionView::TestCase
+  test "#republishable_content_types returns a sorted list combining valid document types and other publishable content types" do
+    # we need to eager load here to ensure we have all the models
+    Rails.application.eager_load!
+
+    feature_flags.switch! :editionable_worldwide_organisations, true
+
+    expected_content_types = [
+      content_types[:omnipresent],
+      content_types[:editionable_worldwide_organisations_enabled],
+    ].flatten.sort
+    result_minus_test_types = republishable_content_types.reject { |type| content_types[:test_specific].include? type }
+
+    assert_equal expected_content_types, result_minus_test_types
+  end
+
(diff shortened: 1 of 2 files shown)


app/models/document.rb
```diff
@@ -60,6 +60,16 @@ def self.at_slug(document_types, slug)
     find_by(document_type: document_types, slug:)
   end
 
+  def self.document_types
```
brucebolt, line 63: Should this be called `edition_types`, since a Document can have multiple editions (all of a different document type)?

test/unit/app/helpers/admin/republishing_helper_test.rb
```diff
@@ -1,6 +1,38 @@
...
 
 class Admin::RepublishingHelperTest < ActionView::TestCase
+  test "#republishable_content_types returns a sorted list combining valid document types and other publishable content types" do
+    # we need to eager load here to ensure we have all the models
+    Rails.application.eager_load!
+
+    stubbed_document_types = %w[DocumentTypeOne DocumentTypeTwo]
+    Document.expects(:document_types).once.returns(stubbed_document_types)
+
+    expected_non_document_republishable_content_types = %w[Contact
```
brucebolt, line 11: Maybe this could just be an array that combines `DocumentTypeOne` and `DocumentTypeTwo`, to avoid the need to merge, flatten and sort on lines 28-31?  But not fussed either way if you prefer leaving as is.

brucebolt (commented):
I like this approach, rather than having a hardcoded list of things that can be republished.

I've just left a couple of really small comments and there's a failing test, but otherwise it's looking great.

app/models/document.rb
yndajas: I did wonder if this should belong on the `Edition` model, but then the `documents` table has a `document_type` column (which I suppose is just the latest edition's type, though I can't see that that's definitely the case from the model?). You'd know better than me though! Which of these three seems the most correct?

```rb
Edition.types # matches a database field
Document.document_types # matches a database field
Document.edition_types
```

brucebolt (approved):

PR #9122, opened 2024-06-05
Add UI for bulk republishing by type

yndajas:
[Trello](https://trello.com/c/2nRbOaFp/1200-add-a-user-interface-for-whitehalls-bulk-republishing-documents-by-type-rake-task)

## Changes in this PR

Adds a UI for bulk republishing all content types by their type to replace the `publishing_api:bulk_republish:document_type[document_type]` rake task.

There are a couple of things that _could_ be a bit neater here, e.g. text formatting of types, but as this isn't an interface that we're hoping will be used very often, I don't think it's worth spending the time making this look any prettier.

## Screenshots

Screenshot 2024-06-13 at 16 04 14

Screenshot 2024-06-13 at 16 04 23

Screenshot 2024-06-13 at 16 04 32

Screenshot 2024-06-13 at 16 04 38

### DIFF
--- a/app/controllers/admin/bulk_republishing_controller.rb
+++ b/app/controllers/admin/bulk_republishing_controller.rb
@@ -3,6 +3,43 @@ class Admin::BulkRepublishingController < Admin::BaseController
 
   before_action :enforce_permissions!
 
+  def new_by_type
+    @republishable_content_types_select_options = republishable_content_types_select_options
+  end
+
+  def new_by_type_redirect
+    redirect_to(admin_bulk_republishing_by_type_confirm_path(params[:content_type]))
+  end
+
+  def confirm_by_type
+    content_type = republishable_content_types.find { |type| type == params[:content_type].underscore.camelcase }
+    return render "admin/errors/not_found", status: :not_found unless content_type
+
+    @bulk_content_type_metadata = bulk_content_type_metadata.fetch(:all_by_type)
+    @content_type = content_type
+    @republishing_event = RepublishingEvent.new
+  end
+
+  def republish_by_type
+    content_type = republishable_content_types.find { |type| type == params[:content_type].underscore.camelcase }
+    return render "admin/errors/not_found", status: :not_found unless content_type
+
+    bulk_content_type_key = :all_by_type
+    @bulk_content_type_metadata = bulk_content_type_metadata.fetch(bulk_content_type_key)
+    action = "#{@bulk_content_type_metadata[:name].upcase_first} '#{content_type}' have been queued for republishing"
+    bulk_content_type_value = RepublishingEvent.bulk_content_types.fetch(bulk_content_type_key)
+    @republishing_event = build_republishing_event(action:, bulk_content_type: bulk_content_type_value, content_type:)
+
+    if @republishing_event.save
+      @bulk_content_type_metadata[:republish_method].call(content_type)
+
+      flash[:notice] = action
+      redirect_to(admin_republishing_index_path)
+    else
+      render "confirm_by_type"
+    end
+  end
+
   def confirm
     bulk_content_type_key = params[:bulk_content_type].underscore.to_sym
     @bulk_content_type_metadata = bulk_content_type_metadata[bulk_content_type_key]
@@ -36,7 +73,7 @@ private
     enforce_permission!(:administer, :republish_content)
   end
 
-  def build_republishing_event(action:, bulk_content_type:)
-    RepublishingEvent.new(user: current_user, reason: params.fetch(:reason), action:, bulk_content_type:, bulk: true)
+  def build_republishing_event(action:, bulk_content_type:, content_type: nil)
+    RepublishingEvent.new(user: current_user, reason: params.fetch(:reason), action:, bulk_content_type:, bulk: true, content_type:)
   end
 end
--- a/app/helpers/admin/republishing_helper.rb
+++ b/app/helpers/admin/republishing_helper.rb
@@ -49,15 +49,13 @@ module Admin::RepublishingHelper
   end
 
   def republishable_content_types
-    document_types = Edition.descendants.select { |descendant|
+    editionable_content_types = Edition.descendants.select { |descendant|
       next(false) if descendant == EditionableWorldwideOrganisation && !Flipflop.editionable_worldwide_organisations?
 
       descendant.descendants.count.zero?
     }.map(&:to_s)
 
-    includes_publish_to_publishing_api = ApplicationRecord.subclasses.select { |subclass| subclass.included_modules.include? PublishesToPublishingApi }.map(&:to_s)
-
-    [document_types, includes_publish_to_publishing_api].flatten.sort
+    [editionable_content_types, non_editionable_content_types].flatten.sort
   end
 
   def republishing_index_bulk_republishing_rows
@@ -74,5 +72,9 @@ module Admin::RepublishingHelper
         },
       ]
     end
+  end
+
+  def non_editionable_content_types
+    ApplicationRecord.subclasses.select { |subclass| subclass.included_modules.include? PublishesToPublishingApi }.map(&:to_s)
   end
 end
--- a/app/helpers/admin/republishing_helper.rb
+++ b/app/helpers/admin/republishing_helper.rb
@@ -45,6 +45,12 @@ module Admin::RepublishingHelper
         confirmation_path: admin_bulk_republishing_confirm_path("all-published-organisation-about-us-pages"),
         republish_method: -> { BulkRepublisher.new.republish_all_published_organisation_about_us_pages },
       },
+      all_by_type: {
+        id: "all-by-type",
+        name: "all by type",
+        new_path: admin_bulk_republishing_by_type_new_path,
+        republish_method: ->(type) { BulkRepublisher.new.republish_all_by_type(type) },
+      },
     }
   end
 
@@ -58,6 +64,15 @@ module Admin::RepublishingHelper
     [editionable_content_types, non_editionable_content_types].flatten.sort
   end
 
+  def republishable_content_types_select_options
+    republishable_content_types.map do |type|
+      {
+        text: type,
+        value: type.underscore.dasherize,
+      }
+    end
+  end
+
   def republishing_index_bulk_republishing_rows
     bulk_content_type_metadata.values.map do |content_type|
       [
@@ -66,7 +81,7 @@ module Admin::RepublishingHelper
         },
         {
           text: link_to(sanitize("Republish #{tag.span(content_type[:name], class: 'govuk-visually-hidden')}"),
-                        content_type[:confirmation_path],
+                        content_type[:new_path] || content_type[:confirmation_path],
                         id: content_type[:id],
                         class: "govuk-link"),
         },
--- a/app/services/bulk_republisher.rb
+++ b/app/services/bulk_republisher.rb
@@ -1,4 +1,6 @@
 class BulkRepublisher
+  include Admin::RepublishingHelper
+
   def republish_all_published_organisation_about_us_pages
     document_ids = Organisation.all.map(&:about_us).compact.pluck(:document_id)
     republish_by_document_ids(document_ids)
@@ -41,11 +43,32 @@ class BulkRepublisher
     republish_by_document_ids(document_ids)
   end
 
+  def republish_all_by_type(type)
+    begin
+      content_type_klass = type.constantize
+      raise NameError unless republishable_content_types.include?(type)
+    rescue NameError
+      raise "Unknown content type #{type}\nCheck the GOV.UK developer documentation for a list of acceptable document types: https://docs.publishing.service.gov.uk/manual/republishing-content.html#whitehall"
+    end
+
+    if non_editionable_content_types.include?(type)
+      republish_non_editionable_content_types(content_type_klass)
+    else
+      republish_by_document_ids(content_type_klass.pluck(:document_id))
+    end
+  end
+
 private
 
   def republish_by_document_ids(document_ids)
     document_ids.each do |document_id|
       PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
+    end
+  end
+
+  def republish_non_editionable_content_types(content_type_klass)
+    content_type_klass.find_each do |record|
+      Whitehall::PublishingApi.bulk_republish_async(record)
     end
   end
 end
--- a/app/views/admin/bulk_republishing/confirm_by_type.html.erb
+++ b/app/views/admin/bulk_republishing/confirm_by_type.html.erb
@@ -1,0 +1,16 @@
+<% content_for :page_title, "Republish #{@bulk_content_type_metadata[:name]} '#{@content_type}'" %>
+<% content_for :title, "Are you sure you want to republish #{@bulk_content_type_metadata[:name]} '#{@content_type}'?" %>
+<% content_for :title_margin_bottom, 6 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @republishing_event)) %>
+
+<div class="govuk-grid-row">
+  <section class="govuk-grid-column-two-thirds">
+    <p class="govuk-body govuk-!-margin-bottom-7">
+      This will queue <%= @bulk_content_type_metadata[:name] %> '<%= @content_type %>' to be republished.
+    </p>
+    <%= render partial: "shared/republishing_form", locals: {
+      republishing_event: @republishing_event,
+      republishing_path: admin_bulk_republishing_by_type_republish_path(@content_type.underscore.dasherize),
+    } %>
+  </section>
+</div>
--- a/features/bulk-republishing-content.feature
+++ b/features/bulk-republishing-content.feature
@@ -35,3 +35,13 @@ Feature: Bulk republishing content
     Given Published organisation "About us" pages exist
     When I request a bulk republishing of all published organisation "About us" pages
     Then I can see all published organisation "About us" pages have been queued for republishing
+
+  Scenario: Republish non-editionable content types
+    Given Contacts exist
+    When I select all of type "Contact" for republishing
+    Then I can see all of type "Contact" have been queued for republishing
+
+  Scenario: Republish editionable content types
+    Given Case Studies exist
+    When I select all of type "CaseStudy" for republishing
+    Then I can see all of type "CaseStudy" have been queued for republishing
--- a/test/functional/admin/bulk_republishing_controller_test.rb
+++ b/test/functional/admin/bulk_republishing_controller_test.rb
@@ -64,4 +64,87 @@ class Admin::BulkRepublishingControllerTest < ActionController::TestCase
     post :republish, params: { bulk_content_type: "all-published-organisation-about-us-pages", reason: "this needs republishing" }
     assert_response :forbidden
   end
+
+  test "GDS Admin users should be able to GET :new_by_type" do
+    get :new_by_type
+    assert_response :ok
+  end
+
+  test "Non-GDS Admin users should not be able to GET :new_by_type" do
+    login_as :writer
+
+    get :new_by_type
+    assert_response :forbidden
+  end
+
+  test "GDS Admin users should be redirected to :confirm_by_type on POST :new_by_type" do
+    post :new_by_type_redirect, params: { content_type: "organisation" }
+    assert_redirected_to admin_bulk_republishing_by_type_confirm_path(content_type: "organisation")
+  end
+
+  test "Non-GDS Admin users should not be able to POST :new_by_type" do
+    login_as :writer
+
+    post :new_by_type_redirect, params: { content_type: "organisation" }
+    assert_response :forbidden
+  end
+
+  test "GDS Admin users should be able to GET :confirm_by_type with a valid content type" do
+    get :confirm_by_type, params: { content_type: "organisation" }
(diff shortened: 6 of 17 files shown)


app/helpers/admin/republishing_helper.rb
```diff
@@ -49,15 +49,15 @@ def bulk_content_type_metadata
   end
 
   def republishable_content_types
-    document_types = Edition.descendants.select { |descendant|
+    editionable_document_types = Edition.descendants.select { |descendant|
```
Gweaton, line 58: rename -> `content_types` and below

app/services/bulk_republisher.rb
```diff
@@ -41,6 +43,20 @@ def republish_all_documents_with_publicly_visible_editions_with_html_attachments
...
   end
 
+  def republish_all_by_type(type)
+    begin
+      document_type = type.constantize
+    rescue NameError
+      raise "Unknown document type #{document_type}\nCheck the GOV.UK developer documentation for a list of acceptable document types: https://docs.publishing.service.gov.uk/manual/republishing-content.html#whitehall"
+    end
+
+    if non_editionable_document_types.map(&:constantize).include?(document_type)
```
Gweaton, line 53: let's remove `:constantize` and use the `type` here.

test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -201,4 +201,67 @@ class BulkRepublisherTest < ActiveSupport::TestCase
       BulkRepublisher.new.republish_all_published_organisation_about_us_pages
     end
   end
+
+  describe "#republish_all_by_type" do
+    context "for editionable document types" do
+      document_types = %w[CaseStudy
```
Gweaton, line 207: We probably don't need to loop over all these if we're stubbing the `non_editionable_document_types` return value. Let's just use one of these. Should we stub `non_editionable_document_types` further up?

app/services/bulk_republisher.rb
```diff
@@ -41,6 +43,20 @@ def republish_all_documents_with_publicly_visible_editions_with_html_attachments
...
+      republish_non_editionable_documents(document_type.all)
```
Gweaton, line 54: Has this method not been brought in from another commit?

app/services/bulk_republisher.rb
```diff
@@ -41,11 +43,31 @@ def republish_all_documents_with_publicly_visible_editions_with_html_attachments
...
+
 private
 
   def republish_by_document_ids(document_ids)
     document_ids.each do |document_id|
       PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
     end
   end
+
+  def republish_non_editionable_documents(documents)
```
Gweaton, line 68: Is it possible to use the document id here like we do above?

Gweaton: No, `BulkRepublishAsync` takes a model instance rather than an ID.

app/services/bulk_republisher.rb
```diff
@@ -41,11 +43,31 @@ def republish_all_documents_with_publicly_visible_editions_with_html_attachments
     republish_by_document_ids(document_ids)
   end
 
+  def republish_all_by_type(type)
+    begin
+      content_type = type.constantize
```
yndajas, line 48: 1. For increased safety, I guess we could search for the type in `republishable_content_types` from the helper before constantising?
2. Lower down when seeing `content_type.pluck` I was a little confused about plucking on a string. Is it worth calling this `type_klass`, `type_model`, `content_type_klass`, or `content_type_model`?

app/services/bulk_republisher.rb
```diff
@@ -41,11 +43,31 @@ def republish_all_documents_with_publicly_visible_editions_with_html_attachments
...
+    rescue NameError
+      raise "Unknown content type #{content_type}\nCheck the GOV.UK developer documentation for a list of acceptable document types: https://docs.publishing.service.gov.uk/manual/republishing-content.html#whitehall"
+    end
+
+    if non_editionable_content_types.include?(type)
+      republish_non_editionable_content_types(content_type.all)
```
yndajas, line 54: I think the benefits of using `find_each` in `republish_non_editionable_content_types` are lost by calling `all` here. Could you just pass in the model instead? I think the other method will continue to work the same

app/services/bulk_republisher.rb
```diff
@@ -41,11 +43,31 @@ def republish_all_documents_with_publicly_visible_editions_with_html_attachments
...
   def republish_by_document_ids(document_ids)
     document_ids.each do |document_id|
       PublishingApiDocumentRepublishingWorker.perform_async_in_queue("bulk_republishing", document_id, true)
     end
   end
+
+  def republish_non_editionable_content_types(content_type)
+    content_type.find_each do |type|
+      Whitehall::PublishingApi.bulk_republish_async(type)
+    end
```
yndajas, line 71: The difference between a "content type" and a "type" (instance) feels unclear - perhaps the enumerator variable here should be `instance` or `record`?

test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -201,4 +201,40 @@ class BulkRepublisherTest < ActiveSupport::TestCase
...
   end
+
+  describe "#republish_all_by_type" do
+    setup do
+      BulkRepublisher.any_instance.stubs(:non_editionable_content_types).returns(%w[Contact])
+    end
+
+    context "for editionable content types, like CaseStudy" do
+      test "republishes all content of the specified type via the PublishingApiDocumentRepublishingWorker" do
+        content_type = create(:case_study)
```
yndajas, line 212: Again this is an instance or record, so I find `content_type` a little misleading as a variable name (same below)

test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -201,4 +201,40 @@ class BulkRepublisherTest < ActiveSupport::TestCase
...
+
+        PublishingApiDocumentRepublishingWorker.expects(:perform_async_in_queue).with(
```
yndajas, line 214: Is it worth looping to create and expect are queued at least two records? At present, this would pass if the method just queued the first instance of the type (same below)

test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -201,4 +201,40 @@ class BulkRepublisherTest < ActiveSupport::TestCase
...
+          content_type = create(:contact)
+          Whitehall::PublishingApi.expects(:bulk_republish_async).with(content_type)
+          BulkRepublisher.new.republish_all_by_type("Contact")
+        end
+      end
+
+      describe "for non-existent content types" do
+        test "it returns an error" do
+          content_type = "SomeDocumentTypeThatDoesntExist"
+          assert_raises(StandardError, /Unknown content type #{content_type}/) do
```
yndajas, line 233: Per #9149, the string here needs to be in the keyword argument `match`, otherwise it's ignored and any message will pass. It doesn't need to be a `//`-style string either

test/functional/admin/bulk_republishing_controller_test.rb
```diff
@@ -64,4 +64,87 @@ class Admin::BulkRepublishingControllerTest < ActionController::TestCase
...
+  end
+
+  test "Non-GDS Admin users should not be able to GET :new_by_type" do
+    login_as :writer
+
+    get :new_by_type
+    assert_response :forbidden
+  end
+
+  test "GDS Admin users should be redirected to :confirm_by_type when posting to '/new'" do
```
yndajas, line 80: `'/new'` -> `:new_by_type` to be more specific?

test/functional/admin/bulk_republishing_controller_test.rb
```diff
@@ -64,4 +64,87 @@ class Admin::BulkRepublishingControllerTest < ActionController::TestCase
...
+  end
+
+  test "Non-GDS Admin users should not be able to access :confirm_by_type" do
+    login_as :writer
+
+    get :confirm_by_type, params: { content_type: "organisation" }
+    assert_response :forbidden
+  end
+
+  test "GDS Admin users should be able to POST :republish_by_type with a valid bulk content type and a reason, creating a RepublishingEvent for the current user" do
```
yndajas, line 109: `valid bulk content type` -> `valid content type`

app/controllers/admin/bulk_republishing_controller.rb
```diff
@@ -3,6 +3,43 @@ class Admin::BulkRepublishingController < Admin::BaseController
...
+    @republishing_event = RepublishingEvent.new
+  end
+
+  def republish_by_type
+    content_type = republishable_content_types.find { |type| type == params[:content_type].underscore.camelcase }
+    return render "admin/errors/not_found", status: :not_found unless content_type
+
+    bulk_content_type_key = :all_by_type
+    @bulk_content_type_metadata = bulk_content_type_metadata.fetch(bulk_content_type_key)
+    action = "#{@bulk_content_type_metadata[:name].upcase_first} '#{params[:content_type]}' have been queued for republishing"
```
yndajas, line 29: Would it be better to use `content_type` rather than `params[:content_type]` in the action here to get the class-y casing (how we present it to users in the select field and docs)?

app/controllers/admin/bulk_republishing_controller.rb
```diff
@@ -37,6 +74,6 @@ def enforce_permissions!
   end
 
   def build_republishing_event(action:, bulk_content_type:)
-    RepublishingEvent.new(user: current_user, reason: params.fetch(:reason), action:, bulk_content_type:, bulk: true)
+    RepublishingEvent.new(user: current_user, reason: params.fetch(:reason), action:, bulk_content_type:, bulk: true, content_type: params.fetch(:content_type, nil))
```
yndajas, line 77: And here, maybe it's better to pass in the upper camelcase content type we get in the action rather than using the kebabified version from the params

test/functional/admin/bulk_republishing_controller_test.rb
```diff
@@ -64,4 +64,87 @@ class Admin::BulkRepublishingControllerTest < ActionController::TestCase
...
+  test "GDS Admin users should encounter an error on POST :republish_by_type without a `reason` and be sent back to the confirm page" do
+    BulkRepublisher.expects(:new).never
+
+    post :republish_by_type, params: { content_type: "organisation", reason: "" }
+
+    assert_equal ["Reason can't be blank"], assigns(:republishing_event).errors.full_messages
+    assert_template "confirm_by_type"
+  end
+
+  test "GDS Admin users should see a 404 page when trying to POST :republish_by_type with an invalid bulk content type" do
```
yndajas, line 135: `bulk content type` -> `content type`

features/bulk-republishing-content.feature
```diff
@@ -35,3 +35,13 @@ Feature: Bulk republishing content
...
     Then I can see all published organisation "About us" pages have been queued for republishing
+
+  Scenario: Republish non-editionable content types
+    Given Contacts exist
+    When I select all of type "Contact" for republishing
+    Then I can see all of type "contact" have been queued for republishing
+
+  Scenario: Republish editionable content types
+    Given Case Studies exist
+    When I select all of type "CaseStudy" for republishing
```
yndajas, line 46: Minor, but these `When` step names would be more consistent with others:
- `I request a bulk republishing of all content of type "Contact"`
- `I request a bulk republishing of all content of type "CaseStudy"`

yndajas (commented):
Generally looks good - just some suggestions for tightening and tidying up!

app/views/admin/bulk_republishing/confirm_by_type.html.erb
```diff
@@ -0,0 +1,16 @@
+<% content_for :page_title, "Republish #{@bulk_content_type_metadata[:name]} '#{@content_type}'" %>
+<% content_for :title, "Are you sure you want to republish #{@bulk_content_type_metadata[:name]} '#{@content_type}'?" %>
+<% content_for :title_margin_bottom, 6 %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @republishing_event)) %>
+
+<div class="govuk-grid-row">
+  <section class="govuk-grid-column-two-thirds">
+    <p class="govuk-body govuk-!-margin-bottom-7">
+      This will schedule <%= @bulk_content_type_metadata[:name] %> '<%= @content_type %>' to be republished.
```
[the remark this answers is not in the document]
Gweaton: We've tended to use `schedule` elsewhere to represent things being added to queues, unless I've misunderstood e.g in `app/views/admin/bulk_republishing/confirm.html.erb` and `app/views/admin/republishing/index.html.erb`.

How strongly do you feel about this one?

test/unit/app/helpers/admin/republishing_helper_test.rb
```diff
@@ -47,6 +47,33 @@ class Admin::RepublishingHelperTest < ActionView::TestCase
...
+    assert_includes options, {
+      text: "CallForEvidence",
+      value: "call-for-evidence",
+    }
+
+    assert_includes options, {
+      text: "Contact",
+      value: "contact",
+    }
+  end
```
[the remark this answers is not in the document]
Gweaton: I opted against this for a couple of reasons:

1. I didn't want to manually write out every select option manually in the test (although I can if you feel it's important to check every type is included) and the option for looping over a list would have involved doing exactly the same as the implementation which also didn't feel right i.e we'd have to do this to get the expected values:
```
     republishable_content_types.map do |type|
        {
          text: type,
          value: type.underscore.dasherize,
        }
    end
```

2. Stubbing a `republishing_helper` method inside the tests for class itself didn't feel right to me. We _could_ stub its constituent parts like `ApplicationRecord.subclasses[...]` and `Edition.descendants[...]` but that didn't feel great either.

I think our tests are good enough for the feature here as we're testing that one of each option is present in the list.

test/unit/app/helpers/admin/republishing_helper_test.rb
```diff
@@ -42,34 +49,38 @@ class Admin::RepublishingHelperTest < ActionView::TestCase
   end
 end
 
+def omnipresent_content_types
+  content_types[:omnipresent_editionable].concat(content_types[:omnipresent_non_editionable]).sort
```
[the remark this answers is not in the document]
Gweaton: Sorting was used to check equality against the real array values in tests. I've moved the sorting up to the test where it's used though.

app/helpers/admin/republishing_helper.rb
```diff
@@ -75,4 +73,8 @@ def republishing_index_bulk_republishing_rows
       ]
     end
   end
+
+  def non_editionable_content_types
+    ApplicationRecord.subclasses.select { |subclass| subclass.included_modules.include? PublishesToPublishingApi }.map(&:to_s).sort
```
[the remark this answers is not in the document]
Gweaton: You're right, I've removed the sorting here.

test/unit/app/services/bulk_republisher_test.rb
Gweaton: Ah, a useful discovery, nice one!

app/services/bulk_republisher.rb
Gweaton: Yeah, fair enough. Have renamed this to `record`

test/unit/app/helpers/admin/republishing_helper_test.rb
```diff
@@ -47,6 +47,33 @@ class Admin::RepublishingHelperTest < ActionView::TestCase
...
+
+  test "#republishing_index_bulk_republishing_rows creates a link to the 'new' path for the 'all_by_type' section" do
+    all_by_type_link = republishing_index_bulk_republishing_rows.flatten.find { |column|
+                         column[:text].include?("all-by-type")
+                       }[:text]
+
+    expected_link = '<a id="all-by-type" class="govuk-link" href="/government/admin/republishing/bulk/by-type/new">Republish <span class="govuk-visually-hidden">all by type</span></a>'
+
+    assert_equal all_by_type_link, expected_link
+  end
```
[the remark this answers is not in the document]
Gweaton: Interesting! I personally find accessing things by their index a little harder to read than the `flatten`, as it feels a little too closely tied to the implementation and a bit magic-numbery, whereas you can _kind of_ ignore the `flatten`.

I've tweaked this to be more of a happy medium to match on the text (including some HTML) to make it hopefully a bit more readable. I've left the previous test above as-is, as I don't want to refactor it in this PR (which is big enough with the number of changes as-is 😓), but can open a separate one if we feel we can follow a similar approach.

app/services/bulk_republisher.rb
Gweaton: Ah good point! I'd used the original implementation from the Rake task but didn't realise we could get rid of the `all` here.

features/bulk-republishing-content.feature
Gweaton: Do you mind if I leave it as-is, despite being less consistent here, as I slightly prefer the way the original scans?

Gweaton:
@yndajas thanks for the review. I've addressed all your comments (I hope!) - I would have re-requested a review from you but as you're the original author of this, I need to ask in a comment.

You might not be able to "approve" your own PR, but let me know if you're happy for this to go in and I'll approve myself.

test/unit/app/helpers/admin/republishing_helper_test.rb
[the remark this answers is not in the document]
yndajas: >Interesting! I personally find accessing things by their index a little harder to read than the flatten, as it feels a little too closely tied to the implementation and a bit magic-numbery, whereas you can kind of ignore the flatten.

Yeah that's fair. I guess I was trying to use the lines that set the cells to a variable to make it less magic number-y: there's an almost immediate label identifying what the row and cell are

New approach is easier to read though!

test/unit/app/helpers/admin/republishing_helper_test.rb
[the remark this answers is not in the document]
yndajas: 1. I was more meaning stub it to be an array of say two or three types, then just write out the full array of three hashes and `assert_equal` on that
2. I did think about that when writing this comment, but you already do this for the `BulkRepublisher` tests: https://github.com/alphagov/whitehall/pull/9122/files#diff-60fd99a4438aa942f0d060eeb263e1a56843bb86050f4aa55e7d2f907edc225fR207-R208

test/functional/admin/bulk_republishing_controller_test.rb
```diff
@@ -64,4 +64,87 @@ class Admin::BulkRepublishingControllerTest < ActionController::TestCase
...
+  end
+
+  test "Non-GDS Admin users should not be able to GET :new_by_type" do
+    login_as :writer
+
+    get :new_by_type
+    assert_response :forbidden
+  end
+
+  test "GDS Admin users should be redirected to :confirm_by_type on POST ':new_by_type'" do
```
yndajas, line 80: Don't think you need the quotes around `:new_by_type`

app/views/admin/bulk_republishing/confirm_by_type.html.erb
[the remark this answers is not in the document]
yndajas: Yeah we discussed this a while back and made a card to fix the old instances of "schedule": https://trello.com/c/IrapH5gq/1177-replace-instances-of-scheduled-with-queued-in-the-republishing-interface

It's because we're just adding it to a queue to be performed when ready, rather than scheduling it for a specific time (using e.g. `perform_at` or `perform_in`)

yndajas (commented):
Added some extra explanation to a couple of the earlier comments, but there's nothing substantial remaining, so I'm happy for you to approve in any case!

test/unit/app/helpers/admin/republishing_helper_test.rb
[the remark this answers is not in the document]
Gweaton: Ah sorry my mistake, I thought `republishable_content_types` was a method in the `BulkRepublisher` rather than an included helper module's method (too many similarly-named methods!), which we confusingly have to stub on the `BulkRepublisher` itself. Yep, I'll sort that.

app/views/admin/bulk_republishing/confirm_by_type.html.erb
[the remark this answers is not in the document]
Gweaton: Ah yeah, this feature has really been going on a while, that conversation had slipped my mind...

Gweaton (approved):
Approving on Ynda's behalf as they're the original author...

test/unit/app/helpers/admin/republishing_helper_test.rb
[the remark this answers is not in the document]
Gweaton: It's stubbing the methods here:
1. https://github.com/alphagov/whitehall/blob/5d79b12529e362bd98f997756966a52048fdb787/app/helpers/admin/republishing_helper.rb#L92
2. https://github.com/alphagov/whitehall/blob/5d79b12529e362bd98f997756966a52048fdb787/app/helpers/admin/republishing_helper.rb#L57

but we have to stub them on the `BulkRepublisher` class as it's an included module, unless there's another way to stub a module method?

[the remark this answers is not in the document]
yndajas: Ohhh! I suppose because of the way inclusion works, they're inserted into the `BulkRepublisher`'s method list, so it would probably be complicated/not possible to stub them via the `RepublishingHelper` itself?

PR #9133, opened 2024-06-11
Fix translations not appearing on English draft when added

brucebolt:
When a translation is added to an editionable document, we send the draft of the translation to Publishing API. However we don't send the draft of the English (or other translations), meaning the link to the newly added translation doesn't appear.

This is an issue with all translatable editionable document types.

Therefore updating the controller to use `draft_updater` instead of `draft_translation_updater`, which will send the draft of all translations to Publishing API when a translation is updated.

[Trello card](https://trello.com/c/8St6Zw4w)


brucebolt:
It appears the translations get added to the content item in Publishing API, not Whitehall, so this is unlikely to solve the problem.

PR #9135, opened 2024-06-11
Empty ministerial links when reshuffle mode is on

brucebolt:
Publishing API doesn't delete links unless you send it an empty array for a specific key.

Therefore updating the presenter to include all the keys that need to be emptied, rather than an empty hash.


ChrisBAshton (approved):
LGTM - thanks Bruce!

PR #9137, opened 2024-06-11
Preview mode for Ministers page when in "reshuffle mode"

ChrisBAshton:
- Encapsulates "reshuffle mode" into its own 'concern' in Rails
- All callbacks that triggered updating of the 'ministers index' page and the 'how government works' pages are now routed through the concern, and only update the draft stack when reshuffle mode is switched on
- Manual republishing of those pages via the new "Republish" UI feature is also disabled when reshuffle mode is switched on
- A preview link is available from the cabinet ministers reorder page 

Tested on Integration. Updates to the ministers index page take about 1-2 minutes to go through, and attempting to access the page for the first 1-2 minutes after turning on reshuffle mode will result in a "Sorry, we’re experiencing technical difficulties" message. This appears to be because the `patch_links` call is slow for Publishing API to process?

Trello: https://trello.com/c/3y1wYXDP/2674-add-preview-to-ministers-page-when-reshuffle-mode-is-turned-on

### DIFF
--- a/app/controllers/admin/republishing_controller.rb
+++ b/app/controllers/admin/republishing_controller.rb
@@ -1,5 +1,6 @@
 class Admin::RepublishingController < Admin::BaseController
   include Admin::RepublishingHelper
+  include ReshuffleMode
 
   before_action :enforce_permissions!
 
@@ -20,6 +21,11 @@ class Admin::RepublishingController < Admin::BaseController
   def republish_page
     page_to_republish = republishable_pages.find { |page| page[:slug] == params[:page_slug] }
     return render "admin/errors/not_found", status: :not_found unless page_to_republish
+
+    if reshuffle_in_progress? && %w[how-government-works ministers].include?(params[:page_slug])
+      flash[:alert] = "Cannot republish #{params[:page_slug]} page while in reshuffle mode"
+      return redirect_to(admin_republishing_index_path)
+    end
 
     action = "The page '#{page_to_republish[:title]}' has been scheduled for republishing"
 
--- a/app/models/sitewide_setting.rb
+++ b/app/models/sitewide_setting.rb
@@ -17,11 +17,13 @@ class SitewideSetting < ApplicationRecord
   def republish_downstream_if_reshuffle
     return unless key == "minister_reshuffle_mode"
 
-    payload = PublishingApi::MinistersIndexPresenter.new
-
-    Services.publishing_api.put_content(payload.content_id, payload.content)
-    Services.publishing_api.publish(payload.content_id, nil, locale: "en")
-
-    PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksPresenter")
+    update_live = true
+    if on
+      PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksEnableReshufflePresenter", update_live)
+      PresentPageToPublishingApiWorker.perform_async("PublishingApi::MinistersIndexEnableReshufflePresenter", update_live)
+    else
+      PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksPresenter", update_live)
+      PresentPageToPublishingApiWorker.perform_async("PublishingApi::MinistersIndexPresenter", update_live)
+    end
   end
 end
--- a/app/models/sitewide_setting.rb
+++ b/app/models/sitewide_setting.rb
@@ -19,8 +19,12 @@ class SitewideSetting < ApplicationRecord
 
     update_live = true
     if on
-      PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksEnableReshufflePresenter", update_live)
-      PresentPageToPublishingApiWorker.perform_async("PublishingApi::MinistersIndexEnableReshufflePresenter", update_live)
+      # These have to be sent synchronously so we can guarantee the order in which they're processed.
+      # First, send 'page is currently being updated' message to Draft and promote to Live
+      PresentPageToPublishingApiWorker.new.perform("PublishingApi::HowGovernmentWorksEnableReshufflePresenter", update_live)
+      PresentPageToPublishingApiWorker.new.perform("PublishingApi::MinistersIndexEnableReshufflePresenter", update_live)
+      # Finally, send normal ministers index payload to draft so that we can use it as a preview
+      PresentPageToPublishingApiWorker.new.perform("PublishingApi::MinistersIndexPresenter", false)
     else
       PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksPresenter", update_live)
       PresentPageToPublishingApiWorker.perform_async("PublishingApi::MinistersIndexPresenter", update_live)
--- a/app/presenters/publishing_api/how_government_works_enable_reshuffle_presenter.rb
+++ b/app/presenters/publishing_api/how_government_works_enable_reshuffle_presenter.rb
@@ -1,0 +1,15 @@
+module PublishingApi
+  class HowGovernmentWorksEnableReshufflePresenter < HowGovernmentWorksPresenter
+    def links
+      {
+        current_prime_minister: [],
+      }
+    end
+
+    def details
+      {
+        reshuffle_in_progress: true,
+      }
+    end
+  end
+end
--- a/app/views/admin/cabinet_ministers/show.html.erb
+++ b/app/views/admin/cabinet_ministers/show.html.erb
@@ -1,6 +1,28 @@
 <% content_for :page_title, "Cabinet ministers ordering" %>
 <% content_for :title, "Cabinet ministers ordering" %>
 
+<% if reshuffle_in_progress? %>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Preview ministers page in reshuffle mode",
+    heading_level: 2,
+    font_size: "m",
+    margin_bottom: 3,
+  } %>
+
+  <p class="govuk-body govuk-!-margin-bottom-9">
+    <%= link_to("Preview on website (opens in new tab)",
+      "#{Plek.external_url_for('draft-origin')}/government/ministers",
+      class: "govuk-link",
+      target: "_blank",
+      data: {
+        module: "gem-track-click",
+        "track-category": "button-clicked",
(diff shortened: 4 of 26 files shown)


app/models/sitewide_setting.rb
```diff
@@ -17,11 +17,13 @@ def name
...
-    payload = PublishingApi::MinistersIndexPresenter.new
-
-    Services.publishing_api.put_content(payload.content_id, payload.content)
-    Services.publishing_api.publish(payload.content_id, nil, locale: "en")
-
-    PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksPresenter")
+    update_live = true
+    if on
+      PresentPageToPublishingApiWorker.perform_async("PublishingApi::HowGovernmentWorksEnableReshufflePresenter", update_live)
+      PresentPageToPublishingApiWorker.perform_async("PublishingApi::MinistersIndexEnableReshufflePresenter", update_live)
```
ChrisBAshton, line 23: I think we'll probably want to do something like this before merge:

```suggestion
      # send 'page is currently being updated' message to Draft and promote to Live
            PresentPageToPublishingApiWorker.perform("PublishingApi::HowGovernmentWorksEnableReshufflePresenter", update_live)
      PresentPageToPublishingApiWorker.perform("PublishingApi::MinistersIndexEnableReshufflePresenter", update_live)
      # Now send normal ministers index payload to draft so that we can use it as a preview
      PresentPageToPublishingApiWorker.perform("PublishingApi::MinistersIndexPresenter", false)
```

app/views/admin/cabinet_ministers/show.html.erb
```diff
@@ -1,6 +1,29 @@
...
+  <div class="govuk-!-margin-bottom-9">
+    <%= render "govuk_publishing_components/components/details", {
+      title: "Share ministers page preview",
+    } do %>
+      <% capture do %>
+        <%= render "govuk_publishing_components/components/copy_to_clipboard", {
+          label: "Copy and send this link to someone and they’ll be able to preview your draft on GOV.UK.",
+          copyable_content: "#{Plek.external_url_for('draft-origin')}/government/ministers",
+          button_text: "Copy link",
+        } %>
```
ryanb-gds, line 21: I'm not 100% sure, on further reflection, that using the "shareable preview" styles with a non shareable link is a great idea. It might be a bit confusing for users who will expect it to work in the same way as shareable preview links. Would just a normal "Preview on website" link be better?

features/step_definitions/cabinet_ministers_steps.rb
```diff
@@ -1,13 +1,27 @@
...
 
+Then(/^I (should|should not) see a preview to the ministers index page$/) do |should_or_should_not|
+  selector = ".govuk-details__summary-text"
+  text = "Share ministers page preview"
+  if should_or_should_not == "should"
+    expect(page).to have_selector(selector, text:)
+  else
+    expect(page).to_not have_selector(selector, text:)
+  end
+end
```
ryanb-gds, line 17: I feel this might actually be less, simpler, code as two separate methods rather than rolling `should`/`should not` into one

test/unit/app/models/reshuffle_mode_test.rb
```diff
@@ -0,0 +1,72 @@
+require "test_helper"
+
+class ReshuffleModeTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  class ClassThatIncludesReshuffleMode < ApplicationRecord
```
ryanb-gds, line 6: If you don't want to bring the whole of ActiveRecord along for the ride, you can do something like this: https://github.com/alphagov/whitehall/blob/7abf9c0587c6818e06748835cebb1fc0ff5fbe3d/test/unit/app/models/user_orderable_test.rb#L4

Not sure if that works in this situation, it can be more awkward if you do need some ActiveRecord behaviours

test/unit/app/models/reshuffle_mode_test.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      ClassThatIncludesReshuffleMode.new.republish_ministerial_pages_to_publishing_api
+    end
+
+    it "only saves to the draft stack when reshuffle mode is switched on" do
+      PresentPageToPublishingApiWorker.expects(:perform_async).with("PublishingApi::HowGovernmentWorksPresenter", false).once
+      PresentPageToPublishingApiWorker.expects(:perform_async).with("PublishingApi::MinistersIndexPresenter", false).once
+
+      reshuffle_mode = ClassThatIncludesReshuffleMode.new
+      reshuffle_mode.stubs(:reshuffle_in_progress?).returns(true)
```
ryanb-gds, line 36: Bit of a nitpick, but I don't love stubbing a method on the public interface of the class under test. I know it's covered above, so I don't feel strongly about it, it's more of an instinctive "don't mock things you don't have to" reaction even though it's probably fine in this case

app/models/sitewide_setting.rb
ryanb-gds: That seems reasonable, if a bit annoying

ryanb-gds (commented):
I think this is a really nice solution, I like the use of the concerns. One suggestion about links on the "how government works" presenter, and I don't know if we want to consult with Nadeem again on the preview link appearance, but overall looking great 👏

features/step_definitions/cabinet_ministers_steps.rb
ChrisBAshton: Fair. I was trying to avoid having to duplicate the selector (less to update if we change the design) but it was a toss-up either way!

test/unit/app/models/reshuffle_mode_test.rb
ChrisBAshton: This didn't work - but it did result in me removing one layer of inheritance by inheriting from `ActiveRecord::Base` rather than `ApplicationRecord`.

test/unit/app/models/reshuffle_mode_test.rb
ChrisBAshton: Fair point, though I like that we've already testing `reshuffle_in_progress?` above, so the rest of the tests don't have to worry about implementation detail (and can be stubbed in a quick, easy, readable way). It means we can change the implementation of `reshuffle_in_progress?` (e.g. maybe one day it will be decided by both a sitewide setting _and_ something else) and we won't have to update these tests.

app/presenters/publishing_api/how_government_works_enable_reshuffle_presenter.rb
```diff
@@ -0,0 +1,13 @@
+module PublishingApi
+  class HowGovernmentWorksEnableReshufflePresenter < HowGovernmentWorksPresenter
+    def links
+      {}
```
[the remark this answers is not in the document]
ChrisBAshton: Good spot!

test/unit/app/models/reshuffle_mode_test.rb
ChrisBAshton: EDIT: not allowed!

```
test/unit/app/models/reshuffle_mode_test.rb:6:42: C: [Correctable] Rails/ApplicationRecord: Models should subclass ApplicationRecord.
  class ClassThatIncludesReshuffleMode < ActiveRecord::Base
```

app/models/sitewide_setting.rb
ChrisBAshton: Just to double-check - you mean the "having to reorder a minister in order to see the preview" workaround is reasonable, rather than the code suggestion above?

ryanb-gds: Sorry, I meant the code suggestion is reasonable. Just a bit annoying that we have to make an extra request and make them all synchronously, which is something our architecture is trying to help us not to have to do

app/controllers/admin/republishing_controller.rb
```diff
@@ -1,5 +1,6 @@
 class Admin::RepublishingController < Admin::BaseController
   include Admin::RepublishingHelper
+  include ReshuffleMode
```
ryanb-gds, line 3: Using this model concern in a controller feels a tad weird. On the other hand it's not really a model concern arguably, we just don't have anywhere more logical to put it 🤔

ryanb-gds: Just realised I missed this first time around, sorry for commenting at second time of asking!

ryanb-gds (approved):
I can live with it, but including the `ReshuffleMode` in both controllers and models definitely doesn't feel great from a layering perspective, if you do have any alternative ideas. I can't suggest a clear improvement though! Thanks for getting this sorted out

app/controllers/admin/republishing_controller.rb
ChrisBAshton: Yeah it felt a bit odd. I think we could probably find a way of removing it from the RepublishingController, as it's only used to make `reshuffle_in_progress?` available (could probably extract that to a helper class we can `require_relative` or something).

But the other controllers make use of the `republish_ministers_index_page_to_publishing_api` callbacks which I don't think we can move to a helper class whilst still allowing neatly referencing it in an `after_save`. And rather than spread the code across both a concern and a helper, I like that we've encapsulated this mess in one place 😅 So think this is good to go for now. Thanks!

lib/present_page_to_publishing_api.rb
```diff
@@ -5,4 +5,10 @@ def publish(presenter_class)
     Services.publishing_api.patch_links(payload.content_id, links: payload.links)
     Services.publishing_api.publish(payload.content_id, nil, locale: payload.content[:locale])
   end
+
+  def save_draft(presenter_class)
+    payload = presenter_class.new
+    Services.publishing_api.put_content(payload.content_id, payload.content)
+    Services.publishing_api.patch_links(payload.content_id, links: payload.links)
```
brucebolt, line 12: Patching the links makes these appear in the live content item as well as the draft.  This could mean a list of the draft ministers can be obtained by accessing the page's content item, even if reshuffle mode is activated.

The only way around this would be to switch the page to use edition links (but then they aren't recursive, so wouldn't work here).

ChrisBAshton: Correct. We're also aware folks could go and look at individual people's pages and see role appointments etc being applied, even while in reshuffle mode. I'll make sure it's raised tomorrow so everyone is aware, but I think it's a known and accepted thing.

PR #9143, opened 2024-06-12
Exclude superseded editions from document collection document search results

ryanb-gds:
This was causing a confusing experience for users where they would be asked to select from numerous documents with identical titles.

The active scope on the Edition model excludes superseded editions.

Trello: https://trello.com/c/USevUJuy


ChrisBAshton (approved):
⭐ nice one!

PR #9147, opened 2024-06-13
Add GA4 tracking to all buttons in Whitehall

dnkrj:
## What
- Introduce a new module to add ga4 data to all buttons on the current page
- Add it to the design system template to be included on all pages
- The module merges any existing attributes defined on the button with the generic set
- Remove the redundant attributes from the visual editor exit button 
 
## Why 
- Clicks will then be tracked by the `ga4-event-tracker` which is already initialised


app/assets/javascripts/admin/modules/ga4-button-setup.js
```diff
@@ -0,0 +1,27 @@
...
+        event_name: button.type === 'submit' ? 'form_response' : 'navigation',
+        type: 'generic_link',
+        text: button.textContent,
+        section: document.title.split(' - ')[0],
+        action: button.textContent,
+        tool_name: document.title.split(' - ')[0]
+      }
+      if (button.dataset.ga4Event) {
+        Object.assign(event, JSON.parse(button.dataset.ga4Event))
+      }
```
ChrisBAshton, line 21: My JS is rusty, but something like this is more readable IMO:

```suggestion
      const event = (button.dataset.ga4Event) ? JSON.parse(button.dataset.ga4Event) : {
        event_name: button.type === 'submit' ? 'form_response' : 'navigation',
        type: 'generic_link',
        text: button.textContent,
        section: document.title.split(' - ')[0],
        action: button.textContent,
        tool_name: document.title.split(' - ')[0]
      }
```

app/assets/javascripts/admin/modules/ga4-button-setup.js
```diff
@@ -0,0 +1,27 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+;(function (Modules) {
+  function Ga4ButtonSetup(module) {
+    this.module = module
+  }
+
+  Ga4ButtonSetup.prototype.init = function () {
+    const buttons = this.module.querySelectorAll('button')
```
ChrisBAshton, line 9: Do we have to worry about [links with a role of 'button'](https://design-system.service.gov.uk/components/button/#start-buttons) too?

ChrisBAshton (approved):
LGTM 👍 a couple of nitpicks below.

app/assets/javascripts/admin/modules/ga4-button-setup.js
dnkrj: The `Object.assign` here is merging the two objects so that you're able to override only one part of the `ga4-event` without having to reconstruct the whole thing. The ternary operator wouldn't allow for this.

I've added an extra test to illustrate and make sure we don't lose the merging in future.

app/assets/javascripts/admin/modules/ga4-button-setup.js
dnkrj: Good point, it's not something that we see on this page, but I've adjusted the query selector and added a test case for it

app/assets/javascripts/admin/modules/ga4-button-setup.js
ChrisBAshton: Ahh makes sense. Thanks for explaining

PR #9148, opened 2024-06-13
Add GA4 to form validation errors

lauraghiorghisor-tw:
[Trello card](https://trello.com/c/nds7d8mH/2689-add-ga4-tracking-for-edition-form-validation-errors)


app/components/admin/error_summary_component.html.erb
```diff
@@ -1,13 +1,4 @@
-<% if flash["alert"].present? %>
```
lauraghiorghisor-tw, line 1: @minhngocd Not super sure what the logic here was meant to achieve.

test/functional/admin/statistics_announcement_publications_controller_test.rb
```diff
@@ -69,7 +69,7 @@ class Admin::StatisticsAnnouncementPublicationsControllerTest < ActionController
 
     assert_response :success
     has_search_results_table
-    assert_select "ul.govuk-error-summary__list a[data-track-action='statistics announcement-error'][data-track-label=\"Publication type does not match: must be statistics\"]", text: "Publication type does not match: must be statistics"
```
lauraghiorghisor-tw, line 72: @minhngocd not sure why this was ever needed

app/components/admin/error_summary_component.html.erb
minhngocd: this is to track flash alerts back in the days i guess - if you check the commit message it says

> Ensure flash-danger Google Analytics event is retained
> 
> We've moved away from using flash[:alert] in favour of the error summary component when there are errors on an object.
> 
> However, we want to retain the GA event that is triggered by the flash. To get around this I've ensured a flash["alert"] is only rendered on the page when the error summary block is empty.
> 
> This will ensure we only ever get 1 banner on the page. To ensure we still get the GA event i've wrapped the error summary component in a div which tracks the flash["alert"].

test/functional/admin/statistics_announcement_publications_controller_test.rb
minhngocd: probably to test that we are sending that tracking data by setting those attributes

app/components/admin/error_summary_component.html.erb
lauraghiorghisor-tw: I dug it up too. I just find it a bit confusing/convoluted. We're still firing 
`data_attributes: track_analytics_data("form-error", analytics_action, error.full_message)` from within the `error_summary_component.rb` so it feels a bit duplicated. 

I can leave it in and we can make this call when we implement whatever schema we want for flash, as part of the migration to GA4.

test/functional/admin/statistics_announcement_publications_controller_test.rb
lauraghiorghisor-tw: It is tested elsewhere, the test is not about tracking, the test does not require tracking selector to pass.

app/components/admin/error_summary_component.html.erb
minhngocd: update on conversations - we've discussed with Carmen, and believe if the user doesn't see the flash as a separate thing, then we should not need to see it as a separate event. we can remove this logic to reduce the duplication

app/components/admin/error_summary_component.rb
```diff
@@ -51,4 +55,10 @@ def errors
                   object.errors
                 end
   end
+
+  def ga4_title
+    return object.class.name.humanize if [ActiveModel::Errors, Array].include?(object.class)
+
+    "#{object.try(:new_record?) ? 'New' : 'Editing'} #{object.model_name.human.downcase.titleize}"
```
minhngocd, line 62: we do this so many times you it warrants a helper method 👀  but also happy to refactor as a separate piece of work

minhngocd (approved):

PR #9149, opened 2024-06-13
Fix error message matching in `assert_raises`

yndajas:
Some tests using `assert_raises` were including a string, presumably to check that the error message looked correct. However, the `assert_raises` method requires that this is a keyword argument using `match`, which the tests were not using. This updates the tests to use this, in some places fixing the expected error messages and in one place making a minor change to the code to match the expected behaviour (see below)

## Note on the `clone_published_html_attachment_to_draft_edition`

The default scope of an `Edition` excludes those with a state of `deleted` (from `Edition::Workflow`), which I believe means that when `.attachable` is called on an attachment with a deleted edition, it will return `nil`

I couldn't find exactly how the `attachable` method works at a glance, but presumably it searches for an instance of the `attachable_type` with the `attachable_id`, which is where the default scope excludes the deleted editions

Using the safe navigation operator, the intended exception-raising behaviour of the rake task as documented in the tests is retained

---

### DIFF
--- a/lib/tasks/clone_published_html_attachment_to_draft_edition.rake
+++ b/lib/tasks/clone_published_html_attachment_to_draft_edition.rake
@@ -3,7 +3,7 @@ task :clone_published_html_attachment_to_draft_edition, %i[html_attachment_id] =
   html_attachment = HtmlAttachment.find(args[:html_attachment_id])
 
   edition = html_attachment.attachable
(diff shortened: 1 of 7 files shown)


Gweaton (approved):
Well done for catching this.

Re: `attachable`, it looks like it's done by the `Edition` subclasses including the `Attachable` module in combination with polymorphic association declared in the `Attachment` base class that `HtmlAttachment` inherits from. I think your explanation of how it's looked up is probably accurate enough.

PR #9153, opened 2024-06-14
Remove hardcoded logic preventing "foreign language only" visibility

ChrisBAshton:
These extra conditionals appear to have been added without context in December 2022:
https://github.com/alphagov/whitehall/commit/1cddcbb7e54855e9daed9558202d081785973827

Consultations and World News Stories continued to work because they were included in the hardcoded logic, but other formats that should support "foreign language only" mode would have the field hidden and unusable. These formats are:

- [Call for Evidence](https://github.com/alphagov/whitehall/blob/62f85ecbbb1aa9bb4a86416f4ec68a28d9ef6d14/app/models/call_for_evidence.rb#L147-L149)
- [Document Collection](https://github.com/alphagov/whitehall/blob/3915f372e344b274e8d198b6fe64c0a3d4fffa9c/app/models/document_collection.rb#L51-L53)

We should be relying solely on the `locale_can_be_changed?` method to drive this view logic.

Solves Zendesk ticket https://govuk.zendesk.com/agent/tickets/5509163


GDSNewt (approved):
This seems like a logical change. 👍

PR #9154, opened 2024-06-17
Componentisation proof of concept

pezholio:
This is a proof of concept that demostrates how we might use [Packwerk](https://github.com/Shopify/packwerk) to organise Whitehall into domain components by following Domain-Driven Design principles.

As a proof of concept, I've taken a small feature in Whitehall (the Emergency Banner) and arranged it into a small package. This package is a Rails engine which serves as a "mini-application" including all the things required for the application (such as controllers, views etc), as well as any tests, and the routes. 

We use Packwerk to define the package, as well as highlight any parts of the package that use code from the main application. 

As well as the Emergency Banner package, I've also added a `Base` package, which includes things like the Base controllers, helpers, and anything else that may be required across the application. We can then define it as an explicit dependency on both the package and the main application, ensuring that any boundaries are well-defined.

Going forward the ambition would be to move more functionality into packages, making code more well defined and easier to reason about. 

This is very much an idea in its draft stage, but if this is something that people are keen to try out / explore adopting, we (the Content Modelling team) can draft a more formal ADR and begin trying this pattern when we build our Object Store within Whitehall.


ryanb-gds:
I'm quite keen on this in principle, but... I'm doubtful that we can invest the time to see this through for the whole of Whitehall, at least if it depends on the Whitehall Experience team to do the work. From a product perspective I think it's going to be very hard for us to justify as I expect it could take quite a long time. I suppose we could just put everything we can't easily separate into one large package, but then would we be gaining anything?

This does offer a nice way to keep the new content block/object store behaviours separate from the rest of Whitehall, which is definitely a good thing. But if we aren't able to do the hard work of decoupling the edition behaviours from each other and the model so that we can leverage this approach for other parts of the app, then does it just become "yet another way of doing things in Whitehall"? Would be very interested to get @richardTowers thoughts - maybe there's an appetite to devote some resource to such an idea?

richardTowers:
I also like the idea. Reminds me a bit of my .NET days ( :trollface: ) when we'd have a bunch of .csproj files separating out assemblies, and [`internal`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/internal) methods and things like that ❤️ 

I agree with Ryan's concerns that we might not be able to commit to this enough for it to get critical mass though.

The value of packwerk feels relatively modest, which would need to balance the modest cost of introducing a new concept. I think we'd get a reasonable benefit if we properly carved things up and used it everywhere, but that's a big bit of work. On the other hand, if we just did a couple of small examples (like this), the cost of the new concept would probably outweigh the benefit. So I think it's a "we should only do this if we're going to really commit to it" thing.

Maybe it would be good to sketch out what sort of module structure we might aim for in the long term? Would we just be using it for things like the emergency banner, which are bolt ons? Or would we have it for more fundamental things like different document types?

pezholio:
I certainly think, in an ideal world, the first thing to be componentised would be the "bolt on" functionality, then potentially the main publishing functionality could be compontentised after. The level of seperation could be as broad or as narrow as we'd like, we could go down to document type level if that makes sense, or just have one "mega component" for all the existing Editionable content. I did have a go at adding a few more components, but didn't push these for the sake of brevity. I'll push them up to show some other use cases too.

pezholio:
Assuming we only included the "bolt on" functionality, I think I'd see the package structure looking a bit like this:

```
packages/
  cabinet_ministers/
  emergency_banner/
  operational_fields/
  get_involved/
  governments/
  groups/
  organisations/ # This could be split up further into "Organisations and Contacts"
  people/
  roles/
  sitewide_settings/
  topical_events/
  world_location_news/
  worldwide_organisations/
  republishing/
```

pezholio:
I think I'd like to see an approach of "if you touch it, you move it", but the challenges there are getting buy-in and also enforcing that approach too

PR #9160, opened 2024-06-17
No need to set SECRET_KEY_BASE_DUMMY here.

sengi:
govuk-ruby-builder image [already sets](https://github.com/alphagov/govuk-ruby-images/blob/6cd1b32/builder.Dockerfile#L16) this.


ryanb-gds (approved):
Thanks @sengi 🙌

PR #9162, opened 2024-06-17
Avoid Oracle-specific collation in db:create.

sengi:
`utf8mb4_0900_ai_ci` is specific to Oracle MySQL and doesn't work with MariaDB. `utf8mb4_unicode_520_ci` is, for all practical purposes, identical\* and works with both MySQL and MariaDB.

This makes it easier to run Whitehall locally and helps avoid us getting locked into Oracle's flavour of MySQL.

Doesn't affect existing tables, only `db:create`, `db:reset` etc. i.e. testing/development.

\* citation needed 🙃

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -868,7 +868,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_06_05_104148) do
     t.index ["user_id"], name: "index_republishing_events_on_user_id"
   end
 
-  create_table "review_reminders", charset: "utf8mb4", collation: "utf8mb4_0900_ai_ci", force: :cascade do |t|
+  create_table "review_reminders", charset: "utf8mb4", collation: "utf8mb4_unicode_520_ci", force: :cascade do |t|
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
This is interesting. Thanks @sengi

sengi:
Been thinking on this in the background and now I'm much less certain that it's worth it 🤔

Mainly because I forgot that `db/schema.rb` is supposed to be generatable from the actual database, i.e. Rails kinda considers the database itself to be the source of truth for defining the schema.

I'm also not 100% sure there isn't whacky data in production Whitehall that might get sorted/compared differently between these two very similar collations.

So I'm now leaning towards "ehh the whole Unicode situation with MySQL/MariaDB is messed up but this probably isn't a clean enough workaround on balance". Sorry for the noise!

sengi:
Argument for I guess would be: "Well it doesn't break any tests, so it's still WAI." I wouldn't buy that though haha.

PR #9163, opened 2024-06-17
Remove Ghostscript and PDF thumb generation.

### LINKED ISSUE #7480: Remove obsolete PDF thumbnail generation feature and Ghostscript dependency
Whitehall currently uses Ghostscript to generate thumbnail icons for PDF attachments.

~~It's not clear whether this is even a useful feature (since most document cover pages look very similar and the thumbnail is too small for text to be readable), but if we want to keep generating these thumbnails then it's worth considering switching from Ghostscript to pdf.js to reduce the attack surface.~~

Ghostscript is 70+ MB of old code, mostly written in C ~~, whereas pdf.js is way smaller and designed for security from the beginning.~~

See Keith's comment below. We just need to delete the thumbnail code and remove Ghostscript.

sengi:
gov-frontend no longer uses PDF thumbnails, so we can drop Ghostscript etc.

For now, we keep the (probably disused) thumbnail_url field and use the "fallback" icon for new PDFs. Can probably remove the field and the icon altogether soon after.

Fixes #7480.

See also:
https://trello.com/c/vjSowrs7 (dup)
https://trello.com/c/oh4MXlUA (related)

### DIFF
--- a/.github/workflows/cucumber.yml
+++ b/.github/workflows/cucumber.yml
@@ -17,12 +17,10 @@ jobs:
       - name: Setup Redis
         uses: alphagov/govuk-infrastructure/.github/actions/setup-redis@main
 
-      - name: Install additional system dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends ghostscript
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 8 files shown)


ryanb-gds (approved):
Excellent, always good to see this sort of thing depart the codebase

PR #9165, opened 2024-06-17
Use Marcel gem for filetype sniffing.

### LINKED ISSUE #9161: Switch from deprecated `MiniMagick::Image#mime_type` to Marcel/MimeMagic gems.
`rake test` output spams this, which appears to be a valid point:

`[MiniMagick] MiniMagick::Image#mime_type has been deprecated, because it wasn't returning correct result for all formats ImageMagick supports. Unfortunately, returning the correct MIME type would be very slow, because it would require ImageMagick to read the whole file. It's better to use Marcel and MimeMagic gems, which are able to determine the MIME type just from the image header.`

The `MiniMagick::Image#mime_type` method [was removed](https://github.com/minimagick/minimagick/commit/6049889) altogether a few days ago.

sengi:
The deprecated `MiniMagick::Image#mime_type` method (recently [removed](https://github.com/minimagick/minimagick/commit/6049889)) was causing a ton of logspew in tests.

[Marcel](https://github.com/rails/marcel#usage) was already an indirect dependency via Rails.

Fixes #9161.


ryanb-gds (approved):
LGTM

PR #9166, opened 2024-06-17
Maintain lead organisations first when publishing HTML attachments

KludgeKML:
Currently HTML Attachments are publishing with organisations in default order (which means that there's no way for frontend apps to order by putting lead organisations first in the list). This is in contrast to the attachables, which seem to retain some ordering where lead organisations are listed first, example the two links in this ticket:

https://govuk.zendesk.com/agent/tickets/5448531

https://www.gov.uk/government/consultations/improving-the-experiences-of-people-with-mecfs-interim-delivery-plan
https://www.gov.uk/government/consultations/improving-the-experiences-of-people-with-mecfs-interim-delivery-plan/consultation-document-the-interim-delivery-plan-on-mecfs

### DIFF
--- a/app/models/consultation_response.rb
+++ b/app/models/consultation_response.rb
@@ -20,6 +20,10 @@ class ConsultationResponse < ApplicationRecord
 
   delegate :organisations, to: :parent_attachable
 
+  delegate :lead_organisations, to: :parent_attachable
+
+  delegate :supporting_organisations, to: :parent_attachable
+
   delegate :alternative_format_contact_email, to: :consultation
 
   delegate :publicly_visible?, to: :parent_attachable
--- a/app/presenters/publishing_api/html_attachment_presenter.rb
+++ b/app/presenters/publishing_api/html_attachment_presenter.rb
@@ -38,7 +38,7 @@ module PublishingApi
     def edition_links
       {
         parent: parent_content_ids, # please use the breadcrumb component when migrating document_type to government-frontend
-        organisations: parent.organisations.pluck(:content_id).uniq,
+        organisations: (lead_org_ids + supporting_org_ids).uniq,
         primary_publishing_organisation:,
         government: government_id,
       }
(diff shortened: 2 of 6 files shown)


app/presenters/publishing_api/html_attachment_presenter.rb
```diff
@@ -38,7 +38,7 @@ def links
     def edition_links
       {
         parent: parent_content_ids, # please use the breadcrumb component when migrating document_type to government-frontend
-        organisations: parent.organisations.pluck(:content_id).uniq,
+        organisations: (parent.lead_organisations.pluck(:content_id) + parent.supporting_organisations.pluck(:content_id)).uniq,
```
ChrisBAshton, line 41: Looks like there's already an [existing method](https://github.com/alphagov/whitehall/blob/3f7a05fd6959784db3463febe977ccc26ee05cec/app/presenters/publishing_api/html_attachment_presenter.rb#L95) for retrieving the `lead_org_id` (and it uses `.try` to fail gracefully). Can you reuse that method, and add an equivalent for `supporting_org_ids` that uses `.try` in the same way?

ChrisBAshton: Ah, just spotted the existing method only returns one lead org ID, you probably want them all. But `lead_org_ids` seems a fair method to add.

app/models/consultation_response.rb
```diff
@@ -20,6 +20,10 @@ def access_limited_object
 
   delegate :organisations, to: :parent_attachable
 
+  delegate :lead_organisations, to: :parent_attachable
+
+  delegate :supporting_organisations, to: :parent_attachable
```
ChrisBAshton, line 25: Is there a test that can be added to ensure CallForEvidenceResponse and ConsultationResponse delegate these methods? As removing these lines would presumably cause some issues down the line!

KludgeKML: Good idea. Have added tests for this.

PR #9170, opened 2024-06-18
Add UI for republishing documents by content IDs

yndajas:
[Trello](https://trello.com/c/3Pz8w9D0/1197-add-a-user-interface-for-whitehalls-bulk-republishing-documents-by-content-ids-rake-task)

This adds a UI for republishing documents by content IDs

## Screenshots

image

image

image

image

---

### DIFF
--- a/app/controllers/admin/bulk_republishing_controller.rb
+++ b/app/controllers/admin/bulk_republishing_controller.rb
@@ -79,6 +79,53 @@ class Admin::BulkRepublishingController < Admin::BaseController
     end
   end
 
+  def new_documents_by_content_ids; end
+
+  def search_documents_by_content_ids
+    content_ids = content_ids_string_to_array(params[:content_ids])
+    return if redirect_if_no_requested_content_ids(content_ids)
+
+    matching_document_content_ids = Document.where(content_id: content_ids).pluck(:content_id)
+    return if redirect_if_unfound_content_ids(matching_document_content_ids, content_ids)
+
+    redirect_to(admin_bulk_republishing_documents_by_content_ids_confirm_path(params[:content_ids]))
+  end
+
+  def confirm_documents_by_content_ids
+    @content_ids_string = params[:content_ids]
+    content_ids = content_ids_string_to_array(@content_ids_string)
+    return if redirect_if_no_requested_content_ids(content_ids)
+
+    @matching_documents = Document.where(content_id: content_ids)
+    return if redirect_if_unfound_content_ids(@matching_documents.pluck(:content_id), content_ids)
+
+    @republishing_event = RepublishingEvent.new
+  end
+
+  def republish_documents_by_content_ids
+    @content_ids_string = params[:content_ids]
+    content_ids = content_ids_string_to_array(@content_ids_string)
+    return if redirect_if_no_requested_content_ids(content_ids)
+
+    @matching_documents = Document.where(content_id: content_ids)
+    return if redirect_if_unfound_content_ids(@matching_documents.pluck(:content_id), content_ids)
+
+    bulk_content_type_key = :all_documents_by_content_ids
+    bulk_content_type_metadata_for_content_ids = bulk_content_type_metadata.fetch(bulk_content_type_key)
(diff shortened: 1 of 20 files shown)


app/helpers/admin/editions_helper.rb
```diff
@@ -212,4 +212,8 @@ def search_results_table_actions(edition)
 
     sanitize(actions)
   end
+
+  def edition_title_or_edition_title_link(edition)
```
Gweaton, line 216: It might be a bit clearer if the logic is reversed here, either in the name of the method or the implementation i.e. `edition_title` is sort of the second return value, but the first part of the method name if that makes sense?

app/controllers/admin/bulk_republishing_controller.rb
```diff
@@ -79,6 +79,81 @@ def republish_documents_by_organisation
...
+
+    if content_ids.empty?
+      flash[:alert] = "No content IDs provided"
+      return redirect_to(admin_bulk_republishing_documents_by_content_ids_new_path)
+    end
+
+    # we could pluck just the IDs and content IDs here for speed, but this is more readable
+    @matching_documents = Document.where(content_id: content_ids)
+
+    unless @matching_documents.count == content_ids.count
```
Gweaton, line 135: Could this be extracted to a private method as we're using it a few times here?

test/functional/admin/bulk_republishing_controller_test.rb
```diff
@@ -205,7 +205,7 @@ class Admin::BulkRepublishingControllerTest < ActionController::TestCase
     assert_response :forbidden
   end
 
-  test "GDS Admin users should be able to POST :republish_documents_by_organisation with an existing organisation slug, creating a RepublishingEvent for the current user" do
+  test "GDS Admin users should be able to POST :republish_documents_by_organisation with an existing organisation slug and a reason, creating a RepublishingEvent for the current user" do
```
Gweaton, line 208: For next time, I think putting things like this (and some other refactorings) into a separate PR would help reduce some of the reviewing burden, as there's a lot of change in this PR!

Gweaton (approved):
I think this looks good, and a lot of hard work, so well done on getting it in - particularly the preview of the thing being republished.

There's quite a lot going on here so I found it a little tricky to go through with a particularly fine-toothed comb but the implementation broadly looks good to me.

With big features like this that introduce a lot of new lines, I'd find it easier to spot things and understand what's required for the feature vs. a refactor if the refactoring commits were moved onto their own branch. I'm probably guilty of doing this myself in recent PRs, but this was a particularly chunky branch!

db/migrate/20240618164903_add_content_ids_to_republishing_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddContentIdsToRepublishingEvents < ActiveRecord::Migration[7.1]
+  def change
+    add_column :republishing_events, :content_ids, :json
```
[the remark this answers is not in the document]
yndajas: Should probably mention that in the commit

app/helpers/admin/editions_helper.rb
yndajas: Yeah, I did have this thought but I was falling over all the words! `edition_title_link_or_edition_title` is probably okay though?

test/functional/admin/bulk_republishing_controller_test.rb
yndajas: Yup - I normally pull them out at the end, but rushed this one a bit!

app/controllers/admin/bulk_republishing_controller.rb
yndajas: Something like [this](https://github.com/alphagov/whitehall/pull/9170/commits/8ae0b5db4b735d46eb372dcaca9fbd55a5e4d7c7)?

Gweaton: Yeah - is there a way we could remove the `return if` and just call `redirect_if` or do we need to do it this way to allow us to exit out of the controller action early?

yndajas: Pretty sure we'd have a multiple render error without it, since the code would reach multiple `redirect_to` or `render` statements

PR #9172, opened 2024-06-19
Clear organisations links

GDSNewt:
**What**

Clear organisations links from people pages.

**Why**

Change brought about by an issue encountered on [2ndline](https://govuk.zendesk.com/agent/tickets/5848753). 

Currently there is an undesirable scenario in which a person can be associated with a department, not in the sense of a role, but as a piece of organisation content. This fix clears those links, which should prevent updates to people pages being associated with the organisation's latest news, which is what happened in this instance. 

**Before**
Screenshot 2024-06-19 at 09 15 36

**After**

Screenshot 2024-06-19 at 09 14 57

### DIFF
--- a/app/presenters/publishing_api/person_presenter.rb
+++ b/app/presenters/publishing_api/person_presenter.rb
@@ -28,7 +28,9 @@ module PublishingApi
     end
 
     def links
-      {}
+      {
+        organisations: [],
+      }
     end
 
     def details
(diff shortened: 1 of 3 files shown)


ryanb-gds (approved):
I feel slightly uncomfortable with this, and I'm starting to wonder whether the more appropriate course of action might have been to see if the web team feel that they might want to filter the results they get from Search API for the latest content section more restrictively. There no real reason why people shouldn't have a link to their organisation, we don't have to render them.

On the other hand this solves a problem and is highly unlikely to cause any new ones, so I'm going to approve it.

PR #9173, opened 2024-06-19
Refactor edition model

pezholio:
This does a bit of tidying up of the Edition model, as well as moves the Search behaviour to a Concern. This will allow the Edition model to be inherited without having to override the search functionality if we don't want to include an Edition's content in search results. 

I've been experimenting with a couple of things, including making a "Lite Edition" as documented in https://github.com/alphagov/whitehall/blob/edition-refactoring/docs/adr/0003-introduce-reusable-content-items.md, but I kept coming across issues with existing relations, which were tricky to get around and would result in messy workarounds. This seemed like the cleanest solution available.

Thoughts welcome!

### DIFF
--- a/app/models/concerns/edition/base_permission_methods.rb
+++ b/app/models/concerns/edition/base_permission_methods.rb
@@ -1,0 +1,29 @@
+# Sets all default overwritable permission methods for an Edition
+module Edition::BasePermissionMethods
+  extend ActiveSupport::Concern
+
+  %w[
+    can_be_associated_with_social_media_accounts?
+    can_be_associated_with_topical_events?
+    can_be_associated_with_roles?
+    can_be_associated_with_role_appointments?
+    can_be_associated_with_worldwide_organisations?
+    can_be_fact_checked?
+    can_be_related_to_mainstream_content?
+    can_be_related_to_organisations?
+    can_apply_to_subset_of_nations?
+    allows_attachments?
+    allows_attachment_references?
+    allows_inline_attachments?
+    can_be_grouped_in_collections?
+    has_operational_field?
+    can_apply_to_local_government?
+    national_statistic?
+    has_consultation_participation?
+    is_associated_with_a_minister?
+    statistics?
+    can_be_tagged_to_worldwide_taxonomy?
+  ].each do |method|
+    define_method(method) { false }
+  end
+end
--- a/app/models/concerns/edition/scopes.rb
+++ b/app/models/concerns/edition/scopes.rb
@@ -1,0 +1,158 @@
+module Edition::Scopes
+  extend ActiveSupport::Concern
+
+  included do
+    scope :with_title_or_summary_containing,
+          lambda { |*keywords|
+            pattern = "(#{keywords.map { |k| Regexp.escape(k) }.join('|')})"
+            in_default_locale.where("edition_translations.title REGEXP :pattern OR edition_translations.summary REGEXP :pattern", pattern:)
+          }
+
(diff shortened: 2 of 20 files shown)


app/models/concerns/edition/scopes.rb
```diff
@@ -0,0 +1,158 @@
...
+            scope = in_default_locale.includes(:document)
+            scope
+              .where("edition_translations.title LIKE :like_clause", like_clause:)
+              .or(scope.where(document: { slug: keywords }))
+          }
+
+    scope :in_pre_publication_state, -> { where(state: Edition::PRE_PUBLICATION_STATES) }
+    scope :force_published, -> { where(state: "published", force_published: true) }
+    scope :not_published, -> { where(state: %w[draft submitted rejected]) }
+    scope :without_not_published, -> { where.not(state: %w[draft submitted rejected]) }
```
ryanb-gds, line 25: In general I'd be keen to see us refactor into domain based concerns rather than technical concerns. So for these scopes, I'd suggest factoring them into the edition workflow concern. I realise this may not be easy everywhere but where the opportunity arises I feel we should take it.

app/models/concerns/edition/scopes.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    scope :announcements, -> { where(type: Announcement.concrete_descendants.collect(&:name)) }
+    scope :consultations, -> { where(type: "Consultation") }
+    scope :call_for_evidence, -> { where(type: "CallForEvidence") }
+    scope :detailed_guides, -> { where(type: "DetailedGuide") }
+    scope :statistical_publications, -> { where("publication_type_id IN (?)", PublicationType.statistical.map(&:id)) }
+    scope :non_statistical_publications, -> { where("publication_type_id NOT IN (?)", PublicationType.statistical.map(&:id)) }
+    scope :corporate_publications, -> { where(publication_type_id: PublicationType::CorporateReport.id) }
```
ryanb-gds, line 33: Could these go on to the Publication model itself, or will that break things?

app/models/concerns/edition/scopes.rb
```diff
@@ -0,0 +1,158 @@
...
+    scope :corporate_information_pages, -> { where(type: "CorporateInformationPage") }
+    scope :publicly_visible, -> { where(state: Edition::PUBLICLY_VISIBLE_STATES) }
```
ryanb-gds, line 35: Another one for the workflow concern, potentially

app/models/concerns/edition/base_permission_methods.rb
```diff
@@ -0,0 +1,29 @@
...
+    can_be_associated_with_roles?
+    can_be_associated_with_role_appointments?
+    can_be_associated_with_worldwide_organisations?
+    can_be_fact_checked?
+    can_be_related_to_mainstream_content?
+    can_be_related_to_organisations?
+    can_apply_to_subset_of_nations?
+    allows_attachments?
+    allows_attachment_references?
+    allows_inline_attachments?
```
ryanb-gds, line 17: I wonder if these could go on the attachable concern?

ryanb-gds (approved):
This is very exciting, I'm really happy to see that there are potentially some legs in factoring out edition behaviour into concerns.

I have a suggestion about using domain based concerns but that can come in a later PR, or I can take a look at doing that later if you'd prefer to move on.

pezholio:
Cool thanks. I'm happy to run with those suggestions in a separate PR. Will see what else falls out too!

PR #9176, opened 2024-06-19
Remove empty ministerial roles from drag and drop

lauraghiorghisor-tw:
In preparation for the election, we are improving the ordering of cabinet ministers. The empty, unappointed roles, were showing in the ordering page (and during drag and drop), which made the process quite cumbersome.

We filtered out the empty roles to streamline the reordering task. The roles still show in the roles page.
Other tabs on the ordering page are not affected (e.g. 'Also attends cabinet ordering').

[Trello card](https://trello.com/c/Yl4ZZBc2/2690-remove-empty-roles-from-drag-and-drop-ministers-list)

### DIFF
--- a/app/controllers/admin/cabinet_ministers_controller.rb
+++ b/app/controllers/admin/cabinet_ministers_controller.rb
@@ -5,14 +5,14 @@ class Admin::CabinetMinistersController < Admin::BaseController
   helper_method :reshuffle_in_progress?
 
   def show
-    @cabinet_minister_roles = MinisterialRole.includes(:translations).where(cabinet_member: true).order(:seniority)
+    @cabinet_minister_roles = MinisterialRole.ministerial_roles_with_current_appointments
     @also_attends_cabinet_roles = MinisterialRole.includes(:translations).also_attends_cabinet.order(:seniority)
     @whip_roles = MinisterialRole.includes(:translations).whip.order(:whip_ordering)
     @organisations = Organisation.ministerial_departments.excluding_govuk_status_closed.order(:ministerial_ordering)
   end
 
   def reorder_cabinet_minister_roles
-    @roles = MinisterialRole.includes(:translations).where(cabinet_member: true).order(:seniority)
+    @roles = MinisterialRole.ministerial_roles_with_current_appointments
   end
 
   def order_cabinet_minister_roles
--- a/app/controllers/admin/cabinet_ministers_controller.rb
+++ b/app/controllers/admin/cabinet_ministers_controller.rb
@@ -6,8 +6,8 @@ class Admin::CabinetMinistersController < Admin::BaseController
 
   def show
     @cabinet_minister_roles = MinisterialRole.ministerial_roles_with_current_appointments
-    @also_attends_cabinet_roles = MinisterialRole.includes(:translations).also_attends_cabinet.order(:seniority)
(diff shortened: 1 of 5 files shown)


app/controllers/admin/cabinet_ministers_controller.rb
```diff
@@ -68,4 +68,8 @@ def order_ministerial_roles_params
   def order_ministerial_organisations_params
     params.require(:ministerial_organisations)["ordering"]
   end
+
+  def ministerial_roles_with_current_appointments
+    MinisterialRole.includes(:translations).where(cabinet_member: true).order(:seniority).select { |role| role.current_role_appointments.any? }
+  end
```
ryanb-gds, line 74: This should probably move belong on the `Ministerial` role class. Might be nice to have a `cabinet_members` scope as well

db/schema.rb
```diff
@@ -79,7 +79,7 @@
     t.string "email"
     t.integer "call_for_evidence_response_form_id"
     t.text "postal_address"
-    t.index ["call_for_evidence_response_form_id"], name: "index_cfes_participations_on_cfes_response_form_id"
+    t.index ["call_for_evidence_response_form_id"], name: "index_cons_participations_on_cons_response_form_id"
```
ryanb-gds, line 82: This shouldn't be here. Weird how this keeps regenrating itself in some people's local envs

features/step_definitions/cabinet_ministers_steps.rb
```diff
@@ -16,8 +16,11 @@
 
 Given(/^there are multiple Cabinet minister roles$/) do
   organisation = create(:organisation)
-  create(:ministerial_role, name: "Role 1", cabinet_member: true, organisations: [organisation], seniority: 0)
-  create(:ministerial_role, name: "Role 2", cabinet_member: true, organisations: [organisation], seniority: 1)
+  person = create(:person, forename: "Tony")
+  minister_role1 = create(:ministerial_role, name: "Role 1", cabinet_member: true, organisations: [organisation], seniority: 0)
+  minister_role2 = create(:ministerial_role, name: "Role 2", cabinet_member: true, organisations: [organisation], seniority: 1)
+  create(:role_appointment, role: minister_role1, person:)
+  create(:role_appointment, role: minister_role2, person:)
```
ryanb-gds, line 23: Can we add `with appointments` or similar to the name of this method to be a bit more explicit?

app/controllers/admin/cabinet_ministers_controller.rb
ryanb-gds: Also is this creating an N+1 query by checking the current role appointments for each role? We should probably just inner join to the role_appointments table which will naturally filter out any ministerial roles without appointments

ryanb-gds (changes requested):
Dodgy DB migration in there. I'm pretty sure we've got an unnecessary N+1 query in there as well

db/schema.rb
KludgeKML: I think this happens if people only ever run migrations and don't do DB resets every so often. (If  this existed once in a branch but never got merged, it'll be in the schema of the actual DB of people who worked on that branch and ran the migration that included it, and the way db/schema is made, I believe, is just dumping the actual DB schema at the end of the migration. If you get the current main branch and then do a db:reset that should fix it from appearing again?)

ryanb-gds (approved):
Beautiful 👌

PR #9177, opened 2024-06-19
Improve unique title error message

ryanb-gds:
This change is intended to make it clearer to users that they may not be able to find the document that has a similar slug, and that they should use a different title anyway.

This should reduce the likelihood of users opening support tickets when they are unable to find a document with a similar title.

Zendesk ticket: https://govuk.zendesk.com/agent/tickets/5853617


sengi (approved):

KludgeKML (approved):
LGTM

PR #9178, opened 2024-06-19
Move publication scopes to the correct model

pezholio:
This was moved to the wrong place in https://github.com/alphagov/whitehall/pull/9175/commits/15bc63f1a3243d212a9c4f1f1a003ae76ac31473


ryanb-gds (approved):
Thanks for jumping on this so quickly!

PR #9179, opened 2024-06-19
Update Publishing API in more predictable order when toggling reshuffle mode

ChrisBAshton:
This PR does two things:

- Makes the calls to Publishing API more specific and synchronous when toggling reshuffle mode, to minimise any risk of hitting a race condition whereby the republishes happen out of order (which could mean continuing to show the ministers index page when we actually want to show a maintenance message, or visa versa). This is all in the penultimate commit.
- Introduces a new `PatchLinksPublishingApiWorker` which simplifies some of the callback code surrounding the ministers index page. It is enqueued when ministers are reordered, so that we can async patch the ministers index page, and stop putting/publishing the page (which are effectively no-op operations).

Tested on Integration.

Trello: https://trello.com/c/w0L32BNJ/2692-investigate-race-condition-on-ministers-order-page

### DIFF
--- a/app/controllers/admin/cabinet_ministers_controller.rb
+++ b/app/controllers/admin/cabinet_ministers_controller.rb
@@ -17,7 +17,7 @@ class Admin::CabinetMinistersController < Admin::BaseController
 
   def order_cabinet_minister_roles
     MinisterialRole.reorder_without_callbacks!(order_ministerial_roles_params, :seniority)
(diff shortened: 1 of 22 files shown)


brucebolt (approved):

ryanb-gds (approved):
Gosh, this is horribly awkward. Well done for figuring out a way through

PR #9180, opened 2024-06-19
SPIKE: creating an Object Store for Content Blocks 

Harriethw:


lib/object_store/app/validators/object_store/content_block_validator.rb
```diff
@@ -0,0 +1,27 @@
+require "json-schema"
+
+# TODO: Rename "validator" to something else
+class ObjectStore::ContentBlockValidator
+  EMAIL_ADDRESS_SCHEMA = Rails.root.join("lib/object_store/config/object_store/schemas/email_address.json")
```
Harriethw, line 5: we now have the `/schemas` endpoint on the Publishing API so maybe this would be a call to that/that would handle validation?

config/application.rb
```diff
@@ -15,6 +15,11 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+# Require all Engines inside the `packages` directory
+Dir.glob("lib/**/engine.rb", base: Rails.root).each do |path|
```
pezholio, line 19: I wonder whether we should have all of our engines in `lib/engines/`, just so we're clearer about what these things are. We should probs change the commit title here too.

lib/object_store/app/models/object_store/content_block.rb
```diff
@@ -0,0 +1,8 @@
+class ObjectStore::ContentBlock
```
pezholio: I think we should be good to inherit from the Edition model now, but I guess that can come later 👍

lib/object_store/app/validators/object_store/content_block_validator.rb
pezholio: Yeah, I think we should be able to use the endpoints we setup before now

pezholio: However, we'll probably need to add a schema first (though we could/should stub out in test) - I'll look into that

db/migrate/20240619101531_create_content_blocks.rb
```diff
@@ -0,0 +1,10 @@
+class CreateContentBlocks < ActiveRecord::Migration[7.1]
```
pezholio, line 1: I think we'll need to inherit from the Edition model and use the `editions` table here. There's some prior art in this PR - https://github.com/alphagov/whitehall/pull/9112

lib/object_store/app/controllers/object_store/content_blocks_controller.rb
```diff
@@ -1,5 +1,58 @@
 class ObjectStore::ContentBlocksController < ApplicationController
+  include ObjectStore::Engine.routes.url_helpers
+
+  before_action :set_content_block, only: %i[show update destroy]
+
+  def info
+    block_type = params.require(:block_type)
+    @schema = ObjectStore::ContentBlockValidator.schema_for(block_type)
```
pezholio, line 8: I'm wondering whether we should break up the fetching of the schema from the Validator itself? It feels a bit weird to fetch a schema from a validator - then we can call the Schema fetching service within the validator when validating

lib/object_store/app/controllers/object_store/content_blocks_controller.rb
```diff
@@ -1,5 +1,58 @@
...
+  def show
+    render json: @content_block
+  end
+
+  def new
+    block_type = params.require(:block_type)
+
+    @schema = ObjectStore::ContentBlockValidator.schema_for(block_type)
+
+    ActionController::Parameters.action_on_unpermitted_parameters = false
```
pezholio, line 25: What does this do here?

config/routes.rb
```diff
@@ -466,4 +466,6 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
   mount Flipflop::Engine => "/flipflop"
 
   mount GovukPublishingComponents::Engine, at: "/component-guide" unless Rails.env.production?
+
+  mount ObjectStore::Engine, at: "/object-store"
```
pezholio, line 470: I _think_ we'll need to mount the Engine inside the `government/admin` namespace here. I'm not sure why we do that (perhaps @ryanb-gds could advise?), but it seems like most of the functionality is there.

pezholio (commented):
Looking good so far, just some general comments that popped into my head as I worked through. Looks to be on the right track though! 👍

config/routes.rb
ryanb-gds: It's a legacy thing, Whitehall used to render public facing GOV.UK pages, and the publishing app was at `government/admin`. But the frontend stuff was broken out into a separate app, `government-frontend`

pezholio: Ah ha - so we don't necessarily _need_ to route within `/government/admin` then?

ryanb-gds: I'm not sure, to be honest. I think there are probably things like layouts, maybe some middlewares, that are automatically applied at that path. But if there's a good reason not to use it then I expect you could make it work.

lib/object_store/app/controllers/object_store/content_blocks_controller.rb
Harriethw: ah this is a temporary workaround as the next line to get the `properties` params threw an error because it's missing `block_type` (`action_on_unpermitted_parameters` is set to error elsewhere)... will tidy up

lib/engines/object_store/app/views/object_store/content_blocks/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1>New <%= @content_block.block_type %></h1>
+
+<%= form_with(model: @content_block, url: "/object-store/content_blocks", local: true) do |form| %>
```
Harriethw, line 3: I had to hardcode the path here because otherwise it would default to `/object-store/assets/whitehall?action=index&controller=object_store%2Fcontent_blocks`... I'm not sure what middleware was causing that

pezholio: I think we might have more predictable results by replacing:

```
ObjectStore::Engine.routes.draw do
```

with:

```
Whitehall::Application.routes.draw do
```

in the routes.rb file

lib/engines/object_store/config/routes.rb
```diff
@@ -0,0 +1,10 @@
+ObjectStore::Engine.routes.draw do
+  namespace :object_store, path: "/" do
+    root to: "content_blocks#index", via: :get
+
+    resources :content_blocks do
+      get "new/:block_type", to: "content_blocks#new", on: :collection, as: :new
+      get ":block_type/schema", to: "content_blocks#info", on: :collection, as: :info
+    end
```
pezholio, line 8: I think I'd prefer the URLs to look like this:

```
object_store/content_blocks # Show all content blocks
object_store/content_blocks/new # Create a new content block (ask what block type is required)
object_store/content_blocks/:object_type/new # Render the form for a new :object_type
object_store/content_blocks/:object_type # Create an :object_type
etc etc
```

I think we can do this like so:

```
    resources :content_blocks do
      scope ":block_type" do
        resources :content_blocks, path: "/"
      end
    end
```

config/application.rb
Harriethw: have made this change now to `lib/engines`

lib/engines/object_store/app/views/object_store/content_blocks/new.html.erb
Harriethw: ah yep, that seems to have fixed things

lib/engines/object_store/docs/adr/0001-create-object-store.md
```diff
@@ -0,0 +1,57 @@
...
+
+All our users who may need to add/edit content blocks should have access to the Whitehall app. 
+
+### 2. We will build it in an engine
+
+We want to ensure that we are 'good neighbours' and can create this service in a way that isn't polluting the Whitehall
+repo. We also want to build with a view to this service being used outside of Whitehall. It makes sense therefore to put 
+the code in an engine separate from the rest of the repo.
+
+### 3. We will inherit from Editions
```
Harriethw, line 35: I've tried to add Editions inheritance to this branch, but coming up against a lot of difficulty both practical and conceptual. As this draft ADR says, [Edition is tightly coupled to Document](https://github.com/alphagov/whitehall/blob/edition-refactoring/docs/adr/0003-introduce-reusable-content-items.md) and that's the first difficult to cross that makes just creating a Content Block difficult. More thoughts below

Using editions

* We don’t currently know for certain whether we need **the exact same** drafting/publishing/auditing/change history process as Editions - even if we needed exactly the same process, could we add the needed features as we go rather than introducing them up front, e.g [via composition](https://github.com/alphagov/whitehall/pull/9112#discussion_r1626128237)? Or is there room to have a simplified process that sits outside of the existing legacy complexity? 
* We’ll have fields and functionality in Content Blocks that don’t relate to Content Blocks - e.g. an Edition requires a Document to be present but Content Blocks will have different relationship to Documents. Lots of validation and other relationships as well just from creating the simplest version of an Edition
* We’ll be adding things that are not related to Editions to an Editions table
* We’ll be dependent on the Editions model/table across the different publishing apps. If we ever wanted to extract this engine into a standalone gem/service, would it not be better to have our own models?
* Maintainability - future changes to either Editions or Content Blocks will have to ensure both types can handle them.

Using JSON blob
* Adding new fields under one column as JSON is a new way of adding to the Editions model
* If we include meta data like object_type, reason for change, auditing etc. in a JSON blob, will it be harder to index/search and find relational data from JSON?

lib/engines/object_store/docs/adr/0001-create-object-store.md
```diff
@@ -0,0 +1,57 @@
+# 3. Build an Object Store in Whitehall to create/edit Content Blocks
```
Harriethw, line 1: duplicating/re-wording a bit of [what's in this doc ](https://docs.google.com/document/d/1b-jgRmr-CrtuUhPkgPZbFBt1UG0f8wJ4hOFLTAS00D0/edit)to try and understand it and to narrow the scope to just the creation of the object store...

lib/engines/object_store/docs/adr/0001-create-object-store.md
Harriethw: pro's to not using inheritance:
* we could potentially get an MVP up and running quicker, while the publishing workflow gets ironed out in parallel

lib/object_store/app/validators/object_store/content_block_validator.rb
Harriethw: I've had a look at the schemas on Publishing API and I'm not sure we can use without changing the current configuration quite dramatically - there's a generation process that adds / assumes a lot about the fields and schema  https://github.com/alphagov/publishing-api/blob/9dc6d2a7c36809069748cb5457465bebf983c558/docs/content_schemas/adding-a-new-schema.md

lib/engines/object_store/docs/adr/0001-create-object-store.md
pezholio: This is a really good summary, thanks! I think, as the very least, we'll need Content Blocks to be versioned and have a relation to a root `Document` element, and I'm worried as we go, we'll keep picking up various different "Edition-like" behaviours that we need. So how about we add a new `Editionable` Concern that's private to us only that (for now):

- Specifies `editions` as the table we want to use (with `self.table_name`) 
- Includes the `Edition::Identifiable` concern (which sets up the relation between the Editions and the Document)

As we progress we can then identify the behaviours we need and import them into our Edtionable concern. This will also have the side effect of seperating out the stuff that makes our model editionable from the stuff that is specific to it being a content block (i.e. the JSON columns, validation etc) and will also help us identify the bare minimum of behaviour required to make a model behave like an edition.

Whaddya reckon?

Harriethw: sounds like a good plan!

Harriethw: > * Specifies editions as the table we want to use (with self.table_name)
> * Includes the Edition::Identifiable concern (which sets up the relation between the Editions and the Document)

I like the idea of starting with our own private concern and adding to it as needed, but I think doing both of these things above would still cause all of the same problems I've outlined... I'll have a go at sketching something but it will mean workarounds/adding more specific behaviour to our use case, e.g. currently a `slug` is needed to create a Document.

pezholio: I think that's fine - at the end of the day, our stuff should really behave in the same way as any other piece of content in the Publishing platform, so it should have a slug, a UUID etc etc.

lib/engines/object_store/docs/adr/0001-create-object-store.md
```diff
@@ -0,0 +1,56 @@
...
+
+* Specifies editions as the table we want to use (with self.table_name)
+* Includes the Edition::Identifiable concern (which sets up the relation between the Editions and the Document)
+
+As we progress we can then identify the behaviours we need and import them into our Edtionable concern. This will also 
+have the side effect of separating out the stuff that makes our model editionable from the stuff that is specific to it 
+being a content block (i.e. the JSON columns, validation etc) and will also help us identify the bare minimum of behaviour 
+required to make a model behave like an edition.
+
+### 4. The source of truth for published Blocks will be the Publishing API
```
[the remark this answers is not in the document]
Harriethw: ah ok, I copied this [over from the original ADR](https://docs.google.com/document/d/1b-jgRmr-CrtuUhPkgPZbFBt1UG0f8wJ4hOFLTAS00D0/edit)
I assumed this meant that for any other services wanting to get what the latest Content Block is they will talk to the Publishing API rather than Whitehall... but Whitehall will be pushing out the latest versions and keeping track of the blocks. Maybe there's a better of way of wording this

lib/engines/object_store/docs/adr/0001-create-object-store.md
tahb: This was a helpful read, thank you. I very much agree with the new direction 👍 

> As we progress we can then identify the behaviours we need and import them into our Edtionable concern.

Given we don't yet have a clear picture of what we'll need, any integration decisions which make now will be extremely difficult to get right first time and very likely need several more adjustments.  

In summary I think we want both to:

1. move fast and learn more about what we'll need 
2. integrate into Whitehall for reuse whilst being a good neighbour

I think a good strategy here will be to **quickly** spin up a proof of concept within Whitehall (given we shouldn't make a prototype?) and separate ourselves **as much as possible**, duplicating things for a time if necessary. 

As we move through and find that "we need versioning" or "part x of the publishing flow", each will become an isolated opportunity to work with the Whitehall team on. We can collaborate on _each part_ to find a jointly compatible approach with more ease rather than trying to plan for _all/multiple_ parts upfront.

If Whitehall aren't able to refactor their side over at the same time then there will at least be agreed direction for them to take. In addition there would be a plan and precedent for any other publishing journey looking to move to Whitehall too. 

Iterating like this should enable us to build quickly and make incremental refactorings to Whitehall? 

Then nearer the end when we have a firm idea of everything we need we can look to consolidate any remaining duplication such as models or tables that we originally duplicated.

pezholio: Yeah, I'm definitely leaning towards this approach, the one tiny voice that rings in my mind after this though:

> Then nearer the end when we have a firm idea of everything we need we can look to consolidate any remaining duplication such as models or tables that we originally duplicated.

With the best will in the world, will we _actually_ do this? Or will other priorities take over and we end up with more technical debt? 

I do agree though, we need to focus on delivering now, so we can find things out. There's definitely a happy medium somewhere there.

tahb: > With the best will in the world, will we actually do this? Or will other priorities take over and we end up with more technical debt?

Absolutely. I think it'll be up to us to explain and advocate for the program of work needed. We could share the trade off we're making now with others nice and early so when it comes to repaying it it won't be a surprise to anyone. Hopefully if it's expected it can be better planned for it and so more likely to happen?  

It could be that we try our best and don't succeed, in which case I'd like the next folks code that's prepared for it and really good notes 🤞

lib/engines/object_store/docs/adr/0001-create-object-store.md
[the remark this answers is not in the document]
tahb: Sounds like Whitehall is the source of truth here and we are synchronising _published_ blocks with the publishing API via a push?

Then when the publishing API has to translate some references it can either check locally in it's copy or ask Whitehall.

If that's correct and Publishing API just has a secondary copy that's only for reading.. I wonder if a cache with a fallback request to Whitehall wouldn't be lighter and simpler?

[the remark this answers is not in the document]
pezholio: We send all versions (drafts and published versions) to the content API. Drafts get sent to the draft content store (for previewing etc), while published versions get sent to the live content store. I did wonder about the wisdom of keeping a copy in the database too, but things like publishing flow and who created an edition and when isn't information that the Publishing API keeps. We'd also have to carry out validations and search etc via the API, which may result in slowness or missing features that we'd have to implement on the publishing API.

pezholio:
Closing this, as we've continued the work elsewhere

PR #9182, opened 2024-06-19
Document how history mode works in Whitehall

ryanb-gds:
It has been difficult to answer questions from content designers about how history mode works in preparation for the July 2024 election.

Hopefully this document will make it easier to prepare for future elections.

Trello: https://trello.com/c/eyXLVSxc


ChrisBAshton (approved):
Thanks for this ⭐

PR #9186, opened 2024-06-21
Add GA4 tracking on page load

lauraghiorghisor-tw:
Sending user and page data to GA4, on page load.

[Trello](https://trello.com/c/R7Bl3Y8n/2693-ga4-add-attributes-tracking-on-page-load)

### DIFF
--- a/app/assets/javascripts/admin/modules/ga4-page-view-tracking.js
+++ b/app/assets/javascripts/admin/modules/ga4-page-view-tracking.js
@@ -1,0 +1,14 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+;(function (Modules) {
+  function GA4PageViewTracking(module) {
+    this.module = module
+  }
+
+  GA4PageViewTracking.prototype.init = function () {
+    this.trackingData = this.module.getAttribute('data-attributes')
+    window.GOVUK.analyticsGa4.core.sendData(JSON.parse(this.trackingData))
+  }
+
+  Modules.GA4PageViewTracking = GA4PageViewTracking
+})(window.GOVUK.Modules)
--- a/app/helpers/admin/analytics_helper.rb
+++ b/app/helpers/admin/analytics_helper.rb
@@ -8,5 +8,18 @@ module Admin
         "track-label" => ActionController::Base.helpers.strip_tags(label),
       }
     end
+
+    def track_analytics_data_on_load(title)
+      {
+        event_name: "page_view",
+        page_view: {
+          government_department_name: "Government Digital Service",
+          access_level: current_user&.permissions,
+          publishing_app: "Whitehall",
+          user_created_at: current_user&.created_at&.to_date,
+          document_type: title,
+        },
+      }.to_json
+    end
   end
 end
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -9,13 +9,16 @@
   <% end %>
 <% end %>
 
+<% sanitized_title = sanitize((yield(:page_title).presence || yield(:title))) %>
+
 <%= render "govuk_publishing_components/components/layout_for_admin",
            product_name: "Whitehall Publisher",
            environment: environment,
-           browser_title: ("Error: " if yield(:error_summary).present?).to_s + sanitize((yield(:page_title).presence || yield(:title))) do %>
+           browser_title: ("Error: " if yield(:error_summary).present?).to_s + sanitized_title do %>
 
   <!-- This element exists to initialise the JS module that configures custom Universal Analytics behaviour -->
   <div data-module="app-analytics"></div>
+  <div data-module="GA4PageViewTracking" data-attributes='<%= track_analytics_data_on_load(sanitized_title) %>'></div>
 
   <%= render "govuk_publishing_components/components/skip_link" %>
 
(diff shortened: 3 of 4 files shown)


app/assets/javascripts/admin/ga4-page-view-tracking.js
```diff
@@ -0,0 +1,20 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+;(function (Modules) {
+  function GA4PageViewTracking(module) {
+    this.module = module
+  }
+
+  GA4PageViewTracking.prototype.init = function () {
+    console.log('inside this thing here')
+    this.trackingDiv = document.querySelector(
```
minhngocd, line 10: you can make this this.module instead of querying it

.eslintrc.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     es6: true
   },
   globals: {
-    GOVUK: 'readonly'
+    GOVUK: 'readonly',
+    dataLayer: 'readonly'
```
lauraghiorghisor-tw, line 11: @minhngocd this is to silence the linter on `ga4-page-view-tracking.js` - can also disable it inline 🤔

app/helpers/admin/analytics_helper.rb
```diff
@@ -8,5 +8,22 @@ def track_analytics_data(category, action, label)
...
+          access_level: access_based_roles(current_user&.permissions),
+          publishing_app: "Whitehall",
+          experience: current_user&.created_at,
+          document_type: title,
+        },
+      }.to_json
+    end
+  end
+
+  def access_based_roles(permissions)
```
lauraghiorghisor-tw, line 26: Was just in a chat with Carmen and another PA and it seems there is a character limit on GTM, so perhaps it's not ideal to send the full role array + most of the roles aren't exactly app-access related

app/helpers/admin/analytics_helper.rb
```diff
@@ -8,5 +8,22 @@ def track_analytics_data(category, action, label)
...
     end
+
+    def track_analytics_data_on_load(title)
+      {
+        event_name: "page_view",
+        page_view: {
+          government_department_name: "Government Digital Service",
+          access_level: access_based_roles(current_user&.permissions),
+          publishing_app: "Whitehall",
+          experience: current_user&.created_at,
```
lauraghiorghisor-tw, line 19: This experience field bothers me. Perhaps we can rename the key so it actually matches what we are sending + prettify the date?

app/helpers/admin/analytics_helper.rb
```diff
@@ -8,5 +8,24 @@ def track_analytics_data(category, action, label)
...
+          government_department_name: "Government Digital Service",
+          access_level: access_based_roles(current_user&.permissions),
+          publishing_app: "Whitehall",
+          user_created_at: current_user&.created_at&.to_date,
+          document_type: title,
+        },
+      }.to_json
+    end
+
+    def access_based_roles(permissions)
```
lauraghiorghisor-tw, line 25: @minhngocd I am actually wondering if we should just send "Admin"/"GDS Admin"/"Editor"?

app/views/layouts/design_system.html.erb
```diff
@@ -5,17 +5,21 @@
     <meta name="govuk:components_gem_version" content="<%= GovukPublishingComponents::VERSION %>">
     <%= javascript_include_tag "admin/domain-config" %>
     <%= javascript_include_tag "govuk_publishing_components/load-analytics" %>
+    <%= javascript_include_tag "admin/modules/ga4-page-view-tracking" %>
```
minhngocd, line 8: I don't think you need to do this - it should be included as part of application.js?

app/assets/config/manifest.js
```diff
@@ -2,4 +2,5 @@
 //= link_tree ../builds
 //= link application.js
 //= link admin/domain-config.js
+//= link admin/modules/ga4-page-view-tracking.js
```
minhngocd, line 5: should this not be in application.js? or is there an issue with it loading order?

app/assets/javascripts/admin/modules/ga4-page-view-tracking.js
```diff
@@ -0,0 +1,15 @@
...
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+;(function (Modules) {
+  function GA4PageViewTracking(module) {
+    this.module = module
+  }
+
+  GA4PageViewTracking.prototype.init = function () {
+    this.trackingData = this.module.getAttribute('data-attributes')
+    window.dataLayer = window.dataLayer || []
+    dataLayer.push(JSON.parse(this.trackingData))
```
minhngocd, line 11: did this not work? `window.GOVUK.analyticsGa4.core.sendData(data)`
do we need to push it directly to datalayer for it to appear as one?
Figuring if we use the publishing component one then if they make any changes or fixes we get it for free

minhngocd: If we use this, then we can get rid of the dataLayer exception in eslint as well

app/helpers/admin/analytics_helper.rb
minhngocd: spoke about it - a separate list to maintain which permissions to exclude sounds like effort to maintain for not a lot of value - leave it as is and see if it causes any issues.Most users won't have all the permissions

minhngocd (approved):

PR #9188, opened 2024-06-21
Close government / update ministers transactionally

richardTowers:
Currently, it's possible for this method to close the government, and then run into trouble later on while closing ministerial appointments. If one of these `save!` calls failed, we'd be left in an inconsistent state.

Wrapping the update and save! calls in a transaction will mean that they either all succeed, or are all rolled back, avoiding the chance of getting into this bad state.

I also suspect (but am not 100% sure) that doing this transactionally will significantly speed up the code, because we won't have to open / close a transaction for every role appointment we save.


ChrisBAshton (approved):
LGTM

PR #9187, opened 2024-06-21
Revert "Remove empty ministerial roles from drag and drop"

ChrisBAshton:
Reverts alphagov/whitehall#9176

This was causing a lot of duplicate roles to appear.

Trello: https://trello.com/c/Yl4ZZBc2/2690-remove-empty-roles-from-drag-and-drop-ministers-list

PR #9193, opened 2024-06-24
Remove empty ministerial roles

lauraghiorghisor-tw:
In preparation for the election, we are improving the ordering of cabinet ministers. The empty, unappointed roles, were showing in the ordering page (and during drag and drop), which made the process quite cumbersome.

We filtered out the empty roles to streamline the reordering task. The roles will only show if they have active, current appointments. The roles still show in the roles page. Other tabs on the ordering page are not affected (e.g. 'Also attends cabinet ordering').

The [original PR ](https://github.com/alphagov/whitehall/pull/9176)missed checking that the associated appointments are still active, causing role duplication, so it was [reverted](https://github.com/alphagov/whitehall/pull/9187).

[Trello card](https://trello.com/c/Yl4ZZBc2/2690-remove-empty-roles-from-drag-and-drop-ministers-list)

### DIFF
--- a/app/models/ministerial_role.rb
+++ b/app/models/ministerial_role.rb
@@ -30,6 +30,18 @@ class MinisterialRole < Role
     where(cabinet_member: true).alphabetical_by_person
   end
 
+  def self.ministerial_roles_with_current_appointments
+    includes(:translations).cabinet_members.order(:seniority).joins(:current_role_appointments)
+  end
+
+  def self.also_attends_cabinet_roles
+    includes(:translations).also_attends_cabinet.order(:seniority)
+  end
+
+  def self.whip_roles
+    includes(:translations).whip.order(:whip_ordering)
+  end
+
   def ministerial?
     true
   end
--- a/app/models/ministerial_role.rb
+++ b/app/models/ministerial_role.rb
@@ -9,6 +9,9 @@ class MinisterialRole < Role
 
   after_save :patch_links_ministers_index_page_to_publishing_api, :republish_how_government_works_page_to_publishing_api
 
+  scope :cabinet_members,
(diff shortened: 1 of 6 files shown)


features/cabinet-ministers.feature
```diff
@@ -17,7 +17,7 @@ Feature: Reordering of Cabinet ministers and Organisations
     Then I should not see a preview link to the ministers index page
 
   Scenario: Reordering Cabinet ministers
-    Given there are multiple Cabinet minister roles
+    Given there are multiple Cabinet minister roles with appointments
```
ryanb-gds, line 20: Maybe this should now say "current appointments" 🫠

ryanb-gds (approved):
Looks good to go, assuming it worked on integration this time

app/models/ministerial_role.rb
```diff
@@ -27,6 +30,18 @@ def self.cabinet
     where(cabinet_member: true).alphabetical_by_person
   end
 
+  def self.ministerial_roles_with_current_appointments
+    includes(:translations).cabinet_members.order(:seniority).joins(:current_role_appointments)
```
ChrisBAshton, line 34: I did see this [Sentry error](https://govuk.sentry.io/issues/5529352718/?referrer=slack&notification_uuid=d2446203-0e50-4fca-a483-d2ea78e13e3b&alert_rule_id=963054&alert_type=issue) appear when your branch was deployed - is there a further tweak required to this PR so that [this bit](https://github.com/alphagov/whitehall/blob/49b6515a11cae20199c4d0b03d33fd0f07610099/app/presenters/publishing_api/ministers_index_presenter.rb#L64-L70) doesn't raise an exception?

PR #9197, opened 2024-06-25
Fix republished old stats announcements

ryanb-gds:
We’ve noticed that since working on the change to the national statistics publication sub type (see attached Trello card), old statistics announcements have reappeared on [GOV.UK](http://gov.uk/), e.g. https://www.gov.uk/government/statistics/announcements/ambulance-quality-indicators-system-indicators-may-2014. This PR contains a data migration that should resolve the issue.

The issue occurred because we ran a republishing rake task on all of the statistics announcements, and the task is not idempotent so it actually created and published a new edition of each announcement, removing the redirect to the actual data set. This PR also includes a fix to the rake task so that it checks that each document can be safely republished.

Trello: https://trello.com/c/9JHwMQJV


ChrisBAshton (approved):
LGTM ✅ Quick work! ⭐

PR #9199, opened 2024-06-25
Always show "Political" checkbox on published editions (all formats)

ChrisBAshton:
All editions already have a `political` property under the hood; its value is determined on publish, [according to documented criteria](https://github.com/alphagov/whitehall/blob/dcdf572a37ff462c54a857679a301be20217da79/docs/history_mode.md#L9).

After publishing, the boolean value can be overridden manually by using the "Political" form in the UI. However, we choose not to show that form on certain formats: namely the Editionable Worldwide Organisation, for which a `can_be_marked_political?` method was added in #8840. This means it is impossible to manually change the `political` value for Editionable Worldwide Orgs. In #9198, the same setting was applied to Fatality Notices, after we'd noticed that the form was visible despite our [guidance to publishers](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#what-content-does-not-get-history-mode) stating that "fatality notices" and "national statistics" do not get "history mode".

In fact, our guidance was partially correct, because checking the box (opting into history mode) had no effect on fatality notices, because its presenters did not include the
[PoliticalDetails](https://github.com/alphagov/whitehall/blob/268e390cf866e631ccc8e7f84252aab99e43ac81/app/presenters/publishing_api/payload_builder/political_details.rb#L3) payload, i.e. we never send the value of `political` to Publishing API, so whether or not its overridden in the UI is irrelevant. In any case, looking at production data, there are no occurrences of Fatality Notices which have been attempted to be marked as political, so the checkbox rendering was removed in #9198.

This is not the case for national statistics, for which the guidance is wrong: the checkbox _is_ visible for that format, and making a national statistic as political _does_ give it 'history mode'. Of the thousands of national statistics (now termed 'accredited official statistics') and official statistics, a small handful have been manually marked as political, so we cannot unilaterally remove the checkbox for that format.

Having [discussed with colleagues in Content](https://gds.slack.com/archives/C06HBEH9TM3/p1719312642793529?thread_ts=1719233645.425039&cid=C06HBEH9TM3), we've decided to make Whitehall simpler and more predictable by simply showing the checkbox for _all_ editionable formats, and trusting publishers to only check the box when appropriate. The guidance will be updated accordingly. The logic for determining whether a format should or shouldn't get `political` status by _default_ remains unchanged.

Trello: https://trello.com/c/jhhdPAgX/2703-remove-or-fix-history-mode-option-in-fatality-notice-and-statistics

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -147,10 +147,6 @@ class Edition < ApplicationRecord
     change_note.present? || minor_change
   end
 
-  def can_be_marked_political?
-    true
-  end
-
   def can_index_in_search?
     false
   end
--- a/app/models/editionable_worldwide_organisation.rb
+++ b/app/models/editionable_worldwide_organisation.rb
@@ -177,10 +177,6 @@ class EditionableWorldwideOrganisation < Edition
     false
   end
 
-  def can_be_marked_political?
-    false
-  end
-
   def skip_world_location_validation?
     false
   end
--- a/app/models/fatality_notice.rb
+++ b/app/models/fatality_notice.rb
@@ -47,8 +47,4 @@ class FatalityNotice < Announcement
   def publishing_api_presenter
     PublishingApi::FatalityNoticePresenter
   end
-
-  def can_be_marked_political?
-    false
-  end
 end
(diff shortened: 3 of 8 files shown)


ChrisBAshton:
This was more complex than hoped. For the rule to be applied universally, we'd have to start showing the checkbox (and sending the `political` boolean downstream) for Editionable Worldwide Organisations too, which feels like a retrograde step, given we'd never want to actually use it. It's more a symptom of having to use the Edition model to get other things for free, and 'political' being coupled to that.

Will close this for now and give it another think.

PR #9204, opened 2024-06-26
Introduce foundations for the Content Object Store with a Rails Engine

tahb:
# Changes in this PR

* We break down [our larger previous spike ](https://github.com/alphagov/whitehall/pull/9180) into a smaller chunk that we think is ready to merge (the rest will follow in varying chunks)
* This slice concerns itself with adding the Content Object Store to Whitehall via a Rails Engine. [The use of a Rails Engine was discussed here](https://gds.slack.com/archives/C0776B04EJU/p1718637232472229?thread_ts=1718613361.943989&cid=C0776B04EJU)
* We add a quick controller called health check to give us some way to prove that the engine is able to respond to a request as well as tests being loaded and run as part of the whole test suite
* ADR for adding code via a Rails Engine. Merging this PR cements that decision.

Screenshot 2024-06-26 at 18 58 19

## Notes

* @pezholio and I briefly discussed the naming of "Content Object Store". The keen eyed will have seen that changed from "Object store" in the original spike. The wider Reuse team are referring to this service as "the object store" and have decided that more specific names such as Content Library aren't right. We as the Reuse dev team think "object store" runs the risk of being too generic and therefore misleading from the perspective of the Whitehall codebase. There's also the chance that the name might need to change from the user point of view so we shouldn't worry about the codebase being exactly the same. To that end we propose to keep "object store" as language for wider Reuse team but prefix it to be "content object store" for the codebase.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -43,3 +43,8 @@ Style/OptionalBooleanParameter:
 Lint/MissingSuper:
   Exclude:
     - 'app/components/**/*.rb'
+    - 'lib/engines/**/app/components/**/*.rb'
+
+Rails/SaveBang:
+  Exclude:
+    - 'Rakefile'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -473,5 +473,5 @@ Whitehall::Application.routes.draw do
 
   mount GovukPublishingComponents::Engine, at: "/component-guide" unless Rails.env.production?
 
-  mount ContentObjectStore::Engine, at: "/content-object-store"
+  mount ContentObjectStore::Engine, at: "government/admin/content-object-store"
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -472,4 +472,6 @@ Whitehall::Application.routes.draw do
   mount Flipflop::Engine => "/flipflop"
 
   mount GovukPublishingComponents::Engine, at: "/component-guide" unless Rails.env.production?
+
+  mount ContentObjectStore::Engine, at: "/content-object-store"
 end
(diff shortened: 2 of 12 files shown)


ryanb-gds (approved):
Sounds good to me, looking forward to seeing how this develops

.rubocop.yml
```diff
@@ -43,3 +43,8 @@ Style/OptionalBooleanParameter:
 Lint/MissingSuper:
   Exclude:
     - 'app/components/**/*.rb'
+    - 'lib/engines/**/app/components/**/*.rb'
+
+Rails/SaveBang:
+  Exclude:
+    - 'Rakefile'
```
Harriethw, line 50: minor nitpick - missing new line here

Harriethw (approved):
thanks for concisely pulling this together 👍

pezholio (approved):
Looks great - nice focused ADR too. LGTM 👍

config/routes.rb
```diff
@@ -472,4 +472,6 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
   mount Flipflop::Engine => "/flipflop"
 
   mount GovukPublishingComponents::Engine, at: "/component-guide" unless Rails.env.production?
+
+  mount ContentObjectStore::Engine, at: "government/admin/content-object-store"
```
Harriethw, line 476: curious as I don't know the answer - could we use `admin_root_path` as a variable here rather than hard coding `government/admin`?

tahb: I'm not aware of a way to do this. Given `admin_root_path` is an output of the configuration earlier in this file I'm not sure it'd be available to us. [I did do a quick read and couldn't see a way to make this dynamic](https://api.rubyonrails.org/v7.1.3.4/classes/ActionDispatch/Routing/Mapper/Base.html#method-i-mount).

PR #9208, opened 2024-06-27
Feature/328 create basic modelling for content object store

tahb:


lib/engines/content_object_store/test/factories/content_block.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :content_block, class: "ContentObjectStore::ContentBlock" do
```
Harriethw, line 2: should we call this `content_block_document` to signify the behaviour we are expecting/imitating?

pezholio: Yeah, I was kinda iffy on that naming, but I think it makes the intention clearer

Harriethw: cool, will rename `content_block` to `content_block_document`

lib/engines/content_object_store/app/models/content_object_store/content_block_edition.rb
```diff
@@ -0,0 +1,2 @@
+class ContentObjectStore::ContentBlockEdition < ApplicationRecord
```
pezholio, line 1: Are we going to add the relations in the next PR?

pezholio (approved):
LGTM - let's keep iterating! 👍

lib/engines/content_object_store/app/models/content_object_store/content_block_edition.rb
Harriethw: For the Edition I've added the `references` in the `content_block_edition` migration, and was thinking that a future PR could handle any logic around validating the presence of the Document (like the Identifiable concern does in the existing code).
In this PR should we add the relation to the Document class, something like this?
```
  has_many :content_block_editions,
           -> { order(created_at: :asc, id: :asc) },
           inverse_of: :content_block_document
```

pezholio: Yeah, that should do the trick 👍

PR #9217, opened 2024-07-02
Remove surprising `Timecop` that freezes time for all tests

ChrisBAshton:
This line of code was setting the system time to 11:11:11am on November 11th, 2011.

Overriding the system time using Timecop is fine, but should really be done on a per-test (or per-test-file) basis.

### DIFF
--- a/test/support/admin_edition_controller_scheduled_publishing_test_helpers.rb
+++ b/test/support/admin_edition_controller_scheduled_publishing_test_helpers.rb
@@ -14,8 +14,9 @@ module AdminEditionControllerScheduledPublishingTestHelpers
 
   module ClassMethods
     def should_allow_scheduled_publication_of(edition_type)
+      Timecop.freeze(2011, 11, 11, 11, 11, 11)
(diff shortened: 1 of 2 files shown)


ChrisBAshton:
Nah, this is going nowhere. So many tests depend on this. Even after copying the value into the relevant module, the test fails (presumably because upstream, we're setting `Timecop.freeze` to another value, which is then overridden in the call to `should_allow_scheduled_publication_of`):

```
Failure:
Admin::NewsArticlesControllerTest#test_update_should_set_scheduled_publication_if_scheduled_publication_active_checked [test/support/admin_edition_controller_scheduled_publishing_test_helpers.rb:179]:
--- expected
+++ actual
@@ -1 +1 @@
-Tue, 03 Jul 2012 09:30:00.000000000 BST +01:00
+nil


bin/rails test test/support/admin_edition_controller_scheduled_publishing_test_helpers.rb:164

................F
```

Looks like the best way forward is to live with this, and to make our own manual `Timecop.freeze` calls if being on a different date to 2011-11-11 is important.

PR #9041, opened 2024-05-14
Remove `GenericEditionPresenter`

ChrisBAshton:
Re-hashing #8093 to see whether the tests are still reliant on the generic edition presenter. Hoping to add more info to https://trello.com/c/9n7FoudZ/32-whitehall-has-a-genericeditionpresenter.

As part of moving the rendering of content out of Whitehall, we have introduced a Publishing API presenter for each type of document.

We therefore no longer need the `GenericEditionPresenter` that was being used to send placeholder content items to Publishing API.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #9226, opened 2024-07-04
Remove `GenericEditionPresenter`

pezholio:
Yet another rehash of https://github.com/alphagov/whitehall/pull/9041 - I think the tests should pass this time 🤞 

As part of moving the rendering of content out of Whitehall, we have introduced a Publishing API presenter for each type of document.

We therefore no longer need the `GenericEditionPresenter` that was being used to send placeholder content items to Publishing API, instead replacing it with a stub presenter which does the bare minimum in the support directory.


ryanb-gds (approved):
This looks a good step. Thanks @pezholio!

PR #9235, opened 2024-07-08
Block type improvements

pezholio:
This makes some improvements to how schemas are fetched / returned as well as making the `block_type` field into an enum to give us faster querying, as well as accessors and scopes to search by block type.


lib/engines/content_object_store/test/unit/app/models/content_block_schema_test.rb
```diff
@@ -27,11 +27,14 @@ class ContentObjectStore::SchemaTest < ActiveSupport::TestCase
       {
         "something" => {},
         "something_else" => {},
-        "content_block_foo" => {
+        "content_block_email_address" => {
```
tahb, line 30: I think I added foo here because I didn't want to tie these tests to the interface of email address. That said, it is more realistic and seems stable enough so perhaps all is well 👍

tahb: In fact, does the fact this change is required point to the fact that our test suite may be dependent on loading real schemas from Publishing API? If so, I think we ought to stub that out instead and these tests might be being helpful?

lib/engines/content_object_store/test/factories/content_block_schema.rb
```diff
@@ -0,0 +1,10 @@
+FactoryBot.define do
+  factory :content_block_schema, class: "ContentObjectStore::ContentBlockSchema" do
+    block_type { "email_address" }
```
tahb, line 3: We could go further while we're here to set these as transient attributes here and in any other factories as it should be easy to do while we're on the subject?

tahb (commented):
Mostly looks good. I wonder if our test suite should be decoupled from schemas defined elsewhere, maybe that wasn't the reason for the changes from foo to email?

lib/engines/content_object_store/test/unit/app/models/content_block_schema_test.rb
pezholio: The reason I changed this was because the block types now need to be valid schemas (as per `ContentObjectStore::ContentBlockSchema::VALID_SCHEMAS`). That said, I'd feel happier if there was a way we could make it feel a bit more fake in the tests. I'll have a think 🤔

pezholio: Sorted! Have broken the logic for defining what a valid schema is into a method on its own, so we can stub it out when testing 👍

lib/engines/content_object_store/test/factories/content_block_schema.rb
pezholio: Done! 👍

lib/engines/content_object_store/test/unit/app/models/content_block_schema_test.rb
```diff
@@ -21,4 +21,116 @@ class ContentObjectStore::SchemaTest < ActiveSupport::TestCase
...
+          },
+        },
+        "content_block_invalid" => {},
+      }
+    end
+
+    before(:all) do
+      Services.publishing_api.expects(:get_schemas).once.returns(response)
+      ContentObjectStore::ContentBlockSchema.stubs(:is_valid_schema?).with(anything).returns(false)
+      ContentObjectStore::ContentBlockSchema.stubs(:is_valid_schema?).with(any_of("content_block_foo", "content_block_bar")).returns(true)
```
tahb, line 62: Nice!

tahb (approved):
LGTM!

PR #9238, opened 2024-07-09
Remove data for non-editionable worldwide organisations

brucebolt:
The code to render these has been deleted and the data migrated into `EditionableWorldwideOrganisation` (and associated models). Therefore we can now delete the data.

Depends on: https://github.com/alphagov/whitehall/pull/9216.

[Trello card](https://trello.com/c/y8BtTlEF)

### DIFF
--- a/db/migrate/20240718091749_remove_non_editionable_worldwide_organisations.rb
+++ b/db/migrate/20240718091749_remove_non_editionable_worldwide_organisations.rb
@@ -1,0 +1,24 @@
+class RemoveNonEditionableWorldwideOrganisations < ActiveRecord::Migration[7.1]
+  def up
+    Asset.where(assetable_type: "WorldwideOrganisation").destroy_all
+    Attachment.where(attachable_type: "WorldwideOrganisation").destroy_all
+    Contact.where(contactable_type: "WorldwideOrganisation").destroy_all
+    EditionDependency.where(dependable_type: "WorldwideOrganisation").destroy_all
+    FeatureList.where(featurable_type: "WorldwideOrganisation").destroy_all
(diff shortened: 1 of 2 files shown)


db/migrate/20240709091749_remove_non_editionable_worldwide_organisations.rb
```diff
@@ -0,0 +1,23 @@
...
+    LinkCheckerApiReport.where(link_reportable_type: "WorldwideOrganisation").destroy_all
+    PolicyGroupDependency.where(dependable_type: "WorldwideOrganisation").destroy_all
+    SocialMediaAccount.where(socialable_type: "WorldwideOrganisation").destroy_all
+
+    drop_table :edition_worldwide_organisations
+    drop_table :sponsorships
+    drop_table :worldwide_organisation_roles
+    drop_table :worldwide_organisation_translations
+    drop_table :worldwide_organisation_world_locations
+    drop_table :worldwide_organisations
```
yndajas, line 19: Probably overkill, but thoughts on checking none of these things exist with an `updated_at` (or whatever date has that meaning) later than the migration to editionable worldwide organisations?

yndajas (approved):
Are you thinking of manually deploying to integration to see if anything fails before merging this?

db/migrate/20240709091749_remove_non_editionable_worldwide_organisations.rb
brucebolt: That's a good idea.  I'll check that after this gets merged.

yndajas (approved):

PR #9240, opened 2024-07-10
Show and list content blocks

pezholio:
This adds some simple UI to show and list content blocks. The design is taken (at least partly) from the designs that are currently undergoing user testing, so this is all subject to change. It's a useful exercise to get us a straight through journey though, as well as allowing ourselves to get familiar with concepts like View Components, which are used heavily in Whitehall

## Screenshots

image

image


lib/engines/content_object_store/app/components/content_object_store/content_block_edition/index/summary_card_component.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def title
+    content_block_edition.document.title
+  end
+
+  def summary_card_actions
+    [
+      {
+        label: "View/edit",
```
tahb, line 20: I haven't seen the prototype, this feels unusual to say just "View" so I wanted to checking it's deliberate.

tahb (approved):
LGTM! Nice work.

lib/engines/content_object_store/app/components/content_object_store/content_block_edition/index/summary_card_component.rb
pezholio: Yeah, this roughly chimes with what's in the prototype. We can always tweak later anyhoo

lib/engines/content_object_store/test/factories/content_block_edition.rb
```diff
@@ -3,6 +3,7 @@
     details { "{}" }
     created_at { Time.zone.now.utc }
     updated_at { Time.zone.now.utc }
+    content_block_document { nil }
```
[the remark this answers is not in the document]
pezholio: Yeah, good call, I defaulted it to `nil` because I didn't want to mess with the model tests (that confirm a document is created), but it's better to specify a `nil` document in those tests and have the factory create the correct thing.

PR #9245, opened 2024-07-11
Fix ContentBlockEdition tests by stubbing the new id

tahb:
I noticed 2 deprecation warnings when running the test suite this morning, eg:

```
DEPRECATED: Use assert_nil if expecting nil from lib/engines/content_object_store/test/unit/app/models/content_block_edition_test.rb:55. This will fail in Minitest 6.
```

I looked at the line `assert_equal @content_block_edition.document.content_id, @new_content_id` and found that both sides of the assertion was nil. So instead of using assert_nil there was probably a problem with th test.

I figure that @new_content_id was only available from within the stub block so we move things around to give it a file wide scope.


pezholio (approved):
Good catch!

PR #9246, opened 2024-07-11
Only show published worldwide organisations on embassies index

brucebolt:
The embassies index page currently includes all non-editionable worldwide organisations, whereas we only want the published editionable worldwide organisations to appear on this page.

[Trello card](https://trello.com/c/y8BtTlEF)


brucebolt:
Closing as I've added it into https://github.com/alphagov/whitehall/pull/9216, since there are a number of changes in other models also required.

PR #9248, opened 2024-07-11
(354) Send content blocks straight to live

pezholio:
This ensures the edition is promoted to live immediately. We aim to have a second pair of eyes process in the longer term, but this gets us over the line while that process is mapped out.

I've also added some belt and braces stuff to delete an edition if a draft is created, but it isn't published. This is unlikely, but worth handling I think.

### DIFF
--- a/lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb
+++ b/lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb
@@ -1,5 +1,7 @@
 module ContentObjectStore
   class CreateEditionService
+    class PublishingFailureError < StandardError; end
+
     def initialize(schema, edition_params)
       @schema = schema
       @edition_params = edition_params
@@ -11,6 +13,9 @@ module ContentObjectStore
         content_id = content_block_edition.document.content_id
         create_publishing_api_edition(content_id:)
         publish_publishing_api_edition(content_id:)
+      rescue PublishingFailureError => e
+        discard_publishing_api_edition(content_id:)
+        raise e
       end
     end
 
@@ -32,6 +37,12 @@ module ContentObjectStore
 
     def publish_publishing_api_edition(content_id:)
       Services.publishing_api.publish(content_id)
+    rescue GdsApi::HTTPErrorResponse => e
+      raise PublishingFailureError, "Could not publish #{content_id} because: #{e.message}"
+    end
+
+    def discard_publishing_api_edition(content_id:)
+      Services.publishing_api.discard_draft(content_id)
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb
```diff
@@ -32,6 +34,9 @@ def create_publishing_api_edition(content_id:)
 
     def publish_publishing_api_edition(content_id:)
       Services.publishing_api.publish(content_id)
+    rescue StandardError => e
```
tahb, line 37: blocker: Rescuing from something so broad as `StandardError` gave me pause here because it would also catch anything we might be unexpectedly do wrong?

```ruby
Services.publishing_api.foo(content_id)
```

But I get the intention: we do want to be somewhat broad to catch any type of error with the integration 👍 

How about `GdsApi::HTTPErrorResponse` which [is defined here](https://github.com/alphagov/gds-api-adapters/blob/9b03c6d5e8ec03ce86d79fdc1ed0efdd0b474095/lib/gds_api/exceptions.rb#L25)?

(I now realise I used `GdsApi::HTTPUnprocessableEntity` in the other tests, perhaps that was realistic but too specific so we ought to change both over?)

lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb
```diff
@@ -8,7 +10,9 @@ def initialize(schema, edition_params)
     def call
       ActiveRecord::Base.transaction do
         content_block_edition = create_whitehall_edition
-        create_publishing_api_edition(content_id: content_block_edition.document.content_id)
+        content_id = content_block_edition.document.content_id
+        create_publishing_api_edition(content_id:)
+        publish_publishing_api_edition(content_id:)
```
tahb, line 15: optional: The `publish_publishing_api_edition` method now also takes care of discarding drafts which isn't clear from this top level composer. It might be clearer to the reader to handle that here instead:

```ruby
def call
  ActiveRecord::Base.transaction do
    content_block_edition = create_whitehall_edition
    content_id = content_block_edition.document.content_id
    
    create_publishing_api_draft_edition(content_id: content_block_edition.document.content_id)
    publish_publishing_api_edition(content_id:)
  rescue => e
    discard_publishing_api_draft_edition(content_id:)
    raise e
  end
end
```

tahb (commented):
Good stuff! 

1 optional code design suggestion and a error thing I think we should change?

lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb
pezholio: Yeah, I thought about that, but we probably don't want to discard the draft if the draft creation has failed. I do agree it's clearer though

pezholio: Ah, but if I rescue from the `PublishingFailureError` method, that'll only recover from that error, so that should work

tahb (approved):
🚀

PR #9252, opened 2024-07-12
Bug fix - Only delete removed translations upon publish or force publish

minhngocd:
At the moment, on every action of the publishing API pusher, deleted translations are unpublished from Publishing API. This has caused some bugs where live translations were deleted when only the draft has been updated. This is due to the unpublish-gone redirect taking only a content ID, and hence unable to differentiate between removing the draft or removing the live edition on Publishing API. The draft of the translation that has been deleted is removed from Publishing API already during the translation deletion step.

This change will fix live translations being pulled prematurely. The only problem we have left to solve is the preview not reflecting exactly the correct translation status of the new draft edition.

https://trello.com/c/CDd3uwCj/2564-fix-bug-that-triggers-deletion-of-live-translation-when-in-draft-state

### DIFF
--- a/app/services/service_listeners/publishing_api_pusher.rb
+++ b/app/services/service_listeners/publishing_api_pusher.rb
@@ -7,13 +7,9 @@ module ServiceListeners
     end
 
     def push(event:, options: {})
-      # This is done synchronously before the rest of the publishing.
-      # Currently (02/11/2016) publishing-api links
-      # are not recalculated on parent documents when their translations are unpublished.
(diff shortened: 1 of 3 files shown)


ryanb-gds (approved):
Looks good, hope it doesn't have any unforeseen knock on effects. I do wonder why `handle_translations` was being called for all events in the first place 🤔 

The translation handling is another thing in Whitehall that would be much more elegant if the document was the first class concept rather than the edition, I think

PR #9257, opened 2024-07-15
Do not attempt to redirect unpublished contacts

brucebolt:
When unpublishing a contact (which is associated with a worldwide office), we can't publish a redirect, since contacts don't have base paths.

Therefore updating to publish 'gone' instead. This mirrors the behaviour in `PublishesToPublishingApi` that was implemented prior to worldwide organisations being made editionable.

[Trello card](https://trello.com/c/T7dF1B7k)

### DIFF
--- a/test/unit/app/services/service_listeners/publishing_api_associated_documents_test.rb
+++ b/test/unit/app/services/service_listeners/publishing_api_associated_documents_test.rb
@@ -268,9 +268,24 @@ module ServiceListeners
           "en",
         )
 
-        PublishingApiRedirectWorker.any_instance.expects(:perform).with(
+        call(new_edition)
+      end
+
+      test "with a contact on the old version of an editionable worldwide organisation publishes gone for the contact" do
+        worldwide_organisation = create(:published_editionable_worldwide_organisation, :with_main_office)
+
+        new_edition = worldwide_organisation.create_draft(create(:writer))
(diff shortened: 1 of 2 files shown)


test/unit/app/services/service_listeners/publishing_api_associated_documents_test.rb
```diff
@@ -268,15 +268,69 @@ class Publish < PublishingApiAssociatedDocumentsTest
...
+        new_edition.publish!
+
+        old_office = worldwide_organisation.main_office
+
+        PublishingApiGoneWorker.any_instance.expects(:perform).with(
+          old_office.contact.content_id,
+          nil,
+          nil,
+          "en",
+        ).never
```
mike21573, line 329: This `any_instance`/`never` pattern feels quite awkward to me, but I'm not familiar enough with Whitehall testing to make an alternative suggestion. I assume we're just limited to something a lot like this by the shape of the code?

mike21573 (approved):

test/unit/app/services/service_listeners/publishing_api_associated_documents_test.rb
brucebolt: It's definitely a pattern we use elsewhere in Whitehall.  I'm not sure there's any other ideal solution really.

PR #9260, opened 2024-07-15
Validate presence of a Content Block's Title

pezholio:
This does some simple validation to ensure the presence of a Content Block's title, as well as pull in the error messages component. There's more work to do around validating the actual details of the block, but I thought it would be better to get this in first to make both bits more reviewable.

## Screenshot

image


tahb (approved):
This looks good to me 👍 

Thanks for rebasing that first commit on my changes so I could merge ❤️

PR #9261, opened 2024-07-15
Content modelling/refactor publishing workflow into module

tahb:
We're currently adding the ability to update a content block which has led to this prior refactoring. We noticed that the create and update flows both share the same logic for sending data to the Publishing API.

Both want to:
- create draft
- publish draft immediately
- discard draft incase of any issues

Instead of replicating this behaviour across `CreateContentBlockService` and `UpdateContentBlockService`, we refactor the existing create behaviour into a module that we can reuse in the upcoming Update PR.


lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
```diff
@@ -2,6 +2,26 @@ module ContentObjectStore
   module Publishable
     class PublishingFailureError < StandardError; end
 
+    def publish_with_rollback(schema:, document_title:, details:)
+      raise ArgumentError, "Local database changes not given" unless block_given?
+
+      ActiveRecord::Base.transaction do
+        content_block_edition = yield
```
tahb, line 9: Here we perform either the Whitehall logic to create or update a content block, before letting the module code take care of the requests and rollbacks.

lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb
```diff
@@ -2,32 +2,14 @@ module ContentObjectStore
...
-          schema_id: @schema.id,
-          document_title: @edition_params[:document_title],
-          details: @edition_params[:details].to_h,
-        )
-        publish_publishing_api_edition(content_id:)
-      rescue PublishingFailureError => e
-        discard_publishing_api_edition(content_id:)
-        raise e
+    def call(edition_params)
+      publish_with_rollback(schema: @schema, document_title: edition_params[:document_title], details: edition_params[:details]) do
```
pezholio, line 10: I like this a lot! 💯

pezholio (approved):
LGTM - looks like a much nicer approach now, and something we can build on 👍

lib/engines/content_object_store/test/unit/app/lib/content_object_store/publishable_test.rb
```diff
@@ -0,0 +1,20 @@
+require "test_helper"
+
+class ContentObjectStore::PublishableTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  class PublishableTestClass
+    include ContentObjectStore::Publishable
+  end
+
+  test "it raises an error if a block isn't passed since changes need to be made locally" do
```
tahb, line 10: @pezholio for transparency I added this test after your review

PR #9262, opened 2024-07-16
Catch ActiveRecord::RecordInvalid errors when creating editions

ChrisBAshton:
Until now, these errors would be swallowed up, showing a generic "Something went wrong" message to the user, but not actually generating any Sentry errors developers could go into later to diagnose (since the PUT is treated as a 422 client error, which is ignored by default).

This happened recently, where the underlying error was:

```
ActiveRecord::RecordInvalid
Validation failed: Summary can't be blank
```

We should at least be surfacing such errors to the user, so that it is quicker and easier to get a handle on what the underlying cause may be.

NB, in #9296 the "summary can't be blank" validation was removed, but there are still other kinds of validation errors that could have been swalloed up - in the test example we're now surfacing the "NoFootnotesInGovspeakValidator" result.

Trello: https://trello.com/c/AVhwYYgP/2737-investigate-consultation-not-allowing-new-edition-to-be-created

### DIFF
--- a/test/functional/admin/editions_controller_test.rb
+++ b/test/functional/admin/editions_controller_test.rb
@@ -81,6 +81,25 @@ class Admin::EditionsControllerTest < ActionController::TestCase
     assert_redirected_to edit_admin_publication_path(draft_edition.reload)
   end
 
+  test "failing to revise an edition should surface any ActiveRecord errors" do
+    published_consultation = create(:consultation_with_public_feedback_html_attachment)
+    public_feedback = published_consultation.public_feedback
+    public_feedback.summary = <<~MARKDOWN
+      Here is a simple footnote[^1]. With some additional text after it.
+
+      [^1]: Footnote which is not allowed by consultations.
+    MARKDOWN
+    public_feedback.save!(validate: false)
+
+    post :revise, params: { id: published_consultation }
+
+    assert_redirected_to admin_consultation_path(published_consultation)
+    assert_match(
+      "Validation failed: Summary cannot include footnotes on this type of document (Summary includes '[^1]')",
+      flash[:alert],
+    )
+  end
+
   test "failing to revise an edition should redirect to the existing draft" do
     published_edition = create(:published_publication)
     existing_draft = create(:draft_publication, document: published_edition.document)
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -158,6 +158,8 @@ class Admin::EditionsController < Admin::BaseController
       redirect_to edit_admin_edition_path(@edition.document.latest_edition),
                   alert: new_draft.errors.full_messages.to_sentence
     end
+  rescue ActiveRecord::RecordInvalid => e
+    redirect_to show_or_edit_path, alert: e.to_s
   end
 
   def diff


test/functional/admin/editions_controller_test.rb
```diff
@@ -81,6 +81,25 @@ class Admin::EditionsControllerTest < ActionController::TestCase
...
+  test "failing to revise an edition should surface any ActiveRecord errors" do
+    error_message = "Validation failed: Summary can't be blank"
+    I18n.stubs(:t).returns("foo") # for any other calls to `I18n.t`
+    I18n.stubs(:t).with(:"activemodel.errors.messages.record_invalid", anything).returns(error_message)
+
+    published_consultation = create(:closed_consultation) do |consultation|
+      Object.stub_const("InvalidRecord", Class.new { include ActiveModel::Model }) do
+        exception = ActiveRecord::RecordInvalid.new(InvalidRecord.new)
+        trait = Edition::Traits::Trait.any_instance.stubs(:process_associations_after_save).raises(exception)
+        consultation.stubs(:traits).returns([trait])
```
ryanb-gds, line 93: I don't really like the amount of stubbing we're doing here... would it be possible to do this using a factory? You can always save an invalid edition by setting the `validate: false` override when saving the model (and `build` rather than `create` on the factory), if that's the problem you're working around

ChrisBAshton: Thanks for the tip - I think I tried and gave up on a `validate: false` option (perhaps because the validation needs to be disabled on the ConsultationPublicFeedback, rather than the Consultation itself). Have given it another go, much cleaner now 👍

test/functional/admin/editions_controller_test.rb
```diff
@@ -81,6 +81,25 @@ class Admin::EditionsControllerTest < ActionController::TestCase
...
   end
 
+  test "failing to revise an edition should surface any ActiveRecord errors" do
+    published_consultation = create(:consultation_with_public_feedback_html_attachment)
+    public_feedback = published_consultation.public_feedback
+    public_feedback.summary = <<~MARKDOWN
+      Here is a simple footnote[^1]. With some additional text after it.
+
+      [^1]: Footnote which is not allowed by consultations.
+    MARKDOWN
```
ChrisBAshton, line 91: NB, I did try exposing the [SafeHtmlValidator](https://github.com/alphagov/whitehall/blob/59b7250a02139f36725fcd495a4df92f92de0f66/app/models/consultation_response.rb#L10) using:

```ruby
public_feedback.summary = "<script>alert('not allowed')</script>"
```

...but it validated fine, despite it failing when checking the console directly:

```
Govspeak::HtmlValidator.new("<script>alert('not allowed')</script>").valid?
=> false
```

That's the [same code as appears in the validator](https://github.com/alphagov/whitehall/blob/108623e34bfcfca22ed5b0d2b46a9ada2d102aba/app/validators/safe_html_validator.rb#L33-L43), so I suspect the [default `@attributes` setting](https://github.com/alphagov/whitehall/blob/108623e34bfcfca22ed5b0d2b46a9ada2d102aba/app/validators/safe_html_validator.rb#L15) doesn't do what we think it does i.e. nothing is being validated unless an explicit `attribute` or `attributes` keyword is passed ([example](https://github.com/alphagov/whitehall/blob/1b4391b328cd3b20afdb54622d9c62833cf7bb09/app/models/govspeak_content.rb#L5)). Out of scope for this PR, but something I intend to look into a bit later 👀

ryanb-gds (approved):
Looks sound to me

PR #9269, opened 2024-07-16
Fail minitest step if any tests fail

pezholio:
We had a [PR where a failure snuck through when one of the matrix steps failed](https://github.com/alphagov/whitehall/actions/runs/9955264703/job/27502756556), and the success step was skipped, but didn't prevent the PR from being merged. This updates the final `run-minitest` job to expressly fail if any of the steps are in a failing, cancelled or skipped state, which should address this issue.


ryanb-gds (approved):
Good catch, thanks @pezholio

PR #9270, opened 2024-07-16
Validate content block detail fields

pezholio:
This builds on https://github.com/alphagov/whitehall/pull/9260 to validate against the `details` fields when creating Content Blocks. 

This was a bit trickier than I first thought, because we need to marshal the error messages into the correct format, which is problematic because JSON schema errors for missing fields apply to the whole object, not individual fields. Also, ActiveRecord errors have to either apply to one field, or none, so we have to do a bit of extra work to make the errors map to an individual item in the `details ` hash.

## Screenshot

image


Harriethw (approved):
LGTM!

PR #9271, opened 2024-07-16
Refactor ContentBlockEdition create to accepted nested document attrs

tahb:
## Changes in this PR

It was becomign cumbersome to remember which attributes were for the document and which were for the edition when it came to creating and updating.

Instead of having to track this we can use the Rails `accepts_nested_attributes_for` to help us take care of this. The main Whitehall repository takes this approach[1].

With it we’re able to send around a single blob of params, pass them to the Edition ORM and it’ll figure out what to do, including validations.

* We have to make one override to the title error message so we see “Title can’t be blank” rather than “Content document title can’t be blank”.
* We also adjust the hidden field for block tag so it puts the params in the same place. Inside `content_block_edition`, rather than `content_block_document_attributes ` at the top level. We’ll now have `content_block_edition: { details: {}, content_block_document_attributes: {}}`. This feels more conventional, what do we think?
* In renaming these attributes we also fix the form so the error messages correctly link to the title field on error

https://github.com/user-attachments/assets/1fea5533-d092-4ffc-9aa1-90e57e77342c


lib/engines/content_object_store/app/models/content_object_store/content_block_edition.rb
```diff
@@ -1,3 +1,5 @@
 class ContentObjectStore::ContentBlockEdition < ApplicationRecord
   include ContentObjectStore::Identifiable
+
+  accepts_nested_attributes_for :content_block_document
```
tahb, line 4: Side question: how are we deciding what lives in this class vs in Identifiable? I can see we have block_type accessors in there but not sure what the boundary of that concern is anymore?

Harriethw: Should `Identifiable` be limited to only copying behaviour/scope of existing `Identifiable`, or what would be shared between the two types of Editions? in which case, `block_type` should be lifted out, as that's unique to our Editions?

lib/engines/content_object_store/test/unit/app/models/content_block_edition_test.rb
```diff
@@ -10,14 +10,14 @@ class ContentObjectStore::ContentBlockEditionTest < ActiveSupport::TestCase
     @details = { "some_field" => "some_content" }
     @title = "Document title"
 
-    @content_block_edition = build(
-      :content_block_edition,
-      :email_address,
+    @content_block_edition = ContentObjectStore::ContentBlockEdition.new(
```
tahb, line 13: We switch from using the factory here as we want to test that ActiveRecord sets up the Edition and Document objects correctly, rather than FactoryBot.

lib/engines/content_object_store/app/models/content_object_store/content_block_edition.rb
pezholio: I think `block_type` is OK in the `Identifiable` concern, as it's required for the relationship between the Edition and the Document. If that's the case, then this `accepts_nested_attributes_for` setup should be in here too I think

lib/engines/content_object_store/test/unit/app/models/content_block_edition_test.rb
pezholio: Nice 👍

lib/engines/content_object_store/app/models/content_object_store/content_block_edition.rb
pezholio: In fact, once we have this in, the `block_type` and `title` stuff can be removed too

pezholio:
Looks good - I totally forgot about `accepts_nested_attributes_for` - the cukes seem to be failing for some reason though...

lib/engines/content_object_store/test/unit/app/models/content_block_edition_test.rb
tahb: I actually only found it because our stub of `create_random_id` was no longer being called 😬

lib/engines/content_object_store/app/models/content_object_store/content_block_edition.rb
tahb: > I find the name Identifiable actually a bit obscure in meaning and maybe that's causing the confusion?...

Yes this is where I'm at, I'm not clear what the meaning is which is half the issue. Is it everything to do with being associated with a Document perhaps, so`Documentable`? Once we're aligned it's easy to move this around.

pezholio: Yeah, that's how I read it, so maybe our naming is a bit off

tahb: I suspect our naming is inherited from [the original Identifable class](https://github.com/alphagov/whitehall/blob/8822d03cb96fc36f69ec92db1a0c87ede55aee3c/app/models/concerns/edition/identifiable.rb#L1). Which itself seems to be doing more than identifying things. 

I propose we rename to `Documentable` or similar and move our changes into it? @Harriethw @pezholio 

That'll leave us diverging from Whitehall main which may make it harder to reuse from later, if we get there.

pezholio: Sounds like a plan 👍

Harriethw: 👍

tahb:
Okay folks after looking at that cucumber test I had to work on customising the error messages. That's now done with help from @pezholio ❤️ 

Whilst I look at this rebase it's ready for another review please. I'm conscious the longer this type of change stays in progress the many more conflicts we'll get.

cc @Harriethw

Harriethw (approved):
think it makes sense to me...thanks for these improvements!

pezholio (approved):
LGTM 👍

lib/engines/content_object_store/config/locales/en.yml
```diff
@@ -1,3 +1,5 @@
 en:
   activerecord:
     attributes:
+      content_object_store/content_block_edition/content_block_document:
```
Harriethw, line 4: would you be able to ELI5 how this is used? Is there magic going on in the view somewhere?

tahb: Rails magic! Which is why it was so hard to find the right combination.

[What I had to go on was this piece of the docs](https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models), if you scroll down slightly you get to this:

```
In the event you need to access nested attributes within a given model, you should nest these under model/attribute at the model level of your translation file:

en:
  activerecord:
    attributes:
      user/role:
        admin: "Admin"
        contributor: "Contributor"
```

Which gave me the clue there was some magic `/` syntax to make it work for nested objects.

The trouble I had was how `content_object_store` came into the picture, i.e why not

```
content_object_store/content_block_edition/content_object_store/content_block_document:
or 
content_object_store/content_block_edition/content_block_document_attributes:
```

PR #9281, opened 2024-07-18
add Postal Address block type schema

Harriethw:
Not to be merged until the schema has been added here: https://github.com/alphagov/publishing-api/pull/2807

Note: the current schema generator puts fields in alphabetical order which is why this doesn't match the GDS order.

Screenshot 2024-07-18 at 16 37 47


tahb (approved):
🚀

PR #9283, opened 2024-07-19
Add feature flag for Content Object Store

Harriethw:
While we are still developing the object store we do not want it released to the public.

By default this will be turned on in development, but off in other environments until we’re ready to test/roll out.

Feature flag docs: https://github.com/alphagov/whitehall/blob/main/docs/feature_flags.md

### DIFF
--- a/lib/engines/content_object_store/features/object_store_feature_flag.feature
+++ b/lib/engines/content_object_store/features/object_store_feature_flag.feature
@@ -1,0 +1,9 @@
+Feature: Create a content object when feature flag is disabled
+
+  Background:
+    Given the content object store feature flag is disabled
+    Given I am a GDS admin
+
+  Scenario: GDS editor visits object store
+    When I visit the object store
+    Then I should see a permissions error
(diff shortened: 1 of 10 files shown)


lib/engines/content_object_store/features/object_store_feature_flag.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Create a content object when feature flag is disabled
+
+  Background:
```
Harriethw, line 3: is this enough of a test for the feature flag working when disabled? Not sure how much coverage we need 🤔

tahb: I think this is a reasonable 👍 

How might we increase this if we wanted to? We could write integration tests for each path, that might better catch any unexpected changes, eg `before_action :check_object_store_feature_flag, except: delete`.

tahb (approved):
Looks good and works when I play around locally 🚀

pezholio (approved):
LGTM 👍

PR #9286, opened 2024-07-19
Content modelling/337 ContentBlockEditions can be updated

tahb:
# Changes in this PR

When a user changes an existing Whitehall Edition we want to create a new Whitehall Edition and a new Edition inside Publishing API. The previous Document (created on the first edition) should not be recreated. In Whitehall the Document is updated to point to the new Edition and the title is changed. 

Within the Publishing API the `title` lives on the `Edition` itself (which threw me at first). Therefore we expect to see the Publishing API `Document` have its `stale_lock_version` number increase for each `Edition` we create. 

In the following diagram blue is the result of create, green is the result of an update, red are some of the fields we change along the way:
Untitled drawing

- We retrofit `ContentBlockDocument.latest_edition_id` and `ContentBlockDocument.latest_edition_id` as I noticed we weren't able to tell the `Document` which the latest `Edition` was - this could be pulled out into another PR
- We can't reuse the same `_form` partial as the create journey, details in the commit
- This does not handle multiple users making modifications at the same time, we will deal with that separately
- We rename the "Save and continue" button from the create journey to "Save and publish" since we have no draft state. The main driver here was I wanted to repurpose the existing cucumber form helpers and actually the button language makes sense in both create and update flows

## Wireframes

https://www.figma.com/design/F2Uq6Ju6fsSk1CEMjXm2va/Content-Modelling?node-id=416-8807&t=O3xRvuT7KEdT4si4-0

Screenshot 2024-07-22 at 10 47 00

Screenshot 2024-07-22 at 10 47 21

## What we've got 

https://github.com/user-attachments/assets/09791cb7-c575-487d-ac52-7f4482196eb9

We aren't yet at the same point of these wireframes since the change form is the same form for every field. We could break this out into one form per attribute with a patch endpoint which would take a bit more effort. 

The problem I see is that we don't have a draft state and are automatically publishing to the Publishing API. To support a form input for each field I think we need to have a draft Edition state that those incremental changes are contributing to.

Without a draft state we will be publishing a new live edition to Whitehall and to Publishing API for every field change. This amount of requests and data created feels excessive and not in the spirit of the existing publishing flow?

## Checking Publishing API

And with `govuk-docker run publishing-api-app bin/rails c` we can hop inside publishing-api to see what changes:

Screenshot 2024-07-19 at 16 53 20

```ruby
# If you'd like to run the same report as in the screenshot
puts "edition count: #{document.editions.count}", "titles:
#{document.editions.pluck(:title).join("=>")}", "details:
#{document.editions.pluck(:details).join("=>")}", 
"document_id: #{document.editions.pluck(:document_id).join("=>")}"
```

### DIFF
--- a/lib/engines/content_object_store/app/services/content_object_store/update_edition_service.rb
+++ b/lib/engines/content_object_store/app/services/content_object_store/update_edition_service.rb
@@ -31,8 +31,16 @@ module ContentObjectStore
   private
 
     def create_new_content_block_edition_for_document(edition_params:)
+      @original_content_block_edition.assign_attributes(
+        filter_params_for_validation_check(edition_params),
+      )
+
+      unless @original_content_block_edition.valid?
+        raise ActiveRecord::RecordInvalid, @original_content_block_edition
+      end
+
       new_content_block_edition = ContentObjectStore::ContentBlockEdition.new(edition_params)
-      new_content_block_edition.document = @original_content_block_edition.document
+      new_content_block_edition.content_block_document_id = @original_content_block_edition.document.id
       new_content_block_edition.save!
       new_content_block_edition
     end
@@ -45,6 +53,21 @@ module ContentObjectStore
       update_document_params.delete(:block_type)
 
       new_content_block_edition.document.update!(update_document_params)
+    end
+
+    def filter_params_for_validation_check(edition_params)
+      # Remove the `creator` as this is not modifiable and will return a false negative
+      validation_params = edition_params.except(:creator)
+
+      # Remove document `block_type`` as this is not modifiable
+      # Add the original Document ID to avoid `valid?` creating a new Document
+      if validation_params.key?(:content_block_document_attributes)
+        validation_params[:content_block_document_attributes] = validation_params[:content_block_document_attributes]
+          .except(:block_type)
+          .merge(id: @original_content_block_edition.document.id)
+      end
+
+      validation_params
     end
   end
 end
--- a/lib/engines/content_object_store/app/services/content_object_store/update_edition_service.rb
+++ b/lib/engines/content_object_store/app/services/content_object_store/update_edition_service.rb
@@ -1,0 +1,50 @@
+module ContentObjectStore
+  class UpdateEditionService
+    include Publishable
+
+    def initialize(schema, original_content_block_edition)
+      @schema = schema
+      @original_content_block_edition = original_content_block_edition
+    end
+
+    def call(edition_params)
+      raise ArgumentError, "Edition params must be provided" if edition_params.blank? || edition_params[:details].blank?
+
+      title = edition_params.dig(:content_block_document_attributes, :title) || @original_content_block_edition.document.title
+      details = edition_params[:details]
+      update_document_params = edition_params[:content_block_document_attributes] || {}
+
+      publish_with_rollback(schema: @schema, title:, details:) do
+        @new_content_block_edition = create_new_content_block_edition_for_document(edition_params:)
+
+        update_content_block_document(
+          new_content_block_edition: @new_content_block_edition,
+          update_document_params:,
+        )
+
+        @new_content_block_edition
+      end
+
+      @new_content_block_edition
+    end
+
+  private
+
+    def create_new_content_block_edition_for_document(edition_params:)
+      new_content_block_edition = ContentObjectStore::ContentBlockEdition.new(edition_params)
+      new_content_block_edition.document = @original_content_block_edition.document
+      new_content_block_edition.save!
+      new_content_block_edition
+    end
+
+    def update_content_block_document(new_content_block_edition:, update_document_params:)
+      update_document_params[:latest_edition_id] = new_content_block_edition.id
+      update_document_params[:live_edition_id] = new_content_block_edition.id
+
+      # Updates to a Document should never change its block type
+      update_document_params.delete(:block_type)
+
+      new_content_block_edition.document.update!(update_document_params)
+    end
+  end
+end
--- a/lib/engines/content_object_store/app/views/content_object_store/content_block_editions/edit.html.erb
+++ b/lib/engines/content_object_store/app/views/content_object_store/content_block_editions/edit.html.erb
@@ -1,0 +1,53 @@
+<% content_for :context, "Object store" %>
+<% content_for :title, "Change #{@schema.name}" %>
+
+<% content_for :back_link do %>
+  <%= render "govuk_publishing_components/components/back_link", {
+    href: content_object_store.content_object_store_content_block_edition_path(@content_block_edition),
+  } %>
+<% end %>
+
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @content_block_edition)) %>
+<%= form_with model: @content_block_edition, as: :content_block_edition, url: content_object_store.content_object_store_content_block_edition_path do |f| %>
+  <%= hidden_field_tag "content_block_edition[content_block_document_attributes][block_type]",
+    @schema.block_type,
+    id: "content_object_store/content_block_edition_content_block_document_block_type" %>
+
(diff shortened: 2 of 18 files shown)


lib/engines/content_object_store/app/views/content_object_store/content_block_editions/edit.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    label: {
+      text: "Title",
+    },
+    name: "content_block_edition[content_block_document_attributes][title]",
+    id: "content_object_store/content_block_edition_content_block_document_title",
+    value: @content_block_edition.content_block_document&.title,
+    error_items: errors_for(@content_block_edition.errors, "content_block_document.title".to_sym),
+  } %>
+
+  <% @content_block_edition.details.each do |field, value| %>
```
tahb, line 25: This block is not in _form and so is the main difference.

We could attempt to partial up other parts of the page but it feels like too much indirection just to dry things up.

lib/engines/content_object_store/features/edit_object.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Edit a content object
```
tahb, line 1: After a convo with @pezholio I haven't continued extending our integration here as it should be covered by this Cucumber test. Between this and the unit tests for the service it feels well covered. If I were to add more to the integration test I'd be repeating a lot of what's written in the units?

lib/engines/content_object_store/test/factories/content_block_edition.rb
```diff
@@ -22,8 +22,10 @@
     end
 
     after(:create) do |content_block_edition, evaluator|
-      document = create(:content_block_document, evaluator.content_block_document_attributes)
-      content_block_edition.update!(content_block_document_id: document.id)
+      unless content_block_edition.content_block_document_id
```
tahb, line 25: When we want to create an Edition and assign our own Document to it then we need a way to avoid creating duplicate Documents.

lib/engines/content_object_store/test/unit/app/services/update_edition_service_test.rb
```diff
@@ -0,0 +1,274 @@
...
+        assert_no_changes -> { @original_content_block_edition.document.title } do
+          ContentObjectStore::UpdateEditionService
+            .new(schema, @original_content_block_edition)
+            .call(edition_params)
+        end
+      end
+    end
+
+    describe "when no parameters are changed" do
+      it "publishes a new edition with the same values as the original" do
```
tahb, line 73: This edge case is something we might want to handle differently in future. For the user to make a change but submit the same results is a redundant action, we could accept the request, recognise no change would be made and prevent any changes, returning success like a 304 Not Modified. That would be a bit more involved so for now I think we just trust the user not to click submit when they've not actually made any changes, since that would make them the author... perhaps that's a actually workaround for someone who wants to take on ownership.

lib/engines/content_object_store/features/edit_object.feature
Harriethw: are these more 'expensive' than the integration tests do you think?
just wondering what we should do going forward...should features be for one main happy path and just test what's expected in the UI, and then integration is for happy and unhappy, and testing the underlying database?
e.g. you could put validation/error handling in integration

Harriethw:
> Without a draft state we will be publishing a new live edition to Whitehall and to Publishing API for every field change. This amount of requests and data created feels excessive and not in the spirit of the existing publishing flow?

Ah I hadn't totally clocked that an Edition can't be updated if it is published already (if I understand that correctly).
Sounds like starting to add a 'state machine' on an edition like Whitehall's `workflow.rb` will be helpful for drafts and for scheduling publication https://github.com/alphagov/whitehall/blob/8ffdfa177159b45fe00173f57f74ce4ed1df08cc/app/models/concerns/edition/workflow.rb

Harriethw (approved):
LGTM! thanks for the v helpful diagrams as well.
The only thing is the validation errors not showing on the form, but maybe that just needs a `rescue` for the errors in the controller, or could be done in another PR?

lib/engines/content_object_store/app/views/content_object_store/content_block_editions/edit.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    id: "content_object_store/content_block_edition_content_block_document_block_type" %>
+
+  <%= render "govuk_publishing_components/components/input", {
+    label: {
+      text: "Title",
+    },
+    name: "content_block_edition[content_block_document_attributes][title]",
+    id: "content_object_store/content_block_edition_content_block_document_title",
+    value: @content_block_edition.content_block_document&.title,
+    error_items: errors_for(@content_block_edition.errors, "content_block_document.title".to_sym),
```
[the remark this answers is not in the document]
tahb: You're absolutely right on the errors, I need to add support for that in before we're ready 👍

tahb:
@Harriethw I've pushed two extra commits for validation shaped work:

https://github.com/alphagov/whitehall/pull/9286/commits/922aea5ca55e82fcf88bc22c0ccd694fbc9ff5e9
https://github.com/alphagov/whitehall/pull/9286/commits/e7146c91310bb6567770f7f06d51321ffbdd4b5d

And updated the video demo to the latest version so we can see it working. 

Could you give me another quick review please 🙏

lib/engines/content_object_store/app/services/content_object_store/update_edition_service.rb
```diff
@@ -31,8 +31,20 @@ def call(edition_params)
   private
 
     def create_new_content_block_edition_for_document(edition_params:)
+      validate_existing_edition_params = edition_params.except(:creator)
+      if validate_existing_edition_params.key?(:content_block_document_attributes)
+        validate_existing_edition_params[:content_block_document_attributes] = edition_params[:content_block_document_attributes]
+          .except(:block_type)
+          .merge(id: @original_content_block_edition.document.id)
+      end
```
Harriethw, line 39: not sure I totally understand what lines 34-39 are doing... is it ensuring that the Document ID is there, and that `creator` and `block_type` are removed from the fields updated? Why do we have to remove them, because they shouldn't be changed in an update?

lib/engines/content_object_store/features/edit_object.feature
tahb: I should think Cucumber will be more expensive to run.

Before reading this I went and followed what the Create journey did and put Updates tests in Cucumber too! These did feel like common and expected scenarios from a users point of view, so felt worthy of the cost?

Defining what and when to cover code with integration tests would be good. 

At the moment we've reasonably moved the majority of the logic out of controllers and into service classes. Those services are stubbing and testing the integrations to Publishing API. 

I think a good example for integration will be something like scheduled publishing or notifications. We wouldn't have cucumber tests but would have integration and units?

Cucumber: user facing happy paths
Integration: edge cases or not user facing paths

Harriethw (approved):
looks good! just had a question to clarify

lib/engines/content_object_store/app/services/content_object_store/update_edition_service.rb
tahb: This is indeed a bit sticky. 

Since we have to call `valid?` on the _original_ object we need to be careful about how we send it information. We want the validation feedback but we need to avoid setter violations and prevent it creating a brand new `Document`.

- `creator` has to be removed since this it not an updatable field and so would falsely fail the validation every time
- `block_type` also cannot be updated since it's `attr_readonly`
- we add Document `id` to associate the records again, preventing it creating a new `Document` object with the new`title` value

tahb: I've lifted this responsibilities and given it a name and some comments as I think it's a rare case where they might be useful.

PR #9289, opened 2024-07-22
Update organisation updater rake task

lauraghiorghisor-tw:
Added a csv for running the rake task. This will be deleted in a future commit.

In the csv, the document type is given with a space. Temporarily changing the rake task to drop spaces from input. In the csv, the slugs are given as paths. Changing the logic to extract the slugs from path.

Changing the log output so that it's easier to follow error cases.


minhngocd (approved):
LGTM

PR #9290, opened 2024-07-22
Remove csv used for running a rake task

lauraghiorghisor-tw:
This is to clean up csv and some code changes done in [this PR](https://github.com/alphagov/whitehall/pull/9289)

[Trello](https://trello.com/c/I2FVtqFh/2750-switch-all-associations-with-dluhc-to-mhclg)


lauraghiorghisor-tw:
@minhngocd Should we also revert the changes to the `BulkOrganisationUpdater`? I think the path splitting and even log changes are good to keep.

minhngocd:
I agree - i think we can keep improvements to the rake task

minhngocd (approved):

PR #9291, opened 2024-07-22
Remove universal analytics code

ryanb-gds:
Universal Analytics was discontinued as of July 1st. This PR removes all Whitehall code relating to Universal Analytics. Sorry it's quite large, I'd recommend reviewing commit-by-commit (although to be honest that doesn't help an awful lot).

Trello: https://trello.com/c/czNzvFq4


app/assets/javascripts/components/govspeak-editor.js
```diff
@@ -67,16 +67,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
     return data
   }
 
-  GovspeakEditor.prototype.maybeTrackEvent = function (label) {
-    if (this.shouldTrackToggle) {
```
dnkrj, line 71: Looks like this can be removed from the init function as well

app/assets/javascripts/components/autocomplete.js
```diff
@@ -15,12 +15,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
       showAllValues: $select.multiple,
       showNoOptionsFound: true,
       onConfirm: function (query) {
-        const category = $select.getAttribute('data-track-category')
```
dnkrj, line 18: This should be removed from the html template as well

app/assets/javascripts/components/govspeak-editor.js
```diff
@@ -67,16 +67,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
...
-      GOVUK.analytics.trackEvent(
-        this.previewButton.getAttribute('data-preview-toggle-track-category'),
```
dnkrj, line 73: These attributes can be removed from the template

dnkrj (commented):
Looks good 🎉 A few missed things I've commented on and then I also found a few extra places that reference tracking listed below:
- https://github.com/alphagov/whitehall/blob/main/app/components/admin/flash_message_component.html.erb#L2
- https://github.com/alphagov/whitehall/blob/main/app/components/admin/flash_alert_component.rb#L7
- https://github.com/alphagov/whitehall/blob/main/app/components/admin/flash_notice_component.rb
- https://github.com/alphagov/whitehall/blob/main/test/support/admin_edition_world_locations_behaviour.rb#L17
- https://github.com/alphagov/whitehall/blob/main/test/support/admin_edition_controller_test_helpers.rb#L413

There are also quite a few references to tracking in the component `.yml` files.

dnkrj (approved):
🎉

PR #9292, opened 2024-07-22
Enable html attachment on foreign languages

GDSNewt:
**What**

While looking at Sentry errors we noticed that a lot of HTML attachments do not have a valid URL. This occurs when an html attachment has special characters in the title, but are not on a foreign only document (consultations mainly).

Documents have a concept of using the id as a fallback if the title of an edition creates an invalid slug. HTML attachments don’t and just return an empty string when special characters are used in the title.

This causes issues when stuff  is passed in. A html attachments slug is constructed by downcasing and underscoring the title (and some other stuff). But when special chars are passed in it returns an empty string.

New attachments blow up and can’t be created as Publishing API rejects them. There’s a small subset of older HTML Attachments which have been pushed to Publishing API and have an invalid URL that can’t be visited. See this page for example https://www.gov.uk/government/publications/china-uk-high-level-security-dialogue-official-statement

**PR changes**

Have modified a method in the html attachment model to ensure that content with locales that are not suitable for direct slug generation from titles or similar attributes (due to potential issues with special characters, non-Latin scripts, etc.) still receives a unique and URL-safe slug. By using the `content_id` as a fallback, the method guarantees that each piece of content can be referenced by a unique and stable URL. 

Have also added a migration to backfill the existing Html Attachments that do not currently have slugs on Whitehall. 

[Trello](https://trello.com/c/ZkPnDPrg/2279-html-attachments-break-if-special-characters-are-used-in-the-title)

### DIFF
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -7,7 +7,7 @@ class HtmlAttachment < Attachment
           autosave: true,
           inverse_of: :html_attachment
 
-  before_validation :clear_slug_if_non_english_locale
+  before_validation :handle_slug_for_foreign_locale
 
   validates :govspeak_content, presence: true
 
@@ -129,9 +129,9 @@ private
     sluggable_locale? ? title : nil
   end
 
-  def clear_slug_if_non_english_locale
-    if locale_changed? && !sluggable_locale?
-      self.slug = nil
+  def handle_slug_for_foreign_locale
+    if !sluggable_locale? && (locale_changed? || slug.blank?)
+      self.slug = content_id.to_s
     end
   end
 end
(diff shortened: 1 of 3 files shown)


ryanb-gds (approved):
I do wonder two things:

1. Why is it like this? Why would we ever have wanted to have an empty slug? I'm a bit worried we're missing something.
2. Should we be using friendly ID as documents do? Friendly ID does support transliteration of most locales, though not all. Usage is like https://github.com/alphagov/whitehall/blob/231a1c2255eb754853edb900a5e156c2cc1a3a2c/app/models/document.rb#L8

I think as long as we've tested this thoroughly in integration and it works then we should probably merge it, but it does alarm me slightly.

One comment to check on the data migration

PR #9293, opened 2024-07-22
Content modelling/add audit trail for creation

Harriethw:


lib/engines/content_object_store/app/components/content_object_store/content_block_edition/show/timeline_component.rb
```diff
@@ -0,0 +1,39 @@
+class ContentObjectStore::ContentBlockEdition::Show::TimelineComponent < ViewComponent::Base
```
pezholio, line 1: It'd be good to get some unit tests for the component I reckon

lib/engines/content_object_store/app/models/content_object_store/content_block_version.rb
```diff
@@ -0,0 +1,5 @@
+class ContentObjectStore::ContentBlockVersion < ApplicationRecord
```
pezholio, line 1: This is a good idea - in my head we'd be leaning on the Editions to get a picture of the history, but would be better to spell it out

lib/engines/content_object_store/app/models/content_object_store/content_block_version.rb
```diff
@@ -0,0 +1,5 @@
...
+  belongs_to :item, polymorphic: true
+  validates :event, presence: true
```
pezholio, line 3: Could `event` be an enum?

pezholio (commented):
Coupla drive by comments, but looking good!

lib/engines/content_object_store/app/models/content_object_store/content_block_version.rb
Harriethw: that's probably a good idea! instead of `create` we would have to use `created` though, because otherwise it conflicts with the ActiveRecord `create` method, but that would actually benefit us as it's how we will render/read it in the UI...

PR #9297, opened 2024-07-23
Content block edition form reused for both create and update journeys

tahb:
# Changes in this PR

[We recently added the ability to update content block editions](https://github.com/alphagov/whitehall/pull/9286). In that change we ended with different create and edit forms due to the differences on where the field values came from.

- Creation takes the form attributes from the Schema
- Updating takes the form attributes from two nested objects (Edition and Document)

In this change we use a slim form object to wrap the objects we need to assemble the form and then remove the duplication.

Some form objects are more involved and take care of things like validation. This is already taken care of, so the use of this pattern may be doing less than what the reader might expect. I've added comments to the form object to set this expectation.

## After

https://github.com/user-attachments/assets/6e48a032-8fb1-482f-aa5d-486a92bc3393

### DIFF
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block_editions_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block_editions_controller.rb
@@ -13,17 +13,9 @@ class ContentObjectStore::ContentBlockEditionsController < ContentObjectStore::B
       @schemas = ContentObjectStore::ContentBlockSchema.all
     else
       @schema = ContentObjectStore::ContentBlockSchema.find_by_block_type(params[:block_type].underscore)
-
-      form_attributes = @schema.fields.each_with_object({}) do |field, hash|
-        hash[field] = nil
-        hash
-      end
-
-      @form = ContentObjectStore::ContentBlockEditionForm.new(
+      @form = ContentObjectStore::ContentBlockEditionForm::Create.new(
         content_block_edition: ContentObjectStore::ContentBlockEdition.new,
         schema: @schema,
-        attributes: form_attributes,
-        back_path: content_object_store.content_object_store_content_block_editions_path,
       )
     end
   end
@@ -35,12 +27,7 @@ class ContentObjectStore::ContentBlockEditionsController < ContentObjectStore::B
 
     redirect_to content_object_store.content_object_store_content_block_editions_path, flash: { notice: "#{@schema.name} created successfully" }
   rescue ActiveRecord::RecordInvalid => e
-    @form = ContentObjectStore::ContentBlockEditionForm.new(
-      content_block_edition: e.record,
-      schema: @schema,
-      attributes: @schema.fields,
-      back_path: content_object_store.content_object_store_content_block_editions_path,
-    )
+    @form = ContentObjectStore::ContentBlockEditionForm::Create.new(content_block_edition: e.record, schema: @schema)
     render :new
   end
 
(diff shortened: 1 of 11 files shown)


lib/engines/content_object_store/app/controllers/content_object_store/content_block_editions_controller.rb
```diff
@@ -12,7 +12,18 @@ def new
       @schemas = ContentObjectStore::ContentBlockSchema.all
     else
       @schema = ContentObjectStore::ContentBlockSchema.find_by_block_type(params[:block_type].underscore)
-      @content_block_edition = ContentObjectStore::ContentBlockEdition.new
+
+      form_attributes = @schema.fields.each_with_object({}) do |field, hash|
+        hash[field] = nil
+        hash
+      end
```
pezholio, line 19: I think I'd feel happier with the logic being inside the Form object. That way we keep from overloading the controller with too much logic and can unit test a bit more reliably. We could potentially have two form objects that inherit from `ContentObjectStore::ContentBlockEditionForm` with the correct values for each, so `ContentObjectStore::ContentBlockEditionForm` contains the base logic:

```ruby
class ContentObjectStore::ContentBlockEditionForm
  include ContentObjectStore::Engine.routes.url_helpers
  
  attr_reader :content_block_edition, :schema

  def initialize(content_block_edition, schema)
    # blah
  end
end
```

And we have two subclasses that contain the rules for edit/update:

```ruby
class ContentObjectStore::ContentBlockEditionForm::Update < ContentObjectStore::ContentBlockEditionForm
   def attributes
     @schema.fields.each_with_object({}) do |field, hash|
         hash[field] = nil
         hash
       end
     end

    def back_path
       content_object_store.content_object_store_content_block_editions_path
    end
end 
```

```ruby
class ContentObjectStore::ContentBlockEditionForm::Edit < ContentObjectStore::ContentBlockEditionForm
   def attributes
     @schema.fields
   end

    def back_path
       content_object_store.content_object_store_content_block_editions_path
    end
end 
```

Happy to chat this through if you want 👍

tahb: Yeah that makes sense 👍

pezholio: Yeah, that's fair, I was trying to think of a more elegant way to specify create/update without a (somewhat ugly) `form_type` parameter

pezholio: I've been using Thoughtbot's Ruby Science book a lot recently to get me into the latest and greatest Ruby practice, so was inspired in part by this https://thoughtbot.com/ruby-science/replace-conditional-with-polymorphism.html

lib/engines/content_object_store/app/forms/content_object_store/content_block_edition_form.rb
```diff
@@ -2,6 +2,8 @@
 # - Creating an object requires dynamic attributes from a schema
 # - Editing an object requires attributes from the object itself
 class ContentObjectStore::ContentBlockEditionForm
+  include ContentObjectStore::Engine.routes.url_helpers
```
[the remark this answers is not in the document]
tahb: Oh good, this was left in from an approach I ditched so isn't needed anymore 👍

pezholio (approved):
Nice work! 👍

PR #9298, opened 2024-07-24
Remove unused callback

brucebolt:
Since the removal on non-editionable worldwide organisations, there is now never a situation where a worldwide office does not have an edition (covered in 650640f4ea160452ac453d758b43db5fc818d27e).

Therefore this callback is redundant.

The tests had already been deleted, so removing the callback itself.

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -23,7 +23,7 @@ class Contact < ApplicationRecord
   after_create :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
 
-  # after_commit :republish_embassies_index_page_to_publishing_api, :republish_worldwide_office_to_publishing_api
+  # after_commit :republish_embassies_index_page_to_publishing_api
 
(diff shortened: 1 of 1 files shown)


app/models/contact.rb
```diff
@@ -23,7 +23,7 @@ class Contact < ApplicationRecord
   after_create :republish_organisation_to_publishing_api
   after_destroy :republish_organisation_to_publishing_api
 
-  # after_commit :republish_embassies_index_page_to_publishing_api, :republish_worldwide_office_to_publishing_api
+  # after_commit :republish_embassies_index_page_to_publishing_api
```
yndajas, line 26: Looks like this was and still is commented out - should it be?

brucebolt: It was intentionally disabled in https://github.com/alphagov/whitehall/pull/9280 until the embassies index page issue is resolved.

yndajas (approved):

PR #9303, opened 2024-07-24
Content modelling/370: Document timeline

Harriethw:
(This depends on work to build the Document user journey, and can be included on the `show` view for a Document.)

We want to show users a history of the
updates made to content, this will be via the Document
as the model that connects the history of all the different
Editions/versions.

This PR adds a component to construct a timeline for a Document
based on the versions of it's editions. (which can now be found by
calling `content_block_versions` on a Document)

Screenshot 2024-07-24 at 11 35 41


lib/engines/content_object_store/app/models/concerns/content_object_store/has_audit_trail.rb
```diff
@@ -3,7 +3,7 @@ module HasAuditTrail
     extend ActiveSupport::Concern
 
     included do
-      has_many :content_block_versions, -> { order(created_at: :asc, id: :asc) }, as: :item
```
Harriethw, line 6: I noticed that this concern is missing unit tests, but found it really hard to write a useful one... welcome any ideas on it.

pezholio: I guess what you could do is create two versions, one with a `created_at` date of 3 days ago, and one with a `created_at` date of today and confirm that the newer one appears first?

pezholio (approved):
Looks great, thanks! I do wonder if we actually need the `TimelineComponent` now anyway?

Harriethw:
> I do wonder if we actually need the TimelineComponent now anyway?

yep that's a good point now it's not being used, I will delete it

PR #9304, opened 2024-07-24
Namespace content block models

pezholio:
It was getting a bit repetitive having to prefix _all_ our models, with `ContentBlock*`, so let's namespace them. We add a `ContentBlock` module with a `table_name_prefix` of `content_block_`, which ensures the table names are correct. 

There's a little bit more moving round we have to do with relations, which is a bit messy, but the tests all pass!


Harriethw (approved):
Nice, this should make things a bit tidier!

PR #9305, opened 2024-07-24
Content modelling/Documents can find their latest_edition

tahb:
With the move to reorganise the user journey around a document we now need to make sure that the document has an easy way to find it's latest edition.

We spend some time updating the factory to support this change. Whilst on the main redesign branch it became clear that we don't have a way to ask our factories to create multiple `Editions` with the same `Document`, we fix this by adding a transient attribute. More context can be found on the commit.


pezholio (approved):
LGTM 👍

Harriethw (approved):
👍

pezholio (approved):

PR #9306, opened 2024-07-24
URL uses content_id in case of special characters in html attachment …

GDSNewt:
**What**

While looking at Sentry errors we noticed that a lot of HTML attachments do not have a valid URL. This occurs when an html attachment has special characters in the title, but are not on a foreign only document.

This causes issues when stuff is passed in. A html attachments slug is constructed by downcasing and underscoring the title (and some other stuff). But when special chars are passed in it returns an empty string.

New attachments blow up and can’t be created as Publishing API rejects them. There’s a small subset of older HTML Attachments which have been pushed to Publishing API and have an invalid URL that can’t be visited. See this page for example https://www.gov.uk/government/publications/china-uk-high-level-security-dialogue-official-statement

**PR changes**

Have modified a method in the html attachment model to ensure that content with locales that are not suitable for direct slug generation from titles or similar attributes (due to potential issues with special characters, non-Latin scripts, etc.) still receives a unique and URL-safe slug. By using the content_id as a fallback, the method guarantees that each piece of content can be referenced by a unique and stable URL.

Screenshot 2024-07-24 at 14 20 15

[Trello](https://trello.com/c/ZkPnDPrg/2279-html-attachments-break-if-special-characters-are-used-in-the-title)

### DIFF
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -83,7 +83,9 @@ class HtmlAttachment < Attachment
   end
 
   def identifier
-    sluggable_locale? ? slug : content_id
+    return slug if slug_eligible?
+
+    content_id
   end
 
   def base_path
@@ -127,6 +129,10 @@ private
 
   def sluggable_string
     sluggable_locale? ? title : nil
+  end
+
+  def slug_eligible?
+    title.ascii_only? && sluggable_locale?
   end
 
   def clear_slug_if_non_english_locale
--- a/test/unit/app/models/html_attachment_test.rb
+++ b/test/unit/app/models/html_attachment_test.rb
@@ -94,6 +94,13 @@ class HtmlAttachmentTest < ActiveSupport::TestCase
     assert_equal "/government/publications/#{edition.slug}/#{attachment.content_id}", attachment.url
   end
 
+  test "#url works when an attachment with english locale has special characters in title" do
+    edition = create(:published_publication, :with_html_attachment)
+    attachment = edition.attachments.first
+    attachment.update!(locale: "en", title: "首次中英高级别安全对话成果声明")
+    assert_equal "/government/publications/#{edition.slug}/#{attachment.content_id}", attachment.url
+  end
+
   test "#url works with statistics" do
     statistics = create(:published_national_statistics)
     attachment = statistics.attachments.last


ryanb-gds:
I think this is probably fine for now, but I'd be really interested to know if you were able to ascertain why we don't use friendly ID for HTML attachments in the same way we do on Documents (even though [we extend the Friendly ID module](https://github.com/alphagov/whitehall/blob/da53fc47aaa0fd2dbeddf8e0505338b205cab939/app/models/html_attachment.rb#L2), we don't actually configure it for use with the slug attribute)

I also wonder if [this method](https://github.com/alphagov/whitehall/blob/da53fc47aaa0fd2dbeddf8e0505338b205cab939/app/models/html_attachment.rb#L63) is still needed or whether that's just dead code, but it looks kinda relevant to what we're doing.

All in all a very confusing issue so well done for getting a working solution together!

GDSNewt:
> I think this is probably fine for now, but I'd be really interested to know if you were able to ascertain why we don't use [friendly ID](https://norman.github.io/friendly_id/file.Guide.html) for HTML attachments in the same way we do on Documents (even though [we extend the Friendly ID module](https://github.com/alphagov/whitehall/blob/da53fc47aaa0fd2dbeddf8e0505338b205cab939/app/models/html_attachment.rb#L2), we don't actually configure it for use with the slug attribute)
> 
> I also wonder if [this method](https://github.com/alphagov/whitehall/blob/da53fc47aaa0fd2dbeddf8e0505338b205cab939/app/models/html_attachment.rb#L63) is still needed or whether that's just dead code, but it looks kinda relevant to what we're doing.
> 
> All in all a very confusing issue so well done for getting a working solution together!

Morning Ryan, I think the rationale behind the use of content_ids for non-english HTML attachments is in this [commit](https://github.com/alphagov/whitehall/commit/eb574b437d1db34e80b2656b01efe7908bf934ef)

I think `def should_generate_new_friendly_id?` is still used by English HTML attachments, removing it breaks these tests so I don't think it's dead just yet

test/unit/app/models/html_attachment_test.rb:167
test/unit/app/models/html_attachment_test.rb:196
test/unit/app/models/html_attachment_test.rb:211
test/unit/app/models/html_attachment_test.rb:252

ryanb-gds:
Thanks @GDSNewt, that commit helped me to see what's going on. I had assumed documents had transliterated slugs for translations but they don't, they use the english slug with a locale on the end, and a translation of a document shares the document's content ID. HTML attachments are completely separate documents with separate content IDs, so we can't share the english slug between them. Not easily changed, I would imagine.

ryanb-gds (approved):
This will do for now. One day I'd love for us to take a long, hard look at HTML attachments

PR #9307, opened 2024-07-24
Content modelling/369 Reorientate the user journey to start with Documents

tahb:
# Changes in this PR

- We temporarily duplicate the feature tests 
- We add in the show+index routing for Documents
- We leave behind the create+update routing for Editions
- We switch the journey to require the Document show+index page after edition creation and consolidate the tests 

### Other bits:
- We needed to make sure we test that only details of the latest_edition are visible on the Documents index page
- We add some Cucumber coverage in for back links, as my changes break them (these commits could arguably be fixed up into b67426d0fb745103b3df342fa574ebe1fc24d7a0 as it's a breaking change, however that commit is already big)

### After

(Note the changes to URL paths)

https://github.com/user-attachments/assets/aafac565-97b6-4afd-9da2-d27dbe8af935


pezholio (approved):
LGTM - thanks for persevering with this!

Harriethw (approved):
Looks good  :)

lib/engines/content_object_store/app/components/content_object_store/content_block_document/index/summary_card_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render "components/summary_card", {
```
Harriethw, line 1: it looks like the components and index/show views are still there under `content_block_edition` but are they are being used any more?

tahb: Good post. I needed to remove these and switch the existing tests over to the new Document based ones 🙌

PR #9311, opened 2024-07-25
Do not republish contacts belonging to worldwide organisations

brucebolt:
We have made worldwide organisations editionable, this means that their offices have contacts that can exist in both draft and live states. These don't use `PublishesToPublishingApi`.

This means we need to ensure when bulk republishing all contacts that we don't send draft contacts to Publishing API as if they were live, otherwise draft content will become live before it is published.

Therefore checking whether the contact (or any other type of object) has `can_publish_to_publishing_api?` before we attempt to republish.

In order to republish contacts for a worldwide organisaion, we need to republish all worldwide organisations and allow
`PublishingApiAssociatedDocuments` handle the republishing of the contact.

This has been implemented in such a way that any other models which have a mixture of values for `can_publish_to_publishing_api?` (e.g. `HistoricalAccount` and `StatisticsAnnouncement`) are also treated correctly.

[Trello card](https://trello.com/c/PxHcUng8)

### DIFF
--- a/app/services/bulk_republisher.rb
+++ b/app/services/bulk_republisher.rb
@@ -78,8 +78,6 @@ class BulkRepublisher
 private
 
   def republish_all_by_non_editionable_type(content_type_klass)
-    content_type_klass.find_each do |record|
-      Whitehall::PublishingApi.bulk_republish_async(record)
-    end
+    content_type_klass.find_each(&:bulk_republish_to_publishing_api_async)
   end
 end
--- a/test/unit/app/models/publishes_to_publishing_api_test.rb
+++ b/test/unit/app/models/publishes_to_publishing_api_test.rb
@@ -101,6 +101,16 @@ class PublishesToPublishingApiTest < ActiveSupport::TestCase
     test_object.publish_gone_to_publishing_api
   end
 
+  test "enqueues a worker to bulk republish the document on request" do
+    test_object = TestObject.new
+    class << test_object
+      include PublishesToPublishingApi
+    end
+
+    Whitehall::PublishingApi.expects(:bulk_republish_async).with(test_object).once
+    test_object.bulk_republish_to_publishing_api_async
+  end
+
   test "enqueues a worker to republish the document on request" do
     test_object = TestObject.new
     class << test_object
--- a/test/unit/app/models/publishes_to_publishing_api_test.rb
+++ b/test/unit/app/models/publishes_to_publishing_api_test.rb
@@ -1,6 +1,8 @@
 require "test_helper"
 
 class PublishesToPublishingApiTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
   class TestObject
     include ActiveModel::Validations
     include ActiveModel::Validations::Callbacks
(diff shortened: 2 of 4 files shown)


test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -232,6 +232,15 @@ class BulkRepublisherTest < ActiveSupport::TestCase
       end
     end
 
+    context "for editionable content types, like Contact attached to a WorldwideOrganisation" do
```
yndajas, line 235: This description confuses me a little, especially since we already have context about editionable types, which are handled differently. The changes here seem to be about non-editionable types. Should the contexts be something like this?
- for editionable content types, like CaseStudy (no change)
- **for non-editionable content types, like Contact, when publishable to Publishing API**
- **for non-editionable content types, like Contact, when not publishable to Publishing API**
- for non-republishable model types (no change: though we should probably change this to content types)
- for non-existant content types (no change)

test/unit/app/services/bulk_republisher_test.rb
```diff
@@ -232,6 +232,15 @@ class BulkRepublisherTest < ActiveSupport::TestCase
...
+      test "does not republish all content of the specified type via the Whitehall::Publishing API" do
+        contact = create(:worldwide_organisation, :with_main_office).offices.first.contact
```
yndajas, line 237: Then here I wonder if we could stub `can_publish_to_publishing_api?`, just because this currently feels like it requires quite a bit context on what's unique about `WorldwideOrganisation` contacts?

test/unit/app/models/publishes_to_publishing_api_test.rb
```diff
@@ -101,6 +101,16 @@ def test_published_gone_handler; end
...
 
+  test "enqueues a worker to bulk republish the document on request" do
+    test_object = TestObject.new
+    class << test_object
+      include PublishesToPublishingApi
+    end
+
+    Whitehall::PublishingApi.expects(:bulk_republish_async).with(test_object).once
+    test_object.bulk_republish_to_publishing_api_async
+  end
```
yndajas, line 112: Should we have a context/test here covering `can_publish_to_publishing_api? => false`?

app/services/bulk_republisher.rb
```diff
@@ -78,8 +78,6 @@ def republish_all_documents_by_ids(ids)
 private
 
   def republish_all_by_non_editionable_type(content_type_klass)
-    content_type_klass.find_each do |record|
-      Whitehall::PublishingApi.bulk_republish_async(record)
-    end
+    content_type_klass.find_each(&:bulk_republish_to_publishing_api_async)
   end
```
yndajas, line 82: I'm curious about the decision to put/hide the conditional logic in the `PublishesToPublishingApi` module. Is it worthy of an explanation in the commit message (without wanting to make it super long!)?

brucebolt: I added this to `PublishesToPublishingApi` to keep the logic in the same place as other similar methods performing that check, e.g. `republish_to_publishing_api_async` which is essentially the same method, just without the bulk flag.

test/unit/app/services/bulk_republisher_test.rb
brucebolt: That's much better than what I wrote, I've updated these.

test/unit/app/models/publishes_to_publishing_api_test.rb
brucebolt: Good idea.  I've added tests for both the republishing methods.

test/unit/app/services/bulk_republisher_test.rb
brucebolt: That makes the code much more understandable.  Have updated this.

test/unit/app/models/publishes_to_publishing_api_test.rb
```diff
@@ -101,23 +107,48 @@ def test_published_gone_handler; end
...
-  test "enqueues a worker to bulk republish the document on request" do
-    test_object = TestObject.new
-    class << test_object
-      include PublishesToPublishingApi
+  context "when object can be published to Publishing API" do
+    test "bulk_republish_async enqueues a worker to bulk republish the document" do
+      test_object = TestObject.new
+      class << test_object
+        include PublishesToPublishingApi
+      end
```
yndajas, line 115: Is it worth moving this to a `before` block like you do for the next context?

yndajas (approved):
Looks good!

test/unit/app/models/publishes_to_publishing_api_test.rb
brucebolt: Ooh, yes, I'll move that into a `before` to keep things consistent.

PR #9313, opened 2024-07-26
Enabled strict mode for all javascripts

ryanb-gds:
This is a pre-requisite for upgrading to GOV.UK Publishing Components v40, which requires us to have JS code compatible with GOV.UK Frontend v5, which requires the use of strict mode.

Trello: https://trello.com/c/W12S9N8P


minhngocd (approved):
👍

PR #9315, opened 2024-07-26
Remove unused edition organisations routes

ryanb-gds:
The routes for editing the relationships between editions and organisations no longer appear to be linked to from anywhere in Whitehall

This commit removes the routes, their view template and also a defunct form builder class, which was only used in the edit template.

Trello: https://trello.com/c/32QbGSkT


dnkrj (approved):
Feel likes if this form has been inaccessible for as long as it has and no one has mentioned it there's potentially some over related code related to the model that could be cleaned up?

ryanb-gds:
> Feel likes if this form has been inaccessible for as long as it has and no one has mentioned it there's potentially some over related code related to the model that could be cleaned up?

I did look into that, but I think the model is still used fairly frequently when making queries for all the orgs associated with an edition, or the editions associated with an orgs. There might be a few bits but I couldn't see anything obvious

PR #9317, opened 2024-07-29
Namespace non-model classes

pezholio:
Following from #9304, there are still a number of classes that would benefit from being namespaced inside the relevant module. I've also removed the healthcheck controller and now redirect to view all the content objects when accessing `government/admin/content-object-store/`


tahb (approved):
LGTM 🚀

PR #9320, opened 2024-07-30
Content-modelling/367: Add a 'review' journey when creating an Edition

Harriethw:
This PR adds a draft -> publish workflow for the 'create' journey for Editions.
Not included:
* How drafts versus live editions should be shown on a Document - currently we just show the latest edition on a Document view whether it's published or not - I don't think we have designs for how to indicate whether an Edition is a draft? And maybe this should happen in another PR as this is quite big already.
* a path for if the edition is just saved as 'draft' and not published
* updating editions and saving those edits as draft.

Included:
* rather than publishing on `create` of an Edition, we just create the Document and Edition within Whitehall, in a state of Draft
* the `create` Controller action now redirects to a `confirm` page which prompts the user to accept the content and publish
* this `publish` action is in a new Workflow Controller (as in existing Whitehall code) and this handles the publishing process that used to be handled by `create`.
* We introduce a `Workflow` concern (as in Whitehall) that for now just transitions an Edition from `draft` to `published` state.

https://app.mural.co/t/contentmodellingtechnotes1539/m/contentmodellingtechnotes1539/1718976902185/df334d013e46889c0bef873eab5d999bd476f9dd?wid=0-1722338637651 
Screenshot 2024-07-30 at 12 28 19

Questions

* Should we also send Drafts to the Publishing API?

### DIFF
--- a/lib/engines/content_object_store/app/components/content_object_store/content_block_edition/show/confirm_summary_list_component.rb
+++ b/lib/engines/content_object_store/app/components/content_object_store/content_block_edition/show/confirm_summary_list_component.rb
@@ -1,0 +1,36 @@
+class ContentObjectStore::ContentBlockEdition::Show::ConfirmSummaryListComponent < ViewComponent::Base
+  def initialize(content_block_edition:)
+    @content_block_edition = content_block_edition
+  end
+
+private
+
+  attr_reader :content_block_edition
+
+  def items
+    details_items.push(confirm_item).push(date_item)
+  end
+
+  def details_items
+    content_block_edition.details.map do |key, value|
+      {
+        field: "New #{key.humanize.downcase}",
+        value:,
+      }
+    end
+  end
+
+  def confirm_item
+    {
+      field: "Confirm",
+      value: "I confirm that I am happy for the content block to be changed on these pages.",
+    }
+  end
+
+  def date_item
+    {
+      field: "Publish date",
+      value: Date.current.strftime("%d %B %Y"),
+    }
+  end
+end
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
@@ -1,0 +1,15 @@
+class ContentObjectStore::ContentBlock::WorkflowController < ContentObjectStore::BaseController
+  def publish
+    edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
+    schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(edition.document.block_type)
+    new_edition = ContentObjectStore::PublishEditionService.new(
+      schema,
+    ).call(edition)
+    redirect_to content_object_store.content_object_store_content_block_document_path(new_edition.document),
+                flash: { notice: "#{new_edition.block_type.humanize} created successfully" }
+    # TODO: error handling
+    # else
+    #   redirect_to admin_edition_path(@edition), alert: edition_publisher.failure_reason
+    # end
+  end
+end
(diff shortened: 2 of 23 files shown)


lib/engines/content_object_store/config/routes.rb
```diff
@@ -4,7 +4,9 @@
 
     namespace :content_block, path: "content-block" do
       resources :documents, only: %i[index show], path_names: { new: "(:block_type)/new" }
-      resources :editions, only: %i[new create edit update], path_names: { new: "(:block_type)/new" }
+      resources :editions, only: %i[new create edit update], path_names: { new: "(:block_type)/new" } do
+        get :confirm, on: :member
```
Harriethw: maybe a better name is `confirm_publish` ?

lib/engines/content_object_store/app/components/content_object_store/content_block_edition/show/confirm_summary_list_component.rb
```diff
@@ -0,0 +1,36 @@
...
+        field: "New #{key.humanize.downcase}",
+        value:,
+      }
+    end
+  end
+
+  def confirm_item
+    {
+      field: "Confirm",
+      value: "I confirm that I am happy for the email to be changed on these pages.",
```
Harriethw, line 26: need to programmatically add in block type rather than hard code `email`

lib/engines/content_object_store/config/routes.rb
pezholio: Maybe even `review`, as that's what we're doing here

lib/engines/content_object_store/app/components/content_object_store/content_block_edition/show/confirm_summary_list_component.rb
pezholio: Maybe "content block" for now?

lib/engines/content_object_store/test/unit/app/models/concerns/workflow_test.rb
```diff
@@ -0,0 +1,9 @@
+require "test_helper"
+
+class ContentObjectStore::WorkflowTest < ActiveSupport::TestCase
+  test "publishing a draft edition transitions it into the published state" do
```
pezholio, line 4: Do we need to test for the initial state too?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
```diff
@@ -0,0 +1,15 @@
+class ContentObjectStore::ContentBlock::WorkflowController < ContentObjectStore::BaseController
```
pezholio, line 1: I quite like this being in a different controller!

lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
```diff
@@ -238,3 +238,16 @@ def should_show_edit_form_for_email_address_content_block(document_title, email_
   assert_text "Email address changed"
   expect(page).to have_selector(".timeline__byline", text: "by #{@user.name}", count: 2)
 end
+
+Then("I am asked to check my answers for the following fields") do |table|
+  fields = table.hashes
+  assert_text "Check your answers"
+  fields.each do |key, field|
```
pezholio, line 245: Could we maybe assert the text is present within the summary list?

pezholio (commented):
Looks on the right track. Just a couple of minor thoughts

lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
pezholio: Perfect 👍

lib/engines/content_object_store/config/routes.rb
Harriethw: ah yeah that's a good one too.... you don't think it needs to have `publish` in there?

pezholio: Nah, I think `review` is clear enough

Harriethw: done!

pezholio (approved):
LGTM 👍

tahb (commented):
I'm a bit late to the party but just wanted to say this looks great! 🚀

PR #9325, opened 2024-07-31
Catch invalid URI when parsing mainstream URL

ChrisBAshton:
Fixes https://govuk.sentry.io/issues/5657864742/?alert_rule_id=963054&alert_type=issue&notification_uuid=6fae429a-a184-4a8d-bcd4-a2a9aa6cbf89&project=202259&referrer=slack

---

### DIFF
--- a/app/models/detailed_guide.rb
+++ b/app/models/detailed_guide.rb
@@ -155,11 +155,13 @@ private
   def parse_base_path_from_related_mainstream_url(url)
     return nil if url.blank?
 
-    parsed_url = URI.parse(url.strip)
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
I feel that ideally rather than returning `nil` we should be capturing the exception higher up the stack and doing something sensible with it, but appreciate you probably don't want to get drawn into that as it could get a bit fiddly. This'll do for now

ChrisBAshton:
I figured we already return nil for some scenarios (`return nil if url.blank?`) so Whitehall already knows how to handle `nil` in this case. But agreed, it does indicate a failure somewhere else in the process!

PR #9328, opened 2024-07-31
Stop republishing Embassies Index so much

mike21573:
[Trello card](https://trello.com/c/u6rgcwqi/1258-incident-action-investigate-and-fix-the-issue-with-the-embassiesindex-page-which-is-no-longer-being-updated-in-response-to-event)

## Changes in this PR

In response to an incident last week (see linked Trello card) where republishes of the Embassies Index page (`/world/embassies`) hogged resources on the `whitehall-admin-worker`, preventing attachments from being uploaded to Asset Manager, we temporarily disabled republishing the embassies index at all [here](https://github.com/alphagov/whitehall/commit/38a978996c515a92cc466108068f21dd0bb14bd4).

We realised we were republishing this page on any edit of a WorldLocation, a Contact or a WorldwideOffice, meaning this would be republished 3 times (once for each model). We should ideally only be republishing it once, so we've moved this republishing `after_commit` to any edits of the parent `WorldwideOrganisation` itself, which should reduce the number of times we republish the index page.

We've left the republishing of a WorldLocation as the model feels independent from a `WorldwideOrganisation` (i.e. it's not governed by it, unlike the other two models, if we're understanding correctly).

## Next steps

The Embassies Index Presenter is still fairly slow, but we vastly improve the speed of its `#content` method in [this next PR](https://github.com/alphagov/whitehall/pull/9329).

### DIFF
--- a/app/controllers/admin/republishing_controller.rb
+++ b/app/controllers/admin/republishing_controller.rb
@@ -193,7 +193,7 @@ private
       "PublishingApi::HowGovernmentWorksPresenter",
       "PublishingApi::OperationalFieldsIndexPresenter",
       "PublishingApi::MinistersIndexPresenter",
-      # "PublishingApi::EmbassiesIndexPresenter",
+      "PublishingApi::EmbassiesIndexPresenter",
       "PublishingApi::WorldIndexPresenter",
       "PublishingApi::OrganisationsIndexPresenter",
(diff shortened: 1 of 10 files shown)


brucebolt (approved):
Looks good to me.

I think the final paragraph in the second commit message can probably be deleted now.

PR #9329, opened 2024-07-31
Speed up EmbassiesIndexPresenter

mike21573:
[Trello card](https://trello.com/c/u6rgcwqi/1258-incident-action-investigate-and-fix-the-issue-with-the-embassiesindex-page-which-is-no-longer-being-updated-in-response-to-event)

## Changes in this PR

In response to an incident last week (see linked Trello card) where republishes of the Embassies Index page (`/world/embassies`) hogged resources on the `whitehall-admin-worker`, preventing attachments from being uploaded to Asset Manager, we temporarily disabled republishing the embassies index at all [here](https://github.com/alphagov/whitehall/commit/38a978996c515a92cc466108068f21dd0bb14bd4).

In #9328 we turn the republishing back on, but reduce the number of times we're calling the "republish" method. Here we speed up the `content` method using Rails caching, lazy loading and memoisation, the combination of which drastically speed up the `#content` method.

On my local branch, using a recent dump of Integration data, these changes reduce the time taken running `PublishingApi::EmbassiesIndexPresenter.new.content` from ~15 seconds to ~0.5 seconds.

### DIFF
--- a/app/presenters/publishing_api/embassies_index_presenter.rb
+++ b/app/presenters/publishing_api/embassies_index_presenter.rb
@@ -8,21 +8,23 @@ module PublishingApi
     end
 
     def content
-      content = BaseItemPresenter.new(
-        nil,
-        title:,
-        update_type: "minor",
-      ).base_attributes
+      @content ||= begin
+        content = BaseItemPresenter.new(
+          nil,
+          title:,
+          update_type: "minor",
+        ).base_attributes
 
-      content.merge!(
-        base_path:,
-        details:,
-        document_type: "embassies_index",
-        rendering_app: Whitehall::RenderingApp::COLLECTIONS_FRONTEND,
-        schema_name: "embassies_index",
-      )
+        content.merge!(
+          base_path:,
+          details:,
+          document_type: "embassies_index",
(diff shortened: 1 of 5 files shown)


app/presenters/publishing_api/embassies_index_presenter.rb
```diff
@@ -8,21 +8,23 @@ def content_id
...
+          details:,
+          document_type: "embassies_index",
+          rendering_app: Whitehall::RenderingApp::COLLECTIONS_FRONTEND,
+          schema_name: "embassies_index",
+        )
 
-      content.merge!(PayloadBuilder::Routes.for(base_path))
+        content.merge!(PayloadBuilder::Routes.for(base_path))
+      end
     end
```
yndajas, line 28: Is it worth adding a test to say that `BaseItemPresenter.new` is only called once when calling `EmbassiesIndexPresenter.new.content` two or more times?

yndajas (approved):

app/presenters/publishing_api/embassies_index_presenter.rb
Gweaton: Good suggestion. Have just updated this.

PR #9337, opened 2024-08-02
Provide GA4 configuration for the staging env

ryanb-gds:
We've been seeing Smokey failures for the staging environment which are reporting that the browser is unable to initialise the GA4 module correctly.

We're not certain this is the actual cause of the Smokey test failure, but we want to rule it out and prevent the error.

GA4 tracking is not needed for the staging env, so we set the `initialiseGA4` property to false.


minhngocd (approved):
🚀

PR #9340, opened 2024-08-02
Add task to bulk remove / replace links to domains

richardTowers:
Occasionally, entire domains which are regularly linked to on GOV.UK break. For example, there are over 1000 links to storify.com which was:

> a social network service that let the user create stories or timelines
> using social media such as Twitter, Facebook and Instagram

(https://en.wikipedia.org/wiki/Storify)

... until it went out of business in 2018.

Sometimes these domains just disappear, but occasionally they're hoovered up by some malicious third party and end up doing spammy things, or, worse, phishy things. (Note: storify isn't an example of that - it's simply down).

In these situations, we've got a choice with no clearly good options:

- Leave the links on GOV.UK pointing to malicious sites (not good)
- Attempt to contact the "owners" of the affected pages on GOV.UK and have them fix the links (not practical, since many pages older than a couple of years are effectively unowned)
- Mark up the links in some way to show that they're broken and warn users before they click on them (confusing / scary UX)
- Archive the content (only works if the content with the broken links can actually be archived, which isn't always true. Also it's never true at the moment, as there's no way to meaningfully "archive" content on GOV.UK yet).
- Replace the links with their bare link text (what this PR does) (potentially confusing in some situations, where the link text might be "click here" or similar).
- Redirect the links to a "this link is broken" page on GOV.UK (looks bad, potentially scary UX)
- Redirect the links to another site (only works if there's a sensible place to redirect to) (this will be addressed in a subsequent commit)
- Remove the links entirely, including their link text (would break sentences / potentially make content unreadable / inaccurate)

These two rake tasks are an attempt at automating the removal or redirection of links to bad domains.

I am aware that I am committing a grave sin by [parsing HTML with regex](https://stackoverflow.com/a/1732454), but given that I had to do it for govspeak (because the govspeak parser won't round trip cleanly), I figured it was best to be consistent. The resulting regexes are pretty horrifix, so I've used the extended syntax to comment them.

I'm interested in thoughts on alternative approaches here - I had fun writing this code, but I'm not wedded to it if there's a better way. I don't think "find all the bad URLs and update them by hand" is a very good option though.

### DIFF
--- a/lib/tasks/bulk_content_updates.rake
+++ b/lib/tasks/bulk_content_updates.rake
@@ -1,0 +1,36 @@
+namespace :bulk_content_updates do
+  def editions_matching_domain(domain)
+    Edition
+      .active
+      .joins("INNER JOIN edition_translations ON edition_translations.edition_id = editions.id")
+      .where("edition_translations.body LIKE ?", "%#{domain}%")
+  end
+
+  desc "Remove all links to a particular domain - "
+  task :remove_links_to_domain, %i[domain mode] => :environment do |_, args|
+    domain, mode = args.values_at(:domain, :mode)
+    unless %w[live dry-run].include?(mode)
+      raise "mode should be 'live' or 'dry-run'"
+    end
+
+    editions_matching_domain(domain).find_each do |edition|
+      link_remover = Govspeak::LinkRemover.new(edition.body, domain)
+      next unless link_remover.match?
+
+      puts "Replacing links in #{edition.base_path} (#{mode})"
+      puts link_remover.describe_replacements
+
+      if mode == "dry-run"
+        puts "Skipping changes in dry-run mode"
+      elsif mode == "live"
+        edition.body = link_remover.remove_links_for_domain
+        edition.save!
+        PublishingApiDocumentRepublishingWorker.perform_async_in_queue(
+          "bulk_republishing",
+          edition.document_id,
+          true,
+        )
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/tasks/bulk_content_updates.rake
```diff
@@ -0,0 +1,64 @@
...
+      if mode == "dry-run"
+        puts "Skipping changes in dry-run mode"
+      elsif mode == "live"
+        edition.body = link_remover.remove_links_for_domain
+        edition.save!
+        PublishingApiDocumentRepublishingWorker.perform_async_in_queue(
+          "bulk_republishing",
+          edition.document_id,
+          true,
+        )
```
richardTowers, line 32: Not sure if this is required, as we might be able to rely on callbacks on Edition when we call save! to republish to publishing-api.

ryanb-gds: There are no callbacks on editions that save content to Publishing API, editions use an imperative workflow where a service is called from the controller to send content to Publishing API

richardTowers:
There's still an outstanding ask to remove / redirect a relatively large number of links ([spreadsheet](https://docs.google.com/spreadsheets/d/1-TN9aOl0FPk8aNZQAJWiy4__xttKiDl15dVYhD6dmZE/edit?gid=1726710970#gid=1726710970)), which I've just dropped the ball on.

I guess we could use this as a one off and then revert it, and hope that an archiving strategy comes along before we need this again. Or we could just fix all the links manually I suppose.

vignesh1507 (commented):
Can you explain more about the changes you've made?

PR #9341, opened 2024-08-05
Load GA4 modules as analytics modules

ryanb-gds:
GA4 tracking code should only be loaded if GA4 analytics is enabled. By default, Whitehall initialises all modules in window.GOVUK.modules, so this commit registers our GA4 modules specifically as analytics modules instead.

This means that GA4 code will only be loaded by the analytics initialising code in GOVUK Publishing Components

Trello: https://trello.com/c/U3Cfl1DD


app/assets/javascripts/admin/modules/ga4-button-setup.js
```diff
@@ -1,6 +1,8 @@
 'use strict'
 window.GOVUK = window.GOVUK || {}
-window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
```
richardTowers, line 5: I wonder if it would be better to add a guard to each of the `init` functions, like:

```
Ga4ButtonSetup.prototype.init = function () {
    return if !analyticsIsEnabled // note - not sure what the variable to use would be here

    ...
```

ryanb-gds: That was my first thought, but actually I think the only way I could do that would be to read the domain config and check the location myself, which is a bit painful especially considering the GA4 code in publishing components already does that.

Plus why load the modules at all if GA4 isn't enabled? These are purely GA4 analytics modules so I think having the GA4 code load them probably makes sense

richardTowers: Yep, no objections from me 👍🏻

andysellick (approved):
Good solution 👍

PR #9345, opened 2024-08-05
[DO NOT MERGE] Whitehall content API proof of concept

ryanb-gds:
Hacked together an API that Whitehall could use to serve content direct to frontend applications, rather than via Publishing API. This would simplify maintaining document state between Whitehall and the frontend, but give us other problems to solve such as performance, auth bypassing, routing etc. I believe the latter set of problems would be easier to solve than achieving distributed transactions across the GOV.UK system.


ryanb-gds:
Closing this as POC no longer relevant

PR #9350, opened 2024-08-06
Fix Edition Roles bug

callumknights:
Adds a simple guard clause to prevent republishing or WorldwideOrganisations being triggered by the updating of associated roles after they have been deleted. 

Trello: https://trello.com/c/6BkaykOY/1276-fix-issue-with-some-roles-being-unable-to-be-edited

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -74,6 +74,8 @@ class Role < ApplicationRecord
 
   def republish_associated_editions_to_publishing_api
     edition_roles.each do |edition_role|
+      next unless edition_role.edition
+
       PublishingApiDocumentRepublishingWorker.perform_async(edition_role.edition.document_id)
     end
   end
--- a/test/unit/app/models/role_test.rb
+++ b/test/unit/app/models/role_test.rb
@@ -249,6 +249,16 @@ class RoleTest < ActiveSupport::TestCase
     role.update!(name: "New role name")
   end
 
+  test "doesn't republish a worldwide organisation when a role is updated if the worldwide org is deleted" do
+    worldwide_organisation = create(:worldwide_organisation, :deleted)
+    role = create(:role_without_organisations)
+    create(:edition_role, role:, edition: worldwide_organisation)
+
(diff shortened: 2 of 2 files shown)


brucebolt (commented):
Could you update the commit message to give a little more information about why we're adding this guard clause, e.g. the user impact of not doing so.  Also, is it possible to make the commit headline fit onto one line?

callumknights:
Confirmed this fix works in integration 
image

test/unit/app/models/role_test.rb
```diff
@@ -280,4 +280,15 @@ class RoleTest < ActiveSupport::TestCase
 
     role.update!(name: "New role name")
   end
+
+  test "doesn't republish a worldwide organisation when a role is updated if the worldwide org is deleted" do
+    worldwide_organisation = create(:worldwide_organisation, :deleted)
+    role = create(:role_without_organisations)
+    create(:edition_role, role:, edition: worldwide_organisation)
+    role.reload
```
yndajas, line 288: Is the `role.reload` needed? I can't see that we're asserting on it

app/models/role.rb
```diff
@@ -74,6 +74,8 @@ def self.prime_minister_role
 
   def republish_associated_editions_to_publishing_api
     edition_roles.each do |edition_role|
+      next unless edition_role.edition
```
yndajas, line 77: Might be worth adding to the commit message an explanation that the reason this guard clause filters out edition roles with deleted editions is because the default scope of `Edition` excludes editions with a state of `deleted` (if my understanding is correct): https://github.com/alphagov/whitehall/blob/main/app/models/concerns/edition/workflow.rb#L21

yndajas (approved):
Looks good - just a couple of minor suggestions

app/models/role.rb
callumknights: Done 👍

test/unit/app/models/role_test.rb
callumknights: Removed 👍

PR #9351, opened 2024-08-06
Add rake task to report on deleted assets that are still live

brucebolt:
This adds a rake task which produces a report listing all attachments that are only attached to non-visible editions (i.e. the publisher thinks they are deleted) but are not deleted (and therefore publicly visible) in Asset Manager.

[Trello card](https://trello.com/c/e9N6IfPR)

### DIFF
--- a/lib/tasks/asset_manager.rake
+++ b/lib/tasks/asset_manager.rake
@@ -1,0 +1,28 @@
+namespace :asset_manager do
+  desc "Report on attachments that are only attached to non-visible editions in Whitehall but not deleted in Asset Manager"
+  task report_deleted_assets: :environment do
+    output = AttachmentData.find_each.map do |attachment_data|
+      attachables = attachment_data.attachments.map(&:attachable).compact
+
+      next unless attachables.any?
+      next if attachables.detect { |attachable| !attachable.is_a?(Edition) }
+      next if (attachables.map(&:state) - %w[unpublished superseded]).any?
+
+      attachment_data.assets.map do |asset|
+        asset_manager_response = GdsApi.asset_manager.asset(asset.asset_manager_id).to_h
+
+        next unless asset_manager_response["deleted"] == false
+
+        {
+          asset_url: asset_manager_response["file_url"],
+          document_path: attachables.last.public_url,
+          edition_state: attachables.last.state,
+        }
+      end
+    end
+
+    output.flatten.compact.each do |item|
+      puts [item[:asset_url], item[:document_path], item[:edition_state]].join(",")
+    end
+  end
+end
--- a/test/unit/lib/tasks/asset_manager_test.rb
+++ b/test/unit/lib/tasks/asset_manager_test.rb
@@ -37,7 +37,7 @@ class AssetManagerTest < ActiveSupport::TestCase
         end
 
         test "it should not include the attachment in the report" do
-          assert_output(/^(?!#{attachment.attachment_data.assets.first.filename},#{edition.public_url},#{edition.state}).*$/) { task.invoke }
+          refute_output(/#{attachment.attachment_data.assets.first.filename},#{edition.public_url},#{edition.state}/) { task.invoke }
         end
       end
     end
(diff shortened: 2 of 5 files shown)


lib/tasks/asset_manager.rake
```diff
@@ -0,0 +1,33 @@
...
+    output = AttachmentData.find_each.map do |attachment_data|
+      attachables = attachment_data.attachments.map(&:attachable).compact
+
+      next unless attachables.any?
+
+      non_edition_attachables = attachables.reject do |attachable|
+        attachable.is_a?(Edition)
+      end
+
+      next if non_edition_attachables.any?
```
yndajas, line 13: Given `non_edition_attachables` isn't used after this, you could simplify this and allow early exits from the iterator:

```rb
next if attachables.detect { |attachable| !attachable.is_a?(Edition) }
```

test/unit/lib/tasks/asset_manager_test.rb
```diff
@@ -0,0 +1,168 @@
...
+
+    context "when asset is deleted from asset manager" do
+      before do
+        attachment.attachment_data.assets.each do |asset|
+          stub_asset_manager_has_an_asset(asset.asset_manager_id, { deleted: true, file_url: asset.filename })
+        end
+      end
+
+      test "it should not include the attachment in the report" do
+        assert_output(/^(?!#{attachment.attachment_data.assets.first.filename},#{edition.public_url},#{edition.state}).*$/) { task.invoke }
```
yndajas, line 39: I wonder if it's worth adding a `refute_output` method somewhere to wrap around `assert_output` and make this less prone to error with the slightly complex regex. It looks like we do this in `test/unit/lib/tasks/reporting_test.rb` as well

test/unit/lib/tasks/asset_manager_test.rb
```diff
@@ -0,0 +1,168 @@
...
+      end
+    end
+  end
+
+  context "for an asset attached only to both published and superseded editions" do
```
yndajas, line 131: `attached only to both` -> `attached to both`?

test/unit/lib/tasks/asset_manager_test.rb
```diff
@@ -0,0 +1,168 @@
...
+        assert_output(/^(?!#{attachment.attachment_data.assets.first.filename},(#{Regexp.escape(edition_1.public_url)}|#{Regexp.escape(edition_2.public_url)}),(#{edition_1.state}|#{edition_2.state})).*$/) { task.invoke }
+      end
+    end
+  end
+
+  context "for an asset attached only to something that is not an edition" do
+    let(:attachment) { create(:file_attachment_with_asset, attachable: create(:double)) }
+
+    test "it should include no output in the report" do
+      assert_output("") { task.invoke }
```
yndajas, line 165: Is there a reason you can't do this in other tests? Will/might they have some output depending on factory randomness?

test/unit/lib/tasks/asset_manager_test.rb
```diff
@@ -0,0 +1,168 @@
...
+        end
+      end
+
+      test "it should not include the attachment in the report" do
+        assert_output(/^(?!#{attachment.attachment_data.assets.first.filename},#{edition.public_url},#{edition.state}).*$/) { task.invoke }
+      end
+    end
+  end
+
+  context "for an asset attached only to a superseded edition" do
```
yndajas, line 44: Is it worth drying up the first two and second two contexts? E.g. for the first two

```rb
%w[unpublished superseded].each do |state|
  context "for an asset attached only to a #{state} edition" do
      let(:edition) { create(:"#{state}_publication") }
  [...]
  end
end
```

yndajas (approved):
Looks good - just a few minor suggestions

lib/tasks/asset_manager.rake
brucebolt: Nice.  I was using this variable later on, but didn't need it in the end as I realised these were out of scope for the problem.

test/unit/lib/tasks/asset_manager_test.rb
brucebolt: Good idea.  I've added a `refute_output` method and used it in both placed.

test/unit/lib/tasks/asset_manager_test.rb
brucebolt: That seems like a good idea.  Have made this change.

test/unit/lib/tasks/asset_manager_test.rb
brucebolt: I was also considering adding a header row to the output, but never did, which is why I'm using a regex, rather than looking at the whole output.  There are some cases where something is output and we want to check the other isn't, so I've tried to keep them consistent.

PR #9353, opened 2024-08-07
Ensure topical events can be removed from organisations

kashifatcha:
The `delete_absent_topical_event_organisations` method was operating directly on the params object, but the `organisation_params` method that was passed to the organisation model was grabbing a fresh copy of the parameters, and therefore the `_destroy` parameter was not included.

By using the `organisation_params` method inside `delete_absent_topical_event_organisations`, we make sure that topical event organisation associations can be removed.

We also added a test to verify this behaviour.

https://trello.com/c/ZqWZd13U/2760-organisations-cant-be-disassociated-from-topical-events

### DIFF
--- a/app/controllers/admin/organisations_controller.rb
+++ b/app/controllers/admin/organisations_controller.rb
@@ -131,10 +131,10 @@ private
   end
 
   def delete_absent_topical_event_organisations
-    return unless params[:organisation] &&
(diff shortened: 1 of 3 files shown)


ryanb-gds (changes requested):
Couple of bits of tidying up to do but I think we're looking good

ryanb-gds (approved):
Good to go :+1:

PR #9357, opened 2024-08-08
Add warning text when creating worldwide organisations

brucebolt:
Since rolling out editionable worldwide organisations, we have seen a few instances of users intending to publish world news stories but actually publishing new worldwide organisations.

Adding some warning text to remind publishers that these are only to be created after discussion with the relevant people.

This is in a consistent position to format specific advice we show for other formats, e.g. [publications](https://github.com/alphagov/whitehall/blob/2d1417a4980d8aa926539ae1d7ea1b4f02e10cf3/app/views/admin/publications/_form.html.erb#L1-L3).

Screenshot showing the form to create a new worldwide organisation.  The text "Do not create a worldwide organisation unless you have permission from your managing editor or GOV.UK department lead." appears alongside a symbol of an exclamation mark at the top of the form.

[Trello card](https://trello.com/c/7jolFBQr)

### DIFF
--- a/app/views/admin/editions/new.html.erb
+++ b/app/views/admin/editions/new.html.erb
@@ -9,6 +9,14 @@
       items: secondary_navigation_tabs_items(@edition, request.path),
     } %>
 
+    <% if @edition.new_content_warning %>
+      <div class="format-advice">
+        <%= render "govuk_publishing_components/components/warning_text", {
+          text: @edition.new_content_warning,
+        } %>
+      </div>
+    <% end %>
+
     <%= render "form", edition: @edition %>
   </div>
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -269,6 +269,8 @@ class Edition < ApplicationRecord
     self.class.format_name
   end
 
+  def new_content_warning; end
+
   def has_parent_type?
     true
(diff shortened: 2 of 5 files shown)


app/views/admin/editions/new.html.erb
```diff
@@ -9,6 +9,14 @@
       items: secondary_navigation_tabs_items(@edition, request.path),
     } %>
 
+    <% if @edition.new_content_warning %>
+      <div class="format-advice">
```
yndajas, line 13: What impact does this class/div have? I can't see that it's used in a stylesheet from this or another alphagov repo, and it doesn't seem to be used as a test selector here

Although this is a kind of format advice, so far it seems to be used for the content that starts "Use this format for:"

yndajas (approved):
Just one question about the enclosing `div`, but generally looks good!

app/views/admin/editions/new.html.erb
brucebolt: As far as I can see, the class is redundant.  But I've chosen to keep it there as it's consistent with other similar pieces of advice, which appear visually in the same place on the form.

Gweaton (approved):

PR #9358, opened 2024-08-08
Stop tests from running multiple times (Minitest Spec DSL bug)

ChrisBAshton:
We have to be very careful about how `test` blocks are defined. Consider:

```
context "foo" do
  test { assert_something }
end

test { assert_something_else }
```

`assert_something_else` will get called _twice_ by Minitest. As described in https://github.com/minitest/minitest/pull/865:

> minitest spec styles deals with tests being included from preceding or ancestor classes within the DSL with the method `Minitest::Spec::DSL#nuke_test_methods!`. This will wipe out existing test methods that have been defined at the time of class initialization (`Minitest::Spec::DSL#create`). However, when tests are defined after a block is declared, they are still included when the instance methods for the class are pulled (since they're pulled with all the ancestor instance methods).

We therefore have to ensure that any `test` definitions are defined in a `describe` or `context` block that is as deeply nested as any preceding `test` definition.

Some tests in this file were running three times, due to their inclusion after the first `context` block. I've therefore moved the `context` blocks to the bottom of the file. This has resulted in fewer runs and assertions, and shaved 2 seconds off the test run time.

Before:

```
$ run rake test TEST=test/unit/app/models/role_test.rb

..................................................

Finished in 7.189464s, 6.9546 runs/s, 9.7365 assertions/s.
50 runs, 70 assertions, 0 failures, 0 errors, 0 skips
```

After:

```
$ run rake test TEST=test/unit/app/models/role_test.rb

........................................

Finished in 5.163601s, 7.7465 runs/s, 10.4578 assertions/s.
40 runs, 54 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/lib/engines/content_object_store/test/unit/app/models/content_block_document_test.rb
+++ b/lib/engines/content_object_store/test/unit/app/models/content_block_document_test.rb
@@ -40,16 +40,6 @@ class ContentObjectStore::ContentBlockDocumentTest < ActiveSupport::TestCase
     assert content_block_document.reload.live_edition_id, 1
   end
 
-  describe "latest_edition" do
-    it "returns the latest edition" do
-      document = create(:content_block_document, :email_address)
-      _first_edition = create(:content_block_edition, document:)
-      second_edition = create(:content_block_edition, document:)
-
-      assert_equal second_edition, document.latest_edition
-    end
-  end
-
   test "it gets its version history from its editions" do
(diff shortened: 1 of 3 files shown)


ryanb-gds (approved):
Fascinating. How did you even notice this?

ChrisBAshton:
@callumknights happened to be adding a new test to `role_test.rb` and noticed the behaviour!

PR #9361, opened 2024-08-09
Add embed code to Content Block Summary List

pezholio:
Going forward, we’d like this to be integrated into the visual editor, but this makes testing easier in the meantime.

## Screenshot

image


Harriethw (approved):
good idea!

PR #9364, opened 2024-08-09
Make government a real association of editions

ryanb-gds:
I should have used a feature flag for this so that I could deploy it in small pieces, but I didn't anticipate this requiring me to touch anywhere near the number of files that I have. The code changes were not that substantial but the number of tests that broke was enormous, which perhaps suggests we're testing some aspects of the codebase at the wrong level.

## Major Changes

Currently, Whitehall infers a relationship between editions and governments using the political boolean attribute and a date relating to the edition (usually first edition publishing date).

This means that it isn't possible for content editors to override the government in the event that a piece of content should actually relate to a different government. This is a common situation where content has been published during the days after an election.

This PR makes the government a real association of a document edition, and allows GDS editors to set the value of the association in the edition form using a select element.

## Other Considerations

It also preserves the previous behaviour of setting the initial political status at the time of the publication of the first edition based on the appropriate date attribute for each edition. In this case we select the government that was in power on the specified date. Any edition with a government association is considered to be political.

We preserve the behaviour of sending a government link for each edition to Publishing API regardless of whether or not the edition is political. We use the default government for each edition to do that if no government is set.

## Migration

I have chosen to add a foreign key index on the new `government_id` column as I expect around election time we will want to query all of the editions associated with a particular government, which would be very slow otherwise.

There is also a data migration required here. The data migration affects approx 125,000 editions (I have excluded superseded editions as they will never appear on GOV.UK again). The data migration took approximately 5 minutes to run on integration, so I think this seems reasonable to run in production.

## Follow Up

Once this has been deployed and is running well in production, I will write a migration to drop the political column on the editions table.

I also want to refactor the history mode and government behaviour on the edition model into a concern. This should include an appropriate scope for political and non-political editions to improve readability. I will create separate PRs for these.

I also need to update the History Mode docs in Whitehall!

## Screenshot

Screenshot 2024-08-13 at 17 09 00

Trello: https://trello.com/c/II8jgMR7

### DIFF
--- a/test/functional/admin/generic_editions_controller_tests/political_document_test.rb
+++ b/test/functional/admin/generic_editions_controller_tests/political_document_test.rb
@@ -2,18 +2,31 @@ require "test_helper"
 
 class Admin::GenericEditionsController::PolticalDocumentsTest < ActionController::TestCase
   tests Admin::NewsArticlesController
+  enable_url_helpers
 
   setup do
     login_as :writer
   end
 
-  view_test "displays the political checkbox for privileged users " do
-    create(:current_government)
+  view_test "displays the history mode form control for managing editor users" do
+    current_government = create(:current_government)
     login_as :managing_editor
     published_edition = create(:published_news_article, first_published_at: 2.days.ago)
     new_draft = create(:news_article, document: published_edition.document, first_published_at: 2.days.ago)
     get :edit, params: { id: new_draft }
-    assert_select "#edition_political"
+    assert_select "#edition_political input[type='checkbox'][name='edition[government_id]'][value='#{current_government.id}']"
+  end
+
+  test "gds editors can override the government associated with an edition" do
+    current_government = create(:current_government)
+    previous_government = create(:previous_government)
+    published_edition = create(:published_news_article, first_published_at: 2.days.ago, government: current_government)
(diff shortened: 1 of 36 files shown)


test/functional/admin/generic_editions_controller_tests/political_document_test.rb
```diff
@@ -2,18 +2,32 @@
...
+    current_government = create(:current_government)
     login_as :managing_editor
     published_edition = create(:published_news_article, first_published_at: 2.days.ago)
     new_draft = create(:news_article, document: published_edition.document, first_published_at: 2.days.ago)
     get :edit, params: { id: new_draft }
-    assert_select "#edition_political"
+    assert_select "#edition_political input[type='checkbox'][name='edition[government_id]'][value='#{current_government.id}']"
+  end
+
+  # TODO: Decide whether to move this into an edition test helper used by all edition types
```
ChrisBAshton, line 20: It looks fine where it is to me - what benefit are you thinking moving it to an edition test helper would give us?

ChrisBAshton (approved):
LGTM 👍 nice work! ⭐

ryanb-gds:
Closing this PR as we're planning to take a different approach that does not require us to make government a true association of the edition

PR #9367, opened 2024-08-13
Select block type changes

pezholio:
This updates the "Select block type" screen to use radio buttons, rather than links, better reflecting what we have in the designs.

### Before

image

### After

image


app/views/layouts/design_system.html.erb
```diff
@@ -50,7 +50,7 @@
         </div>
       <% end %>
 
-      <% if yield(:title).present? %>
+      <% if yield(:title).present? && yield(:hide_title).blank? %>
```
tahb, line 53: I can't see why we need to stop rendering the page title. This might just need a bit more explanation.

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/_schemas.html.erb
```diff
@@ -1,5 +1,23 @@
...
-    <li><%= link_to_new_block_type(schema) %></li>
-  <% end %>
-</ul>
+<% if @error_message %>
+  <%= render ContentObjectStore::ContentBlockEdition::New::ErrorSummaryComponent.new(
+    error_message: @error_message,
+  ) %>
+<% end %>
+
+<%= form_for :block, url: content_object_store.new_content_object_store_content_block_edition_path, method: :get do |f| %>
```
tahb, line 7: I think Rails is pushing us to start using `form_with` [these days](https://guides.rubyonrails.org/form_helpers.html#using-form-tag-and-form-for) if it's possible.

tahb (commented):
👍 Just one query on the need to change layout.rb to hide the title

app/views/layouts/design_system.html.erb
pezholio: It's because if we don't, we end up doubling up the title with the use of the radio component. If we leave out `content_for :title, "title goes here"`, the HTML `title` attribute then gets munged, which could hurt accessibility

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/_schemas.html.erb
pezholio: Done! 👍

app/views/layouts/design_system.html.erb
tahb: Okay I think I've found the problem this is trying to solve.  
Screenshot 2024-08-14 at 11 55 01

Introducing a bit of behaviour into the main layout just to serve one of our pages felt like too much of a reach. I had another look and think we can leverage the existing `page_title` context. I'll give this a go in a new commit.

tahb: @Harriethw I've added a new commit for this so I think it'd be good to get your approval here too. After my commit here's what we've got. We don't lose the page title nor the body title and we're able to avoid changing the global layout.
 
Screenshot 2024-08-14 at 12 20 17

lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
```diff
@@ -37,7 +37,8 @@
 When("I click on the {string} schema") do |schema_id|
   @schema = @schemas[schema_id]
   ContentObjectStore::ContentBlock::Schema.expects(:find_by_block_type).with(schema_id).at_least_once.returns(@schema)
-  click_link @schema.name
+  choose @schema.name
+  click_on "Save and continue"
```
Harriethw, line 41: I've noticed we have a mix of where clicking on the submit button some places is separated into a different step, other times is included in a step - very minor little quibble but would be nice to just have it as a separate step where we can I think, as when/if the button text changes we only have to change it in one place

app/views/layouts/design_system.html.erb
Harriethw: looks good to me - makes sense!

Harriethw (approved):

PR #9369, opened 2024-08-13
Content modelling/364: Review changes (Links)

Harriethw:
Screenshot 2024-08-13 at 17 46 06

Screenshot 2024-08-14 at 11 46 56

When users are editing an Edition/Document, we would like to show users the documents that have embedded that Edition so that users can make an informed decision about the knock-on effects of updating it.

This introduces more custom controller actions for 'interstitial' pages (like `review`) - this an initial solution for a multi-part form - we discussed implementing a form wizard or hand-rolling logic in one action a Controller, and decided to go for the 'dumb' solution for now and iterate when a third step in the form is introduced. (This could be next with Scheduling)

I also have not included:

1. the sort functionality on the table as this would have been added complexity to the PR - I also couldn't find examples of a sortable table in GDS or Whitehall. If we wanted to sort, this could be done in further PR by passing in the `order` param https://github.com/alphagov/publishing-api/blob/main/docs/api.md#get-v2content

2. the checkbox confirmation on the page. This can be done in another PR.

### DIFF
--- a/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
+++ b/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
@@ -59,7 +59,7 @@ private
 
   def edit_action
     {
-      href: helpers.content_object_store.edit_content_object_store_content_block_edition_path(content_block_document.latest_edition),
+      href: helpers.content_object_store.edit_content_object_store_content_block_edition_path(content_block_document.latest_edition, step: ContentObjectStore::ContentBlock::EditionsController::EDIT_FORM_STEPS[:edit_block]),
       link_text: "Change",
     }
   end
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
@@ -31,7 +31,36 @@ class ContentObjectStore::ContentBlock::EditionsController < ContentObjectStore:
     content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
     @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(content_block_edition.document.block_type)
 
-    @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(content_block_edition:, schema: @schema)
+    @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
+      content_block_edition:, schema: @schema, edition_to_update_id: content_block_edition.id,
+    )
+  end
+
+  def review_links
+    @content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
+    @content_block_document = @content_block_edition.document
+    @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(@content_block_edition.document.block_type)
+    @edition_params = edition_params
+
+    new_edition = ContentObjectStore::ContentBlock::Edition.new(@edition_params)
+    new_edition.document.id = @content_block_document.id
+
+    if new_edition.valid?
+      linked_item_service = ContentObjectStore::GetLinkedContentItems.new(
+        content_block_document: @content_block_document,
+        page: params[:page],
+      )
+
+      @linked_content_items = linked_item_service.items
+      @page_data = linked_item_service.page_data
+    else
+      @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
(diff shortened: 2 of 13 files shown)


lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
```diff
@@ -30,7 +30,36 @@ def edit
     content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
     @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(content_block_edition.document.block_type)
 
-    @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(content_block_edition:, schema: @schema)
+    @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
+      content_block_edition:, schema: @schema, edition_to_update_id: content_block_edition.id,
+    )
+  end
+
+  def review_links
```
Harriethw, line 38: as discussed with @pezholio and @tahb  - having a separate controller action per page of the form is not sustainable in long term, but to get an initial PR over the line this is what I am suggesting.

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
```diff
@@ -59,7 +59,7 @@ def creator_item
 
   def edit_action
     {
-      href: helpers.content_object_store.edit_content_object_store_content_block_edition_path(content_block_document.latest_edition),
+      href: helpers.content_object_store.edit_content_object_store_content_block_edition_path(content_block_document.latest_edition, step: ContentObjectStore::ContentBlock::EditionsController::EDIT_FORM_STEPS[:edit_block]),
```
Harriethw, line 62: is there a way not to have to keep referencing the module path...

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
```diff
@@ -27,10 +27,25 @@ def review
...
+  EDIT_FORM_STEPS = {
+    edit_block: "edit_block",
+    review_links: "review_links",
+  }.freeze
+
   def edit
+    step = params[:step]
+
+    case step
+    when EDIT_FORM_STEPS[:edit_block]
```
Harriethw, line 39: could the steps or logic be abstracted to a helper?

tahb: We could try to move the steps to a helper. That will be available to controllers, views and can be included into classes like the form objects. I'm not sure about view components.

```ruby
edit_form_step_for(params[:edit_block])
```

It would also mean we don't have to use the long module name which I can't think around that. We might be able to assign it to a shorter constant in our engine initialiser?

```ruby 
EDIT_FORM_STEPS = ContentObjectStore::ContentBlock::EditionsController::EDIT_FORM_STEPS
```

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
```diff
@@ -27,11 +27,57 @@ def review
...
+      )
+
+      @linked_content_items = linked_item_service.items
+      @page_data = linked_item_service.page_data
+
+      render :review_links
+    else
+      @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
+        content_block_edition: new_edition,
+        schema: @schema, edition_to_update_id: @content_block_edition.id
```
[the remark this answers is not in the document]
Harriethw: ah yeah good shout

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
Harriethw: > We could try to move the steps to a helper. That will be available to controllers, views and can be included into classes like the form objects. I'm not sure about view components.

Shall I leave further refactoring - e.g. to a helper - to when we introduce a third step, to keep this PR smaller and get it over the line?

> We might be able to assign it to a shorter constant in our engine initialiser?

will take a look at this!

Harriethw: > We might be able to assign it to a shorter constant in our engine initialiser?

Tried doing this (also suggested in the engines docs, [to pub configuration in the module](https://guides.rubyonrails.org/engines.html)) but with no success as still didn't seem to recognise it in other files :( 
I will leave as is - I think referencing with long module name is still better than having random strings floating about.

PR #9373, opened 2024-08-15
Remove edition roles when edition has been deleted.

kashifatcha:
Since worldwide organisations became an edition deleting them has tranistioned them to deleted state instead of deleting them from the database. This means that the destroy dependancy option on the active record edition roles association no longer worked.


minhngocd (approved):
Looks good ! was just wondering if the commit messages in the first two commits could be structured a bit clearer about what the issue is and what the fix was? i think it took a while for me to catch onto what was going on. (e.g. why the type update)

PR #9378, opened 2024-08-20
Content modelling/417 get embedded content via a new endpoint, fixing live

tahb:
~~**Requires https://github.com/alphagov/publishing-api/pull/2846 to be merged for this to work.**~~ Merged

## Changes in this PR

We introduce a new way of getting embedded content to fix the timeouts we're seeing in Integration. More details on that can be found [on the Publishing API change](https://github.com/alphagov/publishing-api/pull/2846).

* We aren't using a GDS adapter method yet, we'll look to add that once this fixes the problem on Integration. As we aren't in production we can take this shortcut
* We lose the pagination support offered by the endpoint we're switching away from. Again we're not in production so the lack of pagination is OK. Our priority here is to get the feature working again first.
* A benefit of the new approach is that we're now able to present an organisation's title even when that organisation doesn't exist in Whitehall. We're able to present the name stored in one the Publishing API Edition. We still aren't able to link in this scenario as we don't know what service to redirect people towards

## Before

image
image(1)

## After

Screenshot 2024-08-20 at 14-10-16 Manage an Email address - GOV UK Whitehall Publisher
Screenshot 2024-08-20 at 15-59-08 Where the change will appear - GOV UK Whitehall Publisher


lib/engines/content_object_store/app/controllers/content_object_store/content_block/documents_controller.rb
```diff
@@ -7,12 +7,10 @@ def show
     @content_block_document = ContentObjectStore::ContentBlock::Document.find(params[:id])
     @content_block_versions = @content_block_document.versions
 
-    linked_item_service = ContentObjectStore::GetLinkedContentItems.new(
+    embedded_content = ContentObjectStore::GetEmbeddedContentItems.new(
```
Harriethw, line 10: should this actually be GetHostContentItems? we're returning the documents that host the embedded content if I've got it right?

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/embedded_editions_table_component.html.erb
```diff
@@ -0,0 +1,20 @@
+
+<!-- TODO: The prototype wants to show a counter beneath the title but
+before the table, the existing component doesn't support that so I've left
+it out for now-->
+<%= render "govuk_publishing_components/components/table", {
+  caption:,
```
Harriethw: similar comment to the one above - is this a HostEditionsTable? 🤔

lib/engines/content_object_store/app/controllers/content_object_store/content_block/documents_controller.rb
tahb: Woah yes this needs some work, I've glossed over the naming after looking at them for so long.

We certainly get back `host_content` and the objects and table etc should be named accordingly - shall fix. 

I'm less sure what we do about this service. `GetHostContentItems` conveys what's returned but doesn't speak to how it works. How about `GetHostContentItemsByTarget`?

Harriethw: > We could add a method to help us: GetContentItems.by_embedded_document(doc)?

yep I think that works 👍

PR #9385, opened 2024-08-23
Fix error caused by some edition types not implementing organisations association

ryanb-gds:
This fixes the following Sentry error: https://govuk.sentry.io/issues/5742988523/events/da44c233e7ee47d3b7b30279d7ca1422/

The draft updater checks that the user is not about to limit their own access to a document.

However, not all documents can be access limited, or access is controlled by other means. Corporate information pages are an example of this.

This commit ensures that the updater only validates that the user is not about to limit their own access in contexts where the document is tagged to one or more organisations.

It's pretty ugly. I did initially try encapsulating the validation logic within the access limiting concern that is included in the edition model, rather than in private methods on the updater. However, I didn't want to couple the access limiting concern to the organisations concern, as I think it's possible to use them both independently. I therefore decided on the edition updater for now, but open to revisiting that decision if the reviewer feels strongly about it. This is another good example of having all document types inherit from the edition model is inhibiting the speed of change in Whitehall

Trello: https://trello.com/c/lZMXGcBv

### DIFF
--- a/app/services/draft_edition_updater.rb
+++ b/app/services/draft_edition_updater.rb
@@ -10,7 +10,7 @@ class DraftEditionUpdater < EditionService
   def failure_reason
     if !edition.pre_publication?
       "A #{edition.state} edition may not be updated."
-    elsif edition.access_limited? && @options[:current_user].present? && edition.edition_organisations.map(&:organisation_id).exclude?(@options[:current_user].organisation.id)
+    elsif should_check_current_user_will_retain_access? && access_limit_excludes_current_user?
       "Access can only be limited by users belonging to an organisation tagged to the document"
     elsif !edition.valid?
       "This edition is invalid: #{edition.errors.full_messages.to_sentence}"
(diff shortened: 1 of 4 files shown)


app/services/draft_edition_updater.rb
```diff
@@ -10,7 +10,7 @@ def perform!
   def failure_reason
     if !edition.pre_publication?
       "A #{edition.state} edition may not be updated."
-    elsif edition.access_limited? && @options[:current_user].present? && edition.edition_organisations.map(&:organisation_id).exclude?(@options[:current_user].organisation.id)
+    elsif validates_access_limit_does_not_exclude_current_user? && access_limit_excludes_current_user?
```
minhngocd, line 13: I'm a little confused reading this if statement - is this more like - "user is able to access limit" and "user does not belong to document organisations"?

ryanb-gds: It's not quite that, that but I'll have a think to see if I can come up with some better names because the current ones aren't very clear.

ryanb-gds: Let me know if you think the renamed methods are more expressive @minhngocd 🤞

PR #9387, opened 2024-08-27
Content modelling/410 schedule publishing

Harriethw:
**Note: to make this work you will need `Sidekiq Admin` permissions - locally, this can be added to the Test User in `db/seeds.rb`, on integration this can be added to each user through SignOn.**

We want to allow users to schedule the publication of Content Blocks. This is the first attempt at introducing that in the code. As with our other features, we follow Whitehall patterns where possible/where it makes sense. This has included :
* having any actions related to the `workflow` of an Edition in a separate controller
  * @tahb and I discussed renaming the `workflow_controller` to `editions_workflow_controller` to make this more clear, will do in a future PR
* Using a Worker class to put jobs onto a Sidekiq queue.
  * our Worker is almost identical, but calls a different Service to schedule a Content Block Edition. 
  * while Whitehall have a separate `ScheduledEditionPublisher` to publish scheduled Editions, as we don't have any logic to check for scheduled editions, I've just re-used our `PublishEditionService`.
* Using a similar workflow state on the Edition, and a field called `scheduled_publication` to store the datetime in the db 
* Using the same date/time fields component  and error messages in the view.

Screenshot 2024-08-27 at 12 34 32

Screenshot 2024-08-27 at 17 29 13

Screenshot 2024-08-27 at 17 29 28

Sidekiq jobs can be viewed at http://whitehall-admin.dev.gov.uk/sidekiq/queues 

## steps for manual testing
* create a content block
* go to 'change' the content block details
* schedule the publication for the next minute or so
* observe job at http://whitehall-admin.dev.gov.uk/sidekiq/queues 
* once published, put content id for the block (found in embed code) and check published on Publishing API http://publishing-api.dev.gov.uk/v2/content/{content-id}

## Todo in future PRs

* if there is another scheduled edition, scheduling should not be allowed or we should supersede the existing scheduled edition
* version history/ audit trail
  * the scheduling action should appear on the Document Timeline as `scheduled` by User
  * the actual publication by Sidekiq should appear as `published` (by who?)
* Interaction design for showing the state of the Block and the publication date. 
* Interaction design to cancel or supersede a scheduled Block
* Understanding how this relates to Documents that have embedded the Block - e.g. superseding logic, notifications etc.  

~~⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️~~

~~Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.~~

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_08_22_112802) do
+ActiveRecord::Schema[7.1].define(version: 2024_08_29_130419) do
   create_table "assets", charset: "utf8mb3", force: :cascade do |t|
     t.string "asset_manager_id", null: false
     t.string "variant", null: false
@@ -227,6 +227,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_08_22_112802) do
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
     t.string "state", default: "draft", null: false
+    t.datetime "scheduled_publication", precision: nil
     t.index ["document_id"], name: "index_content_block_editions_on_document_id"
   end
 
--- a/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
+++ b/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
@@ -14,6 +14,8 @@ private
       organisation_item,
       creator_item,
       embed_code_item,
+      state_item,
+      scheduled_item,
     ]
   end
 
@@ -54,6 +56,20 @@ private
     {
       field: "Creator",
       value: content_block_document.latest_edition.creator.name,
+    }
+  end
+
+  def state_item
+    {
+      field: "State",
+      value: content_block_document.latest_edition.state,
+    }
+  end
+
+  def scheduled_item
+    {
+      field: "Scheduled for publication at",
+      value: content_block_document.latest_edition.scheduled_publication&.strftime("%e %B %Y at %I:%M%P"),
     }
   end
 
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
@@ -90,31 +90,7 @@ private
     params.require(:content_object_store_content_block_edition)
   end
 
-  def edition_params
-    params.require(:content_block_edition)
-      .permit(
-        :organisation_id,
-        :creator,
-        "scheduled_publication(1i)",
-        "scheduled_publication(2i)",
-        "scheduled_publication(3i)",
-        "scheduled_publication(4i)",
-        "scheduled_publication(5i)",
-        document_attributes: %w[title block_type],
-        details: @schema.fields,
-      )
-      .merge!(creator: current_user)
-  end
-
   def block_type_param
     params.require(:content_block_edition).require("document_attributes").require(:block_type)
-  end
-
-  def scheduled_publication_params
-    params.require(:scheduled_at).permit("scheduled_publication(1i)",
-                                         "scheduled_publication(2i)",
-                                         "scheduled_publication(3i)",
-                                         "scheduled_publication(4i)",
-                                         "scheduled_publication(5i)")
   end
 end
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
@@ -30,6 +30,7 @@ class ContentObjectStore::ContentBlock::EditionsController < ContentObjectStore:
   EDIT_FORM_STEPS = {
     edit_block: "edit_block",
     review_links: "review_links",
+    schedule_publishing: "schedule_publishing",
   }.freeze
 
   def edit
@@ -42,6 +43,8 @@ class ContentObjectStore::ContentBlock::EditionsController < ContentObjectStore:
       edit_block
     when EDIT_FORM_STEPS[:review_links]
       review_links
+    when EDIT_FORM_STEPS[:schedule_publishing]
+      schedule_publishing
     end
   end
 
@@ -55,7 +58,7 @@ class ContentObjectStore::ContentBlock::EditionsController < ContentObjectStore:
     @content_block_document = @content_block_edition.document
     @edition_params = edition_params
 
-    new_edition = ContentObjectStore::ContentBlock::Edition.new(@edition_params)
+    new_edition = ContentObjectStore::ContentBlock::Edition.new(edition_params)
     new_edition.document.id = @content_block_document.id
 
     if new_edition.valid?
@@ -74,17 +77,34 @@ class ContentObjectStore::ContentBlock::EditionsController < ContentObjectStore:
     end
   end
 
+  def schedule_publishing
+    @content_block_document = @content_block_edition.document
+    @edition_params = edition_params
+
+    render :schedule_publishing
+  end
+
   def update
     content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
     @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(content_block_edition.document.block_type)
 
-    new_content_block_edition = ContentObjectStore::UpdateEditionService.new(
-      @schema,
-      content_block_edition,
-    ).call(edition_params)
+    new_edition = edition_params
+    if params[:schedule_publishing] == "schedule"
+      new_edition = edition_params.merge!(scheduled_publication_params)
+      new_content_block_edition = ContentObjectStore::ScheduleEditionService.new(
+        content_block_edition,
+      ).call(new_edition)
+      flash_text = "#{@schema.name} scheduled successfully"
+    else
+      new_content_block_edition = ContentObjectStore::UpdateEditionService.new(
+        @schema,
+        content_block_edition,
+      ).call(new_edition)
+      flash_text = "#{@schema.name} changed and published successfully"
+    end
 
     redirect_to content_object_store.content_object_store_content_block_document_path(new_content_block_edition.document),
-                flash: { notice: "#{@schema.name} changed and published successfully" }
+                flash: { notice: flash_text }
   rescue ActiveRecord::RecordInvalid => e
     @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
       content_block_edition: e.record, schema: @schema,
@@ -102,11 +122,29 @@ private
 
   def edition_params
     params.require(:content_block_edition)
-      .permit(:organisation_id, :creator, document_attributes: %w[title block_type], details: @schema.fields)
-      .merge(creator: current_user)
+      .permit(
+        :organisation_id,
+        :creator,
+        "scheduled_publication(1i)",
+        "scheduled_publication(2i)",
+        "scheduled_publication(3i)",
+        "scheduled_publication(4i)",
+        "scheduled_publication(5i)",
+        document_attributes: %w[title block_type],
+        details: @schema.fields,
+      )
+      .merge!(creator: current_user)
   end
 
   def block_type_param
     params.require(:content_block_edition).require("document_attributes").require(:block_type)
+  end
+
+  def scheduled_publication_params
+    params.require(:scheduled_at).permit("scheduled_publication(1i)",
+                                         "scheduled_publication(2i)",
+                                         "scheduled_publication(3i)",
+                                         "scheduled_publication(4i)",
+                                         "scheduled_publication(5i)")
   end
 end
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
@@ -84,36 +84,6 @@ class ContentObjectStore::ContentBlock::EditionsController < ContentObjectStore:
     render :schedule_publishing
   end
 
-  def update
-    content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
-    @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(content_block_edition.document.block_type)
-
-    new_edition = edition_params
-    if params[:schedule_publishing] == "schedule"
-      new_edition = edition_params.merge!(scheduled_publication_params)
-      new_content_block_edition = ContentObjectStore::ScheduleEditionService.new(
-        content_block_edition,
-      ).call(new_edition)
-      flash_text = "#{@schema.name} scheduled successfully"
-    else
-      new_content_block_edition = ContentObjectStore::UpdateEditionService.new(
-        @schema,
-        content_block_edition,
-      ).call(new_edition)
-      flash_text = "#{@schema.name} changed and published successfully"
-    end
-
-    redirect_to content_object_store.content_object_store_content_block_document_path(new_content_block_edition.document),
-                flash: { notice: flash_text }
-  rescue ActiveRecord::RecordInvalid => e
-    @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
-      content_block_edition: e.record, schema: @schema,
-      edition_to_update_id: content_block_edition.id
-    )
-
-    render :edit
-  end
-
 private
 
   def root_params
--- a/lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
+++ b/lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
@@ -12,4 +12,59 @@ class ContentObjectStore::ContentBlock::WorkflowController < ContentObjectStore:
     #   redirect_to admin_edition_path(@edition), alert: edition_publisher.failure_reason
     # end
   end
+
+  def update
+    content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
+    @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(content_block_edition.document.block_type)
+
+    new_edition = edition_params
+    if params[:schedule_publishing] == "schedule"
+      new_edition = edition_params.merge!(scheduled_publication_params)
+      new_content_block_edition = ContentObjectStore::ScheduleEditionService.new(
+        content_block_edition,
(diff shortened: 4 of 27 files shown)


lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
```diff
@@ -14,6 +14,8 @@ def items
       organisation_item,
       creator_item,
       embed_code_item,
+      state_item,
```
Harriethw, line 17: TODO: unit tests for this component

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
```diff
@@ -74,17 +77,37 @@ def review_links
...
     content_block_edition = ContentObjectStore::ContentBlock::Edition.find(params[:id])
     @schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(content_block_edition.document.block_type)
 
-    new_content_block_edition = ContentObjectStore::UpdateEditionService.new(
-      @schema,
-      content_block_edition,
-    ).call(edition_params)
+    new_edition = edition_params
+    if params[:schedule_publishing] == "schedule"
+      new_edition = edition_params.merge!(scheduled_publication_params)
```
Harriethw: this gets moved to `workflow_controller` in subsequent commit, and will need to be refactored to handle errors on the scheduling form

lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
```diff
@@ -12,4 +12,59 @@ def publish
...
+    @form = ContentObjectStore::ContentBlock::EditionForm::Update.new(
+      content_block_edition: e.record, schema: @schema,
+      edition_to_update_id: content_block_edition.id
+    )
+
+    render :edit
+  end
+  private
+
+  def scheduled_publication_params
```
Harriethw, line 46: these methods lifted from editions controller, could be shared?

app/workers/content_block_publishing_worker.rb
```diff
@@ -0,0 +1,62 @@
+require "sidekiq/api"
+
+# ContentBlockPublishingWorker is a job that is scheduled by the `ScheduleEditionService`.
+# It is never executed immediately but uses the Sidekiq delay mechanism to
+# execute at the time the edition should be published.
+class ContentBlockPublishingWorker < WorkerBase
```
Harriethw, line 6: perhaps this should be called `ContentBlockSchedulerWorker` or have `schedule` in it? 🤔

app/workers/content_block_publishing_worker.rb
```diff
@@ -0,0 +1,62 @@
+require "sidekiq/api"
```
Harriethw, line 1: still need more investigation into why this wouldn't work in the object store Engine, as we want to avoid having any code in the main repo

lib/engines/content_object_store/features/schedule_object.feature
```diff
@@ -0,0 +1,25 @@
...
+    When I am updating a content block
+    Then I am asked when I want to publish the change
+    When I choose to schedule the change
+    And I enter a date 7 days in the future
+    And I accept and publish
+    Then the edition should have been scheduled successfully
+    And I should be taken back to the document page
+    And I should see the scheduled date on the object
+
+#  Scenario: A scheduled content object is published
```
Harriethw, line 22: would also like to investigate whether it's possible to do this test - setup a scheduling job in the worker, fast forward the time and see that it gets published. But not sure if possible....

app/workers/content_block_publishing_worker.rb
Harriethw: ok it looked like restarting the sidekiq docker container fixed the engine issue - will move the Worker into our object store engine 👍

lib/engines/content_object_store/features/schedule_object.feature
tahb: IIRC we can include [sidekiq/testing](https://github.com/sidekiq/sidekiq/wiki/Testing#testing-delay-and-delay_for-methods) which enables:

1. We can check that our code correctly puts the right job on the right delayed queue with the right time
2. We can then run with inline mode which will process all jobs straight away regardless of what our code says

Harriethw: Nice, I'm using the inline method in a feature test to check it gets published, and have tested that we are pushing the right data to the right queue in the integration test `worker_test.rb`

db/schema.rb
```diff
@@ -226,8 +226,11 @@
     t.bigint "document_id", null: false
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
+    t.bigint "user_id"
```
tahb, line 229: The schema looks to have been changed with more than just `scheduled_publication`. Likely some cross over with a later commit

lib/engines/content_object_store/test/unit/workers/schedule_publishing_worker_test.rb
```diff
@@ -0,0 +1,148 @@
...
+    publish_service_mock.verify
+  end
+
+  test "#perform returns without consequence if the edition is already published" do
+    document = create(:content_block_document, :email_address)
+    edition = create(:content_block_edition, document:, state: "published")
+
+    ContentObjectStore::ContentBlock::Schema.expects(:find_by_block_type).never
+    ContentObjectStore::PublishEditionService.expects(:new).never
+    ContentObjectStore::PublishEditionService.any_instance.expects(:call).never
```
tahb, line 47: Great to have this check!

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
```diff
@@ -30,6 +30,17 @@ def publish_with_rollback(schema:, title:, details:)
       end
     end
 
+    def schedule_with_rollback
+      raise ArgumentError, "Local database changes not given" unless block_given?
+
+      ActiveRecord::Base.transaction do
+        content_block_edition = yield
+
+        content_block_edition.public_send(:schedule!)
```
tahb, line 39: I can't find where `.schedule!` comes from? Where did it come from and what does it do?

tahb (commented):
I've read the first half of the commits but have run out of time so sending this across now in the hopes it's still useful.

lib/engines/content_object_store/app/services/content_object_store/schedule_edition_service.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      new_content_block_edition = ContentObjectStore::ContentBlock::Edition.new(edition_params)
+      new_content_block_edition.document_id = @original_content_block_edition.document.id
+      new_content_block_edition.save!
+      new_content_block_edition
+    end
+
+    def update_content_block_document(new_content_block_edition:, update_document_params:)
+      update_document_params[:latest_edition_id] = new_content_block_edition.id
```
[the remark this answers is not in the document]
Harriethw: that's a good question - I think we'd want to update `latest_edition_id` as it is now the latest edition even if it's not published, but not `live_edition_id`?

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
Harriethw: ah yep this is copy/paste cruft from Whitehall, not sure why they do `public_send`... will re-order commits and just call `.schedule!`

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
```diff
@@ -30,6 +30,17 @@ def publish_with_rollback(schema:, title:, details:)
...
+        ContentObjectStore::SchedulePublishingWorker.queue(content_block_edition)
```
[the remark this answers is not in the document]
Harriethw: ah this was something I wanted to bring up - I'm not sure what we would do if the job fails, and I'm not sure what Whitehall do beyond looking at the Sidekiq dashboard manually. Would this be a subsequent ticket to investigate?

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
```diff
@@ -49,6 +60,12 @@ def publish_publishing_api_edition(content_id:)
       raise PublishingFailureError, "Could not publish #{content_id} because: #{e.message}"
     end
 
+    def schedule_publishing_api_edition(content_id:)
```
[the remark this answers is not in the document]
Harriethw: ah yep I don't think we need this schedule method and can just use the existing one

db/schema.rb
Harriethw: resolved, thank u!

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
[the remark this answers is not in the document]
Harriethw: we chatted about this and TLDR:
* if the function to publish a scheduled Edition fails, the sidekiq admin dashboard and Grafana will catch that and retry the job, will be shown in  https://docs.publishing.service.gov.uk/manual/sidekiq.html#sidekiq-on-gov-uk 
* I don't know how we are alerted to errors on Whitehall. We think Sentry is used, but not sure where alerts go, and it would be good to have Object Store specific ones in our dev channel if possible. Created a ticket here https://trello.com/c/K19zXzAU/469-tech-task-sentry-for-the-object-store

lib/engines/content_object_store/app/services/content_object_store/schedule_edition_service.rb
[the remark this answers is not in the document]
Harriethw: there was some complexity/refactoring around setting these values on the Document so I split into a new commit. 
https://github.com/alphagov/whitehall/pull/9387/commits/1e62e6df6ebae4d3946789c12242b24c32aa2aae

lib/engines/content_object_store/app/services/content_object_store/schedule_edition_service.rb
```diff
@@ -0,0 +1,70 @@
...
+          update_document_params:,
+        )
+
+        @new_content_block_edition
+      end
+
+      @new_content_block_edition
+    end
+
+  private
```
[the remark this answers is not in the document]
Harriethw: yep they are the same, have put in `Publishable` in latest commit 👍

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
```diff
@@ -14,6 +14,8 @@ def items
...
+      scheduled_item,
```
pezholio, line 18: I think it'd be good to only show this row if a block is still scheduled

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
```diff
@@ -57,6 +59,20 @@ def creator_item
...
+    {
+      field: "State",
+      value: content_block_document.latest_edition.state,
+    }
+  end
+
+  def scheduled_item
+    {
+      field: "Scheduled for publication at",
+      value: content_block_document.latest_edition.scheduled_publication&.strftime("%d %B %Y at %I:%M%P"),
```
pezholio, line 72: We should use `%e` here to show the date without as leading zero, as per the [GOV.UK guidance on dates](https://www.gov.uk/guidance/style-guide/a-to-z#dates), i.e:

```suggestion
      value: content_block_document.latest_edition.scheduled_publication&.strftime("%e %B %Y at %I:%M%P"),
```

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/schedule_publishing.html.erb
```diff
@@ -4,15 +4,26 @@
...
 <% end %>
 
+<%
+  year_param = @edition_params["scheduled_publication(1i)"]
+  month_param = @edition_params["scheduled_publication(2i)"]
+  day_param = @edition_params["scheduled_publication(3i)"]
+  hour_param = @edition_params["scheduled_publication(4i)"]
+  minute_param =  @edition_params["scheduled_publication(5i)"]
+  is_scheduled_param = params["schedule_publishing"]
+%>
```
pezholio, line 21: Is there a more elegant way to do this? Too much logic in the view makes me a bit squiffy

pezholio (commented):
All works great testing locally, just a couple of things I noticed from a quick(ish) drive by review

lib/engines/content_object_store/app/models/concerns/content_object_store/workflow.rb
```diff
@@ -1,21 +1,30 @@
 module ContentObjectStore::Workflow
   extend ActiveSupport::Concern
+  # include DateValidation
```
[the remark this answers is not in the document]
Harriethw: ah yeah good spot

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
Harriethw: Tom did bring this up as well, but my argument was why spend time writing logic and tests for a UI that we haven't finalised yet? I wanted it to be as crap as possible to ensure we return to it 😆 but aware that probably isn't great practice so happy to change!

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
Harriethw: good spot - I will do this in a separate commit as we are using this format elsewhere so will update everywhere (maybe needs our own entry in `format` as it looks like Whitehall don't have the same rendering? I got this from the figma designs and think it is more up to date in terms of guidelines...)

lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
```diff
@@ -312,3 +319,61 @@ def should_show_edit_form_for_email_address_content_block(document_title, email_
...
+
+When("I enter a date 7 days in the future") do
+  @future_date = 7.days.since(Time.zone.now).to_date
+  fill_in_date_and_time_field(@future_date)
+end
+
+Then("the edition should have been scheduled successfully") do
+  @schema = @schemas[@content_block.document.block_type]
+  assert_text "#{@schema.name} scheduled successfully"
+  # assert_text @future_date.strftime("%d %B %Y at %I:%M%P")
```
tahb: Shall we comment this in or does it come in in a later commit?

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/schedule_publishing.html.erb
Harriethw: thought it would be ok as we aren't actually doing any logic, just destructuring some params to make it more clear what they are, because of ruby date magic... I got the idea from Whitehall's scheduling component, which does a mix of this and inline digging, which seems less desirable https://github.com/alphagov/whitehall/blob/096d3b551f033a7bcb732dbe905914f889f33067/app/views/admin/editions/_scheduled_publication_fields.html.erb 

what would you recommend, constructing an object in the controller? or just using `@edition_params["scheduled_publication(1i)"]` inline?

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
tahb: Yeah we could go either way with this but given we are skipping ahead with the build I like what we did with embed code: we just popped it on the page without overthinking it and our message to the team was this is here for us, help us make it better.

lib/engines/content_object_store/app/validators/content_object_store/scheduled_publication_validator.rb
```diff
@@ -0,0 +1,14 @@
+class ContentObjectStore::ScheduledPublicationValidator < ActiveModel::Validator
+  attr_reader :edition
+
+  def validate(edition)
+    @edition = edition
```
tahb, line 5: Do we need this instance variable?

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/schedule_publishing.html.erb
tahb: It could be an idea to use these params to create a `DateTime` object and sent that to the view instead. The view would be simplified down slightly as it'll just send methods to a single object, rather than knowing how to compose all the different parts. Though it's hard to see how this rails date splitting interface changing.

I don't think Rails gives us a way to nicely create that object without having to unpick each part :( So we'd just have to move this up to the controller and send a `@published_time` back down. 

I don't have a strong opinion here given the params aren't going to change. I might still spend 10 minutes to see how the idea felt.

tahb (approved):
Excellent work 🙌 

Couple of observations in case they're helpful. In hindsight I think this was probably too big for one PR as I found reviewing trickier than usual. Overall though this is definitely good enough to go in 👍

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
pezholio: Yeah, that's fair

lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
pezholio: It might be worth adding the format to en.yml along with the other date formats (https://github.com/alphagov/whitehall/blob/3abce9256c2ca443ec7b1189d4a6c166ec3f2c2d/config/locales/en.yml#L479) and then using i18n to format the time ie:

```ruby
value: l(content_block_document.latest_edition.scheduled_publication, format: :human_readable)
```

lib/engines/content_object_store/app/validators/content_object_store/scheduled_publication_validator.rb
Harriethw: nope, will remove

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/schedule_publishing.html.erb
Harriethw: I'm going to merge as is as this is now blocking work and already such a mega PR, and there's precedent in the Whitehall code, if it is a big issue happy to pick up in a refactoring PR

PR #9388, opened 2024-08-27
Allow overriding government attempt 3

ryanb-gds:
Trello: https://trello.com/c/II8jgMR7

This PR puts the basic mechanisms in place for enabling GDS editors and admins to override the government. However, it does not include a user interface, which will be hidden behind a feature flag initially so that the content team can test it out.

## Implementation Notes

The `government_id` column that we're adding to the editions table is indexed, because I foresee a need when we come to [facilitating changes to historic governments](https://trello.com/c/VwTpDOEa) to be able to query for all of the editions associated with a government, in order to make sure that any overrides aren't deleted. I wanted to use `on_delete: restrict` for this to prevent governments from being deleted but realised that I wouldn't be able to because we won't be able to change the overrides for superseded editions. So we'll have to do that in the application layer instead :cry:

I originally planned to achieve this feature by replacing the `political` attribute on the edition with the `government_id` column (hence this being attempt 3), but that would make changing governments very difficult indeed because of the number of documents that might need to be updated to make sure the overrides were maintained. I wasn't able to envision an automated process to achieve that and the burden on GDS admins would have been excessive. I've therefore decided to make `government_id` be just an override for the government selection for an edition.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_07_31_143537) do
+ActiveRecord::Schema[7.1].define(version: 2024_08_22_112802) do
   create_table "assets", charset: "utf8mb3", force: :cascade do |t|
     t.string "asset_manager_id", null: false
     t.string "variant", null: false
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -447,10 +447,12 @@
     t.string "logo_formatted_name"
     t.string "analytics_identifier"
     t.boolean "visual_editor"
+    t.integer "government_id"
```
minhngocd, line 450: just wondering if it should be called government_override in pattern with `person_override` to make it explicit that it's an override and not a usual thing?

test/functional/admin/generic_editions_controller_tests/political_document_test.rb
```diff
@@ -8,6 +8,17 @@ class Admin::GenericEditionsController::PolticalDocumentsTest < ActionController
...
+  test "can mark a document as political" do
+    create(:current_government)
+    published_edition = create(:published_news_article)
+    new_draft = create(:news_article, document: published_edition.document)
+
+    put :update, params: { id: new_draft, edition: { political: true } }
+
+    assert_redirected_to admin_edition_path(new_draft)
+    assert new_draft.reload.political?
+  end
```
minhngocd, line 20: 😮  nice one!! thanks for adding

minhngocd (approved):

db/schema.rb
ryanb-gds: I think the person override is that feature for speeches where you can put arbitrary text in when the speech was given by somebody without a profile on GOV.UK.

In our case the government does exist on GOV.UK, it's just not the default one. So I feel that the use case isn't identical, and I also wonder if not making it clear that the column contains an id might lead to confusion.

PR #9394, opened 2024-08-28
Get content that embeds a target via GDS API Adapters

tahb:
## Changes in this PR

* We switch our a request to the Publishing API. Replacing a Net::HTTP request with a conventional GDS API Adapter call
* Change in our error handling. Instead of defaulting to an empty list of results in the case of an error we now throw. When users are managing content it's important they can see the full reach of their reusable content. To see an empty list could be dangerously misleading, implying there was in fact no reuse.

Naming continues to be a battle. Coining a concise term that captures the concept and the relationships of reusable content. We have 'reusable' editions which are _embedded_ and linked to by ‘normal’ editions. When we call this endpoint we want to know where a piece of content has been embedded but our results should be references to the normal editions. We're using "content which embeds our target content" which is accurate but a little awkward.

### DIFF
--- a/lib/engines/content_object_store/features/support/stubs.rb
+++ b/lib/engines/content_object_store/features/support/stubs.rb
@@ -1,12 +1,3 @@
 Before do
-  stub_dependent_content(content_id: anything, total: 0, results: [])
-end
-
-def stub_dependent_content(results:, content_id: anything, total: 0)
-  stub_request(:get, %r{\A#{Plek.find('publishing-api')}/v2/content/[0-9a-fA-F-]{36}/embedded})
-    .to_return(body: {
-      "content_id" => content_id,
-      "total" => total,
-      "results" => results,
-    }.to_json)
+  stub_publishing_api_has_embedded_content(content_id: anything, total: 0, results: [])
 end
--- a/lib/engines/content_object_store/test/unit/app/services/get_host_content_items_test.rb
+++ b/lib/engines/content_object_store/test/unit/app/services/get_host_content_items_test.rb
@@ -6,7 +6,6 @@ class ContentObjectStore::GetHostContentItemsTest < ActiveSupport::TestCase
   let(:described_class) { ContentObjectStore::GetHostContentItems }
 
   let(:target_content_id) { SecureRandom.uuid }
-  let(:expected_url) { "https://publishing-api.test.gov.uk/v2/content/#{target_content_id}/embedded" }
 
   let(:response_body) do
     {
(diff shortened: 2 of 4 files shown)


lib/engines/content_object_store/test/unit/app/services/get_host_content_items_test.rb
```diff
@@ -60,31 +66,23 @@ class ContentObjectStore::GetHostContentItemsTest < ActiveSupport::TestCase
...
-
-      plek_domain = "https://foo.dev.gov.uk"
-      Plek.stubs(:find).with("publishing-api").returns(plek_domain)
-
-      stub_request(:get, %r{\A#{plek_domain}/v2/content/#{target_content_id}/embedded})
-        .to_return(body: response_body.to_json, status: 200)
-
-      described_class.by_embedded_document(
-        content_block_document: document,
+    it "returns an error if the content that embeds the target can't be loaded" do
```
Harriethw, line 69: so if this request fails, are we not able to see the relevant embedded Document at all, whereas before we would have failed silently and shown an empty list?

lib/engines/content_object_store/features/support/stubs.rb
```diff
@@ -1,12 +1,3 @@
...
-end
-
-def stub_dependent_content(results:, content_id: anything, total: 0)
-  stub_request(:get, %r{\A#{Plek.find('publishing-api')}/v2/content/[0-9a-fA-F-]{36}/embedded})
-    .to_return(body: {
-      "content_id" => content_id,
-      "total" => total,
-      "results" => results,
-    }.to_json)
+  stub_publishing_api_has_embedded_content(content_id: anything, total: 0, results: [])
```
Harriethw, line 2: so  `stub_publishing_api_has_embedded_content` is defined in the adapters repo?

tahb: Yeah that's right. https://github.com/alphagov/gds-api-adapters/pull/1283/commits/b206bf80c7a53f1148a6a9cb77e54ee3a101bd9c. I think it's because that way any consumer has easy access to the same stubs, rather than having to redeclare them.

Harriethw (approved):
LGTM!

lib/engines/content_object_store/test/unit/app/services/get_host_content_items_test.rb
tahb: Yup, if it fails now the host document page won't load at all.

The previous behaviour was me skipping the problem of error handling as I knew we'd be coming back to take out the net http request. 

I think this fail safe feels a reasonable starting place. In future we could change things so we still render the page but show an error message where the table was.

pezholio (approved):
LGTM - might be worth bumping the gds-api-adapters version as part of this PR too, rather than waiting for Dependabot to wake up

tahb:
Dependabot beat me to it :) https://github.com/alphagov/whitehall/pull/9399

PR #9401, opened 2024-09-02
Content modelling/467 timeline

Harriethw:
Previous timeline

Screenshot 2024-09-03 at 11 26 16

New timeline

Screenshot 2024-09-03 at 15 45 26

This creates an audit trail for scheduling/updating an Edition, and adds the events to the user-facing timeline of a Document.

Previously we only had events to track when an Edition was created - because our
Content Blocks are published as soon as they are created, this was sufficient, as an Edition would never be 'updated', only superseded by a new Edition.

Now we have introduced Scheduling, an Edition can be updated. It will be created and set in the `scheduled` state, and then `published` by the scheduling worker.

For this reason we introduce the `record_update` method, and create an 'updated' event with the
state - this will also be handy in the future should we ever introduce new states (e.g.withdrawn, draft etc.)

These versions provide an Edition-level database log of events for us, but the user need on a
Document-level view is slightly different. With this new change, we will only show state/update changes to an Edition, as[ this should cover the need-to-know events for now.](https://gds.slack.com/archives/C062Z1FA2ES/p1725375605153839?thread_ts=1725360904.799849&cid=C062Z1FA2ES)

### DIFF
--- a/db/migrate/20240902102908_add_state_to_content_block_versions.rb
+++ b/db/migrate/20240902102908_add_state_to_content_block_versions.rb
@@ -1,0 +1,13 @@
+class AddStateToContentBlockVersions < ActiveRecord::Migration[7.1]
+  def up
+    change_table :content_block_versions, bulk: true do |t|
+      t.text "state"
+    end
+  end
+
+  def down
+    change_table :content_block_versions, bulk: true do |t|
+      t.remove :state
+    end
+  end
+end
--- a/lib/engines/content_object_store/app/workers/content_object_store/schedule_publishing_worker.rb
+++ b/lib/engines/content_object_store/app/workers/content_object_store/schedule_publishing_worker.rb
@@ -48,9 +48,11 @@ module ContentObjectStore
 
       schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(edition.document.block_type)
 
-      ContentObjectStore::PublishEditionService.new(
-        schema,
-      ).call(edition)
+      ContentObjectStore::HasAuditTrail.acting_as(publishing_robot) do
+        ContentObjectStore::PublishEditionService.new(
+          schema,
+        ).call(edition)
+      end
     rescue ContentObjectStore::Publishable::PublishingFailureError => e
       raise SchedulingFailure, e.message
     end
--- a/lib/engines/content_object_store/test/components/content_block/document/show/document_timeline_component_test.rb
+++ b/lib/engines/content_object_store/test/components/content_block/document/show/document_timeline_component_test.rb
@@ -3,44 +3,37 @@ require "test_helper"
 class ContentObjectStore::ContentBlock::Document::Show::DocumentTimelineComponentTest < ViewComponent::TestCase
   test "renders a timeline component with events in correct order" do
     @user = create(:user)
-    @content_block_edition_1 = create(
-      :content_block_edition,
-      document: create(
-        :content_block_document,
-        block_type: "email_address",
-      ),
-    )
-    @content_block_version_1 = create(
+    @version_1 = create(
       :content_block_version,
-      item: @content_block_edition_1,
+      event: "created",
       whodunnit: @user.id,
     )
-    @content_block_edition_2 = create(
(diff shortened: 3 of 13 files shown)


lib/engines/content_object_store/app/workers/content_object_store/schedule_publishing_worker.rb
```diff
@@ -48,9 +48,11 @@ def perform(edition_id)
 
       schema = ContentObjectStore::ContentBlock::Schema.find_by_block_type(edition.document.block_type)
 
-      ContentObjectStore::PublishEditionService.new(
-        schema,
-      ).call(edition)
+      ContentObjectStore::HasAuditTrail.acting_as(publishing_robot) do
```
Harriethw, line 51: this is the "publishing robot" used by Whitehall, and called the "Scheduled Publishing Robot" - should we create our own User for this? Not entirely sure of the best way of doing that in Int/Prod...

lib/engines/content_object_store/test/components/content_block/document/show/document_timeline_component_test.rb
```diff
@@ -3,44 +3,41 @@
...
 
-    assert_selector ".timeline__item", count: 2
-    assert_equal page.all(".timeline__title")[0].text, "Email address changed"
-    assert_equal page.all(".timeline__byline")[0].text, "by #{@user.name}"
-    assert_equal page.all("time[datetime='#{@content_block_version_2.created_at.iso8601}']")[0].text,
-                 @content_block_version_2.created_at.strftime("%d %B %Y at %I:%M%P")
+    assert_selector ".timeline__item", count: 3
+    assert_equal "Email address scheduled", page.all(".timeline__title")[0].text
+    assert_equal "by #{@user.name}", page.all(".timeline__byline")[0].text
+    assert_equal @version_3.created_at.strftime("%d %B %Y at %I:%M%P"),
```
Harriethw, line 30: guess this datetime formatting can be updated in this PR https://github.com/alphagov/whitehall/pull/9402   or I will update it here if that gets merged before

lib/engines/content_object_store/app/workers/content_object_store/schedule_publishing_worker.rb
pezholio: I think let's keep it the same, as, to a user, publishing is the same regardless of where in Whitehall it's done.

lib/engines/content_object_store/test/components/content_block/document/show/document_timeline_component_test.rb
pezholio: Yeah, I'm happy to fix in my PR 👍

pezholio (approved):
This looks good. I think doing the simplest thing possible in the first instance, allowing us to iterate is the right move 👍

lib/engines/content_object_store/test/unit/app/models/concerns/has_audit_trail_test.rb
```diff
@@ -20,19 +20,44 @@ class ContentObjectStore::HasAuditTrailTest < ActiveSupport::TestCase
...
+      second_version = create(
+        :content_block_version,
+        item: edition,
+      )
+      third_version = create(
+        :content_block_version,
+        item: edition,
+      )
+      assert_equal edition.versions.first, third_version
+      assert_equal edition.versions[1], second_version
```
tahb, line 106: I thin Ruby does let us do `[1,2].second => 2`, if you wanted to keep with the human readable 👍

db/migrate/20240902102908_add_state_to_content_block_versions.rb
```diff
@@ -0,0 +1,13 @@
+class AddStateToContentBlockVersions < ActiveRecord::Migration[7.1]
+  def up
+    change_table :content_block_versions, bulk: true do |t|
+      t.text "state"
```
tahb, line 4: Do we need any default value here?

tahb (commented):
I'm too late to the party but LGTM!

PR #9402, opened 2024-09-03
DRY up date formatting

pezholio:
This makes some tweaks to content modelling to ensure we're formatting dates using I18n, rather than handrolling with `strftime` each time

### DIFF
--- a/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/document_timeline_component.rb
+++ b/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/document_timeline_component.rb
@@ -27,9 +27,8 @@ private
   end
 
   def time_html(date_time)
-    formatted_timestamp = date_time.strftime("%d %B %Y at %I:%M%P")
     tag.time(
-      formatted_timestamp,
+      I18n.l(date_time, format: :long_ordinal),
       class: "date",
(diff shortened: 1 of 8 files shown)


lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/document_timeline_component.rb
```diff
@@ -27,9 +27,8 @@ def first_created_edition
   end
 
   def time_html(date_time)
-    formatted_timestamp = date_time.strftime("%d %B %Y at %I:%M%P")
```
tahb, line 30: I wonder if we ought to go the other way here and be consistent by using `long_ordinal`?

I tried to look [for guidance on time](https://www.gov.uk/guidance/style-guide/a-to-z#times) and [dates](https://www.gov.uk/guidance/style-guide/a-to-z#dates) (just in case) but there wasn't anything helpful. Did you find anything?

In addition, adding in `human_readable` as a new time format alongside `long_ordinal` does make it harder to understand which format to use to be consistent. Unless of course we are highly confident that this is in keeping with the style guide?

pezholio: I think you're right. Keeps it simple and consistent then

PR #9405, opened 2024-09-03
Add sentry tag for ContentObjectStore controllers

pezholio:
This adds a `before_action` to all controllers that inherit from the `ContentObjectStore::BaseController` which sets an `engine` tag so we can easily filter Sentry for all content object store related errors.

This won't (yet) work for anything done asynchronously, which is more of a challenge, but we can pick this up once we've confirmed that this works.


Harriethw (approved):
👍

PR #9406, opened 2024-09-03
Content modelling/514 dequeue old scheduled publications

tahb:
## Before

A user could make multiple changes to a reusable piece of content each with different scheduled_at times.
This was problematic as it meant that the changes of old superseded editions could have overwritten later editions.

## After

When we create a new edition (our user journey is called updating) we dequeue any jobs for that original edition. 
This means only the latest edition can have a schedule.
We follow the steer of Whitehall main with the use of `dequeue`

## Changes in this PR

We could make this change quite easily by adding it into the controller - first commit.

This behaviour felt too important to leave to the controller to orchestrate. What would happen if this service was reused somewhere else? The dequeuing wouldn't happen unless a human read through the consuming controller carefully.

That led me onto a refactoring of the service objects. Making the `ScheduleEdition` service redundant by relying more on the `UpdateEdition` service instead. Given the similarity of those services I think this worked quite nicely.

In addition we add a `ResultMondad` pattern and `ChangeDispatcher` into the mix. This allows our controller to tell the service what type of change is needed and for the related success message to be returned.


lib/engines/content_object_store/app/models/content_object_store/result_monad.rb
```diff
@@ -0,0 +1,11 @@
+module ContentObjectStore
+  ResultMonad = Struct.new(:success, :message, :object) do
+    def success?
+      success == true
+    end
+
+    def failure?
```
tahb, line 7: We don't actually end up using this one but I have found it a useful inverse to have when using these.

lib/engines/content_object_store/app/models/content_object_store/result_monad.rb
```diff
@@ -0,0 +1,11 @@
+module ContentObjectStore
+  ResultMonad = Struct.new(:success, :message, :object) do
```
pezholio, line 2: Would it be better to use `Data.define` here? So:

```suggestion
  ResultMonad < Data.define(:success, :message, :object)
```

That way the monad behaves more like a Class (also I think Structs are historically quite slow)

lib/engines/content_object_store/app/models/content_object_store/change_dispatcher.rb
```diff
@@ -0,0 +1,15 @@
+module ContentObjectStore
+  module ChangeDispatcher
+    class Now
+      def verb
+        "changed and published"
```
pezholio, line 5: This is veeeeery nice!

lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
```diff
@@ -20,15 +20,12 @@ def update
     ContentObjectStore::SchedulePublishingWorker.dequeue(content_block_edition)
 
     new_edition_params = edition_params.merge!(scheduled_publication_params)
-
+    change_dispatcher = params[:schedule_publishing] == "schedule" ? ContentObjectStore::ChangeDispatcher::Schedule.new : ContentObjectStore::ChangeDispatcher::Now.new
```
pezholio, line 21: I wonder if you can break this logic down further so you have a `ContentObjectStore::ChangeDispatcher` class that returns `ContentObjectStore::ChangeDispatcher::Schedule` or `ContentObjectStore::ChangeDispatcher::Now` depending on the value of `schedule_publishing`?

lib/engines/content_object_store/app/models/content_object_store/change_dispatcher.rb
pezholio: This feels very much like someone's been writing Kotlin recently 😉

lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
```diff
@@ -26,13 +26,15 @@ def update
     ).call(new_edition_params)
     new_content_block_edition = result.object
 
-    redirect_to content_object_store.content_object_store_content_block_document_path(new_content_block_edition.document),
-                flash: { notice: result.message }
-  rescue ActiveRecord::RecordInvalid => e
-    @content_block_edition = e.record
-    @content_block_document = content_block_edition.document
-    @edition_params = new_edition_params
+    if result.success?
```
pezholio, line 29: 🙌

pezholio (approved):
LGTM - just a couple of suggestions, but no great shakes

lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
tahb: We could do! I was starting to wonder if I was going overboard with too many abstractions and indirections. Leaving the controller with this responsibility (rather than hiding it away) felt clearer to the reader. I suspect that if there was every a third type of dispatcher then another refactor might be worth the move?

lib/engines/content_object_store/app/models/content_object_store/change_dispatcher.rb
tahb: Thanks! I hope it's not because of Kotlin, I was just trying to practice object design from 99 bottles. 

Before I added the dispatcher I was looking the if conditionals we had in UpdateEdition. We had something like:

```ruby
def call
   …
   if schedule? schedule : publish_now
   …
   flash = schedule? "scheduled" : "changed and published"
   …
```

So I tried to name that behaviour and put it into an injectable thing 🤞

lib/engines/content_object_store/app/models/content_object_store/result_monad.rb
tahb: Interestingggggg! [This definitely looks](https://github.com/ruby/ruby/pull/6353) like a more modern way to do it (Couldn't find anything on speed improvements). We also get named params this way which I much prefer and would level this up.

lib/engines/content_object_store/app/controllers/content_object_store/content_block/workflow_controller.rb
pezholio: Yeah, that's fair 👍

lib/engines/content_object_store/app/models/content_object_store/result_monad.rb
tahb: Done 👍

pezholio (approved):

lib/engines/content_object_store/app/models/content_object_store/result_monad.rb
pezholio: Ah, I was getting Structs mixed up with OpenStructs, which are veeeery slow - https://www.honeybadger.io/blog/how-openstruct-and-hashes-can-kill-performance/

PR #9407, opened 2024-09-03
Fix loading of GA4 analytics modules

ryanb-gds:
If using the GOV.UK publishing component analytics module loading, the loader does not expect the same structure as the normal GOV.UK module loader. Instead it expects a simple object with an init function to call.

This PR also ensures consistent usage of kebab case for module data attributes

Trello: https://trello.com/c/JGuDLzro

### DIFF
--- a/app/helpers/admin/analytics_helper.rb
+++ b/app/helpers/admin/analytics_helper.rb
@@ -2,7 +2,7 @@ module Admin
   module AnalyticsHelper
     def track_analytics_data_on_load(title)
       {
-        event_name: "page_view",
+        event: "page_view",
         page_view: {
           publishing_app: "Whitehall",
           user_created_at: current_user&.created_at&.to_date,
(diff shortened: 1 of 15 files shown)


app/helpers/admin/analytics_helper.rb
```diff
@@ -2,7 +2,7 @@ module Admin
   module AnalyticsHelper
     def track_analytics_data_on_load(title)
       {
-        event_name: "page_view",
+        event: "page_view",
```
ryanb-gds, line 5: I wasn't 100% sure if this was supposed to be "event" or "event_name", but all other GA4 events used "event" so I assumed it probably was

dnkrj: Probably worth double checking that this still works once it goes through

dnkrj (approved):

PR #9409, opened 2024-09-04
Fix flaky Sentry tests

pezholio:
The Sentry test introduced in https://github.com/alphagov/whitehall/pull/9405 appears to be a bit flaky, failing in some circumstances. This changes the test to a functional test (which is a better fit) and catches the error directly, and this seems to fix the issue.

I'm not _entirely_ sure why this works now, but I've re-ran the suite a number of times, and haven't experienced an issue, whereas before we were experiencing failures at least one time out of every 2-3, so I'm confident this will unblock CI.


lib/engines/content_object_store/test/functional/sentry_tags_test.rb
```diff
@@ -1,28 +1,32 @@
...
-class ContentObjectStore::SentryTagsTest < ActionDispatch::IntegrationTest
+class ContentObjectStore::SentryTagsTest < ActionController::TestCase
+  tests ContentObjectStore::ContentBlock::DocumentsController
   include Sentry::TestHelper
 
   setup do
     setup_sentry_test
     feature_flags.switch!(:content_object_store, true)
     login_as_admin
+    @routes = ContentObjectStore::Engine.routes
```
tahb, line 12: I couldn't see what was using `@routes` so wanted to why we need this but removing it definitely fails:

```
ContentObjectStore::SentryTagsTest#test_throwing_an_error_includes_tags:
ActionController::UrlGenerationError: No route matches {:action=>"index", :controller=>"content_object_store/content_block/documents"}
```

This feels too mysterious. Might be worth a comment to explain "We have no idea what uses this instance variable but removing this makes it flakey"

lib/engines/content_object_store/test/functional/sentry_tags_test.rb
```diff
@@ -1,28 +1,32 @@
...
-      get content_object_store.content_object_store_content_block_documents_path
-    rescue ArgumentError
-      assert_equal sentry_events.count, 1
-      assert_equal sentry_events[0].tags[:engine], "content_object_store"
+      get :index
+    rescue ArgumentError => e
+      Sentry.capture_exception(e)
     end
+
+    assert_equal sentry_events.count, 1
```
tahb, line 29: Noticed when testing this that we probably want these the other way around so [the expected is first and the actual is second](https://rubydoc.info/gems/minitest/Minitest/Assertions#assert_equal-instance_method). Otherwise it reads backwards on failure:

```
Expected: nil
  Actual: "content_object_store"
```

tahb, line 29: Given we call `Sentry.capture_exception(e)` in the same test there really should always be a count of 1. So this probably doesn't provide any value.

Happily the second assertion around tagging does kick in so that's a valuable one👍

tahb (approved):
Aware we are trying to push through a fix rather than revert so we are under pressure!

I've left some suggestions which I think we ought to do before merging, though they aren't important enough to block.

ryanb-gds (approved):
Agree with Tom's comments and I'm not really clear on how Sentry tests work, but treating this as a feature test does seem more likely to result in consistent behaviour

lib/engines/content_object_store/test/functional/sentry_tags_test.rb
pezholio: Yeah, that's a thing related to Functional Tests - they don't work out of the box in engines. I'll add a comment

PR #9416, opened 2024-09-04
Enable content object store in staging/integration

pezholio:
This ensures we can still work on the service in non-production environments without having to expressly turn on the feature flag


tahb (approved):
That was fast! Looks like just the thing (though I don't know if we want or need staging?) 🚀

pezholio:
I think staging is handy for deploying branches etc, whereas integration is always up to date with main.

PR #9417, opened 2024-09-04
Hide scheduled at row for unscheduled objects

pezholio:
At the moment, we show “Scheduled at” for any blocks that have ever been scheduled, even if they have since been published. Additionally, we show a blank line for blocks that have never been scheduled. 

This hides the "Scheduled at" row unless something is currently awaiting publication.

I've also titleized the block's status, which looks a bit nicer

### Before

image

image

### After

image

image

### DIFF
--- a/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
+++ b/lib/engines/content_object_store/app/components/content_object_store/content_block/document/show/summary_list_component.rb
@@ -8,7 +8,7 @@ private
   attr_reader :content_block_document
 
   def items
-    [
+    items = [
       title_item,
       *details_items,
       organisation_item,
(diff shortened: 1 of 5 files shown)


tahb (commented):
[I thought we were going to leave these knowingly as part of the unintelligent view which needed research?](https://github.com/alphagov/whitehall/pull/9387#discussion_r1741082675)

My hunch is that these fields should be displayed as a separate thing such as on Whitehall main: Screenshot 2024-09-04 at 16 48 15

pezholio:
@tahb Yeah, I thought we were keeping the views as unintelligent for that PR then coming back to them. I like copying the approach from Whitehall, but I feel like this is a smaller step, and gets rid of potentially misleading / confusing info

tahb (approved):
Okay 👍

PR #9418, opened 2024-09-05
Reduce the number of queries used to build ministers index links

ryanb-gds:
The ministers index presenter was not leveraging eager loading and consequently was running huge numbers of queries. This has the potential to cause performance issues in the Sidekiq queues, as indeed was recently the case with the embassies index presenter.

We also remove the unused govspeak helper from the presenter.

I've tested this in integration and it produces exactly the same list of links as the current presenter.


dnkrj (approved):
Any way that we could a regression test for this?

ryanb-gds:
> Any way that we could a regression test for this?

There's this existing one: https://github.com/alphagov/whitehall/blob/main/test/unit/app/presenters/publishing_api/ministers_index_presenter_test.rb

PR #9420, opened 2024-09-05
Fix content block form errors

tahb:
The errors on the new/edit content_block:edition object stopped working. They’d display on the page with anchor links but clicking those wouldn’t move the user to the related input.

The anchor links that the errors were using had changed from “content_block_edition” to “content_block/edition”. This is likely due to a recent refactor that renamed these objects to make them more concise across the code base[1].

Instead of trying to change the automatically generated error links we instead update the ids of the input fields to bring them inline. Hopefully we’ll be more in keeping with Rails conventions this way.

[1] https://github.com/alphagov/whitehall/pull/9304


pezholio (approved):
Happy with this. I agree, if we stick with the conventions, it'll make life much easier!

PR #9421, opened 2024-09-05
Suppress noisy sidekiq logs in test

tahb:
When running the test suite I noticed we can now see `info` logs for Sidekiq, eg `2011-11-11T11:11:11.000Z pid=1 tid=6z5 INFO: performing content block publishing job for Edition 380`. This provides no value and clutters the test output.

We now suppress this in our worker and cucumber tests.

We add config to our engine tests rather than putting it in the main Whitehall test_helper.rb etc. Whilst they do use sidekiq to queue jobs I’m not sure how this hasn’t come up for them. On a quick look it seems like their tests all use `perform_async` so it’s possible that this means jobs never get processed and so no logging of "performed". In which case they don’t have the same problem and it makes sense to just have this for our noisy tests?

## Before

Screenshot 2024-09-05 at 16 55 25

## After
Screenshot 2024-09-05 at 16 55 30

### DIFF
--- a/lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
+++ b/lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
@@ -1,5 +1,10 @@
 require_relative "../support/stubs"
 
+# Suppress noisy Sidekiq logging in the test output
+Sidekiq.configure_client do |cfg|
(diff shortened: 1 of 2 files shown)


pezholio (approved):
Looks good to me - well spotted!

PR #9422, opened 2024-09-06
(Bug) Can't publish a new version when a scheduled version is waiting to be published

pezholio:
This fixes a bug where if a user tries to edit an already scheduled content block, they get an error telling them the scheduled date and time cannot be blank, even if they try and publish the change immediately. This is because when we copy over the new edition, the `state` of the edition is copied too.

This fixes that issue, as well as carrying out a bunch of refactors to push the concerns used in the Edition model to be closer to the Edition itself. I've also moved the behaviour where we create a new edition from an existing edition to be within an Edition concern, this tidies up the Publishable concern, and also makes testing easier.


lib/engines/content_object_store/app/models/content_object_store/content_block/edition/updatable.rb
```diff
@@ -0,0 +1,38 @@
...
+        edition.assign_attributes(
+          filter_params_for_validation_check(edition_params),
+        )
+      end
+
+      unless new_edition.valid?
+        raise ActiveRecord::RecordInvalid, new_edition
+      end
+
+      new_edition.save!
```
tahb, line 19: I think these are all points to consider before moving forwards:

1. I think we need to remember to assign the new edition as the documents `latest_edition_id`. Having that close to the point of creation feels important so that it isn't forgotten.

    We have this happening over in this [CreateEdition service](https://github.com/alphagov/whitehall/blob/9feec79e7c0205cd386c4162d7aaed5627f086a3/lib/engines/content_object_store/app/services/content_object_store/create_edition_service.rb#L11). This class also feels relevant to this change as it has a different way of doing the same(?) thing so;

2. It seems to me like `CreateEditionService` and `UpdateEditionService` would both benefit from using the same method?

3. When we consider the above point 2: the naming of 'updatable' feels off since creating isn't updating. How about 'creatable' as a slight tweak? It goes play nicely with the method name we have of "create_new_edition"

I feel like I'm losing my grasp of what is a Publishable and what would be a Creatable concern. Do you have a definition for how you think about it? Could it be Publishable is about downstream requests (to publishing API or to our queue) and Creatable is only about inside Whitehall? I think I'd value some sort of definition for each concern so I know how to approach each file. What do you think?

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/edit.html.erb
```diff
@@ -7,4 +7,10 @@
   } %>
 <% end %>
 
+<% if @content_block_edition.state == "scheduled" %>
```
tahb, line 10: Minor suggestion: I think we can use enums to do `@content_block_edition.scheduled?` if we want to make this more concise and be less dependent on the actual value (so it could change).

lib/engines/content_object_store/features/step_definitions/content_object_store_steps.rb
```diff
@@ -415,3 +422,25 @@ def should_show_edit_form_for_email_address_content_block(document_title, email_
...
+end
+
+def change_details
+  fill_in "Title", with: "Changed title"
+  fill_in "Email address", with: "changed@example.com"
+  select "Ministry of Example", from: "content_block/edition_lead_organisation"
+  click_save_and_continue
+end
+
+def click_save_and_continue
```
tahb, line 444: Yes, high time we had this!

tahb (commented):
Oh this is a bug! When we discussed it I thought you meant we just didn't have a warning message, I didn't realise the form failed! Good catch and good testing.

lib/engines/content_object_store/app/models/content_object_store/content_block/edition/updatable.rb
pezholio: To me, `Creatable` doesn't feel the right name either because we're making a copy of the Edition and then returning a new one, but then, neither does `Updatable`. Maybe we need to rethink the API altogether. Maybe it would be better to have a class level method that takes the latest edition and the updated params?

This might help with `2` because using it in the context of creating the first edition would be easier because we could make the latest edition optional?

tahb: If it's called on an edition the method could just be create_copy

pezholio: How about `Copyable` for the name of the concern? I like the idea of having a `update_document_reference_to_latest_edition!` method. I know callbacks aren't usually the best idea, because they hide away, but this does feel like a good use case for them. I'll have a think...

pezholio: I've had a think, and no. I remember why I hate callbacks 😆

tahb: Callbacks does feel like the Rails way. I too pushed them away due to their behaviour being hidden and magical.

lib/engines/content_object_store/app/views/content_object_store/content_block/editions/edit.html.erb
pezholio: Good call! 👍

tahb (approved):
I think Copyable is great and the approach to copy_edition is clearer 👍 

I've left a couple of suggestions but they aren't blocking since they are around readability.

lib/engines/content_object_store/test/unit/app/models/content_block_edition/copyable_test.rb
```diff
@@ -0,0 +1,56 @@
+require "test_helper"
+
+class ContentObjectStore::CopyableTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  let(:document) { create(:content_block_document, :email_address) }
+  let(:edition) { create(:content_block_edition, document:, state: "published") }
+
+  it "creates a new edition given parameters" do
+    # When I create a new edition
```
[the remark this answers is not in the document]
pezholio: Yeah, I considered that, but couldn't find a way to share the setup with all expectations, which would mean having to do setup before each individual test, which would add extra time to the test run. This seemed like a good way round it.

PR #9438, opened 2024-09-13
Ensure history mode government select options appear in correct order

ryanb-gds:
Because `find_each` loads results in batches, the governments were not respecting the newest first ordering, so we switch to using a simple each


minhngocd:
@ryanb-gds is there a test that confirms this?

ryanb-gds:
> @ryanb-gds is there a test that confirms this? or is the batch size so large, it's not worth creating a massive setup for?

There is, batch size is too large though as you suspect

minhngocd (approved):

PR #9439, opened 2024-09-13
Remove usages of override government feature flag

ryanb-gds:
This feature is now in production use, so no need to keep the feature flag around. Will follow up with a separate PR to remove the flag itself and the database record of the flag.

Trello: https://trello.com/c/Q7tCtSXy


ChrisBAshton:
Converting to draft so that we're not nudged to merge this until mid-October, as we want to keep it open for a few weeks as per the Trello card.

ChrisBAshton (approved):

PR #9447, opened 2024-09-18
550 cache busting

Harriethw:
When a content block is updated we do not want to wait the default cache time of 5 mins in order to see the changes in Documents that use this content block.

Publish intents allow us to reset the cache time for a Document
https://docs.publishing.service.gov.uk/repos/content-store/publish_intents.html

This PR creates a publish intent for any host Documents whenever a block is published. 
I've put it at the point of publishing, rather than scheduling or updating, because we can instantly override the cache (or almost instantly depending on length of Sidekiq queue) whereas if we were to call it at point of scheduling, or only when we are updating, it gets a bit trickier to ensure that the method definitely gets called  (we would also then have to cancel the intent if a scheduled Block was cancelled)

## questions

I've put the PUT requests onto a Worker to be performed asynchronously as Whitehall do. I haven't stress tested this and would need help to do so if we wanted to do so- I'm not sure how slow it would be if a Block had 100-1000s of host Documents, but if it was slower then 5 mins the cache would be refreshed anyway.

## To see locally

1. follow instructions to be able to see published changes locally https://github.com/alphagov/govuk-docker/blob/ddd2f32114416dd543892436df1c144e2bb44b03/docs/how-tos/publishing-content-on-whitehall.md
1. Create a published Content Block in Whitehall. Get the embed code from the details view e.g. {{embed:content_block_email_address:8603b86f-281d-43ba-b047-9de6d471db05}}
1. embed that code into a Document (.e.g News Article) and force publish it.
1. Create log/breakpoint in the `content-store`  [repo to see the cache time set for a content item](https://github.com/alphagov/content-store/blob/ebe5467a5ecaa32207a18e5976bb23e535fd5916/app/controllers/content_items_controller.rb)
1. Go to your published Document on the govuk-frontend, e.g. http://government-frontend.dev.gov.uk/government/news/w-cache and observe the log in content-store should be 300 
1. Update/edit the original published Content Block and publish changes
1. Observe in Publishing API logs that a PUT request has been made to `publish-intent`
1. Visit the published Document in the frontend and observe cache log should now be the difference between the time you re-published the Content Block and now... after 5 minutes the publishing intent will be ignored and cache time set back to 300. 


db/seeds.rb
```diff
@@ -7,7 +7,7 @@
   gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
   User.create!(
     name: "Test user",
-    permissions: ["signin", "GDS Admin", "GDS Editor", "Managing Editor", "Export data"],
+    permissions: ["signin", "GDS Admin", "GDS Editor", "Managing Editor", "Export data", "Sidekiq Admin"],
```
Harriethw, line 10: should we just add this for the Test User for good? I feel like it will continue to be useful as we continue to use Sidekiq...

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
```diff
@@ -51,6 +52,15 @@ def update_content_block_document(new_content_block_edition:, update_document_pa
 
   private
 
+    def remove_cache_for_dependent_content(content_block_edition:)
```
Harriethw, line 55: just realised I'm switching between dependent and host content... i've seen both used elsewhere, shall we stick with Host in the code as that's what we've got already?

db/seeds.rb
pezholio: Yeah, I can't see the harm 👍

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
Harriethw: cool, have changed to use host

lib/engines/content_object_store/test/unit/app/workers/publish_intent_worker_test.rb
```diff
@@ -0,0 +1,13 @@
+require "test_helper"
+
+class ContentObjectStore::PublishIntentWorkerTest < ActiveSupport::TestCase
+  test "#perform adds a publishing intent to the Publishing API" do
+    base_path = "/base-path"
+    timestamp = Time.zone.now.to_s
+    publish_intent = { foo: "bar" }
+
+    PublishingApi::PublishIntentPresenter.expects(:new).once.returns(publish_intent)
```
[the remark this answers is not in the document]
Harriethw: good spot, have done

PR #9450, opened 2024-09-19
Diagnose and fix flaky tests

pezholio:


pezholio:
I can't replicate this now, even after reverting the reverted code that caused the fail here https://github.com/alphagov/whitehall/actions/runs/10923006590/job/30322335477 - I assume it's because there's been more tests added since then, so the exact order of tests is no longer happening.

PR #9459, opened 2024-09-20
Content modelling/550 fix cache busting

Harriethw:
In this previous PR https://github.com/alphagov/whitehall/pull/9447 we sent Publish Intents for Host Content, but hardcoded Whitehall as the publishing app - this meant jobs for Documents on other apps will fail.

This PR uses the publishing app on a Host Content Item (added here https://github.com/alphagov/publishing-api/pull/2876) and ensures the correct publishing app name is passed when creating the Publish Intent.


pezholio (approved):
LGTM 👍

Issue #9225, opened 2024-07-04
Irish language two-letter code should be ga not gd


nacnudus:
https://github.com/alphagov/whitehall/blob/78fb8bdce8fb45ae61409eacb63580a1f2afb922/config/locales/en.yml#L412

Introduced in https://github.com/alphagov/whitehall/pull/4713

Reference: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

Noticed by a user of GovSearch. "There seems to be a bug - might be related to Whitehall - for documents that are in Irish. They're classified in GovSearch as 'Scottish Gaelic' rather than 'Gaeilge' or 'Irish'. I think this might be picking up that the Irish language urls wrongly end in '.gd' when they should end in '.ga'."

There are currently three such pages:

* https://www.gov.uk/guidance/eu-settlement-scheme-how-we-use-your-personal-information.gd
* https://www.gov.uk/government/publications/union-connectivity-review-interim-report.gd
* https://www.gov.uk/government/publications/union-connectivity-review-final-report.gd

vignesh1507:
Made the necessary changes to the code.

PR #9461, opened 2024-09-21
Update en.yml

vignesh1507:
Updated en.yml file(config/locales/en.yml) as it had few changes in it.


vignesh1507:
- #9225 
- Resolved the above issue.

PR #9473, opened 2024-09-26
Pin govuk_publishing_components to 43.3.0

ChrisBAshton:
Reverts alphagov/whitehall#9470

This is the last known working version of the dependency.
In bumping to 43.4.1 (in https://github.com/alphagov/whitehall/pull/9463), Whitehall components driven by
JS broke. Namely: choosing organisations, tagging ministers, etc.

It appears to be something to do with how the changes in
https://github.com/alphagov/govuk_publishing_components/pull/4246
are incompatible with Whitehall's own accessible-autocomplete
implementation. Until we have time to look into this properly,
let's pin to the last known working version.


minhngocd (approved):

ChrisBAshton:
Follow-on card: https://trello.com/c/l1YEnZCa/405-unpin-govukpublishingcomponents-in-whitehall

PR #9475, opened 2024-09-26
content-modelling: 550 send publish intents when scheduling a block

Harriethw:
we previously sent publish intents for the instant that a block was published, but this was due to misunderstanding of what the intents do and how the cache would work - they only really make sense when scheduling a block https://docs.publishing.service.gov.uk/repos/content-store/publish_intents.html

We are doing this with hopefully a full understanding of the caveats now https://trello.com/c/OEF9GDRy/550-cache-busting-our-object-updates-should-propagate-through-to-citizens-at-the-same-time#comment-66f3d8389b5abea342d04b6a


pezholio (approved):
LGTM 👍 We got there eventually! 😅

PR #9483, opened 2024-09-30
Use only `EMERGENCY_BANNER_REDIS_URL` environment variable

brucebolt:
In 93c708158530678f10fc5eb7bc2bb7af644554a4, we allowed this application to use either `EMERGENCY_BANNER_REDIS_URL` or `REDIS_URL`, as different environments were in differet states.

That is no longer the case, so we can remove the code to default to one over the other.

[Trello card](https://trello.com/c/2RF94Axr)

### DIFF
--- a/test/functional/admin/emergency_banner_controller_test.rb
+++ b/test/functional/admin/emergency_banner_controller_test.rb
@@ -163,44 +163,4 @@ class Admin::EmergencyBannerControllerTest < ActionController::TestCase
       assert_equal expected_response, Redis.new.hgetall("emergency_banner").symbolize_keys
     end
   end
-
-  context "instantiating Redis" do
-    setup do
-      @mock_redis = Minitest::Mock.new
-      def @mock_redis.hgetall(*args); end
-      def @mock_redis.del(*args); end
-    end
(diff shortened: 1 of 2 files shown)


Gweaton (approved):

test/functional/admin/emergency_banner_controller_test.rb
```diff
@@ -170,36 +170,16 @@ class Admin::EmergencyBannerControllerTest < ActionController::TestCase
...
-          Redis.expects(:new).with(
-            url: "redis://my-redis-url",
-            reconnect_attempts: 4,
-            reconnect_delay: 15,
-            reconnect_delay_max: 60,
-          ).twice.returns(@mock_redis)
-
-          delete :destroy
-        end
+    view_test "uses the EMERGENCY_BANNER_REDIS_URL environment variable as the URL for the Redis client" do
```
Gweaton, line 173: We may as well remove this test, I think - you're removing it in the other repo anyway

brucebolt: Good idea, I've removed it now.

PR #9487, opened 2024-10-01
WIP - add tasks to publish history pages content object

KludgeKML:
Currently, the history pages are published by government frontend - this can't simply be replaced with special routes in Publishing API because the contents of the page have to go into the content store to be searchable. The current solution is for them to be published directly from the frontend app, and it's a bit of a mess (there's ERB code in the body). We want to move rendering over to Frontend, but it would be better if we could make the content item a bit more rational, and publish them from an app (Whitehall) - This would also serve as a first step to it having a UI that content people can use to make edits (if it turns out that is necessary).

Unsolved problems:
- neither history nor generic schemas will allow this to work (ideally we'd want a header image and an array of in-page navigation links)
- how do we put images in (do we include them in the same file as the yaml, publish them as assets and then insert them as image govspeak tags?)

Related to:
https://trello.com/c/g3oNeSDH/199-move-document-type-history-from-government-frontend-to-frontend
https://trello.com/c/hJVUjqbs/1242-frontend-apps-publish-content-items-to-publishing-api

### DIFF
--- a/app/services/publish_history_page.rb
+++ b/app/services/publish_history_page.rb
@@ -1,0 +1,61 @@
+class PublishHistoryPage
+  attr_reader :history_page_content_item, :content_id
+
+  def initialize(history_page_details)
+    @history_page_content_item = make_content_item(history_page_details)
+    @content_id = history_page_details[:content_it]
+  end
+
+  def self.call(history_page_details)
+    new(history_page_details).call
+  end
+
+  def call
+    send_to_publishing_api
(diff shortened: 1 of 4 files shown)


ryanb-gds:
@KludgeKML can we close this on the basis that it's being superseded by the flexible pages project?

KludgeKML:
Closing on the basis that flexible pages is probably a better solution to this.

PR #9488, opened 2024-10-01
Fix statistics announcement type mismatch validation

lauraghiorghisor-tw:
On the back of a few Zendesk tickets, we're improving how validation works for announcements and their associated publications. In the context of renaming "National Statistics" to "Accredited Official Statistics", the users were not setting the correct "publication type" either on the announcement or publication. The validation in the code was technically unfit for the models, causing the users to get a generic "Something went wrong" error. 

These changes ensure that: 
- Validation happens on publication so that when a new publication is created off of an announcement, it is validated for type match;
- Subsequent edits of the publication also validate the type match;
- The publication validation errors is also surfaced on the announcement for subsequent edits;
- The errors include a reference to what the label type is now (what the user sees);
- A bit of guidance text has been added to the "Connect a document" page.

Publication create document from announcement/edit page
Screenshot 2024-10-02 at 11 47 19

Announcement edit page
Screenshot 2024-10-02 at 11 46 47

Guidance text on search page
Screenshot 2024-10-02 at 11 46 03

I had a think about why the users were confused, and whether or not we may have missed some migration or something. Going through [Ryan's PRs on the original renaming work](https://trello.com/c/jkaJlgez/1501-update-whitehall-for-changes-to-national-stats), I think that is not the case, since the renaming was only a label change + republish. It's probably a lack of comms/guidance on the renaming. I've therefore raised [a ticket](https://trello.com/c/1fSoVVZu/11-improve-guidance-on-accredited-official-statistics) on the content board to further discuss guidance, and validate the copy of the current changes. 

[Trello](https://trello.com/c/a7naRYKH/2880-statistical-announcements-issue-saving-accredited-official-statistics-type)
[Content trello ticket](https://trello.com/c/1fSoVVZu/11-improve-guidance-on-accredited-official-statistics)
[Zendesk 1](https://govuk.zendesk.com/agent/tickets/5934093)
[Zendesk 2](https://govuk.zendesk.com/agent/tickets/5931376)

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -23,6 +23,7 @@ class Publication < Publicationesque
   has_one :statistics_announcement
   attr_accessor :statistics_announcement_id
 
+  validate :statistics_announcement_is_matching_type, if: :statistics_announcement_exists?
   after_create :assign_statistics_announcement
 
   after_save :touch_statistics_announcement
@@ -30,6 +31,17 @@ class Publication < Publicationesque
   scope :statistical_publications, -> { where("publication_type_id IN (?)", PublicationType.statistical.map(&:id)) }
   scope :non_statistical_publications, -> { where("publication_type_id NOT IN (?)", PublicationType.statistical.map(&:id)) }
   scope :corporate_publications, -> { where(publication_type_id: PublicationType::CorporateReport.id) }
+
+  def statistics_announcement_exists?
+    statistics_announcement.present? || statistics_announcement_id.present?
+  end
+
+  def statistics_announcement_is_matching_type
+    statistics_announcement = self.statistics_announcement || StatisticsAnnouncement.find(statistics_announcement_id)
+    unless statistics_announcement.publication_type == publication_type
+      errors.add(:publication_type_id, message: "does not match announcement type: must be '#{statistics_announcement.publication_type.singular_name}'")
+    end
+  end
 
   def touch_statistics_announcement
     if published? && !statistics_announcement.nil?
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -16,15 +16,14 @@ class Publication < Publicationesque
   include Edition::CanApplyToLocalGovernmentThroughRelatedPolicies
   include Edition::TopicalEvents
 
+  before_validation :set_statistics_announcement
   validates :publication_type_id, presence: true
   validate :only_publications_allowed_invalid_data_can_be_awaiting_type
   validate :attachment_required_before_moving_out_of_draft
+  validate :statistics_announcement_is_matching_type, if: :statistics_announcement
 
   has_one :statistics_announcement
   attr_accessor :statistics_announcement_id
-
-  validate :statistics_announcement_is_matching_type, if: :statistics_announcement_exists?
-  after_create :assign_statistics_announcement
 
   after_save :touch_statistics_announcement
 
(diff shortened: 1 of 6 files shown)


test/functional/admin/statistics_announcement_publications_controller_test.rb
```diff
@@ -25,6 +25,12 @@ class Admin::StatisticsAnnouncementPublicationsControllerTest < ActionController
     refute_select ".govuk-table"
   end
 
+  view_test "GET :index shows inset text" do
+    get :index, params: { statistics_announcement_id: @official_statistics_announcement }
+
+    assert_select ".govuk-inset-text", text: "The type of the document you are trying to connect must match the type of the statistics announcement. Only Official Statistics documents will be shown."
```
lauraghiorghisor-tw, line 31: I've lost all sense of the English language, help.

app/models/statistics_announcement.rb
```diff
@@ -38,7 +38,10 @@ def can_publish_to_publishing_api?
   has_many :statistics_announcement_organisations, inverse_of: :statistics_announcement, dependent: :destroy
   has_many :organisations, through: :statistics_announcement_organisations
 
-  validate  :publication_is_matching_type, if: :publication
+  validates_associated :publication, if: :publication,
```
lauraghiorghisor-tw, line 41: Maybe this is overly smart and just having the old (duplicated) validation is more readable.

Whilst this is validation for the publication "field", this is surfaced on the announcement and I feel like the messaging should be flipped somehow. We're not linking to anywhere else on the page as we would for an actual field, so this can be as generic and as "about announcement" as we want.

lauraghiorghisor-tw: Maybe we can "This announcement has a publication whose type does not match the current selection" -> this validation would only happen when you edit an announcement, as on create you would not have the associated publication yet.

test/functional/admin/statistics_announcement_publications_controller_test.rb
```diff
@@ -69,7 +75,7 @@ class Admin::StatisticsAnnouncementPublicationsControllerTest < ActionController
 
     assert_response :success
     has_search_results_table
-    assert_select "ul.govuk-error-summary__list a", text: "Publication type does not match: must be statistics"
+    assert_select "ul.govuk-error-summary__list a", text: "Publication type does not match announcement type: must be 'Official Statistics'"
```
lauraghiorghisor-tw, line 78: To single quote or not to singe quote. I think Rosa once showed me some messaging with quotes and now it's how I do things. But they looked horrific on the inset text so no quotes there.

test/unit/app/models/publication_test.rb
```diff
@@ -24,6 +24,35 @@ class PublicationTest < ActiveSupport::TestCase
...
+    accredited_official_announcement = create(:statistics_announcement, publication_type_id: PublicationType::NationalStatistics.id)
+    accredited_official_publication = create(:draft_national_statistics, statistics_announcement_id: accredited_official_announcement.id)
+
+    accredited_official_publication.publication_type_id = PublicationType::OfficialStatistics.id
+
+    assert_not accredited_official_publication.valid?
+    assert_equal accredited_official_publication.errors[:publication_type_id], ["does not match announcement type: must be '#{accredited_official_announcement.publication_type.singular_name}'"]
+  end
+
+  test "it saves the announcement association" do
```
lauraghiorghisor-tw, line 48: I really needed this test for removing the `after_create` callback, but perhaps less valuable to commit.

app/models/publication.rb
```diff
@@ -31,6 +32,17 @@ class Publication < Publicationesque
   scope :non_statistical_publications, -> { where("publication_type_id NOT IN (?)", PublicationType.statistical.map(&:id)) }
   scope :corporate_publications, -> { where(publication_type_id: PublicationType::CorporateReport.id) }
 
+  def statistics_announcement_exists?
+    statistics_announcement.present? || statistics_announcement_id.present?
```
ChrisBAshton: Do we know we need both conditions? Wouldn't just the `statistics_announcement.present?` be enough?

app/models/statistics_announcement.rb
ChrisBAshton: Ooh, I've not seen `validates_associated` before 👍 LGTM

test/functional/admin/statistics_announcement_publications_controller_test.rb
ChrisBAshton: 🤷 LGTM

test/unit/app/models/publication_test.rb
ChrisBAshton: Worth moving this into the second commit then? `Remove after create callback`

test/functional/admin/statistics_announcement_publications_controller_test.rb
ChrisBAshton: LGTM 👍

ChrisBAshton (approved):
Nice work! ⭐

test/unit/app/models/publication_test.rb
lauraghiorghisor-tw: Oh yeah 👍🏻

PR #9496, opened 2024-10-04
Refactor create/edit controllers

pezholio:
This refactors the controllers when creating and editing content blocks, as well as saving editions after the first step, rather than passing around params. We'll need to do this in the next feature, as we'll be looking to save to the draft content store to allow previewing to work.

There's a lot of mess here, so I've dispensed with my "every commit has to be green" rule in an attempt to make it a bit easier to follow 😅


lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
```diff
@@ -0,0 +1,74 @@
+class ContentObjectStore::ContentBlock::Editions::StepsController < ContentObjectStore::BaseController
+  VALID_STEPS = {
+    review_links: "review_links",
+    schedule_publishing: "schedule_publishing",
+  }.freeze
+
+  def show
```
Harriethw, line 7: I'm a bit confused about what the `update` method is for - it looks like `show` is the one that's used when updating?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/publications_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ContentObjectStore::ContentBlock::Editions::PublicationsController < ContentObjectStore::BaseController
```
Harriethw, line 1: I have a little bit of a reservation here as we have said we'll try to keep to Whitehall naming as much as poss - is the `create` here actually the equivalent of `publish` in Whitehalls `Edition_Workflow_Controller`? I could see if we started to have draft/superseded states etc that a Workflow Controller would be useful.
And is this now only used by the 'New' user journey?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
```diff
@@ -1,95 +1,35 @@
...
+    new_edition.document.assign_attributes(edition_params[:document_attributes].except(:block_type))
 
-    if new_edition.valid?
-      @host_content_items = ContentObjectStore::GetHostContentItems.by_embedded_document(
-        content_block_document: @content_block_document,
-      )
-
-      render :review_links
+    if new_edition.valid? && new_edition.document.valid?
+      new_edition.save!
```
Harriethw, line 18: so here we create a new Edition before the user has actually got to the end of the form? That means we go back to having unseen Draft Content Block Editions around if they don't complete the form?

Harriethw: ah sorry I think I'm getting confused by the diff

Harriethw: is this `create` method is used when both creating the first Edition and when adding subsequent Editions?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    redirect_to content_object_store.content_object_store_content_block_step_path(id: @content_block_edition.id, step: :schedule_publishing)
+  end
+
+  def schedule_publishing
+    @content_block_document = @content_block_edition.document
+
+    render :schedule_publishing
+  end
+
+  def create_schedule
```
Harriethw, line 54: is this method more like `schedule_or_publish`?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
Harriethw: Ah I think I see, `show` and `update` hand off to each other - show shows the forms and `update` does any necessary actions?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
```diff
@@ -0,0 +1,74 @@
+class ContentObjectStore::ContentBlock::Editions::StepsController < ContentObjectStore::BaseController
```
Harriethw, line 1: Could we be more specific in naming this controller? This are the steps to add a new Edition to an existing Document?

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
Harriethw: I think while this is more strictly RESTful it might miss the realities of the publishing workflow - I think practically there's a big difference between creating a new Edition on a new Document and adding new Editions to an existing Document, so it would be helpful to keep the `edit` step separate to the `create` step. I also would prefer us not to go back to creating orphaned Draft Editions if someone abandons a form.

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
pezholio: `show` is when we're showing the form, which then POSTs to the `update` method

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/publications_controller.rb
pezholio: Ah, fair enough. I was trying to keep it RESTful, but I think it makes sense to keep it aligned

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
pezholio: Yeah, I think we'll have to if we want to preview editions. It's not ideal, but I can't really see a way round it. The happy path _should_ be that users continue through the flow, so the issue of there being draft editions hanging around should be minimal

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
pezholio: Good call. I'll fix that 👍

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions/steps_controller.rb
pezholio: Maybe it's a workflow actually? Then we can combine the steps and the workflow in one controller? Then we have a workflow as follows for an initial edition:

- `review` (move this from the EditionsController)
- `publish`

And for a new edition on an existing document:

- `confirm_links`
- `schedule_or_publish`

I'll give it a go 👍

lib/engines/content_object_store/app/controllers/content_object_store/content_block/documents_controller.rb
```diff
@@ -11,4 +11,30 @@ def show
       content_block_document: @content_block_document,
     )
   end
+
+  def new
```
Harriethw, line 15: I think this comes up against convos we've had before around the relationship of Document/Edition - afaik although Documents are technically like a 'parent' class, in Whitehall everything is driven from Editions Controllers. The `New_Document_Controller` redirects to an Editions Controller once it has the type of Edition - this didn't make sense to me at first, but I've become used to it as there's so much complexity around the publishing workflow... I'd be up for talking more about it

pezholio: How about the `new` action on the DocumentsController shows the schema types, then once it's chosen, we redirect to the EditionsController, which works regardless of if a Document ID is present?

Harriethw: yeah that would make sense - there's actually a method that creates the Document if it doesn't exist https://github.com/alphagov/whitehall/blob/a19ec07326c1342fed12fa1ffff0ae76c09089f8/lib/engines/content_object_store/app/models/content_object_store/content_block/edition/documentable.rb#L22

lib/engines/content_object_store/app/controllers/content_object_store/content_block/editions_controller.rb
Harriethw: ah that's a shame that we have to create the Edition ... I guess if it does become a problem we can have a task to tidy them

lib/engines/content_object_store/app/controllers/content_object_store/content_block/documents_controller.rb
pezholio: I've got this about 90% done, but might belong better in a separate PR

Harriethw (approved):
onwards and upwards 🚀

PR #9501, opened 2024-10-08
Configure hosts for application

brucebolt:
Note: the healthcheck endpoints are requested by IP, not domain, so we need to specifically exclude them from the protection.<br><br>[Trello card](https://trello.com/c/frqFN7D8)


ChrisBAshton (approved):

brucebolt:
I've realised this allows domains like `app-name.dodgygov.uk`.  The regex will need to be updated to `app-name\..*\.gov.uk$`.

PR #9502, opened 2024-10-08
Use Editions controller when creating and updating blocks

pezholio:
This follows on from @Harriethw's comment here https://github.com/alphagov/whitehall/pull/9496#discussion_r1787902413 to simplify the creation of new content blocks by using the Editions controller, rather than the documents controller.

The flow now looks like this:

```mermaid
graph

a[Create new block] --> b[GET /content-block/new => DocumentsController#new]
b --> c[Choose block type]
c --> d[POST /content-block/new_document_options_redirect => DocumentsController#new_document_options_redirect]
d --> e["Redirect to new content block"]
e --> f["GET /content-block/editions/{block-type}/new => EditionsController#new"]
f --> g["Fill in form"]
g --> h["POST /content-block/editions => EditionsController#create"]
h --> i["Redirect to review"]

```


Harriethw (approved):
Looks good! 
I'm still a bit concerned about the unhappy path when user abandon the new/edit flow as I think that could happen a lot - do we need bug tickets to at least hide any unpublished Documents/Editions? 
Currently:
1. If I abandon the form to create a brand new Block, the Document appears on the index page but I can't click through to edit it
2. If I abandon the form to edit a Block, the Edition appears on the Document page as 'Draft' and I can't see the existing live Edition.

pezholio:
Yeah, I'll add some bug tickets for those two things 👍

pezholio:
Ticket is here https://trello.com/c/TkFPNu7U/607-hide-any-unpublished-documents-editions

PR #9503, opened 2024-10-08
(607) Hide any unpublished document's editions

pezholio:
This fixes the issue raised in https://github.com/alphagov/whitehall/pull/9502 where partially completed documents were showing on the listing page. This should fix it 👍


Harriethw (approved):
🐛 🔨

PR #9507, opened 2024-10-09
Content modelling/610 change name to Content Block Manager

Harriethw:
Here we change the name of the service from Content Object Store to Content Block Manager - and while we are here we remove `/government/admin` from the url. 

-----


lib/engines/content_block_manager/app/controllers/content_block_manager/base_controller.rb
```diff
@@ -30,16 +30,16 @@ def edition_params
   end
 
   def set_sentry_tags
-    Sentry.set_tags(engine: "content_object_store")
+    Sentry.set_tags(engine: "content_block_manager")
```
pezholio, line 33: Ooh, I'll need to tweak the Sentry rules to look for the new tag instead of the old ones to ensure we still get the Slack alerts

pezholio (approved):
LGTM 👍

PR #9508, opened 2024-10-09
Allow statistics announcement update

minhngocd:
This adds functionality to automatically update the connected publication type when statistics announcement type is being changed. At the moment, users are blocked from being able to update the statistics announcement type because of the two way validation between the publicatoin and the announcement needing to be of the same type. 

This logic assumes that once you have successfully linked a document, they always need to be in sync. Hence, when you update the statistics announcement type, it should automatically go and update the connected publication. 

We've added hint text that only shows on the edit form and not the new form:
image

https://trello.com/c/5oXMKjP9/2965-unable-to-link-official-statistics-publication-to-accredited-official-statistics

### DIFF
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -22,7 +22,7 @@ class StatisticsAnnouncement < ApplicationRecord
 
   belongs_to :creator, class_name: "User"
   belongs_to :cancelled_by, class_name: "User"
-  belongs_to :publication
+  belongs_to :publication, autosave: true
 
   has_one  :current_release_date,
            lambda {
@@ -40,8 +40,8 @@ class StatisticsAnnouncement < ApplicationRecord
 
   validates_associated :publication, if: :publication,
                                      message: lambda { |_, publication|
-                                                "type #{publication[:value].errors[:publication_type_id].first}"
-                                              }
+                                       "type #{publication[:value].errors[:publication_type_id].first}"
+                                     }
   validate  :redirect_not_circular, if: :unpublished?
   validates :publishing_state, inclusion: %w[published unpublished]
   validates :redirect_url, presence: { message: "must be provided when unpublishing an announcement" }, if: :unpublished?
@@ -95,6 +95,7 @@ class StatisticsAnnouncement < ApplicationRecord
 
   after_touch :publish_redirect_to_publication, if: :publication_has_been_published?
   set_callback :published, :after, :after_publish
+  before_validation :update_associated_publication_type, on: :update, if: :publication_type_id_changed?
   after_commit :notify_unpublished, if: :unpublished?
 
   def notify_unpublished
@@ -232,6 +233,10 @@ class StatisticsAnnouncement < ApplicationRecord
   end
 
 private
+
+  def update_associated_publication_type
+    publication.publication_type_id = publication_type_id if publication
+  end
 
   def publication_has_been_published?
     publication && publication.published?
--- a/app/models/statistics_announcement.rb
+++ b/app/models/statistics_announcement.rb
@@ -22,15 +22,19 @@ class StatisticsAnnouncement < ApplicationRecord
 
   belongs_to :creator, class_name: "User"
   belongs_to :cancelled_by, class_name: "User"
-  belongs_to :publication, autosave: true
+  belongs_to :publication, autosave: true, validate: false
+  validates_associated :publication, if: :publication,
+                                     message: lambda { |_, publication|
+                                       "type #{publication[:value].errors[:publication_type_id].first}"
+                                     }
 
-  has_one  :current_release_date,
-           lambda {
-             joins(:statistics_announcement)
-                .where("statistics_announcements.current_release_date_id = statistics_announcement_dates.id")
-           },
-           class_name: "StatisticsAnnouncementDate",
-           inverse_of: :statistics_announcement
+  has_one :current_release_date,
+          lambda {
+            joins(:statistics_announcement)
+              .where("statistics_announcements.current_release_date_id = statistics_announcement_dates.id")
+          },
+          class_name: "StatisticsAnnouncementDate",
+          inverse_of: :statistics_announcement
   has_many :statistics_announcement_dates,
            -> { order(created_at: :asc, id: :asc) },
            dependent: :destroy
@@ -38,11 +42,7 @@ class StatisticsAnnouncement < ApplicationRecord
   has_many :statistics_announcement_organisations, inverse_of: :statistics_announcement, dependent: :destroy
   has_many :organisations, through: :statistics_announcement_organisations
 
-  validates_associated :publication, if: :publication,
-                                     message: lambda { |_, publication|
-                                       "type #{publication[:value].errors[:publication_type_id].first}"
-                                     }
-  validate  :redirect_not_circular, if: :unpublished?
+  validate :redirect_not_circular, if: :unpublished?
   validates :publishing_state, inclusion: %w[published unpublished]
   validates :redirect_url, presence: { message: "must be provided when unpublishing an announcement" }, if: :unpublished?
   validates :redirect_url, uri: true, allow_blank: true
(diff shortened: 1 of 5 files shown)


app/views/admin/statistics_announcements/_form.html.erb
```diff
@@ -5,6 +5,7 @@
     heading_size: "l",
     name: "statistics_announcement[publication_type_id]",
     id: "statistics_announcement_publication_type_id",
+    hint: statistics_announcement.new_record? ? nil : "Please note that changing the statistics type will also automatically update the type of the connected document.",
```
lauraghiorghisor-tw, line 8: No please: "Changing the statistics type will automatically update the type of the connected document."

app/models/statistics_announcement.rb
```diff
@@ -22,7 +22,7 @@ def can_publish_to_publishing_api?
 
   belongs_to :creator, class_name: "User"
   belongs_to :cancelled_by, class_name: "User"
-  belongs_to :publication
+  belongs_to :publication, autosave: true
```
lauraghiorghisor-tw, line 25: Oh, interesting 🤔 -> is this needed so that the `before_validation` runs?

test/unit/app/models/statistics_announcement_test.rb
```diff
@@ -292,6 +270,107 @@ class StatisticsAnnouncementTest < ActiveSupport::TestCase
...
+    announcement = create(:statistics_announcement, publication_type_id: PublicationType::OfficialStatistics.id)
+
+    announcement.publication_type_id = PublicationType::NationalStatistics.id
+    announcement.save!
+
+    assert announcement.valid?
+    assert announcement.errors.empty?
+  end
+
+  test "should allow updating to same publication type with attached draft publication" do
```
lauraghiorghisor-tw, line 315: Feel like "same publication type *AS* attached draft publication" reads better. 
For this test, the starting type of the announcement is the same type we're changing to. Whilst the validation is still relevant, perhaps it makes more sense to start off with a non-accredited announcement for this test.

test/unit/app/models/statistics_announcement_test.rb
```diff
@@ -292,6 +270,107 @@ class StatisticsAnnouncementTest < ActiveSupport::TestCase
...
+      title: "New title",
+    )
+    announcement.save!
+
+    assert announcement.valid?
+    assert announcement.errors.empty?
+    assert_equal PublicationType::NationalStatistics.id, national_statistics.reload.publication_type_id
+  end
+
+  test "should allow updating to different publication type to attached publication and updates the draft publication" do
```
lauraghiorghisor-tw, line 330: A bit confusing again, maybe try smth like "should update to a different publication type *from* attached publication, and update the draft publication".

test/unit/app/models/statistics_announcement_test.rb
```diff
@@ -292,6 +270,107 @@ class StatisticsAnnouncementTest < ActiveSupport::TestCase
...
+      publication_type_id: PublicationType::OfficialStatistics.id,
+      title: "New title",
+    )
+
+    assert_not announcement.save
+    assert announcement.errors[:publication].include? "type does not match announcement type: must be 'Official Statistics'"
+    assert_equal PublicationType::NationalStatistics.id, national_statistics.reload.publication_type_id
+  end
+
+  test "should not allow updating Publication type with attached published publication" do
```
lauraghiorghisor-tw, line 359: "[...] Publication type, when connected publication is published"?

Maybe we can use "connected" over "attached" in all these tests 🤔

test/unit/app/models/statistics_announcement_test.rb
```diff
@@ -292,6 +270,107 @@ class StatisticsAnnouncementTest < ActiveSupport::TestCase
...
+    announcement = create(:statistics_announcement, publication_type_id: PublicationType::NationalStatistics.id, publication: national_statistics)
+
+    announcement.assign_attributes(
+      publication_type_id: PublicationType::OfficialStatistics.id,
+      title: "New title",
+    )
+
+    assert_not announcement.save
+    assert announcement.errors[:publication].include? "type cannot be modified when edition is in the published state"
+    assert_equal PublicationType::NationalStatistics.id, national_statistics.reload.publication_type_id
```
lauraghiorghisor-tw, line 370: The test is about the type of the announcement, so maybe check that too/instead.

test/functional/admin/statistics_announcements_controller_test.rb
```diff
@@ -132,6 +192,89 @@ class Admin::StatisticsAnnouncementsControllerTest < ActionController::TestCase
     assert_select "ul.govuk-error-summary__list a", text: "Title can't be blank"
   end
 
+  test "PUT :update should update attached draft publication in Publishing API if announcement publication type is updated on save" do
```
lauraghiorghisor-tw, line 195: "connected publication" 🤔 
"if announcement publication type has changed" 🤔

test/functional/admin/statistics_announcements_controller_test.rb
```diff
@@ -132,6 +192,89 @@ class Admin::StatisticsAnnouncementsControllerTest < ActionController::TestCase
...
+    assert_all_requested(expected_requests)
+  end
+
+  test "PUT :update should not update attached draft publication in Publishing API if announcement publication type is not changed on save" do
+    national_statistics = create(:draft_national_statistics, lead_organisations: [@organisation])
+    announcement = create(
+      :statistics_announcement,
+      publication_type_id: PublicationType::NationalStatistics.id,
+      publication: national_statistics,
+      organisation_ids: [@organisation.id],
```
lauraghiorghisor-tw, line 235: Not needed for the announcement to be set up with an org in either test.

app/models/statistics_announcement.rb
minhngocd: no - this is needed because Rubocop doesn't allow publication.save, only publication.save! - and that's in the middle of a transaction, so we don't really want it to be called there, as there could be other validation issues with the publication (e.g. unmodifiable states). 
autosave will also save the publication after saving the announcement and all other validations have been run, so it's at a more suitable point to save.

test/unit/app/models/statistics_announcement_test.rb
minhngocd: hmmmm we can't have that starting state anyways, that's invalid as you cannot create an announcement type different to the publication type anyways. 

This test is more about updating the announcement, but the type has not been changed, so we're sending in the same publication type (as opposed to changing it)

test/unit/app/models/statistics_announcement_test.rb
minhngocd: the announcement failed to save with `assert_not announcement.save`, so the assumption there is nothing's changed - but i can add that assertion. 
but we shouldn't do instead because the point of the test is to test that should the validation fail and announcement fails to save - it should also not save the publication, hence publication type for the connected document needs to be the same

app/views/admin/statistics_announcements/_form.html.erb
minhngocd: I guess we're waiting for content design on this one

test/unit/app/models/statistics_announcement_test.rb
lauraghiorghisor-tw: I thought this was the case that covers the "stuff got mixed up and now we're in the weird different type state" and we're trying to make the types the same, i.e. our zendesk ticket that started it all. 
Fair enough that we cannot do that on create, but then let's use update and replicate this scenario?

test/unit/app/models/statistics_announcement_test.rb
lauraghiorghisor-tw: My thinking was the same, just to show that the type of the announcement has stayed the same. Going back to the test namings... the test is about the announcement type not the publication's type. Or at least that's how I am reading these. There is one test that says "and updates the draft publication" - that's the only one that calls out that we care about the publication type.

test/unit/app/models/statistics_announcement_test.rb
minhngocd: but with your changes from the last PR, that scenario is just impossible. 

how to create mismatched models
- creation: cannot create with mismatched type because now validation is surfaced. also tested within this class
- update: that's what these suite of tests are testing, is that if you update it will automatically sync the types

so you will never get a starting state of mismatched types

test/unit/app/models/statistics_announcement_test.rb
minhngocd: sure - i can add the publication check to the test name

PR #9509, opened 2024-10-09
(613) Publish to draft content store when creating a draft

pezholio:
This updates the `CreateEditionService` to send an edition to the draft content store whenever a new edition is created. This will make previewing possible once we've decided how to handle it, as well as keep parity between what is in Whitehall and what is in the Publishing API.

We still have the issue of what happens if a user changes their mind and keeps something in draft, but I'm hoping that we can address this in this card - https://trello.com/c/fySdBC34/612-remove-draft-editions-on-cancellation

### DIFF
--- a/lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
+++ b/lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
@@ -41,6 +41,28 @@ module ContentObjectStore
       end
     end
 
+    def create_draft_edition(schema)
+      raise ArgumentError, "Local database changes not given" unless block_given?
+
+      ActiveRecord::Base.transaction do
+        content_block_edition = yield
+        content_id = content_block_edition.document.content_id
+        organisation_id = content_block_edition.lead_organisation.content_id
+
+        create_publishing_api_edition(
+          content_id:,
+          schema_id: schema.id,
+          title: content_block_edition.title,
+          details: content_block_edition.details.to_h,
+          links: {
+            primary_publishing_organisation: [
+              organisation_id,
+            ],
+          },
+        )
+      end
+    end
+
     def update_content_block_document(new_content_block_edition:, update_document_params:)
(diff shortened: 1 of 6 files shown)


lib/engines/content_object_store/test/factories/content_block_edition.rb
```diff
@@ -6,7 +6,7 @@
     schema { build(:content_block_schema) }
     creator
 
-    organisation { FactoryBot.create(:organisation) }
+    organisation { FactoryBot.build(:organisation) }
```
Harriethw, line 9: oh nice idea, I remember this causing me a headache before

pezholio: Yeah, it took me a while to work out why the code was calling out to the Publishing API twice!

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
```diff
@@ -41,6 +41,28 @@ def schedule_with_rollback
...
 
+    def create_draft_edition(schema)
+      raise ArgumentError, "Local database changes not given" unless block_given?
+
+      ActiveRecord::Base.transaction do
+        content_block_edition = yield
+        content_id = content_block_edition.document.content_id
+        organisation_id = content_block_edition.lead_organisation.content_id
+
+        create_publishing_api_edition(
```
Harriethw, line 52: do we still need to call `create_publishing_api_edition` up above in `publish`? Because we might be passing in some new updated `details`/`title` when we publish?

Harriethw (approved):
👍

lib/engines/content_object_store/app/lib/content_object_store/publishable.rb
pezholio: Oooh, that's a good point actually. I think we should still push a draft up before publishing, but we probably don't need those args there now.

PR #9515, opened 2024-10-11
content-modelling/fix: send schema ID not block type to Publishing API

Harriethw:
previously using `block_type` just sent `email_address` and was not recognised by the API, because we need to include
`content_block` in front of it. I thiink the ID was designed to be sent to the Publishing API and `block_type` is for internal use.


pezholio (approved):
Good spot! 😅

PR #9521, opened 2024-10-14
Content modelling/612 remove cancelled draft editions

Harriethw:
When a user clicks 'cancel' on any page in the Content Block forms
Then any new draft Editions created by that form should be destroyed
And any orphaned Documents

Creating New Content Block (review page)
Screenshot 2024-10-14 at 15 04 01

Reviewing links for existing Content Block (review_links page)
Screenshot 2024-10-14 at 14 58 19

Schedule or Publish new Edition for existing Content Block (schedule_publishing page)
Screenshot 2024-10-14 at 14 58 30

------

### DIFF
--- a/lib/engines/content_block_manager/app/services/content_block_manager/delete_edition_service.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/delete_edition_service.rb
@@ -1,0 +1,23 @@
+module ContentBlockManager
+  class DeleteEditionService
+    def call(content_block_edition)
+      if content_block_edition.draft?
+        document = content_block_edition.document
+        document.with_lock do
+          content_block_edition.destroy!
+          if document_has_no_more_editions?(document)
+            document.destroy!
+          end
+        end
+      else
+        raise ArgumentError, "Could not delete Content Block Edition #{content_block_edition.id} because it is not in draft"
+      end
+    end
+
+  private
+
+    def document_has_no_more_editions?(document)
+      document.editions.count.zero?
+    end
+  end
+end
--- a/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review.html.erb
+++ b/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review.html.erb
@@ -20,14 +20,21 @@
   <div class="govuk-grid-column-full">
     <%= form_with(url: content_block_manager.content_block_manager_content_block_workflow_path(@content_block_edition, step: ContentBlockManager::ContentBlock::Editions::WorkflowController::NEW_BLOCK_STEPS[:review]), method: :put) do %>
       <div class="govuk-button-group govuk-!-margin-bottom-6">
-        <%= render "govuk_publishing_components/components/button", {
-          text: "Accept and publish",
-          name: "accept_and_publish",
-          value: "Accept and publish",
-          type: "submit",
-        } %>
-        <%= link_to("Cancel", content_block_manager.content_block_manager_content_block_documents_path, class: "govuk-link") %>
+        <div>
+          <%= render "govuk_publishing_components/components/button", {
+            text: "Accept and publish",
+            name: "accept_and_publish",
+            value: "Accept and publish",
+            type: "submit",
+          } %>
+      <% end %>
+        </div>
+        <div>
+          <%= render partial: "content_block_manager/content_block/shared/cancel_delete_button",
+                     locals: {
+                       url:  content_block_manager.content_block_manager_content_block_edition_path(@content_block_edition),
+                     } %>
+          </div>
       </div>
-    <% end %>
   </div>
 </div>
--- a/lib/engines/content_block_manager/app/views/content_block_manager/content_block/shared/_cancel_delete_button.erb
+++ b/lib/engines/content_block_manager/app/views/content_block_manager/content_block/shared/_cancel_delete_button.erb
@@ -1,0 +1,9 @@
+<%= form_with url: url, method: :patch do %>
+  <%= render "govuk_publishing_components/components/button", {
+    text: "Cancel",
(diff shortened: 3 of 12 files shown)


lib/engines/content_block_manager/features/create_object.feature
```diff
@@ -52,6 +52,18 @@ Feature: Create a content object
...
+  Scenario: GDS editor cancels the creation of an object
+    When I visit the Content Block Manager home page
+    And I click to create an object
+    When I click on the "email_address" schema
+    When I complete the form with the following fields:
+      | title            | email_address   | department | organisation |
+      | my email address | foo@example.com | Somewhere  | Ministry of Example |
+    And I click cancel
+    Then I am taken back to Content Block Manager home page
+    And no draft Content Block Edition has been created
```
Harriethw, line 64: I'm diverting a little here from feature tests just testing changes that are visible to the user, but I'm not sure that just using integration tests would give us the coverage needed - I want to be sure that the deletion has happened when the user clicks this button

lib/engines/content_block_manager/app/services/content_block_manager/delete_edition_service.rb
```diff
@@ -0,0 +1,23 @@
+module ContentBlockManager
+  class DeleteEditionService
+    def call(content_block_edition)
+      if content_block_edition.draft?
+        document = content_block_edition.document
+        document.with_lock do
+          content_block_edition.destroy!
+          if document_has_no_editions(document)
```
pezholio, line 8: Should this be `document_has_no_more_editions?` (I tend to prefer appending a `?` on the end of methods that return a boolean)

lib/engines/content_block_manager/app/views/content_block_manager/content_block/shared/_cancel_delete_button.erb
```diff
@@ -0,0 +1,9 @@
+<%= form_with url: url, method: :patch do %>
+  <%= render "govuk_publishing_components/components/button", {
+    text: "Cancel",
+    name: "_method",
```
pezholio, line 4: Could we have a hidden `_method` field within the form instead? Adding a name/value confused me for a sec

lib/engines/content_block_manager/features/create_object.feature
pezholio: Yeah, I think that's cool 👍

lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review.html.erb
```diff
@@ -20,14 +20,21 @@
...
     <%= form_with(url: content_block_manager.content_block_manager_content_block_workflow_path(@content_block_edition, step: ContentBlockManager::ContentBlock::Editions::WorkflowController::NEW_BLOCK_STEPS[:review]), method: :put) do %>
       <div class="govuk-button-group govuk-!-margin-bottom-6">
-        <%= render "govuk_publishing_components/components/button", {
-          text: "Accept and publish",
-          name: "accept_and_publish",
-          value: "Accept and publish",
-          type: "submit",
-        } %>
-        <%= link_to("Cancel", content_block_manager.content_block_manager_content_block_documents_path, class: "govuk-link") %>
+        <div>
```
pezholio, line 23: Do we need these `<div>`s here?

pezholio (approved):
LGTM - just a couple of teeny non-blocking things

lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review.html.erb
Harriethw: I think so for the `govuk-button-group` to align the items correctly, because I've had to put that `<% end %>` of the form in before the Cancel button - otherwise the second button goes wonky.

pezholio: Aaah, makes sense

lib/engines/content_block_manager/app/services/content_block_manager/delete_edition_service.rb
Harriethw: good idea! renamed and added ?

lib/engines/content_block_manager/app/views/content_block_manager/content_block/shared/_cancel_delete_button.erb
Harriethw: hmm I spoke to soon cause now I'm getting issues with the controllers again! will take a look tomorrow

Harriethw: so when I have a `delete` method on the form, or putting in a `hidden` input like below, I get a routing error on the feature test: `ActionController::RoutingError: No route matches [DELETE] {whatever path it is on at the time}` seems like it isn't taking on the `url` that I'm passing into it, or maybe having two forms on the page is conflicting 🤔 
If it isn't a major issue at the moment I will pop the issue in the commit message for now.
`<input type="hidden" name="_method" value="delete" autocomplete="off">`

pezholio: Ah, I wonder if the issue is because we have two nested forms? I seem to remember that being an issue in the past

pezholio: I wonder whether you could get rid of the nested form and go back to having the `name` and `_method` on the button - after all, sending a `DELETE` in browser is just a `POST` request that's faked by sending a `_method=DELETE` param

PR #9522, opened 2024-10-15
add rake task to delete any Draft Content Blocks

Harriethw:
We don't currently have any UX for using Content
Blocks in a draft state, but they are
automatically created if a user hits 'save and
continue' on the `new` form page - this is a rake
task to delete any Blocks that are hanging around
from people abandoning the form without pressing
'cancel' (e.g. by closing the browser window)


lib/engines/content_block_manager/lib/tasks/delete_draft_content_blocks.rake
```diff
@@ -0,0 +1,12 @@
+namespace :content_block_manager do
+  desc "Delete draft Content Block Editions and any orphaned Content Block Documents"
+  task delete_draft_content_blocks: :environment do
+    draft_content_editions = ContentBlockManager::ContentBlock::Edition.where(state: "draft")
```
pezholio, line 4: I wonder if we should only fetch draft editions that are, say 2 days old? If we run this as it stands, we risk deleting in-progress applications. Something like:

```suggestion
    draft_content_editions = ContentBlockManager::ContentBlock::Edition.where(state: "draft").where("updated_at < ?", 2.days.ago)
```

pezholio (changes requested):

pezholio:
Thinking about this, I think this might be better as a Sidekiq Job, which we can then add as a scheduled job using the [`schedule` options in Sidekiq.yml](https://github.com/alphagov/whitehall/blob/890aca554c008b76fc318c77c77a8611d864fda1/config/sidekiq.yml#L13), that way we don't have to do any infrastructure work to schedule the task.

PR #9527, opened 2024-10-15
schedule draft Content Block Editions to be deleted

Harriethw:
We don't currently have any UX for using Content
Blocks in a draft state, but they are
automatically created if a user hits 'save and
continue' on the new form page - this schedules a
Sidekiq job to delete any Blocks that are hanging
around from people abandoning the form without
pressing 'cancel' (e.g. by closing the browser
window) and that are older than two months (so we
avoid deleting any in progress blocks).

### DIFF
--- a/lib/engines/content_block_manager/app/workers/content_block_manager/delete_draft_content_blocks_worker.rb
+++ b/lib/engines/content_block_manager/app/workers/content_block_manager/delete_draft_content_blocks_worker.rb
@@ -1,0 +1,17 @@
+require "sidekiq/api"
+
+module ContentBlockManager
+  class DeleteDraftContentBlocksWorker < WorkerBase
+    sidekiq_options queue: :content_block_publishing
+
+    def perform
+      draft_content_editions = ContentBlockManager::ContentBlock::Edition.draft.where("created_at < ?", 60.days.ago).limit(100)
+      puts "Starting to delete #{draft_content_editions.count} draft Content Blocks"
+      draft_content_editions.each do |draft_content_edition|
+        puts "Deleting draft Content Block Edition #{draft_content_edition.id}"
+        ContentBlockManager::DeleteEditionService.new.call(draft_content_edition)
+      end
+      puts "Finished deleting draft Content Blocks"
+    end
+  end
+end
--- a/lib/engines/content_block_manager/test/unit/app/workers/delete_draft_content_blocks_worker_test.rb
+++ b/lib/engines/content_block_manager/test/unit/app/workers/delete_draft_content_blocks_worker_test.rb
@@ -1,0 +1,49 @@
+require "test_helper"
+
+class ContentBlockManager::DeleteDraftContentBlocksWorkerTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+  include SidekiqTestHelpers
+
+  # Suppress noisy Sidekiq logging in the test output
+  setup do
+    Sidekiq.configure_client do |cfg|
+      cfg.logger.level = ::Logger::WARN
+    end
+  end
+
+  describe "#perform" do
+    it "deletes draft content block editions older than 60 days" do
+      document = create(:content_block_document, :email_address)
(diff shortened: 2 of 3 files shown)


lib/engines/content_block_manager/app/workers/content_block_manager/delete_draft_content_blocks_worker.rb
```diff
@@ -0,0 +1,17 @@
+require "sidekiq/api"
+
+module ContentBlockManager
+  class DeleteDraftContentBlocksWorker < WorkerBase
+    sidekiq_options queue: :content_block_publishing
+
+    def perform
+      draft_content_editions = ContentBlockManager::ContentBlock::Edition.where(state: "draft").where("created_at < ?", 60.days.ago)
```
Harriethw, line 8: wonder if there's a nicer way of writing this where clause?

pezholio: I think we can do:

```
ContentBlockManager::ContentBlock::Edition.draft.where("created_at < ?", 60.days.ago)
```

lib/engines/content_block_manager/app/workers/content_block_manager/delete_draft_content_blocks_worker.rb
```diff
@@ -0,0 +1,17 @@
...
+      puts "Starting to delete #{draft_content_editions.count} draft Content Blocks"
+      draft_content_editions.each do |draft_content_edition|
+        puts "Deleting draft Content Block Edition #{draft_content_edition.id}"
+        ContentBlockManager::DeleteEditionService.new.call(draft_content_edition)
```
pezholio, line 12: I did wonder if we could amend `DeleteEditionService` to handle deleting editions in bulk, but I don't think it's really worth doing, as we'll likely only be dealing with a handful of editions at a time.

pezholio: Maybe limit it to 100?

Harriethw: cool, I've added `limit` to the end of the query... but testing it by creating 101 records in a unit test is really slow :0 any ideas how that might be speeded up?

pezholio: I don't think it's worth creating all those records. We should trust that ActiveRecord does its thing correctly.

lib/engines/content_block_manager/test/unit/app/workers/delete_draft_content_blocks_worker_test.rb
```diff
@@ -0,0 +1,52 @@
...
+      delete_service_mock = Minitest::Mock.new
+      ContentBlockManager::DeleteEditionService.expects(:new).returns(delete_service_mock).at_least_once
+      delete_service_mock.expect :call, nil, [draft_editions_older_than_61_days[0]]
+      delete_service_mock.expect :call, nil, [draft_editions_older_than_61_days[1]]
+
+      ContentBlockManager::DeleteDraftContentBlocksWorker.new.perform
+      delete_service_mock.verify
+    end
+
+    it "will take the first 100 drafts to delete" do
```
pezholio, line 30: It might be worth structuring this as a classic (i.e. no database) unit test, so we document the behaviour. Something like:

```ruby
ar_mock = mock
ContentBlockManager::ContentBlock::Edition.expects(:draft).returns(ar_mock)
ar_mock.stubs(:where).with("created_at < ?", 60.days.ago).returns(ar_mock)
ar_mock.expects(:limit).with(100)
```

Harriethw: nice one, have done 👍

pezholio (approved):
Nice work 👍

PR #9530, opened 2024-10-16
Increase maximum featured documents for No. 10 to 7

KludgeKML:
https://trello.com/c/ey4KM89d/59-make-changes-to-no10-organisation-page

## Before
Screenshot 2024-10-16 at 14 15 53

## After
Screenshot 2024-10-16 at 14 15 41

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -112,6 +112,12 @@ class Organisation < ApplicationRecord
 
   has_many :offsite_links, as: :parent
 
+  def featured_documents_display_limit
+    return 7 if slug == "prime-ministers-office-10-downing-street"
+
+    FEATURED_DOCUMENTS_DISPLAY_LIMIT
+  end
+
   # I'm trying to use a domain centric design rather than a persistence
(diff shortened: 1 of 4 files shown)


app/models/organisation.rb
```diff
@@ -112,6 +112,12 @@ class Organisation < ApplicationRecord
 
   has_many :offsite_links, as: :parent
 
+  def featured_documents_display_limit
```
leenagupte, line 115: I'd have thought this method would need to be updated: [featured_documents_display_limit](https://github.com/alphagov/whitehall/blob/06a1c9e2bbbc31a0b34807f8b8b8a167b3272c53/app/models/organisation.rb#L510) ?

leenagupte (commented):
I have an inline question.

app/models/organisation.rb
KludgeKML: Lol, that's what Ashley thought as well. Featured links and featured documents are two separate things.

leenagupte: Oh! That's my fault, Ash got that from my investigations. Oops!

PR #9535, opened 2024-10-17
Send `last_edited_by_editor_id` to Publishing API

pezholio:
This is a new field that will allow us to identify the last person who has edited an Edition within the Publishing API.

Rather than adding the field to each individual Edition Presenter, I've added a PayloadBuilder class that fetches the ID from the `last_author` attribute if it exists, and is not nil. 

This DRYs up the presenters, which were getting to be a bit repetitive, and also gives us the flexibility to get users from other attributes for other types of item, if we choose to in the future.


ChrisBAshton (approved):
LGTM!

PR #9543, opened 2024-10-18
Content modelling/626 search by title

Harriethw:
As an editor
When I enter keywords from a block title
Then I should see the relevant blocks

Screenshot 2024-10-18 at 15 37 10

--- 

### DIFF
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
@@ -1,0 +1,13 @@
+module ContentBlockManager
+  module ContentBlock::Document::Scopes::SearchableByTitle
+    extend ActiveSupport::Concern
+
+    included do
+      scope :with_title,
+            lambda { |*keywords|
+              pattern = "(#{keywords.map { |k| Regexp.escape(k) }.join('|')})"
+              where("title REGEXP :pattern", pattern:)
+            }
+    end
+  end
+end
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
@@ -4,8 +4,13 @@ module ContentBlockManager
 
     included do
       scope :with_title,
-            lambda { |*keywords|
-              pattern = "(#{keywords.map { |k| Regexp.escape(k) }.join('|')})"
+            lambda { |keywords|
+              split_keywords = keywords.split
+              pattern = ""
+              split_keywords.map do |k|
+                escaped_word = Regexp.escape(k)
+                pattern += "(?=.*#{escaped_word})"
+              end
               where("title REGEXP :pattern", pattern:)
             }
     end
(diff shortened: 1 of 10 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
```diff
@@ -4,8 +4,13 @@ module ContentBlock::Document::Scopes::SearchableByTitle
 
     included do
       scope :with_title,
-            lambda { |*keywords|
-              pattern = "(#{keywords.map { |k| Regexp.escape(k) }.join('|')})"
+            lambda { |keywords|
+              split_keywords = keywords.split
+              pattern = ""
+              split_keywords.map do |k|
```
Harriethw, line 10: not sure if this is better user experience or not! but it's how i would expect a keyword search to work elsewhere.
also might be a neater way of writing but wanted it to be readable

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
```diff
@@ -0,0 +1,13 @@
+module ContentBlockManager
+  module ContentBlock::Document::Scopes::SearchableByTitle
+    extend ActiveSupport::Concern
+
+    included do
+      scope :with_title,
```
Harriethw, line 6: I've adjusted this function in the optional last commit which diverges from Whitehall way of matching

lib/engines/content_block_manager/test/unit/app/models/content_block_edition/document/document_filter_test.rb
```diff
@@ -0,0 +1,28 @@
+require "test_helper"
+
+class ContentBlockManager::DocumentFilterTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
+  describe "documents" do
+    describe "when no filters are given" do
+      it "returns live documents" do
```
pezholio, line 8: I wonder if it might be better if we unit test this functionality, rather than hitting the database directly. So rather than checking what's returned, we just stub `ContentBlock::Document` and check the right calls are made to it?

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_title.rb
pezholio: Yeah, I think this is good - makes sense to be more flexible

pezholio (approved):

lib/engines/content_block_manager/app/views/content_block_manager/content_block/documents/index.html.erb
```diff
@@ -11,9 +11,13 @@
...
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
-  <% @content_block_documents.each do |content_block_document| %>
-    <%= render ContentBlockManager::ContentBlock::Document::Index::SummaryCardComponent.new(content_block_document:) %>
-  <% end %>
+  <div class="govuk-grid-column-one-third">
+    <%= render "filter_options" %>
+  </div>
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-m"><%= @content_block_documents.count %> results</h2>
```
Harriethw, line 18: ooh I forgot to `pluralize` this -note to myself to go back to it

PR #9549, opened 2024-10-22
search by keyword in details and title of Content Block

Harriethw:
Screenshot 2024-10-22 at 16 42 28

This now matches keywords agains the user generated free text on a block. Select options will be added for Organisation, date etc.

If any other free text options were added (such as `internal instructions`, `description`) we'd need to
add them to this keyword match here too.

---


lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_keyword.rb
```diff
@@ -1,17 +1,18 @@
...
+      scope :with_keyword,
             lambda { |keywords|
               split_keywords = keywords.split
               pattern = ""
               split_keywords.map do |k|
                 escaped_word = Regexp.escape(k)
                 pattern += "(?=.*#{escaped_word})"
               end
-              where("title REGEXP :pattern", pattern:)
+              joins("JOIN content_block_editions ON content_block_documents.latest_edition_id = content_block_editions.id")
```
Harriethw, line 14: I've spent way too long today trying to get this `join` to work in a fancier way... Ideally I would be able to re-use the `latest_edition` association on a Document (by doing `includes`?) but have not figured out a way that works

pezholio: I've had a look and this seems to work:

```suggestion
              joins(:latest_edition)
                         .where("content_block_documents.title REGEXP :pattern OR content_block_editions.details REGEXP :pattern", pattern:)
```

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_keyword.rb
```diff
@@ -1,17 +1,18 @@
...
+  module ContentBlock::Document::Scopes::SearchableByKeyword
     extend ActiveSupport::Concern
 
     included do
-      scope :with_title,
+      scope :with_keyword,
             lambda { |keywords|
               split_keywords = keywords.split
               pattern = ""
               split_keywords.map do |k|
```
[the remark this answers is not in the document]
Harriethw: ah yep it should be `each`

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_keyword.rb
Harriethw: ooh great this does work! 🥳 
I'm now realising though that it also matches the keys in the JSON for the `details` - so e.g. searching for `email address` will return all email addresses - I don't think this is a huge problem as user's would be unlikely to put in generic terms like that? But I might spend a little bit more time tinkering to see what if we could only match the values in the JSON blob

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_keyword.rb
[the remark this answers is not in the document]
Harriethw: or yep we can assign it directly to `pattern` as you've done there!

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_keyword.rb
Harriethw: ok having had a bit of a dig around, I think that as the `details` column is `json` rather than `jsonb` and we won't know the structure of the json blob, searching only the values could get a bit expensive (from what I understand we'd have to cast it to [jsonb and then recursively search all the paths?](https://www.postgresql.org/docs/current/functions-json.html)) - will play back to the team today in collab time.

pezholio: I wonder if we could have some kind of indexable content field that we populate on save, which has the title and only the keys of the JSON field stored in it?

Harriethw: I think if it were `jsonb` we would be able to do fancier things on it and also index fields? https://www.postgresql.org/docs/current/datatype-json.html#JSON-INDEXING 
the team were happy with how the search worked when I demo'd it, would it become expensive to change if we wait until later to look at this if it's needed?

pezholio (approved):
LGTM 👍

PR #9555, opened 2024-10-24
Allow setting of base path of landing_pages

KludgeKML:
- Landing pages' documents' slugs are actually base paths, so they _must_ start with a / - other slugs must not. So we create a specialist repathing class that subclasses DocumentReslugger and overrides the errors and editorial remark, then allow the edition slugger UI to use that. The UI will still refer to slugs, but people using the UI will understand that at the moment.

https://trello.com/c/U0Z4s3OW/113-investigate-changing-slugs-in-whitehall

### DIFF
--- a/app/controllers/admin/edition_slug_controller.rb
+++ b/app/controllers/admin/edition_slug_controller.rb
@@ -7,7 +7,7 @@ class Admin::EditionSlugController < Admin::BaseController
   def edit_slug; end
 
   def update_slug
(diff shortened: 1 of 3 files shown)


richardTowers (approved):
Yeah, I think this looks good. Can't think of a much nicer way to do it.

I suppose you could have overridden `run!` and not called the `*_search_index` methods, but then I guess you'd miss out on future changes to the parent method, so I think this is better.

I might have stuck with LandingPageReslugger, rather than LandingPageRepather, but I don't think it's important.

PR #9556, opened 2024-10-24
Make republishable_content_types less clever

richardTowers:
Using `Edition.descendants` makes some assumptions about the inheritance tree of Edition which might not always be true.

Specifically, at the moment the only classes that get published are leaf nodes in the tree - classes without any descendants. In another branch, I would like to make a new LandingPage class which inherits DocumentCollection. Both LandingPage and DocumentCollection should be publishable, so they should both appear in the republishable documents list.

I don't think there's a neat way to account for this, and we have to hardcode the lists of document types in the tests anyway, so I think it's probably fine to just... hardcode them in the production code.

This also removes the need for us to keep calling eager_load in the tests, which was a bit smelly because it indicates that the production code would give incorrect results if the classes in question hadn't been loaded for some reason.

I've also made non_editionable_content_types hardcode its list, even though the heuristic it's using seems to be more accurate / less likely to break with changes to the inheritance structure.

I've removed the non_editionable_content_types test, because with both sides hardcoded it's a bit meaningless.

### DIFF
--- a/app/helpers/admin/republishing_helper.rb
+++ b/app/helpers/admin/republishing_helper.rb
@@ -68,11 +68,23 @@ module Admin::RepublishingHelper
   end
 
   def republishable_content_types
-    editionable_content_types = Edition.descendants.select { |descendant|
-      descendant.descendants.count.zero?
-    }.map(&:to_s)
+    editionable_content_types = %w[
+      CallForEvidence
+      CaseStudy
+      Consultation
+      CorporateInformationPage
+      DetailedGuide
+      DocumentCollection
+      WorldwideOrganisation
+      FatalityNotice
+      LandingPage
+      NewsArticle
+      Publication
+      Speech
+      StatisticalDataSet
+    ]
 
-    [editionable_content_types, non_editionable_content_types].flatten.sort
+    [*editionable_content_types, *non_editionable_content_types].sort
   end
 
   def republishable_content_types_select_options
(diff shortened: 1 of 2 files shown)


app/helpers/admin/republishing_helper.rb
```diff
@@ -68,11 +68,23 @@ def bulk_content_type_metadata
   end
 
   def republishable_content_types
-    editionable_content_types = Edition.descendants.select { |descendant|
-      descendant.descendants.count.zero?
-    }.map(&:to_s)
+    editionable_content_types = %w[
```
yndajas, line 71: **thought:** maintainability

This was always a bit meh. The benefit here - assumptions aside - was that if a new type gets added, the developer who adds it doesn't need to remember to add it to the list(s) of republishable content types, but maybe we can just deal with missing types via support requests if and when they come up?

I encountered a different issue with `.descendants` in some of the GraphQL work, but we're considering still including it in test land to ensure we keep a hard-coded list up-to-date: I suppose we could do something similar here to ensure any new leaf nodes get added to this list, but then we'd still not be checking for new non-leaf republishable descendant nodes, so maybe it's too half-baked to be worth it

yndajas (commented):
Not reviewed this in detail. Just some quick thoughts/context!

---

The labels/first lines of each of the comments here are following the [Conventional Comments](https://conventionalcomments.org/) format, which I'm trying out after watching [a useful DevOpsDays London talk](https://www.youtube.com/watch?v=5TnNlfPU7kY)

app/helpers/admin/republishing_helper.rb
richardTowers: Done in  331793d

yndajas (approved):
Seems like an improvement to me! Is it worth squashing the two commits (not sure if you were already planning that)?

richardTowers:
Yeah, I'll squash the commits (I wouldn't have thought to do that, but makes sense).

Thanks for the review on this - I wasn't super happy with my original change, and I'm much more happy with where we got to in the end. :corporate_fist_bump:

Gweaton (approved):
Nothing to add over what Ynda has said, this looks good.

Optional suggestion, though: I'd squash that second commit into the first so we don't have a failing build on a commit. You may have planned to do that anyway, and left that extra commit for ease of review - in which case, ignore me!

PR #9564, opened 2024-10-28
Extend keyword landing pages

KludgeKML:

### DIFF
--- a/app/presenters/publishing_api/landing_page_presenter.rb
+++ b/app/presenters/publishing_api/landing_page_presenter.rb
@@ -47,8 +47,13 @@ module PublishingApi
     attr_reader :item
 
     def details
-      YAML.load(item.body, permitted_classes: [Date])
-        .merge(PayloadBuilder::Attachments.for(item))
+      body = YAML.load(item.body, permitted_classes: [Date])
+              .merge(PayloadBuilder::Attachments.for(item))
+      extends_slug = body.delete("extends")
+      return body unless extends_slug
+
+      extends = Document.find_by(slug: extends_slug).latest_edition
+      YAML.load(extends.body, permitted_classes: [Date]).merge(body)
     end
   end
 end
--- a/app/models/landing_page.rb
+++ b/app/models/landing_page.rb
@@ -34,6 +34,9 @@ private
     unless body_hash.keys.include?("blocks")
       errors.add(:body, "must contain a root element 'blocks:'")
     end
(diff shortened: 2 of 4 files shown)


app/presenters/publishing_api/landing_page_presenter.rb
```diff
@@ -47,7 +47,11 @@ def document_type
     attr_reader :item
 
     def details
-      YAML.load(item.body, permitted_classes: [Date])
+      body = YAML.load(item.body, permitted_classes: [Date])
+      return body unless body.key?("extends")
+
+      extends = Document.find_by(slug: body.delete("extends")).latest_edition
+      YAML.load(extends.body, permitted_classes: [Date]).merge(body)
```
richardTowers, line 54:
```suggestion
      extends_slug = body["extends"]
      return body if extends_slug.nil?

      extends = Document.find_by(slug: extends_slug).latest_edition
      YAML.load(extends.body, permitted_classes: [Date]).merge(body)
```

richardTowers (commented):
I think this approach should be fine. It's a bit of a hack on top of a hack, but it should work at least.

We already discussed in Slack that the documents won't keep themselves up to date (i.e. if document /ford extends document /car, and /car is updated, /ford won't be updated automatcially). This could be fixed with callbacks, but it's probably more pragmatic to remember to republish things manually when the menu changes.

Longer term, it would be nicer if we could find some approach that uses publishing-api's links concept to keep things in sync automatically, but we've explored document collections at some length, and they don't really work for hierarchical menus as things are at the moment.

app/presenters/publishing_api/landing_page_presenter.rb
KludgeKML: Just to double-check - this does simplify, but also changes the behaviour so that we publish the extends keyword, is that what you want?

richardTowers: That is _not_ what I want, no. I should have looked more closely before assuming you were using `.delete` just to be weird. Sorry 😅 !

I think your implementation is perfectly good in that case. Or maybe with `extends_slug = body.delete("extends")` (so it's got its own line, and it's not hidden inside the `Document.find_by()` call)
```

KludgeKML: Makes sense - I've moved it out onto its own line.

richardTowers (approved):

PR #9569, opened 2024-10-30
Support reverting "foreign language only" mode

ChrisBAshton:
We had a [support ticket on Zendesk](https://govuk.zendesk.com/agent/#/tickets/5956584) recently: a user created an English document, then accidentally set it to "Foreign language only" mode when trying to add a Welsh translation. The English version became unavailable, and they were unable to switch back to English mode - the document was stuck in "foreign language only" mode.

This PR tweaks the controller logic. The `edition[create_foreign_language_only]` field associated with the "Create a foreign language only {format}" should always be included in the form submission as a value of `1` if it is checked. The checking of the box conditionally reveals the list of locale options, which is already set to the previously set locale (`cy`, in this case).

If the "Create a foreign language only {format}" box is unchecked, then `edition[create_foreign_language_only]` is no longer submitted in the form parameters. The underlying `edition[primary_locale]` of `cy` is still passed though, since there is no logic to unset it. So the `primary_locale` was impossible to change back to English. (English is also not an option on the list of locales, so the user couldn't even 'hack' their way back to the default state).

On the basis that a non-English primary locale will only ever be accompanied by the `edition[create_foreign_language_only]` field, we can safely check for the absence of said field, and if it is absent, we can assume the primary locale should revert to English.

It did cross my mind that a developer might rename or delete the checkbox, so I've made the feature test a bit more verbose to explicitly check that the "Foreign language only" box is checked when the primary locale has been changed.

Trello: https://trello.com/c/5B60uybX/3062-unable-to-uncheck-the-create-a-foreign-language-only-box

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -101,29 +101,36 @@ class Admin::EditionsController < Admin::BaseController
   end
 
   def update
-    @edition.assign_attributes(edition_params)
+    ActiveRecord::Base.transaction do
+      @edition.assign_attributes(edition_params)
 
-    if updater.can_perform? && @edition.save_as(current_user)
-      updater.perform!
+      if updater.can_perform? && @edition.save_as(current_user)
+        updater.perform!
+        # TODO: dynamic locale, and check that there are no other locales other than "en"
+        if @remove_non_english_translation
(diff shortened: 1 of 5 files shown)


ChrisBAshton:
Closed in favour of #9570

PR #9570, opened 2024-10-31
Support reverting "foreign language only" mode by choosing English

ChrisBAshton:
This has been successfully tested on Integration 👍 

We had a [support ticket on Zendesk](https://govuk.zendesk.com/agent/#/tickets/5956584) recently: a user created an English document, then accidentally set it to "Foreign language only" mode when trying to add a Welsh translation. The English version became unavailable, and they were unable to switch back to English mode - the document was stuck in "foreign language only" mode.

Whitehall has [three varying levels of support](https://docs.google.com/document/d/1ma5gzJU8EUl3Yo6J6z3XzkEdpqUVLPLPEoSPff9ZduE/edit) for languages:

1. No support (e.g. for Calls for Evidence)
2. Foreign language only (can change primary locale to something other than English, but not provide any other translations)
3. Translatable (can add any number of translations, but not change the primary locale to anything other than English)

There therefore never should be a situation where a publisher has a document with a non-English primary locale AND any translations. Therefore we shouldn't be breaking any existing workflows by triggering the deletion of non-English translations under these circumstances.

NB, I did consider not adding a "English" option to the list of languages, and instead figure out the user's intention from the absence of the checked "foreign language only" checkbox. But this felt rather unsafe - all it would take is for a developer to rename that checkbox and then the absence check would always return `true` and we could find ourselves in a place where non-English documents get mistakenly converted to English ones. See #9569 for the approach if interested.

Trello: https://trello.com/c/5B60uybX/3062-unable-to-uncheck-the-create-a-foreign-language-only-box

### DIFF
--- a/app/models/concerns/edition/translatable.rb
+++ b/app/models/concerns/edition/translatable.rb
@@ -73,7 +73,7 @@ private
   end
 
   def remove_other_translations_if_primary_locale_changed
-    return unless saved_change_to_primary_locale?
+    return unless saved_change_to_primary_locale? && translations.count == 2
 
     translations.each do |translation|
       translation.destroy! unless translation.locale.to_s == primary_locale
--- a/app/models/concerns/edition/translatable.rb
+++ b/app/models/concerns/edition/translatable.rb
@@ -25,7 +25,7 @@ module Edition::Translatable
       on: :update,
     )
     before_validation(
-      :remove_other_translations_if_primary_locale_no_longer_english,
+      :remove_other_translations_if_primary_locale_changed,
       on: :update,
     )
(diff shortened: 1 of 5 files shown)


GDSNewt (approved):

AgaDufrat:
The collection link missing in the GraphQL response but we are accepting this diff (https://gov-uk.atlassian.net/browse/PP-7521)
- Content Store: The Welsh translation is "gone" and content store falls back to the English version Screenshot 2026-01-22 at 12 01 35
- GraphQL Screenshot 2026-01-22 at 12 01 03

PR #9571, opened 2024-11-01
(656) Show pageviews on dependent content

pezholio:
Now Publishing API sends page views in with the Dependent Content payload (see https://github.com/alphagov/publishing-api/pull/2957), we can display this to users

## Screenshot

image


Harriethw (approved):
nice!

PR #9575, opened 2024-11-04
Refactor image sizes so they can be made configurable

richardTowers:
This PR is probably best reviewed commit by commit.

There were several places where Whitehall hardcoded image dimensions to 960px x 640px:
- image size validation
- generating smaller image versions to save bandwidth
- presenting the user with an interface for cropping images larger than the valid size

This PR pulls the configuration required for these three things into a config YAML file, and provides a "default" configuration which matches the existing behaviour.

All ImageData models in this PR use the default configuration by defining `def image_kind = "default"` in the `ImageKind` mixin.

Future work will change `image_kind` from a method which always returns `"default"` to an active record attribute, so that image data can be created for different kinds of images. This will only be permitted where the content item allows different kinds of images, for example on LandingPage models, which can have hero images.

I've chosed "image_kind", rather than "image_type" to try and avoid confusion with jpeg vs. png etc. An image_kind could be "default", or "mobile_hero", or "featured" etc. - it defines the dimensions and the versions which should be generated.

By far the most gnarly bit of this is the interaction with carrierwave. Carrierwave makes heavy use of class methods, which makes it unintuitive to configure it based on the state of the instance - in this case what the value of the model's `image_kind` method is. The way we achieve this is by providing a `lambda` to carrierwave's `if:` configuration parameter, which checks whether the version is applicable for the current image_kind. This approach is heavily inspired by [Using one carrierwave image uploader with dynamic versions](https://www.diowa.com/blog/rails/2014/03/22/using-one-carrierwave-image-uploader-with-dynamic-versions) and [Using one carrierwave image uploader with different sizes on several models](http://andreapavoni.com/blog/2012/3/using-one-carrierwave-image-uploader-with-different-sizes-on-several-models) - credit where credit is due bloggers from > 10 years ago 👏🏻.

This PR is intended to be a straight refactoring - any behaviour changes are unintentional. Future PRs will add support for other image kinds.


dnkrj (commented):
👏 long overdue, I'll leave approval for the moment to give others a chance to take a look but as far as I'm concerned the code side of things all looks good.

I have one broader comment about the UX flow that potentially affects the code here and at the very least having `image_kind` as an attribute and was wondering if it came up in discussions at all.

The proposed flow begins with the user telling Whitehall what kind of image they're trying to upload and receiving an answer of "yes this is big enough for that purpose" or "no this image is too small". The alternative way to go about this would be that a user uploads an image and Whitehall tells them "this can be used as a desktop hero image, mobile hero image or embedded in your body copy".

I feel like the alternative I've outlined gives more long term flexibility to users (e.g. "this was the hero image, but we've got a new one so let's move this one to the body") and treats Whitehall more as a fully featured content management system, rather than a tool for one off publishing. At the same time, it might be more confusing to users why they can use images for one purpose but not another when they've not explicitly chosen that.

Ultimately I don't mind really mind which direction we go in, just want to make sure we've fully thought this through for whatever different kinds of blocks or image formats might come up in future.

richardTowers:
Yeah, that's a good point @dnkrj. It is janky having the user select the kind from a radio button at the same time as uploading the image. Honestly, the reason I did it that way was because the code was easy to write, not because it was good UX.

Thinking about it a bit more, there are probably some missing validations where these images get used. At the moment, I don't think there's anything to stop a user from uploading a big hero image (so long as it's on a landing page), and then including it in some govspeak, other than it'll probably error when it tries to present a bunch of variants that don't exist. I'll have a quick look to see if there's some nicer validation I could do there.

I think your alternative flow is probably better, but it would be a much bigger refactoring. At the moment, we decide whether the image needs cropping and what variants to upload at the point the image is uploaded. If we wanted to delay that, we'd have to store the high res images, and then have a point later in the user journey where they specialise the image so it can be used in a particular context. E.g. if someone uploads a high res image, and then later wants to use it in the default govspeak context, we'd need to present them with a cropping interface at the point where they choose to use the image, and then create all the required variants once it's cropped. I think that would necessitate a different structure in the database, and we'd probably have to drop Carrierwave, as it wants to do everything at file upload time.

Hopefully this change is a small step in the right direction, and doesn't preclude a bigger refactoring if we later decide that Whitehall should have better support for images.

richardTowers:
Added a check so that non-default image kinds can't be used in govspeak in [f885b46](https://github.com/alphagov/whitehall/pull/9575/commits/f885b466bbab481d5d9e98ee9f806e0352936460)

minhngocd:
From a quick review (I think i need to digest it a little bit more), I think the term `image_kind` doesn't really tell me exactly what it is there for (would like to hear other people's opinions too). 

I was wondering if what we are trying to do is abstract `image_dimensions` and `image_usage`? 
Another question I have is - do we want to have that image resizing functionality in Whitehall? Or should we rethink the image / uploading workflow and get asset management to manage the sizes. Would we not want asset manager to manage the sizes? user should be able to upload any image and crop, asset manager manages the storage of sizes for bandwidth, and rendering apps are in charge of querying for the correct size and rendering it.

Throwing a couple of questions out there from a user experience flow: 
- How are we envisioning users uploading images, will they be able to switch an image to become hero images after uploading them as normal size? 
- If the images are not switchable between the usages, does it make sense for them to share the same uploader? 

Going to think a little more on it, but these are the initial questions I have.

richardTowers:
Thanks Minno!

Some initial answers to those questions:

> From a quick review (I think i need to digest it a little bit more), I think the term image_kind doesn't really tell me exactly what it is there for (would like to hear other people's opinions too). I was wondering if what we are trying to do is abstract image_dimensions and image_usage?

Yeah, it's some combination of dimensions and usage. You can get a feel for the things that make up an `image_kind` by looking at [the YAML](https://github.com/alphagov/whitehall/pull/9575/files#diff-39e310b4323fe33d55ea677b13f9886c2fc4ca63c1b3889f7faeffaa033223d9). It determines the valid width and height of the file the user uploads (or what the upload will be cropped down to, if the file is bigger than that), and then the smaller variants of the file that carrierwave will generate.

The reason I didn't go for `image_dimensions` (which sounds like it would be simpler) is that in theory you could have images with the same valid dimensions, but different variants. Say, a featured image might be 960x640, just like a govspeak image, but require different variants because it's used in different ways.

`image_usage` would be an okay name too I think - how the image will be used is the thing that should determine the required size and variants. Maybe that's a better name.

> Another question I have is - do we want to have that image resizing functionality in Whitehall? Or should we rethink the image / uploading workflow and get asset manager to manage the sizes? user should be able to upload any image and crop, asset manager manages the storage of sizes for bandwidth, and rendering apps are in charge of querying for the correct size and rendering it.

Yes, possibly. But I think that's a much bigger refactoring, probably big enough to need pulling up into a team's objectives.

I think it's also good to think about this with @dnkrj 's comment in mind. Maybe Whitehall (or asset manager) shouldn't be determining how the image will be used (and therefore what dimensions are valid, and what variants are required) when it's uploaded. Maybe it should make that decision later, when the image actually gets used. So you could have like, a media library of high res images, and then when someone tries to use one in govspeak they're asked to crop it to the right size, and variants are generated (maybe by asset manager, maybe not).

I had a look at how a couple of third party CMSs do this, and it looks like there's also a common pattern where the cropping / resizing / whatever happens at request time (presumably with some caching). For example this hygraph URL with cropping parameters in the path:

```
https://eu-west-2.graphassets.com/cm3e4kmgk09pu07midhct1mgx/crop=dim:[800,10,1500,1500]/cm3eoqo8qfmdv07mp7lqfi82t
```

Or the [Contentful images API](https://www.contentful.com/developers/docs/references/images-api/).

I think it would definitely be nice to get the functionality for generating image variants out of whitehall, but I don't think it's totally clear that having it happen in asset manager at upload time is the right option.

As to whether this PR helps or hinders that direction of travel... I think it mostly helps. Introducing some concept of "how will this image be used" is still going to be necessary, even if it's asset-manager that does the resizing / variant production.

> Throwing a couple of questions out there from a user experience flow:

> How are we envisioning users uploading images, will they be able to switch an image to become hero images after uploading them as normal size?

Not at the moment - they get some radio buttons at upload time, and then the image kind is fixed from that point on. There's a screenshot on https://github.com/alphagov/whitehall/pull/9579 which kind of shows the idea.

I think @dnkrj 's comment is questioning whether this is the right model, and I take the point. But a system where users can upload images and then later customise them (crop / resize / generate variants) for various use cases would be quite a lot more involved to build.

> If the images are not switchable between the usages, does it make sense for them to share the same uploader?

They're not switchable in this implementation, so I think this is moot.

If I was going to do switchable image kinds, I'd probably do something more like "user uploads some high res image data, and then user can later create many `image_usage` entities from it". Schema something like this:
 
```mermaid
erDiagram
    IMAGE ||--|| IMAGE_DATA : "has one"
    IMAGE_DATA ||--|{ IMAGE_USAGE : "has many"
    IMAGE_USAGE ||--|{ IMAGE_VARIANT : "has many"
```

... but the user journey to populate such a database schema would need some careful design.

> Going to think a little more on it, but these are the initial questions I have.

🙇🏻  thank you very much for the thoughtful review so far.

test/unit/app/helpers/govspeak_helper_test.rb
```diff
@@ -197,6 +197,17 @@ def edition_with_attachment(body:)
     refute_select_within_html html, ".image.embedded"
   end
 
+  test "should ignore images with non-default image kinds" do
```
ChrisBAshton, line 200: It wouldn't be too much extra effort to add a `can_be_embedded_in_govspeak: true` property to the YAML config, so that we can set this at the config level rather than in code. Would also be less surprising for a dev following the breadcrumb trail when a support ticket comes in asking "why isn't my image rendering?" (explicit config as opposed to hardcoded conditional). What do you think?

ChrisBAshton (commented):
Thanks for DRYing up all the 960x640 references 👏 

+1 to the suggestion of, long term, following the example of other CMS's in having a media library instead of uploads-per-document.
And +1 to the suggestion of, long term, moving 'variant generation' out of Whitehall and into Asset Manager.
I agree that your PR doesn't preclude us from doing either of the above in future.

I'm not terribly keen on `image_kinds` as a name, and `image_usage` also doesn't feel right. Perhaps `image_constraints` would be a good fit? Think it lends itself quite well to the domain ("The default image constraints are 960x640, with the following different variants...").

config/image_kinds.yml
```diff
@@ -0,0 +1,29 @@
+---
+default:
+  display_name: "Default (960 by 640)"
+  valid_width: 960
+  valid_height: 640
+  versions:
```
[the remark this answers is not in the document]
richardTowers: Yeah, it's already inconsistent sadly. When we're talking about files, Carrierwave calls them [versions](https://github.com/alphagov/whitehall/blob/main/app/uploaders/image_uploader.rb#L13), but when we're talking about assets we call them variants.

I went with versions because the main place this bit of config is used is in a loop to call the Carrierwave `version` method, so it felt better to be consistent with that than with the assets code (that's mostly on the far side of Carrierwave).

app/models/concerns/image_kind.rb
```diff
@@ -2,8 +2,20 @@ module ImageKind
...
+                             { image_kind:, **attributes }
+                           else
+                             attributes
+                           end
+      super ordered_attributes
+    end
+
     def image_kind
-      attributes.fetch(:image_kind, "default")
+      attributes.fetch("image_kind", "default")
```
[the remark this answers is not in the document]
richardTowers: Done

test/unit/app/helpers/govspeak_helper_test.rb
richardTowers: Done in 5f82aa5 - although I might be getting a bit too fancy with the method_missing stuff there - let me know if you think it's too much. I could go with a `permits?(string)` method instead, or just `permitted_uses.include?(string)` directly.

richardTowers: Actually this is one of those where "if you have to ask 'is this too clever', it's probably too clever". I'll refactor it to use `permits?(use_case)`.

richardTowers: Now done more simply in e0ade76

ChrisBAshton (approved):

PR #9576, opened 2024-11-05
Content modelling/631 add copy code link

pezholio:


lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
```diff
@@ -0,0 +1,69 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+;(function (Modules) {
+  function CopyEmbedCode(module) {
+    this.module = module
```
Harriethw, line 6: clarifying question: I've not used JS modules like this before... so does this `function` get called before the `init` below? and the `module` that's passed in is the html of the component?

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
```diff
@@ -0,0 +1,69 @@
...
+    this.copyLink = document.createElement('a')
```
Harriethw, line 8: nitpick/accessibility: should this be a `button` element rather than an anchor, as it's doing something rather than taking us somewhere? or have the `aria-role` of button? not sure if it makes much of a difference for screen readers though tbh.

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
```diff
@@ -0,0 +1,69 @@
...
+    this.copyLink.textContent = originalText
+  }
+
+  // This is a fallback for browsers that do not support the async clipboard API
+  CopyEmbedCode.prototype.writeToClipboard = function (embedCode) {
+    return new Promise(function (resolve) {
+      // Create a textarea element with the embed code
+      const textArea = document.createElement('textarea')
+      textArea.value = embedCode
+      // This ensures the textarea doesn't get created at the bottom of the page
```
Harriethw, line 54: clarification: why do we care where the text area gets created if it's not seen by the user?

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
```diff
@@ -0,0 +1,69 @@
...
+      textArea.style.top = '0px'
+      textArea.style.left = '0px'
+      textArea.style.position = 'fixed'
+
+      document.body.appendChild(textArea)
+
+      // Select the text in the textarea
+      textArea.focus()
```
Harriethw, line 62: accessibility: Would calling `focus` here affect screen readers and keyboard nav? maybe we can just call `select`?

Harriethw (approved):
wow, so much hidden complexity for a little button!
LGTM, just had a few suggestions and questions.

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
```diff
@@ -4,19 +4,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
...
+
+    this.valueCell = this.module.querySelector('.govuk-summary-list__value')
   }
 
   CopyEmbedCode.prototype.init = function () {
     const dd = document.createElement('dd')
     dd.classList.add('govuk-summary-list__actions')
     dd.append(this.copyLink)
 
+    this.valueCell.style['text-overflow'] = 'ellipsis'
```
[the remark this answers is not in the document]
pezholio: Yeah, good call. Will get rid 👍

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
pezholio: Pretty much, yeah, there's some docs here https://docs.publishing.service.gov.uk/repos/whitehall/javascript.html

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
pezholio: Yeah, I've added `role=button` to this now - ta!

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
pezholio: Good call - I think that was when we were calling `focus()` on it, but we don't need it!

lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js
pezholio: Yeah, that works! 🙌

PR #9579, opened 2024-11-05
Support hero image sizes for landing pages

richardTowers:
This allows users to add various hero image sizes to LandingPage editions (but no other types of edition).

The user interface looks like this:

Screenshot of Whitehall showing the interface for adding images. There are radio buttons for "What kind of image is this?"

The markdown snippets for these images can be included in the blocks YAML on the landing page, and the `recursively_expand_images` method will replace them with image details (including the uploaded variants), irrespective of how deep they appear in the structure.


richardTowers:
Example landing page with images published using this branch (for those with access to integration):

https://draft-origin.integration.publishing.service.gov.uk/landing-page/rory

ryanb-gds (approved):
I don't know if there's really time to do anything about this, but personally I would like to see each block type have a (non-Active Record-backed) model class responsible for the logic of validation (could use ActiveModel validation) and mapping image tags to URLs. On the assumption that we're not in a position to start trying to nail down a well-designed object model just yet I'm approving.

Also, outside the scope of this PR, but I hadn't seen the extending of other landing pages' YAML body before and now I won't sleep tonight 😅

richardTowers:
Yeah, I'm reluctant to take too many steps down the "model the blocks properly" path, because I'm really still not sure that Whitehall is the right place for this kind of publishing at all. Maybe it is (in which case we should model things properly), or maybe it isn't (in which case we should keep it as simple as possible, and then excise this code as soon as possible).

It would be nice to have some activemodel validations on the blocks, so we could tell the user about errors rather than putting the errors in the content item (like I'm doing here) though.

I think I will also need to expand images in "featured" blocks in a future PR though, so I'll have a think about whether it's better to introduce a more proper model at that point.

PR #9580, opened 2024-11-05
Link to draft host document for content block preview

Harriethw:
Screenshot 2024-11-05 at 11 17 02

This PR includes two changes:
1. Link to draft host document for content block preview
We would like to link to the draft store so that users can see the changes they've just made to the block in host documents. 
As the table component is also used on the 'view' page I am passing in a boolean to determine whether to show draft or live frontend.

2. Open links in new tab
Although opening a new tab is not ideal for accessibility, in this case we think it is the preferred behaviour - opening a new tab can be recommended if a user is in the middle of a for or would lose context of what they are working on if not taken to a new tab. In the case of these host documents, if the user is previewing changes they will be in the middle of the form, if they are just viewing the content block it could be confusing to be taken to gov.uk and out of Content Block Manager.
There is precedent across Whitehall GDS for opening in new tabs when needed, and the mitigation is to include a message for screen readers declaring that the link opens in a new tab. We can test how this approach works for our users and change if needed.

----

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/host_editions_table_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/host_editions_table_component.rb
@@ -52,8 +52,16 @@ private
     end
   end
 
+  def content_link_text(content_item)
+    sanitize [
+      content_item.title,
+      tag.span("(opens in new tab)", class: "govuk-visually-hidden"),
+    ].join(" ")
+  end
+
   def content_link(content_item)
-    link_to(content_item.title, frontend_path(content_item), class: "govuk-link")
+    link_to(content_link_text(content_item),
+            frontend_path(content_item), class: "govuk-link", target: "_blank", rel: "noopener")
   end
 
   def organisation_link(content_item)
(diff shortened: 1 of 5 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/host_editions_table_component.rb
```diff
@@ -53,7 +53,8 @@ def frontend_path(content_item)
   end
 
   def content_link(content_item)
-    link_to(content_item.title, frontend_path(content_item), class: "govuk-link")
+    link_to(sanitize(content_item.title + tag.span(" (opens in new tab)", class: "govuk-visually-hidden")),
```
pezholio, line 56: I find this a bit hard to grok with all the brackets, maybe having a method called `content_link_text` with something like this would be better?

```ruby
def content_link_text(content_item)
  sanitize [
    content_item.title,
    tag.span("(opens in new tab)", class: "govuk-visually-hidden"),
  ].join(" ")
end
```

pezholio (approved):
LGTM - just one minor thing

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/host_editions_table_component.rb
Harriethw: good idea!

pezholio (approved):

PR #9582, opened 2024-11-05
(667) Fix back button on review links page

pezholio:
This should take the user back to the new edition path for a document. This will result in a new draft being created, but it’s fine as we’re deleting obsolete drafts now.

### DIFF
--- a/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review_links.html.erb
+++ b/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review_links.html.erb
@@ -3,7 +3,7 @@
 <% content_for :title_margin_bottom, 4 %>
 <% content_for :back_link do %>
   <%= render "govuk_publishing_components/components/back_link", {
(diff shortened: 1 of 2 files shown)


lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/workflow/review_links.html.erb
```diff
@@ -3,7 +3,7 @@
 <% content_for :title_margin_bottom, 4 %>
 <% content_for :back_link do %>
   <%= render "govuk_publishing_components/components/back_link", {
-    href: content_block_manager.new_content_block_manager_content_block_edition_path(@content_block_document.id),
+    href: content_block_manager.new_content_block_manager_content_block_document_edition_path(@content_block_document),
```
Harriethw, line 6: think this is failing cause it's covered by a feature test somewhere?

pezholio: Ah ha, will sort

Harriethw (approved):

PR #9584, opened 2024-11-05
(680) Switch feature flag out for permissions check

pezholio:
This removes the Flipflop feature flag check for a simple check of if the environment is staging or integration OR if the user is a GDS admin. Eventually we’ll allow access to all users, so there’s no point adding a new permission, but we do want to control access in production until we’re ready for launch.

I’ve also updated the `login_as_admin` helper, so the user we log in as is actually a GDS admin.


richardTowers (approved):
Looks good to me!

PR #9588, opened 2024-11-06
(629) Update summary cards

pezholio:
This updates the summary cards on the homepage, as well as some tweaks to the copy.

## Screenshot

image


Harriethw (approved):
this looks great! my only request is would it be possible to get the columns aligned? This bothers me 😅 
383913159-8577fa6b-da6b-46f2-bbec-5d983bbf90a5

PR #9589, opened 2024-11-06
Enable monaco editor for new landing pages

richardTowers:
The form only has the id #edit_edition when we're editing an edition. It has #new_edition when we're creating a new edition. In either case, it has the .edition-form class, so let's use that for our selector instead:

https://github.com/alphagov/whitehall/blob/3bff7dd08ea39b71d5d63305a33b87394287d04a/app/helpers/admin/editions_helper.rb#L105


KludgeKML (approved):
LGTM

PR #9605, opened 2024-11-13
(678) Allow sorting on dependent content

pezholio:
This allows sorting and pagination on dependent content, both on the view and review pages

## Screenshots

_Note: I temporarily changed the number of results per page to demonstrate the pagination_

image

image

### DIFF
--- a/lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
@@ -4,18 +4,22 @@ require "uri"
 
 module ContentBlockManager
   class GetHostContentItems
-    attr_reader :content_id
+    attr_reader :content_id, :page, :order
 
-    def initialize(content_id:)
+    DEFAULT_ORDER = "-unique_pageviews".freeze
+
+    def initialize(content_id:, page: nil, order: nil)
       self.content_id = content_id
+      self.page = page
+      self.order = order || DEFAULT_ORDER
     end
 
-    def self.by_embedded_document(content_block_document:)
-      new(content_id: content_block_document.content_id).items
+    def self.by_embedded_document(content_block_document:, page: nil, order: nil)
+      new(content_id: content_block_document.content_id, page:, order:).items
     end
 
     def items
-      content_items["results"].map do |item|
+      items = content_items["results"].map do |item|
         ContentBlockManager::HostContentItem.new(
           title: item["title"],
           base_path: item["base_path"],
@@ -27,15 +31,17 @@ module ContentBlockManager
           unique_pageviews: item["unique_pageviews"],
         )
       end
+
+      ContentBlockManager::HostContentItems.new(items, content_items["total"], content_items["total_pages"])
     end
 
   private
 
-    attr_writer :content_id
+    attr_writer :content_id, :page, :order
 
     def content_items
       @content_items ||= begin
-        response = Services.publishing_api.get_content_by_embedded_document(@content_id)
+        response = Services.publishing_api.get_content_by_embedded_document(@content_id, { page:, order: }.compact)
         response.parsed_content
       end
     end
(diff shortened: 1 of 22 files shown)


lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
```diff
@@ -4,18 +4,22 @@
...
+    def initialize(content_id:, page: nil, order: nil)
       self.content_id = content_id
+      self.page = page
+      self.order = order || DEFAULT_ORDER
     end
 
-    def self.by_embedded_document(content_block_document:)
-      new(content_id: content_block_document.content_id).items
+    def self.by_embedded_document(content_block_document:, page: nil, order: nil)
+      new(content_id: content_block_document.content_id, page:, order:).items
```
Harriethw, line 18: I wondered if you'd seen the `Kaminara` helper methods that Whitehall uses to paginate? https://github.com/alphagov/whitehall/blob/c88f75b785b92b5d045ac1e2d35c675ed98c414c/app/models/admin/edition_filter.rb#L26 
That means the results can just be passed to the `paginate` function in the `erb` file https://github.com/alphagov/whitehall/blob/adee19eb5cff5c1c83cb0b2ab46ece334539bd0b/lib/engines/content_block_manager/app/views/content_block_manager/content_block/documents/index.html.erb#L40 
it kind of hides away the logic/shape of data but is a bit cleaner, and you wouldn't need the extra data class that's been created?

lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
```diff
@@ -4,18 +4,22 @@
...
   class GetHostContentItems
-    attr_reader :content_id
+    attr_reader :content_id, :page, :order
 
-    def initialize(content_id:)
+    DEFAULT_ORDER = "-unique_pageviews".freeze
+
+    def initialize(content_id:, page: nil, order: nil)
       self.content_id = content_id
+      self.page = page
```
Harriethw, line 13: minor question: is having the `self` attributes here a more old fashioned way of having class attributes, over the `@` ones?

lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
```diff
@@ -4,18 +4,22 @@
 
 module ContentBlockManager
   class GetHostContentItems
-    attr_reader :content_id
+    attr_reader :content_id, :page, :order
 
-    def initialize(content_id:)
+    DEFAULT_ORDER = "-unique_pageviews".freeze
```
Harriethw, line 9: does this mean it shows the pages with least page views first? Is this something we'll revisit, as I guess the more page views the more important the document is?

Harriethw (approved):
Looks good! just had a couple of questions more for clarification really

lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
pezholio: Yeah, we'd have to load all the results into memory then wouldn't we? (although, on reflection, that might have been easier in the first instance 😅 )

lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
pezholio: It's more akin to having an `@` variable with an `attr_reader`, which I kinda prefer and is more Rubyish to my eyes. Think I'll get into the habit of removing this pattern when I see it.

lib/engines/content_block_manager/app/services/content_block_manager/get_host_content_items.rb
pezholio: This confused me too, but when we're doing ascending order with numbers, it shows the smallest first, so to see the most page views first we need descending order.

PR #9623, opened 2024-11-18
Fix yaml-editor for new documents (again)

richardTowers:
The check for the querySelector chain should be that we found a textArea, not that we found a textArea and it has a truthy value.

The empty string is falsey in JavaScript land, so checking !value incorrectly meant we got the "DOM structure did not match" warning.


KludgeKML (approved):
LGTM

PR #9628, opened 2024-11-19
Remove rake task that relies on CSV input

ChrisBAshton:
Rake tasks that rely on CSV input are a throwback to the old Puppet infrastructure of GOV.UK, where each app was running on dedicated VMs, meaning there was always a fairly stable/reliable file system to which we could upload said CSV and then run the rake task against it via Jenkins. Kubernetes pods, in contrast, are ephemeral in nature, so any uploaded CSVs could be lost at any time. We should therefore migrate CSV-based rake tasks into the UI of the respective publishing app, or remove it if it is no longer needed.

This rake task is no longer needed as it's been replaced by a form in the UI:
https://whitehall-admin.publishing.service.gov.uk/government/admin/republishing/bulk/documents-by-content-ids/new

Trello: https://trello.com/c/yi1ExGvz/3179-audit-rake-tasks-that-rely-on-scp-push

---


GDSNewt (approved):
Looks good, satisfied that the tasks have been replaced by UI elements or are no longer relevant 👍

PR #9629, opened 2024-11-19
Validate landing page base paths must start with /

richardTowers:
We're being a bit sneaky with landing pages, and setting base_path to be the entire user entered slug. Unlike with other editions, where the slug gets appended to /government/something, this lets the user set the whole base path.

Consequently, it's important that it starts with a slash, otherwise publishing API will reject it. As slugs can only be provided for landing pages when they're first created, leaving off the slash results in a situation the user can't recover from (the document is created with a bad slug, and there's no interface to change it).


KludgeKML (approved):
LGTM

PR #9631, opened 2024-11-19
Add Landing Pages to the list of edition classes

richardTowers:
This changes behaviour in a couple of ways:
1) It causes landing pages to show up in the "Document types" dropdown on the editions index page (which is my primary motivation for doing this)
2) It adds support for the `[internal admin link](/government/landing-pages/888)` style links for landing pages, which previously wouldn't have worked.

For the first use case, we remove landing pages for non-gds-admins, since other users can't do anything useful with landing pages.

The second use case isn't something I particularly care about, but there are tests for it so I had to make it work.

There's also a third side effect of adding LandingPage to the edition classes, which is that they would be sent to search-api somehow. I don't think we want or need that behaviour for landing pages, as we're already populating the search indexes via publishing-api. So I've excluded landing pages from that process.

### DIFF
--- a/app/helpers/admin/edition_actions_helper.rb
+++ b/app/helpers/admin/edition_actions_helper.rb
@@ -159,6 +159,7 @@ private
   def type_options_container(user)
     Whitehall.edition_classes.map { |edition_type|
       next if edition_type == FatalityNotice && !user.can_handle_fatalities?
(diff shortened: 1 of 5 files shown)


KludgeKML (approved):
LGTM

ChrisBAshton (approved):

PR #9635, opened 2024-11-20
(647) Support embed codes in inline preview

pezholio:
This updates the inline Govspeak preview within Whitehall to support embed codes. As we have the luxury of Whitehall being within the confines of Whitehall, it's fairly easy for us to use the `content_block_tools` gem to parse the Govspeak, pull out the references, fetch the content blocks and then replace the embed code with the rendered blocks.


lib/engines/content_block_manager/app/services/content_block_manager/find_and_replace_embed_codes_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    def initialize(html)
+      @html = html
+    end
+
+    def embed_content_references
+      @embed_content_references ||= ContentBlockTools::ContentBlockReference.find_all_in_document(html)
+    end
+
+    def content_blocks
+      @content_blocks ||= ContentBlockManager::ContentBlock::Edition.joins(:document).where(
```
Harriethw, line 31: This query would return all published Editions (of which there could be many per Document because we don't have a `superseded` workflow) but I think we would only want the `latest_edition` scope?
I'm wondering if the `render` should belong to the Document, as the embed code isn't going to change throughout its lifecycle, so we will always only be referring to the `live` content right?

Harriethw: otherwise when you update an edition the changes wouldn't propagate to other content blocks, or am I misunderstanding?

pezholio: Good point on the latest edition scope. I did consider having it on the document, but I thought we might want to render draft blocks too (for example, when doing the in-app preview)

Harriethw: > we might want to render draft blocks too (for example, when doing the in-app preview)

ah yeah that's true... I guess as long as this service and the API only ever returns the live one then it will be ok, would be worth adding a Document with a couple of Editions in the test

pezholio: I've tweaked it so we have a `current_versions` scope now, which ensures we're only querying the latest published editions.

pezholio:
Trello card https://trello.com/c/cbXSET85/647-investigate-inline-preview-in-whitehall

Harriethw (approved):

PR #9640, opened 2024-11-21
Content modelling/659 Preview Host Documents via iFrame

Harriethw:
https://trello.com/c/e2rpWPUT/659-spike-preview-host-documents

We would like users to be able to click through from the edit journey to view a preview of the Host Document with the new Content Block content within it.
This PR suggests the first stage of enabling it via an iframe embedding the HTML from the gov.uk frontend of the Host Document. 
We have come to this solution of using an iframe after extensive chats and found this to be the least bad option.
https://docs.google.com/document/d/19Nla6imjblR4XK_VI7t1lpob8Jmng2ZqRfpmsSnVuVQ/edit?tab=t.0#heading=h.s3ee1ypp6ehg

Screenshot 2024-11-25 at 11 49 32

### DIFF
--- a/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
@@ -4,19 +4,20 @@ require "uri"
 
 module ContentBlockManager
   class GetPreviewContent
+    def self.for_content_id(content_id:, content_block_edition:)
+      new(content_id:, content_block_edition:).for_content_id
+    end
+
+    def for_content_id
+      ContentBlockManager::PreviewContent.new(title: content_item["title"], html:)
+    end
+
+  private
+
     def initialize(content_id:, content_block_edition:)
       @content_id = content_id
       @content_block_edition = content_block_edition
     end
-
-    def preview_content
-      {
-        title: content_item["title"],
-        html:,
-      }
-    end
-
-  private
 
     def html
       @html ||= preview_html
@@ -50,8 +51,9 @@ module ContentBlockManager
     end
 
     def replace_blocks(nokogiri_html)
+      @preview_content_block_render ||= @content_block_edition.render
       content_block_spans(nokogiri_html).each do |span|
-        span.replace @content_block_edition.render
+        span.replace @preview_content_block_render
       end
     end
 
@@ -59,7 +61,7 @@ module ContentBlockManager
 
     def style_blocks(nokogiri_html)
       content_block_spans(nokogiri_html).each do |span|
-        span["style"] = "background-color: yellow;"
+        span["style"] = BLOCK_STYLE
       end
     end
 
--- a/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
@@ -40,7 +40,7 @@ module ContentBlockManager
 
     def preview_html
       uri = URI(frontend_path)
-      nokogiri_html = Nokogiri::HTML.parse(Net::HTTP.get(uri))
+      nokogiri_html = html_snapshot_from_frontend(uri)
       replace_existing_content_blocks(nokogiri_html)
     end
 
@@ -67,6 +67,17 @@ module ContentBlockManager
 
     def content_block_spans(nokogiri_html)
       nokogiri_html.css("span[data-content-id=\"#{@content_block_edition.document.content_id}\"]")
+    end
+
+    ERROR_HTML = "<html><body><p>Preview not found</p></body></html>".freeze
+
+    def html_snapshot_from_frontend(uri)
+      begin
+        raw_html = Net::HTTP.get(uri)
+      rescue StandardError
+        raw_html = ERROR_HTML
+      end
+      Nokogiri::HTML.parse(raw_html)
     end
   end
 end
--- a/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
@@ -44,13 +44,26 @@ module ContentBlockManager
     end
 
     def replace_existing_content_blocks(nokogiri_html)
-      existing_content_block_spans(nokogiri_html).each do |span|
-        span.replace @content_block_edition.render
-      end
+      replace_blocks(nokogiri_html)
+      style_blocks(nokogiri_html)
       nokogiri_html
     end
 
-    def existing_content_block_spans(nokogiri_html)
+    def replace_blocks(nokogiri_html)
+      content_block_spans(nokogiri_html).each do |span|
+        span.replace @content_block_edition.render
+      end
+    end
+
+    BLOCK_STYLE = "background-color: yellow;".freeze
+
+    def style_blocks(nokogiri_html)
+      content_block_spans(nokogiri_html).each do |span|
+        span["style"] = "background-color: yellow;"
+      end
+    end
+
+    def content_block_spans(nokogiri_html)
       nokogiri_html.css("span[data-content-id=\"#{@content_block_edition.document.content_id}\"]")
     end
   end
--- a/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
@@ -1,0 +1,40 @@
+require "net/http"
+require "json"
+require "uri"
+
+module ContentBlockManager
+  class GetPreviewContent
+    def initialize(content_id:)
+      @content_id = content_id
+    end
+
+    def preview_content
+      {
+        title: content_item["title"],
+        html:,
+      }
+    end
+
+  private
+
+    def content_item
+      @content_item ||= begin
+        response = Services.publishing_api.get_content(@content_id)
+        response.parsed_content
+      end
+    end
+
+    def frontend_base_path
+      Rails.env.development? ? Plek.external_url_for("government-frontend") : Plek.website_root
+    end
+
+    def frontend_path
+      frontend_base_path + content_item["base_path"]
+    end
+
+    def html
+      uri = URI(frontend_path)
+      @html ||= Nokogiri::HTML.parse(Net::HTTP.get(uri))
+    end
+  end
+end
--- a/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
+++ b/lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
@@ -4,8 +4,9 @@ require "uri"
 
 module ContentBlockManager
   class GetPreviewContent
-    def initialize(content_id:)
+    def initialize(content_id:, content_block_edition:)
       @content_id = content_id
+      @content_block_edition = content_block_edition
     end
 
     def preview_content
@@ -16,6 +17,10 @@ module ContentBlockManager
     end
 
   private
+
+    def html
+      @html ||= preview_html
+    end
 
     def content_item
       @content_item ||= begin
@@ -32,9 +37,21 @@ module ContentBlockManager
       frontend_base_path + content_item["base_path"]
     end
 
-    def html
+    def preview_html
       uri = URI(frontend_path)
-      @html ||= Nokogiri::HTML.parse(Net::HTTP.get(uri))
+      nokogiri_html = Nokogiri::HTML.parse(Net::HTTP.get(uri))
+      replace_existing_content_blocks(nokogiri_html)
+    end
+
+    def replace_existing_content_blocks(nokogiri_html)
+      existing_content_block_spans(nokogiri_html).each do |span|
+        span.replace @content_block_edition.render
+      end
+      nokogiri_html
+    end
+
+    def existing_content_block_spans(nokogiri_html)
+      nokogiri_html.css("span[data-content-id=\"#{@content_block_edition.document.content_id}\"]")
     end
   end
 end
--- a/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/host_content/preview.html.erb
+++ b/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/host_content/preview.html.erb
@@ -1,0 +1,16 @@
+<% content_for :page_title, "Preview content block in host document" %>
+<% content_for :context, "Preview content block" %>
+<% content_for :title, @preview_content[:title] %>
+<% content_for :title_margin_bottom, 0 %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds govuk-body govuk-!-margin-bottom-0">
+    <p><strong>Document title: </strong><%= @preview_content[:title] %></p>
+  </div>
+</div>
+<hr class="govuk-section-break govuk-!-margin-bottom-8 govuk-section-break--visible">
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <iframe id="preview" style="width:100%;height:80vh;" srcdoc="<%= @preview_content[:html] %>"></iframe>
+  </div>
+</div>
--- a/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/host_content/preview.html.erb
+++ b/lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/host_content/preview.html.erb
@@ -1,16 +1,16 @@
 <% content_for :page_title, "Preview content block in host document" %>
 <% content_for :context, "Preview content block" %>
-<% content_for :title, @preview_content[:title] %>
+<% content_for :title, @preview_content.title %>
 <% content_for :title_margin_bottom, 0 %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds govuk-body govuk-!-margin-bottom-0">
-    <p><strong>Document title: </strong><%= @preview_content[:title] %></p>
+    <p><strong>Document title: </strong><%= @preview_content.title %></p>
   </div>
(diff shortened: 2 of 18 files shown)


lib/engines/content_block_manager/config/routes.rb
```diff
@@ -12,6 +12,7 @@
       resources :editions, only: %i[new create destroy], path_names: { new: ":block_type/new" } do
         member do
           resources :workflow, only: %i[show update], controller: "editions/workflow", param: :step
+          resources :preview, only: %i[show], controller: "editions/preview", param: :host_content_id
```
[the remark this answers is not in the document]
Harriethw: hmm my only thinking is that this will eventually be getting a preview for a specific content block edition - so by having it be `/content-block/editions/:edition_id/preview/:uuid` we get the content block info 'for free' from its id. But I guess we could pass the content block information as a query param? 
Or is it the `host_content` you think it should have, so could be `/content-block/editions/:edition_id/host_content/:host_id/preview/`?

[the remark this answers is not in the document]
pezholio: Oh yeah, good point - `/content-block/editions/:edition_id/host_content/:host_id/preview/` is better in this case I think

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -0,0 +1,40 @@
...
+      Rails.env.development? ? Plek.external_url_for("government-frontend") : Plek.website_root
+    end
+
+    def frontend_path
+      frontend_base_path + content_item["base_path"]
+    end
+
+    def html
+      # TODO: how to make opening this path secure?
+      @html ||= URI.open(frontend_path).read
```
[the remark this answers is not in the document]
Harriethw: ah yeah good spot... I wasn't able to use Net:HTTP directly as there was some errors about encoding, but wrapping it in Nokogiri (which we use elsewhere) looks like it worked 👍

lib/engines/content_block_manager/test/unit/app/services/get_preview_content_test.rb
```diff
@@ -4,29 +4,44 @@ class ContentBlockManager::GetPreviewContentTest < ActiveSupport::TestCase
...
   end
 
   describe "#preview_content" do
     setup do
       stub_publishing_api_has_item(content_id: host_content_id, title: host_title, base_path: host_base_path)
     end
 
-    it "returns the title and raw frontend HTML for a document" do
+    it "returns the title and preview HTML for a document" do
       Net::HTTP.expects(:get).with(URI(Plek.website_root + host_base_path)).returns(fake_frontend_response)
```
Harriethw, line 31: we are no longer strictly unit testing the Nokogiri functionality, it seemed way to complicated to stub its methods as we iterate through the different elements, and I think it's worth letting that slide to get the assurance we're getting the content we expect?

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -32,9 +37,21 @@ def frontend_path
...
+    end
+
+    def replace_existing_content_blocks(nokogiri_html)
+      existing_content_block_spans(nokogiri_html).each do |span|
+        span.replace @content_block_edition.render
+      end
+      nokogiri_html
+    end
+
+    def existing_content_block_spans(nokogiri_html)
```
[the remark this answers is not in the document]
pezholio: I think this is fine to leave in here. All the tools functionality is functionality we use in multiple places. If we find that we need to use it elsewhere, we can pull it into the gem

lib/engines/content_block_manager/config/routes.rb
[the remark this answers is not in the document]
Harriethw: done 👍

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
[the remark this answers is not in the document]
Harriethw: cool cool

lib/engines/content_block_manager/test/unit/app/services/get_preview_content_test.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  describe "#preview_content" do
+    setup do
+      stub_publishing_api_has_item(content_id: host_content_id, title: host_title, base_path: host_base_path)
+    end
+
+    it "returns the title and raw frontend HTML for a document" do
+      Net::HTTP.expects(:get).with(URI(Plek.website_root + host_base_path)).returns(fake_frontend_response)
+      Nokogiri::HTML4.expects(:parse).with(fake_frontend_response).returns(fake_frontend_response)
```
pezholio: Should this be:

```suggestion
      Nokogiri::HTML.expects(:parse).with(fake_frontend_response).returns(fake_frontend_response)
```

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -0,0 +1,40 @@
+require "net/http"
+require "json"
+require "uri"
+
+module ContentBlockManager
+  class GetPreviewContent
```
pezholio, line 6: I wonder if this class could/should be more of a model-type class, especially as it returns something that looks like a model. I thinking something like:

```ruby
class PreviewContent < Data.define(:title, :html)
  class << self
    def for_content_id(content_id)
        PreviewContent.new(title: content_item["title"], html:)
    end
    
    private

    def content_item
      @content_item ||= begin
        response = Services.publishing_api.get_content(@content_id)
        response.parsed_content
      end
    end

    def frontend_base_path
      Rails.env.development? ? Plek.external_url_for("government-frontend") : Plek.website_root
    end

    def frontend_path
      frontend_base_path + content_item["base_path"]
    end

    def html
      uri = URI(frontend_path)
      @html ||= Nokogiri::HTML.parse(Net::HTTP.get(uri))
    end
  end
end
```

lib/engines/content_block_manager/test/unit/app/services/get_preview_content_test.rb
pezholio: Yeah, that makes sense I think

lib/engines/content_block_manager/app/views/content_block_manager/content_block/editions/host_content/preview.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds govuk-body govuk-!-margin-bottom-0">
+    <p><strong>Document title: </strong><%= @preview_content[:title] %></p>
+  </div>
+</div>
+<hr class="govuk-section-break govuk-!-margin-bottom-8 govuk-section-break--visible">
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <iframe id="preview" style="width:100%;height:80vh;" srcdoc="<%= @preview_content[:html] %>"></iframe>
```
pezholio, line 14: This is ace, I didn't know about `srcdoc`!

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -32,9 +37,21 @@ def frontend_path
...
+    def preview_html
       uri = URI(frontend_path)
-      @html ||= Nokogiri::HTML.parse(Net::HTTP.get(uri))
+      nokogiri_html = Nokogiri::HTML.parse(Net::HTTP.get(uri))
+      replace_existing_content_blocks(nokogiri_html)
+    end
+
+    def replace_existing_content_blocks(nokogiri_html)
+      existing_content_block_spans(nokogiri_html).each do |span|
+        span.replace @content_block_edition.render
```
pezholio, line 54: Tiny suggestion - could we memoize the result of `render` here, so we don't have to potentially call it multiple times?

lib/engines/content_block_manager/test/unit/app/services/get_preview_content_test.rb
```diff
@@ -4,29 +4,44 @@ class ContentBlockManager::GetPreviewContentTest < ActiveSupport::TestCase
...
       }
 
-      assert_equal expected_content, ContentBlockManager::GetPreviewContent.new(content_id: host_content_id).preview_content
+      actual_content = ContentBlockManager::GetPreviewContent.new(
+        content_id: host_content_id,
+        content_block_edition: block_to_preview,
+      ).preview_content
+
+      assert_equal expected_content[:title], actual_content[:title]
+      assert_equal expected_content[:html].to_s, actual_content[:html].to_s
```
pezholio, line 44: I think it might be better to mock the repsonse from `.render` here, rather than bubble up through the call stack into the gem, as any change to the way we render in the gem will mean the tests fail. We can also document what calls we expect to send to the edition too.

Something like:

```ruby
block_to_preview.expects(:render).returns("RENDERED RESPONSE")
expected_html = "<body><p>test</p>RENDERED RESPONSE</body>"
```

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -44,13 +44,24 @@ def preview_html
...
+      content_block_spans(nokogiri_html).each do |span|
         span.replace @content_block_edition.render
       end
-      nokogiri_html
     end
 
-    def existing_content_block_spans(nokogiri_html)
+    def style_blocks(nokogiri_html)
+      content_block_spans(nokogiri_html).each do |span|
+        span["style"] = "background-color: yellow;"
```
pezholio, line 60: Would it be better to add a class here and then have the styling in the CSS? That way it'll be easier to make any tweakes further down the line - thinking maybe `content-embed--preview` to stay true to the [BEM way of doing things](https://getbem.com/). Could alter this with something like:

```suggestion
        span["class"] = "#{span["class"]} content-embed--preview"
```

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -0,0 +1,40 @@
...
+      Rails.env.development? ? Plek.external_url_for("government-frontend") : Plek.website_root
+    end
+
+    def frontend_path
+      frontend_base_path + content_item["base_path"]
+    end
+
+    def html
+      uri = URI(frontend_path)
+      @html ||= Nokogiri::HTML.parse(Net::HTTP.get(uri))
```
pezholio: Could/should we have a fallback if we don't get a response? I'm worried that when we run locally, we'll have to have the frontend stack running too, which could get annoying

pezholio (commented):
This is ace! Just a couple of things to think about. Feel free to push back on any / all of them if I'm barking up the wrong tree.

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
Harriethw: oh I meant to put something in the commit about this - i just hit a slight complication with adding a style to an iframe from outside of it, but I think it is possible

lib/engines/content_block_manager/test/unit/app/services/get_preview_content_test.rb
Harriethw: ah it passed because HTML is an alias for HTML4, but yep can go and change it

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
Harriethw: hmm that's a good point and definitely something that needs some design thinking - what do you think we should show for now? A paragraph element that says "Preview not found"?

pezholio: Yeah, I think that'll be enough for now

lib/engines/content_block_manager/test/unit/app/services/get_preview_content_test.rb
Harriethw: ah good spot that should definitely be stubbed, but I'm still returning HTML because we need to manipulate the HTML in future commit to add styles / replace the original element.

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
Harriethw: good one 👍 liking the use of Data

Harriethw: I've done this in a new refactor commit

lib/engines/content_block_manager/app/services/content_block_manager/preview_content.rb
```diff
@@ -0,0 +1,67 @@
+require "net/http"
+require "json"
+require "uri"
+
+module ContentBlockManager
+  class PreviewContent < Data.define(:title, :html)
```
pezholio, line 6: This might sit better in models now

lib/engines/content_block_manager/test/unit/app/services/preview_content_test.rb
```diff
@@ -0,0 +1,54 @@
...
+  let(:host_base_path) { "/test" }
+  let(:uri_mock) { mock }
+  let(:fake_frontend_response) do
+    "<body><p>test</p><span class=\"content-embed content-embed__content_block_email_address\" data-content-block=\"\" data-document-type=\"content_block_email_address\" data-content-id=\"#{preview_content_id}\">example@example.com</span></body>"
+  end
+  let(:block_render) do
+    "<span class=\"content-embed content-embed__content_block_email_address\" data-content-block=\"\" data-document-type=\"content_block_email_address\" data-content-id=\"#{preview_content_id}\">new@new.com</span>"
+  end
+  let(:block_render_with_style) do
+    "<span class=\"content-embed content-embed__content_block_email_address\" data-content-block=\"\" data-document-type=\"content_block_email_address\" data-content-id=\"#{preview_content_id}\" style=\"background-color: yellow;\">new@new.com</span>"
```
pezholio, line 19: If you made the `BLOCK_STYLE` constant public, you could do this:

```suggestion
    "<span class=\"content-embed content-embed__content_block_email_address\" data-content-block=\"\" data-document-type=\"content_block_email_address\" data-content-id=\"#{preview_content_id}\" style=\"#{ContentBlockManager::PreviewContent::BLOCK_STYLE}\">new@new.com</span>"
```

Harriethw: hmm, I'm in two minds about that, because I quite like it when tests are more literal, and would break when that value is changed - but can use if you think it's more readable

pezholio: Nah, that's fair enough 👍

lib/engines/content_block_manager/app/services/content_block_manager/preview_content.rb
Harriethw: ah ok, is that the case even though it's making a call out to other APIs? Always though Models was more for the actual db models or like flat data

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
pezholio: Yeah, I've just clocked that myself when testing - that's probably fine for now 👍

lib/engines/content_block_manager/app/services/content_block_manager/preview_content.rb
Harriethw: actually making me think what we need is a flat Preview Content and then a Get Preview Content service, like we have for Host Content Item?

pezholio: That would probably be more consistent actually

Harriethw: > making me think what we need is a flat Preview Content and then a Get Preview Content service, like we have for Host Content Item?

I've gone with this approach for consistency and keeping things that call outside of the app in service folder.

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -4,20 +4,21 @@
 
 module ContentBlockManager
   class GetPreviewContent
-    def initialize(content_id:, content_block_edition:)
-      @content_id = content_id
-      @content_block_edition = content_block_edition
+    def self.for_content_id(content_id:, content_block_edition:)
+      new(content_id:, content_block_edition:).for_content_id
```
Harriethw, line 8: TIL more about what `self` means in Ruby... had to instantiate a new instance because otherwise in tests the class kept the same context and so the instance variables were also kept 🤔

Harriethw: hopefully this makes sense now 😅

lib/engines/content_block_manager/app/services/content_block_manager/get_preview_content.rb
```diff
@@ -0,0 +1,83 @@
...
+      nokogiri_html.css("span[data-content-id=\"#{@content_block_edition.document.content_id}\"]")
+    end
+
+    ERROR_HTML = "<html><body><p>Preview not found</p></body></html>".freeze
+
+    def html_snapshot_from_frontend(uri)
+      begin
+        raw_html = Net::HTTP.get(uri)
+      rescue StandardError
+        raw_html = ERROR_HTML
```
pezholio, line 78: Love it! 👍

pezholio (approved):
LGTM - there's some tweaks we can make, but this is really good as a first pass! 🥳

PR #9641, opened 2024-11-21
Add asset_manager_id to file attachments

richardTowers:
This needs to follow on from https://github.com/alphagov/publishing-api/pull/2994

As described there, the lack of asset_manager_id in the data for file asset attachments in content store makes it hard to request the attachments on the server side. Instead, we have to redirect users to the assets, which has led to ugly workarounds like CSV previews being rendered by frontend but served on assets.publishing.service.gov.uk.

Adding asset_manager_id should allow the frontends to request attachments directly, using the API client:

    GdsApi.asset_manager.media(asset_manager_id, filename)

This should make it much more convenient to preview assets, or use them for other rendering purposes (e.g. showing a CSV as a line graph).

EDIT: This description has been updated by @unoduetre to not include details which are no longer relevant.

### DIFF
--- a/app/models/file_attachment.rb
+++ b/app/models/file_attachment.rb
@@ -55,7 +55,7 @@ class FileAttachment < Attachment
       filename:,
       number_of_pages:,
       preview_url:,
-      asset_manager_id:,
+      assets:,
     }
   end
 
@@ -67,10 +67,15 @@ private
     nil
   end
 
-  def asset_manager_id
-    return unless csv? && attachable.is_a?(Edition) && attachment_data.all_asset_variants_uploaded?
+  def assets
+    return unless attachable.is_a?(Edition) && attachment_data.all_asset_variants_uploaded?
 
-    attachment_data.assets.first.asset_manager_id
+    attachment_data.assets.map do |asset|
+      {
+        asset_manager_id: asset.asset_manager_id,
+        filename: asset.filename,
+      }
+    end
   end
 
   def preview_url
--- a/app/models/file_attachment.rb
+++ b/app/models/file_attachment.rb
@@ -55,7 +55,7 @@ class FileAttachment < Attachment
       filename:,
       number_of_pages:,
       preview_url:,
-      attachment_data_id:,
+      asset_manager_id:,
     }
   end
 
@@ -67,10 +67,10 @@ private
     nil
   end
 
-  def attachment_data_id
+  def asset_manager_id
     return unless csv? && attachable.is_a?(Edition) && attachment_data.all_asset_variants_uploaded?
 
-    attachment_data.id
+    attachment_data.assets.first.asset_manager_id
   end
 
   def preview_url
--- a/app/models/file_attachment.rb
+++ b/app/models/file_attachment.rb
@@ -55,6 +55,7 @@ class FileAttachment < Attachment
       filename:,
       number_of_pages:,
       preview_url:,
+      attachment_data_id:,
     }
   end
 
@@ -64,6 +65,12 @@ private
     attachable.alternative_format_contact_email
   rescue NoMethodError
     nil
+  end
+
+  def attachment_data_id
+    return unless csv? && attachable.is_a?(Edition) && attachment_data.all_asset_variants_uploaded?
+
+    attachment_data.id
   end
 
   def preview_url
--- a/test/unit/app/models/attachment_test.rb
+++ b/test/unit/app/models/attachment_test.rb
@@ -8,7 +8,7 @@ class AttachmentTest < ActiveSupport::TestCase
 
     output = attachment.publishing_api_details
     assert_equal output.keys,
-                 %i[attachment_type id title url accessible alternative_format_contact_email content_type filename]
+                 %i[attachment_type id title url accessible alternative_format_contact_email content_type filename assets]
   end
 
   test ".publishing_api_details includes publication attachment details for " \
--- a/test/unit/app/models/file_attachment_test.rb
+++ b/test/unit/app/models/file_attachment_test.rb
@@ -69,9 +69,9 @@ class FileAttachmentTest < ActiveSupport::TestCase
     assert_equal Plek.asset_root + "/media/#{attachment.attachment_data.id}/sample.csv/preview", attachment.publishing_api_details_for_format[:preview_url]
   end
 
-  test "#asset_manager_id returns asset manager id if all_asset_variants_uploaded?" do
+  test "#assets returns assets list if all_asset_variants_uploaded?" do
     attachment = create(:csv_attachment, attachable: create(:edition))
-    assert_not_nil attachment.attachment_data.assets.first.asset_manager_id
-    assert_equal attachment.attachment_data.assets.first.asset_manager_id, attachment.publishing_api_details_for_format[:asset_manager_id]
+
+    assert_equal [{ "asset_manager_id": attachment.attachment_data.assets.first.asset_manager_id, "filename": attachment.attachment_data.assets.first.filename }], attachment.publishing_api_details_for_format[:assets]
   end
 end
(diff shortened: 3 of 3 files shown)


app/models/file_attachment.rb
```diff
@@ -66,6 +67,12 @@ def alternative_format_contact_email
     nil
   end
 
+  def attachment_data_id
+    return unless csv? && attachable.is_a?(Edition) && attachment_data.all_asset_variants_uploaded?
+
+    attachment_data.id
```
catalinailie, line 73: I'm not sure `attachment_data.id` is the ID we want. `GdsApi.asset_manager` expects [the asset's ID](https://github.com/alphagov/gds-api-adapters/blob/main/lib/gds_api/asset_manager.rb#L82:L91), and that one, in whitehall, is `Asset.asset_manager_id`.

I think this is the reason why we have this scenario with `preview_url` formed in whitehall using `attachment_data.id` and the overwritten in `government-frontend` with the URL that has this `asset_manager_id` in it (formed [here](https://github.com/alphagov/whitehall/blob/1dae8f4a1fa7d572ab340befda74004a76407396/lib/whitehall/asset_manager_storage.rb#L33) I think, but I might be wrong).

unoduetre: Yes, you're right. It should be asset_manager_id of the relevant asset. The issue is that there can be more than one asset per attachment. But the solution is to get the attachment with the variant "original", which means the original file as uploaded. The combinations of assetable_id, assetable_type and variant are unique (there is a validation for that). assetable_type in this case is "AttachmentData", assetable_id is the id of the object of the AttachmentData class (the id from the database). And the variant should be specified as "original". In this way we can get the unique asset_manager_id.

catalinailie (commented):
The changes look good, I only have one small suggestion.
And really nice that you covered the changes with tests ⭐

lib/whitehall/asset_manager_storage.rb
```diff
@@ -34,6 +34,10 @@ def url
       end
     end
 
+    def asset_manager_id
+      get_asset&.asset_manager_id
```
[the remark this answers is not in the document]
unoduetre: I think it's better to change the other methods as well instead. What do you think?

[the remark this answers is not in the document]
catalinailie: Hmm not sure. Since this is not our app, I think it would be better for us to follow the format that currently is here. And looking at the url method, I think it would be better to leave it as it is, instead of using a safe navigation operator.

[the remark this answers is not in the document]
unoduetre: OK, I'll use that format then.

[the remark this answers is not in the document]
unoduetre: Changed.

minhngocd:
AttachmentData has `assets` (usually only one now, for file attachments), which also store `asset_manager_id`. Can we not make use of that instead of taking it from `file` which I understand is more the actual file uploader from CarrierWave? 
Are we looking to attach this ID to the content item when publishing it through to content store? if so, I think AttachmentData > Asset > Asset Manager ID might be more straight forward

catalinailie:
@minhngocd Ohh interesting, it make sense. I don't know why we haven't thought of that.
Yes, we are looking to attach this ID to the content item when publishing.

minhngocd:
Another thought i was also having is: what's the need for the asset manager ID? If you're getting it from asset manager via the GDS API, would that not be equivalent to using the attachment URL that comes through with the content item? 

e.g. https://www.integration.publishing.service.gov.uk/api/content/government/publications/pension-credit-toolkit 
File attachments currently come through with the content item as 
```
{
        "accessible": false,
        "alternative_format_contact_email": "accessible.formats@dwp.gov.uk",
        "attachment_type": "file",
        "command_paper_number": "",
        "content_type": "application/zip",
        "file_size": 465016,
        "filename": "pension-credit-customers-english.zip",
        "hoc_paper_number": "",
        "id": "8524176",
        "isbn": "",
        "locale": "en",
        "title": "Pension Credit social media images directed at potential customers",
        "unique_reference": "",
        "unnumbered_command_paper": false,
        "unnumbered_hoc_paper": false,
        "url": "https://assets.publishing.service.gov.uk/media/6756cb787ec19a3516f79a0e/pension-credit-customers-english.zip"
      },
```

The url is a direct link to the asset on asset manager. can that not be used to render previews? Sorry if I'm not getting something - happy to jump on a call and discuss

catalinailie:
> The url is a direct link to the asset on asset manager. can that not be used to render previews? 

This is the current behaviour. Government-frontend uses this asset URL for creating the [`/media/id/filename/preview` url](https://github.com/alphagov/government-frontend/blob/main/app/presenters/content_item/attachments.rb#L10).
But for our change, we need to extract the asset_manager_id and filename to create the new preview `/csv-preview/id/filename` url that will be served from gov.uk domain. And parsing an URL to get an ID isn't the best approach. It will be nicer if that logic moves away from government-frontend and we just use the asset_manager_id from content item.

minhngocd:
ahhh ok thanks for explaining! yes, in which case i think exposing it from AttachmentData > Assets through to the `publishing_api_details` on Attachment would be the most straightforward

minhngocd:
Suggestion, when we expose it, maybe it can be an array of hash object? 
like 
```
assets: [{
  filename: 
  asset_manager_id:
}]
```

Because the relationship between AttachmentData and Assets is one to many due to the multiple versions of image uploads. If we do an array of objects, we can return everything rather than arbitrarily selecting the first asset to send.

catalinailie:
We will be sending this asset_manager_id only for csv file attachments. I might be wrong, but I guess the AttachmentData will have only one Asset linked to it, right? So, in this case, picking the first Asset is actually the only one.

catalinailie:
Ups, I requested a review from the wrong user

app/models/file_attachment.rb
```diff
@@ -66,6 +67,12 @@ def alternative_format_contact_email
     nil
   end
 
+  def asset_manager_id
+    return unless csv? && attachable.is_a?(Edition) && attachment_data.all_asset_variants_uploaded?
+
+    attachment_data.assets.first.asset_manager_id
```
minhngocd, line 73: Personally, I feel like implementing this specifically for the CSV scenario and only returning the first asset manager id feels too narrow to one use-case. This is then also not explicit in the naming as well since the variable being exposed is `asset_manager_id` I would expect it to return all asset manager IDs associated with the file, not just the first CSV ones. 

can we not return something like 
```
assets: [{
  asset_manager_id: abc
  filename: abc_1.csv
}]
```

And then it's up to the consumer of the content item to decide to use the first one if it's a CSV?

minhngocd: Otherwise it feels like we're building consumer logic into the provider

catalinailie: Yeah, I agree the name is not very explicit for what that value is used. 
The initial idea was to send this asset_manager_id (or a better name) only for csv attachments. In this way, government-frontend knows when to add the preview URL to the page.
It would be nice not to have any logic in govenrment-frontend for checking if the file is a csv or not and to have this logic in the publishing side since this check is already there.

minhngocd (approved):
Looks good! Thanks @catalinailie !
I guess the failing tests are related to publishing API changes to the schema going through.

app/models/file_attachment.rb
minhngocd: Discussed offline with @catalinailie , As an FYI update for folks reading the PR. We decided we would require returning all assets array as the file is not necessary always the first one. 

```
Asset.where(assetable_type: "AttachmentData").pluck(:assetable_id).group_by{|e|e}.select{|k,v|v.size>1}.count
=> 782044
```
There are 780k+ records where there are multiple variants of an AttachmentData

example of a live published edition:

```
[#<Asset:0x0000ffff710ae9e0
  id: 3812782,
  asset_manager_id: "67598fff9f669f2e28ce2b01",
  assetable_type: "AttachmentData",
  assetable_id: 1286037,
  filename: "thumbnail_redacted..">,
 #<Asset:0x0000ffff710ae8a0
  id: 3812783,
  asset_manager_id: "67598fff3bb681ccb0d346b7",
  assetable_type: "AttachmentData",
  assetable_id: 1286037,
  filename: "redacted...">]
```
Which belongs to a published edition, still live - and the first entry seems to be the thumbnail

nacnudus:
Just to note that this will be useful to data analysts, who have been asked by other departments to provide lists of attachments, when they were last updated, and what page they are attached to. Getting this information will require a join between the Publishing API and the Asset Manager databases, for all types of attachment (pdf, csv, etc.).

PR #9648, opened 2024-11-25
Organisation description fix

minhngocd:
ZenDesk enquiry to fix up some copy of Organisation descriptions.

https://trello.com/c/oLQQId8c/3195-technical-fault-with-whitehall

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -22,7 +22,7 @@ module OrganisationHelper
   end
 
   def organisation_display_name_and_parental_relationship(organisation)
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
LGTM, though it feels real weird to me that this text is being constructed by Whitehall rather than being built by the frontend app from the content item links

minhngocd:
> LGTM, though it feels real weird to me that this text is being constructed by Whitehall rather than being built by the frontend app from the content item links

+1 to that - maybe we stick something in tech debt... I was also expecting to put the fix into the rendering app but when traced back, it was attached to the content item all along...

PR #9651, opened 2024-11-26
Track all links and buttons in Whitehall as per agreed event schema

ryanb-gds:
There are a few things going on here:

We are preventing button tracking for elements that are not buttons but have the button role because there is a risk of duplicate events. Most non-button elements with the button role are links, and will therefore be tracked via the link tracker module. The link tracker module will send a link event type for links with the button role. We do not need to send additional events via button tracking for such elements.

There was a hardcoded "method" attribute which we unset, as we don't know whether it's actually a primary click or not. The link tracker from Publishing components does apply this attribute.

The link tracking setup was using the wrong data object, "event" instead of "link", so that's been fixed.

Finally we move the data module initialisation for links and buttons higher up the DOM tree so that tracking is applied globally(ish) 🎉

### DIFF
--- a/app/assets/javascripts/admin/analytics-modules/ga4-button-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-button-setup.js
@@ -13,7 +13,8 @@ window.GOVUK.analyticsGa4.analyticsModules =
         const buttons = moduleElement.querySelectorAll('button')
         buttons.forEach((button) => {
           const event = {
-            event_name: 'navigation',
+            event_name:
+              button.type === 'submit' ? 'navigation' : 'select_content',
             type: 'button',
             text: button.textContent
           }
--- a/app/assets/javascripts/admin/analytics-modules/ga4-button-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-button-setup.js
@@ -15,8 +15,7 @@ window.GOVUK.analyticsGa4.analyticsModules =
           const event = {
             event_name: 'navigation',
             type: 'button',
-            text: button.textContent,
-            method: 'primary_click'
+            text: button.textContent
           }
           if (button.dataset.ga4Event) {
             Object.assign(event, JSON.parse(button.dataset.ga4Event))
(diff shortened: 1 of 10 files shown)


dnkrj (approved):

app/assets/javascripts/admin/analytics-modules/ga4-button-setup.js
```diff
@@ -10,15 +10,12 @@ window.GOVUK.analyticsGa4.analyticsModules =
         "[data-module~='ga4-button-setup']"
       )
       moduleElements.forEach(function (moduleElement) {
-        const buttons = moduleElement.querySelectorAll(
-          'button'
```
ChrisBAshton: Nit: this reduction to one line should be fixed up to the first commit.

And the second commit should explain what you've explained in the PR description:
> There was a hardcoded "method" attribute which we unset, as we don't know whether it's actually a primary click or not. The link tracker from Publishing components does apply this attribute.

(NB, that last sentence reads a bit confusingly - I'd suggest rewording along the lines of "the link tracker from Publishing components already applies the attribute, so we don't need to repeat it here")

app/assets/javascripts/admin/analytics-modules/ga4-link-setup.js
```diff
@@ -16,10 +16,10 @@ window.GOVUK.analyticsGa4.analyticsModules =
             event_name: 'navigation',
             type: link.role === 'button' ? 'button' : 'generic_link'
           }
-          if (link.dataset.ga4Event) {
-            Object.assign(event, JSON.parse(link.dataset.ga4Event))
+          if (link.dataset.ga4Link) {
```
ChrisBAshton, line 19: Worth linking to https://github.com/alphagov/govuk_publishing_components/blob/2bf9077e3e091a2151feab56a715035e1883c689/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js#L9 in the commit message to more clearly show why this is being updated?

ChrisBAshton (approved):
LGTM 👍 Nitpick suggestions below

PR #9661, opened 2024-11-27
Validate landing pages with active model

richardTowers:
## What

At the moment, if you attempt to save a landing page which has YAML in a format that frontend won't accept, it will be submitted to publishing-api with the errors. Most of the time, this will result in a 500 error from frontend when the page is loaded.

This PR processes the YAML into a tree of ActiveModel classes, which validate that various properties hold.

It doesn't fully validate the format of every block - any unknown block types will be passed through as-is. It does add validation for hero blocks though, which are a very common source of errors.

These models can also be used to present the YAML to publishing-api, which significantly simplifies the landing page presenter, and should make it much easier to add other kinds of block which need processing (like images and featured images).

## Screenshots

image

### DIFF
--- a/app/models/landing_page/compound_block.rb
+++ b/app/models/landing_page/compound_block.rb
@@ -1,0 +1,20 @@
+class LandingPage::CompoundBlock < LandingPage::BaseBlock
+  include ActiveModel::API
+
+  attr_reader :content_blocks, :content_block_key
+
+  validates :content_blocks, presence: true
+  validate do
+    content_blocks.each { |b| errors.merge!(b.errors) if b.invalid? }
+  end
+
+  def initialize(source, images, content_block_key, content_blocks)
+    super(source, images)
+    @content_block_key = content_block_key
+    @content_blocks = LandingPage::BlockFactory.build_all(content_blocks, images)
+  end
+
+  def present_for_publishing_api
+    super.merge({ content_block_key => { "blocks" => content_blocks.map(&:present_for_publishing_api) } })
+  end
+end
--- a/app/models/landing_page/compound_block.rb
+++ b/app/models/landing_page/compound_block.rb
@@ -1,3 +1,18 @@
+# Compound blocks are blocks which have some of their own configuration,
+# alongside a list of child blocks. For example:
+#
+# - type: hero
+#   image: ...
+#   hero_content:
+#     blocks:
+#       - type: govspeak
+#         content: some text
+#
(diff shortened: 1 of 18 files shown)


KludgeKML (approved):
Looks like a good approach to me (absent failing test)

test/unit/app/models/landing_page/base_block_test.rb
```diff
@@ -11,4 +11,10 @@ class BaseBlockTest < ActiveSupport::TestCase
     assert subject.invalid?
     assert_equal ["Type can't be blank"], subject.errors.to_a
   end
+
+  test "raises error when presenting an invalid block to publishing api" do
+    subject = LandingPage::BaseBlock.new({}, [])
+    assert subject.invalid?
+    assert_raises(StandardError, match: /Type can't be blank/) { subject.present_for_publishing_api }
```
ChrisBAshton, line 18: Should we also assert that `cannot present invalid block to publishing api` is included in the StandardError?
Not immediately obvious the corresponding line has test coverage otherwise.

KludgeKML (approved):
LGTM

app/models/landing_page/compound_block.rb
```diff
@@ -0,0 +1,20 @@
+class LandingPage::CompoundBlock < LandingPage::BaseBlock
```
ChrisBAshton, line 1: One of those rare cases that might save a few future head-scratches with a short comment now:

```suggestion
# Compound blocks are blocks that are made up of several child blocks
# e.g. a 'hero' block contains both an image block and a title overlay block

class LandingPage::CompoundBlock < LandingPage::BaseBlock
```

^ I don't actually know what blocks are involved within the Hero block, as the associated test just specifies `"blocks" => [{ "type" => "some-block-type" }]`. So a comment might help clarify.

Also if I've understood its purpose properly, `CompositeBlock` might be a better name for it? 🤔

ChrisBAshton (approved):
LGTM - I like that the changes are all pretty much self contained to `landing_page/`.
(Optimising code for deletion, even if it's unlikely to be deleted anytime soon...!)

A couple of nitpicks below but no blockers from me. I'd be interested to hear Ryan's views on it though, in case it's not quite what he had in mind.

app/models/landing_page/compound_block.rb
richardTowers: Done in 38c9c2a

PR #9662, opened 2024-11-27
Add support for landing page images in image / featured blocks

richardTowers:
## What

Adds support for "landing page" images, which are a different size to the default images in whitehall.

These can be used in two types of "block" - the image block, and the featured block. The code to present these blocks is updated to transform: `[Image: blah.png]` to the appropriate asset URLs, as we do for the hero blocks.

Largely for consistency with hero blocks, the config we're expecting users to write is:

```
sources:
  desktop: "[Image: blah.png]"
  tablet: "[Image: blah.png]"
  mobile: "[Image: blah.png]"
```

However, as the aspect ratios of each image are the same (just different sizes), rather than adding three image kinds with two variants (1x / 2x) each, I've just added one image kind with six variants. Since we don't need or want to do art direction, I expect that we'll just upload one image and use it for all three sources (desktop / tablet / mobile).

It might be better in future just to do `source:` instead of `sources:` for these images, but I think that's a change which will likely confuse us at this point, so I'm sticking with the same structure as heros.

I'll update our blocks documentation once this merges.

## Screenshots

image


KludgeKML (approved):
LGTM (probably needs whitehall team approve too though)

config/image_kinds.yml
```diff
@@ -29,6 +29,36 @@ default:
...
+    - name: landing_page_desktop_2x
+      width: 1260
+      height: 944
+    - name: landing_page_desktop_1x
+      width: 630
+      height: 472
+      from_version: landing_page_desktop_2x
+    - name: landing_page_tablet_2x
+      width: 1416
+      height: 1062
```
ChrisBAshton, line 48: Hmmm shouldn't this be the value used for `valid_width`/`valid_height`? I understood that to be "the largest of the versions" so that it can scale down...?

ChrisBAshton (approved):
LGTM, but one thing threw me:

config/image_kinds.yml
richardTowers: Ahh yeah - I hadn't realised the tablet image is actually bigger than the desktop image. Yes, you're right. That should be the main one, and the others should scale down from it. Good catch.

PR #9666, opened 2024-11-28
Add an ADR to keep document history in sync

pezholio:
This adds an ADR to document our decision to add a RabbitMQ consumer to Whitehall to allow us to keep a record of when document updates have been triggered by an update to a dependent content block.


docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
```diff
@@ -0,0 +1,64 @@
+# 5. Keep Document history in sync with RabbitMQ
```
Harriethw, line 1: I wonder if it's worth grouping or labelling this and our last ADR somehow, to show that these are changes relating specifically to the content block manager? Or referencing them in our engine? Just aware that if the engine was moved out of Whitehall they might lose their context

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
```diff
@@ -0,0 +1,64 @@
...
+new events, however:
+
+1. This will result in an API call every time a user views a document; and
+2. Carrying out an INSERT query on a GET request isn't a pattern we want to encourage
+
+## Decision
+
+With this in mind, we are proposing adding a new message queue consumer in Whitehall. Rabbit MQ messages
+are already sent by Publishing API when documents are republished, so we can consume the existing 
+`published_documents` queue. We will listen for events with the `host_content` key, so we only listen
```
ryanb-gds, line 41: I think I'm right in saying that in Rabbit/ActiveMQ parlance Publishing API publishes messages to the `published_documents` [topic](https://www.rabbitmq.com/tutorials/amqp-concepts#exchange-topic), and then we set up a queue in Whitehall to consume them?

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
pezholio: I think in this case, it's more related to actual Whitehall code. I do think anything related directly to Content Block Manager should go elsewhere though

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
```diff
@@ -0,0 +1,64 @@
...
+for events triggered by a content object update. When we receive a message, we will:
+
+* Make a call to the `events` endpoint in Publishing API for that Content ID to find the latest 
+`HostContentUpdateJob` event
```
ryanb-gds, line 45: Can you elaborate a little about why we decided upon this route rather than creating a new topic for host content events and a new message that contained the information we need about the change to the content block? I know we discussed it on Slack, would be good to cover that here.

ryanb-gds (commented):
Looks good! Couple of suggestions from me

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
pezholio: Ah, you're right. I'm not super familiar with the nomenclature in RabbitMQ world

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
pezholio: Done 👍

ChrisBAshton (commented):
Thanks for writing this up. A few comments below.

More broadly - I wonder if there's a simpler alternative. If we consider two sources of 'history':

1. Changenotes/editorial remarks in Whitehall, pushed to Publishing API
2. Republishing of content via content-block-updates in Publishing API, pushed back to Whitehall

Have we considered moving _all_ history into Publishing API? On the surface of it, that seems cleaner to me (albeit we'd have the complexity of a one-off data migration from Whitehall to Publishing API):

1. Public changenotes already exist in Publishing API - we'd just need to find a way of pushing 'internal' changenotes up too (editorial remarks etc)
2. Let Publishing API worry about consolidating the history of public/internal/content-block-driven updates
3. On page load of a document, Whitehall can pull a document's entire history from Publishing API

(This idea, or any others you've considered, would make a good "Alternatives considered" section at the bottom of the ADR, as we often do in RFCs 🙏 ).

ryanb-gds:
> Have we considered moving _all_ history into Publishing API? On the surface of it, that seems cleaner to me (albeit we'd have the complexity of a one-off data migration from Whitehall to Publishing API):
> 
> 1. Public changenotes already exist in Publishing API - we'd just need to find a way of pushing 'internal' changenotes up too (editorial remarks etc)
> 2. Let Publishing API worry about consolidating the history of public/internal/content-block-driven updates
> 3. On page load of a document, Whitehall can pull a document's entire history from Publishing API

FWIW this was my first suggestion when I discussed it with Stu as well. I quite like the idea of Publishing API being responsible for all the data about, well, publishing. There would probably be some challenges though, e.g. change notes are part of Whitehall model validation at the moment

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
```diff
@@ -0,0 +1,64 @@
...
+
+* Content block is updated
+* We find all documents that use the content block
+* Each document is then represented to the content store with the updated content block details
+
+This all happens in Publishing API, so there is no record in Whitehall (or any other publishing apps) 
+of when a change to a document has been triggered by an update to a content block.
+
+In order to do this, we need to update the Publishing API to record an event when a document has been
+republished as a result to a change to a content block. We can then have an endpoint that allows us to
```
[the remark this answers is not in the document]
pezholio: I've got a draft PR here, but I didn't want to date it by linking to a PR that will have either been closed or merged by then https://github.com/alphagov/publishing-api/pull/2993. Will add a bit more context.

ChrisBAshton:
To add another voice to the suggestion of changenotes living in Publishing API: we'e just had a [request for internal changenotes to Manuals Publisher](https://govuk.zendesk.com/agent/tickets/5982091). Could obviously replicate the internal changenotes feature in Whitehall, but feels like there's a more widespread need here that Publishing API could accommodate.

ChrisBAshton:
@pezholio we've just chatted this through at our team's Dev Huddle and we think there's a way forward that addresses the need in the ADR, that also goes down an architectural route that's more scaleable long term.

The nice thing about the below approach is that you could stop at the MVP stage and it wouldn't be too bad, but there's also a nice end state in sight without too much additional effort (he says optimistically!). Shall we see about having a chat this week (perhaps with our TA or Lead Dev in attendance to advise)?

# MVP

All that would need doing right now to deliver an MVP:

1. Add a new endpoint to Publishing API to see the events for a particular document (as you've suggested in the ADR).
   The endpoint would expose all public changenote history, including the new 'content block update' changenotes.
2. Query the endpoint when loading the Edition screen in Whitehall, and merge the document histories.
   NB, we'd only really care about merging the first page's worth of document history, if that makes it any simpler (rather than worrying about pagination).

It does mean making an extra call to Publishing API on document load, but that's the direction we feel would be beneficial going all-in on further down the line (and has been done a lot elsewhere, eg Specialist Publisher is all Publishing API calls).

# Future work

The remaining work could be tackled in three phases (and not necessarily by the Content Modelling team!).

First, supporting editorial remarks / internal changenotes in Publishing API:

1. Make a change to Publishing API so that one can send both 'public' and 'internal' changenotes...
2. ...and ensure both kinds are included in the endpoint added earlier.

Second, sending all _new_ Whitehall changenote history to Publishing API:

1. Start sending internal changenotes to Publishing API, and...
3. ...stop storing any new changenotes (public or internal) in Whitehall.

Then (finally) the last phase would be to consolidate and then remove all unused code:

1. Migrate all historic public/internal changenotes out of Whitehall and into Publishing API
2. Delete local changenotes from Whitehall, and remove the 'changenote merging' logic added earlier, as now _all_ changenotes should be being pulled from Publishing API

pezholio:
@ChrisBAshton Sorry, was deep in user research last week, but picking this up now. Happy to give that approach a go, but the only issue I get with the approach of only merging the first page of results is we won't necessarily know what date window(s) to cover. For example:

A document has the following range of event datetimes for the first page:

2024-03-23T09:23:00
.....
2023-12-10T11:13:00

And a range of event datetimes for the second page

2023-11-22T12:27:00
...
2023-09-12T15:17:00

If we have an event that happens between `2023-11-22T12:27:00` (the newest event for the second page) and `2023-12-10T11:13:00` (the oldest even for the first page) it won't get picked up because it doesn't occur within that range of events. I guess there's something we could do with getting the timestamp of the first item of the next page, but it'd be good to have a chat anyway.

pezholio:
I've updated this ADR spelling out the issues with interleaving the events, as well as tweaked the proposal for RabbitMQ to propose a brand new topic (rather than using an existing topic). I'm still leaning towards the RabbitMQ solution, but I'm keeping an open mind. Thoughts welcome!

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
```diff
@@ -0,0 +1,147 @@
...
+
+1. This will result in an API call every time a user views a document; and
+2. Carrying out an INSERT query on a GET request isn't a pattern we want to encourage
+
+## Solution 2: Add a new message consumer in Whitehall
+
+This would involve setting up a new RabbitMQ message topic in Publishing API that sends 
+messages when a content block update triggers a change to a document. This would be a brand new 
+topic that contains a thin message that includes the `content_id` of the document that has
+been updated, when it was updated and information about the content block that triggered the update:
```
ryanb-gds, line 100: This interesting. In my imagination we were going to send just a message for the content block and then leave Whitehall to do the lookup of all the documents that link to the block. However, I realise now that would mean searching through govspeak because Whitehall doesn't have a structured reference to content blocks. I can see now why the original plan was to use the existing message, because we'd now have to send two messages for each document. Hmm.

Whitehall does a thing where it parses out the references from Govspeak to contacts and to other editions each time an edition is saved ([link](https://github.com/alphagov/whitehall/blob/main/app/services/service_listeners/edition_dependencies_populator.rb)). We could do the same with content block references perhaps, but I have not really thought through the consequences of that (performance etc.). It would allow us to just send the content block ID to Whitehall and update the document history though.

ryanb-gds (commented):
Thanks for adding to this Stu. Would be interested to get your thoughts on the below comment.

docs/adr/0005-keep-document-history-in-sync-with-rabbit-mq.md
pezholio: We have got a similar thing (contained within the [Content Block Tools Gem](https://github.com/alphagov/govuk_content_block_tools)) that we use to extract references from Govspeak (It's actually [used in Whitehall at the mo](https://github.com/alphagov/whitehall/blob/50acb8baa46c0fe3da0872486421ad53bc6b922e/app/helpers/admin/admin_govspeak_helper.rb) too), so that's entirely a thing we could do

pezholio:
Superseded by https://github.com/alphagov/whitehall/pull/9754

PR #9668, opened 2024-11-28
(707) Add Content Block guidance to Whitehall in non-production environments

pezholio:
To allow user testing of the "insert" journey for Content Modelling, we need to add a link to the Content Block Manager in the Guidance section of the sidebar when editing content. This is enabled by a feature flag, which is set to `true` by default in Staging/Integration.

## Screenshot

image


pezholio:
https://trello.com/c/XImApU0u/707-update-whitehall-to-allow-insert-journey

ChrisBAshton (approved):

PR #9674, opened 2024-11-29
Remove uses of advisory 

hannako:

### DIFF
--- a/db/data_migration/20241024103203_remove_advisory_govspeak_from_edition_body.rb
+++ b/db/data_migration/20241024103203_remove_advisory_govspeak_from_edition_body.rb
@@ -1,0 +1,142 @@
+CONTENT_CONTAINING_ADVISORY_GOVSPEAK = %w[ae74f681-0eac-4de0-bedc-b4c1703ba5f5
+                                          06a0dc8c-3e2a-46fb-ac4c-655ffd7f0591
+                                          133d6a44-b47d-4698-8b89-033d4a73e19d
+                                          e61139a5-ad01-4ab7-bfb1-719e41e8894a
+                                          1c84cb91-e607-46a5-abb2-6cc4eb206666
+                                          de6b61eb-a4b9-4d73-a577-84637ad51233
+                                          0ce9a360-cd01-4188-993d-91ac80885a95
+                                          e5a4fc6a-dbc7-4061-b93e-f226aa1d2775
+                                          1594fbb4-6d9f-4c4e-9d5e-c318e6ab46f2
+                                          7f254eb4-57ea-4044-8236-4964d9d19e19
+                                          aba62e67-b9b7-487f-98b8-4f2a9dcb8759
+                                          ba7d1732-962b-4685-aa47-0b2d303c5b7b
+                                          461c089f-746e-4918-a987-e4bd80b9a608
+                                          bcae8bb2-f364-4436-8b42-e9926b55254a
+                                          9711f631-4056-49bb-87b9-f65b47fd7368
+                                          329e94f5-9105-453f-8708-d41582eb9301
+                                          62a6b167-120b-4197-be3c-bf24f1cef129
+                                          965bc1a6-22e8-4d6c-87e3-573e1a71ce24
+                                          bea7ebde-98a8-439f-a3c8-2ba688a72de2
+                                          2c355840-93c0-49f6-9c3a-92a7f1798716
+                                          7e34d4ae-2236-4fae-9d7d-9262bcfb2186
+                                          6bdd03c6-cbb6-4996-b341-27a9a7d313be
+                                          d8100dcc-93a9-4c0a-b7ac-c97ef393dcd5
+                                          4e2eca0c-6d23-4482-af26-763a248eeadd
+                                          1d4cdc27-2575-494d-bd9e-61b6990b5647
+                                          377b8abf-10d0-45fa-a7b2-9d2b10d70a01
+                                          433f8123-b314-4154-9a7d-6309fcadb0b7
+                                          b7beb9ed-db49-4aec-9d28-dcab1a566886
+                                          f4e5f66d-e758-4f1b-a58f-c3d6f27607a1
+                                          689b56c4-5799-4407-8e62-a5384fe8e41b
+                                          9d28e1d7-9a82-4dbc-8879-2f675b42902b
+                                          0147506d-2355-415a-960b-d0d2fd975af6
+                                          8f9e4a1e-32d4-4f42-bf5a-210fe99d82c2
+                                          adda2cc0-e956-4912-ad07-10f55151c182
+                                          6407ed1c-f94c-4a36-ae00-3730bf9b0e51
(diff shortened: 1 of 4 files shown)


lib/govspeak/remove_advisory_service.rb
```diff
@@ -0,0 +1,58 @@
...
+        "^#{content}^"
+      end
+      if new_body != body
+        record.update!(body: new_body)
+      end
+    end
+
+  private
+
+    def create_minor_update(new_body)
```
hannako, line 27: If I understand it correctly, we don't need this method as we are not going to be handling drafts that contain advisory markup anymore - is that right?

If so you could simplify this class, and just initialise it with an `edition`? 
- You could have a method that returns the regex, rather than passing it in as it's not needed elsewhere, only in this class. 
- You could still have the `@body` variable if you did
```
def initialize(object)
  @body = object.body
end

attr_reader :body
```

lib/tasks/remove_advisory.rake
```diff
@@ -1,152 +1,101 @@
...
-      if new_body != body
-        edition.update!(minor_change: true)
-
-        edition.update!(body: new_body)
-
-        puts "Modified body for #{slug}"
-      end
+    published_content_containing_advisory_govspeak.each do |document_id|
+      edition = Document.find(document_id).latest_edition
+      Govspeak::RemoveAdvisoryService.new(document_id, regex).process!
```
hannako, line 22: Are you meant to be passing in the edition here, rather than the document_id?

lib/govspeak/remove_advisory_service.rb
```diff
@@ -0,0 +1,58 @@
...
+  class RemoveAdvisoryService
+    attr_reader :object, :regex, :whodunnit
+
+    def initialize(object, regex, whodunnit = User.find_by(name: "GDS Inside Government Team"))
+      @object = object
+      @regex = regex
+      @whodunnit = whodunnit
+    end
+
+    def process!
```
hannako, line 11: I can't figure out how this method could be working. I think `Regexp.last_match(1)` is referencing the `@` rather than the content after the `@`. Should it be `Regexp.last_match(2)` ?

I'm wondering if using the string methods `scan` and `gsub` is making it hard to feel confident about what you're working with, because you have to remember that each subexpression in a `()` in your regexp equates to an object in the array that scan generates. If you'd used the Regexp method `match` it outputs the matches as a hash that you can [reference by key](https://ruby-doc.org/3.3.6/Regexp.html#class-Regexp-label-Named+Captures). The downside would be it only matches once, rather than searching the entire body, so for content with more than one advisory you'd need to think about recursion for example.

TLDR: 
- If you've had this code working with `scan` and `gsub` then stick with it, but please can you add comprehensive tests to show that your regex will match the different types of advisory markup that exists (eg @ only at the start, @ at the start and the end etc), and replace it as expected. 

- If you're struggling to get this to work, I wrote it using `Regexp.match` just to understand the shape of the problem, and I think it works so I could share that with you to see what you think...but I don't want to de-rail if you're almost there! So just let me know.

lib/govspeak/remove_advisory_service.rb
```diff
@@ -0,0 +1,58 @@
...
+      body = object.body
+      matches = body.scan(regex)
+      return if matches.empty?
+
+      new_body = body.gsub(/(^@)([\s\S]*?)(@?)(?=(?:^\$CTA|\r?\n\r?\n|^@|$))/m) do
```
hannako, line 16: I've had a go and splitting your regex:

```
def regexp_for_advisory_markup
  opening_at = '(^@)'
  content_after_at = '([\s\S]*?)'
  closing_at = '(@?)'
  other_possible_line_ends = '(?:^\$CTA|\r?\n\r?\n|^@|$)'
  Regexp.new(opening_at + content_after_at + closing_at + '(?=' + 
  other_possible_line_ends + ')', Regexp::MULTILINE)
end
```

hannako (commented):
I opened your branch as a draft PR so it was easier to review! Hope that's ok.

Great work on figuring out that regex! So mind melting, but think I got there in the end and had a go and splitting it out - see what you think. 

I've left some comments - happy to chat through or pair, whatever is best for you.

db/data_migration/20241024103203_remove_advisory_govspeak_from_edition_body.rb
```diff
@@ -0,0 +1,142 @@
+CONTENT_CONTAINING_ADVISORY_GOVSPEAK = %w[ae74f681-0eac-4de0-bedc-b4c1703ba5f5
```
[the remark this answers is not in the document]
beccapearce: Yes this will be removed!

lib/govspeak/remove_advisory_service.rb
beccapearce: I think it would be good to keep this in, as it will match our flow for normal editions. So the changes will appear in the audit history for the doc

lib/govspeak/remove_advisory_service.rb
beccapearce: YES PLEASE! A solution with `.match` would be so much better! I think this was the issue I was trying to fix in my tests! <3

lib/govspeak/remove_advisory_service.rb
```diff
@@ -0,0 +1,101 @@
...
+          # Find the attachment in the new draft
+          new_attachment = draft.html_attachments.find_by(slug: @object.slug)
+
+          # Replace advisories in the body of the new attachment
+          new_body = replace_all_advisories(new_attachment.body)
+          new_attachment.govspeak_content.update!(body: new_body)
+
+          # Update the draft edition and publish
+          draft.update!(minor_change: true)
+          draft.submit!
```
hannako, line 45: You could probably DRY up the `process!` method by creating a ` submit_and_publish(draft)`

lib/tasks/remove_advisory.rake
```diff
@@ -0,0 +1,68 @@
+namespace :remove_advisory_from_editions do
+  desc "Process advisory govspeak in published editions"
+  task published_editions: :environment do
+    regex = Regexp.new(/(^@)([\s\S]*?)(@?)(?=(?:^\$CTA|\r?\n\r?\n|^@|$))/m).to_s
```
hannako, line 4: There's an `EmbeddedContentPatterns` module in Whitehall that contains regex's used by the various extractor classes in lib/Govspeak. I think we should follow the same pattern here rather than defining the regexp in two files.

i.e. create a `Govspeak::AdvisoryPattern` module and include it in both your new service and your rake task

ChrisBAshton (commented):
Looks good in principle 👍 some comments below.

beccapearce:
Thank you Chris! I've added the changes as a separate commit for now, just to make it easy to rereview. I'll squash them together before we get ready to merge.

catalinailie (approved):

hannako:
I don't think I can approve this PR Becca, as I opened it on your draft branch when you were off. But it looks great to me!

PR #9680, opened 2024-11-29
Content modelling/735 homepage date filter

Harriethw:
https://trello.com/c/auF0TsMT/735-update-homepage-add-last-updated-date-filter

Screenshot 2024-11-29 at 13 00 59

---

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/index/date_filter_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/index/date_filter_component.html.erb
@@ -1,0 +1,50 @@
+<%= render("components/datetime_fields", {
+  heading_size: "s",
+  field_name: "last_updated_from",
+  prefix: "last_updated_from",
+  date_heading: "From",
+  date_only: true,
+  year: {
+    id: "last_updated_from_1i",
+    name: "last_updated_from[1i]",
+    label: "Year",
+    width: 4,
+  },
+  month: {
+    id: "last_updated_from_2i",
+    name: "last_updated_from[2i]",
+    label: "Month",
+    width: 2,
+  },
+  day: {
+    id: "last_updated_from_3i",
+    name: "last_updated_from[3i]",
+    label: "Day",
+    width: 2,
+  },
+}) %>
+<%= render("components/datetime_fields", {
+  heading_size: "s",
+  field_name: "last_updated_to",
+  prefix: "last_updated_to",
+  date_heading: "To",
+  date_only: true,
+  year: {
+    id: "last_updated_to_1i",
+    name: "last_updated_to[1i]",
+    label: "Year",
+    width: 4,
+  },
+  month: {
+    id: "last_updated_to_2i",
+    name: "last_updated_to[2i]",
+    label: "Month",
+    width: 2,
+  },
+  day: {
+    id: "last_updated_to_3i",
+    name: "last_updated_to[3i]",
+    label: "Day",
+    width: 2,
+  },
+}) %>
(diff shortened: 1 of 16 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/scopes/searchable_by_updated_date.rb
```diff
@@ -0,0 +1,10 @@
+module ContentBlockManager
+  module ContentBlock::Document::Scopes::SearchableByUpdatedDate
+    extend ActiveSupport::Concern
+
+    included do
+      scope :from_date, ->(date) { where("content_block_documents.updated_at >= ?", date) }
```
Harriethw, line 6: so it looks like editing an Edition updates the Document (even if we never get to the publish/`latest_edition` stage).
I think a solution would be doing a query here to get the current `latest_edition` (which should always be the latest published/scheduled on) and checking that date.

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/index/date_filter_component.html.erb
```diff
@@ -34,17 +37,20 @@
     name: "last_updated_to[1i]",
     label: "Year",
     width: 4,
+    value: @filters && @filters[:last_updated_from] && @filters[:last_updated_to]["1i"],
```
pezholio, line 42: Should this be:

```suggestion
    value: @filters && @filters[:last_updated_to] && @filters[:last_updated_to]["1i"],
```

pezholio: I'm also wondering if this logic can be inside the component class itself to be a bit more DRY. You could also use [`dig`](https://apidock.com/ruby/v2_4_6/Hash/dig) Something like:

```
def date_value(date, date_part)
  filters&.dig(date, date_part)
end
```

And then call it in the view like:

```
value: date_value(:last_updated_to, "1i")
```

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/document_filter.rb
```diff
@@ -22,23 +46,25 @@ def is_date_present?(date_key)
...
 
     def to_date
       @to_date ||= if is_date_present?(:last_updated_to)
-                     filter = @filters[:last_updated_to]
-                     year = filter["1i"].to_i
-                     month = filter["2i"].to_i
-                     day = filter["3i"].to_i
-                     Time.zone.local(year, month, day).end_of_day
+                     date = date_from_filters(:last_updated_to)
+                     Time.zone.local(date[:year], date[:month], date[:day]).end_of_day
```
pezholio, line 67: I added the call to `end_of_day` rather than using `Time.zone.local` with the time specifically set. Sorry, should have been a suggestion!

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/document_filter.rb
```diff
@@ -1,15 +1,39 @@
...
+      end
+    end
+
   private
 
+    def validate_date(key)
+      return unless is_date_present?(key)
+
+      date = date_from_filters(key)
+      Time.zone.local(date[:year], date[:month], date[:day])
```
pezholio, line 32: This should error if there are missing values right? It's just the commit message seems to suggest otherwise

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/document_filter.rb
```diff
@@ -1,15 +1,39 @@
...
     def paginated_documents
       unpaginated_documents.page(page).per(default_page_size)
     end
 
+    def valid?
+      return @valid if defined?(@valid)
+
+      @valid ||= begin
+        validate_date(:last_updated_from)
+        validate_date(:last_updated_to)
```
pezholio, line 20: Could we make these into variables and add an error if `to` is after `from`?

pezholio (commented):
I've updated the commit where we introduce the scope - it should work now, and reviewed the rest. Just some suggestions and a bit of confusion around the error handling. I think we're almost there with that though.

PR #9682, opened 2024-12-02
Remove thumbnail generation code

ChrisBAshton:
The [frontend rendering apps now handle attachments directly, and just use generic icons](https://github.com/alphagov/whitehall/issues/7480#issuecomment-2165235596). The code to generate a custom PNG thumbnail of a PDF was removed in #9163, but Whitehall was still generating and uploading thumbnails unnecessarily: it was simply falling back to a static image asset, [pub-cover.png](https://github.com/alphagov/whitehall/blob/a89d37baab91e38ed6e6d66be16d4a7e282d6efc/app/assets/images/pub-cover.png). Every time a PDF was uploaded, this static asset would be uploaded under a new filepath to Asset Manager (which presumably has many thousands of this identical thumbnail now).

This PR removes the thumbnail variant logic from PDF attachments altogether. The code was spread across a large number of places, so I've tried to break it into commits and explain why each chunk of code is safe to delete. I'd appreciate a second opinion particularly on the test deletion commits though.

Trello card: https://trello.com/c/vjSowrs7/3224-remove-thumbnail-generation-from-whitehall

---


test/unit/lib/whitehall/asset_manager_storage_test.rb
```diff
@@ -53,22 +53,6 @@ class Whitehall::AssetManagerStorage::FileTest < ActiveSupport::TestCase
     assert_equal "http://assets-host/media/asset_manager_id/sample.docx", model.file.url
   end
 
-  test "returns file url using asset_manager_id when the model has an asset variant" do
```
lauraghiorghisor-tw, line 56: Should work: 
```
  test "returns file url using asset_manager_id when the model has an asset variant" do
    model = build(:image_data)
    model.save!
    model.reload

    assert_equal "http://assets-host/media/asset_manager_id_s216/s216_#{model.filename}", model.file.url(:s216)
  end
```

The original itself is technically a variant too, but i'd keep both tests. 

Info dump: I think at some point we had two big test `context`s going, for `AttachmentData` and `ImageData` respectively, but it was a bit repetitive, albeit thorough. There isn't much difference in how the models behave in the context of this class, so happy to leave this as is without adding in more `ImageData` or other `_Data` tests.

test/unit/lib/whitehall/asset_manager_storage_test.rb
```diff
@@ -53,22 +53,6 @@ class Whitehall::AssetManagerStorage::FileTest < ActiveSupport::TestCase
...
-    model = build(:attachment_data, attachable: build(:draft_edition, id: 1))
-    model.save!
-    model.reload
-
-    assert_equal "http://assets-host/media/asset_manager_id_thumbnail/thumbnail_greenpaper.pdf.png", model.file.url(:thumbnail)
-  end
-
-  test "returns nil when the model has assets but the requested variant is not available" do
```
lauraghiorghisor-tw, line 64: Bit of a confusing one, can't really find a way to rewrite it, and unsure it was doing much. OK to delete.

Info dump from hazy memories: When implementing the new asset IDs, as opposed to legacy paths, we needed a way to not send broken url for variants that may still be uploading, to publishing-api. What would have been nice, but we were unable to do, was to get the url method to just return `nil` when any of these asset variants is not available. And then presumably rerun the workers and patch that data on subsequent retries. 
I believe it was some carrierwave internal logic that causes the `path` value to be returned if `url` is nil, as per the test `returns store path when the model has no assets, although it should (still uploading or error has occurred)`. 
Whilst the ubiquitous method `all_asset_variants_uploaded?` is also used to cause some worker retries, if we could return a proper nil here, that might allow us to simplify the usage of that method. It is used heavily in presenters to ensure we don't send any half-broken attachment url data to publishing-api. Of course, famously failing in the `Person` same name file case.

app/models/attachment_data.rb
```diff
@@ -61,9 +61,8 @@ def indexable?
 
   def all_asset_variants_uploaded?
     asset_variants = assets.map(&:variant).map(&:to_sym)
-    required_variants = pdf? ? AttachmentUploader.versions.keys.push(:original) : [Asset.variants[:original].to_sym]
 
-    (required_variants - asset_variants).empty?
+    (%i[original] - asset_variants).empty?
```
lauraghiorghisor-tw, line 65: Could still use the `Asset.variants[:original].to_sym` syntax. Alternatively, keeping the`AttachmentUploader.versions.keys.push(:original)` logic here future-proofs us from when we might reintroduce a variant to that uploader.

test/unit/app/models/attachment_data_test.rb
```diff
@@ -444,11 +437,4 @@ class AttachmentDataTest < ActiveSupport::TestCase
 
     assert_not attachment_data.all_asset_variants_uploaded?
   end
-
-  test "#all_asset_variants_uploaded? returns false if some asset variants are missing" do
```
lauraghiorghisor-tw, line 448: Covered in the test above (about the original, which is technically a variant). Happy to delete ✅

test/factories/attachment_data.rb
```diff
@@ -7,7 +7,6 @@
 
       after(:build) do |attachment_data|
         attachment_data.assets << build(:asset, asset_manager_id: "asset_manager_id_original", variant: Asset.variants[:original], filename: attachment_data.filename)
-        attachment_data.assets << build(:asset, asset_manager_id: "asset_manager_id_thumbnail", variant: Asset.variants[:thumbnail], filename: "thumbnail_#{attachment_data.filename}.png")
```
lauraghiorghisor-tw, line 10: I recall that, at the time, we added a variety of factories to give us plenty to work with for both tests where "we only care if it has an asset, or not even an asset" and tests where "it would be really good to know that update fires twice". Now that we only have the one variant you could perhaps get rid of the doc factory and the `attachment_data_with_asset` and replace it with pdf. `csv` may have its own use cases so we should probably keep.

ChrisBAshton: Good point - thanks

lauraghiorghisor-tw: Had a quick look, definitely can remove the `attachment_data_with_asset` factory.

app/models/attachment_data.rb
ChrisBAshton: > Could still use the `Asset.variants[:original].to_sym` syntax

Could do, but that seems unnecessarily cryptic. We've started just using the `:original` symbol directly [in other places in Whitehall](https://github.com/alphagov/whitehall/blob/a656267d0bb9fb3e1ccd240cc02d7d64a2e5a2d2/app/models/image_data.rb#L41) so I think it's the direction of travel we want to go in.

> Alternatively, keeping the `AttachmentUploader.versions.keys.push(:original)` logic here future-proofs us from when we might reintroduce a variant to that uploader.

True, but also feels like YAGNI to me. We can always bring the code back in future if we need variants again.

test/integration/attachment_deletion_integration_test.rb
```diff
@@ -14,7 +14,6 @@ class AttachmentDeletionIntegrationTest < ActionDispatch::IntegrationTest
       let(:first_asset_id) { "asset_manager_id" }
       let(:second_attachment) { build(:file_attachment, attachable: edition) }
       let(:second_asset_id_original) { "asset_manager_id_original" }
-      let(:second_asset_id_thumbnail) { "asset_manager_id_thumbnail" }
```
lauraghiorghisor-tw, line 17: Just as a catch-all comment on all these integration tests, they need love, they've always needed love. I should hope they will get some love when we revisit them in the bug fix card.

test/unit/lib/whitehall/asset_manager_storage_test.rb
ChrisBAshton: Thank you - that's the winning combination of code I couldn't quite get to earlier! ⭐ Have replaced my commit.

lauraghiorghisor-tw (commented):
LGTM and OMG so many deletions!! (frantic happy clapping)

- I suggested one test fix that should work, ok with the deletions of the other tests you pointed out. Overall sadness over the occasionally convoluted integration tests, but out of scope for this PR to try and make them better.
- Also suggested another factory deletion that could live in this PR. 

There is one thing that crossed my mind, and am unsure if it is an issue, maybe we can huddle about it. Is there any risk associated with maintaining the data in WH? We still have ALL the thumbnail assets. For documents that are published, that's only potentially an issue when we republish, but not sure if it will cause some problems for editions in draft.

Looking at: 
```
    attachment_data.assets.each do |asset|
      AssetManager::AssetUpdater.call(asset.asset_manager_id, asset_attributes)
    end
```
-> it will always try and loop through all associations. I think a little well orchestrated data migration might be needed.

test/factories/attachment_data.rb
ChrisBAshton: Done!

test/integration/asset_access_options_integration_test.rb
```diff
@@ -256,7 +256,7 @@ def setup_publishing_api_for(edition)
 
     def add_file_attachment_with_asset(filename, to:)
       to.attachments << FactoryBot.build(
-        :file_attachment_with_asset,
```
lauraghiorghisor-tw, line 265: My preferred fix would be to change the `asset_manager_id: "asset_manager_id_original"` to asset_manager_id: "asset_manager_id" which should hopefully allow us to replace `file_attachment_with_asset` with `file_attachment` - the generated URLs should be the same. Unless it, of course, breaks everything. 

I like the above better as it doesn't make me question why a csv was chosen for some tests. The csv usage activates my spidey-senses unnecessarily. And with this approach, csvs would only be used in the csv-relevant tests. But it doesn't really matter which doc type you use for the tests. I can live with this.

lauraghiorghisor-tw: I really hoped that we don't have to spend time looking too much into these tests as they are time sinks, but if we make these changes it might be needed. For example this tests is now making a csv file but we pass it a doc as filename (above in the code).

test/integration/attachment_deletion_integration_test.rb
```diff
@@ -10,7 +10,7 @@ class AttachmentDeletionIntegrationTest < ActionDispatch::IntegrationTest
   describe "attachment deletion" do
     context "given a draft document with multiple file attachments" do
       let(:managing_editor) { create(:managing_editor) }
-      let(:first_attachment) { build(:file_attachment_with_asset, attachable: edition, title: "first attachment") }
+      let(:first_attachment) { build(:csv_attachment, attachable: edition, title: "first attachment") }
```
lauraghiorghisor-tw, line 13: This one works fine with two different types of files actually ✅

lauraghiorghisor-tw (approved):
✅
with a few mentions: 
- small option test fix / the suggestion from that comment
- do we need to get rid of the thumbnail assets from the db first?
- should we do a republish?

PR #9691, opened 2024-12-03
(699) Update view page

pezholio:
This updates the view page to make some text tweaks, as well as allowing the schedule to be edited.

## Mural

image

## Code

image

Trello card: https://trello.com/c/nYOtTtr2/699-update-view-page

### DIFF
--- a/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/application.scss
+++ b/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/application.scss
@@ -4,3 +4,4 @@
 @import "components/timeline-component";
 
 @import "views/content_block_manager_header";
+@import "views/content_block_manager_show";
--- a/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/components/_timeline-component.scss
+++ b/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/components/_timeline-component.scss
@@ -62,7 +62,8 @@
 
 .timeline__latest {
   @include govuk-font($size: 19, $weight: bold);
-  background: #ffdd00;
+  color: govuk-colour("white");
+  background: govuk-colour("blue");
   padding: 0.25rem 0.5rem;
   font-size: 1rem !important;
   line-height: 1.25 !important;
--- a/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/views/_content_block_manager_show.scss
+++ b/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/views/_content_block_manager_show.scss
@@ -1,0 +1,8 @@
+.support-request-wrapper {
+  display: flex;
+
+  .govuk-button--secondary {
+    margin-left: auto;
+    margin-top: govuk-spacing(6);
+  }
+}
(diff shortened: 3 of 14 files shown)


lib/engines/content_block_manager/app/lib/content_block_manager/publishable.rb
```diff
@@ -36,7 +36,12 @@ def schedule_with_rollback
       ActiveRecord::Base.transaction do
         content_block_edition = yield
 
-        content_block_edition.schedule!
+        if content_block_edition.scheduled?
```
Harriethw, line 40: finding all the different possibilities a bit confusing - I guess this is the first time we are actually editing an existing content block rather than just creating a new one 💫 
Would there be an edge case here where we need to check if the content block edition hasn't been published in the mean time? e.g. if the user hadn't refreshed the view page in a while and then clicks to edit...

lib/engines/content_block_manager/app/lib/content_block_manager/publishable.rb
```diff
@@ -21,6 +21,7 @@ def publish_with_rollback(content_block_edition)
           ],
         },
       )
+      dequeue_content_block_edition(content_block_edition) if content_block_edition.scheduled?
```
Harriethw, line 24: should we dequeue all previous editions even if the current one isn't scheduled? Just noticed some weird behaviour 
* if I schedule an edition, then publish a new one
* the scheduled edition will override it when it's publish time comes.
* but it won't override the document title 🤔 
 Maybe a different PR but probably want to do asap?

Harriethw: I wonder if we also need a another `state` for editions that have been dequeued, to avoid any confusion down the line, as otherwise the editions will always be in the state `scheduled` I think? Whitehall have `superseded`

lib/engines/content_block_manager/app/lib/content_block_manager/publishable.rb
pezholio: Maybe in this circumstance we should _only_ let the current edition be edited. If a user clicks to edit an edition that isn't the current edition, we redirect to the current one?

lib/engines/content_block_manager/app/lib/content_block_manager/publishable.rb
pezholio: Oooh, that's a good point. Probably belongs in its own PR though. I'll make a card

pezholio: Have made a card https://trello.com/c/BPc2NY8s/755-dequeue-scheduled-editions-when-publishing-a-new-edition

lib/engines/content_block_manager/app/lib/content_block_manager/publishable.rb
pezholio: That might be a bit tricky actually, as we don't record the latest edition ID until the end of the edit journey, we could have an endpoint like:

```
content-block-manager/content-block/$documentId/schedule/edit
```

Which then finds the latest edition and redirects to the `editions/$editionId/workflow/schedule_publishing` URL

pezholio:
@Harriethw I might just bin off the edit schedule stuff from this PR and do it seperately. We should probably make this a more deliberate act

Harriethw (approved):
👍  (looks like just some feature tests to clean up)

PR #9697, opened 2024-12-04
show errors if confirm not checked on review page

Harriethw:
DO NOT MERGE:  Unless between 11:30-14:00

Because this is not an active record error we have to have some custom behaviour here. Not sure if this is the best way!
Currently the review page is only shown for the create/new journey, so have checked we are on the 'new' step here.

## mural

Screenshot 2024-12-04 at 12 04 11

## code

Screenshot 2024-12-04 at 12 02 55

---

### DIFF
--- a/lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
+++ b/lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
@@ -135,8 +135,15 @@ private
     render "content_block_manager/content_block/editions/workflow/schedule_publishing"
   end
 
+  REVIEW_ERROR = Data.define(:attribute, :full_message)
   def publish
-    new_edition = ContentBlockManager::PublishEditionService.new.call(@content_block_edition)
-    redirect_to content_block_manager.content_block_manager_content_block_workflow_path(id: new_edition.id, step: :confirmation)
+    if params[:step] == NEW_BLOCK_STEPS[:review] && params[:is_confirmed].blank?
+      @confirm_error_copy = "Confirm details are correct"
+      @error_summary_errors = [{ text: @confirm_error_copy, href: "#is_confirmed-0" }]
+      render "content_block_manager/content_block/editions/workflow/review"
+    else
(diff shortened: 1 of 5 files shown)


lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
```diff
@@ -61,6 +61,7 @@ def edit_draft
 
   def review
     @content_block_edition = ContentBlockManager::ContentBlock::Edition.find(params[:id])
+    @review_errors = [REVIEW_ERROR.new(attribute: "is_confirmed", full_message: "Confirm details are correct")]
```
pezholio, line 64: Is this used anywhere?

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
```diff
@@ -119,8 +120,17 @@ def schedule_or_publish
     render "content_block_manager/content_block/editions/workflow/schedule_publishing"
   end
 
+  REVIEW_ERROR = Data.define(:attribute, :full_message)
   def publish
+    if params[:step] == NEW_BLOCK_STEPS[:review] && params[:is_confirmed].blank?
+      raise StandardError
```
pezholio, line 126: Do we need to raise here? Could we not just have the code from line 132 and down inside the conditional, and then wrap the happy path in an else?

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
Harriethw: ah nope

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
Harriethw: true! I guess it doesn't have to be an error, have turned into an if statement

PR #9700, opened 2024-12-04
(746) Update preview email address screen in the edit journey

pezholio:
This makes a bunch of minor copy changes to the preview section of the Content Modelling edit journey

## Screenshot

image

### DIFF
--- a/lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
+++ b/lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
@@ -47,6 +47,11 @@ class ContentBlockManager::ContentBlock::Editions::WorkflowController < ContentB
     end
   end
 
+  def context
+    "Edit content block"
+  end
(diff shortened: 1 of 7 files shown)


lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
```diff
@@ -47,6 +47,11 @@ def update
     end
   end
 
+  def context
```
Harriethw, line 50: this is a good idea!

Harriethw (approved):

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/editions/workflow_controller.rb
pezholio: I do wonder whether we should be a bit better at having microcopy like this in translation files. It's always one of those things that you start off thinking it's not worth doing, then as a project gets bigger you wish you'd have done sooner!

PR #9701, opened 2024-12-04
Add rake task to update missing replacement links

lauraghiorghisor-tw:
We've identified a bug where if the user replaces an attachments on a new draft, without having saved the edition draft first, the supposedly replaced original asset remains live, because it fails to redirect to its replacement.

Rerunning the updater on the original attachment's AttachmentData should send the correct `replacement_id` field to `asset-manager`, and fix the data.

This is only a data patch solution.
A fix for the issue will be implemented separately.

[Trello card](https://trello.com/c/YIs8YZmF/3146-investigate-attachments-on-superseded-editions-that-may-still-be-live)


test/unit/lib/tasks/attachments/fix_missing_asset_replacement_links_test.rb
```diff
@@ -0,0 +1,89 @@
...
+        ERROR - ad: #{attachment_data.id} does not have a replacement.
+      OUTPUT
+
+      output, _err = capture_io { task.invoke(filepath) }
+      assert_equal output, expected_output
+    end
+
+    test "it logs error if update fails" do
+      error = GdsApi::HTTPClientError.new(404, "Not found")
+      Services.asset_manager.expects(:asset).with(original_asset.asset_manager_id).raises(error)
```
minhngocd, line 79: should this be :update_asset rather than :asset?

lib/tasks/attachments/fix_missing_asset_replacement_links.rake
```diff
@@ -0,0 +1,21 @@
...
+    attachment_data = AttachmentData.find(attachment_data_id)
+
+    if attachment_data.replaced_by_id.nil?
+      puts "ERROR - ad: #{attachment_data.id} does not have a replacement."
+      next
+    end
+
+    begin
+      AssetManager::AttachmentUpdater.replace(attachment_data)
+      puts "OK - ad: #{attachment_data_id}" << "\n"
```
minhngocd, line 16: no need to insert via `<<` - can just have 'n directly in the string?

minhngocd (approved):
a few non-blocking comments

test/unit/lib/tasks/attachments/fix_missing_asset_replacement_links_test.rb
lauraghiorghisor-tw: There's actually 2 calls in the updater, one to fetch the asset and one to update it, in this case I am mocking the first one. "Update" is used a tad generically in the test name. I only care that there is some error.

PR #9703, opened 2024-12-04
Support hero blocks without content

richardTowers:
Previously the pattern in BlockFactory would only match hero blocks if they had content. For now, adding another pattern for blocks without content, but this could be refactored more nicely in future I think.


leenagupte (approved):
Thanks!

PR #9715, opened 2024-12-09
(763) Fix mainstream preview bug within guide chapters

pezholio:
Trello card: https://trello.com/c/WZvzmuMI/763-bug-mainstream-preview-bug-within-guide-chapters

This fixes the issue when previewing content types with multiple pages by updating all relative links to link to the preview page with the path added as a query string, allowing any content to be parsed via the `GetPreviewContent` service.

Additionally, because we can't reliably get _all_ the instances within a document when parsing the content randomly, I've updated the call to `instances` to fetch the instance count via the Publishing API (PR here https://github.com/alphagov/publishing-api/pull/3030) rather than relying on the count on the current preview page.

~~This can't be merged until https://github.com/alphagov/publishing-api/pull/3030 is merged~~


Harriethw (approved):
nice!

PR #9730, opened 2024-12-11
Refactor paginated timeline

pezholio:
As discussed in https://github.com/alphagov/whitehall/pull/9666 - the content modelling team will be making some changes to the Paginated Timeline class. It's a little hard to follow at the mo, so I've made some tweaks to separate out the behaviour and make it a littler easier to reason about before adding any additional complexity.


app/presenters/queries/version_presenter.rb
```diff
@@ -0,0 +1,45 @@
+module Queries
```
ryanb-gds, line 1: We typically use presenters for making transformations before API requests, which isn't a major issue.

However, I'm just wondering how frequently this gets used, my suspicion is that it's only used to eventually pass data into the AuditTrailEntry component. I wonder if we could do away with this class and do the presenting in the AuditTrailEntry component which doesn't have much behaviour on it at the moment. May not work but might be worth a try 🤔

app/queries/document/paginated_timeline_query.rb
```diff
@@ -0,0 +1,111 @@
+class Document
```
ryanb-gds, line 1: I'm still finding this file a bit tricky to parse, and I think it's because this query object actually represents four(?) different queries which don't actually need to depend on each other. Would it make sense to lift the remarks and versions queries up to the Document model and pass in the remark/version ID lists as a parameter? Otherwise those queries are quite simple. Then we could leave this file to focus on the gnarly pagination/union behaviour.

ryanb-gds (approved):
I'm approving this because it definitely improves the situation, but I might have a crack at taking things a bit further at some point, because I think the original code is excessively DRY.

Fundamentally there are three queries required I think:

1. Get a paginated list of editorial remarks
2. Get a pagianted list of versions
3. Get a paginated list of the union of editorial remarks and versions

Plus total count for each of the above.

If we were to just write three totally separate queries without trying to share the pagination logic between them, my suspicion is we'd end up with a much simpler piece of code at the cost of a little duplication

pezholio:
Cool, thanks for the review. Will merge as is, but some good thoughts there. I may well tweak as/when I get a chance

PR #9734, opened 2024-12-11
Hide licence in footer

pezholio:
Trello Card: https://trello.com/c/vaeRae01/730-remove-licence-details-from-footer

The licence details make sense on public-facing sites, but we don’t need them on the admin side. We’ve added a `hide_licence` option to the layout_footer component, so let’s use it here.

### Before

image

### After

image


ryanb-gds (approved):
Good spot, thanks Stu!

PR #9745, opened 2024-12-17
Convert all tables to use UTF8MB4 charset

ryanb-gds:
We have seen an increasing number of errors recently relating to publishers attempting to use 4 byte characters in Whitehall. This commit adds a migration which will convert all of the tables to use the UTF8MB4 charset.

The migration will need to be run out of hours as the tables will need to be locked during conversion, meaning publishers will not be able to update content. I will run this in integration one evening to check the downtime required and organise a suitable window to apply the migration to production.

Storing 4 bytes per character can cause issues with VARCHAR and TINYTEXT columns, and with the index sizes for VARCHAR columns. I have checked that we will not be affected using the queries described in [this article](https://www.percona.com/blog/migrating-to-utf8mb4-things-to-consider/).

Trello: https://trello.com/c/aTlLraLX


minhngocd (approved):

ryanb-gds:
Closing this because I'm going to write it as a data migration, which will save me from needing to faff around changing helm charts

PR #9746, opened 2024-12-17
Refactoring: Reduce usages of sti_name to minimum

minhngocd:
`sti_names` methods don't seem to be used anywhere and can be removed.

within Edition::Identifiable:

* When edition is saved, the edition type should be propogated to the document type. Additionally the document_type on an Edition is being used in multiple places so is not easy to unpick. The published as class method is used to fetch document by slug based on the type of the subclass. So unfortunately we still need to use sti_name for this.
* The test however has 1 duplicate unecessary test, as we not only support News and Speeches sharing same slug, but all other document types too, which already has another test above covering that.

https://trello.com/c/EewW5W4A/3264-delete-selfstinames-methods

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -8,10 +8,6 @@ class Announcement < Edition
   include Edition::WorldLocations
   include Edition::TopicalEvents
 
-  def self.sti_names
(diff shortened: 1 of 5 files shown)


ryanb-gds (approved):
🧹

PR #9752, opened 2024-12-18
update logo image hint text

leimdorfer:


leimdorfer:
This adds a hardcoded image_hint to the form, which is then passed to the page to be displayed on the logo uploader instead of the standard image text.
This solves the issue at hand but as a follow up we should:

- Audit for other image uploaders that don't require the 960 by 640 size
- Look into a more strategic image upload approach potentially using the work related to supporting different image sizes https://github.com/alphagov/whitehall/blob/main/config/image_kinds.yml

Note: No new tests added for this change.

minhngocd (approved):
Just amended the comment you made to be part of the commit message. 

Looks good to me 🚀

PR #9754, opened 2024-12-19
(688) Add content block updates to a document's timeline

pezholio:
Trello card: https://trello.com/c/PTDFVt1z/688-pull-in-when-content-has-been-updated-by-an-update-to-a-content-block-within-publishing-apps

As discussed in https://github.com/alphagov/whitehall/pull/9666, we need a way of pulling in content block updates to a document's history. After much to-ing and fro-ing, we have settled on doing this in-app, when fetching the history. This is complicated slightly by the fact that the history is paginated. To compensate for this, we are having to "look ahead" to the next page to get the date of the next event to give us a window to query Publishing API for events.

We also have an extra complication as we don't know what updates "belong" to an edition. We've got around this by adding some event-based helpers to an edition, so we know when an edition was published and superseded. To simplify the logic in the paginated timeline class, we've moved the logic to work out what object belongs to an edition to the model classes themselves.

Additionally, there's a bit more tidying up and refactoring of the history-related work so we're doing away with complexity before adding any more!

Hopefully this isn't too much of a monster to review (especially with 🎄 on the horizon). I've tried to break it down as much as possible, but if you want me to break it up into seperate PRs, I'm happy to do that!

## Screenshot

image


test/unit/app/models/host_content_update_event_test.rb
ChrisBAshton, line 1: Nice tests ⭐

test/unit/app/models/document/paginated_timeline_test.rb
```diff
@@ -17,7 +17,7 @@ class PaginatedTimelineTest < ActiveSupport::TestCase
       assert_equal entry_timestamps.sort.reverse, entry_timestamps
     end
 
-    it "is is a list of VersionDecorator and EditorialRemark objects when no 'only' argument is passed in" do
+    it "is is a list of VersionDecorator and EditorialRemark objects when no 'only' argument is not passed in" do
```
ChrisBAshton, line 20: This is difficult to parse 😅 can we reword to remove the double negative?

Similarly on line 92

app/models/document/paginated_timeline.rb
```diff
@@ -89,4 +95,30 @@ def prev_page
...
     end
   end
+
+private
+
+  def next_page_entries
+    if next_page
+      Document::PaginatedTimelineQuery.new(document:, page: next_page, only:).raw_entries
+    end
+  end
```
ChrisBAshton, line 87: I'd move this below `def date_window` as it isn't called until further down in the file (nice to have methods defined in the order they're called)

app/models/document/paginated_timeline.rb
```diff
@@ -89,4 +95,30 @@ def prev_page
...
+      document:,
+      from: date_window.last,
+      to: date_window.first,
+    )
+  end
+
+  def date_window
+    @date_window ||= begin
+      start = page == 1 ? Time.zone.now : query.raw_entries.first.created_at
+      ends = next_page_entries ? next_page_entries.first.created_at : query.raw_entries.last.created_at
```
ChrisBAshton, line 102: The `next_page` logic is surprising - can you add to the commit detail what you've added to the PR description?

```
This is complicated slightly by the fact that the history is paginated. To compensate for this, we are having to "look ahead" to the next page to get the date of the next event to give us a window to query Publishing API for events.
```

Better yet, perhaps just handle the first page in this first commit, and then add the "next page" logic in a subsequent commit, rather than doing it all at once.

app/models/editorial_remark.rb
```diff
@@ -3,4 +3,16 @@ class EditorialRemark < ApplicationRecord
   belongs_to :author, class_name: "User"
 
   validates :edition, :body, :author, presence: true
+
+  def is_for_newer_edition?(edition)
```
ChrisBAshton, line 7: Seems a shame to have to duplicate this - did you consider any alternatives (e.g. extract into a module?)

I don't feel strongly about it though, agree that duplication is better than the wrong abstraction! And at least the code is simple.

app/models/document/paginated_timeline.rb
```diff
@@ -35,33 +35,15 @@ def query
   end
 
   def entries_on_newer_editions(edition)
-    @entries_on_newer_editions ||= entries.select do |entry|
-      if entry.is_a?(EditorialRemark)
-        entry.edition_id > edition.id
-      else
-        entry.item_id > edition.id
-      end
-    end
```
ChrisBAshton, line 38: Niiice ⭐

ChrisBAshton (approved):
Nice and easy to follow 👍 some suggestions for code refactoring and commit structure, but nothing blocking.

app/decorators/version_decorator.rb
```diff
@@ -0,0 +1,45 @@
+class VersionDecorator < SimpleDelegator
```
ryanb-gds, line 1: I think it would be nice if the name of this decorator said something about why we have chosen to use a decorator rather than simply adding these behaviours on to the version model. If I understand your intentions correctly, it's because we only expect these behaviours to be useful in the context of constructing a document history timeline, but I'm struggling to think of a way to communicate that pithily via a class name

app/models/editorial_remark.rb
ryanb-gds: I agree with Stu's duplication choice - the fact that the algorithm for sorting editorial remarks and versions is identical seems like coincidental rather than fundamental duplication to me.

ryanb-gds (commented):
Thought I'd chuck in my two penneth as well,  hope you don't mind.

On a macro level I do look forward to revisiting the entirety of how document history works at some point, it has always felt a lotharder than it needs to be for some reason

ryanb-gds (approved):
Looks ready to go to me, thanks for taking on the feedback

PR #9756, opened 2024-12-19
Have CompoundBlock extract subblocks

richardTowers:
... instead of doing it with pattern matching in BlockFactory.

We want these to be optional in most cases, but putting them in the patterns means we only use the correct block type if they're present. Instead, we just use type: in the pattern, and have the blocks take care of finding their config inside source.

This cleans up the following piece of tech debt: https://trello.com/c/zfUEbZdo/224-simplify-compound-block-models-in-whitehall

### DIFF
--- a/test/unit/app/models/landing_page/featured_block_test.rb
+++ b/test/unit/app/models/landing_page/featured_block_test.rb
@@ -5,7 +5,6 @@ class FeaturedBlockTest < ActiveSupport::TestCase
     @valid_featured_images = [
       build(:image, image_data: build(:landing_page_image_data, file: upload_fixture("landing_page_image.png", "image/png"))),
     ]
-    @valid_featured_content_blocks = [{ "type" => "some-block-type" }]
     @valid_featured_block_config = {
       "type" => "featured",
       "image" => {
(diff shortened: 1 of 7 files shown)


test/unit/app/models/landing_page/featured_block_test.rb
```diff
@@ -69,13 +68,19 @@ class FeaturedBlockTest < ActiveSupport::TestCase
...
+      @valid_featured_images,
+    )
     assert subject.valid?
   end
 
   test "invalid when featured content blocks are invalid" do
     invalid_blocks_config = [{ "invalid" => "because I do not have a type" }]
-    subject = LandingPage::FeaturedBlock.new(@valid_featured_block_config, @valid_featured_images, invalid_blocks_config)
+    subject = LandingPage::FeaturedBlock.new(
+      @valid_featured_block_config.tap { _1["featured_content"]["blocks"] = invalid_blocks_config },
```
ryanb-gds, line 81: I find this `_1` really unpleasant, excited for being able to use `it` instead in Ruby 3.4 🎉

ryanb-gds (approved):
Nice

test/unit/app/models/landing_page/featured_block_test.rb
richardTowers: I've come to quite like it, mostly through trying to golf my way through advent of code. Agree `it` will be nicer.

I'd be happy to do this if it's less smelly:

```suggestion
      @valid_featured_block_config.tap { |config| config["featured_content"]["blocks"] = invalid_blocks_config },
```

richardTowers: Or I could do this I think:

```suggestion
      @valid_featured_block_config.merge("featured_content" => { "blocks" => invalid_blocks_config }),
```

PR #9764, opened 2024-12-20
Create a maintenance page for Whitehall

ryanb-gds:
We sometimes need periods of planned downtime when maintaining Whitehall, for example when performing long running database migrations.

This commit adds a basic maintenance page that can be toggled on and off using a feature flag

Trello: https://trello.com/c/otYKhOEa


ChrisBAshton (approved):
Quick work! ⭐

PR #9765, opened 2024-12-20
Remove "important board members" image constraint

ChrisBAshton:
Sister PRs:
- https://github.com/alphagov/publishing-api/pull/3057
- https://github.com/alphagov/collections/pull/3919

This feature was added in #4162 (counterpart in Collections: https://github.com/alphagov/collections/pull/754). Neither PR really explains/justifies the feature: it appears to have been added as a workaround for a specific organisation, where they'd neglected to add an image for one of their board members. Instead of adding an image for them, this feature was added to impose an artificial limit on the number of images to display on the page.

This has since caused numerous support tickets for the publishing team, with publishers struggling to understand why board member photos are missing from the org page. The limit, tucked away in the organisation settings, is not well known and is not intuitive.

We've decided to remove the feature, removing complexity from both Whitehall and Collections. It should result in fewer support tickets for the team, and be a net positive for organisations in terms of an easier-to-understand system. If we find there is a requirement to re-introduce something like this in future, we can approach it more holistically. For now, the simplest thing is to remove the feature, as we suspect it's not widely used.

Trello: https://trello.com/c/mz0WLkIR/3282-remove-display-management-team-images-feature

---

### DIFF
--- a/app/components/admin/organisations/show/summary_list_component.rb
+++ b/app/components/admin/organisations/show/summary_list_component.rb
@@ -35,7 +35,6 @@ private
       topical_events_row,
       featured_links_position_row,
       featured_links_row,
-      management_team_row,
       foi_exempt_row,
       analytics_identifier_row,
     ]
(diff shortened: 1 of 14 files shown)


ryanb-gds (approved):
Good to see this gone, most peculiar feature

PR #9766, opened 2024-12-20
Remove default value for jobs url

GDSNewt:
By default organisation pages, except No 10, would have a link to civilservicejobs show up in the corporate info segment. We had a request to remove this from the DFT, as they do not publish jobs on that site. 

We have decided to make this section 'opt in', users will still be able to link to civilservicejobs by explicitly stating they want to when editing the Org's details - the 'Recruitment URL' value. This will make the 'Jobs' link reappear. 

**BEFORE** - with default jobs link
Screenshot 2024-12-20 at 14 40 23

**AFTER** - no link given
Screenshot 2024-12-20 at 13 50 30

[Trello](https://trello.com/c/cS9RH4B6/3265-remove-default-civil-service-jobs-link-from-organisation-pages)
[Zendesk](https://govuk.zendesk.com/agent/tickets/5869890)

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -9,7 +9,6 @@ class Organisation < ApplicationRecord
 
   date_attributes(:closed_at)
 
-  DEFAULT_JOBS_URL = "https://www.civilservicejobs.service.gov.uk/csr".freeze
   FEATURED_DOCUMENTS_DISPLAY_LIMIT = 6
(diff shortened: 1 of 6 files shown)


ChrisBAshton (approved):
LGTM 👍 pleasantly surprised no frontend changes are needed! I'd squash the two commits though.

PR #9767, opened 2024-12-20
Convert all tables to use UTF8MB4 charset

ryanb-gds:
We have seen an increasing number of errors recently relating to publishers attempting to use 4 byte characters in Whitehall. This commit adds a data migration which will convert all of the tables to use the UTF8MB4 charset. This uses a data migration because it is not subject to the 15 minute limit for schema migrations applied in the kubernetes configuration.

The migration will need to be run out of hours as the tables will need to be locked during conversion.

Trello: https://trello.com/c/aTlLraLX

### DIFF
--- a/db/data_migration/2024121711524600_convert_text_columns_to_utf8mb4.rb
+++ b/db/data_migration/2024121711524600_convert_text_columns_to_utf8mb4.rb
@@ -1,0 +1,11 @@
+# Convert all tables to utf8mb4 in order to improve support for non-English languages.
+
+connection = ActiveRecord::Base.connection
+# Disable foreign key constraints because foreign keys using strings will prevent conversion due to a charset mismatch.
+connection.execute "SET foreign_key_checks = 0;"
+connection.tables.each do |table|
+  puts "START: Converting table #{table} to utf8mb4"
+  connection.execute "ALTER TABLE `#{table}` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;"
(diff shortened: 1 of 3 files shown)


db/data_migration/2024121711524600_convert_text_columns_to_utf8mb4.rb
```diff
@@ -0,0 +1,13 @@
+# Convert all tables to utf8mb4 in order to improve support for non-English languages.
+
+connection = ActiveRecord::Base.connection
+# Disable foreign key constraints because foreign keys using strings will prevent conversion due to a charset mismatch.
+connection.execute "SET foreign_key_checks = 0;"
+connection.tables.each do |table|
+  next if table == "schema_migrations"
```
ChrisBAshton, line 7: Probably worth a quick inline comment on why this is necessary

ChrisBAshton (approved):
LGTM. Please drop commits 2 and 3 though, not much value in committing those.

db/data_migration/2024121711524600_convert_text_columns_to_utf8mb4.rb
ryanb-gds: That's a good point, this is a data migration, so this isn't necessary any longer. I think I'm right in saying Active Record migrations lock the migrations table (or used to? don't know if they still do actually) while the migration is running to stop multiple migrations running simultaneously. This would prevent the conversion from running for the table. Data migrations don't appear to do that, so we no longer need that line.

PR #9775, opened 2024-12-27
Allow bulk republishing of World Locations

brucebolt:
World Locations are not rendered on GOV.UK so do not have a rendering app or base path.

However, when first presented to Publishing API, they were rendered by Whitehall Frontend and had a base path.
    
In 865661ef5f7df596b590126cc9cb1069d0d39349, the name of the rendering app was removed from the presenter, but as no replacement value was provided (i.e. nil) we retained the value of Whitehall Frontend in Publishing API.
    
In 1166bb92d545e658bc1e893a44973d545e6b2e26, we removed the `PublisesToPublishingApi` from World Locations to stop them conflicting with World Location News which shared the same base path. This meant the World Locations could never be republished again.
    
Recently all published editions with a rendering app of Whitehall Frontend [were deleted](https://github.com/alphagov/publishing-api/pull/3025) from Publishing API, meaning the World Locations were also deleted.  We therefore need to republish all World Locations again.
    
Therefore adding a nil value to the rendering app, so they can be republished to Publishing API and the legacy Whitehall Frontend value overwritten, and adding a nil value to the base path so they do not conflict with the World Location News documents. Finally, reintroducing `PublishesToPublishingApi` so the world locations can be republished through the republishing user interface.

Depends on https://github.com/alphagov/publishing-api/pull/3061.

[Trello card](https://trello.com/c/6jD9m6Sl)

### DIFF
--- a/app/helpers/admin/republishing_helper.rb
+++ b/app/helpers/admin/republishing_helper.rb
@@ -162,6 +162,7 @@ module Admin::RepublishingHelper
       TakePartPage
       TopicalEvent
       TopicalEventAboutPage
+      WorldLocation
       WorldLocationNews
     ]
   end
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -42,6 +42,8 @@ class WorldLocation < ApplicationRecord
 
   scope :ordered_by_name, -> { with_translations(I18n.default_locale).order("world_location_translations.name") }
 
+  include PublishesToPublishingApi
+
   def self.active
     where(active: true)
   end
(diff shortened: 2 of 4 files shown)


ChrisBAshton (approved):
LGTM 👍 and then presumably we revert this once the republish has happened?

brucebolt:
I think we should leave it in place.  There's no UI for editing world locations, but it'll save having to figure out what code to add back into the app should we need to update a country name in the future.

brucebolt:
Looking at the docs for [renaming a country](https://docs.publishing.service.gov.uk/manual/rename-a-country.html#3-update-whitehall), there's no step to republish the World Location to Publishing API, so they must have been written to rely on `PublishesToPublishingApi` to perform that action.

PR #9783, opened 2025-01-03
Use gds api adapters for search indexing

ryanb-gds:
Trello: https://trello.com/c/FUjE91gQ

This will improve our chances of being able to diagnose issues with Whitehall search indexing, as it will enable the use of GOVUK Request ID for tracing requests.

I dropped some of the logging of each request, as they were only info logs and essentially provide the same information as the request logs from Search API. Duplicating that information from Whitehall didn't make sense, plus the logging was not a high enough log level to be logged in production in any case.


app/services/search_api_service.rb
```diff
@@ -1,33 +0,0 @@
-class SearchApiService
```
ChrisBAshton, line 1: Needs removing from https://github.com/alphagov/whitehall/blob/fe6d2a9fa95fdc175a15f8b02b8137152f8b743d/test/support/search_api_helper.rb#L16-L20 too

ChrisBAshton (approved):
Nice one ⭐ you've spotted some good opportunities for code deletion here!

PR #9788, opened 2025-01-06
Add title to Content Block Editions, stage 1

Harriethw:
https://trello.com/c/kGVJtSCJ/777-bug-block-title-is-saved-even-when-edit-form-is-abandoned 

We want to move a Content Block's `title` field from the `Document` to the `Edition` model - this is because the title should be versionable, as it can change with each update of a block. Having it belong to the Document is causing the bug described in Trello, and we lose the audit history of the title over a block's lifetime. 

To do this without affecting live data I'm suggesting three steps:

### step 1
PR1 (this PR):

* Add a `title` field to CB Editions table - not nullable, but default to an empty string?

* Temporarily rename the existing Edition `title` method to `document_title` so we can keep referring to it while the db gets populated.

* Rake task to populate Editions with string from their Document `title`

* When Edition is created via UI, ensure the Edition `title` is updated as well as document one.

### step 2

Run the rake task, which will mean all old Editions have a populated `title` field with whatever the current Document `title` is.  
NOTE: this will mean old Editions might have a 'wrong' title on them, e.g. they may have been saved originally with a different title, I don't think this matters at this point, as we have no way of knowing what the original title was? And we are not yet using the tool properly in production.

### step 3

PR2:

* Point Document’s `title` method to latest Edition

* Remove `title` column from Documents table

* Remove `title` from Document attributes setting in Create service

---

### DIFF
--- a/db/migrate/20250106111536_add_title_to_content_block_editions.rb
+++ b/db/migrate/20250106111536_add_title_to_content_block_editions.rb
@@ -1,0 +1,13 @@
+class AddTitleToContentBlockEditions < ActiveRecord::Migration[7.1]
+  def up
+    change_table :content_block_editions, bulk: true do |t|
+      t.string "title", default: "", null: false
+    end
+  end
+
+  def down
+    change_table :content_block_editions, bulk: true do |t|
+      t.remove :title
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_12_20_143730) do
+ActiveRecord::Schema[7.1].define(version: 2025_01_06_111536) do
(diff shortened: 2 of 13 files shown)


db/migrate/20250106111536_add_title_to_content_block_editions.rb
```diff
@@ -10,4 +10,4 @@ def down
       t.remove :title
     end
   end
-end
\ No newline at end of file
+end
```
pezholio, line 13: I think this can be squashed into the original commit?

pezholio (commented):
LGTM - seems like a sensible way of going about things!

pezholio (approved):
LGTM - like the use of a dry run too!

db/migrate/20250106111536_add_title_to_content_block_editions.rb
Harriethw: oh yep, good spot!

PR #9790, opened 2025-01-07
Ensure worldwide org office contacts are unpublished effectively

ryanb-gds:
Contacts should be unpublished without a redirect, as they do not have a base path. We therefore invoke the gone worker instead of the redirect worker where the content item does not response to the base_path method.

This was preventing the unpublishing of a world organisation.

I'd desperately love to refactor that associated document handling stuff but it'll take a while, so leaving it alone for now and just throwing another conditional at the problem 😢 

One thing I can't understand - this resulted in a user facing error because the Publishing API unpublishing request returned a 500 response. However, that should have failed in a Sidekiq job. Are we not calling our Sidekiq jobs correctly? 🤔 

Trello: https://trello.com/c/U4QZKdQ7

### DIFF
--- a/app/services/service_listeners/publishing_api_associated_documents.rb
+++ b/app/services/service_listeners/publishing_api_associated_documents.rb
@@ -55,22 +55,30 @@ module ServiceListeners
                     end
 
       current_associated_documents.each do |associated_document|
-        if associated_document.respond_to?(:translations)
-          associated_document.translations.each do |translation|
+        locales_to_unpublish =
+          if associated_document.respond_to?(:translations)
+            associated_document.translations.map(&:locale).map(&:to_s)
+          else
+            [locale_for_document(associated_document).to_s]
+          end
+
+        locales_to_unpublish.each do |locale|
+          if associated_document.respond_to?(:base_path)
             PublishingApiRedirectWorker.new.perform(
               associated_document.content_id,
               destination,
-              translation.locale.to_s,
+              locale,
+              allow_draft,
+            )
+          else
+            PublishingApiGoneWorker.new.perform(
+              associated_document.content_id,
+              nil,
+              nil,
+              locale,
               allow_draft,
             )
           end
-        else
-          PublishingApiRedirectWorker.new.perform(
-            associated_document.content_id,
-            destination,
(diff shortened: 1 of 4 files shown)


app/services/service_listeners/publishing_api_associated_documents.rb
```diff
@@ -57,20 +57,38 @@ def unpublish(allow_draft: false)
...
             allow_draft,
           )
+        else
+          PublishingApiGoneWorker.new.perform(
+            associated_document.content_id,
+            nil,
+            nil,
+            locale_for_document(associated_document).to_s,
+            allow_draft,
+          )
```
ChrisBAshton, line 91: Just trying to think of a slightly less painful way of expressing this. Do you think this would work? It doesn't really seem any more complicated/risky than the currently proposed duplication/branching.

```ruby
locales_to_unpublish =
  associated_document.respond_to?(:translations) ?
    associated_document.translations.map(&:locale).map(&:to_s)
    :
    [locale_for_document(associated_document).to_s]

locales_to_unpublish.each do |locale|
  if associated_document.respond_to?(:base_path)
    PublishingApiRedirectWorker.new.perform(
      associated_document.content_id,
      destination,
      locale,
      allow_draft,
    )
  else
    PublishingApiGoneWorker.new.perform(
      associated_document.content_id,
      nil,
      nil,
      locale,
      allow_draft,
    )
  end
end
```

ChrisBAshton (approved):
Nice one 👍 

RE why the UI shows an error - I don't know either 🤷 
I think we'll need to tackle https://trello.com/c/ZluRj3ec/19-make-publishing-api-interactions-more-consistent to make this sort of head-scratcher go away.

ryanb-gds:
> Nice one 👍
> 
> RE why the UI shows an error - I don't know either 🤷 I think we'll need to tackle https://trello.com/c/ZluRj3ec/19-make-publishing-api-interactions-more-consistent to make this sort of head-scratcher go away.

I'd forgotten how Sidekiq works. Instantiating the job and calling perform just runs the perform method which is defined on the job. In some places we do this and run it inline, in some places we use `PublishingApiRepublishingWorker.perform_async` to run it as a Sidekiq job.

PR #9791, opened 2025-01-07
content-modelling/777 - remove title from Content Block Document and use the title on Content Block Editions

Harriethw:
https://trello.com/c/kGVJtSCJ/777-bug-block-title-is-saved-even-when-edit-form-is-abandoned

Unfortunately most of the changes are done in one mega-commit, I couldn't figure out how best to split it up. 

- Rename `title` column from Documents table to `sluggable_string` to preserve the functionality of the `content_id_alias` (and copy what is done on Whitehall)
- assign the `sluggable_string` whenever we create a new Document. 
- no longer update the Document title from the form data, instead update the Edition title.
- Point Document’s `title` method to latest Edition - so we have a helper method to get the `title` straight from Doc.

---


lib/engines/content_block_manager/app/models/content_block_manager/content_block/document.rb
```diff
@@ -6,7 +6,7 @@ class Document < ApplicationRecord
       include Scopes::SearchableByUpdatedDate
 
       extend FriendlyId
-      friendly_id :title, use: :slugged, slug_column: :content_id_alias, routes: :default
+      friendly_id :sluggable_string, use: :slugged, slug_column: :content_id_alias, routes: :default
```
pezholio, line 9: FWIW I think it's fine to not update the slug when we update the title, as they should be immutable. This will be particularly important when we move over to use slugs over the UUIDs in the embed codes, though I _think_ we can have multiple aliases for blocks.

docs/diagrams/content_block_manager_models.puml
```diff
@@ -17,6 +16,7 @@ together {
         details
         document_id
         state
+        title
```
pezholio, line 19: Good catch! I would have totally forgotten about this! 🤦

pezholio (approved):
LGTM - the test fail looks like some flakiness introduced by me, if you rerun it should work. I'll look at fixing the flake now.

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document.rb
Harriethw: do you think the `content_id_alias` could just be assigned directly then, rather than going through the `sluggable_string`? I wasn't sure what was going on with the FriendlyId stuff tbh.

docs/diagrams/content_block_manager_models.puml
Harriethw: I think it's all a bit out of date but at least this field won't be 😆

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document.rb
Harriethw: I'll leave as it is for now as I don't want to unpick the friendlyID stuff or risk it getting out of sync with how the other aliases are generated elsewhere, but we can always return to it

PR #9796, opened 2025-01-08
Reorder social media worldwide org

GDSNewt:
We've had requests to introduce an ordering feature for social media accounts on worldwide organisations.

[Zendesk](https://govuk.zendesk.com/agent/tickets/5474377)
[Trello](https://trello.com/c/AGtaWRkW/3285-request-to-reorder-social-media-accounts-for-worldwide-organisations)


app/controllers/admin/worldwide_offices_controller.rb
```diff
@@ -94,13 +95,8 @@ def handle_show_on_home_page_param
...
 
-  def extract_items_from_ordering_params
-    item_ordering = params[:ordering] || {}
-    item_ordering.permit!.to_h
-      .map { |item_id, ordering| [WorldwideOffice.find_by(id: item_id), ordering.to_i] }
-      .sort_by { |_, ordering| ordering }
-      .map { |item, _| item }
-      .compact
+  def fetch_ordered_items
+    @ordered_items = extract_items_from_ordering_params(WorldwideOffice)
```
ryanb-gds, line 99: Does this need to be an instance variable?

app/controllers/admin/editionable_social_media_accounts_controller.rb
```diff
@@ -46,6 +47,15 @@ def update
     end
   end
 
+  def reorder
+    if request.put?
```
ryanb-gds, line 51: Still not sure about serving two routes from the same controller method, would be interested to get the rest of the team's thoughts on that. I don't think it's a major issue but I don't think I've seen this done much in other places in Whitehall

app/models/social_media_account.rb
```diff
@@ -24,4 +24,9 @@ def service_name
   def display_name
     title.presence || service_name
   end
+
+  def update_order(new_order)
```
ryanb-gds, line 28: I would probably just use the [update](https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update) method from whatever object calls this - or, if you find there should be a transaction, you might want to figure out how to do a bulk update to the ordering column.

ryanb-gds (commented):
A few thoughts from me - on the face of it I think there should be a database transaction involved for the reordering, but I have a vague recollection of discussing this with @davidgisbey and that wasn't viable for some reason 🤔

ryanb-gds (approved):
I think I'm happy with this

PR #9797, opened 2025-01-08
content modelling/793 accessible autocomplete

Harriethw:
https://trello.com/c/t8Yu5ImI/793-combobox-lead-organisations-names-should-be-announced-with-a-text-reader-severe

Screenshot 2025-01-08 at 17 44 57

Following an accessibility audit, the
select_with_search component was found not to meet
WCAG requirements - "Users should be able to
access this list using a keyboard or other input
methods, with each item being announced along with
a count of the total number of options."

It seems like on Safari Voice Over would not
read out the selected option at all, and in other
browsers would not read out the total count of
options.

I'm replacing the component with the autocomplete
component, which does not use the `Choices`
library and is based on the alphagov accessible
autocomplete, and seems to provide the same
functionality and the required Voice Over
behaviour.

---

### DIFF
--- a/app/views/components/docs/select_with_search.yml
+++ b/app/views/components/docs/select_with_search.yml
@@ -1,6 +1,10 @@
 name: Select with search
 description: A dropdown select with search
 body: |
+  NOTE: this component currently fails WCAG compliance for keyboard navigation, as the total number of options are not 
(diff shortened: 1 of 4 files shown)


pezholio (approved):
LGTM 👍

PR #9801, opened 2025-01-09
Fix multiple replace asset bug

lauraghiorghisor-tw:
[Trello](https://trello.com/c/31CtEP8h/3249-fix-missing-replacement-causing-assets-to-be-live-when-they-should-not-be)


lauraghiorghisor-tw:
Worked, but a risky change, as future create or update callbacks would not work without a context.

PR #9806, opened 2025-01-13
Stop validating change notes for draft editions

ryanb-gds:
We only want to make sure that editions have a change note at the point that they are submitted for publication (or force published/scheduled).

Validating change notes for draft editions could cause bugs. For example, if the user replaced an attachment before adding a change note to the edition, the replacement silently failed, leaving the old attachment live.

We therefore change the validation logic to only validate non-draft editions.

Unfortunately this required us to temporarily update the state of the editions in the edition publisher and scheduler services. Those services trigger the edition validation before the state is updated, so the validation occurs as if the editions are still drafts, which isn't logical behaviour.

Trello: https://trello.com/c/31CtEP8h - note that this is how we fix the bug, even if it looks unrelated

### DIFF
--- a/features/visual-editor.feature
+++ b/features/visual-editor.feature
@@ -11,7 +11,6 @@ Feature: Save edition content with visual editor
     When I fill in the required fields for publication "Publication with visual editor" in organisation "Visual ministry"
     And I save and go to document summary
     Then I should see the visual editor on subsequent edits of the publication
-    And I force publish the publication "Publication with visual editor"
 
   @javascript
   Scenario: I create a new publication and exit the visual editor experience
--- a/test/functional/admin/generic_editions_controller_test.rb
+++ b/test/functional/admin/generic_editions_controller_test.rb
@@ -100,15 +100,4 @@ class Admin::GenericEditionsControllerTest < ActionController::TestCase
     get :show, params: { id: draft_edition }
     assert_select ".govuk-link", text: "Preview on website (opens in new tab)", href: draft_edition.public_url(draft: true)
   end
-
-  view_test "GET :show doesn't render preview link if publically visible, change note is blank and edition is a major version" do
-    published_edition = create(:published_edition)
-    draft_edition = create(:draft_edition, change_note: nil, minor_change: false, document: published_edition.document)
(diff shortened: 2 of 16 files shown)


features/visual-editor.feature
```diff
@@ -11,7 +11,6 @@ Feature: Save edition content with visual editor
     When I fill in the required fields for publication "Publication with visual editor" in organisation "Visual ministry"
     And I save and go to document summary
     Then I should see the visual editor on subsequent edits of the publication
-    And I force publish the publication "Publication with visual editor"
```
ryanb-gds, line 14: This failed because the test was set up invalidly before (the publication is invalid due to missing attachment) but the change note stuff was obscuring the failure. In any case it doesn't seem necessary for the test scenario to prove what it was intended to prove, so I've removed it

test/functional/admin/generic_editions_controller_test.rb
```diff
@@ -100,15 +100,4 @@ class Admin::GenericEditionsControllerTest < ActionController::TestCase
...
-  view_test "GET :show doesn't render preview link if publically visible, change note is blank and edition is a major version" do
-    published_edition = create(:published_edition)
-    draft_edition = create(:draft_edition, change_note: nil, minor_change: false, document: published_edition.document)
-    stub_publishing_api_expanded_links_with_taxons(draft_edition.content_id, [])
-
-    get :show, params: { id: draft_edition }
-
-    assert_select ".govuk-link", text: "Preview on website (opens in new tab)", href: draft_edition.public_url(draft: true), count: 0
-    assert_select ".govuk-inset-text", text: "To see the changes and share a document preview link, add a change note or mark the change type to minor."
-  end
```
ryanb-gds, line 113: There's no reason not to render the preview link in this scenario, so removing this test

ryanb-gds:
Closing this because it requires us to manipulate the `state` of the edition directly, ignoring the API provided by the transitions gem. This can cause weird side effects, such as not triggering validation callbacks depending on the transition state machine state. Unfortunately this happens in quite a few places in Whitehall 😞

PR #9807, opened 2025-01-13
Ensure asset update happens at least once

lauraghiorghisor-tw:
The `CreateAssetWorker` relies on the `DraftUpdaterWorker` successfully running in order to trigger a call to the `MetadataWorker` for editionable attachables. This happens due to the `AttachmentUpdater` service listening on the edition `draft_update` event.

However, if the edition is not valid, e.g. the attachment has been updated/replaced before saving the edition with a change note, then the `DraftUpdaterWorker` will not run, and the MetadataWorker will not be triggered. This means the state of the asset in Asset Manager will not be updated.

There are additional attempts that try to achieve the same result, that also fail:
- The `save_attachment` method in the `AttachmentsController` calls the `draft_updater`, which again fails, as above, because the edition is invalid, causing the `MetadataWorker` to not be enqueued.
- Although the `attachment_updater` in the `AttachmentsController` also triggers the MetadataWorker, it typically fails because the assets are not yet ready. There is currently no retry mechanism in place for this worker.

The current change ensures that the `MetadataWorker` runs after the asset has been created irrespective of whether or not the edition is valid. This ensures that we send an update to Asset Manager.

Related PRs for alternative solutions we considered (now closed): #9801 and #9806.

[Trello](https://trello.com/c/31CtEP8h/3249-fix-missing-replacement-causing-assets-to-be-live-when-they-should-not-be)

### DIFF
--- a/app/sidekiq/asset_manager_create_asset_worker.rb
+++ b/app/sidekiq/asset_manager_create_asset_worker.rb
@@ -42,14 +42,16 @@ private
   end
 
   def enqueue_downstream_service_updates(assetable_id, assetable_type, attachable_model_class, attachable_model_id)
-    assetable = assetable_type.constantize.find(assetable_id)
-    assetable.republish_on_assets_ready if assetable.respond_to? :republish_on_assets_ready
-
     if attachable_model_class
       if attachable_model_class.constantize.ancestors.include?(Edition)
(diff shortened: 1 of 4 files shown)


ryanb-gds (approved):
Just keep running those workers 🥵 💪

PR #9811, opened 2025-01-14
Maintenance banner

ryanb-gds:
We want be able to switch a maintenance banner on to warn users of upcoming downtime.

Trello: https://trello.com/c/qjrtsn2Y

### DIFF
--- a/config/locales/en/admin/maintenance_banner.yml
+++ b/config/locales/en/admin/maintenance_banner.yml
@@ -1,0 +1,6 @@
+en:
+  admin:
+    maintenance_banner:
+      show_banner: true
+      message: |
+        Whitehall Publisher will be unavailable on 3 February from 8pm to 10pm for essential maintenance. Scheduled publishing may be delayed.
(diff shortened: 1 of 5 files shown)


config/locales/en/admin/maintenance_banner.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  admin:
+    maintenance_banner:
+      show_banner: true
+      message: |
+        Whitehall Publishing will be unavailable to all users for approximately 1-2 hours on Mon 3 Feb from 8pm. This is to carry out essential maintenance. Thank you for your patience.
```
ChrisBAshton, line 6:
```suggestion
        Whitehall Publishing will be unavailable to all users for approximately 1-2 hours on Mon 3 Feb from 8pm. This is to carry out essential maintenance. Scheduled publications may also be delayed during this period. Thank you for your patience
```

ChrisBAshton (approved):
LGTM 👍

PR #9815, opened 2025-01-15
Content modelling/814 allow nested fields in content block manager

pezholio:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #9819, opened 2025-01-15
Use components for details items

pezholio:
This was originally part of the draft PR https://github.com/alphagov/whitehall/pull/9815, but as we may not be using nested fields now, I've extracted the more useful code out of that feature, as it will still be helpful to have around.

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/base_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/base_component.rb
@@ -1,0 +1,37 @@
+class ContentBlockManager::ContentBlockEdition::Details::Fields::BaseComponent < ViewComponent::Base
+  include ErrorsHelper
+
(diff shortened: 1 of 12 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/base_component.rb
```diff
@@ -0,0 +1,37 @@
...
+    @value = value
+    @id_suffix = id_suffix
+  end
+
+private
+
+  attr_reader :content_block_edition, :field
+
+  def label
+    @label || field.humanize
```
Harriethw, line 19: does `||` also memoize it, or does it need to be `||=`?

Harriethw: ohh I misunderstood what it's doing, nvm :)

Harriethw (approved):
v cool! this is such a good shout and works well 👍

PR #9820, opened 2025-01-16
Apply GA4 tracking to flash alerts

gtvj:
## What

Adds GA4 tracking to flash alerts via the [GA4 Auto Tracker](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/trackers/ga4-auto-tracker.md).

### Expected datalayer

```javascript
event_data':{
    'event_name':'flash_message',
    'text':'<flash text>',
    'action':error
}
```

## Why 

We want to understand when users have disruptions in their journey. 

### Screen / data layer screenshots after change

<details>

<summary>Data attributes <strong>are</strong> present in flash alert</summary>

Screenshot of data attributes present in Flash alerts

</details>

<details>

<summary>Data attributes <strong>are not</strong> present in flash success</summary>

Screenshot 2025-01-16 at 12 23 36

</details>

## Performance Analyst review

Implementation has been tested by PAs in Integration who confirm the data is present as expected.

### DIFF
--- a/app/components/admin/flash_alert_component.rb
+++ b/app/components/admin/flash_alert_component.rb
@@ -2,4 +2,15 @@ class Admin::FlashAlertComponent < Admin::FlashMessageComponent
   def component_name
     "error_alert"
   end
+
+  def data_attributes
+    {
+      module: "ga4-auto-tracker",
+      "ga4-auto": {
+        event_name: "flash_message",
+        text: message,
+        action: "error",
+      }.to_json,
+    }
+  end
 end
--- a/app/components/admin/flash_message_component.html.erb
+++ b/app/components/admin/flash_message_component.html.erb
@@ -1,5 +1,8 @@
 <%=
 content_tag(:div) do
-  render "govuk_publishing_components/components/#{component_name}", { message: }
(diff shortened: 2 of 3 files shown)


dnkrj (approved):

app/components/admin/flash_alert_component.rb
```diff
@@ -2,4 +2,26 @@ class Admin::FlashAlertComponent < Admin::FlashMessageComponent
   def component_name
     "error_alert"
   end
+
+  def page_title
+    if content_for?(:page_title)
+      content_for(:page_title).titleize
+    elsif content_for?(:title)
```
ryanb-gds, line 9: As far as I can tell from a search of the codebase, these should both be set on more or less every page - I'd be interested to know which pages are missing a page title.

gtvj: Thanks, I'll update the code to be a single `if...else` on `page_title`. Was just including the additional check because I'd understood it was populated from either `page_title` or `title`.

ryanb-gds (approved):
📸

PR #9821, opened 2025-01-16
content-modelling/819 fix cancel button for reschedule form

Harriethw:
with other forms, the schedule publishing page appears after a draft Edition has been created, so if a user abandons the form we want to delete the draft Edition.

With rescheduling, we don't want to delete the whole Edition, because the user is just editing the scheduling of an already scheduled Edition. So we just return to the Document view.

NOTE:

in doing this refactoring/fix I noticed that the form is not hydrating properly - I've created another ticket to fix that https://trello.com/c/aDIYdgfJ/837-bug-scheduled-publishing-form-does-not-populate-data-from-existing-block 

--

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/shared/schedule_publishing_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/shared/schedule_publishing_component.html.erb
@@ -105,13 +105,12 @@
           secondary_solid: true,
         } %>
       <% else %>
-        <%= render partial: "content_block_manager/content_block/shared/cancel_delete_button",
-                   locals: {
-                     url: helpers.content_block_manager.content_block_manager_content_block_edition_path(
-                       content_block_edition,
-                       redirect_path: helpers.content_block_manager.content_block_manager_content_block_document_path(content_block_edition.document),
-                       ),
-                   } %>
+        <%= render ContentBlockManager::Shared::CancelAndDeleteButtonComponent.new(url: helpers.content_block_manager.
+              content_block_manager_content_block_edition_path(
+                content_block_edition, redirect_path:
+                  helpers.content_block_manager.content_block_manager_content_block_document_path(content_block_edition.document)
+              ),
+            ) %>
       <% end %>
     </div>
     </div>
--- a/lib/engines/content_block_manager/app/components/content_block_manager/shared/schedule_publishing_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/shared/schedule_publishing_component.html.erb
@@ -16,7 +16,7 @@
   is_scheduled_param = params["schedule_publishing"]
 %>
 
-<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: @content_block_edition)) %>
+<% content_for :error_summary, render(Admin::ErrorSummaryComponent.new(object: content_block_edition)) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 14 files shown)


lib/engines/content_block_manager/app/views/content_block_manager/content_block/shared/_schedule_publishing.html.erb
```diff
@@ -96,13 +96,23 @@
...
     </div>
     <div>
-      <%= render partial: "content_block_manager/content_block/shared/cancel_delete_button",
-                 locals: {
-                   url: content_block_manager.content_block_manager_content_block_edition_path(
-                     @content_block_edition,
-                     redirect_path: content_block_manager.content_block_manager_content_block_document_path(@content_block_edition.document),
-                     ),
-                 } %>
+      <% if is_rescheduling %>
```
Harriethw, line 99: open to a different naming or method for this! or turning into a component so it's easier to test?

pezholio: Hmmm, maybe a component would be good as we're using it in multiple places. Would make it more testable too.

Harriethw: cool cool, I will be a bit less lazy and turn it into a component :)

Harriethw: split out into quite a large a component, but I still think we need the feature tests in place to test the integration between the component and everything else - I've also noticed another bug so maybe there is more refactoring that could be done in this ticket  https://trello.com/c/aDIYdgfJ/837-bug-scheduled-publishing-form-does-not-populate-data-from-existing-block

lib/engines/content_block_manager/app/components/content_block_manager/shared/schedule_publishing_component.html.erb
```diff
@@ -96,13 +96,23 @@
...
+      <% if is_rescheduling %>
+        <%= render "govuk_publishing_components/components/button", {
+          text: "Cancel",
+          name: "Cancel",
+          value: "Cancel",
+          href: back_link,
+          secondary_solid: true,
+        } %>
+      <% else %>
+        <%= render partial: "content_block_manager/content_block/shared/cancel_delete_button",
```
pezholio, line 108: Sorry to be a stickler, but I think this should be a component too, rather than rendering a partial. My instinct is that components should be able to stand on their own and not require knowledge of the actual application.

Harriethw: you mean the `cancel_delete_button` should be a component, or the whole code that's wrapped in the `if` statement?

pezholio: I think the `cancel_delete_button` - referring to a partial outside of components feels like a bit of a smell

pezholio: It doesn't need to be massive, all it needs to accept is a URL argument I think. Then you can use it in the other places where the partial is used

Harriethw: cool, done in two new commits 👍

pezholio (approved):
Nice one - thanks for sticking with me! 👍

PR #9822, opened 2025-01-16
Add copy to homepage header

pezholio:
Trello card: https://trello.com/c/WnDhkF18/824-add-copy-to-homepage-heading

This will help users know where they are and what the purpose of the system is. The copy is the same as that in Signon.

## Screenshot

image


Harriethw:
Looks ok technically, but is the layout from our designers? Doesn't seem to sit quite right on the horizontal... would  having the text spaced out more be a bit more gds-y? this is keeping the gds inherited margins, but changing `&--column-right` to `align-self: flex-start;`
Screenshot 2025-01-16 at 14 40 31

pezholio:
Ah, that's what I had initially, but it looked a bit odd to my eyes. Will pop it up for some feedback

Harriethw (approved):

PR #9828, opened 2025-01-17
Ensure deleted assets are marked as live

lauraghiorghisor-tw:
We want to separate out the catch-all logic we currently have around asset updates. The `AttachmentUpdater` service listener was listening on all possible events, making it hard to understand which specific updates need to happen in which case. Additionally, in the `MetadataWorker` that it enqueued, the update and delete logic fired together. For deletion, similar updates used to fire on attachment deletion, edition publish and edition discard. 

We now understand that, due to the limitations of how we can represent draft/live states in Asset Manager, we can only delete a pre-existing live asset at the point of publish. Any earlier, and the live edition's attachment would become unavailable. Thus, we decided to separate out the Asset Manager update and delete calls that run at edition publication time.

These changes are being made in the context of a bug that causes assets to remain live when a new draft attachment gets replaced and deleted. The issue was that the `deleted?` method on the replaced `AttachmentData` evaluated to true when attempting to update the asset to live after deleting the attachment. By separating the publish logic we are confident in removing the `deleted?` check on update. 

- These changes make it clear that an attachment deletion should only be sent to Asset Manager at the point of publish. 
- Removed the asset deletion/update call from the `AttachmentsController`'s `destroy` method, which only ever succeeded for first draft editions before. 
- Removed the delete calls to Asset Manager that were mixed in with the updates in the `MetadataWorker` - deletion is now only triggered from the publish listener, at the point of publish. 
- We have now extracted what needs to happen on edition discard into a service listener, making it clear that no updates are firing, only asset deletion. 
- Keeping in line with the async implementation we had before (all updates and deletions were run from the `MetadataWorker`), we added jobs inside the listeners.

In addition to the edition workflow changes, we've had to make sure asset deletion also works for Policy Groups, as they are not editionable. 

[Trello](https://trello.com/c/I2Fmtu3L/3185-fix-draft-setting-causing-assets-to-be-live-when-they-should-not-be)


config/initializers/edition_services.rb
```diff
@@ -1,9 +1,17 @@
...
-    coordinator.subscribe do |_event, edition, _options|
+    coordinator.subscribe(/^(?!(force_publish|publish|delete))/) do |_event, edition, _options|
       ServiceListeners::AttachmentUpdater.call(attachable: edition)
     end
 
+    coordinator.subscribe(/^(force_publish|publish)$/) do |_event, edition, _options|
+      ServiceListeners::AttachmentAssetPublisher.call(attachable: edition)
+    end
+
+    coordinator.subscribe(/^(delete)$/) do |_event, edition, _options|
```
ryanb-gds, line 11: Might as well use just `delete` rather than a regex here, as we're subscribing to a single event

ryanb-gds (approved):
Looks good to me, a few comments notwithstanding. I would like a review from @ChrisBAshton on this before we merge it just to see if he likes the direction of travel. Personally though I feel having service listeners and jobs that respond to particular changes in edition state is making things much easier to understand. The old job that tried to work out the changes it needs to make to Asset Manager from the state of the edition was very confusing.

test/integration/attachment_deletion_integration_test.rb
```diff
@@ -31,43 +33,43 @@ class AttachmentDeletionIntegrationTest < ActionDispatch::IntegrationTest
...
           Services.asset_manager.expects(:delete_asset).once.with(first_asset_id)
           Services.asset_manager.expects(:delete_asset).once.with(second_asset_id)
-          assert_equal AssetManagerAttachmentMetadataWorker.jobs.count, 2
 
-          AssetManagerAttachmentMetadataWorker.drain
+          visit admin_news_article_path(edition)
+          click_link "Delete draft"
+          click_button "Delete"
+
+          assert_equal AssetManagerAttachmentMetadataWorker.jobs.count, 0
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Agreed, deleted all the rest from where we had them ;) - must have missed this one

config/initializers/edition_services.rb
```diff
@@ -1,9 +1,17 @@
 Whitehall::Application.config.to_prepare do
   Whitehall.edition_services.tap do |coordinator|
-    coordinator.subscribe do |_event, edition, _options|
+    coordinator.subscribe(/^(?!(force_publish|publish|delete))/) do |_event, edition, _options|
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Fair enough. I checked the event in the block instead.

config/initializers/edition_services.rb
```diff
@@ -1,7 +1,11 @@
 Whitehall::Application.config.to_prepare do
   Whitehall.edition_services.tap do |coordinator|
-    coordinator.subscribe do |_event, edition, _options|
-      ServiceListeners::AttachmentUpdater.call(attachable: edition)
+    coordinator.subscribe do |event, edition, _options|
+      ServiceListeners::AttachmentUpdater.call(attachable: edition) unless %w[publish force_publish].include?(event)
+    end
+
+    coordinator.subscribe(/^(force_publish|publish)$/) do |_event, edition, _options|
+      ServiceListeners::AttachmentAssetPublisher.call(edition)
```
ChrisBAshton: Seems to me that this would be clearer as a conditional within one `coordinator.subscribe` block, otherwise it's not clear that we're going for an either/or situation.

```suggestion
    coordinator.subscribe do |event, edition, _options|
      if %w[publish force_publish].include?(event)
        ServiceListeners::AttachmentAssetPublisher.call(edition)
      else
        ServiceListeners::AttachmentUpdater.call(attachable: edition)
      end
    end
```

app/sidekiq/asset_manager_attachment_metadata_worker.rb
```diff
@@ -6,10 +6,6 @@ def perform(attachment_data_id)
 
     return if attachment_data.blank?
 
-    # this call needs to be here to run when we do an edition discard,
-    # unless we add a custom service for that
```
ChrisBAshton: Just to make the commit sequence easier to follow, I'd consider tweaking the inline comment to something along the lines of "Temporarily calling this here until we add a custom service to do this - see next commit"

config/initializers/edition_services.rb
```diff
@@ -1,13 +1,17 @@
 Whitehall::Application.config.to_prepare do
   Whitehall.edition_services.tap do |coordinator|
     coordinator.subscribe do |event, edition, _options|
-      ServiceListeners::AttachmentUpdater.call(attachable: edition) unless %w[publish force_publish].include?(event)
+      ServiceListeners::AttachmentUpdater.call(attachable: edition) unless %w[publish force_publish delete].include?(event)
```
ChrisBAshton, line 4: Related to my earlier comment on this file - I can see now why you may want to split these out into separate `.subscribe` calls. In which case... what other events remain, apart from force_publish/publish/delete? It might be time to change the `.subscribe` call to only subscribe to the events we actually care about now.

ChrisBAshton (approved):
Nice one for getting your heads around this! ⭐

PR #9831, opened 2025-01-20
Update attachment asset docs

ryanb-gds:
This expands on the publishing part of Whitehall's attachment asset flow, and also includes more discussion of asset replacements.

Will need rebasing onto main after #9828 has merged.

Trello: https://trello.com/c/AOzsXtG9


docs/assets.md
```diff
@@ -47,10 +47,23 @@ a new Asset model was proposed. Find the related ADR for the asset model [here](
...
 ### File attachments upload flow
 *Editionable Content* types such as **News Articles** and **Consultations**, as well as **Policy Groups**, can have attachments.
-Uploads and republishing actions are managed asynchronously in the background by a Sidekiq worker.
+Uploads and republishing actions are managed asynchronously in the background by a Sidekiq job.
 
 ![Attachment Upload Flow](diagrams/asset_attachment_upload_flow.png)
 
+#### File attachment replacements
+
+Asset Manager supports "replacing" existing assets with a new one. Replacement is essentially a special type of redirect for asset to asset redirects (as opposed to the general redirect URL attribute supported by Asset Manager, which redirects assets to non-asset URLs). Replacement is necessary to ensure that bookmarks and links to existing attachment assets still work even if the binary asset file has changed.
```
lauraghiorghisor-tw, line 56: Maybe "redirect for asset-to-asset redirects" - that was odd to read for a second there.

docs/assets.md
```diff
@@ -47,10 +47,23 @@ a new Asset model was proposed. Find the related ADR for the asset model [here](
...
+
+When a new asset is uploaded as a replacement for the previous one, the flow is identical to the upload flow, but the `AssetManagerAttachmentMetadataWorker` also sends a request to Asset Manager to mark the previous asset as replaced. Note that the replacement does not take effect until the new asset is no longer flagged as a draft. This occurs when the edition is published.
+
+### File attachments publishing flow
+Changes to assets are not applied to Asset Manager until an edition is published, because the previous edition may have attachments that still refer to those assets. Consequently, and somewhat surprisingly, we delete assets associated with deleted attachments during publishing. We also remove the draft flag from all assets associated with the attachment. That includes removing the draft flag from deleted assets because, if the deleted draft asset was a replacement for a previous attachment asset, the replacement will not be applied until the draft flag is removed.
```
lauraghiorghisor-tw, line 61: "...with deleted attachments during publishing." - maybe "at publish time"?

And I don't know if speaking about "removing" the draft flag is accurate, as we are setting it to false rather than removing it.

docs/diagrams/asset_attachment_upload_flow.puml
```diff
@@ -1,4 +1,4 @@
-@startuml asset-attachment-upload-flow
+@startuml asset_attachment_upload_flow
```
lauraghiorghisor-tw, line 1: I had a look at this diagram. We are actually missing the additional call that gets made [here](https://github.com/alphagov/whitehall/blob/826b54f3ecbe01e978fbbd6096c3b556855bcfd2/app/controllers/admin/attachments_controller.rb#L41) - but since we're looking to remove that soon it's probably fine. 

Also, I am wondering if these should be renamed to **edition** asset_flow because this would not be what happens to a policy group, for example.

lauraghiorghisor-tw (approved):

docs/assets.md
ryanb-gds: I did consider "at publish time", but felt it was a bit of a lazy way of saying "during". But maybe I'll go for "at the time of" 🤔 

Good point about the draft flag

docs/diagrams/asset_attachment_upload_flow.puml
ryanb-gds: Mmm, good question about the policy group. I might just call them out as an exception.

PR #9832, opened 2025-01-20
(829) Change button copy on review screen

pezholio:
Trello card: https://trello.com/c/e5kbeUe5/829-change-button-copy-on-review-screen

This changes the button copy on the review screen to be `Publish` if the user is planning to immediately publish the content block, or `Schedule` if they're planning to schedule the publication.

### DIFF
--- a/lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
+++ b/lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
@@ -62,4 +62,17 @@ module CanScheduleOrPublish
       scheduled_publication_params["scheduled_publication(3i)"],
     ]
   end
+
+  def review_update_url
+    schedule_publishing = params[:schedule_publishing]
+    scheduled_at = scheduled_publication_params.to_h
+
+    content_block_manager.content_block_manager_content_block_workflow_path(
+      @content_block_edition,
+      step: ContentBlockManager::ContentBlock::Editions::WorkflowController::UPDATE_BLOCK_STEPS[:review_update],
+      schedule_publishing:,
+      scheduled_at:,
+    )
+  end
+
 end
--- a/lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
+++ b/lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
@@ -4,8 +4,8 @@ module CanScheduleOrPublish
   def schedule_or_publish
     @schema = ContentBlockManager::ContentBlock::Schema.find_by_block_type(@content_block_edition.document.block_type)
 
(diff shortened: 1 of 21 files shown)


lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
```diff
@@ -21,15 +21,22 @@ def publish
...
+    case params[:schedule_publishing]
+    when "schedule"
       validate_scheduled_publication_params
 
-      @content_block_edition.assign_attributes(scheduled_publication_params)
-      @content_block_edition.assign_attributes(state: "scheduled")
-      raise ActiveRecord::RecordInvalid, @content_block_edition unless @content_block_edition.valid?
+      @content_block_edition.update!(scheduled_publication_params)
+      if @content_block_edition.valid?(:scheduling)
+        @content_block_edition.save!
```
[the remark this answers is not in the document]
pezholio: I'm not sure that matters though, as it hasn't at that point been sent to Publishing API, so it's not canonically in Publishing API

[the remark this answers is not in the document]
pezholio: Ah, good call. I'll revisit this then. I still think this is the right way to go, because this is going to be more of an issue when we start introducing other screens.

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/documents/schedule_controller.rb
```diff
@@ -8,12 +8,14 @@ def edit
 
   def update
     document = ContentBlockManager::ContentBlock::Document.find(params[:document_id])
-    @content_block_edition = document.latest_edition
-    validate_scheduled_edition
+    previous_edition = document.latest_edition
```
Harriethw, line 11: this is definitely an improvement! only thing is when we duplicate, the state is in `published` and that would stop the cancel/delete jobs working if the user abandoned the form. think it can be solved by setting state as `draft` (also think there must be some test safety nets missing as this wasn't caught, only saw it as I tried it out in the UI)

Harriethw: I would also double check that changing the state on this new edition doesn't create a new `version` and thus an entry on the timeline 🤔 maybe there's a clever rails way of duplicating except for one field?

pezholio: I think that's fine, as it shows as it's superseded on the timeline, which I think is the correct behaviour. We can revisit if it's confusing

Harriethw (approved):
nice improvements! we got there in the end 😆

PR #9834, opened 2025-01-21
Add UI for retagging content

ChrisBAshton:
Adds UI for "bulk retagging documents" (as it is referred to in the [Support government changes](https://docs.publishing.service.gov.uk/manual/government-changes.html#bulk-retagging-documents) docs) to change their lead and supporting organisations. This UI replaces a rake task we had to run in the past, and removes our last CSV-dependent rake task (which has been [problematic since the introduction of Kubernetes](https://docs.publishing.service.gov.uk/manual/running-rake-tasks.html#working-with-csvs-on-kubernetes)).

Trello: https://trello.com/c/o9iYyNxf/623-add-ui-for-re-tagging-organisations-to-content-in-bulk

## Screenshots

Link added to the "More" page:
> Screenshot 2025-01-23 at 15 59 21

"Retag content" form:

> Screenshot 2025-01-23 at 15 59 34

Form with inputs filled in:

> Screenshot 2025-01-23 at 16 00 40

Preview page / summary of changes:
> Screenshot 2025-01-23 at 16 00 52

Confirmation status:
> Screenshot 2025-01-23 at 16 01 03

---

### DIFF
--- a/features/retagging-content.feature
+++ b/features/retagging-content.feature
@@ -6,20 +6,6 @@ Feature: Retagging documents to different organisations
   Background:
     Given I am a GDS admin
 
-  Scenario: Dry run of the retagging - invalid data
-    Given the documents and organisations I am retagging contain errors
-    When I visit the retagging page
-    And I submit my CSV of documents to be retagged
-    Then I should be on the retagging page with my CSV input still present
-    And I should see a summary of retagging errors
-
-  Scenario: Dry run of the retagging - valid data
-    Given the documents and organisations I am retagging exist
-    When I visit the retagging page
-    And I submit my CSV of documents to be retagged
-    Then I can see a summary of the proposed changes
-    And my CSV input should be in a hidden field ready to confirm retagging
-
   Scenario: Full run of the retagging
     Given the documents and organisations I am retagging exist
     When I visit the retagging page
--- a/features/retagging-content.feature
+++ b/features/retagging-content.feature
@@ -1,0 +1,14 @@
+Feature: Retagging documents to different organisations
+  As an editor
+  I want to be able to retag published documents
+  So that they reflect changes to lead and supporting organisations
+
+  Background:
+    Given I am a GDS admin
+
+  Scenario: Dry run of the retagging - invalid data
+    Given the documents and organisations I am retagging contain errors
+    When I visit the retagging page
+    And I submit my CSV of documents to be retagged
+    Then I should be on the retagging page with my CSV input still present
+    And I should see a summary of retagging errors
--- a/features/retagging-content.feature
+++ b/features/retagging-content.feature
@@ -19,3 +19,14 @@ Feature: Retagging documents to different organisations
     And I submit my CSV of documents to be retagged
     Then I can see a summary of the proposed changes
     And my CSV input should be in a hidden field ready to confirm retagging
+
+  Scenario: Full run of the retagging
+    Given the documents and organisations I am retagging exist
+    When I visit the retagging page
+    And I submit my CSV of documents to be retagged
+    Then I can see a summary of the proposed changes
+    And my CSV input should be in a hidden field ready to confirm retagging
+    And when I click "Publish changes" on this retagging screen
+    Then I am redirected to the retagging index page
+    And I see a confirmation message that my documents are being retagged
+    And the changes should have been actioned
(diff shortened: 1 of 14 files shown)


features/step_definitions/retagging_content_steps.rb
```diff
@@ -73,3 +73,40 @@
...
+  #
+  # NB, the `new_supporting_organisations` parameter in `BulkOrganisationUpdater`'s `update_edition`
+  # method DOES have the organisations coming through in the correct order, so it only seems to
+  # go out of order in the `edition.update(supporting_organisations: new_supporting_organisations)`
+  # call.
+  #
+  # NB, the same issue does not affect `lead_organisations`, presumably because the
+  # "edition_organisations" table has a `lead_ordering` integer to track the order. There is no
+  # equivalent ordering tracking for supporting organisations.
+  expect(doc2.latest_edition.supporting_organisations.map(&:slug).sort).to eq(%w[cabinet-office geospatial-commission])
```
ryanb-gds, line 111: Can we just use a different expectation if we don't care about the order? I'm pretty sure it's an established rule of the Whitehall domain that we don't care about the order of supporting organisations, so I think that would be reasonable. There's [assert_same_elements](https://www.rubydoc.info/gems/minitest-rails-shoulda/0.4.1/MiniTest%2FAssertions%3Aassert_same_elements) but I'm not sure if that's part of Minitest by default or not

features/retagging-content.feature
```diff
@@ -19,3 +19,14 @@ Feature: Retagging documents to different organisations
     And I submit my CSV of documents to be retagged
     Then I can see a summary of the proposed changes
     And my CSV input should be in a hidden field ready to confirm retagging
+
```
ryanb-gds, line 22: Now we've got the full run test below, should we consider moving the dry run tests down to the controller test level, if that's feasible? Our cucumber test suite is already very large and I wonder if these tests are of sufficient value for the time cost. Don't feel strongly about this, just a consideration

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -1,26 +1,72 @@
...
+    def validate
+      expected_headers_sorted = ["Document type", "New lead organisations", "New supporting organisations", "Slug"]
+      @validated_rows = @parsed_csv.each_with_index.map do |row, index|
+        if index.zero? && row.headers.compact.sort != expected_headers_sorted
+          @errors << "Expected the following headers: #{expected_headers_sorted.join(',')}. Detected: #{row.headers.join(',')}"
+          break
+        end
+
+        if row.fields.count != 4
+          @errors << "Exactly four fields expected. Detected: #{row.fields.count} (#{row.fields})"
```
ryanb-gds, line 25: Some places we use the attr_accessor for errors, and some places we use the instance var, which always mildly irks me

test/functional/admin/retagging_controller_test.rb
```diff
@@ -0,0 +1,24 @@
...
+  setup do
+    login_as :gds_admin
+  end
+
+  should_be_an_admin_controller
+
+  view_test "GDS Admin users should be able to GET :index and see a textarea for CSV input" do
+    get :index
+
+    assert_select "textarea[name='csv_input']", count: 1
```
ryanb-gds, line 13: I'm a big fan of an assertion that the form has the right action and method in this sort of test as well. Gives the test a nice completeness, IMO

ryanb-gds (approved):
Looks good. A few suggestions for your consideration, but more just me throwing ideas around than any changes I strongly feel we should make

lib/data_hygiene/bulk_organisation_updater.rb
ChrisBAshton: oops - me too

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -34,6 +34,42 @@ def validate
...
+      else
+        if orgs_added.count.positive?
+          status << "Added #{orgs_added.join(', ')}"
+        end
+        if orgs_removed.count.positive?
+          status << "Removed #{orgs_removed.join(', ')}"
+        end
+      end
+      status.join(", ") + ". Result: #{new_orgs.join(', ')}"
+    end
```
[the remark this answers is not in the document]
ChrisBAshton: That's one I'll leave for a future PR I think - the scope of this PR is just to migrate what currently exists as a rake task, into a UI form, and not to look too closely at the current implementation 😅

PR #9835, opened 2025-01-22
Replace title component with heading

andysellick:

## What / why

- replaces one instance of the title component with the heading component
- this instance was using `margin_top: 0` which is an option we're removing from the title component
- instead use the heading component, which when configured like this should look exactly like the title component, but by default has no margin top
- hopefully therefore no visual changes

Related to this change: https://github.com/alphagov/govuk_publishing_components/pull/4508

Trello card: https://trello.com/c/Y0pDWbHw/390-move-some-shared-helper-options-into-component-wrapper


ChrisBAshton (approved):
LGTM

PR #9837, opened 2025-01-22
(828) Add internal and external change notes when editing blocks

pezholio:

This adds screens to add internal change notes and external change notes (if the change is major). As part of this, I've refactored the Workflow controller to (try) and make it easier to follow. There's still scope for improvement, but I think we're getting there!

This doesn't include showing the notes on the confirm / view screens. I'll do this in a separate PR.

## Screenshots

image

image

### DIFF
--- a/lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
+++ b/lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
@@ -77,7 +77,7 @@ module CanScheduleOrPublish
   def review_update_url
     content_block_manager.content_block_manager_content_block_workflow_path(
       @content_block_edition,
-      step: ContentBlockManager::ContentBlock::Editions::WorkflowController::UPDATE_BLOCK_STEPS[:review_update],
+      step: :review,
     )
   end
 
--- a/lib/engines/content_block_manager/app/controllers/concerns/workflow/update_methods.rb
+++ b/lib/engines/content_block_manager/app/controllers/concerns/workflow/update_methods.rb
@@ -6,6 +6,7 @@ module Workflow::UpdateMethods
   UPDATE_ACTIONS = {
     review_links: :redirect_to_schedule,
     schedule_publishing: :validate_schedule,
+    internal_note: :update_internal_note,
     review_update: :validate_review_page,
     review: :validate_review_page,
   }.freeze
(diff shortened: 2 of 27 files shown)


lib/engines/content_block_manager/app/controllers/concerns/workflow/update_methods.rb
```diff
@@ -0,0 +1,76 @@
...
+      id: @content_block_edition.id,
+      step: :change_note,
+    )
+  end
+
+  def update_change_note
+    @content_block_edition.assign_attributes(change_note: edition_params[:change_note], major_change: edition_params[:major_change])
+    @content_block_edition.save!(context: :change_note)
+
+    redirect_to content_block_manager.content_block_manager_content_block_workflow_path(
```
Harriethw, line 48: the back link isn't working when i hit an error on the change note form, does that happen for you?

lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
```diff
@@ -77,7 +77,7 @@ def scheduled_publication_date_params
   def review_update_url
     content_block_manager.content_block_manager_content_block_workflow_path(
       @content_block_edition,
-      step: ContentBlockManager::ContentBlock::Editions::WorkflowController::UPDATE_BLOCK_STEPS[:review_update],
+      step: :review,
```
Harriethw, line 80: I'm getting really confused about what all the steps are and what they do, especially now they are in different files 😅  (though I don't think it's a bad idea to separate them) e.g. the difference between `review` and `review_update` and then also the difference between the `review` in show and the `review` in update. 
* could the names at least be unique between SHOW and UPDATE? 
I'm also wondering if all the actions could be in some kind of hash that would show their relationship/order 🤔 
* is the difference between review and review_update just the back link?

lib/engines/content_block_manager/app/controllers/concerns/workflow/update_methods.rb
pezholio: Ah, it might do actually. I'll fix that

lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
pezholio: I'll give it a go - would be good if we could combine `review` and `review_update` into one action, at the moment, it's a bit icky

pezholio:
@Harriethw I've also noticed the Cancel buttons should probably be delete buttons too, so I'll look at that too

lib/engines/content_block_manager/app/controllers/concerns/can_schedule_or_publish.rb
pezholio: Maybe `review_new_block` and `review_updated_block` might be better naming?

pezholio: Or actually, if there is only one edition, we know it's a new block

Harriethw: yeah those names are better I think, or yeah if it could be one action that would be even better.

Harriethw (approved):
will add my approval as looks good apart from some refactoring suggestions (that could be picked up later if you wanted to get this in)

pezholio:
Yeah, cool. I'll get it merged in and open another PR for the changes 👍

PR #9838, opened 2025-01-22
content modelling/830 show changes

Harriethw:
- **wip add edition diff**
- **add table to document timeline**

### DIFF
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
@@ -28,8 +28,73 @@ module ContentBlockManager
       unless draft?
         user = Current.user
         state = try(:state)
-        versions.create!(event: "updated", user:, state:)
+        versions.create!(event: "updated", user:, state:, changed_fields:)
       end
+    end
+
+    def changed_fields
+      if document.editions.count <= 1
+        nil
+      else
+        @previous_edition = document.editions[-2]
+        diff = []
+        diff.append(diff_for_lead_organisation)
+        changed_attributes = diff_active_record
+        if changed_attributes
+          diff.append(diffs_for_user_generated_attributes(changed_attributes))
+        end
+        diff.compact.flatten!
+      end
+    end
+
+    def diff_active_record
+      user_generated_fields = %w[title instructions_to_publishers details]
+      all_attributes = (@previous_edition.attributes.to_a - attributes.to_a).map(&:first)
+      all_attributes & user_generated_fields
+    end
+
+    def diffs_for_user_generated_attributes(changed_attributes)
+      changes = []
+      changed_attributes.each do |attribute|
+        if attribute == "details"
+          changes.append(diffs_for_edition_details)
+        else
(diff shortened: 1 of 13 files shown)


db/migrate/20250117163617_add_changed_fields_to_content_block_versions.rb
```diff
@@ -0,0 +1,13 @@
+class AddChangedFieldsToContentBlockVersions < ActiveRecord::Migration[7.1]
+  def up
+    change_table :content_block_versions, bulk: true do |t|
+      t.json "changed_fields"
```
Harriethw, line 4: I put this as a json blob first because I wasn't sure what structure it could have or if it would be different for different versions... but then I formalised it as a list of  a Data type `ChangedFields`. If we're happy with this structure I'm thinking I should create a table of ChangedFields with a many to one relationship with Versions, and then the data is enshrined in the db and will be easier to create etc. But then again it would mean any changes (e.g. if the design of the UI changes) might be harder to make 🤔

lib/engines/content_block_manager/app/models/content_block_manager/content_block/version/changed_field.rb
```diff
@@ -0,0 +1,6 @@
+module ContentBlockManager
```
Harriethw, line 1: I don't like the name `ChangedField` - would `EditionFieldDiff` be better? or something else??

lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
```diff
@@ -28,8 +28,73 @@ def record_update
...
       end
     end
+
+    def changed_fields
+      if document.editions.count <= 1
+        nil
+      else
+        @previous_edition = document.editions[-2]
+        diff = []
+        diff.append(diff_for_lead_organisation)
```
Harriethw, line 41: organisation proved to be a bit complicated as I wasn't able to call `lead_organisation` on the previous edition directly...

pezholio: You might be able to get `lead_organisation` if you add an `includes` to the call to `editions`. Something like:

```
editions.includes(:edition_organisation, :organisation)
```

lib/engines/content_block_manager/app/models/content_block_manager/content_block/version/changed_field.rb
pezholio: I think maybe if you embed it in the `ContentBlock::Edition` namespace, so it's `ContentBlockManager::ContentBlock::Edition::FieldDiff` maybe?

db/migrate/20250117163617_add_changed_fields_to_content_block_versions.rb
pezholio: I do wonder what the benefit of saving this stuff would be. I don't think it'll be a massive strain on the DB to do it at load time (unless, of course, the history gets very long), but then that might be time to introduce pagination anyway.

Harriethw:
closing this as have made a new simpler branch

PR #9840, opened 2025-01-23
Set read_consultation_principles to true for all post publication consultation editions

ryanb-gds:
When the validation requiring publishers to confirm that they have considered the consultation principles was introduced in 797909f215a62261ce763a339a515ccb8833ef4f, we did not backdate this to existing consultations. This means that some consultations cannot be withdrawn, unwithdrawn or unpublished because they are not technically valid, despite being public on GOV.UK. This data migration retroactively confirms that all post-publication consultations have considered the consultation principles.

Trello: https://trello.com/c/nf4GMMi7


ChrisBAshton (approved):
Great work! ⭐

PR #9841, opened 2025-01-24
(828) Additional workflow iterations

pezholio:

This simplifies the workflow code further, having one source of truth for the workflow steps, adding some useful helpers to get the next/previous steps, and removing some confusing code.

### DIFF
--- a/lib/engines/content_block_manager/app/helpers/workflow_helper.rb
+++ b/lib/engines/content_block_manager/app/helpers/workflow_helper.rb
@@ -1,0 +1,15 @@
+module WorkflowHelper
+  def back_path(content_block_edition, current_step)
+    if current_step == "review" && content_block_edition.document.is_new_block?
+      content_block_manager.content_block_manager_content_block_documents_path
+    else
+      step = Workflow::Step.by_name(current_step)&.previous_step
+      return nil unless step
+
+      content_block_manager.content_block_manager_content_block_workflow_path(
+        content_block_edition,
+        step: step.name,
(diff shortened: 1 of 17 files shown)


lib/engines/content_block_manager/app/helpers/workflow_helper.rb
```diff
@@ -0,0 +1,15 @@
+module WorkflowHelper
+  def back_path(content_block_edition, current_step)
+    if current_step == "review" && content_block_edition.document.is_new_block?
+      content_block_manager.content_block_manager_content_block_documents_path
```
Harriethw, line 4: now I'm questing why we take users back to the documents page when reviewing a new block - it seems weird, when my browser button just takes me back to the editing form. Is it because we can't actually provide a link back to the form including the data they may have changed?

Harriethw (approved):
lots of improvements here!

lib/engines/content_block_manager/app/helpers/workflow_helper.rb
pezholio: Yeah, I thought that was weird too. I guess we could throw them back to the edit state for that block. We'd just need to be careful that we don't show them the rest of the edit flow. I think it's doable though. Lemme have a look.

pezholio: Hmmm. I think I'll merge as is. I'll make a card to pick it up at a later date.

PR #9842, opened 2025-01-24
content modelling/830 show changes to a content block

Harriethw:
- **add edition's field diff to content block versions**

we want to show the changes that have been made to
a block when updating - this takes a snapshot of
the difference between the user generated fields
and saves them as a json blob, with the key being
the field name.

I'm hesitant to use json as it isn't typed, but I
explored [1] codifying the data structure
previously and it proved too complicated and
inefficient, especially as the fields need to be
displayed to the user in a certain order of
priority, and we don't know exactly what the
fields we're saving will be. By using the Data
class to show the shape of the items, this is
hopefully a halfway house. Shoving everything in a
json blob seems like a middle ground between just
calculating everything at run time and having a
genuine record of the change.

[1] https://github.com/alphagov/whitehall/pull/9838

- **Add content block changes table to timeline**

As design
Screenshot 2025-01-24 at 16 56 02

---

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
@@ -14,6 +14,7 @@ private
         title: title(version),
         byline: User.find_by_id(version.whodunnit)&.then { |user| helpers.linked_author(user, { class: "govuk-link" }) } || "unknown user",
         date: time_html(version.created_at),
+        table_rows: table_rows(version),
       }
     end
   end
@@ -33,5 +34,17 @@ private
       datetime: date_time.iso8601,
       lang: "en",
     )
+  end
+
+  def table_rows(version)
+    if version.field_diffs.present?
+      version.field_diffs.map do |field|
+        [
+          { text: field["field_name"].humanize },
+          { text: field["previous_value"] },
+          { text: field["new_value"] },
+        ]
+      end
+    end
   end
 end
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
@@ -28,7 +28,7 @@ module ContentBlockManager
       unless draft?
         user = Current.user
         state = try(:state)
-        versions.create!(event: "updated", user:, state:)
+        versions.create!(event: "updated", user:, state:, field_diffs: ContentBlockManager::ContentBlock::FieldDiff.all_for_edition(edition: self))
       end
     end
   end
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/field_diff.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/field_diff.rb
@@ -1,0 +1,41 @@
+module ContentBlockManager
+  module ContentBlock
+    class FieldDiff < Data.define(:field_name, :new_value, :previous_value)
+      def self.all_for_edition(edition:)
+        if edition.document.is_new_block?
+          nil
+        else
+          field_diffs = []
+          previous_edition = edition.document.editions.includes(:edition_organisation, :organisation)[-2]
+          if previous_edition.title != edition.title
+            field_diffs.append(new(field_name: "title", previous_value: previous_edition.title, new_value: edition.title))
+          end
+          if previous_edition.details != edition.details
+            previous_edition.details.each do |key, value|
+              next unless value != edition.details[key]
+
+              field_diffs.append(new(field_name: key, previous_value: value, new_value: edition.details[key]))
+            end
+          end
+          previous_org = previous_edition.lead_organisation
+          new_org = edition.lead_organisation
+          if new_org != previous_org
+            field_diffs.append(new(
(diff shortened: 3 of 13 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
```diff
@@ -28,7 +28,43 @@ def record_update
       unless draft?
         user = Current.user
         state = try(:state)
-        versions.create!(event: "updated", user:, state:)
+        versions.create!(event: "updated", user:, state:, field_diff:)
+      end
+    end
+
+    FIELD_DIFF = Data.define(:previous_value, :new_value)
```
pezholio, line 35: I wonder if you might be better having `FIELD_DIFF` as a separate class with a `.all_for_document` class method. That way you can test it in isolation a bit better, as well as breaking down the methods in `field_diff` a bit more. Something like:

```
class FieldDiff < Data.define(:previous_value, :new_value)
  def self.all_for_document
     # logic from `field_diff` goes here, maybe broken down into smaller methods
  end
end
```

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
```diff
@@ -34,4 +35,16 @@ def time_html(date_time)
       lang: "en",
     )
   end
+
+  def table_rows(version)
```
pezholio, line 39: This is loads better! 👍

lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/has_audit_trail.rb
pezholio: I'm also wondering if you could have a way of translating the JSON into a `FieldDiff` object, that way if there are any changes to the structure, you could handle that programatically, rather than having to deal with JSON objects that may have changed over time

Harriethw: good shout about breaking out the logic, will take a look.

> I'm also wondering if you could have a way of translating the JSON into a FieldDiff object, 
I was wondering if there was way to use jsonschema in some way - basically it should always follow this shape
```
{
  {field_key}: {
    "previous_value": string,
    "new_value": string,
  }
}
```
but with the `field_key` being unknown i don't how to say that in rails land 🤔

pezholio: Maybe having an array of objects like this might be more predictable:

```
[
  { "key": string, "previous_value": string, "new_value": string }
]
```

Harriethw: I tried the array part in the last approach but it's a nightmare to then try and work out the order and display it correctly on the front end...

Harriethw: though maybe now we are doing the ordering in the audit_trail the frontend is less a problem, as long as the order was preserved when we iterate through to make the rows...

Harriethw: ok I've had a go at your suggestions and think it's looking better 👍

lib/engines/content_block_manager/app/models/content_block_manager/content_block/field_diff.rb
```diff
@@ -0,0 +1,41 @@
...
+  module ContentBlock
+    class FieldDiff < Data.define(:field_name, :new_value, :previous_value)
+      def self.all_for_edition(edition:)
+        if edition.document.is_new_block?
+          nil
+        else
+          field_diffs = []
+          previous_edition = edition.document.editions.includes(:edition_organisation, :organisation)[-2]
+          if previous_edition.title != edition.title
+            field_diffs.append(ContentBlockManager::ContentBlock::FieldDiff.new(field_name: "title", previous_value: previous_edition.title, new_value: edition.title))
```
pezholio, line 11: You could probably just do `new(...)` here

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
```diff
@@ -34,4 +35,14 @@ def time_html(date_time)
...
+    if version.field_diffs.present?
+      rows = []
+      version.field_diffs.each do |field|
+        rows.append([{ text: field["field_name"].humanize }, { text: field["previous_value"] }, { text: field["new_value"] }])
+      end
+      rows
```
pezholio, line 45: You could probably just do `map` here:

```suggestion
      version.field_diffs.map do |field|
        [
          { text: field["field_name"].humanize }, 
          { text: field["previous_value"] }, 
          { text: field["new_value"] },
        ]
      end
```

(Added some linebreaks for readability too)

pezholio (approved):
This looks ace. Thanks for struggling through it all! A couple of minor comments, but I think it looks good overall 👍

Harriethw:
@pezholio thank you!! have made those suggestions. I don't think I'll be around to watch it go green so feel free to merge on Monday when I'm not in

PR #9845, opened 2025-01-24
Remove unused locales

ChrisBAshton:
As part of investigating requirements for adding a new language to Whitehall, we came across these locale files and concluded that they're - mostly - not required.

1. Whitehall does not have an option for changing the locale of its UI.
2. Even if it did, the majority of the UI uses hardcoded English as opposed to i18n.
3. Even if it used i18n, a lot of the translation keys in non English locales have been left blank.

We suspect that most of these are a throwback to the days of Whitehall Frontend, or an optimistic view that Whitehall should fully support i18n and that we have to start somewhere. But until we take a proper look at i18n as a whole, this halfway house solution causes confusion for no benefit.

The PR results in a complete removal of all locale files except for `cy.yml`, where we are reliant on Welsh translations for corporate information page titles (and don't want to open up the ability for editors to provide their own translations of page titles in the UI, as - we assume - corporate information pages should have a standard set of page titles).

It also adds a new `config/languages.yml` file for defining the languages Whitehall supports, as well as info about each language's 'native' name and text direction. We will need to [update the Developer Docs](https://docs.publishing.service.gov.uk/manual/add-translation-whitehall.html) accordingly.

Note that there are other properties that are referred to by `i18n.t` - e.g. in presenters - which implies that those locales and properties are required. But having looked into each of them, they do tend to only ever be surfaced in English anyway, or lack translations in all but a handful of languages, so are sporadically implemented, and should be reasonably safe to do away with.

I did consider removing the calls to `i18n.t` altogether and just hardcoding the text in the presenter or view, but that has an impact on testability, as it is then not possible to just mock an i18n value and it's harder to make edits.

Trello: https://trello.com/c/T2neQSyf

## Testing

Tested on integration:

- Creating and publishing a Welsh-only Consultation
- Creating and publishing a publication with an Arabic translation
- Adding a translation to a Worldwide Organisation (and also a translation to a Worldwide Organisation's Contact)
- Creating a "Welsh language scheme" corporate information page, and a Welsh translation for that page
- Editing and publishing a translation of an existing worldwide news story

---


ryanb-gds (approved):
Great stuff. I thought more of these translations manifested on the frontend, amazing that it's just the corporate info page labels. Wonder if we can move them to Government Frontend 🤔

ChrisBAshton:
@ryanb-gds - good shout, I've raised https://trello.com/c/ZR8Vajsc/762-move-remaining-cy-locale-files-out-of-whitehall

PR #9851, opened 2025-01-28
(828) Show changenotes in review and show screens

pezholio:

This adds Change note related content to the review and show screens for a content block:

## Screenshots

image

image

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/summary_list_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/summary_list_component.rb
@@ -15,6 +15,7 @@ private
       organisation_item,
       instructions_item,
       status_item,
+      internal_change_note_item,
       embed_code_item,
     ].compact
   end
@@ -84,6 +85,13 @@ private
         value: last_updated_value,
       }
     end
+  end
+
+  def internal_change_note_item
+    {
+      field: "Internal change note",
+      value: content_block_document.latest_edition.internal_change_note.presence || "None",
+    }
   end
 
   def last_updated_value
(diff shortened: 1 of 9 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/summary_list_component.rb
```diff
@@ -90,16 +91,34 @@ def status_item
   def internal_change_note_item
     {
       field: "Internal change note",
-      value: content_block_document.latest_edition.internal_change_note.presence || "None",
+      value: content_block_edition.internal_change_note.presence || "None",
+    }
+  end
+
+  def change_note_items
+    content_block_edition.major_change ? [major_change_item, external_change_note_item] : [major_change_item]
```
Harriethw, line 99: don't think it really makes a difference at the moment, but an internal change note can't be added _unless_ it's a major change right?  If there is value under the internal change not field but it's _not_ a major change, shouldn't we still show the note anyway? just in case those user-facing rules changed?

pezholio: Internal notes _can_ be added regardless of if it's a major or minor. It's only public change notes that depend on major/minor

Harriethw: >Internal notes can be added regardless of if it's a major or minor. 

sorry I meant external note, not internal - it's only possible to add one if it's a major change at the moment

Harriethw: >If there is value under the internal change not field but it's not a major change, shouldn't we still show the note anyway? just in case those user-facing rules changed?

was trying to say that it doesn't seem quite right to check the `major_change` field instead of just checking if there's an external note, but don't think it needs to hold up the PR!

Harriethw (approved):

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/summary_list_component.rb
pezholio: Ah, gotcha!

PR #9854, opened 2025-01-28
Remove mechanize gem from development dependencies

ryanb-gds:
I'm pretty sure this is no longer being used anywhere


ChrisBAshton (approved):
⭐

PR #9864, opened 2025-01-29
content modelling/833 timeline updates

Harriethw:
https://trello.com/c/JOyjhbTb/833-update-copy-in-content-block-history

figma design: https://www.figma.com/design/F2Uq6Ju6fsSk1CEMjXm2va/Content-Modelling?node-id=2406-17913&t=ujjqo8fVksJJFFQY-0

convo about copy which is still ongoing: https://gds.slack.com/archives/C062Z1FA2ES/p1738168595581529?thread_ts=1738146596.061529&cid=C062Z1FA2ES 

Some slightly finnicky changes needed for the timeline. It sounds like the content is still a bit up in the air, but this should match the design as it stands.

Screenshot 2025-01-29 at 16-11-38 666 - GOV UK Content Block Manager

---

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
@@ -22,7 +22,14 @@ private
   end
 
   def title(version)
-    "#{version.item.block_type.humanize} #{version.state}"
+    case version.state
+    when "published"
+      version.state.capitalize
+    when "scheduled"
+      "Scheduled for publishing on #{version.item.scheduled_publication.to_fs(:long_ordinal_with_at)}"
+    else
+      "#{version.item.block_type.humanize} #{version.state}"
+    end
   end
 
   def first_created_edition
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
@@ -31,7 +31,7 @@ private
 
   def time_html(date_time)
     tag.time(
-      I18n.l(date_time, format: :long_ordinal),
+      date_time.to_fs(:long_ordinal_with_at),
       class: "date",
       datetime: date_time.iso8601,
       lang: "en",
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
@@ -9,7 +9,7 @@ private
   attr_reader :content_block_versions
 
   def items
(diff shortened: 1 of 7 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
```diff
@@ -28,16 +28,20 @@ def show_to_user?(version)
...
       "Scheduled for publishing on #{version.item.scheduled_publication.to_fs(:long_ordinal_with_at)}"
     else
       "#{version.item.block_type.humanize} #{version.state}"
     end
   end
 
-  def first_created_edition
-    content_block_versions.last
+  def first_published_version
+    @first_published_version ||= content_block_versions[-2]
```
pezholio, line 44: This seems a bit magic, could we do something like:

```
@first_published_version ||= content_block_versions.order(&:created_at).find { |v| v.event == "published" }
```

pezholio (commented):
Just one minor comment - I may be barking up the wrong tree though

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline_component.rb
Harriethw: that's a good call actually, it's more specific instead of relying on the order of versions always being the same

Harriethw: hmm ok I'm having trouble with the rails way of sorting... it's an array so I need to `sort_by` and I want to do by descending because I want the second to last oldest, but it doesn't like this? 
`@first_published_version ||= content_block_versions.sort_by{|v| -v[:created_at]}.drop(1).find { |v| v.state == "published" }`

pezholio: Try this maybe?

```
@first_published_version ||= content_block_versions.sort_by{|v| v[:created_at]}.reverse.drop(1).find { |v| v.state == "published" }
```

pezholio: Do you need the `drop(1)` as you're filtering for `published` events?

pezholio: On reflection, it might be more performant to filter first:

```
@first_published_version ||= content_block_versions.filter { |v| v.state == "published" }.sort_by{ |v| v[:created_at]}.reverse.first
```

Harriethw: > Do you need the drop(1) as you're filtering for published events?
oh yeahh, d'oh!

pezholio: Awesome! 🙌

pezholio (approved):

PR #9870, opened 2025-01-30
Make unpublishing operation asynchronous

ChrisBAshton:
The ‘unpublish’ action in Whitehall currently creates a worker to call Publishing API asynchronously. It should be updated to call Publishing API synchronously.

We should then listen for the response from Publishing API, and should catch any 4xx error and avoid updating the state of the document in Whitehall, and display an error message to the user.

Trello: https://trello.com/c/xlvNUI3c/3382-make-whitehalls-unpublishing-call-to-publishing-api-synchronous

---

### DIFF
--- a/app/services/service_listeners/publishing_api_pusher.rb
+++ b/app/services/service_listeners/publishing_api_pusher.rb
@@ -18,7 +18,7 @@ module ServiceListeners
         api.patch_links(edition)
         api.save_draft_translation(edition, options.fetch(:locale))
       when "unpublish"
-        api.unpublish_async(edition.unpublishing)
+        api.unpublish_sync(edition.unpublishing)
       when "withdraw"
         edition.translations.each do |translation|
           api.publish_withdrawal_async(
--- a/app/services/service_listeners/publishing_api_pusher.rb
+++ b/app/services/service_listeners/publishing_api_pusher.rb
@@ -21,7 +21,7 @@ module ServiceListeners
         api.unpublish_sync(edition.unpublishing)
       when "withdraw"
         edition.translations.each do |translation|
-          api.publish_withdrawal_async(
+          api.publish_withdrawal_sync(
             edition.content_id,
             edition.unpublishing.explanation,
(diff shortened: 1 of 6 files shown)


test/integration/attachment_redirect_due_to_unpublishing_integration_test.rb
```diff
@@ -188,6 +189,20 @@ def assert_sets_redirect_url_in_asset_manager_to(redirect_url)
     end
 
     def assert_redirected_in_publishing_api(content_id, redirect_path)
```
ryanb-gds, line 191: This method is wrongly named I think, should be `assert_gone_in_publishing_api`?

app/services/service_listeners/publishing_api_pusher.rb
```diff
@@ -18,10 +18,10 @@ def push(event:, options: {})
         api.patch_links(edition)
         api.save_draft_translation(edition, options.fetch(:locale))
       when "unpublish"
-        api.unpublish_async(edition.unpublishing)
+        api.unpublish_sync(edition.unpublishing)
       when "withdraw"
         edition.translations.each do |translation|
-          api.publish_withdrawal_async(
+          api.publish_withdrawal_sync(
```
ryanb-gds, line 24: I wonder how many editions there are with lots of translations... could get sloooow. Might be worth checking integration to see if this is acceptable (i.e. if only a few editions have say > 2 translations we can probably live with it).

ryanb-gds (changes requested):
Test name needs correcting I think and I'd like to know a bit about the performance implications

app/services/service_listeners/publishing_api_pusher.rb
ryanb-gds: That seems a reasonable trade off. It's probably a smaller number than that actually, because many of those editions are probably superseded.

ryanb-gds (approved):
LGTM

PR #9883, opened 2025-02-03
(872) Surface in progress drafts

pezholio:
Trello card: https://trello.com/c/GylqgRaE/872-surface-in-progress-drafts

This removes the job that deletes old drafts, as well as adds a link to edit an in progress draft if one exists. Additionally, I've added a screen to present to the user if they want to cancel half way through editing a block, as well as made some changes to how existing draft blocks are edited so we don't create a new draft each time.

## Screenshots

image

image


Harriethw (approved):
looks good technically! I think the UX will be a little confusing but like you say, let's see how it gets broken :)

PR #9886, opened 2025-02-04
Output failing tests when a test suite fails

pezholio:
Sometimes there’s issues with a particular order of tests (one specific one we're seeing is with Sidekiq tests - see https://github.com/alphagov/whitehall/actions/runs/13069262982/job/36467157548?pr=9858). If this happens, we output the tests that were run in that shard. This should hopefully help narrow down what tests are leaking state, so we can stop it from happening in future.


ChrisBAshton:
Going to close this as I assume it's superseded by #10015.

PR #9892, opened 2025-02-05
Support human readable slugs for non-English locales

ChrisBAshton:
This PR:

- No longer sets `slug` to `nil` when the locale is non-English. Every locale is considered "sluggable".
- This also means we no longer need the `sluggable_locale?`, `slug_eligible?` methods and ASCII checks. (The latter has caused a number of Zendesk tickets, whereby even English HTML Attachments have been given a content-ID based URL, because they contain a non-ASCII character like a fancy single quote).
- `sluggable_string` doesn't appear to have been in use, so it was deleted. (`sluggable_string` is used in the Document class: `friendly_id :sluggable_string`. But the Attachment class uses `friendly_id :title`, so overriding the `sluggable_string` method had no effect).
- Adds a test for friendly_id's native "uniqueness" feature (which we've opted into via `use: :scoped, scope: :attachable` on the Attachment class), so in theory we should be protected from clashes in the attachment slug namespace. Therefore the issue alluded to in #5321 (which reverts the last time this was attempted) should not happen again.
- Adds a basic safeguard that the slug-generation results in a slug that contains at least 4 a-z or A-Z characters. This is to protect against generating a slug like `-`, which could happen from sluggifying a title like `{chinese}-{chinese}`.
- That said, if we somehow DO get into a state whereby an attachment has what we'd call an invalid slug (<4 azAZ chars), we don't want to fix it if the attachment is already live. So we check for `safely_resluggable?`, introduced in #5942.

Result:

- Most HTML attachments should now have human-readable slugs. Only languages that are non-ASCII and cannot be converted to ASCII, such as Chinese, will continue to fall back to the content-ID based URLs.

Next steps:

- Now that all Whitehall tables use the UTF8MB4 charset (#9767),   we could potentially use CJK character sets natively in   slugs - but it may not always show nicely in the browser. Compare `https://example.com/你好` with `https://example.com/%E4%BD%A0%E5%A5%BD`.
- The option we'd likely want to go with is conversion to their Pinyin representations, e.g. `https://example.com/ni-hao`, probably using the `pinyin` or `ruby-pinyin` gem.

Trello: https://trello.com/c/YGsEG9YV/3418-improve-html-attachment-slug-creation

### DIFF
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -7,7 +7,7 @@ class HtmlAttachment < Attachment
           autosave: true,
           inverse_of: :html_attachment
 
-  before_validation :clear_slug_if_non_english_locale
+  before_save :ensure_slug_is_valid
 
   validates :govspeak_content, presence: true
 
@@ -61,8 +61,6 @@ class HtmlAttachment < Attachment
   end
 
   def should_generate_new_friendly_id?
-    return false unless sluggable_locale?
-
     slug.nil? || attachable.nil? || safely_resluggable?
   end
 
@@ -121,17 +119,12 @@ private
     root + public_path(options)
   end
 
-  def sluggable_locale?
-    locale.blank? || (locale == "en")
-  end
+  # Sense-check that the title converted to ASCII contains more than just hyphens!
+  def ensure_slug_is_valid
+    return unless slug.present? && slug.count("A-Za-z") < 4
 
-  def sluggable_string
-    sluggable_locale? ? title : nil
-  end
-
-  def clear_slug_if_non_english_locale
-    if locale_changed? && !sluggable_locale?
-      self.slug = nil
-    end
+    # Force slug to be `nil` - but only if attachment isn't live already.
+    # A `nil` slug will fall back to using the content_id as the slug.
+    self.slug = nil if safely_resluggable?
   end
 end
--- a/app/models/html_attachment.rb
+++ b/app/models/html_attachment.rb
@@ -61,7 +61,7 @@ class HtmlAttachment < Attachment
   end
 
   def should_generate_new_friendly_id?
-    slug.nil? || attachable.nil? || safely_resluggable?
+    safely_resluggable?
   end
 
   def deep_clone
--- a/test/unit/app/models/html_attachment_test.rb
+++ b/test/unit/app/models/html_attachment_test.rb
@@ -46,18 +46,6 @@ class HtmlAttachmentTest < ActiveSupport::TestCase
     assert_equal expected, actual
   end
 
-  test "#url returns absolute path to the draft stack when previewing for non-english locale" do
-    edition = create(:draft_publication, :with_html_attachment)
-    attachment = edition.attachments.first
-    attachment.update!(locale: "fi")
-
-    expected = "https://draft-origin.test.gov.uk/government/publications/"
-    expected += "#{edition.slug}/#{attachment.content_id}?preview=#{attachment.id}"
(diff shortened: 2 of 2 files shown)


test/unit/app/models/html_attachment_test.rb
```diff
@@ -257,11 +226,66 @@ class HtmlAttachmentTest < ActiveSupport::TestCase
...
+    #  now check slug not overridden
+    draft = edition.create_draft(create(:writer))
+    attachment = draft.attachments.first
+    attachment.title = "a-new-title"
+    attachment.save!
+    attachment.reload
+    assert_equal "foo", attachment.slug
+  end
+
+  # This behaviour is built into the friendly_id gem
```
ChrisBAshton, line 275: Side note: I did spot [in the docs](https://norman.github.io/friendly_id/file.Guide.html):

> Previous versions of FriendlyId appended a numeric sequence to make slugs unique, but this was removed to simplify using FriendlyId in concurrent code.

So I'd expect this test to need an update when we update finder_id. I did try updating finder_id locally but got `Bundler attempted to update friendly_id but its version stayed the same` and it wasn't immediately clear why.

app/models/html_attachment.rb
```diff
@@ -61,9 +62,7 @@ def url(options = {})
   end
 
   def should_generate_new_friendly_id?
-    return false unless sluggable_locale?
-
-    slug.nil? || attachable.nil? || safely_resluggable?
+    slug.nil? || slug.blank? || attachable.nil? || safely_resluggable?
```
ryanb-gds, line 65: I wonder if actually the only one of these we need to check now is `safely_resluggable?`, which as far as I can tell is set to true for newly created attachments and false if they've been copied over from the previous edition. It's quite poorly named, should probably be called `needs_slug?` if I understand the usage correctly, but that would be outside the scope of this PR.

app/models/html_attachment.rb
```diff
@@ -7,7 +7,8 @@ class HtmlAttachment < Attachment
           autosave: true,
           inverse_of: :html_attachment
 
-  before_validation :clear_slug_if_non_english_locale
+  before_create :ensure_slug_is_valid
```
ryanb-gds, line 10: Think you can remove `before_create`, as `before_save` should cover both create and update?

app/models/html_attachment.rb
```diff
@@ -121,17 +120,11 @@ def public_url(options = {})
...
-  def sluggable_string
-    sluggable_locale? ? title : nil
-  end
-
-  def clear_slug_if_non_english_locale
-    if locale_changed? && !sluggable_locale?
-      self.slug = nil
-    end
+    # Force FriendlyId to regenerate the slug, but only if attachment isn't live already
+    self.slug = nil if safely_resluggable?
```
ryanb-gds, line 128: Why would regenerating the slug ensure that it has more than four A-Za-z characters? Would it not just generate the same slug again?

ryanb-gds (commented):
I have questions. Nice work though!

test/unit/app/models/html_attachment_test.rb
ryanb-gds: Interesting 🤔

app/models/html_attachment.rb
ChrisBAshton: "Regenerate the slug" is a misleading comment (apologies). It sets `slug` to `nil`, and in our application code we fall back to using `content_id`.

ryanb-gds (approved):
Good stuff, this makes a lot more sense now

PR #9900, opened 2025-02-05
Ensure assets are deleted when policy group is deleted

lauraghiorghisor-tw:
Since Policy Groups are not editionable, their delete flows are different. Because of this, when deleting a group, the assets associated with their attachments did not get deleted in Asset Manager.

These changes address that bug, and ensure we delete both the attachments themselves and their corresponding assets in Asset Manager. Rather than trying to reuse some of the classes that serve the edition workflow, we preferred to use a callback on the `PolicyGroup` model, only reusing the asset specific `DeleteAttachmentAssetJob`.

[Trello](https://trello.com/c/CWdy6E90/3384-ensure-policy-group-attachments-are-removed-from-asset-manager-when-the-policy-group-is-deleted)

### DIFF
--- a/app/models/policy_group.rb
+++ b/app/models/policy_group.rb
@@ -14,7 +14,7 @@ class PolicyGroup < ApplicationRecord
 
   after_create :extract_dependencies
   after_update :extract_dependencies
-  after_destroy :remove_all_dependencies
+  after_destroy :remove_all_dependencies, :delete_all_attachments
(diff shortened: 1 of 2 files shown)


ryanb-gds (commented):
Surprisingly straightforward 😅

ryanb-gds (approved):
Meant to approve 🤦

PR #9901, opened 2025-02-05
WIP add search index method to landing page

richardTowers:
*Draft because I haven't worked out how to test this, or whether it's best to just nuke the Attachable include*

This is needed because the Attachable module assumes that the class its attached to has a search_index method, which landing page doesn't. This usually comes from the Searchable module, but we don't need any of that behaviour.

Actually we don't need Attachable either, I don't think - I think it's a hangover from when we were attaching CSVs to use for line graphs. But we're not doing line graphs now. We could remove it, but I'm slightly nervous about breaking something.

### DIFF
--- a/app/models/landing_page.rb
+++ b/app/models/landing_page.rb
@@ -41,6 +41,10 @@ class LandingPage < Edition
     end
   end
 
+  def search_index
+    {}
+  end
+
 private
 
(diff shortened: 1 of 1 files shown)


ryanb-gds:
Closing this as editionable pages are all indexed via Search API now (see #10048)

PR #9903, opened 2025-02-05
WIP on attachment state replication reporting task

ryanb-gds:
This needs thinking through carefully. Happily we can probably write some pretty useful tests for it, will try that soon.

Trello: https://trello.com/c/7KgPYYG1


ryanb-gds:
Closing this as we have a data pipeline and analysis being performed via Google BigQuery now

PR #9906, opened 2025-02-06
(867) Allow embedded objects to be created

pezholio:
Trello card: https://trello.com/c/tPWQl692/867-allow-embedded-objects-to-be-created-in-content-block-manager

This adds a new endpoint to create an embedded object if one exists. It's not (yet) linked from anywhere, and we don't have a confirmation screen, but this puts all the pieces in place for us to iterate on.

## Screenshots

image

image

### DIFF
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
@@ -30,6 +30,13 @@ module ContentBlockManager
           details:,
         ).render
       end
+
+      def add_object_to_details(object_type, body)
+        key = body["name"]&.parameterize.presence || SecureRandom.alphanumeric.downcase
+
+        details[object_type] ||= {}
+        details[object_type][key] = body.to_h
+      end
     end
   end
 end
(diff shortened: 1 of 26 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
```diff
@@ -30,6 +30,13 @@ def render
           details:,
         ).render
       end
+
+      def add_object_to_details(object_type, body)
```
Harriethw, line 43: will we be able to do this to editions that are already technically `published`? the workflow state transitions aren't totally clear to me since we added the draft option.

Harriethw (approved):
a big piece of work! looks good and great to see it coming together - some of the logic around the form is still a bit over my head but looks good to me now, think I'll have to just get using it to have it sink in!

lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
pezholio: Yeah, this will work because we're updating the state directly instead of using the transformers

PR #9907, opened 2025-02-06
Replace title component with heading on two views

AshGDS:
## What / Why
- Replaces the components gem `title` component with the `heading` component
- Trello card: https://trello.com/c/cYepys8K/476-replace-title-with-heading-in-whitehall
- Two of the changes are replacing component render statements. 
  - One of the templates required moving the `< Back` link out of the `<main>` element so that the spacing looked correct - it shouldn't be within the `<main>` otherwise that link sits too far down in the page.
- One of the commits is replacing a `gem-c-title__text` class with `gem-c-heading__text` in an app component. However this line of HTML isn't actually used anywhere in the repo - the rails boolean which renders this HTML ( `is_page_heading`) - is never set to `true` in any render statements. I wasn't sure if I should worry about removing functionality from the repo, especially as we don't own this app, so I've left it in, and just replaced the class. I think this change should be OK visually, because if you go to the [title component in our component guide 
](https://components.publishing.service.gov.uk/component-guide/title) and replace `gem-c-title__text` with `gem-c-heading__text` via `Inspect Element`, nothing visually changes, so theoretically nothing visual would change on this element either. CSS wise `.gem-c-title__text` and `.gem-c-heading__text` have minor differences, but I don't think it's enough to result in a major visual change.
- I've checked the changes on Integration.

### "Add translation" heading
- The "Add translation" heading exists when adding a new translation to a document. e.g. 
https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/publications/1647318 - you might need to start a new draft first to see the add translation link

image
image

### "New translation" heading
- Create a new translation of the GDS organisation first if one doesn't exist: https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/organisations/government-digital-service/translations
- Then on the Contacts tab, the contacts will have an "Add translation" link https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/organisations/government-digital-service/contacts
- If you click on "Add translation", you get to the template with the change in this PR

image
image

## Visual Changes

Everything moves up slightly, as the `title` component had some top padding, which will no longer exist.

| Before | After |
| --- | --- |
|image|image|
image|image|

### DIFF
--- a/app/views/components/_select_with_search.html.erb
+++ b/app/views/components/_select_with_search.html.erb
@@ -13,7 +13,7 @@
 
 <%= content_tag :div, class: select_helper.css_classes, data: select_helper.data_attributes do %>
   <% if is_page_heading %>
-    <%= tag.h1 label_tag(id, label, class: select_helper.label_classes, id: "#{id}_label"), class: "gem-c-title__text" %>
+    <%= tag.h1 label_tag(id, label, class: select_helper.label_classes, id: "#{id}_label"), class: "govuk-heading-xl" %>
   <% else %>
     <%= label_tag(id, label, class: select_helper.label_classes, id: "#{id}_label") %>
   <% end %>
--- a/app/views/admin/contact_translations/index.html.erb
+++ b/app/views/admin/contact_translations/index.html.erb
@@ -2,9 +2,12 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/title", {
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "New translation",
       context: @contact.title,
-      title: "New translation",
+      font_size: "xl",
+      heading_level: 1,
+      margin_bottom: 8,
     } %>
 
     <%= form_with url:admin_organisation_contact_translations_path(@contactable,@contact) do |form| %>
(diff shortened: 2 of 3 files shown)


app/views/components/_select_with_search.html.erb
```diff
@@ -13,7 +13,7 @@
 
 <%= content_tag :div, class: select_helper.css_classes, data: select_helper.data_attributes do %>
   <% if is_page_heading %>
-    <%= tag.h1 label_tag(id, label, class: select_helper.label_classes, id: "#{id}_label"), class: "gem-c-title__text" %>
+    <%= tag.h1 label_tag(id, label, class: select_helper.label_classes, id: "#{id}_label"), class: "gem-c-heading__text" %>
```
andysellick, line 16: It'd be nice if we didn't have a gem class outside of a component like this but I see that this makes the least amount of change. Did you consider using one of the `govuk-heading-` classes instead? It wouldn't look the same, but would reduce the assumed dependency.

andysellick (commented):
Changes look good, just one comment. Good thorough PR description 👍

app/views/components/_select_with_search.html.erb
andysellick: Looks like you've changed it to `govuk-heading-xl` - would `govuk-heading-m` not be closer in size? Sorry, can't see this to check.

andysellick: Ah of course, sorry got confused, ignore me.

andysellick (approved):

PR #9908, opened 2025-02-06
Make /government/get-involved an exact route

KludgeKML:
- The prefix route seems to be a hangover from when whitehall was rendering the Take Part items. These all now have their own content item, so anything under get-involved is either the get involved item or has its own route.
- Both of the SpecialRoute instances are now exact path, so make that the default in rake task and test, and remove it from the instances.

### DIFF
--- a/app/models/special_route.rb
+++ b/app/models/special_route.rb
@@ -7,7 +7,6 @@ class SpecialRoute
         title: "Government feed",
         description: "This route serves the feed of published content",
         rendering_app: Whitehall::RenderingApp::COLLECTIONS_FRONTEND,
-        type: "exact",
       },
       {
         base_path: "/government/get-involved",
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Code looks good to me, so assuming you've validated that there aren't risks in this changing this looks good to me.

Good luck with fixing the test though that looked a little confusing

KludgeKML:
The only routes under /government/get-involved are the take-part pages (all linked to from the main get-involved page), which all have their own content items and routes (https://github.com/alphagov/frontend/blob/main/config/routes.rb#L82-L85), so I'm fairly confident that this won't have any knock-on effects.

KludgeKML:
I think that test might be a known flaky one (it passed on re-run), but I might double-check with the Whitehall team to confim.

KludgeKML:
Have confirmed all relevant paths still work on integration after republishing this route.

PR #9909, opened 2025-02-06
Fix date input name bug

ryanb-gds:
Rails nested form inputs should be in the style "parent[level_one_name][level_two_name]", but forms also used to work in the style "parent[level_one_name[level_two_name]]". As of Rails 8 that is no longer the case, so we need to correct the form input names.

Trello: https://trello.com/c/IylHk1B6


lauraghiorghisor-tw (approved):
LGTM 🧐

PR #9914, opened 2025-02-07
Remove unused Search API Document Presenter

ryanb-gds:
This was used to present finder results when Whitehall Frontend rendered finders.


ChrisBAshton (approved):
🎉

PR #9924, opened 2025-02-11
Content modelling/885 add rates to view page

pezholio:


lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
```diff
@@ -31,6 +31,10 @@ def render
         ).render
       end
 
+      def first_class_details
+        details.select { |_k, v| v.is_a?(String) }
```
richardTowers, line 35: 🤔  Is it always going to be the case that entries with String values are "first class" / important enough to be shown on the summary card, and entries with other values are not?

Or is this method likely to develop special cases?

richardTowers (approved):
Non blocking comment.

lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
pezholio: For now, almost definitely. We can revisit as we develop new schemas

PR #9929, opened 2025-02-12
Refactor timeline component

pezholio:
This breaks up each timeline component into a seperate component, as well as making the field changes component into a single component, which makes them easier to test.

I originally did this as some prep for adding embedded object histories, but think it's best to leave this until we have the ability to edit embedded objects.

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.html.erb
@@ -1,0 +1,45 @@
+<div class="timeline__item">
+  <div class="timeline__header">
+    <% if is_latest %>
+      <span class="timeline__latest">Latest</span>
+      <br>
+    <% end %>
+    <h2 class="timeline__title"><%= title %></h2>
+    <p class="timeline__byline">by <%= byline %></p>
+  </div>
+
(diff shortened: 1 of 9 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    <%= date %>
+  </p>
+
+  <% if version.field_diffs.present? %>
+    <div class="timeline__diff-table">
+      <%= render "govuk_publishing_components/components/details", {
+        title: "Details of changes",
+        open: is_latest,
+      } do %>
+        <% capture do %>
```
richardTowers, line 21: Do we need this `capture` block? Looks like we're not assigning it to anything, and we're only calling one method inside the block so if we needed the string we could just call the method?

richardTowers (approved):
Another non-blocking comment.

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.html.erb
pezholio: Ah, just tried that, and it renders outside the details container unfortunately 😞

PR #9593, opened 2024-11-07
SPIKE: display `controller_name` in title

dnkrj:
⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Display the controller name in the page title so its eligibility as a form of categorisation for tracking can be evaluated.

https://trello.com/c/R2kwIDVF/3137-spike-decide-on-page-grouping-categorisation-for-analytics

PR #9934, opened 2025-02-13
Populate GA4 `document_type` using Rails variables

gtvj:
## What

Updates the existing analytics helper to use a combination of the controller name and the action name as an imperfect proxy for the concept of a `document_type` in Whitehall. The Trello card for this ticket is "[Track document_type variable on GTM page load](https://trello.com/c/LQjqDVcZ/3438-track-documenttype-variable-on-gtm-page-load?search_id=c5be5c9c-f3b9-405c-8f3c-88c60d704b9a)".

A range of options were explored in an earlier spike - see [this PR](https://github.com/alphagov/whitehall/pull/9593) and [this Trello card](https://trello.com/c/R2kwIDVF/3137-spike-decide-on-page-grouping-categorisation-for-analytics).

### `dataLayer` examples from this implementation

| Screenshot  | `document_type` passed to `dataLayer`  |
|---|---|
| image | `index-dashboard`  |
| image | `index-new_document`  |
| image | `new-consultations`  |
| image | `index-corporate_information_pages`  |
| image | `index-statistics_announcements` |

## Why 

Performance Analysts would like the GA4 `dataLayer` to be populated with a `document_type` every time a page loads, reflecting the implementation present in GOV.UK. Because Whitehall is a content management system that does not currently have the concept of a `document_type`, an earlier spike explored options. The implementation in this PR builds upon the exploration of using `controller_name` as the document type in #9593 to use an interpolated string of both `action_name` and `controller_name` because this will provide additional information to Performance Analysts.

### DIFF
--- a/app/helpers/admin/analytics_helper.rb
+++ b/app/helpers/admin/analytics_helper.rb
@@ -1,6 +1,6 @@
 module Admin
   module AnalyticsHelper
-    def track_analytics_data_on_load(title)
+    def track_analytics_data_on_load(controller_name, action_name)
       {
         event: "page_view",
         page_view: {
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Code looks good 👍

PR #9942, opened 2025-02-17
content modelling/899 copy embed code

Harriethw:
https://trello.com/c/Hr1mmk1C/899-add-copy-embed-code-to-attributes

Screenshot 2025-02-17 at 16 21 12

Adding the 'copy code' to specific attributes on a rate.
This involves first calculating what the embed code is for the nested field, then adding the module to the row in the summary card.

The copy code should only appear next to Amount for MVP, but that can be done in future PR.

---

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
@@ -20,6 +20,10 @@ private
       {
         key: key.titleize,
         value: object[key],
+        data: {
+          module: "copy-embed-code",
+          "embed-code": content_block_edition.document.embed_code_for_field("#{object_type}/#{object_name}/#{key}"),
+        },
       }
     end
   end
--- a/lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
@@ -1,15 +1,15 @@
 class ContentBlockManager::Shared::EmbeddedObjects::SummaryCardComponent < ViewComponent::Base
-  def initialize(content_block_edition:, object_type:, object_name:, show_edit_action: false, redirect_url: nil)
+  def initialize(content_block_edition:, object_type:, object_name:, is_editable: false, redirect_url: nil)
     @content_block_edition = content_block_edition
     @object_type = object_type
     @object_name = object_name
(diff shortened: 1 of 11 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/document.rb
```diff
@@ -27,7 +27,11 @@ class Document < ApplicationRecord
       scope :live, -> { where.not(latest_edition_id: nil) }
 
       def embed_code
-        "{{embed:content_block_#{block_type}:#{content_id}}}"
+        "#{embed_code_prefix}}}"
+      end
+
+      def embed_code_for_field(field_path)
```
pezholio, line 33: I think this is OK for now, but yeah, I think we can probably do something a bit smarter given the schema etc

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
```diff
@@ -16,10 +16,18 @@ def title
   end
 
   def rows
+    object_keys_rows
```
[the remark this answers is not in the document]
pezholio: Nah, you're right, just move the code from `object_keys_rows` to here I reckon. Cuts down on the misdirection.

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
```diff
@@ -16,10 +16,18 @@ def title
...
+  end
+
+  def object_keys_rows
     object.keys.map do |key|
       {
         key: key.titleize,
         value: object[key],
```
pezholio, line 26: I think we need to have the embed code outputted here too, then hidden when we initialize the `copy-embed-code` JS module.

pezholio (commented):
Looks good so far, just a couple of comments

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.rb
Harriethw: ah yeah I meant to say I'd pick the no-JS option up in another PR!

Harriethw: as the new design is a bit different and I think will mean changes to the actual javascript https://trello.com/c/Hr1mmk1C/899-add-copy-embed-code-to-attributes?filter=label:Dev

pezholio (approved):

PR #9945, opened 2025-02-18
(889) Handle embedded objects in history

pezholio:
Trello card: https://trello.com/c/e38axhHu/889-handle-embedded-objects-in-history

This allows us to show the history of embedded objects more neatly in the history (rather than showing hashes). This has involved a bit of tweaking to change how history has recorded (inspired by [Paper Trail](https://github.com/paper-trail-gem/paper_trail)), so I've also had to add a backfill task that will need to be run on each environment after deploy.

## Screenshot

image


lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/diffable.rb
```diff
@@ -0,0 +1,45 @@
...
+    def generate_details_diff(previous_details = previous_edition.details, current_details = details)
+      diff = {}
+      keys = [*previous_details&.keys, *current_details&.keys].uniq
+      keys.each do |key|
+        previous_value = previous_details&.fetch(key, nil)
+        new_value = current_details&.fetch(key, nil)
+        if previous_value.is_a?(String) || new_value.is_a?(String)
+          next unless previous_value != new_value
+
+          diff[key] = [previous_value, new_value]
```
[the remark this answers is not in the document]
pezholio: I know what you mean - I'll give it a go 👍

lib/engines/content_block_manager/lib/tasks/update_version_diffs.rake
```diff
@@ -0,0 +1,17 @@
+namespace :content_block_manager do
+  desc "Update Content Block Manager Version Diffs"
+  task update_version_diffs: :environment do
+    ContentBlockManager::ContentBlock::Version.where(event: "updated", item_type: "ContentBlockManager::ContentBlock::Edition").find_each do |version|
+      edition = version.item
```
Harriethw, line 5: running this locally I seem to have a version where `item` is `nil` and it fails 🤔

Harriethw: wonder how that happened...maybe an Edition got deleted?

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.html.erb
```diff
@@ -23,6 +23,13 @@
             version:,
             schema:,
             ) %>
+
+          <% embedded_object_diffs.each do |item| %>
```
Harriethw, line 27: I think there needs to be a null check here, as seems to fail if there are no `embedded_object_diffs`

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.rb
```diff
@@ -43,4 +43,12 @@ def internal_change_note
   def change_note
     version.item.change_note
   end
+
+  def embedded_object_diffs
+    schema.subschemas.map { |subschema|
+      version.field_diffs.dig("details", subschema.id).map do |object_id, field_diff|
```
Harriethw, line 49: seemed to fail locally for me if `field_diffs` was `nil`

lib/engines/content_block_manager/lib/tasks/update_version_diffs.rake
pezholio: That's weird, because Versions require an item to be present. Might be something janky in your environment

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.html.erb
pezholio: That's a bit odd too, because it should always return at least an empty array? 🤔

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.rb
pezholio: I've fixed this so `field_diffs` returns an empty hash by default

lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition/diffable.rb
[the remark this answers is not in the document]
pezholio: Have added this in https://github.com/alphagov/whitehall/pull/9945/commits/394973bc852b0b83471f0bfe66aba302fbe98fbf

Harriethw (approved):

PR #9951, opened 2025-02-19
Remove Rates from listing

pezholio:
Trello card: https://trello.com/b/L1j2FYNX/content-modelling-team-trello

Just noticed this, so I thought I'd fix as I go. Also, the SummaryCardComponent was inheriting from a component which is no longer used, so I’ve moved all the logic to the index component.

### Before

image

### After

image


Harriethw (approved):
🧹

PR #9953, opened 2025-02-19
(907) Allow fields with copy code to be configured

pezholio:
Trello card: https://trello.com/c/aSyUk0AR/907-allow-fields-with-copy-code-to-be-configured

This adds a new configuration file that allows us to configure which fields can have a copy code associated (defaults to all fields). This config file will also be useful for us to define fun things like custom fields and ordering of fields too.

## Screenshot

image


Harriethw (approved):
👍

PR #9957, opened 2025-02-20
(900) Update amount field to text input with prefix component

pezholio:
Trello card: https://trello.com/c/3bpVWCMn/900-update-amount-field-to-text-input-with-prefix-component

This adds some configuration to add a prefix to fields. It's not as simple as simply changing the component if we want a prefix, as we still want the value contained in the prefix to be part of the resulting value. To get around this, if we want a prefix we add some Javascript to add the prefix element to the input, removing the prefix from the value of the field. Once the form is submitted, we update the value to add the prefix back.

This gracefully degrades without Javascript to just adding the prefix value to the input field if there is no value present.

## Screenshot

image

### DIFF
--- a/lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/application.js
+++ b/lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/application.js
@@ -1,1 +1,2 @@
 //= require ./modules/copy-embed-code
+//= require ./modules/form-prefix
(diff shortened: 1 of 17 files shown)


Harriethw:
It's a shame we have to introduce a tacked on bit of javascript to make that component work - the split second where you can see the script add a `£` really bothers me! I wonder if it's worth pushing back on UCD a bit? it looks like that component is designed more for public-facing forms, and presumably for different backend use cases than we have hear (e.g. I guess most data wouldn't need to have the `£` explicitly in the string!).  Would having the first step of auto-adding the `£` be enough for now?

I also noticed it's possible to input `££££12.2` so the validation could be a little tightened up there. Could take the chance to allow whole numbers not just decimals?

pezholio:
I see what you're saying. The validation is all schema-side, so we'll need to do that there. Lemme speak to UCD and see what we can do

PR #9958, opened 2025-02-20
Unify APIs for the 'autocomplete' and 'select with search' components

ChrisBAshton:
Trello: https://trello.com/c/pmbL2yId/3411-remove-the-selectwithsearch-component-from-whitehall

This PR is a refactor, to make the data structures used by 'autocomplete' and 'select with search' more alike, to enable us to consolidate the components in a future PR. There are *no user facing changes* here: all the same existing components are used and there should be no change to look/feel/functionality.

Changes:

- Removes some unused features from both 'autocomplete' and 'select with search'
- Changes 'autocomplete' to define its options as an array of hashes, making it more expressive (allowing options to be defined as 'selected') and bringing it in line with 'select with search'
  - The 'autocomplete' component now uses the `SelectWithSearchHelper` class as a result (so that both components have one way of converting the array of hashes into the 'options HTML').
  - And as a result, we've also had to tweak the helper class to accept a `multiple` parameter.
- Changes 'autocomplete' to have a proper `include_blank: true` option instead of having to prefix an empty string
- Changes 'select with search' to define its error messages as an array of errors, making it more flexible and bringing it in line with 'autocomplete' and the vast majority of other components
- Patches some minor functionality that is currently unused, but will make it easier to consolidate the components in the next PR. E.g. overriding data attributes, defaulting a value for `name`.

This leaves the components' data structures with one main difference each:

- The 'autocomplete' component doesn't yet support the 'grouped_options' property for the 'select with search' component
- The 'select with search' component doesn't yet support the 'multiple' property for the 'autocomplete' component
  - NB, 'select with search' also doesn't support `autocomplete_configuration_options`, but given that's specific to the accessible-autocomplete-multiselect library, that's not something we'd want to carry over. It also isn't currently used - it's just a documented feature.

In the next PR, we'll either switch all 'select_with_search' calls to use 'autocomplete' instead, or switch all 'autocomplete' calls to use 'select_with_search'. Whichever component is left unused will be deleted as part of the same PR. It also makes it easier to swap out the component for a [more accessible alternative](https://trello.com/c/tFmr25TL/647-implement-accessible-replacement-for-accessible-autocomplete-multiselect) in future, as we'll only be dealing with one kind of component (and one set of 'parameters' passed to the component) instead of two.

## Background

There are currently 17 instances of "autocomplete" and 24 instances of "select_with_search".

* ‘autocomplete’ was [added in August 2022](https://github.com/alphagov/whitehall/pull/6727) (copied over [from Content Publisher](https://github.com/alphagov/content-publisher/blob/main/app/assets/javascripts/components/autocomplete.js))
* ‘select_with_search’ was [added in March 2023](https://github.com/alphagov/whitehall/pull/7406).

Neither the latter PR nor the [Trello card](https://trello.com/c/G8Uagrjg/1154-create-an-app-component-for-single-select-with-search) explain why ‘autocomplete’ wasn’t used instead (nor indeed makes any mention of ‘autocomplete’ at all). It may have been an attempt to side-step the known problems with accessible-autocomplete component, or it may be that search_with_select was used as a drop in jQuery-less replacement for how Whitehall _was_ using Select2. Some resources Kevin was able to dig up:

- https://github.com/ollietreend/govuk-choices
- [Autocomplete in Whitehall](https://docs.google.com/document/d/199HXtDjT2b8kMGJ-r4FjTius2mM_XPHKbyBR0vgPKZQ/edit)
- [Selects in Whitehall publisher](https://docs.google.com/spreadsheets/d/1cLyqhKfcSv3h6phJhpUQqV83byf0WhU57g0hPKJcgRM/edit)

In any case, both components look very similar and share a lot of functionality. Here is a key table of differences:

| autocomplete | select_with_search |
|---|---|
| Allows multiple | Does not allow multiple |
| Powered by GOV.UK's accessible-autocomplete-multiselect (a fork of accessible-autocomplete) | Powered by Choices.js |
| Defines options as an array `[text, value]` | Defines options as a hash `{ text:, value: }`
| Defines 'selected' option as a separate `selected: <value>` | Defines 'selected' option inline in the hash |
| Defines label as a hash `text:, bold: true` | Defines label as a string, alongside a `heading_size` option to change its size |
| Has no page heading option | Has a `is_page_heading: true` boolean |
| Has `error_items` array | Has `error_message`/`error_id` |
| Doesn't support [grouped options](https://apidock.com/rails/v6.0.0/ActionView/Helpers/FormOptionsHelper/grouped_options_for_select) | Supports grouped options via `grouped_options` property |
| Required `[""] +` prefix when specifying blank option | Has `include_blank: true` boolean |

There is little point in maintaining two highly similar but differently architected components: we should pick only one, and retire the other.

## Deployment note

This has been successfully tested on Integration. One thing to note: because the cached objects in TaggableContentHelper have changed, we need to clear the cache with `Rails.cache.clear` after deploy, otherwise users will see the generic "server issue" error. See https://govuk.sentry.io/issues/6317660943/activity/?environment=integration&project=202259&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=1

---

### DIFF
--- a/app/helpers/admin/taggable_content_helper.rb
+++ b/app/helpers/admin/taggable_content_helper.rb
@@ -1,15 +1,6 @@
 # A bunch of helpers for efficiently generating select options for taggable
 # content, e.g. topics, organisations, etc.
 module Admin::TaggableContentHelper
-  # Returns an Array that represents the current set of taggable topical
-  # events. Each element of the array consists of two values: the name and ID
-  # of the topical event.
-  def taggable_topical_events_container
-    Rails.cache.fetch(taggable_topical_events_cache_digest, expires_in: 1.day) do
-      TopicalEvent.order(:name).map { |te| [te.name, te.id] }
-    end
-  end
-
   # Returns an Array that represents the current set of taggable organisations.
   # Each element of the array consists of two values: the select_name and the
   # ID of the organisation
@@ -51,17 +42,6 @@ module Admin::TaggableContentHelper
   def taggable_role_appointments_container
     Rails.cache.fetch(taggable_role_appointments_cache_digest, expires_in: 1.day) do
       role_appointments_container_for(RoleAppointment)
-    end
-  end
-
-  # Returns an Array that represents the taggable ministerial roles. Each
-  # element of the array consists of two values: the name of the ministerial
-  # role with the organisation and current holder and its ID.
-  def taggable_ministerial_roles_container
-    Rails.cache.fetch(taggable_ministerial_roles_cache_digest, expires_in: 1.day) do
-      MinisterialRole.with_translations.with_translations_for(:organisations).alphabetical_by_person.map do |role|
-        ["#{role.name}, #{role.organisations.map(&:name).to_sentence} (#{role.current_person_name})", role.id]
-      end
     end
   end
 
@@ -112,17 +92,6 @@ module Admin::TaggableContentHelper
     end
   end
 
-  # Returns an Array representing the taggable document collections and their
-  # groups. Each element of the array consists of two values: the
-  # collection/group name and the ID of the group.
-  def taggable_document_collection_groups_container
-    Rails.cache.fetch(taggable_document_collection_groups_cache_digest, expires_in: 1.day) do
-      DocumentCollection.latest_edition.alphabetical.includes(:groups).flat_map do |collection|
-        collection.groups.map { |group| ["#{collection.title} (#{group.heading})", group.id] }
-      end
-    end
-  end
-
   # Returns an Array that represents the taggable worldwide organisations.
   # Each element of the array consists of two values: the name of the worldwide
   # organisation and its ID.
@@ -166,13 +135,6 @@ module Admin::TaggableContentHelper
     @taggable_role_appointments_cache_digest ||= calculate_digest(RoleAppointment.order(:id), "role-appointments")
   end
 
-  # Returns an MD5 digest representing the current set of taggable ministerial
-  # rile appointments. THis will change if any ministerial role is added or
-  # updated.
-  def taggable_ministerial_roles_cache_digest
-    @taggable_ministerial_roles_cache_digest ||= calculate_digest(MinisterialRole.order(:id), "ministerial-roles")
-  end
-
   # Returns an MD5 digest representing all the detailed guides. This wil change
   # if any detailed guides are added or updated.
   def taggable_detailed_guides_cache_digest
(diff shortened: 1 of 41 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "lint:js": "eslint --cache --cache-location .cache/eslint --color --ignore-path .gitignore -- \"**/*.js\"",
     "lint:scss": "stylelint app/assets/stylesheets/",
     "lint:prettier": "prettier --cache --cache-location .cache/prettier --cache-strategy content --check -- \"**/*.{js,scss}\"",
+    "lint:prettier:fix": "prettier --write \"**/*.{js,scss}\"",
```
ryanb-gds, line 12: Hah, I should have done this before now. Thanks

app/views/components/_autocomplete.html.erb
```diff
@@ -4,6 +4,7 @@
   error_items ||= []
   aria = error_id if error_items.any?
   select ||= {}
+  include_blank ||= false
```
ryanb-gds: Nice, the empty string concatenation has long bugged me

app/helpers/admin/taggable_content_helper.rb
```diff
@@ -40,64 +38,66 @@ def taggable_needs_container
...
-    Rails.cache.fetch(taggable_worldwide_organisations_cache_digest, expires_in: 1.day) do
-      WorldwideOrganisation.with_translations.latest_edition.map { |wo| [wo.title, wo.document.id] }
+    cached_taggable_worldwide_organisations.map do |wo|
+      [wo.title, wo.document.id]
+    end
+  end
+
+  def cached_taggable_organisations
+    Rails.cache.fetch(taggable_organisations_cache_digest, expires_in: 1.day) do
+      Organisation.with_translations.order("organisation_translations.name")
```
ryanb-gds, line 112: Some of these queries (this one being a good example) shouldn't really be very slow - if it is it might be a missing index on the translation name. I can very much imagine a MoG change scenario where we put an org live and then have to shell into prod to clear the cache because we can't tag any content to it.

I suppose my question is could we apply the caching a bit more selectively/check we've got the indexes we need? Probably a bit outside the scope of this PR, to be fair

ryanb-gds (approved):
Sterling work Chris! The inconsistency in publishing components with regards to labels and heading sizes is a bit of an irritation, hope that gets sorted soon

app/helpers/admin/taggable_content_helper.rb
ChrisBAshton: Yes, I suspected as much too. Digging into the commit history (read 0a2f1b732f499d8009279647007acd31f6210cb6), it looks like some of the cached queries have a significant impact on page load times, so I didn't want to open that can of worms. Would be ripe for a revisit in a future PR though.

PR #9959, opened 2025-02-20
Retire the 'autocomplete' component in favour of 'select with search'

ChrisBAshton:
*Blocked on #9977*

Trello: https://trello.com/c/pmbL2yId/3411-remove-the-selectwithsearch-component-from-whitehall

Following the unification of the component data structures in #9958, **this PR switches all of the "autocomplete" component calls to "select_with_search"**, allowing us to retire the "autocomplete" component. It also makes it easier to swap out the component for a [more accessible alternative](https://trello.com/c/tFmr25TL/647-implement-accessible-replacement-for-accessible-autocomplete-multiselect) in future, as we'll only be dealing with one kind of component (and one set of 'parameters' passed to the component) instead of two.

## Background

Prior to this PR, there were 17 instances of "autocomplete" and 24 instances of "select_with_search".

* ‘autocomplete’ was [added in August 2022](https://github.com/alphagov/whitehall/pull/6727) (copied over [from Content Publisher](https://github.com/alphagov/content-publisher/blob/main/app/assets/javascripts/components/autocomplete.js))
* ‘select_with_search’ was [added in March 2023](https://github.com/alphagov/whitehall/pull/7406).

Neither the latter PR nor the [Trello card](https://trello.com/c/G8Uagrjg/1154-create-an-app-component-for-single-select-with-search) explain why ‘autocomplete’ wasn’t used instead (nor indeed makes any mention of ‘autocomplete’ at all). It may have been an attempt to side-step the known problems with accessible-autocomplete component, or it may be that search_with_select was used as a drop in jQuery-less replacement for how Whitehall _was_ using Select2. Some resources Kevin was able to dig up:

- https://github.com/ollietreend/govuk-choices
- [Autocomplete in Whitehall](https://docs.google.com/document/d/199HXtDjT2b8kMGJ-r4FjTius2mM_XPHKbyBR0vgPKZQ/edit)
- [Selects in Whitehall publisher](https://docs.google.com/spreadsheets/d/1cLyqhKfcSv3h6phJhpUQqV83byf0WhU57g0hPKJcgRM/edit)

In any case, both components look very similar and share a lot of functionality. Here is a key table of differences:

| autocomplete | select_with_search |
|---|---|
| Allows multiple | Does not allow multiple |
| Powered by GOV.UK's accessible-autocomplete-multiselect (a fork of accessible-autocomplete) | Powered by Choices.js |
| Defines options as an array `[text, value]` | Defines options as a hash `{ text:, value: }`
| Defines 'selected' option as a separate `selected: <value>` | Defines 'selected' option inline in the hash |
| Defines label as a hash `text:, bold: true` | Defines label as a string, alongside a `heading_size` option to change its size |
| Has no page heading option | Has a `is_page_heading: true` boolean |
| Has `error_items` array | Has `error_message`/`error_id` |
| Doesn't support [grouped options](https://apidock.com/rails/v6.0.0/ActionView/Helpers/FormOptionsHelper/grouped_options_for_select) | Supports grouped options via `grouped_options` property |
| Required `[""] +` prefix when specifying blank option | Has `include_blank: true` boolean |

There is little point in maintaining two highly similar but differently architected components: we should pick only one, and retire the other.

## Picking the surviving component

Neither component is perfect. But committing to one component means we're more likely to be able to fix issues as they arise.

The 'select_with_search' component has a known a11y issue for VoiceOver (see #9797). However, implementing 'multiple select' support in this PR was very straightforward.

In contrast, there are a number of usability issues for the 'autocomplete' component:

- unable to 'unselect' an option from the 'single' dropdown
- indeed, no way of even opening the dropdown to just see the available options: 'autocomplete' defaults to selecting the first item in the list when you navigate away from the input
- would need to add support for `grouped_options` (currently unclear how much work that would be)
- has [known a11y issues](https://drive.google.com/file/d/1BgR9Tkpr9nEYoU3mrUwue306XjLAsZ7B/view) with its multi-select mode (see `DAC_Custom_Elements_01` in the report).
- ties us to 'accessible-autocomplete-multiselect', which is a repo we're hoping to retire

The user-facing experience of the 'autocomplete' component seems preferable given it has fewer a11y issues overall than the 'select_with_search' component, but the underlying data structures of the 'select_with_search' component seem more expressive and well thought through.

## miller-columns-element

miller-columns-element used accessible-autocomplete-multiselect in a different way, so has been handled specially. We've swapped it out to use Choices.js under the hood - along with the SelectWithSearch helper - though it doesn't call 'select_with_search' directly.

The UX is near identical:

| Old component | New component|
|------|----|
|1-landing-old|1-landing-new|
|On page load|On page load - we've lost the magnifying glass, but gained some hint text|
|1-searching-old|1-searching-new|
|Searching|Searching - we've lost some highlighting, but otherwise identical|
|1-selected-old|1-selected-new|
|Selected|Selected - the end result is identical|

---


GDSNewt (commented):
Just an FYI; the before/ after links in this commit aren't accessible

PR #9961, opened 2025-02-21
(926) Fix server error when searching for certain keywords

pezholio:
Trello card: https://trello.com/c/Yl8xQYoU/926-bug-server-error-when-searching-for-certain-keywords

Before we were getting server errors when searching for certain keywords in integration. This is because, as the database has grown larger, the method of using regular expressions to search for keywords has become slower, eventually causing the database to time out.

To get around this, we've added a fulltext index to the editions table which includes the title, instructions to publishers and details. Because the details column is  a JSON column, we also need to add a virtual column that converts this to text.

I had a bit of a time testing this too, but finally discovered that they weren't working because all of our tests run in transactions, meaning the indexes didn't get built, so the search didn't work. To get around this, I've disabled transactions for that particular test.

### DIFF
--- a/db/migrate/20250220131003_add_fulltext_index_to_content_block_editions.rb
+++ b/db/migrate/20250220131003_add_fulltext_index_to_content_block_editions.rb
@@ -1,0 +1,17 @@
+class AddFulltextIndexToContentBlockEditions < ActiveRecord::Migration[8.0]
+  INDEX_NAME = "title_details_instructions_to_publishers".freeze
+
+  def up
+    change_table :content_block_editions, bulk: true do |t|
+      t.virtual :details_for_indexing, type: :text, as: "JSON_UNQUOTE(details)", stored: true
+      t.index %i[title details_for_indexing instructions_to_publishers], name: INDEX_NAME, type: :fulltext
+    end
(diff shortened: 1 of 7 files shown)


db/migrate/20250220131003_add_fulltext_index_to_content_block_editions.rb
```diff
@@ -0,0 +1,17 @@
+class AddFulltextIndexToContentBlockEditions < ActiveRecord::Migration[8.0]
+  INDEX_NAME = "title_details_instructions_to_publishers".freeze
+
+  def up
+    change_table :content_block_editions, bulk: true do |t|
+      t.virtual :details_for_indexing, type: :text, as: "JSON_UNQUOTE(details)", stored: true
```
Harriethw, line 6: cool!

Harriethw (approved):
nice one, thanks for researching and describing the solution so well. TIL!

Harriethw:
ooh wait, is there some stray changes on the change_notes in schema.rb that shouldn't be there?

pezholio:
Good catch - have removed, though I suspect these might keep coming back

PR #9977, opened 2025-02-24
Add support for multiple select in 'select_with_search' component

ChrisBAshton:
This is needed to deprecate the 'autocomplete' component. Choices.js supports multiple select mode out of the box: we just need to pass the `multiple: true` boolean to the Rails `select_tag` helper (which adds the `multiple="multiple"` attribute to the rendered `<select>` tag).

## Screenshots of 'multiple' support so far

The resulting UI is a bit different to the 'autocomplete' equivalent, but not so different that it should be detrimental to our users.

[Related guides](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/detailed-guides/new) dropdown. NB: there is a similar ([slow](https://govuk.zendesk.com/agent/tickets/5839628)) performance in both cases.

The landing page view has a similar affordance (we lose the 'dropdown' icon, but gain some placeholder text. Either one is relatively simple to add/remove if necessary). The search experience is identical too. What differs is the 'on choose' behaviour: 'autocomplete' exits the dropdown whereas 'select with search' keeps the dropdown open. The representation of selected options is different too ('autocomplete' shows the options below the input, and 'select with search' shows them inline).

| Before | After |
|-------|-------|
|related_guides-closed-ac|related_guides-closed-sws| | On page load | On page load |
|related_guides-open-ac|Screenshot 2025-02-20 at 14 54 38| | On focus | On focus |
|related_guides-selected-ac|Screenshot 2025-02-20 at 14 51 59| | On choice made | On choice made |
|As above |related_guides-selected-sws| | On blur | On blur |

Trello: https://trello.com/c/jBjHrNou/3492-replicate-ux-of-multiple-autocomplete-into-selectwithsearch-component

### DIFF
--- a/app/assets/javascripts/components/select-with-search.js
+++ b/app/assets/javascripts/components/select-with-search.js
@@ -28,6 +28,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
       searchResultLimit: 100,
       removeItemButton: this.select.multiple,
       labelId: this.select.id + '_label',
+      callbackOnInit: function () {
+        // For the multiple select, move the input field to
+        // the top of the feedback area, so that the selected
+        // 'lozenges' appear afterwards in a more natural flow
+        if (this.dropdown.type === 'select-multiple') {
+          const inner = this.containerInner.element
+          const input = this.input.element
+          inner.prepend(input)
+        }
+      },
       // https://fusejs.io/api/options.html
       fuseOptions: {
         ignoreLocation: true, // matches any part of the string
--- a/app/assets/javascripts/components/select-with-search.js
+++ b/app/assets/javascripts/components/select-with-search.js
@@ -15,7 +15,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
     )
 
     if (placeholderOption && placeholderOption.textContent === '') {
-      placeholderOption.textContent = 'Select one'
+      placeholderOption.textContent = this.select.multiple
+        ? 'Select all that apply'
+        : 'Select one'
     }
 
     this.choices = new window.Choices(this.select, {
@@ -24,6 +26,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
       shouldSort: false, // show options and groups in the order they were given
       itemSelectText: '',
       searchResultLimit: 100,
+      removeItemButton: this.select.multiple,
       labelId: this.select.id + '_label',
       // https://fusejs.io/api/options.html
       fuseOptions: {
--- a/app/views/components/docs/select_with_search.yml
+++ b/app/views/components/docs/select_with_search.yml
@@ -184,3 +184,22 @@ examples:
         value: option2
       - text: Option three
         value: option3
+  with_multiple_select_enabled:
+    description: Allow multiple items to be selected and de-selected.
+    data:
+      id: dropdown-with-multiple
+      label: Select your country
+      include_blank: true
+      select:
+        multiple: true
+      options:
+        - text: France
+          value: fr
+          selected: false
+        - text: Germany
+          value: de
+          selected: false
+        - text: The United Kingdom of Great Britain and Northern Ireland
+          value: uk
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/components/select-with-search.js
```diff
@@ -24,7 +26,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
       shouldSort: false, // show options and groups in the order they were given
       itemSelectText: '',
       searchResultLimit: 100,
+      removeItemButton: this.select.multiple,
       labelId: this.select.id + '_label',
+      callbackOnInit: function () {
+        if (this.dropdown.type === 'select-multiple') {
```
ChrisBAshton, line 32: An inline comment explaining what this is doing / why this is needed would be good.

app/views/components/docs/select_with_search.yml
```diff
@@ -184,3 +184,22 @@ examples:
...
+        - text: France
+          value: fr
+          selected: false
+        - text: Germany
+          value: de
+          selected: false
+        - text: The United Kingdom of Great Britain and Northern Ireland
+          value: uk
+        - text: Democratic Republic of the Congo
+          value: cg
```
ChrisBAshton, line 205: Nitpick: newline at the end of the file please

ChrisBAshton (commented):
Looks good, @matthillco - thanks for persevering with this 🎉 

My only feedback would be that the hover/focus styles aren't great. Could we style it to look more like, say, the focus on the [service navigation component](https://design-system.service.gov.uk/components/service-navigation/)? I was able to achieve the following results with these CSS overrides in my browser:

```
app-c-select-with-search .choices__list--multiple .choices__item {
    padding: 6px 10px;
}

.app-c-select-with-search .choices[data-type*=select-multiple] .choices__button:hover, .app-c-select-with-search .choices[data-type*=select-multiple] .choices__button:focus, .app-c-select-with-search .choices[data-type*=text] .choices__button:hover, .app-c-select-with-search .choices[data-type*=text] .choices__button:focus {
    background-color: #fd0;
    box-shadow: 0 -2px #fd0, 0 4px #0b0c0c;
}

.app-c-select-with-search .choices[data-type*=select-multiple] .choices__button, .app-c-select-with-search .choices[data-type*=text] .choices__button {
    top: -2px;
    padding-top: 6px;
    padding-bottom: 6px;
    padding-right: 16px;
}
```

Before:

Screenshot 2025-03-06 at 16 51 02

After:

Screenshot 2025-03-06 at 16 57 54

matthillco:
@ChrisBAshton 

> My only feedback would be that the hover/focus styles aren't great. Could we style it to look more like, say, the focus on the [service navigation component](https://design-system.service.gov.uk/components/service-navigation/)? I was able to achieve the following results with these CSS overrides in my browser:

Part of the problem is that the multi-select allows each lozenge to be selected by clicking the label (not the remove button). The selected state uses a yellow background. I'm unsure what further actions are available when these items are selected, it's unclear from the `choices.js` docs. We tried to disable that event to no avail. I don't think it would matter if the user selects these items and there's no feedback for that, since we're not using that feature.

So the selected state is a little strange and I'll remove that in favour of improved focus and hover states as you suggested. I'm exploring that currently, will share something later this morning.

app/assets/javascripts/components/select-with-search.js
matthillco: OK done.

app/views/components/docs/select_with_search.yml
matthillco: Done.

matthillco:
@ChrisBAshton New styles for the buttons are now as follows:

1. Default state has better padding/margins and a centred delete button with a border that meets the top and bottom: 
image

2. Hover state matches the styling of the Secondary Solid Button from the components gem:
image

3. Focus state matches the styling of the Secondary Solid Button from the components gem:
image

PR #9985, opened 2025-02-26
(941) Hide details if there is no change history

pezholio:
https://trello.com/c/REP99FWH/941-empty-details-of-changes-in-change-history

Sometimes change history can be present, but doesn’t have items within it. To get around this, we move the logic to the component and only render the details if there are any details rendered.

Additionally, I’ve changed the version factory to build, rather than create so we don’t have to create versions in the database unnecessarily


Harriethw (approved):
👍

PR #9986, opened 2025-02-26
Link checker report edition ids

ChrisBAshton:
Adds `edition_id` association to Link Checker Reports. We'll run the migration and then remove the old `link_reportable_type`/`link_reportable_id` fields which are an unnecessary layer of abstraction.

The polymorphic association was added in November 2017 with no explanation: https://github.com/alphagov/whitehall/pull/3557/commits/03c47344a42243f96990559e33125f73edc4d00b

Since then, there doesn't seem to have been a single link check report which is associated with anything other than an Edition:

```
$ LinkCheckerApiReport.distinct.pluck(:link_reportable_type)
=> ["Edition"]
```

Therefore let's remove the obtuse reference and make it clear
that a link check report is associated with an edition (not a
'linkable'). This should also cut down on quite a lot of DB space,
since we're currently storing the word "Edition" on every single
link check report (almost 12 million rows in the DB).

Loosely related to https://trello.com/c/tmnht4P1/

---


ryanb-gds (approved):
Nice to simplify this a bit. We might need to have a chat to the platform team about how best to run long running migrations like this though

db/data_migration/20250303155753_migrate_link_check_report_ids.rb
```diff
@@ -0,0 +1,29 @@
...
+  # Delete the associated links before deleting each report (this will respect the has_many :links association)
+  reports_to_delete.each do |report|
+    report.links.delete_all
+    report.destroy!
+  end
+
+  puts "#{index + 1}/#{link_reportable_ids_count} - Edition ID #{link_reportable_id}: latest link_checker_api_report ID is #{latest_report.id} (updated at: #{latest_report.updated_at}). Deleted #{reports_to_delete.count} older reports."
+end
+
+puts "Finished cleanup of old LinkCheckerApiReports."
```
ryanb-gds, line 29: Depending on how this tests performance-wise you could probably get this down to three database queries if you wanted, with a bit of subquerying. If the performance is acceptable as is though then might not be worth worrying about.

ChrisBAshton: On my run on Staging earlier, it took somewhere around 45 minutes, from memory. I'm comfortable with that sort of time frame, as it shouldn't lock the DB.

(PS: pushed up to rebase and tweak some commit messages. The code itself is the same).

PR #9995, opened 2025-02-27
(936) Fix weird entries in the change history

pezholio:
Trello card: https://trello.com/c/gW7ssapF/936-weird-entries-in-the-change-history

This ensures that when we clone an edition, we don't copy the change notes, as well as updating the history when adding a new embedded object

## Screenshot

image

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.rb
@@ -13,13 +13,24 @@ private
   attr_reader :version, :schema, :is_first_published_version, :is_latest
 
   def title
-    case version.state
-    when "published"
+    if version.is_embedded_update?
+      "#{updated_subschema_id.humanize.singularize} created"
+    elsif version.state == "published"
       is_first_published_version ? "#{version.item.block_type.humanize} created" : version.state.capitalize
-    when "scheduled"
+    elsif version.state == "scheduled"
       "Scheduled for publishing on #{version.item.scheduled_publication.to_fs(:long_ordinal_with_at)}"
     else
       "#{version.item.block_type.humanize} #{version.state}"
+    end
+  end
+
+  def updated_subschema_id
+    version.updated_embedded_object_type
+  end
+
+  def new_subschema_item_details
+    version.field_diffs.dig("details", updated_subschema_id, version.updated_embedded_object_name).map do |field_name, diff|
+      [field_name.humanize, diff.new_value]
     end
   end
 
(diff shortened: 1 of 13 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/version.rb
```diff
@@ -14,6 +14,10 @@ def field_diffs
       def event_details
         self[:event_details] || {}
       end
+
+      def is_embedded_update?
+        event_details["updated_embedded_object_type"] && event_details["updated_embedded_object_name"]
```
Harriethw, line 19: as we know the shape of this data - it's two string always with the same key right? - would there be anything wrong with just adding columns to the versions table? 
Just wondering as the naming of `event_details` atm feels a bit vague to me, and I miss the security of strong types... If we want to keep the blob could it be under `embedded_object_metadata` or something like that?

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/document_timeline/timeline_item_component.rb
```diff
@@ -13,16 +13,27 @@ def initialize(version:, schema:, is_first_published_version:, is_latest:)
...
-    when "scheduled"
+    elsif version.state == "scheduled"
       "Scheduled for publishing on #{version.item.scheduled_publication.to_fs(:long_ordinal_with_at)}"
     else
       "#{version.item.block_type.humanize} #{version.state}"
     end
   end
 
+  def updated_subschema
+    @updated_subschema ||= @schema.subschema(version.event_details["updated_embedded_object_type"])
```
Harriethw, line 28: do we go and get the schema here in case the embedded object name has changed? Wondering if would be better to use the name at the time the version is created, because otherwise schemas will always have to be backwards compatible or versioned?

lib/engines/content_block_manager/app/models/content_block_manager/content_block/version.rb
pezholio: Yeah, I wasn't 100% happy with this approach TBH. I think individual columns might be OK actually. Lemme give that a go

lib/engines/content_block_manager/app/models/content_block_manager/content_block/version.rb
```diff
@@ -10,6 +10,10 @@ class Version < ApplicationRecord
       def field_diffs
         self[:field_diffs] ? ContentBlock::DiffItem.from_hash(self[:field_diffs]) : {}
       end
+
```
Harriethw, line 13: is this a stray method left over from a refactor?

pezholio: Whoops, yes!

Harriethw (approved):

PR #9998, opened 2025-02-28
Fix schema on `main`

ChrisBAshton:
I've followed the instructions at
https://docs.publishing.service.gov.uk/repos/govuk-docker/troubleshooting.html#rails-database-schema-changed-unexpectedly-when-running-rake-dbmigrate

I cleaned my local docker setup completely, ran `make whitehall` and created the DB and ran `db:migrate` and the schema file changed.

Committing this so that in my next PR there is less of a diff

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -98,7 +98,7 @@ ActiveRecord::Schema[8.0].define(version: 2025_02_20_131003) do
 
   create_table "call_for_evidence_responses", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.integer "edition_id"
-    t.text "summary", size: :medium
+    t.text "summary"
     t.datetime "created_at", precision: nil
     t.datetime "updated_at", precision: nil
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
Ah, I think this is, indirectly, caused by the UTF8MB4 migration. Because UTF8MB4 requires an extra byte per character, text fields created against a UTF8MB3 schema are restricted to a shorter length to make sure they don't get truncated.

I think we need to ask everybody developing against Whitehall to drop their local DBs, in case they are still creating migrations against a UTF8MB3 database. However, that doesn't explain the change here to calls for evidence responses - I'm sure that wouldn't have changed in the last two years 🤔 it's possible I missed it when doing the migration I suppose, though I'm pretty sure I wouldn't have committed some lines of the schema file and not others.

ChrisBAshton:
You've tested locally and have the same changes as I do, plus some others, which you've attributed to stale data - so we're happy to proceed with this 👍

PR #10012, opened 2025-03-04
Ensure each document is only republished once when republishing all documents of a document type

ryanb-gds:
We can do this by making sure that we only queue a document for editions which are the latest edition of the document. Every document has one latest edition. Previously we were kicking off a republishing job for every edition of each document. This caused an excessive number of jobs and could result in table locking issues if republishing jobs for the same document attempt to run concurrently.


ChrisBAshton (approved):
Nice one!

PR #10015, opened 2025-03-04
Fix intermittent test failures in CI

ChrisBAshton:
Fixes our flaky CI tests by:

1. fixing up tests to be more robust where possible
2. where not possible, running the tests in isolation to avoid race conditions

This PR allows us to add a `# RUN_IN_ISOLATION` magic comment to any test that is sensitive to being run in parallel. So far this is just the `ScheduledPublishingWorkerTest`, because it queries the `Sidekiq::ScheduledSet` directly, and that set can have its contents altered by other tests at runtime.

Test files with the magic comment are put onto their own node in GitHub Actions, and automatically configures the parallelize gem to run tests sequentially on that node, so all of our flaky tests can be run one after the other, at a slower speed but a greater reliability.

I really didn't want to do this. I tried switching up this test to use Sidekiq's "inline mode" for isolation, which forces jobs to execute immediately instead of being queued, removing race conditions. It's the approach we've already been applying elsewhere in Whitehall:
https://github.com/alphagov/whitehall/blob/d045ee80abfa1aad6750b779132b4baf3ca03adf/test/unit/lib/whitehall/publishing_api_test.rb#L98

Unfortunately it’s not as simple as setting the `scheduled_publishing_worker_test` into inline mode, as it’s some other test(s) that are polluting the ScheduledSet that are causing it to fail. So we’d have to go and fix all other tests to make this one test file pass consistently. That could quickly turn into whackamole...!

I then looked into [forcing Sidekiq to run inline globally](https://github.com/alphagov/whitehall/commit/a5a0589d89a05c8903592ffc75fea4cb4ce6c494), but that didn't appear to make any difference (nor did it allow me to [delete a whole bunch of 'Sidekiq::Testing.inline!' calls](https://github.com/alphagov/whitehall/commit/7cb6c5cb18a564fe3aef357411ed55b4f98e3c1c?w=1) as I'd hoped).

As a result:

- Test duration is [reduced to 1m11s](https://github.com/alphagov/whitehall/actions/runs/13672935533/job/38226967266?pr=10015) on the 'isolated' node
- 2m40s, 2m40s and 2m57s on the other nodes.
- By comparison, a [recent run on another branch](https://github.com/alphagov/whitehall/actions/runs/13672752674/job/38226418560) took 2m6s, 2m10s, 2m36s and 2m38s.
- Comparing the slowest nodes in each test run, the introduction of this 'isolated' mode has the potential to slow down our overall CI pipeline by 19 seconds.
- In reality, however, that's dwarfed by the (unaffected) Cucumber test suite, which [takes 6 and a half minutes](https://github.com/alphagov/whitehall/actions/runs/13672935533/job/38226966897?pr=10015) - so devs should not really be waiting any longer for the Minitest suite to pass.

---

### DIFF
--- a/bin/minitest-ci
+++ b/bin/minitest-ci
@@ -4,17 +4,32 @@
 # defined in .github/workflows/minitest.yml.
 # It splits the MiniTest suite into randomly-allocated groups
 # which are executed across multiple GitHub Actions 'matrix' nodes.
+
 test_files = [
   Dir["test/**/*_test.rb"],
   Dir["lib/engines/**/test/**/*_test.rb"],
 ].flatten
 
-tests = test_files.
-  sort.
-  shuffle(random: Random.new(ENV['GITHUB_SHA'].to_i(16))).
-  select.
-  with_index do |el, i|
-    i % ENV["CI_NODE_TOTAL"].to_i == ENV["CI_NODE_INDEX"].to_i
-  end
+# Detect isolated test files (those with # RUN_IN_ISOLATION in the first 10 lines)
+isolated_tests, regular_tests = test_files.partition do |file|
+  File.foreach(file).first(10).grep(/# RUN_IN_ISOLATION/).any?
+end
 
-exec "bundle exec rails test #{tests.join(" ")}"
+# Determine which CI node is running
+node_index = ENV["CI_NODE_INDEX"].to_i
+total_nodes = ENV["CI_NODE_TOTAL"].to_i
+
+if node_index == 0
+  # First CI node runs all isolated tests sequentially
+  tests = isolated_tests
+  exec({ "RUN_IN_PARALLEL" => "false" }, "bundle exec rails test #{tests.join(' ')}")
+else
+  # Other nodes distribute regular tests
+  tests = regular_tests.
+    sort.
+    shuffle(random: Random.new(ENV['GITHUB_SHA'].to_i(16))).
+    select.
+    with_index { |_, i| i % (total_nodes - 1) == (node_index - 1) }
+
+  exec({ "RUN_IN_PARALLEL" => "true" }, "bundle exec rails test #{tests.join(' ')}")
+end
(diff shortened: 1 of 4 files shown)


bin/minitest-ci
```diff
@@ -1,20 +1,30 @@
 #!/usr/bin/env ruby
 
-# This script is used during the GitHub Actions workflow
-# defined in .github/workflows/minitest.yml.
-# It splits the MiniTest suite into randomly-allocated groups
-# which are executed across multiple GitHub Actions 'matrix' nodes.
```
ryanb-gds, line 6: I quite liked this comment - at least the first two lines. I wasn't sure initially if this file was specific to GitHub actions or not

bin/minitest-ci
```diff
@@ -1,20 +1,30 @@
...
+isolated_tests, regular_tests = test_files.partition do |file|
+  File.foreach(file).first(10).grep(/# RUN_IN_ISOLATION/).any?
+end
 
-exec "bundle exec rails test #{tests.join(" ")}"
+# Determine which CI node is running
+node_index = ENV["CI_NODE_INDEX"].to_i
+total_nodes = ENV["CI_NODE_TOTAL"].to_i
+
+if node_index == 0
```
ryanb-gds, line 17: Does this mean we've effectively lost a node from our CI pipeline, as one of the nodes is now only running one test? Can we bump it up so we don't increase the CI pipeline time?

ryanb-gds (approved):
Couple of thoughts, but looks good

bin/minitest-ci
ChrisBAshton: Agreed 👍

bin/minitest-ci
ChrisBAshton: It does mean that, yes - but the impact is minimal (see PR description).

ryanb-gds: Ah, yes. We should look at perhaps applying the same parallelization technique to the cucumber tests

PR #10024, opened 2025-03-05
Surface fixes for link validation errors

minhngocd:
Currently fixes for any link validation errors that prevents publishing and scheduling are silent. The user only sees a high level message that there was something wrong with the links and not how to go about fixing it. If we surface the fixes we may receive less support tasks to manage.

https://trello.com/c/zuRPc1sS/3504-unable-to-publish-whitehall-drafts-unclear-broken-links-state

image

image

### DIFF
--- a/app/services/edition_scheduler.rb
+++ b/app/services/edition_scheduler.rb
@@ -25,17 +25,16 @@ class EditionScheduler < EditionService
       reasons << "This edition does not have a scheduled publication date set"
     elsif scheduled_publication_is_not_within_cache_limit?
       reasons << "Scheduled publication date must be at least #{Whitehall.default_cache_max_age / 60} minutes from now"
-    elsif govspeak_link_errors.any?
-      reasons << "This edition contains links which violate linking guidelines."
-      reasons.concat govspeak_link_errors.pluck(:fix).uniq
+    elsif govspeak_link_validator.errors.any?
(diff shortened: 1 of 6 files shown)


minhngocd:
Hmmmm good point. Adding some screen shots. It currently looks like so, which isn't very readable:

Screenshot 2025-03-06 at 11 59 48

Making them an array / bullet point is slightly better but stlil have loads of room for improvement. We need some content / frontend guidance here

Screenshot 2025-03-06 at 12 05 55

minhngocd:
Awaiting Content Guidance

minhngocd:
merged in changes from https://github.com/alphagov/whitehall/pull/10175
@lauraghiorghisor-tw do you want to do a quick re-review?

app/services/edition_scheduler.rb
```diff
@@ -10,21 +10,33 @@ def past_participle
   end
 
   def failure_reason
```
lauraghiorghisor-tw, line 12: There is no screenshot for this. So the scheduler now returns a list of errors (failure reasons). Is this singular `failure_reason` just required by the inheritance from EditionService or still used?

minhngocd: it is. There seems to be different instances in the view that uses the different method `failure_reason` vs `failure_reasons`. I believe the message that says scheduling failed will contain only one single reason. can grab a screenshot for the schedule case

PR #10032, opened 2025-03-06
Don't retry jobs that fail with `GdsApi::HTTPPayloadTooLarge`

ChrisBAshton:
We're seeing a lot of republish job failures off the back of a recent bulk republish, whereby the documents in question are just far too large for the publishing platform to handle (too many attachments etc):
https://govuk.sentry.io/issues/6332398641/?environment=production&project=202259&referrer=issue-stream&statsPeriod=1h&stream_index=0

It's pointless retrying these, as the payload size won't change in between retries, and therefore it's always going to fail. We should instead log the exception, and kill the job.

### DIFF
--- a/test/integration/workers/publishing_api_document_republishing_worker_integration_test.rb
+++ b/test/integration/workers/publishing_api_document_republishing_worker_integration_test.rb
@@ -130,6 +130,18 @@ class PublishingApiDocumentRepublishingWorkerIntegrationTest < ActiveSupport::Te
 
       assert_all_requested(requests)
     end
+
+    test "Should not retry if encountering `GdsApi::HTTPPayloadTooLarge` exception" do
+      edition = create(:published_publication)
+      Whitehall::PublishingApi.expects(:publish).raises(GdsApi::HTTPPayloadTooLarge.new("413 Request Entity Too Large"))
+      worker = PublishingApiDocumentRepublishingWorker.new
+
+      assert_equal(worker.sidekiq_options_hash["retry"], true)
+      assert_raises(GdsApi::HTTPPayloadTooLarge) do
+        worker.perform(edition.document.id)
+      end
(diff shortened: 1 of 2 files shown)


test/integration/workers/publishing_api_document_republishing_worker_integration_test.rb
```diff
@@ -130,6 +130,18 @@ class PublishingApiDocumentRepublishingWorkerIntegrationTest < ActiveSupport::Te
...
+
+    test "Should not retry if encountering `GdsApi::HTTPPayloadTooLarge` exception" do
+      edition = create(:published_publication)
+      Whitehall::PublishingApi.expects(:publish).raises(GdsApi::HTTPPayloadTooLarge.new("413 Request Entity Too Large"))
+      worker = PublishingApiDocumentRepublishingWorker.new
+
+      assert_equal(worker.sidekiq_options_hash["retry"], true)
+      assert_raises(GdsApi::HTTPPayloadTooLarge) do
+        worker.perform(edition.document.id)
+        assert_equal(worker.sidekiq_options_hash["retry"], 0)
```
minhngocd, line 142: Maybe this needs to come out of the do block? the worker.perform will raise an error which will short circuit this block. and the assert_equal never gets called. For example, i've added `assert_false true` to line 142 and the test was still passing

minhngocd (changes requested):

test/integration/workers/publishing_api_document_republishing_worker_integration_test.rb
ChrisBAshton: Oof, good spot! I'm usually good at double-checking my test fails if I change the assertion. Fixed!

GDSNewt (approved):
Makes sense 👍

minhngocd (approved):

PR #9983, opened 2025-02-25
Add GA4 tracking to `select_with_search`

gtvj:
## What

Send data to GA4 upon changes to `select_with_search` component. Details provided in the [Trello card](https://trello.com/c/gm92ppp8/776-component-level-tracking-track-dropdown-interactions-on-whitehall) include requirements to send data to GA4 in response to:

1. `change` events triggered on `select` elements 
2. removal of an item from the multi-select variant of `select_with_search`

## How

The requirements have been met by introducing:
1. new `data-` attributes into the component HTML 
2. a JavaScript analytics module to send data to GA4 in response to `addItem` and `removeItem` events 

`addItem` and `removeItem` are custom events triggered by `choices.js` on the underlying `select` elements when a user performs these actions on a mounted `select_with_search` component.

## Performance analyst review

On 13 March, PAs tested the implementation on Integration and said it seems fine. 

## Paths tested

The following paths have been provided by PAs in response to their testing. Here's what data is being provided to GA4 for each.

### New consultation (path is `/government/admin/consultations/new`)

Lead organisation 1

```json
{
 "event": "event_data",
 "event_data": {
  "action": "select",
  "event_name": "select_component",
  "index": {
   "index_section_count": "1",
   "index_section": "17"
  },
  "section": "Lead organisation 1",
  "text": "HM Revenue & Customs (HMRC)",
  "type": "new-consultations"
 },
 "govuk_gem_version": "54.0.1",
 "timestamp": "1741857295779"
}
```

Lead organisation 2

```json
{
 "event": "event_data",
 "event_data": {
  "action": "select",
  "event_name": "select_component",
  "index": {
   "index_section_count": "1",
   "index_section": "18"
  },
  "section": "Lead organisation 2",
  "text": "HM Revenue & Customs (HMRC)",
  "type": "new-consultations"
 },
 "govuk_gem_version": "54.0.1",
 "timestamp": "1741857401033"
}
```

Lead organisation 3

```json
{
 "event": "event_data",
 "event_data": {
  "action": "select",
  "event_name": "select_component",
  "index": {
   "index_section_count": "1",
   "index_section": "19"
  },
  "section": "Lead organisation 3",
  "text": "HM Revenue & Customs (HMRC)",
  "type": "new-consultations"
 },
 "govuk_gem_version": "54.0.1",
 "timestamp": "1741857446887"
}
```

Lead organisation 4

```json
{
 "event": "event_data",
 "event_data": {
  "action": "select",
  "event_name": "select_component",
[...]

PR #10069, opened 2025-03-17
Track `select` interactions in Whitehall

gtvj:
## What

Sends data to GA4 in response to `change` events that occur on those `<select>` elements which are not part of a `select_with_search` component. Further information is provided in the [Trello card](https://trello.com/c/xQAjZFsB/3446-component-level-tracking-track-select-dropdown-interactions-on-whitehall)

- [x] Introduce new analytics module with tests (See 6a2d54a496013155cc9dec37d2a2f697a20b8ed8)
- [x] Decorate native `<select>` elements within the codebase (See 310e6cc5c31d463e37c7c22d5881d7ad301be7a9)
- [x] Understand how we might decorate `<select>` elements coming from the components Gem with `data_attributes`[^1]
- [x] Decorate all ~29 instances of `govuk_publishing_components/components/select` in Whitehall
- [x] PAs to confirm approach to indexing grouped and visually hidden fields

## How 

Introduces a new analytics `GA4SelectSetup` module that sends data to GA4 in response to `change` events on _only_ those select elements that are not part of a `select_with_search` component (See https://github.com/alphagov/whitehall/pull/9983 for details of how the `select_with_search` components are tracked using custom events provided by choices.js).

Following the implementation of other analytics modules, a new module is added to the HTML (`ga4-select-setup`) to allow for control of when this tracking is enabled.

## Performance analyst review

This has been reviewed and accepted by Jinish Gohil, team PA.

## Related PRs

- https://github.com/alphagov/whitehall/pull/9983 Introduced GA4 tracking to `select_with_search` components
- https://github.com/alphagov/govuk_publishing_components/pull/4723 updated the publishing components gem to allow select elements to accept data attributes
- https://github.com/alphagov/whitehall/pull/10122 introduced a JavaScript module that populates the `ga4-index-section` client-side

[^1]: Achieving this required an update to the publishing components gem, which was done in this PR https://github.com/alphagov/govuk_publishing_components/pull/4723

### DIFF
--- a/app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
@@ -1,0 +1,44 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
+;(function (Modules) {
+  Modules.Ga4SelectSetup = {
+    init: function () {
+      const moduleElements = document.querySelectorAll(
+        "[data-module~='ga4-select-setup']"
+      )
+
+      moduleElements.forEach(function (moduleElement) {
+        moduleElement.addEventListener('change', function (event) {
+          const ga4DocumentType = moduleElement.dataset.ga4DocumentType
+
+          if (
+            event.target.tagName.toLowerCase() === 'select' &&
+            !event.target.getAttribute('hidden')
+          ) {
+            const eventData = {
+              event: 'event_data',
+              event_data: {
+                event_name: 'select_content',
+                type: event.target.dataset.ga4DocumentType || ga4DocumentType,
+                index: {
+                  index_section_count: `${event.target.selectedIndex}`,
+                  index_section: event.target.dataset.ga4IndexSection
+                },
+                text: event.target[event.target.selectedIndex].label,
+                section:
+                  event.target.dataset.ga4Section ||
+                  event.target.labels[0].innerText,
+                action: 'select'
+              }
+            }
+
+            window.GOVUK.analyticsGa4.core.sendData(eventData, 'event_data')
+          }
+        })
+      })
+    }
+  }
+})(window.GOVUK.analyticsGa4.analyticsModules)
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -23,6 +23,7 @@
 //= require admin/analytics-modules/ga4-page-view-tracking.js
 //= require admin/analytics-modules/ga4-paste-tracker.js
 //= require admin/analytics-modules/ga4-select-with-search-tracker.js
+//= require admin/analytics-modules/ga4-select-tracker.js
 
 //= require admin/modules/document-history-paginator
 //= require admin/modules/locale-switcher
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -22,7 +22,8 @@
 
   <%= render partial: "shared/header" %>
 
-  <div class="govuk-width-container" data-module="ga4-event-tracker ga4-paste-tracker ga4-link-setup ga4-link-tracker ga4-button-setup ga4-select-with-search-setup ga4-index-section-setup">
+  <div class="govuk-width-container" data-module="ga4-event-tracker ga4-paste-tracker ga4-link-setup ga4-link-tracker ga4-button-setup ga4-select-with-search-setup ga4-index-section-setup ga4-select-setup" data-ga4-document-type="<%= action_name %>-<%= controller_name %>">
+
     <%= render "shared/phase_banner", {
       show_feedback_banner: t("admin.feedback.show_banner"),
     } %>
--- a/spec/javascripts/admin/analytics-modules/ga4-select-tracker.spec.js
+++ b/spec/javascripts/admin/analytics-modules/ga4-select-tracker.spec.js
@@ -1,0 +1,122 @@
+describe('GOVUK.analyticsGa4.analyticsModules.Ga4SelectSetup', function () {
+  let container, select, option, label, expectedAttributes
+
+  const documentType = 'document-type'
+  const labelContent = 'Select label'
+
+  beforeEach(function () {
+    expectedAttributes = {
+      event: 'event_data',
+      event_data: {
+        event_name: 'select_content',
+        type: 'new-consultations',
+        index: {
+          index_section_count: '0',
+          index_section: '20'
+        },
+        text: 'Email address for ordering attachment files',
+        section: 'Settings',
+        action: 'select'
+      }
+    }
+
+    container = document.createElement('div')
+    container.setAttribute('data-module', 'ga4-select-setup')
+    container.setAttribute('data-ga4-document-type', documentType)
+
+    label = document.createElement('label')
+    label.innerText = labelContent
+    label.setAttribute('for', 'select-test')
+
+    select = document.createElement('select')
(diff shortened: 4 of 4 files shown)


app/views/admin/editions/_settings_fields.html.erb
```diff
@@ -5,7 +5,14 @@
         margin_bottom: 6,
     } %>
 
-    <%= render "accessible_attachment_field", form: form, edition: edition %>
+    <%= render "accessible_attachment_field",
+               form: form,
+               edition: edition,
+               ga_data: {
+                 document_type: "#{action_name}-#{controller_name}",
+                 index_section: 24,
```
ryanb-gds, line 13: Can we always guarantee that this field will be the 24th field in the view (if that is what index_section is intended to communicate)? The number of fields on an edition form varies between edition types

ryanb-gds (changes requested):
One question about the intent for the index section value. There's also an interim commit in there, would you mind squashing it? Ta

app/views/admin/editions/_settings_fields.html.erb
gtvj: Thanks for spotting this @ryanb-gds - I didn't know it was the case that edition form rendering was dynamic in this way. I'll look into this and introduce the necessary change.

ryanb-gds: Worth noting that there may be other forms this applies to, but I'm not sure off of the top of my head. Possibly worth considering a general approach of iterating through all of the inputs and selects on the screen and setting the index section based on the count, but I don't know if that might bring its own issues

gtvj: Thanks @ryanb-gds - I think I see the value in this kind of thing being more dynamic. 

If I've understood what you're describing, this would involve changing the way form fields are rendered in Whitehall so that HTML form fields are the product of iterating over an array of objects (representing things like the type of field, the label text, the value, and optional fields - such as an iterable collection of nested options for selects).

I'm wondering how this would work in circumstances where, for example, the HTML being rendered represents the output of multiple includes, each of which may include conditionals (or potentially looping logic, I suppose) to determine which fields are output? 

I'm thinking of the example you've commented on above. If I've understood the code correctly, this is included in a form _after_ [another include](https://github.com/alphagov/whitehall/blob/908972d9b63d3565c43acd1c63f6cf67119e61d6/app/views/admin/corporate_information_pages/_form.html.erb#L2) that seems to have quite a few conditional fields.

ryanb-gds: That's not quite what I'm suggesting - I'd suggest we run some javascript when the dom has loaded to select all of the form inputs/selects and add the index section attribute for each field. Much like we do for each tracking module anyway. Maybe you could even use the existing module and just set the index_section to the value of the loop index, but I don't know if that would cover all field types

gtvj: Sounds good - I'll look into that now 👍🏽

gtvj:
Hi @ryanb-gds - just spotted your comment above about the interim commit needing to be squashed. There seems to have been a bit of miscommunication because I'd not actually intended for the PR to be reviewed yet. I was only committing a work in progress before going on leave (in case I got hit by a bus or something). The PR is still in Draft status and the interim commit message explains "This is an interim commit to save and share my work. It'll be overwritten before I ask for a PR review."

ryanb-gds:
> Hi @ryanb-gds - just spotted your comment above about the interim commit needing to be squashed. There seems to have been a bit of miscommunication because I'd not actually intended for the PR to be reviewed yet. I was only committing a work in progress before going on leave (in case I got hit by a bus or something). The PR is still in Draft status and the interim commit message explains "This is an interim commit to save and share my work. It'll be overwritten before I ask for a PR review."

Sorry @gtvj, I didn't realise that. It was in the review column on Trello, which made us think it was ready, but I'm guessing that was for a review from Jinish

app/components/admin/editions/document_history_tab_component.html.erb
```diff
@@ -18,6 +18,10 @@
     label: "Show",
     heading_size: "s",
     full_width: true,
+    data_attributes: {
+      ga4_document_type: "#{action_name}-#{controller_name}",
```
patrickpatrickpatrick, line 22: We might be not need to do this for each element, we could just do something like

```
<script type="javascript">
    window.GOVUK.analyticsGa4.documentType = "<%= action_name %>-<%= controller_name %>"
</script>
```

in `application.html.erb`. Then we could reference it whenever we need to fire an event like:

```
          const eventData = {
            event: 'event_data',
            event_data: {
              event_name: 'select_component',
              type: window.GOVUK.analyticsGa4.documentType,
              index: {
                index_section_count: `${event.target.selectedIndex}`,
                index_section: event.target.dataset.ga4IndexSection
              },
              text: event.target[event.target.selectedIndex].label,
              section: event.target.dataset.ga4Section,
              action: 'select'
            }
          }
```

app/components/admin/editions/history_mode_form_controls.erb
```diff
@@ -20,6 +20,10 @@
           name: "edition[government_id]",
           label: "Change the associated government:",
           id: "edition_government_id",
+          data_attributes: {
+            ga4_document_type: "#{action_name}-#{controller_name}",
+            ga4_section: "Change the associated government",
```
patrickpatrickpatrick, line 25: Could we take the label and use that for populating the `section` of the event? If it needed to be different for a specific component, we could add logic that would give a specified document type on the select component being changed precedence.

app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
```diff
@@ -0,0 +1,44 @@
...
+        "[data-module~='ga4-select-setup']"
+      )
+
+      moduleElements.forEach(function (moduleElement) {
+        moduleElement.addEventListener('change', function (event) {
+          const ga4DocumentType = moduleElement.dataset.ga4DocumentType
+
+          if (
+            event.target.tagName.toLowerCase() === 'select' &&
+            !event.target.getAttribute('hidden')
```
ryanb-gds, line 19: Is there actually a scenario where we might get a change event on a hidden select?

ryanb-gds: Ah, I'm guessing from the test that we want to exclude events from Choices.js because they're tracked separately?

spec/javascripts/admin/analytics-modules/ga4-select-setup.spec.js
```diff
@@ -0,0 +1,122 @@
...
+
+    select.dispatchEvent(new Event('change', { bubbles: true }))
+
+    expect(mockGa4SendData).toHaveBeenCalledWith(
+      expectedAttributes,
+      'event_data'
+    )
+  })
+
+  it(`does not trigger GA4 in response to change events on selects that are not replaced by choices.js`, function () {
```
ryanb-gds, line 64: I'm a bit confused by the double negative, but seems to me this test name might not be right?

ryanb-gds (approved):
Looks good to me, just a couple of questions but I think this is essentially ready to go

app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
patrickpatrickpatrick: Yep, that's right. This seemed to be the most consistent way to check if it's a Choices.js select as even if we changed the classes applied to the actual `select`, it will always be hidden with a `hidden` attribute.

spec/javascripts/admin/analytics-modules/ga4-select-setup.spec.js
patrickpatrickpatrick: Ah yep, that should be **does not trigger GA4 in response to change events on selects that are replaced by choices.js**. Have changed that now!

PR #10077, opened 2025-03-19
Change link reverification logic

ChrisBAshton:
The old logic was to grab the 500 live editions with the oldest link check reports, per organisation, and re-verify them (i.e. refresh the link check report via Link Checker API, or just update the 'updated_at' of the noop reports [if no links were detected](https://github.com/alphagov/whitehall/blob/a5b91921476686c5a4b0e1b47e26f690c18ba3cc/app/sidekiq/check_organisation_links_worker.rb#L28-L29)).

At time of writing, only 118 of the 1219 organisations in Whitehall have over 500 editions, meaning that the remaining 1101 orgs were having every single link check report refreshed every single day, which is probably overkill. Conversely, orgs with a large number of editions (such as 'environment-agency', which has 28,142 editions at time of writing) are capped at 500 and therefore in that example would take 57 days to cycle through all of its link check reports.

Splitting by organisation seems an oddly arbitrary choice. A better approach, implemented in this commit, is to set a global cap (in this case, 60,000 editions) to refresh every day, and to work globally from last updated to most recently updated. This means we'll cycle through the entire (at time of writing) 366,771 live editions over the course of 7 days.

60,000 was chosen because it's in the ballpark of the current Sidekiq queue: [around 34k-50k jobs](https://grafana.eks.production.govuk.digital/goto/TJAgLshNR?orgId=1) are enqueued every morning at 4am, taking less than 2 hours to complete. The variability stems from the fact that 'noop' link check reports don't create a worker (they just have their 'updated_at' updated, in-place) and thus we never reach our theoretical maximum of 609,500 jobs being enqueued (500 * 1219). (Indeed, for one org, [499 out of 500 editions](https://kibana.logit.io/s/13d1a0b1-f54f-407b-a4e5-f53ba653fac3/app/discover#/doc/filebeat-*/filebeat-2025.03.18?id=8XN6p5UB9dMuYLNb8ULz) were ‘ignored’).


test/unit/app/sidekiq/revalidate_old_link_check_reports_worker_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    original_report = create(:link_checker_api_report_completed, batch_id: 10, edition: publication)
+    stub_post_request(links, link_checker_api_batch_report_hash(id: 10, links:))
+
+    assert_difference "LinkCheckerApiReport.count", 0 do
+      RevalidateOldLinkCheckReportsWorker.new.perform
+
+      assert_equal 1, LinkCheckerApiReport.where(edition_id: original_report.edition_id).count
+      assert_not_equal original_report.id, LinkCheckerApiReport.where(edition_id: original_report.edition_id).first.id
+    end
+  end
```
ryanb-gds, line 46: I feel that you've got enough guarantees of this elsewhere that you don't need to test that we're using a has one association here again

ryanb-gds (approved):
Looks good to me and @lauraghiorghisor-tw

app/sidekiq/revalidate_old_link_check_reports_worker.rb
```diff
@@ -0,0 +1,46 @@
...
+  MAX_REPORTS_TO_CHECK = 60_000
+  FIND_EACH_BATCH_SIZE = 50
+
+  sidekiq_options queue: "link_checks", retry: 6
+  # Stop retrying jobs within a day as new jobs are enqueued each day,
+  # and also retry less times within a day as this isn't a critical
+  # task.
+  sidekiq_retry_in do |retry_count, _exception|
+    # First retry in 15 seconds, 6th (and last) retry at ~5 hours
+    ((0.5 * (retry_count + 1)**4) * 30)
```
GDSNewt, line 13: Nitpick alert; I'm not sure if these multipliers are ever liable to change, but having them as constants after the batch size constant might make this more readable

app/sidekiq/revalidate_old_link_check_reports_worker.rb
```diff
@@ -0,0 +1,46 @@
...
+private
+
+  def public_editions_with_reports
+    Edition.includes(:link_check_report).publicly_visible.with_translations
+  end
+
+  def least_recently_checked_editions
+    least_recently_checked = public_editions_with_reports.order("link_checker_api_reports.updated_at").limit(MAX_REPORTS_TO_CHECK)
+    Edition.where(id: least_recently_checked.pluck(:id))
+  end
```
GDSNewt, line 45: I think these two methods are fine assuming that `public_editions_with_reports` and `link_checker_api_reports.updated_at` will always have a value.

app/sidekiq/revalidate_old_link_check_reports_worker.rb
```diff
@@ -0,0 +1,46 @@
...
+      editions.find_each(batch_size: FIND_EACH_BATCH_SIZE) do |edition|
+        if LinkCheckerApiService.has_links?(edition)
+          LinkCheckerApiService.check_links(edition, admin_link_checker_api_callback_url(host: Plek.find("whitehall-admin")))
+          rechecked += 1
+        else
+          LinkCheckerApiReport.create_noop_report(edition)
+          touched += 1
+        end
+      end
+      logger.info("[link-checking-debug][job_#{jid}]: Done requesting link checks for #{editions.count}. #{touched} of them were noops, whereas #{rechecked} were rechecked using Link Checker API.")
```
GDSNewt, line 32: I think editions.count is fine and is the only realistic option, I did wonder if logging the exact editions being checked would be useful but then quickly realised that would be a horrendous amount of chaff 😄

test/unit/app/sidekiq/revalidate_old_link_check_reports_worker_test.rb
ChrisBAshton: I had the same thought tbh - now removed!

app/sidekiq/revalidate_old_link_check_reports_worker.rb
ChrisBAshton: I didn't give this much of a read tbh (this code is copied over from the previous worker). I agree with you about the readability, but I think the comment helps.

app/sidekiq/revalidate_old_link_check_reports_worker.rb
ChrisBAshton: They will 👍

GDSNewt (approved):
LGTM 👍

PR #10078, opened 2025-03-19
(882) Allow Content blocks to be "soft deleted"

pezholio:
Trello card: https://trello.com/c/uM6yqDs8/882-delete-all-non-pension-blocks-in-production

This allows content blocks to be soft deleted. Additionally, we add a Rake task to soft delete a content block with a given content ID, as long as it has no dependent content.


lib/engines/content_block_manager/lib/tasks/delete_content_block.rake
```diff
@@ -0,0 +1,24 @@
...
+  desc "Delete content block"
+  task :delete_content_block, [:content_id] => :environment do |_t, args|
+    content_id = args[:content_id]
+    content_block_document = ContentBlockManager::ContentBlock::Document.find_by(content_id:)
+
+    abort("A content block with the content ID `#{content_id}` cannot be found") unless content_block_document
+
+    @host_content_items = ContentBlockManager::HostContentItem.for_document(content_block_document)
+
+    abort("Content block `#{content_id}` cannot be deleted because it has dependent content. Try removing the dependencies and trying again") unless @host_content_items.items.count.zero?
```
Harriethw, line 11: really small thing, but could we keep the term "host documents/content" rather than dependent content?

lib/engines/content_block_manager/lib/tasks/delete_content_block.rake
```diff
@@ -0,0 +1,24 @@
+namespace :content_block_manager do
+  desc "Delete content block"
+  task :delete_content_block, [:content_id] => :environment do |_t, args|
+    content_id = args[:content_id]
+    content_block_document = ContentBlockManager::ContentBlock::Document.find_by(content_id:)
+
+    abort("A content block with the content ID `#{content_id}` cannot be found") unless content_block_document
```
Harriethw, line 7: neat, didn't know about `abort`

lib/engines/content_block_manager/lib/tasks/delete_content_block.rake
pezholio: Done! 👍

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/soft_deletable.rb
```diff
@@ -2,6 +2,10 @@ module ContentBlockManager
   module ContentBlock::Document::SoftDeletable
     extend ActiveSupport::Concern
 
+    included do
+      default_scope { where(deleted_at: nil) }
```
Harriethw, line 6: really like the approach of adding `deleted_at` in general, but wondering if there should be some change to the `state` of the document's editions? e.g. they are now in a `deleted` state. Or do you think that's a bit too complicated at this point?

I'm also wondering if we keep all the scoping logic in once place - we have the `live` scope here that the `DocumentFilter` uses https://github.com/alphagov/whitehall/blob/eb94bca2a74591c37fb8f3f29241b36ef83f2cc9/lib/engines/content_block_manager/app/models/content_block_manager/content_block/document.rb#L29

lib/engines/content_block_manager/lib/tasks/delete_content_block.rake
```diff
@@ -0,0 +1,24 @@
...
+
+    @host_content_items = ContentBlockManager::HostContentItem.for_document(content_block_document)
+
+    abort("Content block `#{content_id}` cannot be deleted because it has host content. Try removing the dependencies and trying again") unless @host_content_items.items.count.zero?
+
+    Services.publishing_api.unpublish(
```
Harriethw, line 13: cool that it's this simple!

Harriethw, line 13: can't quite figure out what it does on Publishing API at first glance - wondering if it will affect this query, whether we need to add some kind of check that it hasn't been unpublished? https://github.com/alphagov/publishing-api/pull/3223/files#diff-f75e6df9ef6e0a772a32e0f688a7796a31b1a892547b49210304a1cd8a5545ed

lib/engines/content_block_manager/app/models/content_block_manager/content_block/document/soft_deletable.rb
pezholio: I liked keeping the default scope in the `SoftDeletebale` concern as it relates to things being soft deleted and communicates that better I think. Without the default scope in there, you could remove the default scope from the model and it would still look like things were being soft deleted, but all it would be doing in effect is adding a timestamp to a column.

The state stuff I did consider, but like you say, I think that might be a bit too complex for now.

lib/engines/content_block_manager/lib/tasks/delete_content_block.rake
pezholio: It sets all editions to `unpublished`, so this part of the query should cover it https://github.com/alphagov/publishing-api/blob/6254843f121efb9e17e559fe58c09daf963356e9/app/queries/get_embedded_editions_from_host_edition.rb#L10

Harriethw (approved):

PR #10081, opened 2025-03-19
Automatically remove dangerous links from published editions

ChrisBAshton:

## What and Why

If we're informed of a domain that is no longer safe, but is referenced across a number of our pages, then it is not feasible for our content designers to step in and manually fix them all in a timely manner. Instead, we need an automated approach. This commit introduces a blunt tool to find and replace all references to the dangerous domain with a `#link-removed` value.

Whilst ideally we'd have a content designer address these on a case by case basis - finding an alternative URL to link to, for example, or removing the link but adding some historical context - we're not able to do that in an automated way. The alternative is to either:

1. remove the link and link text altogether, which could lead to a confusing 'absence' of content
2. remove the link but keep the link text, which could be odd if the link text reads something like "click here"
3. replace the link with a `#`, so that it's still a link but doesn't do anything.
4. replace the link with an anchor tag that leads nowhere, i.e. `#link-removed`, so that it's still a link, it doesn't do anything, and it's clearer to the user and to the publisher what has happened.

I've opted for option 4. I've done so with minimal HTML / GovSpeak wrangling, as I can't envisage a situation where we'd even want to display the name of a dangerous domain, so finding-and-replacing only `<a>` link tags seemed an unnecessary complexity when we can simply find and replace all instances of the string.

## Architecture

This new `RemoveDangerousLinksWorker`, triggered via the Link Checker API batch report callback only if a dangerous link was detected, does the following:

1. Validates that the edition is 'published'. There's little point in fixing up superseded editions, and 'draft' editions haven't yet been published, so we're better off surfacing dangerous link warnings in the UI for the publisher to resolve it properly. 
2. Validates that there is no existing 'draft' edition. We don't want to risk overwriting the publisher's draft with an auto generated draft edition that is identical to the live edition but with dangerous links removed.
3. Validates that there are indeed 'danger' links in the link check report, that need removing.
4. Creates a draft edition, removes the dangerous links, and force publishes the edition as a minor change, with an internal note explaining what has happened.

Since https://github.com/alphagov/whitehall/pull/10077, we should be refreshing every link check report for every published edition on GOV.UK roughly every 7 days, meaning we don't have to wait too long for all dangerous links to be systematically removed. Dangerous links will be driven by the configuration of 'suspicious domains', as will be documented in https://github.com/alphagov/link-checker-api/pull/988.

There is room for improvement, but it can come later. Suggestions:

1. Check that the body of the edition changes after calling `remove_danger_links!`, before bothering to create a draft.
2. Better handle the "published with draft" state.
3. Add support for "scheduled" states.

## Screenshots / testing

With a body containing 'bad' domains [as per Link Checker API's seeds.rb](https://github.com/alphagov/link-checker-api/blob/e56e923da2a0c95ecfb78066ba230ca6d0463994/db/seeds.rb#L10):

> Screenshot 2025-03-20 at 16 39 51

I managed to set up a 'published' edition with dangerous links locally, using `Edition.find(1653798).update!(change_note: "tmp", major_change_published_at: Time.zone.now, state: "published")`:

> Screenshot 2025-03-20 at 16 49 21

After running `RemoveDangerousLinksWorker.new.perform(1653798)` locally, we can see the publication is superseded:

> Screenshot 2025-03-20 at 16 49 38

And the new edition is published and has details in the change history (NB, ignore the jump of 2 from 1653798 to 1653800. 1653799 was a draft edition I'd accidentally created earlier in my testing, and subsequently deleted before the above `RemoveDangerousLinksWorker` call):

> Screenshot 2025-03-20 at 16 50 05

And inspecting the body, the links have been replaced:

> Screenshot 2025-03-20 at 16 50 32

---

### DIFF
--- a/app/models/link_checker_api_report.rb
+++ b/app/models/link_checker_api_report.rb
@@ -26,7 +26,7 @@ NOT EXISTS (
   end
 
   def mark_report_as_completed(batch_report)
-    UpdateFromBatchReport.new(self, batch_report).update
+    UpdateFromBatchReport.new(self, batch_report).call
   end
 
   def completed?
--- a/app/models/link_checker_api_report.rb
+++ b/app/models/link_checker_api_report.rb
@@ -27,6 +27,7 @@ NOT EXISTS (
 
   def mark_report_as_completed(batch_report)
     UpdateFromBatchReport.new(self, batch_report).call
+    RemoveDangerousLinksWorker.perform_async(edition.id) if danger_links.any?
   end
 
   def completed?
--- a/app/models/link_checker_api_report/link.rb
+++ b/app/models/link_checker_api_report/link.rb
@@ -14,7 +14,7 @@ class LinkCheckerApiReport::Link < ApplicationRecord
       uri: payload.fetch("uri"),
       status: payload.fetch("status"),
       checked: payload.fetch("checked"),
-      check_dangers: payload.fetch("dangers", []),
+      check_dangers: payload.fetch("danger", []),
       check_errors: payload.fetch("errors", []),
       check_warnings: payload.fetch("warnings", []),
       problem_summary: payload.fetch("problem_summary"),
--- a/app/models/link_checker_api_report/update_from_batch_report.rb
+++ b/app/models/link_checker_api_report/update_from_batch_report.rb
@@ -4,7 +4,7 @@ class LinkCheckerApiReport::UpdateFromBatchReport
     @payload = payload
   end
 
-  def update
+  def call
     ActiveRecord::Base.transaction do
       update_report
       links = payload.fetch("links", [])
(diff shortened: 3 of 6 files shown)


app/models/link_checker_api_report.rb
```diff
@@ -26,7 +26,8 @@ def self.create_in_progress_report(batch_report, edition)
   end
 
   def mark_report_as_completed(batch_report)
-    UpdateFromBatchReport.new(self, batch_report).update
+    UpdateFromBatchReport.new(self, batch_report).call
```
lauraghiorghisor-tw, line 29: Could this be a class method?

lauraghiorghisor-tw (approved):
LGTM and @ryanb-gds

PR #10083, opened 2025-03-20
Make sure multiple select with search sends a nil value when all selections are removed

ryanb-gds:
HTTP form data is unable to represent empty arrays. If a select has a name like "test[]" and no selected value then the name of the select will be dropped from the form data when the result is submitted. However, by creating a hidden input with the same name we can make sure that the form data will look like 'test[0]=""' instead of 'test=[]'. If the select has a value, then that will override the nil value.

Trello: https://trello.com/c/mZqawRIP

### DIFF
--- a/app/views/components/_select_with_search.html.erb
+++ b/app/views/components/_select_with_search.html.erb
@@ -27,6 +27,10 @@
     } %>
   <% end %>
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Nice fix 🎉 Apologies this slipped through!

PR #10084, opened 2025-03-20
Allow deletion of people without current role appointments or former ministerial roles

ryanb-gds:
We occasionally get duplicates of the same person existing where people have held various board member roles in different organisations, etc.

These people are not linked to on any govuk pages because organisation pages only show current appointment holders, and only ministerial roles have their own pages showing historical office holders.

It therefore seems reasonable to allow the deletion of these "orphaned" people, so that they are no longer present on GOV.UK and also do not appear in the Whitehall people list.

This will enable the user who has reported https://govuk.zendesk.com/agent/tickets/6068030 to at least delete the duplicate people. We may want to consider allow redirection is there is a valid use case.

Trello: https://trello.com/c/zzmAwGPV

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -62,7 +62,7 @@ class Person < ApplicationRecord
   end
 
   def destroyable?
-    role_appointments.empty?
+    current_role_appointments.empty? && ministerial_roles.empty?
   end
 
   def name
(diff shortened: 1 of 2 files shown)


ryanb-gds:
Closing this as according to JP on Slack there are plans to reintroduce a "previous roles in government" page for people with non-ministerial roles.

PR #10086, opened 2025-03-21
Run cucumber tests in parallel

pezholio:
This uses the same approach as minitest.yml to split the Cucumber tests into groups and run them in parallel across multiple containers, which should hopefully speed up the CI. 

The cucumber tests were already running in a matrix (I think as an artefact of there being two suites at one point), so it’s pretty trivial to split the tests up.

From an unscientific survey, this cuts the runtime down from ~7 minutes to ~3 minutes


ChrisBAshton (approved):
Amazing! 👏

PR #10091, opened 2025-03-24
Fix logic loop whereby docs can be continuously republished

ChrisBAshton:
Without this commit, the RemoveDangerousLinksWorker can get into a state where it continually editions and publishes the same content. Consider:

1. Edition ID `123` - ‘published’ - has danger links
2. The first worker creates a draft (124) and removes the links
3. ...which triggers a new link check report, sending the payload of ‘dangerous’ links to Link Checker APi
4. The first worker publishes the draft 124 - we want to finish here.
5. Link Checker API makes a callback to Whitehall saying edition 124 has dangerous links. Whitehall kicks off a new (second) worker, which creates a new draft and attempts to removes the links, but there’s nothing to do, so it silently succeeds
6. ...meanwhile, it may itself have triggered another worker. And so on.

This commit fixes things by checking whether the edition has changed at all as a result of the sanitization. The edition is only turned into a draft and published anew if there is a change to be made.

---

Test instructions:

```
Edition.find(1653805).create_draft(User.find_by(name: "Scheduled Publishing Robot", uid: nil)).update!(body: "[malicious link](https://malicious.example.com)", minor_change: t
rue); Edition.last.update!(change_note: "tmp", state: "published")
```

### DIFF
--- a/app/models/link_checker_api_report.rb
+++ b/app/models/link_checker_api_report.rb
@@ -27,7 +27,7 @@ NOT EXISTS (
 
   def mark_report_as_completed(batch_report)
     UpdateFromBatchReport.new(self, batch_report).call
-    # RemoveDangerousLinksWorker.perform_async(edition.id) if danger_links.any?
+    RemoveDangerousLinksWorker.perform_async(edition.id) if danger_links.any?
   end
 
   def completed?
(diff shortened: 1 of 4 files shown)


GDSNewt (approved):
LGTM

PR #10103, opened 2025-03-26
Remove Attachment deletion rake task

minhngocd:
In light of our current understanding of asset management between Whitehall and Asset Manager, this rake task is no longer necessary. If there is a requirement to remove an HTML attachment, we use `redirect_html_attachments` rake task. If there is a requirement to remove a file attachment, there are instructions to remove them directly from asset manager. The deletion state of attachments in Whitehall is more about the link between Attachments and its corresponding Edition, which means users should be able to break that link within the UI.

https://trello.com/c/UTDPROlf/3576-consolidate-existing-asset-diagrams


lauraghiorghisor-tw (approved):
I agree with the approach of using AM's rake task as per the dev docs ✅

PR #10122, opened 2025-04-02
Add JavaScript to populate `ga4-index-section` client-side

gtvj:
This PR proposes two things: 

1. introducing a JavaScript module that will add the `data-ga4-index-section` attribute to `select` elements and relevant `input` elements client-side. More information is provided in the commit message (See 0e5ea5a4aff3a9efff8a5f88f8bbea1dcba910ff)
2. removing the existing server-side code that had added the `ga4-index-section` to `select` elements in the `select-with-search` component (introduced in a recent PR[^1])

The approach taken in this PR comes from a suggestion[^2] made in a relevant PR. The subsequent conversation revealed it's difficult to know what the correct value for `data-ga4-index-section` would be for some includes.

## Grouped fields

Because PAs have not yet confirmed if grouped inputs and selects (such as dates and times) should share an index, I've omitted this functionality from the PR. We can add it later if needed.

[^1]: See https://github.com/alphagov/whitehall/pull/9983
[^2]: See https://github.com/alphagov/whitehall/pull/10069#discussion_r2022600260

### DIFF
--- a/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
@@ -1,0 +1,23 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
+;(function (Modules) {
+  Modules.Ga4IndexSectionSetup = {
+    init: function () {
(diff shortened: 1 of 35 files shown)


minhngocd (approved):
Looks good to me - had a question around if the module was initialised multiple times if it was included in two or more elements but that looks like it should still yield the same results.

gtvj:
> Looks good to me - had a question around if the module was initialised multiple times if it was included in two or more elements but that looks like it should still yield the same results.

Thanks @minhngocd - that's a good point. The initialisation approach would restart the indexing each time it's initialised but, because the analytics data attributes are applied only in the layout this shouldn't happen.

PR #10128, opened 2025-04-04
Add `search_results` tracking

patrickpatrickpatrick:
## What
Use `Ga4EcommerceTracker` from [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js) to track when a user views a search results page.

### More details 
- Add required data attributes to elements on identified search results pages for `Ga4EcommerceTracker` to function
- Implement `Ga4SearchResultsSetup` which runs `Ga4EcommerceTracker.init` on page load unless search result is within tab (in which case it only runs `Ga4EcommerceTracker.init` if the tab containing the search result is visible)

**NB** Search results should not be within tabs, if all search results are moved out of tab components then we can simply this code significantly

## Why

Part of adding GA4 Tracking to Whitehall.

Trello: https://trello.com/c/4BesXrbX/3622-search-results-page-tracking-on-whitehall-app

---

### DIFF
--- a/app/assets/javascripts/admin/analytics-modules/ga4-search-results-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-search-results-setup.js
@@ -1,0 +1,104 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
+;(function (Modules) {
+  Modules.Ga4SearchResultsSetup = {
+    init: function (noHashChange) {
+      if (!document.querySelector('[data-ga4-ecommerce]')) {
+        return
+      }
+
+      // if we have multiple tabs with search results
+      // then remove the attributes so that we only capture
+      // the search results in the visible tab
+      Array.from(
+        document.querySelectorAll(
+          '[data-module~="govuk-tabs"] [data-ga4-ecommerce]'
+        )
+      ).forEach((search) => {
+        search.removeAttribute('data-ga4-ecommerce')
+        search.setAttribute('data-search-track', true)
+      })
+
+      this.trackEcommerce()
+
+      if (!noHashChange) {
+        this.eventListener = this.trackEcommerce.bind(this)
+
+        window.addEventListener('hashchange', this.eventListener)
+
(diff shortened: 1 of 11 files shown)


spec/javascripts/admin/analytics-modules/ga4-ecommerce-setup.spec.js
```diff
@@ -0,0 +1,128 @@
...
+  container.id = 'container'
+  document.body.appendChild(container)
+
+  describe('if search outside tabs component', function () {
+    it('start Ga4EcommerceTracker', function () {
+      container.innerHTML = `
+        <div data-ga4-ecommerce></div>
+      `
+
+      const mockGa4SendData = spyOn(
```
ryanb-gds, line 12: Should we rename this variable, as we're mocking the `init` method rather than `sendData`?

app/assets/javascripts/admin/analytics-modules/ga4-ecommerce-setup.js
```diff
@@ -0,0 +1,104 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
+;(function (Modules) {
+  Modules.Ga4EcommerceSetup = {
```
ryanb-gds, line 7: Would it be more meaningful to Whitehall devs to call this module `Ga4SearchSetup` or something along those lines? The fact that it uses the ecommerce tracking under the hood should be an implementation detail obscured by the module, I feel

ryanb-gds (approved):
Couple of naming suggestions for your consideration, but I wouldn't insist on them.

I see now what a headache having search within the tabs component is 😅 

Cool that the publishing component ecommerce tracker obviously meets our needs well, doesn't seem like we've had to add much code once we've taken care of the tab situation

app/assets/javascripts/admin/analytics-modules/ga4-ecommerce-setup.js
patrickpatrickpatrick: I'm going to call it `Ga4SearchResultsSetup` since it's a `search_results` event and a `search` event is used when a search actually executed.

PR #10129, opened 2025-04-04
Consolidate search partials ahead of adding analytics to search pages

patrickpatrickpatrick:
## What

Reduce duplicated mark-up in `admin/editions/index.html.erb`, `admin/scheduled_announcements/index.html.erb` and `admin/shared/_featurable_editions.html.erb`.

### More details
- Move `admin/editions/filter_options` to `admin/shared/filter_options`
     - Use `admin/shared/filter_options` in `app/views/admin/shared/_filtered_view.html.erb`
- Create `app/views/admin/shared/_filtered_view.html.erb`
     - Use `filtered_view` in `admin/editions/index.html.erb`, `admin/scheduled_announcements/index.html.erb` and `admin/shared/_featurable_editions.html.erb`

## Why

We want to add analytics to search pages. The first barrier of this task was finding out where search pages existed because of duplicated code in different partials. I noted that `admin/editions/filter_options` had already been implemented as generic partial to be used in multiple places already. Therefore it didn't require a significant change to be shared in different views. Creating `app/views/admin/shared/_filtered_view.html.erb` means we can now add analytics to one partial instead of three partials.

There are further steps we could take to refine the partials further but would be more time consuming to implement and to test:
- `admin/editions/filter_options` could be a ViewComponent
- filterable attributes could be specified in the `MODELNAME_filter` or a presenter
- Move logic for building the content of each cell to a helper or presenter for each model
- Other pages that use `govuk_publishing_components/table` with `filter: true` could also use  `app/views/admin/shared/_filtered_view.html.erb`
- Adding a document to Collection could use the same search as other instances of filtering editions

Trello: https://trello.com/c/4BesXrbX/3622-search-results-page-tracking-on-whitehall-app

### DIFF
--- a/app/views/admin/shared/_filter_options.html.erb
+++ b/app/views/admin/shared/_filter_options.html.erb
@@ -43,20 +43,6 @@
       } %>
     <% end %>
 
-    <% if filter_by.include?(:organisation_id) %>
-      <%= render "components/select_with_search", {
-        label: "Organisation",
-        id: "organisation_filter",
-        name: "organisation_id",
-        heading_size: "s",
-        ga_data: {
-          document_type: "#{action_name}-#{controller_name}",
-          section: "Organisation",
-        },
-        grouped_options: admin_organisation_filter_options(@filter.options[:organisation_id]),
-    } %>
-    <% end %>
-
     <% if filter_by.include?(:organisation) %>
       <%= render "components/select_with_search", {
         label: "Organisation",
--- a/app/views/admin/shared/_filter_options.html.erb
+++ b/app/views/admin/shared/_filter_options.html.erb
@@ -4,7 +4,7 @@
   raise "filter action required" unless defined?(filter_action)
 %>
 
-<div class="app-view-filter govuk-!-margin-right-5 govuk-!-padding-5">
+<div class="app-view-filter govuk-!-padding-5">
   <%= form_with url: filter_action + anchor, method: :get do |form| %>
     <%= render "govuk_publishing_components/components/heading", {
       text: "Filter by",
@@ -41,6 +41,20 @@
           }
         end,
       } %>
+    <% end %>
+
+    <% if filter_by.include?(:organisation_id) %>
+      <%= render "components/select_with_search", {
+        label: "Organisation",
+        id: "organisation_filter",
+        name: "organisation_id",
+        heading_size: "s",
+        ga_data: {
+          document_type: "#{action_name}-#{controller_name}",
+          section: "Organisation",
+        },
+        grouped_options: admin_organisation_filter_options(@filter.options[:organisation_id]),
+    } %>
     <% end %>
 
     <% if filter_by.include?(:organisation) %>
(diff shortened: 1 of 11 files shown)


app/views/admin/shared/_filter_options.html.erb
```diff
@@ -43,6 +43,20 @@
       } %>
     <% end %>
 
+    <% if filter_by.include?(:organisation_id) %>
```
minhngocd, line 46: What's the difference between this one and the filter_by organisation below?

minhngocd (commented):
My gut feel going through this, is that we might hit some difficulty in maintaining this if we try and generalise 3 different search views into one. They exist on different pages and might require different tweaks (e.g. adding a new filter to Statistics Announcement but not Featureable documents search or full document search. 

In that scenario, we'd then have to work in additional conditionals and ifs into this shared search view that could potentially grow out of proportion, and have overlapping conditionals which we already see with the filter by Organisation ID vs filter by Organisastion. 

I feel like maybe we would want all the filters to be small view components, and then we can construct a Document Search Filter component, Statistics Announcement Search Filter component, or Featured Document Search Filter component, which contains a different set of search filters each. However, that would be a load more effort to do as you've mentioned in the PR description. 

For the purpose of this story, I was wondering, if it would be a good first step for the above to extract these 3 search filters into their own view components for now, without removing any of the duplication, and then have the 3 view components include like a Search Analytics component or something?

app/views/admin/shared/_filter_options.html.erb
patrickpatrickpatrick: I had copied it over from the original `StatisticsAnnouncement` filter options. Changing it to `organisation` doesn't seem to impact functionality, not sure why it was referred to as `organisation_id` before in the controller and the filter model.

patrickpatrickpatrick:
> My gut feel going through this, is that we might hit some difficulty in maintaining this if we try and generalise 3 different search views into one. They exist on different pages and might require different tweaks (e.g. adding a new filter to Statistics Announcement but not Featureable documents search or full document search.
> 
> In that scenario, we'd then have to work in additional conditionals and ifs into this shared search view that could potentially grow out of proportion, and have overlapping conditionals which we already see with the filter by Organisation ID vs filter by Organisastion.
> 
> I feel like maybe we would want all the filters to be small view components, and then we can construct a Document Search Filter component, Statistics Announcement Search Filter component, or Featured Document Search Filter component, which contains a different set of search filters each. However, that would be a load more effort to do as you've mentioned in the PR description.
> 
> For the purpose of this story, I was wondering, if it would be a good first step for the above to extract these 3 search filters into their own view components for now, without removing any of the duplication, and then have the 3 view components include like a Search Analytics component or something?

Ah, I was thinking along of the lines of having one `Filter` component because the individual filters are self contained components already. Then the filterable entity would specify the attributes and the possible values of the attributes to be filtered upon. This would then be rendered by the `Filter` component which would call upon a function on the entity, on a presenter of the entity or on a helper of the entity to get the rows. I thought that this change could happen when we need to add additional filters to one of the filterable entities, since it would require adding tests to make sure it works as expected.

PR #10134, opened 2025-04-07
Make 'dangerous' links unclickable in the Link Check Report UI

ChrisBAshton:
For broken and warning links, we create a clickable link in the sidebar so that editors can click on the links and reproduce any of the reported issues.

For 'dangerous' links that may have been taken over by a malicious third party and could be distributing malware etc, we should take a more cautious approach and just render the link as text, so that editors don't accidentally click on a dangerous link.

Before:
> clickable link

After:
> plain text

Trello: https://trello.com/c/PLcGoXXf

---

### DIFF
--- a/app/views/admin/link_check_reports/_link_check_report.html.erb
+++ b/app/views/admin/link_check_reports/_link_check_report.html.erb
@@ -38,7 +38,11 @@
           <ul class="govuk-list app-view-summary__broken-links-report">
             <% links.each do |link| %>
               <li>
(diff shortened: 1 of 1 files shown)


minhngocd (approved):
ohhh makes sense - nice! is there a test for this?

ChrisBAshton:
> ohhh makes sense - nice! is there a test for this?

No - I did look in `whitehall/test/components/admin/editions`, which is where a test _should_ be. But we lack test coverage for the EditionsController#show view in general. Would be good to add some later 🤞

PR #10136, opened 2025-04-08
add Content Modelling team as code owners of Content Block Manager

Harriethw:
https://trello.com/c/7QlOgZ3F/991-add-cbm-team-to-apps-list-in-dev-docs

This should ensure that the team gets tagged in any PRs for the engine, and provides some more
documentation for any devs looking to understand the code.

This was suggested [as we want to ensure that ownership of this part of the repo is clearly defined ](https://github.com/alphagov/whitehall/pull/10136)(as we are currently a separate team to Whitehall)
 
---

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+lib/engines/content_block_manager @alphagov/gov-uk-content-modelling


pezholio:
This is cool! I wonder if we can alter the template based on what code gets touched?

pezholio (approved):

Harriethw:
@pezholio ooh that's a good idea! From a quick look it looks like there is no way to set a different template based on path/codeowners, [a custom template has to be picked manually by query params](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) 🤔 I guess if we had our own template we could select it from the github cli (which is what I use to create PRs, but I'm not sure if most people do) but not sure it's worth having as if someone knows to choose that template then they'll know it's for content modelling...

PR #10149, opened 2025-04-16
(1014) Enable line breaks in instructions to publishers

pezholio:
Trello card: https://trello.com/c/ggVAML1f/1014-enable-line-breaks-in-instructions-to-publishers

This allows instructions to publishers to be lightly formatted, turning URLs into links and replacing line breaks with HTML `<br />` tags.

### Before

image

image

### After

image

image


Harriethw (approved):
This doesn't add the line breaks into the summary list cards on the home page, but I don't think we should have them there as that could really lengthen the page? we probably want to revisit the list page design and e.g. have a cut off point if the instructions are over a certain character count

PR #10157, opened 2025-04-23
Set HTML attachment locale to the same as the attachable's primary locale if the primary locale is not English

ryanb-gds:
This fixes a bug with foreign language only consultations and calls for evidence which results in non-English HTML attachments rendering on GOV.UK with English page furniture.

If the locale isn't set for an HTML attachment, Whitehall sends English to the Publishing API for the locale value as the default.

This isn't a great solution, because if we subsequently choose to have translated documents that have a primary locale other than English, then this will cause problems. Ideally we would check if the attachable was a "foreign language only" content item and only render the hidden field if that is the case. However, because we don't persist the information about whether an attachable is foreign language only, the only way to know when to set the attachment locale is to check if the attachable has a non-English primary locale. I did try to use the `locale_can_be_changed?` method in the condition, but this doesn't work for consultations/calls for evidence that aren't foreign language only. In that case we don't want to override the attachment locale, because it will stop the attachment appearing on the English translation.

Trello: https://trello.com/c/q2C39WwA

### DIFF
--- a/app/views/admin/attachments/_reference_fields.html.erb
+++ b/app/views/admin/attachments/_reference_fields.html.erb
@@ -22,6 +22,10 @@
   </div>
 <% end %>
 
+<% if attachable.respond_to?(:primary_locale) && attachable.primary_locale != :en %>
+  <%= form.hidden_field :locale, value: attachable.primary_locale %>
+<% end %>
+
 <div class="govuk-!-margin-bottom-8">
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
We had an offline chat: https://gds.slack.com/archives/C02L13S214K/p1745489573902839?thread_ts=1745403301.412719&cid=C02L13S214K

LGTM

PR #10161, opened 2025-04-24
Use `Ga4FinderTracker` to track search form interactions

patrickpatrickpatrick:
## What

Use `Ga4FinderTracker` from `govuk_publishing_components` to track search form interactions on `whitehall`.

### More details
- Add `Ga4FinderSetup` to add change tracking to search forms and additional tracking for multiple select
- Change `Ga4IndexSetup` to make the parent `fieldset` of a radio input the index (a group radios will change one value of a form, so generally each radio isn't counted a separate input)
- Add necessary data attributes to search forms and inputs
- Change `Ga4SelectTracker` and `Ga4SelectWithSearchTracker` to not double track on finder forms

## Why

Part of the work to add GA4 tracking in `whitehall`. 

Addresses the following trello cards:

https://trello.com/c/BoARWzeD/3671-use-ga4-finder-tracker-in-whitehall
https://trello.com/c/lVFJZk1M/3715-track-reset-all-fields-button-on-select-with-search-components

Partly addresses the following trello cards:
https://trello.com/c/RAs3Yyss/3616-component-level-tracking-track-radio-button-interactions-on-whitehall
https://trello.com/c/EbHEqYtk/3627-modify-component-type-tracking-parameters
https://trello.com/c/zEY8eHbQ/803-component-level-tracking-track-checkbox-interactions-on-whitehall

### DIFF
--- a/app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
@@ -1,0 +1,54 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
+;(function (Modules) {
+  Modules.analyticsModules = Modules.analyticsModules || {}
+
+  Modules.analyticsModules.Ga4FinderSetup = {
+    init: function () {
+      const finders = Array.from(
+        document.querySelectorAll("[data-module~='ga4-finder-tracker']")
+      )
+
+      if (finders) {
+        finders.forEach((finder) => {
+          finder.addEventListener('change', this.onFinderChange)
+        })
+
+        Modules.Ga4FinderTracker = Modules.Ga4FinderTracker || {}
+        Modules.Ga4FinderTracker.extraSupportedElements = {
+          // for select with search with multiple choice
+          'select-multiple': function (eventTarget, event) {
+            const eventValue = event.detail.value
+            const elementValue = eventTarget.querySelector(
+              `option[value="${eventValue}"]`
+            ).text
+            const selectedEventOption = eventTarget.querySelector(
+              `option[value="${eventValue}"]:checked`
+            )
+
+            return {
+              elementValue,
+              wasFilterRemoved: !selectedEventOption
+            }
+          }
+        }
+      }
+    },
+
+    onFinderChange: function (event) {
+      let ga4ChangeCategory = event.target.closest('[data-ga4-change-category]')
+      if (ga4ChangeCategory) {
+        ga4ChangeCategory = ga4ChangeCategory.getAttribute(
+          'data-ga4-change-category'
+        )
+        window.GOVUK.analyticsGa4.Ga4FinderTracker.trackChangeEvent(
+          event,
+          ga4ChangeCategory
+        )
+      }
+    }
+  }
+})(window.GOVUK.analyticsGa4)
--- a/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
@@ -11,11 +11,39 @@ window.GOVUK.analyticsGa4.analyticsModules =
       )
 
       moduleElements.forEach(function (moduleElement) {
+        // An indexed section is either a single element
+        // or set of elements that changes one value of
+        // a form. Therefore we don't want to index:
+        //
+        // - individual radio buttons because they change
+        //   one value of a single value
+        // - individual checkboxes if they're part of a
+        //   group of checkboxes that change a single value
+        //
+        // Additionally we don't want to index:
+        // - the search within a SelectWithSearch component as it doesn't
+        //   change the value of the form
+        // - a hidden input because the user can't interact with it
         const indexedElements = moduleElement.querySelectorAll(
-          'select, input:not([data-module~="select-with-search"] input)'
+          'select, input:not([data-module~="select-with-search"] input):not([type="radio"]):not([type="hidden"]), fieldset'
         )
+
         indexedElements.forEach((element, index) => {
-          element.dataset.ga4IndexSection = index
+          if (element.tagName === 'FIELDSET' || !element.closest('fieldset')) {
+            const indexData = {
+              index_section: index,
+              index_section_count: indexedElements.length
+            }
+            element.dataset.ga4Index = JSON.stringify(indexData)
+            element.dataset.ga4IndexSection = index
+
+            if (element.closest('[data-module~="ga4-finder-tracker"]')) {
+              // required attribute for `ga4-finder-tracker`
+              // assumes that index values will come from
+              // an element with `ga4-filter-parent` set
+              element.dataset.ga4FilterParent = true
+            }
+          }
         })
       })
     }
--- a/app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
@@ -12,6 +12,8 @@ window.GOVUK.analyticsGa4.analyticsModules =
 
       moduleElements.forEach(function (moduleElement) {
         moduleElement.addEventListener('change', function (event) {
+          if (event.target.dataset.ga4ChangeCategory) return
+
           const ga4DocumentType = moduleElement.dataset.ga4DocumentType
 
           if (
--- a/app/views/admin/document_collection_group_document_search/add_by_title.html.erb
+++ b/app/views/admin/document_collection_group_document_search/add_by_title.html.erb
@@ -10,7 +10,7 @@
 
 <div class="govuk-grid-row" data-ga4-ecommerce data-ga4-search-query="<%= params[:title] %>" data-ga4-list-title="<%= yield(:page_title) %>" data-ga4-ecommerce-start-index="<%= page_start_index(params[:page]) %>">
   <div class="govuk-grid-column-two-thirds app-view-document-collection-document-search-bar" data-ga4-search-section="Search by title">
-    <%= form_with method: :get do |_form| %>
+    <%= form_with method: :get, data: { module: "ga4-finder-tracker" } do |_form| %>
       <%= render "govuk_publishing_components/components/search", {
         name: "title",
         value: params[:title],
--- a/spec/javascripts/admin/analytics-modules/ga4-finder-setup.spec.js
+++ b/spec/javascripts/admin/analytics-modules/ga4-finder-setup.spec.js
@@ -1,0 +1,119 @@
+describe('GOVUK.analyticsGa4.analyticsModules.Ga4FinderSetup', function () {
+  const container = document.createElement('div')
+  const form = document.createElement('form')
+  form.dataset.module = 'ga4-finder-tracker'
+
+  const text = document.createElement('input')
+  text.type = 'text'
+
+  const date = document.createElement('div')
+  date.className = 'govuk-date-input'
+  date.innerHTML = `
+    <div class="govuk-date-input__item">
+        <div class="govuk-form-group">
+            <input class="gem-c-input govuk-input govuk-input--width-4" name="day" type="text" data-ga4-change-category="update-filter date">
+        </div>
+    </div>
+    <div class="govuk-date-input__item">
+        <div class="govuk-form-group">
+            <input class="gem-c-input govuk-input govuk-input--width-4" name="month" type="text" data-ga4-change-category="update-filter date">
+        </div>
+    </div>
+    <div class="govuk-date-input__item">
+        <div class="govuk-form-group">
+            <input class="gem-c-input govuk-input govuk-input--width-4" name="year" type="text" data-ga4-change-category="update-filter date">
+        </div>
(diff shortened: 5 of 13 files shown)


patrickpatrickpatrick:
> It might be a bit late in the day for this, but I'm a bit dubious about using the finder tracker to track all forms. It absolutely makes sense for search pages because the use case is similar enough, but for general forms, although the finder tracker might work, we're pushing it far outside of its intended use case. I think that's borne out by the amount of workarounds we've had to put in place in the finder setup module to get it to meet our needs.
> 
> It is a shame that the form tracker from publishing components doesn't support multiple inputs in a form. I'd strongly consider either extending it or building our own. Alternatively, we could continue down the path of having separate modules for each input type, which I think is a better option than doing this. I'd rather maintain a greater amount of simpler code than a lesser amount of more complex code. This is excessively DRY to my mind.
> 
> Anyway, that's my two cents. Would be worth seeing what the rest of the team reckons.

I’m interested in what aspects of the setup you think are workarounds, as there’s not anything that I would specifically change if we only tracked search forms. Happy to talk through this next week 🙂

app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
```diff
@@ -12,10 +12,21 @@ window.GOVUK.analyticsGa4.analyticsModules =
...
         )
+
         indexedElements.forEach((element, index) => {
-          element.dataset.ga4IndexSection = index
+          if (element.tagName === 'FIELDSET' || !element.closest('fieldset')) {
+            const indexData = {
+              // cast to string otherwise treated as false by
+              // `analyticsGa4.core.applySchemaAndSendData`
+              index_section: String(index),
+              index_section_count: String(indexedElements.length)
```
ChrisBAshton, line 24: Is that comment true? As best I can tell, we're only doing [explicit checks for `undefined` and `null`](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.js#L58-L64) in the corresponding code. Feels as though any casting should be done in the core module rather than in the consumer.

app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
```diff
@@ -0,0 +1,124 @@
...
+  const excludedTypes = ['search', 'hidden']
+
+  Modules.analyticsModules.Ga4FinderSetup = {
+    init: function () {
+      const finders = Array.from(
+        document.querySelectorAll("[data-module~='ga4-finder-tracker']")
+      )
+
+      if (finders) {
+        finders.forEach((finder) => {
```
ChrisBAshton, line 18: It's a bit of a smell that we're still referring to this as a finder tracker, now that we're trying to use it to track... non-finders 😬 When you proposed moving the component back into govuk_publishing_components I think I'd imagined it would be renamed at the same time, e.g. to something like `ga4-form-tracker` - is there a reason we can't/shouldn't do that?

And even if we're stuck with what it's called in govuk_publishing_components, perhaps we can use the more generic name here in Whitehall, and only map it to 'finders' in one place, e.g.

```javascript
// stuck with the name for legacy reasons - we're tracking other forms, not finders
component_name = "ga4-finder-tracker"
module_name = "Ga4FinderSetup"

Modules.analyticsModules[module_name] = {
  init: function () {
      const finders = Array.from(
        document.querySelectorAll("[data-module~='${component_name}']")
        // ...etc
      )
```

app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
```diff
@@ -0,0 +1,124 @@
+'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules =
+  window.GOVUK.analyticsGa4.analyticsModules || {}
+;(function (Modules) {
+  Modules.analyticsModules = Modules.analyticsModules || {}
+
+  const excludedTypes = ['search', 'hidden']
```
ChrisBAshton, line 9: Why is type 'search' excluded here - are we tracking it some other way?
May be useful to put an inline comment here.

app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
```diff
@@ -0,0 +1,124 @@
...
+              changeCategoryType = changeCategoryType || el.type
+
+              if (el.hasAttribute('multiple')) {
+                changeCategoryType = `${changeCategoryType}-multiple`
+              }
+
+              el.setAttribute(
+                'data-ga4-change-category',
+                `update-filter ${changeCategoryType}`
+              )
```
ChrisBAshton, line 50: It's quite difficult to follow the procedural style code above - is there any way it could be broken down into well named methods that are called out to instead? E.g.

```javascript
init: function () {
  finders().forEach(finder => {
    inputs(finder).forEach(input => {
      changeCategoryType = figureOutTheChangeType(input)
      if (changeCategoryType) {
        input.setAttribute(
          'data-ga4-change-category',
        `update-filter ${changeCategoryType}`
        )
    })
  })
```

app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
```diff
@@ -0,0 +1,124 @@
...
+              JSON.stringify({
+                action: 'remove',
+                event_name: 'select_content',
+                text: resetLink.innerText,
+                type:
+                  (filterTypeContainer &&
+                    filterTypeContainer.getAttribute('data-ga4-filter-type')) ||
+                  undefined
+              })
+            )
```
ChrisBAshton, line 103: This should be its own method too

spec/javascripts/admin/analytics-modules/ga4-finder-setup.spec.js
```diff
@@ -0,0 +1,251 @@
...
+
+      if (field.tagName === 'SELECT') {
+        field.querySelectorAll('option')[1].selected = true
+        field.dispatchEvent(new Event('change', { bubbles: true }))
+      } else {
+        field.click()
+        field.value = '123'
+        field.dispatchEvent(new Event('change', { bubbles: true }))
+      }
+    }
```
ChrisBAshton, line 113: I'd move this inside the ` it('calls `trackChangeEvent` on tracked inputs', ` block, as it's only being used there. Better to keep it all together so that it's clear there aren't multiple tests relying on this setup code.

spec/javascripts/admin/analytics-modules/ga4-finder-setup.spec.js
```diff
@@ -0,0 +1,251 @@
...
+          field.dataset.ga4ChangeCategory
+        )
+
+        form.innerHTML = ''
+      })
+    })
+  })
+
+  describe('updates dataset correctly', () => {
+    it('for a clear all fields link', () => {
```
ChrisBAshton, line 136: I think I'd reverse the concepts here so that it reads better:

```js
describe('a clear all fields link', () => {
  it('updates the dataset correctly', () => {
```

app/views/admin/document_collection_group_document_search/add_by_title.html.erb
ChrisBAshton, line 1: For the "Ga4FinderTracker HTML data attribute changes" commit, I'd probably break the bullets down into their own commits actually - so one each for:

- `data-ga4-section` on the closest parent to an element
- `data-ga4-change-category` on the date inputs within
edition filter options
- `data-ga4-change-category` on "Reset all fields" links
- `data-module="ga4-finder-tracker"` on finder forms"

app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
```diff
@@ -16,7 +16,8 @@ window.GOVUK.analyticsGa4.analyticsModules =
 
           if (
             event.target.tagName.toLowerCase() === 'select' &&
-            !event.target.getAttribute('hidden')
+            !event.target.getAttribute('hidden') &&
+            !event.target.closest("[data-module~='ga4-finder-tracker']")
```
ChrisBAshton, line 20: All for avoiding double-tracking, but do you have a plan for consolidating _all_ tracking onto the ga4-finder-tracker approach? It feels like a rabbit hole if we're sometimes having to disable tracking like this and sometimes not. I'd have thought / hoped that pulling in ga4-finder-tracker would give us a tracking module / wrapper that sufficiently encompasses all basic components (i.e. 'select'), if not the more complex ones ('select with search')

ChrisBAshton (commented):
Thanks @patrickpatrickpatrick , and good to get a fresh pair of eyes on the problem - I'm keen to tackle the GA4 challenge with as little overhead as possible.

I do share some of Ryan's concern that we're introducing rather a lot of complexity in this PR, though I'm not clear what the alternative would be. If you were to raise a PR to apply the same tracking _without_ making use of the ga4-finder-tracker component, what would that look like? How many lines of code?

I'm generally in favour of a tracking component that's smart enough to do everything, but from looking at some of the changes here, we're still having to do some manual tracking in other ways, and now have the complexity of having to opt out in some cases where there is a finder-tracker 'wrapper'. The state of GA4 tracking in Whitehall may be clear if you're deep in the weeds of GA4 but it's not hugely clear to me at the moment from an outsider's perspective 👀

app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
patrickpatrickpatrick: Interesting, I remember this behaviour happening but actually when I try it now it doesn't seem to do that. It might have been me getting confused with something else or another index property overwriting it. Can remove this casting for sure.

app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
patrickpatrickpatrick: So there's maybe some confusion here, `Ga4FinderTracker` now does not track all forms in `whitehall`. `data-ga4-finder-tracker` is only applied to specific search forms.

app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
patrickpatrickpatrick: My plan is to remove the select tracking entirely when I track finders and [form changes](https://github.com/alphagov/whitehall/pull/10201/files).

app/assets/javascripts/admin/analytics-modules/ga4-finder-setup.js
patrickpatrickpatrick: I'm now working on [extending the functionality of Ga4FormTracker](https://github.com/alphagov/whitehall/pull/10201/files) in govuk_publishing_components to track form changes as well as form submit (which is the default behaviour within the gem). My intention is that when other publishing applications need to use the change tracking of the form tracker, I'll move it to the gem.

spec/javascripts/admin/analytics-modules/ga4-finder-setup.spec.js
ChrisBAshton: All of them 👍

patrickpatrickpatrick: Ah OK, in that case I probably wouldn't use an `it(updates the dataset correctly)` then. I'd rewrite them to be a `describe('THE INPUT TYPE updates the dataset correctly')` since there isn't a reverse clause. Would that be correct?

ChrisBAshton: You may have to go straight for an `it` - I don't think you can write assertions inside a `describe` block. You're right though, we may not need to nest two blocks.

ryanb-gds (approved):
This is looking great, just one question about whether we are scoping our application of the `ga4FilterParent` attribute appropriately. Would it be feasible to put that into the view markup as well if appropriate? If not then I think this is ready to go, so approving on that basis

app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
```diff
@@ -12,10 +12,19 @@ window.GOVUK.analyticsGa4.analyticsModules =
 
       moduleElements.forEach(function (moduleElement) {
         const indexedElements = moduleElement.querySelectorAll(
-          'select, input:not([data-module~="select-with-search"] input)'
+          'select, input:not([data-module~="select-with-search"] input):not([type="radio"]):not([type="hidden"]), fieldset'
```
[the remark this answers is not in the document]
patrickpatrickpatrick: Sure!

app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
```diff
@@ -12,10 +12,19 @@ window.GOVUK.analyticsGa4.analyticsModules =
...
+
         indexedElements.forEach((element, index) => {
-          element.dataset.ga4IndexSection = index
+          if (element.tagName === 'FIELDSET' || !element.closest('fieldset')) {
+            const indexData = {
+              index_section: index,
+              index_section_count: indexedElements.length
+            }
+            element.dataset.ga4Index = JSON.stringify(indexData)
+            element.dataset.ga4FilterParent = true
```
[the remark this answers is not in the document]
patrickpatrickpatrick: Yeah, that makes sense.

PR #10169, opened 2025-04-29
Social media account editor permissions

ryanb-gds:
This PR ensures that users which only have the basic Whitehall permissions are unable to create, update or delete social media accounts.

We're looking to provide a basic level of authorisation for unprotected resources in Whitehall, ahead of a planned overhaul of the authorisation system.

I have tried as far as possible to remain consistent with existing authorisation implementations within Whitehall.

Trello: https://trello.com/c/RevqdALn


ChrisBAshton (approved):
Nice one ⭐

PR #10173, opened 2025-04-30
search tracking ga4

patrickpatrickpatrick:
- **add js for search tracking**
- **add data attributes to search**


app/assets/javascripts/admin/analytics-modules/ga4-search-setup.js
```diff
@@ -0,0 +1,44 @@
...
+      const searchFormInputs = document.querySelectorAll('[type="search"]')
+
+      searchFormInputs.forEach((searchFormInput) => {
+        const searchForm = searchFormInput.closest('form')
+
+        if (searchForm) {
+          let ga4DocumentType = searchForm.closest('[data-ga4-document-type]')
+
+          if (ga4DocumentType) {
+            ga4DocumentType = ga4DocumentType.dataset.ga4DocumentType
```
ryanb-gds, line 18: I'm a general fan of immutable variables and static types, and try to code that way even when the language I'm working in doesn't apply such constraints. More of a personal thing than any standard we have in Whitehall though, so I wouldn't change the reassignment.

app/views/admin/statistics_announcements/_filter_options.html.erb
```diff
@@ -13,6 +13,7 @@
       name: "title",
       id: "title_filter",
       value: @filter.options[:title],
+      type: "search",
```
ryanb-gds, line 16: This is a good addition, imagine it might improve the accessibility a bit 👏

ryanb-gds (commented):
I'm a little bit dubious about the need to copy the form search data attributes from the design system container element onto the form. Are there such a large number of search forms that just adding the data attributes directly to the form markup is painful? And also what are we going to do if we decide at some point to change one of the values?

patrickpatrickpatrick:
> I'm a little bit dubious about the need to copy the form search data attributes from the design system container element onto the form. Are there such a large number of search forms that just adding the data attributes directly to the form markup is painful? And also what are we going to do if we decide at some point to change one of the values?

A different approach now I think about it could be removing "Filter by" on the design system container and making a default value in the tracker itself instead (since that's in effect what I'm doing here anyway): 

```
const defaultGa4SearchSectionType = "Filter by"

let ga4SearchSectionType = searchForm.closest(
  '[data-ga4-search-section-type]'
)

if (ga4SearchSectionType) {
  ga4SearchSectionType =
    ga4SearchSectionType.dataset.ga4SearchSectionType
}
```

app/assets/javascripts/admin/analytics-modules/ga4-search-setup.js
patrickpatrickpatrick: This is kind of a started so I finish style currently, there's some changes I want to do to all the trackers (such as using `dataset.attribute` and not `getAttribute("data-set-attribute")`.

patrickpatrickpatrick:
> > > I'm a little bit dubious about the need to copy the form search data attributes from the design system container element onto the form. Are there such a large number of search forms that just adding the data attributes directly to the form markup is painful? And also what are we going to do if we decide at some point to change one of the values?
> > 
> > 
> > A different approach now I think about it could be removing "Filter by" on the design system container and making a default value in the tracker itself instead (since that's in effect what I'm doing here anyway):
> > ```
> > let ga4SearchSectionType = "Filter by"
> > 
> > const ga4SearchSection = searchForm.closest(
> >   '[data-ga4-search-section-type]'
> > )
> > 
> > if (ga4SearchSection) {
> >   ga4SearchSectionType =
> >     ga4SearchSection.dataset.ga4SearchSectionType
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could keep the functionality for overriding the section type in the process.
> 
> I think that would be reasonable. Are you proposing keeping the document type behaviour as is?

In my opinion, `document type` still makes sense to kept there because it's always the controller and action name and so would have to be taken from the DOM somehow. Alternatively it could be assigned to a window object to be referenced like so: 

```
<script type="javascript>
window.GOVUK.analyticsGa4.documentType = "<%= controller_name %>-<%= action_name %>"
</script>
```

would then need to change other trackers that use the data attribute from the root container though. Could do that in another PR.

ryanb-gds:
> In my opinion, `document type` still makes sense to kept there because it's always the controller and action name and so would have to be taken from the DOM somehow. Alternatively it could be assigned to a window object to be referenced like so:
> 
> ```
> <script type="javascript>
> window.GOVUK.analyticsGa4.documentType = "<%= controller_name %>-<%= action_name %>"
> </script>
> ```
> 
> would then need to change other trackers that use the data attribute from the root container though. Could do that in another PR. There might be even be a way to actually change the default value of in the schema in `window.GOVUK.analyticsGa4.Schemas` but I'd have to check more into that.

Sorry, I misread the code, I thought the document type was a new addition to the container div when it predates this PR. Carry on 😅

ryanb-gds (approved):
Looks good to me, I'll leave it up to you about the filter by default

PR #10174, opened 2025-04-30
Return an empty map if there are no headings in body

deborahchua:
We're currently returning an empty array in the `headers` details, but instead we should be returning an empty map as there are no headings present in the body.

Related PR - https://github.com/alphagov/whitehall/pull/10143

Trello card: https://trello.com/c/EKDXTDsm/398-move-document-type-corporateinformationpage-from-government-frontend-to-frontend

### DIFF
--- a/app/presenters/publishing_api/payload_builder/body_headings.rb
+++ b/app/presenters/publishing_api/payload_builder/body_headings.rb
@@ -12,7 +12,11 @@ module PublishingApi
       end
 
       def call
-        { headers: body_headings }
+        headers = body_headings
+
+        return {} if headers.empty?
+
+        { headers: }
       end
 
     private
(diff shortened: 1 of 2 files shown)


app/presenters/publishing_api/payload_builder/body_headings.rb
```diff
@@ -12,6 +12,8 @@ def initialize(item)
       end
 
       def call
+        return {} if body_headings.empty?
```
KludgeKML, line 15: How heavy is the call to Govspeak::Document.new in body_headings? Should we store the value of body_headings it in a temp variable here to avoid calling it twice?

deborahchua: Yes good point! This would be an issue with very long documents - I've updated this now.

KludgeKML (approved):
LGTM!

ChrisBAshton (approved):

PR #10177, opened 2025-05-01
Change field_of_operation rendering app to frontend

AshGDS:
As part of app consolidation. See https://github.com/alphagov/frontend/pull/4742


KludgeKML (approved):
LGTM!

PR #10186, opened 2025-05-06
Remove training page from Whitehall

GDSNewt:
Reverts the change made in [this PR](https://github.com/alphagov/whitehall/pull/8754)

We have setup a new space for [publisher guidance](https://github.com/alphagov/govuk-content-publishing-guidance). Training documentation will be moved over there and does not need to remain part of whitehall. 

[This](https://trello.com/c/EoJph8wc/3618-support-content-guidance-migration-off-manuals) new space for publisher guidance feels like a better place for [Whitehall training pages](https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/how-to-use-whitehall-publisher) to live (if we need them at all).


ryanb-gds (approved):
🎉 good to see this finding a new home!

PR #10189, opened 2025-05-07
Cm55 multiple subschemas

pezholio:


lib/engines/content_block_manager/app/controllers/concerns/workflow/steps.rb
```diff
@@ -6,7 +6,7 @@ module Workflow::Steps
   end
 
   def steps
-    @steps ||= @schema.subschemas.any? ? all_steps : all_steps.reject { |s| s.name == :embedded_objects }
+    @steps ||= should_show_subschema_steps? ? all_steps : all_steps.reject { |s| s.name == :embedded_objects }
```
Harriethw, line 9: was this suggested by UCD? I would find it a bit confusing as a user not to be able to add a new mode in the edit flow, because it's like the main CTA at the top of the page and I might miss the 'Add an email' button below, also it is technically editing the block 🤔

pezholio: Yeah, this is what happens at the moment anyway - it's only possible to edit subschema items in the edit flow - adding new ones is only possible from the show action. I remember having a long back and forth with Nadeem about it

Harriethw: ah ok, seems quite unintuitive to me (and I *think* one of the users brought it up), but guess that will all come out in the wash :)

Harriethw (approved):
nice one, good to see this being tidied up

PR #10195, opened 2025-05-08
(CM-66) Remove nils from `embedded_object_diffs`

pezholio:
If there are multiple subschemas for an object, and only one object type has been updated, there is no entry for that subschema in the diff, and we end up trying to cast a `nil` to a hash when initializing the `FieldChangesTableComponent` in line 90.

Simply adding `.compact` to the end of the `map` call removes any nils from the resulting array and allows us to continue error free.

I’ve added multiple subschemas to the tests to reflect this behaviour.


ryanb-gds (approved):
Looks good to me. Dynamic typing strikes again 🙈

PR #10201, opened 2025-05-12
use form tracker

patrickpatrickpatrick:
- **what if we used a form tracker**
- **`Ga4IndexSetup` changes to index calculation**
- **`Ga4FinderTracker` HTML data attribute changes**
- **more form**
- **forms**


app/assets/javascripts/admin/analytics-modules/ga4-form-tracker.js
```diff
@@ -0,0 +1,98 @@
+'use strict'
```
lauraghiorghisor-tw, line 1: Would be good to see a breakdown of where the two implementations diverge (components vs ours) in the commit message. The "why" of the work.

I mean tests explain the end result but still unsure what we extended (without looking into the components code).

app/assets/javascripts/admin/analytics-modules/ga4-select-tracker.js
```diff
@@ -1,46 +0,0 @@
-'use strict'
```
lauraghiorghisor-tw, line 1: Deletion, yes, so satisfying 😌

spec/javascripts/helpers/helpers.js
```diff
@@ -0,0 +1,173 @@
...
+    date,
+    checkbox,
+    text,
+    textarea,
+    radio,
+    select,
+    'select-multiple': selectMultiple
+  }
+
+  class FormHelper {
```
lauraghiorghisor-tw: Would this clash with anything if it were named just "Form" 👀

app/assets/javascripts/admin/analytics-modules/ga4-form-setup.js
```diff
@@ -0,0 +1,51 @@
...
+;(function (Modules) {
+  Modules.Ga4FormSetup = {
+    init: function () {
+      const $modules = document.querySelectorAll(
+        "[data-module~='ga4-form-setup']"
+      )
+
+      $modules.forEach(($module) => {
+        const forms = $module.querySelectorAll(
+          "form:not([data-module~='ga4-finder-tracker'])"
```
lauraghiorghisor-tw, line 15: And what do search forms do? 👀

spec/javascripts/helpers/helpers.js
patrickpatrickpatrick: Yeah, I could just name it form and it would be fine! Maybe was being overly cautious.

app/assets/javascripts/admin/analytics-modules/ga4-form-setup.js
patrickpatrickpatrick: Search forms have their own tracking from [Ga4FinderTracker](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js) that lives in `govuk_publishing_components`. They handle some things a bit differently (such as having removal events for certain interactions which wouldn't make sense on a form for submitting data), so that's why we have different approaches.

app/assets/javascripts/admin/analytics-modules/ga4-form-tracker.js
patrickpatrickpatrick: Sure, I can add a comment here!

patrickpatrickpatrick:
> I soldiered through - no blocking comments - LGTM. Feel free to get another opinion if you need it though. If Ed sees what he needs to see it's (mostly) good enough for me.
> 
> One suggestion for when the work is done - add some documentation that lives in WH, capturing the general technical approach, without going into crazy detail. I know we have the big implementation document but that feels too PA-y (and too overwhelming!).

Yeah that makes sense. Hopefully most of the tracking code itself will not be in whitehall but would still be good to have some docs that outline the approach.

PR #10202, opened 2025-05-13
Remove references to old document hub feature flag

ryanb-gds:
This feature flag was removed some time ago, so I'm quite surprised these tests were still passing. Obviously flipflop doesn't enforce that a feature flag actually exists when you attempt to switch it on 🤔


lauraghiorghisor-tw (approved):
LGTM ✅

PR #10203, opened 2025-05-13
(CM-67) Schema improvements

pezholio:
This adds some improvements to the schema, mainly around calculating which form components to use, as well as ensuring field orders are retained everywhere the details object is used.


Harriethw (commented):
looks good!  just one final Q about not checking the schema's `order` field now.

Harriethw (approved):

Harriethw (approved):
🙌

PR #10206, opened 2025-05-13
Add labeler action

pezholio:
This automatically labels any PRs that have changes to files in the content_block_manager engine with the `Content Modelling` label using the [official Github Labeler action](https://github.com/actions/labeler)


ryanb-gds (approved):
Ooh, nice 🏷️

PR #10213, opened 2025-05-14
Switch to using a separate repo for select with search.

minhngocd:
Use gem from select-with-search-component github repo to import Select With Search component.

A few things were not removed:
- Docs will still need to reside in Whitehall as the component guide overrides loads from this folder and it seems moving it to the Gem engine stops loading that component. Also the component guide will not necessarily need to be ported as it is not a reusable part of the select with search component.
- GA4 Analytics: This is being removed and refactored as part of a separate piece of work. GA4 tracking is being pulled to track the form component as opposed to individual component as part of a separate piece of work. This functionality will be removed in that piece of work and is not ported over to the select-with-search component repo

Working as usual
image

https://trello.com/c/13PXF8wR/3687-design-system-move-selectwithsearch-component-somewhere-shared


ChrisBAshton (approved):
LGTM 👍 

Would be nice to [publish the gem to rubygems](https://docs.publishing.service.gov.uk/manual/publishing-a-ruby-gem.html) so that we don't need to pull it in via Git (which blocks [this card](https://trello.com/c/ZEpN5Ejm/39-upgrade-whitehall-to-yarn-4)) 🙏 I don't see why we can't publish it as a version 1.0.0.

Aside: [our naming convention for gems](https://docs.publishing.service.gov.uk/manual/naming.html#naming-gems) might favour a name like `govuk_select_with_search_component` but as the component _could_ be of use outside of GOVUK, the current name is fine.

patrickpatrickpatrick:
Looks good to me also, can rebase my PR if this is merged before :)

PR #10218, opened 2025-05-16
Fix draft creation to check for clone attachment validity

minhngocd:
Previously we seem to be checking for existing attachment save! as opposed to the newly cloned attachment for the new draft. Whilst debugging a ZenDesk ticket: https://govuk.zendesk.com/agent/tickets/6122813 we find that original attachment can return valid, but the newly cloned attachment fails validation and does not save. The save fails silently and the attachment is removed from the edition.

In doing this, some integration tests are failing because editions are `built` in the tests, not `created`. With the previous code, since we are saving the original edition attachment, that saves the edition to the database and commits the document. With this no longer there, document does not exist for the presenter to read content_id out of, hence the test fail. We need to create the editions for tests to run.

It is now surfacing an error when you try to create a new edition and html attachments fail to clone
image

https://trello.com/c/DICxbdzx/3730-html-attachment-gets-lost-when-creating-new-edition

### DIFF
--- a/app/models/concerns/attachable.rb
+++ b/app/models/concerns/attachable.rb
@@ -35,8 +35,9 @@ module Attachable
       add_trait do
         def process_associations_after_save(edition)
           @edition.attachments.each do |attachment|
-            edition.attachments << attachment.deep_clone
-            raise "Your edition failed to create successfully. Please contact GOV.UK support." unless attachment.save!
+            draft_attachment = attachment.deep_clone
+            edition.attachments << draft_attachment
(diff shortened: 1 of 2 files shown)


minhngocd:
Also to clarify - the draft attachment needs to be associated with the new draft rather than the old edition before we can save it so it is associated with the correct edition.

minhngocd:
One other thing to note - the validation error is for the HTML attachment, but the message on the UI does not make that clear. We might want to look at that later down the line, but for now we want to get this in ASAP so we don't have any issues with HTML attachments being silently dropped from editions.

ChrisBAshton (approved):

PR #10230, opened 2025-05-20
Remove `analytics-ga4` from `application.js`

patrickpatrickpatrick:
## What

- Remove `analytics-ga4.js` from `application.js`
- Add `test-dependencies.js` 
- Update `manifest.js` and `jasmine-browser.json` to use `test-dependencies.js`

## Why

`analytics-ga4.js` is already included in `load-analytics.js` from `govuk_publishing_components`.

### Difference in `application.js` file size

| Before  | After |
| ------------- | ------------- |
| 298kb  | 252kb  |

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,6 +1,7 @@
 //= link_tree ../images
 //= link_tree ../builds
 //= link application.js
+//= link test-dependencies.js
 //= link admin/domain-config.js
 //= link es6-components.js
 //= link components/visual-editor.js
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,7 +1,6 @@
 //= require admin/stop-scripts-nomodule
 
 //= require govuk_publishing_components/dependencies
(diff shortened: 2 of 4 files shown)


app/assets/config/manifest.js
```diff
@@ -1,6 +1,7 @@
 //= link_tree ../images
 //= link_tree ../builds
 //= link application.js
+//= link test-dependencies.js
```
lauraghiorghisor-tw, line 4: Any chance we don't have to include this here at all since they are test dependencies?

lauraghiorghisor-tw (approved):

app/assets/config/manifest.js
patrickpatrickpatrick: Is there a way to only include things in the manifest specifically or having an environment specific manifest? This was the way I found to get the test-dependencies to be present in `public`.

lauraghiorghisor-tw: Thanks for digging it out, all good if it's a common pattern. Ship it 🛳️

PR #10231, opened 2025-05-20
CM-71: remove `email_address` and `postal_address` as top level block

Harriethw:
https://gov-uk.atlassian.net/browse/CM-71?atlOrigin=eyJpIjoiY2UyMWFjZDI5MDFjNDk2YWE2NDhhMTgwNWYzNGU5ZjAiLCJwIjoiaiJ9

We began the project with `email_address` as our first block, so it is weaved into most of our test cases as the default type. I was a bit unsure about removing it but I think it feels like a necessary growing pain - it looks like the `email_address` as a type will never exist for our users so we should remove it and base things on types we are certain of. 

This does mean we no longer have a type that doesn't have embedded objects, but as we kind of create the schemas on the fly in the feature tests I think we can cover every scenario.

Will follow this up with tidying up PR's in Tools and Publishing API as well, basically reversing these instructions https://docs.publishing.service.gov.uk/content-modelling/schemas/adding-a-new-content-block-type.html#create-a-new-schema-in-publishing-api

----


lib/engines/content_block_manager/features/edit_object.feature
```diff
@@ -2,16 +2,16 @@ Feature: Edit a content object
   Background:
     Given I am a GDS admin
     And the organisation "Ministry of Example" exists
-    And a schema "email_address" exists with the following fields:
+    And a schema "pension" exists with the following fields:
       | field         | type   | format | required |
-      | email_address | string | email  | true     |
-    And an email address content block has been created
+      | description   | string | string | true     |
```
Harriethw, line 7: do we think it's ok to setup a pension without an embedded object of rate here? as the subschema functionality isn't under test here...
Maybe the confusion speaks to some tech debt around the tests here, that we are having to create bespoke ones each time we create a type and there could be a better organising of them?

pezholio: Yeah, I had a go at doing this, but I ended up getting tied up in knots

PR #10232, opened 2025-05-21
Content Modelling ADR - Add a new attribute to content blocks CM-111

pezholio:
https://gov-uk.atlassian.net/jira/software/c/projects/CM/boards/134?selectedIssue=CM-111

This spells out our intention to add a new `field_order` attribute to content blocks, giving users the ability to reorder the display of attributes within content blocks, as well as creating new block variants.

### DIFF
--- a/docs/content_block_manager/adr/0003-add-new-field-order-attribute-to-content-blocks.md
+++ b/docs/content_block_manager/adr/0003-add-new-field-order-attribute-to-content-blocks.md
@@ -1,0 +1,167 @@
+# 3. Add a new Field Order attribute to content blocks 
+
+Date: 2025-05-20
+
+## Context
+
+We are introducing the concept of "views" for entire blocks, in addition to being able to embed individual
+attributes. These views can be ordered in specific ways (for example, a user can choose to show the telephone
+number before the address in the case of a contact), as well as allowing custom views to be created to allow
+only specific attributes to be shown.
+
+At the moment, the order and presentation of blocks is controlled by code in the 
+[`content_block_tools`](https://github.com/alphagov/content_block_tools) gem. Each block type has a presenter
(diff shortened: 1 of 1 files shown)


docs/content_block_manager/adr/0003-render-blocks-at-publish-time.md
```diff
@@ -0,0 +1,120 @@
...
+    <p>My contact details</p>
+    <p>Tel: 12345</p>
+    <p>Email: <a href="mailto:abc@example.com">abc@example.com</a></p>
+</div>
+```
+
+This will then be sent in the JSON payload as a default `views` object within the payload sent to Publishing API. An 
+abridged version of the payload could look like this:
+
+```json
```
pezholio, line 51: Another thing I’ve just thought about is rather than sending the HTML to Publishing API, we could just send an ordered array as a `blocks` element, with the order of the blocks specified. So for addresses, the default would look like this:

```
{
  "blocks: {
    "default": ["title", "addresses", "email", "telephones"],
  }
}
```

This might be simpler, but would give us less flexibility if we wanted to introduce (for example) custom styling / presentation to blocks, as well as the ability to add arbitrary text around blocks

Harriethw (approved):
Thanks for working on this! Seems like the best way forward...

I was wondering if we should do yet more nesting and have something like `body` or `content` or something that has all the attributes we'll build the forms out of within `details`? and then this `field_order` key is metadata sitting alongside it? That way we would have to add some ugly config check to ignore `field_order` and we just replace iterating through `details` with iterating through `body`?

Harriethw (changes requested):
changing my review to request changes in light of the pre-work that needs to be done as Stu has commented.

pezholio:
Closing this, as we've decided to pause this piece of work until we know more about user needs

PR #10244, opened 2025-05-27
Fix `HtmlAttachment` reslug rake task

patrickpatrickpatrick:
## What

Update the `HtmlAttachment` reslug rake task

## Why

In a support issue, it was revealed that the rake task does not work as expected. This was because:

- the slug of attachments is generated from the title of the attachment
  - the slug cannot be named differently from it
  - if the slug/title already exists then `--NUMBER` will be appended to it
- if another attachment associated with the edition that has been `deleted` (but not removed from the database) has the same slug, the target attachment cannot be renamed to that slug

So in this approach we don't accept a new slug as a parameter. In light of what we discovered, we could have used the 'title' of the attachment we want to reslug instead. As this can be changed within whitehall already though, personally it makes sense to me to defer to this to the user to make the process of reslugging more straightforward.

Solves: https://trello.com/c/NNssAJ4u/3668-fix-html-attachment-reslugging-rake-task

### DIFF
--- a/lib/tasks/reslugging.rake
+++ b/lib/tasks/reslugging.rake
@@ -47,7 +47,7 @@ namespace :reslug do
   - changes a html attachment slug
   - reindexes the document
   - republishes the document and attachment to Publishing API (automatically handles the redirect)"
-  task :html_attachment, %i[publication_slug old_attachment_slug new_attachment_slug] => :environment do |_task, args|
+  task :html_attachment, %i[publication_slug old_attachment_slug] => :environment do |_task, args|
     documents = Document.where(slug: args.publication_slug)
     if documents.count > 1
       raise "There are multiple documents with the slug '#{args.publication_slug}'. Consider writing a migration and fetching the document with a content_id or document_type to uniquely identify it."
@@ -59,8 +59,20 @@ namespace :reslug do
 
     raise "Could not find existing attachment with slug /#{args.old_attachment_slug}" unless html_attachment
 
-    # update slug of html attachement and add redirect
-    html_attachment.update!(slug: args.new_attachment_slug)
+    new_attachment_slug = html_attachment.title.to_slug.normalize.to_s
+
+    existing_attachment_with_slug = edition.attachments.find_by(slug: new_attachment_slug)
+
+    raise "Attachment with slug '#{new_attachment_slug}' already exists and has been not deleted. Delete this attachment first." if existing_attachment_with_slug
+
+    # need to hard delete attachment with the new slug if exists
+    # as FriendlyID will identify it as a conflict and append an
+    # index to the slug we want to use
+    edition.attachments.unscoped.deleted.delete_by(slug: new_attachment_slug)
+
+    html_attachment.update!(safely_resluggable: true)
+    html_attachment.update!(safely_resluggable: false)
+
     Whitehall::PublishingApi.republish_async(html_attachment)
 
     # send edition to publishing api
--- a/test/unit/lib/tasks/reslugging_test.rb
+++ b/test/unit/lib/tasks/reslugging_test.rb
@@ -2,15 +2,85 @@ require "test_helper"
 require "rake"
 
 class ResluggingTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
   teardown do
+    task.reenable
     Sidekiq::Job.clear_all
   end
 
-  test "it should reslug the world location" do
-    world_location_news = build(:world_location_news, content_id: SecureRandom.uuid)
-    world_location = create(:world_location, slug: "old-name", world_location_news:)
-    Rake.application.invoke_task "reslug:world_location[old-name,new-name]"
+  describe "for world location" do
+    let(:task) { Rake::Task["reslug:world_location"] }
 
-    assert_equal "new-name", world_location.reload.slug
+    test "it should reslug" do
+      world_location_news = build(:world_location_news, content_id: SecureRandom.uuid)
+      world_location = create(:world_location, slug: "old-name", world_location_news:)
+      task.invoke("old-name", "new-name")
+      Rake.application.invoke_task "reslug:world_location[old-name,new-name]"
+      assert_equal "new-name", world_location.reload.slug
+    end
+  end
+
+  describe "for HTML attachments" do
+    let(:task) { Rake::Task["reslug:html_attachment"] }
+    let(:html_attachment_title) { "HTML Attachment" }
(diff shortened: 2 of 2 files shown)


test/unit/lib/tasks/reslugging_test.rb
```diff
@@ -2,15 +2,75 @@
 require "rake"
 
 class ResluggingTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
   teardown do
+    task.reenable
```
patrickpatrickpatrick, line 8: sidenote, I'm not really sure why it needs to be reenabled each time but if it isn't then it doesn't work!

lib/tasks/reslugging.rake
```diff
@@ -59,8 +60,16 @@ namespace :reslug do
 
     raise "Could not find existing attachment with slug /#{args.old_attachment_slug}" unless html_attachment
 
-    # update slug of html attachement and add redirect
-    html_attachment.update!(slug: args.new_attachment_slug)
+    new_attachment_slug = html_attachment.title.to_slug.normalize.to_s
+
+    deleted_attachments_of_edition = edition.attachments.unscoped.deleted
+    deleted_attachments_of_edition.delete_by(slug: new_attachment_slug)
```
patrickpatrickpatrick, line 66: for some reason `delete_by` was the only way I could get actual deletion to happen (`delete` wouldn't do it)

lauraghiorghisor-tw (approved):
Have some questions but I think overall OK as long as you give them some thought 😅

lib/tasks/reslugging.rake
```diff
@@ -59,8 +60,16 @@ namespace :reslug do
...
+
+    safely_resluggable_attachment = html_attachment.safely_resluggable
+
+    html_attachment.update!(safely_resluggable: true)
+    html_attachment.update!(safely_resluggable: safely_resluggable_attachment)
```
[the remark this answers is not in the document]
patrickpatrickpatrick: That's actually a good point wrt to the resluggable state, it probably will always be set to false since we're only looking for the last published edition. So can probably change that.

lib/tasks/reslugging.rake
```diff
@@ -59,8 +60,16 @@ namespace :reslug do
...
-    # update slug of html attachement and add redirect
-    html_attachment.update!(slug: args.new_attachment_slug)
+    new_attachment_slug = html_attachment.title.to_slug.normalize.to_s
+
+    deleted_attachments_of_edition = edition.attachments.unscoped.deleted
+    deleted_attachments_of_edition.delete_by(slug: new_attachment_slug)
+
+    safely_resluggable_attachment = html_attachment.safely_resluggable
+
+    html_attachment.update!(safely_resluggable: true)
```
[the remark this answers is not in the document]
ryanb-gds: Oooff, I did not write that clearly. Let me try again: "if there is a conflicting attachment that is not deleted, abort".

lib/tasks/reslugging.rake
ryanb-gds: Because of the way the FriendlyID gem works, if there's any matching slug in the database, even for a deleted attachment, it will suffix the slug. So we need to perform a hard delete else the `--2` suffix will remain. Thinking about it, we might be able to include "not deleted" in [the scope we use when creating the slug](https://github.com/alphagov/whitehall/blob/9e9e9b7b68c71de2661274bcbdb9ebec71cd14b1/app/models/attachment.rb#L4) ([see FriendlyID docs for how scopes work](https://norman.github.io/friendly_id/file.Guide.html#Unique_Slugs_by_Scope))

I agree that hard deleting is a bit contrary to our general technical approach so perhaps we should explore this. We might also have had conflict issues in Publishing API actually wherein it would already have a content item at the base path (would need to check), so this is a good point @lauraghiorghisor-tw

lib/tasks/reslugging.rake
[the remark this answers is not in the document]
lauraghiorghisor-tw: @ryanb-gds Oooh, I see what you mean 😅

But equally, we could check that indeed the total number of soft deleted attachments with the desired slug is 0 before confidently declaring this is safely resluggable?

lib/tasks/reslugging.rake
```diff
@@ -59,8 +60,19 @@ namespace :reslug do
...
-    html_attachment.update!(slug: args.new_attachment_slug)
+    new_attachment_slug = html_attachment.title.to_slug.normalize.to_s
+
+    existing_attachment_with_slug = edition.attachments.find_by(slug: new_attachment_slug)
+
+    raise "Attachment with slug '#{new_attachment_slug}' exists that has been not deleted. Delete the existing attachment first." if existing_attachment_with_slug
+
+    edition.attachments.unscoped.delete_by(slug: new_attachment_slug)
+
+    html_attachment.safely_resluggable
```
ryanb-gds, line 71: I don't think this line is doing anything?

lib/tasks/reslugging.rake
```diff
@@ -59,8 +60,19 @@ namespace :reslug do
...
 
-    # update slug of html attachement and add redirect
-    html_attachment.update!(slug: args.new_attachment_slug)
+    new_attachment_slug = html_attachment.title.to_slug.normalize.to_s
+
+    existing_attachment_with_slug = edition.attachments.find_by(slug: new_attachment_slug)
+
+    raise "Attachment with slug '#{new_attachment_slug}' exists that has been not deleted. Delete the existing attachment first." if existing_attachment_with_slug
+
+    edition.attachments.unscoped.delete_by(slug: new_attachment_slug)
```
ryanb-gds, line 69: There's a `deleted` scope you can use ([link](https://github.com/alphagov/whitehall/blob/4d3af94543b21be448110667188a0073bf3cc4bd/app/models/attachment.rb#L33C8-L33C17)), which would be more explicit (and safer) than unscoped. Probably worth adding a comment explaining why we need to do this as well (because of how FriendlyID works)

patrickpatrickpatrick: For some reason I have to call `unscoped.deleted` for the `delete_by` to work, `deleted` means that nothing will be deleted.

ryanb-gds: That does make sense, without unscoped we're applying both the "not deleted" and "deleted" scopes, so nothing would be returned from the query

patrickpatrickpatrick: Ah yeah of course that makes sense!

patrickpatrickpatrick:
Tested on integration and works as expected.

PR #10248, opened 2025-05-28
Add `data-ga4-no-copy` to layout

patrickpatrickpatrick:

### DIFF
--- a/app/views/layouts/design_system.html.erb
+++ b/app/views/layouts/design_system.html.erb
@@ -23,7 +23,7 @@
 
   <%= render partial: "shared/header" %>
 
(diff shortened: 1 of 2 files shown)


patrickpatrickpatrick:
> This is fine from a technical perspective, but are we really sure we have no interest in tracking this?
> 
> I'd be quite interested to know, for example, how many documents are more or less copies from another document with maybe one small change. Or if small parts of a piece of content are copied across many docs - could give good insight to the content block modelling team. Might be worth a chat at standup tomorrow.

Sure, I'm happy to discuss this in standup tomorrow 🙂

I'm not super familiar with the content block modelling side of whitehall but the copy tracker is intended for tracking the copying of static content on a published page. It deliberately doesn't track copied content from forms (because of PII concerns). I'm aware there's functionality of [copying embed code in the content block codebase](https://github.com/alphagov/whitehall/blob/9e9e9b7b68c71de2661274bcbdb9ebec71cd14b1/lib/engines/content_block_manager/app/assets/javascripts/content_block_manager/modules/copy-embed-code.js#L5) but the way this functions isn't trackable by the copy tracker either as the workaround uses a `textarea` to copy the text from.

PR #10250, opened 2025-05-30
CM 123 modes table

Harriethw:
- **add tab components to render embedded objects**
- **display tabs on block view**


lib/engines/content_block_manager/app/views/content_block_manager/content_block/documents/show.html.erb
```diff
@@ -46,31 +46,42 @@
   </div>
 </div>
 
-<% @schema.subschemas.each do |subschema| %>
```
Harriethw, line 49: are we happy at least initially to just rely on the count of subschemas, or would it be preferable to define in our config?

pezholio: I think for now, yeah, but we can tweak as we see fit

PR #10261, opened 2025-06-03
CM-131: Subschema default name ("mode" for Contact) config

Harriethw:


lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema.rb
```diff
@@ -77,6 +77,10 @@ def config
         @config ||= self.class.schema_settings.dig("schemas", @id) || {}
       end
 
+      def subschemas_default_name
```
Harriethw: is there a better naming convention to use?

lib/engines/content_block_manager/config/content_block_manager.yml
```diff
@@ -11,6 +11,7 @@ schemas:
           - description
   content_block_contact:
     subschemas:
+      default_name: mode
```
pezholio: I wonder if a better thing to do would be for each subschema to have a `group` variable, which would then allow us to merge multiple subschemas into one or more groups when rendering them on the view page. Something like (in the case of contacts):

```yaml
  content_block_contact:
    subschemas:
      email_addresses:
        group: modes
        embeddable_fields:
          - email_address
      telephones:
        group: modes
        embeddable_fields:
          - telephone
      addresses:
        group: modes
        field_order:
          - title
          - street_address
          - locality
          - region
          - postal_code
          - country
      contact_forms:
        group: modes
        embeddable_fields:
          - url
```

Then, when displaying the subschemas on the show page, we could group any subschemas that are in the same group together and show them in a tab view. T

his would also remove the need for the ugly cludge in https://github.com/alphagov/whitehall/blob/86bb8c55f336da93db3f43b70818a48b44db6608/lib/engines/content_block_manager/app/views/content_block_manager/content_block/documents/show.html.erb#L49 where we're checking if there's one subschema, and also give us the flexibility to have any number of groups, or have some subschemas grouped and some not. 

Whaddya reckon?

Harriethw: ooh yeah that's much more elegant

Harriethw: It's a bit tricky when doing the create/edit flow -  we only have capacity to add all the embedded objects at once at the moment, so knowing where to pull that group name from on a parent schema level is a bit tricky. Would groups maybe be a further iteration of the code, as then we could iterate through the groups in the form flow as well?
Screenshot 2025-06-03 at 15 55 11

Harriethw: or maybe it's not an iteration, it's just a bigger piece of work than i have time to do right now!

pezholio: Maybe let's keep the create/edit flow for a further iteration?

pezholio: I can have a look later this week 👍

Harriethw: on the parent schema, would you have a key like this?
```
content_block_contact:
    subschema_groups:
          - modes
    subschemas: 
          email_addresses:
          ...
```
because we iterate through the `subschemas` key to get the subschemas

pezholio: Ah yeah, that'll work 👍

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/documents/embedded_objects_controller.rb
```diff
@@ -1,10 +1,25 @@
...
   include EmbeddedObjects
 
-  before_action :initialize_document_and_schema
+  before_action :initialize_document_and_schema, except: [:new_embedded_objects_options_redirect]
 
   def new
     @content_block_edition = @content_block_document.latest_edition
+
+    if @subschema
+      @back_link = flash[:back_link] || content_block_manager.content_block_manager_content_block_document_path(@content_block_document)
```
richardTowers, line 12: Does `@back_link` also need to be assigned in the else condition? It's hard to tell from this vantage point.

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/documents/embedded_objects_controller.rb
```diff
@@ -1,10 +1,25 @@
...
+      render :new
+    else
+      @schema = ContentBlockManager::ContentBlock::Schema.find_by_block_type(@content_block_document.block_type)
```
richardTowers: Won't this already be set by `initialize_document_and_schema` ?

richardTowers: Ah, maybe I see what's going on... That before_action only calls `get_schema_and_subschema` now if `params[:object_type]` is truthy.

I think this is a bit of a confusing setup. I wonder if you should split it into an `initialize_document` before action (which you wouldn't need to skip in `new_embedded_objects_options_redirect`), and then call `get_schema_and_subschema` explicitly in your actions?

richardTowers (commented):
I found some of the before_action stuff a bit confusing - I reckon that could be refactored into a better shape. Otherwise looks good.

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/documents/embedded_objects_controller.rb
pezholio: No, it's not used in the other view, but I can see why it would be confusing. Will fix for consistency 👍

lib/engines/content_block_manager/app/controllers/content_block_manager/content_block/documents/embedded_objects_controller.rb
pezholio: Good idea 👍

richardTowers (approved):
Thanks! Looks good!

PR #10262, opened 2025-06-04
(CM 141) Allow attribute blocks to be hidden

pezholio:
If an object is embeddable as a block, we want to show this front and centre, but also give the user the opportunity to use the individual attributes that make up the block if they want.

If an object has a block-level representation, we render a `details` component with a further summary list inside. The user can then expand the details element to see the other blocks.

I’ve altered the styling a bit to make the component look like one unified component, whilst still keeping the HTML neat

## Screenshot

image

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.html.erb
@@ -1,6 +1,25 @@
-<div class="app-c-embedded-objects-blocks-component">
+<%= tag.div(class: component_classes) do %>
   <%= render "govuk_publishing_components/components/summary_card", {
     title: "Content blocks",
-    rows:,
+    rows: summary_card_rows,
   } %>
-</div>
+
+  <% if schema.embeddable_as_block? %>
+    <div class="app-c-embedded-objects-blocks-component__details-wrapper">
+      <%= render "govuk_publishing_components/components/details", {
+        title: "All #{object_name} attributes",
+      } do %>
+        <% capture do %>
+          <div class="app-c-embedded-objects-blocks-component__details-text">
+            These are all the <%= object_name %> attributes that make up the <%= object_name %>. You can use the embed code for each attribute separately in your content if required.
+          </div>
+          <div class="app-c-embedded-objects-blocks-component__details-summary-list">
+            <%= render "govuk_publishing_components/components/summary_list", {
+              items: attribute_rows(:field),
+            } %>
+          </div>
+        <% end %>
+      <% end %>
+    </div>
+  <% end %>
+<% end %>
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
@@ -12,18 +12,33 @@ private
 
   attr_reader :items, :object_type, :object_title, :content_block_document
 
-  def rows
-    rows = schema.embeddable_as_block? ? [block_row] : []
+  def component_classes
+    [
+      "app-c-embedded-objects-blocks-component",
+      ("app-c-embedded-objects-blocks-component--with-block" if schema.embeddable_as_block?),
+    ].compact.join(" ")
+  end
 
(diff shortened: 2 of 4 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
```diff
@@ -12,18 +12,32 @@ def initialize(items:, object_type:, object_title:, content_block_document:)
 
   attr_reader :items, :object_type, :object_title, :content_block_document
 
-  def rows
-    rows = schema.embeddable_as_block? ? [block_row] : []
+  def component_classes
+    classes = ["app-c-embedded-objects-blocks-component"]
+    classes.push("app-c-embedded-objects-blocks-component--with-block") if schema.embeddable_as_block?
+    classes.join(" ")
```
richardTowers: Nit:

```suggestion
    [
      "app-c-embedded-objects-blocks-component",
      ("app-c-embedded-objects-blocks-component--with-block" if schema.embeddable_as_block?)
    ].compact.join(" ")
```

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.html.erb
```diff
@@ -1,6 +1,25 @@
-<div class="app-c-embedded-objects-blocks-component">
+<div class="<%= component_classes %>">
```
richardTowers: Nit / possibly controversial one:

I like to do

```
<%= tag.div(class: component_classes) %>

...

<% end %>
```

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.html.erb
```diff
@@ -1,6 +1,25 @@
...
-    rows:,
+    rows: block_rows,
   } %>
+
+  <% if schema.embeddable_as_block? %>
+    <div class="app-c-embedded-objects-blocks-component__details-wrapper">
+      <%= render "govuk_publishing_components/components/details", {
+        title: "All #{object_name} attributes",
+      } do %>
+        <% capture do %>
```
richardTowers, line 12: Is this capture needed?

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
```diff
@@ -12,18 +12,32 @@ def initialize(items:, object_type:, object_title:, content_block_document:)
...
-    items.each do |key, value|
-      rows << {
-        key: key.titleize,
+  def block_rows
+    if schema.embeddable_as_block?
+      [block_row]
+    else
+      attribute_rows
+    end
+  end
```
richardTowers: As I understand it there's two kinds or row we might get:
- "Block" rows - exactly one if the schema is embeddable as a block
- "Attribute" rows - many, rendered in the summary card if the schema only has attributes, rendered in a details element if the schema is embeddable as a block

I find it a bit confusing to have attribute rows returned from a method called block_rows.

Maybe sticking with the original "rows" name would have been better? Or summary_card_rows? (to be more explicit that these are the rows we're going to pass into the summary_card component, as opposed to things we're going to place in the details element)

richardTowers (approved):
I think mostly just nitpicks. Main substantive comment is on the naming of the block_rows method - if you agree with me that it's a bit confusing, it's probably worth tidying that up. Doesn't need to block the merge if you don't agree though.

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.html.erb
pezholio: Yeah, I was bashing my head against a brick wall with this for a while - turns out if you're rendering details within a component, then you need to use `capture`, otherwise the content of the block gets rendered outside the details component.

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
pezholio: Good shout 👍

PR #10288, opened 2025-06-13
Add a db migration to support flexible pages for the edition model

ryanb-gds:
Trello: https://trello.com/c/eR0sZZ23


db/migrate/20250610095041_add_flexible_page_content_to_editions.rb
```diff
@@ -0,0 +1,9 @@
+class AddFlexiblePageContentToEditions < ActiveRecord::Migration[8.0]
+  def change
+    change_table :editions, bulk: true do |t|
```
ChrisBAshton, line 3: Did we consider using the existing `type` property for the flexible page type, and reusing a `body` property for storing the JSON? I think that was the original idea (for POC at least) and it would be good to expand on alternatives considered in the commit message.

Aside: not sure which 'body' property we'd have hooked into in the end - the one in govspeak_contents or the one in edition_translations? 🤔

ryanb-gds: I think the flexible page content should be on the edition translations table, you're right. I can try to use the existing body field if you want but I don't think saving us from creating one database column is really worth the possible downside of having to treat the body field conditionally depending on the edition type.

I'm also not convinced about using the `type` column for an attribute that only certain types of edition should have. I can't see a benefit to reuse here apart from saving us an index write on creation, which I don't think is worth the potential confusion.

ryanb-gds: Oooofff. The Globalize gem doesn't work with JSON columns https://github.com/globalize/globalize/issues/795

Guess we'll have to use a text column after all

ChrisBAshton: Oof - unexpected! Have documented that as a pain point in https://trello.com/c/urs5AgXp/798-globalize-gem-is-unsupported

ryanb-gds: Actually it seems to work fine - I thought the issue I was seeing was this but turned out to be a bug in my code, as usual 😅

PR #10290, opened 2025-06-13
(CM-159) Support objects and arrays in subschemas

pezholio:
Telephones need multiple “telephone numbers”, so we need a way for CBM to support arrays of objects using the “Add another” pattern.

This adds new components for both objects and arrays, as well as updating the `permitted_params` helpers for subschemas, so Rails will recognise the fields and allow them to be passed through.

The next step is getting these array items to display correctly in the summary lists, but I'll handle this in a separate PR.

## Screenshots

image

image

### DIFF
--- a/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/application.scss
+++ b/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/application.scss
@@ -1,3 +1,4 @@
+@import "components/array-component";
 @import "components/embedded-objects-blocks-component";
 @import "components/embedded-objects-metadata-component";
 @import "components/default-block-component";
--- a/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/components/_array-component.scss
+++ b/lib/engines/content_block_manager/app/assets/stylesheets/content_block_manager/components/_array-component.scss
@@ -1,0 +1,16 @@
+.app-c-content-block-manager-array-component {
+  margin-bottom: govuk-spacing(6);
+
+  .gem-c-fieldset {
+    background-color: govuk-colour("light-grey");
+    padding-left: govuk-spacing(3);
+    padding-right: govuk-spacing(3);
+
+    padding-top: govuk-spacing(6);
+    padding-bottom: govuk-spacing(6);
+
+    .govuk-fieldset__legend {
+      margin-bottom: govuk-spacing(6);
+    }
+  }
+}
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array/item_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array/item_component.html.erb
@@ -1,0 +1,21 @@
+<% if array_items["type"] == "string" %>
+  <%= render "govuk_publishing_components/components/input", {
+    label: {
+      text: field_name.humanize,
+    },
+    name:,
+    id:,
+    value: field_value,
+  } %>
+<% elsif array_items["type"] == "object" %>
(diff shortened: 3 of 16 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_item_component.html.erb
```diff
@@ -0,0 +1,21 @@
+<% if array_items["type"] == "string" %>
+  <%= render "govuk_publishing_components/components/input", {
```
Harriethw: in a future PR, this could have something like the `component_for_field`, so that it can auto-render a component based on the type?

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_item_component.rb
```diff
@@ -0,0 +1,38 @@
+class ContentBlockManager::ContentBlockEdition::Details::Fields::ArrayItemComponent < ViewComponent::Base
```
Harriethw: vv small nitpick, could these components be grouped in a folder together, as guess you can't use the item without the array?

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_item_component.html.erb
pezholio: Yeah, I think that would work, I just didn't want to get too much in the weeds for now

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_item_component.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    label: {
+      text: field_name.humanize,
+    },
+    name:,
+    id:,
+    value: field_value,
+  } %>
+<% elsif array_items["type"] == "object" %>
```
Harriethw: I'm not sure I quite understand how validation is working, but I noticed if I put an incorrect value into the telephone `type` input I don't get a form validation error at the point of save and continue - I would have thought it should validate it against the schema then and there? Could be something we pick up in another PR but just flagging

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_item_component.rb
pezholio: Good call 👍

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_item_component.html.erb
pezholio: Yep, good idea. I'll add that as a separate task

Harriethw (approved):
🙌 looks really good and easy to follow. Love it
Just one other general note - I guess the schema changes on Publishing API will cause the e2e's to break, do you want to update the e2e for creating a Contact in this PR so it can just be merged quickly?

pezholio:
I don't think we have e2es for contacts yet do we?

Harriethw:
oh sorry I meant feature test!

Harriethw:
but actually maybe it won't break them cause we make the schema inside the test? But we should update it at some point to reflect the new schema...

pezholio:
Yeah, I think we should add the feature test once we've got a straight through journey

Harriethw:
I'm impressed that you were able to do this without a feature test tbh, I would have gone crazy just having to click through the form manually 😅

PR #10291, opened 2025-06-13
Allow users to select a flexible page type and create and edit flexible pages

ryanb-gds:
Trello: https://trello.com/c/t3Hosesy

Depends on #10288 

Note that because there isn't a Publishing API schema in place yet, attempting to save the page will fail.

### Implementation Notes

We had to introduce a way to override the new document selection process because flexible pages have an extra type selection step before the user reaches the new edition form. We therefore added the `choose_document_type_form_action` to the edition model, which is overridden for flexible pages to point to the 'choose_type' action on the flexible page controller.

We have assumed that flexible pages do not need the previously published input available for now, but it may be required later.


app/models/flexible_page_type.rb
```diff
@@ -0,0 +1,41 @@
...
+  class << self
+    attr_accessor :types
+  end
+
+  def self.boot
+    filenames = Dir.glob("app/models/flexible_page_types/*.json")
+    @types = filenames.each_with_object({}) do |filename, types|
+      type = JSON.parse(File.read(filename))
+      types[type["key"]] = type
+    end
```
ryanb-gds: Can anybody confirm if this is safe or not? In development the @types class instance variable kept being reset to an empty hash but I assume that was because Rails was hot reloading and it wouldn't happen in production. I'll give it a thorough workout in integration anyway.

ChrisBAshton: Even if it's only on happening on the dev environment, is that acceptable? I'd think probably not as it would really impede any dev work we do on flexible pages 🤔 

> We decided to load all the flexible page type configuration at application startup and persist it in memory. This allows us to provide a more ergonomic API for interacting with flexible page type definitions and allows the app to work without having to be aware of the file names on disk to load the type information.

I'm not sure I see much value in doing that vs having a runtime `def types` method with memoisation?

app/views/admin/flexible_pages/_form.html.erb
```diff
@@ -0,0 +1,49 @@
...
+          },
+          name: "edition[title]",
+          value: edition.title,
+          error_items: errors_for(edition.errors, :title),
+          right_to_left: form.object.translation_rtl?,
+          right_to_left_help: false,
+          rows: 1,
+        },
+        id: "edition_title",
+        maxlength: 65,
```
ChrisBAshton, line 20: Side note: it feels like we should be consolidating this maxlength value (or title input) in one template, as presumably we want the same title length limit across Whitehall. (Out of scope for PR but one to raise a backlog card for?)

test/unit/app/models/flexible_page_test.rb
```diff
@@ -0,0 +1,11 @@
+require "test_helper"
+
+class FlexiblePageTest < ActiveSupport::TestCase
+  test "does not require some of the standard edition fields" do
+    page = FlexiblePage.new
+    assert_not page.summary_required?
+    assert_not page.body_required?
+    assert_not page.can_set_previously_published?
+    assert_not page.previously_published
```
ChrisBAshton, line 9: Another out of scope observation: I had to do a little digging to figure out what this is. (My initial instinct was that it was something behavioural around whether or not the format allows ediitng of existing content, and my second guess was that it is something to do with the changelog in the sidebar)

 I wonder if it should be renamed to something like `overridden_first_published_date` and `can_override_first_published_date?` to better map to the input users see?

app/models/flexible_page_type.rb
ryanb-gds: I was trying to avoid putting the burden of reading and parsing the JSON files onto a user request. It'd be negligible at the moment, but if we had, say, 50 JSON files, it could make a noticeable difference. Happy to go with the memoised method for now

test/unit/app/models/flexible_page_test.rb
ryanb-gds: That probably would be a nice improvement

app/views/admin/flexible_pages/_form.html.erb
ChrisBAshton: https://github.com/alphagov/whitehall/pull/10305

test/unit/app/models/flexible_page_test.rb
ChrisBAshton: Had a quick go at this but I don't think I'd clocked that `previously_published` is a boolean and `first_published_at` is the actual datetime we want to use. So introducing `overridden_first_published_date` probably not helpful.

And if we're not doing that bit, probably also not helpful to rename `can_set_previously_published?` as we'd then be defining a `can_set_{x}?` method that refers to a property by a different name.

PR #10292, opened 2025-06-13
Render flexible page content inputs

ryanb-gds:
Trello: https://trello.com/c/t3Hosesy

Depends on #10291 

We are using [JSON schema](https://json-schema.org/docs) to configure the flexible page types, which allows us to easily validate the content against the schema.

We are currently only handling object and string inputs, we can add additional input types as we add more flexible page blocks. Objects automatically wrap their child inputs in a fieldset. We expect to provide an abstractions for defining these blocks in a later PR.

The flexible page helper uses recursion to walk the JSON schema tree and build the HTML form for the flexible content.


app/helpers/admin/flexible_page_helper.rb
```diff
@@ -0,0 +1,28 @@
+module Admin::FlexiblePageHelper
+  def render_flexible_page_content_fields(schema, edition, path = "", output_html = "")
+    schema["properties"].each do |key, property|
+      next_path = path.empty? ? key : "#{path}.#{key}"
+      case property["type"]
+      when "string"
+        output_html += render "govuk_publishing_components/components/input", {
```
ChrisBAshton, line 7: I'd move these `render` calls out into their own (private?) methods, in anticipation of this helper growing.

app/models/flexible_page.rb
```diff
@@ -28,4 +29,11 @@ def previously_published
   def rendering_app
     Whitehall::RenderingApp::FRONTEND
   end
+
+  def content_conforms_to_schema
+    type_instance = FlexiblePageType.find(flexible_page_type)
+    unless type_instance.validator.valid?(flexible_page_content)
+      errors.add(:flexible_page_content, "does not conform to schema for the #{type_instance.label}.")
```
ChrisBAshton, line 36: I wonder if we can surface more useful diagnostic info here? But that can come post-POC.

app/models/flexible_page_types/history_page.json
```diff
@@ -1,3 +1,31 @@
...
+          "heading_text": {
+            "title": "Heading text",
+            "description": "The main heading text for the page title",
+            "type": "string"
+          },
+          "context": {
+            "title": "Context",
+            "description": "Some additional context for the page title",
+            "type": "string"
+          }
```
ChrisBAshton, line 24: I think we're missing the `lead_paragraph` property?  See https://github.com/alphagov/govuk-rfcs/pull/182/files#diff-98512ba43959e2b34beee8dfddb363eb84dcfbac857576b27ecf04355fef21e3R135

app/helpers/admin/flexible_page_helper.rb
ryanb-gds: I am planning to develop a component abstraction around these, but I want a few more use cases such as images and the contents list within the code before trying to anticipate what the abstraction design needs to be.

app/models/flexible_page.rb
ryanb-gds: I think we can, the JSONschemer gem does expose the errors. I also marked it post POC

app/models/flexible_page_types/history_page.json
ryanb-gds: I don't think that's used anywhere, looking at the [history](https://www.gov.uk/government/history/10-downing-street) [pages](https://www.gov.uk/government/history/11-downing-street) [themselves](https://www.gov.uk/government/history/1-horse-guards-road). It is present on the [landing page](https://www.gov.uk/government/history), but that will have to be a separate schema file. We could get into defining sub-schemas and referencing them (so that we can share certain fieldsets easily across schemas, but that seems a bit ambitious for the PoC

app/models/flexible_page.rb
ryanb-gds: Wrote a ticket up for this: https://trello.com/c/aL0KSKiF

PR #10297, opened 2025-06-16
attempt to use locale in govspeak

patrickpatrickpatrick:

### DIFF
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -39,9 +39,10 @@ module GovspeakHelper
   def govspeak_html_attachment_to_html(html_attachment)
     # HTML attachments can embed images - but not attachments - from their parent Edition
     images = prepare_images(html_attachment.attachable.try(:images) || [])
+    locale = html_attachment.translated_locales.first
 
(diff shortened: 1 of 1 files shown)


GDSNewt (approved):

app/helpers/govspeak_helper.rb
```diff
@@ -38,9 +38,10 @@ def bare_govspeak_edition_to_html(edition)
   def govspeak_html_attachment_to_html(html_attachment)
     # HTML attachments can embed images - but not attachments - from their parent Edition
     images = prepare_images(html_attachment.attachable.try(:images) || [])
+    locale = html_attachment.translated_locales.first
```
ChrisBAshton, line 41: The only "?" here (from a cursory look) is the `.last` - do we ever expect to have to handle more than one locale in this helper? Are we missing something, eg potentially choosing the 'wrong' locale?

patrickpatrickpatrick: The HTML Attachment presenter derives the locale in the same way. The ensure that the locale used in the HTML and locale attribute match in the API response, I thought it was best to use the same logic here.

PR #10298, opened 2025-06-17
Add strong_migrations gem

ChrisBAshton:
Following the precedent set in https://github.com/alphagov/publishing-api/pull/3251, this PR adds the strong_migrations gem.

From the README:

>  Catch unsafe migrations in development
>
>    ✓  Detects potentially dangerous operations
>    ✓  Prevents them from running by default
>    ✓  Provides instructions on safer ways to do what you want

This should help prevent accidents like adding indexes non-concurrently,
which can lock tables and cause incidents.

Lots of information about what this tool does in the README:

https://github.com/ankane/strong_migrations/blob/master/README.md

--

### DIFF
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,0 +1,2 @@
+StrongMigrations.lock_timeout = 10.seconds
+StrongMigrations.statement_timeout = 1.hour
--- a/Gemfile
+++ b/Gemfile
@@ -56,6 +56,7 @@ gem "select_with_search_component", github: "alphagov/select-with-search-compone
 gem "sentry-sidekiq"
 gem "sidekiq-scheduler"
 gem "sprockets-rails"
+gem "strong_migrations"
 gem "terser"
 gem "transitions", require: ["transitions", "active_record/transitions"]
 gem "validates_email_format_of"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -995,6 +995,8 @@ GEM
     string_pattern (2.3.0)
       regexp_parser (~> 2.5, >= 2.5.0)
     stringio (3.1.7)
+    strong_migrations (2.3.0)
+      activerecord (>= 7)
     sync (0.5.0)
     sys-uname (1.3.1)
       ffi (~> 1.1)
@@ -1138,6 +1140,7 @@ DEPENDENCIES
   sidekiq-scheduler
   simplecov
   sprockets-rails
+  strong_migrations
   terser
   timecop
   transitions


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,2 @@
+StrongMigrations.lock_timeout = 10.seconds
+StrongMigrations.statement_timeout = 1.hour
```
ryanb-gds, line 5: Did we bump the migration kubernetes job timeout to an hour in the end?

To be honest I'm not sure what happens when a timeout occurs - presumably MySQL has received its instructions and continues with its operation regardless of whether or not a client is connected. Or does it roll back the transaction?

ryanb-gds (approved):
Nice. Bit of a shame a lot of the checks seem to be Postgres specific, but better than nothing

config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,2 @@
+StrongMigrations.lock_timeout = 10.seconds
```
richardTowers, line 4: You might want `StrongMigrations.start_after` too - I needed it in publishing-api 
https://github.com/alphagov/publishing-api/commit/734d6b9947e3f13ba4d15d1d81a2b4c036b12c72

ChrisBAshton: Thanks - done!

config/initializers/strong_migrations.rb
ChrisBAshton: Have now raised https://github.com/alphagov/govuk-helm-charts/pull/3395

> To be honest I'm not sure what happens when a timeout occurs - presumably MySQL has received its instructions and continues with its operation regardless of whether or not a client is connected. Or does it roll back the transaction?

(Full disclosure, this is according to ChatGPT...!)

> A lock timeout (controlled via lock_wait_timeout) causes MySQL to abort the waiting statement and raise error 1205. Only the statement is rolled back; the rest of the transaction stays open and must be committed or rolled back manually. A statement timeout (via max_execution_time) aborts the long‑running query. Again, only the timed‑out statement is undone. If autocommit is enabled then no explicit COMMIT is needed; otherwise you need to handle the transaction. If the client disconnects, MySQL generally finishes the running query and will commit or roll back depending on autocommit.

> In summary, the code in the PR correctly follows the strong‑migrations recommended defaults by using a very short lock timeout and a long statement timeout [github.com](https://github.com/ankane/strong_migrations/blob/master/README.md#:~:text=It%E2%80%99s%20extremely%20important%20to%20set,can%20run%20for%20a%20while). No changes are needed to the Ruby code; however, you do need to bump the Kubernetes migration job’s activeDeadlineSeconds to slightly over an hour to prevent the job runner from timing out before the migration completes.

PR #10307, opened 2025-06-17
Add new `/view` route to support "View {current state} edition" action

ChrisBAshton:
## What

Adds a "View {current state} edition" button in the sidebar, so that:

1. We can see the contents of a published Edition without creating a new edition(!).
2. We can easily see what the contents of superseded editions used to be.

See individual commits for details of which document type has this "View {state} edition" feature. Note that the Flexible Page type has been omitted, for now, as its implementation is liable to change quite rapidly in the coming weeks. 

The view reuses all the same partials as the "Edit" action, but with the publish/settings actions removed and with an overall `<fieldset>` wrapping the form inputs to force every input to be disabled / readonly. It does require some duplicate route definitions in routes.rb, because we can't use the `edition/{id}/view` slug because the form itself relies on knowing the content type, so the content type has to be part of the URL. The standard new/edit/show RESTful routes are included in Rails by default, so aren't explicitly defined in routes.rb, but the 'view' route needs defining manually.

Notably absent is the ability to view HTML attachments (or indeed any attachments) from this screen, but I figure that's something that can be added later in response to publisher demand.

NB, we don't add the button to the sidebar of _draft_ editions, since the publisher would just press the "Edit draft" button.

Trello: https://trello.com/c/wCnq2kT4/3836-add-view-current-edition-feature

## Why

Not being able to see an edition's GovSpeak contents impedes our ability to troubleshoot when responding to support requests. We have to go through the somewhat arduous process of switching to integration, so that we can take the non-idempotent action of creating a new edition.

Even worse is the experience of trying to find out the GovSpeak contents of past editions, which either means trying to construct the correct [diff url](https://github.com/alphagov/whitehall/blob/3aa1268058476ddfd04b135cc53cd342db51cdec/config/routes.rb#L210) to compare the superseded edition with the latest edition, or else fire up a Rails console to dig out the details manually. There is surely a tonne of benefit to more easily surfacing this information for publishers.

## Screenshots

Sidebar on superseded editions:
> sidebar on superseded editions

Sidebar on 'current' editions:
> sidebar on 'current' editions

Sidebar on draft/submitted/rejected editions (unchanged):
> sidebar on draft editions

Read-only form:
> read-only form)

---

### DIFF
--- a/app/components/admin/editions/show/sidebar_actions_component.rb
+++ b/app/components/admin/editions/show/sidebar_actions_component.rb
@@ -32,6 +32,7 @@ class Admin::Editions::Show::SidebarActionsComponent < ViewComponent::Base
       add_unpublish_action
       add_review_reminder_action
       add_view_action
+      add_view_on_website_action
       add_data_action
     end
     @actions
@@ -172,6 +173,16 @@ private
     end
   end
 
+  def add_view_action
+    if @enforcer.can?(:see) && %w[draft submitted rejected].exclude?(@edition.state)
+      actions << render("govuk_publishing_components/components/button", {
+        text: "View #{@edition.state} edition",
+        href: helpers.view_path_for_edition(@edition),
+        secondary_quiet: true,
+      })
+    end
+  end
+
   def add_review_reminder_action
     if @edition.publicly_visible? && @edition.document.latest_edition == @edition
       review_reminder = @edition.document.review_reminder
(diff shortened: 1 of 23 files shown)


ChrisBAshton:
Superseded by #10332, proposed by @ryanb-gds , which ended up a more elegant solution 👍

PR #10308, opened 2025-06-17
CM-155: description as textarea component

Harriethw:
- **set correct field order for Contact**
- **add a textarea component**
- **add description component configuration**

Screenshot 2025-06-17 at 16 56 35

### DIFF
--- a/lib/engines/content_block_manager/config/content_block_manager.yml
+++ b/lib/engines/content_block_manager/config/content_block_manager.yml
@@ -1,5 +1,9 @@
 schemas:
   content_block_pension:
+    fields:
+      description:
+        component:
+          textarea
     subschemas:
       rates:
         embeddable_fields:
@@ -15,6 +19,10 @@ schemas:
       - title
       - description
       - contact_type
+    fields:
+      description:
+        component:
+          textarea
     subschemas:
       email_addresses:
         group: modes


lib/engines/content_block_manager/config/content_block_manager.yml
```diff
@@ -1,5 +1,9 @@
 schemas:
   content_block_pension:
+    fields:
```
Harriethw, line 3: it's a bit of a shame that we have to add this twice... I think there is a better fix upstream. As we weren't sure if `description` would be a shared field on all blocks, we put it under `details` and it's generated programmatically from the schema. However,  `description` is actually a [top-level field in the Publishing API](https://github.com/alphagov/publishing-api/blob/4b63ff83ad6547c17be70b729357515406e91df7/content_schemas/formats/shared/default_format.jsonnet#L5) outside of `details`, so if we're certain it will be shared across all blocks we could use that.
I think that would need a bit of migrating of the existing Pension data though, and a bit of a rejig of the form (the Description component could be hardcoded like Instructions to Publishers). And we might not want to commit to it yet without experimenting with more types of blocks

pezholio: Yeah, I think for now the repetition is fine, but let's see where things go from here. Alternatively there could be a kind of root component rule that says "for all fields called description, render a textarea"

PR #10311, opened 2025-06-18
(CM-162) Handle validation of objects and arrays

pezholio:
This fixes the validator to ensure we can support nested objects and arrays, as well as updating the array component to ensure the errors show correctly. I've also done a bit of refactoring of the Cucumber steps to make them a bit easier to follow and iterate on.

### DIFF
--- a/lib/engines/content_block_manager/features/create_contact_object.feature
+++ b/lib/engines/content_block_manager/features/create_contact_object.feature
@@ -37,12 +37,36 @@ Feature: Create a contact object
     """
     {
       "type":"object",
-      "required": ["title", "telephone"],
+      "required": ["title", "telephone_numbers"],
       "properties": {
-        "title": {
-          "type": "string"
+        "telephone_numbers": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "required": [
+              "type",
+              "label",
+              "telephone_number"
+            ],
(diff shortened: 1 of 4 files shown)


lib/engines/content_block_manager/features/create_contact_object.feature
```diff
@@ -3,17 +3,51 @@ Feature: Create a contact object
...
-      | field            | type   | format | required | enum           | pattern          |
-      | title            | string | string | true     |                |                  |
-      | email_address    | string | string | true     |                |                  |
-    And the schema "contact" has a subschema with the name "telephones" and the following fields:
-      | field     | type   | format | required | enum           | pattern          |
-      | title     | string | string | true     |                |                  |
-      | telephone | string | string | true     |                |                  |
+    And a schema "contact" exists:
+    """
+    {
```
Harriethw, line 8: I like this change, it might be a bit more fiddly to setup but at least it makes it more transparent what we are actually doing. I guess if needed we could go one step further and have the json strings saved somewhere so we could re-use them?

pezholio: I did wonder that, but I wasn't sure if it'd make it less visible? We could give it a go though if it gets unweildy

PR #10321, opened 2025-06-19
Change GovSpeak Contact Validator to check HTML Attachment validation…

minhngocd:
… from parent edition

Before this change, the validator in triggered on the parent edition only on publish, however the validator on HTML attachment is triggered all the time, as there is no direct publish method on the HTML attachment. This means that validation happens all the time on HTML attachments. This is causing problems for publishers as when they are trying to create a new edition but the HTML attachment contains invalid GoVSpeak, it causes validation to fail on the parent edition and no new edition is allowed to be created. I'm solving this by getting the validator to check html attachments from the parent edition at the time of publish for the parent edition. We can't hook the validator for HTML attachment on anything as publishing for the attachments are triggered from Publishing Workeres, enqueued from pulishing parent editions.

https://trello.com/c/NIgzxsJF/3835-technical-fault-with-whitehall


ryanb-gds (approved):
Fix looks good, test failures appear unrelated. Personally I feel the edition test is duplicating the validator test really - one test on the edition model which proves the validator is configured appropriately would be sufficient.

minhngocd:
@ryanb-gds thanks 🙏  
I'm not fussed about taking either out - i added the validator tests from TDD and then added the edition tests because the validator ones don't actually check edition validation after publish.

PR #10325, opened 2025-06-20
Govspeak flexible page block

ryanb-gds:
## What

Add support for govspeak to flexible page schemas

## Implementation Notes

Out next move will be to create a cohesive block type abstraction so that we can keep the form input rendering and publishing API presentation behaviour for a block together. We'll put that in a separate PR.

Trello: https://trello.com/c/RLzz9VJe


ryanb-gds:
> Looks good, with the current flexible page types I don't think we need to be too concerned about testing massive schemas with dozens of properties...
> 
> I do wonder if a test around the sanitisation of the govspeak might be good? I can have a go at adding one? It's not a blocker though!

I think the Govspeak gem itself ought to have some sanitization tests, so hopefully we can just rely on that rather than adding the tests to Whitehall.

PR #10327, opened 2025-06-23
Use fake test types for flexible pages controller test

ryanb-gds:
We want to avoid relying on real flexible page types, so that changes to the type configuration won't break our tests.


minhngocd (approved):
Makes sense to me!

PR #10332, opened 2025-06-24
Add "View {current state} edition" action to editionable content

ChrisBAshton:
## What

Adds a "View {current state} edition" button in the sidebar, so that:

1. We can see the contents of a published Edition without creating a new edition(!).
2. We can easily see what the contents of superseded editions used to be.

The view reuses the "Edit" view, but with the publish/settings actions removed and with an overall `<fieldset>` wrapping the form inputs to force every input to be disabled / readonly. NB, we don't add the button to the sidebar of _draft_ editions, since the publisher would just press the "Edit draft" button. Note also that we've had to fix up some tests that were referring to the wrong path and relying on the redirect from `/edit` to the parent path: now that we're allowed to view a (readonly version of a) non-draft edition on the Edit screen, there's no longer a redirect and error message.

Notably absent is the ability to view HTML attachments (or indeed any attachments) from this screen, but I figure that's something that can be added later in response to publisher demand.

Trello: https://trello.com/c/wCnq2kT4/3836-add-view-current-edition-feature

## Why

Not being able to see an edition's GovSpeak contents impedes our ability to troubleshoot when responding to support requests. We have to go through the somewhat arduous process of switching to integration, so that we can take the non-idempotent action of creating a new edition.

Even worse is the experience of trying to find out the GovSpeak contents of past editions, which either means trying to construct the correct [diff url](https://github.com/alphagov/whitehall/blob/3aa1268058476ddfd04b135cc53cd342db51cdec/config/routes.rb#L210) to compare the superseded edition with the latest edition, or else fire up a Rails console to dig out the details manually. There is surely a tonne of benefit to more easily surfacing this information for publishers.

## Screenshots

Sidebar on superseded editions:
> sidebar on superseded editions

Sidebar on 'current' editions:
> sidebar on 'current' editions

Sidebar on draft/submitted/rejected editions (unchanged):
> sidebar on draft editions

Read-only form:
> read-only form)

---

### DIFF
--- a/test/support/admin_edition_controller_test_helpers.rb
+++ b/test/support/admin_edition_controller_test_helpers.rb
@@ -243,6 +243,7 @@ module AdminEditionControllerTestHelpers
 
         get :edit, params: { id: published_edition }
 
+        assert_select ".govspeak div[role='note']", text: "This is a read-only view of the current (published) edition. To edit, please return to the summary page and choose “Create new edition”."
         assert_select "form fieldset[disabled='disabled']" do
           assert_select "textarea[name='edition[body]']", published_edition.body
           assert_select "textarea[name='edition[summary]']", published_edition.summary
--- a/app/views/admin/editions/edit.html.erb
+++ b/app/views/admin/editions/edit.html.erb
@@ -37,6 +37,8 @@
         items: secondary_navigation_tabs_items(@edition, request.path),
       } if @edition.editable? %>
 
+      <%= govspeak_to_html("^This is a read-only view of the current (#{@edition.state}) edition. To edit, please return to the summary page and choose \"Create new edition\".^") unless @edition.editable? %>
+
       <%= render "form", edition: @edition %>
     </div>
 


test/support/admin_edition_controller_test_helpers.rb
```diff
@@ -753,32 +753,6 @@ def should_allow_role_appointments_for(edition_type)
       end
     end
 
-    def should_prevent_modification_of_unmodifiable(edition_type)
```
ChrisBAshton, line 756: The `prevent_modification_of_unmodifiable_edition` is still used by several (non-standalone Edition) controllers (such as Call for Evidence responses, or attachments) - but these don't appear to be tested. Could probably repurpose this deleted method to test those, but feels out of scope for this PR given I'm not making anything worse.

leimdorfer:
Couple of points:
1. The "View published edition" button might get confused with "View on Website" link. Had a chat with JP - we think "View edition" might be enough.
2. When the user is on the read only edition, there should be a message to say "This is a read-only view of the current published edition. To edit please return to the summary page and choose "Create new edition". Or something to that effect.

GDSNewt (approved):
I suppose the absence of buttons on the read only form should make it obvious that it is read only... my only thought is maybe adding a visual cue to make this more obvious. Not a dealbreaker, just trying to find things to suggest 😃. 

Also, I assume this change is fairly robust when it comes to users with different roles and permissions (e.g., managing editors, writers). That's the only edge cases scenario I can envision at the moment.

ChrisBAshton:
@leimdorfer 

> 1. The "View published edition" button might get confused with "View on Website" link. Had a chat with JP - we think "View edition" might be enough.

I'm not too sure about this, because it's not always just published editions. E.g. the "View scheduled edition" button feels more useful and clear than just "View edition" (which otherwise might imply viewing the currently live edition, for example). There's also something nice about being explicit about "View withdrawn edition" to make it clear we're viewing the contents of the edition at the point it was withdrawn, and likewise for unpublished editions.

What do you reckon? Merge as-is or make the button always "View edition"?

> 2. When the user is on the read only edition, there should be a message to say "This is a read-only view of the current published edition. To edit please return to the summary page and choose "Create new edition". Or something to that effect.

This is done now, and an updated screenshot provided 👍 

@GDSNewt 

> I assume this change is fairly robust when it comes to users with different roles and permissions (e.g., managing editors, writers). That's the only edge cases scenario I can envision at the moment.

Yeah, that should be covered by the [existing permissions enforced for the :edit route](https://github.com/alphagov/whitehall/blob/72a6951bcbc36715b0c74380462c53b8b2408a4c/app/controllers/admin/editions_controller.rb#L34-L35). But you did prompt me to take another look at the controller action and I've now added a test for the `open_for_editing_as` logic - the logic should be skipped when viewing the readonly form!

PR #10342, opened 2025-06-26
Flexible page block refactor

ryanb-gds:
This refactor provides a flexible page block abstraction which we can develop to add further flexible block types as user requirements emerge.


app/models/flexible_page_content_blocks/default_object.rb
```diff
@@ -0,0 +1,42 @@
...
+        id: path.form_control_id,
+      }) do
+        output = ""
+        schema["properties"].each do |property_key, property_schema|
+          block = Factory.build(property_schema["type"], property_schema["format"] || "default")
+          safe_content = content && Context.renderer.sanitize(content[property_key])
+          property_required = schema["required"].include?(property_key)
+          output += block.render(property_schema, safe_content, path.push(property_key), required: property_required)
+        end
+        # Rails recommends using sanitize instead of html_safe, but it removes a lot of the HTML and I don't know why
```
patrickpatrickpatrick, line 39: `sanitize` requires a list of permitted attributes and tags, otherwise it'll strip them. Not sure if this is globally configurable somewhere but I've had to do this before when calling this function.

ryanb-gds: Ah, that'll be the problem. I'll have a quick look into that, thanks @patrickpatrickpatrick

app/models/flexible_page_content_blocks/default_object.rb
```diff
@@ -0,0 +1,46 @@
+module FlexiblePageContentBlocks
```
lauraghiorghisor-tw, line 1: Ryan and I discussed the dual use of this default object - i.e. it is used to render the entire schema, as well as subsequent "object" blocks. In the next [refactor PR](https://github.com/alphagov/whitehall/pull/10352/files) it becomes a bit clearer that there are distinct behaviours for the "root" object.

app/models/flexible_page_content_blocks/default_object.rb
lauraghiorghisor-tw: I made a note of this in the last commit - I decided to actually bring over most of the tests in the [next PR](https://github.com/alphagov/whitehall/pull/10352/files) here (we really should have had them here already) - and skip the failing one. I think it makes this PR more complete (albeit buggy) and it makes the next one clearer in scope.

app/models/flexible_page_content_blocks/image_select.rb
```diff
@@ -0,0 +1,62 @@
...
+      return nil if content.blank?
+
+      if (selected_image = Context.page.images.find { |image| image.id == content.to_i })
+        {
+          src: selected_image.url,
+          caption: selected_image.caption,
+          alt_text: selected_image.alt_text,
+        }
+      else
+        Rails.logger.warn("Flexible page with ID #{Context.page.id} does not have an image with ID #{content}, so the image has been excluded from the Publishing API payload.")
```
lauraghiorghisor-tw, line 29: Fixes this error we saw earlier today https://govuk.sentry.io/issues/6717510802/?alert_rule_id=963054&alert_type=issue&notification_uuid=c40e8d06-beb4-4cbe-9a43-6ee6802c5c22&project=202259&referrer=slack

lauraghiorghisor-tw (commented):
LGTM but will open it for review by other team members.

app/models/flexible_page_content_blocks/default_object.rb
patrickpatrickpatrick: As this isn't 'live', that makes sense to me!

PR #10343, opened 2025-06-26
Flexible page image block

ryanb-gds:
Trello: https://trello.com/c/kVJ6zEW1

This is a first implementation of the image flexible block.

It provides a drop down to select one of the images that has been uploaded to the edition using the images tab. We also add a setting to the flexible page config files so that the images tab can be made available to publishers on particular flexible page types.

This block required a change to some of the APIs, because it needs the entire edition as context so that the images can be loaded from the edition images association.

In a [future PR](https://github.com/alphagov/whitehall/pull/10342) I am going to factor the blocks out into separate classes so that all behaviour relating to a block is located in one place.

### DIFF
--- a/app/helpers/admin/flexible_page_helper.rb
+++ b/app/helpers/admin/flexible_page_helper.rb
@@ -14,8 +14,28 @@ private
 
     case property["type"]
     when "string"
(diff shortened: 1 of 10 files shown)


lauraghiorghisor-tw (approved):
LGTM

PR #10351, opened 2025-06-27
HMRC internal manuals finder

richardTowers:
This is a new finder that lets users search through all HMRC manuals / manual sections.

I've been working behind the scenes with HMRC to agree the content and settings for this, which is now done. We're planning on deploying it to production as a trial, and HMRC are going to get feedback on its effectiveness from external the users of their manuals.

As this is only configuration, and I haven't had to change any code or reindex anything in search, I don't think this will add to our maintenance burden significantly. I've also done everything I can to manage down HMRC's expectations of ongoing support - they understand that there's no promise of ongoing feature improvements to this finder, and they're still keen to have it over the current situation (where it's difficult to search across HMRC manuals).

[You can currently see the finder deployed to integration here](https://www.integration.publishing.service.gov.uk/government/hmrc-manuals-search)

### DIFF
--- a/lib/finders/hmrc_internal_manuals.json
+++ b/lib/finders/hmrc_internal_manuals.json
@@ -1,0 +1,69 @@
+{
+  "base_path": "/government/hmrc-manuals-search",
+  "title": "HMRC manuals",
+  "description": "Find HMRC technical guidance.",
+  "locale": "en",
+  "document_type": "finder",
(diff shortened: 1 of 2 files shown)


richardTowers:
Going to move this to finder-frontend, as per https://gov-uk.atlassian.net/browse/WHIT-1833

PR #10352, opened 2025-06-27
[WHIT-1472 ] Use templates to render flexible page content blocks

ryanb-gds:
## Implementation notes 

The previous rendering solution for content blocks risked introducing XSS attack vectors through its use of `html_safe`. This solution makes use of erb templates to render the content blocks, which escape all content by default. It also saves us from passing the view context around in the block context, which was a bit of a hack anyway. This is a safer and more "Railsy" approach to rendering the content blocks.

This also fixes a known bug with the sanitization code we have previously - see details in [previous PR](https://github.com/alphagov/whitehall/pull/10342).

## Testing

Also did a wrap-up manual test in integration. We cannot test against Publishing API's `main` code as of now, as the Whitehall changes invalidate the schema there, so we tested with allowed additional properties in the schema (see branch `allow-additional-properties-for-history-page`).

- Saved successfully 
Screenshot 2025-06-30 at 18 22 07

- Full journey currently not working (expected)
Screenshot 2025-06-30 at 18 23 08

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-1472)


app/views/admin/flexible_pages/content_blocks/_default_object.html.erb
```diff
@@ -0,0 +1,30 @@
+<% required = required || false %>
```
lauraghiorghisor-tw, line 1: I think we can rewrite this whole thing as 
```
<%= render "govuk_publishing_components/components/fieldset", {
  legend_text: if root
                 ""
               else
                 schema["title"] + (required ? " (required)" : "")
               end,
  heading_size: "m",
  id: path.form_control_id,
} do %>
  <% schema["properties"].each do |property_key, property_schema| %>
    <% block = FlexiblePageContentBlocks::Factory.build(property_schema["type"], property_schema["format"] || "default") %>
    <%= render block.to_partial_path, {
      schema: property_schema,
      content: content && content[property_key],
      path: path.push(property_key),
      required: schema["required"].present? ? schema["required"].include?(property_key) : false,
    } %>
  <% end %>
<% end %>
```
- It would wrap the schema block in an unnecessary?  fieldset with `form_control_id` of  `      "edition_flexible_page_content_` (as `Path.segments` is [])   - which I don't think is a problem? Looks the same on the page.

test/unit/app/models/flexible_page_content_blocks/default_object_test.rb
```diff
@@ -63,12 +60,11 @@ class FlexiblePageContentBlocks::DefaultObjectRenderingTest < ActionView::TestCa
       },
       "required" => %w[test_attribute],
     }
-    render html: FlexiblePageContentBlocks::DefaultObject.new.render(schema, {})
+    render partial: "admin/flexible_pages/content_blocks/default_object", locals: { schema:, content: {}, path: Path.new, required: false, root: true }
     refute_dom "legend", text: schema["title"]
   end
 
   test "it renders non-required child attribute" do
```
lauraghiorghisor-tw, line 67: I've been thinking about the language we use here - all these things we're calling attributes are probably best called properties. It's not incomprehensible though.

app/views/admin/flexible_pages/content_blocks/_default_object.html.erb
ChrisBAshton: Can we not do something like this, which avoids the duplication but also avoids creating an unnecessary fieldset?

```ruby
<% content = content || {} %>
<% root = root || false %>

<% content_for :properties_loop do %>
  <% schema["properties"].each do |property_key, property_schema| %>
    <% block = FlexiblePageContentBlocks::Factory.build(property_schema["type"], property_schema["format"] || "default") %>
    <%= render block.to_partial_path, {
      schema: property_schema,
      content: content[property_key],
      path: path.push(property_key),
      required: schema["required"]&.include?(property_key) || false,
    } %>
  <% end %>
<% end %>

<% unless root %>
  <%= render "govuk_publishing_components/components/fieldset", {
    legend_text: schema["title"] + (required ? " (required)" : ""),
    heading_size: "m",
    id: path.form_control_id,
  } do %>
    <%= yield :properties_loop %>
  <% end %>
<% else %>
  <%= yield :properties_loop %>
<% end %>
```

PR #10356, opened 2025-06-30
(CM-197) Auto populate the label for a telephone number

pezholio:
When selecting a "Type" for a telephone number, the "label" will usually be the same as the type. This gives a subtle hint to the user by autopopulating the label. They will then be able to change it if they need to.

Untitled


edavey (approved):
This general purpose data-attribute mechanism looks very handy. 

I hope the rather complex JS for manipulating the form for the telephone label isn't tricky to maintain.

PR #10367, opened 2025-07-01
(CM-169) Disable deletion of array items which have already been published

pezholio:
When an object is published, we should treat any objects within an array as immutable, as these have embed codes that may be used on pages.

This PR fixes some of the deletion functionality, so it works as expected, as well as disabling the ability to delete any telephone number or opening hour objects that have already been published.


lib/engines/content_block_manager/test/unit/app/models/content_block_edition_test.rb
```diff
@@ -216,6 +216,27 @@ class ContentBlockManager::ContentBlockEditionTest < ActiveSupport::TestCase
 
       assert_equal content_block_edition.details["something"], { "random-string" => { "something" => "else" } }
     end
+
+    it "removes deleted items from the array" do
```
edavey: is it worth including in the narrative the fact that the yes/no `_destroy` marker is also removed from all items? e.g.

`it "removes deleted items from the array (as well as stripping out all _destroy markers)" do`

lib/engines/content_block_manager/test/components/content_block/edition/details/embedded_objects/form_component_test.rb
```diff
@@ -122,4 +122,60 @@ class ContentBlockManager::ContentBlockEdition::Details::EmbeddedObjects::FormCo
 
     render_inline(component)
   end
+
+  it "sends the object_title if provided" do
+    object_title = "something"
```
edavey, line 127: If I understand correctly, this is always the subschema `block_type`? A value like `"email_addresses"`?

If so maybe it would be helpful (to the uninitiated) to make this explicit in the spec e.g

```rb
it "sends the object_title (subschema 'block_type') if provided" do
    object_title = "email_addresses"
```

lib/engines/content_block_manager/test/components/content_block/edition/details/fields/array_component_test.rb
```diff
@@ -64,8 +64,72 @@ class ContentBlockManager::ContentBlockEdition::Details::Fields::ArrayComponentT
...
+    describe "all items have previously been published" do
+      let(:details) do
+        {
+          object_title => {
+            field.name => field_value,
+          },
+        }
+      end
+
+      it "does not render any checkboxes" do
```
edavey: it might be worth being more explicit that we're talking about *deletion* checkboxes? e.g.
```
it "does not render an deletion checkboxes do
```

lib/engines/content_block_manager/test/components/content_block/edition/details/fields/array_component_test.rb
```diff
@@ -64,8 +64,72 @@ class ContentBlockManager::ContentBlockEdition::Details::Fields::ArrayComponentT
...
+    let(:field_value) { %w[foo bar] }
+    let(:object_title) { "field" }
+
+    let(:latest_edition) { build(:content_block_edition, :contact, details:) }
+
+    before do
+      content_block_edition.document.stubs(:latest_edition).returns(latest_edition)
+    end
+
+    describe "all items have previously been published" do
```
edavey: just a tiny typo here I think? We're adding an additional context where our example consists of published objects? 

```
describe "when all items have previously been published" do
```

lib/engines/content_block_manager/test/components/content_block/edition/details/fields/array_component_test.rb
```diff
@@ -64,8 +64,72 @@ class ContentBlockManager::ContentBlockEdition::Details::Fields::ArrayComponentT
...
+    describe "one item has been previously published" do
+      let(:details) do
+        {
+          object_title => {
+            field.name => %w[foo],
+          },
+        }
+      end
+
+      it "renders a checkbox for the second item" do
```
edavey: again, it may be worth drawing attention to the fact that we're describing the appearance or non-appearance of *deletion* checkboxes? e.g.

```
it "renders a deletion checkbox for the unpublished item only" do
```

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_component.rb
```diff
@@ -45,10 +45,31 @@ def component(index)
...
+    else
+      hidden_field_tag(field_name, 0)
+    end
   end
 
   def errors
     content_block_edition.errors
   end
+
+  def can_be_deleted?(index)
```
edavey, line 61: Perhaps the fact that this critical `can_be_deleted?` method is private and therefore not easily unit-testable is telling us that we ought to consider refactoring to extract this deletion-related logic into a new class which has this single important responsibility? 

Perhaps that could be something like a `EmbeddedObjectImmutabilityCheck.new(field:, subschema:, document:)`

Just an idea...

lib/engines/content_block_manager/test/components/content_block/edition/details/fields/array/array_item_component_test.rb
```diff
@@ -32,30 +34,52 @@ class ContentBlockManager::ContentBlockEdition::Details::Fields::Array::ItemComp
...
 
-    it "renders a text field" do
-      render_inline(component)
+    describe "if the item can be deleted" do
+      let(:can_be_deleted) { true }
 
-      assert_selector "label", text: "Bar"
-      assert_selector "input[type='text'][value='Some text'][name='foo[bar][]'][id='foo_bar_1']"
-    end
+      it "renders with the correct class" do
```
edavey: perhaps it's worth being explicit here about the presence / absence of the `--immutable` class and it's purpose? e.g.

```
it "renders with the '--immutable' class to display the 'delete' button" do
```

lib/engines/content_block_manager/test/components/content_block/edition/details/fields/array/array_item_component_test.rb
```diff
@@ -32,30 +34,52 @@ class ContentBlockManager::ContentBlockEdition::Details::Fields::Array::ItemComp
...
+            form_group.assert_selector ".govuk-error-message", text: "Bar cannot be blank"
+          end
         end
       end
     end
+
+    describe "if the item cannot be deleted" do
+      let(:can_be_deleted) { false }
+
+      it "renders with the correct class" do
```
edavey: again, perhaps I'd find it easier if we spell out explicitly which class we expect and why, e.g

```
it "renders without the '--immutable' class, to suppress the 'delete' button" do
```

edavey (approved):
I _think_ I follow this: thanks for laying out the commits so carefully.

I do feel that it's not ideal to have to rely on logic in the code to protect the integrity of published content. I wonder if we ought to think about how we might model edition details in a way where we can record their "published" state in the db. 

No doubt this is very naive and probably I'm misunderstanding a good deal but something like e.g. 

```
EditionDetail
  edition_id
  field_name
  block_type
  published_at
```

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/array_component.rb
pezholio: Good plan! Have added that 👍

PR #10374, opened 2025-07-03
CM 170: Extend the email address object within the "Contact" content block

edavey:
Extend the email address object within the "Contact" content block

In this PR we add 2 new fields to the email address objects which can be embedded as "modes of contact" within a "Contact" content block. In addition, we:

- make the `description` use a text area rather than a regular text input field
- define a `field_order` for the email address fields in the form
- use that `field_order` to order the address fields when displayed in the metadata component's summary component

This work depends on changes to the Publishing API: see [PR 3434](https://github.com/alphagov/publishing-api/pull/3434) for additions to the content schema at `content_block_contact.jsonnet`

## Form for adding / editing an embedded email address

form_for_extended_email

## Display of the email address within a "Contact" content block

email_address_content_block

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/metadata_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/metadata_component.rb
@@ -11,11 +11,29 @@ private
   attr_reader :items
 
   def rows
+    unordered_rows.sort_by { |row| row_ordering_rule(row) }
+  end
+
+  def unordered_rows
     items.map do |key, value|
       {
         field: humanized_label(key, @object_type),
         value: translated_value(value),
       }
     end
+  end
+
+  def row_ordering_rule(row)
+    if field_order
+      # If a field order is found in the config, order by the index. If a field is not found, put it to the end
+      field_order.index(row.fetch(:field).downcase) || Float::INFINITY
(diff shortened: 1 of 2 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/metadata_component.rb
```diff
@@ -11,11 +11,29 @@ def initialize(items:, object_type:, schema:)
...
         field: humanized_label(key, @object_type),
         value: translated_value(value),
       }
     end
   end
+
+  def row_ordering_rule(row)
+    if field_order
+      # If a field order is found in the config, order by the index. If a field is not found, put it to the end
+      field_order.index(row.fetch(:field).downcase) || 99
```
pezholio: Using `Float::INFINITY` here might be better than having a magic number (`99`)

pezholio (approved):
LGTM - just one minor suggestion (I've done this in other places, so makes sense to be consistent)

lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/metadata_component.rb
edavey: 👍 thks, have made that change

PR #10380, opened 2025-07-07
Add `revalidation_passed` property to Editions (WHIT-1507)

ChrisBAshton:
We know that there are a number of editions in Whitehall that are no longer considered 'valid', e.g. they embed a Contact that no longer exists. It's currently very difficult to get a handle on the numbers of invalid editions as edition validity is somewhat expensive to compute and can't be quickly enumerated on the rails console.

We're therefore introducing a cache boolean which we will default to `true` (it is fair to assume that all past editions are valid, for now, as they were obviously valid at one point). We will then, in a future PR:

1. Update the boolean on a per-edition basis, at the point of validation (so we can slowly build up a better picture of the number of invalid editions just through normal publisher interaction).
2. Update the booleans in bulk by looping through all editions.

The 'invalid editions' count will be useful in reporting on Whitehall system health.

---

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_06_10_095041) do
+ActiveRecord::Schema[8.0].define(version: 2025_07_04_084716) do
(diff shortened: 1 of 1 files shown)


ChrisBAshton:
Superseded by #10391.

PR #10381, opened 2025-07-07
Dynamically update `revalidated_at` property (WHIT-1507)

ChrisBAshton:
Dependent on #10391. With that PR safely merged and the DB migration run, we are free to start referring to the new table property in application code. Spreading this across two PRs as we don't want to accidentally deploy code that refers to DB properties that don't yet exist - [we've been bitten by that before](https://docs.google.com/document/d/1GsLZTyuFBaAyt-VNFk2guV2XIzCoui1KNfe2_gN0lX4/edit?tab=t.0#heading=h.p99426yo0rbv).

This will cause the `revalidated_at` to be set to a timestamp if `valid?(:publish)` returns `true`, indicating whether or not the edition is considered valid to publish. We'll be able to query editions via this value, in a separate PR.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,6 @@
 /attachment-cache
 /*_import_errors.csv
 /bulk-upload-zip-file-tmp
-.yardoc/
 /doc/
 /failures
 /node_modules
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -73,7 +73,8 @@ class Edition < ApplicationRecord
 
   before_create :set_auth_bypass_id
   before_save :set_public_timestamp
-  after_validation :cache_revalidation_result
+  after_validation :update_revalidated_at, if: -> { validation_context == :publish }
+  after_commit :trigger_revalidation, on: %i[create update]
   after_create :update_document_edition_references
   after_update :update_document_edition_references, if: :saved_change_to_state?
 
@@ -108,20 +109,18 @@ class Edition < ApplicationRecord
     FROZEN_STATES.include?(state)
   end
 
-  def cache_revalidation_result
-    return unless new_record? || changed? || validation_context == :publish
+  def update_revalidated_at
+    new_value = errors.empty? ? Time.current : nil
 
-    # If there were no errors, the edition passed re-validation
-    new_timestamp = errors.empty? ? Time.current : nil
-
-    if validation_context == :publish && !changed?
-      # Someone called valid?(:publish) on an already-saved record
-      # There may be no subsequent save, so hit the DB directly
-      update_column(:revalidated_at, new_timestamp)
+    if persisted?
+      update_column(:revalidated_at, new_value)
     else
-      # We’re inside a create/update flow; assigning is enough:
-      self.revalidated_at = new_timestamp # will be persisted with the save
+      self.revalidated_at = new_value
     end
+  end
+
+  def trigger_revalidation
+    valid?(:publish)
   end
 
   def unmodifiable?
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -109,13 +109,18 @@ class Edition < ApplicationRecord
   end
 
   def cache_revalidation_result
-    # Set `nil` if invalid edition, otherwise use current timestamp
-    value_to_set = errors.empty? ? Time.zone.now : nil
+    return unless new_record? || changed? || validation_context == :publish
 
-    if new_record?
-      self.revalidated_at = value_to_set
+    # If there were no errors, the edition passed re-validation
+    new_timestamp = errors.empty? ? Time.current : nil
+
+    if validation_context == :publish && !changed?
+      # Someone called valid?(:publish) on an already-saved record
+      # There may be no subsequent save, so hit the DB directly
+      update_column(:revalidated_at, new_timestamp)
     else
-      update_column(:revalidated_at, value_to_set)
+      # We’re inside a create/update flow; assigning is enough:
+      self.revalidated_at = new_timestamp # will be persisted with the save
     end
   end
 
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -73,7 +73,7 @@ class Edition < ApplicationRecord
 
   before_create :set_auth_bypass_id
   before_save :set_public_timestamp
-  after_create :cache_revalidation_result
+  after_validation :cache_revalidation_result
   after_create :update_document_edition_references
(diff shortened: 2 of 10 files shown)


app/models/edition.rb
```diff
@@ -73,12 +73,10 @@ class Edition < ApplicationRecord
   POST_PUBLICATION_STATES = %w[published superseded withdrawn unpublished].freeze
   PUBLICLY_VISIBLE_STATES = %w[published withdrawn].freeze
 
-  # @!group Callbacks
```
lauraghiorghisor-tw, line 76: It appears to be a syntax for the documentation tool [yard](https://rubygems.org/gems/yard) (according to copilot, thanks, copilot) - and we have some stuff related to it including this [options file](https://github.com/alphagov/whitehall/blob/main/.yardopts) and a [gitignore mention](https://github.com/alphagov/whitehall/blob/02cf8bdce70460f6e972f7bde277addab0a60964/.gitignore#L27).

It seems to use tags like @param and @return (look for `# @param` or `# @return` in the codebase, they exist) and ofc this directive thing that this commit deletes, and others, haven't dug in too deep.

We do not have a gem nor a npm package so I imagine then that we can just delete all yard references? Maybe worth making sure it's not used and then cleaning it all up (including all these other references) in a separate PR.

test/unit/app/models/edition_test.rb
```diff
@@ -978,6 +978,47 @@ class EditionTest < ActiveSupport::TestCase
     assert_nil edition.revalidated_at
   end
 
+  test "should update cached 'revalidated_at' value on each `valid?` call" do
```
lauraghiorghisor-tw: One could argue that the validation test is enough (no need to cover create and update separately) - but I won't argue that 😅

app/models/edition.rb
```diff
@@ -111,13 +111,18 @@ def skip_main_validation?
   end
 
   def cache_revalidation_result
-    # Set `nil` if invalid edition, otherwise use current timestamp
-    value_to_set = errors.empty? ? Time.zone.now : nil
+    return unless new_record? || changed? || validation_context == :publish
```
lauraghiorghisor-tw: We should only be in the code after the return, if `changed?` is true but then the first `if` checks that it's not changed. That's confusing.

I'm half thinking it might be clearer to say which cases are we actively excluding? If there is an update with no change, no?  But then that's the case we expect to be in for the publish context. Maybe remove the changed? check then.


So the idea here is that all these validators run `on: publish` and that's going to set the context for this logic here? 🤔

test/unit/app/models/edition_test.rb
```diff
@@ -978,25 +979,50 @@ class EditionTest < ActiveSupport::TestCase
...
+    assert_not edition.valid?(:publish)
+    # It should reset `revalidated_at` to nil
     assert_nil edition.revalidated_at
-
-    # And ensure the updated value is persisted to the DB
+    # ...and ensure the updated value is persisted to the DB
     assert_nil edition.reload.revalidated_at
   end
 
+  test "should not update cached 'revalidated_at' value on each general `valid?` call" do
```
lauraghiorghisor-tw: The naming is a bit confusing here, especially when you read the next test, because technically we do update it outside the publish context too... It's more like specific publish-time validators only reset the timestamp on publish? 🤔

test/unit/app/models/edition_test.rb
ChrisBAshton: One probably could argue that... it was more that I didn't want to have a case of someone creating a new edition, going back to the search page and then finding their edition is flagged as invalid just because the `revalidate_at` property is still `nil`. In practice that shouldn't really happen once #10384 is merged, where we call the `valid?(:publish)` method on page load. Until then, there's an outside possibility of it, so felt safer to account for it here 🤷

app/models/edition.rb
ChrisBAshton: Thanks - I'll fix up those removals to this commit. The reason I'm doing it here (even though it's not really related to the PR) is that I'm adding an `after_validation` callback in the same area of code and I had to make a decision about whether or not to include it inside the weird 'group' comment block of callbacks. So felt it better to remove that complexity first. 👍

app/models/edition.rb
ChrisBAshton: Thanks! Your comment led to me discovering an underlying bug. I've fixed the bug and refactored this code (and the tests) in a [new commit](https://github.com/alphagov/whitehall/pull/10381/commits/ce916f1c3fc1984310ca43bd2f56633b8de9fb58).

test/unit/app/models/edition_test.rb
ChrisBAshton: Fixed in https://github.com/alphagov/whitehall/pull/10381/commits/ce916f1c3fc1984310ca43bd2f56633b8de9fb58

.gitignore
```diff
@@ -24,7 +24,6 @@
 /attachment-cache
 /*_import_errors.csv
 /bulk-upload-zip-file-tmp
-.yardoc/
```
lauraghiorghisor-tw, line 27: For this commit or elsewhere - all occurrences of @abstract, @return, @param:
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/lib/whitehall/decorators/collection_decorator.rb#L14 & others in this file
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/lib/whitehall/decorators/collection_decorator.rb#L32 & others in this file
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/lib/taxonomy/taxon.rb#L52 and others in this file
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/test/support/view_rendering.rb#L55 and other lines
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/app/models/role.rb#L1
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/app/models/announcement.rb#L1
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/app/models/edition.rb#L2
https://github.com/alphagov/whitehall/blob/974eb5e8d53f500e61d2d4d8c4b776514bfba2c9/app/models/publicationesque.rb#L9

PR #10382, opened 2025-07-07
Bulk update `revalidated_at` for all applicable Editions (WHIT-1507)

ChrisBAshton:
## What

Every edition has a `revalidated_at` timestamp, which is currently nil, as the property was added in #10391. In #10381, we added logic to overwrite the property with the current timestamp whenever `valid?(:publish)` is called (and succeeds). We now want to revalidate all of Whitehall's editions (except for deleted, superseded and unpublished editions) in bulk, via a worker running on a cron. This PR adds that worker and cron configuration.

## Why

This will accommodate changing circumstances in future (e.g. a more strict GovSpeak validator, or a situation where an embedded Contact has been deleted): it would revalidate all of the relevant Whitehall editions within roughly one week.

We'll then be able to easily grab all the invalid editions (`Edition.only_invalid_editions`) and report on that as a system health metric. We'll also be able to surface invalid editions for publishers, for them to fix up (in #10383).

## Testing on Integration

```
$ RevalidateEditionsSchedulerWorker.new.perform
# enqueues lots of workers...

$ puts Sidekiq::Queue.new("edition_revalidation").size; puts Edition.only_invalid_editions.count
1000 # jobs
398974 # editions where `revalidated_at` is `nil` - i.e. all of them

# a bit later... (and after running `RevalidateEditionsSchedulerWorker.new.perform` a couple more times)
$ puts Sidekiq::Queue.new("edition_revalidation").size; puts Edition.only_invalid_editions.count
0
213841

# and finally...
$ puts Sidekiq::Queue.new("edition_revalidation").size; puts Edition.only_invalid_editions.count
0
48045
```

Publishing API seemed to cope fine with this bulk-enqueuing:

Screenshot 2025-07-14 at 10 25 38
Screenshot 2025-07-14 at 10 26 00

---

### DIFF
--- a/app/sidekiq/revalidate_editions_scheduler_worker.rb
+++ b/app/sidekiq/revalidate_editions_scheduler_worker.rb
@@ -1,0 +1,24 @@
+class RevalidateEditionsSchedulerWorker
+  include Sidekiq::Job
+  sidekiq_options queue: "edition_revalidation", retry: 0
+
+  BATCH_SIZE  = ENV.fetch("REVALIDATE_BATCH_SIZE", 100).to_i
+  MAX_BATCHES = ENV.fetch("REVALIDATE_MAX_BATCHES", 1000).to_i # 1000 × 100 = 100 000 editions/run
+  def self.stale_after
+    1.week.ago
+  end
+
+  def perform
+    scope = Edition.not_validated_since(1.week.ago.strftime("%d/%m/%Y"))
+    logger.info("[RevalidateEditionsSchedulerWorker] #{scope.count} editions need revalidating")
+
+    # Pull batches in random order so permanently invalid editions
+    # don't prevent us from revalidating 'old' revalidated editions
+    scope.order(Arel.sql("RAND()")) # MySQL / MariaDB
+         .limit(BATCH_SIZE * MAX_BATCHES)
+         .pluck(:id)
+         .each_slice(BATCH_SIZE) do |ids|
+      RevalidateEditionBatchWorker.perform_async(ids)
+    end
+  end
+end
--- a/app/sidekiq/revalidate_edition_batch_worker.rb
+++ b/app/sidekiq/revalidate_edition_batch_worker.rb
@@ -1,0 +1,10 @@
+class RevalidateEditionBatchWorker
+  include Sidekiq::Job
+  sidekiq_options queue: "edition_revalidation", retry: 0, backtrace: false
+
+  def perform(edition_ids)
+    Edition.where(id: edition_ids).find_each do |edition|
+      edition.valid?(:publish) # revalidates the edition and updates `revalidated_at`
+    end
+  end
+end
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -10,8 +10,12 @@
   - bulk_republishing
   - link_checks
   - asset_migration
+  - edition_revalidation
 :scheduler:
   :schedule:
     revalidate_old_link_check_reports:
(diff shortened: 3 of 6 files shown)


ChrisBAshton:
> Might be overkill, and i'm not sure we test it anywhere else, but do we need a test around the batching? I.e can it handle large batches and doesn't fall over....

It's been manually tested on Integration and double-checked with Publishing Platform, so should be fine 👍

app/sidekiq/revalidate_editions_scheduler_worker.rb
```diff
@@ -0,0 +1,24 @@
+class RevalidateEditionsSchedulerWorker
+  include Sidekiq::Job
+  sidekiq_options queue: "edition_revalidation", retry: 0
+
+  BATCH_SIZE  = ENV.fetch("REVALIDATE_BATCH_SIZE", 100).to_i
+  MAX_BATCHES = ENV.fetch("REVALIDATE_MAX_BATCHES", 1000).to_i # 1000 × 100 = 100 000 editions/run
+  def self.stale_after
```
lauraghiorghisor-tw, line 7: How are we going to use this method?

app/sidekiq/revalidate_editions_scheduler_worker.rb
```diff
@@ -0,0 +1,24 @@
+class RevalidateEditionsSchedulerWorker
```
lauraghiorghisor-tw, line 1: I might be wrong but are these not meant to be all named "Job" now? Not that that's consistent with the rest of the code 😅

app/sidekiq/revalidate_editions_scheduler_worker.rb
```diff
@@ -0,0 +1,24 @@
...
+  include Sidekiq::Job
+  sidekiq_options queue: "edition_revalidation", retry: 0
+
+  BATCH_SIZE  = ENV.fetch("REVALIDATE_BATCH_SIZE", 100).to_i
```
lauraghiorghisor-tw, line 5: Would these realistically be set as env vars?

lauraghiorghisor-tw (commented):
Late drive-be comments sorrryyy

app/sidekiq/revalidate_editions_scheduler_worker.rb
ChrisBAshton: Quite right, on all counts, Laura! I'll tidy this up in another PR

app/sidekiq/revalidate_editions_scheduler_worker.rb
ChrisBAshton: You're right, but I'll leave this for now - I was just following the convention in the code. We should probably do a bulk renaming of all the workers / jobs at some point.

app/sidekiq/revalidate_editions_scheduler_worker.rb
ChrisBAshton: https://github.com/alphagov/whitehall/pull/10407

app/sidekiq/revalidate_editions_scheduler_worker.rb
ChrisBAshton: https://github.com/alphagov/whitehall/pull/10407

app/sidekiq/revalidate_editions_scheduler_worker.rb
ChrisBAshton: Yes, this is captured in https://gov-uk.atlassian.net/browse/WHIT-1836

PR #10383, opened 2025-07-07
Add "Only invalid editions" filter on Whitehall search (WHIT-1507)

ChrisBAshton:
Depends on the scope added in #10392, and the end-to-end journey enabled by #10384. This will allow publishers to easily see which of their documents needs editing, and what needs editing, respectively.

New option in sidebar:

"Only invalid editions" checkbox

Search results in action±:

search results showing issue with `Accelerator themed competition: the future of aviation security`

± after manually running `.valid?(:publish)` against an edition we know references a missing Contact.

---

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

This application is owned by the Whitehall Experience team. Please let us know in [#govuk-whitehall-experience-tech](https://gds.slack.com/archives/C02L13S214K) when you raise any PRs.

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #10384, opened 2025-07-07
Surface validation errors in the sidebar (WHIT-1507)

ChrisBAshton:
This PR adds a new "This edition is invalid" section to the sidebar, surfacing any and all± validation errors that would prevent publishing of the content. This section is also surfaced on earlier editions on the document (e.g. superseded ones) which provides a useful audit trail of whether or not past editions were considered 'valid'.

± Where validation fails on a dependency, such as an attachment (or attachment's attachment_data, or the attachment_data's asset), these are surfaced as a high level "Attachments is invalid" as per Rails' default. We're hoping to explore this further in https://gov-uk.atlassian.net/browse/WHIT-1517 to see if we can surface more specific validation messages in these cases at the Edition level.

It's worth noting that Whitehall did already have "This edition cannot be force-published" / "This edition cannot be scheduled" sections, which our new "This edition is invalid" section would now be duplicating. The former weren't backed up by any tests, were only visible in specific circumstances, and ultimately didn't flag all the publisher-blocking issues (see screenshots), so have been removed.

The context for this change is that in #10383, we're offering publishers a way of seeing all their 'invalid' editions directly from the search screen, but then it would be all too common for them to click through to an edition and see no issues until they try to edit or publish the content. We need a better journey for them to actually find out what the validation issues are, so we now run the `valid?(:publish)` checks on page render, and surface any validation issues in the sidebar.

## Screenshots

| Before | After |
|--------|------|
|whitehall-admin integration publishing service gov uk_government_admin_collections_502122|whitehall-admin dev gov uk_government_admin_collections_502122|
| No visible issues | Validation error now surfaced on the main screen, in the sidebar 🎉  |
|whitehall-admin integration publishing service gov uk_government_admin_news_505814|whitehall-admin dev gov uk_government_admin_news_505814|
| Visible issues ✅ | Visible issues, but now with proper tests 🎉  |

### DIFF
--- a/app/views/admin/editions/show/_sidebar_notices.html.erb
+++ b/app/views/admin/editions/show/_sidebar_notices.html.erb
@@ -6,6 +6,23 @@
 %>
 
 <div class="app-view-summary__sidebar-notices">
+  <% unless @edition.valid?(:publish) %>
+    <%= render "components/inset_prompt", {
+      data_attributes: {
+        module: "ga4-auto-tracker",
+        ga4_auto: {
+          type: "show invalid edition warning",
+          text: @edition.errors.map(&:full_message).join(", "),
+        }.merge(ga4_auto_attributes).as_json,
+      },
+      title: "This edition is invalid",
+      description: render("govuk_publishing_components/components/list", {
+        visible_counters: true,
+        items:  @edition.errors.map(&:full_message),
+      }),
+      error: true,
+    } %>
+  <% end %>
   <% if @edition.scheduled_publication %>
     <% if force_scheduler.can_transition? && !force_scheduler.can_perform? %>
       <%= render "components/inset_prompt", {
--- a/app/views/admin/editions/show/_sidebar_notices.html.erb
+++ b/app/views/admin/editions/show/_sidebar_notices.html.erb
@@ -23,44 +23,6 @@
       error: true,
     } %>
   <% end %>
-  <% if @edition.scheduled_publication %>
-    <% if force_scheduler.can_transition? && !force_scheduler.can_perform? %>
-      <%= render "components/inset_prompt", {
-        title: "This edition cannot be scheduled",
-        data_attributes: {
-          module: "ga4-auto-tracker",
-          ga4_auto: {
-            type: "scheduled",
-            text: force_scheduler.failure_reasons_plaintext,
-          }.merge(ga4_auto_attributes).as_json,
-        },
-        description: render("govuk_publishing_components/components/list", {
-          visible_counters: true,
-          items: force_scheduler.failure_reasons,
-        }),
-        error: true,
-      } %>
-    <% end %>
-  <% else %>
-    <% if force_publisher.can_transition? && !force_publisher.can_perform? %>
-      <%= render "components/inset_prompt", {
-        title: "This edition cannot be force-published",
-        module: "ga4-auto-tracker",
-        data_attributes: {
-          module: "ga4-auto-tracker",
-          ga4_auto: {
-            type: "force published",
-            text: force_scheduler.failure_reasons_plaintext,
-          }.merge(ga4_auto_attributes).as_json,
-        },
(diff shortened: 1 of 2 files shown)


joao-krauchenco-mottu (approved):

test/functional/admin/generic_editions_controller_test.rb
```diff
@@ -111,4 +112,20 @@ class Admin::GenericEditionsControllerTest < ActionController::TestCase
...
+  view_test "visiting an invalid edition should show the validation errors" do
+    bad_id = "9999999999999"
+    edition = create(:draft_edition, body: "[Contact:#{bad_id}]")
+    edition.save!(validate: false)
+    stub_publishing_api_expanded_links_with_taxons(edition.content_id, [])
+
+    get :show, params: { id: edition }
+
+    assert_select ".app-view-summary__sidebar .app-c-inset-prompt--error" do |elements|
+      expected_message = "This edition is invalid"
```
lauraghiorghisor-tw, line 125: Not PR-blocking. Musings:

Just to be the devil's advocate here... but there is some value in the kind of message we had before, where we included the exact action that was blocked. So, for example, a missing changenote blocks publishing (but you can save without it? I think?) - whereas now it's all grouped into one block of invalid things. 

Actually, saying that, for new editions, all validations preventing save would still block save, so anything showing here would be post-save validations, right? It's because we have to render this for both old and new editions that we have to make the message generic... Hmm, maybe that's not such a crazy split (draft vs other states?) if we want to be specific on what action is blocked.

app/views/admin/editions/show/_sidebar_notices.html.erb
```diff
@@ -6,43 +6,22 @@
 %>
 
 <div class="app-view-summary__sidebar-notices">
-  <% if @edition.scheduled_publication %>
-    <% if force_scheduler.can_transition? && !force_scheduler.can_perform? %>
```
lauraghiorghisor-tw, line 10: I wonder (and hope that) if this will unblock some refactoring - and our favourite, code deletion 🔥  - from all these edition workflow classes. It might be that we don't need all these transition check methods. For now they're still in use though.

lauraghiorghisor-tw (approved):
LGTM

app/views/admin/editions/show/_sidebar_notices.html.erb
ChrisBAshton: Agreed. I think once this is merged and publishers are used to the new sidebar section of errors, we'll double down on that as the primary means of funnelling validation and we can delete a bunch of things 🎉

test/functional/admin/generic_editions_controller_test.rb
ChrisBAshton: Yes, this is post-save validation - we've not introduced any additional 'blocking' at the save stage. So publishers will still be able to edit drafts that are 'invalid' to their hearts' content; we're just surfacing potential issues down the road, earlier.

I did think there may be value in differentiating between "this is an error that will stop you from publishing - but is fine for now" vs "this is an error that will stop you from publishing, and should also be fixed right now, e.g. maybe even prevents you from *saving*". But I'm not sure the latter exists? I suppose a publisher might edit a draft and set a schedule date for the past, and face a form validation error (rather than our sidebar). I can't really think of a solid example of something we'd report differently for draft and live editions.

lauraghiorghisor-tw: My thinking is old editions (in post-publish states) will still have pre-save errors but all new ones after this merge should only have post-save errors, therefore only publish/schedule-blocking errors. 

You could thus safely say that "the edition cannot be force published" because it is invalid in x ways for all new things (if the above logic is valid) - but then it does not rightly apply to pre-merge editions. 

And also as time passes by, published editions that were valid at the time can become invalid as new rules are introduced without cleaning up data... so what can we say about them then?

All that to say, on second thought, "invalid" is sound 😅

PR #10392, opened 2025-07-08
Add only invalid editions scope (WHIT-1507)

ChrisBAshton:
Adds a scope (`Edition.only_invalid_editions` / `Edition.not_validated_since("01/01/1970")`) to easily filter only the editions considered to be 'invalid' - see property added in #10391.

This surfaces any edition that has a `nil` value for `revalidated_at`. Initially, all editions will have a `nil` value, but in #10382 we're going to bulk-revalidate all the editions, so that will no longer be the case. It should also not be possible to create a new edition that is valid but has a `revalidated_at` nil value, once #10384 is merged.

The purpose of this scope is twofold:

1. Initially, for dev use only (not exposed to publishers at this point) - to have a figure we can use for reporting system health. And the `not_validated_since` scope will be used for the bulk-revalidation, as we'll want to grab editions that have not been revalidated since a certain date (e.g. 1 week ago)
2. In #10383, expose the scope via the Whitehall search UI, so that publishers can easily find the editions they need to patch.

---

### DIFF
--- a/app/models/concerns/edition/scopes/filterable_by_invalid.rb
+++ b/app/models/concerns/edition/scopes/filterable_by_invalid.rb
@@ -6,5 +6,11 @@ module Edition::Scopes::FilterableByInvalid
       where(revalidated_at: nil)
         .where.not(state: %w[superseded unpublished deleted]) # We don't care if editions are invalid if they're superseded
     }
+
+    scope :not_validated_since, lambda { |from_date|
+      cutoff = Time.zone.parse(from_date.to_s)
+      where("revalidated_at IS NULL OR revalidated_at < ?", cutoff)
+        .where.not(state: %w[superseded unpublished deleted]) # We don't care if editions are invalid if they're superseded
+    }
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/concerns/edition/scopes/filterable_by_invalid.rb
```diff
@@ -0,0 +1,16 @@
+module Edition::Scopes::FilterableByInvalid
+  extend ActiveSupport::Concern
+
+  included do
+    scope :only_invalid_editions, lambda {
+      where(revalidated_at: nil)
+        .where.not(state: %w[superseded unpublished deleted]) # We don't care if editions are invalid if they're superseded
```
lauraghiorghisor-tw, line 7: Musings:
- Not sure where we've left the whole saga of "creating new drafts from invalid unpublished editions" and "skipping the draft creation altogether if we need to publish a fix" (I think this was in the context of those suspicious sites) ->  is this truly not useful to have the scope include unpublished?

ChrisBAshton: Hmmm I'd argue it is not useful to have the scope include unpublished, **but** we should also never be in a situation where a publisher is unable to create a new draft from an unpublished edition. (We should have some sort of validation bypassing on-create). That way they'll encounter validation issues they need to fix before they can publish, but will at least be in a position to be able to fix up those issues.

ChrisBAshton: We're going to ensure the above in https://gov-uk.atlassian.net/browse/WHIT-1517?focusedCommentId=74653

PR #10402, opened 2025-07-14
CM 226: Model "telephones" call charges info URL (within "Contact" content block)

edavey:
(See also https://github.com/alphagov/publishing-api/pull/3452 for schema changes)

**Jira: [CM-226](https://gov-uk.atlassian.net/browse/CM-226)**

In this PR we extend the `telephones` embedded object within the "Contact" content block schema to model "call charges". These comprise 3 properties:

1. `show_call_charges_info_url`: a boolean indicating whether or not the information is to be shown
2. `label`: a label or heading or legend with a default value of "Find out about call charges"
3. `call_charges_info_url`: the url at which the information about call charges can be found. This has a default value of `https://gov.uk/call-charges`

This PR also:

- exposes any default values of nested fields via `ContentBlock::Schema::Field::NestedField`
- adds a `Field#nested_field(name)` method to facilitate access to nested fields

In addition this PR includes a tweak to the repo's Capybara config to set the Capybara `asset_host` to a value given in the `ASSET_HOST` environment variable so that Capybara pages can be saved and viewed locally with assets served from a local development server. For example, if you run a test with:

```
govuk-docker-run env  
ASSET_HOST=http://whitehall-admin.dev.gov.uk  
RAILS_ENV=test  
bundle exec cucumber path/to/feature
```
this will set the following `<base>` element in the HTML HEAD:

```
<base href="http://whitehall-admin.dev.gov.uk">
```

---

### Form
call_charges_form

---

### Review stage
call_charges_confirmation

[CM-226]: https://gov-uk.atlassian.net/browse/CM-226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
@@ -1,9 +1,5 @@
 <div class="app-c-content-block-manager-call-charges-component">
 
-  <% show_call_charges_info_url = field.nested_field("show_call_charges_info_url") %>
-  <% call_charges_info_url = field.nested_field("call_charges_info_url") %>
-  <% label = field.nested_field("label") %>
-
   <% call_charges_url_input = capture do %>
     <%= render("govuk_publishing_components/components/input", {
               label: {
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.rb
@@ -1,2 +1,13 @@
 class ContentBlockManager::ContentBlockEdition::Details::Fields::CallChargesComponent < ContentBlockManager::ContentBlockEdition::Details::Fields::ObjectComponent
+  def show_call_charges_info_url
+    field.nested_field("show_call_charges_info_url")
+  end
+
+  def call_charges_info_url
+    field.nested_field("call_charges_info_url")
+  end
+
+  def label
+    field.nested_field("label")
+  end
 end


features/support/capybara.rb
```diff
@@ -14,6 +14,10 @@
                                    headless: true)
 end
 
+Capybara.configure do |config|
+  config.asset_host = ENV.fetch("ASSET_HOST", nil)
+end
```
pezholio, line 19: Nice catch! Would be good to add this env var in the docker-compose file in `govuk-docker` too https://github.com/alphagov/govuk-docker/blob/main/projects/whitehall/docker-compose.yml

lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema/field.rb
```diff
@@ -4,7 +4,11 @@ class Schema
       class Field
         attr_reader :name, :schema
 
-        NestedField = Data.define(:name, :format, :enum_values)
+        NestedField = Data.define(:name, :format, :enum_values, :default_value) do
```
pezholio, line 7: Oh neat! I forgot this was a nested field, and so didn't have the default value method

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="app-c-content-block-manager-call-charges-component">
 
-  <% show_call_charges_info_url = field.nested_fields.find{ |nf| nf.name == "show_call_charges_info_url" } %>
-  <% call_charges_info_url = field.nested_fields.find{ |nf| nf.name == "call_charges_info_url" } %>
+  <% show_call_charges_info_url = field.nested_field("show_call_charges_info_url") %>
+  <% call_charges_info_url = field.nested_field("call_charges_info_url") %>
```
pezholio: Lovely stuff!

pezholio: It might be better to have these as methods in the class, eg:

```
def show_call_charges_info_url
  field.nested_field("show_call_charges_info_url")
end
```

I tend to get a bit antsy when declaring variables in a template. No biggy though

edavey: thanks, yes that is WAY better!

features/support/capybara.rb
edavey: 👍 yes good idea, I'll open PR to add this ASSET_HOST="http://whitehall-admin.dev.gov.uk"/
environment setting to [Whitehall's Docker config][]

[Whitehall's Docker config]:
https://github.com/alphagov/govuk-docker/blob/main/projects/whitehall/docker-compose.yml#L31-L37

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
edavey: done

PR #10405, opened 2025-07-14
(CM-148) Don’t send a `content_block` update type

pezholio:
In #10386 we updated the update type to be major or minor, instead of sending a “special” `content_block` updated type, but neglected to update this part of the code where we set the update type. 

Now we leave this blank and let Publishing API work out the update type from the edition that is being published (sending the update type to the Publish endpoint is deprecated anyway, see https://github.com/alphagov/publishing-api/blob/main/docs/api.md#json-attributes-1)


edavey (approved):
In the [API docs](https://github.com/alphagov/publishing-api/blob/d7088dfd83b10a8af864ebaa56af272bc76eee19/docs/api.md?plain=1#L201) I don't really understand this directive about `update_type`:

> clients should instead set the update type when
    putting the content.

I guess this is referring to some other message which is sent using the PUT method?

pezholio:
Yeah, this is what happens in https://github.com/alphagov/whitehall/blob/407f2b91dced755a2f003370737a0cb8a2017d9c/lib/engines/content_block_manager/app/services/content_block_manager/publish_edition_service.rb#L35

pezholio:
I'll hold off merging this until https://github.com/alphagov/publishing-api/pull/3442 is in

PR #10410, opened 2025-07-15
Remove support for `.zip` files in Whitehall [WHIT-1873]

patrickpatrickpatrick:
### Change bulk upload to use multiple files instead of `zip`

- Rename `upload_zip` to `upload_files`
- Remove the bulk upload zip functionality
  - Delete `BulkUpload::Zip`
  - Remove frontend references to "Bulk upload from zip file"
- Use updated `FileUpload` component
  - Add `multiple` to upload multiple files
  - Use the [Javascript enhanced version of the File Upload component](https://design-system.service.gov.uk/components/file-upload/)
- Make the `Attachment` partials reusable for `BulkUpload`
- Various accessibility fixes
  - More descriptive errors that link to specific errored fields
  - Use `fieldset` and `legend` in the individual groups of file attributes
  - Heading levels in post upload screen fixed to have hierarchy (previously all headings and labels were `h2`)
- Adjust test suite for the changes made

### Remove `.zip` functionality from Whitehall
- Remove `zip` from allowed file extensions
- Remove `ZipFile` from `AttachmentUploader`
- Remove `broken_links_export`
  - Remove rake task
  - Remove mailer
  - Remove controller
  - Remove routes
- Remove additional references to `zip` files

### Visual Differences

| Case  | Before | After |
| ------------- | ------------- | ------------- |
| `editions/ID/attachments`  | Screenshot 2025-07-21 at 14 11 31  | Screenshot 2025-07-21 at 14 11 34  |
| `editions/ID/bulk_uploads/new`  |  Screenshot 2025-07-21 at 14 12 46  | Screenshot 2025-07-21 at 14 12 42  |
| `editions/ID/bulk_uploads/new` with an invalid file  |  Screenshot 2025-07-21 at 14 14 20  | Screenshot 2025-07-21 at 14 19 25  |
| `editions/ID/bulk_uploads` with an invalid fields  |  Screenshot 2025-07-21 at 15 48 41  |  Screenshot 2025-07-21 at 15 46 21  |

## Why

The current implementation of uploading multiple files is difficult to maintain, unintuitive for the users (and developers) and inaccessible (error messages). It's also recently caused some support tickets which have been difficult to address due to the complexity of the bulk uploader.

Changing the implementation of the bulk uploader to use the new enhanced File Upload component that supports `multiple` should address these issues.

### Next steps

Ideally we shouldn't have a separate controller and routing for the bulk uploading at all as `BulkUpload` now supports uploading a single file. In my view, this PR is is a good intermediate step for the work to remove 'BulkUpload' to start. In my opinion the next step would be to remove the "single" file upload and only have the multiple file upload as an option. This would require more work to the `AttachmentController` and associated tests however. Leaving `BulkUpload` as a separate route also means we can get feedback on how users use the new journey. 


app/models/bulk_upload.rb
```diff
@@ -1,49 +1,66 @@
...
     @edition = edition
-    @attachments = []
+    @attachments = if files
+                     errors.add(:files, message: "not selected for upload") if files.compact.blank?
+                     files.compact.map { |file| build_attachment_for_file(file) }
+                   elsif attachments_params.present?
+                     attachments_params.values.map(&method(:find_and_update_existing_attachment))
+                   else
+                     []
+                   end
```
ChrisBAshton: This branching of logic reads as quite surprising (and is at least explained in the deleted comment that was on lines 4-11). Can we change the API of this class to do less stuff in the `initialize` method? I'm thinking something like:

``` 
  def new
    @bulk_upload = BulkUpload.new(@edition)
  end

  def upload_zip
    files = params.fetch(:bulk_upload, {})[:files] || []

    @bulk_upload = BulkUpload.new(@edition)
    @bulk_upload.populate_files(files:)

    if @bulk_upload.errors.present?
      render :new
    else
      render :set_titles
    end
  end

  def create
    @bulk_upload = BulkUpload.new(@edition)
    @bulk_upload.populate_attachments(attachments_params:)
```

app/models/bulk_upload.rb
```diff
@@ -1,49 +1,66 @@
...
+    sanitized_file.move_to(File.join(File.dirname(sanitized_file.path), sanitized_file.filename))
+
+    filename = sanitized_file.filename
+    file = File.open(sanitized_file.path)
+
+    attachment = find_attachment_with_file(filename) || FileAttachment.new(attachment_data_attributes: { file: })
+
+    unless AttachmentUploader::MIME_ALLOW_LIST.include?(sanitized_file.content_type)
+      @invalid_file_types << File.extname(filename)
+      errors.clear
```
ChrisBAshton: What is the `errors.clear` bit for? Don't we want to preserve all errors? May need an inline comment here

app/models/bulk_upload.rb
```diff
@@ -1,49 +1,66 @@
...
+    # attachment without losing other attachment data, so we need to "move"
+    # the temporary file so that it has the original filename on save.
+    #
+    # Using CarrierWave::SanitizedFile also means that the filename is
+    # sanitized in the same way as all other uploaded files. This prevents
+    # inconsistency if we decided to change our uploaded filename format
+    # in the future.
+    #
+    sanitized_file = CarrierWave::SanitizedFile.new(uploaded_file)
+    sanitized_file.move_to(File.join(File.dirname(sanitized_file.path), sanitized_file.filename))
```
ChrisBAshton: I'd be tempted to move this whole chunk to a private `preserve_original_filename` method

```ruby
  sanitized_file = preserve_original_filename(uploaded_file)
```

As otherwise there's quite a lot to follow in this method.

app/views/admin/attachments/_attachment_data_fields.html.erb
```diff
@@ -1,25 +1,37 @@
...
+ heading_size ||= "l"
+ heading_level ||= 2
+%>
+
 <%= form.fields_for(:attachment_data, include_id: false) do |attachment_data_fields| %>
   <% if attachment_data_fields.object.filename %>
-    <%= attachment_data_fields.hidden_field(:to_replace_id, value: attachment_data_fields.object.to_replace_id || attachment_data_fields.object.id) %>
-    <p class="govuk-body">Current file: <%= link_to_attachment_data(attachment_data_fields.object) %></p>
+    <%= hidden_field_tag "#{name}[attachment_data_attributes][to_replace_id]", attachment_data_fields.object.to_replace_id || attachment_data_fields.object.id %>
+    <% unless controller_name == "bulk_uploads" %>
```
ChrisBAshton, line 10: The branching of view logic depending on whether or not we're coming from the 'bulk uploads' controller is a bit of a smell 😬 but I think it's justifiable on the basis that the next PR following this one would remove the 'single file upload' controller/view altogether and just have everything use the bulk uploader, which would be a great opportunity for simplification. So I'm happy not to quibble about this! 👍

app/helpers/attachments_helper.rb
```diff
@@ -72,4 +72,34 @@ def block_attachments(attachments = [], alternative_format_contact_email = nil)
...
     end
   end
+
+  def bulk_attachment_errors(attachments = nil)
+    return if attachments.blank?
+
+    items = attachments.map(&:errors).flat_map.with_index do |errors, index|
+      errors.map do |error|
+        {
+          text: text = "#{attachments[index].attachment_data.filename}: #{error.full_message.to_s.humanize}",
```
ChrisBAshton: `error.full_message` is already a string, IIRC - so no `.to_s` necessary?

test/functional/admin/bulk_uploads_controller_test.rb
```diff
@@ -62,39 +63,32 @@ def post_to_upload_zip(filename)
     assert_response :forbidden
   end
 
-  view_test "POST :upload_zip with no zip file requests that zip file be specified" do
+  view_test "POST :upload_zip with no files requests that files be specified" do
```
ChrisBAshton: Can we update all these references to uploading zips, now that we're just bulk-native-uploading individual files? I.e. change the `upload_zip` action, remove `.zip` from [allowed extensions](https://github.com/alphagov/whitehall/blob/857d6e561fc41a36663adf9d1d7be040ef8b3958/app/uploaders/attachment_uploader.rb#L4), etc.

ChrisBAshton (commented):
This is shaping up really nicely ⭐ 

I think there are several places we can delete some Ruby code now though? 🙏 

- [x] https://github.com/alphagov/whitehall/blob/857d6e561fc41a36663adf9d1d7be040ef8b3958/app/controllers/admin/bulk_uploads_controller.rb#L11-L21
- [x] https://github.com/alphagov/whitehall/blob/857d6e561fc41a36663adf9d1d7be040ef8b3958/app/uploaders/attachment_uploader.rb#L30-L188
- [x] https://github.com/alphagov/whitehall/blob/6bde85ed0d60db2a65f07cf6929708ff4de46a05/test/unit/app/uploaders/attachment_uploader_test.rb#L75
- [x] https://github.com/alphagov/whitehall/blob/857d6e561fc41a36663adf9d1d7be040ef8b3958/app/uploaders/attachment_uploader.rb#L39
- [x] https://github.com/alphagov/whitehall/blob/857d6e561fc41a36663adf9d1d7be040ef8b3958/lib/whitehall.rb#L39-L44
- [x] https://github.com/alphagov/whitehall/blob/86402bc5d1e4918b17e7ab84285b7c295d2e93e8/test/unit/lib/whitehall_test.rb#L4-L10

...and also places we can clean up zip dependencies:

- [x] https://github.com/alphagov/whitehall/blob/a41fb5715767d07a808ab718850730bc3a545e2e/Dockerfile#L21
- [x] https://github.com/alphagov/whitehall/blob/93d56474056dc9bd6b98c6ffdefcee0f9cfa3b89/Gemfile#L54
- [x] https://github.com/alphagov/whitehall/blob/99385ffc650d1692e2019ef241de2800fe10c42f/lib/tasks/broken_link_reporting.rake#L2

^ that last one feels like a bit of scope creep for the PR - I guess it could be tackled separately - it's already been [identified as a rake task to delete](https://docs.google.com/spreadsheets/d/1fhHGi9TNJhOjWBhTh-QwAdTUsmtHxwLUG-iCpuLgSi0/edit?gid=0#gid=0&range=14:14) so we could have a separate PR to delete it and its 'system' dependencies on zips. But if you feel so inclined, would also be great to just kill ALL the zips in this one PR 🎉 🔥

app/views/admin/attachments/_attachment_data_fields.html.erb
patrickpatrickpatrick: Yeah, this is very much a temporary measure until we can refactor this out in a subsequent PR. I can maybe even include a comment in the template so this isn't forgotten about.

test/functional/admin/bulk_uploads_controller_test.rb
patrickpatrickpatrick: Ah, I didn't know that we were removing all `zip` file uploading! So not as any attachment?

ChrisBAshton: Correct - I can't think of any situation where we wouldn't want to use the new multi-file upload instead.

app/uploaders/attachment_uploader.rb
```diff
@@ -1,11 +1,9 @@
 class AttachmentUploader < WhitehallUploader
   PDF_CONTENT_TYPE = "application/pdf".freeze
   INDEXABLE_TYPES = %w[csv doc docx ods odp odt pdf ppt pptx rdf rtf txt xls xlsx xml].freeze
-  EXTENSION_ALLOW_LIST = %w[chm csv diff doc docx dot dxf eps gif gml ics jpg kml odp ods odt pdf png ppt pptx ps rdf ris rtf sch txt vcf wsdl xls xlsm xlsx xlt xml xsd xslt zip].freeze
+  EXTENSION_ALLOW_LIST = %w[chm csv diff doc docx dot dxf eps gif gml ics jpg kml odp ods odt pdf png ppt pptx ps rdf ris rtf sch txt vcf wsdl xls xlsm xlsx xlt xml xsd xslt].freeze
```
ChrisBAshton, line 4: Let's also remove zip from https://github.com/alphagov/whitehall/blob/3b717cab95e965fda5583b7381d88488823e4878/app/helpers/document_helper.rb#L85

and corresponding test:
https://github.com/alphagov/whitehall/blob/3b717cab95e965fda5583b7381d88488823e4878/test/unit/app/helpers/document_helper_test.rb#L47-L49

lib/tasks/broken_link_reporting.rake
ChrisBAshton, line 1: Worth referring to the rake task audit [spreadsheet](https://docs.google.com/spreadsheets/d/1fhHGi9TNJhOjWBhTh-QwAdTUsmtHxwLUG-iCpuLgSi0/edit?gid=0#gid=0) and/or [Trello](https://trello.com/c/zUBmSn3L/2080-audit-rake-tasks) in the commit message.

config/brakeman.ignore
```diff
@@ -168,29 +145,6 @@
       ],
       "note": ""
     },
-    {
-      "warning_type": "Command Injection",
```
ChrisBAshton, line 172: Ooh nice, I hadn't realised we've got Brakeman warnings we can delete 🎉

ChrisBAshton (approved):
Amazing job, Patrick ⭐ I've been banging this drum for a year or two and it's lovely to see it now being implemented! A couple of small comments below, but happy to pre-approve.

patrickpatrickpatrick:
Discussion led to us deciding that "Change bulk upload to use multiple files instead of zip" and "Change bulk upload to use multiple files instead of zip" should be two different PRs. 

There is still a use-case for `zip` attachments in Whitehall e.g. a large set of connected forms. The simplest way to overcome this would be just to add `zip` to the allowed extensions but the application that handles uploads https://github.com/alphagov/asset-manager doesn't validate files within `zip` archives. This would mean users could upload arbitrary files.

PR #10411, opened 2025-07-15
(CM-239) Generate object keys more intelligently

pezholio:
Up until now, we worked on the assumption that all subobjects would require a title, which we would then use to generate an object key.

For example, if we added an email object to a contact with the name “My email address”, this would be added to the `email` object with the key `my-email-address`. For example:

```
…

{
  “emails”: {
      “my-email-address”: { … }
}
…
```

We decided early on to make the titles mandatory, because otherwise we would have no way of generating a predictable identifier for an individual object.

As we’ve moved along the road, we’ve discovered that actually, mandating the use of a title causes more problems than it solves, so we’ve decided to make `title` an optional field and fall back to generating a key if a title is not present.

So, in the example above, if we were to add an email to our contact, and there was no title provided, the key would default to `email-1`.

Additionally, if there was already an email object present, we would use the count of the current number of email objects, and increment this count to generate the key `email-2`

We already did have some fallback code to generate a random identifier, but this never got hit, becuase all subobjects had mandatory titles.

### DIFF
--- a/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
+++ b/lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
@@ -48,7 +48,7 @@ module ContentBlockManager
       end
 
       def add_object_to_details(object_type, body)
-        key = key_for_object(body)
+        key = key_for_object(object_type, body)
 
         details[object_type] ||= {}
         details[object_type][key] = remove_destroyed body.to_h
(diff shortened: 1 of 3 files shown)


lib/engines/content_block_manager/app/models/content_block_manager/content_block/edition.rb
```diff
@@ -58,8 +58,13 @@ def update_object_with_details(object_type, object_title, body)
...
       end
 
-      def key_for_object(object)
-        object["title"]&.parameterize.presence || SecureRandom.alphanumeric.downcase
+      def key_for_object(object_type, object)
+        if object["title"].present?
+          object["title"].parameterize
+        else
+          current_count = details[object_type]&.values&.count || 0
+          "#{object_type.parameterize}-#{current_count + 1}"
```
edavey, line 66: just to nit-pick the terminology in the commit message, I think you meant to write "**increment** this count" rather than "prepend this count"

edavey (approved):
This make sense!

PR #10414, opened 2025-07-16
(CM-176) Update country component

pezholio:
This updates the address mode to use the previously added country component, which inherits from the Enum component. I've also made some updates to ensure only geographical countries are included, and the default (blank) option is the United Kingdom.

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/enum_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/enum_component.rb
@@ -5,17 +5,29 @@ class ContentBlockManager::ContentBlockEdition::Details::Fields::EnumComponent <
     super(**args)
   end
 
-private
-
   def options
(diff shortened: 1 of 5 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/enum_component.rb
```diff
@@ -7,14 +7,26 @@ def initialize(enum:, default: "", **args)
...
+      options.push({
         text: item,
-        value: item,
+      value: item,
         selected: selected?(item),
-      }
+      })
     end
+
+    options
```
edavey, line 25: do you think it's worth using the EnumComponent unit test to illustrate the behaviour here? I'm not clear what's going on...

pezholio: I've made the `options` method public and tested it directly now, so hopefully that will simplify it a bit 👍

PR #10416, opened 2025-07-16
Content Modelling: Tidy: extract call_charges label copy to locale file

edavey:
We extract all label text from the `CallChargesComponent` to the local file. We'll try to keep components free of "copy".

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
@@ -3,7 +3,7 @@
   <% call_charges_url_input = capture do %>
     <%= render("govuk_publishing_components/components/input", {
               label: {
-              text: "Label",
+              text: label_for("label"),
             },
             name: name_for_field(label) ,
             id: id_for_field(label),
@@ -12,7 +12,7 @@
 
     <%= render("govuk_publishing_components/components/input", {
                 label: {
-                text: "URL to find out about call charges",
+                text: label_for("call_charges_info_url"),
               },
               name: name_for_field(call_charges_info_url) ,
               id: id_for_field(call_charges_info_url),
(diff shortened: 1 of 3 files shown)


edavey:
@pezholio do you agree with this or do you think that the component should in fact be free of "copy" and that all labels etc should be part of the localisation system?

pezholio:
Ideally, I'd like copy to be in the localisation system, but I know we've been lazy in the past!

edavey:
> Ideally, I'd like copy to be in the localisation system, but I know we've been lazy in the past!

I've now extracted the label copy to the locale file, so this is now ready for another look, please

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.rb
```diff
@@ -10,4 +10,8 @@ def call_charges_info_url
   def label
     field.nested_field("label")
   end
+
+  def label_for(name)
```
pezholio: There's actually a helper method here you can use - https://github.com/alphagov/whitehall/blob/8803ac6006763f7d08490f1b25ad958e3fa33442/lib/engines/content_block_manager/app/helpers/content_block_manager/content_block/translation_helper.rb#L2

edavey: 👍 thanks, will tweak

edavey: @pezholio this is now tweaked to use the existing helper, so ready for another look.

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/call_charges_component.html.erb
```diff
@@ -25,13 +25,13 @@
...
     id: id_for_field(show_call_charges_info_url),
-    heading: "Show call charges info link?",
+    heading: humanized_label("call_charges.show_call_charges_info_url"),
     visually_hide_heading: true,
     no_hint_text: true,
     small: true,
     items: [
       {
-        label: "Show hyperlink to 'Find out about call charges'",
+        label: humanized_label("call_charges.show_call_charges_info_url"),
```
pezholio: I think the way we've gone before is nest the translations within `telephones` and add the `telephones` `object_type` argument, ie:

```
        label: humanized_label("call_charges.show_call_charges_info_url", "telephones"),
```

edavey:
@pezholio Could I get another review on this now, please?

PR #10420, opened 2025-07-17
Fix call charges validation

pezholio:
`show_call_charges_info_url` is a boolean, but is set as a string. This wasn't getting captured by the validator, so was causing an error when passing this to Publishing API.

This updates the telephone preprocessor to cast the field as a boolean, as well as some other tweaks to fix boolean handling.


edavey (approved):
👍 Thanks Stu, I will extend the pre-processor's work for the "video relay service" component I'm building which similarly needs to be toggled with a boolean...

PR #10423, opened 2025-07-17
(CM-243) Add BSL guidance

pezholio:
This adds support for the fields added in https://github.com/alphagov/publishing-api/pull/3462

## Screenshot

image


edavey (approved):
Looks good!

PR #10424, opened 2025-07-17
CM 227: model video relay service within telephones

edavey:


pezholio:
I've had to make a few tweaks after discussion with Nadeem and others - the relay service is seperate from Relay UK, so we have two models for each thing. Will merge this to ensure we get it in in time for user research

pezholio (approved):

PR #10437, opened 2025-07-22
(CM-244) Fix mistake in the url when deleting newly added telephone numbers in review stage of edit journey

pezholio:
This fixes an issue when clicking "Edit" on a subschema when reviewing during the edit process. In https://github.com/alphagov/whitehall/pull/10319 we added the concept of "Groups" for subschemas, which meant after editing an embedded object, the user is redirected to the page for the group, rather than the subschema. We forgot to take this into account at the review stage, so we still assumed each subschema had its own page, which resulted in a 404.

This tweak should fix it


edavey (approved):
👍 I think this fixes the "Step embedded_telephones does not exist" error which I noted locally this morning?

pezholio:
That's the fella :)

PR #10438, opened 2025-07-22
Fix 'native name' of each non-English language

ChrisBAshton:
These values were set in PR #9845, where we can see what the values used to be in the `language_names: {locale}` property of the corresponding file (e.g. [ar.yml](https://github.com/alphagov/whitehall/blob/10b15007275cce466b33ac14020c59381c617867/config/locales/ar.yml#L730)).

From memory, I think I used Copilot to perform the laborious task of copying the correct value from the old locale file to the new `languages.yml` file, and evidently, it hallucinated some translations. So I've manually patched these back to the values they were prior to the change, even if that meant lowercasing them when we probably shouldn't(!) - and then fixed up the lowercasing in the next commit, just so we have a full audit trail.

Thanks to @yndajas for spotting the issue.

---

### DIFF
--- a/config/languages.yml
+++ b/config/languages.yml
@@ -1,11 +1,11 @@
 ar:
   direction: rtl
   english_name: Arabic
-  native_name: العربية
+  native_name: العربيَّة
 az:
   direction: ltr
   english_name: Azeri
-  native_name: Azərbaycanca
+  native_name: Azeri
 be:
   direction: ltr
   english_name: Belarusian
@@ -13,7 +13,7 @@ be:
 bg:
   direction: ltr
   english_name: Bulgarian
-  native_name: Български
+  native_name: Bulgarian
 bn:
   direction: ltr
   english_name: Bangla
@@ -65,7 +65,7 @@ fa:
 fi:
   direction: ltr
   english_name: Finnish
-  native_name: Suomi
+  native_name: suomi
 fr:
   direction: ltr
   english_name: French
@@ -77,7 +77,7 @@ gd:
 gu:
   direction: ltr
   english_name: Gujarati
-  native_name: ગુજરાતી
+  native_name: જોર્જિયન
 he:
   direction: rtl
   english_name: Hebrew
@@ -85,7 +85,7 @@ he:
 hi:
   direction: ltr
   english_name: Hindi
-  native_name: हिन्दी
+  native_name: हिंदी
 hr:
   direction: ltr
   english_name: Croatian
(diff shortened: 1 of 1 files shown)


config/languages.yml
```diff
@@ -1,19 +1,19 @@
...
+  native_name: Azeri
 be:
   direction: ltr
   english_name: Belarusian
   native_name: Беларуская
 bg:
   direction: ltr
   english_name: Bulgarian
-  native_name: Български
+  native_name: Bulgarian
```
yndajas: **note:** interesting. I guess Whitehall wasn't quite in line with (Government) Frontend: https://github.com/alphagov/frontend/blob/main/config/locales/bg.yml#L721

config/languages.yml
```diff
@@ -241,23 +241,23 @@ ur:
...
+  native_name: יידיש
 zh:
   direction: ltr
   english_name: Chinese
   native_name: 中文
 zh-hk:
   direction: ltr
   english_name: Cantonese
-  native_name: 粵語
+  native_name: 中文
```
yndajas, line 260: **note:** this makes 'Cantonese' and 'Chinese' equivalent 👀 I suppose it was that way before, but still divergent from Frontend

ChrisBAshton: Hmm I've just double-checked Frontend and this matches?

It was a useful prompt though - Frontend has lowercase `norsk` instead of `Norsk`, so maybe I should remove my 'capitalisation' commit. I'll loop in @KludgeKML for an opinion (I believe he added the translations to Frontend).

config/languages.yml
ChrisBAshton: I've now pushed up a commit to take Frontend's (presumably more accurate!) value.

config/languages.yml
KludgeKML: I just copied them from government-frontend, I'm afraid (GF had them to support the translations controls on localisable pages, when I moved them over it was to support other languages in frontend, which until then had only had english and welsh localisation).

yndajas: I must have been hallucinating yesterday 🙈 maybe I thought I was looking at Frontend but it was actually Whitehall

KludgeKML: Ynda is right about the various chinese names, though, and I think frontend has the same bug: Simplfied Chinese and Traditional Chinese look right, but I believe Cantonese does have another set of characters specifically for Cantonese. We should probably correct that here and in frontend. (It shouldn't be too hard to find someone to confirm, given how big the Hong Kong diaspora is in the UK)

PR #10439, opened 2025-07-22
Amendments to tracked form events [WHIT-2325]

patrickpatrickpatrick:
## What

- Use `fieldset` for `datetime_fields`
- Move "Support organisations" select outside of "Lead Organisation" `fieldset`
- Updates to `Ga4IndexSectionSetup` to take into account multiple choice fields and split fields
- Updates to `Ga4FormTracker` to use correct `section` for fields within fieldsets

## Why

Addresses indexing issues that have occurred within forms in the previous implementation, which based the index on if the field was a child of `fieldset` and not on what the field actually controlled within the form. Using `name` and `id` as a basis for the indexing will allow more accurate generation of index sections.

### DIFF
--- a/app/views/components/_datetime_fields.html.erb
+++ b/app/views/components/_datetime_fields.html.erb
@@ -41,82 +41,79 @@
 %>
 <% if prefix && field_name %>
   <%= tag.div class: root_classes, data: data_attributes, id: id do %>
-    <% unless date_only && !date_heading %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: date_heading || "Date (required)",
-        heading_level: heading_level || 3,
-        font_size: heading_size || "m",
-        padding: true,
+    <% date_input = capture do %>
+      <%= render "govuk_publishing_components/components/date_input", {
+        id: date_id,
+        hint: date_hint,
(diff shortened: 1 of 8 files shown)


app/views/components/_datetime_fields.html.erb
```diff
@@ -41,82 +41,79 @@
 %>
 <% if prefix && field_name %>
   <%= tag.div class: root_classes, data: data_attributes, id: id do %>
-    <% unless date_only && !date_heading %>
-      <%= render "govuk_publishing_components/components/heading", {
-        text: date_heading || "Date (required)",
-        heading_level: heading_level || 3,
-        font_size: heading_size || "m",
-        padding: true,
+    <% date_input = capture do %>
```
lauraghiorghisor-tw, line 44: That double negation + OR made my mind collapse, so sorry if I'm off, but I guess there was a conditional here that only showed a heading for date when there was also a time? Whereas now we get a heading all the time? I don't imagine that's a bit issue. As long as you've tested and you're ok with the behaviour, I guess.

lauraghiorghisor-tw (approved):
LGTM

PR #10443, opened 2025-07-23
(CM-168) Handle name collisions in object keys

pezholio:
If an object has the same title, then we add a number as appropriate to prevent name collisions. Additionally, I’ve simplified the code so if a title is missing we use the object_type as the key.

I’ve had to tweak the tests, so the first time an object_type is used, a number isn’t appended, but I think this makes sense in reality.


edavey (approved):
Probably I'm not understanding properly but might it be simpler if we always included a SHA or UUID in the key? Rather than figuring out whether to use a number and then whether to increment the number?

pezholio:
Yeah, I was thinking more about how they're presented when they get used by users when they get embedded. It's slightly friendlier to use something human readable, rather than shas or UUIDs, especially as they'll be visible when editing.

PR #10445, opened 2025-07-23
CM-249: Order "Opening times" nested object's properties in summary card

edavey:
See [Jira CM-249](https://gov-uk.atlassian.net/browse/CM-249)

In this PR we:

### 1) extend `Document::Show::EmbeddedObjects::BlocksComponent`
to  order the rows in the summary cards rendered according to the `field_order`
setting, if present.

### 2) define 'field_order' for "Opening times" nested object
within the "telephones" sub-schema of the "Contact" content block

ordering_of_opening_hours_properties

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
@@ -61,12 +61,22 @@ private
   end
 
   def rows_for_nested_items(items, nested_name, index)
-    items.map do |key, value|
+    rows = items.map do |key, value|
       {
         key: key_to_title(key),
         value: content_for_row(embed_code_identifier(nested_name, index, key), value),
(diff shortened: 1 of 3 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block/document/show/embedded_objects/blocks_component.rb
```diff
@@ -61,13 +61,23 @@ def nested_blocks
...
   def rows_for_nested_items(items, nested_name, index)
-    items.map do |key, value|
+    rows = items.map do |key, value|
       {
         key: key_to_title(key),
         value: content_for_row(embed_code_identifier(nested_name, index, key), value),
         data: data_attributes_for_row(embed_code_identifier(nested_name, index, key)),
       }
     end
+    ordered_by_field_order(rows, nested_name)
```
pezholio, line 71: Would it be easier to do the sorting before we map the items to rows? I'm just a bit concerned that the call to `parameterize` will mean some things get missed (especially if they're translated)

PR #10447, opened 2025-07-23
WIP: experimental:  Add summary component for editing/reviewing opening hours

edavey:
## EXPERIMENT: for comments

This new `shared/embedded_objects/summary_card/opening_hours_component.html.erb` component is rendered by `shared/embedded_objects/summary_card_component.html.erb` when:

- editing a new edition and reviewing a work in progress on the `group_modes` step at a path like:

```
/content-block-manager/content-block/editions/10/workflow/group_modes
```

- and at the `review` step at a path like:

```
content-block-manager/content-block/editions/10/workflow/review
```

custom_summary_card_for_opening_hours

### Notes:

- I understand that we may be stepping back from the structured approach to Opening times so this particular custom summary card component may not be needed. However, the general pattern may be useful?
- automated tests need to be backfilled
- a reflexive module or static lookup table to "constantise" potential custom components based on the value of the nested  item key (in this case `"opening_hours"` -> `ContentBlockManager::Shared::EmbeddedObjects::SummaryCard::OpeningHoursComponent`) is probably needed

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card/opening_hours_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card/opening_hours_component.rb
@@ -1,0 +1,28 @@
+class ContentBlockManager::Shared::EmbeddedObjects::SummaryCard::OpeningHoursComponent < ViewComponent::Base
+  include ContentBlockManager::ContentBlock::TranslationHelper
+  with_collection_parameter :opening_hours
+
+  def initialize(opening_hours)
+    @hours = opening_hours.fetch(:opening_hours)
+  end
+
+  def title
+    "Opening hours"
+  end
+
+  def rows
+    [{ key: "Hours", value: formatted_hours }]
+  end
+
+private
+
+  def formatted_hours
+    [
+      @hours.fetch("day_from"),
+      @hours.fetch("time_from").downcase,
+      "to",
+      @hours.fetch("day_to"),
+      @hours.fetch("time_to").downcase,
+    ].join(" ")
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.html.erb
```diff
@@ -17,7 +17,10 @@
     } %>
 
     <% nested_items(items).each do |key, items| %>
-      <% if items.is_a?(Array) %>
+      <% if key == "opening_hours"%>
```
pezholio, line 20: We could possibly have this as a config setting in the same way we do for fields

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card/opening_hours_component.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def formatted_hours
+    [
+      @hours.fetch("day_from"),
+      @hours.fetch("time_from").downcase,
+      "to",
+      @hours.fetch("day_to"),
+      @hours.fetch("time_to").downcase,
+    ].join(" ")
+  end
```
pezholio, line 27: I wonder if we go down this route if we should use `content_block_tools` to format the hours, so we ensure the formatting is the same and we only have to change it in one place. We'd probably have to break the formatting out into a helper method or something so it's publicly accessible.

pezholio (commented):
This is something I've been thinking about myself, so thanks for doing it! A couple of things to think about, but I think the general idea is sound 👍

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card_component.html.erb
edavey: > We could possibly have this as a config setting in the same way we do for fields

that sounds like a good place for it

edavey:
I'm closing this now because our latest thinking, following round 6 of user testing, is that rather than consisting of structured data (date_from, time_from, date_to, time_to) "opening hours" should be described using a simple text block.

PR #10452, opened 2025-07-25
Requested amendments to analytics [WHIT-2323] [WHIT-2306]

patrickpatrickpatrick:
## What

- Add `govuk:content-id` to metatags
  - Remove `data-ga4-content-id` from root container
  - Remove logic for getting `content-id` from container from `GA4FormTracker`
- Add `govuk:user-id` to metatags
- Use `document` in `Ga4IndexSectionSetup`
- Add `CSS.escape` to programmatic CSS selectors

## Why

- If `content-id` is present in `pageview` event on load, then it will be present in subsequent events. We want to send the `content-id` in different scenarios, so it makes sense to have in the `pageview` event.
- `user-id` has been requested to be added to `pageview` events.
- Issue in which `Ga4IndexSectionSetup` would `error` if element was not part of a form (such as copy-and-paste embed code components)
- Issue in which programmatic CSS selectors were invalid as the name or id of an element contained characters that need to be escaped in queries. This has been fixed using `CSS.escape`.

### DIFF
--- a/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
@@ -35,7 +35,7 @@ window.GOVUK.analyticsGa4.analyticsModules =
             // field could have a hidden element if it is
             // a checkbox so need to exclude hidden
             const multipleElements = moduleElement.querySelectorAll(
-              `[name='${element.name}']:not([type="hidden"])`
+              `[name='${CSS.escape(element.name)}']:not([type="hidden"])`
             )
             // if name is split then `name` will be of format
             // `name[Ni]` where `N` is index of split field
--- a/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
+++ b/app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
@@ -32,10 +32,9 @@ window.GOVUK.analyticsGa4.analyticsModules =
 
         Array.from(indexedElements)
           .map((element) => {
-            const form = element.form
             // field could have a hidden element if it is
             // a checkbox so need to exclude hidden
-            const multipleElements = form.querySelectorAll(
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
```diff
@@ -32,11 +32,10 @@ window.GOVUK.analyticsGa4.analyticsModules =
 
         Array.from(indexedElements)
           .map((element) => {
-            const form = element.form
             // field could have a hidden element if it is
             // a checkbox so need to exclude hidden
-            const multipleElements = form.querySelectorAll(
-              `[name='${element.name}']:not([type="hidden"])`
+            const multipleElements = document.querySelectorAll(
+              `[name='${CSS.escape(element.name)}']:not([type="hidden"])`
```
ryanb-gds, line 38: I'm a bit confused by what this code was doing before. It looks at first glance as though we're searching for inputs with duplicate names, which in my naive view of things shouldn't really exist but maybe in some cases such as with choices JS does? In any case I'm wondering if we should be more specific than `document` here - can we search the `element`, or if not then the `moduleElement`? Breaking out to the global document scope from inside our module feels like something we should avoid.

ryanb-gds (commented):
I have a thought, but it may not be accurate/useful. Happy to approve if there's nothing to be done on it

app/assets/javascripts/admin/analytics-modules/ga4-index-section-setup.js
patrickpatrickpatrick: A `section` is defined as a group of inputs that edit the same attribute within a form. Inputs with the same name will exist if they are multiple or single choice (radio or checkbox inputs) for an attribute of a form. This selector consistently now determines if an `input` is part of a `section` or is a `section` itself (a text input or a select).

`moduleElement` is probably a better shout here instead of `document`, I'll make that change.

ryanb-gds (approved):
:shipit:

PR #10453, opened 2025-07-27
[WHIT-1460] Create Rake Task to Export Design Decisions From Whitehall

TonyGDS:
## What

A rake task that exports Publications (including attachments) from a supplied Organisation and Document Collection. It exports all published Publications under the document collection including the first PDF attachment for each decision. The exported csv will be used by a subsequent rake task in Specialist Publisher to create entries for the Design Decisions finder.

## Why

[[WHIT-1460] IPO Design decisions Finder](https://gov-uk.atlassian.net/browse/WHIT-1460) Involved the creation of a new finder for Design Decisions. IPO asked if the existing records could be bulk imported rather than manually created, due to the amount of existing decisions. The script could be extended to support other publications in future.


lib/tasks/publications_export.rake
```diff
@@ -0,0 +1,12 @@
+namespace :publications do
+  desc "Export all publications for a document collection to a CSV file."
+  task export_for_document_collection: :environment do
+    require "csv"
+
+    filepath = Rails.root.join("publictions_export.csv")
+    puts "Exporting publications to #{filepath}..."
+
+    CSV.open(filepath, 'w') do |csv|
```
ChrisBAshton: This threw me for a minute - might be good to add an inline comment like `# TODO: implement in next commit`

lib/tasks/publications_export.rake
```diff
@@ -7,6 +7,9 @@ namespace :publications do
     puts "Exporting publications to #{filepath}..."
 
     CSV.open(filepath, 'w') do |csv|
+      publlication = Publication.first
```
ChrisBAshton: Typo:

```suggestion
      publication = Publication.first
```

lib/tasks/publications_export.rake
```diff
@@ -8,8 +8,8 @@ namespace :publications do
 
     CSV.open(filepath, 'w') do |csv|
       publlication = Publication.first
-      csv << %w"title summary"
-      csv << [publlication.title, publlication.summary] if publlication
+      csv << %w"title summary body"
```
ChrisBAshton: I think this can be fixed up to the 'title/summary' commit - no reason for it to be its own commit.
(Saves an additional conflict commit when patching up the typo too!)

lib/tasks/publications_export.rake
```diff
@@ -3,13 +3,22 @@ namespace :publications do
   task export_for_document_collection: :environment do
     require "csv"
 
-    filepath = Rails.root.join("publictions_export.csv")
+    filepath = Rails.root.join("publications_export.csv")
```
ChrisBAshton: Nitpick: this typo fix should be fixed up to the first commit

lib/tasks/publications_export.rake
```diff
@@ -3,13 +3,22 @@ namespace :publications do
...
-      publlication = Publication.first
-      csv << %w"title summary body"
-      csv << [publlication.title, publlication.summary, publlication.body] if publlication
+      publication = Publication.first
+      csv << %w"title summary body attachment_title attachment_filename attachment_url attachment_created_at attachment_updated_at"
+      csv << [
+        publication.title,
+        publication.summary,
+        publication.body,
+        publication.attachments[0].title,
```
ChrisBAshton: Hmmm... how does this cope with publications that have more than 1 attachment?
Or where the attachment is a HtmlAttachment rather than FileAttachment?

I could go and look - it may be that neither case applies - but would be good to have it spelled out in the commit message for future readers.

test/unit/lib/tasks/export_publication_data_test.rb
```diff
@@ -59,4 +59,17 @@ class ExportPublicationDataTest < ActiveSupport::TestCase
     assert_empty csv[0]['attachment_updated_at']
   end
 
+  test "export_for_document_collection gets the first pdf attachment" do
```
ChrisBAshton: Ah ok, this answers my earlier question. Is only getting the first attachment acceptable?

In any case, I'd be tempted to fix this up to the commit that makes a change to the rake task. Adding tests in isolation always feels a bit odd - it should be bundled with the change that caused the test to pass.

test/unit/lib/tasks/export_publication_data_test.rb
```diff
@@ -72,4 +72,19 @@ class ExportPublicationDataTest < ActiveSupport::TestCase
     assert_equal additional_attachment.url, csv[0]['attachment_url']
   end
 
+  test "export_for_document_collection gets the first pdf attachment if there are multiple" do
```
ChrisBAshton: Same here

test/unit/lib/tasks/export_publication_data_test.rb
```diff
@@ -85,6 +85,16 @@ class ExportPublicationDataTest < ActiveSupport::TestCase
     assert_equal additional_attachment1.filename, csv[0]['attachment_filename']
     assert_equal additional_attachment1.url, csv[0]['attachment_url']
   end
+  test "export_for_document_collection only exports published publications" do
```
ChrisBAshton:
```suggestion

  test "export_for_document_collection only exports published publications" do
```

lib/tasks/publications_export.rake
```diff
@@ -1,14 +1,15 @@
 namespace :publications do
   desc "Export all publications for a document collection to a CSV file."
-  task :export_for_document_collection, [:organisation_slug] => :environment do |_, args|
+  task :export_for_document_collection, [:organisation_slug, :documents_collection_slug] => :environment do |_, args|
```
ChrisBAshton: It would be good to document the rake tasks (and how to run them) in the PR description. I wasn't really expecting to see a documents_collection_slug parameter.

lib/tasks/publications_export.rake
```diff
@@ -1,19 +1,20 @@
 namespace :publications do
   desc "Export all publications for a document collection to a CSV file."
-  task :export_for_document_collection, [:organisation_slug, :documents_collection_slug] => :environment do |_, args|
+  task :export_for_document_collection, %i[organisation_slug documents_collection_slug] => :environment do |_, args|
```
ChrisBAshton: This should be patched up to previous commit

test/unit/lib/tasks/export_publication_data_test.rb
```diff
@@ -139,8 +139,19 @@ class ExportPublicationDataTest < ActiveSupport::TestCase
 
     csv = CSV.read(CSV_PATH, headers: true)
     assert_equal 1, csv.size
-    assert_equal publication1.title, csv[0]['title']
-    refute_includes csv.map { |row| row['title'] }, publication2.title
+    assert_equal publication1.title, csv[0]["title"]
+    assert_not_includes csv.map { |row| row["title"] }, publication2.title
   end
 
+  test "export_for_document_collection only exports publications for published document collections" do
```
ChrisBAshton: This test should be fixed up to the point where the corresponding app code was added

test/unit/lib/tasks/export_publication_data_test.rb
```diff
@@ -0,0 +1,161 @@
...
+      Rake.application.invoke_task "publications:export_for_document_collection[#{publication.organisations.first.slug},#{document_collection.slug}]"
+
+      assert File.exist?(CSV_PATH), "CSV file should be created"
+      csv = CSV.read(CSV_PATH, headers: true)
+      assert_equal 1, csv.size
+      assert_equal publication.title, csv[0]["title"]
+      assert_equal publication.summary, csv[0]["summary"]
+      assert_equal publication.body, csv[0]["body"]
+      assert_equal publication.attachments[0].title, csv[0]["attachment_title"]
+      assert_equal publication.attachments[0].filename, csv[0]["attachment_filename"]
```
ChrisBAshton: What do we intend to do with the filename property in the CSV?
I imagine we'd only want the attachment_url...?

TonyGDS: We need it when we build the design decision in SP. The filename is used for the clickable link which points to the URL: https://github.com/alphagov/specialist-publisher/blob/895b34f05e64d98449b5861671d098a02a0679f7/lib/tasks/bulk_import_documents_from_csv.rake#L58

test/unit/lib/tasks/export_publication_data_test.rb
TonyGDS: Sure. This was an additional written after the fact test because the `find` method on an enumeration already handles this case but I wanted to make sure the business logic was documented. Also this was in the spirit of a making small commits which I should have squashed, so complete agree with the logical grouping.

TonyGDS: In regards to getting the first attachment, in the case of the design decisions they will have one pdf attachment and I didn't think it was worth it to extract multiple docs/types for now as it would lead to multiple lines or transposed output in the csv that would have to be handled when importing into SP. But in the future it could be changed to make it more flexible.

lib/tasks/publications_export.rake
TonyGDS: Answered below

test/unit/lib/tasks/export_publication_data_test.rb
TonyGDS: Same as above, test written after the fact to document business logic

test/unit/lib/tasks/export_publication_data_test.rb
TonyGDS: This test was added for implementation code `.where(document_collections_documents: { state: :published })` in this same commit. Apologies there was a rubocop run at this time so it makes it a bit hard to read

test/unit/lib/tasks/export_publication_data_test.rb
ChrisBAshton: Makes sense - ta

lib/tasks/publications_export.rake
```diff
@@ -1,14 +1,15 @@
 namespace :publications do
   desc "Export all publications for a document collection to a CSV file."
-  task :export_for_document_collection, [:organisation_slug] => :environment do |_, args|
+  task :export_for_document_collection, %i[organisation_slug documents_collection_slug] => :environment do |_, args|
```
ChrisBAshton: Would be good to document why this arg is needed, in the commit message (thanks for doing it in the PR description!)

test/unit/lib/tasks/export_publication_data_test.rb
ChrisBAshton: Lots of unrelated lint fixes in this "only export published document collections" commit

test/unit/lib/tasks/export_publication_data_test.rb
```diff
@@ -2,156 +2,160 @@
...
-    publication = create(:published_publication, :with_file_attachment, organisations: [organisation])
-    document_collection = create(:draft_document_collection, :with_group, organisations: [organisation])
-    create(:document_collection_group_membership, document_collection_group: document_collection.groups.first, document: publication.document)
-
-    Rake.application.invoke_task "publications:export_for_document_collection[#{organisation.slug},#{document_collection.slug}]"
-
-    csv = CSV.read(CSV_PATH, headers: true)
-    assert_equal 0, csv.size
+  describe "#publications:export_for_document_collection" do
+    test "creates an empty file when there are no publications" do
```
ChrisBAshton: Would be nice to fix up these test name tweaks to the relevant commits, to do away with the need for the "clean up tests" commit entirely.

PR #10458, opened 2025-07-29
(CM-181) Update Contact Link

pezholio:
Following on from https://github.com/alphagov/publishing-api/pull/3480 this updates the name of the `contact_form` to `contact_link`, as well as adding some config, as well as updating the feature specs to add a contact form subschema.


lib/engines/content_block_manager/config/content_block_manager.yml
```diff
@@ -115,5 +115,9 @@ schemas:
         group: contact_methods
         embeddable_as_block: true
         group_order: 4
+        fields:
```
edavey, line 118: tiny typo in commit messsage, should be:

> **Use** a textarea...

PR #10470, opened 2025-07-30
[WHIT-1454] Include headers in flexible page publishing-api payload

lauraghiorghisor-tw:
The headers are used FE-side to render the contents list.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-1454)

### DIFF
--- a/app/presenters/publishing_api/flexible_page_presenter.rb
+++ b/app/presenters/publishing_api/flexible_page_presenter.rb
@@ -15,7 +15,7 @@ module PublishingApi
       content = BaseItemPresenter.new(item, update_type:).base_attributes
       content.merge!(
         details: {
-          **FlexiblePageContentBlocks::DefaultObject.new.publishing_api_payload(type.schema, item.flexible_page_content),
+          **flatten_headers(FlexiblePageContentBlocks::DefaultObject.new.publishing_api_payload(type.schema, item.flexible_page_content)),
         },
         document_type: type.settings["publishing_api_document_type"],
         public_updated_at: item.public_timestamp || item.updated_at,
@@ -38,6 +38,17 @@ module PublishingApi
     end
 
   private
+
+    def flatten_headers(content)
+      if content[:body]
+        html = content.dig(:body, :html)
+        headers = content.dig(:body, :headers)
+        content[:body] = html
+        content[:headers] = headers
+      end
+
+      content.compact
+    end
 
     def type
       item.type_instance
--- a/test/unit/app/presenters/publishing_api/flexible_page_presenter_test.rb
+++ b/test/unit/app/presenters/publishing_api/flexible_page_presenter_test.rb
@@ -143,4 +143,52 @@ class PublishingApi::FlexiblePagePresenterTest < ActiveSupport::TestCase
     assert_equal expected_hash_for_chunk_of_content_two[:html], content[:details][:chunk_of_content_two][:html].squish
     assert_equal expected_hash_for_chunk_of_content_two[:headers], content[:details][:chunk_of_content_two][:headers]
   end
+
+  test "it includes headers once, one layer up, if there is a govspeak block with a 'body' key" do
+    type_key = "test_type_key"
+    FlexiblePageType.setup_test_types({
+      type_key => {
+        "key" => type_key,
+        "schema" => {
+          "type" => "object",
+          "title" => "An object",
+          "properties" => {
+            "body" => {
+              "title" => "Body",
+              "description" => "The main content for the page",
+              "type" => "string",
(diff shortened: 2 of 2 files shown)


test/unit/app/presenters/publishing_api/flexible_page_presenter_test.rb
```diff
@@ -75,4 +75,72 @@ class PublishingApi::FlexiblePagePresenterTest < ActiveSupport::TestCase
     assert_equal page.flexible_page_content["property_one"], content[:details][:property_one]
     assert_equal page.flexible_page_content["property_two"], content[:details][:property_two]
   end
+
+  test "it includes headers for each govspeak type block" do
```
lauraghiorghisor-tw, line 79: Maybe this test is superfluous, but it really helped my figure it out.

app/presenters/publishing_api/flexible_page_presenter.rb
```diff
@@ -39,6 +39,17 @@ def edition_links
 
   private
 
+    def flatten_headers(content)
```
lauraghiorghisor-tw, line 42: This is so that we can get the POC over the line. Can remove too, not precious about it.

Throw away as this might be, there is also a point here - things that have a big chunk of body govspeak likely don't have other chunks (not in the short term road map, anyway). It is very likely that the final iteration of this has a separate headers key that includes all headers of the page, even if they might correspond to funkily presented, and distinct, chunks of govspeak. 

Or it could be that the best solution, for all cases, is to wrap the headers in the govspeak hash as per the previous commit (html:, headers:).

app/presenters/publishing_api/payload_builder/body_headings.rb
```diff
@@ -12,29 +14,7 @@ def initialize(item)
...
-      def body_headings
-        if item
-          body_headings = Govspeak::Document.new(item.body).structured_headers
-          remove_empty_headers(body_headings.map(&:to_h))
-        else
-          []
-        end
-      end
-
-      def remove_empty_headers(body_headings)
```
lauraghiorghisor-tw, line 33: This one's a bit annoying - would be nice if we did not need to do this.

app/helpers/presenters/publishing_api/payload_headings_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+
+        body_headings = Govspeak::Document.new(govspeak).structured_headers
+        headers = remove_empty_headers(body_headings.map(&:to_h))
+
+        headers.empty? ? {} : { headers: headers }
+      end
+
+    private
+
+      def remove_empty_headers(body_headings)
```
ryanb-gds, line 15: We really ought to move this into the gem, I think 🤔 would save us from having this helper. Maybe worth opening a PR? Could make it optional, but I don't know why we'd ever want a list of empty headings

app/presenters/publishing_api/flexible_page_presenter.rb
ryanb-gds: I'd probably rather go with the derived field abstraction that do this, but we can live with it for now. The problem is... will one of us ever come back and change it? Maybe...

test/unit/app/models/flexible_page_content_blocks/default_object_test.rb
```diff
@@ -25,7 +25,7 @@ class FlexiblePageContentBlocks::DefaultObjectTest < ActiveSupport::TestCase
       },
     }
     payload = FlexiblePageContentBlocks::DefaultObject.new.publishing_api_payload(schema, content)
-    assert_equal(Whitehall::GovspeakRenderer.new.govspeak_to_html(content["test_attribute"]), payload[:test_attribute])
+    assert_equal(Whitehall::GovspeakRenderer.new.govspeak_to_html(content["test_attribute"]), payload[:test_attribute][:html])
```
ryanb-gds, line 28: I was a bit lazy, I should have created a test block class here so that a change to the Govspeak block API didn't break this test. Bonus points if you feel like giving that a go.

ryanb-gds (approved):
I'm still a bit dubious about the worth of flattening out the headers, but if you're really strongly against changing our currently very informal API contract then I guess we can live with it

app/presenters/publishing_api/flexible_page_presenter.rb
lauraghiorghisor-tw: If we never think about these blocks again, then we have bigger problems 😅

I kind of just want to meet the FE code sooner rather than later and call this POC stage done. I've shared this PR on FP channel.

PR #10486, opened 2025-08-04
Fix `500` for `WorldwideOrganisation` offices [WHIT-1843]

patrickpatrickpatrick:
## What

Use 

```
WorldwideOrganisation.find(params[:worldwide_organisation_id])
```

instead of

```
Edition.find(params[:worldwide_organisation_id])
```

## Why

This error was occurring because `find_worldwide_organisation` used `Edition.find` instead of `WorldwideOrganisation.find`. This meant that an Edition could exist with the `worldwide_organisation_id` in the param but it might be not actually be a `WorldwideOrganisation`. This would then cause an error when `offices` was called on the `Edition`.

### DIFF
--- a/app/controllers/admin/worldwide_offices_controller.rb
+++ b/app/controllers/admin/worldwide_offices_controller.rb
@@ -108,7 +108,7 @@ private
   end
 
   def find_worldwide_organisation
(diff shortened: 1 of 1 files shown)


matthillco (approved):
Looks good to me.

PR #10487, opened 2025-08-04
CM-283: render GovSpeak content in workflow steps

edavey:


lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema/embedded_schema.rb
```diff
@@ -40,6 +40,20 @@ def permitted_params
...
         end
 
+        def govspeak_enabled?(field:, nested_object_key: nil)
+          return top_level_govspeak_enabled_fields.include?(field) unless nested_object_key
+
+          govspeak_enabled_fields_for_nested_object(nested_object_key).include?(field)
+        end
+
+        def govspeak_enabled_fields_for_nested_object(object_key)
+          body.dig("properties", object_key, "govspeak_enabled") || []
```
pezholio: We'll need to update this to `x-govspeak_enabled` once https://github.com/alphagov/publishing-api/pull/3493 is updated

lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema/embedded_schema.rb
```diff
@@ -40,6 +40,20 @@ def permitted_params
...
+        end
+
+        def top_level_govspeak_enabled_fields
+          body.dig("properties", "govspeak_enabled") || []
```
pezholio: As above, we'll need to make sure we rename this

pezholio: Might be worth having a `GOVSPEAK_ENABLED_PROPERTY_KEY` constant so we don't have to reference it every time

lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema/embedded_schema.rb
```diff
@@ -40,6 +40,20 @@ def permitted_params
           end
         end
 
+        def govspeak_enabled?(field:, nested_object_key: nil)
```
pezholio: I wonder if this makes more sense to sit in the `Schema::Field` class?

pezholio: Actually, scratch that, as we don't have access to the `Field` object here. With that in mind, it might be better to call the `field` argument `field_name` to make it clearer it's a string we're passing in. Oh! My kingdom for a type checker!

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card/nested_item_component.rb
```diff
@@ -21,10 +25,14 @@ def title
   end
 
   def rows
-    nested_items.map do |key, value|
+    nested_items.map do |field, value|
```
pezholio: Wonder if we should rename this to `field_name`, so we're clearer this will be a string, rather than a field object.

pezholio (commented):
Looks to be on the right track - cheers! Just some minor tweaks :)

lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema/embedded_schema.rb
edavey: 👍

lib/engines/content_block_manager/app/components/content_block_manager/shared/embedded_objects/summary_card/nested_item_component.rb
edavey: good idea, thks, I've renamed `field` -> `field_name`

lib/engines/content_block_manager/app/models/content_block_manager/content_block/schema/embedded_schema.rb
edavey: done, thks

PR #10497, opened 2025-08-06
(CM-292) Add recipient field to address

pezholio:
Requires https://github.com/alphagov/publishing-api/pull/3501 first

Before, we treated the “title” as a “recipient” field, but going forward we want the title to describe the object itself, so we’re adding a specific field for the recipient and removing the translation


lib/engines/content_block_manager/config/locales/en.yml
```diff
@@ -87,8 +87,6 @@ en:
             call_charges_info_url: "URL to find out about call charges"
         contact_links:
           url: URL
-        addresses:
-          title: Recipient
```
edavey, line 91: tiny typo in the commit msg: "translation"

PR #10504, opened 2025-08-06
Remove the world editor and world writer roles from Whitehall [WHIT-2377]

ryanb-gds:
## What

These roles are no longer used. There are some GDS users that have them in SignOn, but all of the users also have the GDS editor or admin role, meaning that the world writer and editor roles are redundant.

Here's the list of active users in signon with the world editor or world writer role: https://signon.publishing.service.gov.uk/users?filter=&statuses%5B%5D=active&option-select-filter=&option-select-filter=Whitehall+Wo&permissions%5B%5D=1145&permissions%5B%5D=1146

## Why

It is helpful to simplify the edition authorisation model as much as possible before introducing the ability to configure edition type access via the content type configuration in [WHIT-2377](https://gov-uk.atlassian.net/browse/WHIT-2377)


app/models/user.rb
ChrisBAshton, line 1: > These roles are no longer used. There are some GDS users that have them in SignOn, but all of the users also have the GDS editor or admin role, meaning that the world writer and editor roles are redundant.

Should we run a Signon data migration to remove the roles from users so that they're expunged forever and we don't stumble across them in future and wonder what they're for?

ChrisBAshton (approved):
Nice find! Gold star ⭐

app/models/user.rb
ryanb-gds: Yes, I think so. I think there might be a few other permissions that need cleaning up as well.

PR #10505, opened 2025-08-06
[WHIT-2392] Revert "Move 255 URL maxlength limit into the UriValidator"

lauraghiorghisor-tw:
## What

Reverts alphagov/whitehall#10491

## Why

We got a zendesk ticket relating to this, because we have now made organisations invalid (due to the `custom_jobs_url` being too long) https://govuk.zendesk.com/agent/tickets/6192793.

URI character limit validation was introduced in #10491, extending the validation to all URIs in the codebase. Previously, it was set per model, e.g. for `alternative_url` on `nation_inapplicabilities`.

We investigated what data types the various urls we use are. Here's a list:
```
external_url - all string
link_url - all string
contact_form_url - all string
alternative_url - string for nation_inapplicabilities but a text for unpublishings
organisation_chart_url - all string
url - some string, some text for social_media_account_translations, featured_link_translations
redirect_url - all string
custom_jobs_url - text for organisations
other fields named `<>_url` are all string type
```

So there are only a few outliers that are of `text` type and should thus not be limited to a 255 character limit. 
It seems we should not proceed with applying the validation universally. If we do so,  we should correct the data before applying the validation. 

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2392)

### DIFF
--- a/app/models/nation_inapplicability.rb
+++ b/app/models/nation_inapplicability.rb
@@ -10,6 +10,7 @@ class NationInapplicability < ApplicationRecord
 
   validates :nation_id, inclusion: { in: Nation.potentially_inapplicable.map(&:id) }
   validates :alternative_url, uri: true, allow_blank: true
+  validates :alternative_url, length: { maximum: 255 }
 
   attr_accessor :excluded
 
--- a/app/validators/uri_validator.rb
+++ b/app/validators/uri_validator.rb
@@ -3,15 +3,8 @@ require "addressable/uri"
 
 # Accepts options[:message] and options[:allowed_protocols]
 class UriValidator < ActiveModel::EachValidator
-  MAX_LENGTH = 255
-
   def validate_each(record, attribute, value)
     return if value.nil?
-
-    if value.length > MAX_LENGTH
-      record.errors.add(attribute, "is too long. Please shorten the URL to less than #{MAX_LENGTH} characters.")
-      return
-    end
 
     uri = URI.parse(value)
 
--- a/test/unit/app/models/nation_inapplicability_test.rb
+++ b/test/unit/app/models/nation_inapplicability_test.rb
@@ -1,9 +1,37 @@
 require "test_helper"
 
 class NationInapplicabilityTest < ActiveSupport::TestCase
-  test "uses UriValidator on :alternative_url" do
-    validators = NationInapplicability.validators_on(:alternative_url)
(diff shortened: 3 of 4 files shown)


lauraghiorghisor-tw:
> I don't think a revert is necessary. The scale of the issue is small - see details pulled from integration:

There are also:
`alternative_url` - text type for `unpublishings`
`url` - type text for `social_media_account_translations`, `featured_link_translations`
-> they need to be checked!

I was chatting to Ryan yesterday and he did make a valid point, which was that there is no need to limit character for URLs really - and we can see by the choice of data type that these were intentionally not limited before (well, as intentional as it gets in this codebase...). 

I like that we can have all the validation logic in one place which is neat (as per your original PR) and we should if the usage is limited. Nonetheless, a case could be made that even the original `alternative_url` should not be validated for length.

ChrisBAshton:
That's fair, @lauraghiorghisor-tw - I'm all for removing the validation altogether.

@ryanb-gds can you remember what was being referred to in https://github.com/alphagov/whitehall/pull/9087 here?

> URLs which are longer than 255 characters break the database limit, so we are introducing a character limit for URLs to prevent this from happening. Users should seek to shorten their URLs.

I don't see any sort of a "database limit" in our schema 🤔

lauraghiorghisor-tw:
> That's fair, @lauraghiorghisor-tw - I'm all for removing the validation altogether.

Most of these URL fields in the schema are still "string" - and that should be 255 limited.
So we either 
- correct the remaining text type instances (url and alternative_url) and implement the validation universally (which enforces a limit on users, which can cause potential support tickets)
- remove validation and change the data types to text for remaining url fields too - as it does feel a bit uncontrolled that they'd remain string type with no char limit validation. 

> I don't see any sort of a "database limit" in our schema 🤔

I think it's just that the `alterntive_url` type is `string` so that's inherently limited as it maps to a `VARCHAR(255)`.

ChrisBAshton:
> I think it's just that the `alterntive_url` type is `string` so that's inherently limited as it maps to a `VARCHAR(255)`.

Ah of course - thanks for clarifying.

Looks like `t.string` is only really useful if we're wanting to index the URL, and I don't think we really have a use case for that 🤔 so dropping the validation and doing a DB migration to change all URL fields to `t.text` feels like the way to go IMO. Do you agree? I can write up a card if so 👍

ChrisBAshton:
Raised https://gov-uk.atlassian.net/browse/WHIT-2400. Let's close this in the meantime.

PR #10506, opened 2025-08-06
Raise an error when duplicate content id's are fed to bulk republish

callumknights:
This caused an issue recently which was initially misdiagnosed as an issue with providing a large list of content id's.

Only after someone had a proper look at the list was it realised the problem was a duplicate and nothing to do with the size. So this PR adds similar functionality to what is present is `search_documents_by_content_ids` to prevent this happening again.

### DIFF
--- a/app/controllers/admin/bulk_republishing_controller.rb
+++ b/app/controllers/admin/bulk_republishing_controller.rb
@@ -109,6 +109,8 @@ class Admin::BulkRepublishingController < Admin::BaseController
     content_ids = content_ids_string_to_array(@content_ids_string)
     return if redirect_if_no_requested_content_ids(content_ids)
 
+    return if redirect_if_duplicated_content_ids(content_ids)
+
     @matching_documents = Document.where(content_id: content_ids)
     return if redirect_if_unfound_content_ids(@matching_documents.pluck(:content_id), content_ids)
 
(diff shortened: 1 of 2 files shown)


app/controllers/admin/bulk_republishing_controller.rb
```diff
@@ -109,6 +109,8 @@ def republish_documents_by_content_ids
     content_ids = content_ids_string_to_array(@content_ids_string)
     return if redirect_if_no_requested_content_ids(content_ids)
 
+    return if redirect_if_duplicated_content_ids(content_ids)
```
yndajas, line 112: **question:** I'd be slightly surprised if anyone's getting to this action with duplicates after `redirect_if_duplicated_content_ids` being introduced in the search action. Since that fix was only introduced two days ago, are we sure this is still an issue? I think you'd have to do some fairly manual param hacking to get here without going through the search action, which has that new guard in place already

Separately, I wonder why redirecting with an error was chosen over deduplication - perhaps to be extra super sure that the input is intentional?

callumknights: I think you might be right, and I think I've slightly misunderstood the flow of the how the republishing features work, and combined with me not realising how recently the search action fix was implemented, this is probably completely unnecessary. If you agree I'd be happy to close this then?

yndajas: Yeah I reckon so - might just be worth checking with Patrick who wrote that fix?

callumknights:
Closing as a fix in already in place

PR #10508, opened 2025-08-07
Send only the English version in world_location_names

catalinailie:
On non-english worldwide organisation pages, the link that should send the user to the world location news page is not working. For example,  see Česká republika on this [embassy page](https://www.gov.uk/world/organisations/british-embassy-prague.cs), that links to `/world/ceska-republika/news` , but it should be `/world/czechia/news`.
This is because in government-frontend we are using the country name from world_location_names, and content data returns the native country name for translated pages, and not the English version, which is the one we need.

Because `world_locations` doesn't return only the English version for location name, this issue is affecting links on Woldwide Organisation, News Articles, Case Studies and Speech pages.

Trello card: https://trello.com/c/dXfX6R0o/3433-fix-the-broken-link-in-the-head-of-some-non-english-worldwide-organisation-pages-that-links-to-the-world-location-news-page

### DIFF
--- a/app/presenters/publishing_api/worldwide_organisation_presenter.rb
+++ b/app/presenters/publishing_api/worldwide_organisation_presenter.rb
@@ -166,7 +166,7 @@ module PublishingApi
       item.world_locations.map do |world_location|
         {
           content_id: world_location.content_id,
-          name: world_location.name,
+          name: world_location.translations.find_by(locale: "en")&.name || world_location.name,
         }
       end
     end


catalinailie:
Actually I'm not really sure this is the correct solution. `world_location_names` is expected to store the native version of location name. (see [commit](https://github.com/alphagov/publishing-api/commit/f6baa92af2288535d977a7ce9392dc87dda5fe4e) where it was added) And it seems that `world_locations` is expected to have the locations in English. 
At the moment both of them store the location in the language that the page is currently being rendered in. Not sure why or when the behaviour of `world_locations` has changed.

PR #10514, opened 2025-08-07
[WHIT-2257] Surface nested validation errors

GDSNewt:
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-1517)

### Why

This refactoring improves the user experience of Whitehall's publishing workflow—specifically in how embedded contact IDs in Govspeak content are validated within both Editions and HTML attachments.

Previously, Whitehall allowed users to create and publish content containing HTML attachments that may have become invalid over time. This could happen due to:

- Validation rules being introduced or tightened after initial publication
- Contacts being deleted or changed, rendering existing embeds invalid

As a result, users might only discover issues at publish time(usually when making a new edition)—despite having made no changes to the attachment.

### What
This PR introduces a context-aware validation system for embedded contacts in Govspeak content that:

- Surfaces clear, specific errors in the UI when validation fails
- Skips certain validations during draft creation to unblock editors from fixing invalid editions

**BEFORE**
1 - validation error on html attachment

**AFTER**
Screenshot 2025-08-15 at 09 48 23

**BEFORE**
the problem

**AFTER**
Screenshot 2025-08-15 at 09 47 38

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -146,15 +146,4 @@ module ApplicationHelper
   def get_content_id(edition)
     edition&.content_id
   end
-
-  def filtered_validation_errors(edition)
-    all_messages = edition.errors.map(&:full_message)
-    has_contact_errors = all_messages.any? { |msg| msg.include?("Contact ID") || msg.include?("contact reference") }
-
-    if has_contact_errors
-      all_messages.reject { |msg| msg.match?(/\A(Attachments|Html attachments) is invalid\z/) }
-    else
-      all_messages
-    end
-  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -144,10 +144,17 @@ module ApplicationHelper
   end
 
   def get_content_id(edition)
-    return if edition.nil?
+    edition&.content_id
+  end
 
-    return unless edition.respond_to?("content_id")
+  def filtered_validation_errors(edition)
+    all_messages = edition.errors.map(&:full_message)
+    has_contact_errors = all_messages.any? { |msg| msg.include?("Contact ID") || msg.include?("contact reference") }
 
-    edition.content_id
+    if has_contact_errors
+      all_messages.reject { |msg| msg.match?(/\A(Attachments|Html attachments) is invalid\z/) }
+    else
+      all_messages
+    end
   end
 end
--- a/app/models/concerns/attachable.rb
+++ b/app/models/concerns/attachable.rb
@@ -199,19 +199,34 @@ module Attachable
 private
 
   def check_html_attachments_on_publish
-    add_attachment_errors(html_attachments, "HTML Attachment") do |msg|
-      if msg.include?("Body contains invalid contact reference:")
-        msg.sub("Body contains invalid contact reference: Contact ID ", "contains invalid contact reference. Contact ")
-      else
-        "error: #{msg}"
-      end
+    add_attachment_errors(html_attachments, "HTML Attachment") do |attachment, msg|
+      format_attachment_error_message(attachment, msg)
     end
     errors.delete(:html_attachments, :invalid)
   end
 
   def check_attachments_on_publish
-    add_attachment_errors(attachments.reject { |a| a.is_a?(HtmlAttachment) }, "Attachment") { |msg| "error: #{msg}" }
+    add_attachment_errors(attachments.reject(&:html?), "Attachment") do |attachment, msg|
+      format_attachment_error_message(attachment, msg)
+    end
     errors.delete(:attachments, :invalid)
+  end
+
+  def format_attachment_error_message(attachment, msg)
+    if has_invalid_contact_error?(attachment)
+      contact_id = attachment.errors.details[:base].find { |error| error[:error] == :invalid_contact }[:contact_id]
+      I18n.t("errors.messages.invalid_contact_reference", contact_id: contact_id)
+    else
+      I18n.t("errors.messages.attachment_error", msg: msg)
+    end
+  end
+
+  def has_invalid_contact_error?(attachment)
+    attachment.errors.details[:base].any? { |error| error[:error] == :invalid_contact }
+  end
+
+  def attachment_identifier(attachment)
+    attachment.title.present? ? "'#{attachment.title}'" : attachment.id
   end
 
   def add_attachment_errors(collection, prefix)
@@ -219,9 +234,8 @@ private
       next if attachment.valid?
 
       attachment.errors.full_messages.each do |msg|
-        identifier = attachment.title.present? ? "'#{attachment.title}'" : attachment.id
-        formatted_msg = yield(msg)
-        errors.add(:base, "#{prefix} #{identifier} #{formatted_msg}")
+        formatted_msg = yield(attachment, msg)
+        errors.add(:base, "#{prefix} #{attachment_identifier(attachment)} #{formatted_msg}")
       end
     end
   end
--- a/app/models/concerns/attachable.rb
+++ b/app/models/concerns/attachable.rb
@@ -45,6 +45,7 @@ module Attachable
     end
 
     validate :check_html_attachments_on_publish, on: :publish
+    validate :check_attachments_on_publish, on: :publish
 
     def attachments_ready_for_publishing
       attachments.select { |attachment| !attachment.file? || attachment.attachment_data.all_asset_variants_uploaded? }
@@ -198,12 +199,29 @@ module Attachable
 private
 
   def check_html_attachments_on_publish
-    html_attachments&.each do |attachment|
+    add_attachment_errors(html_attachments, "HTML Attachment") do |msg|
+      if msg.include?("Body contains invalid contact reference:")
+        msg.sub("Body contains invalid contact reference: Contact ID ", "contains invalid contact reference. Contact ")
+      else
+        "error: #{msg}"
+      end
+    end
+    errors.delete(:html_attachments, :invalid)
+  end
+
+  def check_attachments_on_publish
+    add_attachment_errors(attachments.reject { |a| a.is_a?(HtmlAttachment) }, "Attachment") { |msg| "error: #{msg}" }
+    errors.delete(:attachments, :invalid)
+  end
+
+  def add_attachment_errors(collection, prefix)
+    collection&.each do |attachment|
       next if attachment.valid?
 
       attachment.errors.full_messages.each do |msg|
         identifier = attachment.title.present? ? "'#{attachment.title}'" : attachment.id
-        errors.add(:base, "HTML Attachment #{identifier} error: #{msg}")
+        formatted_msg = yield(msg)
+        errors.add(:base, "#{prefix} #{identifier} #{formatted_msg}")
       end
     end
   end
--- a/app/models/concerns/attachable.rb
+++ b/app/models/concerns/attachable.rb
@@ -36,12 +36,15 @@ module Attachable
         def process_associations_after_save(edition)
           @edition.attachments.each do |attachment|
             draft_attachment = attachment.deep_clone
+            draft_attachment.instance_variable_set(:@created_during_draft, true)
             edition.attachments << draft_attachment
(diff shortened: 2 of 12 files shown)


app/models/html_attachment.rb
```diff
@@ -10,6 +10,7 @@ class HtmlAttachment < Attachment
   before_save :ensure_slug_is_valid
 
   validates :govspeak_content, presence: true
+  validates_with GovspeakContactEmbedValidator
```
ChrisBAshton: Would be good to elaborate on the first commit (for the reviewer's sake) to make it clear why the reversion is necessary.

app/models/concerns/attachable.rb
```diff
@@ -36,8 +36,9 @@ module Attachable
         def process_associations_after_save(edition)
           @edition.attachments.each do |attachment|
             draft_attachment = attachment.deep_clone
+            draft_attachment.instance_variable_set(:@created_during_draft, true)
```
ChrisBAshton, line 39: I can sort of figure out what this is doing, but it would be easier to review (and best practice) if there were an associated test in the same commit. It'd make it clear under what conditions we expect validation to happen, vs not.

app/validators/govspeak_contact_embed_validator.rb
```diff
@@ -1,8 +1,51 @@
...
+    end
+  end
+
+  def build_context_hint_for_attachment(record)
+    edition = record.respond_to?(:attachable) ? record.attachable : nil
+    document = edition&.document
+    return "" unless document&.editions&.any?
+
+    if document.editions.count == 1
+      ". You may need to create this contact first or use an existing contact ID"
```
ChrisBAshton: I'm a bit nervous about having branches of code like this that are untested (no corresponding test for "You may need to create this contact first").

It's a nice UX, and something I'd be open to if there are tests to back it up, but in the first instance I'd also be quite happy to strip the main error message back to something like "HTML attachment 'attachment title' has validation issues: Contact 123 doesn't exist" (without any additional contextual hint needed). Certainly the contextual stuff could/should be a later self-contained commit.

app/validators/govspeak_contact_embed_validator.rb
```diff
@@ -1,8 +1,51 @@
...
+
+private
+
+  def add_error_for_invalid_contact(record, contact_id)
+    case record
+    when Edition
+      record.errors.add(:body, "Contact ID #{contact_id} doesn't exist")
+    when HtmlAttachment
+      identifier = record.title.present? ? "'#{record.title}'" : record.id
+      base_message = "HTML Attachment #{identifier} contains invalid contact reference: Contact ID #{contact_id} doesn't exist"
```
ChrisBAshton: It feels to me like this is happening at the wrong level. The `GovspeakContactEmbedValidator` should only care about receiving a body, checking it for bad contact IDs, and reporting back on those. It shouldn't have knowledge of what it has been passed (Edition vs HtmlAttachment) as that violates the separation of concerns principle.

The logic around skipping the validation should probably be at the model level, i.e.

```ruby
class Edition < ApplicationRecord
  validates :body, govspeak_contact_embed: true, if: -> { validation_context == :publish }
end

class HtmlAttachment < ApplicationRecord
  validates :body, govspeak_contact_embed: true, unless: :skip_contact_validation?
end
```

And finally, the validation error message should probably be constructed outside of the validator. This is a bit of a pain to bubble up manually, but should be doable with something like:

```ruby
class HtmlAttachment < ApplicationRecord
  validates :body, govspeak_contact_embed: true

  def invalid_contact_messages
    errors.details[:body]
      .select { |detail| detail[:error] == :invalid_contact }
      .map { |detail| "Contact #{detail[:contact_id]} doesn't exist" }
  end
```

...and then on the Attachable module, something like

```ruby
 validate :check_html_attachments_on_publish, on: :publish

  private

  def check_html_attachments_on_publish
    html_attachments.each do |attachment|
      attachment.invalid_contact_messages.each do |msg|
        errors.add(:base,
          "Unable to publish. Problem HtmlAttachment '#{attachment.title}': #{msg}")
      end
    end
  end
```

(I think that means the `Implement error filtering and improve sidebar display` commit would be made redundant too, as we'd again only be surfacing 'edition' validation issues, even if really those issues are bubbled up from the html attachment)

test/unit/app/validators/govspeak_contact_embed_validator_test.rb
ChrisBAshton, line 1: Some excellent new tests in here 👏

app/models/concerns/attachable.rb
```diff
@@ -191,4 +194,23 @@ def reorder_attachments(ordered_attachment_ids)
...
+
+private
+
+  def check_html_attachments_on_publish
+    return unless respond_to?(:html_attachments)
+
+    html_attachments.each do |attachment|
+      next if attachment.valid?
+
+      attachment.invalid_contact_messages.each do |msg|
```
ChrisBAshton: I realise this is exactly what I [suggested earlier](https://github.com/alphagov/whitehall/pull/10514#discussion_r2260653469), but in hindsight, it feels a bit weird for Attachable to know specifically about contact validation. It should know about Html Attachments, because those are a type of Attachment, which Attachable defines the behaviour for, but we might validate all kinds of things on HTML Attachments beyond just contact embedding.

So I think this method needs genericising to something like:

```ruby
def check_html_attachments_on_publish
  html_attachments&.each do |attachment|
    next if attachment.valid?

    attachment.errors.full_messages.each do |msg|
      identifier = attachment.title.present? ? "'#{attachment.title}'" : attachment.id
      errors.add(:base, "HTML Attachment #{identifier} error: #{msg}")
```

app/models/html_attachment.rb
```diff
@@ -104,8 +104,23 @@ def publishing_api_presenter
...
+      .map { |detail| "Contact #{detail[:contact_id]} doesn't exist" }
+  end
+
 private
 
+  def skip_contact_validation?
+    return true if instance_variable_get(:@created_during_draft)
+
+    return false unless attachable.respond_to?(:draft?) && attachable.draft?
+    return false unless attachable.document&.ever_published_editions&.exists?
```
ChrisBAshton: Hmmm shouldn't we drop this line? I imagine we still want to surface missing contact embeds even if the parent document has been published.

app/models/html_attachment.rb
```diff
@@ -104,8 +104,23 @@ def publishing_api_presenter
...
+  def invalid_contact_messages
+    errors.details[:body]
+      .select { |detail| detail[:error] == :invalid_contact }
+      .map { |detail| "Contact #{detail[:contact_id]} doesn't exist" }
+  end
+
 private
 
+  def skip_contact_validation?
+    return true if instance_variable_get(:@created_during_draft)
```
ChrisBAshton: Isn't this made redundant by line 118 below? 🤔 I'd have thought at the point the HTML attachment is created in `process_associations_after_save` it should be responding to `.draft?` with `true`?

ChrisBAshton (commented):
Shaping up nicely, Alex 👍 last couple of comments below

app/models/html_attachment.rb
```diff
@@ -104,8 +104,20 @@ def publishing_api_presenter
...
+  def invalid_contact_messages
+    errors.details[:base]
+      .select { |detail| detail[:error] == :invalid_contact }
+      .map { |detail| "Contact #{detail[:contact_id]} doesn't exist" }
+  end
+
 private
 
+  def skip_contact_validation?
+    return false unless attachable.respond_to?(:draft?) && attachable.draft?
```
ChrisBAshton:
```suggestion
    return false unless attachable&.draft?
```

app/models/html_attachment.rb
```diff
@@ -104,8 +104,20 @@ def publishing_api_presenter
     PublishingApi::HtmlAttachmentPresenter
   end
 
+  def invalid_contact_messages
+    errors.details[:base]
+      .select { |detail| detail[:error] == :invalid_contact }
+      .map { |detail| "Contact #{detail[:contact_id]} doesn't exist" }
+  end
```
ChrisBAshton: I think we agreed to delete this?
(And remove it from the commit message too)

app/helpers/application_helper.rb
```diff
@@ -144,10 +144,17 @@ def starts_with_vowel?(word_or_phrase)
...
+  end
 
-    return unless edition.respond_to?("content_id")
+  def filtered_validation_errors(edition)
+    all_messages = edition.errors.map(&:full_message)
+    has_contact_errors = all_messages.any? { |msg| msg.include?("Contact ID") || msg.include?("contact reference") }
 
-    edition.content_id
+    if has_contact_errors
+      all_messages.reject { |msg| msg.match?(/\A(Attachments|Html attachments) is invalid\z/) }
```
ChrisBAshton: Sorry for throwing another spanner in the works - I've got a sequence of thoughts to feed back on this one:

1. If we keep this, we can drop the conditional, now that we've genericised the attachable so that we're just bubbling up _all_ errors to the parent edition ("Html attachment 'foo' has error: 'this could be anything'"). So `filtered_validation_errors` should really just drop all "Attachments|Html attachments is invalid" errors.
2. BUT you have to be careful there, as we're currently missing an equivalent `check_attachments_on_publish` (non HTML version), meaning we could get into a situation where an attachment is invalid and if we try to publish an edition we get an empty error message.
3. Taking a step back, I'm not sure we should be defining a global `filtered_validation_errors` helper to filter out those base error messages. I think the Edition model (or better yet, its Attachable module) should be responsible for filtering out its own irrelevant error messages.

TLDR: we should delete the `filtered_validation_errors` method, and expand on the work you've already done to the Attachable module. E.g. in the `check_html_attachments_on_publish` method, add `errors.delete(:html_attachments, :invalid)`.

(Does that make sense? Happy to chat on a call about this or pair if it's useful).

app/models/concerns/attachable.rb
```diff
@@ -36,13 +36,17 @@ module Attachable
         def process_associations_after_save(edition)
           @edition.attachments.each do |attachment|
             draft_attachment = attachment.deep_clone
+            draft_attachment.instance_variable_set(:@created_during_draft, true)
             edition.attachments << draft_attachment
-            raise "Your edition failed to create successfully. Please contact GOV.UK support." unless draft_attachment.save!
+            draft_attachment.save!
```
ryanb-gds, line 41: Is there ever actually a scenario where we would want to validate this? Would it be acceptable to just skip validation rather than hacking around with instance vars?

app/models/concerns/attachable.rb
```diff
@@ -191,4 +195,34 @@ def reorder_attachments(ordered_attachment_ids)
       end
     end
   end
+
+private
+
+  def check_html_attachments_on_publish
+    add_attachment_errors(html_attachments, "HTML Attachment") do |msg|
+      if msg.include?("Body contains invalid contact reference:")
```
ryanb-gds: I don't really like that we're matching on a raw string in the error message here, if we were to change the govspeak contact validator error message later we'd subtly break this without realising. Can we check for the `:invalid_contact` error key instead somehow?

app/models/html_attachment.rb
```diff
@@ -105,6 +106,15 @@ def publishing_api_presenter
 
 private
 
+  def skip_contact_validation?
+    return false if persisted? && !@created_during_draft
+
+    @created_during_draft ||
+      (attachable.respond_to?(:draft?) && attachable.draft? &&
+       attachable.respond_to?(:document) && attachable.document.present? &&
+       !attachable.document.editions.published.empty?)
```
ryanb-gds: What do these extra checks add that the @created_during_draft flag doesn't cover?

app/models/concerns/attachable.rb
```diff
@@ -191,4 +195,34 @@ def reorder_attachments(ordered_attachment_ids)
...
+        msg.sub("Body contains invalid contact reference: Contact ID ", "contains invalid contact reference. Contact ")
```
ryanb-gds: Would be good to make this translatable. I know a lot of Whitehall error messages aren't, but we should be aiming to move in that direction

ChrisBAshton:
Superseded by #10588

PR #10515, opened 2025-08-07
CM-285 implement Govspeak-enabled textarea component for nested objects

edavey:
See Jira [CM-285](https://gov-uk.atlassian.net/browse/CM-285)

In this PR we extract 2 `govspeak_enabled` textareas from 2 of the "Contact" content-block's nested `telephones` items to a new `GovspeakEnabledTextAreaComponent`. This component:

- introduces a 'Govspeak supported' indicator. If the field in question is declared to be  `govspeak_enabled`
in the subschema we add a [label hint element][] to the textarea's `label` element.

- will (in a future PR) add the facility to "Preview" govspeak content as rendered HTML

### Screenshot

govspeak_supported_textareas_hint

### A note on aria-describedby
The GOV.UK Design System requires that a "hint" element has
an `id` whose value matches that of the `aria-describedby`
attribute of the input to which it relates e.g. as per
[their example][]:

```html
<div class="govuk-form-group">
  <h1 class="govuk-label-wrapper">
    <label
      class="govuk-label govuk-label--l"
      for="event-name"
      >
      What is the name of the event?
    </label>
  </h1>
  <div
    id="event-name-hint"
    class="govuk-hint"
    >
    The name you’ll use on promotional material
  </div>
  <input
    class="govuk-input"
    id="event-name"
    name="eventName"
    type="text"
    aria-describedby="event-name-hint"
    >
</div>
```

They write that:

> Hint text must:
>
> - be associated with an input. The `hint_id` must match the
> `aria-describedby` property on the input your label is associated
> with.
>
> If hint text is within a label it will be announced in its
> entirety by screen readers. By putting the hint alongside labels
> and associating hints with inputs using `aria-describedby`,
> then screen readers will read the label, describe the type of
> input (eg radio) and then read additional text. It means users
> of screen readers can scan and skip options as easy as people
> making choices with sight. [A discussion of this approach](https://github.com/alphagov/govuk_elements/issues/574)

[label hint element]:
https://components.publishing.service.gov.uk/component-guide/label#with_hint

[their example]:
https://design-system.service.gov.uk/components/text-input/input-hint-text/

[CM-285]: https://gov-uk.atlassian.net/browse/CM-285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.rb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.rb
@@ -1,0 +1,49 @@
+class ContentBlockManager::ContentBlockEdition::Details::Fields::GovspeakEnabledTextareaComponent < ContentBlockManager::ContentBlockEdition::Details::Fields::BaseComponent
+  def initialize(content_block_edition:, value:, nested_object_key:, field:, subschema:)
+    @nested_object_key = nested_object_key
+    @field = field
+    @value = value
+    @content_block_edition = content_block_edition
+
+    super(
+      subschema: subschema,
+      content_block_edition: content_block_edition,
+      field: field,
+      value: value,
+    )
+  end
+
+  attr_reader :content_block_edition, :nested_object_key, :field, :subschema, :errors
+
+  def id_attribute
+    "#{PARENT_CLASS}_details_#{subschema_block_type}_#{nested_object_key}_#{field.name}"
+  end
+
+  def label
+    helpers.humanized_label(
+      relative_key: field.name,
+      root_object: "#{subschema_block_type}.#{nested_object_key}",
+    )
+  end
+
+  def name_attribute
+    "content_block/edition[details][#{subschema_block_type}][#{nested_object_key}][#{field.name}]"
+  end
+
+  def value_or_default
+    value_for_field(field) || field.default_value
+  end
+
+  def error_items
+    errors_for(
+      content_block_edition.errors,
+      "details_#{subschema_block_type}_#{nested_object_key}_#{field.name}".to_sym,
+    )
+  end
+
+private
+
+  def value_for_field(field)
+    value&.fetch(field.name, nil)
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.rb
```diff
@@ -15,10 +15,24 @@ def initialize(content_block_edition:, value:, nested_object_key:, field:, subsc
...
   end
 
+  def aria_described_by
+    "#{id_attribute}-hint"
+  end
+
+  def hint_text
+    return nil unless govspeak_enabled?
+
+    "Markdown supported"
```
pezholio: Just wondering if we should refer to "Govspeak" rather than "Markdown". Is "Govspeak" better understood, or do content designers see the terms as interchangable?

edavey: thanks for this, I've now made it "Govspeak" all the way down!

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.rb
```diff
@@ -15,10 +15,24 @@ def initialize(content_block_edition:, value:, nested_object_key:, field:, subsc
...
+  def govspeak_enabled?
+    subschema.govspeak_enabled?(nested_object_key: nested_object_key, field_name: field.name)
+  end
+
   def id_attribute
     "#{PARENT_CLASS}_details_#{subschema_block_type}_#{nested_object_key}_#{field.name}"
   end
 
+  def aria_described_by
+    "#{id_attribute}-hint"
```
pezholio, line 27: Good catch on this - I wonder if this is something we're missing elsewhere?

pezholio (approved):
LGTM - going forward I wonder if we can automate this for any field with `x-govspeak_enabled` set?

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.rb
edavey: as far as I can see there are (only) 3 other uses of the hint, though I believe this is about to expand following the latest round of user research. I've created a bug ticket to fix this up, in `StringComponent` and `TextareaComponent` https://gov-uk.atlassian.net/browse/CM-331

PR #10516, opened 2025-08-07
Stop 'invalid' organisations from blocking edits to publications (WHIT-2396)

ChrisBAshton:
If an organisation is invalid (e.g. has a custom_jobs_url that is longer than 255 characters, which is one of our current restrictions) then `organisation.valid?` is `false`. This is good and expected.

If a publisher tries to edit a document where one of its associated organisations is the 'bad' organisation, the sidebar will have the contents "Edition is invalid" => "Organisations is invalid". Edits of the edition are possible, but publishing is blocked.

This is an overstep of the organisation validation. The organisation is already live and the edition's relation to it is, frankly, minimal. Publishers should not be prevented from editing their documents just because the object at the end of a bit of metadata isn't quite perfect.

We therefore need to stop the organisation validation logic from being called when evaluating `edition.valid?(:publish)`. A very puzzling thing is that, calling `Edition.find(<id>).valid?(:publish)` of a known problematic edition in the rails console would return `true` rather than `false`. Digging into the issue further locally, we were able to determine that that same edition returns `false` on `.valid?(:publish)` after the following bit of code is evaluated in `admin/editions/show/_main.html.erb`:

```
@edition.organisations.map { |o| o.name }) }
```

That map { |o| o.name } forces ActiveRecord to load each Organisation and call the name accessors. Because Organisation uses globalize translations (translates :name, :logo_formatted_name, :acronym), a missing translation will be lazily built on the fly when name is read. The freshly built translation is unsaved, so the associated Organisation becomes "dirty".

The fundamental issue is that Rails automatically defines methods named `validate_associated_records_for_<association>` whenever an association is set to autosave or validate its children. For Edition this happens for both the edition_organisations and organisations associations. Those methods run validations on each associated record, so as soon as you read a translated attribute on an Organisation (which builds an unsaved translation), Rails will attempt to validate the organisation when you later call `edition.valid?(:publish)`. If any organisation is invalid, the entire edition becomes invalid.

The simplest way to stop an invalid organisation from blocking publication is to override those auto-generated validation methods to no-ops. This prevents Rails from running validations on the associated organisations or edition_organisations when validating the edition.

JIRA: https://gov-uk.atlassian.net/browse/WHIT-2396

---

### DIFF
--- a/test/unit/app/models/edition/organisations_test.rb
+++ b/test/unit/app/models/edition/organisations_test.rb
@@ -43,6 +43,31 @@ class Edition::OrganisationsTest < ActiveSupport::TestCase
     end
   end
 
+  test "should remain publishable when linked organisations are invalid" do
+    # Build and persist an organisation that will be invalid.
+    invalid_org = create(:organisation)
+    invalid_org.update_column(:homepage_type, "invalid")
+    assert_not invalid_org.valid?, "Homepage type is not included in the list"
+
+    # Create an edition using the invalid organisation as the lead organisation and
+    # disable default organisation creation. Persisting the edition mirrors real usage.
+    edition = create(
+      :publication,
+      create_default_organisation: false,
+      lead_organisations: [invalid_org],
+      supporting_organisations: [],
+    )
+
+    # The edition should be valid for publishing before any side-effects.
+    assert edition.valid?(:publish), "edition should be valid in publish context before touching organisations"
+
+    # Reading organisation names used to build unsaved translations and make the edition
+    # invalid via autosave validations:contentReference[oaicite:0]{index=0}. After overriding those validations,
+    # the edition should remain valid even after accessing organisation names.
+    edition.organisations.map(&:name)
+    assert edition.valid?(:publish), "edition should still be valid in publish context after reading organisation names"
+  end
+
   test "#sorted_organisations returns organisations in alphabetical order" do
     organisation1 = create(:organisation, name: "Ministry of Jazz")
     organisation2 = create(:organisation, name: "Free Jazz Foundation")
--- a/app/models/concerns/edition/organisations.rb
+++ b/app/models/concerns/edition/organisations.rb
@@ -90,6 +90,25 @@ module Edition::Organisations
     true
   end
 
+  # Rails automatically defines methods called
+  # `validate_associated_records_for_<association>` for any association that is
+  # autosaved or validated. These methods run validations on each associated
+  # EditionOrganisation and Organisation, which causes an otherwise-valid edition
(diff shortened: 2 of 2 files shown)


test/unit/app/models/edition/organisations_test.rb
```diff
@@ -43,6 +43,34 @@ class Edition::OrganisationsTest < ActiveSupport::TestCase
     end
   end
 
+  # In: whitehall/test/unit/app/models/edition/organisations_test.rb
+  test "should remain publishable when linked organisations are invalid" do
```
lauraghiorghisor-tw, line 46: Oooh, so we DON'T want to surface these errors at all. Ah well, then no need to worry about showing better association errors as I was saying 😅 

There was some secret value in surfacing the invalid orgs in that it would create a healthy self-fix cycle as users created drafts under that organisation, but that sounds more like a pain in the neck :)

test/unit/app/models/edition/organisations_test.rb
```diff
@@ -43,6 +43,34 @@ class Edition::OrganisationsTest < ActiveSupport::TestCase
...
+    # Build and persist an organisation that will be invalid. A URL longer than 255
```
lauraghiorghisor-tw: If we are going to make the URL text type in upcoming work, maybe choose a different way to invalidate this?

lauraghiorghisor-tw (approved):
👌

lauraghiorghisor-tw:
AHA so this thing was it! I did put it down as a step on the card comment, I did wonder if that was what fixed it!

`Digging into the issue further locally, we were able to determine that that same edition returns false on .valid?(:publish) after the following bit of code is evaluated in admin/editions/show/_main.html.erb:`

test/unit/app/models/edition/organisations_test.rb
ChrisBAshton: Yeah it could have gone either way, but this was the minimal/safest change to fix the workflow issue. The alternative would have been something like adding a noop validation method on the Edition model just to actively load the organisation association and then call `.valid?` on those organisations - I'm not sure what the performance impact of those would be!

I've [bookmarked my comment](https://gov-uk.atlassian.net/browse/WHIT-2392?focusedCommentId=89931) on currently invalid orgs, and intend to take a look at those when I get a minute. I suspect their invalidation issues are few enough in number to be manually resolvable and we should now (hopefully) be properly validating that they can't be re-broken in future 😅

test/unit/app/models/edition/organisations_test.rb
ChrisBAshton: Very good point! Will revisit now.

PR #10521, opened 2025-08-08
Use multiple file uploading throughout Whitehall (WHIT-2338)

patrickpatrickpatrick:
## What

- Reimplement `AttachmentController` to be more modular and reflect models
  - AttachmentsController refactored to be a base attachment controller
  - HTMLAttachmentsController, ExternalAttachmentsController and FileAttachmentsController created
  - Split previously centralised attachment partials to specific partials for each type of attachment 
  - Move unique filename validation to `AttachmentData`
- Use `BulkUpload` for all file attachments

#### Reimplement `AttachmentController` more to be modular and reflect models 

`AttachmentController` acted as the controller for different types of attachments. This resulted in instances of branching logic within the controller and views. As we already have defined models for the different types of attachments, I thought it made sense to split out the specific logic for each type of attachment into it's own controller that extends `AttachmentController`. As a result of this splitting, we can refactor the views to be specific to each `Attachment` type and in the process remove partial branching based on attachment type.

#### Use `BulkUpload` for all file attachments

As `BulkUpload` allowed the user to upload single files, there was no need to have both the single file upload and bulk uploader. Therefore this PR removes the single file attachment upload and replaces it with `BulkUpload` across Whitehall. The changes required for this included:

- `BulkUploadController` extends `AttachmentController`, this means that it now has the same validation, security and updating methods that all other attachments implement (the previous implementation was missing the call to `attachment_updater`)
- `BulkUploadController` will now display a message on successful upload (which it previously didn't do even though single file attachments did)
- `BulkUpload` now uses `attachable` and is not tied to `Edition`, as a result it now functions with all content that supports file attachments

### Visual Differences

| Location | Before | After |
| -------- | -------- | ------- |
| From editing an edition with attachments | Screenshot 2025-08-19 at 14 33 59  |  Screenshot 2025-08-19 at 14 33 34 |
| First screen to upload an attachment | Screenshot 2025-08-19 at 14 35 54  |   Screenshot 2025-08-19 at 14 36 51 |
| Screen to edit uploaded attachments |  Screenshot 2025-08-19 at 14 35 54 | Screenshot 2025-08-19 at 14 37 11 |

## Why

This continues the work previously done to make bulk uploading use the new multiple file input component. It addresses a previous comment about referencing the BulkUploadController in the view and ensures that the BulkUploadController implements the same functionality as other attachment controllers. 

### DIFF
--- a/app/controllers/admin/html_attachments_controller.rb
+++ b/app/controllers/admin/html_attachments_controller.rb
@@ -16,7 +16,6 @@ private
       :hoc_paper_number,
       :unnumbered_hoc_paper,
       :parliamentary_session,
-      :visual_editor,
       govspeak_content_attributes: %i[id body manually_numbered_headings],
     ).merge(attachable:)
   end
@@ -41,9 +40,6 @@ private
   def build_attachment
     HtmlAttachment.new(attachment_params).tap do |attachment|
       attachment.build_govspeak_content if attachment.govspeak_content.blank?
-      if attachment.visual_editor.nil?
-        attachment.visual_editor = Flipflop.govspeak_visual_editor? && current_user.can_see_visual_editor_private_beta?
-      end
     end
   end
 end
--- a/app/views/admin/html_attachments/_fields.html.erb
+++ b/app/views/admin/html_attachments/_fields.html.erb
@@ -28,25 +28,6 @@
   </div>
 
   <div class="govuk-!-margin-bottom-<%= margin_bottom %> app-view-attachments__form-body js-locale-switcher-field">
-    <% if Flipflop.govspeak_visual_editor? && current_user.can_see_visual_editor_private_beta? && form.object.visual_editor %>
-        <%= render "components/visual_editor", {
-          label: {
-            text: "Body (required)",
-            heading_size: "l",
-          },
-          name: "#{name}[govspeak_content_attributes][body]",
-          rows: 20,
-          id: "#{name}_govspeak_content_body",
-          value: form.object.govspeak_content.body,
-          error_items: errors_for(attachment.errors, :"govspeak_content.body"),
-          right_to_left:  form.object.rtl_locale?,
-          data_attributes: {
-            image_ids: edition_image_ids,
-            attachment_ids: [], # HTML attachments cannot embed Attachments from their parent Edition
-          },
-          hidden_field_name: "#{name}[visual_editor]",
-      } %>
-    <% else %>
       <%= render "components/govspeak_editor", {
         label: {
           heading_size:,
@@ -63,8 +44,6 @@
           attachment_ids: [], # HTML attachments cannot embed Attachments from their parent Edition
         },
       } %>
-      <%= hidden_field_tag "#{name}[visual_editor]", false %>
-    <% end %>
   </div>
 
   <% if !attachment.new_record? %>
--- a/test/functional/admin/html_attachments_controller_test.rb
+++ b/test/functional/admin/html_attachments_controller_test.rb
@@ -39,15 +39,6 @@ class Admin::HtmlAttachmentsControllerTest < ActionController::TestCase
     assert_not_nil(Attachment.find_by(title: attachment[:title]))
   end
 
-  test "POST :create saves an html attachment with a visual editor flag" do
-    attachment = valid_html_attachment_params.merge(title: SecureRandom.uuid, visual_editor: true)
-
-    post :create, params: { edition_id: @edition.id, attachment: }
-
-    attachment = Attachment.find_by(title: attachment[:title])
-    assert_equal true, attachment.visual_editor
-  end
-
   test "POST :create for an HtmlAttachment updates the publishing api" do
     attachment = valid_html_attachment_params
 
@@ -88,54 +79,10 @@ class Admin::HtmlAttachmentsControllerTest < ActionController::TestCase
     assert_select "option[value='#{Attachment.parliamentary_sessions.first}']"
   end
 
-  view_test "GET :new shows visual editor and no markdown help if permission and feature flag are enabled" do
-    feature_flags.switch!(:govspeak_visual_editor, true)
-    current_user.permissions << User::Permissions::VISUAL_EDITOR_PRIVATE_BETA
-
-    publication = create(:publication)
-    get :new, params: { edition_id: publication }
-
-    assert_select(".app-c-visual-editor__container")
-    assert_select ".govspeak-help", visible: false, count: 1
-  end
-
   test "POST :create with bad data does not save the attachment and re-renders the new template" do
     post :create, params: { edition_id: @edition, attachment: { attachment_data_attributes: {} } }
     assert_template :new
     assert_equal 0, @edition.reload.attachments.size
-  end
-
-  view_test "GET :edit shows visual editor and no markdown help if permission and feature flag are enabled, and attachment has been saved with visual editor" do
-    feature_flags.switch!(:govspeak_visual_editor, true)
-    current_user.permissions << User::Permissions::VISUAL_EDITOR_PRIVATE_BETA
-
-    attachment = create(:html_attachment, attachable: @edition, visual_editor: true)
-    get :edit, params: { edition_id: @edition, id: attachment }
-
-    assert_select(".app-c-visual-editor__container")
-    assert_select ".govspeak-help", visible: false, count: 1
-  end
-
-  view_test "edit form does not render visual editor, and renders the markdown help, for exited attachments" do
-    feature_flags.switch!(:govspeak_visual_editor, true)
-    current_user.permissions << User::Permissions::VISUAL_EDITOR_PRIVATE_BETA
-
-    attachment = create(:html_attachment, attachable: @edition, visual_editor: false)
-    get :edit, params: { edition_id: @edition, id: attachment }
-
-    assert_select ".app-c-visual-editor__container", count: 0
-    assert_select ".govspeak-help", count: 1
-  end
-
-  view_test "edit form does not render visual editor, and renders the markdown help, for pre-existing attachments" do
-    feature_flags.switch!(:govspeak_visual_editor, true)
-    current_user.permissions << User::Permissions::VISUAL_EDITOR_PRIVATE_BETA
-
-    attachment = create(:html_attachment, attachable: @edition, visual_editor: nil)
-    get :edit, params: { edition_id: @edition, id: attachment }
-
-    assert_select ".app-c-visual-editor__container", count: 0
-    assert_select ".govspeak-help", count: 1
   end
 
   test "PUT :update for HTML attachment updates the attachment" do


app/views/admin/attachments/_attachments.html.erb
```diff
@@ -31,8 +31,8 @@
         <% end %>
       </div>
       <div class="app-view-edition-resource__actions govuk-grid-column-full govuk-button-group">
-        <%= link_to("Edit attachment", [:edit, :admin, typecast_for_attachable_routing(attachable), attachment.becomes(Attachment)], class: "govuk-link govuk-link--no-visited-state", "aria-label": "Edit attachment: #{attachment.title}") %>
```
ChrisBAshton, line 34: Wow, I don't even know what `.becomes(Attachment)` does. Nice to kill this off!

config/routes.rb
```diff
@@ -298,20 +313,42 @@ def redirect(path, options = { prefix: Whitehall.router_prefix })
         resource :public_feedback, controller: "consultation_responses", type: "ConsultationPublicFeedback", except: %i[new destroy]
       end
 
-      resources :attachments, path: "consultation-responses/:consultation_response_id/attachments", as: "consultation_response_attachments" do
-        put :order, on: :collection
-        get :confirm_destroy, on: :member
-        get :reorder, on: :collection
+      resources :consultation_responses, only: [] do
```
ChrisBAshton: Should `only` be blank array?

app/views/admin/attachments/index.html.erb
```diff
@@ -24,13 +24,8 @@
     } %>
     <ul class="govuk-list">
       <li>
-        <%= link_to "Upload new file attachment", new_polymorphic_path([:admin, typecast_for_attachable_routing(attachable), FileAttachment]), class: "govuk-link govuk-link--no-visited-state" %>
+        <%= link_to "Upload new file attachments".html_safe, new_polymorphic_path([:admin, typecast_for_attachable_routing(attachable), BulkUpload]), class: "govuk-link govuk-link--no-visited-state" %>
```
ChrisBAshton: Is the `html_safe` needed here, when it wasn't before?

app/controllers/admin/bulk_uploads_controller.rb
```diff
@@ -30,6 +32,21 @@ def create
...
+    notices = @bulk_upload.attachments.map { |attachment| "Attachment '#{attachment.title}' #{attachment.attachment_data.to_replace_id.present? ? 'updated' : 'uploaded'}".html_safe }
+
+    notice = "".html_safe
+    notice << content_tag(:ul) do
+      inner_html = "".html_safe
+      notices.each { |notice| inner_html << content_tag(:li, notice, class: "gem-c-success-alert__message") }
+      inner_html
+    end
+
+    flash[:html_safe] = notice
```
ChrisBAshton: Does this do anything?

app/controllers/admin/bulk_uploads_controller.rb
```diff
@@ -30,6 +32,21 @@ def create
...
 
+  def notice
+    notices = @bulk_upload.attachments.map { |attachment| "Attachment '#{attachment.title}' #{attachment.attachment_data.to_replace_id.present? ? 'updated' : 'uploaded'}".html_safe }
+
+    notice = "".html_safe
+    notice << content_tag(:ul) do
+      inner_html = "".html_safe
+      notices.each { |notice| inner_html << content_tag(:li, notice, class: "gem-c-success-alert__message") }
+      inner_html
+    end
```
ChrisBAshton: This method can be refactored to just this, I think:

```ruby
def notice
  content_tag(:ul) do
    @bulk_upload.attachments.map do |attachment|
      content_tag(
        :li,
        "Attachment '#{attachment.title}' #{attachment.attachment_data.to_replace_id.present? ? 'updated' : 'uploaded'}".html_safe,
        class: "gem-c-success-alert__message",
      )
    end
end
```

app/controllers/admin/attachments_controller.rb
```diff
@@ -69,47 +62,16 @@ def attachment
...
-    FileAttachment.new(attachment_params).tap do |file_attachment|
-      file_attachment.build_attachment_data unless file_attachment.attachment_data
-      file_attachment.attachment_data.attachable = attachable
-    end
+    Attachment.new(attachment_params)
   end
 
   def attachment_params
-    attachment_params = params.fetch(:attachment, {}).permit(
+    params.fetch(:attachment, {}).permit(
```
ChrisBAshton, line 74: Can this method be simplified/removed now? E.g. `command_paper_number` is still explicitly permitted here even though we only really want to allow it in the FileAttachmentsController.

test/functional/admin/attachments_controller_test.rb
```diff
@@ -3,29 +3,6 @@
 class Admin::AttachmentsControllerTest < ActionController::TestCase
   should_be_an_admin_controller
 
-  def valid_file_attachment_params
```
ChrisBAshton, line 6: Full disclaimer - I've skipped review of the `Split AttachmentController tests` commit. I trust it's just the same code, moved around.

app/models/bulk_upload.rb
```diff
@@ -111,6 +110,8 @@ def find_and_update_existing_attachment(attachment_params)
 
     attachment.attachment_data = AttachmentData.new(attachment_data_attributes)
 
+    attachment.attachment_data.attachable = @attachable
```
ChrisBAshton, line 113: I find it a bit odd we have to set this here and on `attachment.attachable`. Would have thought `attachment.attachable` would defer to `attachment.attachment_data.attachable` so only need setting once?

config/routes.rb
ChrisBAshton, line 1: The routes.rb file has some added bloat now and a lot of repetition. It looks as though that could be DRY'd up using Rails route concerns?

```ruby
# Reusable snippets
concern :confirm_destroyable do
  get :confirm_destroy, on: :member
end

concern :bulk_uploadable do
  resources :bulk_uploads, except: %i[show edit update] do
    post :upload_files, on: :collection
    get :set_titles,   on: :member
  end
end

# Everything an "attachable" needs
concern :attachable do
  resources :attachments do
    put :order, on: :collection
    concerns :confirm_destroyable
  end

  resources :file_attachments,    except: %i[index show], concerns: :confirm_destroyable
  resources :html_attachments,    except: %i[index show], concerns: :confirm_destroyable
  resources :external_attachments,except: %i[index show], concerns: :confirm_destroyable

  concerns :bulk_uploadable
end

# Use the concern wherever those blocks appeared
resources :policy_groups, path: "groups", except: [:show], concerns: :attachable
# ...and any other parents that have these routes:
# resources :documents, concerns: :attachable
# resources :whatever_else, concerns: :attachable
```

My LLM also suggests looking into 'shallow routing':

> If your nesting is deep (e.g., /parents/:parent_id/.../attachments/:id), consider:

```ruby
resources :policy_groups, path: "groups",
                          except: [:show],
                          concerns: :attachable,
                          shallow: true
```

> shallow: true keeps creation routes nested (good for scoping) but flattens member routes to /attachments/:id, which is often nicer.

(It would be fine to do one or both of these in a tack-on commit at the end of the PR if you like)

ChrisBAshton (approved):
Really excellent work, Patrick ⭐ greatly simplifies Whitehall IMO.

Approving subject to manual testing on Integration, and comms going out on Basecamp.

app/controllers/admin/bulk_uploads_controller.rb
patrickpatrickpatrick: Yes, if it's not set then the html will be rendered as plain text because of the `FlashMessageComponent`. It could actually be set to `true` instead of the string and it would still work.

https://github.com/alphagov/whitehall/blob/24190ec81a45bda7b053cad4cda063a639f3e87f/app/components/admin/flash_message_component.rb#L7-L13

app/controllers/admin/attachments_controller.rb
patrickpatrickpatrick: Oh, are the reference fields actually only meant for files? From reading the code I didn't know that, since previously it would display them for any attachment type if the attachable allowed it.

https://github.com/alphagov/whitehall/blob/dbe3a3a6164ff0aedf23501f371e271f6de9eb67/app/views/admin/attachments/_form_fields.html.erb#L25-L27

app/models/bulk_upload.rb
patrickpatrickpatrick: I think this is required because it's an `attr_accessor` in `AttachmentData` so it's not actually accessed through the associated `attachment` and not saved. Not sure why this is the case (but did not change it).

https://github.com/alphagov/whitehall/blob/34402df0c4ad70f4c80e5730345bcbcb09148055/app/models/attachment_data.rb#L20

app/controllers/admin/attachments_controller.rb
patrickpatrickpatrick: Based on this, in `external_attachments_controller` we could have

```
  def attachment_params
    params.fetch(:attachment, {}).permit(
      :title,
      :locale,
      :external_url,
    ).merge(attachable:)
  end
```

and also in `html_attachments_controller` we could have

```
def attachments_params
    params.fetch(:attachment, {}).permit(
      :title,
      :locale,
      :visual_editor,
      govspeak_content_attributes: %i[id body manually_numbered_headings],
    ).merge(attachable:)
end
```

if I'm making the right assumptions here.

config/routes.rb
patrickpatrickpatrick: The way attachments works means I can't do this exactly, but I have refactored to use routes. Shallow routing is an interesting idea but I might do that in another PR since that would introduce some URL changes.

app/controllers/admin/attachments_controller.rb
ChrisBAshton: Ahhh my bad, sorry Patrick - command paper numbers are allowed on HTML attachments too:
https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/editions/572828/attachments/new?type=html

(and indeed on external ones) https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/editions/572828/attachments/new?type=external

Sorry I led you down the wrong path here.

app/controllers/admin/bulk_uploads_controller.rb
```diff
@@ -1,5 +1,8 @@
 class Admin::BulkUploadsController < Admin::AttachmentsController
   before_action :build_bulk_upload
+  include ActionView::Helpers::TagHelper
+  include ActionView::Helpers::TextHelper
+  attr_accessor :output_buffer
```
ChrisBAshton, line 5: What does this do?

patrickpatrickpatrick: These are for the `notice` method, to get `concat` [working](https://stackoverflow.com/a/7562194). To be honest, I'm not totally sure where `notice` should live. I suppose it's really a view helper but then not sure where it could be called for the `flash` to pick up the content.

PR #10522, opened 2025-08-08
(CM-133) Remove the ability to add an embedded object from the view page

pezholio:
When we were dealing with Pension objects, we added the ability to add another subschema (Rate) on an existing Pension from the view page. This would publish the Pension immediately without having to preview or add change notes. We decided this was OK because if a new rate was added it would, by definition, not be used anywhere, so would never trigger an update.

However, now we're dealing with Contacts, which have a composed block, it's possible that adding a new Contact Method would trigger an update on a page, so we should ensure the user has the ability to review those changes and add change notes as neccessary before publishing.

The easiest way to do this is to remove the ability to add embedded objects from the view page. This has the advantage of allowing us to delete a ton of code, as well as making the journey more streamlined. 

The only slight drawback is if a user adds (say) a new pension rate to an already in-use pension, they will be prompted to preview the changes on the page, which won't show any differences. We decided we would live with this for now, and think about how to simplify this when we come back to making improvements to the preview step.


edavey (approved):
Nice to simplify / remove code!

PR #10523, opened 2025-08-11
Remove force publish anything permission and refactor edition auth rules [WHIT-2377]

ryanb-gds:
Happy to separate these into independent PRs if reviewers would prefer.

## What

The "force publish anything" permission is no longer present in SignOn, so we can remove all reference to it from Whitehall 🎉 

Likewise, we never actually check the permissions for the Scheduled Publishing Robot, as we know exactly who it is, so we can get rid of those rules too.

This PR also refactors the edition authorisation rules quite substantially. The following changes have been made:

1. Raise an error if an unknown action is submitted for authorisation. We want to let our fellow developers know if they have misused an action instead of silently returning false.
2. Deal with historic editions using a separate private method, as a historic edition has completely different authorisation rules from a normal edition. The normal edition rules were complex enough without bringing historic editions into the mix.
3. Group authorisation logic by action rather than by role. This brings the instance authorisation handling in line with the class authorisation handling, and makes it much easier to see at a glance how an action is handled.
4. Inline a number of low value private methods. This makes it easier to reason about an authorisation decision without having to navigate several layers of indirection.

I believe the above changes make it significantly easier to reason about Whitehall's authorisation decision flow.

## Why

We want to simplify the edition authorisation layer as much as we can before introducing additional complexity in [WHIT-2377](https://gov-uk.atlassian.net/browse/WHIT-2377)

[WHIT-2377]: https://gov-uk.atlassian.net/browse/WHIT-2377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ChrisBAshton (approved):
These are great improvements ⭐

PR #10530, opened 2025-08-12
(CM-293) Set default values globally

pezholio:
This simplifies the way we handle default values, so the way they are handled when set in the schema is more predictable.


edavey (approved):
Thanks for this improvement.

PR #10532, opened 2025-08-12
CM-286: Provide preview on govpress enabled textarea component

edavey:
See Jira [CM-286](https://gov-uk.atlassian.net/browse/CM-286)

## Use the govspeak-editor to provide HTML previews on govspeak enabled textareas

In the case of fields which are `govspeak-enabled` we use
the existing `component/govspeak-editor` which has all the
functionality we require for toggling between Govspeak
markup and a 'Preview' rendered into HTML.

In the case of textareas which aren't govspeak-enabled, we
use our regular textarea component:

`govuk_publishing_components/components/textarea`

toggle_preview_govspeak

[CM-286]: https://gov-uk.atlassian.net/browse/CM-286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.html.erb
+++ b/lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.html.erb
@@ -1,10 +1,21 @@
 <div class="app-c-content-block-manager-govspeak-enabled-textarea-component">
-  <%= render("govuk_publishing_components/components/textarea", {
-              label: { text: label, hint_text: hint_text, hint_id: aria_described_by },
-              name: name_attribute,
-              id: id_attribute,
-              value:,
-              error_items: error_items,
-              describedby: aria_described_by,
-  }) %>
+  <% if govspeak_enabled? %>
+    <%= render "components/govspeak_editor", {
+                name: name_attribute,
+                id: id_attribute,
+                label: { text: label, hint_text: hint_text, hint_id: aria_described_by },
+                value:,
+                error_items: error_items,
+                hint_id: aria_described_by,
+                rows: 5,
+    } %>
+  <% else %>
+    <%= render "govuk_publishing_components/components/textarea", {
+                label: { text: label },
+                name: name_attribute,
+                id: id_attribute,
+                value:,
+                error_items: error_items,
+    } %>
+  <% end %>
 </div>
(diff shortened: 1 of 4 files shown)


lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.html.erb
```diff
@@ -1,10 +1,21 @@
...
-  <%= render("govuk_publishing_components/components/textarea", {
-              label: { text: label, hint_text: hint_text, hint_id: aria_described_by },
-              name: name_attribute,
-              id: id_attribute,
-              value:,
-              error_items: error_items,
-              describedby: aria_described_by,
-  }) %>
+  <% if govspeak_enabled? %>
+    <%= render "components/govspeak_editor", {
```
pezholio, line 3: Oh wow, I didn't realise it'd be this easy! Worth pointing out that this will be another thing we'll need to move if/when we replatform

lib/engines/content_block_manager/features/step_definitions/video_relay_service_steps.rb
```diff
@@ -26,3 +27,13 @@
 def label_for(field_name)
   I18n.t("content_block_edition.details.labels.telephones.video_relay_service.#{field_name}")
 end
+
+def should_be_able_to_preview_the_govspeak_enabled_field
```
pezholio, line 31: Yeah, I had a look myself, and couldn't get anything to load either 😢

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.html.erb
pezholio: Might be good to get this into the components gem as something portable

pezholio (approved):
LGTM 👍

lib/engines/content_block_manager/app/components/content_block_manager/content_block_edition/details/fields/govspeak_enabled_textarea_component.html.erb
```diff
@@ -1,10 +1,21 @@
 <div class="app-c-content-block-manager-govspeak-enabled-textarea-component">
```
pezholio, line 1: Also I wonder if we can make this just the default textarea component now?

edavey: yes, I'll make a ticket for this as the current `GovspeakEnabledTextareaComponent` works only for nested objects, so there's a little work to do to make it general purpose.

PR #10533, opened 2025-08-12
Add a way to restrict usage of a flexible page type to users of specific organisations [WHIT-2377]

ryanb-gds:
We use the content ID to identify the organisation, on the basis that Whitehall organisation IDs are not well known, and the organisation slug is more likely to change than the content ID.

The organisations setting is optional. If the setting is omitted from the type config, it is assumed that all users can use the page type. This seemed more intuitive than making the settings required and using an empty array to indicate that the page type is not restricted by organisation.

### DIFF
--- a/docs/flexible_pages.md
+++ b/docs/flexible_pages.md
@@ -10,17 +10,18 @@ The JSON for each type has these top level keys:
 
 - 'key': The unique identifier for the flexible page type. This is what will be stored in the edition's `flexible_page_type` column.
 - 'schema': The schema for the flexible page type, defined as [JSON schema](https://json-schema.org/docs). Each schema must have a root schema of the type "object".
-- 'settings': The settings for the flexible page type. All settings are required.
+- 'settings': The settings for the flexible page type.
 
-These are the settings available for flexible page types:
+These are the settings available for flexible page types. All settings are required.
 
-| Key                          | Description                                                                                                                                            |
-|------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------|
-| base_path_prefix             | The prefix for the base path at which the flexible page will be published. E.g. /government/history for the page /government/history/10-downing-street | 
-| publishing_api_schema_name   | The Publishing API schema name for flexible pages of this type                                                                                         |
-| publishing_api_document_type | The Publishing API document type for flexible pages of this type                                                                                       |
-| rendering_app                | The rendering app for the flexible page type                                                                                                           |
-| images_enabled               | Whether or not users should be able to upload images for this flexible page type using the images tab on the edition form                              |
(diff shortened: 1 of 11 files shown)


docs/flexible_pages.md
```diff
@@ -10,17 +10,18 @@ The JSON for each type has these top level keys:
...
+| base_path_prefix         | The prefix for the base path at which the flexible page will be published. E.g. /government/history for the page /government/history/10-downing-street | yes      |
+| publishing_api_schema_name | The Publishing API schema name for flexible pages of this type                                                                                         | yes      |
+| publishing_api_document_type | The Publishing API document type for flexible pages of this type                                                                                       | yes      |
+| rendering_app            | The rendering app for the flexible page type                                                                                                           | yes      |
+| images_enabled           | Whether or not users should be able to upload images for this flexible page type using the images tab on the edition form                              | no       |
+| organisations            | An array of organisation content IDs. Only users from one of the listed organisations will be able to use the flexible page type                       | yes      |
```
ChrisBAshton: This is interesting! I had every intention of putting together the ADR last week, and one of my vague thoughts I was going to get onto paper was "configuration over convention". Every property should be explicit in every schema, so that we can quickly and easily answer questions like "how many document types support translations" etc, and also avoid getting into config drift where some doctypes are arbitrarily different to others just because one thing didn't get copied over.

In that spirit, I'd be tempted to suggest a property name like `restricted_to_organisations`, which is a bit more self documenting, and where a `nil` or empty array value can be inferred as 'unrestricted'.

ryanb-gds: I also considered this quite hard. I quite like the idea of using `null` as the value for "all orgs", so I can go with that if you're keen. On the naming side of things I suspect we're going to want  to potentially define other authorisation controls at some point (e.g. for a type to use lead organisations, or to restrict access to particular users), at which point we'll likely want a top-level authorisation key. So I decided to put a decision for a "proper" naming for this off until a later date and see how our needs develop. I also thought about consistency with the specialist publisher schemas, which also just use organisations - might be a chance this makes a migration easier (though probably not). If you're quite liking `restricted_to_organisations` though I'm happy to roll with that - what do you think?

ChrisBAshton: Nice that we're agreed on `null` for 'no restrictions' 👍 

Hmmm yeah I see what you mean. We may end up with a `restrictions` hash, or a `policy` hash, or a `access_limited` hash, all with pros and cons, that would give us the flexibility to define all sorts of authorisation rules. But we aren't there yet.

I don't think I mind too much about the naming just yet, my only strong feeling is that every schema should define a value, even if it's just `null`. So `organisations` or `restricted_to_organisations` both seem ok.

ChrisBAshton (approved):
Looks great!

PR #10540, opened 2025-08-13
Changes to history page definition [WHIT-2268]

ryanb-gds:
1. Remove the page_title attribute from the schema. We are going to use the default edition title for now. We have therefore changed the label for that field to "Title" from "Admin Title" in the UI, as that will now be the title used on GOV.UK.
2. Change the rendering app for history pages to "frontend"

### DIFF
--- a/app/models/flexible_page_content_blocks/image_select.rb
+++ b/app/models/flexible_page_content_blocks/image_select.rb
@@ -21,9 +21,8 @@ module FlexiblePageContentBlocks
 
       if (selected_image = Context.page.images.find { |image| image.id == content.to_i })
         {
-          src: selected_image.url,
+          url: selected_image.url,
           caption: selected_image.caption,
-          alt_text: selected_image.alt_text,
         }
       else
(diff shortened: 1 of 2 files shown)


app/models/flexible_page_content_blocks/image_select.rb
```diff
@@ -21,9 +21,8 @@ def publishing_api_payload(content)
 
       if (selected_image = Context.page.images.find { |image| image.id == content.to_i })
         {
-          src: selected_image.url,
+          url: selected_image.url,
```
lauraghiorghisor-tw, line 24: So the alt_text cannot be null, but it can be missing? (I would have thought it missing from the payload is technically the same as it being null.) Or do we want to provide a default value of sorts for now?

lauraghiorghisor-tw (approved):
LGTM

app/models/flexible_page_content_blocks/image_select.rb
ryanb-gds: Discussed offline, this comment referred to the `alt_text`, which cannot be null but is not required by the schema. So missing from the payload is correct in this case

PR #10546, opened 2025-08-14
Allow embedding of inline images in Govspeak blocks

ryanb-gds:
History pages need to support images embedded in the body of the content. To make this possible, we need to pass the images to the govspeak renderer so that any inline image shortcodes can be converted to `` tags. Luckily we have the page available in the flexible block context.


lauraghiorghisor-tw (approved):
LGTM

PR #10547, opened 2025-08-14
Content Modelling: ADR to record decision to move Content Block Manager to own application

edavey:


edavey:
Hi @richardTowers 
Stu has asked me to make some progress with this "re-platforming" whilst he's away. I thought I'd start by organising my understanding in an ADR before having a go at setting up the new application.

Do you have the capacity to review these notes and then have a chat with me?

docs/content_block_manager/adr/0003-move-out-to-new-application.md
```diff
@@ -0,0 +1,171 @@
...
+
+- view a "hot-reloading" preview of a complex content block whilst editing a
+nested object. For example, whilst editing a Govspeak-enabled textarea within
+the "telephone 2" nested object, the user will be able to generate a preview of
+the composite Contact content-block including the unsaved contents of that
+Govspeak textarea.
+
+To offer these features we plan to use Rails' "Hotwire" components, such as
+[Turbo Frames][]. We believe that it would not be practical to introduce this
+library to Whitehall.
```
pezholio: Might be worth adding some context to this - something like:

```
We believe that it would not be practical to introduce this library to Whitehall, as introducing Turbo as a dependency to Whitehall will mean that all pages within Whitehall will be loaded via Turbo, which may be confusing to users, and also have unintended consequences elsewhere in the app. 
```

docs/content_block_manager/adr/0003-move-out-to-new-application.md
```diff
@@ -0,0 +1,171 @@
...
+that behaviour is not inadvertently changed, this would be a secondary step,
+after initially porting the Minitest test suite.
+
+### Lift and shift code
+
+We will "lift and shift" our source code, deferring any refactoring.
+
+### Move from MySQL to PostgreSQL
+
+We will use Postgres rather than MySQL in the new Content Block Manager service.
```
pezholio, line 119: Would be good to say why we're doing that - we could reference [this part of the Conventions for Rails apps page](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#data-storage) to state that it's the default for new apps now.

docs/content_block_manager/adr/0003-move-out-to-new-application.md
edavey: thks, done

docs/content_block_manager/adr/0003-move-out-to-new-application.md
edavey: good idea, thks, done

edavey:
@pezholio I think we ought to merge this now?

richardTowers (commented):
GitHub won't let me formally approve this as it's already merged, but retrospectively here's a ✅  from me!

PR #10548, opened 2025-08-14
Refactor away flexible blocks context

ryanb-gds:
The flexible block objects previously used a "Context" object to make the flexible page model available to the blocks so that they can do things such as access the value of other associations that belong to the page. The Context object was set up in a controller "before" action for each request.

This was a poor design, because the context was essentially a big hairy global object and resulted in blocks having access to far more information than they needed to do their jobs.

We therefore have refactored the app to give more responsibility to the block factory, which receives the flexible page model instance and is now responsible for extracting any information required from the model instance and passing it to each block in the build method. This allows us to hide the page model from the blocks.

We also take advantage of Rails magical partial rendering so that we can access the block from inside its erb template. We add attribute readers to blocks so that information passed to the block initializers can be used in the erb template.


test/unit/app/models/flexible_page_content_blocks/factory_test.rb
```diff
@@ -1,16 +1,52 @@
...
+  end
+
+  test "it can build a govspeak string block" do
+    page = FlexiblePage.new
+    page.images = [build(:image)]
+    factory = FlexiblePageContentBlocks::Factory.new(page)
+    block = mock("FlexiblePageContentBlocks::Govspeak")
+    FlexiblePageContentBlocks::Govspeak.expects(:new).with(page.images).returns(block)
+    assert_equal block, factory.build("string", "govspeak")
+  end
```
ryanb-gds, line 51: I'm not entirely sure if this is the right way to test this, it's more mocking than I really feel comfortable with. I could create a real instance of each block, but unless they publicly expose a method that depends on the arguments passed to their initialiser then I can't really test that the factory has done its job. Alternatively, we could test every block in a feature, but as the number of blocks grows we could end up with an enormous and complex feature. Thoughts? Better ideas?

patrickpatrickpatrick: Thinking about this some more, I don't think this is actually that bad. We're only testing the creation of the block type and the factory isn't responsible for what the block does. So as long as the block type is also tested, this is fine right?

ryanb-gds: I think so, it just feels like I'm testing the mock as much as anything. Lets go with it for now if you're happy enough with it, can always remove or change later if we decide we don't like it

PR #10569, opened 2025-08-19
Fixed deprecation warnings, fixed Ruby warnings, fixed missing assertions, removed test logging output [WHIT-1861]

ryanb-gds:
The Whitehall tests have gradually become extremely cluttered with warnings and log messages which are untidy and distracting.

Many of the deprecation warnings related to Rack's renaming of the unprocessable entity response to unprocessable content. This commit fixes all but one of these which is emitted from inside Rails and will be fixed in a later version of Rails.

In some cases, the missing assertions were for very low value tests that were essentially testing active record. These tests have been deleted.

Other missing assertions were seeming caused by Minitest not recognising Capybara DSL usage. I have generally been able to use non Capybara assertions instead, and we might consider removing this from the test base altogether in future to improve test style consistency.

I have been unable to work out how to prevent the Prometheus exporter from being active in the test environment, so we still get a message from that.


test/unit/app/models/organisation_test.rb
```diff
@@ -1042,24 +1031,6 @@ class OrganisationTest < ActiveSupport::TestCase
     )
   end
 
-  test "#save triggers organisation with a changed default news organisation image to republish news articles" do
```
ryanb-gds, line 1045: This test never did anything, because the documents list was empty. I tried to fix the test but couldn't get it to pass, so I removed it on the basis that we haven't been testing anything anyway for presumably quite a while.

ChrisBAshton (approved):
LGTM! ⭐ 

Are we also planning to raise a PR to add a lint rule that prevents us from adding tests that don't make assertions in future? (An AC I added to the [ticket](https://gov-uk.atlassian.net/browse/WHIT-1861) yesterday, so you may not have seen it!)

ryanb-gds:
> Are we also planning to raise a PR to add a lint rule that prevents us from adding tests that don't make assertions in future? (An AC I added to the [ticket](https://gov-uk.atlassian.net/browse/WHIT-1861) yesterday, so you may not have seen it!)

I can't see an obvious way to do this. I could monkey patch [this](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/testing/tests_without_assertions.rb) to raise instead of warn, perhaps. What do you think?

ChrisBAshton:
> > Are we also planning to raise a PR to add a lint rule that prevents us from adding tests that don't make assertions in future? (An AC I added to the [ticket](https://gov-uk.atlassian.net/browse/WHIT-1861) yesterday, so you may not have seen it!)
> 
> I can't see an obvious way to do this. I could monkey patch [this](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/testing/tests_without_assertions.rb) to raise instead of warn, perhaps. What do you think?

Sounds reasonable. ChatGPT came up with this:

```ruby
# test/test_helper.rb
require "active_support/testing/tests_without_assertions"

if ENV["CI"] # only fail on CI
  module FailOnNoAssertions
    def after_teardown
      if assertions.zero? && !skipped? && !error?
        file, line = method(name).source_location
        flunk "Test is missing assertions: `#{name}` #{File.expand_path(file)}:#{line}"
      end
      super
    end
  end

  ActiveSupport::Testing::TestsWithoutAssertions.prepend(FailOnNoAssertions)
end
```

ryanb-gds:
> > > Are we also planning to raise a PR to add a lint rule that prevents us from adding tests that don't make assertions in future? (An AC I added to the [ticket](https://gov-uk.atlassian.net/browse/WHIT-1861) yesterday, so you may not have seen it!)
> > 
> > 
> > I can't see an obvious way to do this. I could monkey patch [this](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/testing/tests_without_assertions.rb) to raise instead of warn, perhaps. What do you think?
> 
> Sounds reasonable. ChatGPT came up with this:
> 
> ```ruby
> # test/test_helper.rb
> require "active_support/testing/tests_without_assertions"
> 
> if ENV["CI"] # only fail on CI
>   module FailOnNoAssertions
>     def after_teardown
>       if assertions.zero? && !skipped? && !error?
>         file, line = method(name).source_location
>         flunk "Test is missing assertions: `#{name}` #{File.expand_path(file)}:#{line}"
>       end
>       super
>     end
>   end
> 
>   ActiveSupport::Testing::TestsWithoutAssertions.prepend(FailOnNoAssertions)
> end
> ```

We're actually reopening the `ActiveSupport::TestCase` module in our test helper already, so I have chosen to just invoke the `after_teardown` hook there rather than monkey patching `ActiveSupport::Testing::TestsWithoutAssertions`. The effect is the same, but I think there's marginally less risk of the patch causing problems if Rails' underlying definitions change.

ryanb-gds:
On further reflection we can't add the failure for missing assertions because content block manager tests often do not have assertions. Once they move out to their own repo we can add it.

PR #10573, opened 2025-08-20
fix: Filter markdown links from collected headers

KludgeKML:
- The page https://www.gov.uk/guidance/employment-tribunal-offices-and-venues (before and after links at the bottom), has h2 elements in govspeak that are links. Government Frontend, picking the text out of HTML on the fly to create the contents list, handled this somehow. When we moved it to publish time (picking the links out of govspeak), we introduced a bug where the markdown would be used as the link title in the content list.
- This pattern (of H2s as links) is itself not a good idea and may cause accessibility issues.  We should try to dissuade people from using it. That said, we should at least handle it gracefully. This commit allows us to filter out the markdown link and extract only the link text as our header item.

- before headers collected in Whitehall: https://webarchive.nationalarchives.gov.uk/ukgwa/20250716183242/https://www.gov.uk/guidance/employment-tribunal-offices-and-venues
- after headers collected in Whitehall: https://webarchive.nationalarchives.gov.uk/ukgwa/20250813183618/https://www.gov.uk/guidance/employment-tribunal-offices-and-venues

### DIFF
--- a/app/helpers/presenters/publishing_api/payload_headings_helper.rb
+++ b/app/helpers/presenters/publishing_api/payload_headings_helper.rb
@@ -6,6 +6,7 @@ module Presenters
 
         body_headings = Govspeak::Document.new(govspeak).structured_headers
         headers = remove_empty_headers(body_headings.map(&:to_h))
+        headers = remove_links_in_headers(headers)
 
         headers.empty? ? {} : { headers: headers }
       end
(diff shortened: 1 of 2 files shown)


KludgeKML:
This really needs to be done in the Govspeak gem, so closing this in favour of: https://github.com/alphagov/govspeak/pull/439

PR #10574, opened 2025-08-20
Move updating `main_office` into `Edition#edit` form [WHIT-1956]

patrickpatrickpatrick:
## What

- Move `main_office_id` into `_form` of `WorldwideOrganisations`
- Remove `WorldwideOrganisationsMainOfficesController`

### Visual Differences

Within `government/admin/worldwide-organisations/ID/edit`:

Screenshot 2025-08-20 at 16 27 07

`government/admin/worldwide-organisations/ID/choose_main_office` has been deleted

## Why

To prevent an error in which editing the main office of the draft of an already published Worldwide Organisation would cause an error due to a missing change note.

Main office is now editable within the main edition form of Worldwide Organisation. This means we can remove the controller which previously handled this edit and the associated views and tests.

The link to `Set main office` now links the user to the field within the edition edit which controls this.


app/views/admin/worldwide_organisations/_form.html.erb
```diff
@@ -36,6 +36,21 @@
     id: "associations",
   } do %>
     <div class="govuk-!-margin-bottom-4">
+      <% if edition.offices.many? %>
```
lauraghiorghisor-tw, line 39: Would it make sense to show it all the time, blank on new form? I think it would suggest to the user that this is where this information will show once selected. Perhaps we can add a hint if needed.

app/views/admin/worldwide_offices/index.html.erb
```diff
@@ -30,7 +30,7 @@
 
     <% if @worldwide_organisation.offices.many? %>
       <p class="govuk-body govuk-!-margin-bottom-6">
-        <%= link_to "Set main office", choose_main_office_admin_worldwide_organisation_path(@worldwide_organisation), class: "govuk-link" %>
+        <%= link_to "Set main office", edit_admin_edition_path(@worldwide_organisation) + "#edition_main_office_id-label", class: "govuk-link" %>
```
lauraghiorghisor-tw, line 33: Yeah I don't know how I feel about this. Would an information callout or something (with the link in it?) work better?

lauraghiorghisor-tw (commented):
Happy with the changes. Nonetheless...

I do think the flow as we had it before (well, if it worked) was preferable. I think flipping between the Offices page to the select is a bit intense for the user. Maybe showing the select at all times and providing a bit more context will help. 

My main point against this is that, since we have an offices tab, I would expect that we can do all the work related to offices there, so jumping back to the org form seems weird to me. 

---- 
I did revisit the previous flow (implementation you suggested, the one liner change). 

I understand your concern that if we make the change minor, a previous selection of major+change note would be overridden. 

What I assumed would happen when reviewing earlier, and it does, is that the user will have to re-save the org to get to publish so they will again see the screen with the update type and change note selection. The change note is persisted in the DB, so setting it back to major would bring back the change note. I am not sure if pub-api complains if we send a change note with a minor update type?

I guess it's easier to miss the selection, if the user thinks they've done it before, so that's a fair risk. 

We can alternatively:
-  save the org with validation false or better still
-  call `valid?` and drop the errors if the only validation is about change note, and then save without validating. Something like that. 

I would try this last one out, as I like the previous workflow better (i.e. keeping things in the offices controller). But if you consider the flow in this PR to be better, then I am happy to approve. Do consider the other in-line comments.

lauraghiorghisor-tw (approved):

PR #10576, opened 2025-08-21
[WHIT-2270] Fix images for configurable documents

lauraghiorghisor-tw:
Images were not included in the govspeak correctly. 

Also added the preview function to the govspeak block.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2270)

### DIFF
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -22,6 +22,13 @@ module GovspeakHelper
     wrapped_in_govspeak_div(bare_govspeak_to_html(body, [], attachments))
   end
 
+  def govspeak_to_html_with_images_and_attachments(govspeak, images = [], attachments = [], alternative_format_contact_email = nil)
+    mapped_images = prepare_images(images)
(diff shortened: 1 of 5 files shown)


app/helpers/govspeak_helper.rb
```diff
@@ -22,6 +22,13 @@ def govspeak_with_attachments_to_html(body, attachments = [], alternative_format
     wrapped_in_govspeak_div(bare_govspeak_to_html(body, [], attachments))
   end
 
+  def govspeak_to_html_with_images_and_attachments(govspeak, images = [], attachments = [], alternative_format_contact_email = nil)
```
ChrisBAshton, line 25: Would be good to know why you've defined this as a separate method rather than building the functionality into `govspeak_to_html` by default?

lauraghiorghisor-tw: It felt like that was specifically designed to not have images and attachments. Thought I'd make a new thing and be explicit about it. We'll likely touch on this again in the govspeak work.

PR #10588, opened 2025-08-22
[WHIT-2257] Display clearer embedded-contact errors on Editions & HTML attachments; relax validation on draft creation

GDSNewt:
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-1517)

### **Why**

Editors were not only confronted with vague error messages (e.g. “attachments is invalid”) but also completely blocked by validation when trying to create new editions or publish content whose HTML attachments had silently become invalid over time. This could happen when:

- Validation rules were introduced or tightened after initial publication, or
- Contacts referenced in Govspeak were deleted/changed.

As a result, users often discovered issues only at publish time—despite making no changes to the attachment—leading to both confusing feedback and an inability to progress with their workflow in Whitehall.

### **What**

We’ve introduced a context-aware validation approach for embedded contact IDs in Govspeak that:

- Surfaces clear, specific errors in the UI for both Editions and HTML attachments (e.g. helpful “missing contact” messages now appear on the edition summary page).
- Improves labelling: adds a more human-friendly label for the HTML attachments edition attribute.
- Improves validator copy: the Govspeak contact-embedding validator now explains what’s wrong and how to fix it.
- Unblocks draft creation: skips certain validations when creating a new draft, allowing editors to get past legacy/externally-introduced invalid states so they can fix issues in the UI.

### **Technical notes**

- Skip validation during new draft creation: Previously it wasn’t possible to create a new draft if an HTML attachment was invalid. We now skip validation in that specific flow, which is a safe choice that lets editors repair content without being blocked.
- Disable association validation in attachable trait: We explicitly turned off association validation for attachments and html_attachments when using custom validation contexts (e.g. :publish). Rails will otherwise validate loaded associations under the custom context—an unexpected behaviour that produced generic, unhelpful errors like “attachments is invalid.” See: https://github.com/rails/rails/issues/45209

Test updates: Fixed tests that were passing erroneously due to the previous behaviour; added/adjusted specs to assert clearer error messaging and the relaxed validation during draft creation.

### **User impact**

- Editors see actionable, field-level errors when embedded contacts are invalid.
- Creating a new draft no longer fails due to historical or external changes—editors can create, then fix.
- Overall: smoother, more predictable publishing flow with better guidance.

**BEFORE**

Screenshot 2025-08-15 at 09 29 01

Screenshot 2025-08-22 at 10 46 10

**AFTER**

[
Screenshot 2025-08-22 at 09 53 17
Screenshot 2025-08-22 at 09 53 39
](url)

### DIFF
--- a/app/models/concerns/attachable.rb
+++ b/app/models/concerns/attachable.rb
@@ -15,11 +15,13 @@ module Attachable
     has_many :attachments,
              -> { not_deleted.order("attachments.ordering, attachments.id") },
              as: :attachable,
-             inverse_of: :attachable
+             inverse_of: :attachable,
+             validate: false
 
     has_many :html_attachments,
              -> { not_deleted.order("attachments.ordering, attachments.id") },
-             as: :attachable
+             as: :attachable,
+             validate: false
 
     has_many :deleted_html_attachments,
              -> { deleted },
@@ -37,7 +39,7 @@ module Attachable
           @edition.attachments.each do |attachment|
             draft_attachment = attachment.deep_clone
             edition.attachments << draft_attachment
-            raise "Your edition failed to create successfully. Please contact GOV.UK support." unless draft_attachment.save!
+            raise "Your edition failed to create successfully. Please contact GOV.UK support." unless draft_attachment.save!(validate: false)
           end
         end
       end
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -19,6 +19,8 @@ en:
               invalid_date: First published date is not in the correct format
               blank: Enter a first published date
               inclusion: First published date must be between 1/1/1900 and the present
+            html_attachments:
+              missing_contact_reference: "- \"%{html_attachment_title}\" - embedded Contact (ID %{contact_id}) doesn't exist"
         nation_inapplicability:
           attributes:
             alternative_url:
--- a/test/unit/app/validators/govspeak_contact_embed_validator_test.rb
+++ b/test/unit/app/validators/govspeak_contact_embed_validator_test.rb
@@ -44,16 +44,25 @@ class GovspeakContactEmbedValidatorTest < ActiveSupport::TestCase
     assert_equal 0, edition.html_attachments.first.errors.size
   end
 
-  test "should be invalid if the HTML attachment contains a Contact that doesn't exist" do
+  test "should be invalid if the HTML attachment contains a Contact that has been deleted" do
     contact = create(:contact)
-    bad_id = "9999999999999"
-    edition = create(:case_study, body: "[Contact:#{contact.id}]")
-
-    edition.html_attachments = [create(:html_attachment, body: "[Contact:#{bad_id}]")]
+    edition = create(:case_study, html_attachments: [create(:html_attachment, body: "[Contact:#{contact.id}]")])
+    contact.destroy!
 
     GovspeakContactEmbedValidator.new.validate(edition)
 
-    assert_equal 0, edition.errors.size
-    assert_equal 1, edition.html_attachments.first.errors.size
+    assert_equal 1, edition.errors.where(:html_attachments).size
+  end
+
+  test "should add a specific error message if the HTML attachment contains a Contact that doesn't exist" do
+    bad_id = "9999999999999"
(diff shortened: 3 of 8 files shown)


config/locales/en/errors.yml
```diff
@@ -19,6 +19,8 @@ en:
               invalid_date: First published date is not in the correct format
               blank: Enter a first published date
               inclusion: First published date must be between 1/1/1900 and the present
+            html_attachments:
+              missing_contact_reference: "- Contact with ID %{contact_id} embedded in \"%{html_attachment_title}\" doesn't exist"
```
ChrisBAshton: Looking at the screenshot, I'd suggest this should be:

```suggestion
              missing_contact_reference: "- \"%{html_attachment_title}\" - embedded Contact (ID %{contact_id}) doesn't exist"
```

Which would render as:

> HTML attachments - "Test HTML attachment" - embedded Contact (ID 123) doesn't exist

app/models/concerns/attachable.rb
ChrisBAshton, line 1: Please reword the commit so that the first line of the commit doesn't spill over into the body in GitHub (if that makes sense)

ChrisBAshton (approved):
This bit was key:

> Disable association validation in attachable trait: We explicitly turned off association validation for attachments and html_attachments when using custom validation contexts (e.g. :publish). Rails will otherwise validate loaded associations under the custom context—an unexpected behaviour that produced generic, unhelpful errors like “attachments is invalid.” See: https://github.com/rails/rails/issues/45209

I didn't realise that - and it does suggest that dropping validation of associations altogether is the way to go 👍 

But then I don't understand how a HTML attachment validation issue is surfaced when trying to publish the parent edition. If we've disabled validation on associations, wouldn't they be skipped?

Is it because publishing a parent edition _triggers publication of its HTML attachments_? And if so, does the HTML attachment validation error prevent (or roll back) the publishing of the parent edition?

It looks as though you've tested it from the screenshots, so am happy to approve on the basis you and Ryan have considered the above already ✅

test/unit/app/validators/govspeak_contact_embed_validator_test.rb
```diff
@@ -44,16 +44,25 @@ class GovspeakContactEmbedValidatorTest < ActiveSupport::TestCase
...
+    bad_id = "9999999999999"
+    html_attachment = build(:html_attachment, title: "Test Doc", body: "[Contact:#{bad_id}]")
+    edition = build(:case_study, html_attachments: [html_attachment])
 
     GovspeakContactEmbedValidator.new.validate(edition)
 
-    assert_equal 0, edition.errors.size
-    assert_equal 1, edition.html_attachments.first.errors.size
+    error_message = edition.errors[:html_attachments].first
+    expected_message = "Contact with ID #{bad_id} embedded in \"Test Doc\" doesn't exist"
```
ryanb-gds: Probably best to load this using the internationalisation API so if we change the message we don't have to update the test. Minor nitpick though but something to consider in future

ryanb-gds (approved):
Thanks for seeing this through @GDSNewt, a big UX improvement 🎉

GDSNewt:
> This bit was key:
> 
> > Disable association validation in attachable trait: We explicitly turned off association validation for attachments and html_attachments when using custom validation contexts (e.g. :publish). Rails will otherwise validate loaded associations under the custom context—an unexpected behaviour that produced generic, unhelpful errors like “attachments is invalid.” See: [rails/rails#45209](https://github.com/rails/rails/issues/45209)
> 
> I didn't realise that - and it does suggest that dropping validation of associations altogether is the way to go 👍
> 
> But then I don't understand how a HTML attachment validation issue is surfaced when trying to publish the parent edition. If we've disabled validation on associations, wouldn't they be skipped?
> 
> Is it because publishing a parent edition _triggers publication of its HTML attachments_? And if so, does the HTML attachment validation error prevent (or roll back) the publishing of the parent edition?
> 
> It looks as though you've tested it from the screenshots, so am happy to approve on the basis you and Ryan have considered the above already ✅

I think it's because custom validators—such as GovspeakContactEmbedValidator—still explicitly check the contents of html_attachments during the publish workflow, but i'm not sure about that...

PR #10589, opened 2025-08-22
Remove GovSpeak post-processing (WHIT-2391)

ChrisBAshton:
---


app/helpers/govspeak_helper.rb
```diff
@@ -97,35 +93,28 @@ def govspeak_header_hierarchy(govspeak)
...
+  def bare_govspeak_to_html(govspeak = "", images = [], attachments = [], options = {})
+    processed_govspeak = preprocess_govspeak(govspeak, attachments, options)
+    html = markup_to_nokogiri_doc(processed_govspeak, images, attachments, locale: options[:locale])
       .to_html
 
     "<div class=\"govspeak\">#{html}</div>".html_safe
   end
 
+  def preprocess_govspeak(govspeak, attachments, options)
+    govspeak ||= ""
```
ryanb-gds, line 84: Slightly intrigued by this. Necessary for testing?

app/helpers/govspeak_helper.rb
```diff
@@ -97,35 +93,28 @@ def govspeak_header_hierarchy(govspeak)
...
+    govspeak = ContentBlockManager::FindAndReplaceEmbedCodesService.call(govspeak) if options[:preview]
+    govspeak = convert_attachment_syntax(govspeak, attachments)
+    govspeak = render_embedded_contacts(govspeak, options[:contact_heading_tag])
+    govspeak = replace_internal_admin_links(govspeak, options[:preview] == true)
+    govspeak = add_heading_numbers(govspeak) if options[:heading_numbering] == :auto
+    govspeak = add_manual_heading_numbers(govspeak) if options[:heading_numbering] == :manual
```
ryanb-gds: Would love to know when each of these options is used, if you know. Might be worth a comment?

test/unit/app/helpers/govspeak_helper_test.rb
```diff
@@ -1,6 +1,8 @@
 require "test_helper"
 
 class GovspeakHelperTest < ActionView::TestCase
+  extend Minitest::Spec::DSL
```
ryanb-gds, line 4: Do we need to add this here? I know we use it quite a bit in Whitehall but I don't like when we've got some minitest style tests and some rspec style tests in the same file. I'd actually quite like, if possible, to set an LLM agent to trying to consolidate all the Whitehall tests one way or the other at some point.

app/helpers/govspeak_helper.rb
```diff
@@ -36,7 +20,19 @@ def bare_govspeak_edition_to_html(edition)
...
-    bare_govspeak_to_html(edition.body, images, attachments)
+    bare_govspeak_to_html(edition.body, images, attachments, options)
+  end
+
+  def govspeak_with_attachments_to_html(body, attachments = [], alternative_format_contact_email = nil, options = {})
+    attachments = prepare_attachments(attachments, alternative_format_contact_email)
+    bare_govspeak_to_html(body, [], attachments, options)
+  end
+
+  def govspeak_to_html_with_images_and_attachments(govspeak, images = [], attachments = [], alternative_format_contact_email = nil, options = {})
```
ryanb-gds: How feasible would it be to make this the one method exposed by this helper and just expose all of the options/args as needed (and rename it just `govspeak_to_html`? I'd have thought we could do without `govspeak_with_attachments_to_html` and `bare_govspeak_to_html`, and `govspeak_edition_to_html` could just pass the body straight to the main method (or we could do that from the call sites)

ryanb-gds (approved):
No reason not to merge this as-is but a few thoughts

app/helpers/govspeak_helper.rb
ChrisBAshton: Yep. Doesn't feel great! I can live with it for now (especially as this is consolidated to one line, compared to similar lines that were there before).

test/unit/app/helpers/govspeak_helper_test.rb
ChrisBAshton: Agreed about the style mix. I was intending to change the existing tests in govpeak_helper_test.rb _first_, and then bring over the admin ones. I'll do that now 👍

app/helpers/govspeak_helper.rb
ChrisBAshton: I had the same thought 👍 and I'd like to make these keyword arguments too. Will give that a go.

app/helpers/govspeak_helper.rb
ChrisBAshton: Your comment prompted me to take a look! This option is only set on the [Govspeak rendering for HTML attachments](https://github.com/alphagov/whitehall/blob/4ff454bb5e6365b491eb64f86e917d24604a7f9e/app/helpers/govspeak_helper.rb#L43-L44). `auto` vs `manual` is set by this checkbox:

> Screenshot 2025-08-26 at 12 11 36

```
## Foo

...

### Bar

...

### Baz

...
```

If the box is unchecked ('auto'), the above renders as:

```
<h2><span class="number">1. </span>Foo</h2>
<h3><span class="number">1.1 </span>Bar</h3>
<h3><span class="number">1.2 </span> Baz</h3>
```

If the box is checked ('manual'), then... the user's input is retained.

Which made me take a step back and wonder what does the `add_manual_heading_numbers` method actually do?? It ensures that the number in the heading is wrapped in a `<span class="number">`, but that doesn't actually seem to do anything on the frontend (no screen reader only rules there either), so I assume it's either a legacy requirement or it was maybe needed to avoid including the number into the ID?

If you can't see a reason why it should exist (I can't!) then I'll delete this feature as part of this PR.

app/helpers/attachments_helper.rb
```diff
@@ -19,6 +19,14 @@ def preview_path_for_attachment(attachment)
...
 
+  def prepare_attachments(attachments, alternative_format_contact_email)
+    attachments
+      .select { |attachment| !attachment.file? || attachment.attachment_data&.all_asset_variants_uploaded? }
+      .map do |attachment|
+      attachment_component_params(attachment, alternative_format_contact_email:)
+    end
+  end
+
   def attachment_component_params(attachment, alternative_format_contact_email: nil)
```
TonyGDS, line 30: The actually seems to be more govspeak related than attachments related, would it make more sense to move this into the `govspeak_helper` rather than moving `prepare_attachments` into here?

app/helpers/govspeak_helper.rb
```diff
@@ -30,13 +30,6 @@ def govspeak_edition_to_html(edition, options = {})
     govspeak_to_html(edition.body, options.merge(images: images, attachments: attachments))
   end
 
-  def govspeak_to_html_with_images_and_attachments(govspeak, images = [], attachments = [], alternative_format_contact_email = nil, options = {})
```
TonyGDS: Is there a reason why we don't call `prepare_attachments` and `prepare_images` [here](https://github.com/alphagov/whitehall/pull/10589/commits/b9b2e6187d9004b319239ccb9509d428dbd035c5#diff-521ba147a44c2faeb0975bfeb8d1db1b805b1c3dd275d2e109faa3211a5599daL8)? It seems that other methods such as `govspeak_html_attachment_to_html` also need to call at least one of them before calling `govspeak_to_html`. This would also remove the need for having to call them outside of the `govspeak_helper` such as in `preview.html.erb`.

app/models/configurable_content_blocks/govspeak.rb
```diff
@@ -24,7 +25,7 @@ def json_schema_validator
 
     def publishing_api_payload(content)
       {
-        html: Whitehall::GovspeakRenderer.new.govspeak_to_html_with_images_and_attachments(content, @images),
+        html: Whitehall::GovspeakRenderer.new.govspeak_to_html(content, images: prepare_images(@images)),
```
TonyGDS: Related to the comment above, although the `prepare_images` method is public, it must be called before `govspeak_to_html` in order for that method to succeed. It puts responsibility on the caller to know they have to do this which they might not. It feels like this is the responsibility of `govspeak_to_html` to orchestrate.

TonyGDS (approved):

app/helpers/attachments_helper.rb
ChrisBAshton: Fair point, but AttachmentsHelper itself seems to be mostly Govspeak related. I think I'll leave this one for whoever picks up the work to [move attachments rendering out of Whitehall](https://gov-uk.atlassian.net/browse/WHIT-1922).

app/models/configurable_content_blocks/govspeak.rb
ChrisBAshton: Fixed in https://github.com/alphagov/whitehall/pull/10589/commits/2a52eb63d9c731e8984cdc9d829f670bcdc62a28!

app/helpers/govspeak_helper.rb
ChrisBAshton: Good idea. I did try it but had some failures at the time - it looked as though we were passing two different data structures at various places in Whitehall. On closer inspection, there were just some branches that ended up calling `prepare_images` _twice_. So this was easy enough to do and has now been applied 👍

ChrisBAshton: Fixed in https://github.com/alphagov/whitehall/pull/10589/commits/f4183a1c7e1bba485b56c39006d0c0a87ded488b!

app/helpers/attachments_helper.rb
ChrisBAshton: Fair point, but AttachmentsHelper itself seems to be mostly Govspeak related. I think I'll leave this one for whoever picks up the work to [move attachments rendering out of Whitehall](https://gov-uk.atlassian.net/browse/WHIT-1922).

PR #10592, opened 2025-08-26
News story configurable document type [WHIT-2381]

ryanb-gds:
This includes a new news story configurable document type.

In order to achieve parity with base Whitehall editions, we needed to:

- Add the summary field to standard editions
- Add the first publishing backdating behaviour
- Add history mode

These can all be disabled in the settings.

We also made the test organisation seed use the real GDS content ID, which helped when testing the behaviour of history pages.


GDSNewt (approved):
LGTM :shipit:

PR #10594, opened 2025-08-26
[WHIT-2270] Additional fixes for configurable documents images

lauraghiorghisor-tw:
Few tweaks for Standard Editions & other clean-ups, including:
- Fix bug where the (sidebar) image is not persisted.
- Change rendering of title so we can refer to a "history page" rather than a "standard edition"
- Include a description field in the publishing api payload.

Screenshots
Screenshot 2025-08-27 at 10 41 20
Screenshot 2025-08-27 at 10 41 08

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2270)

### DIFF
--- a/app/models/configurable_document_types/history_page.json
+++ b/app/models/configurable_document_types/history_page.json
@@ -19,7 +19,8 @@
         "type": "string",
         "format": "image_select"
       }
-    }
+    },
+    "required": ["body"]
   },
   "settings": {
     "base_path_prefix": "/government/history",
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -8,6 +8,14 @@ class StandardEdition < Edition
     "choose_type_admin_standard_editions_path"
   end
 
+  def format_name
+    type_instance.label.downcase
+  end
+
+  def display_type
+    type_instance.label
+  end
+
   def publishing_api_presenter
     PublishingApi::StandardEditionPresenter
   end
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -20,10 +20,6 @@ class StandardEdition < Edition
     PublishingApi::StandardEditionPresenter
   end
 
-  def summary_required?
-    false
-  end
-
   def body_required?
     false
   end
--- a/app/views/admin/configurable_content_blocks/_image_select.html.erb
+++ b/app/views/admin/configurable_content_blocks/_image_select.html.erb
@@ -7,8 +7,8 @@
    ] + image_select.images.map do |image|
   {
     text: image.filename,
-    value: image.id,
-    selected: image.id == content.to_i,
+    value: image.image_data.id,
+    selected: image.image_data.id == content.to_i,
   }
 end %>
 
--- a/app/views/admin/standard_editions/_form.html.erb
+++ b/app/views/admin/standard_editions/_form.html.erb
@@ -21,6 +21,25 @@
       } %>
     </div>
 
+    <div class="govuk-!-margin-bottom-6 js-locale-switcher-field">
+      <%= render "govuk_publishing_components/components/character_count", {
+        textarea: {
+          label: {
+            text: "Summary (required)",
+            heading_size: "m",
+          },
+          name: "edition[summary]",
+          value: edition.summary,
+          error_items: errors_for(edition.errors, :summary),
+          right_to_left: form.object.translation_rtl?,
+          right_to_left_help: false,
+          rows: 1,
+        },
+        id: "edition_summary",
+        maxlength: MaxLengths::SUMMARY,
+      } %>
+    </div>
+
     <% if edition.document&.slug.present? %>
       <div class="app-view-edit-edition__page-address govuk-!-margin-bottom-6">
         <%= render "govuk_publishing_components/components/heading", {
--- a/features/step_definitions/standard_edition_steps.rb
+++ b/features/step_definitions/standard_edition_steps.rb
@@ -16,6 +16,7 @@ When(/^I draft a new "([^"]*)" configurable document titled "([^"]*)"$/) do |con
   click_button("Next")
   page.choose(configurable_document_type)
   click_button("Next")
+  expect(page).to have_content("New test")
   within "form" do
     fill_in "edition_title", with: title
     fill_in "edition_block_content_page_title_heading_text", with: title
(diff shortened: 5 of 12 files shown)


app/models/configurable_document_types/history_page.json
```diff
@@ -19,7 +19,8 @@
         "type": "string",
         "format": "image_select"
       }
-    }
+    },
+    "required": ["body"]
```
lauraghiorghisor-tw, line 23: Thought I'd add this since it is [required](https://github.com/alphagov/publishing-api/blob/549513cd1c03a9b320a27119b3e8935493a73563/content_schemas/formats/history.jsonnet#L7) in pub-api. In fact, I'd argue that both the `body` and the `sidebar_image` are required properties, but perhaps the latter is a bit unclear given [the parent history page](https://www.gov.uk/government/history) with the multiple images on it. But for all other history pages, we always require the sidebar image 🤔

app/views/admin/standard_editions/_form.html.erb
```diff
@@ -21,6 +21,25 @@
       } %>
     </div>
 
+    <div class="govuk-!-margin-bottom-6 js-locale-switcher-field">
```
lauraghiorghisor-tw, line 24: Some musings on this one as well:

Shouldn't we look to include the title and summary as schema defined properties? If the schema is thought to be the "place where you go look so that you get a full picture of what this document type is all about", having the title and summary visible seems reasonable.

app/views/admin/configurable_content_blocks/_image_select.html.erb
```diff
@@ -7,8 +7,8 @@
    ] + image_select.images.map do |image|
   {
     text: image.filename,
-    value: image.id,
-    selected: image.id == content.to_i,
+    value: image.image_data.id,
+    selected: image.image_data.id == content.to_i,
```
lauraghiorghisor-tw, line 11: As per the commit message, we know this is not the greatest solution, or rather not one that will work for attachments too. More thinking required in this space.

app/models/standard_edition.rb
```diff
@@ -8,6 +8,10 @@ def self.choose_document_type_form_action
     "choose_type_admin_standard_editions_path"
   end
 
+  def display_type
```
ryanb-gds, line 15: Where is this being used? Doesn't seem to be part of this commit

app/helpers/admin/editions_helper.rb
```diff
@@ -251,4 +251,8 @@ def featurable_search_results_table_actions(edition, feature_path)
   def edition_title_link_or_edition_title(edition)
     edition.public_url ? sanitize(link_to(edition.title, edition.public_url, { class: "govuk-link" })) : edition.title
   end
+
+  def extract_format_name(edition)
```
ryanb-gds: Can't we just override the `format_name` method for standard editions instead of adding this helper?

app/views/admin/standard_editions/_form.html.erb
ryanb-gds: Maybe 🤷 not practical at the moment because we're extending the edition model. If we come across a document type that doesn't want them we can consider it

features/step_definitions/standard_edition_steps.rb
```diff
@@ -28,27 +28,30 @@
 
 When(/^I publish a submitted draft of a test configurable document titled "([^"]*)"$/) do |title|
   submitter = create(:user)
-  standard_edition = StandardEdition.new
+  image = create(:image)
+  @standard_edition = StandardEdition.new
```
ryanb-gds: Try not to use instance variables to persist test state across steps, I know it's done a lot in Whitehall but it's a [mystery guest problem](https://thoughtbot.com/blog/mystery-guest) waiting to happen. Instead load the edition anew using an argument passed to the step if possible - you're passing the title in the step below so that should do it.

I'm not actually that bothered about covering this in a feature test - as long as it's covered by the image select block test I'm happy with that. We can't cover every block with a feature.

app/helpers/admin/editions_helper.rb
ryanb-gds: And even if we're not going to override the format name we should use the "title" field from the root schema instead of humanizing the machine name

test/unit/app/presenters/publishing_api/standard_edition_presenter_test.rb
```diff
@@ -75,6 +75,29 @@ class PublishingApi::StandardEditionPresenterTest < ActiveSupport::TestCase
     assert_equal page.block_content["property_two"], content[:details][:property_two]
   end
 
+  test "it includes a title and a description" do
```
ryanb-gds, line 78: I don't know how much value there is in this test which is why I omitted it first time around. I can't see it would ever help us refactor anything and there are no conditionals or loops to test, it's a bit like testing config really. But I don't feel very strongly about it

app/helpers/admin/editions_helper.rb
lauraghiorghisor-tw: I thought that wouldn't work because it's a class method - but managed to override it just fine actually! Thanks ;)
Tapped into the title from schema as suggested.

app/models/standard_edition.rb
lauraghiorghisor-tw: This is the summary page - added some screenshots to PR summary now! - https://github.com/alphagov/whitehall/blob/8fabeb901f09820a7a614fd94a3635f922a51465/app/views/admin/editions/show/_main.html.erb#L10 -> and then the `edition_type` uses the `display_type` under the hood.

test/unit/app/presenters/publishing_api/standard_edition_presenter_test.rb
lauraghiorghisor-tw: Thinking of it as documentation. Would cover us if we want to move the title and summary into the schema I guess.

features/step_definitions/standard_edition_steps.rb
lauraghiorghisor-tw: Changed, ta ✅

Hmm, I was covering persisting across editions which is not covered elsewhere - though it should be an integration level test rather than here.

ryanb-gds (approved):
Nice one, thanks for sorting this out

PR #10595, opened 2025-08-27
Move `invalid_edition` filter test to features [WHIT-2458]

patrickpatrickpatrick:
## What

Move the `only_invalid_edition` test to `features/admin-filtering-documents.feature`

### Implementation note

All (apart from one type) of the documents derived from `GenericEdition` instances created by `FactoryBot` are invalid, as none of them are tagged (will return `false` for `has_been_tagged?`) since we don't stub `Services.publishing_api.get_expanded_links`. To get around this for the filter test (we need at least one invalid edition and one valid edition), we can use `WorldwideOrganisation` which does not require tags to be a valid. I suspect this difficulty in getting examples of valid and invalid editions is why this test was moved to the controller test suite.

## Why

The previous test for this functionality kept randomly failing. Reading the test, it seemed to be an anomaly within the controller test suite. There are no other examples of controller tests that interact with the rendered page (apart from checking to see what has been rendered). So therefore it made sense to move this test to a feature test suite.

### DIFF
--- a/features/admin-filtering-documents.feature
+++ b/features/admin-filtering-documents.feature
@@ -41,3 +41,17 @@ Feature: Filtering documents by author in admin
     When I filter by review overdue
     Then I should see the publication "I moustache you a question"
     And I should not see the publication "Standard Beard Lengths"
+
+  Scenario: Viewing only invalid documents
+    Given I am a writer
+    And a world location "United Kingdom" exists
+    When I draft a new worldwide organisation "Test Worldwide Organisation" assigned to world location "United Kingdom"
+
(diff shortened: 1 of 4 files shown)


ChrisBAshton (approved):
Seems reasonable 👍 I don't really get why it was failing before, but it's disruptive enough to our workflow that I'm happy to see it moved to a feature test and draw a line under it!

ChrisBAshton:
Given this has just blocked a deployment (!) (https://github.com/alphagov/whitehall/actions/runs/17288606391/job/49070591850), I'll merge this now 👍

PR #10599, opened 2025-08-27
Moved 'Find PDF attachment's parent page' bookmarklet into Whitehall …

leimdorfer:
Moved 'Find PDF attachment's parent page' bookmarklet into Whitehall and changed name of 'find in admin' bookmarklet' to 'Bookmarklets'.

https://gov-uk.atlassian.net/browse/WHIT-1898


.erb-lint.yml
```diff
@@ -3,7 +3,8 @@ glob: "**/*.erb"
 exclude:
   - '**/vendor/**/*'
   - 'test/fixtures/**'
-  - "app/views/admin/find_in_admin_bookmarklet/_bookmarklet.erb"
+  - "app/views/admin/bookmarklets/_bookmarklet.erb"
+  - "app/views/admin/bookmarklets/_bookmarklet_find_parent.erb"
```
ChrisBAshton: Nitpick: I'd rename these so it's clearer what each does (and remove the repetition of 'bookmarklet'):

```suggestion
  - "app/views/admin/bookmarklets/_find_in_admin.erb"
  - "app/views/admin/bookmarklets/_find_parent_document.erb"
```

app/controllers/admin/bookmarklets_controller.rb
```diff
@@ -0,0 +1,3 @@
+class Admin::BookmarkletsController < Admin::BaseController
```
ChrisBAshton, line 1: Nitpick: I'd rename/move the existing controller etc in the first commit. And then in the second commit, add the second bookmarklet. Rather than do it all in one commit.

ChrisBAshton (commented):
LGTM! I have one suggestion about renaming that I think would be worth doing, and there's also something about how the commits are structured (but I'd also be happy for you to just squish it all into one commit if you like - using `git rebase -i main`).

PR #10615, opened 2025-08-28
[WHIT-1882] Remove unused base permission methods

GDSNewt:
### **What**

This PR removes unused edition feature methods from Edition::BasePermissionMethods that always returned false and were never overridden or called anywhere in the Whitehall codebase.

These methods were investigated as part of the review into whether the base permission methods list could be pruned.

**Background**

Whitehall defines a set of “base permission methods” in Edition::BasePermissionMethods.
Each edition type may override these methods via concerns (e.g. Edition::FactCheckable, Edition::Organisations), enabling features such as associations with organisations, fact-checking, topical events, etc.

By default, all editions return false for unsupported methods to provide a uniform interface. However, some methods are never overridden or used in the application.

Method | Overridden in | Supported by (Edition Types)
-- | -- | --
can_be_associated_with_social_media_accounts? | Edition::SocialMediaAccounts | WorldwideOrganisation
can_be_associated_with_topical_events? | Edition::TopicalEvents | Speech, FatalityNotice, Consultation, DocumentCollection, DetailedGuide, Publication, NewsArticle, CallForEvidence
can_be_associated_with_roles? | Edition::Roles | WorldwideOrganisation
can_be_associated_with_role_appointments? | Edition::RoleAppointments | FatalityNotice, Consultation, NewsArticle, Publication, CallForEvidence
can_be_associated_with_worldwide_organisations? | Edition::WorldwideOrganisations | CaseStudy, NewsArticle
can_be_fact_checked? | Edition::FactCheckable | FatalityNotice, Consultation, CallForEvidence, CaseStudy, DetailedGuide, NewsArticle, Publication
can_be_related_to_mainstream_content? | (None – base only) | (Always returns false)
can_be_related_to_organisations? | Edition::Organisations | Speech, FatalityNotice, LandingPage, Consultation, StatisticalDataSet, WorldwideOrganisation, DocumentCollection, CallForEvidence, CaseStudy, DetailedGuide, NewsArticle, Publication
can_apply_to_subset_of_nations? | Edition::NationalApplicability | CallForEvidence, DetailedGuide, Publication, Consultation
can_be_grouped_in_collections? | Edition::HasDocumentCollections | Speech, Consultation, CallForEvidence, StatisticalDataSet, NewsArticle, Publication, DetailedGuide, CaseStudy
has_operational_field? | (None – base only) | (Always returns false)
national_statistic? | (None – base only) | (Always returns false)
has_consultation_participation? | (None – base only) | (Always returns false)
is_associated_with_a_minister? | Edition::RoleAppointments, Edition::Appointment | Speech (via Appointment), FatalityNotice, Consultation, NewsArticle, Publication, CallForEvidence (via RoleAppointments)
statistics? | (None – base only) | (Always returns false)
can_be_tagged_to_worldwide_taxonomy? | Edition::TaggableOrganisations | Speech, FatalityNotice, Consultation, StatisticalDataSet, DocumentCollection, CallForEvidence, CaseStudy, DetailedGuide, Publication, NewsArticle

### DIFF
--- a/app/models/concerns/edition/base_permission_methods.rb
+++ b/app/models/concerns/edition/base_permission_methods.rb
@@ -9,15 +9,10 @@ module Edition::BasePermissionMethods
     can_be_associated_with_role_appointments?
     can_be_associated_with_worldwide_organisations?
     can_be_fact_checked?
-    can_be_related_to_mainstream_content?
     can_be_related_to_organisations?
     can_apply_to_subset_of_nations?
     can_be_grouped_in_collections?
-    has_operational_field?
-    national_statistic?
-    has_consultation_participation?
     is_associated_with_a_minister?
-    statistics?
     can_be_tagged_to_worldwide_taxonomy?
   ].each do |method|
     define_method(method) { false }


app/models/concerns/edition/base_permission_methods.rb
```diff
@@ -9,15 +9,10 @@ module Edition::BasePermissionMethods
     can_be_associated_with_role_appointments?
     can_be_associated_with_worldwide_organisations?
     can_be_fact_checked?
-    can_be_related_to_mainstream_content?
     can_be_related_to_organisations?
     can_apply_to_subset_of_nations?
     can_be_grouped_in_collections?
-    has_operational_field?
-    national_statistic?
```
lauraghiorghisor-tw, line 17: I mean I have not like looked into this a lot but this - and the `statistics?` down below -  seems like it is in use to me (in `publication.rb`):
```
  def national_statistic?
    publication_type == PublicationType::NationalStatistics
  end

  def statistics?
    PublicationType.statistical.include?(publication_type)
  end
```

app/models/concerns/edition/base_permission_methods.rb
```diff
@@ -9,15 +9,10 @@ module Edition::BasePermissionMethods
     can_be_associated_with_role_appointments?
     can_be_associated_with_worldwide_organisations?
     can_be_fact_checked?
-    can_be_related_to_mainstream_content?
     can_be_related_to_organisations?
     can_apply_to_subset_of_nations?
     can_be_grouped_in_collections?
-    has_operational_field?
```
lauraghiorghisor-tw, line 16: In `fatality_notice.rb` `has_operational_field` is overridden and returns true. That said, it is not used elsewhere so it can be deleted from there (and the one associated test).

app/models/concerns/edition/base_permission_methods.rb
```diff
@@ -9,15 +9,10 @@ module Edition::BasePermissionMethods
...
-    national_statistic?
-    has_consultation_participation?
```
lauraghiorghisor-tw, line 18: I see this used in `consultation.rb`
```
  def has_consultation_participation?
    consultation_participation.present?
  end
```

lauraghiorghisor-tw (changes requested):
I think I am missing something here as all these (minus the mainstream one) look like they're in use to me. And `has_operational_field` is fine to delete, but I'd remove the usage too.

Enlighten me, please 🥲

app/models/concerns/edition/base_permission_methods.rb
GDSNewt: I think for cases like this, the edition models are not inheriting behaviour from BasePermissionMethods, so having them defined there is effectively redundant.

lauraghiorghisor-tw (approved):
On second thought and after chatting a bit:
- It seems to me that the usages for these are very well scoped to certain edition subclasses ->  there is no need to define these methods globally then.
You can argue that there is some sense in having these here as "documentation" so that if there is ever a need to "say" this in another subclass, you don't write yet another method - which I assume was the original intention. But I don't see how something about consultation can be anywhere but in a consultation; same for statistics.

I would still remove the last usage of `has_operational_field` from fatality notice and its test.

app/models/concerns/edition/base_permission_methods.rb
lauraghiorghisor-tw: They are inheriting the methods because the concern is included in the edition model [here](https://github.com/alphagov/whitehall/blob/55610ea58a9a9bb322efcad560353be02b4a9847/app/models/edition.rb#L8). But that does nothing for most of those subclasses 😅

PR #10616, opened 2025-08-28
Fix error when generated notice becomes too large

patrickpatrickpatrick:
## What

If the number of `content_id`s specified for the bulk republisher is over 100, don't include them in the `flash[:notice]`.

## Why

`republish_documents_by_content_ids` can generate a notice that will cause an `ActionDispatch::Cookies::CookieOverflow` error to occur if the list of `content_ids` provided is large enough. This is because the session generated as a result of the notice exceeds the size limit due to the notice including the list of all the `content_ids`.


GDSNewt (approved):
Sensible change 👍

PR #10617, opened 2025-08-28
Consolidate 'dry run' and 'actual run' variants of `merge_people` (WHIT-1876)

ChrisBAshton:
We are moving away from defining separate 'dry run' and 'actual run' equivalents of rake tasks due to:
- Duplication of code
- Risk of running the wrong variant

We've therefore added a confirmation step, using Thor, to have the user explicitly opt in to the 'actual' run, once presented with details of what will change. And we have deleted the old 'dry run' rake task.

We've also consolidated the test suite accordingly. Some of the 'exception like' messages were duplicated across both rake tasks, but some were specific to just the one rake task, so we've had to copy some of the `puts` statements over and reorder them such that the existing test assertions continue to pass.

JIRA: https://gov-uk.atlassian.net/browse/WHIT-1876


GDSNewt:
Looks good @kashifatcha, only thing i'd check is if there is documentation for these rake tasks anywhere. Might be a good idea to update that if so 🤓

lib/tasks/data_hygiene.rake
```diff
@@ -52,11 +41,16 @@ namespace :data_hygiene do
       next
     end
 
-    if person_to_merge.translations.count > 1
-      puts "Please manually migrate non-English translations from the person you want to merge to the person you want to keep, and retry."
+    if person_to_merge == person_to_keep
```
ryanb-gds: Am I horribly misreading this diff or is this an exact copy of the code ten lines above?

test/unit/lib/tasks/data_hygiene/merge_people_test.rb
```diff
@@ -10,141 +10,143 @@ class MergePeopleTest < ActiveSupport::TestCase
...
+        "en" => { biography: "english-biography" },
+        "fr" => { biography: "french-biography" },
+      })
+      person_to_keep = create(:person)
+
+      out, _err = capture_io { task.invoke(person_to_merge.id, person_to_keep.id) }
+
+      assert_includes out, "Please manually migrate non-English translations from the person you want to merge to the person you want to keep, and retry."
+      assert person_to_merge.reload
+    end
```
ryanb-gds: Personally I wouldn't duplicate these tests for the "yes" context, because the different context does not affect the outcome of the test. I think it would be sufficient to deal with the failures as part of the "no" context and then only check that changes have been made in the "yes" context

lauraghiorghisor-tw: Done

PR #10619, opened 2025-08-28
[WHIT-2381] Configurable document types test refactor

TonyGDS:
## What
- Added missing test coverage for checking default fields are rendered in a Standard Document
- Made `DefaultObject` more resilient when dealing with missing block content
- Refactored tests to take advantage of the StandardEditionsFactory
- Created `ConfigurableDocumentTypeHelper` to reduce boilerplate during test setup
## Why
- Test set up for configurable document types had started to get verbose and could be reduced with the use of factories and helpers

### DIFF
--- a/test/functional/admin/standard_editions_controller_test.rb
+++ b/test/functional/admin/standard_editions_controller_test.rb
@@ -22,45 +22,11 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
   end
 
   view_test "GET choose_type scopes the list of types to types that the user has permission to use" do
-    configurable_document_types = {
-      "test_type_one" => {
-        "key" => "test_type_one",
-        "schema" => {
-          "$schema": "https://json-schema.org/draft/2020-12/schema",
-          "$id": "https://www.gov.uk/schemas/test_type/v1",
-          "title" => "Test Type One",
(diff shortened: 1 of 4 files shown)


app/models/configurable_content_blocks/default_object.rb
```diff
@@ -21,6 +21,10 @@ def to_partial_path
     def publishing_api_payload(schema, content)
       output = {}
       schema["properties"].each do |property_key, property_schema|
+        unless content&.key?(property_key)
+          logger.warn "Property #{property_key} is missing from content, skipping..."
```
ChrisBAshton, line 25: This is OK for optional fields, but what if one of the missing properties is required?

ChrisBAshton (approved):
LGTM 👍

app/models/configurable_content_blocks/default_object.rb
ryanb-gds: The only time this should really ever happen is if we're republishing an old doc that's missing a new field, and we can never add "required" fields anyway without having it be optional first and then backfilling the data. A warning seems appropriate in this case

ryanb-gds (approved):
Thanks for wrapping this up @TonyGDS

PR #10625, opened 2025-09-02
bulk image upload

patrickpatrickpatrick:
- **use image width/hiegh**
- **so far**


app/controllers/admin/edition_images_controller.rb
```diff
@@ -12,59 +12,118 @@ def destroy
...
+    @images = if images_params.length.positive?
+                images_params.map { |image| create_image(image) }
+              else
+                invalid_image = @edition.images.build
+                invalid_image.build_image_data({})
+                [invalid_image]
+              end
+
+    if @images.all?(&:valid?)
+      @images.each(&:save)
```
patrickpatrickpatrick: These lines mimic the behaviour of the bulk upload, which is that if any of the images chosen are invalid then none of them will be uploaded. I do wonder if instead we could:

- save the images that are valid
- not save the images that are invalid
- render a summary of what images were uploaded and not uploaded in the index page

Otherwise it's quite resource intensive if the user is uploading multiple images and includes an incorrect one multiple times over.

app/uploaders/image_uploader.rb
```diff
@@ -1,22 +1,76 @@
...
+  end
+
+  def width_range
+    return unless bitmap?(file)
+
+    if model.respond_to?(:image_kind_config)
+      model.image_kind_config.valid_width..
+    else
+      0..
+    end
```
patrickpatrickpatrick: This commit moves the validation out of the `ImageValidator` and into the `ImageUploader` itself. As carrierwave has already inbuilt validation for content type and dimensions, I thought it made sense to make use of these? It wasn't clear to me why `ImageUploader` is used instead as this appears to result in the same validation behaviour.

[FeaturedImageUploader](https://github.com/alphagov/whitehall/blob/main/app/uploaders/featured_image_uploader.rb#L1) still uses ImageValidator so for now we can't totally remove it. Doesn't seem like it would be a lot of effort to open a follow-up PR that removes `ImageValidator` and merges FeatureImageUploader into ImageUploader.

test/unit/app/validators/image_validator_test.rb
```diff
@@ -3,7 +3,7 @@
 
 class ImageValidatorTest < ActiveSupport::TestCase
   def setup
-    @example_model = ImageData
+    @example_model = FeaturedImageData
```
patrickpatrickpatrick: The changes in this file are a result of `ImageData` no longer using ImageValidator. `FeaturedImageData` still does, so this is why this line has changed.

test/unit/app/validators/image_validator_test.rb
```diff
@@ -82,11 +78,6 @@ def image_kind_config
     end
   end
 
-  test "it allows SVG" do
```
patrickpatrickpatrick, line 85: There's a test for uploading specified MIME types above (and `FeaturedImage` does not support SVG files), so we don't need this test.

app/components/admin/edition_images/image_component.html.erb
```diff
@@ -1,24 +1,36 @@
...
-      <img src="<%= image.url %>" alt="<%= preview_alt_text %>" class="app-view-edition-resource__preview">
+    <% if image.image_data&.original_uploaded? %>
+      <% if image.image_data&.requires_crop? %>
+        <img src="<%= image.url %>" alt="<%= preview_alt_text %>" class="app-view-edition-resource__preview">
+      <% elsif image.image_data&.all_asset_variants_uploaded? %>
+        <img src="<%= image.url(:s960) %>" alt="<%= preview_alt_text %>" class="app-view-edition-resource__preview">
+      <% end %>
     <% else %>
-      <span class="govuk-tag govuk-tag--green">Processing</span>
+      <span class="govuk-tag govuk-tag--green">Uploading</span>
```
ChrisBAshton: Any strong reason for the change of label here? 

What's changed about the workflow? I can't quite tell from the code. When an image (or images) is/are uploaded, is there some sort of Ajax refresh or does the user have to reload?

"Processing" (and the associated label around "The image is being processed. Try refreshing the page.", which I see you've removed further down) covered the fact that images are not necessarily ready once they're in asset manager - they may also be being virus scanned etc. Is that no longer something to communicate to the user?

app/controllers/admin/edition_images_controller.rb
```diff
@@ -16,10 +16,44 @@ def destroy
...
   def update
-    if image.update(params.require(:image).permit(:caption, :alt_text))
+    image_params_hash = image_params.to_h
+
+    begin
+      image_params_hash["image_data"]["crop_data"] = JSON.parse(image_params_hash["image_data"]["crop_data"])
+    rescue JSON::ParserError
+      @valid_width = image_kind_config.valid_width
+      @valid_height = image_kind_config.valid_height
+      render :edit, notice: "Error processing crop data, try again"
```
ChrisBAshton: Shouldn't this be...?

```suggestion
      flash.now[:alert] = "Error processing crop data, try again"
      render :edit
```

app/models/image_data.rb
```diff
@@ -47,24 +60,33 @@ def svg?
     content_type == SVG_CONTENT_TYPE
   end
 
-private
+  # if there is no height, we can assume that this
+  # image was uploaded before the changes to save
+  # the dimensions on upload which means it must have
+  # had a valid height to have been saved
```
ChrisBAshton: When you say "valid height" do you mean 640 pixels exactly (ie 960x640) or do you mean _at least_ 640? If the latter, I'd clarify:

```suggestion
  # been at least the minimum valid height to have been saved
```

app/uploaders/image_uploader.rb
```diff
@@ -10,13 +12,43 @@ def extension_allowlist
...
 
+  def store_dimensions
+    if file && bitmap?(file) && model
+      begin
+        image = ::MiniMagick::Image.open(file.file)
+        model.dimensions ||= {}
+        model.dimensions[:width], model.dimensions[:height] = image[:dimensions]
+      rescue MiniMagick::Error, MiniMagick::Invalid
+        logger.warn("Error opening #{file.file}")
+        # model.errors.add(:file, "could not be read. The file may not be an image or may be corrupt")
```
ChrisBAshton: TODO here

app/validators/image_validator.rb
```diff
@@ -46,14 +49,10 @@ def validate_size(record, image)
     target_height = record.image_kind_config.valid_height
 
     too_small = actual_width < target_width || actual_height < target_height
-    too_large = actual_width > target_width || actual_height > target_height
 
-    return unless too_small || too_large
```
ChrisBAshton, line 51: Niiiiice 🎉

config/initializers/carrierwave.rb
```diff
@@ -4,7 +4,7 @@
 Whitehall::Application.config.to_prepare do
   CarrierWave.configure do |config|
     config.storage Storage::DefaultStorage
-    config.enable_processing = false if Rails.env.test?
+    config.enable_processing = true if Rails.env.test?
```
ChrisBAshton: Shouldn't this be enabled in all envs now?

```suggestion
    config.enable_processing = true
```

spec/javascripts/components/image-cropper-spec.js
```diff
@@ -38,13 +41,20 @@ describe('GOVUK.Modules.ImageCropper', () => {
...
-  it('should attach an image to the input when the form is submitted', (done) => {
-    form.submit = () => {
+  it('should add cropping data to input when crop takes place', (done) => {
+    const image = document.querySelector('.app-c-image-cropper__image')
+    image.addEventListener('crop', () => {
       const input = document.querySelector('.js-cropped-image-input')
-      expect(input.files.length).toBe(1)
+      expect(input.value).toBe(
+        // the cropper supports width and height that are bigger
+        // that the height and width specified if the aspect ratio
```
ChrisBAshton:
```suggestion
        // than the height and width specified if the aspect ratio
```

spec/javascripts/components/image-cropper-spec.js
```diff
@@ -38,13 +41,20 @@ describe('GOVUK.Modules.ImageCropper', () => {
...
+        // of the cropped width and height is the same (since we
+        // use "fit to size" image processing on our images)
+        JSON.stringify({ x: 0, y: 0, width: 1920, height: 1280 })
```
ChrisBAshton: Not keen on these [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) 🤔

test/unit/app/validators/image_validator_test.rb
```diff
@@ -18,10 +18,6 @@ def setup
     assert_validates_as_valid(ImageValidator.new, "960x640_jpeg_with_uppercase_extension.JPG")
   end
 
-  test "should not accept a corrupt image" do
-    assert_validates_as_invalid(ImageValidator.new, "not_an_image.jpg")
-  end
```
ChrisBAshton, line 23: This should be kept, shouldn't it?

app/assets/javascripts/components/image-cropper.js
```diff
@@ -29,15 +31,49 @@ window.GOVUK.Modules = window.GOVUK.Modules || {}
...
+
         this.initKeyboardControls()
         this.updateAriaLabel()
+
+        const cropBoxData = this.cropper.getCropBoxData()
+
+        cropBoxData.left = this.$croppingX
+        cropBoxData.top = this.$croppingY
+
+        this.cropper.setCropBoxData(cropBoxData)
```
ChrisBAshton, line 42: Can we add some tests for the changes in this file?

app/controllers/admin/edition_images_controller.rb
ChrisBAshton: Can revisit that behaviour in a later commit - this is a good MVP 👍

app/controllers/admin/edition_images_controller.rb
```diff
@@ -77,22 +77,42 @@ def create
...
+      # Uploaded files are renamed by Rails but we want to retain
+      # `original_filename` so a file can be cropped and not saved
+      # with a different name
+      sanitized_file.move_to(File.join(File.dirname(sanitized_file.path), existing_image.image_data.carrierwave_image))
 
-    new_image.image_data.validate_on_image = new_image
+      existing_image.image_data.file.store!(sanitized_file)
 
-    new_image.image_data.images << new_image
+      # remove crop data since it is a new image
```
ChrisBAshton: Lots of this code duplicates that of the first commit -  are you planning to consolidate it into some sort of helper/service?

app/models/image_data.rb
```diff
@@ -87,6 +86,10 @@ def too_large?
 
 private
 
+  def file_is_not_blank
+    errors.add(:file, message: "cannot be uploaded. Choose a valid JPEG, PNG, SVG or GIF.") if file.blank? && errors[:file].blank?
```
ChrisBAshton: I think the team consensus is that we want to start moving these error messages into the i18n locale files. Would be good to have an image-related section of error messages in the YML.

app/components/admin/edition_images/image_component.html.erb
patrickpatrickpatrick: I think the changes made here will be clearer when I make the screenshots comparing the flow

app/models/image_data.rb
patrickpatrickpatrick: The former, since I'm referring to how images used to be validated (they had to be exactly 640 otherwise they would not save).

config/initializers/carrierwave.rb
patrickpatrickpatrick: Oh yeah, probably don't need this line at all now since I guess `true` would be default lol

app/models/image_data.rb
patrickpatrickpatrick: Have moved this to the locale file for errors.

app/assets/javascripts/components/image-cropper.js
patrickpatrickpatrick: I think these are covered by the tests now but happy to add more tests if you think coverage could be improved further.

test/unit/app/validators/image_validator_test.rb
patrickpatrickpatrick: This test now errors consistently because if the image is corrupt then the image isn't saved by the model. So I thought it could be removed.

app/models/image_data.rb
```diff
@@ -76,7 +107,29 @@ def filename_is_unique
...
+  rescue CarrierWave::DownloadError
+    errors.add(:file, message: "could not crop file. Please try again.")
+  end
+
   def image_changed?
-    changes["carrierwave_image"].present?
+    # if the dimensions have changed then
+    # an image with the same name has been
+    # reuploaded and we need to revalidate
+    changes["carrierwave_image"].present? || changes["dimensions"].present?
```
patrickpatrickpatrick: Aware that this isn't actually foolproof and a user could upload a different image with the same name and same dimensions as the currently saved image.

Ideally the name of the file the user uploads and the actual filename that is saved should be different. Both names could be saved to `ImageData` so the reuploading behaviour would still work. Separation of names like this would make a lot of validation with files much easier throughout Whitehall but seems like a different issue.

config/locales/en/errors.yml
```diff
@@ -12,6 +12,10 @@ en:
       messages:
         invalid_date: "is not in the correct format"
       models:
+        image_data:
+          attributes:
+            file:
+              blank: cannot be uploaded. Choose a valid JPEG, PNG, SVG or GIF.
```
patrickpatrickpatrick: So I think actually what would best here instead would be:

```
en:
  file:
    errors:
      messages:
        blank: "cannot be uploaded. Choose a valid %{extension_list}."
```

and then using 

```
I18n.t("file.errors.blank", extension_list:)
```

for all the uploaders within Whitehall. This could be a different PR though since it would add extra lines that aren't too pertinent to the core changes in this PR.

test/unit/app/validators/image_validator_test.rb
patrickpatrickpatrick: I also think we could remove the `ImageValidator` entirely and just use the in-built Carrierwave validation. [See this follow-on PR](https://github.com/alphagov/whitehall/pull/10748) (the tests currently failing are due to the different error message content, not the validations themselves not working). We currently use Carrierwave validation for other file validation in Whitehall.

app/components/admin/edition_images/image_component.html.erb
```diff
@@ -10,11 +10,15 @@
   <div class="govuk-grid-column-two-thirds">
     <p class="govuk-body"><strong>Caption: </strong><%= caption %></p>
     <p class="govuk-body"><strong>Alt text: </strong><%= alt_text %></p>
-    <%= render "govuk_publishing_components/components/copy_to_clipboard", {
-      label: tag.strong("Markdown code:"),
-      copyable_content: image_markdown,
-      button_text: "Copy Markdown",
-    } %>
+    <% if image.image_data&.requires_crop? %>
+      <span class="govuk-tag govuk-tag--red">Requires crop</span>
```
ryanb-gds, line 14: Out of interest, what would happen if the user constructed the markdown themselves and embedded it before cropping took place?

app/components/admin/edition_images/image_component.rb
```diff
@@ -40,6 +40,6 @@ def find_image_index
   end
 
   def can_be_custom_lead_image?
-    edition.can_have_custom_lead_image? && !image.svg?
+    edition.can_have_custom_lead_image? && !image.svg? && !image.image_data.requires_crop?
```
ryanb-gds, line 43: @lauraghiorghisor-tw we'll need to add this to the lead image select block if this merges

app/controllers/admin/edition_images_controller.rb
```diff
@@ -16,7 +16,26 @@ def destroy
...
   def update
-    if image.update(params.require(:image).permit(:caption, :alt_text))
+    @valid_width = image_kind_config.valid_width
+    @valid_height = image_kind_config.valid_height
+
+    image_params_hash = image_params.to_h
+
+    begin
+      if image_params_hash["image_data"]["crop_data"]
+        image_params_hash["image_data"]["crop_data"] = JSON.parse(image_params_hash["image_data"]["crop_data"])
```
ryanb-gds: This feels a bit off, I wouldn't expect you to need to do this manually, Rails is good at dealing with this kind of stuff generally. Probably the "Rails way" of doing this would be to have four separate hidden inputs for your crop data rather than a single one storing a JSON string as a value. I can see why that would feel a bit ick, but I'd much rather handle as much of this declaratively at the view layer as possible than be doing a lot of parameter wrangling in a controller, because it's far less prone to unexpected bugs such as nil pointer errors etc.

app/controllers/admin/edition_images_controller.rb
```diff
@@ -16,7 +16,26 @@ def destroy
...
+      end
+    rescue JSON::ParserError
+      render :edit, alert: "Error processing crop data, try again"
+    end
+
+    image.assign_attributes(image_params_hash.except("image_data"))
```
ryanb-gds: If you want to stop the image data getting updated you can just remove [the `accepts_nested_attributes_for` call from the image model](https://github.com/alphagov/whitehall/blob/86feed7e76921e4415e5dbe3263971bcec340d07/app/models/image.rb#L11), but I suspect if you take my advice above you won't have this problem.

app/controllers/admin/edition_images_controller.rb
```diff
@@ -16,7 +16,26 @@ def destroy
...
+
+    image.image_data.crop_data = image_params_hash["image_data"]["crop_data"]
+
+    image.image_data.validate_on_image = image
```
ryanb-gds, line 31: Oooofff, wish we didn't do this, I know it's existing code though. But yeah, we should really have a unique filename validation for images at the edition level, not set transitive attributes on the image data to walk back up the association chain. Grim.

app/controllers/admin/edition_images_controller.rb
```diff
@@ -25,46 +44,66 @@ def update
...
   def create
-    @new_image = @edition.images.build
-    @new_image.build_image_data(image_params["image_data"])
-
-    @new_image.image_data.validate_on_image = @new_image
-    # so that auth_bypass_id is discoverable by AssetManagerStorage
-    @new_image.image_data.images << @new_image
-
-    if @new_image.save
+    @images = if images_params.length.positive?
```
ryanb-gds: Would `images_params.any?` work here?

app/controllers/admin/edition_images_controller.rb
```diff
@@ -25,46 +44,66 @@ def update
...
+      @images.each { |image| image.id.nil? && @edition.images.delete(image) }
+    end
+
+    render :index
+  end
+
+  def create_image(image)
+    existing_image = @edition.images.joins(:image_data).where(["image_data.carrierwave_image = ?", image["image_data"]["file"].original_filename]).first
+
+    if existing_image.present?
```
ryanb-gds: This approach of just loading the existing image for an existing filename has led to a Zendesk ticket today, should we consider just raising a validation error on upload? I suspect that would keep our code a lot simpler too

app/controllers/admin/edition_images_controller.rb
```diff
@@ -16,7 +16,26 @@ def destroy
   end
 
   def update
-    if image.update(params.require(:image).permit(:caption, :alt_text))
+    @valid_width = image_kind_config.valid_width
+    @valid_height = image_kind_config.valid_height
```
ryanb-gds: Would probably prefer to do this in the template rather than have mystery instance vars floating around in the controller, especially as I expect that the image is already in the template scope. You could add an image helper method if you don't what to navigate the object tree that deeply in the template

app/controllers/admin/edition_images_controller.rb
```diff
@@ -25,46 +44,66 @@ def update
...
+
+      new_image
     end
   end
 
   def edit
-    image = Image.find(params[:id])
-    flash.now.notice = "The image is being processed. Try refreshing the page." unless image&.image_data&.all_asset_variants_uploaded?
+    @valid_width = image_kind_config.valid_width
+    @valid_height = image_kind_config.valid_height
```
ryanb-gds: Same comment as above about instance vars

app/models/image.rb
```diff
@@ -10,14 +10,24 @@ class Image < ApplicationRecord
...
   default_scope -> { order(:id) }
 
   def url(*args)
     image_data.file_url(*args)
   end
 
+  def thumbnail
+    return image_data.file_url unless bitmap? && !requires_crop?
+
+    variant = image_data.assets.find { |asset| asset.variant != "original" }&.variant&.to_sym
```
ryanb-gds, line 34: Should we be looking for something specific here rather than any variant that isn't the original?

app/models/image_data.rb
```diff
@@ -76,7 +107,29 @@ def filename_is_unique
...
+  def crop_data_changed?
+    too_large? && changes["crop_data"].present?
+  end
+
+  def recreate_cropped_assets
+    file.download! file.url
+
+    assets.each do |asset|
+      unless asset.original?
+        AssetManager::AssetDeleter.call(asset.asset_manager_id)
```
ryanb-gds: This might be problematic. We don't want to actually delete those assets from Asset Manager until publishing time, because they are likely still referenced by the production content. I'm not sure how best to solve that, you'd need to sort of mark them as deleted in the WH database and then call the deleter for each marked asset at publish time. I imagine the existing image deletion behaviour works something like this though so it should be possible.

app/uploaders/image_uploader.rb
```diff
@@ -1,22 +1,80 @@
...
+
+  def downloader
+    WhitehallDownloader
   end
 
   def extension_allowlist
     %w[jpg jpeg gif png svg]
   end
 
+  def height_range
```
ryanb-gds: I can only see this used in one place in the PR and only it's beginning value is used, so perhaps not really necessary?

app/uploaders/image_uploader.rb
```diff
@@ -1,22 +1,80 @@
...
+    end
+  end
+
+  def store_dimensions
+    if file && bitmap?(file) && model
+      begin
+        image = ::MiniMagick::Image.open(file.file)
+
+        model.dimensions = {}
+        model.dimensions["width"], model.dimensions["height"] = image[:dimensions]
```
ryanb-gds: You may like to use [Active Record store accessors](https://api.rubyonrails.org/classes/ActiveRecord/Store.html) for the dimensions and crop data attributes, saves you dealing with hashes

app/uploaders/image_uploader.rb
```diff
@@ -1,22 +1,80 @@
...
+      rescue MiniMagick::Error, MiniMagick::Invalid
+        raise CarrierWave::IntegrityError, "could not be read. The file may not be an image or may be corrupt"
+      end
+    end
+  end
+
   Whitehall.image_kinds.each do |image_kind, image_kind_config|
     use_versions_for_this_image_kind_proc = lambda do |uploader, opts|
-      uploader.model.image_kind == image_kind && uploader.bitmap?(opts[:file])
+      uploader.model.image_kind == image_kind && uploader.bitmap?(opts[:file]) && !uploader.model.requires_crop?
```
ryanb-gds, line 34: I'm slightly confused about the addition of the check to see that the image doesn't require cropping. Might we not still want to run the `resize_to_fill` process? Also we seem to check `requires_crop` again inside the `crop_image?` method on the version. But tbh I don't really know how carrierwave works so suspect I just need to read the docs again

features/step_definitions/image_steps.rb
```diff
@@ -1,5 +1,8 @@
 Given("a draft document with images exists") do
-  images = [build(:image), build(:image)]
+  svg_image_data = build(:image_data, file: File.open(Rails.root.join("test/fixtures/images/test-svg.svg")))
```
ryanb-gds, line 2: I think there might be a trait on the image factory now (soon? might be waiting on the lead image PR, can't remember)

app/controllers/admin/edition_images_controller.rb
patrickpatrickpatrick: Ah, that actually makes a lot more sense. Would much rather do that than handling the JSON.

app/controllers/admin/edition_images_controller.rb
patrickpatrickpatrick: This is unfortunately how we handle files in the bulk uploader too, I'd much prefer if it the filename was stored separately from the uploaded file within the `AttachmentData`/`ImageData` as it would make a lot of things much easier to validate and implement. It does feel like a separate task to rework this.

app/uploaders/image_uploader.rb
patrickpatrickpatrick: This is used [here](https://github.com/carrierwaveuploader/carrierwave/blob/master/lib/carrierwave/uploader/file_size.rb) in Carrierwave to validate the image dimensions.

app/uploaders/image_uploader.rb
patrickpatrickpatrick: I'll take a look at this, not heard of them before!

app/models/image_data.rb
patrickpatrickpatrick: Ah, am I not understanding assets correctly? I assumed that when a user made a draft, then the images and their assets would be duplicated. So I thought that this would be safe to do without impacting the actually published edition.

ryanb-gds: The files in Asset Manager are not duplicated. Image records are duplicated, image_data records and asset records are not.

app/uploaders/image_uploader.rb
patrickpatrickpatrick: Have updated to use Active Record store accessors, a lot better.

app/uploaders/image_uploader.rb
patrickpatrickpatrick: It's to prevent making versions of an image that needs to be cropped. I thought this was a good way to stop unnecessary processing from happening.

app/models/concerns/replaceable.rb
```diff
@@ -0,0 +1,49 @@
...
+  def replaced?
+    replaced_by.present?
+  end
+
+  def replacement_asset_for(asset)
+    replaced_by.assets.where(variant: asset.variant).first || replaced_by.assets.where(variant: Asset.variants[:original]).first
+  end
+
+  def replace_with!(replacement)
+    # NOTE: we're doing this manually because carrierwave is setup such
```
ryanb-gds: I reckon you can probably drop this comment, because Asset Manager handles virus checking these days, not Whitehall

app/models/image.rb
patrickpatrickpatrick: This is just to present to the user a cropped version of the uploaded image. It might be a specific variant hasn't yet uploaded on page load, so this is why I've implemented it like this.

app/components/admin/edition_images/image_component.html.erb
patrickpatrickpatrick: I think an error would likely occur since there's logic which stops variants from being generated if the crop hasn't been provided. I'm wondering if I should perhaps remove that which although would generate unnecessary images would make edition validation significantly easier.

app/controllers/admin/edition_images_controller.rb
```diff
@@ -25,47 +39,50 @@ def update
...
-    # so that auth_bypass_id is discoverable by AssetManagerStorage
-    @new_image.image_data.images << @new_image
-
-    if @new_image.save
+    @images = if images_params.any?
+                images_params.map { |image| create_image(image) }
+              else
+                invalid_image = @edition.images.build
+                invalid_image.build_image_data({})
+                [invalid_image]
```
ryanb-gds: Apologies if I should have caught this on the first review, but this building an invalid image seems a bit weird. I can kind of see why we'd do it to save ourselves checking if the @images value is set but I think to be honest we'd be better off just doing the null checks.

Or perhaps empty checks - would `@images = images_params.map { |image| create_image(image) }` work?

app/controllers/admin/edition_images_controller.rb
```diff
@@ -25,47 +39,50 @@ def update
...
-      @valid_height = image_kind_config.valid_height
-      @data_url = image_data_url
-      render :crop
+      flash[:notice] = "Images successfully uploaded"
     else
-      @new_image.errors.delete(:"image_data.file", :too_large)
-      # Remove @new_image from @edition.images array, otherwise the view will render it in the 'Uploaded images' list
-      @edition.images.delete(@new_image)
-      render :index
+      flash[:notice] = nil
```
ryanb-gds: Should be nil by default, shouldn't it?

app/controllers/admin/edition_images_controller.rb
```diff
@@ -25,47 +39,50 @@ def update
...
+  end
+
+  def create_image(image)
+    new_image = @edition.images.build
+
+    new_image.build_image_data(image["image_data"])
+
+    new_image.image_data.validate_on_image = new_image
+
+    new_image.image_data.images << new_image
```
ryanb-gds, line 67: I don't think you should need this line, the association keys should be set automatically when you save the image on line 51

app/models/attachment_data.rb
```diff
@@ -2,6 +2,10 @@
 require "timeout"
 
 class AttachmentData < ApplicationRecord
+  include Replaceable
```
ryanb-gds, line 5: Nice to be able to share this behaviour across "data" classes, good work

app/controllers/admin/edition_images_controller.rb
patrickpatrickpatrick: It seems we do need it otherwise the comment is correct, `AssetManagerStorage` [isn't able to get](https://github.com/alphagov/whitehall/actions/runs/18779502346/job/53581868480) the `auth_bypass_id`.

app/components/admin/edition_images/image_component.rb
ChrisBAshton: In that case please hold off on merging this PR until #10635 is done 🙏 I don't want to have to change the lead image setting (and do all the integration import testing) again if I can help it!

db/migrate/20250902091240_add_fields_to_image_data.rb
```diff
@@ -0,0 +1,8 @@
+class AddFieldsToImageData < ActiveRecord::Migration[8.0]
+  def change
+    change_table :image_data, bulk: true do |t|
+      t.json "dimensions"
+      t.json "crop_data"
```
ChrisBAshton: Strictly speaking I'd move this into its own, earlier, commit. (Not entirely sure of the sequencing of db migrate vs pod deployment, if we roll out both the DB changes and the code that references the new fields, in the same PR. Don't we need to ship the DB migration first as a separate PR?)

app/uploaders/image_uploader.rb
```diff
@@ -1,22 +1,58 @@
...
   include CarrierWave::MiniMagick
 
+  process :store_dimensions
+
   configure do |config|
     config.remove_previously_stored_files_after_update = false
     config.storage = Storage::PreviewableStorage
   end
 
+  def downloader
```
ChrisBAshton, line 11: Where is this called?

app/controllers/admin/edition_images_controller.rb
patrickpatrickpatrick: It's a bit strange looking but it's to avoid having to make any changes to the `ErrorSummaryComponent` in this PR. Happy to make a follow-up PR in which I make those changes.

features/step_definitions/image_steps.rb
```diff
@@ -100,15 +100,30 @@
...
 
 And(/^I upload a (\d+)x(\d+) image$/) do |width, height|
-  within "input.gem-c-file-upload" do
-    if width == 960 && height == 640
-      attach_file jpg_image
-    elsif width == 64 && height == 96
-      attach_file Rails.root.join("test/fixtures/horrible-image.64x96.jpg")
-    elsif width == 960 && height == 960
-      attach_file Rails.root.join("test/fixtures/images/960x960_jpeg.jpg")
+  if running_javascript?
```
ChrisBAshton: Nice!

app/uploaders/image_uploader.rb
patrickpatrickpatrick: This is [overloading the base](https://www.rubydoc.info/gems/carrierwave/CarrierWave/Downloader/Base) `downloader` from `CarrierWave::Uploader::Base`. We need to override [skip_ssrf_protection](https://github.com/alphagov/whitehall/blob/a9ca8ed1a0b188a2b4ec9d6e59f63a40cb1c4bc3/app/uploaders/whitehall_downloader.rb), without doing this an error will occur [when EditionImagesController calls](https://github.com/alphagov/whitehall/blob/a9ca8ed1a0b188a2b4ec9d6e59f63a40cb1c4bc3/app/controllers/admin/edition_images_controller.rb#L26) `download!` in test and development environments.

patrickpatrickpatrick: I can put a comment saying this in the `image_uploader` if that makes it clearer.

patrickpatrickpatrick:
> Looking good! I have every confidence this is an improvement on the current code and I wouldn't block merging it.
> 
> But where did we arrive at with the decision around overwriting files of the same name? I thought we'd said we should raise a validation error in that scenario, forcing the publisher to delete the original if they want to upload a new file of the same name.

Yep, that's what happens now! The overwriting files functionality was in a different commit, so I just dropped it.

db/migrate/20250902091240_add_fields_to_image_data.rb
patrickpatrickpatrick: Yeah, I wasn't sure about what to do with regards to the migration. What you're saying about the separate PR for the migrations makes sense.

app/uploaders/image_uploader.rb
ChrisBAshton: Sure, thanks!

PR #10626, opened 2025-09-02
Prevent H4s being converted to H3s inadvertently during govspeak header numbering [WHIT-2473]

ryanb-gds:
In [this commit](https://github.com/alphagov/whitehall/pull/10589/commits/e026691e77a49add1672dd6a114387eafa2183f5) we moved header numbering into a pre govspeak processing rather than a post govspeak processing step. Unfortunately the regex we used to capture the H2 and H3 headings also captured lower level headings and converted them into H3s, leaving the extra hashes in the heading text. This commit prevents heading numbers from being applied for headings lower than H3.

### DIFF
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -134,9 +134,12 @@ private
     h2 = 0
     h3 = 0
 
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Oops - thanks!

PR #10628, opened 2025-09-02
Remove HTTP URI scheme checking from Whitehall's link validation

ChrisBAshton:
http, https and mailto schemes are already allowed by Link Checker API, as is the `tel` scheme (which probably should always have been allowed in Whitehall), here:
https://github.com/alphagov/link-checker-api/blob/62c5a13760da736fa7d9c05e348a410f511506f1/app/lib/link_checker/uri_checker/valid_uri_checker.rb#L26-L47

The only other allowable 'prefix' according to Whitehall's validator, apart from an internal link (`/`) is a `#` (anchor link). Anchor links aren't checked by Link Checker API, but Whitehall doesn't offer any additional security here either: in my test, I had a link `[anchor](#foo)` with no corresponding `Foo` heading in the GovSpeak and Whitehall raised no validation concerns. So I argue that the code deleted in this commit had no value beyond what is already offered by Link Checker API.


app/validators/internal_path_links_validator.rb
```diff
@@ -5,11 +5,15 @@ def validate(record)
     matches(record.body, /\[.*?\]\((\S*?)(?:\s+"[^"]+")?\)/) do |match|
       link = match[1]
 
-      fix = if self.class.is_internal_admin_link?("/#{link}")
+      fix = if link.start_with?("//")
```
lauraghiorghisor-tw, line 8: What if it start with `///`? 😅 Perhaps it's valid to change to `/\/\/+/`

app/validators/internal_path_links_validator.rb
```diff
@@ -5,11 +5,15 @@ def validate(record)
...
+              # Collapse //, ///, //// etc down to a single leading slash for the suggestion
+              suggestion = "/#{link.gsub(/\A\/+/, '')}"
+              "This is an invalid admin link.  Did you mean #{suggestion} instead of #{link}?"
```
lauraghiorghisor-tw: Is the double space here `link.  Did` intentional?

test/unit/app/validators/govspeak_link_validator_test.rb
```diff
@@ -69,16 +42,4 @@ class GovspeakLinkValidatorTest < ActiveSupport::TestCase
...
     assert_not GovspeakLinkValidator.is_internal_admin_link?(nil)
   end
-
-  test "should permit mailto links" do
-    test_model = Edition.new(body: "[example text](mailto:test@example.com)")
-    GovspeakLinkValidator.new.validate(test_model)
-    assert_equal 0, test_model.errors.size
-  end
-
-  test "should permit anchor links" do
```
lauraghiorghisor-tw, line 79: I agree with your reasoning here 😎

lauraghiorghisor-tw (approved):
LGTM

app/validators/internal_path_links_validator.rb
ChrisBAshton: This will work for that case too (since it still starts with `//`)

app/validators/internal_path_links_validator.rb
ChrisBAshton: Nope - good spot, thanks!

app/validators/internal_path_links_validator.rb
lauraghiorghisor-tw: ...oh 😂

PR #10633, opened 2025-09-03
Refactor edition organisations association trait module

ryanb-gds:
This is a pure readability improvement. The previous implementation of this module was hard to follow and to reason about.

It's worth noting why we don't use normal associations for lead and supporting organisations. Rails persists association changes to has_many associations as soon as they're assigned, meaning that validation never runs. This meant when using real associations, users experienced a server error instead of a validation error. We get around this by implementing faux associations. The new implementation isn't really any different from the old, but it's more concise and avoids keeping state in instance variables.

We also delete a few unused methods.

I started this thinking it would make [WHIT-2386](https://gov-uk.atlassian.net/browse/WHIT-2386). I'm not sure it really does, on reflection, but at least there aren't any occurrences of the word "mange" in Whitehall code now, so I think that it has been worth it 🤞 

[WHIT-2386]: https://gov-uk.atlassian.net/browse/WHIT-2386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/admin/edition_access_limited_controller.rb
+++ b/app/controllers/admin/edition_access_limited_controller.rb
@@ -67,9 +67,7 @@ private
 
   def changed?
     if @edition.can_be_related_to_organisations?
-      @edition.changed? ||
-        @edition.lead_organisation_ids.map(&:to_s) != edition_params[:lead_organisation_ids] ||
-        @edition.supporting_organisations.map(&:id).map(&:to_s) != edition_params[:supporting_organisation_ids]
+      @edition.changed? || @edition.edition_organisations != Edition.find(params[:id]).edition_organisations
     else
       @edition.changed?
     end
--- a/app/models/concerns/edition/organisations.rb
+++ b/app/models/concerns/edition/organisations.rb
@@ -10,11 +10,8 @@ module Edition::Organisations
   end
 
   included do
-    has_many :edition_organisations, foreign_key: :edition_id, dependent: :destroy, autosave: true
-    has_many :organisations, -> { includes(:translations) }, through: :edition_organisations
-
-    before_validation :mark_for_destruction_all_edition_organisations_for_destruction
-    after_save :clear_edition_organisations_touched_or_destroyed_by_lead_or_supporting_organisations_setters
+    has_many :edition_organisations, foreign_key: :edition_id, dependent: :destroy, autosave: true, validate: false
+    has_many :organisations, -> { includes(:translations) }, through: :edition_organisations, validate: false
 
     validate :at_least_one_lead_organisation
     validate :no_duplication_of_organisations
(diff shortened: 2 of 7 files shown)


app/controllers/admin/edition_access_limited_controller.rb
```diff
@@ -67,9 +67,7 @@ def clean_organisation_params
 
   def changed?
     if @edition.can_be_related_to_organisations?
-      @edition.changed? ||
-        @edition.lead_organisation_ids.map(&:to_s) != edition_params[:lead_organisation_ids] ||
-        @edition.supporting_organisations.map(&:id).map(&:to_s) != edition_params[:supporting_organisation_ids]
```
ryanb-gds, line 72: This is called after attribute assignment, so really this code shouldn't work as designed. It did because the previous organisations implementation didn't provide a single view of the organisations associated with an edition - you could update the edition organisations association value in memory but lead_organisations_ids would fetch the data from the database.

app/models/concerns/edition/organisations.rb
```diff
@@ -10,64 +10,53 @@ def process_associations_before_save(edition)
...
     add_trait Trait
   end
 
+  def sorted_organisations
+    organisations.alphabetical
+  end
+
   def lead_edition_organisations
-    edition_organisations.where(lead: true).order("edition_organisations.lead_ordering")
+    edition_organisations.select(&:lead?).sort { |a, b| a.lead_ordering <=> b.lead_ordering }
```
ryanb-gds, line 27: Changing this to use the in memory association rather than querying the database ensures consumers always get the current state of the edition organisations. Same for other similar methods below.

app/models/concerns/edition/organisations.rb
```diff
@@ -118,89 +84,49 @@ def at_least_one_lead_organisation
...
+  # If one does, then update it, add it to our new list of edition organisations and record it as updated.
+  # We then exclude already updated edition organisations from further updates, because we want to allow duplicates
+  # to be created if any are included in the new IDs so that they can be validated, rather than silently dropped.
+  # If no matching edition organisation exists, build a new one and add it to the new list.
+  # Next, add the existing lead organisations to the new supporting organisations, or vice versa.
+  # Finally, replace the existing edition organisations with the new list in memory using the Active Record Collection Proxy
+  # replace method (https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-replace),
+  # which effectively deletes any old edition organisations that were not present in the input.
+  # We then rely on the autosave attribute on the edition_organisations association to ensure these
+  # changes are saved when the edition is saved.
```
ryanb-gds: I hope this makes sense 😅

test/support/admin_edition_controller_test_helpers.rb
```diff
@@ -548,8 +548,10 @@ def should_allow_lead_and_supporting_organisations_for(edition_type)
         organisation2 = create(:organisation)
         organisation3 = create(:organisation)
 
-        edition = create(edition_type, organisations: [organisation1, organisation2])
-        edition.organisations << organisation3
+        edition = build(edition_type)
+        edition.lead_organisations = [organisation1, organisation2]
+        edition.supporting_organisations = [organisation3]
+        edition.save!
```
ryanb-gds, line 554: These tests started failing due to a validation error of no lead org assigned after my changes, so I believe the old code must have set lead to true by default. I wouldn't think in real world usage we would assign to plain `organisations` so I have made the test more explicit rather than seek to handle that issue.

test/unit/lib/data_hygiene/bulk_organisation_updater_test.rb
```diff
@@ -183,25 +183,26 @@ def process(raw_csv)
 
     process(csv_file)
 
-    assert_equal [organisation], edition.lead_organisations
+    assert_equal [organisation], edition.reload.lead_organisations
```
ryanb-gds, line 186: Previously `lead_organisations` always hit the database, but that's no longer the case, so we need to reload now.

app/models/concerns/edition/organisations.rb
```diff
@@ -82,33 +71,10 @@ def skip_organisation_validation?
...
-  # Presence and uniqueness of associated organisations are already enforced
-  # by our own custom validators, so skipping these auto-generated validations
-  # has no negative side-effects.
-  def validate_associated_records_for_edition_organisations
-    # no-op: prevent associated EditionOrganisation validations from running
-  end
-
-  def validate_associated_records_for_organisations
-    # no-op: prevent associated Organisation validations from running
-  end
```
ChrisBAshton, line 110: 🤮

app/models/concerns/edition/organisations.rb
ChrisBAshton: I'd break up the above comment block and move the relevant bits inline in the code below. Easier to follow without having to scroll up and down, and more likely to not fall out of sync with the code.

app/models/concerns/edition/organisations.rb
```diff
@@ -118,89 +84,49 @@ def at_least_one_lead_organisation
...
+  # to be created if any are included in the new IDs so that they can be validated, rather than silently dropped.
+  # If no matching edition organisation exists, build a new one and add it to the new list.
+  # Next, add the existing lead organisations to the new supporting organisations, or vice versa.
+  # Finally, replace the existing edition organisations with the new list in memory using the Active Record Collection Proxy
+  # replace method (https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-replace),
+  # which effectively deletes any old edition organisations that were not present in the input.
+  # We then rely on the autosave attribute on the edition_organisations association to ensure these
+  # changes are saved when the edition is saved.
+  def assign_organisation_ids(new_organisation_ids, is_lead)
+    new_organisation_int_ids = new_organisation_ids.map(&:to_i)
```
ChrisBAshton, line 96: I'd have thought the `new_organisation_ids` param was already an array of integers. Can it be renamed to `new_organisation_string_ids` or whatever it is that's being passed, so that we can call this variable `new_organisation_ids`?

ryanb-gds: Problem is it might already be an array of integers, e.g. from the bulk organisation importer, but if it's from an HTTP request it isn't. So calling the main input "strings" is probably more misleading potentially. I'll have a look to see if I can trace down all of the call sites to confirm they're always strings, but it'l be tricky because obviously they can be mass assigned

PR #10635, opened 2025-09-04
Import news articles from Content Publisher (WHIT-2440)

ChrisBAshton:
# What & Why

This PR takes the output from the [Content Publisher export script](https://github.com/alphagov/content-publisher/pull/3311) and imports those news stories and press releases into Whitehall. It carries over associations, images and attachments, as well as a condensed document history. Migrating content out of Content Publisher will allow us to retire that long-deprecated app 🎉  It will also be a good first test case for our new config-driven news articles format.

In terms of the comprehensiveness of transfer of content, we're striking a balance of replicating just the published news articles, as best as we can, without worrying about replicating the entire document history. We're also not bothering to transfer drafts nor deleted/'gone' content. There is no draft newer than early 2024 and the nature of news articles is such that it's extremely unlikely an update will need to be applied. Publishers will be able to create new drafts on the migrated content once it's in Whitehall. Similarly, a 'gone' news article is extremely unlikely to need to be brought back, and publishers can always replicate the 'gone' news article manually if needed.

Post merge, we will bulk import the news articles as above, and then **revert this PR** (none of the changes are required to remain once the articles have been migrated, with the exception of the tweak to `app/components/admin/editions/editorial_remark_component.html.erb`).

JIRA: https://gov-uk.atlassian.net/browse/WHIT-2440

# Screenshots

## Frontend

The before and after rendering should be identical. Example: <https://www.gov.uk/government/news/secretary-of-state-for-work-and-pensions-attends-g7-labour-and-employment-ministerial-meeting-in-germany>

| Before | After |
|--------|------|
|  |  |

### Footnotes

As described in [13e038443a22a64ac9b61a43dac5bc398fcdcc68](https://github.com/alphagov/whitehall/pull/10635/commits/13e038443a22a64ac9b61a43dac5bc398fcdcc68), there are a small handful of articles that make use of footnotes, which is not allowed in Whitehall. I've taken the pragmatic decision to move the footnote contents inline, rather than risk dropping something potentially important.

| Before | After |
|--------|------|
|  |  |

### Change history

The condensed public change history for <https://www.gov.uk/government/news/1-million-to-help-nhs-reclaim-debts-from-overseas-visitors>:

| Before | After |
|--------|------|
|  |  |

This is the one downside of exporting only the published edition. The alternative would have been to export _all_ editions and manually recreate each one, as it is only possible to associate one changenote per edition (we can't send an array of changenotes with one edition).

This is deemed tolerable, given these are short-lived news articles and that ultimately the alternative was to withdraw them altogether.

### Publishing new versions

Successfully tested publishing a new major version of the above document:

> 

## Admin

Example doc on Content Publisher: <https://content-publisher.integration.publishing.service.gov.uk/documents/36d03d5e-eac6-4c18-9d29-f02f6bbf6cc1:en>

| Content Publisher | Whitehall |
|-------------------|-----------|
| |  |

# How to run

Importing a single news article via its JSON:

```
$ kubectl get pods
# returns list of pods, including
# whitehall-admin-c4c7c957c-9q966

$ kubectl cp ~/Downloads/sample-article.json whitehall-admin-c4c7c957c-9q966:/tmp
# copies the file

$ kubectl exec whitehall-admin-c4c7c957c-9q966 -- rake import:news_article["/tmp/sample-article.json"]
# runs the rake task
```

Importing all of the news articles:

```
$ kubectl get pods
# returns list of Whitehall application pods (such as whitehall-admin-c4c7c957c-9q966)
# and Whitehall worker pods (such as whitehall-admin-worker-777ff7f7b7-lhfkb)

# Recursively copy the local directory into the worker pods
$ kubectl cp /path/to/your/news-articles whitehall-admin-worker-777ff7f7b7-lhfkb:/tmp/news-articles
$ kubectl cp /path/to/your/news-articles whitehall-admin-worker-another-one-lhfkb:/tmp/news-articles

# Now do the same for one of the application pods
$ kubectl cp /path/to/your/news-articles whitehall-admin-c4c7c957c-9q966:/tmp/news-articles

# Finally, run the rake task on the application pod you uploaded to.
# (Best do this directly on the pod, in case of time-outs)
$ kubectl exec whitehall-admin-c4c7c957c-9q966 -- rake import:news_articles_in_directory["/tmp/news-articles"]
# runs the rake task
```

Explanation:

1. We need to upload the JSON files to the application pod so that the workers get enqueued in the first place. (The results of `Dir.glob("#{args[:dir_path]}/**/*.json").each` is what powers the enqueuing - so we need to ensure all the files are there to be globbed!)
2. We need to upload the JSON files to ALL the worker pods so that every worker has access to the given JSON file. (Workers will just look locally for `/tmp/news-articles/<file>.json`, so it needs to exist locally).
3. **Note that** you will need to re-upload all of the JSON files to all of the pods as above, every time there is a deployment to Whitehall (which tears down the pods and rebuilds them)

Sense check that it all imported correctly:

```
Document.where(document_type: "StandardEdition").map(&:live_edition).select { |ed| ed && ["press_release", "news_story"].include?(ed.configurable_document_type)}.count
=> 2979
^ there should be 2979, the same as the number of JSON files in the local export

expected_slugs = Dir.glob("/tmp/news-articles/**/*.json").map { |filepath| filepath.split("/").last.sub(".json", "") }
=> 
["1-3-billion-investment-to-deliver-homes-infrastructure-and-jobs",
...
actual_slugs = Document.where(document_type: "StandardEdition").pluck(:slug)
=> 
["10-downing-street",
...

expected_slugs - actual_slugs
=> ["helicopter-services-deal-raises-competition-concerns-1", "non-executive-director-appointed-to-gov-facility-services-limited-1"]
```

^ This last one should be an empty array. If for whatever reason there are some stragglers, they can be imported by running the bulk task again, or importing individually (`rake import:news_article["/tmp/news-articles/non-executive-director-appointed-to-gov-facility-services-limited-1.json"]`).

Finally, we can destroy all StandardEdition news articles (`Edition.where(configurable_document_type: ["news_article", "press_release"]).map(&:document).map(&:destroy)`) if we want to test re-import everything (e.g. after making a tweak).

---

### DIFF
--- a/test/unit/lib/whitehall/document_importer_test.rb
+++ b/test/unit/lib/whitehall/document_importer_test.rb
@@ -43,6 +43,7 @@ module Whitehall
         "attachments" => [],
         "images" => [],
       }
+      ConfigurableDocumentType.reset_document_types!
     end
 
     describe ".import!" do
--- a/app/models/configurable_document_type.rb
+++ b/app/models/configurable_document_type.rb
@@ -4,20 +4,22 @@ class ConfigurableDocumentType
   @types_mutex = Mutex.new
 
   def self.types
-    return @types if @types
-
-    @types_mutex.synchronize do
-      return @types if @types
-
-      @types = Dir.glob("app/models/configurable_document_types/*.json").each_with_object({}) do |filename, hash|
-        data = JSON.parse(File.read(filename))
-        hash[data["key"]] = data
-      end
-    end
+    @types_mutex.synchronize { @types ||= load_document_types }
   end
 
   def self.setup_test_types(test_types)
-    @types = test_types
+    @types_mutex.synchronize { @types = test_types }
+  end
+
+  def self.reset_document_types!
+    @types_mutex.synchronize { @types = load_document_types }
+  end
+
+  def self.load_document_types
+    Dir.glob("app/models/configurable_document_types/*.json").each_with_object({}) do |filename, hash|
+      data = JSON.parse(File.read(filename))
+      hash[data["key"]] = data
+    end
   end
 
   def self.find(type_key)
--- a/test/unit/app/sidekiq/document_import_worker_test.rb
+++ b/test/unit/app/sidekiq/document_import_worker_test.rb
@@ -14,6 +14,7 @@ class DocumentImportWorkerTest < ActiveSupport::TestCase
         document: build(:document, slug: "press-release-for-test-purposes"),
       ),
     )
+    ConfigurableDocumentType.reset_document_types!
   end
 
   test "imports a document into Whitehall" do


lib/whitehall/document_importer.rb
```diff
@@ -87,6 +87,16 @@ def self.set_publishing_metadata(edition, data)
     edition.major_change_published_at = data["change_notes"].first["public_timestamp"]
   end
 
+  def self.pre_process_body(body)
+    # Content Publisher has embeds like `[Contact: c1f13fd8-9feb-4028-9323-7cb3383323b4]`.
+    # Here we find-and-replace for Whitehall's equivalent: `[Contact:171]`
+    body.gsub!(/\[Contact: ?(.+?)\]/) do |_match|
+      contact = Contact.find_by(content_id: ::Regexp.last_match(1))
+      contact ? "[Contact:#{contact.id}]" : ""
```
TonyGDS, line 114: Would it be worth logging instances when the contact is missing, incase it requires investigation?

lib/whitehall/document_importer.rb
```diff
@@ -44,7 +44,7 @@ def self.create_base_edition!(data)
         government_id: Government.find_by(content_id: data["government_id"])&.id,
         change_note: combined_change_notes(data["change_notes"]),
         lead_organisations: Organisation.where(content_id: data["tags"]["primary_publishing_organisation"]),
-        supporting_organisations: Organisation.where(content_id: data["tags"]["organisations"]),
+        supporting_organisations: Organisation.where(content_id: (data["tags"]["organisations"] - data["tags"]["primary_publishing_organisation"])),
```
TonyGDS: Again should we log when this has happened?

lib/whitehall/document_importer.rb
ChrisBAshton: I'm happy to not log it, as there's not much we can do about it. There are already a large number of missing Contact references in Whitehall, so we wouldn't really be making anything any worse, and the user-facing page already drops the Contact embed (see [commit message](https://github.com/alphagov/whitehall/pull/10635/commits/2404009f1d80a4cba4528ff87466a54684a9c82e)) so it's not like we're removing anything user-visible.

We'll keep backups of the Content Publisher DB before retiring it, so can always get it back if necessary :)

lib/whitehall/document_importer.rb
ChrisBAshton: I'm not sure there's much benefit to that. Whitehall just has a different way of modelling organisations. We're still retaining the lead/primary orgs, and the supporting orgs - Whitehall just doesn't cope with orgs that are marked as both (which apparently was fine in Content Publisher).

test/unit/lib/whitehall/document_importer_test.rb
```diff
@@ -0,0 +1,520 @@
...
+          ],
+        }
+      end
+
+      it "carries over the image 'created_at' date" do
+        Whitehall::DocumentImporter.save_images(@data, @edition)
+        assert_equal "2022-05-25 08:17:06 +0100", @edition.images.first.created_at.to_s
+      end
+
+      it "carries over the image 'alt_text'" do
```
lauraghiorghisor-tw, line 477: Is the idea that we merge this before running the alt text rake tasks then?

ChrisBAshton: Ah yes, this was written before the alt text cull workstream was a thing! I do intend to run this on Monday, and the alt text cull workstream won't be run before then - so yes 👍

PR #10638, opened 2025-09-04
[WHIT-2383] Add file attachments to news story

lauraghiorghisor-tw:
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2383)

### DIFF
--- a/app/helpers/attachments_helper.rb
+++ b/app/helpers/attachments_helper.rb
@@ -9,25 +9,15 @@ module AttachmentsHelper
     { host: Plek.website_root, protocol: "https" }
   end
 
-  def previewable?(attachment)
-    attachment.csv? && attachment.attachable.is_a?(Edition)
-  end
-
-  def preview_path_for_attachment(attachment)
-    if attachment.attachment_data.all_asset_variants_uploaded?
-      "/csv-preview/#{attachment.attachment_data.assets.first.asset_manager_id}/#{attachment.attachment_data.assets.first.filename}"
-    end
-  end
-
-  def prepare_attachments(attachments, alternative_format_contact_email)
+  def prepare_attachments(attachments)
     attachments
       .select { |attachment| !attachment.file? || attachment.attachment_data&.all_asset_variants_uploaded? }
       .map do |attachment|
-      attachment_component_params(attachment, alternative_format_contact_email:)
+      attachment_component_params(attachment)
     end
   end
 
-  def attachment_component_params(attachment, alternative_format_contact_email: nil)
+  def attachment_component_params(attachment)
     params = {
       type: ATTACHMENT_COMPONENT_TYPES.fetch(attachment.class),
       title: attachment.title,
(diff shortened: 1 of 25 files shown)


app/helpers/attachments_helper.rb
```diff
@@ -9,25 +9,21 @@ def default_url_options
     { host: Plek.website_root, protocol: "https" }
   end
 
-  def previewable?(attachment)
-    attachment.csv? && attachment.attachable.is_a?(Edition)
-  end
-
   def preview_path_for_attachment(attachment)
```
lauraghiorghisor-tw: Think this needs to go, as it's replaced by `preview_path`?

app/helpers/attachments_helper.rb
```diff
@@ -55,26 +51,26 @@ def attachment_component_params(attachment, alternative_format_contact_email: ni
     end
 
     # CSV attachments on an Edition get a "View online" preview link
-    if previewable?(attachment)
-      params[:preview_url] = preview_path_for_attachment(attachment)
+    if attachment.respond_to?(:previewable?) && attachment.previewable?
```
lauraghiorghisor-tw: Do you need to check that it responds to the method? Perhaps this is one of those cases where you'd rather have it scream if someone deletes that method.

app/helpers/attachments_helper.rb
```diff
@@ -55,26 +51,26 @@ def attachment_component_params(attachment, alternative_format_contact_email: ni
...
+      params[:preview_url] = attachment.preview_path
     end
 
     # Inaccessible attachments can have alt format contact info
-    unless attachment.accessible?
-      params[:alternative_format_contact_email] = alternative_format_contact_email
+    if attachment.respond_to?(:alternative_format_contact_email) && !attachment.accessible?
```
lauraghiorghisor-tw: Same here. This one is even rescued and returns nil so suffices to check that it's not nil.

app/views/admin/standard_editions/_form.html.erb
```diff
@@ -80,7 +80,18 @@
 
     <%= render Admin::Editions::HistoryModeFormControls.new(@edition, current_user) %>
 
-    <%= render("settings_fields", form:, edition:) %>
+    <div class="govuk-!-margin-bottom-8">
```
lauraghiorghisor-tw, line 83: within the `_accessible_attachment_field.html.erb` we check ` <% if edition.respond_to?(:alternative_format_provider) %>` so I think using the _settings partial is fine?

ryanb-gds: All standard editions will respond to `alternative_format_provider` though, because the module is included. We need to check whether file attachments are active for the edition type. I didn't want to use `alternative_format_provider_required?` for all editions though, because that's currently `false` for legacy news articles. I'm not sure it should be but I don't want to change existing behaviour.

app/helpers/attachments_helper.rb
ryanb-gds: Yes, in case of Policy Groups.

app/helpers/attachments_helper.rb
ryanb-gds: Yes, in case of Policy Groups. The fact there's a rescue on the edition trait is neither here nor there if we're not calling this in the context of an edition!

app/helpers/attachments_helper.rb
ryanb-gds: Although in this case we're dealing with attachments, not editions, so you're right there's probably no need for this. If somebody passes in an HTML attachment this should blow up

app/helpers/attachments_helper.rb
ryanb-gds: Although in this case we're dealing with attachments, not editions, so you're right there's probably no need for this. If somebody passes in an HTML attachment this should blow up

lauraghiorghisor-tw: Yeah both `previewable` and `alternative_format_contact_email`  are on the  file attachment class.

PR #10642, opened 2025-09-05
[WHIT-2490] Fix html headers with leading symbols/numbers

lauraghiorghisor-tw:
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2490)

**What**

When using auto-numbering for Govspeak headings, certain headings would produce "rogue code" in the HTML output. Specifically:

Currency symbol headings like `## £75,000 fixed payment impact` would generate invalid HTML IDs starting with numbers (e.g., `{#75-000-fixed-payment-impact}`)

This PR seeks to fix that. 

**BEFORE**
Screenshot 2025-09-08 at 14 48 30

**AFTER**
Screenshot 2025-09-08 at 16 24 44

### DIFF
--- a/app/helpers/govspeak_helper.rb
+++ b/app/helpers/govspeak_helper.rb
@@ -151,10 +151,11 @@ private
         num = "#{h2}.#{h3}"
       end
 
-      # We have to manually derive and append a slug otherwise when Govspeak
-      # generates the HTML, it includes the <span> and number in the ID. Hence
-      # the `heading_text.parameterize`
-      "#{hashes} <span class=\"number\">#{num} </span>#{heading_text} {##{heading_text.parameterize}}"
+      custom_id_match = heading_text.match(/\{#([^}]+)\}/)
+      clean_heading_text = custom_id_match ? heading_text.gsub(/\s*\{#[^}]+\}/, "") : heading_text
+      id = (custom_id_match&.[](1) || clean_heading_text).gsub(/^[^a-zA-Z]+/, "").parameterize
+
+      "#{hashes} <span class=\"number\">#{num} </span>#{clean_heading_text} {##{id}}"
     end
   end
 
--- a/test/unit/app/helpers/govspeak_helper_test.rb
+++ b/test/unit/app/helpers/govspeak_helper_test.rb
@@ -293,7 +293,7 @@ class GovspeakHelperTest < ActionView::TestCase
   it "adds numbers to h2 headings" do
     input = "# main\n\n## first\n\n## second"
     output = '<div class="govspeak"><h1 id="main">main</h1> <h2 id="first"> <span class="number">1. </span>first</h2> <h2 id="second"> <span class="number">2. </span>second</h2></div>'
-    assert_equivalent_html output, govspeak_to_html(input, heading_numbering: :auto).gsub(/\s+/, " ")
+    assert_equivalent_html output, collapse_whitespace(govspeak_to_html(input, heading_numbering: :auto))
   end
 
   it "adds sub-numbers to h3 tags" do
@@ -303,7 +303,7 @@ class GovspeakHelperTest < ActionView::TestCase
     expected_output_1b = '<h3 id="first-point-two"> <span class="number">1.2 </span>first point two</h3>'
     expected_output2 = '<h2 id="second"> <span class="number">2. </span>second</h2>'
     expected_output_2a = '<h3 id="second-point-one"> <span class="number">2.1 </span>second point one</h3>'
-    actual_output = govspeak_to_html(input, heading_numbering: :auto).gsub(/\s+/, " ")
+    actual_output = collapse_whitespace(govspeak_to_html(input, heading_numbering: :auto))
     assert_match %r{#{expected_output1}}, actual_output
     assert_match %r{#{expected_output_1a}}, actual_output
     assert_match %r{#{expected_output_1b}}, actual_output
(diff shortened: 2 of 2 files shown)


test/unit/app/helpers/govspeak_helper_test.rb
```diff
@@ -586,4 +598,8 @@ def image_data.image_kind_config = Whitehall::ImageKind.new(
   def collapse_whitespace(string)
     string.gsub(/\s+/, " ").strip
   end
+
+  def normalized_html(html)
```
lauraghiorghisor-tw: The method just above, `collapse_whitespace`, is likely good enough. There are some additional gsub usages in the test file where you can introduce it if you care to.

app/helpers/govspeak_helper.rb
```diff
@@ -138,23 +138,30 @@ def add_heading_numbers(govspeak)
       hashes = Regexp.last_match(1)
       heading_text = Regexp.last_match(2).strip
 
-      # Heading numbers aren't applied for H4s and below, so skip replacement.
       next "#{hashes} #{heading_text}" if hashes.length > 3
```
lauraghiorghisor-tw: You don't like Ryan's comment? 😅 I don't hate that comment, nor the `# "###"` one.

app/helpers/govspeak_helper.rb
```diff
@@ -138,23 +138,30 @@ def add_heading_numbers(govspeak)
...
 
+      custom_id_match = heading_text.match(/\{#([^}]+)\}/)
+      custom_id = custom_id_match&.[](1)
+      clean_heading_text = custom_id_match ? heading_text.gsub(/\s*\{#[^}]+\}/, "") : heading_text
+      clean_heading_text.force_encoding("UTF-8")
```
lauraghiorghisor-tw: Why do we need to do this? Tests pass without. If it's important, please add a test to capture it.

app/helpers/govspeak_helper.rb
```diff
@@ -138,23 +138,30 @@ def add_heading_numbers(govspeak)
       hashes = Regexp.last_match(1)
       heading_text = Regexp.last_match(2).strip
 
-      # Heading numbers aren't applied for H4s and below, so skip replacement.
       next "#{hashes} #{heading_text}" if hashes.length > 3
 
+      custom_id_match = heading_text.match(/\{#([^}]+)\}/)
+      custom_id = custom_id_match&.[](1)
```
lauraghiorghisor-tw: Ok this is the kind of thing I need a comment on (or copilot) 😅

test/unit/app/helpers/govspeak_helper_test.rb
```diff
@@ -332,6 +332,41 @@ def image_data.image_kind_config = Whitehall::ImageKind.new(
...
+
+  it "leaves heading numbers not occuring at the start of the heading text alone when using manual heading numbering" do
+    input = "## Number 8"
+    result = Nokogiri::HTML::DocumentFragment.parse(govspeak_to_html(input))
+    assert_equal "Number 8", result.css("h2").first.text
+  end
+
+  it "handles leading numbers and symbols" do
+    input = "## £100,000 header text here"
+    expected_output = '<div class="govspeak"><h2 id="heading-100-000-header-text-here"><span class="number">1. </span>£100,000 header text here</h2></div>'
```
lauraghiorghisor-tw: I don't actually like the "heading" suggestion. 

I had a vague memory of govspeak dropping leading numbers - and indeed if you check it out here https://govspeak-preview.publishing.service.gov.uk/preview/ with `## £10,000 some thing here` you will see that the generated html is 
```
<h2 id="some-thing-here">£10,000 some thing here</h2>
```

SO, perhaps the better solution is to revert the test to what it was, i.e.:
```
  it "handles/drops? leading numbers and symbols" do
    input = "## £100,000 header text here"
    expected_output = '<div class="govspeak"><h2 id="header-text-here"><span class="number">1. </span>£100,000 header text here</h2></div>'
    assert_equivalent_html expected_output, collapse_whitespace(govspeak_to_html(input, heading_numbering: :auto))
  end
```

lauraghiorghisor-tw: And then I'd also add, or change the previous test so that we also cover this: 
```
  it "can manage a custom ID with leading numbers" do
    input = "## £100,000 header text here {#10-custom-id}"
    expected_output = '<div class="govspeak"><h2 id="custom-id"><span class="number">1. </span>£100,000 header text here</h2></div>'
    assert_equivalent_html expected_output, collapse_whitespace(govspeak_to_html(input, heading_numbering: :auto))
  end
```

app/helpers/govspeak_helper.rb
```diff
@@ -138,23 +138,30 @@ def add_heading_numbers(govspeak)
...
         h2 = 1 if h2.zero?
         h3 += 1
         num = "#{h2}.#{h3}"
       end
 
-      # We have to manually derive and append a slug otherwise when Govspeak
-      # generates the HTML, it includes the <span> and number in the ID. Hence
-      # the `heading_text.parameterize`
-      "#{hashes} <span class=\"number\">#{num} </span>#{heading_text} {##{heading_text.parameterize}}"
+      if custom_id
```
lauraghiorghisor-tw: To pass the tests as per my other comment, I suggest:

```
  def add_heading_numbers(govspeak)
    h2 = 0
    h3 = 0

    govspeak.gsub(/^(##+)\s*(.+)$/) do
      hashes = Regexp.last_match(1)
      heading_text = Regexp.last_match(2).strip

      next "#{hashes} #{heading_text}" if hashes.length > 3

      if hashes == "##"
        h2 += 1
        h3 = 0
        num = "#{h2}."
      else
        h2 = 1 if h2.zero?
        h3 += 1
        num = "#{h2}.#{h3}"
      end
      
      custom_id_match = heading_text.match(/\{#([^}]+)\}/)
      clean_heading_text = custom_id_match ? heading_text.gsub(/\s*\{#[^}]+\}/, "") : heading_text
      id = (custom_id_match&.[](1) || clean_heading_text).gsub(/^[^a-zA-Z]+/, "").parameterize
      
      "#{hashes} <span class=\"number\">#{num} </span>#{clean_heading_text} {##{id}}"
  end
```

lauraghiorghisor-tw (commented):
Nice one, Alex.

But I think introducing the header- thing is unnecessary. Requesting changes.
See comments.

app/helpers/govspeak_helper.rb
GDSNewt: I actually like comments but I think there is a general crusade against comments, in this case I think I just searched for hash and deleted everything 😅

lauraghiorghisor-tw (commented):
LGTM now 👍

@ChrisBAshton if you'd like to have a look as it touches/fixes code you changed recently.

GDSNewt (approved):

PR #10647, opened 2025-09-09
Simplify asset upload check to verify only the original asset (WHIT-2440)

ChrisBAshton:
This PR replaces `all_asset_variants_uploaded?` with a simpler `asset_uploaded?` method for both images and attachments. Instead of requiring all variant files (e.g. `s960`, `s300`, etc.) to be present before allowing publishing or previewing, we now consider the asset “ready” if the original file has successfully uploaded.

In practice, all variants are generated and uploaded within milliseconds of the original upload, and failures nearly always stem from the original file itself. For example, if Asset Manager has an outage or detects a virus in the uploaded asset, then the original variant will fail to successfully upload, just as the variants will not. If the original variant has successfully uploaded, then we can be highly confident (though can't _guarantee_) that its variants will have to. The minimal risk hypothetical edge case is a risk we can live with to allow for the following benefits:

* Supports content with partial variant sets (we will be importing news articles from Content Publisher, which have only the `s960` and `s300` variants).
* Enables future support for dynamically generated variants (e.g. via Fastly image resizing) without requiring full variant pre-generation.
* Greatly simplifies the codebase and removes brittle assumptions.

See https://github.com/alphagov/content-publisher/pull/3311#discussion_r2329354424 for a discussion on the different variant sizes and where they're used.

As you can see in this commit, there is a lot of room for improvement and consolidation across Whitehall. We have several implementations of `asset_uploaded?` that are identical across modules. In practice, most of these implementations were already only checking for the 'original' variant, and have now been simplified to make that absolutely explicit. We should revisit this commit later to consider how to extract some of this behaviour out into a shared module so that we're not duplicating identical logic in multiple places.

Supersedes #10646 
Blocks #10635

---


GDSNewt (approved):
Main comment would be on the potential to DRY things up, but you've acknowledge we can do that in a subsequent batch of work. I'll let Laura comment on the more Assety intricacies 👍

GDSNewt:
Failing tests in CI but look easily fixable _**"all_asset_variants_uploaded?' for an instance of AttachmentData"**_

ryanb-gds:
Whilst we can be highly confident that the resized assets will upload if the original does because kubernetes networking over an AWS network is pretty reliable, we're still falling foul of [distributed system fallacy number 1](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing). But maybe we don't care that much if one or more fail, I'm not really sure what the impact would be. Maybe the smallish risk is worth the simplification of our code

ChrisBAshton:
> Whilst we can be highly confident that the resized assets will upload if the original does because kubernetes networking over an AWS network is pretty reliable, we're still falling foul of [distributed system fallacy number 1](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing). But maybe we don't care that much if one or more fail, I'm not really sure what the impact would be. Maybe the smallish risk is worth the simplification of our code

I personally think it's worth the risk, especially if asset variants are going to be revisited longer term (Fastly image optimisations, [this tech debt card](https://trello.com/c/oh4MXlUA/52-publishing-apps-generate-their-own-image-size-variations), possible replacements to Asset Manager, [RFC-181](https://github.com/alphagov/govuk-rfcs/pull/181)). To me this feels like a pragmatic step in the right direction in the meantime. But worth a discussion - I'd say at "sit-down" but the next one we're both available for is next week. I may put a short meeting in with the team for today.

app/components/admin/organisations/show/summary_list_component.rb
```diff
@@ -88,7 +88,7 @@ def default_news_image_row
 
     {
       field: "Default news image",
-      value: organisation.default_news_image.all_asset_variants_uploaded? ? image_tag(organisation.default_news_image.file.url(:s300)) : tag.span("Processing", class: "govuk-tag govuk-tag--green"),
+      value: organisation.default_news_image.asset_uploaded? ? image_tag(organisation.default_news_image.file.url(:s300)) : tag.span("Processing", class: "govuk-tag govuk-tag--green"),
```
lauraghiorghisor-tw, line 91: Organisations and other image-using, non-editionable models, are particularly annoying because the changes go live immediately.

I think this would show a broken image to the user if the s300 asset is not available yet (rather than showing the processing tag). And then it will subsequently show as some form of broken img on the FE.

Current behaviour:
Screenshot 2025-09-10 at 10 01 39

app/helpers/attachments_helper.rb
```diff
@@ -51,7 +51,7 @@ def attachment_component_params(attachment)
 
   def block_attachments(attachments = [])
     attachments
-      .select { |attachment| !attachment.file? || attachment.attachment_data.all_asset_variants_uploaded? }
+      .select { |attachment| !attachment.file? || attachment.attachment_data.asset_uploaded? }
```
lauraghiorghisor-tw, line 54: This one's fine as the `attachment_component_params` uses the original version - `attachment.url`.

Nice to see in one place all these usages, where we check for all variants only to then use a generic url call with no version - basically all attachments, most embedded govspeak images.

app/helpers/presenters/publishing_api/default_news_image_helper.rb
```diff
@@ -2,7 +2,7 @@ module Presenters::PublishingApi::DefaultNewsImageHelper
 private
 
   def present_default_news_image(item)
-    return unless item.default_news_image && item.default_news_image.all_asset_variants_uploaded?
+    return unless item.default_news_image && item.default_news_image.asset_uploaded?
```
lauraghiorghisor-tw, line 5: The biggest downside is here, and other presenters' level. This default news image helper is used in org and ww-org presenters.

We'd now check on the original version but then immediately proceed to use the 300 version.
When this is not available the behaviour is rather ugly - I don't recall exactly what the url method does in this case, but we've seen this with the Person image on occasion, the "same file name upload" error, where some tmp path ends up being returned, resulting in a broken img on the FE.

**Action**: Can we check what the behaviour is when the variant is not available/ write a test?
**Action**: Also on this one, we also specifically send the :s960 as high-res, but we could send original instead? (which is required to be 960).

app/helpers/presenters/publishing_api/featured_documents_helper.rb
```diff
@@ -4,7 +4,7 @@ module FeaturedDocumentsHelper
       def featured_documents(featurable_item, document_limit)
         featurable_item
           .feature_list_for_locale(I18n.locale).current.limit(document_limit)
-          .select { |feature| feature.image&.all_asset_variants_uploaded? }
+          .select { |feature| feature.image&.asset_uploaded? }
```
lauraghiorghisor-tw, line 7: Again this image is important for the featured documents, having it broken on the FE is not ideal.

app/helpers/presenters/publishing_api/default_news_image_helper.rb
lauraghiorghisor-tw: I dug it out - https://github.com/carrierwaveuploader/carrierwave/blob/a51a97071fec5cf4a0b693b8271dd4c59d6b80ab/lib/carrierwave/uploader/url.rb#L17

If I am reading that somewhat right, when the call to url(:s_something) returns nil, then we end up involving path - a method we have here https://github.com/alphagov/whitehall/blob/85524666caaf34369c4cadf15a88f9aaca8959c0/lib/whitehall/asset_manager_storage.rb#L37 - which we could not remove because it was needed elsewhere. 

Meaning, the url(:s__) call will return some gibberish path which gets sent to pub-api which shows as a broken img on the FE. Which is why we opted for sending nothing if the variants are not available, be it no image, or for example no feature document if the image is absolutely important for that rendering, and showing these processing tags everywhere.

Behaviour covered in test here (though it is not exactly desirable...) - https://github.com/alphagov/whitehall/blob/85524666caaf34369c4cadf15a88f9aaca8959c0/test/unit/lib/whitehall/asset_manager_storage_test.rb#L64 

I wonder if we can, instead, keep the change you have here and  add an additional check, in the places where we need a non-original variant. [...continued in next comment]

app/presenters/publishing_api/person_presenter.rb
```diff
@@ -36,7 +36,7 @@ def links
     def details
       details_hash = {}
 
-      if item.image&.all_asset_variants_uploaded? && item.image.url(:s465)
+      if item.image&.asset_uploaded? && item.image.url(:s465)
```
lauraghiorghisor-tw, line 39: [...] continued from previous comment.

We could do smth like this maybe where we check the asset_uploaded and then the shape of the returned url, a bit like what we did in the logger logic below. 

So in all the places outlined here:
- s712 is used on [Featured documents](https://github.com/alphagov/whitehall/blob/e3cf4b69965b4b43924a3ad296c2d1af9424d99d/app/helpers/presenters/publishing_api/featured_documents_helper.rb#L64) (as its "high resolution" equivalent. Other formats [seem to use s960](https://github.com/alphagov/whitehall/blob/57b3047635af936adce493ed964949a7cb380baf/app/helpers/presenters/publishing_api/default_news_image_helper.rb#L9) for that.
- s630 is used on [Topical Events](https://github.com/alphagov/whitehall/blob/43f9817018a56dc63bd3f98ea00bbb437ae7942b/app/presenters/publishing_api/topical_event_presenter.rb#L56) as its "medium resolution" format.
- s465 is used on [Person](https://github.com/alphagov/whitehall/blob/a6975babadb1eacc66d96a0c7691dc19c6a024aa/app/presenters/publishing_api/person_presenter.rb#L39-L41), [Featured documents](https://github.com/alphagov/whitehall/blob/e3cf4b69965b4b43924a3ad296c2d1af9424d99d/app/helpers/presenters/publishing_api/featured_documents_helper.rb#L63) ("medium resolution") and [Topical Events](https://github.com/alphagov/whitehall/blob/43f9817018a56dc63bd3f98ea00bbb437ae7942b/app/presenters/publishing_api/topical_event_presenter.rb#L73)

we will do if/unless `item.image&.asset_uploaded? && item.image.url(:s_relevant_variant) && !item.image.url(:s_relevant_variant).include?("carrierwave-tmp")`.

lauraghiorghisor-tw (changes requested):
My take is that this will cause temporary carrierwave paths to be returned from the url(:s_variant) calls, and those will get sent to pub-api, and show as broken images on the FE. 

I suggested an alternative fix in a comment below, i.e. to check explicitly for non-tmp-like `url` method call return values in the specific places where we need them.

Alternatively, park this until we have removed the need for versions, and then apply the changes, unless that's a blocker for the Content Publisher import?

ChrisBAshton:
Thanks all! Discussed in meeting today. The summary was that we need to also support at least the s712 variant, ideally, as that is what is used when 'featuring' a document, and if someone wanted to go and feature one of these old news articles (however unlikely), they wouldn't be able to.

That downside, alongside the other risks highlighted, mean this is probably a no-go for now. I'm going to attempt to reupload and re-generate variants of all of these images as the point of import (though no need to do the same for attachments, which have no variants).

lauraghiorghisor-tw:
> though no need to do the same for attachments, which have no variants.

I'd probably do it for attachments too, in case at any point someone will need to do `<unknown future thing>` and find unexpected discrepancies in some random News Articles 🙈

lauraghiorghisor-tw:
> The summary was that we need to also support at least the s712 variant, ideally, as that is what is used when 'featuring' a document, and if someone wanted to go and feature one of these old news articles (however unlikely), they wouldn't be able to.

Sorry, delayed braining on this one, but actually the two things should not be related. When we feature a document, the image used is a fresh upload: https://whitehall-admin.integration.publishing.service.gov.uk/government/admin/feature_lists/1170/features/new?edition_id=1714651. It has nothing to do with the original News Article image. Unless there's a different feature place you had in mind?

That said, I prefer the "trigger the carrierwave flow" approach, but if that's a pain, I think we're safe to do this if we can address the problematic `url` returns/concurrency risks.

PR #10651, opened 2025-09-11
Add confirmation step to election rake tasks [WHIT-1876]

kashifatcha:
Tasks include:
- remove_mp_letters
- republish_political_content
- end_ministerial_appointments
- identify_political_content_for

I have added a confirmation step, using the Thor gem, to the above rake tasks. This is to ensure that the person running these tasks explicitly opts in to making these changes. Also corrected description for "identify_political_content_for" and updated test file name.

This is part of an ongoing task to add confirmation steps to our dangerous rake tasks following our rake task audit.

### DIFF
--- a/lib/tasks/election.rake
+++ b/lib/tasks/election.rake
@@ -1,16 +1,41 @@
+require "thor"
+
+def shell
+  @shell ||= Thor::Shell::Basic.new
+end
+
 namespace :election do
   desc "Remove MP from MP's letters"
   task remove_mp_letters: :environment do
     puts "Removing MP from MP's letters:"
-    Person.where('letters LIKE "%MP%"').find_each do |person|
+    query = Person.where('letters LIKE "%MP%"')
+    people_to_change = []
+
+    puts "----------DRY RUN----------"
+    query.find_each do |person|
       new_letters = person.letters.split(" ").reject { _1 == "MP" }.join(" ")
       if person.letters == new_letters
-        puts "skipped #{person.name} - includes MP (case insensitive), but doesn't match exactly"
+        puts "Skipped #{person.name} - includes 'MP' (case insensitive), but doesn't match exactly"
       else
-        old_name = person.name
-        person.update!(letters: new_letters)
-        puts "updated #{old_name} to #{person.name}"
+        puts "Found #{person.name} that matches 'MP'"
+        people_to_change << {
+          id: person.id,
+          old_name: person.name,
+          new_letters: new_letters,
+        }
       end
+    end
+
+    unless shell.yes?("Proceed with the above changes? (yes/no)")
+      shell.say_error "Aborted"
+      next
+    end
+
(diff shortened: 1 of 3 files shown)


test/unit/lib/tasks/election_mark_organisation_documents_as_political_test.rb
```diff
@@ -22,6 +26,7 @@ class IdentifyPoliticalContentFor < ActiveSupport::TestCase
     end
 
     it "marks eligible published editions of documents first published after the specified date and any associated drafts as political" do
+      # Thor::Shell::Basic.any_instance.stubs(:yes?).returns(true)
```
TonyGDS: Probably want to remove this comment @kashifatcha

kashifatcha: Whoops, done

lib/tasks/election.rake
```diff
@@ -8,6 +8,11 @@ namespace :election do
         puts "skipped #{person.name} - includes MP (case insensitive), but doesn't match exactly"
       else
         old_name = person.name
+        puts "You are about to change '#{old_name}' current letters '#{person.letters}' to '#{new_letters}"
+        unless Thor::Shell::Basic.new.yes?("Proceed with this change? (yes/no)")
```
ryanb-gds: I suspect there isn't much value in getting the user to confirm every instance of these changes, perhaps it would be best to get a count of the people, ask for confirmation, and then enter the loop to process them?

ryanb-gds (approved):
This is fine but there are quite a lot of MPs, so might be worth reworking the task slightly to avoid having to confirm them individually.

lib/tasks/election.rake
```diff
@@ -1,7 +1,20 @@
...
+
+def shell
+  @shell ||= Thor::Shell::Basic.new
+end
+
 namespace :election do
   desc "Remove MP from MP's letters"
   task remove_mp_letters: :environment do
     puts "Removing MP from MP's letters:"
+    people_count = Person.where('letters LIKE "MP %" OR letters LIKE "% MP %" OR letters LIKE "% MP" OR letters = "MP"').count
```
ryanb-gds: We should use the same query as below to make sure the numbers are correct. Also you might want to have a careful look into how the LIKE clause works in MySQL, because you don't need most of the query above.

You can build the query as follows:

```
query = Person.where('letters LIKE "%MP%"')
```
and then do `query.count` and `query.find_each`

ryanb-gds (changes requested):
Needs a bit of tidying up, but I think this is a better direction

ryanb-gds (approved):
Ready to go I think

lib/tasks/election.rake
lauraghiorghisor-tw: Added a single confirmation after going through all people 👍

PR #10654, opened 2025-09-11
CM-425: Obtain content block preview from publishing api

edavey:
See [Jira CM-425](https://gov-uk.atlassian.net/browse/CM-425)

The `GovspeakHelper` used by `Admin::PreviewController#preview` depended on the `FindAndReplaceEmbedCodesService` within the `ContentBlockManager` Rails engine in order to render content blocks as "previews". This is no longer possible as Content Block Manager is moving out of Whitehall into its own application. See [ADR 3][] and the [new Content Block Manager repo][].

This PR removes that dependency by:

- moving the `FindAndReplaceEmbedCodesService` out of the CBM engine and into Whitehall proper
- changing the behaviour of `FindAndReplaceEmbedCodesService` so that it now fetches content blocks from the Publishing API rather than from Whitehall's database. (This required a [change to the Publishing API][]'s `GET /v2/get_content)` endpoint to allow editions to be found by `content_id_alias`.

[ADR 3]:
https://github.com/alphagov/whitehall/blob/main/docs/content_block_manager/adr/0003-move-out-to-new-application.md

[new Content Block Manager repo]:
https://github.com/alphagov/content-block-manager

[change to the Publishing API]:
https://github.com/alphagov/publishing-api/commit/135d3dc34dcf4bd9525c5bf0802a16b50cd69c59

### DIFF
--- a/app/services/content_block/find_and_replace_embed_codes_service.rb
+++ b/app/services/content_block/find_and_replace_embed_codes_service.rb
@@ -1,4 +1,4 @@
-module ContentBlockManager
+module ContentBlock
   class FindAndReplaceEmbedCodesService
     def self.call(html)
       new(html).call
--- a/app/services/content_block/find_and_replace_embed_codes_service.rb
+++ b/app/services/content_block/find_and_replace_embed_codes_service.rb
@@ -6,14 +6,14 @@ module ContentBlock
 
     def call
       embed_content_references.uniq.each do |reference|
-        content_block = content_blocks.find do |c|
-          c.fetch("content_id_aliases").first.fetch("name") == reference.identifier
+        content_item = content_items.find do |item|
+          item.content_id_alias == reference.identifier
         end
-        next if content_block.nil?
+        next if content_item.nil?
 
         html.gsub!(
           reference.embed_code,
-          render_block(content_block: content_block, embed_code: reference.embed_code),
+          render_block(content_item: content_item, embed_code: reference.embed_code),
         )
       end
 
@@ -36,8 +36,10 @@ module ContentBlock
       embed_content_references.uniq.map(&:identifier)
     end
 
-    def content_blocks
-      @content_blocks ||= get_content_items_from_publishing_api
+    def content_items
+      @content_items ||= get_content_items_from_publishing_api.map do |result|
+        PublishingApiContentItem.new(result)
+      end
     end
 
     def get_content_items_from_publishing_api
@@ -50,14 +52,25 @@ module ContentBlock
         )["results"]
     end
 
-    def render_block(content_block:, embed_code:)
+    def render_block(content_item:, embed_code:)
       ContentBlockTools::ContentBlock.new(
-        document_type: content_block.fetch("document_type"),
-        content_id: content_block.fetch("content_id"),
-        title: content_block.fetch("title"),
-        details: content_block.fetch("details"),
+        document_type: content_item.document_type,
+        content_id: content_item.content_id,
+        title: content_item.title,
+        details: content_item.details,
         embed_code: embed_code,
       ).render
+    end
+
+    class PublishingApiContentItem
+      def initialize(result_hash)
+        @document_type = result_hash.fetch("document_type")
+        @content_id = result_hash.fetch("content_id")
+        @title = result_hash.fetch("title")
+        @details = result_hash.fetch("details")
+        @content_id_alias = result_hash.fetch("content_id_aliases").first.fetch("name")
+      end
+      attr_reader :document_type, :content_id, :title, :details, :content_id_alias
     end
   end
 end
--- a/app/services/content_block/find_and_replace_embed_codes_service.rb
+++ b/app/services/content_block/find_and_replace_embed_codes_service.rb
@@ -5,16 +5,8 @@ module ContentBlock
     end
 
     def call
-      embed_content_references.uniq.each do |reference|
-        content_item = content_items.find do |item|
-          item.content_id_alias == reference.identifier
-        end
-        next if content_item.nil?
-
-        html.gsub!(
-          reference.embed_code,
-          render_block(content_item: content_item, embed_code: reference.embed_code),
-        )
+      unique_embedded_content_references.each do |reference|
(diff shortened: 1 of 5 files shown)


app/services/content_block/find_and_replace_embed_codes_service.rb
```diff
@@ -7,11 +7,11 @@ def self.call(html)
     def call
       embed_content_references.uniq.each do |reference|
         content_block = content_blocks.find do |c|
-          c.document.content_id == reference.identifier || c.document.content_id_alias == reference.identifier
+          c.fetch("content_id_aliases").first.fetch("name") == reference.identifier
```
pezholio: I know we will _usually_ only have one `content_id_alias`, but I wonder if it's more futureproof to do an `includes` here? Something like:

```ruby
c.fetch("content_id_aliases").map { |alias| alias["name"] }.includes?(reference.identifier)
```

app/services/content_block/find_and_replace_embed_codes_service.rb
```diff
@@ -29,16 +29,35 @@ def embed_content_references
...
+        .fetch("results")
+    end
+
+    def render_block(content_block)
+      ContentBlockTools::ContentBlock.new(
+        document_type: content_block.fetch("document_type"),
+        content_id: content_block.fetch("content_id"),
+        title: content_block.fetch("title"),
+        details: content_block.fetch("details"),
+        embed_code: content_block.fetch("content_id_aliases").first.fetch("name"),
```
pezholio: `embed_code` here should come from the reference, so we know what exactly to render. For example, if the embed code is:

```
{{embed:content_block_contact:my-contact/addresses/bar}}
```

Content Block Tools knows to only render the `address` `bar`.

app/services/content_block/find_and_replace_embed_codes_service.rb
```diff
@@ -50,14 +50,25 @@ def get_content_items_from_publishing_api
...
-        embed_code: content_block.fetch("content_id_aliases").first.fetch("name"),
+        document_type: content_item.document_type,
+        content_id: content_item.content_id,
+        title: content_item.title,
+        details: content_item.details,
+        embed_code: content_item.embed_code,
       ).render
     end
+
+    class PublishingApiContentItem
```
pezholio, line 70: Nice!

app/services/content_block/find_and_replace_embed_codes_service.rb
edavey: thks. have you got an example scenario where we would have more than one `content_id_alias`? I'm curious to know why we need to have multiple ones? It's more than just having the a "friendly" human-readable "slug"?

pezholio: We don't yet, but the way the data is structured means there might be a case where that does happen. Just thinking in terms of future proofing.

app/services/content_block/find_and_replace_embed_codes_service.rb
edavey: ok, thanks, I think I understand now!

app/services/content_block/find_and_replace_embed_codes_service.rb
edavey: OK, thks, I've added support for this.

edavey:
@pezholio this is ready for another look

app/services/content_block/find_and_replace_embed_codes_service.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def get_content_items_from_publishing_api
+      Services
+        .publishing_api
+        .get_content_items(
+          content_id_aliases: identifiers,
+          fields: %w[title content_id_aliases details document_type],
+          states: %w[published],
+        )
+        .fetch("results")
```
pezholio: Just tried this locally and I get the error:

```
NoMethodError - undefined method `fetch' for an instance of GdsApi::Response:
  app/services/content_block/find_and_replace_embed_codes_service.rb:46:in `get_content_items_from_publishing_api'
  app/services/content_block/find_and_replace_embed_codes_service.rb:33:in `content_items'
   app/services/content_block/find_and_replace_embed_codes_service.rb:50:in `replace_embed_code_with_rendered_block_for'
  app/services/content_block/find_and_replace_embed_codes_service.rb:9:in `block in call'
   app/services/content_block/find_and_replace_embed_codes_service.rb:8:in `call'
   app/services/content_block/find_and_replace_embed_codes_service.rb:4:in `call'
   app/helpers/govspeak_helper.rb:85:in `preprocess_govspeak'
   app/helpers/govspeak_helper.rb:11:in `govspeak_to_html'
   app/views/admin/preview/preview.html.erb:4
   app/controllers/admin/preview_controller.rb:10:in `preview'
```

I guess this is because `GdsApi::Response` acts "like" a hash, but isn't actually a hash. Replacing it with `["results"]` seems to work.

app/services/content_block/find_and_replace_embed_codes_service.rb
```diff
@@ -0,0 +1,82 @@
...
+        PublishingApiContentItem.new(result)
+      end
+    end
+
+    def get_content_items_from_publishing_api
+      Services
+        .publishing_api
+        .get_content_items(
+          content_id_aliases: identifiers,
+          fields: %w[title content_id_aliases details document_type],
```
pezholio: `content_id` seems to be missing from the `fields` here - got an error when initialising the `PublishingApiContentItem`, but adding `content_id` like so fixed it:

```
          fields: %w[title content_id content_id_aliases details document_type],
```

app/services/content_block/find_and_replace_embed_codes_service.rb
edavey: thanks, I've addressed this now.

app/services/content_block/find_and_replace_embed_codes_service.rb
edavey: 🤦 thanks, this is fixed now.

ChrisBAshton (approved):
LGTM!

PR #10665, opened 2025-09-16
[WHIT-2574] Add lead image behaviour to config driven news stories

lauraghiorghisor-tw:
## What and why

Add full legacy behaviour for lead image. We have previously introduced the image "config driven block", but without some of the behaviours we had before. This is now porting over existing behaviour. 
We decided to create a new block `lead_image_select` rather than try to fit all this logic in the `image_select`.

For news article lead image the behaviour was:
- Custom images can be uploaded in the Images tab. A custom image is not required. 
- Guidance on lead images appears in the Images tab.
- Defaults to org “default news image” if no img is uploaded (actually it defaults to lead organisation, then organisation, then worldwide organisation, in this order, based on whether or not they have an image set) -> the “default lead image” behaviour.
- If there is one custom img uploaded, it is promoted to lead.
- If multiple images are uploaded you can set any as lead images -> the “custom lead image” behaviour.
- Any subsequent images after the lead one, get the option to be embedded in govspeak.
- If nothing is sent for a document type that should have a lead image, a placeholder is provided by the [frontend rendering app](https://github.com/alphagov/frontend/blob/057edd16b90b58724981fc207b29c67089fe91f0/app/models/concerns/news_image.rb#L15).
- Validations: you cannot use the lead image in the body as an embedded image as well; you cannot use an svg as a lead image.

Note: frontend also has duplicated logic for falling back to organisation image. I originally wanted to move more logic into there, but am now inclined to keep things wh-side as "source of truth", and potentially clean up FE logic once we standardise the lead image behaviour wh-side for news articles, case studies and speeches.

New behaviour:
- We still upload the images in the Images tab. 
- Move the "set image as lead" to the edition form, in the shape of an image select block.
- There is no automatic promotion of image to lead. The selection of the image is now explicit.
- Keep svg restriction.
- Drop the validation "you cannot use the lead image in the body as an embedded image as well" since it should be very visible now what images are in use. 
- Show the default org image or placeholder, alongside the select for the custom image.
- Hide the default image if a custom image selection has been made.
- Show custom hint text for the default image if it shows the organisation image, or the placeholder, respectively.
- Show custom hint text for the custom lead image if the default is the organisation image, or the placeholder, respectively.
- We are reintroducing the placeholder image removed in b7be3164 because we want to render it to the users. This was not the case before.

Additional changes to the lead/image select block:
- Check for the presence of assets on the selected image (we do not want to send temp files to pub api) - this logic existed prior.
- Do not send caption if nil - as it's not a required field.

Previously, logic for lead image lived in:
- `LeadImage`
- `CustomLeadImage`
Now it is mostly contained to the `DefaultLeadImage` concern and utility functions for composing the payload and rendering are contained within the `lead_image_select` block. The code is much reduced.

## Testing

| Publisher side |
| --- |
| Showing org default news image on edition form |
| Screenshot 2025-10-17 at 17 37 25 |
| Showing placeholder image on edition form |
| Screenshot 2025-10-17 at 17 36 56 |
| Hide default placeholder if custom selection has been made |
| Screenshot 2025-10-17 at 17 38 17 |
| Hide default org image if custom selection has been made |
| Screenshot 2025-10-17 at 17 38 37 |

| Before / After |
| --- |
| Placeholder when lead org 1 has no default image |
| t1 lead org 1 with no image
| Lead 2 org has a default image |
| Lead 2 has an image |
| Supporting org has default image |
| Supporting org has lead |
| No svg allowed - it is now filtered out in the select |
| no svg |
| Changed: validation removed for no lead image in body |
| Different no validation for repeat image |
| now allowed to do this |
| Translations - can now show/edit a different lead image for translations |
| Screenshot 2025-10-22 at 15 09 50 |
| Screenshot 2025-10-22 at 15 14 38 |

| Other integration tests |
| --- |
| News story can be featured |
| image |
| News story appears in Latest News |
| image |

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2574) and originally on  [this card](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2384)

### DIFF
--- a/test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
+++ b/test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
@@ -34,12 +34,10 @@ class ConfigurableContentBlocks::LeadImageSelectTest < ActiveSupport::TestCase
 
   test "does not have a publishing api payload if selected image's assets are not ready" do
     images = create_list(:image, 3)
-    page = StandardEdition.new
-    page.images = images
     images[1].image_data.assets = []
     images[1].image_data.save!
 
-    payload = ConfigurableContentBlocks::LeadImageSelect.new(page.images).publishing_api_payload(images[1].image_data.id)
+    payload = ConfigurableContentBlocks::LeadImageSelect.new(images).publishing_api_payload(images[1].image_data.id)
 
     assert_nil payload
   end
@@ -58,23 +56,19 @@ class ConfigurableContentBlocks::LeadImageSelectRenderingTest < ActionView::Test
         },
       },
     }
-    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { schema: }))
-
-    page = StandardEdition.new
-    page.configurable_document_type = "test_type"
-    page.images = create_list(:image, 3)
-    page.block_content = { "test_attribute" => page.images.last.image_data.id.to_s }
-    block = ConfigurableContentBlocks::LeadImageSelect.new(page.images)
+    images = create_list(:image, 3)
+    block_content = { "test_attribute" => images.last.image_data.id.to_s }
+    block = ConfigurableContentBlocks::LeadImageSelect.new(images)
 
     render block, {
       schema: schema["properties"]["test_attribute"],
-      content: page.block_content["test_attribute"],
+      content: block_content["test_attribute"],
       path: Path.new.push("test_attribute"),
     }
 
     assert_dom "select[name=?]", "edition[block_content][test_attribute]"
-    assert_dom "option[selected]", text: page.images.last.filename
-    page.images.each do |image|
+    assert_dom "option[selected]", text: images.last.filename
+    images.each do |image|
       assert_dom "option", text: image.filename
     end
   end
@@ -91,27 +85,24 @@ class ConfigurableContentBlocks::LeadImageSelectRenderingTest < ActionView::Test
         },
       },
     }
-    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { schema: }))
-
-    page = StandardEdition.new
-    page.configurable_document_type = "test_type"
-    page.images = create_list(:image, 3)
-    page.block_content = { "test_attribute" => page.images.last.image_data.id.to_s }
-    block = ConfigurableContentBlocks::LeadImageSelect.new(page.images)
+    images = create_list(:image, 3)
+    block_content = { "test_attribute" => images.last.image_data.id.to_s }
+    block = ConfigurableContentBlocks::LeadImageSelect.new(images)
 
     render block, {
       schema: schema["properties"]["test_attribute"],
-      content: page.block_content["test_attribute"],
-      translated_content: page.images.first.image_data.id,
+      content: block_content["test_attribute"],
+      translated_content: images.first.image_data.id,
       path: Path.new.push("test_attribute"),
     }
 
     assert_dom "select[name=?]", "edition[block_content][test_attribute]"
-    assert_dom "option[selected][value=?]", page.images.first.image_data.id
+    assert_dom "option[selected][value=?]", images.first.image_data.id
   end
 
   test "it renders any validation errors when they are present" do
     schema = {
+      "title" => "Test object",
       "type" => "object",
       "properties" => {
         "test_attribute" => {
@@ -127,22 +118,22 @@ class ConfigurableContentBlocks::LeadImageSelectRenderingTest < ActionView::Test
         },
       },
     }
-    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { "schema" => schema }))
-
-    page = StandardEdition.new
-    page.configurable_document_type = "test_type"
-    page.images = create_list(:image, 3)
-    page.block_content = { "test_attribute" => nil }
-    page.validate
-    block = ConfigurableContentBlocks::LeadImageSelect.new(page.images)
+    errors = [mock("object"), mock("object")]
+    messages = %w[foo bar]
+    errors.each_with_index do |error, index|
+      error.expects(:attribute).returns(:test_attribute)
+      error.expects(:full_message).returns(messages[index])
+    end
+    block = ConfigurableContentBlocks::LeadImageSelect.new([create(:image)])
 
     render block, {
-      schema: schema["properties"]["test_attribute"],
-      content: page.block_content["test_attribute"],
+      schema:,
+      content: nil,
       path: Path.new.push("test_attribute"),
-      errors: page.errors,
+      errors:,
     }
-    assert_dom ".govuk-error-message", "Error: #{page.errors.where(:test_attribute).map(&:full_message).join}"
+
+    assert_dom ".govuk-error-message", "Error: #{messages.join}"
   end
 
   test "it renders the default lead image, if no custom lead image has been selected" do
@@ -157,7 +148,6 @@ class ConfigurableContentBlocks::LeadImageSelectRenderingTest < ActionView::Test
         },
       },
     }
-    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { schema: }))
     default_image = build(:organisation, :with_default_news_image).default_news_image
     block = ConfigurableContentBlocks::LeadImageSelect.new([], default_lead_image: default_image)
 
@@ -183,7 +173,6 @@ class ConfigurableContentBlocks::LeadImageSelectRenderingTest < ActionView::Test
         },
       },
     }
-    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { schema: }))
     block = ConfigurableContentBlocks::LeadImageSelect.new([], default_lead_image: nil)
 
     render block, {
@@ -208,21 +197,18 @@ class ConfigurableContentBlocks::LeadImageSelectRenderingTest < ActionView::Test
         },
       },
     }
-    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { schema: }))
-    page = StandardEdition.new
-    page.configurable_document_type = "test_type"
     default_image = build(:organisation, :with_default_news_image).default_news_image
-    page.images = create_list(:image, 2)
-    page.block_content = { "test_attribute" => page.images.last.image_data.id.to_s }
-    block = ConfigurableContentBlocks::LeadImageSelect.new(page.images, default_lead_image: default_image)
+    images = create_list(:image, 2)
+    block_content = { "test_attribute" => images.last.image_data.id.to_s }
+    block = ConfigurableContentBlocks::LeadImageSelect.new(images, default_lead_image: default_image)
 
     render block, {
       schema: schema["properties"]["test_attribute"],
-      content: page.block_content["test_attribute"],
+      content: block_content["test_attribute"],
       path: Path.new.push("test_attribute"),
     }
 
-    assert "a", text: page.images.last.url
+    assert "a", text: images.last.url
     assert_dom "h2", text: "Default lead image", count: 0
     assert "a", text: default_image.url, count: 0
   end
--- a/test/unit/app/models/configurable_content_blocks/image_select_test.rb
+++ b/test/unit/app/models/configurable_content_blocks/image_select_test.rb
@@ -22,12 +22,10 @@ class ConfigurableContentBlocks::ImageSelectTest < ActiveSupport::TestCase
 
   test "does not have a publishing api payload if selected image's assets are not ready" do
     images = create_list(:image, 3)
-    page = StandardEdition.new
-    page.images = images
     images[1].image_data.assets = []
     images[1].image_data.save!
 
-    payload = ConfigurableContentBlocks::ImageSelect.new(page.images).publishing_api_payload(images[1].image_data.id)
+    payload = ConfigurableContentBlocks::ImageSelect.new(images).publishing_api_payload(images[1].image_data.id)
 
     assert_nil payload
   end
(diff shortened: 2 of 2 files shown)


app/models/concerns/standard_edition/default_lead_image.rb
```diff
@@ -0,0 +1,17 @@
+module StandardEdition::DefaultLeadImage
+  extend ActiveSupport::Concern
+
+  def default_lead_image
+    if lead_organisations.any? && lead_organisations.first.default_news_image
```
ryanb-gds: On reflection, I think it is dangerous to assume that the edition will definitely respond to lead_organisations. Or perhaps this module should include the `Organisations` module just in case it isn't included already

app/models/configurable_content_blocks/image_select.rb
```diff
@@ -23,5 +23,12 @@ def publishing_api_payload(content)
     def to_partial_path
       "admin/configurable_content_blocks/image_select"
     end
+
+  private
+
+    def lead_image_caption(image)
```
ryanb-gds: This doesn't seem as though it should be in the image_select block?

app/models/configurable_content_blocks/lead_image_select.rb
```diff
@@ -0,0 +1,65 @@
...
+      @default_lead_image = default_lead_image
+    end
+
+    def json_schema_type
+      "string"
+    end
+
+    def json_schema_format
+      "lead_image_select"
+    end
```
ryanb-gds: Drop these, they're no longer part of the block interface since dropping JSON schema validation

app/models/configurable_content_blocks/lead_image_select.rb
```diff
@@ -0,0 +1,65 @@
...
+      selected_image = images.find { |image| image.image_data.id == content.to_i } if content.present?
+      if selected_image&.image_data&.all_asset_variants_uploaded?
+        return {
+          high_resolution_url: selected_image.image_data.file.url(:s960),
+          url: selected_image.image_data.file.url(:s300),
+          caption: lead_image_caption(selected_image),
+        }.compact
+      end
+
+      Rails.logger.warn("The page does not have an image with ID #{content}, so the image has been excluded from the Publishing API payload.") unless selected_image
```
ryanb-gds: I think we want to log this if `content.present?` and not `selected_image`, because otherwise we'll be logging this even if the user has deliberately chosen the fallback, won't we?

app/models/configurable_content_blocks/lead_image_select.rb
```diff
@@ -0,0 +1,65 @@
...
+      default_lead_image_url || placeholder_image_url
+    end
+
+    def default_lead_image_url
+      lead_image_url(default_lead_image)
+    end
+
+  private
+
+    def lead_image_url(image_data)
```
ryanb-gds: Worth creating a method on the image data object so that we don't have to do `.file.url(:size)` all the time? Surprised it's not there already tbh

app/views/admin/configurable_content_blocks/_lead_image_select.html.erb
```diff
@@ -0,0 +1,50 @@
+<% default_lead_image_hint =  lead_image_select.default_lead_image_url ? "The organisation's default news image is used as a lead image. It appears at the top of the document." : "The organisation does not have a default news image, so a placeholder image will be shown at the top of the document." %>
+<% custom_lead_image_hint = "Select a lead image. #{lead_image_select.default_lead_image_url ?  "If no image is provided, the organisation's default news image will be shown." : "If no image is provided, a placeholder will be shown."}" %>
```
ryanb-gds: I'd be tempted to put these in locale files, because I hate having long strings in templates that I have to scroll past before I can read the template code. Up to you though

app/views/admin/configurable_content_blocks/_lead_image_select.html.erb
```diff
@@ -0,0 +1,50 @@
...
+<% select_options = [
+     {
+       text: "No image selected",
+       value: "",
+     },
+   ] + lead_image_select.images.map do |image|
+  {
+    text: image.filename,
+    value: image.image_data.id,
+    selected: image.image_data.id == (translated_content || content).to_i,
```
ryanb-gds, line 37: `translated_content || content` definitely evaluates to a string and not a boolean, right? 😅

What if they're both nil?

test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
```diff
@@ -0,0 +1,235 @@
...
+      "properties" => {
+        "test_attribute" => {
+          "type" => "string",
+          "title" => "Test attribute",
+          "description" => "A test attribute",
+          "format" => "lead_image_select",
+        },
+      },
+    }
+    ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type", { schema: }))
```
ryanb-gds: It would be good actually if we didn't set these up as heavily as we do, because it's not strictly necessary, I just hadn't got around to bringing the image select block tests into line with the other blocks. But if you look at the rendering tests for the [govspeak block](https://github.com/alphagov/whitehall/blob/86feed7e76921e4415e5dbe3263971bcec340d07/test/unit/app/models/configurable_content_blocks/govspeak_test.rb#L38), you'll see they don't use a real standard edition or setup the document type fully, so they're quicker to run and less coupled to other parts of the system. It would be nice to bring these tests into line, though appreciate we may want to deal with that in a separate PR

test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
```diff
@@ -0,0 +1,235 @@
...
+    block = ConfigurableContentBlocks::LeadImageSelect.new([], default_lead_image: nil)
+
+    render block, {
+      schema: schema["properties"]["test_attribute"],
+      content: "",
+      path: Path.new.push("test_attribute"),
+    }
+
+    assert_dom "h2", text: "Default lead image"
+    assert_match %r{placeholder.jpg}, rendered
```
ryanb-gds: What's this rendered thing? Any reason not to do a (presumably more specific) `assert_dom "img[src=?]"` kind of assertion?

app/models/configurable_content_blocks/lead_image_select.rb
lauraghiorghisor-tw: I went through too many permutations of this code and probably ended up with something less readable that only looks more readable after all. There is a little if content.present? on l. 19 😅

Will push a new version.

app/models/configurable_content_blocks/image_select.rb
lauraghiorghisor-tw: I thought about it. Can't really put it in the concern, as then I don't have access to the edition. I could put it on the image model, but because it's a little manipulation it belongs more around this payload scope. Could be in a helper (all these lead-y methods could be)?

app/models/configurable_content_blocks/lead_image_select.rb
lauraghiorghisor-tw: Can use the attachment data approach and delegate to file. Will push commit.

app/views/admin/configurable_content_blocks/_lead_image_select.html.erb
lauraghiorghisor-tw: That code was already there, from `image_select`, so I have not read it 😅 I did not test translations behaviours at all, actually. Good call. Will redeploy and give that a go as well. 

Well:
```
a = 4 
# numerical image ID
a == ("4" || nil).to_i
=> true
a == (nil || "4").to_i
=> true
a == (nil || nil).to_i
=> false
 a == ("" || "").to_i
=> false # I am not sure what content comes in as ??
 "" || ""
=> ""
[13] pry(main)> nil || nil
=> nil
[14] pry(main)> 4 || nil
=> 4
[15] pry(main)> nil || 4
=> 4
```

app/models/concerns/standard_edition/default_lead_image.rb
lauraghiorghisor-tw: Hmmmmm, it's a funny one because the org association is conditionally enabled by the schema, so then including it here would invalidate that. Inclined to just check for the methods, all of them. Will push.

app/views/admin/configurable_content_blocks/_lead_image_select.html.erb
lauraghiorghisor-tw: I gave it a go, hard to tell if that's improved readability... There's a bit of computation in that ternary, so I can't just put it all in the locale.

 - view:
```
<% default_lead_image_hint =  lead_image_select.default_lead_image_url ? t("configurable_content_blocks.lead_image_select.default_hint_for_news_image") : t("configurable_content_blocks.lead_image_select.default_hint_for_placeholder") %>
<% custom_lead_image_hint = "Select a lead image. #{lead_image_select.default_lead_image_url ? t("configurable_content_blocks.lead_image_select.custom_hint_for_news_image") : t("configurable_content_blocks.lead_image_select.custom_hint_for_placeholder")}" %>
```

- and `en.yml`:
```
  configurable_content_blocks:
    lead_image_select:
      default_hint_for_news_image: "The organisation's default news image is used as a lead image. It appears at the top of the document."
      default_hint_for_placeholder: "The organisation does not have a default news image, so a placeholder image will be shown at the top of the document."
      custom_hint_for_news_image: "If no image is provided, the organisation's default news image will be shown."
      custom_hint_for_placeholder: "If no image is provided, a placeholder will be shown."
```

How do you feel about it, should I commit this?

test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
lauraghiorghisor-tw: I don't think I cared too much about specificity - but we will introduce another placeholder for the world news stories anyway so might as well make it specific.

The rendered thing is that bit of html:
```
rendered
"\n  \n<div class=\"gem-c-heading govuk-!-margin-bottom-2\">\n\n  <h2 class=\"gem-c-heading__text govuk-heading-m\">\n    Default lead image\n</h2></div>\n\n  <div id=\"hint-9ab555a0\" class=\"gem-c-hint govuk-hint govuk-!-margin-bottom-3\">\n  The organisation does not have a default news image, so a placeholder image will be shown at the top of the document.\n</div>\n\n  <div class=\"app-c-edition-images-lead-image-component__default_lead_image govuk-!-margin-bottom-4\">\n    <div class=\"govuk-grid-row\">\n      <div class=\"govuk-grid-column-one-third\">\n        <img src=\"https://www.test.gov.uk/assets/whitehall/placeholder.jpg\" alt=\"Default lead image\" class=\"app-view-edition-resource__preview\">\n      </div>\n    </div>\n  </div>\n\n  <div class=\"govuk-form-group gem-c-select\">\n    \n  <label for=\"edition_test_attribute\" id=\"edition_test_attribute-label\" class=\"gem-c-label govuk-label govuk-label--m\">Test attribute</label>\n\n  <div id=\"hint-524e89b2\" class=\"gem-c-hint govuk-hint\">\n  Select a lead image. If no image is provided, a placeholder will be shown.\n</div>\n\n\n\n    <select name=\"edition[block_content][test_attribute]\" id=\"edition_test_attribute\" class=\"govuk-select gem-c-select__select--full-width\" aria-describedby=\"hint-524e89b2\"><option value=\"\">No image selected</option></select>\n</div>\n"
```

test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
lauraghiorghisor-tw: I can give that a go (wip for now)

lauraghiorghisor-tw: Done for lead and regular image ✅

app/models/configurable_content_blocks/image_select.rb
ryanb-gds: Maybe just call it `image_caption` instead of `lead_image_caption`, as we're on the `image_select` block?

app/models/configurable_content_blocks/lead_image_select.rb
ryanb-gds: I now feel this private method doesn't really justify its existence because it's so simple, but minor nitpick 🤷

app/models/image_data.rb
```diff
@@ -19,7 +19,7 @@ class ImageData < ApplicationRecord
   validate :filename_is_unique
 
   delegate :width, :height, to: :dimensions
-  delegate :content_type, to: :file
+  delegate :url, :content_type, to: :file
```
ryanb-gds, line 22: Nice.

app/views/admin/configurable_content_blocks/_lead_image_select.html.erb
ryanb-gds: Personally I prefer it. You may wish to store the various hints in intermediate variables before doing the ternary I guess if it helps, or maybe move the ternary into the template. Up to you though

app/views/admin/configurable_content_blocks/_lead_image_select.html.erb
ryanb-gds: Looks probably alright 😅

test/unit/app/models/configurable_content_blocks/lead_image_select_test.rb
```diff
@@ -0,0 +1,221 @@
...
+      "properties" => {
+        "test_attribute" => {
+          "type" => "string",
+          "title" => "Test attribute",
+          "description" => "A test attribute",
+          "format" => "lead_image_select",
+        },
+      },
+    }
+    default_image = create(:organisation, :with_default_news_image).default_news_image
```
ryanb-gds: Dunno if you could get away with `build` rather than `create` here or not

app/models/configurable_content_blocks/lead_image_select.rb
lauraghiorghisor-tw: You're right, it was at some point used in the view -> I can inline it, and I can move `default_lead_image_url` to private too, and check on whether we have a `default_news_image` for the ternaries, rather than checking on the `url`.

PR #10666, opened 2025-09-16
Avoid sending `details.headers` for unsupported content types (WHIT-2440)

ChrisBAshton:
The top level `details.headers` property is required for History Pages:
https://github.com/alphagov/publishing-api/blob/4005b72ce5996ade8b83eda568c38afcbe63fd79/content_schemas/formats/history.jsonnet#L19-L20

It is not allowed by the schema for News Articles, nor do we see much purpose in allowing it:
https://gds.slack.com/archives/C02L13S214K/p1757428495759379

It is therefore a configurable property in our JSON. We can always revisit this later (e.g. split out into separate presenter for History pages, or adding support for headers to the News Article jsonnet, etc).

---

### DIFF
--- a/app/models/configurable_document_types/history_page.json
+++ b/app/models/configurable_document_types/history_page.json
@@ -34,6 +34,7 @@
     "publishing_api_document_type": "history",
     "rendering_app": "frontend",
     "images_enabled": true,
+    "send_headings": true,
     "organisations": ["af07d5a5-df63-4ddc-9383-6a666845ebe9"],
     "backdating_enabled": false,
     "history_mode_enabled": false,
--- a/app/models/configurable_document_types/news_story.json
+++ b/app/models/configurable_document_types/news_story.json
@@ -26,6 +26,7 @@
     "publishing_api_document_type": "news_story",
     "rendering_app": "frontend",
     "images_enabled": true,
+    "send_headings": false,
     "file_attachments_enabled": true,
     "organisations": null,
     "backdating_enabled": true,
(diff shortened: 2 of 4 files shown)


lauraghiorghisor-tw (approved):
I guess that's fine, although I only half remember the chats we had about it.

I'll give it a 👍  because it makes things OBVIOUS which I feel is the vibe we're going for with the config-driven approach.

Nitpicking: Naming-wise, we say headings and headers interchangeably, which is fine, but because the key we set is "headers", maybe call the setting also "send_headers"?

ChrisBAshton:
> I guess that's fine, although I only half remember the chats we had about it.
> 
> I'll give it a 👍 because it makes things OBVIOUS which I feel is the vibe we're going for with the config-driven approach.
> 
> Nitpicking: Naming-wise, we say headings and headers interchangeably, which is fine, but because the key we set is "headers", maybe call the setting also "send_headers"?

That is a very fair nitpick 👍 I chose "send_headings" off the back of Monday's "Regular tech catchup on Flexible Pages" meeting, where there was a consensus that the original choice of "headers" should probably have been "headings" and that there's a rough plan to go and rename it now. I could go with "send_headers" and then leave it to someone to rename en masse later? But even in the short term it potentially feels like an improvement, because anyone glancing at the config of a "send_headers" might assume that's to do with HTTP requests. So I think I'll leave as-is.

lauraghiorghisor-tw:
> send_headers

Oooh, ook, fair.

PR #10670, opened 2025-09-17
Add more properties to StandardEdition's details hash (WHIT-2498)

ChrisBAshton:
We want to replicate the News Articles payload as closely as possible, in migrating them to the new 'StandardEdition' format. The [relevant NewsArticlePresenter code](https://github.com/alphagov/whitehall/blob/d8f5013d0ab3eefbb78c007e860942aa43e970a5/app/presenters/publishing_api/news_article_presenter.rb#L72-L80) is:

```
 details
      base_details
        .merge(ChangeHistory.for(news_article))
        .merge(Image.for(news_article))
        .merge(PayloadBuilder::FirstPublicAt.for(news_article))
        .merge(PayloadBuilder::PoliticalDetails.for(news_article))
        .merge(PayloadBuilder::TagDetails.for(news_article))
        .merge(PayloadBuilder::Attachments.for(news_article))
    end
```

This PR carries over the relevant properties from the above, into the StandardEditionPresenter, with the following exceptions:

1. No `first_public_at` property, which is just a [duplication](https://github.com/alphagov/whitehall/blob/aa9cdc7d53e68f1bf75443e60d7d3186534208de/app/presenters/publishing_api/payload_builder/first_public_at.rb#L9) of the existing `first_published_at` property. It isn't included on the live news articles originating from Content Publisher ([example](https://www.gov.uk/api/content/government/news/covid-19-vaccine-surveillance-report-published)) and is [considered deprecated](https://github.com/alphagov/content-publisher/blob/7a999c738bb982db267a23577af80db66a66312d/docs/adr/0016-publishing-times-and-change-history.md?plain=1#L40). government-frontend [falls back to first_published_at](https://github.com/alphagov/government-frontend/blob/f38f22922249a57e9df5ec9fe9a520ad91e8b0dd/app/presenters/content_item/updatable.rb#L18), as does Frontend ([for detailed guides](https://github.com/alphagov/frontend/blob/10bd509470017853ec103c075cb9bfcc54d067ca/app/views/detailed_guide/show.html.erb#L41) - but [not for case studies](https://github.com/alphagov/frontend/blob/10bd509470017853ec103c075cb9bfcc54d067ca/app/views/case_study/show.html.erb#L65)) - one can imagine how any rendering app references to `first_public_at` can be swapped out for `first_published_at`.
2. No `tags` (`{ browse_pages: [] }`) property, which has been [recognised as legacy and unused](https://github.com/alphagov/email-alert-api/pull/2136#discussion_r1594207581).
3. No `image` property (yet) while we're figuring out how to represent lead images (WHIT-2384](https://gov-uk.atlassian.net/browse/WHIT-2384).

---

### DIFF
--- a/app/models/configurable_document_types/history_page.json
+++ b/app/models/configurable_document_types/history_page.json
@@ -35,6 +35,7 @@
     "rendering_app": "frontend",
     "images_enabled": true,
     "send_headings": true,
+    "send_change_history": false,
     "organisations": ["af07d5a5-df63-4ddc-9383-6a666845ebe9"],
     "backdating_enabled": false,
     "history_mode_enabled": false,
--- a/app/presenters/publishing_api/standard_edition_presenter.rb
+++ b/app/presenters/publishing_api/standard_edition_presenter.rb
@@ -45,6 +45,7 @@ module PublishingApi
       }
       details.merge!(PayloadBuilder::ChangeHistory.for(item)) if type.settings["send_change_history"] == true
       details.merge!(PayloadBuilder::PoliticalDetails.for(item)) if type.settings["history_mode_enabled"] == true
+      details.merge!(PayloadBuilder::Attachments.for(item)) if type.settings["file_attachments_enabled"] == true
       details
     end
 
--- a/app/presenters/publishing_api/standard_edition_presenter.rb
+++ b/app/presenters/publishing_api/standard_edition_presenter.rb
@@ -40,9 +40,11 @@ module PublishingApi
 
     def details
       root_block = ConfigurableContentBlocks::Factory.new(item).build("object")
-      {
+      details = {
         **flatten_headers(root_block.publishing_api_payload(type.schema, item.block_content)),
       }
+      details.merge!(PayloadBuilder::PoliticalDetails.for(item)) if type.settings["history_mode_enabled"] == true
+      details
     end
 
     def flatten_headers(content)
--- a/app/presenters/publishing_api/standard_edition_presenter.rb
+++ b/app/presenters/publishing_api/standard_edition_presenter.rb
@@ -43,6 +43,7 @@ module PublishingApi
       details = {
         **flatten_headers(root_block.publishing_api_payload(type.schema, item.block_content)),
       }
+      details.merge!(PayloadBuilder::ChangeHistory.for(item)) if type.settings["send_change_history"] == true
       details.merge!(PayloadBuilder::PoliticalDetails.for(item)) if type.settings["history_mode_enabled"] == true
       details
     end
--- a/app/models/configurable_document_types/news_story.json
+++ b/app/models/configurable_document_types/news_story.json
@@ -27,6 +27,7 @@
     "rendering_app": "frontend",
     "images_enabled": true,
     "send_headings": false,
+    "send_change_history": true,
     "file_attachments_enabled": true,
     "organisations": null,
     "backdating_enabled": true,
--- a/test/unit/app/presenters/publishing_api/standard_edition_presenter_test.rb
+++ b/test/unit/app/presenters/publishing_api/standard_edition_presenter_test.rb
@@ -307,4 +307,41 @@ class PublishingApi::StandardEditionPresenterTest < ActiveSupport::TestCase
 
     assert_not content[:details].key?(:change_history)
   end
+
+  test "it includes attachments in the details if file attachments are enabled" do
+    type_key = "test_type_key"
+    ConfigurableDocumentType.setup_test_types(build_configurable_document_type(type_key, {
+      "settings" => {
+        "file_attachments_enabled" => true,
+      },
+    }))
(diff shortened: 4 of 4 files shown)


app/models/configurable_document_types/history_page.json
```diff
@@ -35,6 +35,7 @@
     "rendering_app": "frontend",
     "images_enabled": true,
     "send_headings": true,
+    "send_change_history": false,
```
ryanb-gds, line 38: I think in an ideal world we would send the change history for all pages and it would be a top level key, outside the details hash. I suspect that would be a painful migration though.

app/presenters/publishing_api/standard_edition_presenter.rb
```diff
@@ -40,9 +40,13 @@ def edition_links
...
     def details
       root_block = ConfigurableContentBlocks::Factory.new(item).build("object")
-      {
+      details = {
         **flatten_headers(root_block.publishing_api_payload(type.schema, item.block_content)),
       }
+      details.merge!(PayloadBuilder::ChangeHistory.for(item)) if type.settings["send_change_history"] == true
+      details.merge!(PayloadBuilder::PoliticalDetails.for(item)) if type.settings["history_mode_enabled"] == true
+      details.merge!(PayloadBuilder::Attachments.for(item)) if type.settings["file_attachments_enabled"] == true
+      details
```
ryanb-gds, line 49: I'm starting to think we need an abstraction around settings, much like we have for blocks (and will have for associations), so that we can give the settings objects a chance to manipulate the payload.

Or, should the political field/history mode and the change note field just be normal blocks? I'm not sure how realistic that is because there is some very particular code around those.

ryanb-gds (approved):
Looks good for now, but I think we'll want to encapsulate some of the complexity of these settings away from the presenter (and indeed the HTML form, though not affected here) going forward

app/presenters/publishing_api/standard_edition_presenter.rb
ChrisBAshton: Sounds good. It does feel like a bit of a surprise that a setting in the config would be queried by the presenter and that the StandardEdition model itself has no idea about it. Some better way of organising these going forward would be good.

app/presenters/publishing_api/standard_edition_presenter.rb
```diff
@@ -45,6 +45,7 @@ def details
       }
       details.merge!(PayloadBuilder::ChangeHistory.for(item)) if type.settings["send_change_history"] == true
       details.merge!(PayloadBuilder::PoliticalDetails.for(item)) if type.settings["history_mode_enabled"] == true
+      details.merge!(PayloadBuilder::Attachments.for(item)) if type.settings["file_attachments_enabled"] == true
```
lauraghiorghisor-tw, line 48: Oooh, we missed this bit when we implemented the other PR. I am never quite sure if the govspeak handling is enough or not for attachments (because we do already include them in the preprocessing, so the url is there, as well as a lot of metadata). Just not very sure what FE does with it. 

I know I am always militating for too much digging into things, but can we spike this/study what FE is doing here?

I added this to the doc https://docs.google.com/document/d/1vG8SJN8o4BM0MaZGJDt1RL9UPwhoscdpXRXrc_56Wbo/edit?tab=t.0#heading=h.xtnkxrta6vqc. I only had a glance at it, but it seems to me that there is a lot of redundancy in the payload which makes it a bit unclear what is truly needed in the FE. It could well be that everything is.

ryanb-gds: DIdn't miss it, just chose not to do it because news articles only work with inline attachments so this wasn't required.

lauraghiorghisor-tw: So then this line of code has no business existing? https://github.com/alphagov/whitehall/blob/bfffe6bc568e8bd676ff6faaf7cd7d4d3fdf20b4/app/presenters/publishing_api/news_article_presenter.rb#L79

In which case, it can be excluded from the `StandardEditionPresenter` too.

lauraghiorghisor-tw: That's what I mean when I say, I think there's unnecessary duplication here.

ryanb-gds: Yes I think you're correct

lauraghiorghisor-tw: If you're in the region with that PR you said you're doing about files - are you ok to have a look to see if FE actually needs this for NewsArticles? 

I figure I/you/we tested the [Files PR](https://github.com/alphagov/whitehall/pull/10638) and it must have worked as expected without this in the Presenter (sorry, no memory of it though).

I imagine this only needs to be included for Publications and the like, but even then we might be sending enough information through some other bits we include in the `details` hash. 

If not, I can make a new card/short spike to dig into this.

PR #10674, opened 2025-09-18
Add database migration to copy alt text into empty caption fields

ChrisBAshton:
We only support alt text on legacy images. Newly uploaded images have no alt text option.

Rather than continuing to maintain two different policies on images, we're consolidating by dropping the alt text column - but there are cases where the alt text is actually useful (rather than merely decorative) and would be best retained as a caption (provided no caption exists already). So this is a first step before we can drop the alt text property.

JIRA: https://gov-uk.atlassian.net/browse/WHIT-2475

---


leimdorfer:
I checked with content and I think we've missed the cases where there is an existing caption, but the alt_text is a better description. Specifically there are many cases where the caption only has a credit/copyright notice and no description, but there is a description in the alt_text

lauraghiorghisor-tw:
@GDSNewt should this be closed in favour of https://github.com/alphagov/whitehall/pull/10724?

ChrisBAshton:
Superseded

PR #10679, opened 2025-09-23
Remove govspeak renderer

ryanb-gds:
This refactor simplifies a number of aspects of presenting content to Publishing API, mainly around parts of the content that need to be "rendered" as HTML before we send them to Publishing API (govspeak and attachments)


TonyGDS (approved):
LGTM

PR #10683, opened 2025-09-25
Add support for translation [WHIT-2385]

ryanb-gds:
JIRA: [WHIT-2385](https://gov-uk.atlassian.net/browse/WHIT-2385?atlOrigin=eyJpIjoiMTQ0NGNjY2M2M2Q3NDgzNmE2MWU4ZDZlYTYxMWYwZjciLCJwIjoiaiJ9)

This PR allows standard editions to support translations.

The translated content and RTL support being drilled down through the configurable block templates to the leaf blocks feels a bit clunky. Blocks are starting to receive a lot of arguments, either via their initialiser or via template partial local assigns, so we probably need to look at a refactoring soon to manage that complexity.

[WHIT-2385]: https://gov-uk.atlassian.net/browse/WHIT-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/features/standard-edition.feature
+++ b/features/standard-edition.feature
@@ -14,3 +14,24 @@ Feature: Standard Editions
     When I publish a submitted draft of a test configurable document titled "The history of GOV.UK"
     Then I can see that the draft edition of "The history of GOV.UK" was published successfully
     And a new draft of "The history of GOV.UK" is created with the correct field values
+
+  Scenario: Adding translations with all content block types
+    Given I am a GDS admin
+    And the configurable document types feature flag is enabled
+    And the test configurable document type is defined with translations enabled
+    And I have drafted an English configurable document titled "Digital transformation report"
+    When I add a Welsh translation "Adroddiad trawsnewid digidol"
+    Then configured content blocks should appear on the translation page
+    And the Welsh translation fields should be pre-populated with primary locale content
+    And the image selections should be preserved from the primary locale
+    And I should see the original English content in "original text" sections
+
+  Scenario: Creating a non-English primary locale configurable document
+    Given I am a GDS admin
+    And the configurable document types feature flag is enabled
+    And the test configurable document type is defined with translations enabled
+    When I create a new configurable document with Welsh as the primary locale titled "Strategaeth Ddigidol Cymru"
+    Then the document should be saved with Welsh as the primary locale
+    And English should not appear as a translation option
+
+
--- a/features/step_definitions/standard_edition_steps.rb
+++ b/features/step_definitions/standard_edition_steps.rb
@@ -1,9 +1,58 @@
+def create_test_image
+  create(:image)
+end
+
+def create_configurable_document(title:, locale: "en", summary: nil, body: nil)
+  image = create_test_image
+  defaults = default_content_for_locale(locale)
+  create(
+    :draft_standard_edition,
+    {
+      configurable_document_type: "test",
+      images: [image],
+      title: title,
+      summary: summary || defaults[:summary],
+      primary_locale: locale,
+      block_content: {
+        "image" => image.image_data.id.to_s,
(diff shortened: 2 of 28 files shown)


features/step_definitions/standard_edition_steps.rb
```diff
@@ -63,3 +115,83 @@
...
+And(/^the Welsh translation fields should be pre-populated with primary locale content$/) do
+  expect(page).to have_field("edition_block_content_body", with: /govspeak/)
+end
+
+And(/^the image selections should be preserved from the primary locale$/) do
+  edition = @standard_edition || StandardEdition.last
+  expect(page).to have_select("Image", selected: edition.images.first.filename)
+end
+
+And(/^I should see the original English content in "original text" sections$/) do
```
ryanb-gds, line 174: We probably ought to limit the selection to the appropriate DOM location using a CSS selector because otherwise this assertion might potentially pick up the original content and pass when it shouldn't

features/step_definitions/standard_edition_steps.rb
```diff
@@ -63,3 +115,83 @@
...
+  add_translation(
+    @welsh_edition,
+    "Français",
+    "Document français",
+    "Résumé français du document.",
+    "## Contenu français\n\nCeci est le contenu du corps en français",
+  )
+end
+
+Then(/^all content blocks should appear on the translation page$/) do
```
ryanb-gds: Maybe `configured content blocks` would be a better name than `all content blocks`? I initially thought this test was designed to use every type of block because of the use of "all".

features/step_definitions/standard_edition_steps.rb
```diff
@@ -1,9 +1,58 @@
...
+  click_button "Save"
+end
+
 Given(/^the configurable document types feature flag is (enabled|disabled)$/) do |enabled|
   @test_strategy ||= Flipflop::FeatureSet.current.test!
   @test_strategy.switch!(:configurable_document_types, enabled == "enabled")
 end
 
-Given(/^the test configurable document type is defined$/) do
+Given(/^the test configurable document type is defined(?: with translations enabled)?$/) do
```
ryanb-gds, line 55: This is a nice touch to make the scenario read more explicitly ⭐

features/standard-edition.feature
```diff
@@ -14,3 +14,31 @@ Feature: Standard Editions
...
+    And the configurable document types feature flag is enabled
+    And the test configurable document type is defined with translations enabled
+    And I have drafted an English configurable document titled "Digital transformation report"
+    When I add a Welsh translation "Adroddiad trawsnewid digidol"
+    Then all content blocks should appear on the translation page
+    And the Welsh translation fields should be pre-populated with primary locale content
+    And the image selections should be preserved from the primary locale
+    And I should see the original English content in "original text" sections
+
+    Scenario: Creating a non-English primary locale configurable document
```
ryanb-gds: Indentation is off here. Surprised we haven't got a linter for cucumber

features/standard-edition.feature
```diff
@@ -14,3 +14,31 @@ Feature: Standard Editions
...
+    Given I am a GDS admin
+    And the configurable document types feature flag is enabled
+    And the test configurable document type is defined with translations enabled
+    When I create a new configurable document with Welsh as the primary locale titled "Strategaeth Ddigidol Cymru"
+    Then the document should be saved with Welsh as the primary locale
+    And English should not appear as a translation option
+
+  Scenario: Non-English to non-English translations exclude English options
```
ryanb-gds: This feels like a very low-level scenario to cover in a feature - any way to write a unit test for this? Features are expensive so we only want to add them where they're really valuable.

PR #10684, opened 2025-09-25
[WHIT-2386] Fix association links

TonyGDS:
## What

Links for role appointments and organisations have been fixed

## Why
* Role appointment links did not conform to the Publishing API schema
* Organisations links tried to send the primary lead organisation with an order of 0, which is not possible

### DIFF
--- a/app/models/configurable_associations/ministerial_role_appointments.rb
+++ b/app/models/configurable_associations/ministerial_role_appointments.rb
@@ -5,10 +5,12 @@ module ConfigurableAssociations
     end
 
     def links
-      @association.preload(:person)
-      {
-        role_appointments: @association.map { |role_appointment| role_appointment.person.content_id },
-      }
+      @association.includes(:person, :role)
+
+      @association.each_with_object({ people: [], roles: [] }) do |role_appointment, links|
+        links[:people] << role_appointment.person.content_id
+        links[:roles] << role_appointment.role.content_id
+      end
     end
 
     def selected_ids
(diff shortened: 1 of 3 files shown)


app/models/configurable_associations/ministerial_role_appointments.rb
```diff
@@ -5,10 +5,12 @@ def initialize(association)
     end
 
     def links
-      @association.preload(:person)
-      {
-        role_appointments: @association.map { |role_appointment| role_appointment.person.content_id },
-      }
+      @association.includes(:person, :role)
```
ryanb-gds, line 8: Ah, nice that you can call includes here and get the same result as preload. Includes is much better documented.

test/unit/app/presenters/publishing_api/standard_edition_presenter_test.rb
```diff
@@ -375,8 +375,11 @@ class PublishingApi::StandardEditionPresenterTest < ActiveSupport::TestCase
     edition.edition_organisations.build([{ organisation: organisations.first, lead: true, lead_ordering: 0 }, { organisation: organisations.last, lead: false }])
     presenter = PublishingApi::StandardEditionPresenter.new(edition)
     links = presenter.links
-    expected_role_appointments = ministerial_role_appointments.map { |appointment| appointment.person.content_id }
-    assert_equal expected_role_appointments, links[:role_appointments]
+    expected_people, expected_roles = ministerial_role_appointments
+                                        .map { |appointment| [appointment.person.content_id, appointment.role.content_id] }
+                                        .transpose
```
ryanb-gds, line 380: Nice use of transpose, hadn't seen that one before

ryanb-gds (approved):
Thanks for fixing this up

PR #10685, opened 2025-09-25
[WHIT-2386] Fix indexing of lead organisations for StandardEdition form

TonyGDS:
## What 
Changed the method used to determine what lead organisations are selected to be displayed on the page to use array index rather than lead organisation order

## Why
Lead ordering is not always zero based. Retrieving in this way ensures we can display pre selected options in from the first position in regardless of the starting lead ordering number.


ryanb-gds (approved):
LGTM

PR #10697, opened 2025-09-29
(CM-434) Remove Content Block Manager from Whitehall

pezholio:
Now Content Block Manager has been migrated to [its own repo](https://github.com/alphagov/content-block-manager/), we can safely remove all the Content Block Manager related code from the Whitehall repo.


ryanb-gds (approved):
Thanks Stu. I'm sure you will anyway, but I'd appreciate you giving this a thorough look over on integration before you merge, just in case. There's a lot of code going away 😅

pezholio:
I've deployed to integration and had a poke around, as well as running the E2E tests, and all seems to be working well 🙌

PR #10702, opened 2025-09-30
[WHIT-2518] Send Government links to Publishing API for configurable documents

TonyGDS:
## What
* Use `history_mode_enabled` in a configurable document to determine if Government links should be sent to Publishing API
* Move logic to build links into its own Payload Builder for configurable document types

## Why
News stories need to send Government links but this was not yet supported by the `StandardEditionPresenter`

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-2518)


ryanb-gds (approved):
Looks good to me, good to have that test outside of the general edition presenter test

PR #11210, opened 2026-03-05
Allow publishing of conflicting topical events to one base_path - attempt two

ChrisBAshton:
This is a second attempt of #11198, which had to be reverted because role appointments don't have a `base_path` sent to Publishing API, which caused Whitehall to crash on the introduced line. See [Sentry error](https://govuk.sentry.io/issues/7312203252/?alert_rule_id=9401721&alert_type=issue&environment=production&notification_uuid=d0d266c2-afbd-4eaa-86b0-7715c16eda1d&project=202259&referrer=slack).

That got reverted, and this PR is a second attempt at the change but accounting for the possibility of the `base_path` not being present.

Original PR description below.

---

Currently, Whitehall has a safeguard that raises an exception if you try to publish a Whitehall-owned piece of content under the same public_path as another Whitehall-owned piece of content.

In the short term, however, we'd like to allow the publish (effectively an overwrite) in the case of topical events. This is to aid in the migration of topical events.

Proposed migration workflow:

1. A legacy topical event exists
2. We write a migration script that creates a StandardEdition _copy_ of the legacy topical event
3. We can save and publish the StandardEdition version and see how that looks on the frontend
4. We can also interchangably swap back to the legacy version if for whatever reason the config-driven version is problematic. We would simply edit and 'save' the legacy version, which automatically publishes to Publishing API.

Once all legacy topical events have been migrated, we will revert this change.

---

### DIFF
--- a/lib/whitehall/publishing_api.rb
+++ b/lib/whitehall/publishing_api.rb
@@ -182,6 +182,8 @@ module Whitehall
     end
 
     def self.ensure_base_path_is_associated_with_this_content_id!(base_path, content_id)
+      return if base_path.nil?
+
       existing_content_id = Services.publishing_api.lookup_content_id(base_path:)
       return if existing_content_id.nil? || existing_content_id == content_id
 
--- a/lib/whitehall/publishing_api.rb
+++ b/lib/whitehall/publishing_api.rb
@@ -182,7 +182,9 @@ module Whitehall
     end
 
     def self.ensure_base_path_is_associated_with_this_content_id!(base_path, content_id)
-      return if base_path.nil?
+      # Temporary hack to allow both legacy and config-driven topical events to co-exist,
+      # to aid in the migration. This early return should be removed once the migration is finished.
+      return if base_path.nil? || base_path.match?(%r{^/government/topical-events/})
 
       existing_content_id = Services.publishing_api.lookup_content_id(base_path:)
       return if existing_content_id.nil? || existing_content_id == content_id
--- a/test/unit/lib/whitehall/publishing_api_test.rb
+++ b/test/unit/lib/whitehall/publishing_api_test.rb
@@ -314,6 +314,14 @@ class Whitehall::PublishingApiTest < ActiveSupport::TestCase
     )
   end
 
+  test "#ensure_base_path_is_associated_with_this_content_id! skips Publishing API lookup if no base_path provided" do
+    Whitehall::PublishingApi.unstub(:ensure_base_path_is_associated_with_this_content_id!)
+
+    Services.publishing_api.expects(:lookup_content_id).with(base_path: nil).never
+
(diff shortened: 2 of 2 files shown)


lauraghiorghisor-tw (commented):
I am ok with this, but am not super convinced that it's something we should merge now. 

I really like the idea that you can flexibly publish one thing or another. But let's consider a scenario where we republish in bulk, and then maybe something goes wrong -> what measures do we have to check the data, other than logs? Publishing API won't know which sort of "type" it got, as far as it knows it is "topical_event". And in whitehall both those docs (legacy and config) would appear as published? I'd be inclined to have more rigid release steps where we know better what is to be migrated, which paths we need to release, and what has been republished. But also cool if we can do it with this, I'd just need to see exactly how we know what's going on. 

And if we merge this, don't we still risk that issue that you first thought about, i.e. someone might create a new topical event at the same path as an old one and override it. I know there is governance around what topical events people can create, but it's still a technical possibility. So I'd suggest only enabling this when we are ready to migrate/test the migration, rather than by default 🤔

lauraghiorghisor-tw (approved):
Approved but holding off the merge for now, until we are ready to consider migration.

lauraghiorghisor-tw:
👌 again - we can figure out the specifics of those questions I had above, when we do the migration card - e.g. how to tell the migration has been successful.

ChrisBAshton:
> I really like the idea that you can flexibly publish one thing or another. But let's consider a scenario where we republish in bulk, and then maybe something goes wrong -> what measures do we have to check the data, other than logs? Publishing API won't know which sort of "type" it got, as far as it knows it is "topical_event". And in whitehall both those docs (legacy and config) would appear as published? I'd be inclined to have more rigid release steps where we know better what is to be migrated, which paths we need to release, and what has been republished. But also cool if we can do it with this, I'd just need to see exactly how we know what's going on.

This is the main risk of merging the PR. If cloned StandardEdition topical events and their legacy equivalents co-exist, currently in a bulk-republish scenario then the live topical event would be either legacy or StandardEdition depending on which was the last one to be processed.

In practice, I think the risk is minimal because:

1. We're only intending to keep this PR merged for a short while (while we're developing and testing the migration). Perhaps even only a day or two.
2. We would delete the duplicated legacy topical events before bulk-republishing topical events.
3. Even if a legacy topical event somehow got published, it should have no impact on what the user sees on the frontend (identical template and content).

> And if we merge this, don't we still risk that issue that you first thought about, i.e. someone might create a new topical event at the same path as an old one and override it. I know there is governance around what topical events people can create, but it's still a technical possibility. So I'd suggest only enabling this when we are ready to migrate/test the migration, rather than by default 🤔

Yes, but highly unlikely, given the governance etc. A risk we can accept.

To reiterate, this PR is a tool to aid in testing that a duplicated/migrated topical event is unchanged before-and-after. It's not strictly necessary but it's quite nice to be able to do so. Given we're picking up the migration card this sprint, I'll go ahead and merge it for now.

PR #11262, opened 2026-03-18
WIP: remove organisation_image option [WHIT-2801]

ChrisBAshton:
The behaviour designated by this option is the same behaviour designated by a `nil` value.

The reason we've had to support both `nil` and `organisation_image` in the past is because of the legacy CaseStudy behaviour in which, when deleting an image, the next applicable image would become the new lead image. If the publisher had not yet chosen a lead image and was using the organisation fallback image, then this would have led to a bug where they suddenly have a lead image set when they didn't expect to. Thus the check for 'organisation_image' (no lead image set) vs 'nil' (we should auto-select a lead image when an image is deleted).

This workflow doesn't make sense in the new world where lead images have their own dedicated workflow - they are no longer selected from a 'pool' of images all uploaded to the edition.

We can't merge this commit as-is until we write a data migration to make all "organisation_image" image_display_options now 'nil'.

---

### DIFF
--- a/app/components/admin/edition_images/lead_image_card_component.html.erb
+++ b/app/components/admin/edition_images/lead_image_card_component.html.erb
@@ -13,7 +13,7 @@
       key: "Image",
       value: thumbnail,
     },
-    image_usage.caption_enabled? ? {
+    image.present? && image_usage.caption_enabled? ? {
       key: "Caption",
       value: caption,
     } : nil,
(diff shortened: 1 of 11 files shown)


ChrisBAshton:
Now written up as a backlog item: https://gov-uk.atlassian.net/browse/WHIT-3251

PR #11344, opened 2026-04-09
Add a data migration to apply slug overrides for live documents [WHIT-3063]

ryanb-gds:
There are a number of Whitehall documents that have slugs which are different from the edition slug generated from the edition title. We want to ensure these do not change following the release of per-edition slugs and automatic slug updates based on the title.

To achieve this, we run a data migration which sets the slug override column on published and withdrawn editions to have the same value as the document slug, ensuring that the current live GOV.UK slug will be respected even if the edition is republished.

Prerequisite for [this JIRA ticket](https://gov-uk.atlassian.net/browse/WHIT-3063)


ChrisBAshton (approved):
LGTM 👍 nice one

PR #11345, opened 2026-04-09
Remove document slugs

ryanb-gds:
Follows deployment of https://github.com/alphagov/whitehall/pull/11315 and toggling of edition slugs feature flag to on in production

A few notable elements of this cleanup:

1. As agreed with content team, we are removing the `edit_slug` route
2. Landing pages now make use of the `slug_override` column for their custom slugs
3. Deleted the "export publications for document collections" rake task, as I don't remember ever using it and am pretty sure it already depended on an active record association that no longer exists. I also deleted the document reslugger rake task and the landing page one that extended it, as they're hopefully no longer required.
4. Updated the edition filter so it now searches for edition slugs. This does mean Whitehall search doesn't work as you'd expect for editions with overridden slugs, if this becomes a problem we could add another index for the slug_override column and an "OR" query but it doesn't seem worth it initially. We may also fix this if we move to the alternative "keep slug as default" approach
5. We stop adding the "deleted" prefix to the slugs of deleted editions, it makes more sense to just exclude deleted edition slugs when checking for existing slugs, which happens by default because of the default scope.
6. The bulk organisation updater and the republishing UI now rely on looking up the document via edition slugs. This should work because even if the latest edition has a slug override, there will be a superseded edition with a reference to the document that needs retagging/republishing.
7. For the similar slugs warning, we can just check if the generated slug ends with two dashes and one or more digits
8. I found some dead "published_as" "unpublished_as" methods, so removed those

Ideally these would all have been separate commits, but I approached this by adding `self.ignored_columns += [:slug]` to the `Document` model and fixing everything that broke, so it won't be terribly easy to work out which file changes apply to which change

Now that we have been in production for several days it seems safe to remove the feature flag too


app/models/edition.rb
```diff
@@ -410,7 +401,7 @@ def attribute_names
   end
 
   def base_path
-    url_slug = slug || id.to_param
+    url_slug = slug || document_id.to_param
```
ryanb-gds, line 404: This was always the wrong behaviour but never gets hit outside of tests as all edition classes override the `base_path` method

lauraghiorghisor-tw (approved):
LGTM

PR #11356, opened 2026-04-13
Remove slug column from documents table

ryanb-gds:
Tidying up following #11345

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,5 +1,4 @@
 class Document < ApplicationRecord
-  self.ignored_columns += [:slug]
   after_destroy :destroy_all_editions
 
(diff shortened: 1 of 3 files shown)


app/models/document.rb
```diff
@@ -1,5 +1,4 @@
 class Document < ApplicationRecord
-  self.ignored_columns += [:slug]
```
TonyGDS, line 2: Should this not be removed after the migration runs? Is there any chance a query could get cached with `:slug` in it?

ryanb-gds: It will be, migrations run before code is deployed

TonyGDS (approved):

PR #11358, opened 2026-04-14
Add `title` to `ImageUsage` [WHIT-3298]

patrickpatrickpatrick:
## What

- Add `title` to `ImageUsage` which returns either the `label` (if defined), `KEY image` or `image` (if an embeddable or default image).
- Add `articleize` function to use `a` or `an` correctly depending on label name
- Update `ImageUpload`, `ImageCard` and `LeadImageCard` to use `image_usage.title`
- Update `edition_images#new` to use `image_usage.title`

### Visual Differences
Screenshot 2026-04-20 at 11 24 38

Screenshot 2026-04-20 at 10 44 55

Screenshot 2026-04-14 at 10 45 14

Screenshot 2026-04-14 at 10 45 09

Screenshot 2026-04-14 at 10 44 56

## Why

Ensure consistency when referring to an image in the UI. Means that `label` does not have to be filled in each time an image is defined in the configuration of a document (unless `title` needs to be overridden). 

### DIFF
--- a/app/components/admin/edition_images/image_card_component.html.erb
+++ b/app/components/admin/edition_images/image_card_component.html.erb
@@ -1,7 +1,7 @@
 <%= render "govuk_publishing_components/components/summary_card", {
   id: "uploaded_#{image_usage.key}_image_card",
-  title: image_usage.label&.titleize || image_usage.key.titleize,
   summary_card_actions: summary_card_actions,
+  title: image_usage.title.upcase_first,
   rows: [
     {
       key: "Image",
--- a/app/components/admin/edition_images/image_upload_component.html.erb
+++ b/app/components/admin/edition_images/image_upload_component.html.erb
@@ -1,9 +1,10 @@
 <%
   image_kind = image_usage.kinds.first
+  image_usage_title = image_usage.title.downcase
 
-  hint_text = ["A #{image_kind.display_name} image must be at least <strong>#{image_kind.valid_width}x#{image_kind.valid_height}</strong>."]
+  hint_text = ["#{image_usage_title.articleize.upcase_first} must be at least <strong>#{image_kind.valid_width}x#{image_kind.valid_height}</strong>."]
   unless image_usage.multiple
-    hint_text << " If you are uploading more than one image, <a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos\">read the image guidance</a> on using unique file names."
+    hint_text << " If you are uploading more than one #{image_usage_title}, <a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos\">read the image guidance</a> on using unique file names."
   end
 %>
 
--- a/app/models/image_usage.rb
+++ b/app/models/image_usage.rb
@@ -21,4 +21,10 @@ class ImageUsage
   def lead?
     key == "lead"
   end
+
+  def title
+    return "image" if embeddable?
+
+    label || "#{key} image"
+  end
 end
--- a/lib/patches/string.rb
+++ b/lib/patches/string.rb
@@ -11,4 +11,8 @@ class String
       "#{self}’s"
     end
   end
+
+  def articleize
+    %w[a e i o u].include?(self[0].downcase) ? "an #{self}" : "a #{self}"
+  end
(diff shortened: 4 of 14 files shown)


app/components/admin/edition_images/image_upload_component.html.erb
```diff
@@ -1,7 +1,8 @@
 <%
   image_kind = image_usage.kinds.first
+  image_usage_label = @image_usage&.label || "image"
```
jamiestamp: I think this should be `image_usage` to match the rest of the template.

```suggestion
  image_usage_label = image_usage&.label || "image"
```

app/components/admin/edition_images/image_card_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= render "govuk_publishing_components/components/summary_card", {
   id: "uploaded_#{image_usage.key}_image_card",
-  title: image_usage.label&.titleize || image_usage.key.titleize,
+  title: image_usage.label&.upcase_first || image_usage.key.upcase_first,
```
jamiestamp: Question on this - I believe this is a pattern we use to get a label or key in a couple of places already -  `image_usage.label || image_usage.key`. I am wondering if this is worth creating a method on the image usage class that returns the text we want instead so it is easier to maintain and less logic in the view.

jamiestamp (changes requested):
Nice job, looks very close to being done! I have added a couple of comments and noticed in my testing that the text on the images page and image upload page on news story isn't quite right, or at least it doesn't look right to me.

## `image-workflow-related` branch
I just see 'Lead' and was expecting to see 'Lead Image'

Image

On the upload page it correctly lowercases the text, but again it just says 'lead' whereas I would expect 'lead image'

Image

## `main` branch

This is the same as the feature branch, it just says 'Lead' for the header

Image

The upload page is just saying 'Default' so the feature branch is a bit better, but it doesn't read right.

Image

---

We likely need to get guidance on what should be displayed on news story as well as part of fixing this, or perhaps spin off another ticket to look at news stories separately.

lib/patches/string.rb
```diff
@@ -11,4 +11,9 @@ def possessive
       "#{self}’s"
     end
   end
+
+  # https://stackoverflow.com/a/24987997
+  def articleize
```
ChrisBAshton, line 15: Nice. I did wonder why I hadn't come across this Rails helper before - and it's because it's not a Rails helper!

I'd move this to its own commit first, with corresponding tests - and delete the StackOverflow comment (can put that in the commit message instead).

app/models/image_usage.rb
```diff
@@ -21,4 +21,10 @@ def embeddable?
   def lead?
     key == "lead"
   end
+
+  def title
```
ChrisBAshton, line 25: This method creation (and corresponding tests) also feels worthy of its own commit (with 'Why' etc).

app/components/admin/edition_images/image_upload_component.html.erb
```diff
@@ -1,9 +1,10 @@
 <%
   image_kind = image_usage.kinds.first
+  image_usage_title = image_usage.title.downcase
 
-  hint_text = ["A #{image_kind.display_name} image must be at least <strong>#{image_kind.valid_width}x#{image_kind.valid_height}</strong>."]
+  hint_text = ["#{image_usage_title.articleize.upcase_first} must be at least <strong>#{image_kind.valid_width}x#{image_kind.valid_height}</strong>."]
   unless image_usage.multiple
-    hint_text << " If you are uploading more than one image, <a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos\">read the image guidance</a> on using unique file names."
+    hint_text << " If you are uploading more than one #{image_usage_title}, <a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/how-to-publish-on-gov-uk/images-and-videos\">read the image guidance</a> on using unique file names."
```
ChrisBAshton, line 7: ...then you'll be left with just these few logic changes in the third commit and it should be clearer to follow 👍

ChrisBAshton (commented):
Code LGTM 👍 Some suggestions on commit breakdown before merging. Thanks

ChrisBAshton (approved):
Nice one - thanks

PR #11360, opened 2026-04-14
Improve error message for social media [WHIT-3156]

GDSNewt:
## Overview
This update improves the clarity, consistency, and usability of validation errors when publishers add social media accounts.

## Why
Previously, validation feedback for social media accounts was unclear and inconsistent:
- Errors only appeared in the summary, not inline  
- It wasn’t obvious which field (Channel or URL) needed fixing  
- Terminology used “service” instead of “channel”  
- Only one error appeared even when multiple fields were invalid  

## What Changed
Validation behaviour and messaging have been enhanced:
- Terminology updated from **service → channel** to match the UI  
- Errors are now attached to **individual fields (Channel, URL)** instead of a shared attribute  
- Errors display **both inline and in the summary**  
- Validation runs **independently per field**  
- Multiple errors are shown when applicable (e.g. both fields blank)  
- Invalid fields are highlighted with a **red border**  

## ACs

Meets the ACs as described in the [Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?assignee=5caf4d888c2c1a75f3e46ef6&selectedIssue=WHIT-3156)

---

### 1. Both fields blank

Screenshot 2026-04-14 at 11 33 54

---

### 2. Channel selected, URL missing

Screenshot 2026-04-14 at 11 34 13

---

### 3. URL entered, channel missing

Screenshot 2026-04-14 at 11 34 25

---

### 4. Multiple accounts, mixed errors

Screenshot 2026-04-14 at 11 35 34

---

### DIFF
--- a/app/models/configurable_document_type.rb
+++ b/app/models/configurable_document_type.rb
@@ -113,6 +113,11 @@ class ConfigurableDocumentType
     end
   end
 
+  def title_for_attribute(attribute)
+    segments = attribute.split(".").reject { |s| s.match?(/\A\d+\z/) }
+    find_field_title(form, segments)
+  end
+
   def schema_for_fields(field_keys)
     field_keys = field_keys.map(&:to_s)
 
@@ -126,6 +131,21 @@ class ConfigurableDocumentType
   end
 
 private
+
+  def find_field_title(config, segments)
+    return config["title"] if segments.empty?
+    return nil unless config["fields"]
+
+    config["fields"].each_value do |field|
+      path = Array(field["attribute_path"]) - %w[block_content]
+      next if path.empty? || segments.first(path.length) != path
+
+      result = find_field_title(field, segments.drop(path.length))
+      return result if result
+    end
+
+    nil
+  end
 
   def validations_for_fields(field_keys)
     (@schema["validations"] || {}).each_with_object({}) do |(validation_name, options), result|
--- a/app/models/configurable_document_type.rb
+++ b/app/models/configurable_document_type.rb
@@ -113,6 +113,11 @@ class ConfigurableDocumentType
     end
   end
 
+  def title_for_attribute(attribute)
+    segments = attribute.split(".").reject { |s| s.match?(/\A\d+\z/) }
+    find_field_title(form, segments)
+  end
+
   def schema_for_fields(field_keys)
     field_keys = field_keys.map(&:to_s)
 
@@ -126,6 +131,19 @@ class ConfigurableDocumentType
   end
 
 private
+
+  def find_field_title(config, segments)
+    return config["title"] if segments.empty?
+    return nil unless config["fields"]
+
+    config["fields"].each_value do |field|
+      path = Array(field["attribute_path"]) - %w[block_content]
+      next if path.empty? || segments.first(path.length) != path
+
+      result = find_field_title(field, segments.drop(path.length))
+      return result if result
+    end
+  end
 
   def validations_for_fields(field_keys)
     (@schema["validations"] || {}).each_with_object({}) do |(validation_name, options), result|
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -167,6 +167,10 @@ class StandardEdition < Edition
                           .map { |block| block.path.validation_error_attribute }
   end
 
+  def self.human_attribute_name(attribute, options = {})
+    options[:base]&.type_instance&.title_for_attribute(attribute.to_s) || super
+  end
+
 private
 
   def field_paths(&block)
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -162,6 +162,10 @@ class StandardEdition < Edition
                           .map { |block| block.path.validation_error_attribute }
   end
 
+  def self.human_attribute_name(attribute, options = {})
+    options[:base]&.type_instance&.title_for_attribute(attribute.to_s) || super
+  end
+
 private
 
   def field_paths(&block)
--- a/app/models/standard_edition/block_content.rb
+++ b/app/models/standard_edition/block_content.rb
@@ -64,7 +64,13 @@ private
   end
 
   def method_missing(symbol, *args)
-    if attributes.class.instance_methods.include?(symbol)
+    # Errors on array item fields use dotted attribute names like
+    # :"social_media_links.0.url" so they can be targeted inline per
+    # field. Rails calls these as methods during error processing;
+    # returning nil here prevents a NoMethodError.
+    if symbol.to_s.include?(".")
+      nil
+    elsif attributes.class.instance_methods.include?(symbol)
       attributes.public_send(symbol, *args)
     else
       super
@@ -72,7 +78,7 @@ private
   end
 
   def respond_to_missing?(method_name, _include_all)
-    attributes.class.instance_methods.include?(method_name) || super
+    method_name.to_s.include?(".") || attributes.class.instance_methods.include?(method_name) || super
   end
 
   def attributes_class_for(attribute_config)
--- a/app/validators/social_media_links_validator.rb
+++ b/app/validators/social_media_links_validator.rb
@@ -1,7 +1,7 @@
 class SocialMediaLinksValidator < ActiveModel::Validator
   def initialize(opts = {})
     @attributes = opts[:attributes]
-    @service_field = opts[:fields]["service_field"]
+    @channel_field = opts[:fields]["service_field"]
     @url_field = opts[:fields]["url_field"]
     super
   end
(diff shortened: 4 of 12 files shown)


ryanb-gds (changes requested):
I think there's a much better solution available for this, which is to configure the validation for the sub fields rather than trying to use the "all in one" validator.

You'll need to keep the uniqueness validation at the level of the links object, but for the URL validation and the presence validation you should validate the individual fields. That'll save you from a lot of the complexity you've had to add to force this to work properly

test/unit/app/validators/social_media_links_validator_test.rb
```diff
@@ -72,18 +97,20 @@ class SocialMediaLinksValidatorTestClass
...
     block_content.social_media_links = [
-      { "social_media_service_name" => "Twiter", "url" => "http://facebook.com/govuk" },
+      { "social_media_service_name" => "Twitter", "url" => "http://facebook.com/govuk" },
       { "social_media_service_name" => "Facebook", "url" => "http://facebook.com/govuk" },
     ]
     @validator.validate(block_content)
 
-    assert_equal ["Social media links already has an account with a URL of \"http://facebook.com/govuk\"."], block_content.errors.full_messages
+    assert_empty block_content.errors[:"social_media_links.0.url"]
+    assert_includes block_content.errors[:"social_media_links.1.url"], "must be unique"
```
ChrisBAshton: Throughout, I'd use `"social_media_links.0.url".to_sym` rather than the `:"` approach which is a bit less idiomatic I think.

test/unit/app/validators/social_media_links_validator_test.rb
```diff
@@ -14,6 +14,18 @@ class SocialMediaLinksValidatorTest < ActiveSupport::TestCase
...
+    # Rails tries to call these as methods, which would raise NoMethodError since dots
+    # aren't valid in Ruby method names. BlockContent handles this in production; we
+    # replicate it here.
+    def method_missing(method_name, *args)
+      method_name.to_s.include?(".") ? nil : super
+    end
+
+    def respond_to_missing?(method_name, include_private = false)
+      method_name.to_s.include?(".") || super
+    end
```
ChrisBAshton: I'm not loving this repetition of monkeypatching in several places in the codebase. Can we not hook in and convert `test_array_attribute.0.social_media_service_id` to something like `test_array_attribute__0__social_media_service_id` so we can avoid having to do this?

app/models/configurable_document_type.rb
```diff
@@ -113,6 +113,11 @@ def dynamic_tabs
     end
   end
 
+  def title_for_attribute(attribute)
```
ChrisBAshton, line 116: Please pull this into its own commit and add an associated unit test for it.

app/models/standard_edition.rb
```diff
@@ -162,6 +162,10 @@ def error_field_order
                           .map { |block| block.path.validation_error_attribute }
   end
 
+  def self.human_attribute_name(attribute, options = {})
+    options[:base]&.type_instance&.title_for_attribute(attribute.to_s) || super
```
ChrisBAshton, line 171: Please pull this into its own commit and add an associated unit test for it.

app/validators/social_media_links_validator.rb
```diff
@@ -9,61 +9,60 @@ def initialize(opts = {})
...
+
       arr.each_with_index do |social_media_account, index|
         channel_name = social_media_account[@channel_field]
-        if validate_social_media_channel(channel_name, record, attribute_name, index)
-          validate_social_media_link(social_media_account[@url_field], channel_name, record, attribute_name)
-        end
+        url = social_media_account[@url_field]
+
+        validate_social_media_channel(channel_name, channels_seen, index, record, attribute_name)
+        validate_social_media_url(url, urls_seen, index, record, attribute_name)
```
ChrisBAshton: I think rather than re-injecting the `channels_seen` / `urls_seen` values as method args, I'd have them as variables (`@channels_seen`...) and manage the state that way.

ChrisBAshton (commented):
I'm as keen to get this over the line as you are :D let's see if we can remove some of the code smells first - comments below.

app/models/configurable_document_type.rb
```diff
@@ -127,6 +132,21 @@ class NotFoundError < StandardError
 
 private
 
+  def find_field_title(config, segments)
+    return config["title"] if segments.empty?
```
ChrisBAshton, line 136: Looks like we're missing a test for this behaviour ^

app/models/configurable_document_type.rb
```diff
@@ -127,6 +132,21 @@ class NotFoundError < StandardError
...
+    return nil unless config["fields"]
+
+    config["fields"].each_value do |field|
+      path = Array(field["attribute_path"]) - %w[block_content]
```
ChrisBAshton, line 140: I'd expect some reference to `block_content` in the unit tests 🤔

app/models/configurable_document_type.rb
```diff
@@ -127,6 +132,21 @@ class NotFoundError < StandardError
...
+      next if path.empty? || segments.first(path.length) != path
+
+      result = find_field_title(field, segments.drop(path.length))
```
ChrisBAshton, line 143: This method seems a bit over-engineered - the recursion is only necessary if `field_attribute` could appear any arbitrary level of nesting deep, but in reality we're always expecting it to be at the `forms.{tab_name}.fields[n]` level, aren't we?

test/unit/app/validators/social_media_links_validator_test.rb
```diff
@@ -15,6 +13,7 @@ class SocialMediaLinksValidatorTest < ActiveSupport::TestCase
 
   class SocialMediaLinksValidatorTestClass
     include ActiveModel::API
+    include WithNestedAttributeErrors
```
ChrisBAshton, line 16: This is introduced in the `Add SocialMediaLinksValidator` commit but the validator doesn't exist yet.

test/components/admin/error_summary_component_test.rb
```diff
@@ -153,6 +153,16 @@ class Admin::ErrorSummaryComponentTest < ViewComponent::TestCase
...
     assert_equal third_link[:href], "#labelled_error_summary_test_object_date"
   end
+
+  test "renders full_message for dotted attributes" do
+    object = DottedAttributeErrorSummaryTestObject.new("title", Time.zone.today)
+    object.errors.add("social_media_links.0.url".to_sym, :blank, message: "cannot be blank")
+    render_inline(Admin::ErrorSummaryComponent.new(object:))
+
+    link = page.find(".gem-c-error-summary__list-item a")
+    assert_equal "Social media links 0 url cannot be blank", link.text
```
ChrisBAshton: This is accurate, but - will editors know about zero-based indexes? 🤔 or should the message be `Social media link number 1: URL cannot be blank`?

test/support/with_nested_attribute_errors.rb
ChrisBAshton, line 1: Move this commit earlier (before the module is first referenced) and add corresponding tests?

app/models/standard_edition/block_content.rb
```diff
@@ -64,15 +64,21 @@ def run_schema_validations
   end
 
   def method_missing(symbol, *args)
-    if attributes.class.instance_methods.include?(symbol)
+    # Errors on array item fields use dotted attribute names like
+    # :"social_media_links.0.url" so they can be targeted inline per
+    # field. Rails calls these as methods during error processing;
+    # returning nil here prevents a NoMethodError.
```
ChrisBAshton, line 70: Did you consider my [suggestion](https://github.com/alphagov/whitehall/pull/11360#discussion_r3152502201) that we switch from `test_array_attribute.0.social_media_service_id` to something like `test_array_attribute__0__social_media_service_id` so that we don't need to fight Rails's default method name construction? (Could do it in a refactor commit at the end if you want to avoid a painful rebase)

ChrisBAshton (commented):
Getting there 👏

app/models/configurable_document_type.rb
GDSNewt: I think the recursion is needed because field_attribute doesn’t always live directly at forms.{tab_name}.fields[n]. In topical_event.json, for example, social_media_links contains nested fields like url, each with their own attribute_path.

So for something like social_media_links.0.url, we have to traverse two levels (social_media_links → url) to resolve the title. The recursion is just the simplest way to handle this existing two-level nesting.

Happy to simplify if there’s a flatter approach, but given the current structure, we can’t rely on everything being at a single level.

app/models/standard_edition/block_content.rb
ChrisBAshton: I suggested this on the assumption that it means you don't need to define these `method_missing` / `respond_to_missing` method definitions...? (On the basis that method names can't include a `.` in them, but can include a `__`).

If all this suggestion does is a find-and-replace across the code then it's not a useful suggestion 😅 Did you check what else could be cleaned up as a result? 🙏

app/models/configurable_document_type.rb
ChrisBAshton: I'm not sure we're testing this behaviour yet? `segments` is "field_attribute" in your first unit test. The only situation I see this line returning anything is if we basically call `title_for_attribute(nil)`, which shouldn't ever be a thing, right? Can we just delete this line?

app/models/configurable_document_type.rb
```diff
@@ -127,6 +132,21 @@ class NotFoundError < StandardError
 
 private
 
+  def find_field_title(config, segments)
+    return config["title"] if segments.empty?
+    return nil unless config["fields"]
+
+    config["fields"].each_value do |field|
+      path = Array(field["attribute_path"]) - %w[block_content]
+      next if path.empty? || segments.first(path.length) != path
```
ChrisBAshton, line 141: Can you explain what the code after the `||` is doing? Again, I'm not sure it's needed 🤔

app/models/configurable_document_type.rb
ChrisBAshton: Oh yes - d'uh! Thanks!

test/unit/app/models/configurable_document_type_test.rb
```diff
@@ -1,6 +1,76 @@
...
+        },
+      },
+    })
+    ConfigurableDocumentType.setup_test_types(type_config)
+    document_type = ConfigurableDocumentType.find("test_type")
+
+    assert_equal "Inner Field Title", document_type.title_for_attribute("outer_field.0.inner_field")
+  end
+
+  test "#title_for_attribute finds a field whose attribute_path includes block_content" do
```
ChrisBAshton:
```suggestion
  test "#title_for_attribute ignores 'block_content' in the attribute_path because..." do
```

...because why? I assume it's because we want to avoid surfacing the words "block content" anywhere in the UI...?

app/models/standard_edition/block_content.rb
ChrisBAshton: From a quick look, I'm hoping the move to `__` means we can delete the WithNestedAttributeErrors helper entirely?

test/unit/app/models/standard_edition/block_content_test.rb
```diff
@@ -177,6 +177,26 @@ class BlockContentTest < ActiveSupport::TestCase
 
     page.attributes = { "test_array_attribute" => [{ "social_media_service_id" => "1", "url" => "broken url" }] }
     assert_not page.valid?
-    assert_not page.errors.where("test_array_attribute", :invalid_social_media_link).empty?
+    assert_not page.errors.where(:"test_array_attribute.0.url", :invalid).empty?
```
ChrisBAshton:
```suggestion
    assert_not page.errors.where("test_array_attribute.0.url".to_sym, :invalid).empty?
```

and elsewhere

app/models/standard_edition/block_content.rb
GDSNewt: So I don't think it actually eliminates the need for method_missing guard or WithNestedAttributeErrors — Rails calls the attribute as a method on the model during error processing regardless of whether the separator is . or __, so the guard is needed either way. 

I think it ends up being a pure find-and-replace. 

One small improvement we could make as a follow-up: BlockContent currently has its own inline copy of the method_missing/respond_to_missing? logic, while WithNestedAttributeErrors has an identical copy for use in test doubles. We could de-duplicate by moving the helper to concerns and having BlockContent include it, rather than defining it inline. Wanted to check you were happy with that first, given it's making WithNestedAttributeErrors even more needed than before 😬

app/models/configurable_document_type.rb
GDSNewt: I tried removing it, and it breaks several tests (including the one that expects "Test Attribute"), because the final recursive call with [] no longer returns the title.

For title_for_attribute("field_attribute.0"), we remove the index and get ["field_attribute"], match the field, then recurse with []. That recursive call hits this line because segments.empty? is true, and that’s where we return the title.

So it’s not just for nil input—it’s the stop condition for valid lookups as well. Without it, the recursion wouldn’t know when to stop and would end up returning nil.

I don’t think we can safely delete it for that reason, even though it’s not obvious from the tests. We could perhaps add a test for the empty case?

app/models/configurable_document_type.rb
GDSNewt: it ensures we only recurse into a field if its attribute_path matches the start of the segments we're looking up.

Without it, every field would match regardless and we'd always return the first field's title. 

I'm adding a test to demonstrate this: looking up "body" in a config with two fields (social_media_links defined first, body second).

app/models/configurable_document_type.rb
ChrisBAshton: Yep I'd add a test for the empty case 👍

test/unit/app/models/standard_edition/block_content_test.rb
ChrisBAshton: There are still 12 places where we use `:""` instead of `"".to_sym` (including the above example).
But looking around, it seems `:""` is the idiomatic choice - just a bit jarring to my eyes!
Happy to leave.

app/models/standard_edition/block_content.rb
ChrisBAshton: OK - thanks for removing the "." to "__" replacement.

De-duplicating feels like a good move 👍

ChrisBAshton (approved):
Have spent half an hour spiking trying to remove the dot notation. It should be doable but wasn't doable in that time! Happy to proceed and maybe we can have a backlog item to explore refactoring it back out again later.

A couple of comments above where we've agreed to add a couple more things ^ but otherwise this is good to go.

PR #11361, opened 2026-04-14
Ensure landing page body always extends latest edition [WHIT-3327]

ryanb-gds:
Fixes https://gov-uk.atlassian.net/browse/WHIT-3327

### DIFF
--- a/app/models/landing_page/body.rb
+++ b/app/models/landing_page/body.rb
@@ -52,7 +52,7 @@ class LandingPage::Body
     @body_to_extend ||= begin
       return if extends.blank?
 
-      edition = LandingPage.find_by(slug_override: extends)
+      edition = LandingPage.where(slug_override: extends).last
       return if edition.nil?
 
       parsed_body = begin


TonyGDS (approved):

app/models/landing_page/body.rb
```diff
@@ -52,7 +52,7 @@ def body_to_extend
     @body_to_extend ||= begin
       return if extends.blank?
 
-      edition = LandingPage.find_by(slug_override: extends)
+      edition = LandingPage.where(slug_override: extends).last
```
TonyGDS, line 55: do we have test coverage to catch this?

ryanb-gds: Well, the line is covered but the query behaviour isn't - as long as an edition is returned the tests will pass. I never feel great about testing SQL queries by example, it's always a lot of effort and the result is never exhaustive. I'm quite interested in maybe applying snapshot or property-based testing to SQL queries generated by ActiveRecord, but that's a scheme for another day 😅

PR #11366, opened 2026-04-15
Remove `RenderedAttachmentsHelper` [WHIT-1922]

patrickpatrickpatrick:
## What

- Remove `RenderedAttachmentsHelper`
- Remove fields referencing output of `RenderedAttachmentsHelper` from Publishing API payload presenters (`details.documents`, `details.final_outcome_documents`, `details.public_feedback_documents`) that are no longer referenced by frontend rendering applications
- Use `locale` when building `details.attachments`, preserve previous behaviour of attachments and update tests

## Why

Frontend rendering applications now render attachments themselves instead of using pre-rendered markup delivered by Whitehall. Therefore we can remove the `RenderedAttachmentsHelper` as well as the fields in the payload that contained the output of this helper.


app/presenters/publishing_api/payload_builder/attachments.rb
```diff
@@ -19,7 +19,10 @@ def call
       def attachments
         items.flat_map do |item|
           if item
-            item.attachments_ready_for_publishing.map(&:publishing_api_details)
+            attachments = item.attachments_ready_for_publishing
+            # nil/"" locale should always be returned
+            locales_that_match = [I18n.locale.to_s, ""]
+            attachments.to_a.select { |attachment| locales_that_match.include?(attachment.locale.to_s) }.map(&:publishing_api_details)
```
ChrisBAshton, line 25: This commit seems a bit rushed - no commit message explaining the 'why', no corresponding test to ensure we don't accidentally regress, etc. Please revisit 🙏

app/helpers/presenters/publishing_api/featured_documents_helper.rb
ChrisBAshton, line 1: Great to delete all this code 🚀 Would be good to amend the commit message with the 'why', so a future dev doesn't have to navigate to the PR itself to find out why.

app/presenters/publishing_api/payload_builder/attachments.rb
ChrisBAshton: Thanks, the revised commit message helps a lot. So this first commit is essentially moving this bit from the Publication Presenter:

```
    def documents
      return [] unless item.attachments.any?

      render_attachments(attachments_for_current_locale)
    end
```

Makes sense 👍

ChrisBAshton, line 25: I still don't really get why we have some attachments that have a nil or empty string locale 🤔 but I can see from the `test "nil locale attachments are present for all locales" do` test that that pre-dates this PR, so out of scope for investigation.

ChrisBAshton (approved):
LGTM ⭐

PR #11371, opened 2026-04-16
[WHIT-3323] Store accurate slug value

lauraghiorghisor-tw:
## What & why

We have recently introduced a third column for storing slug-related data, `slug_from_title`, and set its value to the value of the title-based `slug`. This PR's migration and adjacent changes will allow us to make use of this new column - the `slug_from_title` column will only hold the title-based slug, whilst the `slug` will store the "accurate" slug value.

We have previously backfilled `slug_from_title` values for pre-publication editions in https://github.com/alphagov/whitehall/pull/11359. For live editions, we have run [a distinct data migration](https://github.com/alphagov/whitehall/pull/11424) to set the `slug_from_title` to the historical value of the `slug`, where the slug wasn't already set to the value of the override. This ensures we keep a record of the slug column, before repurposing it to store the "accurate" slug value with this commit's migration.

This PR includes the following changes:
1. Add migration to set `slug` to `slug_override`, if override is present
- This migration prepares the data. It sets the value of the `slug` column to the value of the `slug_override`, if the latter is present, for all edition states (bar superseded and deleted). We will be running this as a regular migration to ensure that no mismatched data gets created between the migration run and the code merge.

2. Update callbacks for slug generation:
- set `slug_from_title` on title updates
- set `slug` to either `slug_override` or `slug_from_title`

3. Correct remaining usages of `slug`, `slug_override` and `slug_from_title`
This wraps up the changes whereby we now have distinct usages for our three slug-related columns:
- `slug_from_title` holds the title-based slug
- `slug_override` holds a custom slug, different from the title-based
- `slug` holds the accurate value between the other two fields, `slug_override` if override is present, otherwise `slug_from_title`.
The only place where we use the `slug_from_title` is now the override checkbox in the `_page_address_controls.html.erb` view.

## Test plan

Types coverage:
1. ✅ Publication (as a regular edition, no support for non-en primary)
2. ✅ News story (as a regular standard edition, english primary locale)
3. ✅ World news story (non-en primary support, plus legacy data for supported locales)
4. ✅ Collection (non-en edition)
5. ✅ CIPs
6. ✅ LPs

Data combinations that we need to look into:
- ✅ published with no SO
- ✅ published with SO
- ✅ published with no SO and slug is not title based e.g. suffix
- ✅ draft with SO
- ✅ draft with no SO
- ✅ withdrawn with SO
- ✅ withdrawn without SO
- ✅ withdrawn no SO and slug not title-based
- ✅ unpublished with SO
- ✅ unpublished without SO
- ✅ unpublished with a deleted prefix
- ✅ scheduled (already scheduled and newly scheduled, with/without SO)
- ✅ blank SO
- ✅ translations
- ✅ New draft for regular edition 
- ✅ New draft for config driven edition 
- ✅ New draft for editions of non-en primary with a babosa supported locale
- ✅ New draft for editions of non-en primary with a babosa non-supported locale

Test sequence: 
Live states:
- Republish
- Make new draft from live
- Does checkbox look correct?
- Save with "keep live slug" checked; preview
- Save with "keep new slug" checked; preview
- Revert to previous checkbox option and save; preview
- Publish; preview
- Unpublish/Withdraw; preview

Existing Draft From Published (created before migration):
- Change title, save, preview
- Does checkbox look correct?
- Save with "keep live slug" checked; preview 
- Save with "keep new slug" checked; preview
- Revert to previous checkbox option and save; preview
- Publish

Create new document:
- Change title, save, preview
- Does checkbox look correct?
- Publish
- Redraft
- Does checkbox look correct?
- Save with "keep live slug" checked; preview
- Save with "keep new slug" checked; preview
- Revert to previous checkbox option and save; preview
- Publish

### DIFF
--- a/db/migrate/20260503161913_set_slug_to_slug_override_when_slug_override_present.rb
+++ b/db/migrate/20260503161913_set_slug_to_slug_override_when_slug_override_present.rb
@@ -1,0 +1,27 @@
+class SetSlugToSlugOverrideWhenSlugOverridePresent < ActiveRecord::Migration[8.1]
+  def change
+    find_slug_overrides_sql = <<-SQL
+      SELECT e.id
+      FROM editions e
+      WHERE e.slug_override IS NOT NULL
+      AND e.slug_override != ''
+      AND e.slug != e.slug_override
+      AND e.state NOT IN ('deleted', 'superseded');
+    SQL
+
+    edition_ids = ActiveRecord::Base.connection.select_values(find_slug_overrides_sql)
+    Rails.logger.info "Found #{edition_ids.size} editions where the slug does not match the slug_override."
+
+    update_sql = <<-SQL
+      UPDATE editions e
+      SET e.slug = e.slug_override
+      WHERE e.slug_override IS NOT NULL
+      AND e.slug_override != ''
+      AND e.slug != e.slug_override
+      AND e.state NOT IN ('deleted', 'superseded');
+    SQL
+
+    updated_editions = ActiveRecord::Base.connection.update(update_sql)
+    Rails.logger.info "\nMigration complete. Slug set to the value of the slug_override for #{updated_editions} editions."
+  end
+end
--- a/test/unit/app/models/edition/identifiable_test.rb
+++ b/test/unit/app/models/edition/identifiable_test.rb
@@ -64,6 +64,8 @@ class Edition::IdentifiableTest < ActiveSupport::TestCase
 end
 
 class Edition::SluggingTest < ActiveSupport::TestCase
+  extend Minitest::Spec::DSL
+
   class SluggableEdition < Edition
     def self.create!(attributes)
       super({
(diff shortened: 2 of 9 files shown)


test/unit/app/models/edition/identifiable_test.rb
```diff
@@ -183,4 +214,53 @@ def string_for_slug
...
+    test "slug can change for post-publication editions when saved with validate: false, if slug_from_title changed" do
+      published_edition = SluggableEdition.create_published!(title: "Original Title", slug: "original-title", slug_from_title: "original-title")
+      assert_equal "original-title", published_edition.slug
+      assert_equal "original-title", published_edition.slug_from_title
+
+      published_edition.slug_from_title = "new-title" # implicitly slug_from_title would change if the title is changed
+      published_edition.save!(validate: false)
+
+      assert_equal "new-title", published_edition.reload.slug
+      assert_equal "new-title", published_edition.reload.slug
```
TonyGDS: remove this duplicate assertion

test/factories/editions.rb
```diff
@@ -89,6 +95,8 @@
       force_published { true }
       published_major_version { 1 }
       published_minor_version { 0 }
+      slug_from_title { title.to_slug.normalize(to_ascii: true, max_length: 150).to_s if title.present? }
```
TonyGDS: maybe we can create a `with_slug` trait and assign it for the post-published editions

db/migrate/20260503161800_backfill_slug_from_title_on_editions.rb
```diff
@@ -0,0 +1,7 @@
+class BackfillSlugFromTitleOnEditions < ActiveRecord::Migration[8.1]
+  def change
+    Edition.where(slug_override: nil, state: %w[published withdrawn unpublished]).in_batches do |batch|
```
TonyGDS: should this also filter out where `slug_from_title` is not nil?

db/migrate/20260503161913_set_slug_to_slug_override_when_slug_override_present.rb
```diff
@@ -0,0 +1,23 @@
+class SetSlugToSlugOverrideWhenSlugOverridePresent < ActiveRecord::Migration[8.1]
+  def change
+    find_slug_overrides_sql = <<-SQL
+      SELECT e.id, COUNT(*)#{' '}
```
TonyGDS: can remove count here

TonyGDS (approved):
LGTM :shipit:

PR #11381, opened 2026-04-20
Validate social media link fields individually [WHIT-3156]

GDSNewt:
## Overview
This update improves the clarity, consistency, and usability of validation errors when publishers add social media accounts.

## Why
Previously, validation feedback for social media accounts was unclear and inconsistent:
- Errors only appeared in the summary, not inline  
- It wasn’t obvious which field (Channel or URL) needed fixing  
- Terminology used “service” instead of “channel”  
- Only one error appeared even when multiple fields were invalid  

## What Changed
Validation behaviour and messaging have been enhanced:
- Terminology updated from **service → channel** to match the UI  
- Errors are now attached to **individual fields (Channel, URL)** instead of a shared attribute  
- Errors display **both inline and in the summary**  
- Validation runs **independently per field**  
- Multiple errors are shown when applicable (e.g. both fields blank)  
- Invalid fields are highlighted with a **red border**  
- Presence and URL format validation are now declared in the document type schema and run per-item, replacing bespoke logic in SocialMediaLinksValidator which is now scoped to uniqueness checking only.

## ACs

Meets the ACs as described in the [Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?assignee=5caf4d888c2c1a75f3e46ef6&selectedIssue=WHIT-3156)

---

| # | Channel| URL  | Expected Channel Error | Expected URL Error | Summary Box | Red Border |
|---|----------------|------------|------------------------|-------------------|-------------|------------|
| 1 | Blank | Blank | Channel cannot be blank | URL cannot be blank | Both errors | Both fields |
| 2 | Selected | Blank | — | URL cannot be blank | URL error | URL field |
| 3 | Blank | Valid | Channel cannot be blank | — | Channel error | Channel field |
| 4 | Selected | Invalid (e.g. `foo`) | — | Invalid URL - use the full URL, including `https://` | URL error | URL field |
| 5 | Blank | Invalid (e.g. `foo`) | Channel cannot be blank | Invalid URL - use the full URL, including `https://` | Both errors | Both fields |
| 6 | Duplicate (e.g. Facebook already exists) | Valid | Channel must be unique | — | Channel error | Channel field |
| 7 | Different | Duplicate URL (already exists) | — | URL must be unique | URL error | URL field |
| 8 | Selected | Valid | — | — | Happy path (no errors) | None |
| 9 | Both “Other” with different URLs | Both valid | — | — | Happy path (no errors) | None |

### DIFF
--- a/app/models/configurable_content_blocks/default_array.rb
+++ b/app/models/configurable_content_blocks/default_array.rb
@@ -29,6 +29,20 @@ module ConfigurableContentBlocks
       end
     end
 
+    def each_leaf(&block)
+      return enum_for(:each_leaf) unless block_given?
+
+      items.each_with_index do |_item, index|
+        field_blocks(index).each do |field_block|
+          if field_block.respond_to? :each_leaf
+            field_block.each_leaf(&block)
+          else
+            block.call(field_block)
+          end
+        end
+      end
+    end
+
   private
 
     def template_name
--- a/app/models/configurable_document_types/topical_event.json
+++ b/app/models/configurable_document_types/topical_event.json
@@ -131,6 +131,14 @@
           "url": {
             "type": "string"
           }
+        },
+        "validations": {
+          "presence": {
+            "attributes": ["social_media_service_name"]
+          },
+          "uri": {
+            "attributes": ["url"]
+          }
         }
       }
     },
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -146,23 +146,26 @@ class StandardEdition < Edition
 
   def error_labels
     {}.tap do |labels|
-      ConfigurableContentBlocks::DefaultObject.root_block_for(self, "documents").each do |block|
-        # we only want labels for leaf fields, so do nothing if this field isn't one
-        next if block.respond_to? :field_blocks
-
+      all_leaf_blocks.each do |block|
         labels[block.path.validation_error_attribute] = block.title
       end
     end
   end
 
   def error_field_order
-    %w[title summary] + ConfigurableContentBlocks::DefaultObject
-                          .root_block_for(self, "documents")
-                          .each_leaf
-                          .map { |block| block.path.validation_error_attribute }
+    %w[title summary] + all_leaf_blocks.map { |block| block.path.validation_error_attribute }
   end
 
 private
+
+  def all_leaf_blocks
+    type_instance.dynamic_tabs
+      .map { |tab| tab["id"] }
+      .prepend("documents")
+      .flat_map do |tab_key|
+        ConfigurableContentBlocks::DefaultObject.root_block_for(self, tab_key).each_leaf.to_a
+      end
+  end
 
   def field_paths(&block)
     [].tap do |paths|
--- a/app/models/standard_edition/block_content.rb
+++ b/app/models/standard_edition/block_content.rb
@@ -50,9 +50,7 @@ class StandardEdition::BlockContent
 private
 
   def run_schema_validations
-    return unless @schema.key?("validations")
-
-    @schema["validations"].each do |key, options|
+    @schema["validations"]&.each do |key, options|
       validator_class = VALIDATORS[key] or raise ArgumentError, "undefined validator type #{key}"
       opts = options.symbolize_keys
 
@@ -60,6 +58,20 @@ private
       next if on && validation_context&.to_sym != on.to_sym
 
       validator_class.new(opts).validate(self)
+    end
+
+    @schema["attributes"]&.each do |attr_name, attr_schema|
+      next unless attr_schema["attributes"]
+
+      item_class = attributes_class_for(attr_schema["attributes"], attr_schema["validations"])
+      (send(attr_name.to_sym) || []).each_with_index do |item, index|
+        item_instance = item_class.new(item)
+        next if item_instance.valid?
+
+        item_instance.errors.each do |error|
+          errors.import(error, attribute: "#{attr_name}.#{index}.#{error.attribute}".to_sym)
+        end
+      end
     end
   end
 
@@ -75,7 +87,7 @@ private
     attributes.class.instance_methods.include?(method_name) || super
   end
 
-  def attributes_class_for(attribute_config)
+  def attributes_class_for(attribute_config, validations_config = nil)
     attributes_class = Class.new do
       include ActiveModel::API
       include ActiveModel::Attributes
@@ -98,6 +110,12 @@ private
         attributes
       end
     end
+
+    validations_config&.each do |validator_key, options|
+      attrs = Array(options["attributes"]).map(&:to_sym)
+      attributes_class.validates(*attrs, validator_key.to_sym => true)
+    end
+
     attributes_class.set_temporary_name("Block content attributes")
     attributes_class
   end
--- a/app/validators/social_media_links_validator.rb
+++ b/app/validators/social_media_links_validator.rb
@@ -1,7 +1,7 @@
 class SocialMediaLinksValidator < ActiveModel::Validator
   def initialize(opts = {})
     @attributes = opts[:attributes]
-    @service_field = opts[:fields]["service_field"]
+    @channel_field = opts[:fields]["service_field"]
     @url_field = opts[:fields]["url_field"]
     super
   end
@@ -10,7 +10,7 @@ class SocialMediaLinksValidator < ActiveModel::Validator
     @attributes.each do |attribute_name|
       arr = record.send(attribute_name.to_sym) || []
       arr.each_with_index do |social_media_service, index|
-        service_name = social_media_service[@service_field]
+        service_name = social_media_service[@channel_field]
         if validate_social_media_service(service_name, record, attribute_name, index)
           validate_social_media_link(social_media_service[@url_field], service_name, record, attribute_name)
         end
--- a/app/validators/social_media_links_validator.rb
+++ b/app/validators/social_media_links_validator.rb
@@ -9,68 +9,25 @@ class SocialMediaLinksValidator < ActiveModel::Validator
   def validate(record)
     @attributes.each do |attribute_name|
       arr = record.send(attribute_name.to_sym) || []
-      arr.each_with_index do |social_media_service, index|
-        service_name = social_media_service[@channel_field]
-        if validate_social_media_service(service_name, record, attribute_name, index)
-          validate_social_media_link(social_media_service[@url_field], service_name, record, attribute_name)
-        end
+
+      duplicate_values(arr, @channel_field, exclude: "Other").each do |channel|
+        record.errors.add(attribute_name.to_sym, :taken, message: "already has an account with a channel of \"#{channel}\"")
+      end
+
+      duplicate_values(arr, @url_field).each do |url|
+        record.errors.add(attribute_name.to_sym, :taken, message: "already has an account with a URL of \"#{url}\"")
       end
     end
   end
 
 private
 
-  def validate_social_media_service(service_name, record, attribute_name, index)
-    @services ||= []
-
-    if service_name.blank?
-      record.errors.add(
-        attribute_name.to_sym,
-        :invalid_social_media_link,
-        message: "contains an account (\"Social media account #{index + 1}\") without a service selected.",
-      )
-      return false
-    end
-
-    if service_name != "Other" && @services.include?(service_name)
-      record.errors.add(
-        attribute_name.to_sym,
-        :invalid_social_media_link,
-        message: "contains another account with a service of \"#{service_name}\".",
-      )
-      return false
-    end
-    @services << service_name
-  end
-
-  def validate_social_media_link(url, service_name, record, attribute_name)
-    if url.blank?
-      record.errors.add(
-        attribute_name.to_sym,
-        :invalid_social_media_link,
-        message: "contains a \"#{service_name}\" account without a URL.",
-      )
-    elsif !valid_url?(url)
-      record.errors.add(
-        attribute_name.to_sym,
-        :invalid_social_media_link,
-        message: "contains a \"#{service_name}\" account with an invalid URL - use the full URL, including https://",
-      )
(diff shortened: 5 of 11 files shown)


app/validators/social_media_links_validator.rb
```diff
@@ -1,76 +1,37 @@
 class SocialMediaLinksValidator < ActiveModel::Validator
   def initialize(opts = {})
     @attributes = opts[:attributes]
-    @service_field = opts[:fields]["service_field"]
+    @channel_field = opts[:fields]["service_field"]
     @url_field = opts[:fields]["url_field"]
     super
   end
 
   def validate(record)
```
patrickpatrickpatrick, line 9: I think this could be simplified a bit. Looking into it, it's possible to the get duplicated items in an array with the following: 

```
array.group_by{ |e| e }.select { |k, v| v.size > 1 }.map(&:first)
```

this could be used in a method which you could call for each field.

app/validators/social_media_links_validator.rb
```diff
@@ -1,76 +1,37 @@
 class SocialMediaLinksValidator < ActiveModel::Validator
   def initialize(opts = {})
     @attributes = opts[:attributes]
-    @service_field = opts[:fields]["service_field"]
+    @channel_field = opts[:fields]["service_field"]
```
ChrisBAshton, line 4: I'd start with this commit. First commit could be a rename from service_field to channel_field (and explaining why). Includes updating the corresponding tests etc.

public/configurable-document-type.schema.json
```diff
@@ -33,6 +33,7 @@
           "presence",
           "safe_html",
           "valid_internal_path_links",
+          "uri",
```
ChrisBAshton, line 36: Second commit could be expanding the schema of schemas here to add uri as a validator (plus explaining why).

app/models/standard_edition/block_content.rb
```diff
@@ -75,17 +87,14 @@ def respond_to_missing?(method_name, _include_all)
...
       include ActiveModel::API
       include ActiveModel::Attributes
       include ActiveModel::Serializers::JSON
 
       attribute_config.each do |key, property_schema|
-        case property_schema["type"]
-        when "object"
-          attribute key
-        when "array"
+        if %w[object array].include?(property_schema["type"])
```
ChrisBAshton: This is the kind of drive-by refactor an LLM tends to do...?
It doesn't seem necessary for the changes we're making (though is ultimately harmless here).
But it's one more bit of context for the reviewer to have to read through. I'd avoid making the change unless it's clearly better, which I'd argue is not the case here.

app/models/standard_edition/block_content.rb
```diff
@@ -61,6 +59,20 @@ def run_schema_validations
...
+    @schema["attributes"]&.each do |attr_name, attr_schema|
+      next unless attr_schema["attributes"]
+
+      item_class = attributes_class_for(attr_schema["attributes"], attr_schema["validations"])
+      (send(attr_name.to_sym) || []).each_with_index do |item, index|
+        item_instance = item_class.new(item)
+        next if item_instance.valid?
+
+        item_instance.errors.each do |error|
+          errors.import(error, attribute: :"#{attr_name}.#{index}.#{error.attribute}")
```
ChrisBAshton: I think we tend to prefer an explicit `to_sym` casting rather than rely on reading `:"` properly.

```suggestion
          errors.import(error, attribute: "#{attr_name}.#{index}.#{error.attribute}".to_sym)
```

app/models/standard_edition.rb
```diff
@@ -146,24 +146,27 @@ def permitted_image_usages
...
-    %w[title summary] + ConfigurableContentBlocks::DefaultObject
-                          .root_block_for(self, "documents")
-                          .each_leaf
-                          .map { |block| block.path.validation_error_attribute }
+    %w[title summary] + all_leaf_blocks.map { |block| block.path.validation_error_attribute }
   end
 
 private
 
+  def all_leaf_blocks
```
ChrisBAshton, line 161: I'd put the creation of this method, and of the `each_leaf` method in default_array.rb, into its own commit, explaining clearly why it's needed.

app/models/standard_edition/block_content.rb
```diff
@@ -61,6 +59,20 @@ def run_schema_validations
 
       validator_class.new(opts).validate(self)
     end
+
+    @schema["attributes"]&.each do |attr_name, attr_schema|
```
ChrisBAshton, line 63: The addition of this block feels commit sized (and should have only its corresponding tests in the same commit)

app/models/standard_edition/block_content.rb
```diff
@@ -98,6 +107,12 @@ def to_h
         attributes
       end
     end
+
+    validations_config&.each do |validator_key, options|
```
ChrisBAshton, line 114: And here too - another commit

ChrisBAshton (commented):
The approach is looking good, and I know we're keen to get this over the line, so I hate to add friction here. But I'm not sure I 100% understand the changes, as there are a lot of changes in one commit here, some of which are quite deep in the StandardEdition architecture, making it a bit tricky and risky to review.

Please could you break it up into several commits? I've suggested how you can break it down below. (Breaking up the code isn't enough, mind - we also need the commit messages themselves to explain what's going on).

Thanks 🙏

app/models/configurable_document_types/topical_event.json
```diff
@@ -131,6 +131,14 @@
           "url": {
             "type": "string"
           }
+        },
```
lauraghiorghisor-tw, line 134: I think at the time when we introduced fields (under forms) we decided against wrapper objects and removed veeery similar looking code, whereby we were able to apply nested-like rules. I think at the time we also said something about how validation would apply only to "true" (block content) model attributes (which in this case nested field attributes are not), i.e. only attributes listed under "schema". I don't think this was necessarily enforced anywhere, but it was understood to be a constraint, or a constraint we'd be keen to remove in the future. Maybe @eYinka or @ChrisBAshton can remember more.

I think after the wrapper object removal work, we added the array block, which had its own needs and likely has to save its children in the json.

Anyway, feels like we're going in circles with the implementation. We now have some top-level model attributes validations and then these nested ones for the children of an array. Few questions:

1. Are we ok to now adjust our previous agreement of validation? i.e. we can have nested fields validated as well.
2. Does this also mean that we can validate nested fields in an object (a nested object, I guess, not a root level one)? Should we restrict the logic added for validation to specify we only want it to run for arrays? Unsure of the implications, but it feels like an edge case that needs to be considered, what do we want?
3. The shape of the schema is now a bit confusing, you have validations at the top level in the schema, then you also have nested validations under an attribute. If we're going to allow listing validations in the attributes section, maybe we should just fully switch to defining them there. So maybe body attribute:
```
    "attributes": {
      "body": {
        "type": "string",
        "validations": blah
      }
```

test/unit/app/models/standard_edition_test.rb
```diff
@@ -840,6 +840,58 @@ class StandardEditionTest < ActiveSupport::TestCase
 
       assert_equal %w[title summary field_attribute object_attribute.nested_attribute], page.error_field_order
     end
+
```
TonyGDS, line 843: We should also have a test for the order for dynamic tabs, currently we only have one for the documents tab.

app/validators/social_media_links_validator.rb
```diff
@@ -9,68 +9,25 @@ def initialize(opts = {})
...
   end
 
-  def valid_url?(url)
-    uri = URI.parse(url)
-    uri.is_a?(URI::HTTP) && uri.host.present?
-  rescue URI::InvalidURIError
-    false
+private
+
+  def duplicate_values(arr, field, exclude: nil)
```
TonyGDS, line 25: the name implies this method actually duplicates the values, it should be `duplicates_values?`

public/configurable-document-type.schema.json
TonyGDS, line 36: this does not meet the AC for the message "Invalid URL - use the full URL, including https://", the validator gives the error "URL is not a valid URI. Make sure it starts with http(s)"

app/models/standard_edition/block_content.rb
```diff
@@ -98,6 +110,12 @@ def to_h
...
+      attrs = Array(options["attributes"]).map(&:to_sym)
```
TonyGDS, line 115: What was the reason we limit to attributes here? If we preserve full options we could provide customer error messages etc

app/models/configurable_content_blocks/default_array.rb
```diff
@@ -29,6 +29,20 @@ def each(&block)
       end
     end
 
+    def each_leaf(&block)
```
TonyGDS, line 32: We have very similar logic in `default_object`, maybe we can DRY this up in BaseBlock?

app/models/configurable_document_types/topical_event.json
TonyGDS, line 166: should this now be `channel_field` to match the name change?

app/models/standard_edition/block_content.rb
TonyGDS, line 63: maybe this can be moved to a `run_item_validations` method? Then another method `run_validations` could run both `run_schema_validations` and `run_item_validations` together.

app/models/standard_edition.rb
```diff
@@ -146,24 +146,27 @@ def permitted_image_usages
...
+    type_instance.dynamic_tabs
```
TonyGDS, line 162: We already get the tab keys in a similar way in `valid_tab_key`. We could DRY this up this sort of logic with a `tab_keys` method.

public/configurable-document-type.schema.json
```diff
@@ -158,6 +159,10 @@
         "attributes": {
           "description": "For array types, defines the schema of each array item using the same minimal attribute format.",
           "$ref": "#/$defs/block_attributes"
+        },
+        "validations": {
```
lauraghiorghisor-tw, line 163: See the other comment about allowing validations at this level.

test/unit/app/models/standard_edition_test.rb
```diff
@@ -840,6 +840,58 @@ class StandardEditionTest < ActiveSupport::TestCase
...
+    test "validation errors on list row fields use the field title, not the internal field path" do
```
lauraghiorghisor-tw, line 844: "list row fields"? - let's call them array fields or smth

Also this test feels a bit weird -  you're composing these messages on which you expect things, but it's not clear why that messages shape is important to us. Wonder if it makes sense to just cover the `error_labels` on the st_ed test. Maybe the whole logic can live in error summary component test.

app/models/standard_edition.rb
lauraghiorghisor-tw, line 162: For my own understanding - so this code doesn't really do anything different from what it did before other than ordering (now also) dynamic tab fields so they come at the top?

I.e. our famous change note missing error would show last, rather than first (this is turned off for TE now though).

Bit out of scope, but I would write a test describing that as is behaviour - i.e. dynamic tabs would for the time being render the documents tab error too (and then we can write the correct test when we run the tabs validation card).

app/models/standard_edition/block_content.rb
lauraghiorghisor-tw, line 63: When you have multiple blocks with nested fields, would the errors be mixed up, i.e. rendering in the summary box as:
- top-level-1
- top-level-2
- nested-1-1
- nested-1-2
- nested-2-1
- nested-2-2

or would the label ordering from previous commit actually render them in config order - i.e. parent > children, next parent > its children?

app/models/standard_edition/block_content.rb
```diff
@@ -98,6 +110,12 @@ def to_h
         attributes
       end
     end
+
```
lauraghiorghisor-tw, line 113: I think I understand how it works, but the methods are becoming a bit polluted, you now have validation logic in the attributes method.

test/unit/app/models/standard_edition/block_content_test.rb
```diff
@@ -175,8 +175,32 @@ class BlockContentTest < ActiveSupport::TestCase
...
     page = StandardEdition::BlockContent.new(schema)
 
-    page.attributes = { "test_array_attribute" => [{ "social_media_service_id" => "1", "url" => "broken url" }] }
+    page.attributes = { "test_array_attribute" => [{ "social_media_service_id" => "Facebook", "url" => "http://facebook.com" }, { "social_media_service_id" => "Facebook", "url" => "http://facebook.com/other" }] }
     assert_not page.valid?
-    assert_not page.errors.where("test_array_attribute", :invalid_social_media_link).empty?
+    assert_not page.errors.where(:test_array_attribute, :taken).empty?
+  end
+
+  test "runs per-item validations for array attributes with sub-field validations" do
```
lauraghiorghisor-tw, line 183: Unclear test name - is the item the sub-field? also we tend to call these things "nested" something.

app/views/admin/configurable_content_blocks/default_select.html.erb
```diff
@@ -4,5 +4,5 @@
   name: block.path.form_control_name,
   options: block.select_options,
   hint: block.hint_text,
-  error_items: errors_for(block.edition.errors, block.path.validation_error_attribute.to_sym),
+  error_items: errors_for(block.edition.errors, block.path.validation_error_attribute.to_sym, block.title),
```
lauraghiorghisor-tw, line 7: Was this absolutely required here? Does it have any implications for anywhere else where we use this?

lauraghiorghisor-tw (commented):
LGTM but maybe let's have a sit-down chat about the schema structure.

app/views/admin/configurable_content_blocks/default_select.html.erb
lauraghiorghisor-tw: Ah got it, I think this belongs with that test on the error labels on commit 2 or whatever - I don't think this is the right commit.

PR #11392, opened 2026-04-23
Remove org restriction on new topical events [WHIT-3372]

kashifatcha:
As things stand only GDS and DCMS users can create New Topical events
/government/admin/standard-editions/new?configurable_document_type=topical_event
We are making the feature more widely accessible for new Topical events (without “about pages”)
The plan is to send out a basecamp post and update the guidance accordingly.
Before doing that we should remove the gds/dcms restriction.
Publishers will still need to have a conversation with GDS before being granted the option to create an TE.

AC:
As a user who is not in GDS or DCMS
I want to access Topical Event pages
So that I can create a new topical event

### DIFF
--- a/app/models/configurable_document_types/topical_event.json
+++ b/app/models/configurable_document_types/topical_event.json
@@ -197,7 +197,7 @@
     "send_change_history": false,
     "file_attachments_enabled": false,
     "features_enabled": true,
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM 👍 please also remember to add the jira ticket to the PR title so that it's automatically linked in Jira.

PR #11393, opened 2026-04-23
First publishing date validation [WHIT-3076]

davidsauntson:
## What

This PR:

- adds validation to the `Edition` model to prevent users changing the `first_published_at` date to a date after the first change note was created
- stops two contradictory errors messages being shown to the user when `first_published_at` is set to a future date on an edition with changes notes in associated editions

## Why

Allowing users to change `first_published_at` to a date after the earliest change note causes undesirable behaviour including items appearing in the wrong order in the 'See all updates' section of a rendered page.

The `first_published_preceeds_change_notes` validation caused two contradictory error messages when the `first_published_at` date was set to the future on editions with `change_notes`, ie:

- "First published date must be between 01/01/1900 and the present"
- "First published must be before the first change note (<change note date>)"

This change fixes this by only running the `first_published_at, inclusion` validation if there are no `other_editions` present for the draft edition in question.

https://gov-uk.atlassian.net/browse/WHIT-3076

### before

image

image

### After

image

### DIFF
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -238,6 +238,8 @@ class Edition < ApplicationRecord
   end
 
   def other_editions
+    return [] if document.nil?
+
     if persisted?
       document.editions.where(self.class.arel_table[:id].not_eq(id))
     else
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -68,7 +68,7 @@ class Edition < ApplicationRecord
   validates :summary, presence: true, if: :summary_required?, length: { maximum: 65_535 }
   validates :previously_published, inclusion: { in: [true, false], message: "You must specify whether the document has been published before" }
   validates :first_published_at, presence: true, if: -> { previously_published || published_major_version }
-  validates :first_published_at, inclusion: { in: proc { Date.parse("1900-01-01")..Time.zone.now } }, if: :draft?, allow_blank: true
+  validates :first_published_at, inclusion: { in: proc { Date.parse("1900-01-01")..Time.zone.now } }, if: -> { draft? && other_editions.empty? }, allow_blank: true
   validates :scheduled_publication, inclusion: { in: proc { Time.zone.now.. }, message: "must be in the future" }, if: :draft?, allow_blank: true
   validates :political, inclusion: { in: [true, false] }
   validates :image_display_option, inclusion: { in: ["no_image", "organisation_image", "custom_image", nil] }
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -73,6 +73,8 @@ class Edition < ApplicationRecord
   validates :political, inclusion: { in: [true, false] }
   validates :image_display_option, inclusion: { in: ["no_image", "organisation_image", "custom_image", nil] }
 
+  validate :first_published_preceeds_change_notes, if: :draft?
+
   UNMODIFIABLE_STATES = %w[scheduled published superseded deleted unpublished].freeze
   FROZEN_STATES = %w[superseded deleted].freeze
   PRE_PUBLICATION_STATES = %w[draft submitted rejected scheduled].freeze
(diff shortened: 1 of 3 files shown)


config/locales/en/errors.yml
```diff
@@ -23,6 +23,7 @@ en:
               invalid_date: First published date is not in the correct format
               blank: Enter a first published date
               inclusion: First published date must be between 1/1/1900 and the present
+              after_change_notes: "First published date must before the first change note (%{latest})"
```
GDSNewt: Think there may be a Typo/omission. 

Do we need a 'be' after must?

GDSNewt (approved):
Sorry it took so long to get this review out David!

Following our chat the other day I'm satisfied that the validation logic covers both legacy edition types and StandardEdition. 

Optional suggestion: I think the existing test only exercises GenericEdition via the :edition_with_document factory. Would it be worth adding a test using a StandardEdition subclass just to make that coverage explicit?

test/unit/app/models/edition_test.rb
```diff
@@ -678,6 +678,13 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal "First published date must be between 1/1/1900 and the present", edition.errors.full_messages.first
   end
 
+  test "first_published_at cannot be after the date of the first change note" do
+    edition_with_change_note = create(:edition_with_document, :published, change_note: "changed", major_change_published_at: 2.days.ago)
+    edition = build(:edition, document: edition_with_change_note.document, first_published_at: 1.day.ago)
+
+    assert edition.invalid?
```
ChrisBAshton, line 691: I'd also assert the error message here.

app/models/edition.rb
```diff
@@ -238,6 +238,8 @@ def body_without_markup
   end
 
   def other_editions
+    return [] if document.nil?
```
ChrisBAshton, line 241: I'd pull this change out to a separate commit, and add a test for it.

test/unit/app/models/edition_test.rb
```diff
@@ -684,6 +684,13 @@ class EditionTest < ActiveSupport::TestCase
 
     assert edition.invalid?
   end
+  test "first_published_at dates in the future only generate one error message" do
```
ChrisBAshton: Missing newline between 686 and 687

test/unit/app/models/edition_test.rb
```diff
@@ -684,6 +684,13 @@ class EditionTest < ActiveSupport::TestCase
...
+    edition_with_change_note = create(:edition_with_document, :published, change_note: "changed", major_change_published_at: 2.days.ago)
+    edition = build(:edition, document: edition_with_change_note.document, first_published_at: 10.years.from_now)
+    edition.validate
+
+    assert_equal 1, edition.errors.size
```
ChrisBAshton, line 700: I think I'd assert the error message here too (and rephrase the test to be more like `test "'after_change_notes' error message takes priority if multiple validation errors on first_published_at"`)

ChrisBAshton (commented):
LGTM - nice first PR! 🎉  Just a few comments below.

ChrisBAshton (approved):
Nice 🚢

davidsauntson:
> Sorry it took so long to get this review out David!
> 
> Following our chat the other day I'm satisfied that the validation logic covers both legacy edition types and StandardEdition.
> 
> Optional suggestion: I think the existing test only exercises GenericEdition via the :edition_with_document factory. Would it be worth adding a test using a StandardEdition subclass just to make that coverage explicit?

Thanks Alex!

I had a look at the `StandardEdition` point and I thought that the inheritance from `Edition` on the `StandardEdition` model makes the coverage explicit enough anyway, so I wonder if repeating the same test using the `StandardEditionFactory` is needed?  I think it would be odd to just include it for `Edition` and `StandardEdition` and not other models that inherit from `Edition`, eg `Publication` etc?  LMK what you think.

PR #11396, opened 2026-04-24
Standard Edition Migrator can support non-edition types [WHIT-3268]

GDSNewt:
## Summary

Extends `StandardEditionMigrator` and `StandardEditionMigratorJob` to support migrating legacy content types that do not use the Edition model (e.g. `TopicalEvent`), in addition to the existing editionable migration path.

Previously, passing a non-`Document` scope would fail. Now, any ActiveRecord model scope can be passed, and the migrator routes to the appropriate code path.

---

### Flexible scope handling

- Before: Only `Document` scopes were supported  
- Now: Any ActiveRecord model scope is supported  

### Preview changes

- Editionable: unchanged (edition-based summary)  
- Non-editionable:
  ```ruby
  { unique_records: N }
  

### `StandardEditionMigrator#migrate!`
- Enqueues jobs with an additional `model_class` argument

### `StandardEditionMigratorJob`
- Loads records using `model_class`
- For each record:
  - Creates a new `Document`
  - Creates a new `StandardEdition`
  - Maps fields using the recipe
- Leaves the original record untouched (manual cleanup required)

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-3268)

### DIFF
--- a/app/services/standard_edition_migrator.rb
+++ b/app/services/standard_edition_migrator.rb
@@ -4,21 +4,38 @@ class StandardEditionMigrator
   end
 
   def preview
-    total_editions = @scope.sum { |doc| Edition.unscoped.where(document: doc).count }
-
-    {
-      unique_documents: @scope.count,
-      total_editions: total_editions,
-    }
-  end
-
-  def migrate!(republish: false, compare_payloads: true)
-    @scope.each do |document|
-      StandardEditionMigratorJob.perform_async(document.id, { "republish" => republish, "compare_payloads" => compare_payloads })
+    if document_scope?
+      total_editions = @scope.sum { |doc| Edition.unscoped.where(document: doc).count }
+      { unique_documents: @scope.count, total_editions: total_editions }
+    else
+      { unique_records: @scope.count }
     end
   end
 
-  def self.recipe_for(edition)
-    raise "No migration recipe defined for Edition type #{edition.type}"
+  def migrate!(republish: false, compare_payloads: true)
+    @scope.each do |record|
+      StandardEditionMigratorJob.perform_async(
+        record.id,
+        { "republish" => republish, "compare_payloads" => compare_payloads, "model_class" => model_class_name },
+      )
+    end
+  end
+
+  def self.recipe_for(model)
+    if model.is_a?(Edition)
+      raise "No migration recipe defined for Edition type #{model.type}"
+    end
+
+    raise "No migration recipe defined for #{model.class.name}"
+  end
+
+private
+
+  def model_class_name
+    @scope.model.name
+  end
+
+  def document_scope?
+    model_class_name == "Document"
   end
 end
--- a/app/sidekiq/standard_edition_migrator_job.rb
+++ b/app/sidekiq/standard_edition_migrator_job.rb
@@ -7,10 +7,21 @@ class StandardEditionMigratorJob < JobBase
   # and any failure is unlikely to resolve itself on a retry.
   sidekiq_options queue: "standard_edition_migration", retry: 0
 
-  def perform(document_id, args)
+  def perform(record_id, args)
     republish = args["republish"]
     compare_payloads = args["compare_payloads"]
+    model_class_name = args["model_class"]
 
+    if model_class_name != "Document"
+      perform_for_non_editionable(record_id, model_class_name, republish:, compare_payloads:)
+    else
+      perform_for_document(record_id, republish:, compare_payloads:)
+    end
+  end
+
+private
+
+  def perform_for_document(document_id, republish:, compare_payloads:)
     ActiveRecord::Base.transaction do
       document = Document.find(document_id)
       migrate_editions!(document, compare_payloads)
@@ -19,7 +30,16 @@ class StandardEditionMigratorJob < JobBase
     PublishingApiDocumentRepublishingJob.new.perform(document_id, true) if republish
   end
 
-private
+  def perform_for_non_editionable(record_id, model_class_name, republish:, compare_payloads:)
+    record = model_class_name.constantize.find(record_id)
+    recipe = StandardEditionMigrator.recipe_for(record)
+
+    new_document_id = ActiveRecord::Base.transaction do
+      migrate_non_editionable!(record, recipe, compare_payloads)
+    end
+
+    PublishingApiDocumentRepublishingJob.new.perform(new_document_id, true) if republish
+  end
 
   def migrate_editions!(document, compare_payloads)
     editions_to_migrate = Edition.unscoped.where(document: document)
@@ -34,6 +54,36 @@ private
         migrate_edition!(edition, recipe)
       end
     end
+  end
+
+  def migrate_non_editionable!(record, recipe, compare_payloads)
+    document = Document.create!(document_type: "StandardEdition", content_id: record.content_id)
+    edition = StandardEdition.new(
+      document:,
+      configurable_document_type: recipe.configurable_document_type,
+      state: "published",
+    )
+    edition.save!(validate: false)
+
+    record.translations.each do |record_translation|
+      edition.translations.find_or_create_by!(locale: record_translation.locale).update_columns(
+        title: recipe.title(record_translation),
+        summary: recipe.summary(record_translation),
+        block_content: recipe.map_legacy_fields_to_block_content(record, record_translation),
+      )
+    end
+
+    if compare_payloads
+      old_presenter = recipe.presenter.new(record)
+      new_presenter = PublishingApi::StandardEditionPresenter.new(edition)
+      compare_presenter_payloads(
+        old_presenter.content, new_presenter.content,
+        old_presenter.links, new_presenter.links,
+        recipe, subject: "#{record.class.name} ID #{record.id}"
+      )
+    end
+
+    document.id
   end
 
   def migrate_edition!(edition, recipe)
@@ -62,12 +112,19 @@ private
     new_content = new_presenter.content
     new_links = new_presenter.links
 
+    compare_presenter_payloads(
+      old_content, new_content, old_links, new_links,
+      recipe, subject: "Edition ID #{edition.id}"
+    )
+  end
+
+  def compare_presenter_payloads(old_content, new_content, old_links, new_links, recipe, subject:)
     diff = diff_values(
       recipe.ignore_legacy_content_fields(old_content),
       recipe.ignore_new_content_fields(new_content),
     )
     unless diff.to_s.empty?
-      raise "Presenter content mismatch after migration for Edition ID #{edition.id}: #{diff}"
+      raise "Presenter content mismatch after migration for #{subject}: #{diff}"
     end
 
     diff = diff_values(
@@ -75,7 +132,7 @@ private
       recipe.ignore_new_links(new_links),
     )
     unless diff.to_s.empty?
-      raise "Presenter links mismatch after migration for Edition ID #{edition.id}: #{diff}"
+      raise "Presenter links mismatch after migration for #{subject}: #{diff}"
     end
   end
 
--- a/docs/migrating_to_standard_edition.md
+++ b/docs/migrating_to_standard_edition.md
@@ -89,4 +89,32 @@ With that in mind, the steps for migrating a legacy content type to being config
 
 ### The legacy content type does not use the Edition model
 
-You'll need to perform a bespoke migration.
+If the legacy content type doesn't use the Edition model (e.g. `TopicalEvent`), you can still use the `StandardEditionMigrator` — just pass a scope of the legacy records directly rather than a `Document` scope.
+
+For each record in scope, the migrator creates a brand new `Document` and `StandardEdition`, preserving the original record's `content_id` on the new document. Note that for the base path overwrite to work in Publishing API you'll need to do the groundwork described in [#11210](https://github.com/alphagov/whitehall/pull/11210) — this lets you interchangeably publish either the legacy content item or the new `StandardEdition` at the same path for testing purposes. The original record is left untouched, so you can remove it manually afterwards (or leave the two co-existing for a while if that's useful).
+
+The recipe works the same way as for editionable content types, with two differences:
+
+- Use the same `StandardEditionMigrator.recipe_for` method (it accepts any model, not just editions)
+- Add `title(record_translation)` and `summary(record_translation)` methods — these are called once per source record translation, so if the record has multiple locales, each gets its own edition translation. The argument is the Globalize translation object, so you can read locale-specific attributes directly from it (e.g. `record_translation.name`).
+
+`map_legacy_fields_to_block_content` takes `(record, record_translation)` — the first argument is your record (for any non-translated data) and the second is the source translation being processed. Everything else — `presenter`, `configurable_document_type`, and the `ignore_*` methods — works identically to the editionable recipe.
+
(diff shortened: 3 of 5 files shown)


app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -7,10 +7,21 @@ class StandardEditionMigratorJob < JobBase
   # and any failure is unlikely to resolve itself on a retry.
   sidekiq_options queue: "standard_edition_migration", retry: 0
 
-  def perform(document_id, args)
+  def perform(record_id, args)
     republish = args["republish"]
     compare_payloads = args["compare_payloads"]
+    model_class_name = args["model_class"]
 
+    if model_class_name.present? && model_class_name != "Document"
```
ChrisBAshton: We should always assume `model_class_name` is present

```suggestion
    if model_class_name != "Document"
```

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -36,6 +56,37 @@ def migrate_editions!(document, compare_payloads)
     end
   end
 
+  def migrate_non_editionable!(record, recipe, compare_payloads)
+    old_presenter = recipe.presenter.new(record) if compare_payloads
+
+    document = Document.create!(document_type: "StandardEdition", content_id: SecureRandom.uuid)
```
ChrisBAshton: We need the content_id to remain identical. E.g. TopicalEvent with content_id `abc123` needs to be copied to StandardEdition whose Document has a content_id `abc123`.

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -36,6 +56,37 @@ def migrate_editions!(document, compare_payloads)
...
+      title: recipe.title(record),
+      summary: recipe.summary(record),
+    )
+    edition.save!(validate: false)
+
+    edition.translations.each do |translation|
+      translation.update_columns(
+        block_content: recipe.map_non_editionable_to_block_content(record, translation),
+      )
+    end
```
ChrisBAshton: I'm not sure this works 🤔 

1. `edition.translations` will always just be an array with 'en' as the only translation (the default), since we're just creating the edition from scratch here. The `edition.translations` iterator in `migrate_edition!` was necessary because it was dealing with existing editions some of which had translations.
2. I'm not sure if we have any non-editionable content types that use translations? 🤔 If we do, then they need copying over here. If we don't, then we don't have to touch any translation code in this migrator service.

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
```diff
@@ -373,12 +373,98 @@ class StandardEditionMigratorJobTest < ActiveSupport::TestCase
...
+          perform_non_editionable_migration
+        end
+      end
+
+      new_document = Document.last
+      assert_equal "StandardEdition", new_document.document_type
+
+      new_edition = Edition.last
+      assert_equal "draft", new_edition.state
+      assert_equal "test_type", new_edition.configurable_document_type
```
ChrisBAshton: Nowhere in the recipe have we specified the configurable_document_type. I think this is just default behaviour in the tests you're relying on - it doesn't give us value. You should have the recipe specify the configurable document type to convert to, and assert that here.

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
```diff
@@ -373,12 +373,98 @@ class StandardEditionMigratorJobTest < ActiveSupport::TestCase
...
+    setup do
+      ConfigurableDocumentType.setup_test_types(build_configurable_document_type("test_type"))
+      StandardEditionMigrator.stubs(:recipe_for_non_editionable).returns(StandardEditionMigratorJobTest::TestNonEditionableRecipe.new)
+
+      TestNonEditionablePresenter.any_instance.stubs(:content).returns({ some: "content" })
+      PublishingApi::StandardEditionPresenter.any_instance.stubs(:content).returns({ some: "content" })
+      TestNonEditionablePresenter.any_instance.stubs(:links).returns({ some: "links" })
+      PublishingApi::StandardEditionPresenter.any_instance.stubs(:links).returns({ some: "links" })
+
+      @topical_event = create(:topical_event, name: "My Event", summary: "Event summary", description: "Event description")
```
ChrisBAshton: Topical events will be deleted this quarter, at which point we'd have to rewrite all these tests. Would be better to use a different class, or better yet, define a fake test class in this file.

docs/migrating_to_standard_edition.md
```diff
@@ -89,4 +89,33 @@ With that in mind, the steps for migrating a legacy content type to being config
 
 ### The legacy content type does not use the Edition model
 
-You'll need to perform a bespoke migration.
+If the legacy content type doesn't use the Edition model (e.g. `TopicalEvent`), you can still use the `StandardEditionMigrator` — just pass a scope of the legacy records directly rather than a `Document` scope.
+
+For each record in scope, the migrator creates a brand new `Document` and `StandardEdition`. The original record is left untouched, so you can remove it manually afterwards (or leave the two co-existing for a while if that's useful).
```
ChrisBAshton: Maybe make a reference to https://github.com/alphagov/whitehall/pull/11210 here, to clarify that you can interchangably overwrite a path with either the legacy content item or the new StandardEdition content item (for testing purposes) but need to do the groundwork as per that PR.

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -19,7 +30,16 @@ def perform(document_id, args)
...
-private
+  def perform_for_non_editionable(record_id, model_class_name, republish:, compare_payloads:)
+    record = model_class_name.constantize.find(record_id)
+    recipe = StandardEditionMigrator.recipe_for_non_editionable(record)
+
+    new_document_id = ActiveRecord::Base.transaction do
+      migrate_non_editionable!(record, recipe, compare_payloads)
+    end
+
+    PublishingApiDocumentRepublishingJob.new.perform(new_document_id, true) if republish
```
ChrisBAshton, line 41: Indeed, this line would fail unless the dev has done the ground work in terms of allowing overwrites of the base path as per my comment above.

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -19,7 +30,16 @@ def perform(document_id, args)
     PublishingApiDocumentRepublishingJob.new.perform(document_id, true) if republish
   end
 
-private
+  def perform_for_non_editionable(record_id, model_class_name, republish:, compare_payloads:)
+    record = model_class_name.constantize.find(record_id)
+    recipe = StandardEditionMigrator.recipe_for_non_editionable(record)
```
ChrisBAshton: I don't think we need a separate `recipe_for_non_editionable` method. We can just extend the existing `recipe_for` method to do something like:

```ruby
def self.recipe_for(model)
  if model.is_a?(TopicalEvent)
    return StandardEditionMigrator::TopicalEventRecipe.new
  end

  if model.is_a?(Edition)
    # TODO: we would fill this out for something like Speeches [as per news articles](https://github.com/alphagov/whitehall/pull/10814/changes#diff-b76be167ce63dc60dda7002623049b32253884decfecd8341932553c05c0f37cR21-R36) 
    raise "No migration recipe defined for Edition type #{edition.type}"
  end
end
```

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -36,6 +56,37 @@ def migrate_editions!(document, compare_payloads)
...
+    if compare_payloads
+      new_presenter = PublishingApi::StandardEditionPresenter.new(edition)
+      compare_presenter_payloads(
+        old_presenter.content, new_presenter.content,
+        old_presenter.links, new_presenter.links,
+        recipe, subject: "#{record.class.name} ID #{record.id}"
+      )
+    end
+
+    document.id
```
ChrisBAshton, line 86: I don't think we have a corresponding test that we return the created document's ID?

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
```diff
@@ -432,4 +518,22 @@ def ignore_new_links(links)
...
+
+    def title(record)
+      record.name
+    end
+
+    def summary(record)
+      record.summary
+    end
+
+    def map_non_editionable_to_block_content(record, _translation)
```
ChrisBAshton: This new method isn't needed? We have `map_legacy_fields_to_block_content` already and it does the same thing. Whether we're calling it against an Edition or something else doesn't affect the first parameter. (You're probably right that there is no second parameter, but then we just wouldn't use it in the recipe).

ChrisBAshton (changes requested):
Hi Alex,

This is a good start 👍 One major blocker is the converted document must have the same content ID - not SecureRandom. And then there are a few other suggestions for improvements - I'm not sure we need to create lots of extra methods on the Recipe class to support non-Edition classes.

app/sidekiq/standard_edition_migrator_job.rb
ChrisBAshton: EDIT: we do have translations of non-editionable things, e.g. Organisations.

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -36,6 +56,37 @@ def migrate_editions!(document, compare_payloads)
     end
   end
 
+  def migrate_non_editionable!(record, recipe, compare_payloads)
+    old_presenter = recipe.presenter.new(record) if compare_payloads
```
ChrisBAshton: I'd move this inside the `if compare_payloads` as it's only needed there

app/sidekiq/standard_edition_migrator_job.rb
```diff
@@ -19,7 +30,16 @@ def perform(document_id, args)
...
-private
+  def perform_for_non_editionable(record_id, model_class_name, republish:, compare_payloads:)
+    record = model_class_name.constantize.find(record_id)
+    recipe = StandardEditionMigrator.recipe_for(record)
+
+    new_document_id = ActiveRecord::Base.transaction do
+      migrate_non_editionable!(record, recipe, compare_payloads)
+    end
+
+    PublishingApiDocumentRepublishingJob.new.perform(new_document_id, true) if republish
```
ChrisBAshton, line 41: This won't publish the document - it only [republishes an existing published document](https://github.com/alphagov/whitehall/blob/34ee818fdec2b111a6358b68f95803564df08d31/app/sidekiq/publishing_api_document_republishing_job.rb#L84-L85). You've set state to `draft` on line 66. I think we need to set state to `published` right off the bat.

Could alternatively call a different Publishing API method here, but that's more complicated for little benefit imo.

test/unit/app/services/standard_edition_migrator_test.rb
```diff
@@ -81,6 +81,15 @@ class StandardEditionMigratorTest < ActiveSupport::TestCase
 
       assert_equal summary, migrator.preview
     end
+
+    test "summarises how many non-editionable records will be migrated" do
+      create(:topical_event)
+      create(:topical_event)
```
ChrisBAshton: As before - these tests will fail the moment we delete topical events. Please change for a different class / fake class.

test/unit/app/services/standard_edition_migrator_test.rb
```diff
@@ -116,8 +125,22 @@ class StandardEditionMigratorTest < ActiveSupport::TestCase
...
       migrator = StandardEditionMigrator.new(scope: Document.all)
 
-      StandardEditionMigratorJob.expects(:perform_async).with(some_doc.document.id, { "republish" => true, "compare_payloads" => false }).once
+      StandardEditionMigratorJob.expects(:perform_async).with(some_doc.document.id, { "republish" => true, "compare_payloads" => false, "model_class" => "Document" }).once
       migrator.migrate!(republish: true, compare_payloads: false)
     end
+
+    test "enqueues a migration job for each non-editionable record, passing the model class" do
+      topical_event_1 = create(:topical_event)
+      topical_event_2 = create(:topical_event)
```
ChrisBAshton: Ditto

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
```diff
@@ -367,18 +367,128 @@ class StandardEditionMigratorJobTest < ActiveSupport::TestCase
...
+          perform_non_editionable_migration
+        end
+      end
+
+      new_document = Document.last
+      assert_equal "StandardEdition", new_document.document_type
+      assert_equal @test_record.content_id, new_document.content_id
+
+      new_edition = Edition.last
+      assert_equal "draft", new_edition.state
```
ChrisBAshton: We want this to be `published`

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
```diff
@@ -432,4 +542,56 @@ def ignore_new_links(links)
...
+    def initialize(id:, name:, summary:, description:, content_id:)
+      @id = id
+      @name = name
+      @summary = summary
+      @description = description
+      @content_id = content_id
+    end
+
+    def translations
+      [Translation.new(locale: :en, name: @name, summary: @summary, description: @description)]
```
ChrisBAshton: Please add a corresponding test that the created edition has translations carried over.

app/services/standard_edition_migrator.rb
```diff
@@ -4,21 +4,38 @@ def initialize(scope:)
...
+      raise "No migration recipe defined for Edition type #{model.type}"
     end
+
+    raise "No migration recipe defined for #{model.class.name}"
+  end
+
+private
+
+  def model_class_name
+    @scope.is_a?(ActiveRecord::Relation) ? @scope.model.name : @scope.first.class.name
```
ChrisBAshton: This suggests you can pass both `TopicalEvent.all` and `TopicalEvent`. We haven't documented that the latter is allowed. I'd vote to simplify the `model_class_name` method instead of supporting both.

ChrisBAshton (commented):
Thanks Alex - getting there. Main thing to sort now is the edition state.

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
```diff
@@ -367,18 +367,138 @@ class StandardEditionMigratorJobTest < ActiveSupport::TestCase
...
+      assert_equal "My Event", new_edition.title
+      assert_equal "Event summary", new_edition.summary
+      assert_equal({ "description" => "Event description" }, new_edition[:block_content])
+    end
+
+    test "carries over translations from the non-editionable record to the new edition" do
+      perform_non_editionable_migration
+
+      new_edition = Edition.last
+      en_translation = new_edition.translations.find_by(locale: "en")
```
ChrisBAshton, line 443: Probably ought to test this more thoroughly by having a test subject with at least two translations (e.g. en and cy) and checking they're both carried over.

ChrisBAshton (approved):
LGTM 👍 one suggestion below

test/unit/app/sidekiq/standard_edition_migrator_job_test.rb
GDSNewt: Translation test elaborated on 👍

PR #11387, opened 2026-04-21
Improve Whitehall's search performance

ChrisBAshton:
## What

This PR:

- Adds a benchmarking script we can use to track search performance before-and-after edits
- Makes a low hanging fruit improvement in terms of detecting if the input is likely a slug (by checking that the search term contains at least one hyphen), and if so, doing a wildcard search on `slug`, otherwise making that search on `title`. This makes slug searches 99% faster (! 🎉  !) and general title search about 40% faster (as we're only searching against one field now, not two.

There is an edge case where a title could be 'slug-like', e.g. "Covid-19", and in that case no results would be returned if the slug itself is different. But most of the time the slug should match, since it is derived from the title. So it should be _extremely_ rare for that to be an issue.

## Why

We've known for a long time that Whitehall search (by title) is _sloooooow_. We've also had some recent reports of related 5xx errors. This is a first iteration attempt at speeding up the search performance.

---

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

This application is owned by the Whitehall Experience team. Please let us know in [#govuk-whitehall-experience-tech](https://gds.slack.com/archives/C02L13S214K) when you raise any PRs.

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #11397, opened 2026-04-24
Update search scope for slug and add slug_override index

eYinka:
Since the introduction of `slug_override`, editions using an overridden slug were not found when searching by that slug value. I've extended the slug search condition to also check `slug_override`, and add a database index on the column to keep lookup performance on par with the existing slug index.

Related PR: https://github.com/alphagov/whitehall/pull/11387

---


ChrisBAshton (approved):
LGTM 👍 I'd personally make the DB migration its own commit, then make the searchable_by_title change in a separate commit, but only nitpicking.

PR #11400, opened 2026-04-24
Automatically fix up associations when converting news article subtypes [WHIT-2883]

ChrisBAshton:
## What

Adds a new `StandardEdition::HygieneEnforcer` class that contains all the hard-coded rules around which associations to drop when converting from one configurable document type to another. The class is invoked via a callback on the `update_configurable_document_type` method call.

This is the least worst solution we have for now, as it's relatively lightweight, scalable, and follows the Open-Closed principle. The alternatives are:

1. Continue as we are, patching up news articles manually in response to support tickets.
2. Running the remediation rake task on a cron, so that we don't have to run it manually (but still may get support tickets for newly converted documents).
3. Build fully fledged 'association management' [as outlined here](https://gov-uk.atlassian.net/browse/WHIT-2883?focusedCommentId=185941) - quite a heavyweight solution for what is pretty niche behaviour.

### 'Links' behaviour when changing type

The original rake task had a [guardrail](https://github.com/alphagov/whitehall/pull/10952/changes#diff-ae83cbb49dd3aa3f1e0567300c835676c0d70a3f0ae5fb84b0fb2e2445f1291fR119-R121) in it that only did the association removal if there was a new 'equivalent' value specified. E.g. a news_story, converted to a world_news_story, would only have its 'organisations' dropped if there were at least one worldwide organisation set.

We couldn't carry this over as-is because we're now calling this cleanup code immediately after changing the document type. At that point, the user hasn't had a chance to set a value for worldwide organisations yet - meaning that association would always be empty, and thus the HygieneEnforcer wouldn't do anything.

Instead of calling the HygieneEnforcer class on changing document type, I considered calling it after every single save, so that we drop the old association as soon as we have a new equivalent. But that felt surprising, very unclean and potentially could lead to poor performance.

Therefore what I've done instead is drop the guardrail. We drop the obsolete associations (e.g. organisations) **as soon as the document type is changed**, without checking if there is an equivalent set (e.g. worldwide organisations).

This does cause some interesting behaviour. Consider the following:

1. A news story is created, with organisations set (as [edition links](https://github.com/alphagov/whitehall/blob/1fd2d4701c9f92beb025168cae30fc850b98aa27/app/presenters/publishing_api/standard_edition_presenter.rb#L24) _and_ as [link set links](https://github.com/alphagov/whitehall/blob/1fd2d4701c9f92beb025168cae30fc850b98aa27/app/presenters/publishing_api/standard_edition_presenter.rb#L32)), and is published. This is (live) `Edition 1`.

Here's how it might look (note the organisation link):

> Screenshot 2026-04-27 at 13 20 17

2. A new draft of it is created and it is converted to a world news story, which **removes 'organisations' from the link set links via the `patch_links` call in `HygieneEnforcer`**. It's edited to have a worldwide organisation added (again, as edition links and link set links). This is (draft) `Edition 2`.

Now when we render the live page again (edition 2), we see **both organisations (via edition links) and worldwide organisations (via link set links)**.

> Screenshot 2026-04-27 at 12 59 26

When we render the _draft_ page (edition 2): we see only worldwide organisations (via edition links). We don't see any organisations because they're not on the edition as edition_links, and their link set link has been removed.

> Screenshot 2026-04-27 at 12 59 30

Explanation: we send all links as both Edition Links and Link Set Links, and in link expansion, [edition links take priority](https://docs.publishing.service.gov.uk/repos/publishing-api/link-expansion.html#put-content-edition-links). Organisations were sent as link set links and edition links *on the original publish* (of `Edition 1`), but they've been dropped from link set links in the HygieneEnforcer `patch_links` call when we changed the document type. They still exist as _edition_ links, so are visible on live, but on draft they don't exist as either link type, so they're not visible.

After clicking 'Publish', the draft edition goes live, and thus only its edition links (and any remaining link set links) are rendered, and thus the organisations are dropped from the page, leaving only the worldwide organisations:

> Screenshot 2026-04-27 at 13 23 50

### So why merge this PR?

Without this PR, we already have this bug, but it's not user-remediable. Prior to this PR, if a publisher converts a news story to a world news story, the 'organisations' will not be carried over as edition links but they _do_ persist as link set links, meaning even when the publisher publishes their new (world news story) draft, the 'organisations' remain on the page. It's then a support ticket for us to step in and manually purge the link set links. So document conversion is already likely to lead to user-visible weirdness on org/worldwide-org links.

This PR, however, means that the bug is eradicated the moment the new draft is published. And as a bonus, the draft _preview_ for the publisher is always accurate.

## Why

When converting from one news article subtype to another, namely to or from world news stories, we carry over associations that are no longer relevant. In the case of 'news story' to 'world news story', the edition's `organisations` are carried over, even though only `worldwide_organisations` are supported on world news stories.

The impact of this is that the wrong organisations may be surfaced to users on the frontend, as the edition now has both organisations _and_ worldwide organisations set, and Frontend defaults to the former. There is no way for the publisher to manually remove the organisations from the edition, as the 'world news story' configurable document type doesn't have organisations listed as a field and thus no form is rendered through which they could remove existing organisations.

We therefore need to purge the old associations when changing content type, and also ensure Publishing API has the corresponding links purged too, since sending a `patch_links` call which _omits_ the old association doesn't remove them - we have to explicitly send an empty array value for the old association.

Jira: https://gov-uk.atlassian.net/browse/WHIT-2883

---

### DIFF
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -37,7 +37,11 @@ class StandardEdition < Edition
       ConfigurableDocumentType.convertible_from(configurable_document_type).none? { |type| type.key == new_type_key }
 
     self.configurable_document_type = new_type_key
-    save!(validate: false)
+    if save(validate: false)
+      StandardEdition::HygieneEnforcer.new(self).cleanup!
+      return true
+    end
+    false
   end
 
   def body_required?
--- a/app/services/standard_edition/hygiene_enforcer.rb
+++ b/app/services/standard_edition/hygiene_enforcer.rb
@@ -51,14 +51,30 @@ class StandardEdition::HygieneEnforcer
   end
 
   def cleanup!
-    if (edition.configurable_document_type == "world_news_story") && edition.organisations.any?
+    if dirty_world_news_story?(edition)
       edition.organisations.delete_all
-      # Temporarily pretend this is still a news story, so that we send the correct
-      # associations (empty 'organisations' array) as links to Publishing API.
-      # The change below isn't persisted. The patch_links method acts on the instance,
-      # so the change is only in-memory and only for the duration of this method call.
-      edition.configurable_document_type = "news_story"
-      Whitehall::PublishingApi.patch_links(edition)
+      patch_links_as_document_type(edition, "news_story")
+    elsif dirty_other_news_article_subtype?(edition)
+      edition.worldwide_organisation_documents.delete_all
+      patch_links_as_document_type(edition, "world_news_story")
     end
+  end
+
+private
+
+  def dirty_world_news_story?(edition)
+    edition.configurable_document_type == "world_news_story" && edition.organisations.any?
+  end
+
+  def dirty_other_news_article_subtype?(edition)
+    edition.configurable_document_type.in?(%w[news_story press_release government_response]) && edition.worldwide_organisations.any?
+  end
+
+  def patch_links_as_document_type(edition, document_type)
+    # Temporarily pretend this document if of document type 'document_type', so that we send the correct
+    # associations (e.g. empty 'worldwide_organisations' and 'world_locations' arrays for a doc that has
+    # been converted from a "world_news_story") as links to Publishing API.
+    edition.configurable_document_type = document_type
+    Whitehall::PublishingApi.patch_links(edition)
   end
 end
--- a/app/services/standard_edition/hygiene_enforcer.rb
+++ b/app/services/standard_edition/hygiene_enforcer.rb
@@ -1,0 +1,64 @@
+# This class is considered technical debt. We hope to remove it one day if we implement
+# https://gov-uk.atlassian.net/browse/WHIT-2883?focusedCommentId=185941
+#
+# ==== WHAT THIS CODE DOES ====
+#
+# We allow converting from one configurable document type to another (within reason),
+#  such as converting a world news story to a news story. Most conversions work fine
+# because their associations are identical (e.g. news_story -> government_response)
+# but world_news_story has different associations to the other subtypes. When
+# converting to or from world_news_story, we want to clean up the redundant associations
+# as otherwise that can present as a bug on the frontend (e.g. a world news story
+# having its 'organisations' being surfaced as the organisations on the page, instead of
+# the selected 'worldwide_organisations'. And given we hide the organisations field on
+# the world news story form, it's not possible for publishers to fix it up themselves.)
+#
+# We therefore need to:
+#   1. Remove the redundant associations in Whitehall's database, and
+#   2. Patch the links in the Publishing API to remove the redundant associations from there too.
+#
+# But that's complicated by the fact that we dynamically determine which subset of associations
+# to send to Publishing API (see `PublishingApi::PayloadBuilder::ConfigurableDocumentLinks`)
+# so we can't just call `patch_links` and call it a day. We instead need to 'spoof' the
+# configurable document type to pretend to be of the _previous_ content type, so that we can
+# patch links to send an empty array (to remove the values in Publishing API).
+#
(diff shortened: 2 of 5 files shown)


app/models/standard_edition.rb
```diff
@@ -37,7 +37,11 @@ def update_configurable_document_type(new_type_key)
       ConfigurableDocumentType.convertible_from(configurable_document_type).none? { |type| type.key == new_type_key }
 
     self.configurable_document_type = new_type_key
-    save!(validate: false)
+    if save!(validate: false)
+      StandardEdition::HygieneEnforcer.new(self).cleanup!
+      return true
+    end
+    false
```
GDSNewt, line 44: I'm a bit confused, does this false ever get hit? I think save! either returns true or raises an ActiveRecord right?

app/services/standard_edition/hygiene_enforcer.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def dirty_other_news_article_subtype?(edition)
+    edition.configurable_document_type.in?(%w[news_story press_release government_response]) && edition.worldwide_organisations.any?
+  end
+
+  def patch_links_as_document_type(edition, document_type)
+    # Temporarily pretend this document if of document type 'document_type', so that we send the correct
+    # associations (e.g. empty 'worldwide_organisations' and 'world_locations' arrays for a doc that has
+    # been converted from a "world_news_story") as links to Publishing API.
+    edition.configurable_document_type = document_type
```
GDSNewt, line 77: Is the the type ever restored? I think after` cleanup! `it isn't remembering the real type - or is it 🤔?

app/models/standard_edition.rb
ChrisBAshton: Ahh good spot! `save` returns false on failure, but `save!` raises an exception on failure. I'll take another look at this.

app/services/standard_edition/hygiene_enforcer.rb
ChrisBAshton: This type change only happens in memory - we never call `save`. Then the `patch_links` method works on the _model instance_ that's passed in - not read from the database - so it patches links as though it's of the old document type.

davidsauntson (approved):
This make sense to me 🎉  

A couple of questions, but nothing blocking:
- do `organisations` exist on other document types? If so, do you need a test that checks they aren't affected by the `HygieneEnforcer`?
- do you think it might be hard to diagnose problems in `Whitehall::PublishingApi` when the type has been (temporarily) changed like this?  Is it worth having a method in `Whitehall::PublishingApi` that handles this case, and calling it explicitly from the `HygieneEnforcer`?  I mean smth like:
```
# Whitehall::PublishingApi
def patch_links_as(edition, override_type)
  edition.configurable_document_type = document_type
  self.patch_links(edition)
end
```
I guess this is just spilling the tech debt into another class so it's probably not worth it, but thought I'd ask anyway.

ChrisBAshton:
> do organisations exist on other document types? If so, do you need a test that checks they aren't affected by the HygieneEnforcer?

Organisations do exist on other configurable document types, yes (topical events, history pages, etc) - but they shouldn't be affected as the `cleanup!` method only references the specific doc types that need cleaning up, and then has specific cleanup actions for those. Not sure there's a huge amount of value in a test for something that isn't referenced at all 🤔 

> do you think it might be hard to diagnose problems in Whitehall::PublishingApi when the type has been (temporarily) changed like this? Is it worth having a method in Whitehall::PublishingApi that handles this case, and calling it explicitly from the HygieneEnforcer? I mean smth like:

Not a bad shout, though it's StandardEdition specific logic, which we'd rather not have bleed into the generic Publishing API helper class. 🙏

PR #11410, opened 2026-04-29
[SPIKE] Prototype targeted approach for config-driven tabs [WHIT-3351]

eYinka:
Tabs like Attachments, Images, and Featured are currently in legacy hardwired form and they're toggled by boolean flags in `settings`; then added to the nav by some `if edition.allows_attachments?` checks in the helper. The goal of this spike was to understand whether we could instead declare them in `forms` (alongside dynamic tabs like Social Media Accounts), making the JSON config the single source of truth for all tabs on a StandardEdition.

As part of [WHIT-3351](https://gov-uk.atlassian.net/browse/WHIT-3351), I've made a quick prototype on how we can make these legacy tabs (Featured, Images, Attachements) expressed in the config-driven way; maintaining a minimal approach to support existing code.

In this approach, I have introduced `"linked_to": "foo"` in the `forms` object to declare a tab that routes to a separate controller rather than rendering inline.

I struggled with the naming here. Had to decide between `linked_tabs` and `external_tabs`. I thought `linked_tabs` will communicate the idea more effectively.

**Key things I learned:**

- The targeted approach works well with the dynamic tab (Understood as in-line tabs) pipeline with minimal changes and no new abstractions
- Linked tabs don't need `valid_tab_key?` changes, since they never submit via `?current_tab=`, so scoped validation is not a concern for them. **Or not?**
- We can safely (I guess) get rid of `settings.file_attachments_enabled`  in a configurable document type.

**Note:** The `type.settings["..."] == true` fallback on the Attachments line is intentional for transition purposes. Once all document type configs have been updated to declare their tabs via linked_to, the settings flags (file_attachments_enabled, features_enabled, etc) and the `||` fallbacks can be removed from the presenter and StandardEdition model methods.

To add other tabs, follow the approach in `news_story.json`. For example, to add features tab to News Story:

``` json

  "forms": {
    "documents": {
      "fields": { ... }
    },
    "attachments": {
      "label": "Attachments",
      "linked_to": "edition_attachments"
    },
    "features": {
      "label": "Features",
      "linked_to": "edition_features"
    },
  }
```
 
Then update the presenter: `app/presenters/publishing_api/standard_edition_presenter.rb`:

**Existing code:** 
`details.merge!(PayloadBuilder::Features.for(item)) if type.settings["features_enabled"] == true`

**Updated code:**
`details.merge!(PayloadBuilder::Features.for(item)) if type.settings["features_enabled"] == true || type.has_linked_tab?("edition_features")` 

**TODO:**
- [x]  Run test suites to catch edge cases
- [ ] Thorough testing on the UI

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-3351)

--- 

[WHIT-3351]: https://gov-uk.atlassian.net/browse/WHIT-3351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/helpers/admin/tabbed_nav_helper.rb
+++ b/app/helpers/admin/tabbed_nav_helper.rb
@@ -51,6 +51,7 @@ module Admin::TabbedNavHelper
           current: on_dynamic_tab && current_tab == tab["id"],
         }
       end
+      nav_items.concat(linked_tabs_nav_items(edition, current_path))
     end
   end
 
@@ -215,5 +216,21 @@ module Admin::TabbedNavHelper
         current: current_path == admin_worldwide_organisation_page_attachments_path(page),
       },
     ]
+  end
+
+  def linked_tabs_nav_items(edition, current_path)
+    edition.type_instance.linked_tabs.map do |tab|
+      href = href_for_linked_tab(tab, edition)
+      { label: tab["label"], href:, current: current_path == href }
+    end
+  end
+
+  def href_for_linked_tab(tab, edition)
+    builders = {
+      "edition_attachments" => -> { admin_edition_attachments_path(edition) },
+      "edition_images" => -> { admin_edition_images_path(edition) },
+      "edition_features" => -> { features_admin_standard_edition_path(edition, locale: edition.primary_locale) },
+    }
+    builders[tab["linked_to"]]&.call
   end
 end
--- a/app/models/configurable_document_types/news_story.json
+++ b/app/models/configurable_document_types/news_story.json
@@ -58,6 +58,10 @@
           "translatable": false
         }
       }
+    },
+    "attachments": {
+      "label": "Attachments",
+      "linked_to": "edition_attachments"
     }
   },
   "schema": {
@@ -122,7 +126,6 @@
       }
     },
     "send_change_history": true,
-    "file_attachments_enabled": true,
     "organisations": null,
     "backdating_enabled": true,
     "history_mode_enabled": true,
--- a/app/models/configurable_document_type.rb
+++ b/app/models/configurable_document_type.rb
@@ -113,6 +113,16 @@ class ConfigurableDocumentType
     end
   end
 
+  def linked_tabs
+    @linked_tabs ||= @forms.filter_map do |id, form|
+      { "id" => id, "label" => form["label"], "linked_to" => form["linked_to"] } if form["linked_to"]
+    end
+  end
+
+  def has_linked_tab?(tab_key)
+    linked_tabs.any? { |tab| tab["linked_to"] == tab_key }
+  end
+
   def schema_for_fields(field_keys)
     field_keys = field_keys.map(&:to_s)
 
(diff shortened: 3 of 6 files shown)


app/models/configurable_document_types/news_story.json
```diff
@@ -122,7 +126,6 @@
       }
     },
     "send_change_history": true,
-    "file_attachments_enabled": true,
```
ChrisBAshton, line 125: I think you're probably right that we can remove this. Just needs double-checking.

The presence of the Attachments tab doesn't fully describe the attachments behaviour we want for the content type: historically we sometimes allow HtmlAttachment, sometimes not, sometimes allow InlineAttachment, sometimes not, etc. I know `file_attachments_enabled` doesn't cover all that either, but we may still need some sort of toggling in the `settings` hash.

app/helpers/admin/tabbed_nav_helper.rb
```diff
@@ -216,4 +217,20 @@ def worldwide_organisation_page_nav_items(page, current_path)
...
   end
+
+  def linked_tabs_nav_items(edition, current_path)
+    edition.type_instance.linked_tabs.map do |tab|
+      href = href_for_linked_tab(tab, edition)
+      { label: tab["label"], href:, current: current_path == href }
+    end
+  end
+
+  def href_for_linked_tab(tab, edition)
```
ChrisBAshton, line 228: I can see why you went for 'linked tab', and why you considered 'external tab'.

Tbh I think 'hardcoded tab' is more expressive here (since they are!).

ChrisBAshton (commented):
I really like this approach! Think it's a solid improvement, and minimally invasive code-wise 🎉 Nice job ⭐ 

To answer some points in your PR:

> Linked tabs don't need `valid_tab_key?` changes, since they never submit via `?current_tab=`, so scoped validation is not a concern for them. Or not?

I think that's correct, yes. The hardcoded tabs basically already have their own validation scope.

> In this approach, I have introduced `"linked": true`

Have you? I only see `"linked_to": "<kind>"`

Btw, I can see us skipping the fallback you've added and just migrating all configurable content type schemas en-masse.

app/helpers/admin/tabbed_nav_helper.rb
eYinka: Fairs

eYinka:
@ChrisBAshton 

> Have you? I only see "linked_to": "<kind>"

My bad. I refactored to get rid of the `linked` property as it was redundant. Will update the description.

app/models/configurable_document_types/news_story.json
eYinka: Agreed. We will probably have to do something like this later on: 

``` json

  "settings": { 
    "attachments": { 
      "file": true, 
      "html": false 
    } 
  }

```

PR #11429, opened 2026-05-07
First publishing date [WHIT-3076]

davidsauntson:
## What

This PR:
- re-implements the previously merged `first_published_at` validation in #11393  
- adds validation to prevent users from backdating the `first_published_at` date before the start of the current government
- updates some of the test code to just build the test edition models rather than validate and persist them
- updates the default `first_published_at` date in the behavioural tests so it is in line with how the current government is created in those tests

## Why

#11393  was merged but reverted because of problematic interaction with history mode (see https://govuk.zendesk.com/agent/tickets/6565061), where:
- a user followed the new validation instructions 
- backdated an edition into a previous government
- the edition entered history mode
- the user could not edit or delete the edition because they didn't have sufficient privileges

This PR fixes that by adding the validation described above.

Some unit tests use FactoryBot's `#create` build strategy, which meant they failed the new validation.  Changing to `#build` allows the tests to assert on the same behaviour without requiring the models to be valid - this makes sense since the changed tests relate to behaviour of models in previous governments rather than specific details of persistence / associations that come with using `#create`(**AFAICT - it would be great if someone could check this** - it's in commit ~99c5882902921476245c192d4f11c7aaf9a86eba~ 69d30438a3574e8564d59bb899266fcc7d48f0c2 🙏🏻 )

---

### DIFF
--- a/test/unit/app/models/edition_test.rb
+++ b/test/unit/app/models/edition_test.rb
@@ -684,6 +684,14 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal "First published date must be between 1/1/1900 and the present", edition.errors.full_messages.first
   end
 
+  test "first_published_at cannot be after the date of the first change note" do
+    edition_with_change_note = create(:edition_with_document, :published, change_note: "changed", major_change_published_at: 2.days.ago)
+    edition = build(:edition, document: edition_with_change_note.document, first_published_at: 1.day.ago)
+
+    assert edition.invalid?
+    assert_equal "First published date must be before the first change note (09/11/2011 11:11)", edition.errors.full_messages.first
+  end
+
   test "#government returns the associated government when the edition has a specific government_id" do
     create(:current_government)
     previous_government = create(:previous_government)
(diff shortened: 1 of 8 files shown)


test/unit/app/models/edition_test.rb
```diff
@@ -692,6 +692,14 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal "First published date must be before the first change note (09/11/2011 11:11)", edition.errors.full_messages.first
   end
 
+  test "first_published_at cannot be before the start of the current government" do
+    create(:current_government)
+    edition = build(:edition, first_published_at: 10.years.ago)
+
+    assert edition.invalid?
+    assert_equal "First published date must be after the start of the current government (11/11/2009)", edition.errors.full_messages.first
+  end
```
ChrisBAshton, line 701: Maybe add a test that exercises the early return if the document is not political.

ChrisBAshton (approved):
Excellent work, David 👏 Thanks for signposting which commits were cherry-picked - saved me from reviewing them twice.

test/unit/app/models/edition_test.rb
davidsauntson: I've now added a test that covers the guard clause 👍🏻

PR #11431, opened 2026-05-07
Implement StandardEdition Parent/Child relationships with semi-independent publishing workflows [WHIT-3341]

ChrisBAshton:
# What

This PR implements a generic multi-page architecture and applies it to Topical Events and their 'About' pages (dependent on https://github.com/alphagov/publishing-api/pull/3860).

Child documents have largely independent publishing workflows, by design - this allows publishers to make changes to their child documents without having to do everything through the parent. They can't be _totally_ independent though, otherwise we could end up with orphaned children, live children with unpublished parents and so on. So we've encoded four simple, sensible rules that constrain the lifecycles and help keep states predictable and valid. See [Mural](https://app.mural.co/t/govukdelivery7534/m/govukdelivery7534/1769703142448/85c09ba01ccbad8ae78b6dc8a1654d48dc849c88) for some of the thinking.

### Rule 1: New child drafts

> A child document can only be created on a parent in a pre-publication state.

### Rule 2: Prevent orphaned children

> A parent draft may not be discarded while it has associated child documents whose only edition is linked to that parent draft.

We could offer to drop the draft documents automatically when dropping the parent. But the simpler solution for now is to raise a validation error and force the publisher to drop the child documents manually first.

### Rule 3: Child publishability

> A draft child may only be published if its parent _document_’s `live_edition` is "published" **and** the parent _edition_ it is linked to is not in a pre-publication state.

The simplest implementation, applied here, is to force the publisher to publish the parent (draft) document, then publish the child document.

In future, I expect we'd offer some sort of "publish all (or a subset of) draft children with the parent" feature in future, where the operation order would be that the parent is published first, thereby allowing said children to be published in turn. However, whilst we can surface that kind of UX on the Publish / Force-Publish pages, we can't do the equivalent for schedule-publishing until we tackle [WHIT-1887](https://gov-uk.atlassian.net/browse/WHIT-1887).

### Rule 4: Parent visibility ceiling

> A child document may not be more publicly visible than its parent document.

| Parent          | Allowed child states                             |
|-------------|-------------------------------------|
| published     | published / withdrawn / unpublished |
| withdrawn    | withdrawn / unpublished                     |
| unpublished | unpublished                                         |

Implementation wise, we **could** auto-cascade such that if you unpublish the parent, you unpublish its children. But it gets messy if there exists a draft child - do you auto-discard the draft in order to unpublish it? Etc. So to begin with at least, we will force the publisher to manually withdraw/unpublish (as appropriate) all of a document's children before withdrawing/unpublishing the parent document. This simplified interpretation can be summarised as "You cannot withdraw/unpublish a parent while any children remain more publicly visible than it."

## Out of scope (future tickets)

* Limiting the number of About pages that can be created.
* (if deemed necessary) Hard-coding the slug to "/about", rather than having it determined by title.
* Redirect locations. Currently when publishing a Child document, you're redirected to the Document Search screen, but should probably be redirected to the Parent summary screen instead? Same with the cancel link on the "New child document" form.
* Hide Topical Event About pages from Whitehall search (if needed)
* Offer option to - or just go ahead and action - deletion of 'new' child document when deleting its draft parent. (See Rule 2)
* Offer option to publish all or a subset of child pages at the same time as the parent (see Rule 3)
* "Through the parent" publishing, e.g. Publications & HTML Attachments, where the child document is 'part of the whole' and has a shared change history, shared last_updated timestamp, can't be edited independently, etc. Architecturally this should be possible by writing some additional constraints and config options on top of what we have here.

# Why

Whitehall has committed to migrating Topical Events in their entirety, to our config-driven 'StandardEdition' model. Unlike the already migrated content types, Topical Events aren't standalone content items - many of them have corresponding 'About' pages too, which need migrating.

Looking more widely, there are other microsite-like and multi-part document structures we hope to consolidate into StandardEdition in future, including:

- Publications (and HTML attachments)
- Plan for Change pages (and their children)
- Organisations (and their Corporate Information Pages)
- Travel Advice (summary and individual sections)
- Manuals (and manual sections)

So we've had to implement a multi-part architecture that can scale to accommodate all of those content types and their requirements. As in the 'out of scope' section, we're likely to need to add some configuration options and support 'top-down' / 'through-the-parent' publishing for some of them. We're also likely to need to introduce the concept of 'ordered' child pages (e.g. for HTML attachments). None of what we have built here locks us out from those future iterations.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3341

## Screenshots

We have a new "Child documents" section (hidden behind a feature flag) on Topical Event document summary pages:

> Screenshot 2026-05-14 at 16 07 25

Creating a new child document takes you to a screen where you can choose from the subset of allowed child document types:

> Screenshot 2026-05-14 at 16 08 26

You're taken to a StandardEdition new document form as normal - the only difference being the additional call-out reinforcing that this is being created as a child document of the given parent:

> Screenshot 2026-05-14 at 16 08 43

The newly created document is again, just a StandardEdition - but with a callout at the top of the page, linking it back to the parent. Note that the publish/force-publish button is deliberately hidden, since we need the parent to be published before we allow the 'first publish' of the child (subsequent publishes are unrestricted).

(Note also the warning about the same title/slug - a growing problem and a symptom of the fact our global uniqueness check logic isn't compatible with consolidating onto StandardEdition, as we check `document_type` and `slug` - in this case 'StandardEdition' and 'About' - and unnecessarily add a `--2`/`--3`/... suffix, as the base path prefix here is actually unique. This will be tackled in [WHIT-3371](https://gov-uk.atlassian.net/browse/WHIT-3371).)

> Screenshot 2026-05-14 at 16 09 30 note validation error and lack of publish button

The child document now appears on the parent document summary screen:

> Screenshot 2026-05-14 at 16 10 18

After publishing the parent, the callout text at the top of the child document is now simplified, and we can see the publish actions in the sidebar:

> Screenshot 2026-05-14 at 16 12 16 after publishing parent

After publishing the child document, it is visible in Whitehall Search. We may want to change this (out of scope for this PR):

> Screenshot 2026-05-14 at 16 14 35

---

[WHIT-1887]: https://gov-uk.atlassian.net/browse/WHIT-1887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/admin/editions/new.html.erb
```diff
@@ -16,6 +16,19 @@
         } %>
       </div>
     <% end %>
+
+    <%
+      if params[:parent_edition_id]
```
eYinka, line 21: Just wondering if this should be moved to the controller?

db/migrate/20260507125603_create_parent_child_relationships_table.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    # strong_migrations prefers indexes outside create_table
+    # rubocop:disable Rails/BulkChangeTable
+    add_index :parent_child_relationships,
+              %i[parent_edition_id child_document_id],
+              unique: true,
+              name: "idx_pcr_unique_parent_child"
+
+    add_index :parent_child_relationships,
+              %i[parent_edition_id ordering],
```
TonyGDS: need to remove the ordering here or the migration won't work

test/unit/app/models/parent_child_relationship_test.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    relationship = build(
+      :parent_child_relationship,
+      parent_edition: existing_relationship.parent_edition,
+      child_document: existing_relationship.child_document,
+    )
+
+    assert_not relationship.valid?
+  end
+
```
TonyGDS, line 42: Are we missing a test here for a child with multiple parents (different documents)?

db/migrate/20260507125603_create_parent_child_relationships_table.rb
ChrisBAshton: Good spot. Original implementation was with ordering, and then I simplified.

app/models/configurable_document_type.rb
```diff
@@ -59,6 +59,11 @@ def self.where_group(group)
     all.filter { |t| t.settings["configurable_document_group"] == group }
   end
 
+  def self.allowed_child_document_types_of(parent_edition)
```
lauraghiorghisor-tw, line 62: Nice one!

test/functional/admin/editions_controller_test.rb
```diff
@@ -50,6 +50,72 @@ class Admin::EditionsControllerTest < ActionController::TestCase
     assert_equal publication, assigns(:audit_trail_entry)
   end
 
+  test "attach_to_parent_if_this_is_a_child! creates relationship when parent_edition_id is present" do
```
lauraghiorghisor-tw: I can see these being a more regular-looking controller test (post :blah) in the future, maybe a view test once the flow is agreed. But quite nicely tested like this too 😎

app/controllers/admin/editions_controller.rb
```diff
@@ -92,6 +92,7 @@ def new; end
 
   def create
     if updater.can_perform? && @edition.save
+      attach_to_parent_if_this_is_a_child!
```
TonyGDS: I feel this would be better inside the `child_documents_controller`, the parent shouldn't always have to check if it's a child. I appreciate this would mean some duplication in the method, but I think that's ok for clarity.

public/configurable-document-type.schema.json
```diff
@@ -343,22 +356,14 @@
           "enum": [
             "history",
             "topical_event",
+            "topical_event_about_page",
```
lauraghiorghisor-tw, line 364: Can probably extract the list of schemas at the top too.

test/functional/admin/editions_controller_test.rb
```diff
@@ -50,6 +50,72 @@ class Admin::EditionsControllerTest < ActionController::TestCase
...
+    child_edition = create(:edition)
+
+    controller.params = ActionController::Parameters.new(
+      parent_edition_id: parent_edition.id,
+    )
+
+    controller.instance_variable_set(:@edition, child_edition)
+
+    assert_difference("ParentChildRelationship.count", 1) do
+      controller.send(:attach_to_parent_if_this_is_a_child!)
```
TonyGDS: should this not be a private method that we test through `:create`?

app/models/configurable_document_types/topical_event.json
```diff
@@ -177,6 +177,11 @@
     "publishing_api_schema_name": "topical_event",
     "publishing_api_document_type": "topical_event",
     "rendering_app": "frontend",
+    "allowed_child_document_types": [
+      {
```
lauraghiorghisor-tw, line 181: An object so we can add future options, nice 👌

app/models/concerns/standard_edition/parent_document.rb
```diff
@@ -28,6 +28,10 @@ def process_associations_after_save(new_edition)
     add_trait Trait
   end
 
+  def allows_child_documents?
```
lauraghiorghisor-tw, line 43: Do we need some additional validation here - maybe before save - to ensure that only allowed children types can be created? -> if `allows_child_documents?`, then validate against `allowed_child_document_types_of`? Doubling down on the check in the choose type controller action.

test/functional/admin/editions_controller_test.rb
```diff
@@ -50,6 +50,72 @@ class Admin::EditionsControllerTest < ActionController::TestCase
...
+    child_edition = create(:draft_publication)
+
+    controller.params = ActionController::Parameters.new(
+      parent_edition_id: parent_edition.id,
+    )
+    controller.instance_variable_set(:@edition, child_edition)
+    controller.stubs(:enforce_permission!).with(:update, parent_edition).raises(Whitehall::Authority::Errors::PermissionDenied.new(:update, parent_edition))
+
+    assert_raise(Whitehall::Authority::Errors::PermissionDenied) do
+      controller.send(:attach_to_parent_if_this_is_a_child!)
```
TonyGDS: test through `:create`

test/functional/admin/editions_controller_test.rb
```diff
@@ -50,6 +50,72 @@ class Admin::EditionsControllerTest < ActionController::TestCase
...
+
+  test "attach_to_parent_if_this_is_a_child! does nothing when parent_edition_id is absent" do
+    child_edition = create(:edition)
+
+    controller.params = ActionController::Parameters.new
+
+    controller.instance_variable_set(:@edition, child_edition)
+
+    assert_no_difference("ParentChildRelationship.count") do
+      controller.send(:attach_to_parent_if_this_is_a_child!)
```
TonyGDS: test through `:create`

test/functional/admin/editions_controller_test.rb
```diff
@@ -50,6 +50,72 @@ class Admin::EditionsControllerTest < ActionController::TestCase
...
+
+    controller.instance_variable_set(:@edition, child_edition)
+
+    parent_edition.expects(:with_lock).once.yields
+
+    Edition.stubs(:find)
+          .with(parent_edition.id)
+          .returns(parent_edition)
+
+    controller.send(:attach_to_parent_if_this_is_a_child!)
```
TonyGDS: test through `:create`

app/controllers/admin/editions_controller.rb
```diff
@@ -532,4 +533,19 @@ def redirect_param(fallback: nil)
...
   end
+
+  def attach_to_parent_if_this_is_a_child!
+    parent_id = params[:parent_edition_id].presence
+    return unless parent_id
+
+    parent = Edition.find(parent_id)
+    enforce_permission!(:update, parent)
+
+    parent.with_lock do
```
TonyGDS: Maybe this logic in future actually belongs in the model, given we need to run rake tasks etc. If we don't go through the controller we'd miss this guard agains the state of the parent changing

app/controllers/admin/child_documents_controller.rb
```diff
@@ -0,0 +1,24 @@
+class Admin::ChildDocumentsController < Admin::EditionsController
+  rescue_from ConfigurableDocumentType::NotFoundError, with: :render_not_found
+
+  def choose_type
+    @permitted_child_document_types = ConfigurableDocumentType.allowed_child_document_types_of(StandardEdition.find(params[:parent_edition_id])).select { |type| can?(current_user, type) }
```
lauraghiorghisor-tw, line 5: Settings for parent and child could fall out of sync across schemas (human error). Could add specific validation to ensure settings match, or use parent's setting, or fall back to parent's when child setting nil.

Do we have any validation here to say that the orgs passed into this setting, for restriction, should also be listed in the org association?

app/models/concerns/standard_edition/child_document.rb
```diff
@@ -11,9 +13,29 @@ module StandardEdition::ChildDocument
     has_one :parent_edition,
```
TonyGDS, line 13: Is this guaranteed to always return the latest edition for the document id, given the document may have more than one edition e.g. a published edition and a new draft?

app/models/concerns/standard_edition/child_document.rb
```diff
@@ -11,9 +13,29 @@ module StandardEdition::ChildDocument
...
   end
 
   def is_child_document?
     parent_edition.present?
   end
+
+  def can_be_published?
+    return true unless is_child_document?
+
+    parent_edition.document.live_edition&.state == "published" && !parent_edition.pre_publication?
```
TonyGDS, line 27: works if we can guarantee `parent_edition` is the latest edition.

app/models/concerns/standard_edition/child_document.rb
```diff
@@ -11,9 +13,29 @@ module StandardEdition::ChildDocument
...
             source: :parent_edition
+
+    before_update :ensure_publishing_is_allowed!, if: :publishing?
   end
 
   def is_child_document?
     parent_edition.present?
   end
+
+  def can_be_published?
```
TonyGDS: `parent_allows_publishing`?

app/controllers/admin/editions_controller.rb
```diff
@@ -90,9 +90,25 @@ def show
...
+  # That seems to be how we handle this sort of thing elsewhere.
   def create
-    if updater.can_perform? && @edition.save
-      updater.perform!
+    success =
+      if updater.can_perform?
+        begin
+          ActiveRecord::Base.transaction do
+            @edition.save!
+            attach_to_parent_if_this_is_a_child!
```
lauraghiorghisor-tw: Since we already have a `ChildDocumentsController` which inherits from this `EditionsController`, I wonder if it reads better to override this action in the child controller, call super, then call the `attach_to...`. The edition creation itself is separate from what the `attach_to...` does, which is saving a different table altogether.

lauraghiorghisor-tw: Since we already have a `ChildDocumentsController` which inherits from this `EditionsController`, I wonder if it reads better to override this action in the child controller, call `super`, then call the `attach_to...`. The edition creation itself is separate from what the `attach_to...` does, which is saving a different table altogether.

app/models/concerns/standard_edition/parent_document.rb
```diff
@@ -0,0 +1,80 @@
...
+    has_many :child_relationships,
+             class_name: "ParentChildRelationship",
+             foreign_key: :parent_edition_id,
+             inverse_of: :parent_edition,
+             dependent: :destroy
+
+    has_many :child_documents,
+             lambda {
+               joins(:editions)
+                 .merge(Edition.all) # applies default scope (i.e. excludes deleted)
```
lauraghiorghisor-tw, line 31: You can't dependent destroy the child because they're linked as document IDs, and would delete all entries in the relationship table, right? - so you need to solve the child deletion with scope 🤔

app/models/standard_edition.rb
```diff
@@ -103,7 +105,16 @@ def change_note_required?
   end
 
   def base_path
-    "#{type_instance.settings['base_path_prefix']}/#{slug}"
+    # Would be nice to restrict all of the child document logic to the ChildDocument
```
lauraghiorghisor-tw, line 108: Apparently we could do:

```
  module BasePath
    def base_path
      return super unless is_child_document?

      "#{parent_edition.base_path}/#{slug}"
    end
  end

  included do
    prepend BasePath
    [...]
  end
```

or

```
included do
[...]

    prepend(Module.new do
      def base_path
        return super unless is_child_document?

        "#{parent_edition.base_path}/#{slug}"
      end
    end)
```

app/models/standard_edition.rb
```diff
@@ -103,7 +105,16 @@ def change_note_required?
...
+    # concern, but we can't override base_path in there and fall back to the generic
+    # implementation here. So here it lives!
+    base_path_prefix = if is_child_document?
+                         parent_edition.base_path
+                       else
+                         type_instance.settings["base_path_prefix"]
+                       end
+
+    "#{base_path_prefix}/#{slug}"
```
lauraghiorghisor-tw, line 117: Are we asking this question in any past or future UR? -> whether to introduce a dynamic slug for about page? 

Also,  think this needs a change in the FE router if we go with the title-based slugs https://github.com/alphagov/frontend/blob/df18d3eee51ccd5836070962ae4daa62b5242b02/config/routes.rb#L142

app/presenters/publishing_api/payload_builder/configurable_document_links.rb
```diff
@@ -7,6 +7,12 @@ def self.for(item)
       }.compact
     end
 
+    def self.parent(item)
+      if item.is_child_document?
```
lauraghiorghisor-tw, line 11: I'd argue that if the link is there the presenter should just send it (handle the nil, but without referencing `is_child_document` inside here).

app/views/admin/editions/show/_main.html.erb
```diff
@@ -118,6 +118,36 @@
     </section>
   <% end %>
 
+  <% if @edition.allows_child_documents? && Flipflop.enabled?(:configurable_document_types) %>
```
lauraghiorghisor-tw, line 121: Nice toggle. We could toggle more of the code - e.g. protect the controller actions as well - but... this might just suffice 🤔

db/migrate/20260507125603_create_parent_child_relationships_table.rb
lauraghiorghisor-tw: Ah, man, this got my DB in a 🥒

features/step_definitions/standard_edition_child_pages_steps.rb
```diff
@@ -0,0 +1,101 @@
...
+
+Then("I am on the document creation screen") do
+  assert_current_path(/^\/government\/admin\/standard-editions\/new\?parent_edition_id=#{@standard_edition.id}&configurable_document_type=topical_event_about_page$/)
+end
+
+Then(/^I can see a "(.+)" callout$/) do |callout|
+  assert page.has_css?(".govuk-inset-text", text: callout)
+end
+
+Then("when I fill in and create the child document") do
```
lauraghiorghisor-tw, line 51: Speaking of, what changes are needed to pub API for this to actually save? Should this PR come with a link to a publishing API PR?

I had to set links and remove the requirement for the read more link to make it work locally 🤔 Although the links bit was a bit confusing because it looked like it _should_ have been inheriting some links (with parent) from whatever generic definitions, but then there just weren't any in the generated schema.

public/configurable-document-type.schema.json
```diff
@@ -322,6 +336,10 @@
       "type": "object",
       "description": "A list of configurations to enable certain edition behaviours.",
       "properties": {
+        "is_child_document": {
```
lauraghiorghisor-tw, line 339: I know the suffixed slug issue isn't solved yet  and the proposed fix is to save base path in the DB, but just noting as an FYI here, that not setting `base_path_prefix`  causes some weird behaviours on the slug clashing checks in the `identifiable` concern - the `where_base_path_prefix_matches` ends up only checking against other types where `base_path_prefix` is `nil`) - which is sort of wrong, although it ends up checking against its own nil-base-path type implicitly (and all other types that might have nil base path in the schema).

lauraghiorghisor-tw: I am also not entirely sold on the saving of the base path prefix solution (NO. MORE. SLUGS.) I think we could - for now at least? - adjust the clash checking logic for our new child usecase. Think this works:
```
    def where_base_path_prefix_matches(edition)
      scope = Edition.where(type: edition.type)
      if edition.configurable_document_type.present?
            if edition.respond_to?(:is_child_document?) && edition.is_child_document?
          parent_edition = edition.parent_edition
               if parent_edition
               scope = scope.joins("INNER JOIN parent_child_relationships ON parent_child_relationships.child_document_id = editions.document_id")
                         .where(parent_child_relationships: { parent_edition_id: parent_edition.id })
          end
        else
          configurable_types = ConfigurableDocumentType.find_by_base_path_prefix(edition.type_instance.settings["base_path_prefix"])
          scope = scope.where(configurable_document_type: configurable_types.map(&:key))
        end
      end
      scope
    end
```

Basically, if child, the only relevant scope clash is within the parent.

app/models/concerns/standard_edition/child_document.rb
lauraghiorghisor-tw: Think we pass the edition ID as a query param, we never deduce it from the parent doc

app/models/concerns/standard_edition/child_document.rb
```diff
@@ -0,0 +1,41 @@
...
+  class IncompatibleParentState < ::WhitehallError; end
+
+  included do
+    has_one :parent_relationship,
+            class_name: "ParentChildRelationship",
+            foreign_key: :child_document_id,
+            primary_key: :document_id,
+            inverse_of: :child_document
+
+    has_one :parent_edition,
```
lauraghiorghisor-tw, line 13: Just thinking about the data a bit here. So throughout the life cycle of a child, the parent edition will always point to that edition on which the child got created? Even 50 redrafts of parent or child later, the entry in the relationships table will be `(superseded) edition ID, and (same) child document ID` 🤔 Not saying it's bad, just trying to think if that's useful for debugging etc., should we update any if it as we redraft?

I am inclined to think it is OK as is, i.e. that "snapshot" of data is correct at that point in time, and persists that trace.

features/standard-edition-child-pages.feature
```diff
@@ -0,0 +1,44 @@
+Feature: Standard Editions - Child Pages
```
lauraghiorghisor-tw, line 1: Sweet feature tests 😍

lauraghiorghisor-tw (commented):
Nice one, @ChrisBAshton 😍 The changes seem minimal and well encapsulated. Minimal if-else-ing on the views as well (I thought it would have to be far worse ngl). Few suggestions and musings here and there, but generally 👍

My only concern is around the flow, as I've already mentioned -> the fact that the parent HAS to go live before the child can be published, and how that might be confusing to publishers. But with the underlying models, and the rule commits last, I think we can rejig the logic if we have to. And scheduling - or some "cascade publish" ™️ feature that is only available for parents 👀 -  could override the rule in the future 😎

app/models/standard_edition.rb
lauraghiorghisor-tw: I have now made a note of this question in the [UCD doc](https://docs.google.com/document/d/1hd2vJHSv1XlLuDvT08YIgxmtH992a3_mDD09QBjCGkI/edit?tab=t.0) and also the [UR questions doc](https://docs.google.com/document/d/1A2TsYhFQhzCP3a8YN1OPllQMg2MleY7vUUs1W8Ylhzk/edit?tab=t.0).

features/step_definitions/standard_edition_child_pages_steps.rb
ChrisBAshton: Yep, this is linked in the PR description: https://github.com/alphagov/publishing-api/pull/3860

lauraghiorghisor-tw: I missed that 🥲

app/models/concerns/standard_edition/parent_document.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+    end
+  end
+
+  included do
+    has_many :child_relationships,
+             class_name: "ParentChildRelationship",
+             foreign_key: :parent_edition_id,
+             inverse_of: :parent_edition,
+             dependent: :destroy
```
lauraghiorghisor-tw, line 26: From my local testing, it appears that even if I drop a parent draft it does not remove the row from the `ParentChildRelationship` - am I misreading this?

app/views/admin/editions/show/_main.html.erb
```diff
@@ -118,6 +118,36 @@
...
+    <section class="app-view-summary__section app-view-summary__section--child_documents">
```
lauraghiorghisor-tw, line 122: I was just thinking about this, the multiple children types allowed. That would definitely need to be well validated with IA. At the minute we're sort of half-agreed with the very flexible schema idea, and still unclear on whether there needs to be a split between parent and child config. 

If it's all one big type of thing, then you never really need additional screens for the child type selection, it's all one potentially larger type with optional fields/tabs or something. We can skip some screens if allowed list count is 1. If you need variation between children, what do we name them, _1 and _2? 😅

Would this implementation support a self-reference case, if we don't need to split between parent and child? For example a `landing_page` has an allowed set of children with a single `landing_page` in it? 

So far I have heard that a split will help with sending distinct analytics data. On the other hand, it's likely that for the microsite case we won't need to render the `document_type` on the screen so the distinction between parent and child types is irrelevant. We'll chat about some of these on Wed in the tech session 🕵🏻‍♀️

app/models/concerns/standard_edition/child_document.rb
lauraghiorghisor-tw: In terms of how the references get copied over on redraft, it's just the normal `create_draft` logic on the edition model doing it, right? 🤔

TonyGDS:
Right now we have a have a `has_one` relationship between parent and child and when we patch link set links (on save draft) it updates the live version. It works now because we assume the parent will always be the same for all editions of the child. 
If we decided to try to use the same mechanism for Document Collections, A document in a collection could potentially have multiple parents and also the parent could change (meaning it could mess the reverse link from the child). Something to consider for Collections.

ChrisBAshton:
I very much anticipate this for modelling classic parent/child relationships _only_, @TonyGDS . Document Collections are out of scope and would be handled differently. I should have added a `Rule 0: a Child document only ever has one parent`. (In fact I think we've documented this rule elsewhere in Murals etc and it's now just an unspoken one, but I can make it clearer in the PR description).

app/models/concerns/standard_edition/child_document.rb
ChrisBAshton: Indeed, Laura.

It is possible for the publisher to change the parent_edition_id param in the URL. I've added validation (to ParentChildRelationship, in the `Rule 1: New child drafts` commit) to ensure the edition ID corresponds to a draft parent that is configured to allow child documents, and that the publisher has the right to edit the parent. So yes it should be pretty much guaranteed that `parent_edition` is the latest edition for the document at the time.

app/models/concerns/standard_edition/child_document.rb
ChrisBAshton: > In terms of how the references get copied over on redraft, it's just the normal `create_draft` logic on the edition model doing it, right? 🤔

Yep, in `StandardEdition::ParentDocument`, introduced in the "Add process_associations_after_save on Parent edition creation" commit.

> So throughout the life cycle of a child, the parent edition will always point to that edition on which the child got created?

Yep. Is it useful? I think so - we'd have to refer to the Mural for the alternatives.

1. Document -> Document relationship: simplest to model, but breaks audit trail entirely. Could go to Parent Edition number 1 and it'll just surface all the child documents that have ever been created, even if they were added on Parent Edition 50.
2. Edition -> Edition relationship: most complex to model, requiring duplicating records on draft creation of parent or child, and quickly gets hairy if we allow creating a draft child on a published parent (as you could then create a draft parent, meaning _another_ ParentChildRelationship row, and then if you discard either one you need to somehow signal the previous ParentChildRelationship as now being the latest, etc...). And then it's quite difficult to actually 'execute' its usefulness. If any edition can be created at any time, how can we say which child editions were published 'alongside' a given parent? If they're all independent anyway, what does it give us?

Which leaves us with:

3. Edition -> Document relationship: least worst option. It's an append-only history: we can see for any given parent edition, exactly which child documents were in the system at the time. We can cross-reference each document's edition histories and stitch together a consolidated timeline if necessary. Everything is 'parent-led' so it's not too difficult to keep a mental model or follow a thread.

app/models/concerns/standard_edition/child_document.rb
ChrisBAshton: I believe so, as per https://github.com/alphagov/whitehall/pull/11431/changes/BASE..f3a86585ec2cbd8aae5804df6e9623662226f3ec#r3267632414

app/models/concerns/standard_edition/parent_document.rb
ChrisBAshton: That's because dropping a draft only soft-deletes. If you actually `destroy` the parent, this relationship record is deleted too.

```
$ ParentChildRelationship.where(parent_edition_id: edition.id)
=> [#<ParentChildRelationship:0x0000ffff83786ff8
  id: 9,
  parent_edition_id: 1773361,
  child_document_id: 612060,
  ordering: nil,
  created_at: "2026-05-19 16:59:49.646612000 +0100",
  updated_at: "2026-05-19 16:59:49.646612000 +0100">]

$ edition.destroy

$ ParentChildRelationship.where(parent_edition_id: edition.id)
=> []
```

app/models/concerns/standard_edition/parent_document.rb
ChrisBAshton: Yeah I'm not seeing an issue here? Worked example:

1. Parent Edition 1, Child Edition 2, Child Document ID 7000. ParentChildRelationship(parent_edition_id: 1, child_document_id: 7000)
2. Create new child draft (Edition 3). No new ParentChildRelationship records are created, since the relationship records are only created on _parent_ draft creation.
3. Call `destroy`  on the child draft. No harm done. As far as ParentEdition 1 is concerned, it still has a Child Document 7000, irrespective of the child's edition.

I guess it's a bit different if we're talking the _first_ child edition:

1. Parent Edition 1
2. Create a draft child document. ParentChildRelationship(parent_edition_id: 1, child_document_id: 7000)
3. Call `destroy`  on the child draft. No harm done. As far as ParentEdition 1 is concerned, it still has a Child Document 7000, even though in Whitehall it no longer exists.

So maybe I _do_ need some sort of callback where if a child edition getting `destroy`ed (or even a child edition getting soft-`delete`ed) should delete the corresponding `ParentChildRelationship`? 🤔

app/models/configurable_document_types/topical_event.json
ChrisBAshton: Yep, I was thinking `ordering` or `required` or `maxChildrenOfThisType` or `consolidateChangenotes` or all sorts of other things will come into the mix at some point!

test/unit/app/models/parent_child_relationship_test.rb
ChrisBAshton: Have pushed one up now 👍

public/configurable-document-type.schema.json
ChrisBAshton: Agreed. Future PR!

app/controllers/admin/editions_controller.rb
ChrisBAshton: Agreed. I think I'd missed that ChildDocumentsController inherits from EditionsController and can hook into the update method :)

ChrisBAshton: Oh wait, it's not as straightforward as that. Have moved it into StandardEditionsController instead though, as that seems to be the best place. And have refactored the tests to test via create as suggested 👍

app/controllers/admin/editions_controller.rb
ChrisBAshton: Have moved to StandardEditionsController 👍 it's not appropriate in ChildDocumentsController as it's going through the standard document creation flow. We'd have to conditionally route to either this controller or the latter, depending on the kind of doc being created.

app/views/admin/editions/new.html.erb
ChrisBAshton: Yep 👍 can be tackled in a future PR once the designs are nailed in https://gov-uk.atlassian.net/browse/WHIT-3332

app/controllers/admin/editions_controller.rb
ChrisBAshton: As in a `parent.create_child_relationship!` method? Good shout.

test/functional/admin/editions_controller_test.rb
ChrisBAshton: Now implemented 👍

app/models/standard_edition.rb
ChrisBAshton: Out of scope for this PR 👍

public/configurable-document-type.schema.json
ChrisBAshton: Raised https://gov-uk.atlassian.net/browse/WHIT-3494 for this

app/models/concerns/standard_edition/parent_document.rb
ChrisBAshton: Raised https://gov-uk.atlassian.net/browse/WHIT-3501

app/controllers/admin/child_documents_controller.rb
ChrisBAshton: YAGNI at the moment. We've agreed to let organisations be set individually per parent/child. Out of scope for this PR

lauraghiorghisor-tw: Oh - I would have thought the implication is the opposite. We said we would allow setting any orgs on child but we won't allow setting access limiting on child. This bit here is more about access so my assumption would be that the org restriction should be based on parent (if access limiting is too)? But we did not touch specifically on org-based access limitations in the session 🤔

app/controllers/admin/editions_controller.rb
lauraghiorghisor-tw: If the flow we end up with is not necessarily going through the new form - none of Sumi's designs are atm - then routing as you say will just make sense.

lauraghiorghisor-tw:
> I very much anticipate this for modelling classic parent/child relationships _only_

I do wonder if that's going to be the case though 😅 It might just be that at that point extending this relationship makes most sense.  Either way, I agree with _only_ for now, because trying to cater for all types in this implementation, from the start, is too much.

I like the idea of Rule 0 for clarity 👍

PR #11432, opened 2026-05-07
[WHIT-3323] Keep page URL by default when re-editing a published document

TonyGDS:
## Summary

- Populate `slug_override` on new drafts so the page URL is preserved by default when a writer re-edits a published document.
- "Keep the current page URL" checkbox now renders ticked on title change; writers explicitly untick to opt the URL into the rename.

### DIFF
--- a/app/assets/javascripts/admin/views/edition-form.js
+++ b/app/assets/javascripts/admin/views/edition-form.js
@@ -160,25 +160,25 @@
   EditionForm.prototype.setupTitleInputEventListener = function () {
     const form = this.module
     const titleInput = form.querySelector('#edition_title')
-    const checkboxContainer = form.querySelector('.js-keep-slug-form-group')
+    const keepSlugControls = form.querySelector('.js-keep-slug-form-group')
 
-    if (!titleInput || !checkboxContainer) {
+    if (!titleInput || !keepSlugControls) {
       return
     }
 
     const originalTitle = titleInput.value
 
-    const setCheckboxVisibility = () => {
+    const setKeepSlugControlsVisibility = () => {
       const titleHasChanged = originalTitle !== titleInput.value
       if (titleHasChanged) {
-        checkboxContainer.removeAttribute('hidden')
+        keepSlugControls.removeAttribute('hidden')
       } else {
-        checkboxContainer.setAttribute('hidden', 'hidden')
+        keepSlugControls.setAttribute('hidden', 'hidden')
       }
     }
 
-    titleInput.addEventListener('input', () => setCheckboxVisibility())
-    setCheckboxVisibility()
+    titleInput.addEventListener('input', () => setKeepSlugControlsVisibility())
+    setKeepSlugControlsVisibility()
   }
 
   Modules.EditionForm = EditionForm
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -192,7 +192,7 @@ class Edition < ApplicationRecord
                                     force_published
                                     scheduled_publication]
       draft_attributes = attributes.except(*ignorable_attribute_keys)
-        .merge("state" => "draft", "creator" => user, "previously_published" => previously_published)
+        .merge("state" => "draft", "creator" => user, "previously_published" => previously_published, "slug_override" => slug)
 
       self.class.new(draft_attributes).tap do |draft|
         traits.each { |t| t.process_associations_before_save(draft) }
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -192,7 +192,7 @@ class Edition < ApplicationRecord
                                     force_published
                                     scheduled_publication]
       draft_attributes = attributes.except(*ignorable_attribute_keys)
-        .merge("state" => "draft", "creator" => user, "previously_published" => previously_published, "slug_override" => slug)
+        .merge("state" => "draft", "creator" => user, "previously_published" => previously_published)
 
       self.class.new(draft_attributes).tap do |draft|
         traits.each { |t| t.process_associations_before_save(draft) }
--- a/app/views/admin/editions/_page_address_controls.html.erb
+++ b/app/views/admin/editions/_page_address_controls.html.erb
@@ -13,17 +13,20 @@
   </div>
 
   <div class="<%= (edition.document.live_edition.slug == edition.slug_from_title) && "js-keep-slug-form-group" %> govuk-!-margin-bottom-6">
-    <%= render "govuk_publishing_components/components/hint", {
-      text: "The title of this document has changed since the page URL was generated. Check the box below to keep the existing URL, or leave it unchecked to update the URL for this page to match the new title when you publish this edition.",
-    } %>
-    <%= hidden_field(:edition, :slug_override, value: "") %>
-    <%= render "govuk_publishing_components/components/checkboxes", {
+    <%= render "govuk_publishing_components/components/radio", {
+      heading: "Page URL",
+      hint: "The title of this document has changed since the page URL was generated. Choose whether to keep the existing URL, or update it to match the new title when you publish this edition.",
       name: "edition[slug_override]",
       items: [
         {
-          label: "Keep the current page URL",
           value: edition.document.live_edition.slug,
+          text: "Keep the current page URL (#{edition.document.live_edition.public_url})",
           checked: edition.slug_override.present?,
+        },
+        {
+          value: "",
+          text: "Update the page URL to match the new title",
+          checked: edition.slug_override.blank?,
         },
       ],
     } %>
--- a/app/views/admin/editions/_page_address_controls.html.erb
+++ b/app/views/admin/editions/_page_address_controls.html.erb
@@ -1,15 +1,16 @@
 <% if edition.document&.live_edition.present? %>
   <div class="app-view-edit-edition__page-address govuk-!-margin-bottom-4">
     <%= render "govuk_publishing_components/components/heading", {
-      text: "Live page address",
+      text: "URL",
       heading_level: 3,
       font_size: "m",
       margin_bottom: 2,
     } %>
 
-    <%= render "govuk_publishing_components/components/hint", {
-      text: edition.document&.live_edition.public_url,
-    } %>
+    <p class="govuk-body">
+      Current GOV.UK URL:<br>
+      <%= link_to edition.document.live_edition.public_url, edition.document.live_edition.public_url, class: "govuk-link" %>
+    </p>
   </div>
 
   <div class="<%= (edition.document.live_edition.slug == edition.slug_from_title) && "js-keep-slug-form-group" %> govuk-!-margin-bottom-6">
--- a/app/views/admin/editions/_page_address_controls.html.erb
+++ b/app/views/admin/editions/_page_address_controls.html.erb
@@ -14,19 +14,21 @@
   </div>
 
   <div class="<%= (edition.document.live_edition.slug == edition.slug_from_title) && "js-keep-slug-form-group" %> govuk-!-margin-bottom-6">
+    <p class="govuk-body">
+      The title has changed since this URL was created. Choose whether to keep the current URL or update it to match the title.
+    </p>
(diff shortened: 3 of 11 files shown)


app/assets/javascripts/admin/views/edition-form.js
```diff
@@ -160,25 +160,25 @@
   EditionForm.prototype.setupTitleInputEventListener = function () {
     const form = this.module
     const titleInput = form.querySelector('#edition_title')
-    const checkboxContainer = form.querySelector('.js-keep-slug-form-group')
+    const keepSlugControls = form.querySelector('.js-keep-slug-form-group')
```
lauraghiorghisor-tw, line 163: Not overly sold on the _Controls name, but I get what it means.

lauraghiorghisor-tw: Also on commit message 1 - what do we mean with the dormant control?

app/models/edition.rb
```diff
@@ -192,7 +192,7 @@ def create_draft(user, allow_creating_draft_from_deleted_edition: false)
                                     force_published
                                     scheduled_publication]
       draft_attributes = attributes.except(*ignorable_attribute_keys)
-        .merge("state" => "draft", "creator" => user, "previously_published" => previously_published)
+        .merge("state" => "draft", "creator" => user, "previously_published" => previously_published, "slug_override" => slug)
```
lauraghiorghisor-tw: One of the main con of this (PR's) approach, when we discussed, was that it now sets an override all the time, introducing "misleading" data in the DB - overrides that are not overrides. 

Would be great if we could restrict the setting of the override, maybe by using the same check we use to hide the radio buttons - if live slug == SFT.

Problem with that would be that you haven't yet got the fresh SFT for your draft here... and live SFTs are historically unreliable so comparing the self fields here is no good.

But actually, shoving slug logic in here feels a bit inappropriate so am actually wondering if the best place is actually our identifiable concern, the slug setting callbacks -> and with that, you'd have the fields available. But that's also sounding a bit circular now - we set SO = slug, then... slug to SO? Maybe it can just live here after all 🥲

app/assets/javascripts/admin/views/edition-form.js
lauraghiorghisor-tw: And could we also capture in the commit message that we are slightly changing the behaviour of the override - from an "always override" to um... whatever it's doing now. Would be good to run this by Chris too if this version of the code is final.

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,23 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal [edition_with_slug], Edition.with_title_containing("https://www.gov.uk/government/news/klingons-rule?foo=bar#details")
   end
 
+  test "#create_draft populates slug_override with the parent's slug when the parent has none" do
```
lauraghiorghisor-tw: I'd go for "published" edition rather than "parent".

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,23 @@ class EditionTest < ActiveSupport::TestCase
...
+    published_edition = create(:published_edition, title: "Klingons rule")
+    assert published_edition[:slug_override].blank?
+
+    draft = published_edition.create_draft(create(:writer))
+
+    assert_equal published_edition.slug, draft.slug_override
+  end
+
+  test "#create_draft preserves an existing slug_override on the parent" do
+    published_edition = create(:published_edition, title: "Klingons rule", slug_override: "kept-url")
```
lauraghiorghisor-tw: Without a flow of steps to show that you have "kept" the url, it doesn't communicate too much - maybe just go for "custom-url" or the trusty "slug-override" 😅

features/edition-update-slug.feature
```diff
@@ -16,3 +16,84 @@ Feature: Edition update slug
     And I opt out of keeping the live slug
     And I save the edition and go to the document summary
     Then I can see the preview URL of the publication "Ten facts that will shock you" contains "ten-facts-that-will-shock-you"
+
+  Scenario: Editor saves a published document (saved live slug, previously changed title)
```
lauraghiorghisor-tw: I suppose these change in later commits but try this format -
`Scenario: Editor redrafts and saves a published document with custom slug, and opts to keep the live slug` or something

features/edition-update-slug.feature
```diff
@@ -1,21 +1,54 @@
...
+    When I change the title to "Ten facts that will shock you"
+    Then I can see the option to keep the current page URL
+
   Scenario: Writer updates the title of a document and keeps the live slug
     Given I am a writer
     And a published publication "You will never guess" exists
     When I edit the publication "You will never guess"
     And I change the title to "Ten facts that will shock you"
-    And I save the edition and go to the document summary
+    And I save my changes to the publication
```
lauraghiorghisor-tw, line 33: If this is useful, it should probably be on all tests where you care to see what happens immediately after redrafts - before save. Or just remove from all now.

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,23 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
     assert_select "legend", text: "Review date"
   end
 
+  view_test "GET edit shows the current GOV.UK URL as a link when the edition has a live edition" do
```
lauraghiorghisor-tw, line 189: Just to check, this is only for the live link, not re-introducing it for draft?

app/views/admin/editions/_page_address_controls.html.erb
```diff
@@ -1,15 +1,16 @@
...
       heading_level: 3,
       font_size: "m",
       margin_bottom: 2,
     } %>
 
-    <%= render "govuk_publishing_components/components/hint", {
-      text: edition.document&.live_edition.public_url,
-    } %>
+    <p class="govuk-body">
+      Current GOV.UK URL:<br>
```
lauraghiorghisor-tw, line 11: I still find "current" confusing 🫤 esp. in this case which is strictly about the live URL. It is not the current url unless chosen so as override.

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,71 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    assert_response :ok
+    refute_select "label", text: /Keep the current page URL/
+  end
+
+  view_test "GET edit does not show the keep current URL option for a draft with no live edition" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type")
```
lauraghiorghisor-tw, line 203: do you need the orgs?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,71 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    assert_response :ok
+    refute_select "label", text: /Keep the current page URL/
+  end
+
+  view_test "GET edit pre-selects the keep current URL option when slug_override is set" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, :with_organisations, configurable_document_type: "test_type")
```
lauraghiorghisor-tw, line 215: Same about the orgs?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,71 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type", document: published_edition.document, slug_override: published_edition.slug)
```
lauraghiorghisor-tw, line 216: Would `create_draft` not work here?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,71 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the update URL option when slug_override is blank" do
```
lauraghiorghisor-tw, line 226: Blank or nil?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,71 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type")
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    refute_select "label", text: /Keep the current page URL/
+  end
+
+  view_test "GET edit pre-selects the keep current URL option when slug_override is set" do
```
lauraghiorghisor-tw, line 211: Put the option name in quotes.

test/functional/admin/standard_editions_controller_test.rb
lauraghiorghisor-tw: Or have tests for both for now?

features/edition-update-slug.feature
```diff
@@ -1,132 +1,51 @@
...
-    And I save the edition and go to the document summary
-    And I can see the preview URL of the publication "Ten facts that will shock you" contains "ten-facts-that-will-shock-you"
-    And I force publish the publication "Ten facts that will shock you"
-    And I edit the publication "Ten facts that will shock you"
-    And I change the title to "Remember this person from the 90's"
-    And I opt out of keeping the live slug
-    And I save the edition and go to the document summary
-    Then I can see the preview URL of the publication "Remember this person from the 90's" contains "remember-this-person-from-the-90s"
\ No newline at end of file
+  Rule: The URL control only appears when the URL can be changed
```
lauraghiorghisor-tw, line 3: Not sure I love "the URL control".

features/edition-update-slug.feature
```diff
@@ -1,132 +1,51 @@
...
+  Rule: Renaming a draft keeps the live URL unless an editor opts out
+
+    Scenario: Renaming a draft keeps the live URL by default
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      When I edit the publication "You will never guess"
+      And I change the title to "Ten facts that will shock you"
+      And I save the edition and go to the document summary
+      Then I can see the preview URL of the publication "Ten facts that will shock you" contains "you-will-never-guess"
+      And the saved URL choice is reflected on re-entering the edit page
```
lauraghiorghisor-tw, line 23: And the URL choice is persisted when revisiting the edition?

features/edition-update-slug.feature
```diff
@@ -1,132 +1,51 @@
...
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      When I edit the publication "You will never guess"
+      And I change the title to "Ten facts that will shock you"
+      And I opt out of keeping the live slug
+      And I save the edition and go to the document summary
+      Then I can see the preview URL of the publication "Ten facts that will shock you" contains "ten-facts-that-will-shock-you"
+      And the saved URL choice is reflected on re-entering the edit page
+
+    Scenario: A force-published edition with the original URL still keeps it on the next rename
```
lauraghiorghisor-tw, line 35: "Custom" URL?

lauraghiorghisor-tw:
Closed in favour of https://github.com/alphagov/whitehall/pull/11445

PR #11436, opened 2026-05-08
 Validate each document tab form separately [WHIT-3214]

eYinka:
## What
Move the edition validation off of the edition model on to the configurable form object, and validate each form independently. Make sure that all forms are validated before the edition can be published.

In the scope of this implementation:

- Errors for the form are rendered to the user at the top of the page and next to the invalid form field, with a link from the top of the page to the invalid form
- Any edition tab’s form can be saved regardless of the validity of the other tabs
- All edition forms are validated before the “Publish” button is displayed to the user, and any errors are displayed to the user on the summary page. 
- Only a globally valid edition is updated in Publishing API
- Users can see a hint text letting them know draft previews may be out of date if an edition has invalid tabs

## Why

- Accelerates future development of configurable document types with a more appropriate abstraction
- Decouples form validation and attributes from the Edition model

### Edition summary page with invalid tab errors

Screenshot 2026-06-09 at 16 51 39

### Clicking through the error link
image

## Testing
Experiencing the full benefit of this work is tricky, because you will almost never have a saved multi-tab edition in bad state unless you manually manipulate the data. 
Currently, the only way to access dynamic tabs is by saving the document tab first, which means the draft is fully valid before any subsequent updates can be made. Therefore, it's not inherently clear that we are scoping validations per tab. 

For anyone that wants to give it a go locally, an easy way is to configure a document type (like topical event) with ` "send_change_history": true`, then: 
- Redraft a published topical event - this should surface the change history form on the document tab. 
- Without filling the change history form, try to save another tab (i.e Social media links) with invalid data. You should only see errors related to social media tab and nothing about the invalid change history should be displayed. Also you will be able to see a summary of the invalid tabs on the edition summary page.
- Alternatively, you can grab a draft edition and manipulate the data by doing the following:

```
# Grab the edition
edition = Edition.find(INSERT_EDITION_ID)

# Manipulate the block content by setting the body to nil and storing an invalid social media link
new_content = edition.translation[:block_content].merge("body" => nil, "social_media_links" => [{ "url" => nil, "social_media_service_name" => "Facebook" }])

# Persist the data
edition.translation.update_column(:block_content, new_content)
``` 

## Future Improvements:
- Before or after [migrating other tabs to be config-driven tab forms](https://gov-uk.atlassian.net/browse/WHIT-3241), we should be able to remove `validate_block_content` and `scoped_block_content` from [has_block_content.rb](https://github.com/alphagov/whitehall/blob/cf2bb1655575689f339dca40707e5c5633fae229/app/models/concerns/standard_edition/has_block_content.rb#L29-L49). 
TabForm will then be the single point for handling all `block_content` validation. For now, it's fine to keep the validation in both places because `edition.valid?` is called in several places that don't go through TabForm. Created [WHIT-3591](https://gov-uk.atlassian.net/browse/WHIT-3591). 
- Add `title` and `summary` to the documents schema of all config driven document types Created [WHIT-3590](https://gov-uk.atlassian.net/browse/WHIT-3590) 
- In the spirit of maintaining consistency, enforce adding a `label` to each form in document type configurations. Added to [WHIT-3241](https://gov-uk.atlassian.net/browse/WHIT-3241). 
- Maintain a single way of accessing the Documents tab. Currently, it is accessible via `/standard-editions/:id/edit` and `/standard-editions/:id/edit?current_tab=documents`. The latter is necessary for validating only the documents tab while the former will assume full edition validation. Added to [WHIT-3241](https://gov-uk.atlassian.net/browse/WHIT-3241) .

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-3214)

---

[WHIT-3591]: https://gov-uk.atlassian.net/browse/WHIT-3591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/admin/editions/show/preview_component.html.erb
+++ b/app/components/admin/editions/show/preview_component.html.erb
@@ -7,6 +7,11 @@
     margin_bottom: 3,
   } %>
 
+  <% if local_edition.is_a?(StandardEdition) && invalid_tab_forms&.any? %>
+    <%= render "govuk_publishing_components/components/inset_text", {
+      text: "This preview may be out of date because the edition is invalid.",
+    } %>
+  <% end %>
   <% if versioning_completed %>
     <p class="govuk-body">
       <%= preview_link(primary_locale_link_text, preview_url) %>
--- a/app/controllers/admin/standard_editions_controller.rb
+++ b/app/controllers/admin/standard_editions_controller.rb
@@ -37,9 +37,33 @@ class Admin::StandardEditionsController < Admin::EditionsController
     end
   end
 
-  def update
-    @edition.current_tab_context = @current_tab_context
+  def edit
     super
+    return if @current_tab_context.blank?
+
+    tab_form = StandardEdition::TabForm.new(@edition, @current_tab_context)
+    apply_tab_errors_to_edition(tab_form) unless tab_form.valid?
+  end
+
+  def update
+    @edition.assign_attributes(edition_params)
+
+    if @current_tab_context.present?
+      tab_form = StandardEdition::TabForm.new(@edition, @current_tab_context)
+
+      if tab_form.valid?
+        @edition.save_as(current_user, validate: false)
+        after_update_operations
+        redirect_to redirect_param(fallback: show_or_edit_path), saved_confirmation_notice
+      else
+        apply_tab_errors_to_edition(tab_form)
+        render_edition_update_failure
+      end
+    else
+      # non-tabbed forms in standard edition, or non-tab pages like translations
+      @edition.current_tab_context = nil
+      super
+    end
   end
 
   def features
@@ -65,6 +89,15 @@ class Admin::StandardEditionsController < Admin::EditionsController
     render :features
   end
 
+  def show
+    super
+    type_instance = @edition.type_instance
+    @invalid_tab_forms = type_instance.form_keys.filter_map do |tab_key|
+      tab_form = StandardEdition::TabForm.new(@edition, tab_key)
+      { tab_key:, label: type_instance.form(tab_key)["label"] || tab_key.humanize } unless tab_form.valid?
+    end
+  end
+
 private
 
   def edition_class
@@ -87,6 +120,11 @@ private
     else
       super
     end
+  end
+
+  def apply_tab_errors_to_edition(tab_form)
+    @edition.errors.clear
+    tab_form.errors.each { |e| @edition.errors.add(e.attribute, e.message) }
   end
 
   def render_not_found
--- a/app/controllers/admin/standard_editions_controller.rb
+++ b/app/controllers/admin/standard_editions_controller.rb
@@ -41,7 +41,7 @@ class Admin::StandardEditionsController < Admin::EditionsController
     super
     return if @current_tab_context.blank?
 
-    tab_form = StandardEdition::TabForm.new(@edition, @current_tab_context)
+    tab_form = StandardEdition::TabForm.new(@edition.reload, @current_tab_context)
     apply_tab_errors_to_edition(tab_form) unless tab_form.valid?
   end
 
--- a/app/helpers/admin/tabbed_nav_helper.rb
+++ b/app/helpers/admin/tabbed_nav_helper.rb
@@ -42,8 +42,9 @@ module Admin::TabbedNavHelper
   def config_driven_nav_items(edition, current_path, current_tab)
     base_url = tab_url_for_edition(edition)
     on_dynamic_tab = current_tab.present? || current_path == base_url
+    on_default_tab = current_tab == edition.default_tab || (current_tab.nil? && current_path == base_url)
 
-    [{ label: "Document", href: base_url, current: on_dynamic_tab && current_tab.nil? }].tap do |nav_items|
+    [{ label: "Document", href: base_url, current: on_default_tab }].tap do |nav_items|
       edition.type_instance.dynamic_tabs.each do |tab|
         nav_items << {
           label: tab["label"],
--- a/app/models/concerns/edition/workflow.rb
+++ b/app/models/concerns/edition/workflow.rb
@@ -100,8 +100,8 @@ module Edition::Workflow
     Edition::PRE_PUBLICATION_STATES.include?(state.to_s)
   end
 
-  def save_as(user)
-    if save
+  def save_as(user, options = {})
+    if save(**options)
       edition_authors.create!(user:)
       recent_edition_openings.where(editor_id: user).delete_all
     end
--- a/app/models/concerns/standard_edition/tab_form.rb
+++ b/app/models/concerns/standard_edition/tab_form.rb
@@ -13,6 +13,10 @@ class StandardEdition::TabForm
 
 private
 
+  def default_tab?
+    tab_key == edition.default_tab
+  end
+
   def form_config
     @form_config ||= edition.type_instance.form(tab_key).tap do |config|
       raise ArgumentError, "Unknown tab key '#{tab_key}'" unless config
@@ -62,11 +66,32 @@ private
 
     edition.current_tab_context = tab_key
     edition.valid?(validation_context)
-    edition.current_tab_context = nil
 
+    if default_tab?
+      import_non_block_content_edition_errors
+    else
+      import_scoped_edition_errors
+    end
+    edition.current_tab_context = nil
+  end
+
+  def import_scoped_edition_errors
     edition_attribute_keys.each do |attr|
       edition.errors.where(attr.to_sym).each { |error| errors.import(error, attribute: error.attribute.to_s) }
     end
-    edition.errors.clear # clear unscoped edition errors so only the imported tab-scoped errors reach the controller.
+  end
+
+  def import_non_block_content_edition_errors
+    edition.errors.each do |error|
+      next if block_content_error_attribute?(error.attribute.to_s)
+
+      errors.import(error, attribute: error.attribute.to_s)
+    end
+  end
+
+  def block_content_error_attribute?(attribute)
+    block_content_field_keys.any? do |field_key|
+      attribute == field_key
+    end
   end
 end
--- a/app/models/concerns/standard_edition/tab_form.rb
+++ b/app/models/concerns/standard_edition/tab_form.rb
@@ -1,0 +1,71 @@
+class StandardEdition::TabForm
+  include ActiveModel::Model
+
+  attr_reader :edition, :tab_key
+
+  validate :validate_block_content
+  validate :validate_edition_fields
+
+  def initialize(edition, tab_key)
+    @edition = edition
+    @tab_key = tab_key
+  end
+
+private
+
+  def form_config
+    @form_config ||= edition.type_instance.form(tab_key).tap do |config|
+      raise ArgumentError, "Unknown tab key '#{tab_key}'" unless config
+    end
+  end
+
+  # Fields whose data lives inside block_content (e.g. "body", "social_media_links")
+  def block_content_field_keys
+    (form_config["fields"] || {}).filter_map do |_key, field|
+      attr_path = Array(field["attribute_path"])
+      attr_path.last if attr_path.first == "block_content"
+    end
+  end
+
+  # Fields whose data lives directly on the edition (e.g. "lead_organisation_ids")
+  def edition_attribute_keys
+    keys = (form_config["fields"] || {}).filter_map do |_key, field|
+      attr_path = Array(field["attribute_path"])
+      attr_path.first unless attr_path.first == "block_content"
+    end
+
+    # TODO: Move these fields into document form configuration
+    # so that we don't have to inject them here
+    if tab_key == edition.default_tab
+      keys.concat(%w[title summary])
+    end
+
+    keys
+  end
+
+  def scoped_block_content
+    schema = edition.type_instance.schema_for_fields(block_content_field_keys)
+    StandardEdition::BlockContent.new(schema).tap do |bc|
+      bc.assign_attributes(edition[:block_content] || {})
+    end
+  end
+
+  def validate_block_content
+    block_content = scoped_block_content
+    return if block_content.valid?(validation_context)
+
+    block_content.errors.each { |error| errors.import(error, attribute: error.attribute.to_s) }
+  end
+
+  def validate_edition_fields
+    return unless edition_attribute_keys.any?
+
+    edition.current_tab_context = tab_key
+    edition.valid?(validation_context)
+    edition.current_tab_context = nil
+
+    edition_attribute_keys.each do |attr|
+      edition.errors.where(attr.to_sym).each { |error| errors.import(error, attribute: error.attribute.to_s) }
+    end
+  end
+end
--- a/app/models/concerns/standard_edition/tab_form.rb
+++ b/app/models/concerns/standard_edition/tab_form.rb
@@ -67,5 +67,6 @@ private
     edition_attribute_keys.each do |attr|
       edition.errors.where(attr.to_sym).each { |error| errors.import(error, attribute: error.attribute.to_s) }
     end
+    edition.errors.clear # clear unscoped edition errors so only the imported tab-scoped errors reach the controller.
   end
 end
--- a/app/models/configurable_document_type.rb
+++ b/app/models/configurable_document_type.rb
@@ -104,6 +104,10 @@ class ConfigurableDocumentType
     end
   end
 
+  def form_keys
+    @forms.keys
+  end
+
   def presenter(key)
     @presenters[key]
   end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -492,4 +492,8 @@ private
 
     first_version_with_state.try(:user)
   end
+
+  def current_tab_context_includes_field?(_attribute_name)
+    true
+  end
 end
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -443,6 +443,10 @@ class Edition < ApplicationRecord
     {}
   end
 
+  def invalid_tab_messages
+    []
+  end
+
 private
 
   def date_for_government
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -179,6 +179,16 @@ class StandardEdition < Edition
     options[:base]&.type_instance&.title_for_attribute(attribute.to_s) || super
   end
 
+  def invalid_tab_messages
+    type_instance.form_keys.filter_map do |tab_key|
+      tab_form = StandardEdition::TabForm.new(self, tab_key)
+      unless tab_form.valid?(:publish)
+        tab_label = type_instance.form(tab_key)["label"] || tab_key.humanize
+        "#{tab_label} tab is invalid"
+      end
+    end
+  end
+
 private
 
   def field_paths(&block)
--- a/app/models/standard_edition.rb
+++ b/app/models/standard_edition.rb
@@ -199,4 +199,15 @@ private
   def string_for_slug
     title if primary_locale.to_sym == translation.locale
   end
+
+  def current_tab_context_includes_field?(attribute_name)
+    return true if current_tab_context.blank? # assume we are checking against the entire edition
+
+    form = type_instance.form(current_tab_context)
+    return true if form.nil?
+
+    (form["fields"] || {}).any? do |_key, field|
+      Array(field["attribute_path"]).include?(attribute_name)
+    end
+  end
 end
--- a/app/services/edition_publisher.rb
+++ b/app/services/edition_publisher.rb
@@ -10,6 +10,7 @@ class EditionPublisher < EditionService
     reasons << "This edition is invalid: #{edition.errors.full_messages.to_sentence}" unless edition.valid?(:publish)
     reasons << "An edition that is #{edition.current_state} cannot be #{past_participle}" unless can_transition?
     reasons << "Scheduled editions cannot be published. This edition is scheduled for publication on #{edition.scheduled_publication}" if scheduled_for_publication?
+    reasons.concat(edition.invalid_tab_messages)
 
     @failure_reasons = reasons
   end
--- a/app/views/admin/editions/show/_main.html.erb
+++ b/app/views/admin/editions/show/_main.html.erb
@@ -34,8 +34,7 @@
   </section>
 
   <%= render partial: "admin/editions/show/main_notices", locals: { edition: @edition } %>
-
-  <%= render Admin::Editions::Show::PreviewComponent.new(edition: @edition) %>
+  <%= render Admin::Editions::Show::PreviewComponent.new(edition: @edition, invalid_tab_forms: @invalid_tab_forms) %>
 
   <% if @edition.change_note_required? %>
     <section class="app-view-summary__section">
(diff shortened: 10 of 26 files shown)


app/models/standard_edition.rb
```diff
@@ -115,15 +115,18 @@ def group
   end
 
   def organisation_association_enabled?
-    field_paths.include?(ConfigurableContentBlocks::Path.new("lead_organisation_ids"))
+    current_tab_context_includes_field?("lead_organisation_ids") &&
+      field_paths.include?(ConfigurableContentBlocks::Path.new("lead_organisation_ids"))
```
ChrisBAshton: I'm a bit concerned this is in the wrong place in the stack. We want to be able to ask "for this type of StandardEdition, are organisations enabled?". That's a business logic / model decision. Whether or not to enforce validation on that association is more of a controller-level one, I'd say.

Making the changes here risks introducing bugs in, say, the Edition Access Limited controller's `changed?` method, or risks us not calling the `build_default_organisation` method on the Edition Controller, or there may be other places too where we rely on this method call being a true reflection of what is/isn't enabled on a document type.

I think really this sort of branching should move to:

https://github.com/alphagov/whitehall/blob/5521aeecac194c4adbc7c83ffdce35f435ff189b/app/models/concerns/edition/organisations.rb#L16-L17

Something like:

```
    validate :at_least_one_lead_organisation, if: -> { organisation_association_enabled? && current_tab_context_includes_field?("lead_organisation_ids") }
```

app/models/concerns/standard_edition/tab_form.rb
```diff
@@ -0,0 +1,65 @@
+class StandardEdition::TabForm
+  include ActiveModel::Model
+
+  attr_reader :edition, :tab_key
+
+  validate :validate_block_content
+  validate :validate_edition_fields, if: -> { edition_attribute_keys.any? }
```
ChrisBAshton: I'd move the conditional inside `validate_edition_fields` as an early return, instead of here.

app/models/concerns/standard_edition/tab_form.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  # Fields whose data lives directly on the edition (e.g. "lead_organisation_ids")
+  def edition_attribute_keys
+    keys = (form_config["fields"] || {}).filter_map do |_key, field|
+      attr_path = Array(field["attribute_path"])
+      attr_path.first unless attr_path.first == "block_content"
+    end
+
+    tab_key == edition.default_tab ? keys + ADDITIONAL_DEFAULT_TAB_FIELDS : keys # Need to move these fields into document form configuration
```
ChrisBAshton: I think we can drop the const as it's only used here.
Can also make it a bit more optimised for deletion as below:

```suggestion
    # TODO: Move these fields into document form configuration
    # so that we don't have to inject them here
    if tab_key == edition.default_tab
      keys << %w[title summary]
    end
    
    keys
```

app/models/concerns/edition/workflow.rb
```diff
@@ -100,8 +100,8 @@ def pre_publication?
     Edition::PRE_PUBLICATION_STATES.include?(state.to_s)
   end
 
-  def save_as(user)
-    if save
+  def save_as(user, options = {})
+    if save(**options)
```
ChrisBAshton, line 104: I'd expect a corresponding unit test here

app/controllers/admin/standard_editions_controller.rb
```diff
@@ -38,8 +38,26 @@ def apply_change_type
   end
 
   def update
-    @edition.current_tab_context = @current_tab_context
-    super
```
ChrisBAshton, line 42: We're moving away from calling `super` on every update, to _only_ calling `super` when saving a non-tabbed form. Have we explored the downsides of that?

https://github.com/alphagov/whitehall/blob/52e595369be78c248fc8a8dcf0ab41ccb8b2140d/app/controllers/admin/editions_controller.rb#L124-L148

Calling `super` means calling `updater.perform!`, rather than `@edition.save_as` - does the DraftEditionUpdater service do something important we'd lose otherwise?

Also currently `super` means a fresh LinkCheckReport is generated, which we'd currently be losing by not calling that.

ChrisBAshton: (it may be that we want to call `super` when updating the Document tab, but otherwise call our own logic if saving any other tab 🤔 )

app/controllers/admin/standard_editions_controller.rb
```diff
@@ -38,8 +38,26 @@ def apply_change_type
...
+
+      if tab_form.valid?
+        @edition.save_as(current_user, validate: false)
+        redirect_to redirect_param(fallback: show_or_edit_path), saved_confirmation_notice
+      else
+        apply_tab_errors_to_edition(tab_form)
+        build_edition_dependencies
+        fetch_version_and_remark_trails
+        construct_similar_slug_warning_error
+        render :edit
```
ChrisBAshton: I did wonder if it's worth pulling out where this is called in EditionsController, into its own `protected` method, so that we can refer to that same method here rather than duplicate it?

app/models/configurable_document_type.rb
```diff
@@ -103,6 +103,10 @@ def form(key = nil)
     end
   end
 
+  def form_keys
```
ChrisBAshton, line 107: Add corresponding unit test please

app/components/admin/editions/show/preview_component.html.erb
```diff
@@ -11,6 +11,11 @@
     <p class="govuk-body">
       <%= preview_link(primary_locale_link_text, local_edition.public_url(draft: true)) %>
     </p>
+    <% if local_edition.is_a?(StandardEdition) && invalid_tab_forms&.any? %>
+      <p class="govuk-hint">
+        This preview may be out of date. Kindly fix the errors in the invalid tabs before the preview reflects your latest changes.
```
ChrisBAshton: I'd move this above the preview link (above line 11) rather than having it sit between the main and the translated preview links.

I'd also get Sally to suggest a line of content. 🙏

ChrisBAshton: And a view test to ensure this message appears when a tab is invalid. (So probably worth pulling this whole thing into its own commit)

app/views/admin/editions/show/_main.html.erb
```diff
@@ -34,8 +34,7 @@
   </section>
 
   <%= render partial: "admin/editions/show/main_notices", locals: { edition: @edition } %>
-
-  <%= render Admin::Editions::Show::PreviewComponent.new(edition: @edition) %>
+  <%= render Admin::Editions::Show::PreviewComponent.new(edition: @edition, invalid_tab_forms: @invalid_tab_forms) %>
```
ChrisBAshton, line 37: Hmmm this extra param is being passed, but doesn't appear to be being used anywhere?

app/helpers/admin/tabbed_nav_helper.rb
```diff
@@ -42,8 +42,9 @@ def legacy_document_nav_items(edition, current_path)
   def config_driven_nav_items(edition, current_path, current_tab)
     base_url = tab_url_for_edition(edition)
     on_dynamic_tab = current_tab.present? || current_path == base_url
+    on_default_tab = current_tab.nil? || current_tab == edition.default_tab
 
-    [{ label: "Document", href: base_url, current: on_dynamic_tab && current_tab.nil? }].tap do |nav_items|
+    [{ label: "Document", href: base_url, current: on_dynamic_tab && on_default_tab }].tap do |nav_items|
```
ChrisBAshton:
```suggestion
    [{ label: "Document", href: base_url, current: on_default_tab }].tap do |nav_items|
```

app/services/draft_edition_updater.rb
```diff
@@ -1,7 +1,7 @@
 class DraftEditionUpdater < EditionService
   def perform!
     if can_perform?
-      update_publishing_api!
+      update_publishing_api! if can_push_draft?
```
ChrisBAshton: Is this not already caught by the `elsif !edition.valid?` check in `failure_reason`?

If not, `failure_reason` feels like the right place to capture a StandardEdition specific check - rather than introducing a change here.

But I'm also confused because in my earlier comment, it looks as though we never call the `DraftEditionUpdater` service with the current changes to the StandardEditionController 🤔 
And actually maybe that's a better approach. By the time we get to the DraftEditionUpdater, we perhaps should already be confident that every tab's individual validation issues have been fixed.

app/services/draft_edition_updater.rb
```diff
@@ -30,4 +30,12 @@ def should_check_current_user_will_retain_access?
   def access_limit_excludes_current_user?
     edition.organisation_association_enabled? && edition.edition_organisations.map(&:organisation_id).exclude?(@options[:current_user].organisation.id)
   end
+
+  def can_push_draft?
+    return true unless edition.is_a?(StandardEdition)
+
+    edition.type_instance.form_keys.all? do |tab_key|
```
ChrisBAshton: If we _do_ need to expose the tab form validation at a higher level than the StandardEditionController, can we at least wrap it in a simpler helper method so we're not having to reference form keys etc? Something like:


```suggestion
    edition.valid?(:on_all_tabs)
```

app/services/edition_service.rb
```diff
@@ -67,4 +67,14 @@ def prepare_edition
   def fire_transition!
     edition.public_send("#{verb}!")
   end
+
+  def invalid_tab_reasons
+    edition.type_instance.form_keys.filter_map do |tab_key|
+      tab_form = StandardEdition::TabForm.new(edition, tab_key)
+      unless tab_form.valid?(:publish)
```
ChrisBAshton: Yeah I'm not keen that we've overridden the meaning of `edition.valid?` or `edition.valid?(:publish)` and now have to update the rest of Whitehall accordingly. I think one of the AC's for this work, in hindsight, should be that calling `edition.valid?(:publish)` checks for validation errors across all tabs. That way we don't have to update any of these higher level workflows.

And then at the lower StandardEdition level, we just stop calling `edition.valid?(:publish)` and start being tab-aware instead, e.g. 

```ruby
edition.validation_tab = tab
edition.valid?(:tabbed)
```

ChrisBAshton (commented):
Well done for working through such a tricky problem ⭐ 

Whilst what you have works, it does spread the tab-awareness throughout some quite key Whitehall publishing workflows. I'd hope that given we have a dedicated StandardEditionsController, we could be a bit more precise about keeping tab awareness 'contained' and not having to reference low-level StandardEdition concepts from high level Edition Services etc.

It may be easier said than done , but I think it's worth spending a few days considering how self-contained we can make the tab validation, and also how self-contained we can make the 'associations' validation (rather than changing the meaning of the associations helper methods in the first commit). 🙏  Give me a shout if you'd like to talk any of this through.

app/views/admin/editions/show/_main.html.erb
eYinka: It's actually being used  to conditionally show the "preview may be out of date..." bit on the preview component.

See: https://github.com/alphagov/whitehall/blob/13c251e75d525c7d57e1fc7e0ad27665bf252973/app/components/admin/editions/show/preview_component.html.erb#L14-L18

app/controllers/admin/standard_editions_controller.rb
eYinka: Well, I don't think calling `super` on the Document tab would be the ideal choice, as this would mean running all validations and could lead to a save failure if the Document tab itself is valid but another tab is invalid.

I've do a bit of re-organisation to that flow and should be cleaner now.

Essentially, we now have something like this:

| Scenario | Validation | Save | Post-save |
|---|---|---|---|
| Form tab save (any tab, including Document) | `TabForm#valid?` on current tab fields only | `save_as(current_user, validate: false)` | `after_update_operations` -->  `updater.perform!` + link checker |
| Non-tabbed save (e.g. non-standard-editions, translations) | Full model validation via `updater.can_perform?` + `save_as` | `save_as(current_user)` | `after_update_operations` --> `updater.perform!` + link checker |


The `after_update_operations` allows us to share a single source for post-save tasks across board.

app/services/draft_edition_updater.rb
eYinka: Good spot. Initially this bit was going to be skipped, because `DraftEditionUpdater` service for for tabbed edition was never called. With the refactor I've done now, it will now be called so I'll safely get rid of this check and the workflow should continue as-is.

app/services/draft_edition_updater.rb
eYinka: Gotten rid of that bit of code now

eYinka: But good opportunity to take another look at simplifying the check outside Standard Editions Controller.

app/models/edition.rb
```diff
@@ -492,4 +492,8 @@ def author_of_latest_state_change_to(state)
 
     first_version_with_state.try(:user)
   end
+
+  def current_tab_context_includes_field?(_attribute_name)
+    true
```
ChrisBAshton, line 501: Ah nice, I was wondering how non-StandardEdition would cope with the changes. Makes sense 👍

app/models/standard_edition.rb
```diff
@@ -191,4 +191,15 @@ def required_field_paths
   def string_for_slug
     title if primary_locale.to_sym == translation.locale
   end
+
+  def current_tab_context_includes_field?(attribute_name)
```
ChrisBAshton, line 213: Missing unit tests for this.

app/models/concerns/standard_edition/tab_form.rb
ChrisBAshton, line 1: This second commit is a joy 🎉

app/controllers/admin/standard_editions_controller.rb
```diff
@@ -89,6 +105,11 @@ def show_or_edit_path
     end
   end
 
+  def apply_tab_errors_to_edition(tab_form)
```
ChrisBAshton, line 125: Clever

test/unit/app/models/configurable_document_type_test.rb
```diff
@@ -306,27 +306,30 @@ class ConfigurableDocumentTypeTest < ActiveSupport::TestCase
     assert_not result["validations"].key?("length")
   end
 
-  test "#schema_for_fields filters the attributes array within a shared validation" do
```
ChrisBAshton, line 309: Uhhh shouldn't we keep this test in? `schema_for_fields` still exists at this point 🤔 I think you wanted to duplicate this test and tweak, rather than overwrite it...?

app/views/admin/editions/show/_sidebar_notices.html.erb
```diff
@@ -7,6 +7,17 @@
 
 <div class="app-view-summary__sidebar-notices">
   <% unless @edition.valid?(:publish) %>
+    <% invalid_items = if @edition.is_a?(StandardEdition) && @invalid_tab_forms&.any?
+      @invalid_tab_forms.map do |tab|
+        link_to(
+          "#{tab[:label]} tab is invalid",
```
ChrisBAshton, line 13: Just to check, does this surface errors for the default Document tab too?
So where we used to surface all the individual errors for the Document tab, we're now just saying there's something wrong with the Document tab and linking them through to the Edit screen?

I think that's the right approach - but also how do they know what needs fixing when they get to that tab? Don't they only see the individual validation error messages when trying to save the tab?

app/views/admin/editions/show/_main.html.erb
ChrisBAshton, line 37: This change isn't necessary (yet - I'm in the `Show invalid tab errors on edition summary page` commit)

app/services/edition_publisher.rb
```diff
@@ -10,6 +10,7 @@ def failure_reasons
     reasons << "This edition is invalid: #{edition.errors.full_messages.to_sentence}" unless edition.valid?(:publish)
     reasons << "An edition that is #{edition.current_state} cannot be #{past_participle}" unless can_transition?
     reasons << "Scheduled editions cannot be published. This edition is scheduled for publication on #{edition.scheduled_publication}" if scheduled_for_publication?
+    reasons.concat(invalid_tab_reasons) if edition.is_a?(StandardEdition)
```
ChrisBAshton: Smart!

ChrisBAshton (commented):
Well, this is looking swell! 👏 Really excellent work, Ola.

My main question is around the new workflow for **how** publishers should know which individual errors to fix (once they've clicked on the link in the sidebar). A screenshot of one of the invalid tabs (on page load, rather than post-save) would be useful.

test/unit/app/models/configurable_document_type_test.rb
eYinka: Ooh! Yes I was duplicating the test and missed the 'paste' somehow. Fixed now 👍

app/views/admin/editions/show/_sidebar_notices.html.erb
eYinka: Have to think through that UX. But to answer your question

> Don't they only see the individual validation error messages when trying to save the tab?

Yes, this is the only way they see the individual errors. Do you reckon we find a way to flash the tab errors on initial load? 🤔

app/views/admin/editions/show/_main.html.erb
eYinka: Yup. Good spot. Didn't carry it over during the previous commit split. Fixed now!

app/views/admin/editions/show/_sidebar_notices.html.erb
eYinka: Got that to work and I've updated the PR description with a screenshot.

ChrisBAshton: Nice - I see the [screenshot](https://private-user-images.githubusercontent.com/38944994/605642927-276506a4-9b39-412a-917d-8c56321e5f34.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODEwODI3OTUsIm5iZiI6MTc4MTA4MjQ5NSwicGF0aCI6Ii8zODk0NDk5NC82MDU2NDI5MjctMjc2NTA2YTQtOWIzOS00MTJhLTkxN2QtOGM1NjMyMWU1ZjM0LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MTAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjEwVDA5MDgxNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTNjNzg3Yjk4NTcwMTI5MWZlNjJhOTcxZWM3MDUwMWU2ZGIwMWNiMGJjODkyNjA2NWUwMzQxYWE4YjEyNDIwMDUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRnBuZyJ9.SIX1TUsXNNjFbweJ4UxDjMFJzXBi8Nz5kGZ3I13LNIc).

Does this also show on document creation? (We wouldn't want it to).
If it's only on document edit, that seems good.

eYinka: No it doesn't. It's limited to the edit route only, and if current tab context is present.

app/services/edition_publisher.rb
eYinka: Actually on a second thought, I'll move this bit to the model level so that the edition service no longer knows about the tab error computing for Standard Editions. I think you touched it a little in your previous comments.

In the new way, I'm thinking of having something like:

```
# EditionPublisher / EditionScheduler
reasons.concat(edition.invalid_tab_reasons) 
```

So that the `invalid_tab_reasons` will be computed on the StandardEdition model - and with a default `[]` response from the base Edition model. 

Looks cleaner to me that way!

test/unit/app/models/standard_edition_test.rb
```diff
@@ -996,4 +996,94 @@ class StandardEditionTest < ActiveSupport::TestCase
...
+    edition = create(:draft_standard_edition, :with_organisations,
+                     configurable_document_type: "test_type",
+                     block_content: { body: "Some content", sidebar: "Some sidebar content" })
+
+    assert_equal [], edition.invalid_tab_messages
+  end
+
+  test "#invalid_tab_messages returns a reason for each invalid tab" do
+    type = build_configurable_document_type(
+      "test_type", {
```
ChrisBAshton: The repetition of these big setup blocks is not ideal - makes it harder to see what conditions you're actually trying to test. I'd be tempted to put both these unit tests in one "context" block with a shared "setup" block that defines the `@type` - then let the unit tests themselves more clearly show what they're testing (ie that you're just calling `update_column` on the latter, to break validation)

Might be worth a step back and doing this across several of the new tests in the PR. It can be an extra refactor commit rather than rebased into each original commit if that helps 👍

ChrisBAshton (approved):
LGTM 👍 One non-blocking suggestion below.

app/controllers/admin/standard_editions_controller.rb
```diff
@@ -41,7 +41,7 @@ def edit
     super
     return if @current_tab_context.blank?
 
-    tab_form = StandardEdition::TabForm.new(@edition, @current_tab_context)
+    tab_form = StandardEdition::TabForm.new(raw_edition_for_edit_validation, @current_tab_context)
```
ChrisBAshton: Hmmm can't we just call `@edition.reload`?

If not, I'd probably rename method to something like `load_edition_from_database(@edition)`. And also we probably need some sort of unit test?

ChrisBAshton (approved):
LGTM. One suggestion below.

Also feels like we ought to have at least one Cucumber feature test covering the tabbed validation behaviour 🤔 But happy to approve as-is. (Probably worth waiting until Monday for merge, in case of unexpected side-effects).

app/controllers/admin/standard_editions_controller.rb
eYinka: I'm being careful there, as `@edition.reload` could introduce un-intended side effects and we might lose the usage of `build_edition_dependencies` before the page loads. Unless `@edition.reload` doesn't re-mutate what gets sent to the view (?)  🤔

ChrisBAshton: Happy to leave it to you. But the code certainly reads as though the intention is to initialize a clean edition instance from the database, which is exactly what `.reload` does :D I'd give it a try and see if there are indeed any side effects or if it works as we'd hope.

eYinka: I've tested that and it seems to do the work. Thanks!

eYinka: Actually, I've just seen that the review reminder fields disappear when the `current_tab` param is visible. That's a side effect, and it fits my guess above that `@edition.reload` re-mutates what gets sent to the view template.

eYinka: I've raised a fix-up https://github.com/alphagov/whitehall/pull/11560

PR #11440, opened 2026-05-12
Fix definitive article [WHIT-3396]

davidsauntson:
## what

Changes how organisation names are displayed in:
- corporate information page titles and links (eg `Working at Environment Agency` -> `Working at the Environment Agency`
- organisation / parent organisation relationship text

## why

Corporate information page titles and text on the organisation home page was missing the definitive article.  Raised as a support ticket https://govuk.zendesk.com/agent/tickets/6543938

Before (current prod):
image
image

After: (tested on integration)

image
image

---

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -1,7 +1,7 @@
 module OrganisationHelper
   include ApplicationHelper
 
-  def organisation_display_name(organisation)
+  def organisation_relationship_display_name(organisation)
     if organisation.acronym.present?
       tag.abbr(organisation.acronym, title: organisation.name)
     else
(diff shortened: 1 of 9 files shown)


app/helpers/organisation_helper.rb
```diff
@@ -1,9 +1,20 @@
...
   include ApplicationHelper
 
+  PATTERNS_EXEMPT_FROM_DEFINITIVE_ARTICLE = [
+    /civil service resourcing/,
+    /^hm/,
+    /ordnance survey/,
+    /homes england/,
+    /british wool/,
+    /building law and hygiene/,
+  ].freeze
```
GDSNewt, line 11: Just curious where this list has come from, I assume content guidance? I wonder if we need some documentation changes. 

Something like this? https://docs.publishing.service.gov.uk/manual/howto-add-organisation-brand-colour-logo.html

Maybe we stick a pin in that and see if this list actually gets changed over the next quarter. Could be that it's a rare event.

GDSNewt (approved):
LGTM, just had that one question about the exception list.

app/helpers/organisation_helper.rb
davidsauntson: The top four were already in the code and the last two were ones I noticed during testing.  It isn't an exhaustive list to be fair, but I'm not really sure where to get one from.  

Do you think it needs to be checked by someone / complete before it's merged?  This PR would fix the issue for the Environment Agency but it might cause an issue for an org that is missing from the exemptions list.

PR #11442, opened 2026-05-12
Allow Welsh-language mainstream URLs [WHIT-3436]

GDSNewt:
Welsh-only GOV.UK pages (e.g. /talu-treth-twe) were failing validation with "Url must reference a GOV.UK page" when added to document collection groups. The root cause was that get_content in the Publishing API defaults to locale: en, so Welsh-only content — which only exists in locale: cy — returned a 404.

This PR replaces get_content with the Content Store for fetching content items. The Content Store serves content regardless of locale, so Welsh URLs now resolve correctly.

The guide sub-page check (which validates that a two-segment path like /guidance/subpage belongs to a guide) continues to work because the Content Store returns the parent guide's content item for sub-page paths, making [document_type](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) available directly from that response.

**BEFORE**

https://github.com/user-attachments/assets/6afa595b-7db7-4b94-ab02-4aa3d1531de4

**AFTER**

https://github.com/user-attachments/assets/6d77a1e9-56e7-4fb1-9d7d-a3dca9d3c5a0

[Jira](https://gov-uk.atlassian.net/browse/WHIT-3436)

### DIFF
--- a/app/models/document_collection_non_whitehall_link/govuk_url.rb
+++ b/app/models/document_collection_non_whitehall_link/govuk_url.rb
@@ -31,24 +31,24 @@ class DocumentCollectionNonWhitehallLink::GovukUrl
   end
 
   def content_item
-    @content_item ||= Services.publishing_api.get_content(content_id).to_h
+    @content_item ||= content_item_from_content_store
   end
 
   def content_id
-    @content_id ||= Services.publishing_api.lookup_content_id(base_path: parsed_url.path, with_drafts: true)
+    @content_id ||= content_item["content_id"]
+  end
 
-    if @content_id.blank?
-      toplevel_path_segment = parsed_url.path.split("/").second
-      @content_id = Services.publishing_api.lookup_content_id(base_path: "/#{toplevel_path_segment}", with_drafts: true)
-      if @content_id.blank?
-        raise GdsApi::HTTPNotFound, 404
-      else
-        unless content_item["document_type"] == "guide"
-          raise GdsApi::HTTPNotFound, 404
-        end
-      end
+  def content_item_from_content_store
+    path = parsed_url.path
+
+    item = Services.content_store.content_item(path).to_h
+
+    if item["base_path"] != path && item["document_type"] != "guide"
+      raise GdsApi::HTTPNotFound, 404
     end
 
-    @content_id
+    item
+  rescue GdsApi::ContentStore::ItemNotFound
+    raise GdsApi::HTTPNotFound, 404
   end
 end
--- a/test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
+++ b/test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
@@ -3,16 +3,15 @@ require "test_helper"
 class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
   setup do
     @content_id = SecureRandom.uuid
-    stub_publishing_api_has_lookups("/test" => @content_id)
-    stub_publishing_api_has_item(
-      content_id: @content_id,
-      title: "Test",
-      base_path: "/test",
-      publishing_app: "content-publisher",
+    Services.content_store.stubs(:content_item).with("/test").returns(
+      "content_id" => @content_id,
+      "title" => "Test",
+      "base_path" => "/test",
+      "publishing_app" => "content-publisher",
     )
   end
 
-  test "should be valid without a GOV.UK url that Publishing API knows" do
+  test "should be valid when content store has the path" do
     url = DocumentCollectionNonWhitehallLink::GovukUrl.new(
       url: "https://www.gov.uk/test",
       document_collection_group: build(:document_collection_group),
@@ -32,12 +31,13 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
 
   test "should be valid when a mainstream guide sub-page url is used" do
     content_id = SecureRandom.uuid
-    stub_publishing_api_has_lookups("/foo" => content_id)
-    stub_publishing_api_has_item(content_id:,
-                                 title: "Foo Bar",
-                                 base_path: "/foo",
-                                 document_type: "guide",
-                                 publishing_app: "content-publisher")
+    Services.content_store.stubs(:content_item).with("/foo/subpage").returns(
+      "content_id" => content_id,
+      "title" => "Foo Bar",
+      "base_path" => "/foo",
+      "publishing_app" => "content-publisher",
+      "document_type" => "guide",
+    )
 
     url = DocumentCollectionNonWhitehallLink::GovukUrl.new(
       url: "https://www.gov.uk/foo/subpage",
@@ -85,18 +85,21 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
     assert url.errors.full_messages.include?("Url must be a valid GOV.UK URL")
   end
 
-  test "should be invalid when a GOV.UK URL that isn't in the Publishing API" do
+  test "should be valid for a Welsh-only GOV.UK page (locale: cy)" do
+    welsh_content_id = SecureRandom.uuid
+    content_store_response = { "content_id" => welsh_content_id, "title" => "Talu TWE y cyflogwr", "base_path" => "/talu-treth-twe", "publishing_app" => "publisher", "locale" => "cy" }
+    Services.content_store.stubs(:content_item).with("/talu-treth-twe").returns(content_store_response)
+
     url = DocumentCollectionNonWhitehallLink::GovukUrl.new(
-      url: "https://www.gov.uk/different-path",
+      url: "https://www.gov.uk/talu-treth-twe",
       document_collection_group: build(:document_collection_group),
     )
 
-    assert_not url.valid?
-    assert url.errors.full_messages.include?("Url must reference a GOV.UK page")
+    assert url.valid?
   end
 
-  test "should be invalid when Publishing API returns a 404" do
-    stub_any_publishing_api_call_to_return_not_found
+  test "should be invalid when content store returns a 404" do
+    Services.content_store.stubs(:content_item).raises(GdsApi::ContentStore::ItemNotFound.new(404))
 
     url = DocumentCollectionNonWhitehallLink::GovukUrl.new(
       url: "https://www.gov.uk/test",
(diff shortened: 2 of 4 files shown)


app/models/document_collection_non_whitehall_link/govuk_url.rb
jamiestamp: Just wondering - where we find a content ID VIA `content_id_from_content_store`, would we still need to include the check on `content_item["document_type"] == "guide"`?

GDSNewt: As I understand it the document_type == "guide" check only applies when a content ID is found via the top-level path segment (e.g. resolving /some-guide/some-part → /some-guide). In that case we’re looking up a parent item, so we need to validate that it’s a guide.

When both Publishing API lookups fail and we fall back to the content store, the URL maps directly to its own content item — there’s no parent relationship involved, so the guide check isn’t applicable.

The content store will still return a 404 if the item doesn’t exist, so invalid URLs are rejected either way.

jamiestamp (approved):
LGTM

app/models/document_collection_non_whitehall_link/govuk_url.rb
```diff
@@ -31,7 +31,12 @@ def parsed_url
   end
 
   def content_item
-    @content_item ||= Services.publishing_api.get_content(content_id).to_h
+    content_id
+    @content_item ||= begin
+      Services.publishing_api.get_content(@content_id).to_h
+    rescue GdsApi::HTTPNotFound
+      content_item_from_content_store
```
ChrisBAshton: Rather than falling back to the content store in the event of a 404, is there a disadvantage to calling content store in the first place and dropping the Publishing API call?

GDSNewt: Yeah think that makes sense, I'll give it a try on integration. I think I left it in there as the content Store only serves live content - but that is absolutely fine for doc collections! 🤦

app/models/document_collection_non_whitehall_link/govuk_url.rb
```diff
@@ -31,7 +31,8 @@ def parsed_url
   end
 
   def content_item
-    @content_item ||= Services.publishing_api.get_content(content_id).to_h
+    content_id
+    @content_item ||= content_item_from_content_store
```
ChrisBAshton: This kind of 'side effect' method (`content_id`) isn't very readable. Probably best being more open that one is dependent on the other:

```suggestion
    @content_item ||= content_item_from_content_store(content_id)
```

app/models/document_collection_non_whitehall_link/govuk_url.rb
ChrisBAshton: I thought we agreed to drop this call and just call Content Store right off the bat? It would simplify things a lot.

ChrisBAshton (commented):
Looks broadly fine - document collections need a bit more love in general anyway - but we do have an opportunity to simplify with Content Store calls and dropping Publishing API calls 🤞

TonyGDS:
I assume we're willing to live with any lag between publishing-api pushing the latest content to the content store? Worst that would happen is the user would see the same error as now.

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -125,7 +164,7 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
   end
 
   test "should be invalid when Publishing API is down" do
-    stub_publishing_api_isnt_available
+    Services.content_store.stubs(:content_item).raises(GdsApi::HTTPIntermittentServerError.new(503))
```
ChrisBAshton, line 132: Test name needs updating

.gitignore
```diff
@@ -33,3 +33,4 @@
 /app/assets/builds/*
 !/app/assets/builds/.keep
 
+console_history.txt
```
ChrisBAshton: This appears to have slipped in?

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -110,11 +147,13 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
...
     stub_publishing_api_has_lookups("/foo" => content_id)
-    stub_publishing_api_has_item(content_id:,
-                                 title: "Foo Bar",
-                                 base_path: "/foo",
-                                 document_type: "other",
-                                 publishing_app: "content-publisher")
+    Services.content_store.stubs(:content_item).with("/foo/subpage").returns(
+      "content_id" => content_id,
+      "title" => "Foo Bar",
+      "base_path" => "/foo",
```
ChrisBAshton, line 118: Should this not be `/foo/subpage`?

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -91,12 +99,41 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
...
+      url: "https://www.gov.uk/talu-treth-twe",
+      document_collection_group: build(:document_collection_group),
+    )
+
+    assert url.valid?
+  end
+
   test "should be invalid when Publishing API returns a 404" do
-    stub_any_publishing_api_call_to_return_not_found
+    Services.content_store.stubs(:content_item).raises(GdsApi::ContentStore::ItemNotFound.new(404))
```
ChrisBAshton, line 102: Test name needs updating

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -10,6 +10,12 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
...
     )
+    Services.content_store.stubs(:content_item).with("/test").returns(
+      "content_id" => @content_id,
+      "title" => "Test",
+      "base_path" => "/test",
+      "publishing_app" => "content-publisher",
+    )
   end
 
   test "should be valid without a GOV.UK url that Publishing API knows" do
```
ChrisBAshton: Test name needs updating

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
ChrisBAshton: Do these Publishing API stubs need to exist any more?

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -91,12 +99,41 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
       document_collection_group: build(:document_collection_group),
     )
 
+    Services.content_store.stubs(:content_item).raises(GdsApi::ContentStore::ItemNotFound.new(404))
+
     assert_not url.valid?
     assert url.errors.full_messages.include?("Url must reference a GOV.UK page")
   end
 
+  test "should be valid when a Welsh-language GOV.UK URL is used that is not in the Publishing API path reservations" do
```
ChrisBAshton: test name needs updating

ChrisBAshton: also I don't really understand the relevance of the second half of the test name 🤔

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -91,12 +99,41 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
...
+
+    url = DocumentCollectionNonWhitehallLink::GovukUrl.new(
+      url: "https://www.gov.uk/talu-cosb-hunanasesiad",
+      document_collection_group: build(:document_collection_group),
+    )
+
+    assert url.valid?
+  end
+
+  test "should be valid when a Welsh-language GOV.UK URL is in the Publishing API but only has a Welsh locale" do
```
ChrisBAshton: I'm not clear what this test is trying to exercise differently to the one above it?

ChrisBAshton (commented):
Hi Alex - the code broadly looks good, just some tidying up to do before merge. Thanks

.gitignore
GDSNewt: Oops thanks Chris, this is from when I was trying to diagnose the permissions error stuff for access limiting. Wanted to see the commands I ran a few days later 😬

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
GDSNewt: It's intentional as is; we stub "/foo/subpage" but return an item with "base_path" => "/foo" to simulate a base_path mismatch, which should be invalid unless the document type is "guide".

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -85,18 +85,34 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
...
+      url: "https://www.gov.uk/talu-treth-twe",
+      document_collection_group: build(:document_collection_group),
+    )
+
+    assert url.valid?
   end
 
-  test "should be invalid when Publishing API returns a 404" do
-    stub_any_publishing_api_call_to_return_not_found
+  test "should be invalid when content lookup returns a 404" do
```
ChrisBAshton:
```suggestion
  test "should be invalid when content store lookup returns a 404" do
```

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -124,8 +141,8 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
     assert_not url.valid?
   end
 
-  test "should be invalid when Publishing API is down" do
-    stub_publishing_api_isnt_available
+  test "should be invalid when content lookup is unavailable" do
```
ChrisBAshton:
```suggestion
  test "should be invalid when Content Store is unavailable" do
```

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
```diff
@@ -85,18 +85,34 @@ class DocumentCollectionNonWhitehallLink::GovukUrlTest < ActiveSupport::TestCase
     assert url.errors.full_messages.include?("Url must be a valid GOV.UK URL")
   end
 
-  test "should be invalid when a GOV.UK URL that isn't in the Publishing API" do
+  test "should be valid for a Welsh-language GOV.UK page (no locale set)" do
+    welsh_content_id = SecureRandom.uuid
+    content_store_response = { "content_id" => welsh_content_id, "title" => "Talu cosb hunanasesiad", "base_path" => "/talu-cosb-hunanasesiad", "publishing_app" => "publisher" }
```
ChrisBAshton: I'm still not clear what the purpose of this test is vs the one below it?
For one thing, the live content item you're referencing here _does_ have a locale set (`"locale": "cy"`) so what real world scenario are we trying to account for here?

Is it that we need a test for a content item that has an English `/foo` route as well as a Welsh `/foo.cy` kind of route?

ChrisBAshton (approved):
LGTM in principle, subject to a thorough test on integration 👍 
One query over one of the tests below.

test/unit/app/models/document_collection_non_whitehall_link/govuk_url_test.rb
GDSNewt: I think it was more relevant in an earlier iteration of this PR, but yeah is redundant now - at least for testing real world scenarios.  I'll remove it as it doesn't really add too much value. 

Devils advocate for leaving in: it communicates that the validator is not reliant on locale, which I suppose might be interesting to note...

PR #11445, opened 2026-05-13
[WHIT-3323] Keep current url on rename

lauraghiorghisor-tw:
## What and why

Previously, when the user changed the title of an edition in Whitehall, they were presented with a checkbox allowing them to preserve the document’s current live URL. If they did not check the box, the URL was updated to match the new title of the edition. Several departments have requested that we only update the edition's URL if the publisher specifically selects that behaviour.

The previous checkbox has been replaced by radio buttons, to better express what options the user has:
- keeping live url
- using the new title-based url

Keeping the live url is always our pre-checked option, unless the user has actively selected the title-based option. 
The copy has been updated to reflect the new functionality. Feature tests now provide extensive coverage.

## UI changes

| Before | After |
| --- | --- |
| image | image |

Related PRs:
https://github.com/alphagov/whitehall/pull/11371
https://github.com/alphagov/whitehall/pull/11424
https://github.com/alphagov/whitehall/pull/11359

### DIFF
--- a/app/assets/javascripts/admin/views/edition-form.js
+++ b/app/assets/javascripts/admin/views/edition-form.js
@@ -160,25 +160,25 @@
   EditionForm.prototype.setupTitleInputEventListener = function () {
     const form = this.module
     const titleInput = form.querySelector('#edition_title')
-    const checkboxContainer = form.querySelector('.js-keep-slug-form-group')
+    const urlRadioGroup = form.querySelector('.js-url-radio-group')
 
-    if (!titleInput || !checkboxContainer) {
+    if (!titleInput || !urlRadioGroup) {
       return
     }
 
     const originalTitle = titleInput.value
 
-    const setCheckboxVisibility = () => {
+    const setUrlRadioGroupVisibility = () => {
       const titleHasChanged = originalTitle !== titleInput.value
       if (titleHasChanged) {
-        checkboxContainer.removeAttribute('hidden')
+        urlRadioGroup.removeAttribute('hidden')
       } else {
-        checkboxContainer.setAttribute('hidden', 'hidden')
+        urlRadioGroup.setAttribute('hidden', 'hidden')
       }
     }
 
-    titleInput.addEventListener('input', () => setCheckboxVisibility())
-    setCheckboxVisibility()
+    titleInput.addEventListener('input', () => setUrlRadioGroupVisibility())
+    setUrlRadioGroupVisibility()
   }
 
   Modules.EditionForm = EditionForm
--- a/app/models/concerns/edition/identifiable.rb
+++ b/app/models/concerns/edition/identifiable.rb
@@ -7,6 +7,7 @@ module Edition::Identifiable
     before_validation :ensure_presence_of_document, on: :create
     before_validation :propagate_type_to_document
     before_save :set_slug_from_title, if: -> { title_changed? }
+    before_save :nullify_redundant_slug_override, if: -> { slug_override.present? }
     before_save :set_slug, if: -> { slug_from_title_changed? || slug_override_changed? }
 
     scope :latest_edition, -> { joins(:document).where("editions.id = documents.latest_edition_id") }
@@ -62,6 +63,10 @@ module Edition::Identifiable
 
   def set_slug
     self[:slug] = slug_override.presence || slug_from_title
+  end
+
+  def nullify_redundant_slug_override
+    self.slug_override = nil if slug_override == slug_from_title
   end
 
   def linkable?
--- a/app/views/admin/editions/_page_address_controls.html.erb
+++ b/app/views/admin/editions/_page_address_controls.html.erb
@@ -13,18 +13,25 @@
     </p>
   </div>
 
-  <div class="<%= (edition.document.live_edition.slug == edition.slug_from_title) && "js-keep-slug-form-group" %> govuk-!-margin-bottom-6">
-    <%= render "govuk_publishing_components/components/hint", {
-      text: "The title of this document has changed since the page URL was generated. Check the box below to keep the existing URL, or leave it unchecked to update the URL for this page to match the new title when you publish this edition.",
-    } %>
-    <%= hidden_field(:edition, :slug_override, value: "") %>
-    <%= render "govuk_publishing_components/components/checkboxes", {
+  <% live_slug = edition.document.live_edition.slug %>
+  <% keep_current_url_checked = edition.slug_override.present? || edition.slug == live_slug %>
+  <div class="<%= (live_slug == edition.slug_from_title) && "js-url-radio-group" %> govuk-!-margin-bottom-6">
+    <p class="govuk-body">
+      The title has changed since this URL was created. Choose whether to keep the current URL or update it to match the title.
+    </p>
+
+    <%= render "govuk_publishing_components/components/radio", {
       name: "edition[slug_override]",
       items: [
         {
-          label: "Keep the current page URL",
-          value: edition.document.live_edition.slug,
-          checked: edition.slug_override.present?,
+          value: live_slug,
+          text: "Keep current URL",
+          checked: keep_current_url_checked,
+        },
+        {
+          value: "",
+          text: "Update URL to match title",
+          checked: !keep_current_url_checked,
         },
       ],
     } %>
--- a/app/views/admin/editions/_page_address_controls.html.erb
+++ b/app/views/admin/editions/_page_address_controls.html.erb
@@ -1,15 +1,16 @@
 <% if edition.document&.live_edition.present? %>
   <div class="app-view-edit-edition__page-address govuk-!-margin-bottom-4">
     <%= render "govuk_publishing_components/components/heading", {
-      text: "Live page address",
+      text: "URL",
       heading_level: 3,
       font_size: "m",
       margin_bottom: 2,
     } %>
 
-    <%= render "govuk_publishing_components/components/hint", {
-      text: edition.document&.live_edition.public_url,
-    } %>
+    <p class="govuk-body">
+      Current GOV.UK URL:<br>
+      <%= link_to edition.document.live_edition.public_url, edition.document.live_edition.public_url, class: "govuk-link" %>
+    </p>
   </div>
 
   <div class="<%= (edition.document.live_edition.slug == edition.slug_from_title) && "js-keep-slug-form-group" %> govuk-!-margin-bottom-6">
--- a/features/edition-update-slug.feature
+++ b/features/edition-update-slug.feature
@@ -1,18 +1,33 @@
 Feature: Edition update slug
-  Scenario: Writer updates the title of a document
-    Given I am a writer
-    And a published publication "You will never guess" exists
-    When I edit the publication "You will never guess"
-    And I change the title to "Ten facts that will shock you"
-    And I save the edition and go to the document summary
-    Then I can see the preview URL of the publication "Ten facts that will shock you" contains "ten-facts-that-will-shock-you"
 
-  @javascript
-  Scenario: Writer updates the title of a document and opts out of slug update
-    Given I am a writer
-    And a published publication "You will never guess" exists
-    When I edit the publication "You will never guess"
-    And I change the title to "Ten facts that will shock you"
-    And I opt out of updating the slug
-    And I save the edition and go to the document summary
-    Then I can see the preview URL of the publication "Ten facts that will shock you" contains "you-will-never-guess"
+  Rule: The URL control only appears when the title changes
+
+    @javascript
+    Scenario: The URL control is revealed when the title changes
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      When I edit the publication "You will never guess"
+      Then I cannot see the option to keep the current page URL
+      When I change the title to "Ten facts that will shock you"
+      Then I can see the option to keep the current page URL
+
+  Rule: Renaming a draft keeps the live URL unless an editor opts out
+
+    Scenario: Renaming a draft keeps the live URL by default
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      When I edit the publication "You will never guess"
+      And I change the title to "Ten facts that will shock you"
+      And I save the edition and go to the document summary
+      Then I can see the preview URL of the publication "Ten facts that will shock you" contains "you-will-never-guess"
+      And the saved URL choice for "Ten facts that will shock you" is reflected when revisiting the edit page
+
+    Scenario: Opting to use the title based slug updates the URL to match the title
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      When I edit the publication "You will never guess"
+      And I change the title to "Ten facts that will shock you"
+      And I opt out of keeping the live slug
+      And I save the edition and go to the document summary
+      Then I can see the preview URL of the publication "Ten facts that will shock you" contains "ten-facts-that-will-shock-you"
+      And the saved URL choice for "Ten facts that will shock you" is reflected when revisiting the edit page
--- a/features/step_definitions/edition_update_slug_steps.rb
+++ b/features/step_definitions/edition_update_slug_steps.rb
@@ -1,20 +1,19 @@
-And(/^I visit the edit slug page for "([^"]*)"$/) do |edition_title|
-  @edition = Edition.find_by!(title: edition_title)
-  visit edit_slug_admin_edition_path(@edition)
-end
-
-And(/^I update the slug to "([^"]*)"$/) do |new_slug|
-  fill_in "Slug", with: new_slug
-  click_button "Update"
-end
-
-Then(/^I can see the edition's public URL contains "([^"]*)"$/) do |new_slug|
-  visit admin_edition_path(@edition)
-  click_button "Create new edition"
-  expect(page.find(".app-view-edit-edition__page-address .govuk-hint")).to have_content new_slug
-end
-
 Then(/^I can see the preview URL of the publication "([^"]*)" contains "([^"]*)"$/) do |title, new_slug|
-  visit admin_edition_path(Publication.find_by(title:))
+  visit admin_edition_path(Publication.latest_edition.find_by!(title:))
   expect(page).to have_link "Preview on website (opens in new tab)", href: "https://draft-origin.test.gov.uk/government/publications/#{new_slug}"
+end
+
+Then(/^I cannot see the option to keep the current page URL$/) do
+  expect(page).not_to have_field("Keep current URL")
+end
+
+Then(/^I can see the option to keep the current page URL$/) do
+  expect(page).to have_css(".js-url-radio-group:not([hidden])")
+end
+
+Then(/^the saved URL choice for "([^"]*)" is reflected when revisiting the edit page$/) do |title|
+  draft = Publication.latest_edition.find_by!(title:)
+  visit edit_admin_edition_path(draft)
+  expected_choice = draft.slug_override.present? ? "Keep current URL" : "Update URL to match title"
+  expect(page).to have_checked_field(expected_choice)
 end
--- a/test/functional/admin/standard_editions_controller_test.rb
+++ b/test/functional/admin/standard_editions_controller_test.rb
@@ -203,6 +203,121 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
     end
   end
 
+  view_test "GET new does not show the URL controls" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    get :new, params: { configurable_document_type: "test_type" }
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit does not show the URL controls for a draft with no live edition" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    edition = create(:draft_standard_edition, configurable_document_type: "test_type")
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit pre-selects the 'Keep current URL' option when slug_override is set" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, configurable_document_type: "test_type", title: "Original title")
+    edition = create(:draft_standard_edition, configurable_document_type: "test_type", document: published_edition.document, title: "Renamed title", slug_override: published_edition.slug)
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the 'Keep current URL' option when slug_override is nil and the draft slug matches the live slug" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, configurable_document_type: "test_type", title: "Shared title")
+    edition = create(:draft_standard_edition, configurable_document_type: "test_type", document: published_edition.document, title: "Shared title")
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the 'Keep current URL' option when slug_override is blank and the draft slug matches the live slug" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, configurable_document_type: "test_type", title: "Shared title")
+    edition = create(:draft_standard_edition, configurable_document_type: "test_type", document: published_edition.document, title: "Shared title")
+    edition.update!(slug_override: "")
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the 'Update URL to match title' option when the draft slug differs from the live slug and slug_override is nil" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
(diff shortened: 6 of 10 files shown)


app/models/concerns/edition/identifiable.rb
```diff
@@ -6,6 +6,8 @@ module Edition::Identifiable
     validates :document, presence: true
     before_validation :ensure_presence_of_document, on: :create
     before_validation :propagate_type_to_document
+    before_save :default_slug_override_to_live_edition_slug,
```
lauraghiorghisor-tw: I take it trying to conditionally set the `slug_override` was too circular? 🤔 I suppose there's no other way but to always set it?

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,22 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal [edition_with_slug], Edition.with_title_containing("https://www.gov.uk/government/news/klingons-rule?foo=bar#details")
   end
 
+  test "a new edition on a document with a live edition defaults slug_override to the live edition's slug" do
```
lauraghiorghisor-tw: Alternatively: "a new draft from published"

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,22 @@ class EditionTest < ActiveSupport::TestCase
...
+    published_edition = create(:published_edition, title: "Klingons rule")
+    draft = create(:draft_edition, document: published_edition.document, title: "Klingons returneth")
```
lauraghiorghisor-tw: I think we should be using `create_draft` here to mimic the real flow.

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,22 @@ class EditionTest < ActiveSupport::TestCase
...
+
+    assert_equal published_edition.slug, draft.slug_override
+    assert_equal published_edition.slug, draft.slug
+  end
+
+  test "a new edition does not overwrite an explicitly set slug_override" do
+    published_edition = create(:published_edition, title: "Klingons rule")
+    draft = create(:draft_edition, document: published_edition.document, slug_override: "chosen-slug", title: "Klingons returneth")
```
lauraghiorghisor-tw: As above, not sure the set up is exactly capturing the flow. Can we use `create_draft`?

Also, for this one, don't you want the override to originally be on the published edition?

app/models/concerns/edition/identifiable.rb
lauraghiorghisor-tw: Could we capture in the commit message that we are changing the behaviour of the `slug_override` column - from a "true override" to whatever this is doing now, which is, I guess, just storing the live slug value, which may or may not be different from the title based value.

This is to allow us to always have the "checked" radio button for keeping the live slug.

With this change, the roles of the database columns become even more muddled in terms of what they store, there are no queries you can run to validate anything, since it's now perfectly valid to have slug = slug_from_title = slug_override = a title based slug.

@ChrisBAshton - can you please sanity check if this is ok with you 🕵🏻‍♀️

app/views/admin/editions/_page_address_controls.html.erb
```diff
@@ -1,15 +1,16 @@
...
       heading_level: 3,
       font_size: "m",
       margin_bottom: 2,
     } %>
 
-    <%= render "govuk_publishing_components/components/hint", {
-      text: edition.document&.live_edition.public_url,
-    } %>
+    <p class="govuk-body">
+      Current GOV.UK URL:<br>
```
lauraghiorghisor-tw, line 11: I wanted to argue that "current" is still unclear, but seeing how it's used three times in this section, it'll have to do 😅

lauraghiorghisor-tw: Just to confirm, the decision is to only show the "Live page address" (now: "URL") where a live page is available. For first ever drafts we will not be showing anything?

lauraghiorghisor-tw: On this commit message "Show the current GOV.UK URL as a link on the edit page" -> would it make sense to explain why you opted to move test coverage from collections to standard edition?

app/assets/javascripts/admin/views/edition-form.js
```diff
@@ -160,25 +160,25 @@
   EditionForm.prototype.setupTitleInputEventListener = function () {
     const form = this.module
     const titleInput = form.querySelector('#edition_title')
-    const checkboxContainer = form.querySelector('.js-keep-slug-form-group')
+    const keepSlugControls = form.querySelector('.js-keep-slug-form-group')
```
lauraghiorghisor-tw: Not too bothered, but we _could_ call it by its name, like a radio buttons container or something.

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,74 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
     end
   end
 
+  view_test "GET new does not show the keep current URL option" do
```
lauraghiorghisor-tw: Can we add some quotes around the name of the option - 'Keep the current URL'?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,74 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    get :new, params: { configurable_document_type: "test_type" }
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit does not show the keep current URL option for a draft with no live edition" do
```
lauraghiorghisor-tw: Same with the 'Keep the current URL'.

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,74 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type")
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit pre-selects the keep current URL option when slug_override is set" do
```
lauraghiorghisor-tw: 'Keep the current URL'

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,74 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the update URL option when slug_override is blank" do
```
lauraghiorghisor-tw: 'Update URL to match title' option

lauraghiorghisor-tw: While we still maintain both blanks and nils for SO, can we have two tests? And then you can delete the blank tests if we change it to only use nil.

features/edition-update-slug.feature
```diff
@@ -1,18 +1,51 @@
...
-  @javascript
-  Scenario: Writer updates the title of a document and opts out of slug update
-    Given I am a writer
-    And a published publication "You will never guess" exists
-    When I edit the publication "You will never guess"
-    And I change the title to "Ten facts that will shock you"
-    And I opt out of updating the slug
-    And I save the edition and go to the document summary
-    Then I can see the preview URL of the publication "Ten facts that will shock you" contains "you-will-never-guess"
+  Rule: The URL control only appears when the URL can be changed
```
lauraghiorghisor-tw: Again with "The URL control" 😅 - it's shorter than "radio buttons", I'll give you that.

features/edition-update-slug.feature
```diff
@@ -1,18 +1,51 @@
...
+
+    @javascript
+    Scenario: The URL control is revealed when the title changes
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      And I edit the publication "You will never guess"
+      And I cannot see the option to keep the current page URL
```
lauraghiorghisor-tw: **Then** I cannot see the option to keep the current page URL?

features/edition-update-slug.feature
lauraghiorghisor-tw: I see how you mean it but maybe `...when the title changes`?

features/edition-update-slug.feature
```diff
@@ -1,18 +1,51 @@
...
+      Given I am a writer
+      And a published publication "You will never guess" exists
+      When I edit the publication "You will never guess"
+      And I change the title to "Ten facts that will shock you"
+      And I opt out of keeping the live slug
+      And I save the edition and go to the document summary
+      Then I can see the preview URL of the publication "Ten facts that will shock you" contains "ten-facts-that-will-shock-you"
+      And the saved URL choice is reflected on re-entering the edit page
+
+    Scenario: A force-published edition with the original URL still keeps it on the next rename
```
lauraghiorghisor-tw: `...edition with a custom URL still keeps it...`?

features/step_definitions/edition_update_slug_steps.rb
```diff
@@ -1,24 +1,36 @@
...
-  click_button "Create new edition"
-  expect(page.find(".app-view-edit-edition__page-address .govuk-hint")).to have_content new_slug
+Then(/^I can see the option to keep the current page URL$/) do
+  expect(page).to have_css(".js-keep-slug-form-group:not([hidden])")
 end
 
-Then(/^I can see the preview URL of the publication "([^"]*)" contains "([^"]*)"$/) do |title, new_slug|
-  visit admin_edition_path(Publication.find_by(title:))
-  expect(page).to have_link "Preview on website (opens in new tab)", href: "https://draft-origin.test.gov.uk/government/publications/#{new_slug}"
+Then(/^the saved URL choice is reflected on re-entering the edit page$/) do
```
lauraghiorghisor-tw: How about `revisiting`?

features/edition-update-slug.feature
```diff
@@ -1,18 +1,51 @@
...
+      And the saved URL choice is reflected on re-entering the edit page
+
+    Scenario: A force-published edition with a title-based URL still keeps it on the next rename
+      Given I am an editor
+      And the publication "You will never guess" has been renamed to "Ten facts that will shock you" with the URL updated to match the new title
+      When I edit the publication "Ten facts that will shock you"
+      And I change the title to "Remember this person from the 90's"
+      And I save the edition and go to the document summary
+      Then I can see the preview URL of the publication "Remember this person from the 90's" contains "ten-facts-that-will-shock-you"
+      And the saved URL choice is reflected on re-entering the edit page
```
lauraghiorghisor-tw: Not sure what to make of the last two tests to be honest. So they cover:
- Custom override edition -> redraft -> chose to use live slug
- Blank override edition -> redraft -> choose to use live slug

Good to have some cross-state coverage, but I wouldn't say the scenarios are exhaustive. I am not sure in which specific ways these two scenarios are more important than any other combination of states? Who's to say the other permutations don't have a secret bug? 🐞 E.g. going:
- From override to title based
- From title based - to a different title based again

features/step_definitions/edition_update_slug_steps.rb
```diff
@@ -1,24 +1,36 @@
...
+  draft = Publication.where(state: "draft").last
```
lauraghiorghisor-tw: Can this not be `Publication.latest_edition.find_by!(title:)`?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,91 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
     assert_select "legend", text: "Review date"
   end
 
+  view_test "GET edit shows the current GOV.UK URL as a link when the edition has a live edition" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, :with_organisations, configurable_document_type: "test_type")
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type", document: published_edition.document)
```
lauraghiorghisor-tw: Do you need `:with_organisations`?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,91 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit does not show the keep current URL option for a draft with no live edition" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type")
```
lauraghiorghisor-tw: Do you need `:with_organisations`?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,91 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit pre-selects the keep current URL option when slug_override is set" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, :with_organisations, configurable_document_type: "test_type")
```
lauraghiorghisor-tw: Do you need `:with_organisations`?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,91 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the update URL option when slug_override is blank" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, :with_organisations, configurable_document_type: "test_type")
```
lauraghiorghisor-tw: Do you need `:with_organisations`?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -186,6 +186,91 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+    assert_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+    refute_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+  end
+
+  view_test "GET edit shows the URL change explanation and option labels" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, :with_organisations, configurable_document_type: "test_type")
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type", document: published_edition.document)
```
lauraghiorghisor-tw: Do you need `:with_organisations`?

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,74 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit pre-selects the keep current URL option when slug_override is set" do
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    published_edition = create(:published_standard_edition, :with_organisations, configurable_document_type: "test_type")
+    edition = create(:draft_standard_edition, :with_organisations, configurable_document_type: "test_type", document: published_edition.document, slug_override: published_edition.slug)
```
lauraghiorghisor-tw: Same as for the model tests, does it make any difference if you just use the factory to create a draft as opposed to using the `create_draft`?

lauraghiorghisor-tw (commented):
LGTM 👏 🏁🏁🏁🏁🏁🏁🏁🏁🏁

Small bits and namings and commit message additions.

features/edition-update-slug.feature
TonyGDS: Sure, will change the `And I edit the publication "You will never guess"` to `When I edit the publication "You will never guess"`. I can then put a **Then** here.

features/edition-update-slug.feature
TonyGDS: Was debating whether to keep them at all due to lower level test coverage. I'll have a think about this.

app/models/concerns/edition/identifiable.rb
```diff
@@ -6,9 +6,8 @@ module Edition::Identifiable
     validates :document, presence: true
     before_validation :ensure_presence_of_document, on: :create
     before_validation :propagate_type_to_document
-    before_save :default_slug_override_to_live_edition_slug,
-                if: -> { new_record? && document&.live_edition_id.present? && slug_override.blank? }
     before_save :set_slug_from_title, if: -> { title_changed? }
+    before_save :nullify_redundant_slug_override, if: -> { slug_override.present? }
```
lauraghiorghisor-tw, line 10: Is this a must or is it only for some specific flows?

app/views/admin/editions/_page_address_controls.html.erb
```diff
@@ -13,7 +13,9 @@
     </p>
   </div>
 
-  <div class="<%= (edition.document.live_edition.slug == edition.slug_from_title) && "js-keep-slug-form-group" %> govuk-!-margin-bottom-6">
+  <% live_slug = edition.document.live_edition.slug %>
+  <% keep_checked = edition.slug_override.present? || edition.slug == live_slug %>
```
lauraghiorghisor-tw: I like this!

test/unit/app/models/edition_test.rb
```diff
@@ -554,6 +554,40 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal "chosen-slug", draft.slug
   end
 
+  test "slug_override is cleared on create when it already matches slug_from_title" do
```
lauraghiorghisor-tw: This is what I mean, feels like another sanity setting to ensure that override is removed when users eventually comes back to the same title - but feels like a separate thing, new commit?

lauraghiorghisor-tw: Also for first ever drafts you'd have no way to set an override, so let's set this test up with a published edition as well?

test/unit/app/models/edition_test.rb
```diff
@@ -554,6 +554,40 @@ class EditionTest < ActiveSupport::TestCase
...
+    draft = create(:draft_edition, slug_override: "matching-title", title: "Matching title")
+
+    assert_nil draft.slug_override
+    assert_equal "matching-title", draft.slug
+  end
+
+  test "slug_override is kept on create when it differs from slug_from_title" do
```
lauraghiorghisor-tw: on redraft?

test/unit/app/models/edition_test.rb
```diff
@@ -554,6 +554,40 @@ class EditionTest < ActiveSupport::TestCase
...
+    published_edition = create(:published_edition, title: "Klingons rule")
+    draft = create(:draft_edition, document: published_edition.document, title: "Klingons returneth", slug_override: "klingons-rule")
+
+    draft.update!(title: "Klingons rule")
+
+    assert_nil draft.slug_override
+    assert_equal "klingons-rule", draft.slug
+  end
+
+  test "slug_override is cleared on update when posted equal to the current slug_from_title" do
```
lauraghiorghisor-tw: Is this possible on first ever drafts? Feels redundant.

test/functional/admin/standard_editions_controller_test.rb
lauraghiorghisor-tw: Let's also add a scenario for blank override here just to be exhaustive.

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -229,8 +229,23 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+
+    get :edit, params: { id: edition }
+
+    assert_response :ok
+    assert_select "input[type=radio][name=?]", "edition[slug_override]", count: 2
+    assert_select "input[type=radio][name=?][value=?][checked]", "edition[slug_override]", published_edition.slug
+    refute_select "input[type=radio][name=?][value=''][checked]", "edition[slug_override]"
+  end
+
+  view_test "GET edit pre-selects the keep current URL option when the draft slug already matches the live slug" do
```
lauraghiorghisor-tw: and the "and there is no override" in this case too

test/functional/admin/standard_editions_controller_test.rb
TonyGDS: no

test/functional/admin/standard_editions_controller_test.rb
TonyGDS: no

test/functional/admin/standard_editions_controller_test.rb
TonyGDS: no

test/functional/admin/standard_editions_controller_test.rb
TonyGDS: no

app/models/concerns/edition/identifiable.rb
TonyGDS: if the user changes the title back to the title from slug, it would be the same in the override. Moving to a separate commit

test/functional/admin/standard_editions_controller_test.rb
TonyGDS: `create_draft` requires a change note. In this test we aren't testing the model directly just the view logic

app/models/concerns/edition/identifiable.rb
ChrisBAshton: Yes this is ok with me in the short term. I do think once this is all merged and the dust has settled, we should look at getting rid of one or two of the slug columns 😅 

I understand why they were necessary with the previous design - we had to persist rather than make inferences about whether or not someone had checked a box and what the state of the document was at the time. But we did discuss last week the idea of bringing back the 'edit slug' screen - which, AFAICR, was only abandoned because of complications arising from slug being on Document rather than Edition. Now that the slug scoped to each Edition, it should be pretty straightforward (he says!) to show "this is your slug, this is your title - feel free to manually edit either one".

We could have a "generate new slug from title" button that takes the grunt work out of manually changing the slug value to match whatever you've set your title to. And we could introduce some sort of Ajax lookup to account for `--2` variances etc (or just accept that the proposed slug may be a little inaccurate until the edition is Saved, at which point the 'true' new slug is set).

Not one to tackle any time soon, but that's where I see this going ultimately - in which case, it doesn't matter too much what the meaning behind `slug_override` is at this point.

lauraghiorghisor-tw: Should have updated this sooner - Tony found a neat way to handle this. So we still have good data and meaningful columns moving forward. 

It is annoying to have the three columns but I think it's one of those scenarios where you have to go through the weird phase before the clarity of a new way becomes visible 😅

lauraghiorghisor-tw: I am just typing out some ideas from the slug confluence and your comments - and it seems that the suggestion there is also allowing more control over the slug in the future - including overriding the override. But as you say, "later" bucket.

test/unit/app/models/edition_test.rb
TonyGDS: will drop this test as its covered by an actual case in the next one

test/unit/app/models/edition_test.rb
TonyGDS: no it isn't, this test is actually covered by "cleared on update when a title change makes it redundant"

app/models/concerns/edition/identifiable.rb
TonyGDS: Have moved this logic to the front end so we don't set the override up front

app/views/admin/editions/_page_address_controls.html.erb
TonyGDS: Yes — the whole `_page_address_controls` partial is wrapped in `if edition.document&.live_edition.present?`, so first-ever drafts get nothing. There's no live URL to keep, and the title-derived slug isn't shown anywhere on draft because it can still drift before publishing.

test/functional/admin/standard_editions_controller_test.rb
TonyGDS: no

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,121 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
     end
   end
 
+  view_test "GET new does not show the 'Keep current URL' option" do
```
lauraghiorghisor-tw: The expectation here is that the radios (neither) are not shown at all, right?

So maybe `GET new does not show the URL controls`? I am totally brainwashed with the url controls now...
The expectation can probs stay the same.

test/functional/admin/standard_editions_controller_test.rb
```diff
@@ -203,6 +203,121 @@ class Admin::StandardEditionsControllerTest < ActionController::TestCase
...
+    configurable_document_type = build_configurable_document_type("test_type")
+    ConfigurableDocumentType.setup_test_types(configurable_document_type)
+
+    get :new, params: { configurable_document_type: "test_type" }
+
+    assert_response :ok
+    refute_select "label", text: /Keep current URL/
+  end
+
+  view_test "GET edit does not show the 'Keep current URL' option for a draft with no live edition" do
```
lauraghiorghisor-tw: Same - no radio group shown?

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,24 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal [edition_with_slug], Edition.with_title_containing("https://www.gov.uk/government/news/klingons-rule?foo=bar#details")
   end
 
+  test "a new draft from published leaves slug_override nil" do
```
lauraghiorghisor-tw: This now feels like it is covering smth that does not exist? Can delete?

test/unit/app/models/edition_test.rb
```diff
@@ -538,6 +538,24 @@ class EditionTest < ActiveSupport::TestCase
...
+    published_edition = create(:published_edition, title: "Klingons rule")
+    draft = published_edition.create_draft(create(:writer))
+    draft.update!(title: "Klingons returneth", change_note: "Renamed")
+
+    assert_nil draft.slug_override
+    assert_equal "klingons-returneth", draft.slug
+  end
+
+  test "a slug_override on the published edition is preserved when redrafting and renaming" do
```
lauraghiorghisor-tw: This test exists in identifiable test L241 - can delete

test/unit/app/models/edition_test.rb
```diff
@@ -556,6 +556,24 @@ class EditionTest < ActiveSupport::TestCase
     assert_equal "chosen-slug", draft.slug
   end
 
+  test "slug_override is cleared on redraft when the new title makes it redundant" do
```
lauraghiorghisor-tw: Move these two tests in the sluggy section of the identifiable test file

TonyGDS (approved):
LGTM & @lauraghiorghisor-tw 🚢

PR #11448, opened 2026-05-14
Whit 2073 access limit documents to named editors

jamiestamp:
## Summary

Adds support for restricting access to an edition by specific publisher emails and replaces the `access_limited` boolean with an enum:

- `not_limited` (0)  
- `organisations` (1)  
- `named_users` (2)  

Editors can now choose **“Limit access to named publishers”** and provide a list of email addresses. 

---

## Feature flag

The named-user access UI is gated behind the `access_limited_named_users` Flipflop feature flag (`features.rb`).

This allows a safe, incremental release without exposing the feature immediately to all users.

---

https://github.com/user-attachments/assets/2da50119-c9c3-48c3-8872-14a961756701

### DIFF
--- a/app/models/concerns/edition/limited_access.rb
+++ b/app/models/concerns/edition/limited_access.rb
@@ -2,6 +2,8 @@ module Edition::LimitedAccess
   extend ActiveSupport::Concern
 
   included do
+    enum :access_limited, { disabled: 0, organisations: 1, named_users: 2 }
+    has_many :edition_user_accesses, dependent: :destroy, inverse_of: :edition
     after_initialize :set_access_limited
   end
 
@@ -16,15 +18,22 @@ module Edition::LimitedAccess
   end
 
   def access_limited?
-    self[:access_limited]
+    organisations? || named_users?
   end
 
   delegate :access_limited_by_default?, to: :class
 
+  def access_limited=(value)
+    @_access_limited_explicitly_set = true
+    super
+  end
+
   def set_access_limited
-    if new_record? && access_limited.nil?
-      self.access_limited = access_limited_by_default?
-    end
+    return unless new_record?
+    return if @_access_limited_explicitly_set
+
+    self.access_limited = access_limited_by_default? ? :organisations : :disabled
+    @_access_limited_explicitly_set = false
   end
 
   def accessible_to?(user)
--- a/app/models/edition.rb
+++ b/app/models/edition.rb
@@ -472,6 +472,16 @@ class Edition < ApplicationRecord
     end
   end
 
+  def access_limited_named_users=(users)
+    return unless Flipflop.enabled?(:access_limited_named_users)
+
+    user_emails = users.split(",").map(&:strip).reject(&:empty?)
+
+    user_emails.each do |email|
+      edition_user_accesses.create!(email: email)
+    end
+  end
+
 private
 
   def date_for_government
--- a/app/models/edition_user_access.rb
+++ b/app/models/edition_user_access.rb
@@ -1,0 +1,16 @@
+class EditionUserAccess < ApplicationRecord
+  belongs_to :edition, inverse_of: :edition_user_accesses
+
+  validates :email,
+            presence: true,
+            uniqueness: { scope: :edition_id, case_sensitive: false },
+            format: { with: URI::MailTo::EMAIL_REGEXP }
+
+  before_destroy :prevent_destroy_if_locked
+
+private
+
+  def prevent_destroy_if_locked
+    throw(:abort) if locked?
+  end
+end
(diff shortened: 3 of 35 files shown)


app/models/edition_user_access.rb
```diff
@@ -0,0 +1,16 @@
...
+            uniqueness: { scope: :edition_id, case_sensitive: false },
+            format: { with: URI::MailTo::EMAIL_REGEXP }
+
+  before_destroy :prevent_destroy_if_locked
+
+private
+
+  def prevent_destroy_if_locked
+    throw(:abort) if locked?
+  end
```
ChrisBAshton, line 15: This is unusual for Whitehall - looks LLM-generated. Can you explain why it's needed (and write a corresponding test for it) or else drop it?

app/models/edition_user_access.rb
```diff
@@ -0,0 +1,16 @@
+class EditionUserAccess < ApplicationRecord
+  belongs_to :edition, inverse_of: :edition_user_accesses
+
+  validates :email,
+            presence: true,
+            uniqueness: { scope: :edition_id, case_sensitive: false },
+            format: { with: URI::MailTo::EMAIL_REGEXP }
```
ChrisBAshton, line 7: We ought to have a corresponding unit test file which exercises this validation.

db/schema.rb
```diff
@@ -347,7 +357,7 @@
   end
 
   create_table "editions", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.boolean "access_limited", null: false
+    t.integer "access_limited", default: 0, null: false
```
ChrisBAshton, line 360: Can you explain in the commit message what enum values you're expecting to use?

Or actually, does it even need to be an integer, or can we go with a string? Doing int-to-string lookups is a pain, for the sake of a few extra bytes.

Rails supports string enum mappings, so it should lead to very little change in your second commit 🙏 

```
enum :access_limited, {
  disabled: "disabled",
  organisations: "organisations",
  named_users: "named_users"
}
```

app/models/edition_user_access.rb
```diff
@@ -0,0 +1,16 @@
+class EditionUserAccess < ApplicationRecord
```
ChrisBAshton, line 1: Not a huge fan of the name - could we go with `AccessLimitingUsers`?
Scales nicely for WHIT-3505 where we want to introduce `AccessLimitingOrganisations`.

app/models/edition.rb
```diff
@@ -472,6 +472,16 @@ def first_published_within_current_govt
     end
   end
 
+  def access_limited_named_users=(users)
```
ChrisBAshton, line 475: This (and associated tests) should be in its own commit I think.
Changes to `set_access_limited` should follow.

app/models/concerns/edition/limited_access.rb
```diff
@@ -16,15 +18,22 @@ def access_limited_object
...
+    @_access_limited_explicitly_set = true
+    super
+  end
+
   def set_access_limited
-    if new_record? && access_limited.nil?
-      self.access_limited = access_limited_by_default?
-    end
+    return unless new_record?
+    return if @_access_limited_explicitly_set
```
ChrisBAshton, line 33: Can you put these changes in their own commit (see previous comment) explaining why this transient class variable is needed?

features/step_definitions/admin_limit_access_steps.rb
```diff
@@ -30,8 +30,8 @@
   select @org.name, from: "edition_lead_organisation_ids_1"
 end
 
-When(/^I check the "(.+)" box$/) do |box_name|
-  check box_name
+When(/^I pick the "(.+)" radio button$/) do |radio_label|
```
ChrisBAshton, line 33: Merge this into previous commit

db/migrate/20260506100009_migrate_access_limited_to_integer_enum.rb
```diff
@@ -0,0 +1,13 @@
+class MigrateAccessLimitedToIntegerEnum < ActiveRecord::Migration[8.0]
+  def up
+    safety_assured do
+      change_column :editions, :access_limited, :integer, null: false, default: 0
```
ChrisBAshton, line 4: What happens to existing access limited documents in this case? Would they suddenly lose access limiting?
May need to do this via expand-and-contract approach instead, i.e. make a new column row and write to both.

ChrisBAshton (commented):
I love the new design 👏 this looks very close. One major blocker is how we handle the migration - see last comment below.

db/migrate/20260506100009_migrate_access_limited_to_integer_enum.rb
jamiestamp: Existing rows are safe — MySQL stores booleans as tinyint(1) (0 or 1), and change_column ... :integer preserves those values. That's why the enum is `disabled: 0, organisations: 1` so the current state of the table is correct.

That said, you're right that expand-and-contract along with string enum mappings is the cleaner approach, rather than memorising numbers to debug access in the Rails console.

jamiestamp:
Closing this PR in favour of rolling out the functionality as defined in the [release plan](https://gov-uk.atlassian.net/browse/WHIT-3468?focusedCommentId=193623)

PR #11451, opened 2026-05-14
Remove legacy case study code [WHIT-3250]

kashifatcha:
This commit removes legacy case study MVC:

- model
- views
- controllers
- factories
- presenters

It also removes lead image component related code.


config/routes.rb
ChrisBAshton: This needs removing too.

ChrisBAshton: (And all references to `update_image_display_option`)!

test/components/admin/edition_images/image_component_test.rb
```diff
@@ -26,29 +25,10 @@ class Admin::EditionImages::ImageComponentTest < ViewComponent::TestCase
...
-  test "does not render a button to update the image to the lead image if the image is an SVG for case studies" do
-    svg_image_data = build(:image_data, file: File.open(Rails.root.join("test/fixtures/images/test-svg.svg")))
-    image = build_stubbed(:image, caption: "caption", image_data: svg_image_data)
-    edition = build_stubbed(:draft_case_study, images: [image])
-    render_inline(Admin::EditionImages::ImageComponent.new(edition:, image:, image_usage: ImageUsage.new))
-
-    assert_selector ".govuk-button", text: "Select as lead image", count: 0
-  end
-
   test "does not render a button to update the image to the lead image if the image usage is not govspeak_embed" do
```
ChrisBAshton: This test can be dropped - CaseStudy was the only remaining format that still had a "Set this image as the Lead Image" button under the 'embeddable images' section itself, rather than setting the lead image via its own separate Lead Image Upload Workflow a la News Articles etc.

test/integration/shareable_preview_test.rb
```diff
@@ -4,30 +4,31 @@
 class ShareablePreviewIntegrationTest < ActionDispatch::IntegrationTest
   extend Minitest::Spec::DSL
   include TaxonomyHelper
+  include Admin::EditionRoutesHelper
```
ChrisBAshton, line 7: Nice

test/unit/lib/whitehall/publishing_api_test.rb
```diff
@@ -44,20 +44,6 @@ class Whitehall::PublishingApiTest < ActiveSupport::TestCase
     assert_all_requested(requests)
   end
 
-  test ".publish sends case studies to the content store" do
```
ChrisBAshton, line 47: This test repeats the test on line 20. I don't know why we ever needed this test 🤔 agreed, let's kill it

ChrisBAshton (commented):
Looks good, Kash! Just a couple more things to delete (feel free to fix it up to the same commit), then we're good to go.

ChrisBAshton (approved):

PR #11469, opened 2026-05-18
Fix parent_document_url not being sent to asset manager for consultation outcome attachments [WHIT-3466]

eYinka:
When a consultation or call for evidence is published, `PublishAttachmentAssetJob` updates each attachment's metadata in asset manager. For attachments belonging to a `ConsultationOutcome`, `ConsultationPublicFeedback`, or `CallForEvidenceOutcome`, `parent_document_url` was never sent because the job guards with `respond_to?(:public_url)`, and neither `ConsultationResponse` nor `CallForEvidenceResponse` implemented that method.

Similarly, when we save a draft or update the metadata of an attachment, `AssetManagerAttachmentMetadataJob` calls `attachable_url` on `AttachmentData` to get the `parent_document_url`. This also returned `nil` for these types because `attachable_url` only handled Edition subclasses and Policy Group; neither `ConsultationResponse` nor `CallForEvidenceResponse` was covered.

We've made two fixes:

1. Extended `ConsultationResponse` and `CallForEvidenceResponse` to delegate `public_url` to their `parent_attachable`, so `PublishAttachmentAssetJob` can use their parent document's public URL for `parent_document_url`.

2. Extended `attachable_url` in `AttachmentData` to handle non-edition attachables that respond to `parent_attachable`, so `AssetManagerAttachmentMetadataJob` now correctly sets `parent_document_url` at every metadata update, including for draft consultations and calls for evidence.

Finally, I've added a data migration which enqueues `AssetManagerAttachmentMetadataJob` for all existing attachment data records belonging to `ConsultationResponse`, `CallForEvidenceResponse`, and `PolicyGroup`, so that historically missing `parent_document_url` values are corrected in asset manager.

Issue was found through a support ticket: https://gov-uk.atlassian.net/browse/WHIT-3466

## Testing

| Scenario | **Before** - WH `attachable_url` return value | **Before** - AM `parent_document_url` | **After** - WH `attachable_url` return value | **After** - AM `parent_document_url` | Comment |
| --- | --- | --- | --- | --- | --- |
| News article |   | | | | |
| Create draft | draft url | set to draft url | draft url | set to draft url | |
| Discard draft | nil | set to draft url | nil | set to draft url | The nil does not get sent downstream because the asset is likely deleted at this point, which is OK, but the code and tests can be misleading. |
| Publish | live url | set to live url | live url | set to live url |  |
| Unpublish | nil | "" | nil | Observed both in testing, sometimes it gets set to "" in AM, other times it remains live url. ??? | Note: asset remains draft: false, but with a redirect |
| Redraft from unpublished and save |  draft url | "" | draft url | "" | Should be setting the draft parent url, but it is not doing it. Possibly because the asset is not draft.  |
| Publish draft from unpublished | live url | set to live url | live url | set to live url | The extension does work, but the user would have to cachebust. |
| Consultation  |   | | | | |
| Create draft | nil | "" | draft url | set to draft url | **New**, OK |
| Discard draft | nil | "" | nil | set to draft url | The nil does not get sent downstream because the asset is likely deleted at this point, which is OK, but the code and tests can be misleading |
| Publish | nil | "" | set to live url | live url | **New**, OK |
| Unpublish | nil | "" | nil | Observed both in testing, sometimes it gets set to "" in AM, other times it remains live url. ??? | Note: asset remains draft: false, but with a redirect |
| Redraft from unpublished and save |  nil | "" | draft url | "" | Should set it to draft in AM |
| Publish draft from unpublished | nil | "" | live url | set to live url |  |

Summary:
- Although the `attachable_url` technically returns a nil parent url when discarding a draft,  the only job that runs in that case is `DraftAttachmentAssetDiscarder`, so the update is not sent downstream. That's generally OK, we still want to persist the draft url in this case. It's just a bit confusing when reading the code in the `attachment_data` model and the tests, as it suggests that it would override the parent url with nil.
- When unpublishing, the parent url is sometimes set to nil, sometimes not. Unsure why this behaviour is the case. Did not notice this discrepancy when testing the "before".
- When redrafting from unpublished, we have noticed some buggy behaviour (surely already noted somewhere in the backlog):
   - The assets on a redrafted edition from unpublished remain "live" i.e. draft: false, with a redirect. This makes sense as the live urls need to keep redirecting. But being able to "preview" the assets is impossible. This is likely a limitation of how we can represent things in assets manager. It could be that we could change the preview link from within wh to go to the draft stack, and allow asset manager to serve a draft view rather than redirect in that case.
   - The assets on a redrafted edition also don't get draft parent URLs - they keep whatever was set on the unpublishing step, either "" or live url. When the edition is eventually published, the url is set correctly.
- In some instances, caching can make the experience of using the browser extension quite confusing.

---


jamiestamp (approved):
LGTM

app/models/consultation_response.rb
```diff
@@ -33,6 +33,8 @@ def access_limited_object
 
   delegate :unpublishing, to: :parent_attachable
 
+  delegate :public_url, to: :parent_attachable
```
ChrisBAshton, line 36: Please roll this out to CallForEvidence too - near-identical document type.

ChrisBAshton (commented):
Approach looks good 👍 nice one getting to the bottom of it.

> In essence, parent_document_url is only reliably set at document publish time. Extending attachable_url to handle non-edition attachables with a public URL would be great as a follow-up work.

Interested in what that looks like. Care to write up a backlog item for it? 🙏

lauraghiorghisor-tw:
Had a quick chat with Ola - changed PR to draft and added a do not merge label to this as I think running the publish job would incorrectly publish draft assets. We'll be looking into using the Metadata updater.

eYinka:
> Had a quick chat with Ola - changed PR to draft and added a do not merge label to this as I think running the publish job would incorrectly publish draft assets. We'll be looking into using the Metadata updater.

Just to clarify, the rake task isn't on this PR yet. My original intention was to make a separate PR after merging this update. Nonetheless, let's see if we can do a global fix for both cases in this PR. My rough thought for the known limitation would be adding something like:

```
  elsif significant_attachable.respond_to?(:public_url)
    significant_attachable.public_url
  end
```

to https://github.com/alphagov/whitehall/blob/0f5896d030917c8fa6051bd5312fdc230010236a/app/models/attachment_data.rb#L176-L183

app/models/attachment_data.rb
```diff
@@ -173,16 +173,22 @@ def attachable_url
...
     elsif significant_attachable.respond_to?(:parent_attachable) && significant_attachable.parent_attachable.is_a?(Edition)
-      parent_attachable_in_pre_publication_state = Edition::PRE_PUBLICATION_STATES.include?(significant_attachable.parent_attachable.state)
-      parent_attachable_in_pre_publication_state ? significant_attachable.parent_attachable.public_url(draft: true) : significant_attachable.parent_attachable.public_url
+      url_for(significant_attachable.parent_attachable)
     elsif significant_attachable.is_a?(PolicyGroup)
       significant_attachable.public_url
     end
   end
 
+  def url_for(edition)
```
lauraghiorghisor-tw: I am certain this method exists elsewhere - more or less the same

eYinka: I've checked for this but didn't find any instance.

test/unit/app/sidekiq/publish_attachment_asset_job_test.rb
```diff
@@ -59,5 +59,68 @@ class PublishAttachmentAssetJobTest < ActiveSupport::TestCase
         job.perform(attachment_data.id)
       end
     end
+
+    context "attachment belongs to a consultation outcome" do
+      let(:consultation) { create(:published_consultation, title: "my-consultation") }
```
lauraghiorghisor-tw, line 64: I see this entire test file was written in the more verbose way but wondering if we can use the shorter set up syntax from a few commits after:

```
      let(:policy_group) { create(:policy_group, :with_file_attachment) }
      let(:attachment_data) { policy_group.attachments.first.attachment_data }
```

I think the `:with_file_attachment` gives us what we want here and you don't need a before save at all.

test/unit/app/models/attachment_data_test.rb
```diff
@@ -445,4 +445,158 @@ class AttachmentDataTest < ActiveSupport::TestCase
...
+      end
+
+      it "returns nil for deleted edition" do
+        edition = create(:draft_standard_edition, :with_file_attachment)
+        attachment_data = edition.attachments.first.attachment_data
+        assert attachment_data.attachments.first.attachable
+
+        edition.destroy!
+        edition.delete_all_attachments
+        # Tests would still pass without this line which is executed as part of the EditionDeleter, but wanted to make sure it doesn't have anything to do with the attachment being deleted.
```
lauraghiorghisor-tw: Delete comment

test/unit/app/models/attachment_data_test.rb
```diff
@@ -445,4 +445,158 @@ class AttachmentDataTest < ActiveSupport::TestCase
...
+
+      it "returns nil for deleted consultation" do
+        consultation = create(:consultation)
+        outcome = create(:consultation_outcome, :with_file_attachment, consultation:)
+        attachment_data = outcome.attachments.first.attachment_data
+        assert attachment_data.attachments.first.attachable
+
+        consultation.destroy!
+        consultation.delete_all_attachments
+        # Tests would still pass without this line which is executed as part of the EditionDeleter, but wanted to make sure it doesn't have anything to do with the attachment being deleted.
```
lauraghiorghisor-tw: Delete comment

test/unit/app/models/attachment_data_test.rb
```diff
@@ -445,4 +445,158 @@ class AttachmentDataTest < ActiveSupport::TestCase
...
+
+      it "returns nil for unpublished edition" do
+        edition = create(:unpublished_standard_edition, :with_file_attachment)
+        attachment_data = edition.attachments.first.attachment_data
+
+        assert_nil attachment_data.attachable_url
+      end
+    end
+
+    context "when the attachable is a consultation outcome" do
```
lauraghiorghisor-tw: This is a lot of coverage... but probs ok to just keep it so we don't get into these issues again. We can also use some test class with the `parent_attachable` subbed - which would match what we do in the code. But that's usually how we end up with edge cases 😅 Up to you.


Since you are using the parameterized approach anyway, you can also shorten this by doing:
```
    %w[consultation call_for_evidence].each do |parent_attachable_type|
      context "when the attachable is a #{parent_attachable_type} outcome" do
        Edition::PRE_PUBLICATION_STATES.each do |state|
          it "returns draft url for #{state} #{parent_attachable_type}" do
            edition = create(:"#{state}_#{parent_attachable_type}")
            outcome = create(:"#{parent_attachable_type}_outcome", :with_file_attachment, "#{parent_attachable_type}": edition)
            attachment_data = outcome.attachments.first.attachment_data

            assert_equal edition.public_url(draft: true), attachment_data.attachable_url
          end
        end

        Edition::PUBLICLY_VISIBLE_STATES.each do |state|
          it "returns live url for #{state} #{parent_attachable_type}" do
            edition = create(:"#{state}_#{parent_attachable_type}")
            outcome = create(:"#{parent_attachable_type}_outcome", :with_file_attachment, "#{parent_attachable_type}": edition)
            attachment_data = outcome.attachments.first.attachment_data

            assert_equal edition.public_url, attachment_data.attachable_url
          end
        end

        it "returns nil for deleted #{parent_attachable_type}" do
          edition = create(:"#{parent_attachable_type}")
          outcome = create(:"#{parent_attachable_type}_outcome", :with_file_attachment, "#{parent_attachable_type}": edition)
          attachment_data = outcome.attachments.first.attachment_data
          assert attachment_data.attachments.first.attachable

          edition.destroy!
          edition.delete_all_attachments

          assert_equal true, attachment_data.reload.attachments.first.deleted?
          assert_nil attachment_data.attachments.first.attachable
          assert_nil attachment_data.attachable_url
        end

        it "returns nil for unpublished #{parent_attachable_type}" do
          edition = create(:"unpublished_#{parent_attachable_type}")
          outcome = create(:"#{parent_attachable_type}_outcome", :with_file_attachment, "#{parent_attachable_type}": edition)
          attachment_data = outcome.attachments.first.attachment_data

          assert_nil attachment_data.attachable_url
        end
      end
    end
```

lib/tasks/attachments/backfill_parent_document_url_for_non_editionable_attachables.rake
```diff
@@ -0,0 +1,17 @@
+desc "Backfill parent_document_url in asset manager for non-editionable attachables (i.e ConsultationResponse, CallForEvidenceResponse, PolicyGroup)"
+task backfill_parent_document_url_for_non_editionable_attachables: :environment do
+  attachable_types = %w[ConsultationResponse CallForEvidenceResponse PolicyGroup]
+
+  attachment_data_ids = Attachment
```
lauraghiorghisor-tw: You can further filter out the CR and CFER whose parents are superseded and deleted themselves - because there is no need to touch those, and would be set to nil. I think unpublished too, since those should be already nil and you'd only set them to nil.

You can use the now public `parent_attachable` to tap into that state. If it's too costly maybe you can extract these out using a bit of sql instead.

lib/tasks/attachments/backfill_parent_document_url_for_non_editionable_attachables.rake
```diff
@@ -0,0 +1,17 @@
+desc "Backfill parent_document_url in asset manager for non-editionable attachables (i.e ConsultationResponse, CallForEvidenceResponse, PolicyGroup)"
```
lauraghiorghisor-tw: This can be a data migration -> goes into that special folder of data migrations rather than DB migrations, and you can run it manually from the terminal once your PR merges. 

I don't imagine it to be smth we need to run again, so not a rake task.

lib/tasks/attachments/backfill_parent_document_url_for_non_editionable_attachables.rake
```diff
@@ -0,0 +1,17 @@
...
+
+  attachment_data_ids = Attachment
+    .where(attachable_type: attachable_types)
+    .where.not(deleted: true)
+    .distinct
+    .pluck(:attachment_data_id)
+  puts "Enqueueing #{attachment_data_ids.count} attachment data records"
+
+  attachment_data_ids.each do |id|
+    AssetManagerAttachmentMetadataJob.perform_async(id)
```
lauraghiorghisor-tw: I think this should be safe enough to run now, and it hopefully contains enough of the logic that you don't need to worry about edge cases.

I would validate the following scenarios. Get some data prior to migration, then run migration, then check data:
- live response parent attachable
- draft response parent attachable
- deleted/unpublished response parent attachable
- policy group

Now that we changed that logic in the attachable_url, could you please also validate an e2e flow on regular edition attachables, as well as response type attachables and policy groups? Ensure that there are draft urls being set when creating and editing, and live urls when publishing, then draft again when you redraft from published. Basically between the code changes and the migration both your old data and your new data should have the correct URLs 🤩

lauraghiorghisor-tw (approved):
LGTM if testing shows no issues 👍 Few comments on tests

app/models/attachment_data.rb
```diff
@@ -103,10 +103,6 @@ def visible_to?(user)
     !deleted? && (!draft? || (draft? && accessible_to?(user)))
   end
 
-  def visible_attachment_for(user)
```
ChrisBAshton, line 106: Nice!

app/models/call_for_evidence_response.rb
```diff
@@ -33,6 +33,8 @@ def access_limited_object
 
   delegate :unpublishing, to: :parent_attachable
 
+  delegate :public_url, to: :parent_attachable
```
ChrisBAshton, line 36: Please update the first commit to make it clear that this bit:


> Extending `attachable_url` to handle non-edition attachables with a public URL would be great as a follow-up work.


...is now implemented a few commits later.

ChrisBAshton (approved):
Nice work 👏

app/models/attachment_data.rb
```diff
@@ -169,12 +169,23 @@ def redirect_url
...
+    if significant_attachable.is_a?(Edition)
+      url_for(significant_attachable)
+    elsif significant_attachable.respond_to?(:parent_attachable) && significant_attachable.parent_attachable.is_a?(Edition)
+      url_for(significant_attachable.parent_attachable)
+    elsif significant_attachable.is_a?(PolicyGroup)
+      significant_attachable.public_url
+    end
+  end
+
+  def url_for(edition)
```
lauraghiorghisor-tw: Can make this a private method now.

test/unit/app/models/attachment_data_test.rb
```diff
@@ -445,4 +445,114 @@ class AttachmentDataTest < ActiveSupport::TestCase
...
+
+        edition.destroy!
+        edition.delete_all_attachments
+
+        assert_equal true, attachment_data.reload.attachments.first.deleted?
+        assert_nil attachment_data.attachments.first.attachable
+        assert_nil attachment_data.attachable_url
+      end
+
+      it "returns nil for unpublished edition" do
```
lauraghiorghisor-tw, line 435: Can we add a little explanation in the commit message please - to cover why we chose to return nil for deleted and unpublished. For example:

> The new logic returns:
> -  A live url when the attachable (or the attachable parent) is in a visible state (published, withdrawn). This matches what the `visible_to?(user)` check accomplished before: only returned true when the attachable was `!deleted? && !draft?`, i.e. not deleted and visibly live. 
> - A draft url when the attachable (or the attachable parent) is in a draft state (pre publication states). This is the logic previously accomplished by the `draft_edition` call. 
> - Exits the if/else and returns `nil` for any other states. This matches the behaviour from before, where the `visible_to(user)` would have evaluated `!deleted (T) && (!draft? (F) && accessible_to(user) (F)` as F because the user is always nil, meaning `visible_edition.present?` was F too.

lauraghiorghisor-tw: Funnily enough when we unpublish an edition (on main) - the `parent_document_url` is still set on the asset as live. I actually think this is correct/ok, the url can still reference the parent. Same for discarding a draft, the parent url is still set to draft. But judging by the previous code, it should be making that url nil... It's possible that the update job does not run when we unpublish or delete. Can't remember.

I am fairly confident that we have preserved the logic for non live and non draft states though - if I run the edition context on main, all tests pass. So even with the new changes, the URL would still be set on the asset when you unpublish or delete, since we've kept the same logic. So if you're seeing that in your manual tests, it's fine.

The code is saying that the url would be nil-ed here, when it is not, which is not great, but out of scope for this PR. It would be good to capture these scenarios in a single place, maybe revisit an integration level test and cover exactly what updates we expect to see for draft/ live/ delete/ unpublish.

PR #11472, opened 2026-05-19
[WHIT-3424] Restore Welsh translations for corporate information

jamiestamp:
A bulk cleanup (3b7e550a (https://github.com/alphagov/whitehall/commit/3b7e550a22)) deleted cy.yml and 64 other locale files on the basis that Whitehall's UI is hardcoded English and the files were considered unused. This assumption was incorrect and we have restored most of the Welsh back. These were missed from earlier restorations:

  - about_our_services_html
  - personal_information_charter_html
  - publication_scheme_html
  - social_media_use_html
  - welsh_language_scheme_html

This PR brings those back to fix missing translations in the corporate information section.

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -37,3 +37,10 @@ cy:
       jobs: Swyddi
       jobs_and_contacts: Swyddi a chontractau
       organisation_chart: Ein siart sefydliadol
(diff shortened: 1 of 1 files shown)


TonyGDS (approved):
LGTM

PR #11474, opened 2026-05-19
Display only publications in `in_pre_publication_state` in `EditionFilter` of `StatisticsAnnouncementPublicationController` [WHIT-3132]

patrickpatrickpatrick:
## What

- Update filter of `StatisticsAnnouncementPublicationController`

## Why

Only present to the user unpublished editions in the results page for filtering documents to connect to a `StatisticsAnnouncement`. This prevents a user connecting a published edition which had unintended consequences.

### DIFF
--- a/app/controllers/admin/statistics_announcement_publications_controller.rb
+++ b/app/controllers/admin/statistics_announcement_publications_controller.rb
@@ -25,7 +25,7 @@ private
   end
 
   def edition_scope
-    Edition.with_translations(I18n.locale)
+    Edition.with_translations(I18n.locale).in_pre_publication_state
   end
 
   def params_filters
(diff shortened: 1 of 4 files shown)


app/controllers/admin/statistics_announcement_publications_controller.rb
ChrisBAshton: Could you not append `.in_pre_publication_state` here rather than add a new edition filter option? https://github.com/alphagov/whitehall/blob/8ffdfa177159b45fe00173f57f74ce4ed1df08cc/app/models/concerns/edition/workflow.rb#L85

davidsauntson (commented):
I could be confused, but I think might be the other way around from what is in the ticket.

https://gov-uk.atlassian.net/browse/WHIT-3132 says that you should _only_ be able to link the statistics announcements to a draft, since when that draft is published, the statistics announcement redirects to the published thing and not the other way around.

ChrisBAshton (approved):
LGTM

PR #11476, opened 2026-05-20
Raise validation error if badly access-limiting draft on-create [WHIT-3458]

ChrisBAshton:
Prior to this commit, it was possible to:

1. Start drafting a new document
2. Change the lead organisation to an org other than the one the user is in
3. Check the "Limit access" checkbox
4. Click "Save"

This would result in the publisher seeing a "You do not have permission to access this page" 403 response. They'd have to hit the 'back' button in the browser to return to their form and correct their inputs. In my local testing, getting into this state did not create the document, but Zendesk ticket 6597730 implies that a document _did_ get created when a user did this in the wild.

As a GDS Admin, the bug is even stranger, leading to a Server Error (https://govuk.sentry.io/issues/7494056061/?project=202259&query=is%3Aunresolved&referrer=issue-stream):
> No route matches {:action=>"edit", :controller=>"admin/edition_access_limited", :id=>nil}, missing required keys: [:id] (ActionView::Template::Error)

This was because of a [recent change to the forbidden.html.erb](https://github.com/alphagov/whitehall/commit/edb745f58baacefdfdb2d6fc71aa711d355c5ec7) template, whereby GDS Admins are linked through to the `edit_access_limited_admin_edition_path` of the Edition. That path requires an edition ID, but there is no ID if the Edition failed to persist, hence we saw a runtime error instead.

And fixing that error led to another one:

```
undefined method `id' for nil (NoMethodError) ./app/services/draft_edition_updater.rb:31:in `access_limit_excludes_current_user?'
```

^ This one was because by default, any user in Cucumber has not been configured to have an Organisation, so the `access_limit_excludes_current_user?` check would raise an exception when trying to query the user's organisations.

We know from [WHIT-3179](https://gov-uk.atlassian.net/browse/WHIT-3179) that it _is_ possible for some users to lack an Organisation, so we do need to account for that too rather than just patch up the fixture data.

The fix in this commit is to:

1. Only enforce the 'create' permission on Editions if the Edition is persisted
2. Ensure that creation failures are passed through to the view via the `flash.now[:alert]` mechanism
3. "Fail closed" on the access limiting check, for any user who is not part of an organisation
4. Only attempt to render the helper link for GDS Admins if the edition is persisted.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3458

## Screenshots

> Screenshot 2026-05-20 at 10 42 41

---

[WHIT-3179]: https://gov-uk.atlassian.net/browse/WHIT-3179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -32,7 +32,7 @@ class Admin::EditionsController < Admin::BaseController
     when "choose_type", "change_type", "change_type_preview", "apply_change_type"
       enforce_permission!(:create, edition_class || Edition)
     when "create"
-      enforce_permission!(:create, @edition)
+      enforce_permission!(:create, @edition) if @edition.persisted?
     when "edit", "update", "revise", "diff", "update_bypass_id", "features"
       enforce_permission!(:update, @edition)
     when "destroy", "confirm_destroy"
@@ -95,6 +95,7 @@ class Admin::EditionsController < Admin::BaseController
       updater.perform!
       redirect_to show_or_edit_path, saved_confirmation_notice
     else
+      flash.now[:alert] = updater.failure_reason
       build_edition_dependencies
       render :new
     end
--- a/app/services/draft_edition_updater.rb
+++ b/app/services/draft_edition_updater.rb
@@ -28,6 +28,9 @@ private
   end
 
   def access_limit_excludes_current_user?
+    # Some users may not necessarily belong to an organisation. We should fail-closed for them.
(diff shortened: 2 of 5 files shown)


GDSNewt (approved):
Looks good, I think if we do end up consolidating the form in one place we may wish to do some renaming/commenting in the tests. At the mo it's a bit confusing to think of where the user is in the flow.

PR #11456, opened 2026-05-15
Selectively allow unpublishing of historical editions [WHIT-3402]

davidsauntson:
Previously only GDS Editors and GDS Admins could unpublish historical content.  The DBT are about to unpublish and archive a lot of content, so an exception has been made to allow specific users outside of those two GDS roles to unpublish historic content.  See https://gov-uk.atlassian.net/browse/WHIT-3402

This change creates the relevant unpublishing permission and allows users with it to perform the unpublishing action for historic documents.

---

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

This application is owned by the Whitehall Experience team. Please let us know in [#govuk-whitehall-experience-tech](https://gds.slack.com/archives/C02L13S214K) when you raise any PRs.

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #11479, opened 2026-05-20
Allow unpublishing of historic content in `HistoricContentConcern` [WHIT-3402]

davidsauntson:
The `HistoricContentConcern` previously prevented any user without permission to `update` a historical edition from doing anything to it at all, since the `before_action` was called for every action.

This was correct (and mostly still is), but now we want some users to be able to unpublish historical content but not otherwise edit it.

This change (along with #11456) allows that by
- creating a new `before_action` restriction that checks the user can unpublish (rather than update) historical content
- applies the new restriction only for the unpublish and confirmation routes
- applies the existing editing restriction to the other, non-unpublishing related routes

---


ChrisBAshton (approved):
Great work ⭐

PR #11483, opened 2026-05-21
[WHIT-3425] Add Welsh translation for non-ministerial department pages

jamiestamp:
This PR lays the foundation on which to translate organisations and their relationships to Welsh. The text had been all hard-coded and built dynamically before. Initially we just have the single org and relationship translated.

- Added Welsh translation for "About us" section on non-ministerial departments' organisation pages.
- Refactored organisation parental relationship templates for better locale-specific grammar within YAML files.

[Jira](https://gov-uk.atlassian.net/browse/WHIT-3425)

### DIFF
--- a/app/helpers/organisation_helper.rb
+++ b/app/helpers/organisation_helper.rb
@@ -10,6 +10,17 @@ module OrganisationHelper
     /building law and hygiene/,
   ].freeze
 
+  SPONSORED_TYPE_KEYS = %i[
+    advisory_ndpb
+    executive_agency
+    executive_ndpb
+    special_health_authority
+  ].freeze
+
+  IDENTIFICATION_ONLY_TYPE_KEYS = %i[
+    non_ministerial_department
+  ].freeze
+
   def organisation_relationship_display_name(organisation)
     if organisation.acronym.present?
       tag.abbr(organisation.acronym, title: organisation.name)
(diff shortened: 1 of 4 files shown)


test/unit/app/helpers/organisation_helper_test.rb
```diff
@@ -177,4 +177,41 @@ def assert_display_name_text(organisation, expected_text)
...
+  test "non-ministerial department with parents renders Welsh identification sentence" do
+    parent = create(:ministerial_department, name: "Department of Testing")
+    org = create(:organisation, acronym: "CC", name: "Comisiwn Elusennau", organisation_type: OrganisationType.non_ministerial_department, parent_organisations: [parent])
+
+    I18n.with_locale(:cy) do
+      assert_display_name_text org, "Adran anweinidogol yw'r CC."
+    end
+  end
+
+  test "non-ministerial department with supporting bodies renders mixed Welsh/English sentence in Welsh" do
```
ChrisBAshton, line 199:
```suggestion
  # TODO: delete test once we have full Welsh translations
  test "non-ministerial department with supporting bodies renders mixed Welsh/English sentence in Welsh" do
```

test/unit/app/helpers/organisation_helper_test.rb
```diff
@@ -177,4 +177,41 @@ def assert_display_name_text(organisation, expected_text)
...
+    org = create(:organisation, acronym: "CC", name: "Comisiwn Elusennau", organisation_type: OrganisationType.non_ministerial_department)
+    org.stubs(:supporting_bodies).returns([child])
+
+    I18n.with_locale(:cy) do
+      description = organisation_display_name_including_parental_and_child_relationships(org)
+      assert_equal "Adran anweinidogol yw'r CC, supported by 1 public body.", strip_html_tags(description)
+    end
+  end
+
+  test "non-Welsh-translated type falls back to English in Welsh locale" do
```
ChrisBAshton, line 211:
```suggestion
  # TODO: delete test once we have full Welsh translations
  test "non-Welsh-translated type falls back to English in Welsh locale" do
```

app/helpers/organisation_helper.rb
```diff
@@ -33,26 +40,38 @@ def organisation_type_name(organisation)
...
                     end
                   else
-                    type_name != "other" ? "#{name} is #{relationship}." : name.to_s
+                    case type_key
+                    when :other
+                      template_args[:name]
+                    when :non_ministerial_department
+                      I18n.t("organisation.relationship.type_only_html", **template_args)
+                    else
+                      I18n.t("organisation.relationship.without_parents.default_html", **template_args)
```
ChrisBAshton: These evaluate to the same string, don't they? Why do they need distinct cases?

app/helpers/organisation_helper.rb
```diff
@@ -33,26 +40,38 @@ def organisation_type_name(organisation)
...
+                    end
```
ChrisBAshton: Generally struggling to parse this mega `description=` setter now that it has multiple case statements, references to i18n text, references to consts defined elsewhere in the file, etc.

Can it be made more readable by splitting into small, well-defined methods?

ChrisBAshton (commented):
LGTM in principle, Jamie 👍 can we refactor it a little before merge?

test/unit/app/helpers/organisation_helper_test.rb
jamiestamp: TODO added

test/unit/app/helpers/organisation_helper_test.rb
jamiestamp: TODO added

app/helpers/organisation_helper.rb
jamiestamp: Good eye! It was due to a difference with some of the machine generated Welsh sentence structure. As we don't need it until we do the wider translation work and have the translations locked down, I have removed the unneeded state.

app/helpers/organisation_helper.rb
jamiestamp: Yep, it was definitely on the gnarly side. I have done a bit of a refactor to hopefully make the logic a bit easier to read.

app/helpers/organisation_helper.rb
```diff
@@ -10,6 +10,17 @@ module OrganisationHelper
     /building law and hygiene/,
   ].freeze
 
+  SPONSORED_TYPE_KEYS = %i[
+    advisory_ndpb
+    executive_agency
+    executive_ndpb
+    special_health_authority
```
ChrisBAshton, line 17: Aren't we missing `tribunal` from this list?

app/helpers/organisation_helper.rb
```diff
@@ -33,29 +44,49 @@ def organisation_type_name(organisation)
...
-                    type_name != "other" ? "#{name} is #{relationship}." : name.to_s
-                  end
-
-    description.html_safe
+    type_key = organisation.organisation_type_key
+    parent_organisations = organisation.parent_organisations
+
+    if type_key == :other && parent_organisations.empty?
+      return ERB::Util.h(organisation_relationship_display_name(organisation)).strip.html_safe
+    end
```
ChrisBAshton: Feels a little odd that this case is called out separately when we also have line 65 dealing with parentless organisations. Is it just because we're only returning the org name, so no interpolation needed?

I'd be inclined to add a noop i18n row instead, and move this case to `relationship_i18n_key`:

```diff
    relationship:
+     none_html: "%{name}"
      identification_html: "%{name} is %{relationship}."
```

app/helpers/organisation_helper.rb
```diff
@@ -33,29 +44,49 @@ def organisation_type_name(organisation)
...
+
+  def relationship_i18n_key(type_key, parent_organisations)
+    return "organisation.relationship.identification_html" unless display_parent_relationships?(type_key, parent_organisations)
+
+    case type_key
+    when :other then "organisation.relationship.works_with_html"
+    when :sub_organisation then "organisation.relationship.part_of_html"
+    when *SPONSORED_TYPE_KEYS then "organisation.relationship.sponsored_html"
+    else "organisation.relationship.default_with_parents_html"
+    end
```
ChrisBAshton, line 69: Nice 👍

ChrisBAshton (approved):
One potential blocker on the tribunal thing, but otherwise good to go 👍

app/helpers/organisation_helper.rb
jamiestamp: That tribunal string match was dead code as of the 2019 removal of the `:tribunal_ndpb` org type. The string was that type's name and has matched nothing since. `:tribunal` falls through to the default as expected.

I have made an extra commit removing the dead code before the refactor so it is clear from the history.

app/helpers/organisation_helper.rb
jamiestamp: Good shout, I've implemented that now.

